PR #25683, opened 2025-04-28
Use UUIDv7 for OpenLineage runId

### LINKED ISSUE #25534: [Enhancement] Use UUIDv7 for QueryId
QueryId is currently [generated](https://github.com/trinodb/trino/blob/474/core/trino-main/src/main/java/io/trino/execution/QueryIdGenerator.java) as `{timestamp:yyyyMMdd_HHmmss}_{index:05}_{coordinator_id:05}`. Could it be changed to match [RFC 9562](https://datatracker.ietf.org/doc/rfc9562/) UUID versions 6-8?

Benefits are:
* UUIDv7 format implementations already uses timestamp + counter + random part for generation, so the value ordering and uniqueness should be just the same. More than that, UUIDv7 timestamp precision is milliseconds compairing to whole seconds of QueryId.
* UUID is 128 bit compairing to 216 bit of string in UTF-8 encoding.
* Better support by third-party systems. Imagine storing audit logs in some database, and user who wants to filter data by queryId field. UUID data type is quite common in modern DBMS, and index takes less space than same index for a string field.

dolfinus:
## Description

Previously OpenLineage run.id was generated as `UUIDv3(hash(queryId))`. Now it is generated as `UUIDv7(queryCreateTime, hash(queryId))`, using [UUIDUtils.generateStaticUUID](https://github.com/OpenLineage/OpenLineage/pull/3672) method.

According to [OpenLineage documentation](https://openlineage.io/docs/spec/object-model#run), it is recommended to use UUID v7 for run.id values.
This also makes values more compatible with B-Tree indexes and allows using internal timestamp as a partition column.

Fixes #25534

## Release notes

```markdown
## OpenLineage
* Use UUIDv7 for OpenLineage `runId` ({issue}`25534`)
```

### DIFF
--- a/docs/src/main/sphinx/admin/event-listeners-openlineage.md
+++ b/docs/src/main/sphinx/admin/event-listeners-openlineage.md
@@ -23,7 +23,7 @@ not limited to) Spark, Airflow, Flink.
     - Trino
     - OpenLineage
 *
-    - `{UUID(Query Id)}`
+    - `{UUIDv7(Query.createTime, hash(Query.Id))}`
     - Run ID
 *
     - `{queryCreatedEvent.getCreateTime()} or {queryCompletedEvent.getEndTime()} `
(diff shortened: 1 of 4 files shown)


mgorsk1:
lgtm 🙏🏻

dolfinus:
Could someone take a look? This is important change to make integration compatible with some OpenLineage backends

ebyhr (approved):

dolfinus:
Thank you very much!

PR #25684, opened 2025-04-28
Simplify and test segment serialization/deserialization

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/client/trino-client/src/main/java/io/trino/client/QueryDataJacksonModule.java
+++ b/client/trino-client/src/main/java/io/trino/client/QueryDataJacksonModule.java
@@ -123,8 +123,15 @@ public class QueryDataJacksonModule
         public void serialize(Segment value, JsonGenerator gen, SerializerProvider provider)
                 throws IOException
         {
+            serializeWithType(value, gen, provider, segmentSerializer(provider));
+        }
+
+        @Override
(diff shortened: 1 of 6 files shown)


client/trino-client/src/main/java/io/trino/client/QueryDataJacksonModule.java
```diff
@@ -127,6 +127,13 @@ public void serialize(Segment value, JsonGenerator gen, SerializerProvider provi
                     .serializeWithType(value, gen, provider, segmentSerializer(provider));
         }
 
+        @Override
+        public void serializeWithType(Segment value, JsonGenerator gen, SerializerProvider provider, TypeSerializer typeSer)
```
losipiuk, line 131: why is typeSer not used?

wendigo: It is used if you want to do a custom representation using JsonGenerator

wendigo: This will be an instance of `AsPropertyTypeSerializer`

wendigo: I've used the type explicitly

losipiuk (approved):

PR #25689, opened 2025-04-28
Remove unused InterpretedHashGenerator#hashPosition method

raunaqmorarka:
## Release notes

```markdown
## Section
```


core/trino-main/src/main/java/io/trino/operator/InterpretedHashGenerator.java
```diff
@@ -86,17 +85,6 @@ public long hashPosition(int position, Page page)
         return result;
     }
 
-    public long hashPosition(int position, IntFunction<Block> blockProvider)
```
pettyjamesm, line 89: The inline comment in the other method should be removed too since it refers to this method.

pettyjamesm (approved):

wendigo (approved):

PR #4313, opened 2020-07-02
Implement aggregate pushdown for druid.

Parth-Brahmbhatt:
Fixes https://github.com/prestosql/presto/issues/4109

PR #25706, opened 2025-04-29
[WIP] Add support for aggregation pushdown in Druid connector

conor-mcavoy:
This change adds support for aggregation pushdown in the Druid connector, for some common aggregate functions, namely min/max, avg, sum, count, and variations of these. Actually adding this support requires little more than boilerplate, as the standard "Implements*" AggregateFunctionRules can be used for most functions.

This change also required some edits to tests. For example, the generic BaseJdbcConnector tests assume connectors support table creation in some places, which is not the case for Druid.

## Release notes

```markdown
## Section
```


docs/src/main/sphinx/connector/druid.md
```diff
@@ -68,6 +68,21 @@ properties files.
...
+:::{list-table}
+:widths: 30, 58, 12
+:header-rows: 1
+
+* - Property name
+  - Description
+  - Default value
+* - ``druid.count-distinct-strategy``
+  - Defines Druid behavior for pushed-down ``DISTINCT`` aggregations by setting (or leaving unset) ``useApproximateCountDistinct``. Must be one of ``DEFAULT``, ``APPROXIMATE``, or ``EXACT``. When ``DEFAULT`` is set, the behavior is left up to the Druid cluster. ``APPROXIMATE`` sets ``useApproximateCountDistinct`` explicitly to ``true``, while ``EXACT`` sets it to ``false``. Note that Druid will reject pushed-down ``DISTINCT`` aggregations unless ``useApproximateCountDistinct=false``. This can either be done on the Trino side by setting this property to ``EXACT``, or it can be done on the Druid cluster. 
+  - ``DEFAULT``
```
Praveen2112, line 83: So if we set the this property to `DEFAULT` and if the Druid cluster has `DEFAULT` has `useApproximateCountDistinct` set to true does it uses approximate computation ?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
Praveen2112, line 1: Can we have a Test class for this config ?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
```diff
@@ -630,4 +666,21 @@ public RemoteIdentifiers getRemoteIdentifiers(Connection connection)
...
+    private static Optional<JdbcTypeHandle> toTypeHandle(DecimalType decimalType)
+    {
+        return Optional.of(new JdbcTypeHandle(Types.NUMERIC, Optional.of("decimal"), Optional.of(decimalType.getPrecision()), Optional.of(decimalType.getScale()), Optional.empty(), Optional.empty()));
+    }
+
+    @Override
+    public Optional<ParameterizedExpression> convertPredicate(ConnectorSession session, ConnectorExpression expression, Map<String, ColumnHandle> assignments)
+    {
+        return connectorExpressionRewriter.rewrite(session, expression, assignments);
+    }
```
Praveen2112, line 685: This PR focuses only on aggregation pushdown right ? So we could implement this as a follow up

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
```diff
@@ -585,4 +598,149 @@ public void testExecuteProcedureWithInvalidQuery()
...
+    @Override
+    public void testDistinctAggregationPushdown()
+    {
+        // SELECT DISTINCT
+        assertThat(query("SELECT DISTINCT regionkey FROM nation")).isFullyPushedDown();
+        assertThat(query("SELECT min(DISTINCT regionkey) FROM nation")).isFullyPushedDown();
+        assertThat(query("SELECT DISTINCT regionkey, min(nationkey) FROM nation GROUP BY regionkey")).isFullyPushedDown();
+
+        Session withMarkDistinct = Session.builder(getSession())
+                .setSystemProperty(DISTINCT_AGGREGATIONS_STRATEGY, "mark_distinct")
```
Praveen2112, line 676: Any reason for it to work on this configuration ?

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestingDruidServer.java
```diff
@@ -66,7 +66,7 @@ public class TestingDruidServer
     private static final int DRUID_HISTORICAL_PORT = 8083;
     private static final int DRUID_MIDDLE_MANAGER_PORT = 8091;
 
-    private static final String DRUID_DOCKER_IMAGE = "apache/druid:0.18.0";
+    private static final String DRUID_DOCKER_IMAGE = "apache/druid:0.19.0";
```
Praveen2112, line 69: Updating druid should be as a prep commit

docs/src/main/sphinx/connector/druid.md
Praveen2112: If we are planning to pushdown aggregation then I think we need to forcefully perform exact calculation irrespective of Druid's configuration. Can we add a single config property and session property to toggle whether do we need to perform the count distinct operation or not.

conor-mcavoy: > So if we set the this property to DEFAULT and if the Druid cluster has DEFAULT has useApproximateCountDistinct set to true does it uses approximate computation ?

Yes. And it would fail on queries that use, for example, `sum(distinct)` (though this is the same behavior Druid would exhibit natively).

> If we are planning to pushdown aggregation then I think we need to forcefully perform exact calculation irrespective of Druid's configuration

We can do that, essentially by always setting useApproximateCountDistinct=false in the connection properties. The reason I worried about this is if someone actually wanted to use approximate count when using the `query` table function. This would block them from doing so.

> Can we add a single config property and session property to toggle whether do we need to perform the count distinct operation or not

Can you explain this more? Do you essentially mean a boolean property to replace the ternary property that is currently in the PR, essentially `druid.use-approx-count-distinct=true/false` rather than having a "default" option?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
conor-mcavoy: Correct, this PR is focusing on aggregation pushdown. A `connectorExpressionRewriter` must be defined to pass into the `new AggregateFunctionRewriter()` constructor. My understanding is that the `connectorExpressionRewriter` is actually used as part of the aggregation pushdown, when calls are made to `aggregateFunctionRewriter.rewrite()`, for example [here](https://github.com/trinodb/trino/blob/f373b7461c237583776f8a7ee250dd366b2d61ba/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/aggregation/ImplementAvgDecimal.java#L65) in `ImplementAvgDecimal`.

So if I didn't include this `convertPredicate` as well, this PR would essentially make it so that expressions are rewritten when aggregation pushdown happens but not at other times. I figured it would be more consistent to rewrite expressions in more general scenarios too.

Let me know your thoughts. Another approach might be to make the `connectorExpressionRewriter` actually have no rules attached (not even standard rules) so that the behavior is not changing at all - then more rules could be added in the future.

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestingDruidServer.java
conor-mcavoy: You mean just as an isolated commit at the start of the branch (not a separate PR)? Sure I can do that

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: I wanted to avoid potential unknown behavior with AUTOMATIC, which is the default distinct aggregations strategy. I noticed this pattern of setting MARK_DISTINCT in other similar tests, so I did the same. I thought perhaps SPLIT_TO_SUBQUERIES in particular could change the plan if Trino decides to use this strategy on the multiple `distinct` queries, which might fail the test. However I just tested this to be sure and seems like it's not an issue, so I can remove this.

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
```diff
@@ -585,4 +598,149 @@ public void testExecuteProcedureWithInvalidQuery()
         assertUpdate("CALL system.execute('SELECT 1')");
         assertQueryFails("CALL system.execute('invalid')", ".*Non-query expression encountered in illegal context.*");
     }
+
+    @Test
+    @Override
+    public void testAggregationPushdown()
```
chenjian2664, line 604: Why override the test

docs/src/main/sphinx/connector/druid.md
```diff
@@ -17,7 +17,7 @@ database from Trino.
 
 To connect to Druid, you need:
 
-- Druid version 0.18.0 or higher.
+- Druid version 0.19.0 or higher.
```
chenjian2664, line 20: Do we need to fully update this, I am wondering that some users may still use lower version?

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
```diff
@@ -585,4 +598,149 @@ public void testExecuteProcedureWithInvalidQuery()
...
+    {
+        // count()
+        assertThat(query("SELECT count(*) FROM nation")).isFullyPushedDown();
+        assertThat(query("SELECT count(nationkey) FROM nation")).isFullyPushedDown();
```
chenjian2664, line 608: Does those aggregation works same with Trino on empty or one row table now?
I saw this is mentioned in https://github.com/trinodb/trino/pull/4313

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
```diff
@@ -585,4 +598,149 @@ public void testExecuteProcedureWithInvalidQuery()
...
+        assertThat(query("SELECT count(1) FROM nation")).isFullyPushedDown();
+        assertThat(query("SELECT count() FROM nation")).isFullyPushedDown();
+        assertThat(query("SELECT regionkey, count(1) FROM nation GROUP BY regionkey")).isFullyPushedDown();
+        assertThat(query("SELECT regionkey, count(*) FROM nation GROUP BY regionkey")).isFullyPushedDown();
```
chenjian2664, line 612: Can we verify the behaviors that covers all types about the `druid.count-distinct-strategy`?

docs/src/main/sphinx/connector/druid.md
```diff
@@ -68,6 +68,21 @@ properties files.
...
+Druid has the ability to use [an "approximate cardinality algorithm"](https://druid.apache.org/docs/latest/querying/sql-query-context/#sql-query-context-parameters) when calling `COUNT(DISTINCT <col>)`, similar to Trino's {func}`approx_distinct` function. However, when Druid uses this approximation algorithm it cannot accept some [aggregations pushed down](druid-pushdown) from Trino that contain a `DISTINCT`.
+
+:::{list-table}
+:widths: 30, 58, 12
+:header-rows: 1
+
+* - Property name
+  - Description
+  - Default value
+* - ``druid.count-distinct-strategy``
```
chenjian2664, line 81: It's just control the computing strategy in Druid or it will also influence the result?

docs/src/main/sphinx/connector/druid.md
```diff
@@ -68,6 +68,21 @@ properties files.
...
+  - Defines Druid behavior for pushed-down ``DISTINCT`` aggregations by setting (or leaving unset) ``useApproximateCountDistinct``. Must be one of ``DEFAULT``, ``APPROXIMATE``, or ``EXACT``. When ``DEFAULT`` is set, the behavior is left up to the Druid cluster. ``APPROXIMATE`` sets ``useApproximateCountDistinct`` explicitly to ``true``, while ``EXACT`` sets it to ``false``. Note that Druid will reject pushed-down ``DISTINCT`` aggregations unless ``useApproximateCountDistinct=false``. This can either be done on the Trino side by setting this property to ``EXACT``, or it can be done on the Druid cluster.
```
chenjian2664, line 82: Can we have a test for the "reject case"

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
```diff
@@ -585,4 +598,149 @@ public void testExecuteProcedureWithInvalidQuery()
...
+    {
+        String caseSensitiveTable = "case_sensitive";
+        MaterializedResult rows = MaterializedResult.resultBuilder(getSession(), VarcharType.createVarcharType(1), BIGINT, TIMESTAMP_MILLIS)
+                .columnNames(List.of("string_col", "bigint_col", "__time"))
+                .row("A", 1, 0L)
+                .row("a", 2, 0L)
+                .build();
+        try {
+            copyAndIngestTpchData(rows, druidServer, caseSensitiveTable);
+        } catch (Exception e) {
```
chenjian2664, line 702: Why catch it then fail?

docs/src/main/sphinx/connector/druid.md
chenjian2664: Maybe we could extract `BaseDruidConnectorTest`, see the `BaseMysqlConnectorTest`

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: The test in BaseJdbcConnectorTest will fail because it requires `create table` support. Druid cannot support table creation in this manner.

docs/src/main/sphinx/connector/druid.md
conor-mcavoy: > Do we need to fully update this

To support certain types of aggregation pushdown, yes. Because of this issue https://github.com/apache/druid/issues/9949, which is fixed in https://github.com/apache/druid/pull/9959 and included in 0.19.0. Without this, queries like `SELECT agg(*) FROM (SELECT * FROM <table> LIMIT N)` would fail due to a bug in Druid's planner. Note that this is pretty much the only type of query affected, regular aggregation pushdown would still work on 0.18.0.

Druid 0.19.0 was released in July 2020, 0.18.0 in April of 2020, so these are relatively old versions. I suppose we could add some kind of flag to disable pushdown selectively so that users can still use the older version without this one type of query breaking. But I don't feel that we should be beholden to work around 5-year-old bugs in Druid's query planner.

> Maybe we could extract `BaseDruidConnectorTest`, see the `BaseMysqlConnectorTest`

`TestDruidConnectorTest` already extends `BaseJdbcConnectorTest`, just like `BaseMysqlConnectorTest` does. What is it about `BaseMysqlConnectorTest` that you wish to be replicated in the Druid tests? And how does this relate to your previous comment about updating the Druid version?

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: `copyAndIngestTpchData` throws `IOException` and `InterruptedException`. If these are not caught, they would need to be added to the method signature. But the overridden method, from `BaseJdbcConnectorTest`, does not throw these exceptions, so we wouldn't be able to override.

docs/src/main/sphinx/connector/druid.md
conor-mcavoy: It can influence the result. If Druid uses an approximate distinct strategy, the results could vary from query to query. This would happen if you set APPROXIMATE or if you accepted the DEFAULT and the Druid cluster was configured to use approximate.

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: Yes will do. First want to resolve some of the other conversations about how to actually approach this config property, since it may change the implementation.

docs/src/main/sphinx/connector/druid.md
conor-mcavoy: @Praveen2112 I am still working on this, but can you provide any info on my question above? Are you looking for a boolean property to enable/disable use-approx-count-distinct?

At the moment actually, I am leaning towards just enforcing that approx-count-distinct is always off, no configuration property. This could make it impossible for users that actually want this behavior (via the `query` table function, for example) to use it, but it prioritizes correctness.

docs/src/main/sphinx/connector/druid.md
conor-mcavoy: @chenjian2664 any further thoughts on this? Are you good with using the newer version of Druid?

And can you clarify your comment on extracting `BaseDruidConnectorTest`, does `TestDruidConnectorTest` itself not meet what you are looking for?

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: Removed these session properties from the tests

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: ~It's not possible to make an empty table in Druid. Druid creates a table when data is actually ingested. I saw this discussion on that issue as well, it was't totally clear to me if they had actually succeeded in making an empty table but as far as I know it is not possible.~ Actually perhaps this is possible by dropping the segments, I see that was done in #4313, will try it out.

We can replicate something similar to an empty table by adding a filter that will be pushed down to Druid (i.e. not just `WHERE false`) that matches nothing. It turns out that in this case, yes Druid is going to return no rows for functions like `sum`, when standard Trino behavior is to return NULL. Good catch. I'm not sure what to do about this, just rewriting the expression like `case when count(%s) = 0 then null else sum(%s) end` does not seem to work, it still returns 0 rows. In SQL mode Druid returns NULL, and [the Druid docs](https://druid.apache.org/docs/latest/querying/aggregations/#expression-aggregator) describe a parameter `isNullUnlessAggregated` that controls this behavior so it's clearly something Druid _can_ do, just have to figure out how to make it happen. Will look into it.

As for tables with one row, those concerns are for statistical functions like variance and std deviation that don't always have a defined value for a dataset with one row, since I'm not implementing those functions here I think we are good.

docs/src/main/sphinx/connector/druid.md
conor-mcavoy: To fix Druid behavior when pushing down aggregations with empty results (i.e. Druid should return NULL), I had to upgrade this even further to 0.22.0. I spoke to a Druid PMC about this, his feedback was that Druid has been working to support SQL-standard behavior more and more in recent releases, and that we should even target Druid 30.0.0 or higher due to [its bug fixes](https://github.com/apache/druid/releases/tag/druid-30.0.0) in this area.

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConnectorTest.java
conor-mcavoy: I still cannot manage to make an empty table even by dropping the segments, it just deletes the table in Druid, so not sure about how that was done in the past. But as I mentioned above, the exact same behavior can be tested using a filter that returns nothing, so I'm not really concerned about this.

On that front, I was able to verify that Druid does not provide the correct behavior in older versions. Druid 0.22.0 introduced the fix, see https://github.com/apache/druid/pull/11188. I bumped the test version up to 0.22.0 and added a test to confirm.

docs/src/main/sphinx/connector/druid.md
abhishekrb19: >Do we need to fully update this, I am wondering that some users may still use lower version?

I'm wondering if there's a concern around upgrading the pinned Druid version (currently 0.18.0)? As @conor-mcavoy pointed out, 0.18.0 is over 5 years old, and IMO using a newer Trino version with an outdated Druid version should be discouraged. As long as we're testing against these older releases, we're likely to keep encountering bugs that are difficult to work around (like force-setting the runtime property in this case).

Fwiw, in my experience, I haven’t seen users running Druid clusters older than 0.25.0. In the Druid community, we typically encourage users to stay on more "recent" versions, since the project evolves quickly and supporting releases older than ~2 years becomes increasingly challenging.

PR #25713, opened 2025-04-30
[wip] Fix empty join conditions query failure

SemionPar:
## Description

Reproduce `"joinConditions is empty" ` error

## Release notes

```markdown
## Section
```


core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/PushJoinIntoTableScan.java
```diff
@@ -107,6 +107,9 @@ public Result apply(JoinNode joinNode, Captures captures, Context context)
         }
 
         Expression effectiveFilter = getEffectiveFilter(joinNode);
+        if (effectiveFilter.equals(Booleans.TRUE)) {
```
Praveen2112, line 110: I'm afraid we might need some additional checks beyond `TRUE` - What if `SELECT n.name FROM nation n %s orders o ON n.comment is null` i.e apply a condition on probe side ?

Praveen2112: Why are we specific to this effective filter ?

chenjian2664: The joinCondition later is derived from this `effectiveFilter`.
The `joinConditions is empty` is comes from the result `DefaultJdbcMetadata#applyJoin#extractConjuncts(joinCondition)`, that means the results of the `ConnectorExpressionTranslator.translateConjuncts` is all `TRUE`.

SemionPar: > I'm afraid we might need some additional checks beyond TRUE - What if SELECT n.name FROM nation n %s orders o ON n.comment is null i.e apply a condition on probe side ?

Great suggestion! Added more assertions and it revealed that there is another issue here as well.

For LEFT JOIN with lone predicate `ON n.comment is null` on the probe side, there is an error on pushdown:

```sql
--trino
SET SESSION postgresql.join_pushdown_strategy = 'eager';
SELECT n.name FROM postgresql.tpch.nation n LEFT JOIN postgresql.tpch.orders o ON n.comment is null;
--postgresql
SELECT "name_0" FROM (SELECT "name_0", "comment_1",  FROM (SELECT "name" AS "name_0", "comment" AS "comment_1" FROM (SELECT "name", "comment" FROM "tpch"."nation") l) l LEFT JOIN (SELECT  FROM (SELECT 1 x FROM "tpch"."orders") r) r ON (("comment_1") IS NULL)) o
---------------------------------------------------^
--error
--double-checking with sqlserver
SELECT "name_0" FROM (SELECT "name_0", "comment_1",  FROM (SELECT "name" AS "name_0", "comment" AS "comment_1" FROM (SELECT "name", "comment" FROM "database_950c7af8f1894a50beef3d652468ffeb"."dbo"."nation") l) l LEFT JOIN (SELECT  FROM (SELECT 1 x FROM "database_950c7af8f1894a50beef3d652468ffeb"."dbo"."orders") r) r ON (("comment_1") IS NULL)) o
--same issue
```

This is with&without my change to `PushJoinIntoTableScan`.

Lone predicate on the build side is fine:

```sql
--trino
SET SESSION postgresql.join_pushdown_strategy = 'eager';
SELECT n.name FROM postgresql.tpch.nation n LEFT JOIN postgresql.tpch.orders o ON o.comment is null
--postgresql, 2 scans, join not pushded down
SELECT 1 x FROM "tpch"."orders" WHERE "comment" IS NULL
SELECT "name" FROM "tpch"."nation"
```

SemionPar: > Why are we specific to this effective filter ?

IIUC empty effective filter (=BOOLEAN::TRUE) is what causing `joinConditions is empty` error downstream

SemionPar: > Maybe put a check in DefaultJdbcMetadata#applyJoin is more direct

Great idea @chenjian2664 - applied!

Adding separate commit with query builder fix to `SELECT "name_0" FROM (SELECT "name_0", "comment_1",  FROM ...` misshaped query.

anusudarsan (approved):

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
```diff
@@ -690,6 +690,11 @@ public Optional<JoinApplicationResult<ConnectorTableHandle>> applyJoin(
             joinConditions.add(converted.get());
         }
 
+        ImmutableList<ParameterizedExpression> convertedJoinConditions = joinConditions.build();
```
chenjian2664, line 693: nit, using `List`

SemionPar:
Oracle ci failure is unrelated, PTAL @Praveen2112

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultQueryBuilder.java
```diff
@@ -129,9 +129,12 @@ public PreparedQuery prepareJoinQuery(
         String query = format(
                 // The subquery aliases (`l` and `r`) are needed by some databases, but are not needed for expressions
                 // The joinConditions and output columns are aliased to use unique names.
-                "SELECT %s, %s FROM (SELECT %s FROM (%s) l) l %s (SELECT %s FROM (%s) r) r ON %s",
-                formatProjectionAliases(client, leftProjections.values()),
-                formatProjectionAliases(client, rightProjections.values()),
+                "SELECT %s FROM (SELECT %s FROM (%s) l) l %s (SELECT %s FROM (%s) r) r ON %s",
+                formatProjectionAliases(
+                        client, ImmutableList.<String>builder()
```
Praveen2112, line 134: nit: ImmutableList to new line

Praveen2112 (approved):
% nit

Praveen2112:
@SemionPar Can you please update the release notes.

PR #25716, opened 2025-05-01
Ensure that createViewSession preserves the trace token

huw0:
## Description

`Session`.`createViewSession` creates a new Session that is used when analysing views.

Presently this does not pass through the trace token. This is breaking the use of views with our custom connectors which mandate that this token is populated.

Therefore this is a very simple PR to ensure that the trace token is populated in `createViewSession` if it is present on the outer session.

## Release notes

```markdown
## General
* createViewSession should preserve the client trace token ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/Session.java
+++ b/core/trino-main/src/main/java/io/trino/Session.java
@@ -661,6 +661,7 @@ public final class Session
                 .setRemoteUserAddress(getRemoteUserAddress().orElse(null))
                 .setUserAgent(getUserAgent().orElse(null))
                 .setClientInfo(getClientInfo().orElse(null))
+                .setTraceToken(getTraceToken())
                 .setStart(getStart())
                 .build();
     }
--- a/core/trino-main/src/test/java/io/trino/TestSession.java
+++ b/core/trino-main/src/test/java/io/trino/TestSession.java
@@ -13,9 +13,17 @@
  */
 package io.trino;
 
+import io.trino.spi.QueryId;
+import io.trino.spi.security.Identity;
+import io.trino.spi.type.TimeZoneKey;
+import io.trino.transaction.TransactionId;
 import org.junit.jupiter.api.Test;
 
+import java.time.Instant;
+import java.util.Collections;
+import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 
 import static io.trino.testing.TestingSession.testSessionBuilder;
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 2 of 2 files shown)


core/trino-main/src/main/java/io/trino/Session.java
```diff
@@ -661,6 +661,7 @@ public Session createViewSession(Optional<String> catalog, Optional<String> sche
                 .setRemoteUserAddress(getRemoteUserAddress().orElse(null))
                 .setUserAgent(getUserAgent().orElse(null))
                 .setClientInfo(getClientInfo().orElse(null))
+                .setTraceToken(traceToken)
```
Praveen2112, line 664: nit:  `.setTraceToken(getTraceToken())`

huw0: Thanks, I should've caught that.

dain (approved):

dain:
This looks good to me, but if a test is possible, please add that.  If not (or really annoying), please let us know, and one of us will merge it as is.

Praveen2112:
@huw0 For testing I think we could use `MockConnectorMetadata` which could make use of the trace token at the time of its execution.

huw0:
I've added some unit testing around `createViewSession` and rebased the branch.
There seem to be some test failures caused by github rate limiting and some random plugin failures that change every time. Expect these should pass once any wider issues resolve.

Praveen2112:
Instead of this test - WDYT about https://github.com/trinodb/trino/pull/25716#issuecomment-2853389163 ?

huw0:
@Praveen2112 - I've added an additional commit with an alternative test using the approach you've outlined.

huw0:
> I'd be in favour of dropping this 3rd commit and sticking with the first two. The unit test alone increases the existing coverage.

@dain @Praveen2112 - Please see comment above. I'd like to get this merged soon if there are no objections?

huw0:
Dropped 3rd commit and rebased.

huw0:
PR updated against latest master - but otherwise unchanged.

core/trino-main/src/test/java/io/trino/TestSession.java
```diff
@@ -61,4 +69,53 @@ public void testAddSecondCatalogProperty()
...
+                .setLocale(locale)
+                .setRemoteUserAddress(remoteUserAddress)
+                .setUserAgent(userAgent)
+                .setClientInfo(clientInfo)
+                .setTraceToken(traceToken)
+                .setStart(start)
+                .build();
+
+        Session viewSession = originalSession.createViewSession(catalog, schema, identity,
+                Collections.emptyList());
```
raunaqmorarka: `Collections.emptyList()` -> `ImmutableList.of()`
Please keep all arguments on same line, or each argument on new line

core/trino-main/src/test/java/io/trino/TestSession.java
```diff
@@ -61,4 +69,53 @@ public void testAddSecondCatalogProperty()
                         "first_property", "some_value",
                         "second_property", "another_value")));
     }
+
+    @Test
+    public void testCreateViewSession()
+    {
+        final var catalog = Optional.of("test_catalog");
```
raunaqmorarka: We don't declare local variables as `final`

core/trino-main/src/test/java/io/trino/TestSession.java
```diff
@@ -61,4 +69,53 @@ public void testAddSecondCatalogProperty()
...
+        final var schema = Optional.of("test_schema");
```
raunaqmorarka: We avoid using `var`.
Please look at https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#code-style to get an idea of the overall development guidelines

huw0:
Thanks @raunaqmorarka for review and merge.

PR #25725, opened 2025-05-03
JDBC Client: fix Exception Handling Bug in TrinoResultSet

kilnamkim:
### Description
When using the configuration setting hive.query-partition-filter-required=true in Trino's JDBC client, a SQLException is expected to be thrown when a query is missing a required partition column. However, no error was raised in this case.
The root cause is that when a query fails due to a missing partition filter, results.getColumns() returns an empty List<Column> object (i.e., a list with a size of 0). As a result, the exception handling logic that depends on the presence of columns was not triggered as expected.
This fix ensures that when the query execution result is FAILED, the column retrieval will also fail accordingly, allowing the exception handling code to run as intended.

---

Change Summary  
- Type: Bug fix  
- Impact Area: JDBC Client  
- User-facing Description: Fixes a bug in Trino JDBC where exception handling in TrinoResultSet was not triggered when a query failed due to missing partition filters.

### DIFF
--- a/client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoResultSet.java
+++ b/client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoResultSet.java
@@ -129,7 +129,7 @@ public class TrinoResultSet
             QueryStatusInfo results = client.currentStatusInfo();
             progressCallback.accept(QueryStats.create(results.getId(), results.getStats()));
             List<Column> columns = results.getColumns();
-            if (columns != null) {
+            if (columns != null && !results.getStats().getState().equals("FAILED")) {
                 return columns;
             }
             client.advance();
--- a/client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcPartitionFilterRequired.java
+++ b/client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcPartitionFilterRequired.java
@@ -1,0 +1,164 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.jdbc;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.Module;
+import com.google.inject.Scopes;
+import io.airlift.log.Logging;
+import io.trino.plugin.blackhole.BlackHolePlugin;
+import io.trino.plugin.hive.HivePlugin;
+import io.trino.server.testing.TestingTrinoServer;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.connector.ConnectorTransactionHandle;
+import io.trino.spi.connector.InMemoryRecordSet;
+import io.trino.spi.connector.RecordCursor;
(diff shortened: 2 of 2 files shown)


client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcPartitionFilterRequired.java
```diff
@@ -0,0 +1,164 @@
...
+    private Connection createConnection(String extra)
+            throws SQLException
+    {
+        String url = format("jdbc:trino://%s/hive/default?%s", server.getAddress(), extra);
+        return DriverManager.getConnection(url, "admin", null);
+    }
+
+    private static class ExtraCredentialsSystemTable
+            implements SystemTable
+    {
```
Praveen2112, line 136: Do we need this System table ?

client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcPartitionFilterRequired.java
```diff
@@ -0,0 +1,164 @@
...
+import static java.lang.String.format;
+import static java.util.concurrent.Executors.newCachedThreadPool;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+import static org.junit.jupiter.api.parallel.ExecutionMode.CONCURRENT;
+
+@TestInstance(PER_CLASS)
+@Execution(CONCURRENT)
+public class TestJdbcPartitionFilterRequired
```
Praveen2112, line 60: Is the issue is very specific to PartitionFilter for hive ? If it is a race condition then it should be seen by other connectors as well.

chenjian2664, line 60: Please follow the [guideline](https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations): `Test classes should be defined as package-private and final.`

client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcPartitionFilterRequired.java
```diff
@@ -0,0 +1,164 @@
...
+            throws Exception
+    {
+        closeAll(
+                server,
+                executor::shutdownNow);
+        server = null;
+    }
+
+    @Test
+    public void testFilterRequiredPartitions()
```
chenjian2664, line 107: Please follow the guideline: `Test methods should be defined as package-private.`

client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoResultSet.java
```diff
@@ -129,7 +129,7 @@ private static List<Column> getColumns(StatementClient client, Consumer<QuerySta
             QueryStatusInfo results = client.currentStatusInfo();
             progressCallback.accept(QueryStats.create(results.getId(), results.getStats()));
             List<Column> columns = results.getColumns();
-            if (columns != null) {
+            if (columns != null && !results.getStats().getState().equals("FAILED")) {
```
chenjian2664, line 132: Do you know how does this get empty columns? @wendigo 
I think it's a bug that should fix in the place that return empty columns

client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcPartitionFilterRequired.java
```diff
@@ -0,0 +1,164 @@
...
+    {
+        String sql = "SELECT id FROM test_required_partition_filter WHERE a = '1' ";
+        assertThatThrownBy(() -> {
+            try (Connection connection = createConnection()) {
+                Statement statement = connection.createStatement();
+                statement.execute(sql);
+                ResultSet resultSet = statement.getResultSet();
+                assertThat(resultSet.next()).isTrue();
+            }
+        }).isInstanceOf(SQLException.class);
```
chenjian2664, line 118: What's the error message ?

PR #25729, opened 2025-05-04
Allow row column for equality delete

### LINKED ISSUE #25720: Iceberg equality deletes with struct type columns should be supported
From Iceberg documentation on [equality-delete-files](https://iceberg.apache.org/spec/?h=equality#equality-delete-files) and [identifier-field-Ids](https://iceberg.apache.org/spec/?h=equality#identifier-field-ids), row type should be allowed in the equality delete columns but we get
```
Caused by: java.lang.UnsupportedOperationException: Type not supported as partition column: row(a varchar, b varchar, c varchar)
	at io.trino.plugin.iceberg.IcebergPageSink.getIcebergValue(IcebergPageSink.java:417)
	at io.trino.plugin.iceberg.delete.TrinoRow.<init>(TrinoRow.java:35)
	at io.trino.plugin.iceberg.delete.EqualityDeleteFilter.readEqualityDeletes(EqualityDeleteFilter.java:74)
	at io.trino.plugin.iceberg.IcebergPageSourceProvider.readDeletes(IcebergPageSourceProvider.java:417)
```
The cause of this is that we are reusing `IcebergPageSink::getIcebergValue` which is tailored for partitioning columns that have different restrictions.

Since `io.trino.plugin.iceberg.delete.TrinoRow` is only used in the delete files, we can have a separate method to load the values without applying the same restrictions for partitioning columns.

From the spec, using required `REAL`/`FLOAT` or `DOUBLE` columns for equality deletes should be failed.

Aside from these, the error message right now is misleading to the end users.
```
Caused by: java.lang.UnsupportedOperationException: Type not supported as partition column: row(a varchar, b varchar, c varchar)
```

weijiii:
## Description
* Fixes #25720 

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/TrinoRow.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/TrinoRow.java
@@ -13,26 +13,49 @@
  */
 package io.trino.plugin.iceberg.delete;
 
+import io.trino.spi.block.Block;
+import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.SourcePage;
+import io.trino.spi.type.DecimalType;
+import io.trino.spi.type.RowType;
 import io.trino.spi.type.Type;
+import io.trino.spi.type.VarbinaryType;
+import io.trino.spi.type.VarcharType;
 import org.apache.iceberg.StructLike;
 
 import java.util.Arrays;
+import java.util.List;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static io.trino.plugin.iceberg.IcebergPageSink.getIcebergValue;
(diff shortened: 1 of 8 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/TrinoRow.java
```diff
@@ -59,4 +82,84 @@ public String toString()
     {
         return "TrinoRow" + Arrays.toString(values);
     }
+
+    static Object getObjectValue(ConnectorSession session, Block block, int position, Type type)
```
chenjian2664, line 86: Can we move this logic to IcebergPageSink, maybe an overload with `getIcebergValue`, or even maybe we could extract a util

PR #25742, opened 2025-05-06
Use interface for declarations

losipiuk:


martint (approved):

Praveen2112:
Is it possible to find them via error-prone compilers or maven plugin ?

chenjian2664 (approved):

losipiuk:
> Is it possible to find them via error-prone compilers or maven plugin ?

It is not trivial. E.g. when sometimes you really want to have variable declared as Immutable... given it hase a few more public methods than interface

PR #25749, opened 2025-05-07
Improve cache invalidation by keying on path and table name

churromorales:
## Description
This change updates the cache key to use both Path and
SchemaTableName, enabling more precise identification of cached entries.
It allows us to iterate over cache keys effectively and target only the
relevant entries during invalidation.

This is especially useful for partitioned tables, where coarse-grained
invalidation can lead to unnecessary cache evictions and performance
degradation.

In response to this in the `CachingDirectoryListener`
```
//TODO use a cache key based on Path & SchemaTableName and iterate over the cache keys
// to deal more efficiently with cache invalidation scenarios for partitioned tables.
```

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
@@ -18,6 +18,7 @@ import com.google.common.cache.Cache;
 import com.google.common.cache.Weigher;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
+import io.airlift.log.Logger;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import io.trino.cache.EvictableCacheBuilder;
@@ -51,9 +52,9 @@ import static java.util.function.Predicate.not;
 public class CachingDirectoryLister
         implements DirectoryLister
 {
-    //TODO use a cache key based on Path & SchemaTableName and iterate over the cache keys
-    // to deal more efficiently with cache invalidation scenarios for partitioned tables.
-    private final Cache<Location, ValueHolder> cache;
+    private static final Logger log = Logger.get(CachingDirectoryLister.class);
+
+    private final Cache<CacheKey, ValueHolder> cache;
     private final Predicate<SchemaTableName> tablePredicate;
     private final Predicate<FileEntry> filterPredicate;
 
@@ -82,7 +83,7 @@ public class CachingDirectoryLister
         requireNonNull(filterPredicate, "filterPredicate is null");
         this.cache = EvictableCacheBuilder.newBuilder()
                 .maximumWeight(maxSize.toBytes())
-                .weigher((Weigher<Location, ValueHolder>) (key, value) -> toIntExact(estimatedSizeOf(key.toString()) + value.getRetainedSizeInBytes()))
+                .weigher((Weigher<CacheKey, ValueHolder>) (key, value) -> toIntExact(key.getRetainedSizeInBytes() + value.getRetainedSizeInBytes()))
                 .expireAfterWrite(expireAfterWrite.toMillis(), TimeUnit.MILLISECONDS)
                 .shareNothingWhenDisabled()
                 .recordStats()
@@ -123,18 +124,19 @@ public class CachingDirectoryLister
             return new TrinoFileStatusRemoteIterator(fs.listFiles(location), filterPredicate);
         }
 
-        return listInternal(fs, location);
+        return listInternal(fs, location, table.getSchemaTableName());
     }
 
-    private RemoteIterator<TrinoFileStatus> listInternal(TrinoFileSystem fs, Location location)
+    private RemoteIterator<TrinoFileStatus> listInternal(TrinoFileSystem fs, Location location, SchemaTableName schemaTableName)
             throws IOException
     {
-        ValueHolder cachedValueHolder = uncheckedCacheGet(cache, location, ValueHolder::new);
+        CacheKey cacheKey = new CacheKey(location, schemaTableName);
+        ValueHolder cachedValueHolder = uncheckedCacheGet(cache, cacheKey, ValueHolder::new);
         if (cachedValueHolder.getFiles().isPresent()) {
             return new SimpleRemoteIterator(cachedValueHolder.getFiles().get().iterator());
         }
 
-        return cachingRemoteIterator(cachedValueHolder, createListingRemoteIterator(fs, location, filterPredicate), location);
+        return cachingRemoteIterator(cachedValueHolder, createListingRemoteIterator(fs, location, filterPredicate), cacheKey);
     }
 
     private static RemoteIterator<TrinoFileStatus> createListingRemoteIterator(TrinoFileSystem fs, Location location, Predicate<FileEntry> filterPredicate)
@@ -144,9 +146,10 @@ public class CachingDirectoryLister
     }
 
     @Override
-    public void invalidate(Location location)
+    public void invalidate(Location location, SchemaTableName schemaTableName)
     {
-        cache.invalidate(location);
+        log.debug("Invalidating cache for schemaTableName: %s and location: %s", schemaTableName, location);
+        cache.invalidate(new CacheKey(location, schemaTableName));
     }
 
     @Override
@@ -154,11 +157,14 @@ public class CachingDirectoryLister
     {
         if (isCacheEnabledFor(table.getSchemaTableName()) && isLocationPresent(table.getStorage())) {
             if (table.getPartitionColumns().isEmpty()) {
-                cache.invalidate(Location.of(table.getStorage().getLocation()));
+                log.debug("Invalidating cache for unpartitioned table: %s", table.getSchemaTableName());
+                cache.invalidate(new CacheKey(Location.of(table.getStorage().getLocation()), table.getSchemaTableName()));
             }
             else {
                 // a partitioned table can have multiple paths in cache
-                cache.invalidateAll();
+                SchemaTableName tableName = table.getSchemaTableName();
+                log.debug("Invalidating cache for partitioned table: %s", table.getSchemaTableName());
+                cache.asMap().keySet().removeIf(key -> key.schemaTableName().equals(tableName));
             }
         }
     }
@@ -167,17 +173,20 @@ public class CachingDirectoryLister
     public void invalidate(Partition partition)
     {
         if (isCacheEnabledFor(partition.getSchemaTableName()) && isLocationPresent(partition.getStorage())) {
-            cache.invalidate(Location.of(partition.getStorage().getLocation()));
+            log.debug("Invalidating partition cache for table: %s partition: %s", partition.getSchemaTableName(), partition.getStorage().getLocation());
+            Location partitionLocation = Location.of(partition.getStorage().getLocation());
+            cache.invalidate(new CacheKey(partitionLocation, partition.getSchemaTableName()));
         }
     }
 
     @Override
     public void invalidateAll()
     {
+        log.debug("Invalidating partition cache (all)");
         cache.invalidateAll();
     }
 
-    private RemoteIterator<TrinoFileStatus> cachingRemoteIterator(ValueHolder cachedValueHolder, RemoteIterator<TrinoFileStatus> iterator, Location location)
+    private RemoteIterator<TrinoFileStatus> cachingRemoteIterator(ValueHolder cachedValueHolder, RemoteIterator<TrinoFileStatus> iterator, CacheKey cacheKey)
     {
         return new RemoteIterator<>()
         {
@@ -191,7 +200,7 @@ public class CachingDirectoryLister
                 if (!hasNext) {
                     // The cachedValueHolder acts as an invalidation guard. If a cache invalidation happens while this iterator goes over
                     // the files from the specified path, the eventually outdated file listing will not be added anymore to the cache.
-                    cache.asMap().replace(location, cachedValueHolder, new ValueHolder(files));
+                    cache.asMap().replace(cacheKey, cachedValueHolder, new ValueHolder(files));
                 }
                 return hasNext;
             }
@@ -244,9 +253,9 @@ public class CachingDirectoryLister
     }
 
     @Override
-    public boolean isCached(Location location)
+    public boolean isCached(Location location, SchemaTableName schemaTableName)
     {
-        ValueHolder cached = cache.getIfPresent(location);
+        ValueHolder cached = cache.getIfPresent(new CacheKey(location, schemaTableName));
         return cached != null && cached.getFiles().isPresent();
     }
 
@@ -291,6 +300,24 @@ public class CachingDirectoryLister
         public long getRetainedSizeInBytes()
         {
             return INSTANCE_SIZE + sizeOf(files, value -> estimatedSizeOf(value, TrinoFileStatus::getRetainedSizeInBytes));
+        }
+    }
+
+    private record CacheKey(Location location, SchemaTableName schemaTableName)
+    {
+        private static final long INSTANCE_SIZE = instanceSize(CacheKey.class);
+
+        private CacheKey(Location location, SchemaTableName schemaTableName)
+        {
+            this.location = requireNonNull(location, "location is null");
+            this.schemaTableName = requireNonNull(schemaTableName, "schemaTableName is null");
+        }
+
+        public long getRetainedSizeInBytes()
+        {
+            return INSTANCE_SIZE +
+                    estimatedSizeOf(location.toString()) +
+                    schemaTableName.getRetainedSizeInBytes();
         }
     }
 }
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionDirectoryListingCacheKey.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionDirectoryListingCacheKey.java
@@ -14,6 +14,7 @@
 package io.trino.plugin.hive.fs;
 
 import io.trino.filesystem.Location;
+import io.trino.spi.connector.SchemaTableName;
 
 import java.util.Objects;
 
@@ -28,11 +29,13 @@ public class TransactionDirectoryListingCacheKey
 
     private final long transactionId;
     private final Location path;
+    private final SchemaTableName schemaTableName;
 
-    public TransactionDirectoryListingCacheKey(long transactionId, Location path)
+    public TransactionDirectoryListingCacheKey(long transactionId, Location path, SchemaTableName schemaTableName)
     {
         this.transactionId = transactionId;
         this.path = requireNonNull(path, "path is null");
+        this.schemaTableName = requireNonNull(schemaTableName, "schemaTableName is null");
     }
 
     public Location getPath()
@@ -40,9 +43,16 @@ public class TransactionDirectoryListingCacheKey
         return path;
     }
 
+    public SchemaTableName getSchemaTableName()
+    {
+        return schemaTableName;
+    }
+
     public long getRetainedSizeInBytes()
     {
-        return INSTANCE_SIZE + estimatedSizeOf(path.toString());
+        return INSTANCE_SIZE +
+               estimatedSizeOf(path.toString()) +
+                schemaTableName.getRetainedSizeInBytes();
     }
 
     @Override
@@ -55,13 +65,15 @@ public class TransactionDirectoryListingCacheKey
             return false;
         }
         TransactionDirectoryListingCacheKey that = (TransactionDirectoryListingCacheKey) o;
-        return transactionId == that.transactionId && path.equals(that.path);
+        return transactionId == that.transactionId &&
+               path.equals(that.path) &&
+               schemaTableName.equals(that.schemaTableName);
     }
 
     @Override
     public int hashCode()
     {
-        return Objects.hash(transactionId, path);
+        return Objects.hash(transactionId, path, schemaTableName);
     }
 
     @Override
@@ -70,6 +82,7 @@ public class TransactionDirectoryListingCacheKey
         return toStringHelper(this)
                 .add("transactionId", transactionId)
                 .add("path", path)
+                .add("schemaTableName", schemaTableName)
                 .toString();
     }
 }
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionScopeCachingDirectoryLister.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionScopeCachingDirectoryLister.java
@@ -22,6 +22,7 @@ import io.trino.filesystem.TrinoFileSystem;
 import io.trino.metastore.Partition;
 import io.trino.metastore.Storage;
 import io.trino.metastore.Table;
+import io.trino.spi.connector.SchemaTableName;
 import jakarta.annotation.Nullable;
 
 import java.io.IOException;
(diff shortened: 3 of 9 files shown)


plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -293,4 +302,22 @@ public long getRetainedSizeInBytes()
...
+        {
+            this.location = requireNonNull(location, "location is null");
+            this.schemaTableName = requireNonNull(schemaTableName, "schemaTableName is null");
+        }
+
+        public long getRetainedSizeInBytes()
+        {
+            return INSTANCE_SIZE +
+                    estimatedSizeOf(location.toString()) +
+                    (schemaTableName != null ? estimatedSizeOf(schemaTableName.toString()) : 0);
```
pettyjamesm, line 320: `schemaTableName` is never null as enforced by the constructor validation, and you can use `schemaTableName.getRetainedSizeInBytes()`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -146,19 +147,23 @@ private static RemoteIterator<TrinoFileStatus> createListingRemoteIterator(Trino
...
     {
-        cache.invalidate(location);
+        log.debug("Invalidating cache for location: %s", location);
+        cache.asMap().keySet().removeIf(key -> key.location().equals(location));
     }
 
     @Override
     public void invalidate(Table table)
     {
         if (isCacheEnabledFor(table.getSchemaTableName()) && isLocationPresent(table.getStorage())) {
```
pettyjamesm, line 157: The `isLocationPresent` check here seems suspicious (for both `invalidate(Table)` and `invalidate(Partition)`). It would be good to look back at usage sites and see whether this can be simplified to a single `invalidate(Table, Location)` method that always has both arguments- because skipping an invalidation due to a missing location seems like a bug waiting to happen.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -246,8 +254,9 @@ public long getRequestCount()
     @Override
     public boolean isCached(Location location)
```
pettyjamesm, line 255: This is the method in a hot path (via `BackgroundHiveSplitLoader`) that should be able to provide both the table name and location to do a point lookup instead of traversing the cache entries.

churromorales: oh cool, i wasn't sure if i could change these API signatures, ill go ahead and take a look and see what i can do. Thanks for the headsup

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
churromorales: cool yeah this seems totally doable, most of the changes are from the tests.  I can make this work.  Thanks for the heads up :)

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -146,19 +147,23 @@ private static RemoteIterator<TrinoFileStatus> createListingRemoteIterator(Trino
     @Override
     public void invalidate(Location location)
     {
-        cache.invalidate(location);
+        log.debug("Invalidating cache for location: %s", location);
+        cache.asMap().keySet().removeIf(key -> key.location().equals(location));
```
pettyjamesm, line 151: I think this will need to change into `CacheUtils.invalidateAllIf(cache, key -> key.location().equals(location))` to avoid the `UnsupportedOperationException` issue you're seeing. Also, if we're invalidating purely on the basis of location (which might be valid, or maybe not) then maybe we actually want something more similar to `.startsWith` instead of `.equals` on that check?

churromorales: looking at the callers of this method, I see that the schema table name is always available to pass, want me to try and give it a shot to change the signature to `public void invalidate(Location location, SchemaTableName schemaTableName)`?

pettyjamesm: Yeah it looks like location based invalidation is always in the context of some table or partition, so it looks like we can either:
- change the signature to `invalidate(SchemaTableName schemaTableName, Location)` and update all usage sites to pass it
- keep separate `invalidate(Table)` and `invalidatePartition(Partition)` methods and do that unwrapping internally

Similar changes will probably need to be made to `TransactionScopeCachingDirectoryLister` whichever route makes the most sense.

churromorales: There was also a TODO in the `TransactionScopeCachingDirectoryLister` so I am going to go with the first option and change the signature, good to make the cache efficient.  I just pushed up the changes, there is one test failing locally which I am looking at, want to run the rest of the build in GHA though to make sure I didn't miss anything.  Thanks for the review, really appreciate it.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
churromorales: done! good call!

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
churromorales: i added the table parameter for this

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -244,10 +252,9 @@ public long getRequestCount()
     }
 
     @Override
-    public boolean isCached(Location location)
+    public boolean isCached(Location location, SchemaTableName schemaTableName)
     {
-        ValueHolder cached = cache.getIfPresent(location);
-        return cached != null && cached.getFiles().isPresent();
+        return cache.asMap().containsKey(new CacheKey(location, schemaTableName));
```
pettyjamesm, line 257: The previous check was ensuring that the cache entry also had files materialized, which I think we want to preserve here. Basically, it's not sufficient to say that the location is cached just because there's an entry- the entry has to have completed a file listing in cache

churromorales:
the checkstyle issue is sorted, i don't know why the GHA didn't pick it up.  The test `testConcurrentDirectoryListingException` was failing and I think it was due to the thread safety on the test itself rather than the code, but would love another set of eyes on this.  I'll squash the commits and fix the message at the end, would just love a review before I do that.  Thank you again for taking the time to review this PR.

Praveen2112:
@findinpath / @churromorales Can we get an explanation on why ? 
> use a cache key based on Path & SchemaTableName and iterate over the cache key to deal more efficiently with cache invalidation scenarios for partitioned tables. Is there is any test case which mimics the same behaviour ?

findinpath:
> Can we get an explanation on why ?

>> use a cache key based on Path & SchemaTableName and iterate over the cache key to deal more efficiently with cache invalidation scenarios for partitioned tables. Is there is any test case which mimics the same behaviour ?

Invalidating a table leads to losing all the paths from the cache at the moment.

https://github.com/trinodb/trino/blob/993a08840651a3e7c7a6e16ffbbc6bc2c02346af/plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java#L160-L161

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -51,9 +52,8 @@
 public class CachingDirectoryLister
         implements DirectoryLister
 {
-    //TODO use a cache key based on Path & SchemaTableName and iterate over the cache keys
-    // to deal more efficiently with cache invalidation scenarios for partitioned tables.
-    private final Cache<Location, ValueHolder> cache;
+    private static final Logger log = Logger.get(CachingDirectoryLister.class);
+    private final Cache<CacheKey, ValueHolder> cache;
```
Praveen2112, line 56: nit: A new line after `    private static final Logger log = Logger.get(CachingDirectoryLister.class);`

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/TestTransactionScopeCachingDirectoryLister.java
```diff
@@ -159,8 +161,8 @@ private static class CountingDirectoryLister
             implements DirectoryLister
     {
         private final Map<Location, List<TrinoFileStatus>> fileStatuses;
-        private int listCount;
-        private boolean throwException;
+        private final AtomicInteger listCount = new AtomicInteger();
+        private volatile boolean throwException;
```
Praveen2112, line 165: nit: This can be as a prep commit right ? Before fixing the actual issue

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/BaseCachingDirectoryListerTest.java
```diff
@@ -315,13 +315,13 @@ public void testDropPartitionedTable()
...
+        assertThat(isCached(tableGroup1PartitionLocation, SchemaTableName.schemaTableName(TPCH_SCHEMA, "drop_partitioned_table"))).isTrue();
+        assertThat(isCached(tableGroup2PartitionLocation, SchemaTableName.schemaTableName(TPCH_SCHEMA, "drop_partitioned_table"))).isTrue();
+        assertThat(isCached(tableGroup3PartitionLocation, SchemaTableName.schemaTableName(TPCH_SCHEMA, "drop_partitioned_table"))).isTrue();
         assertUpdate("DROP TABLE drop_partitioned_table");
-        assertThat(isCached(tableGroup1PartitionLocation)).isFalse();
-        assertThat(isCached(tableGroup2PartitionLocation)).isFalse();
-        assertThat(isCached(tableGroup3PartitionLocation)).isFalse();
+        assertThat(isCached(tableGroup1PartitionLocation, SchemaTableName.schemaTableName(TPCH_SCHEMA, "drop_partitioned_table"))).isFalse();
+        assertThat(isCached(tableGroup2PartitionLocation, SchemaTableName.schemaTableName(TPCH_SCHEMA, "drop_partitioned_table"))).isFalse();
+        assertThat(isCached(tableGroup3PartitionLocation, SchemaTableName.schemaTableName(TPCH_SCHEMA, "drop_partitioned_table"))).isFalse();
```
Praveen2112, line 324: Can we add an additional assertion which ensure other tables are not removed ? Or was it added ?

Praveen2112 (commented):
Can we group the logical set of commits ? I think we would have two commits - 
1. Performing the cleanup
2. Improve cache invalidation based on path and table name

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionScopeCachingDirectoryLister.java
```diff
@@ -123,7 +122,8 @@ public void invalidate(Table table)
     public void invalidate(Partition partition)
     {
         if (isLocationPresent(partition.getStorage())) {
-            cache.invalidate(new TransactionDirectoryListingCacheKey(transactionId, Location.of(partition.getStorage().getLocation())));
+            SchemaTableName schemaTableName = new SchemaTableName(partition.getDatabaseName(), partition.getTableName());
```
pettyjamesm, line 125: Nit: you can just call `Partition#getSchemaTableName()` here

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/TestTransactionScopeCachingDirectoryLister.java
```diff
@@ -125,7 +126,8 @@ public void testConcurrentDirectoryListingException()
         Location path = Location.of("file:/x");
 
         CountingDirectoryLister countingLister = new CountingDirectoryLister(ImmutableMap.of(path, ImmutableList.of(file)));
-        DirectoryLister cachingLister = new TransactionScopeCachingDirectoryListerFactory(DataSize.ofBytes(600), Optional.empty()).get(countingLister);
+        // Set concurrencyLevel to 1 to ensure deterministic behavior
+        DirectoryLister cachingLister = new TransactionScopeCachingDirectoryListerFactory(DataSize.ofBytes(600), Optional.of(1)).get(countingLister);
```
pettyjamesm, line 130: Hrm... setting the concurrency level to 1 seems like "cheating" for the purposes of the test, I'd be interested to know how this fails without setting it.

Unrelated to this change, but `TransactionScopeCachingDirectoryListerFactory` also includes the same `// TODO` comment that we're addressing in the other places- we should delete the comment there in this PR since it will no longer be relevant.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -244,9 +252,10 @@ public long getRequestCount()
     }
 
     @Override
-    public boolean isCached(Location location)
+    public boolean isCached(Location location, SchemaTableName schemaTableName)
     {
-        ValueHolder cached = cache.getIfPresent(location);
+        CacheKey cacheKey = new CacheKey(location, schemaTableName);
```
pettyjamesm, line 257: Nit: no need to store a local variable here since the result is only used to pass into the `cache.getIfPresent` call

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionDirectoryListingCacheKey.java
```diff
@@ -28,21 +29,29 @@ public class TransactionDirectoryListingCacheKey
...
+    public SchemaTableName getSchemaTableName()
+    {
+        return schemaTableName;
+    }
+
     public long getRetainedSizeInBytes()
     {
-        return INSTANCE_SIZE + estimatedSizeOf(path.toString());
+        return INSTANCE_SIZE +
+               estimatedSizeOf(path.toString()) + schemaTableName.getRetainedSizeInBytes();
```
pettyjamesm, line 54: Nit: let's line break for each element being added

pettyjamesm (commented):
Added some mostly minor comments, although I'm curious about the test failure you saw that required reducing concurrency on the test cache- maybe there's a better way to address it.

Other than addressing the other comments and squashing the commits down, the change looks good to me.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/TestTransactionScopeCachingDirectoryLister.java
churromorales: So yeah i thought so as well, so when we set the concurrency optional to empty, the assertion on line 135 fails, where the expected count is `2` when expected is `1`.  Would you happen to have any insight into why?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/TransactionDirectoryListingCacheKey.java
churromorales: of course!

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/TestTransactionScopeCachingDirectoryLister.java
pettyjamesm: Ok, I see what's going on here.

We're creating the cache with a maximum size limit of `DataSize.ofBytes(600)` but internally the cache distributes that maximum weight across each "segment" (segment count is driven by the target concurrency). If you don't specify a concurrency limit, the default value used internally is `4` - so we get a maximum weight per segment of `600 / 4 = 150`. This particular cache entry has a weight of 240 after your code changes, so it's immediately evicted from the cache and doesn't get reused on the second call.

Given that we already set the concurrency to 1 in `testConcurrentDirectoryListing()` I think we're good to use the same approach here.

churromorales: ahhh, great find!  Makes sense now finally.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
churromorales: done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -144,21 +145,25 @@ private static RemoteIterator<TrinoFileStatus> createListingRemoteIterator(Trino
...
             if (table.getPartitionColumns().isEmpty()) {
-                cache.invalidate(Location.of(table.getStorage().getLocation()));
+                log.debug("Invalidating (table) cache for table: %s", table.getSchemaTableName());
+                cache.invalidate(new CacheKey(Location.of(table.getStorage().getLocation()), table.getSchemaTableName()));
             }
             else {
                 // a partitioned table can have multiple paths in cache
-                cache.invalidateAll();
+                SchemaTableName tableName = table.getSchemaTableName();
+                log.debug("Invalidating (partitions) cache for table: %s", table.getSchemaTableName());
```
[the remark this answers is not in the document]
churromorales: done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
churromorales: done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/BaseCachingDirectoryListerTest.java
```diff
@@ -68,19 +68,19 @@ public void testCacheInvalidationIsAppliedSpecificallyOnTheNonPartitionedTableBe
         // The listing for the invalidate_non_partitioned_table1 should be in the directory cache after this call
         assertQuery("SELECT sum(col1) FROM partial_cache_invalidation_table1", "VALUES (6)");
         String cachedTable1Location = getTableLocation(TPCH_SCHEMA, "partial_cache_invalidation_table1");
-        assertThat(isCached(cachedTable1Location)).isTrue();
+        assertThat(isCached(cachedTable1Location, SchemaTableName.schemaTableName(TPCH_SCHEMA, "partial_cache_invalidation_table1"))).isTrue();
```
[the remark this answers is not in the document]
churromorales: done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/BaseCachingDirectoryListerTest.java
churromorales: good call, ill add verification to this test.

pettyjamesm (approved):
Changes LGTM. @Praveen2112, do you want to take another look?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/TestTransactionScopeCachingDirectoryLister.java
churromorales: yep split it into 2

plugin/trino-hive/src/test/java/io/trino/plugin/hive/fs/TestTransactionScopeCachingDirectoryLister.java
```diff
@@ -159,8 +162,8 @@ private static class CountingDirectoryLister
             implements DirectoryLister
     {
         private final Map<Location, List<TrinoFileStatus>> fileStatuses;
-        private int listCount;
-        private boolean throwException;
+        private AtomicInteger listCount;
```
pettyjamesm, line 165: Should be `final` and initialized, woops!

churromorales: oops ill fix that :(

Praveen2112 (approved):
Thanks for fixing this.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -144,21 +146,25 @@ private static RemoteIterator<TrinoFileStatus> createListingRemoteIterator(Trino
...
```
findinpath, line 166: I find it confusing that i see sometimes (partitions) and sometimes (table). Just leave them out. It is cache.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
```diff
@@ -144,21 +146,25 @@ private static RemoteIterator<TrinoFileStatus> createListingRemoteIterator(Trino
...
     {
         if (isCacheEnabledFor(table.getSchemaTableName()) && isLocationPresent(table.getStorage())) {
             if (table.getPartitionColumns().isEmpty()) {
-                cache.invalidate(Location.of(table.getStorage().getLocation()));
+                log.debug("Invalidating (table) cache for table: %s", table.getSchemaTableName());
+                cache.invalidate(new CacheKey(Location.of(table.getStorage().getLocation()), table.getSchemaTableName()));
             }
             else {
                 // a partitioned table can have multiple paths in cache
-                cache.invalidateAll();
```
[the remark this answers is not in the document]
Praveen2112: I think we have a test for the case in `testDropPartitionedTable`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
pettyjamesm: Maybe:
- `log.debug("Invalidating cache for partitioned table: %s", table.getSchemaTableName());`
- and `log.debug("Invalidating cache for unpartitioned table: %s", table.getSchemaTableName());` ?

churromorales: i can just remove logging, no worries.  would that make things easier?

churromorales:
i'll wait till everyone agrees on what path they want to go before making changes.  i don't have a strong opinion either way so happy to accommodate.  just let me know what you want done to get merged

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
findinpath: this is a  nit comment - no need to overfocus on it :)

plugin/trino-hive/src/main/java/io/trino/plugin/hive/fs/CachingDirectoryLister.java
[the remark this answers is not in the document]
churromorales: we do have a test, thank you @Praveen2112 for suggesting to add this in the PR :)

pettyjamesm:
Merged, thanks @churromorales!

PR #24851, opened 2025-01-30
Set write compression codec in Iceberg

SemionPar:
#### Before this change

Trino does not set `write.<fileType>.compression-codec` table property value when creating new tables. As effect of that, all new tables created with Trino would have `"write.parquet.compression-codec"="zstd"` (Iceberg lib is setting this default):

```bash
trino> set session iceberg.compression_codec = 'SNAPPY';
SET SESSION
trino> create table iceberg.default.le_table(data integer) with (format='parquet');
CREATE TABLE
trino> select * from iceberg.default."le_table$properties";
               key               |  value  
---------------------------------+---------
 write.format.default            | PARQUET 
 write.parquet.compression-codec | zstd    
 commit.retry.num-retries        | 4       
(3 rows)
```

#### With this change

When a new Iceberg table is created with Trino, it will use session write compression codec to set `write.<fileType>.compression-codec` value in Iceberg table properties. Compression codec values are set as spelled by the spec - details in https://github.com/trinodb/trino/pull/24851#issuecomment-2674059110

```bash
trino> set session iceberg.compression_codec = 'SNAPPY';
SET SESSION
trino> create table iceberg.default.le_table(data integer) with (format='parquet');
CREATE TABLE
trino> select * from iceberg.default."le_table$properties";
               key               |  value  
---------------------------------+---------
 write.format.default            | PARQUET 
 write.parquet.compression-codec | snappy    
 commit.retry.num-retries        | 4       
(3 rows)
```

This change is only about table metadata - it does not have an effect on how Trino reads or writes data.

## Additional context and related issues
[Iceberg is setting "write.parquet.compression-codec"="zstd" by default](https://github.com/apache/iceberg/pull/8593/files#diff-c540a31e66b157a8f080433c82a29a070096d0e08c6578a0099153f1229bdb7aR93-R96) when no `write.parquet.compression-codec` is provided by the engine, even if _fileType != parquet_: https://github.com/apache/iceberg/issues/9490

## Release notes

```markdown
## Iceberg
* Set `write.<filetype>.compression-codec` table property when creating new tables
```

PR #25755, opened 2025-05-08
Support setting compression_codec table property for Iceberg

SongChujun:
This PR makes compression_codec a table property and drops support for compression_codec iceberg session property for simplicity.

Users are able to run the following command now to specify the compression_codec to create a new table with compression_codec set to ZSTD.
```
CREATE TABLE example (
    c1 INTEGER,
    c2 DATE,
    c3 DOUBLE
)
WITH (
    format = 'AVRO',
    partitioning = ARRAY['c1', 'c2'],
    sorted_by = ARRAY['c3'],
    compression_codec = 'ZSTD'
);
```
Users are able able to change the compression_codec via statement ```Alter Table Set Properties```.

Example
```
ALTER TABLE example SET PROPERTIES compression_codec = 'GZIP';
```

The write_compression users specify will replace the session variable ```iceberg.compression-codec```. If the user change file_format without changing write_compression, it will inherit the write_compression set for the original file format, if the write_compression for the previous file format is set.

If the user is trying to set a write_compression that is inconsistent with the file format, the system will throw an exception.

The compatibility matrix for write_compression and file format is the following
|  |  Orc  | Parquet |  Avro  |
|:-----------------:|:-----:|:-------:|:------:|
| **NONE**          |   ✅   |    ✅   |    ✅   |
| **SNAPPY**        |   ✅   |    ✅   |    ✅   |
| **LZ4**           |   ✅   |    ❌   |    ❌   |
| **ZSTD**          |   ✅   |    ✅   |    ✅   |
| **GZIP**          |   ✅   |    ✅   |    ✅   |

## Release notes

```markdown
## Iceberg connector
* Make `compression_codec` a table property and drop support for `compression_codec` session property. ({issue}`25755`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -973,8 +973,13 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
   - Description
 * - `format`
   - Optionally specifies the format of table data files; either `PARQUET`,
-    `ORC`, or `AVRO`. Defaults to the value of the `iceberg.file-format` catalog
-    configuration property, which defaults to `PARQUET`.
+    `ORC`, or `AVRO`. Defaults to the value of the `iceberg.file-format` 
+    catalog configuration property, which defaults to `PARQUET`.
+* - `compression_codec`
+  - Optionally specifies the compression-codec used for writing the table; 
+    either `NONE`, `ZSTD`, `SNAPPY`, `LZ4`, or `GZIP`. Defaults to the value 
+    of the `iceberg.compression-codec` catalog configuration property, which 
+    defaults to `ZSTD`.
 * - `partitioning`
   - Optionally specifies table partitioning. If a table is partitioned by
     columns `c1` and `c2`, the partitioning property is `partitioning =
@@ -1032,9 +1037,9 @@ WITH (
     location = '/var/example_tables/test_table');
 ```
 
-The table definition below specifies to use ORC files, bloom filter index by columns
-`c1` and `c2`, fpp is 0.05, and a file system location of
-`/var/example_tables/test_table`:
+The table definition below specifies to use ORC files with compression_codec
+SNAPPY, bloom filter index by columns `c1` and `c2`, fpp is 0.05, and a file
+system location of `/var/example_tables/test_table`:
 
 ```sql
 CREATE TABLE test_table (
@@ -1043,6 +1048,7 @@ CREATE TABLE test_table (
     c3 DOUBLE)
 WITH (
     format = 'ORC',
+    compression_codec = 'SNAPPY',
     location = '/var/example_tables/test_table',
     orc_bloom_filter_columns = ARRAY['c1', 'c2'],
     orc_bloom_filter_fpp = 0.05);
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergAvroFileWriter.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergAvroFileWriter.java
@@ -35,6 +35,8 @@ import static io.airlift.slice.SizeOf.instanceSize;
 import static io.trino.plugin.iceberg.IcebergAvroDataConversion.toIcebergRecords;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_WRITER_CLOSE_ERROR;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_WRITER_OPEN_ERROR;
+import static io.trino.plugin.iceberg.IcebergFileFormat.AVRO;
+import static io.trino.plugin.iceberg.IcebergTableProperties.validateCompression;
 import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static java.util.Objects.requireNonNull;
 import static org.apache.iceberg.TableProperties.AVRO_COMPRESSION;
@@ -62,6 +64,8 @@ public final class IcebergAvroFileWriter
         this.rollbackAction = requireNonNull(rollbackAction, "rollbackAction null");
         this.icebergSchema = requireNonNull(icebergSchema, "icebergSchema is null");
         this.types = ImmutableList.copyOf(requireNonNull(types, "types is null"));
+
+        validateCompression(AVRO, Optional.of(hiveCompressionCodec));
 
         try {
             avroWriter = Avro.write(file)
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
@@ -31,6 +31,7 @@ import io.trino.orc.OutputStreamOrcDataSink;
 import io.trino.parquet.writer.ParquetWriterOptions;
 import io.trino.plugin.base.metrics.FileFormatDataSourceStats;
 import io.trino.plugin.hive.HiveCompressionCodec;
+import io.trino.plugin.hive.HiveCompressionOption;
 import io.trino.plugin.hive.NodeVersion;
 import io.trino.plugin.hive.orc.OrcWriterConfig;
 import io.trino.plugin.iceberg.fileio.ForwardingOutputFile;
@@ -60,7 +61,9 @@ import static io.trino.plugin.hive.HiveMetadata.TRINO_VERSION_NAME;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_INVALID_METADATA;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_WRITER_OPEN_ERROR;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_WRITE_VALIDATION_FAILED;
-import static io.trino.plugin.iceberg.IcebergSessionProperties.getCompressionCodec;
+import static io.trino.plugin.iceberg.IcebergFileFormat.AVRO;
+import static io.trino.plugin.iceberg.IcebergFileFormat.ORC;
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.getOrcStringStatisticsLimit;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.getOrcWriterMaxDictionaryMemory;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.getOrcWriterMaxRowGroupRows;
@@ -74,6 +77,7 @@ import static io.trino.plugin.iceberg.IcebergSessionProperties.getParquetWriterP
 import static io.trino.plugin.iceberg.IcebergSessionProperties.getParquetWriterPageValueCount;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.isOrcWriterValidate;
 import static io.trino.plugin.iceberg.IcebergTableProperties.ORC_BLOOM_FILTER_FPP_PROPERTY;
+import static io.trino.plugin.iceberg.IcebergUtil.getHiveCompressionCodec;
 import static io.trino.plugin.iceberg.IcebergUtil.getOrcBloomFilterColumns;
 import static io.trino.plugin.iceberg.IcebergUtil.getOrcBloomFilterFpp;
 import static io.trino.plugin.iceberg.IcebergUtil.getParquetBloomFilterColumns;
@@ -97,6 +101,7 @@ public class IcebergFileWriterFactory
     private final NodeVersion nodeVersion;
     private final FileFormatDataSourceStats readStats;
     private final OrcWriterStats orcWriterStats = new OrcWriterStats();
+    private final HiveCompressionOption hiveCompressionOption;
     private final OrcWriterOptions orcWriterOptions;
 
     @Inject
@@ -104,12 +109,14 @@ public class IcebergFileWriterFactory
             TypeManager typeManager,
             NodeVersion nodeVersion,
             FileFormatDataSourceStats readStats,
+            IcebergConfig icebergConfig,
             OrcWriterConfig orcWriterConfig)
     {
         checkArgument(!orcWriterConfig.isUseLegacyVersion(), "the ORC writer shouldn't be configured to use a legacy version");
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.nodeVersion = requireNonNull(nodeVersion, "nodeVersion is null");
         this.readStats = requireNonNull(readStats, "readStats is null");
+        this.hiveCompressionOption = icebergConfig.getCompressionCodec();
         this.orcWriterOptions = orcWriterConfig.toOrcWriterOptions();
     }
 
@@ -132,7 +139,7 @@ public class IcebergFileWriterFactory
             // TODO use metricsConfig https://github.com/trinodb/trino/issues/9791
             case PARQUET -> createParquetWriter(MetricsConfig.getDefault(), fileSystem, outputPath, icebergSchema, session, storageProperties);
             case ORC -> createOrcWriter(metricsConfig, fileSystem, outputPath, icebergSchema, session, storageProperties, getOrcStringStatisticsLimit(session));
-            case AVRO -> createAvroWriter(fileSystem, outputPath, icebergSchema, session);
+            case AVRO -> createAvroWriter(fileSystem, outputPath, icebergSchema, storageProperties);
         };
     }
 
@@ -146,7 +153,7 @@ public class IcebergFileWriterFactory
         return switch (fileFormat) {
             case PARQUET -> createParquetWriter(FULL_METRICS_CONFIG, fileSystem, outputPath, POSITION_DELETE_SCHEMA, session, storageProperties);
             case ORC -> createOrcWriter(FULL_METRICS_CONFIG, fileSystem, outputPath, POSITION_DELETE_SCHEMA, session, storageProperties, DataSize.ofBytes(Integer.MAX_VALUE));
-            case AVRO -> createAvroWriter(fileSystem, outputPath, POSITION_DELETE_SCHEMA, session);
+            case AVRO -> createAvroWriter(fileSystem, outputPath, POSITION_DELETE_SCHEMA, storageProperties);
         };
     }
 
@@ -178,7 +185,9 @@ public class IcebergFileWriterFactory
                     .setBloomFilterColumns(getParquetBloomFilterColumns(storageProperties))
                     .build();
 
-            HiveCompressionCodec hiveCompressionCodec = toCompressionCodec(getCompressionCodec(session));
+            HiveCompressionCodec compressionCodec = getHiveCompressionCodec(PARQUET, storageProperties)
+                    .orElse(toCompressionCodec(hiveCompressionOption));
+
             return new IcebergParquetFileWriter(
                     metricsConfig,
                     outputFile,
@@ -189,8 +198,8 @@ public class IcebergFileWriterFactory
                     makeTypeMap(fileColumnTypes, fileColumnNames),
                     parquetWriterOptions,
                     IntStream.range(0, fileColumnNames.size()).toArray(),
-                    hiveCompressionCodec.getParquetCompressionCodec()
-                            .orElseThrow(() -> new TrinoException(NOT_SUPPORTED, "Compression codec %s not supported for Parquet".formatted(hiveCompressionCodec))),
+                    compressionCodec.getParquetCompressionCodec()
+                            .orElseThrow(() -> new TrinoException(NOT_SUPPORTED, "Compression codec %s not supported for Parquet".formatted(compressionCodec))),
                     nodeVersion.toString());
         }
         catch (IOException | UncheckedIOException e) {
@@ -234,6 +243,9 @@ public class IcebergFileWriterFactory
                 });
             }
 
+            HiveCompressionCodec compressionCodec = getHiveCompressionCodec(ORC, storageProperties)
+                    .orElse(toCompressionCodec(hiveCompressionOption));
+
             return new IcebergOrcFileWriter(
                     metricsConfig,
                     icebergSchema,
@@ -242,7 +254,7 @@ public class IcebergFileWriterFactory
                     fileColumnNames,
                     fileColumnTypes,
                     toOrcType(icebergSchema),
-                    toCompressionCodec(getCompressionCodec(session)).getOrcCompressionKind(),
+                    compressionCodec.getOrcCompressionKind(),
                     withBloomFilterOptions(orcWriterOptions, storageProperties)
                             .withStripeMinSize(getOrcWriterMinStripeSize(session))
                             .withStripeMaxSize(getOrcWriterMaxStripeSize(session))
@@ -287,7 +299,7 @@ public class IcebergFileWriterFactory
             TrinoFileSystem fileSystem,
             Location outputPath,
             Schema icebergSchema,
-            ConnectorSession session)
+            Map<String, String> storageProperties)
     {
         Closeable rollbackAction = () -> fileSystem.deleteFile(outputPath);
 
@@ -295,11 +307,14 @@ public class IcebergFileWriterFactory
                 .map(column -> toTrinoType(column.type(), typeManager))
                 .collect(toImmutableList());
 
+        HiveCompressionCodec compressionCodec = getHiveCompressionCodec(AVRO, storageProperties)
+                .orElse(toCompressionCodec(hiveCompressionOption));
+
         return new IcebergAvroFileWriter(
                 new ForwardingOutputFile(fileSystem, outputPath),
                 rollbackAction,
                 icebergSchema,
                 columnTypes,
-                toCompressionCodec(getCompressionCodec(session)));
+                compressionCodec);
     }
 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -42,6 +42,7 @@ import io.trino.plugin.base.classloader.ClassLoaderSafeSystemTable;
 import io.trino.plugin.base.filter.UtcConstraintExtractor;
 import io.trino.plugin.base.projection.ApplyProjectionUtil;
 import io.trino.plugin.base.projection.ApplyProjectionUtil.ProjectedColumnRepresentation;
+import io.trino.plugin.hive.HiveCompressionCodec;
 import io.trino.plugin.hive.HiveStorageFormat;
 import io.trino.plugin.hive.HiveWrittenPartitions;
 import io.trino.plugin.iceberg.aggregation.DataSketchStateSerializer;
@@ -270,6 +271,8 @@ import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_FILESYSTEM_ERROR;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_INVALID_METADATA;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_MISSING_METADATA;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_UNSUPPORTED_VIEW_DIALECT;
+import static io.trino.plugin.iceberg.IcebergFileFormat.ORC;
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
 import static io.trino.plugin.iceberg.IcebergMetadataColumn.FILE_MODIFIED_TIME;
 import static io.trino.plugin.iceberg.IcebergMetadataColumn.FILE_PATH;
 import static io.trino.plugin.iceberg.IcebergMetadataColumn.PARTITION;
@@ -292,6 +295,7 @@ import static io.trino.plugin.iceberg.IcebergTableName.isDataTable;
 import static io.trino.plugin.iceberg.IcebergTableName.isIcebergTableName;
 import static io.trino.plugin.iceberg.IcebergTableName.isMaterializedViewStorage;
 import static io.trino.plugin.iceberg.IcebergTableName.tableNameFrom;
+import static io.trino.plugin.iceberg.IcebergTableProperties.COMPRESSION_CODEC;
 import static io.trino.plugin.iceberg.IcebergTableProperties.DATA_LOCATION_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergTableProperties.EXTRA_PROPERTIES_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergTableProperties.FILE_FORMAT_PROPERTY;
@@ -304,6 +308,7 @@ import static io.trino.plugin.iceberg.IcebergTableProperties.PARTITIONING_PROPER
 import static io.trino.plugin.iceberg.IcebergTableProperties.SORTED_BY_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergTableProperties.getPartitioning;
 import static io.trino.plugin.iceberg.IcebergTableProperties.getTableLocation;
+import static io.trino.plugin.iceberg.IcebergTableProperties.validateCompression;
 import static io.trino.plugin.iceberg.IcebergUtil.buildPath;
 import static io.trino.plugin.iceberg.IcebergUtil.canEnforceColumnConstraintInSpecs;
 import static io.trino.plugin.iceberg.IcebergUtil.checkFormatForProperty;
@@ -315,7 +320,9 @@ import static io.trino.plugin.iceberg.IcebergUtil.firstSnapshot;
 import static io.trino.plugin.iceberg.IcebergUtil.firstSnapshotAfter;
 import static io.trino.plugin.iceberg.IcebergUtil.getColumnHandle;
 import static io.trino.plugin.iceberg.IcebergUtil.getColumnMetadatas;
+import static io.trino.plugin.iceberg.IcebergUtil.getCompressionPropertyName;
 import static io.trino.plugin.iceberg.IcebergUtil.getFileFormat;
+import static io.trino.plugin.iceberg.IcebergUtil.getHiveCompressionCodec;
 import static io.trino.plugin.iceberg.IcebergUtil.getIcebergTableProperties;
 import static io.trino.plugin.iceberg.IcebergUtil.getPartitionKeys;
 import static io.trino.plugin.iceberg.IcebergUtil.getPartitionValues;
@@ -422,6 +429,7 @@ public class IcebergMetadata
             .add(EXTRA_PROPERTIES_PROPERTY)
             .add(FILE_FORMAT_PROPERTY)
             .add(FORMAT_VERSION_PROPERTY)
+            .add(COMPRESSION_CODEC)
             .add(MAX_COMMIT_RETRY)
             .add(OBJECT_STORE_LAYOUT_ENABLED_PROPERTY)
             .add(DATA_LOCATION_PROPERTY)
@@ -2474,10 +2482,13 @@ public class IcebergMetadata
             }
         }
 
+        IcebergFileFormat oldFileFormat = getFileFormat(icebergTable.properties());
+        IcebergFileFormat newFileFormat = oldFileFormat;
+
         if (properties.containsKey(FILE_FORMAT_PROPERTY)) {
-            IcebergFileFormat fileFormat = (IcebergFileFormat) properties.get(FILE_FORMAT_PROPERTY)
+            newFileFormat = (IcebergFileFormat) properties.get(FILE_FORMAT_PROPERTY)
                     .orElseThrow(() -> new IllegalArgumentException("The format property cannot be empty"));
-            updateProperties.defaultFormat(fileFormat.toIceberg());
+            updateProperties.defaultFormat(newFileFormat.toIceberg());
         }
 
         if (properties.containsKey(FORMAT_VERSION_PROPERTY)) {
@@ -2486,6 +2497,14 @@ public class IcebergMetadata
                     .orElseThrow(() -> new IllegalArgumentException("The format_version property cannot be empty"));
             updateProperties.set(FORMAT_VERSION, Integer.toString(formatVersion));
         }
+
+        Map<String, String> propertiesForCompression = calculateTableCompressionProperties(oldFileFormat, newFileFormat, icebergTable.properties(), properties.entrySet().stream()
+                .filter(e -> e.getValue().isPresent())
+                .collect(toImmutableMap(
+                        Map.Entry::getKey,
+                        e -> e.getValue().get())));
+
+        propertiesForCompression.forEach(updateProperties::set);
 
         if (properties.containsKey(MAX_COMMIT_RETRY)) {
             int maxCommitRetry = (int) properties.get(MAX_COMMIT_RETRY)
@@ -2540,6 +2559,22 @@ public class IcebergMetadata
         }
 
         commitTransaction(transaction, "set table properties");
+    }
+
+    public static Map<String, String> calculateTableCompressionProperties(IcebergFileFormat oldFileFormat, IcebergFileFormat newFileFormat, Map<String, String> existingProperties, Map<String, Object> inputProperties)
+    {
+        ImmutableMap.Builder<String, String> newCompressionProperties = ImmutableMap.builder();
+
+        Optional<HiveCompressionCodec> oldCompressionCodec = getHiveCompressionCodec(oldFileFormat, existingProperties);
+        Optional<HiveCompressionCodec> newCompressionCodec = IcebergTableProperties.getCompressionCodec(inputProperties);
+
+        Optional<HiveCompressionCodec> compressionCodec = newCompressionCodec.or(() -> oldCompressionCodec);
+
+        validateCompression(newFileFormat, compressionCodec);
+
+        compressionCodec.ifPresent(hiveCompressionCodec -> newCompressionProperties.put(getCompressionPropertyName(newFileFormat), hiveCompressionCodec.name()));
+
+        return newCompressionProperties.buildOrThrow();
     }
 
     private static void updatePartitioning(Table icebergTable, Transaction transaction, List<String> partitionColumns)
@@ -4013,7 +4048,7 @@ public class IcebergMetadata
         IcebergTableHandle tableHandle = (IcebergTableHandle) connectorTableHandle;
         IcebergFileFormat storageFormat = getFileFormat(tableHandle.getStorageProperties());
 
-        return storageFormat == IcebergFileFormat.ORC || storageFormat == IcebergFileFormat.PARQUET;
+        return storageFormat == ORC || storageFormat == PARQUET;
     }
 
     @Override
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
@@ -13,9 +13,13 @@
  */
 package io.trino.plugin.iceberg;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
+import io.trino.plugin.hive.HiveCompressionCodec;
+import io.trino.plugin.hive.HiveCompressionCodecs;
+import io.trino.plugin.hive.HiveCompressionOption;
 import io.trino.plugin.hive.orc.OrcWriterConfig;
 import io.trino.spi.TrinoException;
 import io.trino.spi.session.PropertyMetadata;
@@ -33,7 +37,10 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static io.trino.plugin.iceberg.IcebergConfig.FORMAT_VERSION_SUPPORT_MAX;
 import static io.trino.plugin.iceberg.IcebergConfig.FORMAT_VERSION_SUPPORT_MIN;
+import static io.trino.plugin.iceberg.IcebergFileFormat.AVRO;
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
 import static io.trino.spi.StandardErrorCode.INVALID_TABLE_PROPERTY;
+import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static io.trino.spi.session.PropertyMetadata.booleanProperty;
 import static io.trino.spi.session.PropertyMetadata.doubleProperty;
 import static io.trino.spi.session.PropertyMetadata.enumProperty;
@@ -56,6 +63,7 @@ public class IcebergTableProperties
     public static final String SORTED_BY_PROPERTY = "sorted_by";
     public static final String LOCATION_PROPERTY = "location";
     public static final String FORMAT_VERSION_PROPERTY = "format_version";
+    public static final String COMPRESSION_CODEC = "compression_codec";
     public static final String MAX_COMMIT_RETRY = "max_commit_retry";
     public static final String ORC_BLOOM_FILTER_COLUMNS_PROPERTY = "orc_bloom_filter_columns";
     public static final String ORC_BLOOM_FILTER_FPP_PROPERTY = "orc_bloom_filter_fpp";
@@ -66,6 +74,7 @@ public class IcebergTableProperties
 
     public static final Set<String> SUPPORTED_PROPERTIES = ImmutableSet.<String>builder()
             .add(FILE_FORMAT_PROPERTY)
+            .add(COMPRESSION_CODEC)
             .add(PARTITIONING_PROPERTY)
             .add(SORTED_BY_PROPERTY)
             .add(LOCATION_PROPERTY)
@@ -102,6 +111,12 @@ public class IcebergTableProperties
                         "File format for the table",
                         IcebergFileFormat.class,
                         icebergConfig.getFileFormat(),
+                        false))
+                .add(enumProperty(
+                        COMPRESSION_CODEC,
+                        "Write compression codec for the table",
+                        HiveCompressionOption.class,
+                        null,
                         false))
                 .add(new PropertyMetadata<>(
                         PARTITIONING_PROPERTY,
@@ -220,6 +235,12 @@ public class IcebergTableProperties
         return (IcebergFileFormat) tableProperties.get(FILE_FORMAT_PROPERTY);
     }
 
+    public static Optional<HiveCompressionCodec> getCompressionCodec(Map<String, Object> inputProperties)
+    {
+        return Optional.ofNullable((HiveCompressionOption) inputProperties.get(COMPRESSION_CODEC))
+                .map(HiveCompressionCodecs::toCompressionCodec);
+    }
+
     @SuppressWarnings("unchecked")
     public static List<String> getPartitioning(Map<String, Object> tableProperties)
     {
@@ -250,6 +271,24 @@ public class IcebergTableProperties
             throw new TrinoException(INVALID_TABLE_PROPERTY,
                     format("format_version must be between %d and %d", FORMAT_VERSION_SUPPORT_MIN, FORMAT_VERSION_SUPPORT_MAX));
         }
+    }
+
+    public static void validateCompression(IcebergFileFormat fileFormat, Optional<HiveCompressionCodec> compressionCodec)
+    {
+        if (compressionCodec.isPresent()) {
+            if (!isCompressionCodecSupportedForFormat(fileFormat, compressionCodec.get())) {
+                throw new TrinoException(NOT_SUPPORTED, format("Compression codec %s not supported for %s", compressionCodec.get(), fileFormat.humanName()));
+            }
+        }
+    }
+
+    @VisibleForTesting
+    static boolean isCompressionCodecSupportedForFormat(IcebergFileFormat fileFormat, HiveCompressionCodec codec)
+    {
+        return switch (codec) {
+            case LZ4 -> !(fileFormat == AVRO || fileFormat == PARQUET);
+            default -> true;
+        };
     }
 
     public static int getMaxCommitRetry(Map<String, Object> tableProperties)
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
@@ -28,6 +28,7 @@ import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystem;
 import io.trino.filesystem.TrinoInputFile;
 import io.trino.plugin.hive.HiveCompressionCodec;
+import io.trino.plugin.hive.HiveCompressionOption;
 import io.trino.plugin.iceberg.PartitionTransforms.ColumnTransform;
 import io.trino.plugin.iceberg.catalog.IcebergTableOperations;
 import io.trino.plugin.iceberg.catalog.IcebergTableOperationsProvider;
@@ -88,7 +89,6 @@ import java.time.Instant;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.List;
-import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
@@ -112,7 +112,6 @@ import static io.airlift.slice.Slices.utf8Slice;
 import static io.trino.metastore.Table.TABLE_COMMENT;
 import static io.trino.parquet.writer.ParquetWriter.SUPPORTED_BLOOM_FILTER_TYPES;
 import static io.trino.plugin.base.io.ByteBuffers.getWrappedBytes;
-import static io.trino.plugin.hive.HiveCompressionCodecs.toCompressionCodec;
 import static io.trino.plugin.iceberg.ColumnIdentity.createColumnIdentity;
 import static io.trino.plugin.iceberg.IcebergColumnHandle.fileModifiedTimeColumnHandle;
 import static io.trino.plugin.iceberg.IcebergColumnHandle.fileModifiedTimeColumnMetadata;
@@ -123,7 +122,9 @@ import static io.trino.plugin.iceberg.IcebergColumnHandle.pathColumnMetadata;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_FILESYSTEM_ERROR;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_INVALID_METADATA;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_INVALID_PARTITION_VALUE;
-import static io.trino.plugin.iceberg.IcebergSessionProperties.getCompressionCodec;
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
+import static io.trino.plugin.iceberg.IcebergMetadata.calculateTableCompressionProperties;
+import static io.trino.plugin.iceberg.IcebergTableProperties.COMPRESSION_CODEC;
 import static io.trino.plugin.iceberg.IcebergTableProperties.DATA_LOCATION_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergTableProperties.FILE_FORMAT_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergTableProperties.FORMAT_VERSION_PROPERTY;
@@ -139,6 +140,7 @@ import static io.trino.plugin.iceberg.IcebergTableProperties.SORTED_BY_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergTableProperties.SUPPORTED_PROPERTIES;
 import static io.trino.plugin.iceberg.IcebergTableProperties.getPartitioning;
 import static io.trino.plugin.iceberg.IcebergTableProperties.getSortOrder;
+import static io.trino.plugin.iceberg.IcebergTableProperties.validateCompression;
 import static io.trino.plugin.iceberg.PartitionFields.parsePartitionFields;
 import static io.trino.plugin.iceberg.PartitionFields.toPartitionFields;
 import static io.trino.plugin.iceberg.SortFieldUtils.parseSortFields;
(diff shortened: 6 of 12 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2531,6 +2554,29 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
...
     }
 
+    public static Map<String, String> calculateTableCompressionProperties(IcebergFileFormat oldFileFormat, IcebergFileFormat newFileFormat, Map<String, String> existingProperties, Map<String, Object> inputProperties)
+    {
+        ImmutableMap.Builder<String, String> newCompressionProperties = ImmutableMap.builder();
+
+        Optional<HiveCompressionCodec> oldCompressionCodec = getHiveCompressionCodec(oldFileFormat, existingProperties);
+        Optional<HiveCompressionCodec> newCompressionCodec = IcebergTableProperties.getHiveCompressionCodec(inputProperties);
+
+        Optional<HiveCompressionCodec> compressionCodec = newCompressionCodec.or(() -> oldCompressionCodec);
```
pettyjamesm, line 2564: Nit: no need for a lambda here, just `newCompressionCodec.orElse(oldCompressionCodec)`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -220,6 +252,19 @@ public static IcebergFileFormat getFileFormat(Map<String, Object> tablePropertie
...
+    {
+        return Optional.ofNullable((HiveCompressionOption) inputProperties.get(WRITE_COMPRESSION))
+                .map(HiveCompressionCodecs::toCompressionCodec);
+    }
+
+    public static OptionalInt getCompressionLevel(Map<String, Object> tableProperties)
+    {
+        return Optional.ofNullable((Integer) tableProperties.get(COMPRESSION_LEVEL))
+                .map(OptionalInt::of)
+                .orElseGet(OptionalInt::empty);
```
pettyjamesm, line 265: Nit: no need for a lambda here, just `orElse(OptionalInt.empty())`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -252,6 +297,48 @@ private static void validateFormatVersion(int version)
         }
     }
 
+    public static void validateCompression(IcebergFileFormat fileFormat, Optional<HiveCompressionCodec> compressionCodec, OptionalInt compressionLevel)
+    {
+        if (compressionCodec.isPresent()) {
+            if (!isCompressionCodecSupportedForFormat(fileFormat, compressionCodec.get())) {
+                throw new TrinoException(NOT_SUPPORTED, format("Compression codec LZ4 not supported for %s", fileFormat.humanName()));
```
pettyjamesm, line 304: Error message here mentions `LZ4` explicitly, but it seems like this should be a parameter based on `compressionCodec.get()`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -252,6 +297,48 @@ private static void validateFormatVersion(int version)
...
+            }
+        }
+        if (compressionLevel.isPresent()) {
+            if (compressionCodec.isEmpty()) {
+                throw new TrinoException(INVALID_TABLE_PROPERTY, "write_compression must be set when compression_level is set");
+            }
+            else {
+                if (!(VALID_ICEBERG_FILE_FORMATS_FOR_COMPRESSION_LEVEL_PROPERTY.contains(fileFormat)
```
pettyjamesm, line 312: I wonder whether this branch would actually be easier to read and understand as an enum switch statement. That way we would fail to compile if a new enum were added and we would know we needed to add new handling at that point.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -252,6 +297,48 @@ private static void validateFormatVersion(int version)
...
+                else {
+                    throw new IllegalStateException("Unsupported compression codec: " + compressionCodec);
+                }
+            }
+        }
+    }
+
+    public static boolean isCompressionCodecSupportedForFormat(IcebergFileFormat fileFormat, HiveCompressionCodec codec)
+    {
+        return !((fileFormat.equals(AVRO) || fileFormat.equals(PARQUET)) && codec.equals(HiveCompressionCodec.LZ4));
```
pettyjamesm, line 339: Let's rewrite this as a switch over the file formats and compression codecs so that it's easier to read and understand the valid combinations.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -310,11 +316,20 @@ public static List<Integer> buildPath(Map<Integer, Integer> indexParents, int fi
     public static Map<String, Object> getIcebergTableProperties(BaseTable icebergTable)
     {
         ImmutableMap.Builder<String, Object> properties = ImmutableMap.builder();
+        IcebergFileFormat fileFormat = getFileFormat(icebergTable);
         properties.put(FILE_FORMAT_PROPERTY, getFileFormat(icebergTable));
```
pettyjamesm, line 320: duplicate call to `getFileFormat`, you can pass the instance from the previous line here instead

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -975,6 +989,57 @@ public static void verifyExtraProperties(Set<String> basePropertyKeys, Map<Strin
...
     }
 
+    public static OptionalInt getCompressionLevel(IcebergFileFormat fileFormat, Map<String, String> storageProperties)
+    {
+        if (!VALID_ICEBERG_FILE_FORMATS_FOR_COMPRESSION_LEVEL_PROPERTY.contains(fileFormat)) {
+            return OptionalInt.empty();
+        }
+        String compressionLevelProperty = getCompressionLevelName(fileFormat);
+        String compressionLevelValue = storageProperties.get(compressionLevelProperty);
+        if (!Strings.isNullOrEmpty(compressionLevelValue)) {
```
pettyjamesm, line 999: I would just inline `compressionLevelValue != null && !compressionLevelValue.isEmpty()` instead of bringing in `Strings` to do that

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -471,19 +486,19 @@ public static Map<PartitionField, Integer> getIdentityPartitions(PartitionSpec p
         return columns.buildOrThrow();
     }
 
-    public static List<Types.NestedField> primitiveFields(Schema schema)
+    public static List<NestedField> primitiveFields(Schema schema)
```
pettyjamesm, line 489: Let's move these cleanup / refactoring changes to a separate commit

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +878,46 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
+            txn = catalog.newCreateOrReplaceTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, tableLocation, createTableProperties(tableMetadata, allowedExtraProperties));
+        }
+        else {
+            txn = catalog.newCreateTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, Optional.ofNullable(tableLocation), createTableProperties(tableMetadata, allowedExtraProperties));
         }
-        return catalog.newCreateTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, Optional.ofNullable(tableLocation), createTableProperties(session, tableMetadata, allowedExtraProperties));
+
+        String parquetCompressionValue = txn.table().properties().get(PARQUET_COMPRESSION);
+        if (parquetCompressionValue != null && parquetCompressionValue.isEmpty()) {
+            txn.updateProperties()
```
pettyjamesm, line 892: Is this related to the iceberg parquet compression handling bug? If so, we'll want an inline comment explaining what's going on here

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +878,46 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
+
+        validateCompression(fileFormat, compressionCodec, compressionLevel);
+
+        Map<String, String> tableCompressionProperties = calculateTableCompressionProperties(fileFormat, fileFormat, ImmutableMap.of(), tableMetadata.getProperties());
+
+        tableCompressionProperties.forEach(propertiesBuilder::put);
+
+        // Iceberg will set write.parquet.compression-codec to zstd by default if this property is not set: https://github.com/trinodb/trino/issues/20401,
+        // but we don't want to set this property if this is not explicitly set by customer via set table properties.
+        if (!(fileFormat == IcebergFileFormat.PARQUET && compressionCodec.isPresent())) {
```
pettyjamesm, line 919: We'll definitely want a test around this quirk

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1281,6 +1292,7 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
                     .orElseGet(() -> catalog.defaultTableLocation(session, tableMetadata.getTable()));
         }
         transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties);
+
```
pettyjamesm, line 1295: Unintentional new line insertion?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6802,6 +6813,414 @@ public void testUpdatingFileFormat()
...
+        String tableName = "test_updating_file_format_compression_mixed" + randomNameSuffix();
+
+        assertUpdate(format("CREATE TABLE %s WITH (format = 'AVRO') AS SELECT * FROM nation", tableName), "SELECT count(*) FROM nation");
+
+        assertUpdate(format("ALTER TABLE %s SET PROPERTIES write_compression = 'ZSTD'", tableName));
+        assertQuery(format("SELECT value FROM \"%s$properties\" WHERE key = 'write.format.default'", tableName), format("VALUES '%s'", AVRO));
+        assertQuery(format("SELECT value FROM \"%s$properties\" WHERE key = 'write.avro.compression-codec'", tableName), "VALUES 'ZSTD'");
+
+        assertUpdate(format("ALTER TABLE %s SET PROPERTIES compression_level = 5", tableName));
+        assertQuery(format("SELECT value FROM \"%s$properties\" WHERE key = 'write.avro.compression-level'", tableName), ("VALUES 5"));
```
pettyjamesm, line 7217: Looks like checkstyle is complaining about this line, but giving the wrong line number in the error.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
SongChujun: orElse requires the parameter to be T and return T, but we are processing on optional<T> and want to return optional<T>

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: it should be covered by the test implicitly, but I still added a test on it

SongChujun:
```
assertUpdate(format("CREATE TABLE %s WITH (format = 'AVRO') AS SELECT * FROM nation", tableName), "SELECT count(*) FROM nation");

assertQuery(format("SELECT COUNT(*) FROM \"%s$properties\" WHERE key = 'write.parquet.compression-codec'", tableName), "SELECT 0");
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
pettyjamesm: Ah, I missed that- still, you want to avoid spinning a lambda for a cheap operation like this. Maybe a cleaner way to write it is: `newCompressionCodec.isPresent() ? newCompressionCodec : oldCompressionCodec;` like we do later on for compression level.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +878,48 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
+        Optional<HiveCompressionCodec> compressionCodec = IcebergTableProperties.getHiveCompressionCodec(tableMetadata.getProperties());
+        OptionalInt compressionLevel = IcebergTableProperties.getCompressionLevel(tableMetadata.getProperties());
+
+        validateCompression(fileFormat, compressionCodec, compressionLevel);
+
+        Map<String, String> tableCompressionProperties = calculateTableCompressionProperties(fileFormat, fileFormat, ImmutableMap.of(), tableMetadata.getProperties());
+
+        tableCompressionProperties.forEach(propertiesBuilder::put);
+
+        // Iceberg will set write.parquet.compression-codec to zstd by default if this property is not set: https://github.com/trinodb/trino/issues/20401,
```
raunaqmorarka, line 918: The issue you've linked to is closed, we need to either re-open that (with explanation why) or file a new issue and refer to that.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +878,48 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
+        // but we don't want to set this property if this is not explicitly set by customer via set table properties.
```
raunaqmorarka, line 919: Why ? If user has not set table property, the catalog config property for compression codec still applies

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
raunaqmorarka: Could you clarify what is the bug being referring to here ? I responded at https://github.com/trinodb/trino/pull/24851#issuecomment-2893138245 to the concern raised there.

SongChujun: If we let Iceberg write the table property ```write.parquet.compression-codec``` to zstd by default for us, then this PR will introduce some non-backward-compatible behavior. Suppose we have a user that didn't set ```write.parquet.compression-codec``` in their table property, if they are setting ```write.parquet.compression-codec``` to snappy in their iceberg config/iceberg session property, when they write a new file, on a previous version of Trino, the compression-codec being used would be snappy. But because this change will try to also read compression-codec from table properties when writing a file, this time the compression-codec being used would be zstd instead of snappy, which is not backward compatible. So we want to disable Iceberg from setting this table property if that's not being set by user.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -56,16 +63,30 @@ public class IcebergTableProperties
     public static final String SORTED_BY_PROPERTY = "sorted_by";
     public static final String LOCATION_PROPERTY = "location";
     public static final String FORMAT_VERSION_PROPERTY = "format_version";
+
+    public static final String WRITE_COMPRESSION = "write_compression";
```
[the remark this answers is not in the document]
SongChujun: I can change the property name. But I don't understand why session property should have a higher precedence over table property, if a user set table property for a particular table, it should have higher precedence compared to the **default** one from session property right?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -220,6 +252,19 @@ public static IcebergFileFormat getFileFormat(Map<String, Object> tablePropertie
         return (IcebergFileFormat) tableProperties.get(FILE_FORMAT_PROPERTY);
     }
 
+    public static Optional<HiveCompressionCodec> getHiveCompressionCodec(Map<String, Object> inputProperties)
```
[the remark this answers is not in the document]
SongChujun: will update the name

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +878,48 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
         if (replace) {
-            return catalog.newCreateOrReplaceTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, tableLocation, createTableProperties(session, tableMetadata, allowedExtraProperties));
+            txn = catalog.newCreateOrReplaceTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, tableLocation, createTableProperties(tableMetadata, allowedExtraProperties));
+        }
+        else {
+            txn = catalog.newCreateTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, Optional.ofNullable(tableLocation), createTableProperties(tableMetadata, allowedExtraProperties));
         }
-        return catalog.newCreateTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, Optional.ofNullable(tableLocation), createTableProperties(session, tableMetadata, allowedExtraProperties));
+
+        // If user doesn't set compression-codec for parquet, we need to remove write.parquet.compression-codec property,
```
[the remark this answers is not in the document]
SongChujun: Already replied in a previous comment.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -56,16 +63,30 @@ public class IcebergTableProperties
...
+    public static final int GZIP_MAX_COMPRESSION_LEVEL = 9;
+
     public static final String MAX_COMMIT_RETRY = "max_commit_retry";
     public static final String ORC_BLOOM_FILTER_COLUMNS_PROPERTY = "orc_bloom_filter_columns";
     public static final String ORC_BLOOM_FILTER_FPP_PROPERTY = "orc_bloom_filter_fpp";
     public static final String PARQUET_BLOOM_FILTER_COLUMNS_PROPERTY = "parquet_bloom_filter_columns";
     public static final String OBJECT_STORE_LAYOUT_ENABLED_PROPERTY = "object_store_layout_enabled";
     public static final String DATA_LOCATION_PROPERTY = "data_location";
     public static final String EXTRA_PROPERTIES_PROPERTY = "extra_properties";
+    public static final Set<IcebergFileFormat> VALID_ICEBERG_FILE_FORMATS_FOR_COMPRESSION_LEVEL_PROPERTY = ImmutableSet.of(IcebergFileFormat.AVRO);
```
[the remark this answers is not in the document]
SongChujun: Sure, I can remove support for setting compression_level.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: Sure, we may re-open it after we reach agreement on https://github.com/trinodb/trino/pull/25755#discussion_r2100937840

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: explained in a previous comment

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -252,6 +274,25 @@ private static void validateFormatVersion(int version)
...
+        if (compressionCodec.isPresent()) {
+            if (!isCompressionCodecSupportedForFormat(fileFormat, compressionCodec.get())) {
+                throw new TrinoException(NOT_SUPPORTED, format("Compression codec %s not supported for %s", compressionCodec.get(), fileFormat.humanName()));
+            }
+        }
+    }
+
+    public static boolean isCompressionCodecSupportedForFormat(IcebergFileFormat fileFormat, HiveCompressionCodec codec)
+    {
+        switch (codec) {
```
pettyjamesm: Nit: can be a switch expression

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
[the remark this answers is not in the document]
pettyjamesm: @raunaqmorarka - I would tend to agree with @SongChujun that the precedence order should prefer the table defined codec over the session property (when present) since the table definition seems more specific. I guess you could make an argument that the order should be: session property (if present, but default value is empty and not the default value from catalog config), table property (if present), catalog config default value. Is that what you're suggesting here?

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -932,6 +932,10 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
   - Optionally specifies the format of table data files; either `PARQUET`,
     `ORC`, or `AVRO`. Defaults to the value of the `iceberg.file-format` catalog
     configuration property, which defaults to `PARQUET`.
+* - `compression-codec`
+  - Optionally specifies the compression-codec used for writing the table; either `NONE`,
+    `ZSTD`, `SNAPPY`, `LZ4`,  or `GZIP`. Defaults to the value of the `iceberg.compression-codec` catalog
```
raunaqmorarka: Please wrap lines at 80 characters for anything in the docs

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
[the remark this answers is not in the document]
raunaqmorarka: Usually session property has precedence when it is set, because that is meant to allow users to override configured behaviour on a per query level. Per query write operation is more granular that per table, so if the session property exists it should get higher precedence.
I haven't been able to find a similar scenario in the code base where something is a session and catalog property while also being a table property. I think trying to be both a session property and table property at the same time is making this code and resulting behaviour quite complex, even though it is a seemingly simple thing.
I think we should consider dropping the existing session property entirely and just having table property and catalog config. I can't think of a great use case for being able to override compression per query for a table. Getting rid of the session property should simplify things here.

docs/src/main/sphinx/connector/iceberg.md
SongChujun: sure, will do.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergAvroFileWriter.java
```diff
@@ -63,12 +66,17 @@ public IcebergAvroFileWriter(
...
+        Map<String, String> compressionProperties = new HashMap<>();
+        compressionProperties.put(AVRO_COMPRESSION, toIcebergAvroCompressionName(hiveCompressionCodec));
+
         try {
             avroWriter = Avro.write(file)
                     .schema(icebergSchema)
                     .createWriterFunc(DataWriter::create)
                     .named(AVRO_TABLE_NAME)
-                    .set(AVRO_COMPRESSION, toIcebergAvroCompressionName(hiveCompressionCodec))
+                    .setAll(compressionProperties)
```
pettyjamesm: This change seems unnecessary now, no?

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -972,6 +972,11 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
   - Optionally specifies the format of table data files; either `PARQUET`,
     `ORC`, or `AVRO`. Defaults to the value of the `iceberg.file-format` catalog
     configuration property, which defaults to `PARQUET`.
+* - `compression-codec`
```
pettyjamesm: I think table properties are underscored instead of hyphenated, the code seems to use `"compression_codec"` in `IcebergTableProperties`

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1040,6 +1045,7 @@ CREATE TABLE test_table (
     c3 DOUBLE)
 WITH (
     format = 'ORC',
+    compression-codec = 'SNAPPY',
```
pettyjamesm: Match above documentation (assuming underscores)

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1029,7 +1034,7 @@ WITH (
     location = '/var/example_tables/test_table');
 ```
 
-The table definition below specifies to use ORC files, bloom filter index by columns
+The table definition below specifies to use ORC files with compression-codec SNAPPY, bloom filter index by columns
```
pettyjamesm: Match above documentation (assuming underscores)

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -8579,66 +8834,16 @@ public void testCreateTableWithCompressionCodec()
 
     private void testCreateTableWithCompressionCodec(HiveCompressionCodec compressionCodec)
     {
-        Session session = Session.builder(getSession())
-                .setCatalogSessionProperty(getSession().getCatalog().orElseThrow(), "compression_codec", compressionCodec.name())
-                .build();
         String tableName = "test_table_with_compression_" + compressionCodec;
-        String createTableSql = format("CREATE TABLE %s AS SELECT * FROM tpch.tiny.nation WITH NO DATA", tableName);
-        if (format == AVRO && compressionCodec == HiveCompressionCodec.LZ4) {
-            assertTrinoExceptionThrownBy(() -> computeActual(session, createTableSql))
```
pettyjamesm, line 8593: Shouldn't we still have tests asserting that tables created with invalid combinations still fail?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergAvroFileWriter.java
SongChujun: reverted this change

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -252,6 +274,23 @@ private static void validateFormatVersion(int version)
...
+    public static void validateCompression(IcebergFileFormat fileFormat, Optional<HiveCompressionCodec> compressionCodec)
+    {
+        if (compressionCodec.isPresent()) {
+            if (!isCompressionCodecSupportedForFormat(fileFormat, compressionCodec.get())) {
+                throw new TrinoException(NOT_SUPPORTED, format("Compression codec %s not supported for %s", compressionCodec.get(), fileFormat.humanName()));
+            }
+        }
+    }
+
+    public static boolean isCompressionCodecSupportedForFormat(IcebergFileFormat fileFormat, HiveCompressionCodec codec)
```
pettyjamesm: Nit: looks like this is only used within `validateCompression` and tests, so maybe make package-default visibility and annotate with `@VisibleForTesting`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +872,47 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
+            txn = catalog.newCreateTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, Optional.ofNullable(tableLocation), createTableProperties(tableMetadata, allowedExtraProperties));
+        }
+
+        // If user doesn't set compression-codec for parquet, we need to remove write.parquet.compression-codec property,
+        // Otherwise Iceberg will set write.parquet.compression-codec to zstd by default.
+        String parquetCompressionValue = txn.table().properties().get(PARQUET_COMPRESSION);
+        if (parquetCompressionValue != null && parquetCompressionValue.isEmpty()) {
+            txn.updateProperties()
+                    .remove(PARQUET_COMPRESSION)
+                    .commit();
```
pettyjamesm, line 890: Calling `.commit()` here seems wrong since we don't want to commit the returned transaction until after the create table operation is ready to commit- or am I misunderstanding the semantics of the Iceberg library on this part?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -863,25 +872,47 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
         PartitionSpec partitionSpec = parsePartitionFields(schema, getPartitioning(tableMetadata.getProperties()));
         SortOrder sortOrder = parseSortFields(schema, getSortOrder(tableMetadata.getProperties()));
 
+        Transaction txn;
```
pettyjamesm: Minor: avoid using abbreviated variable names, use `transaction` instead

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
pettyjamesm: Maybe I'm misunderstanding the implication of calling `.commit()` since I see there also exists a `Transaction#commitTransaction()` method.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2492,10 +2498,13 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
             }
         }
 
+        IcebergFileFormat oldFileFormat = getFileFormat(icebergTable.properties());
+        IcebergFileFormat newFileFormat = getFileFormat(icebergTable.properties());
```
raunaqmorarka: `newFileFormat = oldFileFormat` ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
```diff
@@ -287,19 +295,22 @@ private IcebergFileWriter createAvroWriter(
...
-            ConnectorSession session)
+            Map<String, String> storageProperties)
     {
         Closeable rollbackAction = () -> fileSystem.deleteFile(outputPath);
 
         List<Type> columnTypes = icebergSchema.columns().stream()
                 .map(column -> toTrinoType(column.type(), typeManager))
                 .collect(toImmutableList());
 
+        HiveCompressionCodec compressionCodec = getHiveCompressionCodec(IcebergFileFormat.AVRO, storageProperties)
```
raunaqmorarka: nit: static import AVRO here and in the other files

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -970,8 +970,13 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
   - Description
```
raunaqmorarka, line 973: Commit message: 
```
Support setting compression_codec table property in Iceberg

Makes compression_codec a table property and drops support for
compression_codec session property for simplicity
```

raunaqmorarka (commented):
lgtm overall

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
SongChujun: changed all IcebergFileFormat.[Parquet, Orc, Avro] to [Parquet, Orc, Avro]

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
SongChujun: changed it here.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: I would say txn is a very common abbr for transaction, change it to transaction though.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: Yes, I just followed the similar calling convention from other similar places.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
```diff
@@ -97,19 +100,22 @@ public class IcebergFileWriterFactory
     private final NodeVersion nodeVersion;
     private final FileFormatDataSourceStats readStats;
     private final OrcWriterStats orcWriterStats = new OrcWriterStats();
+    private final IcebergConfig icebergConfig;
```
ebyhr: Don't put mutable objects in fields.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
```diff
@@ -97,19 +100,22 @@ public class IcebergFileWriterFactory
...
             NodeVersion nodeVersion,
             FileFormatDataSourceStats readStats,
+            IcebergConfig icebergConfig,
             OrcWriterConfig orcWriterConfig)
     {
         checkArgument(!orcWriterConfig.isUseLegacyVersion(), "the ORC writer shouldn't be configured to use a legacy version");
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.nodeVersion = requireNonNull(nodeVersion, "nodeVersion is null");
         this.readStats = requireNonNull(readStats, "readStats is null");
+        this.icebergConfig = requireNonNull(icebergConfig, "readStats is null");
```
ebyhr: The message is wrong. However, see another comment in fields.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -56,6 +63,9 @@ public class IcebergTableProperties
     public static final String SORTED_BY_PROPERTY = "sorted_by";
     public static final String LOCATION_PROPERTY = "location";
     public static final String FORMAT_VERSION_PROPERTY = "format_version";
+
```
ebyhr: Remove redundant empty line.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -974,6 +983,32 @@ public static void verifyExtraProperties(Set<String> basePropertyKeys, Map<Strin
...
 
+    public static Optional<HiveCompressionCodec> getHiveCompressionCodec(IcebergFileFormat icebergFileFormat, Map<String, String> storageProperties)
+    {
+        String compressionProperty = getCompressionPropertyName(icebergFileFormat);
+
+        return Optional.ofNullable(storageProperties.get(compressionProperty))
+                .filter(value -> !value.isEmpty())
+                .map(value -> {
+                    try {
+                        return HiveCompressionCodec.valueOf(value.toUpperCase(Locale.ENGLISH));
```
ebyhr: Static import `ENGLISH`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/snowflake/IcebergSnowflakeCatalogModule.java
```diff
@@ -35,7 +35,7 @@ protected void setup(Binder binder)
         binder.bind(TrinoCatalogFactory.class).to(TrinoIcebergSnowflakeCatalogFactory.class).in(Scopes.SINGLETON);
 
         IcebergConfig icebergConfig = buildConfigObject(IcebergConfig.class);
-        if (icebergConfig.getFileFormat() != IcebergFileFormat.PARQUET) {
+        if (icebergConfig.getFileFormat() != PARQUET) {
```
ebyhr: Please separate a commit. This change is unrelated to `compression_codec` table property change.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestFileBasedConflictDetection.java
```diff
@@ -41,6 +41,7 @@
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.trino.plugin.iceberg.ColumnIdentity.TypeCategory.PRIMITIVE;
 import static io.trino.plugin.iceberg.ColumnIdentity.TypeCategory.STRUCT;
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
```
ebyhr: Please separate a commit. This change is unrelated to `compression_codec` table property change.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergRegisterTableProcedure.java
```diff
@@ -55,6 +55,7 @@
 import static com.google.common.io.RecursiveDeleteOption.ALLOW_INSECURE;
 import static io.trino.plugin.hive.HiveTestUtils.HDFS_FILE_SYSTEM_FACTORY;
 import static io.trino.plugin.hive.metastore.file.TestingFileHiveMetastore.createTestingFileHiveMetastore;
+import static io.trino.plugin.iceberg.IcebergFileFormat.ORC;
```
ebyhr: Please separate a commit. This change is unrelated to `compression_codec` table property change.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/BaseTrinoCatalogTest.java
```diff
@@ -66,6 +65,7 @@
 import static io.trino.metastore.TableInfo.ExtendedRelationType.TRINO_MATERIALIZED_VIEW;
 import static io.trino.metastore.TableInfo.ExtendedRelationType.TRINO_VIEW;
 import static io.trino.plugin.hive.HiveErrorCode.HIVE_DATABASE_LOCATION_ERROR;
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
```
ebyhr: Please separate a commit. This change is unrelated to `compression_codec` table property change.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
         assertUpdate("DROP TABLE " + tableName);
     }
 
+    @Test
+    public void testCreateTableAsWithCompressionCodecs()
+    {
+        String compressionProperty = getCompressionPropertyName(format);
+
+        for (HiveCompressionCodec compressionCodec : getCompressionCodecs(Optional.empty())) {
+            String tableName = String.format("test_ctas_%s_codec_%s_%s", compressionCodec.name(), format, randomNameSuffix());
```
ebyhr: Use `formatted` method or static import `format`. See https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#prefer-string-formatting
> note that format() should always be statically imported

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+                        "SELECT count(*) FROM nation WHERE nationkey >= 10");
+
+                assertUpdate(format("DROP TABLE %s", tableName));
+
+            } else {
+                assertQueryFails(
+                        format("CREATE TABLE %s WITH (format = '%s', compression_codec = '%s') AS SELECT * FROM nation", tableName, format, compressionCodec.name()),
+                        "Compression codec LZ4 not supported for .*");
+            }
+
```
ebyhr: Remove redundant empty line. Same for other new test methods.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+        }
+    }
+
+    private List<HiveCompressionCodec> getCompressionCodecs(Optional<HiveCompressionCodec> codecToExclude)
```
ebyhr: This method can be static.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+    {
+        return Arrays.stream(HiveCompressionCodec.values())
+                .filter(codec -> !(codecToExclude.isPresent() && codec.equals(codecToExclude.get())))
+                .collect(ImmutableList.toImmutableList());
+    }
+
+    @Test
+    public void testCreateTableAsWithCompressionCodecUnsupported()
+    {
+        String tableName = String.format("test_ctas_unsupported_%s_%s", format, randomNameSuffix());
```
ebyhr: Static import `format`.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+
+        assertQuery(format("SELECT COUNT(*) FROM \"%s$properties\" WHERE key = 'write.parquet.compression-codec'", tableName), "SELECT 0");
+
+        assertUpdate(format("ALTER TABLE %s SET PROPERTIES compression_codec = 'ZSTD'", tableName));
+        assertQuery(format("SELECT value FROM \"%s$properties\" WHERE key = 'write.format.default'", tableName), format("VALUES '%s'", AVRO));
+        assertQuery(format("SELECT value FROM \"%s$properties\" WHERE key = 'write.avro.compression-codec'", tableName), "VALUES 'ZSTD'");
+
+        assertUpdate(format("DROP TABLE %s", tableName));
+    }
+
```
ebyhr, line 7039: Fix format of the new test methods.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+                assertQuery(format("SELECT count(*) FROM \"%s$files\" WHERE file_path LIKE '%%.%s'", tableName, format.name().toLowerCase(Locale.ENGLISH)), "VALUES 2");
+
+            } else {
+                assertQueryFails(
+                        format("ALTER TABLE %s SET PROPERTIES compression_codec = '%s'", tableName, newCompressionCodec),
+                        "Compression codec LZ4 not supported for .*");
+            }
+
+            assertUpdate(format("DROP TABLE %s", tableName));
```
ebyhr: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#prefer-string-formatting
> if you only need to append something, consider using the + operator.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+
+        for (HiveCompressionCodec compressionCodec : getCompressionCodecs(Optional.empty())) {
+            String tableName = String.format("test_ctas_%s_codec_%s_%s", compressionCodec.name(), format, randomNameSuffix());
+            if (isCompressionCodecSupportedForFormat(format, compressionCodec))
+            {
+                assertUpdate(
+                        format("CREATE TABLE %s WITH (format = '%s', compression_codec = '%s') AS SELECT * FROM nation", tableName, format, compressionCodec.name()),
+                        "SELECT count(*) FROM nation");
+
+                assertQuery(
```
ebyhr: `assertQuery` isn't recommended anymore. Please use `assertThat(query(...))` instead.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+                        format("SELECT value FROM \"%s$properties\" WHERE key = 'write.format.default'", tableName),
+                        format("VALUES '%s'", format));
```
ebyhr: Please add a helper method like `BaseIcebergSystemTables#getTableProperties`:
```java
    private Map<String, String> getTableProperties(String tableName)
    {
        return computeActual("SELECT key, value FROM \"" + tableName + "$properties\"").getMaterializedRows().stream()
                .collect(toImmutableMap(row -> (String) row.getField(0), row -> (String) row.getField(1)));
    }
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6820,259 @@ public void testUpdatingFileFormat()
...
+
+                assertQuery(
+                        format("SELECT value FROM \"%s$properties\" WHERE key = '%s'", tableName, compressionProperty),
+                        format("VALUES '%s'", compressionCodec.name()));
+                assertQuery(format("SELECT * FROM %s", tableName), "SELECT * FROM nation");
+                assertQuery(format("SELECT count(*) FROM \"%s$files\" WHERE file_path LIKE '%%.%s'", tableName, format.name().toLowerCase(Locale.ENGLISH)), "VALUES 1");
```
ebyhr: Static import `ENGLISH`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
SongChujun: I find it hard to pass the result of ``icebergConfig.getCompressionCodec()`` we want if IcebergConfig is not included as a field. And it seems to me similar practice is also used in other places like ``TrinoIcebergRestCatalogFactory``.  So I want to keep it here. If you have other ways to remove this field while still being able to  pass the result of ``icebergConfig.getCompressionCodec()``, please let me know.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
SongChujun: changed description

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/snowflake/IcebergSnowflakeCatalogModule.java
SongChujun: change removed

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: removed String here

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: changed to static method

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: removed format here

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: removed format here

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: changed to code to use assertThat(query(...)) instead

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: help method to get table property by key added

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
ebyhr: Why do you need the entire config class instead of only the compression codec here?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6823,255 @@ public void testUpdatingFileFormat()
...
+
+    @Test
+    public void testCreateTableAsWithCompressionCodecs()
+    {
+        String compressionProperty = getCompressionPropertyName(format);
+
+        for (HiveCompressionCodec compressionCodec : getCompressionCodecs(Optional.empty())) {
+            String tableName = format("test_ctas_%s_codec_%s_%s", compressionCodec.name(), format, randomNameSuffix());
+            if (isCompressionCodecSupportedForFormat(format, compressionCodec))
+            {
```
ebyhr: Move `{` to the previous line.

Same for other places.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6823,255 @@ public void testUpdatingFileFormat()
...
+                assertUpdate(
+                        format("CREATE TABLE %s WITH (format = '%s', compression_codec = '%s') AS SELECT * FROM nation", tableName, format, compressionCodec.name()),
+                        "SELECT count(*) FROM nation");
+
+                assertThat(getTableProperty(tableName, DEFAULT_FILE_FORMAT))
+                        .isEqualTo(Optional.of(format.toString()));
```
ebyhr: `Optional.of` can be removed once we use `contains` method which verifies that the actual Optional contains the given value.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6823,255 @@ public void testUpdatingFileFormat()
         assertUpdate("DROP TABLE " + tableName);
     }
 
+    private Optional<String> getTableProperty(String tableName, String propertyName)
```
ebyhr: We put helper methods after their callers. 

Same for `getCompressionCodecs` method.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6823,255 @@ public void testUpdatingFileFormat()
...
+    {
+        return computeActual("SELECT key, value FROM \"" + tableName + "$properties\"" + " WHERE key = '" + propertyName + "'").getMaterializedRows().stream()
+                .map(row -> (String) row.getField(1)).findFirst();
+    }
```
ebyhr: Returning map will reduce redundant calls to `$properties` table. 
```java
    private Map<String, String> getTableProperties(String tableName)
    {
        return computeActual("SELECT key, value FROM \"" + tableName + "$properties\"").getMaterializedRows().stream()
                .collect(toImmutableMap(row -> (String) row.getField(0), row -> (String) row.getField(1)));
    }
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6818,6 +6823,255 @@ public void testUpdatingFileFormat()
...
+
+                assertThat(query("SELECT * FROM " + tableName)).matches("SELECT * FROM nation");
+                assertThat(query(format("SELECT count(*) FROM \"%s$files\" WHERE file_path LIKE '%%.%s'", tableName, format.name().toLowerCase(ENGLISH)))).matches("SELECT CAST(1 AS BIGINT)");
+
+                assertUpdate(
+                        format("INSERT INTO %s SELECT * FROM nation WHERE nationkey >= 10", tableName),
+                        "SELECT count(*) FROM nation WHERE nationkey >= 10");
+
+                assertUpdate("DROP TABLE " + tableName);
+            } else {
```
ebyhr: Move `else {` to the new line.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: done and also done for other places

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
SongChujun: used contains instead

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
SongChujun: Because the entire config is initialized via injection.

ebyhr: See other fields. OrcWriterOptions is retrieved via injected OrcWriterConfig. 
Note that we extract only necessary variables into fields from config classes in this project. 
Let me know if you still don't understand it. I will simply push the change.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
```diff
@@ -97,19 +101,23 @@ public class IcebergFileWriterFactory
...
 
     @Inject
     public IcebergFileWriterFactory(
             TypeManager typeManager,
             NodeVersion nodeVersion,
             FileFormatDataSourceStats readStats,
+            IcebergConfig icebergConfig,
             OrcWriterConfig orcWriterConfig)
     {
+        checkArgument(icebergConfig != null);
```
ebyhr: We don't check nullability for config classes. This line can be removed.  
https://github.com/trinodb/trino/pull/13940

SongChujun: line removed

pettyjamesm (approved):
LGTM, thanks @SongChujun and @ebyhr!

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -868,25 +872,47 @@ public static Transaction newCreateTableTransaction(TrinoCatalog catalog, Connec
...
+        validateCompression(fileFormat, compressionCodec);
+
+        Map<String, String> tableCompressionProperties = calculateTableCompressionProperties(fileFormat, fileFormat, ImmutableMap.of(), tableMetadata.getProperties());
+
+        tableCompressionProperties.forEach(propertiesBuilder::put);
+
+        // Iceberg will set write.parquet.compression-codec to zstd by default if this property is not set: https://github.com/trinodb/trino/issues/20401,
+        // but we don't want to set this property if this is not explicitly set by customer via set table properties.
+        if (!(fileFormat == PARQUET && compressionCodec.isPresent())) {
+            propertiesBuilder.put(PARQUET_COMPRESSION, "");
```
ebyhr, line 915: FYI, this change broke an integration with Databricks Unity catalog via Iceberg REST catalog. They throw `Malformed request: Malformed request: INVALID_PARAMETER_VALUE: Table properties contain prohibited keys: write.parquet.compression-codec` when the connector creates a new table. 

I just found the regression when I was manually testing Unity catalog.

SongChujun: Is there any problem with this logic here that we need to change?

ebyhr: No, not for now. Databricks Unity Catalog isn't supported by our Iceberg REST catalog integration. I'll check the details and get back to you.

hqbhoho:
Hi, if user doesn't set `compression-codec` for parquet, it will remove `write.parquet.compression-codec property`.  But in iceberg will use `GZIP`  as default compresssion codec in fact. Could we not remove `write.parquet.compression-codec` property  If user doesn't set `compression-codec` for parquet and only not expose this table properties when use other format like `avro`?

chenjian2664:
@hqbhoho You could start a discussion on slack channel, the pr is merged anyway, FYI

hqbhoho:
I actually raised this on Slack a while back, but it seems it got missed.
https://app.slack.com/client/TFKPX15NC/activity

PR #25756, opened 2025-05-09
Restore SelectedRole propagation through identity enabled roles

homar:
## Description
A [prior change ](https://github.com/trinodb/trino/commit/3b251c6015b0b5641108aa70fa3b3e2a9ee97f43)inadvertently disrupted the contract for propagating the intended role from the engine to plugins. 
This commit reinstates the expected behaviour, ensuring that the desired role is correctly enabled. While this change does not directly affect the project, it corrects functionality for third-party plugins that depend on this specific behaviour.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
+++ b/core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
@@ -254,12 +254,17 @@ public class HttpRequestSessionContextFactory
         String user = trinoUser != null ? trinoUser : authenticatedIdentity.map(Identity::getUser).orElse(null);
         assertRequest(user != null, "User must be set");
         SelectedRole systemRole = parseSystemRoleHeaders(protocolHeaders, headers);
+        ImmutableSet.Builder<String> systemEnabledRoles = ImmutableSet.builder();
+        if (systemRole.getType() == Type.ROLE) {
+            systemEnabledRoles.add(systemRole.getRole().orElseThrow());
+        }
         Identity newIdentity = authenticatedIdentity
                 .map(identity -> Identity.from(identity).withUser(user))
                 .orElseGet(() -> Identity.forUser(user))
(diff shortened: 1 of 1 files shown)


lukasz-walkiewicz (approved):
Link the previous change that you mention in the description

core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
```diff
@@ -254,13 +254,19 @@ private Identity buildSessionIdentity(Optional<Identity> authenticatedIdentity,
...
+        }
         Identity newIdentity = authenticatedIdentity
                 .map(identity -> Identity.from(identity).withUser(user))
                 .orElseGet(() -> Identity.forUser(user))
                 .withAdditionalConnectorRoles(parseConnectorRoleHeaders(protocolHeaders, headers))
                 .withAdditionalExtraCredentials(parseExtraCredentials(protocolHeaders, headers))
                 .withAdditionalGroups(groupProvider.getGroups(user))
+                .withEnabledRoles(systemEnabledRoles.build())
                 .build();
+
```
kokosing, line 269: unrelated empty line

core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
```diff
@@ -254,13 +254,19 @@ private Identity buildSessionIdentity(Optional<Identity> authenticatedIdentity,
         String user = trinoUser != null ? trinoUser : authenticatedIdentity.map(Identity::getUser).orElse(null);
```
kokosing, line 254: I would add to the commit message. 

> That the previous commit broke the contract between engine and plugins, and it was no longer passing desired role to be enabled. And that this commit restores that. The change has no impact to the project but affects 3rd party plugins that relied on the previous behavior.

Please do not copy paste blindly, I wanted to give you content, but form most likely requires some polishing. 

Please also update PR titlte and description to match that.

PR #25772, opened 2025-05-12
Remove dependency on airlift units from client

wendigo:
## Release notes

```markdown
## Section
```


martint:
What's the purpose of this change? It just seems to make things more complicated.

wendigo:
@martint we have long-standing issue with a mixed dependency versions and pinning. This removes this problem.

martint:
What's the issue? It's the first time I hear about it and haven't seen any PRs or builds fail due to that.

wendigo:
@martint trino-jdbc pulls in old version of units while trino-main depends on the newer one. In the trino-jdbc tests, when you want to bring TestingTrinoServer up with some additional catalog configuration that depends on the Duration or DataSize being in a newer version, this will fail.

wendigo:
We don't need units in the client to make it work. So removing it, removes this dependency mess

martint:
There's a more fundamental problem. `trino-jdbc` and `trino-client` target Java 11, but `TestingTrinoServer` is in `trino-main`, which is built with Java 24. Any transitive dependencies in trino-main (via TrinoTestingServer) that are incompatible with Java 11 will cause similar problems. At a minimum, they may cause trino-jdbc/trino-client to run tests with a dependency that doesn't match the dependency that's actually bundled in the package during final build.

wendigo:
@martint yes but since we keep versions in sync across modules (except for units) we are sure that jdbc uses dependencies supporting JDK11+

martint:
That's not guaranteed. If we need to bump a dependency in trino-main, we'll do that, even if that dependency is not compatible with Java 11.

wendigo:
@martint I understand but we are not in this situation right now except for airlift:units which I'm trying to remove in this PR

wendigo:
@martint I'd like to move forward with this

PR #25775, opened 2025-05-12
Recategorize NPE in StatementAnalyzer

artem-sokolov:
## Description

Recategorize `NullPointerException` that gets thrown when there's a mismatch between columns and insert layout from `Metadata` during `StatementAnalyzer#createMergeAnalysis`.

Instead, it now throws a `TrinoException` with a corresponding message, e.g.:
```
Missing field index for column: 'year', available columns: ('nationkey', 'regionkey', 'name', 'comment')
```

## Release notes

( ) Release notes are required, with the following suggested text


xkrogen:
Resubmitted at https://github.com/trinodb/trino/pull/26023

Issue #25780, opened 2025-05-13
Trino forgets vended-credentials in S3FileSystem::deleteObjects


twuebi:
Hi, 

## Problem

When executing `expire_snapshots` on an iceberg table with Trino 475, we're encountering missing credentials errors that didn't occur on Trino 467. The root cause seems to be in `S3FileSystem::delete_objects` where the second call to `overrideConfiguration()` on the builder overwrites the `credentialsProvider` set by the first call. 

## Reproduction

`expire_snapshots` on an iceberg table on an s3-compatible storage with a rest catalog & "iceberg.rest-catalog.vended-credentials-enabled" = 'true' using Trino 475

## Details

In `src/main/java/io/trino/filesystem/s3/S3FileSystem.java`, two consecutive calls to `Builder::overrideConfiguration` occur:

1. First call: `overrideConfiguration(context::applyCredentialProviderOverride)`, sets credential, essential for vended-credentials to work in iceberg 
2. Second call: `overrideConfiguration(disableStrongIntegrityChecksums())`, a workaround for missing strong integrity checks in some storage providers:

The second `overrideConfiguration` nulls the previously set `credentialsProvider` in the `awsRequestOverrideConfig`. This then causes the delete request to fail with:

[...]

PR #25782, opened 2025-05-13
Fix iceberg REST catalog expire_snapshots

pettyjamesm:
## Description
Fixes https://github.com/trinodb/trino/issues/25780

## Release notes

```markdown
## Iceberg connector
* Fix expire_snapshots functionality when using an Iceberg REST catalog and S3 compatible storage ({issue}`25782`)
```

### DIFF
--- a/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
+++ b/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
@@ -196,10 +196,9 @@ final class S3FileSystem
                         .toList();
 
                 DeleteObjectsRequest request = DeleteObjectsRequest.builder()
-                        .overrideConfiguration(context::applyCredentialProviderOverride)
+                        .overrideConfiguration(builder -> context.applyCredentialProviderOverride(disableStrongIntegrityChecksums(builder)))
                         .requestPayer(requestPayer)
                         .bucket(bucket)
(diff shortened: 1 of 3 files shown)


lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -398,10 +397,8 @@ private static void validateS3Location(Location location)
     // TODO (https://github.com/trinodb/trino/issues/24955):
     // remove me once all of the S3-compatible storage support strong integrity checks
     @SuppressWarnings("deprecation")
-    static AwsRequestOverrideConfiguration disableStrongIntegrityChecksums()
+    static AwsRequestOverrideConfiguration.Builder disableStrongIntegrityChecksums(AwsRequestOverrideConfiguration.Builder builder)
```
ebyhr, line 400: We added `LegacyMd5Plugin` in #25711. Why do we still need this method?
cc: @wendigo

ebyhr (commented):
Can we add a test?

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
wendigo: https://github.com/trinodb/trino/pull/25791 this should help I think

PR #25786, opened 2025-05-14
Various code cleanups

electrum:
## Release notes

### DIFF
--- a/core/trino-parser/src/main/java/io/trino/sql/tree/Query.java
+++ b/core/trino-parser/src/main/java/io/trino/sql/tree/Query.java
@@ -133,29 +133,30 @@ public class Query
     @Override
     public List<Node> getChildren()
     {
-        ImmutableList.Builder<Node> nodes = ImmutableList.builder();
-        nodes.addAll(functions);
-        nodes.addAll(sessionProperties);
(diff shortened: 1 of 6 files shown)


core/trino-parser/src/main/java/io/trino/sql/tree/Query.java
```diff
@@ -133,29 +133,30 @@ public <R, C> R accept(AstVisitor<R, C> visitor, C context)
...
-        with.ifPresent(nodes::add);
-        nodes.add(queryBody);
-        orderBy.ifPresent(nodes::add);
-        offset.ifPresent(nodes::add);
-        limit.ifPresent(nodes::add);
-        return nodes.build();
+        return ImmutableList.<Node>builder()
+                .addAll(functions)
+                .addAll(sessionProperties)
+                .addAll(with.stream().toList())
```
wendigo, line 139: it can be also: `with.map(ImmutableList::of).orElseGet(ImmutableList::of)`

electrum: Yeah, but that's harder to read, and I doubt it makes a difference in practice.

PR #25787, opened 2025-05-14
Fix missing file extension in Hive connector output files

chenjian2664:
## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-metastore/src/main/java/io/trino/metastore/StorageFormat.java
+++ b/lib/trino-metastore/src/main/java/io/trino/metastore/StorageFormat.java
@@ -28,18 +28,20 @@ import static java.util.Objects.requireNonNull;
 @Immutable
 public class StorageFormat
 {
-    public static final StorageFormat NULL_STORAGE_FORMAT = new StorageFormat(null, null, null);
+    public static final StorageFormat NULL_STORAGE_FORMAT = new StorageFormat(null, null, null, null);
     public static final StorageFormat VIEW_STORAGE_FORMAT = NULL_STORAGE_FORMAT;
 
     private final String serde;
     private final String inputFormat;
     private final String outputFormat;
+    private final String fileExtension;
 
-    private StorageFormat(String serde, String inputFormat, String outputFormat)
+    private StorageFormat(String serde, String inputFormat, String outputFormat, String fileExtension)
     {
         this.serde = serde;
         this.inputFormat = inputFormat;
         this.outputFormat = outputFormat;
+        this.fileExtension = fileExtension;
     }
 
     public String getSerde()
@@ -66,6 +68,14 @@ public class StorageFormat
         return outputFormat;
     }
 
+    public String getFileExtension()
+    {
+        if (fileExtension == null) {
+            throw new TrinoException(HIVE_UNSUPPORTED_FORMAT, "FileExtension is not present in StorageFormat: " + this.outputFormat);
+        }
+        return fileExtension;
+    }
+
     @JsonProperty("serde")
     public String getSerDeNullable()
     {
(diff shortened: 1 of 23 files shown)


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -379,7 +379,8 @@ public class DeltaLakeMetadata
     public static final StorageFormat DELTA_STORAGE_FORMAT = create(
             LAZY_SIMPLE_SERDE_CLASS,
             SEQUENCEFILE_INPUT_FORMAT_CLASS,
-            HIVE_SEQUENCEFILE_OUTPUT_FORMAT_CLASS);
+            HIVE_SEQUENCEFILE_OUTPUT_FORMAT_CLASS,
+            ".seq");
```
Praveen2112, line 383: Do we need `.` as a prefix - why not as `seq`

lib/trino-metastore/src/main/java/io/trino/metastore/StorageFormat.java
```diff
@@ -28,18 +28,20 @@
...
     private final String outputFormat;
+    private final String fileExtension;
 
-    private StorageFormat(String serde, String inputFormat, String outputFormat)
+    private StorageFormat(String serde, String inputFormat, String outputFormat, String fileExtension)
     {
         this.serde = serde;
         this.inputFormat = inputFormat;
         this.outputFormat = outputFormat;
+        this.fileExtension = fileExtension;
```
wendigo, line 44: rnn here (and for other fields, in a separate commit)

lib/trino-metastore/src/main/java/io/trino/metastore/StorageFormat.java
```diff
@@ -66,6 +68,14 @@ public String getOutputFormat()
         return outputFormat;
     }
 
+    public String getFileExtension()
+    {
+        if (fileExtension == null) {
+            throw new TrinoException(HIVE_UNSUPPORTED_FORMAT, "File extension is not present in StorageFormat");
```
wendigo, line 74: File extension is missing from StorageFormat

Praveen2112 (commented):
Any testing based on "$path" column ? Does hive also output the data with their extension ?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveStorageFormat.java
```diff
@@ -61,57 +61,70 @@ public enum HiveStorageFormat
     ORC(
             ORC_SERDE_CLASS,
             ORC_INPUT_FORMAT_CLASS,
-            ORC_OUTPUT_FORMAT_CLASS),
+            ORC_OUTPUT_FORMAT_CLASS,
+            ".orc"),
```
wendigo, line 65: can we extract these to constants? or an enum? can we omit dot and add it where we use this extension?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveWriterFactory.java
```diff
@@ -675,7 +675,7 @@ public static int getBucketFromFileName(String fileName)
     public static String getFileExtension(HiveCompressionCodec compression, StorageFormat format)
     {
         // text format files must have the correct extension when compressed
-        return compression.getHiveCompressionKind()
+        return format.getFileExtension() + compression.getHiveCompressionKind()
```
wendigo, line 678: move "." here

plugin/trino-hive/src/test/java/io/trino/plugin/hive/metastore/TestStorage.java
```diff
@@ -29,7 +29,7 @@ public class TestStorage
     public void testRoundTrip()
     {
         Storage storage = Storage.builder()
-                .setStorageFormat(StorageFormat.create("abc", "in", "out"))
+                .setStorageFormat(StorageFormat.create("abc", "in", "out", "ext"))
```
wendigo, line 32: it's inconsistent, in other cases we add a leading dot

wendigo (approved):

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveStorageFormat.java
ebyhr: I disagree with extracting as constants. We must jump to constants to see the value.

lib/trino-metastore/src/main/java/io/trino/metastore/StorageFormat.java
ebyhr: I believe all parameters are nullable.

lib/trino-metastore/src/main/java/io/trino/metastore/StorageFormat.java
chenjian2664: I am following the get method in this class. updated the `File extension` to `fileExtension`

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: It's a file extension, so I think leading '.' is make sense.

chenjian2664: Do you think which class is most suitable for the adding

chenjian2664: added in the `BaseHiveConnectorTest`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/HiveWriteUtils.java
```diff
@@ -240,7 +240,12 @@ public static Optional<Boolean> directoryExists(TrinoFileSystem fileSystem, Loca
 
     public static boolean isFileCreatedByQuery(String fileName, String queryId)
     {
-        return fileName.startsWith(queryId) || fileName.endsWith(queryId);
+        return fileName.startsWith(queryId)
```
chenjian2664, line 243: we are using the [logic](https://github.com/trinodb/trino/blob/6ff430abf9742f36ac94eec5c63ebd979a170705/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveWriterFactory.java#L427-L429) to build fileName/Path, and expect the fileNames always startsWith or endsWith queryId, while the `HiveWriteFactory#getFileExtension` possible returns extension, it seems a bug previously

PR #25811, opened 2025-05-16
Allow using GCS without providing credentials

### LINKED ISSUE #25810: Should be possible to use GCS without authentication
My use-case is local development and testing using [fake-gcs-server](https://github.com/fsouza/fake-gcs-server). Auth is not required to use its GCS-compatible API.

However, Trino does not appear to have a way to not attempt authentication with Google's servers, even when setting the `gcs.endpoint` property.

If no auth-related properties (like `gcs.json-key`) are set, then [GcsStorageFactory](https://github.com/trinodb/trino/blob/475/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java#L99) calls `GoogleCredentials.getApplicationDefault()`.

However, this method throws an exception if no credentials are found:

```
Caused by: java.io.IOException: Your default credentials were not found. To set up Application Default Credentials for your environment, see https://cloud.google.com/docs/authentication/external/set-up-adc.
```

In contrast, if [`StorageOptions.Builder#setCredentials`](https://github.com/trinodb/trino/blob/475/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java#L116) is simply never called, the [underlying GCS library](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L385-L391) populates it by calling `GoogleCredentials.getApplicationDefault()` but ignoring any exceptions it throws.

I have a temporary workaround to just not call `StorageOptions.Builder#setCredentials` in `GcsStorageFactory` if `gcs.endpoint` is set:
```
diff --git a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
index f8ea12c452..b6b20537fd 100644
--- a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
+++ b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
@@ -93,6 +93,9 @@ public class GcsStorageFactory
                     credentials = GoogleCredentials.fromStream(inputStream).createScoped(DEFAULT_SCOPES);
                 }
             }
+            else if (endpoint.isPresent()) {
+                credentials = null;
+            }
             else {
                 credentials = jsonGoogleCredential.orElseGet(() -> {
                     try {
@@ -110,10 +113,13 @@ public class GcsStorageFactory

             endpoint.ifPresent(storageOptionsBuilder::setHost);

+            if (credentials != null) {
+                storageOptionsBuilder.setCredentials(credentials);
+            }
+
             // Note: without uniform strategy we cannot retry idempotent operations.
             // The trino-filesystem api does not violate the conditions for idempotency, see https://cloud.google.com/storage/docs/retry-strategy#java for details.
             return storageOptionsBuilder
-                    .setCredentials(credentials)
                     .setStorageRetryStrategy(getUniformStorageRetryStrategy())
                     .setRetrySettings(RetrySettings.newBuilder()
                             .setMaxAttempts(maxRetries + 1)
```

The three solutions I see are, in order of preference:
- where `GcsStorageFactory` [currently calls](https://github.com/trinodb/trino/blob/475/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java#L97-L104) `getApplicationDefault()`, just set `credentials` to null instead, and if it is indeed null, do not call `StorageOptions.Builder#setCredentials`, allowing the underlying library's behavior to apply
- suppress exceptions thrown by `GoogleCredentials.getApplicationDefault()` like the GCS client library itself does by default
- add an additional config property to disable auth for GCS altogether, perhaps resulting in passing `NoCredentials.getInstance()` as the credentials

I'm happy to open a PR for this.

patricklucas:
## Description

If a JSON key or key file path are not provided, and using access tokens is not enabled, do not attempt to get the Application Default Credentials, instead allowing the GCS client library to use its own default behavior.

This will allow using Trino with custom GCS-compatible endpoints without auth, such as during development and testing.

Fixes #25810

## Release notes

```markdown
## Section
* Add support for using GCS without credentials (#25810)
```

### DIFF
--- a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
+++ b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
@@ -99,7 +99,8 @@ public class GcsStorageFactory
                         return GoogleCredentials.getApplicationDefault();
                     }
                     catch (IOException e) {
-                        throw new UncheckedIOException(e);
+                        // This is consistent with the GCP SDK when no credentials are available in the environment
+                        return null;
                     }
                 });
             }
@@ -108,12 +109,15 @@ public class GcsStorageFactory
                 storageOptionsBuilder.setProjectId(projectId);
             }
 
+            if (credentials != null) {
+                storageOptionsBuilder.setCredentials(credentials);
+            }
+
             endpoint.ifPresent(storageOptionsBuilder::setHost);
 
             // Note: without uniform strategy we cannot retry idempotent operations.
             // The trino-filesystem api does not violate the conditions for idempotency, see https://cloud.google.com/storage/docs/retry-strategy#java for details.
             return storageOptionsBuilder
-                    .setCredentials(credentials)
                     .setStorageRetryStrategy(getUniformStorageRetryStrategy())
                     .setRetrySettings(RetrySettings.newBuilder()
                             .setMaxAttempts(maxRetries + 1)
--- a/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsStorageFactory.java
+++ b/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsStorageFactory.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.filesystem.gcs;
+
+import com.google.auth.Credentials;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageOptions;
+import io.trino.spi.security.ConnectorIdentity;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestGcsStorageFactory
+{
+    @Test
+    void testDefaultCredentials()
+            throws Exception
+    {
+        Credentials expectedCredentials = StorageOptions.newBuilder().build().getCredentials();
+
+        // No credentials options are set
+        GcsFileSystemConfig config = new GcsFileSystemConfig();
+
+        GcsStorageFactory storageFactory = new GcsStorageFactory(config);
+
+        Credentials actualCredentials;
+        try (Storage storage = storageFactory.create(ConnectorIdentity.ofUser("test"))) {
+            actualCredentials = storage.getOptions().getCredentials();
+        }
+
+        assertThat(actualCredentials)
+                .as("if credentials are not explicitly configured, should have same behavior as the GCS client")
+                .isEqualTo(expectedCredentials);
+    }
+}


patricklucas:
I will take a look at the CLA once I get positive indication this change would be accepted.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
```diff
@@ -94,26 +94,22 @@ public Storage create(ConnectorIdentity identity)
...
             else {
-                credentials = jsonGoogleCredential.orElseGet(() -> {
-                    try {
-                        return GoogleCredentials.getApplicationDefault();
-                    }
-                    catch (IOException e) {
-                        throw new UncheckedIOException(e);
-                    }
-                });
+                credentials = jsonGoogleCredential.orElse(null);
```
findinpath, line 97: ideally the PR would have an integration test showcasing the use-case without any credentials.

ebyhr: Is fake-gcs-server for production use? Allowing empty credentials just for tests doesn't make sense to me.

patricklucas: > Does real GCS allow accessing without credentials?

Yes, buckets can be open to the public and allow anonymous access. For example, the datasets listed on this page: https://cloud.google.com/storage/docs/public-datasets

Someone might, for example, make an Iceberg table available to users without authentication, and I would argue Trino should support that.

This snippet demonstrates accessing one of those buckets with null credentials:

```kotlin
import com.google.cloud.storage.StorageOptions

fun main() {
  val storage = StorageOptions.newBuilder().build().service
  println("Credentials: ${storage.options.credentials}")
  val bucket = storage.get("gcp-public-data-arco-era5")
  println(bucket)
}
```

Output:

```
Credentials: null
Bucket{name=gcp-public-data-arco-era5}
```

However, a call to `GoogleCredentials.getApplicationDefault()` fails with an `IOException`.

> Is fake-gcs-server for production use?

No, it is not a complete implementation, but very useful for local/offline development and testing.

> Why don't you fix fake-gcs-server instead?

There's not really anything to "fix" in fake-gcs-server here. The issue is that Trino requires connectivity to Google's OAuth servers due to how it creates `Credentials` objects to pass to the SDK.

Trino is the only software with GCS integration I have used where it is not possible to leave credentials unconfigured, allowing the GCS client library to apply its default behavior.

patricklucas: Also, I would highlight that the GCS client library itself calls `GoogleCredentials.getApplicationDefault()` if no `Credentials` object is provided, meaning the behavior is nearly identical to what Trino does now.

The only difference is that the client library allows anonymous access by suppressing the `IOException` thrown when no credentials are found:

[`ServiceOptions.java`](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L347):
```java
  protected ServiceOptions(
      Class<? extends ServiceFactory<ServiceT, OptionsT>> serviceFactoryClass,
      Class<? extends ServiceRpcFactory<OptionsT>> rpcFactoryClass,
      Builder<ServiceT, OptionsT, ?> builder,
      ServiceDefaults<ServiceT, OptionsT> serviceDefaults) {
...
    credentials = builder.credentials != null ? builder.credentials : defaultCredentials();
...
  }
```

[`ServiceOptions.java`](https://github.com/googleapis/sdk-platform-java/blob/v2.53.0/java-core/google-cloud-core/src/main/java/com/google/cloud/ServiceOptions.java#L385-L391):
```java
  private static GoogleCredentials defaultCredentials() {
    try {
      return GoogleCredentials.getApplicationDefault();
    } catch (Exception ex) {
      return null;
    }
  }
```

Whereas Trino throws an exception:

[`GcsStorageFactory.java`](https://github.com/trinodb/trino/blob/475/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java#L98-L103):
```java
                    try {
                        return GoogleCredentials.getApplicationDefault();
                    }
                    catch (IOException e) {
                        throw new UncheckedIOException(e);
                    }
```

patricklucas: > ideally the PR would have an integration test showcasing the use-case without any credentials.

Thinking about it further, I don't think it's appropriate to add an integration test with fake-gcs-server here, since the purpose of this change is not to add support for it in particular. Instead, what I would do is add a unit test that verifies the behavior that if no credentials options are configured, the `Storage` instance returned from `GcsStorageFactory#create` has credentials equal to a default `Storage` instance:

```java
    @Test
    void testDefaultCredentials()
            throws IOException
    {
        Credentials expectedCredentials = StorageOptions.newBuilder().build().getCredentials();

        // No credentials options are set
        GcsFileSystemConfig config = new GcsFileSystemConfig();

        GcsStorageFactory storageFactory = new GcsStorageFactory(config);

        Credentials actualCredentials;
        try (Storage storage = storageFactory.create(ConnectorIdentity.ofUser("test"))) {
            actualCredentials = storage.getOptions().getCredentials();
        }
        catch (Exception e) {
        }

        assertThat(actualCredentials)
                .as("if no GCS credentials configured, should let default flow from the GCS client library")
                .isEqualTo(expectedCredentials);
    }
```

wendigo: Can we do the same and return null when IOException is catched? The proposed unit test looks fine.

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsStorageFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import com.google.cloud.storage.StorageOptions;
+import io.trino.spi.security.ConnectorIdentity;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+
+import java.io.IOException;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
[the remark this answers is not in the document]
patricklucas: We don't; I erroneously thought it was a convention from looking at other test classes. Fixed.

patricklucas:
CLA submitted.

Praveen2112 (commented):
What is the default mode for these Credentials ?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java
```diff
@@ -99,7 +99,8 @@ public Storage create(ConnectorIdentity identity)
                         return GoogleCredentials.getApplicationDefault();
                     }
                     catch (IOException e) {
-                        throw new UncheckedIOException(e);
+                        // This is consistent with the GCP SDK when no credentials are available in the environment
+                        return null;
```
Praveen2112, line 103: Or should we use `NoCredentials` ?

patricklucas:
> What is the default mode for these Credentials ? The default behaviour is to use the ApplicationDefaultCredentials right

The default behavior of the library as linked to above/in the original issue is to call `GoogleCredentials.getApplicationDefault()`, but importantly, to return null if it throws an exception (which signifies no application default credentials are available in the environment). This is different from the current behavior of Trino which calls that method, but propagates the exception.

Passing `NoCredentials` is a way for a user of the library to explicitly indicate that they do not want to authenticate. I don't think that Trino should _assume_ that in any case, but a possible route would be to add a configuration option to do so.

My preference is what I had originally—if no credentials options are set, simply fall back to the default behavior of the library by not calling `setCredentials` at all, rather than trying to emulate its behavior by calling `GoogleCredentials.getApplicationDefault()` ourselves.

patricklucas:
@cla-bot check

patricklucas:
@wendigo I submitted the CLA to cla@trino.io 25 days ago (19 May)—does it usually take this long, or is it worth checking with someone about getting that verified?

wendigo:
@martint can you check whether you have received @patricklucas CLA?

wendigo:
@patricklucas I can see that the last time CLAs were processed was 2.05: https://github.com/trinodb/cla/commits/master

patricklucas:
@cla-bot check

PR #25816, opened 2025-05-18
Workaround Jetty async thread hanging under contention

wendigo:
## Release notes

```markdown
## General
* Fix rare bug when server can hang under load ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-main/pom.xml
+++ b/core/trino-main/pom.xml
@@ -353,6 +353,11 @@
         </dependency>
 
         <dependency>
+            <groupId>org.eclipse.jetty</groupId>
+            <artifactId>jetty-io</artifactId>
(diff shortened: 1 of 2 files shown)


wendigo:
From the Jetty's javadoc:
```
A Jetty specialization of EOFException.
This is thrown by Jetty to distinguish between EOF received from the connection, vs and EOF thrown by some application talking to some other file/socket etc. The only difference in handling is that Jetty EOFs are logged less verbosely.
```

electrum:
Please change the title to be self-descriptive rather than referencing an issue

electrum (approved):

raunaqmorarka (approved):

PR #25657, opened 2025-04-24
Fix failure with specifying format for date partition projection in Hive connector

chenjian2664:
## Description

Closes #25642 

In the Hive connector, the partition formats for date and timestamp types were previously fixed. For example, the date type used HiveUtils.HIVE_DATE_FORMATTER and HiveWriteUtils.HIVE_DATE_FORMATTER for reading and writing partition values.
However, partition projection supports user-specified formats, which led to inconsistencies and failures when custom formats were used, especially during parsing partition values at read or write time.

This PR addresses the issue by delegating the reading and writing of partition values to the partition projection mechanism when the column is projected.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
## Hive
* Fix failure with specifying format for date partition projection in Hive connector. ({issue}`25642`)
```

PR #25817, opened 2025-05-18
Fix reading specifying format for date partition projection in Hive connector

chenjian2664:
## Description

Make the partition value Hive format compatible during partition generation

Alternative for #25657 

## Release notes

```markdown
## Hive
* Fix reading specifying format for date partition projection in Hive connector. ({issue}`issuenumber`)
```


plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -52,6 +53,31 @@ public PartitionProjection(Optional<String> storageLocationTemplate, Map<String,
         this.columnProjections = ImmutableMap.copyOf(requireNonNull(columnProjections, "columnProjections is null"));
     }
 
+    // TODO: support writing partition projection
+    public boolean allowWrite()
```
pettyjamesm, line 57: Control flow here doesn't handle the possibility of a `storageLocationTemplate` being present that doesn't match Hive's path name assumptions (this can happen with any projection type).

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -52,6 +53,31 @@ public PartitionProjection(Optional<String> storageLocationTemplate, Map<String,
...
+    {
+        for (Projection projection : columnProjections.values()) {
+            // DateProjection may contain a user-defined format that is incompatible with Hive.
+            // Currently, we only support writing partitions in Hive's default date format.
+            // Allowing custom date formats in partition projection can lead to inconsistencies
+            // between reading and writing of the projection partition.
+            // To ensure compatibility and avoid incorrect query results, disable writing for now.
+            if (projection instanceof DateProjection dateProjection) {
+                return dateProjection.allowWrite();
```
pettyjamesm, line 66: As written, this logic would return the result of `dateProjection.allowWrite()` on the first encountered `DateProjection` and would not check any remaining date projections.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
chenjian2664: updated

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
```diff
@@ -1719,12 +1722,13 @@ public void finishStatisticsCollection(ConnectorSession session, ConnectorTableH
                     .map(HiveColumnHandle::getType)
                     .collect(toImmutableList());
 
+            Optional<PartitionProjection> partitionProjection = getPartitionProjectionFromHiveTable(handle);
```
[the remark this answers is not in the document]
chenjian2664: Where is the "unpartitioned tables are handled" ?
You are right, it seems make no sense to handle in the "finish" stage, I am going to put the the Optional.empty here, as the partition will mismatch at the begining.

metadaddy:
Hi folks - any progress on this one? I'm writing an article explaining how to use Trino to query access logs, and I've had to include a caveat explaining that partition projection currently doesn't work, so time-based queries scan all rows, rather than just those included by the query conditions.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -96,20 +149,22 @@ private Partition buildPartitionObject(Table table, String partitionName)
...
                 .withStorage(storage -> storage
                         .setStorageFormat(table.getStorage().getStorageFormat())
-                        .setLocation(storageLocationTemplate
-                                .map(template -> expandStorageLocationTemplate(
-                                        template,
-                                        table.getPartitionColumns().stream()
-                                                .map(Column::getName).collect(Collectors.toList()),
-                                        partitionValues))
-                                .orElseGet(() -> getPartitionLocation(table.getStorage().getLocation(), partitionName)))
+                        .setLocation(expandStorageLocationTemplate(
```
pettyjamesm, line 168: The logic to expand the template should probably be an instance field on `PartitionProjection` since you'd want to avoid doing that for each partition individually.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -123,7 +178,7 @@ private static String expandStorageLocationTemplate(String template, List<String
         while (matcher.find()) {
             String columnPlaceholder = matcher.group(1);
             String columnName = columnPlaceholder.substring(2, columnPlaceholder.length() - 1);
-            matcher.appendReplacement(location, partitionValues.get(partitionColumns.indexOf(columnName)));
+            matcher.appendReplacement(location, escapePathName(partitionValues.get(partitionColumns.indexOf(columnName))));
```
pettyjamesm: I don't think the partition values inserted should be escaped, at least it doesn't appear to be that way in the code that I'm seeing. Otherwise the date format values wouldn't allow path separators in there.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -52,6 +53,46 @@ public PartitionProjection(Optional<String> storageLocationTemplate, Map<String,
...
+    {
+        if (storageLocationTemplate.isPresent()) {
+            String template = storageLocationTemplate.get();
+            if (template.endsWith("/")) {
+                template = template.substring(0, template.length() - 1);
+            }
+            List<String> columns = ImmutableList.copyOf(columnProjections.keySet());
+            String partition = columns.stream()
+                    .map(column -> column + "=${" + column + "}")
+                    .collect(Collectors.joining("/"));
```
pettyjamesm: This logic looks suspiciously similar to the fallback template generation logic for projections, maybe we should call that code directly instead of reimplementing it locally.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/Projection.java
```diff
@@ -22,4 +22,9 @@ sealed interface Projection
         permits DateProjection, EnumProjection, InjectedProjection, IntegerProjection
 {
     List<String> getProjectedValues(Optional<Domain> partitionValueFilter);
+
+    default String transformValue(String value)
```
pettyjamesm: The name here should probably imply that we're specifically "transforming" the value for use inside of the location template so that the intention is more clear.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
chenjian2664: Yeah, that’s what I wanted to discuss. 
We could choose not to escape the path when a template is provided, but my concern is that object storage services like S3 might reject certain characters in unescaped paths, considering the existing behavior in the Hive connector, where paths are already escaped, aligning with that behavior would be safer IMO

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/DateProjection.java
```diff
@@ -152,20 +163,77 @@ public List<String> getProjectedValues(Optional<Domain> partitionValueFilter)
...
+        }
+
+        if (columnType instanceof TimestampType timestampType && timestampType.isShort()) {
+            long epochMicros = parseHiveTimestamp(value);
+            long epochSeconds = floorDiv(epochMicros, MICROSECONDS_PER_SECOND);
+            int nanosOfSecond = floorMod(epochMicros, MICROSECONDS_PER_SECOND) * NANOSECONDS_PER_MICROSECOND;
+            return LocalDateTime.ofEpochSecond(epochSeconds, nanosOfSecond, ZoneOffset.UTC).format(dateFormat);
+        }
+
+        return value;
```
pettyjamesm, line 184: Probably throw an exception here since this path should be unreachable

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
chenjian2664: As talked offline, we are talking about the code: `storageLocationTemplate.orElseGet(() -> getPartitionLocation(table.getStorage().getLocation(), partitionColumns))`
However, this line requires `tableLocation`, which is not easily get from the caller method `getPartitionProjectionHiveTableProperties` when we construct the `PartitionProjection`, I am planning to leave it as is, please suggests if you have other idea

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/DateProjection.java
chenjian2664: We can't, see test case `BaseTestHiveOnDataLake#testDatePartitionProjectionOnVarcharColumnWithHoursInterval`, there is a varchar type Date partition projection.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
chenjian2664: extracted `toPartitionLocationTemplate` method

chenjian2664:
@pettyjamesm comments addressed, PTAL

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/Projection.java
```diff
@@ -22,4 +22,9 @@ sealed interface Projection
         permits DateProjection, EnumProjection, InjectedProjection, IntegerProjection
 {
     List<String> getProjectedValues(Optional<Domain> partitionValueFilter);
+
+    default String transforming(String value)
```
pettyjamesm: Maybe rename to: `toPartitionLocationTemplateValue` to make the purpose of the transformed result more clear.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -87,7 +126,20 @@ public Map<String, Optional<Partition>> getProjectedPartitionsByNames(Table tabl
 
     private Partition buildPartitionObject(Table table, String partitionName)
     {
+        List<String> partitionColumns = table.getPartitionColumns().stream()
```
pettyjamesm, line 148: Maybe inline `table.getPartitionColumns().get(i)` into the loop below to avoid materializing a list copy on each invocation (remember that partition projection tables are expected to have high cardinality, so extra costs here add up).

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -87,7 +126,20 @@ public Map<String, Optional<Partition>> getProjectedPartitionsByNames(Table tabl
...
+                .map(Column::getName)
+                .collect(toImmutableList());
+
         List<String> partitionValues = toPartitionValues(partitionName);
+        ImmutableList.Builder<String> partitions = ImmutableList.builder();
```
pettyjamesm: Nit: mabye rename to `partitionTemplateValues` and use `ImmutableList.builderWithExpectedSize(partitionValues.size())`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -52,6 +53,44 @@ public PartitionProjection(Optional<String> storageLocationTemplate, Map<String,
         this.columnProjections = ImmutableMap.copyOf(requireNonNull(columnProjections, "columnProjections is null"));
     }
 
+    // TODO: support writing partition projection
+    public boolean allowWrite()
+    {
+        if (storageLocationTemplate.isPresent()) {
```
pettyjamesm, line 95: I think this whole check for whether we can support writing boils down to `storageTemplateLocation.isEmpty()`. In the event that the custom template is provided but _happens_ to match the hive default template construction, the values would still need to be escaped in order to be created with the correct path.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
chenjian2664: The column names are also used later in `expandStorageLocationTemplate`, which requires accessing values by name. So we need to compute the names regardless—we can't avoid it

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
chenjian2664: You are right, the logic for the template relates part has changed after this code has created.

We should allow the storage template is compatible with hive -- escape the value for the case as well.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
```diff
@@ -2164,6 +2166,14 @@ private HiveInsertTableHandle beginInsertOrMerge(ConnectorSession session, Conne
                 throw new TrinoException(NOT_SUPPORTED, format("%s Hive table with value of %s property greater than 1 is not supported", description, SKIP_HEADER_COUNT_KEY));
             }
         });
+
+        if (partitionProjectionEnabled) {
+            Optional<PartitionProjection> partitionProjection = getPartitionProjectionFromTable(table, typeManager);
+            if (!partitionProjection.map(PartitionProjection::allowWrite).orElse(true)) {
+                throw new TrinoException(NOT_SUPPORTED, "Hive connector does not support writing such partition with projection");
```
pettyjamesm: Let's see if we can add a more helpful error message here that explains why the operation isn't supported (note how above operations use `description` to indicate whether this is a disallowed insert or merge). Maybe `boolean PartitionProjection#allowWrite()` needs to be more of a `void PartitionProjection#checkWriteSupported(String description)` and throw an exception with a more detailed explanation when rejecting the write.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -87,7 +133,19 @@ public Map<String, Optional<Partition>> getProjectedPartitionsByNames(Table tabl
...
+        List<String> partitionColumns = table.getPartitionColumns().stream()
+                .map(Column::getName)
+                .collect(toImmutableList());
+
         List<String> partitionValues = toPartitionValues(partitionName);
+        ImmutableList.Builder<String> partitionTemplateValues = ImmutableList.builderWithExpectedSize(partitionValues.size());
+        for (int i = 0; i < partitionColumns.size(); i++) {
+            String partitionColumn = partitionColumns.get(i);
+            String partitionValue = columnProjections.get(partitionColumn).toPartitionLocationTemplateValue(partitionValues.get(i));
+            // Don't escape partition value if storage location template is defined
```
pettyjamesm: Nit: expand on this comment to mention that when the template is defined but equivalent to the hive default we still escape.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/projection/PartitionProjection.java
```diff
@@ -44,14 +46,58 @@ public final class PartitionProjection
     private static final Pattern PROJECTION_LOCATION_TEMPLATE_PLACEHOLDER_PATTERN = Pattern.compile("(\\$\\{[^}]+\\})");
 
     private final Optional<String> storageLocationTemplate;
+    private final boolean templateCompatibleWithHive;
```
pettyjamesm: Nit: maybe name this field after the specific action taken, e.g.: `escapeInjectedColumnValues`

pettyjamesm (commented):
Mostly looks good to me, test coverage looks good. Some code style comments / nits.

pettyjamesm (approved):
The changes here look good to me, thanks for taking this over the finish line @chenjian2664!

chenjian2664:
@pettyjamesm I really appreciate your guidance and the time you took to assist me

metadaddy:
Thanks for your work on this, @chenjian2664. I'll test it against the Backblaze B2 Bucket Access Log and let you know how it works out.

PR #25826, opened 2025-05-20
Increase delta.max-split-size to 128MB

raunaqmorarka:
## Description
The default row group size for parquet files in delta is 128MB. 
The default for split size should match that. 
The previous smaller value of 64MB generated unnecessary splits.

## Release notes

```markdown
## Delta Lake
* Increase delta.max-split-size to 128MB. ({issue}`25826`)
```


losipiuk (approved):

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeConfig.java
```diff
@@ -68,7 +68,7 @@ public class DeltaLakeConfig
     private int domainCompactionThreshold = 1000;
     private int maxOutstandingSplits = 1_000;
     private int maxSplitsPerSecond = Integer.MAX_VALUE;
-    private DataSize maxSplitSize = DataSize.of(64, MEGABYTE);
+    private DataSize maxSplitSize = DataSize.of(128, MEGABYTE);
```
ebyhr, line 71: Could you update `delta-lake.md`? It has the default value.

PR #25843, opened 2025-05-21
Allow configuring heap relative to maximum

wendigo:
## Release notes

```markdown
## General
* Allow configuring `query.max-memory-per-node` and `memory.heap-headroom-per-node` relative to maximum heap size ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/HeapSizeParser.java
+++ b/core/trino-main/src/main/java/io/trino/execution/HeapSizeParser.java
@@ -1,0 +1,64 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.execution;
+
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Suppliers;
+import io.airlift.units.DataSize;
+
+import java.util.function.Supplier;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Double.parseDouble;
+import static java.util.Objects.requireNonNull;
+
+public class HeapSizeParser
+{
+    private static final String RELATIVE_SUFFIX = "%";
+
+    // Memoize the max heap memory to avoid calling Runtime.getRuntime().maxMemory() multiple times
+    // and to ensure that the value will not change over JVM lifetime.
+    private static final Supplier<Long> AVAILABLE_HEAP_MEMORY = Suppliers.memoize(Runtime.getRuntime()::maxMemory);
+
+    public static final HeapSizeParser DEFAULT = new HeapSizeParser(AVAILABLE_HEAP_MEMORY);
+
+    private final Supplier<Long> maxHeapMemory;
+
+    @VisibleForTesting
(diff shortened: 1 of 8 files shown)


wendigo:
I'll add a documentation for these as well.

core/trino-main/src/main/java/io/trino/memory/NodeMemoryConfig.java
```diff
@@ -41,9 +40,9 @@ public DataSize getMaxQueryMemoryPerNode()
     }
 
     @Config("query.max-memory-per-node")
```
raunaqmorarka, line 42: Please update `properties-resource-management.md` as well

core/trino-main/src/main/java/io/trino/execution/HeapSizeParser.java
```diff
@@ -0,0 +1,62 @@
...
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Double.parseDouble;
+import static java.util.Objects.requireNonNull;
+
+public class HeapSizeParser
+{
+    private static final String RELATIVE_SUFFIX = "%";
+    private static final Supplier<Long> AVAILABLE_HEAP_MEMORY = Runtime.getRuntime()::maxMemory;
```
raunaqmorarka, line 29: Making it a supplier instead of static constant may be problematic.
We have a code comment in DeltaLakeConfig
```
// Runtime.getRuntime().maxMemory() is not 100% stable and may return slightly different value over JVM lifetime. We use
// constant so default configuration for cache size is stable.
```

wendigo: This is stable as the configuration is resolved once during startup. Since this is part of the `trino-main` it can't be used in the connector.

wendigo: I'll memoize the value so it's computed once. Once Stable values are introduced in Java we will be able to use it instead (which is a nicer thing as they are constant folded)

wendigo: Memoized + documentation added

core/trino-main/src/main/java/io/trino/memory/NodeMemoryConfig.java
wendigo: Updated

losipiuk (approved):

core/trino-main/src/main/java/io/trino/execution/HeapSizeParser.java
```diff
@@ -0,0 +1,66 @@
...
+    public DataSize parse(String value)
+    {
+        long maxHeapMemory = this.maxHeapMemory.get();
+        checkState(maxHeapMemory > 0, "maxHeapMemory must be positive");
+
+        if (value.endsWith(RELATIVE_SUFFIX)) {
+            double multiplier = parseDouble(value.substring(0, value.length() - RELATIVE_SUFFIX.length()).trim()) / 100.0;
+            return checkHeapSizeMemory(DataSize.ofBytes(Math.round(maxHeapMemory * multiplier)).succinct(), maxHeapMemory);
+        }
+        else {
```
raunaqmorarka, line 54: nit: else is unnecessary

core/trino-main/src/main/java/io/trino/execution/HeapSizeParser.java
```diff
@@ -0,0 +1,66 @@
...
+            return checkHeapSizeMemory(DataSize.valueOf(value), maxHeapMemory);
+        }
+    }
+
+    private static DataSize checkHeapSizeMemory(DataSize heapSize, long maxHeapMemory)
+    {
+        System.out.println("Data size is " + heapSize + " max is " + maxHeapMemory);
```
raunaqmorarka, line 61: Is the `println` intentional ?

raunaqmorarka (approved):

core/trino-main/src/main/java/io/trino/execution/HeapSizeParser.java
wendigo: oh no :)

PR #25847, opened 2025-05-21
Reduce memory usage of remove_orphan_files procedure

grantatspothero:
## Description
Reduces memory usage of coordinator during iceberg remove orphan files procedure for tables with large numbers of snapshots. 

See comment for description of fix:
```
    /**
     * Use instead of loadAllManifestsFromSnapshot when loading manifests from multiple distinct snapshots
     * Each BaseSnapshot object caches manifest files separately, so loading manifests from multiple distinct snapshots
     * results in O(num_snapshots^2) copies of the same manifest file metadata in memory
     */
```

To show decreased memory usage:

1. Created a table with 1000 snapshots/1000 manifest files and ran orphan files
```
    @Test
    public void testRemoveOrphanFilesOnTableWithManySnapshots()
    {
        String tableName = "test_remove_orphan_files_on_table_with_many_snapshots_" + randomNameSuffix();
        String fullyQualifiedTableName = "iceberg.%s.%s".formatted(getSession().getSchema().orElseThrow(), tableName);

        // Create table with 10K snapshots
        assertUpdate("SET SESSION iceberg.merge_manifests_on_write = false");
        assertUpdate("CREATE TABLE %s (a) AS VALUES 0".formatted(fullyQualifiedTableName), 1);
        for (int i = 1; i < 1000; i++) {
            assertUpdate("INSERT INTO %s VALUES %d".formatted(fullyQualifiedTableName, i), 1);
        }

        assertQuerySucceeds("ALTER TABLE %s EXECUTE REMOVE_ORPHAN_FILES".formatted(fullyQualifiedTableName));
        assertThat(computeScalar("SELECT count(*) FROM %s".formatted(fullyQualifiedTableName))).isEqualTo(1000L);
        assertUpdate("DROP TABLE %s".formatted(fullyQualifiedTableName));
    }
```
2. Attempted to take heapdump, but it did not show a difference in memory usage attributed to the caching of manifest files on the `BaseSnapshot` object. Root cause was incorrect attribution of memory usage of the List<ManifestFile>, because `BaseSnapshot#cacheManifests` calls `ManifestLists.read` which returns a LinkedList. The heapdump memory attribution logic was not correctly attributing all the objects in the LinkedList to `BaseSnapshot`, only the first node. 
3. Manually inspected `table.snapshots()` in a debugger to confirm the manifest caching was not occurring. 

(before change)
Screenshot 2025-05-21 at 3 03 13 PM
(after change)
Screenshot 2025-05-21 at 3 06 21 PM

## Release notes

```markdown
## Iceberg
* Reduce memory usage of remove_orphan_files procedure. ({issue}`25847`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -145,6 +145,7 @@ import org.apache.iceberg.DeleteFiles;
 import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.FileMetadata;
 import org.apache.iceberg.FileScanTask;
+import org.apache.iceberg.IcebergManifestUtils;
 import org.apache.iceberg.IsolationLevel;
 import org.apache.iceberg.ManifestFile;
 import org.apache.iceberg.ManifestFiles;
@@ -2260,11 +2261,18 @@ public class IcebergMetadata
         ImmutableSet.Builder<String> validDataFileNames = ImmutableSet.builder();
 
         for (Snapshot snapshot : table.snapshots()) {
-            if (snapshot.manifestListLocation() != null) {
-                validMetadataFileNames.add(fileName(snapshot.manifestListLocation()));
+            String manifestListLocation = snapshot.manifestListLocation();
+            List<ManifestFile> allManifests;
+            if (manifestListLocation != null) {
+                validMetadataFileNames.add(fileName(manifestListLocation));
+                allManifests = loadAllManifestsFromManifestList(table, manifestListLocation);
+            }
+            else {
+                // This is to maintain support for V1 tables which have embedded manifest lists
+                allManifests = loadAllManifestsFromSnapshot(table, snapshot);
             }
 
-            for (ManifestFile manifest : loadAllManifestsFromSnapshot(table, snapshot)) {
+            for (ManifestFile manifest : allManifests) {
                 if (!processedManifestFilePaths.add(manifest.path())) {
                     // Already read this manifest
                     continue;
@@ -3463,6 +3471,21 @@ public class IcebergMetadata
     {
         try {
             return snapshot.allManifests(icebergTable.io());
+        }
+        catch (NotFoundException | UncheckedIOException e) {
+            throw new TrinoException(ICEBERG_INVALID_METADATA, "Error accessing manifest file for table %s".formatted(icebergTable.name()), e);
+        }
+    }
+
+    /**
+     * Use instead of loadAllManifestsFromSnapshot when loading manifests from multiple distinct snapshots
+     * Each BaseSnapshot object caches manifest files separately, so loading manifests from multiple distinct snapshots
+     * results in O(num_snapshots^2) copies of the same manifest file metadata in memory
+     */
+    private static List<ManifestFile> loadAllManifestsFromManifestList(Table icebergTable, String manifestListLocation)
+    {
+        try {
+            return IcebergManifestUtils.read(icebergTable.io(), manifestListLocation);
         }
         catch (NotFoundException | UncheckedIOException e) {
             throw new TrinoException(ICEBERG_INVALID_METADATA, "Error accessing manifest file for table %s".formatted(icebergTable.name()), e);
--- a/plugin/trino-iceberg/src/main/java/org/apache/iceberg/IcebergManifestUtils.java
+++ b/plugin/trino-iceberg/src/main/java/org/apache/iceberg/IcebergManifestUtils.java
@@ -1,0 +1,31 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.iceberg;
+
+import org.apache.iceberg.io.FileIO;
+
+import java.util.List;
+
+public class IcebergManifestUtils
+{
+    private IcebergManifestUtils() {}
+
+    public static List<ManifestFile> read(FileIO fileIO, String manifestListLocation)
+    {
+        // Avoid using snapshot.allManifests() when processing multiple snapshots,
+        // as each Snapshot instance internally caches `org.apache.iceberg.BaseSnapshot.allManifests`
+        // and leads to high memory usage
+        return ManifestLists.read(fileIO.newInputFile(manifestListLocation));
+    }
+}


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3736,7 +3756,7 @@ public Optional<ConnectorOutputMetadata> finishRefreshMaterializedView(
                         icebergTable.schema().findType(field.sourceId())))
                 .toArray(Type[]::new);
 
-        AppendFiles appendFiles = transaction.newFastAppend();
+        AppendFiles appendFiles = isMergeManifestsOnWrite(session) ? transaction.newAppend() : transaction.newFastAppend();
```
raunaqmorarka, line 3759: Is this change intentional ? Looks unrelated to remove_orphan_files

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3460,6 +3465,21 @@ private static List<ManifestFile> loadAllManifestsFromSnapshot(Table icebergTabl
         }
     }
 
+    /**
+     * Use instead of loadAllManifestsFromSnapshot when loading manifests from multiple distinct snapshots
+     * Each BaseSnapshot object caches manifest files separately, so loading manifests from multiple distinct snapshots
+     * results in O(num_snapshots^2) copies of the same manifest file metadata in memory
+     */
+    private static List<ManifestFile> loadAllManifestsFromManifestList(Table icebergTable, String manifestListLocation)
```
raunaqmorarka, line 3473: Can this move to ManifestUtils ?

plugin/trino-iceberg/src/main/java/org/apache/iceberg/ManifestUtils.java
```diff
@@ -0,0 +1,29 @@
...
+package org.apache.iceberg;
+
+import org.apache.iceberg.io.FileIO;
+
+import java.util.List;
+
+public class ManifestUtils
+{
+    private ManifestUtils() {}
+
```
raunaqmorarka, line 23: `loadAllManifestsFromSnapshot` also probably belongs here

raunaqmorarka (commented):
minor comments, lgtm

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -4179,4 +4199,18 @@ private static TableStatistics mergeColumnStatistics(TableStatistics currentStat
         newStats.getColumnStatistics().forEach(statisticsBuilder::setColumnStatistics);
         return statisticsBuilder.build();
     }
+
+    /**
+     * Table.snapshots() returns Iterable but the actual implementation returned is a List
+     * snapshot.allManifests() caches manifests, so by keeping a List around GC is prevented from collecting any in memory manifests which leaks memory
+     * Workaround: use Queue, Queue#poll will unlink old Snapshots allowing for GC during traversal
+     */
+    private static Queue<Snapshot> getAllSnapshots(Table icebergTable)
```
ebyhr, line 4208: This method seems unused.

grantatspothero: Whoops meant to remove.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
grantatspothero: Note the `org.apache.iceberg` package. 

The intention of this class was to utilize the package private method `ManifestLists.read`.

raunaqmorarka: Is it possible to restrict the fix to V2 iceberg tables while we look for a solution for V1 tables in https://apache-iceberg.slack.com/archives/C03LG1D563F/p1747921709647279 ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2251,12 +2255,13 @@ private void removeOrphanFiles(Table table, ConnectorSession session, SchemaTabl
         ImmutableSet.Builder<String> validMetadataFileNames = ImmutableSet.builder();
         ImmutableSet.Builder<String> validDataFileNames = ImmutableSet.builder();
 
-        for (Snapshot snapshot : table.snapshots()) {
-            if (snapshot.manifestListLocation() != null) {
-                validMetadataFileNames.add(fileName(snapshot.manifestListLocation()));
+        ArrayList<String> manifestListLocations = newArrayList(Iterables.transform(table.snapshots(), Snapshot::manifestListLocation));
```
chenjian2664:
```suggestion
        Iterable<String> manifestListLocations = Iterables.transform(table.snapshots(), Snapshot::manifestListLocation);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: @grantatspothero i've pushed changes here to restrict the fix to v2 tables, PTAL

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2260,11 +2261,18 @@ private void removeOrphanFiles(Table table, ConnectorSession session, SchemaTabl
         ImmutableSet.Builder<String> validDataFileNames = ImmutableSet.builder();
 
         for (Snapshot snapshot : table.snapshots()) {
-            if (snapshot.manifestListLocation() != null) {
+            String manifestListLocation = snapshot.manifestListLocation();
+            List<ManifestFile> allManifests;
+            if (manifestListLocation != null) {
                 validMetadataFileNames.add(fileName(snapshot.manifestListLocation()));
```
tbaeg: Reuse `manifestListLocation` variable here.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
tbaeg: Not something we have to do this PR, but can we start moving these types of utilities out into a separate utility class? We have lots of utilities throughout with similar/close functionality repeated, and I imagine most occasional contributors (like myself) don't even know they exist because they are placed in the calling class as a private method.

plugin/trino-iceberg/src/main/java/org/apache/iceberg/ManifestUtils.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.iceberg;
+
+import org.apache.iceberg.io.FileIO;
+
+import java.util.List;
+
+public class ManifestUtils
```
tbaeg: It might be worth prefixing the class as `IcebergManifestUtils`/`TrinoManifestUtils` or something to differentiate it from an iceberg provided class. From my prospective as an occasional contributor, this type of convention helps with cognitive load, especially since it's in the same package path.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2260,11 +2261,18 @@ private void removeOrphanFiles(Table table, ConnectorSession session, SchemaTabl
         ImmutableSet.Builder<String> validDataFileNames = ImmutableSet.builder();
 
         for (Snapshot snapshot : table.snapshots()) {
-            if (snapshot.manifestListLocation() != null) {
-                validMetadataFileNames.add(fileName(snapshot.manifestListLocation()));
+            String manifestListLocation = snapshot.manifestListLocation();
+            List<ManifestFile> allManifests;
+            if (manifestListLocation != null) {
```
grantatspothero, line 2266: I do not think we have tests using v1 iceberg format for remove orphan files. We should add at least one as a sanity check.

grantatspothero (commented):
This is fine as it fixes a real issue but I'm not happy with adding separate code paths for different iceberg spec versions.

For context I tried asking iceberg maintainers for advice on what to do here:
https://apache-iceberg.slack.com/archives/C03LG1D563F/p1747921709647279

Suggestion was to implement a per table metadata cache of manifest files. I spent a few hours doing this and then had to pivot to other things. If someone is familiar with the avro reader iceberg uses would love to pair with them.

raunaqmorarka:
> This is fine as it fixes a real issue but I'm not happy with adding separate code paths for different iceberg spec versions.

I don't see the point of doing extra work for V1 tables, why would anyone still be using those ?

grantatspothero:
> > This is fine as it fixes a real issue but I'm not happy with adding separate code paths for different iceberg spec versions.
> 
> I don't see the point of doing extra work for V1 tables, why would anyone still be using those ?

This would break remove orphan files for v1 iceberg tables, the trino docs say `The connector supports Apache Iceberg table spec versions 1 and 2.`: https://trino.io/docs/current/connector/iceberg.html

raunaqmorarka:
> > > This is fine as it fixes a real issue but I'm not happy with adding separate code paths for different iceberg spec versions.
> > 
> > 
> > I don't see the point of doing extra work for V1 tables, why would anyone still be using those ?
> 
> This would break remove orphan files for v1 iceberg tables, the trino docs say `The connector supports Apache Iceberg table spec versions 1 and 2.`: https://trino.io/docs/current/connector/iceberg.html

I'm not talking about dropping support, I'm talking about doing extra work to have the same code path or same fix for V1.
The intent of the current change is to leave everything as it was for V1. If someone comes with an issue for V1, I'd rather tell them to update to V2 than write more code for V1.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: This is a special case of V1 tables where `write.manifest-lists.enabled` is set to `false`. However, the iceberg library removed the ability to write embedded manifest list a few years ago (https://github.com/apache/iceberg/pull/5773). So we don't have a way to produce such data in the tests.
It might be possible to generate such data from an old version and test on that pre-generated dataset, but I think that is overkill at the moment. The fallback logic looks safe enough to me.

raunaqmorarka (approved):

PR #25856, opened 2025-05-23
Refactor `TransactionLogTail#loadNewTail` method

chenjian2664:
Refactor `TransactionLogTail#loadNewTail` to support loading transaction log to based on param `startVersion` and `endVersion` conditional. Previously, the method only supported forward traversal from head to tail, which posed a risk when early transaction logs were missing -- maybe due to cleanup. This refactor enables both forward and backward traversal, increasing robustness and flexibility in tail loading logic.

Additionally, remove `BaseTransactionsTable#loadNewTailBackward` and use the enhanced loadNewTail method.

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/BaseTransactionsTable.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/BaseTransactionsTable.java
@@ -13,7 +13,6 @@
  */
 package io.trino.plugin.deltalake;
 
-import com.google.common.collect.ImmutableList;
 import io.airlift.units.DataSize;
 import io.trino.filesystem.TrinoFileSystem;
 import io.trino.filesystem.TrinoFileSystemFactory;
@@ -21,7 +20,6 @@ import io.trino.plugin.deltalake.metastore.DeltaMetastoreTable;
 import io.trino.plugin.deltalake.transactionlog.TableSnapshot;
 import io.trino.plugin.deltalake.transactionlog.Transaction;
 import io.trino.plugin.deltalake.transactionlog.TransactionLogAccess;
-import io.trino.plugin.deltalake.transactionlog.TransactionLogEntries;
 import io.trino.plugin.deltalake.util.PageListBuilder;
 import io.trino.spi.Page;
 import io.trino.spi.TrinoException;
@@ -43,10 +41,10 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.stream.IntStream;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.MoreCollectors.onlyElement;
 import static io.trino.plugin.deltalake.DeltaLakeErrorCode.DELTA_LAKE_INVALID_SCHEMA;
-import static io.trino.plugin.deltalake.transactionlog.TransactionLogUtil.getTransactionLogDir;
-import static io.trino.plugin.deltalake.transactionlog.checkpoint.TransactionLogTail.getEntriesFromJson;
+import static io.trino.plugin.deltalake.transactionlog.checkpoint.TransactionLogTail.loadNewTail;
 import static java.util.Objects.requireNonNull;
 
 public abstract class BaseTransactionsTable
@@ -140,7 +138,8 @@ public abstract class BaseTransactionsTable
         TrinoFileSystem fileSystem = fileSystemFactory.create(session);
         PageListBuilder pagesBuilder = PageListBuilder.forTable(tableMetadata);
         try {
-            List<Transaction> transactions = loadNewTailBackward(fileSystem, table.location(), startVersionExclusive, endVersionInclusive.get()).reversed();
+            checkArgument(endVersionInclusive.isPresent(), "endVersionInclusive must be present");
+            List<Transaction> transactions = loadNewTail(fileSystem, table.location(), startVersionExclusive, endVersionInclusive, DataSize.ofBytes(0)).getTransactions();
             return new FixedPageSource(buildPages(session, pagesBuilder, transactions, fileSystem));
         }
         catch (TrinoException e) {
@@ -149,39 +148,6 @@ public abstract class BaseTransactionsTable
         catch (IOException | RuntimeException e) {
             throw new TrinoException(DELTA_LAKE_INVALID_SCHEMA, "Error getting commit info entries from " + table.location(), e);
         }
-    }
-
-    // Load a section of the Transaction Log JSON entries. Optionally from a given end version (inclusive) through an start version (exclusive)
-    private static List<Transaction> loadNewTailBackward(
-            TrinoFileSystem fileSystem,
-            String tableLocation,
-            Optional<Long> startVersion,
-            long endVersion)
-            throws IOException
-    {
-        ImmutableList.Builder<Transaction> transactionsBuilder = ImmutableList.builder();
-        String transactionLogDir = getTransactionLogDir(tableLocation);
-
-        long version = endVersion;
-        long entryNumber = version;
-        boolean endOfHead = false;
-
-        while (!endOfHead) {
-            Optional<TransactionLogEntries> results = getEntriesFromJson(entryNumber, transactionLogDir, fileSystem, DataSize.of(0, DataSize.Unit.BYTE));
-            if (results.isPresent()) {
-                transactionsBuilder.add(new Transaction(version, results.get()));
-                version = entryNumber;
-                entryNumber--;
-            }
-            else {
-                // When there is a gap in the transaction log version, indicate the end of the current head
-                endOfHead = true;
-            }
-            if ((startVersion.isPresent() && version == startVersion.get() + 1) || entryNumber < 0) {
-                endOfHead = true;
-            }
-        }
-        return transactionsBuilder.build();
     }
 
     protected abstract List<Page> buildPages(ConnectorSession session, PageListBuilder pagesBuilder, List<Transaction> transactions, TrinoFileSystem fileSystem)
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
@@ -19,6 +19,7 @@ import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystem;
 import io.trino.filesystem.TrinoInputFile;
 import io.trino.plugin.deltalake.transactionlog.DeltaLakeTransactionLogEntry;
+import io.trino.plugin.deltalake.transactionlog.MissingTransactionLogException;
 import io.trino.plugin.deltalake.transactionlog.Transaction;
 import io.trino.plugin.deltalake.transactionlog.TransactionLogEntries;
 
@@ -31,6 +32,7 @@ import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.airlift.slice.SizeOf.SIZE_OF_LONG;
 import static io.airlift.slice.SizeOf.estimatedSizeOf;
 import static io.airlift.slice.SizeOf.instanceSize;
+import static io.trino.plugin.deltalake.transactionlog.TransactionLogUtil.getTransactionLogDir;
 import static io.trino.plugin.deltalake.transactionlog.TransactionLogUtil.getTransactionLogJsonEntryPath;
 import static java.util.Objects.requireNonNull;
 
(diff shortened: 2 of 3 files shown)


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,118 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+            String tableLocation,
+            long version,
+            long startVersion,
+            Optional<Long> endVersion,
+            DataSize transactionLogMaxCachedFileSize)
+            throws IOException
+    {
+        ImmutableList.Builder<Transaction> entriesBuilder = ImmutableList.builder();
+        long entryNumber = startVersion;
+        checkArgument(endVersion.isEmpty() || entryNumber <= endVersion.get(), "Invalid start/end versions: %s, %s", startVersion, endVersion);
```
raunaqmorarka: entryNumber -> startVersion

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,118 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+        boolean endOfTail = false;
+        while (!endOfTail) {
+            Optional<TransactionLogEntries> results = getEntriesFromJson(entryNumber, transactionLogDir, fileSystem, transactionLogMaxCachedFileSize);
+            if (results.isPresent()) {
+                entriesBuilder.add(new Transaction(entryNumber, results.get()));
+                version = entryNumber;
+                entryNumber++;
+            }
+            else {
+                if (endVersion.isPresent() && entryNumber > startVersion) {
```
raunaqmorarka: Why doesn't `entryNumber == startVersion` case generate exception here ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,118 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+
+        while (!endOfHead) {
+            Optional<TransactionLogEntries> results = getEntriesFromJson(entryNumber, transactionLogDir, fileSystem, transactionLogMaxCachedFileSize);
+            if (results.isPresent()) {
+                transactionsBuilder.add(new Transaction(entryNumber, results.get()));
+                version = entryNumber;
+                entryNumber--;
+            }
+            else {
+                // When there is a gap in the transaction log version, indicate the end of the current head
```
raunaqmorarka, line 182: if start is present and we reached end before that, then should there be an exception ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,118 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+            if (endVersion.isPresent() && version == endVersion.get()) {
+                endOfTail = true;
+            }
+        }
+
+        return new TransactionLogTail(entriesBuilder.build(), version);
+    }
+
+    // Load a section of the Transaction Log JSON entries. Optionally from a given end version (inclusive) through an start version (exclusive)
+    private static TransactionLogTail loadNewTail(
```
raunaqmorarka: move below public methods

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,118 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+                endOfHead = true;
+            }
+        }
+        return new TransactionLogTail(transactionsBuilder.build().reversed(), endVersion);
+    }
+
+    public Optional<TransactionLogTail> getUpdatedTail(TrinoFileSystem fileSystem, String tableLocation, Optional<Long> endVersion, DataSize transactionLogMaxCachedFileSize)
+            throws IOException
+    {
+        checkArgument(endVersion.isEmpty() || endVersion.get() > version, "Invalid endVersion, expected higher than %s, but got %s", version, endVersion);
```
raunaqmorarka: Make the exception message here consistent with the one in loadNewTail

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
chenjian2664: Consider an update scenario: You get the current table version -- `currentVersion` and want to check if there are any new incoming logs. In this case, you will call `loadNewTail` with `startVersion` = `Optional.of(currentVersion)` and `endVersion` = `Optional.empty()`
Once the request arrives(in this method), the `startVersion` becomes `currentVersion + 1`. However, it’s possible that this version does not exist yet

chenjian2664: Just check the logic, actually we will never meet this condition since this endVersion will always be `Optional.empty()`

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -95,4 +139,67 @@ public long getRetainedSizeInBytes()
...
+            entriesBuilder.add(new Transaction(entryNumber, results.get()));
+            version = entryNumber;
+            entryNumber++;
+        }
+
+        return new TransactionLogTail(entriesBuilder.build(), version);
+    }
+
+    // Load a section of the Transaction Log JSON entries. Optionally from a given end version (inclusive) through an start version (exclusive)
+    private static TransactionLogTail loadNewTail(
```
raunaqmorarka: Can we keep the name `loadNewTailBackward` ?
Makes it easier to distinguish from the other method

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -95,4 +139,67 @@ public long getRetainedSizeInBytes()
                 + SIZE_OF_LONG
                 + estimatedSizeOf(entries, Transaction::getRetainedSizeInBytes);
     }
+
+    private static TransactionLogTail loadNewTail(
+            TrinoFileSystem fileSystem,
+            String tableLocation,
+            long version,
```
raunaqmorarka, line 135: What is `version` here ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -95,4 +139,67 @@ public long getRetainedSizeInBytes()
                 + SIZE_OF_LONG
                 + estimatedSizeOf(entries, Transaction::getRetainedSizeInBytes);
     }
+
```
raunaqmorarka, line 125: Please add a code comment for this method as well

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
chenjian2664: The determined maximum version in the log, which will be used as the table snapshot version

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,47 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+        }
+
+        return loadNewTail(fileSystem, tableLocation, 0L, 0L, transactionLogMaxCachedFileSize);
+    }
+
+    public Optional<TransactionLogTail> getUpdatedTail(TrinoFileSystem fileSystem, String tableLocation, Optional<Long> endVersion, DataSize transactionLogMaxCachedFileSize)
+            throws IOException
+    {
+        checkArgument(endVersion.isEmpty() || endVersion.get() > version, "Invalid start/end versions: %s, %s", version, endVersion);
+        TransactionLogTail newTail = loadNewTail(fileSystem, tableLocation, Optional.of(version), endVersion, transactionLogMaxCachedFileSize);
```
raunaqmorarka: Can we call `loadNewTailBackward` directly here ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -95,4 +139,68 @@ public long getRetainedSizeInBytes()
                 + SIZE_OF_LONG
                 + estimatedSizeOf(entries, Transaction::getRetainedSizeInBytes);
     }
+
+    // Load a section of transaction log JSON entries starting from `startVersion`(inclusive) up to the latest version
+    private static TransactionLogTail loadNewTail(
```
raunaqmorarka, line 132: I'm finding the input paramaters to this method hard to understand.
The code comment doesn't describe what should be populated in `version`.
One of the callers is using `startVersion` for `version` parameter and `startVersion + 1` for `startVersion` parameter.
Another is using 0 for both, so its confusing to figure out what the caller is supposed to populate.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/BaseTransactionsTable.java
```diff
@@ -140,7 +137,7 @@ public ConnectorPageSource pageSource(ConnectorTransactionHandle transactionHand
         TrinoFileSystem fileSystem = fileSystemFactory.create(session);
         PageListBuilder pagesBuilder = PageListBuilder.forTable(tableMetadata);
         try {
-            List<Transaction> transactions = loadNewTailBackward(fileSystem, table.location(), startVersionExclusive, endVersionInclusive.get()).reversed();
+            List<Transaction> transactions = loadNewTail(fileSystem, table.location(), startVersionExclusive, endVersionInclusive, DataSize.ofBytes(0)).getTransactions();
```
raunaqmorarka, line 142: It looks like `endVersionInclusive` will never be empty, so we should retain that clarity and unwrap the optional like before

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
```diff
@@ -47,6 +50,47 @@ public TransactionLogTail(List<Transaction> entries, long version)
...
+
+        if (endVersion.isPresent()) {
+            return loadNewTailBackward(fileSystem, tableLocation, startVersion, endVersion.get(), transactionLogMaxCachedFileSize);
+        }
+
+        if (startVersion.isPresent()) {
+            return loadNewTail(fileSystem, tableLocation, startVersion.get(), startVersion.get() + 1, transactionLogMaxCachedFileSize);
+        }
+
+        return loadNewTail(fileSystem, tableLocation, 0L, 0L, transactionLogMaxCachedFileSize);
```
raunaqmorarka, line 74: This looks unnecessary, I don't see where both versions would be empty

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
chenjian2664: removed this unused method

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
chenjian2664: from caller DeltaLakeMetadata#getTableHandle, it is possible that both start,end version are absent

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/checkpoint/TransactionLogTail.java
chenjian2664: Added more comments

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/BaseTransactionsTable.java
chenjian2664: added a check for it

raunaqmorarka (approved):

PR #24117, opened 2024-11-13
Add support for fetching Redshift query results using Redshift unload command

mayankvadariya:
## Description

Redshift supports unloading select query results to S3 using UNLOAD command. PR aims to use unload command to unload Redshift query results on S3 and to read generated redshift query results files in parallel.

Official documentation on Redshift unload command:
https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD.html
https://docs.aws.amazon.com/redshift/latest/dg/r_UNLOAD_command_examples.html

Connector internally converts Trino SQL to Redshift's unload SQL and runs it. Resultant Parquet files containing query results will be read in parallel to generate Trino query result. Only Parquet file format is supported in Redshift unload command.

Below are certain limitations for which Trino internally force fallback to traditional JDBC even if connector is configured to use unload.

- select query containing `TimeType` and `VarbinaryType` column type is unsupported as unload doesn't support unloading to Parquet format.
- select query ending with limit clause is unsupported as unload command doesn't support limit clause.
- select query containing filter condition on `DecimalType` column is unsupported with unload as unload query is generated by Trino doesn't add precision to `numeric` cast function.
- select query projecting static literals.

---
### Setup information:

|   Table    |    Schema     |   Records   |  Data size(agg. parquet file size)  |
|:----------:|:-------------:|:-----------:|:-----------------------------------:|
|   nation   |  tpch.sf1000  |     25      |                3.5K                 |
|  supplier  |  tpch.sf1000  |  10000000   |                1.4G                 |
|  customer  |  tpch.sf1000  |  150000000  |                22.8G                |

Coordinator size: `m6g.2xlarge` 
Worker size: 8 X `m7g.8xlarge`

Redshift cluster size: 2(nodes) X `dc2.large`

### Benchmarking stats

UNLOAD performs better as the volume of query results increases .

|  Query  |                                                                                    Query text                                                                                    | Average JDBC flow execution time(ms) | Average UNLOAD flow execution time(ms) |
|:-------:|:--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------:|:------------------------------------:|:--------------------------------------:|
|   q01   |  SELECT checksum(custkey), checksum(name), checksum(address), checksum(nationkey), checksum(phone), checksum(acctbal), checksum(mktsegment), checksum(comment) FROM "customer";  |               622,571                |                240,893                 |
[...]

PR #25866, opened 2025-05-26
Redshift batch inserts using COPY FROM operation

### LINKED ISSUE #24546: Support Redshift COPY for bulk loads
When loading data into Redshift using a CTAS statement, the Redshift connector defaults to BaseJdbcClient behaviour of using batched insert statements.  While this behaviour works well for most database systems, Redshift's handling of INSERT statements is very slow and, according to [this article](https://aws.amazon.com/blogs/big-data/top-10-performance-tuning-techniques-for-amazon-redshift/) by AWS, considered an anti-pattern.  This is also the case for most OLAP systems.  In real world performance, we see about 300 rows per second per Redshift node.

I was wondering if there was an appetite to improve this using the Redshift `COPY` statement? I think it would work as follows:

- Administrator enables inserts using copy as part of the catalog properties of Redshift (new property)
- Administrator provides AWS Access Key ID, Secret (or optionally role if running on AWS), and bucket + prefix (new properties)
- When running a CTAS statement, Trino still creates the table as it did before, but instead of streaming data using INSERT statements, it writes query results (using parquet or csv) to S3 using at location `s3://my-bucket/my-prefix/trino_query_id/parts`
- When all parts are written, Trino issues a `COPY ... FROM ...`

I can put my hand up to implement this if it is felt that this is something that would be useful to others.  It seems like the way to do so based on looking through similar implementations would be to implement the PageSink and then the PageSinkProvider for this operation, as well as any surrounding credential providers.  Please let me know if I am overlooking anything thinking about it this way or not considering anything major.

I have noticed that @mayankvadariya is implementing an adjacent feature in #24117 though so I don't want to step on any toes or duplicate effort if this is already being worked on.

brendanstennett:
## Description

Fixes #24546

This PR aims to allow the use of `COPY FROM` statements when sinking data into Redshift.  

The Redshift connector inherits `BaseJdbcConnector` which uses batched INSERT statements to execute sink operations.  Even when using non transactional mode, this can only push about 1000 rows per second.  This change stages the rows to a parquet file first, then issues a `COPY FROM` statement to load the table.  We are noticing 250K rows per second or more using this method.

This has been running in production for 2+ months on our own branch. 

This functionality needs to be enabled by specifying the following config option:
```
redshift.batched-inserts-copy-location=s3://my-bucket/my-prefix
```

The following options are also required when specifying the above:
```
redshift.batched-inserts-copy-iam-role=arn:aws:iam::123456789000:role/redshift_iam_role
s3.region=region
s3.aws-access-key=KEY
s3.aws-secret-key=SECRET
```

A suggested IAM Policy to for this role and user:
```
{
	"Version": "2012-10-17",
	"Statement": [
		{
			"Effect": "Allow",
			"Action": [
				"s3:ListBucket"
			],
			"Resource": "arn:aws:s3:::my-bucket"
		},
		{
			"Effect": "Allow",
			"Action": [
				"s3:GetObject",
				"s3:PutObject",
				"s3:DeleteObject"
			],
			"Resource": "arn:aws:s3:::my-bucket/*"
		}
	]
}
```

- This is designed to work alongside and compliments the new Redshift UNLOAD feature #24117
- When the above configuration is not supplied, it will default to base JDBC behaviour.
- There needs to be a new GHA var added to the repo `REDSHIFT_S3_COPY_ROOT` similar to `REDSHIFT_S3_UNLOAD_ROOT`

## Release notes

```markdown
## Redshift
* Add support for Redshift COPY FROM statements for batch insert operations ({issue}`24546`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/redshift.md
+++ b/docs/src/main/sphinx/connector/redshift.md
@@ -262,3 +262,35 @@ potentially re-activate it again afterward.
 Additionally, define further required [S3 configuration such as IAM key, role,
 or region](/object-storage/file-system-s3), except `fs.native-s3.enabled`,
 
+### Batch inserts from S3 using Redshift `COPY FROM`
+
+The connector supports the Redshift `COPY FROM` command to 
+efficiently write large batches of data to Redshift by staging them as 
+Parquet files on Amazon S3. This method significantly improves sink 
+performance compared to the default JDBC batch inserts for Redshift.
+
+This feature is disabled by default.  To enable this feature, configure a 
+writeable S3 location with the following configuration properties:
+
+:::{list-table} Parallel write configuration properties
+:widths: 30, 60
+:header-rows: 1
+
+* - Property
+   - Description
+* - `redshift.batched-inserts-copy-location`
+   - A writeable location in Amazon S3 in the same AWS region as the 
+     Redshift cluster. Used for temporary Parquet staging files during 
+     insert operations. These files are automatically cleaned up after the load.
+* - `redshift.batched-inserts-copy-iam-role`
+   - Fully specified ARN of the IAM Role to use for the `COPY FROM` 
+     command. This role must have read access to the S3 bucket.
+
+:::
+
+Use the `batched_inserts_copy_enabled` [catalog session property](/sql/set-session) to
+deactivate the batch inserts using copy for a specific query, and
+potentially re-activate it again afterward.
+
+Additionally, define further required [S3 configuration such as IAM key, role,
+or region](/object-storage/file-system-s3).
--- a/plugin/trino-redshift/src/main/java/io/trino/plugin/redshift/RedshiftBatchedInsertsCopyPageSink.java
+++ b/plugin/trino-redshift/src/main/java/io/trino/plugin/redshift/RedshiftBatchedInsertsCopyPageSink.java
@@ -1,0 +1,225 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.redshift;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.log.Logger;
+import io.airlift.slice.Slice;
+import io.trino.filesystem.Location;
+import io.trino.filesystem.TrinoFileSystem;
+import io.trino.filesystem.TrinoFileSystemFactory;
+import io.trino.parquet.writer.ParquetSchemaConverter;
+import io.trino.parquet.writer.ParquetWriterOptions;
+import io.trino.plugin.hive.parquet.ParquetFileWriter;
+import io.trino.spi.Page;
+import io.trino.spi.connector.ConnectorPageSink;
+import io.trino.spi.connector.ConnectorPageSinkId;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.type.ArrayType;
+import io.trino.spi.type.MapType;
+import io.trino.spi.type.RowType;
+import io.trino.spi.type.TimestampWithTimeZoneType;
+import io.trino.spi.type.Type;
+import io.trino.spi.type.TypeOperators;
+import org.apache.parquet.format.CompressionCodec;
+
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.ArrayList;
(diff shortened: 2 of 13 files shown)


brendanstennett:
@ebyhr Added requested documentation.  We need the ENV var set on the repo for the tests to pass.

`REDSHIFT_S3_COPY_ROOT` which would be a similar location to the ENV var already set of `REDSHIFT_S3_UNLOAD_ROOT`

plugin/trino-redshift/src/main/java/io/trino/plugin/redshift/RedshiftBatchedInsertsCopyPageSink.java
```diff
@@ -0,0 +1,225 @@
...
+    private ParquetFileWriter parquetWriter;
+    private long rowsInCurrentFile;
+    private int filePartNumber;
+
+    public RedshiftBatchedInsertsCopyPageSink(
+            ConnectorSession session,
+            ConnectorPageSinkId pageSinkId,
+            TrinoFileSystemFactory fileSystemFactory,
+            TypeOperators typeOperators,
+            Location copyLocationWithPrefix,
```
chenjian2664: How about rename to copyLocation

plugin/trino-redshift/src/main/java/io/trino/plugin/redshift/RedshiftConfig.java
```diff
@@ -69,4 +71,30 @@ public RedshiftConfig setUnloadIamRole(String unloadIamRole)
...
+        this.batchedInsertsCopyLocation = batchedInsertsCopyLocation;
+        return this;
+    }
+
+    public Optional<String> getBatchedInsertsCopyIamRole()
+    {
+        return Optional.ofNullable(batchedInsertsCopyIamRole);
+    }
+
+    @Config("redshift.batched-inserts-copy-iam-role")
```
chenjian2664, line 93: Do we need a toggle to control that we enable the COPY FROM or not, otherwise we only control it on session level?

docs/src/main/sphinx/connector/redshift.md
```diff
@@ -262,3 +262,35 @@ potentially re-activate it again afterward.
...
+   - A writeable location in Amazon S3 in the same AWS region as the 
+     Redshift cluster. Used for temporary Parquet staging files during 
+     insert operations. These files are automatically cleaned up after the load.
+* - `redshift.batched-inserts-copy-iam-role`
+   - Fully specified ARN of the IAM Role to use for the `COPY FROM` 
+     command. This role must have read access to the S3 bucket.
+
+:::
+
+Use the `batched_inserts_copy_enabled` [catalog session property](/sql/set-session) to
```
chenjian2664, line 291: We should mention current implementation set the value to true by default

plugin/trino-redshift/src/main/java/io/trino/plugin/redshift/RedshiftBatchedInsertsCopyPageSink.java
brendanstennett: Sure! I can do that.

plugin/trino-redshift/src/main/java/io/trino/plugin/redshift/RedshiftConfig.java
brendanstennett: Currently the functionality is only enabled when the `redshift.batched-inserts-copy-location` is set.  See:
https://github.com/trinodb/trino/pull/25866/files/0d4c2da0b150293a8c3b005aa9b9260b5ee098ae#diff-17fd662125cb688e826b7ba43089b3fe313d2f28d124b8378e893f58f3d924ecR80

This is then SESSION set able back to false I believe based on how this is implemented:
https://github.com/trinodb/trino/pull/25866/files/0d4c2da0b150293a8c3b005aa9b9260b5ee098ae#diff-332ac3ff4e15ae6df7a4cbdf8aecb64cfaa8ef4ac6613861b0fba9149a613563

Tests for that:
https://github.com/trinodb/trino/pull/25866/files/0d4c2da0b150293a8c3b005aa9b9260b5ee098ae#diff-e74287f17b924780580d36fd14d86785680657403ccdd1df18154b24bbb96be0R90-R95

docs/src/main/sphinx/connector/redshift.md
brendanstennett: Are you suggesting that we enable this to true by default or just highlight that it is.  As implemented it's actually false by default and only true if `redshift.batched-inserts-copy-location` is set.

brendanstennett:
@chenjian2664 Thank you for having a look.  I dropped some replies to your comments.  What is the best way to get the additional env var added to GHA? Once that's is in, I can rebase and we can rerun the test suite.

docs/src/main/sphinx/connector/redshift.md
brendanstennett: Updated the docs to mention that this is disabled by default.

lewispeckover:
This would be great to get picked up, Redshift insert performance over jdbc is a pain point.

brendanstennett:
I can likely pick this back up in the next week or so. We have been using a forked version in production for almost two years without issues. @chenjian2664 is this something that we can get merged if I rebase / fixup?

PR #25879, opened 2025-05-28
Release hash aggregation memory on output

pettyjamesm:
## Description
Incrementally releases memory from `FlatGroupByHash` when `HashAggregationOperator` starts producing output. Previously, the entire hash table contents were kept in memory until output data was produced from hash aggregations.

## Release notes

```markdown
## General
* Reduce memory usage of aggregations by incrementally releasing memory as output rows are are produced ({issue}`25879`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
+++ b/core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
@@ -138,6 +138,27 @@ public final class AppendOnlyVariableWidthData
         return chunks.get(chunkIndex);
     }
 
+    public void freeChunksBefore(byte[] pointer, int pointerOffset)
+    {
+        int chunkIndex = getChunkIndex(pointer, pointerOffset);
+        if (chunks.isEmpty()) {
+            verify(chunkIndex == 0);
+            return;
+        }
+        checkIndex(chunkIndex, chunks.size());
+        // Release any previous chunks until a null chunk is encountered, which means it and any previous
+        // batches have already been released
+        int releaseIndex = chunkIndex - 1;
+        while (releaseIndex >= 0) {
+            byte[] releaseChunk = chunks.set(releaseIndex, null);
+            if (releaseChunk == null) {
+                break;
+            }
+            chunksRetainedSizeInBytes -= sizeOf(releaseChunk);
+            releaseIndex--;
+        }
+    }
+
     // growth factor for each chunk doubles up to 512KB, then increases by 1.5x for each chunk after that
     private static long nextChunkSize(long previousChunkSize)
     {
--- a/core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
+++ b/core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
@@ -393,11 +393,10 @@ public class FlatGroupByHash
         {
             verify(canProcessDictionary(blocks), "invalid call to addDictionaryPage");
             this.dictionaryBlock = (DictionaryBlock) blocks[0];
-
-            this.dictionaries = Arrays.stream(blocks)
-                    .map(block -> (DictionaryBlock) block)
-                    .map(DictionaryBlock::getDictionary)
-                    .toArray(Block[]::new);
+            this.dictionaries = blocks;
+            for (int i = 0; i < dictionaries.length; i++) {
+                dictionaries[i] = ((DictionaryBlock) dictionaries[i]).getDictionary();
+            }
             updateDictionaryLookBack(dictionaries[0]);
         }
 
@@ -620,13 +619,12 @@ public class FlatGroupByHash
             verify(canProcessDictionary(blocks), "invalid call to processDictionary");
 
             this.dictionaryBlock = (DictionaryBlock) blocks[0];
-            this.groupIds = new int[dictionaryBlock.getPositionCount()];
-
-            this.dictionaries = Arrays.stream(blocks)
-                    .map(block -> (DictionaryBlock) block)
-                    .map(DictionaryBlock::getDictionary)
-                    .toArray(Block[]::new);
+            this.dictionaries = blocks;
+            for (int i = 0; i < dictionaries.length; i++) {
+                dictionaries[i] = ((DictionaryBlock) dictionaries[i]).getDictionary();
+            }
             updateDictionaryLookBack(dictionaries[0]);
+            this.groupIds = new int[dictionaryBlock.getPositionCount()];
         }
 
         @Override
--- a/core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
+++ b/core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
@@ -130,6 +130,16 @@ public class FlatGroupByHash
     }
 
     @Override
+    public void startReleasingOutput()
+    {
+        currentHashes = null;
+        dictionaryLookBack = null;
+        Arrays.fill(currentBlocks, null);
+        currentPageSizeInBytes = 0;
+        flatHash.startReleasingOutput();
+    }
+
+    @Override
     public void appendValuesTo(int groupId, PageBuilder pageBuilder)
     {
         BlockBuilder[] blockBuilders = currentBlockBuilders;
--- a/core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
+++ b/core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
@@ -510,7 +510,7 @@ public class FlatGroupByHash
         public GetNonDictionaryGroupIdsWork(Block[] blocks)
         {
             this.blocks = blocks;
-            this.groupIds = new int[currentBlocks[0].getPositionCount()];
+            this.groupIds = new int[blocks[0].getPositionCount()];
         }
 
         @Override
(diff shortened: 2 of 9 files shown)


core/trino-main/src/main/java/io/trino/operator/aggregation/builder/InMemoryHashAggregationBuilder.java
```diff
@@ -248,12 +248,18 @@ public WorkProcessor<Page> buildResult()
         for (GroupedAggregator groupedAggregator : groupedAggregators) {
             groupedAggregator.prepareFinal();
         }
-        return buildResult(consecutiveGroupIds(), new PageBuilder(buildTypes()), false);
+        // Only incrementally release memory for final aggregations, since partial aggregations have a fixed
+        // memory limit and can be expected to fully flush and release their output quickly
```
sopel39, line 235: Does it matter in practice? Without this condition the code could be simplified and `startReleasingOutput` could be removed

core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
```diff
@@ -138,6 +138,27 @@ public byte[] getChunk(byte[] pointer, int pointerOffset)
         return chunks.get(chunkIndex);
     }
 
+    public void freeChunksBefore(byte[] pointer, int pointerOffset)
+    {
+        int chunkIndex = getChunkIndex(pointer, pointerOffset);
+        if (chunks.isEmpty()) {
+            verify(chunkIndex == 0);
```
sopel39, line 145: is `chunkIndex=0` a valid value when there are no chunks?

core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
```diff
@@ -138,6 +138,27 @@ public byte[] getChunk(byte[] pointer, int pointerOffset)
...
+            return;
+        }
+        checkIndex(chunkIndex, chunks.size());
```
sopel39, line 148: doesn't `getChunkIndex` guarantee correct index?

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
```diff
@@ -153,13 +153,36 @@ public int getCapacity()
...
 
+    /**
+     * Releases memory associated with the hash table which is no longer necessary to produce output. Subsequent
+     * calls to insert new elements are rejected, and calls to {@link FlatHash#appendTo(int, BlockBuilder[])} will
+     * incrementally release memory associated with prior groupId values assuming that the caller will only call into
+     * the method to produce output in a sequential fashion.
+     */
+    public void startReleasingOutput()
+    {
+        if (isReleasingOutput()) {
```
sopel39: A better place for it might be `BigintGroupByHash` and `FlatGroupByHash` since they also release internal structures

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
```diff
@@ -153,13 +153,36 @@ public int getCapacity()
...
+    private boolean isReleasingOutput()
+    {
+        return control == null;
+    }
+
     public long hashPosition(int groupId)
     {
-        if (groupId < 0) {
-            throw new IllegalArgumentException("groupId is negative");
+        if (groupId < 0 || groupId >= nextGroupId) {
```
sopel39, line 173: unrelated change, separate commit?

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
sopel39: use checkState instead

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
```diff
@@ -153,13 +153,36 @@ public int getCapacity()
...
+            throw new IllegalArgumentException("groupId out of range: " + groupId);
         }
         byte[] fixedSizeRecords = getFixedSizeRecords(groupId);
         int fixedRecordOffset = getFixedRecordOffset(groupId);
+        if (isReleasingOutput() && fixedSizeRecords == null) {
```
sopel39: why `fixedSizeRecords == null` check?

use `checkState` instead

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
```diff
@@ -251,6 +287,9 @@ public int putIfAbsent(Block[] blocks, int position, long hash)
 
     private int getIndex(Block[] blocks, int position, long hash)
     {
+        if (isReleasingOutput()) {
```
sopel39: use `checkState`

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
```diff
@@ -351,6 +390,9 @@ private void setControl(int index, byte hashPrefix)
 
     public boolean ensureAvailableCapacity(int batchSize)
     {
+        if (isReleasingOutput()) {
```
sopel39: use `checkState`

core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
```diff
@@ -403,11 +403,10 @@ public AddDictionaryPageWork(Block[] blocks)
         {
             verify(canProcessDictionary(blocks), "invalid call to addDictionaryPage");
             this.dictionaryBlock = (DictionaryBlock) blocks[0];
-
-            this.dictionaries = Arrays.stream(blocks)
-                    .map(block -> (DictionaryBlock) block)
-                    .map(DictionaryBlock::getDictionary)
-                    .toArray(Block[]::new);
+            this.dictionaries = blocks;
```
sopel39, line 396: what's the benefit?

core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
pettyjamesm: Yes, calling `(AppendOnly)VariableWidthData#allocate(0)` will yield `EMPTY_CHUNK` but that chunk isn't added to the `chunks` list. You'll see the same logic in `(AppendOnly)VariableWidthData#getChunk` method asserting that `chunkIndex == 0` when `chunks.isEmpty()`

core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
pettyjamesm: `getChunkIndex` just reads the int value out of the fixed region pointer, so this check validates that the value read is actually pointing to a valid chunk in the list as a sanity check.

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
pettyjamesm: This is the only place where we have structures released that definitively prove that we've transitioned into releasing output. In `FlatGroupByHash` and `BigintGroupByHash` we null out values that are lazily initialized and therefore are allowed to be `null` without necessarily indicating that a terminal transition to releasing output has occurred.

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
pettyjamesm: `fixedSizeRecords` will be null if we've already released the record batch (and are releasing output), otherwise it's valid to refer to the current non-null batch.

core/trino-main/src/main/java/io/trino/operator/aggregation/builder/InMemoryHashAggregationBuilder.java
pettyjamesm: 1. I think you need `startReleasingOutput()` no matter what since there are structures in the `GroupByHash` that can be released immediately before any `groupId` is emitted.
2. Avoiding releasing memory in partial aggregations can help avoid lock contention in the memory tracking system for high parallelism operations (partial aggregations are likely to appear on top of table scans). It may not matter much in practice but it seemed like incrementally releasing there was unlikely to be helpful since we're flushing anyway and the memory released is expected to be negligible

core/trino-main/src/main/java/io/trino/operator/FlatHash.java
pettyjamesm: Extracted into separate commit

core/trino-main/src/main/java/io/trino/operator/FlatGroupByHash.java
pettyjamesm: No need to allocate the extra blocks array on every invocation, we can reuse the `currentBlocks` array here like we do in the other scenarios.

core/trino-main/src/main/java/io/trino/operator/AppendOnlyVariableWidthData.java
sopel39: wouldn't a -1 be a better value when there are no chunks?

pettyjamesm: I would think so, but that would be a stylistic choice that can be deferred to another PR

PR #25902, opened 2025-06-02
Document Loki connector result types

jeschkies:
## Description
The original Loki connector release was missing a description of the result types of a metric query. This changes adds them.

## Release notes

```markdown
## Section
```

### DIFF
--- a/docs/src/main/sphinx/connector/loki.md
+++ b/docs/src/main/sphinx/connector/loki.md
@@ -174,3 +174,12 @@ FROM
     )
   )
 ;
+```
+
+The columns have the following names and types 
+
+| Column Name | Type                        |
+|-------------|-----------------------------|
+| timestamp   | timestamp(3) with time zone |
+| value       | double                      |
+| province    | varchar                     |
--- a/docs/src/main/sphinx/connector/loki.md
+++ b/docs/src/main/sphinx/connector/loki.md
@@ -176,7 +176,25 @@ FROM
 ;
 ```
 
-The columns have the following names and types 
+The output types are as follows
+
+```sql
+PREPARE result FROM
+SELECT 
+  timestamp, 
+  value, 
+  labels['province'] AS province
+FROM
+  TABLE(
+    example.system.query_range(
+      '{origin="CA"}',
+      TIMESTAMP '2025-01-01 00:00:00',
+      TIMESTAMP '2025-01-02 00:00:00'
+    )
+  )
+;
+DESCRIBE OUTPUT result;
+```
 
 ```
  Column Name | Type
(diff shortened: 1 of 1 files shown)


docs/src/main/sphinx/connector/loki.md
```diff
@@ -174,3 +174,12 @@ FROM
     )
   )
 ;
+```
+
+The columns have the following names and types 
+
+| Column Name | Type                        |
```
mosabua: Please use a list-table and same formatting as in other data type mapping tables

Also sentence case titles so 

Column names

mosabua (changes requested):

docs/src/main/sphinx/connector/loki.md
```diff
@@ -174,3 +174,12 @@ FROM
     )
   )
 ;
+```
+
+The columns have the following names and types
```
ebyhr: We use `text` code block for a result example. Please take a look at https://trino.io/docs/current/sql/select.html.

ebyhr (changes requested):

docs/src/main/sphinx/connector/loki.md
jeschkies: Do you have an example of a data type mapping table?

jeschkies: I've used the text block instead.

jeschkies:
I'm unsure how you would like to have the types documented. I've used the `DESCRIBE OUTPUT` query and result now. It's a little verbose IMHO. It would help if you could point me to an example on how to document the output type.

mosabua:
A list-table similar to 

https://github.com/trinodb/trino/blob/master/docs/src/main/sphinx/connector/postgresql.md?plain=1#L136

and we dont need the example query but if you want to keep it thats fine as well. Just add it after the table.

docs/src/main/sphinx/connector/loki.md
ebyhr: Please show result's example and avoid prepread statement.

ebyhr:
I don't think we should add another `list-table`. We already have https://trino.io/docs/current/connector/loki.html#loki-to-trino-type-mapping

mosabua:
> I don't think we should add another `list-table`. We already have https://trino.io/docs/current/connector/loki.html#loki-to-trino-type-mapping

Oh .. I totally forgot .. what is the PR trying to achieve then @jeschkies  .. seems like all the needed info is there already.

Maybe we just need to tweak the wording in the title of that table and add some more info there.

mosabua:
I chatted with @jeschkies as part of the TCB work .. I think we can close this PR .. can you confirm @jeschkies ?

ebyhr:
Please feel free to reopen if you continue the work.

PR #25903, opened 2025-06-02
Sanitize schema and catalog case in Session

losipiuk:
It was possible to provide non-lowercase catalog/schema via connection uri used with JDBC driver. That caused failures of checks during analysis that both schema and catalog must be lowercase.

The fix sanitizes schema and catalog case when Session objects, which holds these, is consturucted.

This fixes JDBC failure scenario and potential other scenarios where different clients would send non-lowercase schema/catalog over Trino protocol.

## Release notes

```markdown
## General
* Fix query failures if catalog/schema passed over Trino connection string were not lower case. ({issue}`25903`)
```

### DIFF
--- a/client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcConnection.java
+++ b/client/trino-jdbc/src/test/java/io/trino/jdbc/TestJdbcConnection.java
@@ -163,6 +163,8 @@ public class TestJdbcConnection
                 Statement statement = connection.createStatement()) {
             statement.execute("SET ROLE admin IN hive");
             statement.execute("CREATE SCHEMA default");
+            statement.execute("CREATE TABLE hive.default.dummy(a bigint)");
+
             statement.execute("CREATE SCHEMA fruit");
             statement.execute(
                     "CREATE TABLE blackhole.default.devzero(dummy bigint) " +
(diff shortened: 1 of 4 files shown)


core/trino-main/src/main/java/io/trino/Session.java
```diff
@@ -129,8 +129,8 @@ public Session(
         this.identity = requireNonNull(identity, "identity is null");
         this.originalIdentity = requireNonNull(originalIdentity, "originalIdentity is null");
         this.source = requireNonNull(source, "source is null");
-        this.catalog = requireNonNull(catalog, "catalog is null");
-        this.schema = requireNonNull(schema, "schema is null");
+        this.catalog = requireNonNull(catalog, "catalog is null").map(value -> value.toLowerCase(Locale.ENGLISH));
```
wendigo: I think that sanitization should happen in the `HttpRequestSessionContextFactory`

losipiuk: I put it there orignally. IDK. This one is more general. Why do you think that one is better?

wendigo: Catalog and schema are read from the headers so it should happen on read

losipiuk:
AC @wendigo

wendigo (approved):

wendigo:
Related failure:

```
Error:  Failures: 
Error:    TestQuerySessionSupplier.testCreateSession:89 
expected: "testCatalog"
 but was: "testcatalog"
[INFO] 
```

PR #25906, opened 2025-06-02
Add support for View Refresh operation

Praveen2112:
## Description

We introduce a new syntax to perform the refresh operation on a view definition - i.e say if we create a view on an existing table like `CREATE VIEW v1 AS SELECT * FROM table_1` and if the table_1 has a schema changes like adding or dropping a column then the view is unaccessible and we need to run `CREATE OR REPLACE VIEW ...` to update the view definition - instead on running `CREATE OR REPLACE VIEW ...` we add support for `ALTER VIEW ... REFRESH` which gets the view definition, analyzes and refreshes view object accordingly. 

## Release notes

```markdown
## General
* Add support for view refresh operation. ({issue}`25906`)

## Memory
* Add support for view refresh operation. ({issue}`25906`)

```


core/trino-main/src/test/java/io/trino/execution/BaseDataDefinitionTaskTest.java
```diff
@@ -143,6 +144,7 @@ MATERIALIZED_VIEW_PROPERTY_1_NAME, longProperty(MATERIALIZED_VIEW_PROPERTY_1_NAM
                 MATERIALIZED_VIEW_PROPERTY_2_NAME, stringProperty(MATERIALIZED_VIEW_PROPERTY_2_NAME, "property 2", MATERIALIZED_VIEW_PROPERTY_2_DEFAULT_VALUE, false));
         materializedViewPropertyManager = new MaterializedViewPropertyManager(CatalogServiceProvider.singleton(TEST_CATALOG_HANDLE, properties));
         queryStateMachine = stateMachine(transactionManager, createTestMetadataManager(), new AllowAllAccessControl(), testSession);
+        metadata.createSchema(testSession, new CatalogSchemaName(TEST_CATALOG_NAME, SCHEMA), ImmutableMap.of(), new TrinoPrincipal(PrincipalType.ROLE, "role"));
```
marcinsbd: nit: import static for PrincipalType.ROLE?

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        Statement viewDefinitionSql = sqlParser.createStatement(viewDefinition.getOriginalSql());
+
+        Analysis analysis = analyzerFactory.createAnalyzer(viewSession, parameters, ImmutableMap.of(), stateMachine.getWarningCollector(), stateMachine.getPlanOptimizersStatsCollector())
+                .analyze(viewDefinitionSql);
+
+        List<ViewColumn> columns = analysis.getOutputDescriptor(viewDefinitionSql)
+                .getVisibleFields().stream()
+                .map(field -> new ViewColumn(field.getName().get(), field.getType().getTypeId(), Optional.empty()))
```
ebyhr: Column comments disappear after refresh:
```sql
trino:default> CREATE VIEW test AS SELECT 1 a;
CREATE VIEW
trino:default> COMMENT ON COLUMN test.a IS 'test comment';
COMMENT
trino:default> DESC test;
 Column |  Type   | Extra |   Comment
--------+---------+-------+--------------
 a      | integer |       | test comment
(1 row)

trino:default> ALTER VIEW test REFRESH;
REFRESH VIEW
trino:default> DESC test;
 Column |  Type   | Extra | Comment
--------+---------+-------+---------
 a      | integer |       |
(1 row)
```

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -114,6 +114,7 @@ statement
         SET PROPERTIES propertyAssignments                             #setMaterializedViewProperties
     | DROP VIEW (IF EXISTS)? qualifiedName                             #dropView
     | ALTER VIEW from=qualifiedName RENAME TO to=qualifiedName         #renameView
+    | ALTER VIEW viewName=qualifiedName REFRESH                        #refreshView
```
kasiafi, line 117: For materialized views we have syntax `REFRESH MATERIALIZED VIEW qualifiedName`.
Maybe we should do `REFRESH VIEW qualifiedName` here for consistency? Other statements have the same syntax for views and materialized views.

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,117 @@
...
+        Metadata metadata = plannerContext.getMetadata();
+        Session session = stateMachine.getSession();
+        QualifiedObjectName viewName = createQualifiedObjectName(session, refreshView, refreshView.getName());
+
+        ViewDefinition viewDefinition = metadata.getView(session, viewName)
+                .orElseThrow(() -> semanticException(TABLE_NOT_FOUND, refreshView, "View '%s' not found", viewName));
+
+        accessControl.checkCanRefreshView(session.toSecurityContext(), viewName);
+
+        Session viewSession = stateMachine.getSession();
```
kasiafi: `Session viewSession = session;`

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,117 @@
...
+        if (!viewDefinition.isRunAsInvoker()) {
+            checkArgument(viewDefinition.getRunAsIdentity().isPresent(), "View owner detail is missing");
+            viewSession = viewSession.createViewSession(viewDefinition.getCatalog(), viewDefinition.getSchema(), viewDefinition.getRunAsIdentity().get(), viewDefinition.getPath());
```
kasiafi: Comparing the code of `StatementAnalyzer.analyzeView()`, it seems that we should do `createViewSession` in both cases (run as invoker and run as identity) so that we use the catalog, schema and path from the view definition.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
Praveen2112: Since we are just modifying the view metadata information `ALTER VIEW ... REFRESH` would be more of a consistent approach right ?

cc: @martint Any guidence would be helpful.

Praveen2112:
@ebyhr , @kasiafi , @marcinsbd Thanks a lot for the review. AC

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -989,6 +990,51 @@ public void testView()
...
+        if (!hasBehavior(SUPPORTS_CREATE_TABLE) && !hasBehavior(SUPPORTS_ADD_COLUMN)) {
+            throw new AssertionError("Cannot test ALTER VIEW REFRESH without CREATE TABLE, the test needs to be implemented in a connector-specific way");
+        }
+
+        try (TestTable table = new TestTable(getQueryRunner()::execute, "test_table", "(id BIGINT, column_to_dropped BIGINT, column_with_comment BIGINT)", ImmutableList.of("1, 2, 3"));
+                TestView view = new TestView(getQueryRunner()::execute, "test_view", " SELECT * FROM %s".formatted(table.getName()))) {
+            assertQuery("SELECT * FROM %s".formatted(view.getName()), "VALUES (1, 2, 3)");
+
+            assertUpdate("COMMENT ON COLUMN %s.column_with_comment IS 'test comment'".formatted(view.getName()));
+            assertQuery("SELECT remarks FROM system.jdbc.columns WHERE table_name = '%s' AND column_name = 'column_with_comment'".formatted(view.getName()), "VALUES 'test comment'");
```
ebyhr: nit: We could use `getColumnComment` helper method.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -989,6 +990,51 @@ public void testView()
...
+            assertUpdate("ALTER VIEW %s REFRESH".formatted(view.getName()));
+            assertQuery("SELECT * FROM %s".formatted(view.getName()), "VALUES (1, 2, 3, null)");
+            assertQuery("SELECT remarks FROM system.jdbc.columns WHERE table_name = '%s' AND column_name = 'column_with_comment'".formatted(view.getName()), "VALUES 'test comment'");
+
+            assertUpdate("ALTER TABLE %s RENAME COLUMN column_with_comment TO renamed_column".formatted(table.getName()));
+            assertQueryFails(
+                    "SELECT * FROM %s".formatted(view.getName()),
+                    ".*is stale or in invalid state: column \\[renamed_column] of type bigint projected from query view at position 2 has a different name from column \\[column_with_comment] of type bigint stored in view definition");
+            assertUpdate("ALTER VIEW %s REFRESH".formatted(view.getName()));
+            assertQuery("SELECT * FROM %s".formatted(view.getName()), "VALUES (1, 2, 3, null)");
```
ebyhr: nit: We could use `+` operator. 

https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#prefer-string-formatting
> Sometimes, if you only need to append something, consider using the + operator.

core/trino-main/src/test/java/io/trino/execution/TestRefreshViewTask.java
```diff
@@ -0,0 +1,252 @@
...
+import static io.airlift.concurrent.MoreFutures.getFutureValue;
+import static io.trino.spi.connector.SaveMode.FAIL;
+import static io.trino.spi.type.BigintType.BIGINT;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static io.trino.sql.analyzer.StatementAnalyzerFactory.createTestingStatementAnalyzerFactory;
+import static io.trino.testing.TestingHandles.TEST_CATALOG_NAME;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class TestRefreshViewTask
```
ebyhr:
```suggestion
final class TestRefreshViewTask
```

https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
> Test classes should be defined as package-private and final.

core/trino-main/src/test/java/io/trino/execution/TestRefreshViewTask.java
```diff
@@ -0,0 +1,252 @@
...
+                        plannerContext,
+                        new AllowAllAccessControl(),
+                        new TablePropertyManager(CatalogServiceProvider.fail()),
+                        new AnalyzePropertyManager(CatalogServiceProvider.fail())),
+                new StatementRewrite(ImmutableSet.of()),
+                plannerContext.getTracer());
+    }
+
+    @Test
+    public void testAddNewColumnsOnTableWhenRefreshingExistingView()
```
ebyhr: Same for other test methods.
```suggestion
    void testAddNewColumnsOnTableWhenRefreshingExistingView()
```
https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
> Test methods should be defined as package-private.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -185,6 +185,7 @@ protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
                  SUPPORTS_RENAME_FIELD,
                  SUPPORTS_RENAME_SCHEMA,
                  SUPPORTS_SET_COLUMN_TYPE,
+                 SUPPORTS_REFRESH_VIEW,
```
ebyhr: Could you put this entry in alphabetical order? Same for other tests.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/security/SqlStandardAccessControl.java
```diff
@@ -238,6 +239,14 @@ public void checkCanSetViewComment(ConnectorSecurityContext context, SchemaTable
         }
     }
 
+    @Override
+    public void checkCanRefreshView(ConnectorSecurityContext context, SchemaTableName viewName)
+    {
+        if (!checkTablePermission(context, viewName, UPDATE, false)) {
+            denyRefreshView(viewName.toString());
+        }
```
ebyhr, line 247: The use of UPDATE privilege looked weird at first, but I see it's consistent with  `checkCanRefreshMaterializedView`.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -989,6 +990,51 @@ public void testView()
...
+
+        if (!hasBehavior(SUPPORTS_CREATE_TABLE) && !hasBehavior(SUPPORTS_ADD_COLUMN)) {
+            throw new AssertionError("Cannot test ALTER VIEW REFRESH without CREATE TABLE, the test needs to be implemented in a connector-specific way");
+        }
+
+        try (TestTable table = new TestTable(getQueryRunner()::execute, "test_table", "(id BIGINT, column_to_dropped BIGINT, column_with_comment BIGINT)", ImmutableList.of("1, 2, 3"));
+                TestView view = new TestView(getQueryRunner()::execute, "test_view", " SELECT * FROM %s".formatted(table.getName()))) {
+            assertQuery("SELECT * FROM %s".formatted(view.getName()), "VALUES (1, 2, 3)");
+
+            assertUpdate("COMMENT ON COLUMN %s.column_with_comment IS 'test comment'".formatted(view.getName()));
```
ebyhr: Should we check `SUPPORTS_COMMENT_ON_TABLE` and `SUPPORTS_RENAME_COLUMN` behaviors?

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
Praveen2112: Sometimes I think we could a consistent behavious on using formatted behaviour for query. I'm fine with both the options.

Praveen2112:
@ebyhr Thanks a lot for the review. AC

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,136 @@
...
+        Statement viewDefinitionSql = sqlParser.createStatement(viewDefinition.getOriginalSql());
+
+        Analysis analysis = analyzerFactory.createAnalyzer(viewSession, parameters, ImmutableMap.of(), stateMachine.getWarningCollector(), stateMachine.getPlanOptimizersStatsCollector())
+                .analyze(viewDefinitionSql);
+
+        Map<String, String> columnComments =
+                viewDefinition.getColumns()
+                        .stream()
+                        .filter(viewColumn -> viewColumn.comment().isPresent())
+                        .collect(toImmutableMap(ViewColumn::name, viewColumn -> viewColumn.comment().get()));
```
kasiafi, line 128: Can there be duplicate column names? I guess that in such case they cannot have a comment, right?

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,136 @@
...
+        Metadata metadata = plannerContext.getMetadata();
+        Session session = stateMachine.getSession();
+        QualifiedObjectName viewName = createQualifiedObjectName(session, refreshView, refreshView.getName());
+
+        ViewDefinition viewDefinition = metadata.getView(session, viewName)
+                .orElseThrow(() -> semanticException(TABLE_NOT_FOUND, refreshView, "View '%s' not found", viewName));
+
+        accessControl.checkCanRefreshView(session.toSecurityContext(), viewName);
+
+        Identity identity = session.getIdentity();
```
kasiafi, line 96: I think that ideally the view definition should always be refreshed by the same identity that initially created the view.
I understand it is not possible in the current model because we don't keep this information for the "run as invoker" views. Refreshing the view by the same identity who would run the view is the second best option imo.

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,136 @@
...
+            identity = Identity.from(owner)
+                    .withGroups(groupProvider.getGroups(owner.getUser()))
+                    .build();
+        }
+
+        Session viewSession = session.createViewSession(viewDefinition.getCatalog(), viewDefinition.getSchema(), identity, viewDefinition.getPath());
+
+        Statement viewDefinitionSql = sqlParser.createStatement(viewDefinition.getOriginalSql());
+
+        Analysis analysis = analyzerFactory.createAnalyzer(viewSession, parameters, ImmutableMap.of(), stateMachine.getWarningCollector(), stateMachine.getPlanOptimizersStatsCollector())
```
kasiafi: Given that there are possibly two identities involved, should we create special access control, like we do in `StatementAnalyzer.analyzeView()`?

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/security/FileBasedAccessControl.java
```diff
@@ -449,6 +450,15 @@ public void checkCanRenameView(ConnectorSecurityContext context, SchemaTableName
         }
     }
 
+    @Override
+    public void checkCanRefreshView(ConnectorSecurityContext context, SchemaTableName viewName)
+    {
+        // check if user owns the existing view, and if they will be an owner of the view after the rename
```
kasiafi: Update comment.

kasiafi (approved):

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
Praveen2112: Today for tables and views we don't support duplicate column names.

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
Praveen2112: I would like to retain the secuirty mode for refresh like we do for execution - Ideally the same could be enforced via access control mechansim as well. 

>  Refreshing the view by the same identity who would run the view is the second best option imo.

Can you explain it a bit.

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
Praveen2112: I think the view access control just wraps the access control error message. Do we need something similar here as well ?

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -989,6 +990,70 @@ public void testView()
...
+                assertQueryFails("CREATE VIEW sample_view AS SELECT * FROM nation", "This connector does not support creating views");
+            }
+            return;
+        }
+
+        if (!hasBehavior(SUPPORTS_CREATE_TABLE) && !hasBehavior(SUPPORTS_ADD_COLUMN)) {
+            throw new AssertionError("Cannot test ALTER VIEW REFRESH without CREATE TABLE, the test needs to be implemented in a connector-specific way");
+        }
+
+        try (TestTable table = new TestTable(getQueryRunner()::execute, "test_table", "(id BIGINT, column_to_dropped BIGINT, column_to_be_renamed BIGINT, column_with_comment BIGINT)", ImmutableList.of("1, 2, 3, 4"));
```
chenjian2664: we can use `newTrinoTable` to create test table

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -989,6 +990,70 @@ public void testView()
...
+    public void testRefreshView()
+    {
+        if (!hasBehavior(SUPPORTS_REFRESH_VIEW)) {
+            if (hasBehavior(SUPPORTS_CREATE_VIEW)) {
+                try (TestView testView = new TestView(getQueryRunner()::execute, "test_view", " SELECT * FROM nation")) {
+                    assertQueryFails("ALTER VIEW %s REFRESH".formatted(testView.getName()), "This connector does not support refreshing view definition");
+                }
+            }
+            else {
+                assertQueryFails("CREATE VIEW sample_view AS SELECT * FROM nation", "This connector does not support creating views");
```
chenjian2664: This should already be covered by `testView`

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,136 @@
...
+import java.util.Optional;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.util.concurrent.Futures.immediateVoidFuture;
+import static io.trino.metadata.MetadataUtil.createQualifiedObjectName;
+import static io.trino.spi.StandardErrorCode.TABLE_NOT_FOUND;
+import static io.trino.sql.analyzer.SemanticExceptions.semanticException;
+import static java.util.Objects.requireNonNull;
+import static org.weakref.jmx.$internal.guava.collect.ImmutableMap.toImmutableMap;
```
chenjian2664: Wrong import?

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
kasiafi: It is also supposed to check if the view owner grants permissions to the current identity, which matters in case when we run the view refresh as owner but the current identity is different.

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
kasiafi: > I would like to retain the secuirty mode for refresh like we do for execution

It is not obvious to me that we should refresh in the same mode as we execute the view. 
To keep the view semantics as close to the original as possible, we should re-define it as the original creator. This is not always possible though, since we lose the original creator info in case of the "run as invoker" views. 
I think that ideally we should decouple the view creator from the "run as identity". For now, we can just refresh as the "executing" identity. What do you think?

Praveen2112: > To keep the view semantics as close to the original as possible, we should re-define it as the original creator.

I agree with this - we should redfine it via the owner identity. 

> For now, we can just refresh as the "executing" identity. What do you think?

But it wouldn;t work for the definer views right ? Let’s say if we have an access to the table only for the owner and accessing/refreshing via  `executing` identity might not fail to refresh the views.

kasiafi: Agreed. The refresh can fail when run by a non-owner. I was thinking of a hypothetical situation when the view semantics actually differs for the owner and the executing identity -- for example if there was a mechanism to hide some columns from some users. If a random user refreshed the view, everyone would get their subset of columns.

This is why I said we should always refresh as the owner.
Do you know what it takes to do so?

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
```diff
@@ -0,0 +1,144 @@
...
+        Identity identity = session.getIdentity();
+        AccessControl viewAccessControl = accessControl;
+
+        if (!viewDefinition.isRunAsInvoker()) {
+            checkArgument(viewDefinition.getRunAsIdentity().isPresent(), "View owner detail is missing");
+            Identity owner = viewDefinition.getRunAsIdentity().get();
+            identity = Identity.from(owner)
+                    .withGroups(groupProvider.getGroups(owner.getUser()))
+                    .build();
+            viewAccessControl = new ViewAccessControl(accessControl);
```
kasiafi: Similarly a in `StatemenAnalyzer.analyzeView()`, we don't need to modify the access control when the current identity happens to be the view owner.

core/trino-main/src/main/java/io/trino/execution/RefreshViewTask.java
Praveen2112: > for example if there was a mechanism to hide some columns from some users

But the same would happen for executing the views as well right ? Today a user could create a view `SELECT * ...` which has 10 columns with invoker mode and now when another user praveen access them it would fail with column mismatch. It is more of an invoker mode limitation which is being seen here.

kasiafi: Agreed, I think we're getting to the same point. I the owner refreshes the view, we get the original intended semantics. If another user refreshes the view, they will impose their semantics not only for themselves but also for everyone who will execute that view later.

Regarding the current implementation: the view is refreshed
- by the owner for "definer" views
- by the current identity for "invoker" views

We want it to be always the owner. But the issue is that for "invoker" views we don't know who the owner is.

Praveen2112: > We want it to be always the owner. But the issue is that for "invoker" views we don't know who the owner is.

Yes - either we need to disable refreshing those views and maybe we could `set authorization` to set the owner and run them once again.

Praveen2112:
@kasiafi / @chenjian2664 Thanks a lot for the review. AC

Praveen2112:
@ebyhr / @kasiafi I did an additional test coverage for refreshing views with access control. PTAL.

chenjian2664 (approved):

PR #25907, opened 2025-06-02
Add tables with authorization information to system.metadata

homar:
## Description
Add tables with authorization information to system.metadata #25907

Trino supports ALTER (TABLE | FUNCTION | SCHEMA) SET AUTHORIZATION for quite some time.
However there is no way to retrieve this information. This commit fixes this by
introducing system.metadata.(tables | schemas | functions)_authorization tables.

## Release notes

```markdown
## Section
* Introduces system.metadata.(tables | schemas | functions)_authorization tables that expose the information about the authorization for given entities.
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
+++ b/core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
@@ -1,0 +1,159 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.connector.system;
+
+import com.google.common.collect.ImmutableList;
+import com.google.inject.Inject;
+import io.trino.FullConnectorSession;
+import io.trino.Session;
+import io.trino.metadata.Metadata;
+import io.trino.metadata.QualifiedObjectPrefix;
+import io.trino.security.AccessControl;
+import io.trino.spi.ErrorCodeSupplier;
+import io.trino.spi.TrinoException;
+import io.trino.spi.connector.CatalogSchemaName;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.connector.ConnectorTransactionHandle;
+import io.trino.spi.connector.InMemoryRecordSet;
+import io.trino.spi.connector.RecordCursor;
+import io.trino.spi.connector.SchemaTableName;
+import io.trino.spi.connector.SystemTable;
+import io.trino.spi.function.SchemaFunctionName;
+import io.trino.spi.predicate.Domain;
+import io.trino.spi.predicate.TupleDomain;
+import io.trino.spi.security.FunctionAuthorization;
+import io.trino.spi.security.TrinoPrincipal;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static io.trino.connector.system.jdbc.FilterUtil.tryGetSingleVarcharValue;
+import static io.trino.metadata.MetadataListing.listAllAvailableSchemas;
+import static io.trino.metadata.MetadataUtil.TableMetadataBuilder.tableMetadataBuilder;
+import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
+import static io.trino.spi.connector.SystemTable.Distribution.SINGLE_COORDINATOR;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static io.trino.spi.type.VarcharType.createUnboundedVarcharType;
+import static java.util.Objects.requireNonNull;
+
+public class FunctionsAuthorization
+        implements SystemTable
+{
+    public static final SchemaTableName FUNCTIONS_AUTHORIZATION_NAME = new SchemaTableName("metadata", "functions_authorization");
+
+    public static final ConnectorTableMetadata FUNCTIONS_AUTHORIZATION = tableMetadataBuilder(FUNCTIONS_AUTHORIZATION_NAME)
+            .column("catalog", createUnboundedVarcharType())
+            .column("schema", createUnboundedVarcharType())
+            .column("name", createUnboundedVarcharType())
+            .column("authorization_type", createUnboundedVarcharType())
+            .column("authorization", createUnboundedVarcharType())
+            .build();
+
+    private final Metadata metadata;
+    private final AccessControl accessControl;
+
+    @Inject
+    public FunctionsAuthorization(Metadata metadata, AccessControl accessControl)
+    {
+        this.metadata = requireNonNull(metadata, "metadata is null");
+        this.accessControl = requireNonNull(accessControl, "accessControl is null");
+    }
+
+    @Override
+    public Distribution getDistribution()
+    {
+        return SINGLE_COORDINATOR;
+    }
+
+    @Override
+    public ConnectorTableMetadata getTableMetadata()
+    {
+        return FUNCTIONS_AUTHORIZATION;
+    }
+
+    @Override
+    public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, ConnectorSession connectorSession, TupleDomain<Integer> constraint)
+    {
+        Session session = ((FullConnectorSession) connectorSession).getSession();
+        InMemoryRecordSet.Builder table = InMemoryRecordSet.builder(FUNCTIONS_AUTHORIZATION);
+        for (CatalogFunctionAuthorization functionAuthorization : getFunctionsAuthorization(session, constraint)) {
+            SchemaFunctionName schemaFunctionName = functionAuthorization.functionAuthorization().schemaFunctionName();
+            TrinoPrincipal trinoPrincipal = functionAuthorization.functionAuthorization().trinoPrincipal();
+            table.addRow(
+                    functionAuthorization.catalog(),
+                    schemaFunctionName.getSchemaName(),
+                    schemaFunctionName.getFunctionName(),
+                    trinoPrincipal.getType().toString(),
+                    trinoPrincipal.getName());
+        }
+        return table.build().cursor();
+    }
+
+    private List<CatalogFunctionAuthorization> getFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetFunctionsAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
+            ErrorCodeSupplier result = GENERIC_INTERNAL_ERROR;
+            if (exception instanceof TrinoException trinoException) {
+                result = trinoException::getErrorCode;
+            }
+            throw new TrinoException(
+                    result,
+                    "Error access functions_authorization metadata table",
+                    exception);
+        }
+    }
+
+    private List<CatalogFunctionAuthorization> doGetFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Domain catalogDomain = constraint.getDomain(0, VARCHAR);
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
+                session,
+                metadata,
+                accessControl,
+                catalogDomain,
+                schemaDomain);
+
+        Optional<String> functionName = tryGetSingleVarcharValue(constraint.getDomain(2, VARCHAR));
+        ImmutableList.Builder<CatalogFunctionAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
+            Set<FunctionAuthorization> allFunctionsAuthorization = metadata.getFunctionsAuthorizationInfo(session, new QualifiedObjectPrefix(catalogSchemaName.getCatalogName(), Optional.of(catalogSchemaName.getSchemaName()), functionName));
+            Set<SchemaFunctionName> filteredFunctions = accessControl.filterFunctions(
+                    session.toSecurityContext(),
+                    catalogSchemaName.getCatalogName(),
+                    allFunctionsAuthorization.stream().map(FunctionAuthorization::schemaFunctionName).collect(toImmutableSet()));
+
+            allFunctionsAuthorization.stream()
+                    .filter(functionAuthorization -> filteredFunctions.contains(functionAuthorization.schemaFunctionName()))
+                    .map(functionAuthorization -> new CatalogFunctionAuthorization(catalogSchemaName.getCatalogName(), functionAuthorization))
+                    .forEach(result::add);
+        });
+        return result.build();
+    }
+
+    private record CatalogFunctionAuthorization(String catalog, FunctionAuthorization functionAuthorization)
+    {
+        public CatalogFunctionAuthorization
+        {
+            requireNonNull(catalog, "catalog is null");
+            requireNonNull(functionAuthorization, "functionAuthorization is null");
+        }
+    }
+}
--- a/core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
+++ b/core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
@@ -1,0 +1,148 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.connector.system;
+
+import com.google.common.collect.ImmutableList;
+import com.google.inject.Inject;
+import io.trino.FullConnectorSession;
+import io.trino.Session;
+import io.trino.metadata.Metadata;
+import io.trino.metadata.QualifiedSchemaPrefix;
+import io.trino.security.AccessControl;
+import io.trino.spi.TrinoException;
+import io.trino.spi.connector.CatalogSchemaName;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.connector.ConnectorTransactionHandle;
+import io.trino.spi.connector.InMemoryRecordSet;
+import io.trino.spi.connector.RecordCursor;
+import io.trino.spi.connector.SchemaTableName;
+import io.trino.spi.connector.SystemTable;
+import io.trino.spi.predicate.Domain;
+import io.trino.spi.predicate.TupleDomain;
+import io.trino.spi.security.SchemaAuthorization;
+import io.trino.spi.security.TrinoPrincipal;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import static io.trino.connector.system.jdbc.FilterUtil.tryGetSingleVarcharValue;
+import static io.trino.metadata.MetadataListing.listAllAvailableSchemas;
+import static io.trino.metadata.MetadataUtil.TableMetadataBuilder.tableMetadataBuilder;
+import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
+import static io.trino.spi.connector.SystemTable.Distribution.SINGLE_COORDINATOR;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static io.trino.spi.type.VarcharType.createUnboundedVarcharType;
+import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.groupingBy;
+
+public class SchemasAuthorization
+        implements SystemTable
+{
+    public static final SchemaTableName SCHEMAS_AUTHORIZATION_NAME = new SchemaTableName("metadata", "schemas_authorization");
+
+    public static final ConnectorTableMetadata SCHEMAS_AUTHORIZATION = tableMetadataBuilder(SCHEMAS_AUTHORIZATION_NAME)
+            .column("catalog", createUnboundedVarcharType())
+            .column("schema", createUnboundedVarcharType())
+            .column("authorization_type", createUnboundedVarcharType())
+            .column("authorization", createUnboundedVarcharType())
+            .build();
+
+    private final Metadata metadata;
+    private final AccessControl accessControl;
+
+    @Inject
+    public SchemasAuthorization(Metadata metadata, AccessControl accessControl)
+    {
+        this.metadata = requireNonNull(metadata, "metadata is null");
+        this.accessControl = requireNonNull(accessControl, "accessControl is null");
+    }
+
+    @Override
+    public Distribution getDistribution()
+    {
+        return SINGLE_COORDINATOR;
+    }
+
+    @Override
+    public ConnectorTableMetadata getTableMetadata()
+    {
+        return SCHEMAS_AUTHORIZATION;
+    }
+
+    @Override
+    public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, ConnectorSession connectorSession, TupleDomain<Integer> constraint)
+    {
+        Session session = ((FullConnectorSession) connectorSession).getSession();
+        InMemoryRecordSet.Builder table = InMemoryRecordSet.builder(SCHEMAS_AUTHORIZATION);
+        for (CatalogSchemaAuthorization catalogSchemaAuthorization : getSchemasAuthorization(session, constraint)) {
+            TrinoPrincipal trinoPrincipal = catalogSchemaAuthorization.schemaAuthorization().trinoPrincipal();
+            table.addRow(
+                    catalogSchemaAuthorization.catalog(),
+                    catalogSchemaAuthorization.schemaAuthorization().schemaName(),
+                    trinoPrincipal.getType().toString(),
+                    trinoPrincipal.getName());
+        }
+        return table.build().cursor();
+    }
+
+    private List<CatalogSchemaAuthorization> getSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetSchemasAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
+            throw new TrinoException(
+                    GENERIC_INTERNAL_ERROR,
+                    "Error access schemas_authorizations metadata table",
+                    exception);
+        }
+    }
+
+    private List<CatalogSchemaAuthorization> doGetSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Domain catalogDomain = constraint.getDomain(0, VARCHAR);
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
+                session,
+                metadata,
+                accessControl,
+                catalogDomain,
+                schemaDomain);
+        Optional<String> schemaName = tryGetSingleVarcharValue(schemaDomain);
+        Map<String, Set<CatalogSchemaName>> groupedByCatalog = availableSchemas.stream()
+                .collect(groupingBy(CatalogSchemaName::getCatalogName, Collectors.toSet()));
+        ImmutableList.Builder<CatalogSchemaAuthorization> result = ImmutableList.builder();
+        for (String catalog : groupedByCatalog.keySet()) {
+            Set<SchemaAuthorization> allSchemasAuthorization = metadata.getSchemasAuthorizationInfo(session, new QualifiedSchemaPrefix(catalog, schemaName));
+            allSchemasAuthorization.stream()
+                    .filter(schemaAuthorization -> groupedByCatalog.get(catalog).contains(new CatalogSchemaName(catalog, schemaAuthorization.schemaName())))
+                    .map(schemaAuthorization -> new CatalogSchemaAuthorization(catalog, schemaAuthorization))
+                    .forEach(result::add);
+        }
+        return result.build();
+    }
+
+    private record CatalogSchemaAuthorization(String catalog, SchemaAuthorization schemaAuthorization)
+    {
+        public CatalogSchemaAuthorization
+        {
+            requireNonNull(catalog, "catalog is null");
+            requireNonNull(schemaAuthorization, "schemaAuthorization is null");
+        }
+    }
+}
--- a/core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
+++ b/core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
@@ -1,0 +1,147 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.connector.system;
+
+import com.google.common.collect.ImmutableList;
+import com.google.inject.Inject;
+import io.trino.FullConnectorSession;
+import io.trino.Session;
+import io.trino.metadata.Metadata;
+import io.trino.metadata.QualifiedTablePrefix;
+import io.trino.security.AccessControl;
+import io.trino.spi.TrinoException;
+import io.trino.spi.connector.CatalogSchemaName;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.connector.ConnectorTransactionHandle;
+import io.trino.spi.connector.InMemoryRecordSet;
+import io.trino.spi.connector.RecordCursor;
+import io.trino.spi.connector.SchemaTableName;
+import io.trino.spi.connector.SystemTable;
+import io.trino.spi.predicate.TupleDomain;
+import io.trino.spi.security.TableAuthorization;
+import io.trino.spi.security.TrinoPrincipal;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.Set;
+
+import static io.trino.connector.system.jdbc.FilterUtil.tryGetSingleVarcharValue;
+import static io.trino.metadata.MetadataListing.listAllAvailableSchemas;
+import static io.trino.metadata.MetadataListing.listTables;
+import static io.trino.metadata.MetadataUtil.TableMetadataBuilder.tableMetadataBuilder;
+import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
+import static io.trino.spi.connector.SystemTable.Distribution.SINGLE_COORDINATOR;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static io.trino.spi.type.VarcharType.createUnboundedVarcharType;
+import static java.util.Objects.requireNonNull;
+
+public class TablesAuthorization
+        implements SystemTable
+{
+    public static final SchemaTableName TABLES_AUTHORIZATION_NAME = new SchemaTableName("metadata", "tables_authorization");
+
+    public static final ConnectorTableMetadata TABLES_AUTHORIZATION = tableMetadataBuilder(TABLES_AUTHORIZATION_NAME)
+            .column("catalog", createUnboundedVarcharType())
+            .column("schema", createUnboundedVarcharType())
+            .column("name", createUnboundedVarcharType())
+            .column("authorization_type", createUnboundedVarcharType())
+            .column("authorization", createUnboundedVarcharType())
+            .build();
+
+    private final Metadata metadata;
+    private final AccessControl accessControl;
+
+    @Inject
+    public TablesAuthorization(Metadata metadata, AccessControl accessControl)
+    {
+        this.metadata = requireNonNull(metadata, "metadata is null");
+        this.accessControl = requireNonNull(accessControl, "accessControl is null");
+    }
+
+    @Override
+    public Distribution getDistribution()
+    {
+        return SINGLE_COORDINATOR;
+    }
+
+    @Override
+    public ConnectorTableMetadata getTableMetadata()
+    {
+        return TABLES_AUTHORIZATION;
+    }
+
+    @Override
+    public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, ConnectorSession connectorSession, TupleDomain<Integer> constraint)
+    {
+        Session session = ((FullConnectorSession) connectorSession).getSession();
+        InMemoryRecordSet.Builder table = InMemoryRecordSet.builder(TABLES_AUTHORIZATION);
+        for (CatalogTableAuthorization catalogTableAuthorization : getTablesAuthorization(session, constraint)) {
+            SchemaTableName schemaTableName = catalogTableAuthorization.tableAuthorization().schemaTableName();
+            TrinoPrincipal trinoPrincipal = catalogTableAuthorization.tableAuthorization.trinoPrincipal();
+            table.addRow(
+                    catalogTableAuthorization.catalog(),
+                    schemaTableName.getSchemaName(),
+                    schemaTableName.getTableName(),
+                    trinoPrincipal.getType().toString(),
+                    trinoPrincipal.getName());
+        }
+        return table.build().cursor();
+    }
+
+    private List<CatalogTableAuthorization> getTablesAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetTablesAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
+            throw new TrinoException(
+                    GENERIC_INTERNAL_ERROR,
+                    "Error access tables_authorization metadata table",
+                    exception);
+        }
+    }
+
+    private List<CatalogTableAuthorization> doGetTablesAuthorization(
+            Session session,
+            TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(session,
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
+                constraint.getDomain(1, VARCHAR));
+        Optional<String> tableName = tryGetSingleVarcharValue(constraint.getDomain(2, VARCHAR));
+        ImmutableList.Builder<CatalogTableAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
+            QualifiedTablePrefix prefix = new QualifiedTablePrefix(catalogSchemaName.getCatalogName(), Optional.of(catalogSchemaName.getSchemaName()), tableName);
+            Set<SchemaTableName> accessibleNames = listTables(session, metadata, accessControl, prefix);
+            Set<TableAuthorization> allTablesAuthorization = metadata.getTablesAuthorizationInfo(session, prefix);
+            allTablesAuthorization.stream()
+                    .filter(tableAuthorization -> accessibleNames.contains(tableAuthorization.schemaTableName()))
+                    .map(tableAuthorization -> new CatalogTableAuthorization(catalogSchemaName.getCatalogName(), tableAuthorization))
+                    .forEach(result::add);
+        });
+        return result.build();
+    }
+
+    private record CatalogTableAuthorization(String catalog, TableAuthorization tableAuthorization)
+    {
+        public CatalogTableAuthorization
+        {
+            requireNonNull(catalog, "catalog is null");
+            requireNonNull(tableAuthorization, "tableAuthorization is null");
+        }
+    }
+}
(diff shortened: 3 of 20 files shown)


core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -853,4 +856,19 @@ default boolean isMaterializedView(Session session, QualifiedObjectName viewName
...
+
+    /**
+     * Returns list of functions authorization info
+     */
+    Set<FunctionAuthorization> getAllFunctionsAuthorizationInfo(Session session, QualifiedTablePrefix prefix);
+
+    /**
+     * Returns list of schemas authorization info
+     */
+    Set<SchemaAuthorization> getAllSchemasAuthorizationInfo(Session session, QualifiedTablePrefix prefix);
```
lukasz-walkiewicz: nit: put schema first, then tables and functions last?

core/trino-main/src/main/java/io/trino/metadata/SystemSecurityMetadata.java
```diff
@@ -240,4 +243,19 @@ default void validateEntityKindAndPrivileges(Session session, String entityKind,
      * to be fully qualified, i.e., if the entity is a table, the name is of size three.
      */
     void setEntityOwner(Session session, EntityKindAndName entityKindAndName, TrinoPrincipal principal);
+
+    /**
+     * Returns all tables authorization information
+     */
+    Set<TableAuthorization> getAllTablesAuthorizationInfo(Session session, QualifiedTablePrefix prefix);
```
lukasz-walkiewicz: nit: I'd say `All` infix is redundant

testing/trino-tests/src/test/java/io/trino/security/TestingSystemSecurityMetadata.java
```diff
@@ -298,8 +311,76 @@ public void setEntityOwner(Session session, EntityKindAndName entityKindAndName,
...
+    @Override
+    public Set<TableAuthorization> getAllTablesAuthorizationInfo(Session session, QualifiedTablePrefix prefix)
+    {
+        return ImmutableSet.<TableAuthorization>builder()
+                .addAll(getOwnershipInfoInternal(tableOwners, prefix))
+                .addAll(getOwnershipInfoInternal(viewOwners, prefix))
+                .build();
+    }
+
+    private Set<TableAuthorization> getOwnershipInfoInternal(Map<CatalogSchemaTableName, Identity> owners, QualifiedTablePrefix prefix)
```
lukasz-walkiewicz: nit: just combine maps instead of extracting this method?

core/trino-main/src/main/java/io/trino/metadata/MetadataListing.java
```diff
@@ -85,6 +86,20 @@ public static List<CatalogInfo> listCatalogs(Session session, Metadata metadata,
                 .collect(toImmutableList());
     }
 
+    public static Set<CatalogSchemaName> listAllAvailableSchemas(
```
lukasz-walkiewicz, line 89: Is this method actually needed if all it does is delegating to `listCatalogNames` and `listSchemas`?

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
```diff
@@ -1472,6 +1476,164 @@ public void testAccessControlWithRolesAndRowFilter()
...
+        Session tableOwner2 = TestingSession.testSessionBuilder()
+                .setIdentity(Identity.ofUser(tableOwnerName2))
+                .setCatalog(getSession().getCatalog())
+                .setSchema(getSession().getSchema())
+                .build();
+
+        getQueryRunner().execute(tableOwner1, "CREATE TABLE %s (id INT)".formatted(table1));
+        assertQuery(
+                "SELECT * FROM system.metadata.tables_authorization",
+                "VALUES('blackhole', 'default', '%s', 'USER', '%s')".formatted(table1, tableOwnerName1));
```
lukasz-walkiewicz: That's the only test using blackhole catalog instead of memory_test, why?

core/trino-main/src/main/java/io/trino/metadata/MetadataListing.java
homar: it only does delegation BUT i would need to reuse this code in 3 places so I decided it is a good idea to actually extract it to a separated method

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
homar: because I was not able to create a function in blackhole catalog so functions are tested using memory (and memory_test to check that table returns functions from different catalogs). 
I can change this to also use memory

core/trino-main/src/main/java/io/trino/metadata/SystemSecurityMetadata.java
```diff
@@ -240,4 +243,19 @@ default void validateEntityKindAndPrivileges(Session session, String entityKind,
      * to be fully qualified, i.e., if the entity is a table, the name is of size three.
      */
     void setEntityOwner(Session session, EntityKindAndName entityKindAndName, TrinoPrincipal principal);
```
kokosing, line 245: Don't we need a generic function like this one? Also I am not a fun that in one place we call it "authorization" and in other "owner". Maybe we should rename `setEntityOwner` to `setEntityAuthorization`?

```
Map<String, TrinoPrincipal> getEntityAuthorizations(Session session, String entityKind);
Optional<TrinoPrincipal> getEntityAuthorization(Session session, EntityKindAndName entityKindAndName);
```
?

Also, if we would go this path then it means we just need `system.metadata.entity_authorization` table.

core/trino-main/src/main/java/io/trino/metadata/SystemSecurityMetadata.java
```diff
@@ -240,4 +243,19 @@ default void validateEntityKindAndPrivileges(Session session, String entityKind,
...
+
+    /**
+     * Returns all tables authorization information
```
kokosing: nit: This javadocs does not say more that the method name. I would drop it.

core/trino-main/src/main/java/io/trino/metadata/SystemSecurityMetadata.java
```diff
@@ -240,4 +243,19 @@ default void validateEntityKindAndPrivileges(Session session, String entityKind,
...
+
+    /**
+     * Returns all functions authorization information
+     */
+    Set<FunctionAuthorization> getAllFunctionsAuthorizationInfo(Session session, QualifiedTablePrefix prefix);
+
+    /**
+     * Returns all schemas authorization information
+     */
+    Set<SchemaAuthorization> getAllSchemasAuthorizationInfo(Session session, QualifiedTablePrefix prefix);
```
kokosing: It is odd to pass `QualifiedTablePrefix` as it is not a table. I mean if one passed `schema_name` parameter then you should return just that schema.

homar: when schema name is passed then it is easy, but this is an optimization for situation when only catalog is known, instead of returning all schemas from all catalogs when can just get schemas for a specific catalog

core/trino-main/src/main/java/io/trino/metadata/SystemSecurityMetadata.java
homar: I see your point, however I am not sure which path to follow now. Both makes sense, I think the generic one may be little more complex and there will be place in the code where finally we will map those generic types to tables + schemas + functions so I am not sure if this is worth the effort. 
Your call

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,136 @@
...
+        catch (RuntimeException exception) {
+            throw handleListingException(exception, "tables_authorization", "system");
+        }
+    }
+
+    private List<CatalogTableAuthorization> doGetTablesAuthorization(
+            Session session,
+            TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(session, metadata, accessControl, constraint.getDomain(0, VARCHAR));
```
kokosing: How do you know what is domain `0`?

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,138 @@
...
+            throw handleListingException(exception, "functions_authorization", "system");
+        }
+    }
+
+    private List<CatalogFunctionAuthorization> doGetFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(session, metadata, accessControl, constraint.getDomain(0, VARCHAR));
+
+        ImmutableList.Builder<CatalogFunctionAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
```
kokosing, line 136: I think it would be nice to enumerate only functions that in `constraint`

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    private List<CatalogSchemaAuthorization> doGetSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(session, metadata, accessControl, constraint.getDomain(0, VARCHAR));
+        Map<String, Set<CatalogSchemaName>> groupedByCatalog = availableSchemas.stream()
+                .collect(groupingBy(CatalogSchemaName::getCatalogName, Collectors.toSet()));
+        ImmutableList.Builder<CatalogSchemaAuthorization> result = ImmutableList.builder();
+        for (String catalog : groupedByCatalog.keySet()) {
+            Set<SchemaAuthorization> allSchemasAuthorization = metadata.getSchemasAuthorizationInfo(session, new QualifiedTablePrefix(catalog));
```
kokosing: I think it would be nice to enumerate only schemas that in `constraint`

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,136 @@
...
+        ImmutableList.Builder<CatalogTableAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
+            QualifiedTablePrefix prefix = new QualifiedTablePrefix(catalogSchemaName.getCatalogName(), catalogSchemaName.getSchemaName());
```
kokosing: I think it would be nice to enumerate only tables that in constraint

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
```diff
@@ -222,6 +225,24 @@ public void setEntityOwner(Session session, EntityKindAndName entityKindAndName,
         throw notSupportedException(entityKindAndName.name().get(0));
     }
 
+    @Override
+    public Set<SchemaAuthorization> getSchemasAuthorizationInfo(Session session, QualifiedTablePrefix prefix)
```
kokosing: We should not use `QualifiedTablePrefix` here as table has no sense

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
```diff
@@ -222,6 +225,24 @@ public void setEntityOwner(Session session, EntityKindAndName entityKindAndName,
...
+    }
+
+    @Override
+    public Set<TableAuthorization> getTablesAuthorizationInfo(Session session, QualifiedTablePrefix prefix)
+    {
+        return Set.of();
+    }
+
+    @Override
+    public Set<FunctionAuthorization> getFunctionsAuthorizationInfo(Session session, QualifiedTablePrefix prefix)
```
kokosing: We should not use QualifiedTablePrefix here as table has no sense

core/trino-main/src/main/java/io/trino/metadata/MetadataListing.java
```diff
@@ -347,7 +362,7 @@ private static Map<SchemaTableName, List<ColumnMetadata>> doListTableColumns(Ses
         return result.buildOrThrow();
     }
 
-    private static TrinoException handleListingException(RuntimeException exception, String type, String catalogName)
+    public static TrinoException handleListingException(RuntimeException exception, String type, String catalogName)
```
kokosing, line 367: can you add listing methods here and not make this method `public`?

core/trino-main/src/main/java/io/trino/metadata/SystemSecurityMetadata.java
kokosing: We discussed this offline and we believe making this generic, but we found problems like how to expose an entity in such table in generic way. Like not all entities have table name or schema name.

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
```diff
@@ -1472,6 +1476,165 @@ public void testAccessControlWithRolesAndRowFilter()
...
+                "SELECT * FROM system.metadata.schemas_authorization WHERE catalog = 'memory_test'",
+                "VALUES('memory_test', '%s', 'USER', '%s')".formatted(schema2, schemaOwnerName1));
+
+        getQueryRunner().execute(schemaOwner2, "DROP SCHEMA memory.%s".formatted(schema1));
+        getQueryRunner().execute(schemaOwner1, "DROP SCHEMA memory_test.%s".formatted(schema2));
+        assertQueryReturnsEmptyResult("SELECT * FROM system.metadata.schemas_authorization");
+    }
+
+    @Test
+    public void testTablesAuthorization()
```
kokosing, line 1550: add test for VIEW and MATERIALIZED VIEW

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
homar: I don't, I saw it in other places like https://github.com/trinodb/trino/blob/0f360f7f77a3e87deeda009532d7c9417baeb247/core/trino-main/src/main/java/io/trino/connector/system/TableCommentSystemTable.java#L99 and from my testing it looks like it works this way. So 0 is catalog, 1 is schema etc.

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
homar: So I believe it is a naming problem, I can copy code of `QualifiedTablePrefix` and name it `QualifiedObjectPrefix` but would that really make sense? 
I may rename `QualifiedTablePrefix` to `QualifiedObjectPrefix` but I don't believe it is a big deal actually

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
homar: https://github.com/trinodb/trino/pull/25907#discussion_r2157579826

core/trino-main/src/main/java/io/trino/metadata/MetadataListing.java
homar: it is not so easy, I don't see a good place for helper classes like `CatalogFunctionAuthorization` `CatalogSchemaAuthorization` and `CatalogTableAuthorization`

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
kokosing: > So I believe it is a naming problem, I can copy code of QualifiedTablePrefix and name it QualifiedObjectPrefix but would that really make sense?

Yes. Please. Also you may want them to become records. So then there is not that much copying :)

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
homar: VIEWS are covered by `testTablesAuthorization`  and MATERIALIZED VIEWS won't work without https://github.com/trinodb/trino/pull/25910/files#diff-438bae34efa7405d6c70ca5d7250d6348c6cab58176314eda60d24275d652c2f

homar: VIEWS are covered in `testViewColumnAccessControl`, unfortunately there is no  available connector in `TestAccessControl` that can create MATERIALIZED VIEWs.. So testing that is nor feasible right now

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+
+    private List<CatalogFunctionAuthorization> doGetFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
+                session,
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
```
kokosing: nit: extracting this as variable would make the code more readable. I need to go up and down to understand what `0` or `2` means.

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,146 @@
...
+                constraint.getDomain(1, VARCHAR));
+
+        Optional<String> functionName = tryGetSingleVarcharValue(constraint.getDomain(2, VARCHAR));
```
kokosing, line 134: But it can be a `Set` too.

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,146 @@
...
+        ImmutableList.Builder<CatalogFunctionAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
+            Set<FunctionAuthorization> allFunctionsAuthorization = metadata.getFunctionsAuthorizationInfo(session, new QualifiedObjectPrefix(catalogSchemaName.getCatalogName(), Optional.of(catalogSchemaName.getSchemaName()), functionName));
+            Set<String> accessibleFunctions = metadata.listFunctions(session, new CatalogSchemaName(catalogSchemaName.getCatalogName(), catalogSchemaName.getSchemaName())).stream()
+                    .flatMap(functionMetadata -> functionMetadata.getNames().stream())
+                    .collect(toImmutableSet());
+
+            allFunctionsAuthorization.stream()
+                    .filter(functionAuthorization -> accessibleFunctions.contains(functionAuthorization.schemaFunctionName().getFunctionName()))
```
kokosing: Hm.... maybe I was wrong. Now when I see this code I think we don't need to do it as it will be filtered anyway. When I requested filtering I wanted to avoid metadata calls, but once the call is done filtering more is not useful.

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
+                session,
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
+                schemaDomain);
+        Optional<String> schemaName = tryGetSingleVarcharValue(schemaDomain);
+        Map<String, Set<CatalogSchemaName>> groupedByCatalog = availableSchemas.stream()
+                .collect(groupingBy(CatalogSchemaName::getCatalogName, Collectors.toSet()));
```
kokosing, line 128: toImmutableSet()?

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+    private List<CatalogSchemaAuthorization> doGetSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
+                session,
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
+                schemaDomain);
+        Optional<String> schemaName = tryGetSingleVarcharValue(schemaDomain);
```
kokosing, line 126: same comments. It could be a set.

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+        Map<String, Set<CatalogSchemaName>> groupedByCatalog = availableSchemas.stream()
+                .collect(groupingBy(CatalogSchemaName::getCatalogName, Collectors.toSet()));
+        ImmutableList.Builder<CatalogSchemaAuthorization> result = ImmutableList.builder();
+        for (String catalog : groupedByCatalog.keySet()) {
+            Set<SchemaAuthorization> allSchemasAuthorization = metadata.getSchemasAuthorizationInfo(session, new QualifiedObjectPrefix(catalog, schemaName, Optional.empty()));
```
kokosing: `QualifiedObjectPrefix` does not fit here. `name` is not here and schema should be a set.

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    private List<CatalogSchemaAuthorization> doGetSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
+                session,
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
```
kokosing: nit: extract variable for it too

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+    private List<CatalogTableAuthorization> doGetTablesAuthorization(
+            Session session,
+            TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(session,
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
+                constraint.getDomain(1, VARCHAR));
+        Optional<String> tableName = tryGetSingleVarcharValue(constraint.getDomain(2, VARCHAR));
```
kokosing, line 125: same comments

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
```diff
@@ -222,6 +225,24 @@ public void setEntityOwner(Session session, EntityKindAndName entityKindAndName,
         throw notSupportedException(entityKindAndName.name().get(0));
     }
 
+    @Override
+    public Set<SchemaAuthorization> getSchemasAuthorizationInfo(Session session, QualifiedObjectPrefix prefix)
+    {
+        return Set.of();
```
kokosing, line 231: why not to throw? It sounds more defensive than that.

core/trino-main/src/main/java/io/trino/metadata/QualifiedObjectPrefix.java
```diff
@@ -0,0 +1,139 @@
...
+import java.util.Objects;
+import java.util.Optional;
+
+import static com.google.common.base.Verify.verify;
+import static io.trino.metadata.MetadataUtil.checkCatalogName;
+import static io.trino.metadata.MetadataUtil.checkSchemaName;
+import static io.trino.metadata.MetadataUtil.checkTableName;
+
+@Immutable
+public class QualifiedObjectPrefix
```
kokosing: why not record? So much boilerplate..

core/trino-main/src/test/java/io/trino/metadata/AbstractMockMetadata.java
```diff
@@ -1027,4 +1030,22 @@ public void setEntityAuthorization(Session session, EntityKindAndName entityKind
     {
         throw new UnsupportedOperationException();
     }
+
+    @Override
+    public Set<SchemaAuthorization> getSchemasAuthorizationInfo(Session session, QualifiedObjectPrefix prefix)
+    {
+        return ImmutableSet.of();
```
kokosing: throw?

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
homar: no, this set is being populated during collection

core/trino-main/src/main/java/io/trino/metadata/DisabledSystemSecurityMetadata.java
homar: Because other methods in this class don't throw

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
homar: How/When? Even if so, this doesn't break anything. If that domain contains more than 1 value Optional.empty will be returned. So the worst that can happen is lack of optimization. And this is exactly what is being done in other methods in `MetataListing`

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
homar: I think it is better this way

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
homar: If there is a single schema then this is an optimization, if instead there are many schemas, then Optional.empty will be here and all schemas will be checked.

kokosing: but still `name` has no value and it is misleading toto see `QualifiedObjectPrefix` which is a triple name, to find things that is a pair

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
kokosing: 👍

kokosing: > How/When? 

`WHERE function_name IN ('one', 'two")`

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
kokosing: I think it unnecessarly complicates the logic here. I wanted to try eliminate the number of metadata calls. Filtering the set here does not improve the performance much and makes code here complex.

Please apply the same comment in other places.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -2891,4 +2894,56 @@ private static boolean cannotExist(QualifiedObjectName name)
     {
         return name.catalogName().isEmpty() || name.schemaName().isEmpty() || name.objectName().isEmpty();
     }
+
+    @Override
+    public Set<SchemaAuthorization> getSchemasAuthorizationInfo(Session session, QualifiedObjectPrefix prefix)
```
kokosing: Instead of passing `QualifiedObjectPrefix`, pass `String catalog, Optional<String> schema`, or some `QualifiedSchemaPrefix`

core/trino-main/src/main/java/io/trino/metadata/QualifiedObjectPrefix.java
```diff
@@ -0,0 +1,75 @@
...
+        String catalogName,
+        Optional<String> schemaName,
+        Optional<String> objectName)
+{
+    public QualifiedObjectPrefix(String catalogName, Optional<String> schemaName)
+    {
+        this(catalogName, schemaName, Optional.empty());
+    }
+
+    @JsonCreator
```
kokosing: I don't think you need this if you use `record`. Same for getters.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
homar: Sorry but I don't agree. I believe `QualifiedObjectPrefix` should be treated as generic and could be used in any scenario - can represent as short path as the one to  catalog and as long as the one to table. Currently it is kind of hardcoded but I believe it can be reimplemented to support prefixes of any length if necessary

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,149 @@
...
+        return table.build().cursor();
+    }
+
+    private List<CatalogFunctionAuthorization> getFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetFunctionsAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
+            throw handleListingException(exception, "functions_authorization", "system");
```
kokosing: This is wrong. See https://github.com/trinodb/trino/commit/e9b931f500d1dee968cee81ded4ae0f86ee4158e

It is obvious that query on `system` catalog in Trino fails in this case. We care here about the catalog from `constraint.getDomain(0, VARCHAR);`.

However ownership is taken from `SystemSecurityMetadata` not a connector metadata so we could potentially skip that.

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
kokosing: ping

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,144 @@
...
+        return table.build().cursor();
+    }
+
+    private List<CatalogSchemaAuthorization> getSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetSchemasAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
+            throw handleListingException(exception, "schemas_authorization", "system");
```
kokosing: same

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,144 @@
...
+                catalogDomain,
+                schemaDomain);
+        Optional<String> schemaName = tryGetSingleVarcharValue(schemaDomain);
+        Map<String, Set<CatalogSchemaName>> groupedByCatalog = availableSchemas.stream()
+                .collect(groupingBy(CatalogSchemaName::getCatalogName, Collectors.toSet()));
+        ImmutableList.Builder<CatalogSchemaAuthorization> result = ImmutableList.builder();
+        for (String catalog : groupedByCatalog.keySet()) {
+            Set<SchemaAuthorization> allSchemasAuthorization = metadata.getSchemasAuthorizationInfo(session, new QualifiedSchemaPrefix(catalog, schemaName));
+            allSchemasAuthorization.stream()
+                    .filter(schemaAuthorization -> groupedByCatalog.get(catalog).contains(new CatalogSchemaName(catalog, schemaAuthorization.schemaName())))
```
kokosing, line 133: There is no need to filter it. You already passed schema filter to `getSchemasAuthorizationInfo`

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+        return table.build().cursor();
+    }
+
+    private List<CatalogTableAuthorization> getTablesAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetTablesAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
+            throw handleListingException(exception, "tables_authorization", "system");
```
kokosing: same

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,143 @@
...
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
+                constraint.getDomain(1, VARCHAR));
+        Optional<String> tableName = tryGetSingleVarcharValue(constraint.getDomain(2, VARCHAR));
+        ImmutableList.Builder<CatalogTableAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
+            QualifiedTablePrefix prefix = new QualifiedTablePrefix(catalogSchemaName.getCatalogName(), Optional.of(catalogSchemaName.getSchemaName()), tableName);
+            Set<SchemaTableName> accessibleNames = listTables(session, metadata, accessControl, prefix);
+            Set<TableAuthorization> allTablesAuthorization = metadata.getTablesAuthorizationInfo(session, prefix);
+            allTablesAuthorization.stream()
```
kokosing, line 131: same with filtering. You already passed prefix to `getTablesAuthorizationInfo`

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -2891,4 +2894,56 @@ private static boolean cannotExist(QualifiedObjectName name)
...
+        if (catalog.isPresent()) {
+            CatalogMetadata catalogMetadata = catalog.get();
+
+            if (catalogMetadata.getSecurityManagement() == SYSTEM) {
+                schemasAuthorization.addAll(systemSecurityMetadata.getSchemasAuthorizationInfo(session, prefix));
+            }
+        }
+        return schemasAuthorization.build();
+    }
+
```
kokosing:
```java
        requireNonNull(prefix, "prefix is null");

        Optional<CatalogMetadata> catalog = getOptionalCatalogMetadata(session, prefix.getCatalogName());
        if (catalog.filter(SYSTEM::equals).present) {
                return systemSecurityMetadata.getSchemasAuthorizationInfo(session, prefix);
        }
        return ImmutableSet.of();
        
```

core/trino-main/src/main/java/io/trino/metadata/QualifiedObjectPrefix.java
```diff
@@ -0,0 +1,68 @@
...
+            String catalogName,
+            Optional<String> schemaName,
+            Optional<String> objectName)
+    {
+        this.catalogName = checkCatalogName(catalogName);
+        this.schemaName = schemaName;
+        this.objectName = objectName;
+    }
+
+    public String getCatalogName()
```
kokosing: there is no need to write getters for record

core/trino-main/src/main/java/io/trino/metadata/QualifiedSchemaPrefix.java
```diff
@@ -0,0 +1,54 @@
...
+        String catalogName,
+        Optional<String> schemaName)
+{
+    public QualifiedSchemaPrefix(String catalogName, Optional<String> schemaName)
+    {
+        this.catalogName = checkCatalogName(catalogName);
+        this.schemaName = schemaName;
+    }
+
+    public String getCatalogName()
```
kokosing: same here.

core/trino-main/src/main/java/io/trino/metadata/QualifiedObjectPrefix.java
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    public QualifiedObjectPrefix(
+            String catalogName,
+            Optional<String> schemaName,
+            Optional<String> objectName)
+    {
+        this.catalogName = checkCatalogName(catalogName);
+        this.schemaName = schemaName;
+        this.objectName = objectName;
```
kokosing: there is no need to assign things in record constructor. 

You can call `requireNonNull` if you like

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
homar: `schemaName` is taken from the constraint, it can be empty, in that case `getSchemasAuthorizationInfo` will return all schemas so I still need to filter. In case it is provided there will be only 1 result so it is fine

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
homar: same answer, that filtering may not happen as the source of that value is in constraint `TupleDomain` which may be empty

core/trino-main/src/main/java/io/trino/metadata/QualifiedObjectPrefix.java
homar: ehh i used copilot and it failed miserably.. (i did fail to as I didn't check what it did..)

core/trino-main/src/main/java/io/trino/metadata/QualifiedObjectPrefix.java
homar: yes, right, again copilot generated too much

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
```diff
@@ -1479,6 +1484,191 @@ public void testAccessControlWithRolesAndRowFilter()
         assertAccessAllowed(session, "SELECT nationkey FROM nation");
     }
 
+
+    @Test
+    public void testSchemasAuthorization()
```
kokosing, line 1489: Do you have a test where user has no access to a schema so they cannot see it in `schemas_authorization`?

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
```diff
@@ -1479,6 +1484,191 @@ public void testAccessControlWithRolesAndRowFilter()
...
+                "VALUES('memory', '%s', 'USER', '%s'), ('memory_test', '%s', 'USER', '%s')".formatted(schema1, schemaOwnerName2, schema2, schemaOwnerName1));
+
+        getQueryRunner().execute(schemaOwner2, "DROP SCHEMA memory.%s".formatted(schema1));
+        getQueryRunner().execute(schemaOwner1, "DROP SCHEMA memory_test.%s".formatted(schema2));
+        getQueryRunner().execute(schemaOwner1, "DROP SCHEMA memory_test.%s".formatted(deniedSchema));
+        assertQueryReturnsEmptyResult("SELECT * FROM system.metadata.schemas_authorization");
+    }
+
+    @Test
+    public void testTablesAuthorization()
```
kokosing, line 1550: Do you have a test where user has no access to a table/view so they cannot see it in `tables_authorization`?

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
```diff
@@ -1479,6 +1484,191 @@ public void testAccessControlWithRolesAndRowFilter()
...
+
+        getQueryRunner().execute(tableOwner2, "DROP TABLE memory.default." + table1);
+        getQueryRunner().execute(tableOwner1, "DROP TABLE memory_test.default." + table2);
+        getQueryRunner().execute(tableOwner1, "DROP VIEW memory.default." + view);
+        getQueryRunner().execute(tableOwner1, "DROP TABLE memory_test.default." + deniedTable);
+        assertQueryReturnsEmptyResult("SELECT * FROM system.metadata.tables_authorization");
+    }
+
+    @Test
+    public void testFunctionsAuthorization()
```
kokosing, line 1612: Do you have a test where user has no access to a schema so they cannot see it in `function_authorization`?

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+        return table.build().cursor();
+    }
+
+    private List<CatalogFunctionAuthorization> getFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        try {
+            return doGetFunctionsAuthorization(session, constraint);
+        }
+        catch (RuntimeException exception) {
```
kokosing, line 111: What if `RuntimeException` is already `TrinoException`?

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,148 @@
...
+                session,
+                metadata,
+                accessControl,
+                catalogDomain,
+                schemaDomain);
+        Optional<String> schemaName = tryGetSingleVarcharValue(schemaDomain);
+        Map<String, Set<CatalogSchemaName>> groupedByCatalog = availableSchemas.stream()
+                .collect(groupingBy(CatalogSchemaName::getCatalogName, Collectors.toSet()));
+        ImmutableList.Builder<CatalogSchemaAuthorization> result = ImmutableList.builder();
+        for (String catalog : groupedByCatalog.keySet()) {
```
kokosing, line 130: I don't see `io.trino.security.AccessControl#filterSchemas` called below.

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
```diff
@@ -0,0 +1,148 @@
...
+                    "Error access schemas_authorizations metadata table",
+                    exception);
+        }
+    }
+
+    private List<CatalogSchemaAuthorization> doGetSchemasAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Domain catalogDomain = constraint.getDomain(0, VARCHAR);
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
```
kokosing, line 120: is this using `io.trino.security.AccessControl#filterCatalogs` or `io.trino.security.AccessControl#filterSchemas`?

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
```diff
@@ -0,0 +1,154 @@
...
+                    "Error access functions_authorization metadata table",
+                    exception);
+        }
+    }
+
+    private List<CatalogFunctionAuthorization> doGetFunctionsAuthorization(Session session, TupleDomain<Integer> constraint)
+    {
+        Domain catalogDomain = constraint.getDomain(0, VARCHAR);
+        Domain schemaDomain = constraint.getDomain(1, VARCHAR);
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(
```
kokosing, line 127: is this using `io.trino.security.AccessControl#filterCatalogs` or `io.trino.security.AccessControl#filterSchemas`?

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,147 @@
...
+                    "Error access tables_authorization metadata table",
+                    exception);
+        }
+    }
+
+    private List<CatalogTableAuthorization> doGetTablesAuthorization(
+            Session session,
+            TupleDomain<Integer> constraint)
+    {
+        Set<CatalogSchemaName> availableSchemas = listAllAvailableSchemas(session,
```
kokosing, line 120: is this using `io.trino.security.AccessControl#filterCatalogs` or `io.trino.security.AccessControl#filterSchemas`?

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
```diff
@@ -0,0 +1,147 @@
...
+                metadata,
+                accessControl,
+                constraint.getDomain(0, VARCHAR),
+                constraint.getDomain(1, VARCHAR));
+        Optional<String> tableName = tryGetSingleVarcharValue(constraint.getDomain(2, VARCHAR));
+        ImmutableList.Builder<CatalogTableAuthorization> result = ImmutableList.builder();
+        availableSchemas.forEach(catalogSchemaName -> {
+            QualifiedTablePrefix prefix = new QualifiedTablePrefix(catalogSchemaName.getCatalogName(), Optional.of(catalogSchemaName.getSchemaName()), tableName);
+            Set<SchemaTableName> accessibleNames = listTables(session, metadata, accessControl, prefix);
```
kokosing, line 129: I don't see `io.trino.security.AccessControl#filterTables` called here.

core/trino-main/src/main/java/io/trino/metadata/QualifiedSchemaPrefix.java
```diff
@@ -0,0 +1,54 @@
...
+import static io.trino.metadata.MetadataUtil.checkCatalogName;
+
+public record QualifiedSchemaPrefix(
+        String catalogName,
+        Optional<String> schemaName)
+{
+    public QualifiedSchemaPrefix(String catalogName, Optional<String> schemaName)
+    {
+        this.catalogName = checkCatalogName(catalogName);
+        this.schemaName = schemaName;
```
kokosing: nit: no need to assign anything here :)

core/trino-main/src/main/java/io/trino/metadata/QualifiedSchemaPrefix.java
kokosing: ping

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
homar: Check last assertion in this test

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
homar: check last assertion in this test

testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java
homar: check last assertion in this test

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
homar: That's good because it is not here :) It is called above inside `io.trino.metadata.MetadataListing#listAllAvailableSchemas`

core/trino-main/src/main/java/io/trino/connector/system/SchemasAuthorization.java
homar: actually both

core/trino-main/src/main/java/io/trino/connector/system/FunctionsAuthorization.java
homar: both

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
homar: both

core/trino-main/src/main/java/io/trino/connector/system/TablesAuthorization.java
homar: good, it is called inside `io.trino.metadata.MetadataListing#listTables`

kokosing:
Thank you!

PR #25910, opened 2025-06-03
[WIP] Add support for ALTER MATERIALIZED VIEW .. SET AUTHORIZATION

codyzwief:
This commit adds support for `SET AUTHORIZATION` on a materialized view. Previously, this was only available on views.

## Description

WIP.

## Release notes

```markdown
## Section
```


codyzwief:
Hi @martint and/or @kokosing  -- would either of you be able to take a look at this PR when you get a free moment? Please feel free to reassign if there is someone who would be better equipped to review this change. Thanks!

core/trino-parser/src/main/java/io/trino/sql/parser/AstBuilder.java
```diff
@@ -511,9 +513,13 @@ public Node visitRenameSchema(SqlBaseParser.RenameSchemaContext context)
     @Override
     public Node visitSetAuthorization(SqlBaseParser.SetAuthorizationContext context)
     {
+        OwnedEntityKindContext ownedEntityKindContext = context.ownedEntityKind();
+        String ownedEntityKind = IntStream.range(0, ownedEntityKindContext.getChildCount())
+                .mapToObj(i -> ownedEntityKindContext.getChild(i).getText())
+                .reduce("", (a, b) -> String.format("%s %s", a, b).trim());
```
kokosing, line 519: Why this is needed?

lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/security/BaseFileBasedSystemAccessControlTest.java
```diff
@@ -1712,57 +1716,62 @@ public void testViewAuthorization()
         assertAccessDenied(
                 () -> accessControl.checkCanSetEntityAuthorization(user("user", "group", "role"), new EntityKindAndName("VIEW", table), new TrinoPrincipal(ROLE, "new_role")),
                 "Cannot set authorization for view some-catalog.test.table to ROLE new_role");
-
```
kokosing, line 1715: I think it would be better to extract a method and have to `@Test` methods one per each view kind.

kokosing (approved):
% comments

homar:
Are you sure we don't need specific methods for materialized views both in `SystemAccessControl` and in `ConnectorMetadata` ? I may be wrong about that but I believe it is safer to have materialized views distinguished from views.

codyzwief:
> Are you sure we don't need specific methods for materialized views both in SystemAccessControl and in ConnectorMetadata ? I may be wrong about that but I believe it is safer to have materialized views distinguished from views.

I mentioned that in the PR description:

> In this PR, I reused checkCanSetViewAuthorization for access control for materialized views. Certainly open to creating a new one to differentiate between views & MVs.

I think you're probably right that it's safer, even if in most cases it'll be the same code. I'll make that change.

lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/security/BaseFileBasedSystemAccessControlTest.java
codyzwief: Done!

core/trino-parser/src/main/java/io/trino/sql/parser/AstBuilder.java
codyzwief: It's not necessary, but it's for keeping the owned entity kind to keep the spaces that `MATERIALIZED VIEW` has in it. Otherwise, the entity kind string would need to be `MATERIALIZEDVIEW`. I thought the former was more user friendly.

Happy to revert this change back and use `MATERIALIZEDVIEW` instead -- I can see the pros and cons of both approach.

homar (approved):

lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/security/BaseFileBasedSystemAccessControlTest.java
kokosing: Way better to review!

kokosing (approved):

kokosing:
I retried the failed job, let me know when it becomes green so I will merge it.

codyzwief:
@kokosing Thank you for rerunning the failure. It succeeded, is it OK to merge?

PR #25924, opened 2025-06-03
Ignore rowsCount during serialization

wendigo:
## Release notes

```markdown
## Section
```


wendigo:
@raunaqmorarka unnecessary info that is ignored by the client during deserialization but can break other clients (non-java). It was introduced in that version (still unreleased) so not affecting users

PR #25945, opened 2025-06-05
Drop ConnectorSession from Type.getObjectValue

wendigo:
Type's object value retrieval should be independent of the session
but previously it was coupled to JSON serialization.

Making types independent of the serialization, moves the formatting logic
to protocol layer.

## Release notes

```markdown
## SPI
* Remove ConnectorSession from Type.getObjectValue ({issue}`issuenumber`)
```


wendigo:
@dain wdyt?

martint:
If I recall, this was only needed for legacy timestamps, which relied on the session timezone to compute their representation. I think it's a good change.

The only thing I'd raise is that it removes the only way to customize serialization dynamically in case we need to introduce a change in behavior and want to allow users to roll it out gradually.

wendigo:
@martint legacy timestamps are still supported, it just happens in the serialization code, not in the Type SPI code.

wendigo:
@martint I'm still thinking about adding a SPI that allows a custom type (not a built-in) to describe how it wants to be represented over the wire (using built-in types)

PR #25954, opened 2025-06-06
Random improvements to random things in random places

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
+++ b/core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
@@ -30,8 +30,12 @@ import jakarta.ws.rs.core.Response;
 
 import javax.crypto.SecretKey;
 
+import java.time.Instant;
 import java.time.ZonedDateTime;
 import java.util.Date;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.function.Function;
+import java.util.function.Supplier;
 
 import static io.airlift.http.client.Request.Builder.fromRequest;
 import static io.jsonwebtoken.security.Keys.hmacShaKeyFor;
@@ -41,18 +45,23 @@ import static io.trino.server.security.jwt.JwtUtil.newJwtParserBuilder;
 import static jakarta.ws.rs.core.MediaType.TEXT_PLAIN_TYPE;
 import static jakarta.ws.rs.core.Response.Status.UNAUTHORIZED;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.time.temporal.ChronoUnit.MINUTES;
 import static java.util.Objects.requireNonNull;
 
 public class InternalAuthenticationManager
         implements HttpRequestFilter
 {
     private static final Logger log = Logger.get(InternalAuthenticationManager.class);
+    private static final Supplier<Instant> DEFAULT_EXPIRATION_SUPPLIER = () -> ZonedDateTime.now().plusMinutes(6).toInstant();
+    // Leave a 5 minute buffer to allow for clock skew and GC pauses
+    private static final Function<Instant, Instant> TOKEN_REUSE_THRESHOLD = instant -> instant.minus(5, MINUTES);
 
     private static final String TRINO_INTERNAL_BEARER = "X-Trino-Internal-Bearer";
 
     private final SecretKey hmac;
     private final String nodeId;
     private final JwtParser jwtParser;
+    private final AtomicReference<InternalToken> currentToken;
 
     @Inject
     public InternalAuthenticationManager(InternalCommunicationConfig internalCommunicationConfig, SecurityConfig securityConfig, NodeInfo nodeInfo)
@@ -84,6 +93,7 @@ public class InternalAuthenticationManager
         this.hmac = hmacShaKeyFor(Hashing.sha256().hashString(sharedSecret, UTF_8).asBytes());
         this.nodeId = nodeId;
         this.jwtParser = newJwtParserBuilder().verifyWith(hmac).build();
+        this.currentToken = new AtomicReference<>(createJwt());
     }
 
     public static boolean isInternalRequest(ContainerRequestContext request)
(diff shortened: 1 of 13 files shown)


losipiuk:
Fuzzer driven development?

core/trino-main/src/main/java/io/trino/sql/analyzer/TypeSignatureTranslator.java
```diff
@@ -65,6 +66,15 @@
 
 public final class TypeSignatureTranslator
 {
+    private static final CharMatcher FIRST_CHAR_DISALLOWED_MATCHER = CharMatcher.inRange('0', '9')
```
losipiuk: name it IS_DIGIT

core/trino-main/src/main/java/io/trino/sql/analyzer/TypeSignatureTranslator.java
```diff
@@ -65,6 +66,15 @@
...
+            .precomputed();
+
+    private static final CharMatcher ALLOWED_CHARS_MATCHER = CharMatcher.inRange('a', 'z')
```
losipiuk: IS_VALID_IDENTIFIER_CHAR

core/trino-main/src/main/java/io/trino/sql/planner/iterative/RuleIndex.java
```diff
@@ -14,34 +14,54 @@
...
 import com.google.common.reflect.TypeToken;
+import io.trino.cache.EvictableCacheBuilder;
 import io.trino.matching.Pattern;
 import io.trino.matching.pattern.TypeOfPattern;
 
 import java.util.Set;
-import java.util.stream.Stream;
+import java.util.concurrent.ExecutionException;
 
 public class RuleIndex
```
losipiuk, line 29:
```java
public class RuleIndex
{
    private final ListMultimap<Class<?>, Rule<?>> rulesByRootType;
    private final Cache<Class<?>, Set<Rule<?>>> rulesByClass;

    private RuleIndex(ListMultimap<Class<?>, Rule<?>> rulesByRootType)
    {
        this.rulesByRootType = ImmutableListMultimap.copyOf(rulesByRootType);

        rulesByClass = EvictableCacheBuilder.newBuilder()
                .maximumSize(128) // we have a limited number of node types, so this is more than enough
                .build();
    }

    public Set<Rule<?>> getCandidates(Object object)
    {
        try {
            return rulesByClass.get(object.getClass(), () -> computeCandidates(object.getClass()));
        }
        catch (ExecutionException e) {
        }
    }

    public Set<Rule<?>> computeCandidates(Class<?> key)
    {
        ImmutableSet.Builder<Rule<?>> builder = ImmutableSet.builder();
        TypeToken.of(key).getTypes().forEach(clazz -> {
            Class<?> rawType = clazz.getRawType();
            if (rulesByRootType.containsKey(rawType)) {
                builder.addAll(rulesByRootType.get(rawType));
            }
        });
        return builder.build();
    }
```

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/ExpressionRewriteRuleSet.java
```diff
@@ -304,9 +304,11 @@ public Result apply(ValuesNode valuesNode, Captures captures, Context context)
                 Expression rewritten;
                 if (row instanceof Row value) {
                     // preserve the structure of row
-                    rewritten = new Row(value.items().stream()
-                            .map(item -> rewriter.rewrite(item, context))
-                            .collect(toImmutableList()));
+                    ImmutableList.Builder<Expression> rowValues = ImmutableList.builderWithExpectedSize(value.items().size());
```
losipiuk, line 307: original looks nicer.

core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
```diff
@@ -118,17 +125,26 @@ public void handleInternalRequest(ContainerRequestContext request)
     public Request filterRequest(Request request)
     {
         return fromRequest(request)
-                .addHeader(TRINO_INTERNAL_BEARER, generateJwt())
+                .addHeader(TRINO_INTERNAL_BEARER, getOrGenerateJwt())
                 .build();
     }
 
-    private String generateJwt()
+    private synchronized String getOrGenerateJwt()
```
losipiuk: this is at cost of synchronization. Not sure worth it.

...

Actually I think this is wrong. With this change you can use a token which is "almost" expired. Risking communicatino failure. If you really need this change (which I think you do not) you should refresh token significantly earlier than expiration time. Maybe half way.

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/ExpressionRewriteRuleSet.java
wendigo: Yeah but resizes the list instead of preallocating it

wendigo: > Returns a new builder, expecting the specified number of elements to be added.
If expectedSize is exactly the number of elements added to the builder before ImmutableList.Builder.build is called, the builder is likely to perform better than an unsized builder() would have.

wendigo: and the `toImmutableList()` is:

```
  private static final Collector<Object, ?, ImmutableList<Object>> TO_IMMUTABLE_LIST =
      Collector.of(
          ImmutableList::builder,
          ImmutableList.Builder::add,
          ImmutableList.Builder::combine,
          ImmutableList.Builder::build);
```

wendigo: Nicer != more performant :P

losipiuk: Meh - I doubt it matters at all, so I would keep nicer. I will not fight though.

martint: Have we observed a measurable performance improvement? If not, then readability is more important.

Issue #8742, opened 2021-08-01
Trino-Prometheus connector: Enable filter labels pushdown to the api/v1/query endpoint in addition to the timestamp and values filters


nikitag55:
As of now, In a query like below:

cur.execute("""SELECT * FROM http_request_total WHERE labels['instance'] = 'grafana' AND timestamp > (NOW() - INTERVAL '1' DAY)""")

Trino sends the request to Prometheus query endpoint is **without** `labels['instance']` and the request to the query endpoint looks like this: 

`api/v1/query?query= http_request_total[5m]&time=123456`

and when the above request is successful, Trino applies these filters `labels['instance']` later on.

Instead the request which Trino should send to Prometheus should be like this:

`api/v1/query?query= http_request_total{instance='grafana'}[5m]&time=123456`

No availability of labels pushdown results in lot of issues:

When the user is aware of what exact labels they want to query, Trino should not drop those labels and apply the filters later on. If it doesn't do so, lot of time, memory, network is wasted in touching all the raw data from Prometheus which doesn't make sense and sometimes requests just timeout's because of retrieving large amount of data from Prometheus. Instead the filter labels should go in the `/query` endpoint request itself.

jirislav:
Hey Trino team, I have implemented this feature, can you please take a look at my Pull Request [here](https://github.com/trinodb/trino/pull/25957)?

PR #25957, opened 2025-06-09
Pushdown Prometheus Predicates #8742

jirislav:
## Description

Implements pushdown for Prometheus predicates, as requested in #8742.

## Release notes

```markdown
## Prometheus Connector
* Support predicate pushdown for equality, non-equality, LIKE & IN operators. (#8742)
```

### DIFF
--- a/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusMetadata.java
+++ b/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusMetadata.java
@@ -17,6 +17,8 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
+import io.trino.plugin.base.expression.ConnectorExpressionRewriter;
+import io.trino.plugin.prometheus.expression.LabelFilterExpression;
 import io.trino.spi.TrinoException;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ColumnMetadata;
@@ -31,6 +33,8 @@ import io.trino.spi.connector.RelationColumnsMetadata;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.SchemaTablePrefix;
 import io.trino.spi.connector.TableNotFoundException;
+import io.trino.spi.expression.ConnectorExpression;
+import io.trino.spi.predicate.TupleDomain;
 
 import java.util.HashMap;
 import java.util.Iterator;
@@ -82,7 +86,7 @@ public class PrometheusMetadata
             return null;
         }
 
-        return new PrometheusTableHandle(tableName.getSchemaName(), tableName.getTableName(), Optional.empty());
+        return new PrometheusTableHandle(tableName.getSchemaName(), tableName.getTableName(), TupleDomain.all(), ImmutableList.of());
     }
 
     @Override
@@ -173,8 +177,13 @@ public class PrometheusMetadata
     @Override
     public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(ConnectorSession session, ConnectorTableHandle handle, Constraint constraint)
     {
+        ConnectorExpression expression = constraint.getExpression();
+
+        ConnectorExpressionRewriter<List<LabelFilterExpression>> rewriter = new ConnectorExpressionRewriter<>(LabelFilterExpression.APPLICABLE_EXPRESSION_RULES);
+
         PrometheusTableHandle tableHandle = ((PrometheusTableHandle) handle)
-                .withPredicate(constraint.getSummary());
+                .withPredicate(constraint.getSummary())
+                .withExpressions(rewriter.rewrite(session, expression, constraint.getAssignments()).orElseGet(ImmutableList::of));
         return Optional.of(new ConstraintApplicationResult<>(tableHandle, constraint.getSummary(), constraint.getExpression(), false));
     }
 }
--- a/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusQueryResponseParse.java
+++ b/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusQueryResponseParse.java
@@ -53,17 +53,15 @@ public class PrometheusQueryResponseParse
                         this.status = true;
                         while (!parser.isClosed()) {
                             parser.nextToken();
-                            if (JsonToken.FIELD_NAME.equals(jsonToken)) {
-                                if (parser.currentName().equals("resultType")) {
-                                    parser.nextToken();
-                                    resultType = parser.getValueAsString();
-                                }
-                                if (parser.currentName().equals("result")) {
-                                    parser.nextToken();
-                                    ArrayNode node = mapper.readTree(parser);
-                                    result = node.toString();
-                                    break;
-                                }
+                            if (parser.currentName().equals("resultType")) {
+                                parser.nextToken();
+                                resultType = parser.getValueAsString();
+                            }
+                            if (parser.currentName().equals("result")) {
+                                parser.nextToken();
+                                ArrayNode node = mapper.readTree(parser);
+                                result = node.toString();
+                                break;
                             }
                         }
                     }
@@ -85,19 +83,17 @@ public class PrometheusQueryResponseParse
             }
         }
         if (result != null && resultType != null) {
-            switch (resultType) {
-                case "matrix":
-                case "vector":
-                    results = mapper.readValue(result, new TypeReference<List<PrometheusMetricResult>>() {});
-                    break;
-                case "scalar":
-                case "string":
+            results = switch (resultType) {
+                case "matrix", "vector" -> mapper.readValue(result, new TypeReference<List<PrometheusMetricResult>>() {});
+                case "scalar", "string" -> {
                     PrometheusTimeSeriesValue stringOrScalarResult = mapper.readValue(result, new TypeReference<PrometheusTimeSeriesValue>() {});
                     Map<String, String> madeUpMetricHeader = new HashMap<>();
                     madeUpMetricHeader.put("__name__", resultType);
                     PrometheusTimeSeriesValueArray timeSeriesValues = new PrometheusTimeSeriesValueArray(singletonList(stringOrScalarResult));
-                    results = singletonList(new PrometheusMetricResult(madeUpMetricHeader, timeSeriesValues));
-            }
+                    yield singletonList(new PrometheusMetricResult(madeUpMetricHeader, timeSeriesValues));
+                }
+                default -> null;
+            };
         }
     }
 
--- a/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusSplitManager.java
+++ b/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusSplitManager.java
@@ -17,6 +17,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.inject.Inject;
 import io.airlift.http.client.HttpUriBuilder;
 import io.airlift.units.Duration;
+import io.trino.plugin.prometheus.expression.LabelFilterExpression;
 import io.trino.spi.TrinoException;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ConnectorSession;
@@ -45,6 +46,7 @@ import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.trino.plugin.prometheus.PrometheusClient.TIMESTAMP_COLUMN_TYPE;
 import static io.trino.plugin.prometheus.PrometheusErrorCode.PROMETHEUS_UNKNOWN_ERROR;
 import static io.trino.plugin.prometheus.PrometheusSessionProperties.getMaxQueryRange;
@@ -89,37 +91,95 @@ public class PrometheusSplitManager
         Duration maxQueryRangeDuration = getMaxQueryRange(session);
         Duration queryChunkSizeDuration = getQueryChunkSize(session);
 
+        // Build a PromQL-compatible label matcher clause from any pushed-down label predicates
+        String labelMatchers = generateLabelMatchers(tableHandle.expressions());
+
         List<ConnectorSplit> splits = generateTimesForSplits(prometheusClock.now(), maxQueryRangeDuration, queryChunkSizeDuration, tableHandle)
                 .stream()
                 .map(time -> {
                     try {
+                        // Pass in the label matchers (e.g., "{instance=\"app1\",job!=\"test\"}")
                         return new PrometheusSplit(buildQuery(
                                 prometheusURI,
                                 time,
                                 table.name(),
-                                queryChunkSizeDuration).toString());
+                                queryChunkSizeDuration,
+                                labelMatchers).toString());
                     }
                     catch (URISyntaxException e) {
                         throw new TrinoException(PROMETHEUS_UNKNOWN_ERROR, "split URI invalid: " + e.getMessage());
                     }
-                }).collect(Collectors.toList());
+                })
+                .collect(toImmutableList());
         return new FixedSplitSource(splits);
     }
 
-    // HttpUriBuilder handles URI encode
-    private static URI buildQuery(URI baseURI, String time, String metricName, Duration queryChunkSizeDuration)
-            throws URISyntaxException
+    /**
+      * Inspect the TupleDomain predicate to extract any label-based constraints
+      * (PrometheusColumnHandle instances that are not the timestamp column),
+      * and convert them into a PromQL "{...}" matcher. If no label filters are present,
+      * returns an empty string.
+      */
+    private static String generateLabelMatchers(List<LabelFilterExpression> expressions)
     {
+        if (expressions.isEmpty()) {
+            return "";
+        }
+
+        // Convert each LabelFilterExpression into a PromQL matcher string
+        List<String> matchers = expressions.stream()
+                .map(expr -> {
+                    String labelName = expr.labelName();
+                    String labelValue = expr.labelValue();
+
+                    // If the value is a regex, use =~, otherwise use =
+                    if (expr.isRegex()) {
+                        if (expr.isNegative()) {
+                            return labelName + "!~\"" + labelValue + "\"";
+                        }
+                        return labelName + "=~\"" + labelValue + "\"";
+                    }
+                    else {
+                        if (expr.isNegative()) {
+                            return labelName + "!=\"" + labelValue + "\"";
+                        }
+                        return labelName + "=\"" + labelValue + "\"";
+                    }
+                })
+                .collect(toImmutableList());
+
+        // Join all individual matchers with commas inside braces
+        return "{" + String.join(",", matchers) + "}";
+    }
+
+    private static URI buildQuery(
+             URI baseURI,
+             String time,
+             String metricName,
+             Duration queryChunkSizeDuration,
+             String labelMatchers)
+             throws URISyntaxException
+    {
+        // Construct PromQL expressions: e.g. metricName{…}[5m]
+        StringBuilder promql = new StringBuilder(metricName);
+        if (!labelMatchers.isEmpty()) {
+            promql.append(labelMatchers);
+        }
+        promql.append("[")
+                .append(queryChunkSizeDuration.roundTo(queryChunkSizeDuration.getUnit()))
+                .append(Duration.timeUnitToString(queryChunkSizeDuration.getUnit()))
+                .append("]");
+
         return HttpUriBuilder.uriBuilderFrom(baseURI)
                 .appendPath("api/v1/query")
-                .addParameter("query", metricName + "[" + queryChunkSizeDuration.roundTo(queryChunkSizeDuration.getUnit()) + Duration.timeUnitToString(queryChunkSizeDuration.getUnit()) + "]")
+                .addParameter("query", promql.toString())
                 .addParameter("time", time)
                 .build();
     }
 
     /**
      * Utility method to get the end times in decimal seconds that divide up the query into chunks
-     * The times will be used in queries to Prometheus like: {@code http://localhost:9090/api/v1/query?query=up[21d]&time=1568229904.000"}
+     * The times will be used in queries to Prometheus like: {@code http://localhost:9090/api/v1/query?query=up[21d]&time=1568229904.000}
      *
      * <p>
      * NOTE: Prometheus instant query wants the duration and end time specified.
@@ -131,8 +191,7 @@ public class PrometheusSplitManager
     protected static List<String> generateTimesForSplits(Instant defaultUpperBound, Duration maxQueryRangeDurationRequested, Duration queryChunkSizeDurationRequested,
             PrometheusTableHandle tableHandle)
     {
-        Optional<PrometheusPredicateTimeInfo> predicateRange = tableHandle.predicate()
-                .flatMap(PrometheusSplitManager::determinePredicateTimes);
+        Optional<PrometheusPredicateTimeInfo> predicateRange = determinePredicateTimes(tableHandle.predicate());
 
         EffectiveLimits effectiveLimits = new EffectiveLimits(defaultUpperBound, maxQueryRangeDurationRequested, predicateRange);
         Instant upperBound = effectiveLimits.getUpperBound();
--- a/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusTableHandle.java
+++ b/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusTableHandle.java
@@ -13,17 +13,20 @@
  */
 package io.trino.plugin.prometheus;
 
+import com.google.common.collect.ImmutableList;
+import io.trino.plugin.prometheus.expression.LabelFilterExpression;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ConnectorTableHandle;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.predicate.TupleDomain;
 
+import java.util.HashSet;
+import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 
 import static java.util.Objects.requireNonNull;
 
-public record PrometheusTableHandle(String schemaName, String tableName, Optional<TupleDomain<ColumnHandle>> predicate)
+public record PrometheusTableHandle(String schemaName, String tableName, TupleDomain<ColumnHandle> predicate, List<LabelFilterExpression> expressions)
         implements ConnectorTableHandle
 {
     public PrometheusTableHandle
@@ -31,6 +34,9 @@ public record PrometheusTableHandle(String schemaName, String tableName, Optiona
         requireNonNull(schemaName, "schemaName is null");
         requireNonNull(tableName, "tableName is null");
         requireNonNull(predicate, "predicate is null");
+
+        // Throws NullPointerException if collection contains null element
+        expressions = ImmutableList.copyOf(expressions);
     }
 
     public SchemaTableName toSchemaTableName()
@@ -40,13 +46,18 @@ public record PrometheusTableHandle(String schemaName, String tableName, Optiona
 
     public PrometheusTableHandle withPredicate(TupleDomain<ColumnHandle> predicate)
     {
-        return new PrometheusTableHandle(schemaName, tableName, Optional.of(predicate));
+        return new PrometheusTableHandle(schemaName, tableName, predicate, expressions);
+    }
+
+    public PrometheusTableHandle withExpressions(List<LabelFilterExpression> expressions)
+    {
+        return new PrometheusTableHandle(schemaName, tableName, predicate, expressions);
     }
 
     @Override
     public int hashCode()
     {
-        return Objects.hash(schemaName, tableName);
+        return Objects.hash(schemaName, tableName, expressions);
     }
 
     @Override
@@ -61,7 +72,9 @@ public record PrometheusTableHandle(String schemaName, String tableName, Optiona
 
         PrometheusTableHandle other = (PrometheusTableHandle) obj;
         return Objects.equals(this.schemaName, other.schemaName) &&
-                Objects.equals(this.tableName, other.tableName);
+                Objects.equals(this.tableName, other.tableName) &&
+                this.expressions.size() == other.expressions.size() &&
+                new HashSet<>(this.expressions).containsAll(other.expressions);
     }
 
     @Override
--- a/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/LabelFilterExpression.java
+++ b/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/LabelFilterExpression.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.prometheus.expression;
+
+import io.trino.plugin.base.expression.ConnectorExpressionRule;
+
+import java.util.List;
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+public record LabelFilterExpression(String labelName, String labelValue, boolean isRegex, boolean isNegative)
+{
+    public static final Set<ConnectorExpressionRule<?, List<LabelFilterExpression>>> APPLICABLE_EXPRESSION_RULES = Set.of(
+            new RewriteEqAndNeq(),
+            new RewriteIn(),
+            new RewriteAnd(),
+            new RewriteLikeWithCaseSensitivity(),
+            new RewriteLikeEscapeWithCaseSensitivity());
+
+    public LabelFilterExpression
+    {
+        requireNonNull(labelName, "labelName is null");
+        requireNonNull(labelValue, "labelValue is null");
+    }
+}
--- a/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteAnd.java
+++ b/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteAnd.java
@@ -1,0 +1,63 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.prometheus.expression;
+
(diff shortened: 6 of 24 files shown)


plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteIn.java
```diff
@@ -0,0 +1,84 @@
...
+    @Override
+    public Optional<ImmutableList<LabelFilterExpression>> rewrite(Call call, Captures captures, RewriteContext<ImmutableList<LabelFilterExpression>> context)
+    {
+        String labelName = extractLabelName(captures, LABEL_NAME);
+        List<ConnectorExpression> expressions = captures.get(LABEL_VALUES);
+
+        List<String> terms = new ArrayList<>(expressions.size());
+        for (ConnectorExpression expression : expressions) {
+            terms.add(extractVarchar(expression));
+        }
```
etolbakov: nit: this section could be rewritten with Java Streams:
```
List<String> terms = expressions.stream()
    .map(this::extractVarchar)
    .collect(Collectors.toList());
 ```
 
 or even (if you consider inlining `expressions`
 ```
 List<String> terms = captures.get(LABEL_VALUES).stream()
    .map(this::extractVarchar)
    .collect(Collectors.toList());
```

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/LoggedRequest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public String extractLimit()
+    {
+        return extractGetParameter("limit");
+    }
+
+    @Override
+    public String toString()
+    {
+        return "[" + method + "] " + uri + " -> " + status;
```
etolbakov: nit:
```suggestion
        return String.format("[%s] %s -> %s", method, uri, status);
```

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteIn.java
jirislav: Great idea. Actually I think we can push it even further and even avoid creating the list:

```java
        String labelName = extractLabelName(captures, LABEL_NAME);
        List<ConnectorExpression> expressions = captures.get(LABEL_VALUES);

        return Optional.of(ImmutableList.of(new LabelFilterExpression(
                labelName,
                expressions.stream()
                        .map(RewriteHelpers::extractVarchar)
                        .collect(Collectors.joining("|", "(", ")")),
                true,
                false)));
```

jirislav:
> Thank you for your pull request and welcome to the Trino community. We require contributors to sign our [Contributor License Agreement](https://github.com/trinodb/cla/raw/master/Trino%20Foundation%20Individual%20CLA.pdf), and we don't seem to have you on file. Continue to work with us on the review and improvements in this PR, and submit the signed CLA to [cla@trino.io](mailto:cla@trino.io). Photos, scans, or digitally-signed PDF files are all suitable. Processing may take a few days. The CLA needs to be on file before we merge your changes. For more information, see https://github.com/trinodb/cla

I have the CLA signed and sent like 6 hours ago, not sure how long it usually takes to propagate to the related failing CI job.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteIn.java
etolbakov: @jirislav  It's great that you found it useful!
Btw, I'm not trino contributor/maintainer (as yourself sent a signed agreement but haven't got reply back), I'm just following the project and find your PR interesting to read!

jirislav:
@ebyhr could you please review & merge? I see you [made recent changes to Prometheus Connector](https://github.com/trinodb/trino/blame/6028e809b8b3993665191766af504e0561607674/plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusMetadata.java) as well.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteEqAndNeq.java
```diff
@@ -79,6 +79,6 @@ public Optional<ImmutableList<LabelFilterExpression>> rewrite(Call call, Capture
                 labelName,
                 labelValue,
                 false,
-                call.getFunctionName() == NOT_EQUAL_OPERATOR_FUNCTION_NAME)));
+                call.getFunctionName().equals(NOT_EQUAL_OPERATOR_FUNCTION_NAME))));
```
ebyhr: This change should go to the 1st commit.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -53,13 +53,13 @@ public static String extractLabelName(Captures captures, Capture<Call> captureKe
         // This method assumes that
         List<ConnectorExpression> arguments = captures.get(captureKey).getArguments();
         Object labelNameObject = ((Constant) arguments.getLast()).getValue();
-        if (!(labelNameObject instanceof Slice)) {
+        if (!(labelNameObject instanceof Slice slice)) {
```
ebyhr: This change should go to the 1st commit.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -53,13 +53,13 @@ public static String extractLabelName(Captures captures, Capture<Call> captureKe
...
             if (labelNameObject == null) {
                 throw new IllegalArgumentException("Label name cannot be null");
             }
             throw new IllegalArgumentException("Expected label name to be a Slice, but got: " + labelNameObject.getClass().getSimpleName());
         }
-        return ((Slice) labelNameObject).toStringUtf8();
+        return slice.toStringUtf8();
```
ebyhr: This change should go to the 1st commit.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusTableHandle.java
```diff
@@ -23,14 +25,15 @@
 
 import static java.util.Objects.requireNonNull;
 
-public record PrometheusTableHandle(String schemaName, String tableName, Optional<TupleDomain<ColumnHandle>> predicate)
+public record PrometheusTableHandle(String schemaName, String tableName, Optional<TupleDomain<ColumnHandle>> predicate, ImmutableList<LabelFilterExpression> expressions)
```
ebyhr:
```suggestion
public record PrometheusTableHandle(String schemaName, String tableName, Optional<TupleDomain<ColumnHandle>> predicate, List<LabelFilterExpression> expressions)
```

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusTableHandle.java
```diff
@@ -23,14 +25,15 @@
...
         implements ConnectorTableHandle
 {
     public PrometheusTableHandle
     {
         requireNonNull(schemaName, "schemaName is null");
         requireNonNull(tableName, "tableName is null");
         requireNonNull(predicate, "predicate is null");
+        requireNonNull(expressions, "expressions is null");
```
ebyhr: Remove requireNonNull and add ImmutableList.copyOf.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/LabelFilterExpression.java
```diff
@@ -0,0 +1,37 @@
...
+import com.google.common.collect.ImmutableList;
+import io.trino.plugin.base.expression.ConnectorExpressionRule;
+
+import java.util.Set;
+
+import static java.util.Objects.requireNonNull;
+
+public record LabelFilterExpression(String labelName, String labelValue, boolean isRegex, boolean isNegative)
+{
+    public static final Set<ConnectorExpressionRule<?, ImmutableList<LabelFilterExpression>>> APPLICABLE_EXPRESSION_RULES = Set.of(
```
ebyhr: ImmutableList → List

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteAnd.java
```diff
@@ -0,0 +1,70 @@
...
+import static com.google.common.base.Verify.verify;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static io.trino.plugin.base.expression.ConnectorExpressionPatterns.call;
+import static io.trino.plugin.base.expression.ConnectorExpressionPatterns.functionName;
+import static io.trino.plugin.base.expression.ConnectorExpressionPatterns.type;
+import static io.trino.spi.expression.StandardFunctions.AND_FUNCTION_NAME;
+import static io.trino.spi.type.BooleanType.BOOLEAN;
+
+class RewriteAnd
+        implements ConnectorExpressionRule<Call, ImmutableList<LabelFilterExpression>>
```
ebyhr: ImmutableList → List. Same for other rewrite classes.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteAnd.java
```diff
@@ -0,0 +1,70 @@
...
+{
+    private final Pattern<Call> pattern;
+
+    public RewriteAnd()
+    {
+        this.pattern = call()
+                .with(functionName().equalTo(AND_FUNCTION_NAME))
+                .with(type().equalTo(BOOLEAN));
+    }
```
ebyhr: The `pattern` can be a constant. Same for other rewrite classes.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteAnd.java
```diff
@@ -0,0 +1,70 @@
...
+    {
+        return pattern;
+    }
+
+    @Override
+    public Optional<ImmutableList<LabelFilterExpression>> rewrite(Call call, Captures captures, RewriteContext<ImmutableList<LabelFilterExpression>> context)
+    {
+        List<ConnectorExpression> arguments = call.getArguments();
+        verify(!arguments.isEmpty(), "no arguments");
+        List<LabelFilterExpression> terms = new ArrayList<>(arguments.size());
```
ebyhr:
```java
ImmutableList.Builder<LabelFilterExpression> terms = ImmutableList.builderWithExpectedSize(arguments.size());
```

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteEqAndNeq.java
```diff
@@ -0,0 +1,84 @@
...
+{
+    private static final Capture<Call> LABEL_NAME = newCapture();
+    private static final Capture<Constant> LABEL_VALUE = newCapture();
+
+    private final Pattern<Call> pattern;
+
+    public RewriteEqAndNeq()
+    {
+        pattern = call()
+                .with(functionName().matching(fn -> ImmutableSet.of(
```
ebyhr: Please avoid abbreviations. https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -0,0 +1,78 @@
...
+import io.trino.spi.type.CharType;
+import io.trino.spi.type.Type;
+import io.trino.spi.type.VarcharType;
+
+import java.util.List;
+import java.util.function.Predicate;
+
+import static com.google.common.base.Verify.verify;
+
+public class RewriteHelpers
```
ebyhr: Add `final`

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/LoggedRequest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    public String extractStep()
+    {
+        return extractGetParameter("step");
+    }
+
+    public String extractLimit()
+    {
+        return extractGetParameter("limit");
+    }
```
ebyhr: Remove unused methods.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/LoggedRequest.java
```diff
@@ -0,0 +1,82 @@
...
+
+    @Override
+    public String toString()
+    {
+        return String.format("[%s] %s -> %s", method, uri, status);
```
ebyhr: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#prefer-string-formatting

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
```diff
@@ -64,13 +68,143 @@ public void testAggregation()
...
-    public void testPushDown()
+    public void testTimestampPushDown()
     {
         // default interval on the `up` metric that Prometheus records on itself is about 15 seconds, so this should only yield one or two row
         MaterializedResult results = computeActual("SELECT * FROM prometheus.default.up WHERE timestamp > (NOW() - INTERVAL '15' SECOND)");
         assertThat(results).hasSizeBetween(1, 2);
     }
 
+    @Test
+    public void testLabelEqPushDown()
```
ebyhr: Can we use `QueryAssert#isFullyPushedDown` to  ensure pushdown?

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
```diff
@@ -64,13 +68,143 @@ public void testAggregation()
...
+        MaterializedResult results = server.recordRequestsFor(recorder, () ->
+                getQueryRunner().execute("SELECT * FROM prometheus.default.up WHERE labels['job'] = 'prometheus'"));
+
+        assertThat(recorder.getRequestsWithoutDiscovery()).hasSizeGreaterThan(0);
+        recorder.getRequestsWithoutDiscovery().forEach(request -> {
+            assertThat(request.method).isEqualTo("GET");
+            assertThat(request.extractPromQL()).isEqualTo("up{job=\"prometheus\"}[1d]");
+        });
+
+        assertThat(results).hasSizeBetween(1, 2);
```
ebyhr: Why is the result size non-deterministic? Please leave a code comment.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
```diff
@@ -64,13 +68,143 @@ public void testAggregation()
...
+        MaterializedResult results = server.recordRequestsFor(recorder, () ->
+                getQueryRunner().execute("SELECT * FROM prometheus.default.up WHERE NOT (labels['job'] LIKE 'prom%' AND labels['instance'] = 'localhost:9090' AND labels['__name__'] != 'down')"));
+
+        assertThat(recorder.getRequestsWithoutDiscovery()).hasSizeGreaterThan(0);
+        recorder.getRequestsWithoutDiscovery().forEach(request -> {
+            assertThat(request.method).isEqualTo("GET");
+            assertThat(request.extractPromQL()).isEqualTo("up[1d]");
+        });
+
+        assertThat(results).hasSize(0);
```
ebyhr: There is `isEmpty()` method.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusTableHandle.java
```diff
@@ -47,6 +51,11 @@ public void testEquivalence()
 
     public static PrometheusTableHandle newTableHandle(String schemaName, String tableName)
     {
-        return new PrometheusTableHandle(schemaName, tableName, Optional.empty());
+        return newTableHandle(schemaName, tableName, Optional.empty(), ImmutableList.of());
+    }
+
+    public static PrometheusTableHandle newTableHandle(String schemaName, String tableName, Optional<TupleDomain<ColumnHandle>> predicate, ImmutableList<LabelFilterExpression> expressions)
```
ebyhr:
```suggestion
    public static PrometheusTableHandle newTableHandle(String schemaName, String tableName, Optional<TupleDomain<ColumnHandle>> predicate, List<LabelFilterExpression> expressions)
```

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
```diff
@@ -31,8 +33,10 @@
 import static java.util.concurrent.TimeUnit.DAYS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.parallel.ExecutionMode.SAME_THREAD;
 
-public class TestPrometheusIntegration
+@Execution(SAME_THREAD)
```
ebyhr: Please leave a code comment why specifying `SAME_THREAD`.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/PrometheusServer.java
```diff
@@ -14,57 +14,129 @@
...
     public void close()
     {
-        dockerContainer.close();
+        prometheusContainer.close();
+
+        if (proxyContainer != null) {
+            proxyContainer.close();
+        }
+
+        network.close();
```
ebyhr: You can use Guava's `Closer`.

ebyhr (commented):
Just skimmed.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/PrometheusServer.java
```diff
@@ -14,57 +14,129 @@
...
 import static org.testcontainers.utility.MountableFile.forClasspathResource;
 
 public class PrometheusServer
         implements Closeable
 {
+    private static final int PROXY_PORT = 8080;
+    private static final String DEFAULT_NGINX_PROXY_VERSION = "1.28.0-alpine";
+
     private static final int PROMETHEUS_PORT = 9090;
-    private static final String DEFAULT_VERSION = "v2.15.1";
```
chenjian2664: Does this changes required?

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -0,0 +1,78 @@
...
+{
+    private RewriteHelpers() {}
+
+    public static final Predicate<Call> LABEL_MAP_CHECK_CORRECT_TYPE = connectorExpression -> {
+        List<ConnectorExpression> arguments = connectorExpression.getArguments();
+
+        if (!(arguments.getFirst() instanceof Variable)) {
```
chenjian2664: Add a check if arguments is empty or not

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -0,0 +1,78 @@
...
+    };
+
+    public static String extractLabelName(Captures captures, Capture<Call> captureKey)
+    {
+        // Extract the label name from the call expression, which is actually the key of the labels map.
+        // This method assumes that
+        List<ConnectorExpression> arguments = captures.get(captureKey).getArguments();
+        Object labelNameObject = ((Constant) arguments.getLast()).getValue();
+        if (!(labelNameObject instanceof Slice)) {
+            if (labelNameObject == null) {
```
chenjian2664: Should we perform this check earlier?

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return isCharOrStringType(connectorExpression.getType());
+    };
+
+    public static String extractLabelName(Captures captures, Capture<Call> captureKey)
+    {
+        // Extract the label name from the call expression, which is actually the key of the labels map.
+        // This method assumes that
+        List<ConnectorExpression> arguments = captures.get(captureKey).getArguments();
+        Object labelNameObject = ((Constant) arguments.getLast()).getValue();
```
chenjian2664: add arguments check if it is empty

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
jirislav: Maybe we could, but it wouldn't give us the option to inspect the resulting PromQL query. Having this enables us to validate that the push-down resulted in expected query to Prometheus. We don't have that with `QueryAssert#isFullyPushedDown`. Additionally, it's easier to debug as we can see exactly what does the PromQL look like in case it fails.

And lastly, using `AssertProvider` is not compatible with recording the actual queries, so the queries would have to run twice with nearly duplicate code in order to test both.

If we are to keep my proposed solution, I think adding `QueryAssert#isFullyPushedDown` is redundant.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/PrometheusServer.java
jirislav: Guava's Closer [currently doesn't support `AutoCloseable` instances](https://github.com/google/guava/issues/3068), only `Closable` instances. All of these being closed here are `AutoCloseable`.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/PrometheusServer.java
jirislav: Your question is ambiguous, what change exactly?

If you mean the change of scope from private to public, then yes, we're using this constant in the `TestPrometheusIntegration` class.

If you mean the change of name from `DEFAULT_VERSION` to `DEFAULT_PROMETHEUS_VERSION`, then it's required to avoid ambiguity of what container exactly belongs this version to (prometheus vs. nginx).

jirislav: You mean `private -> public` or `DEFAULT_VERSION -> DEFAULT_PROMETHEUS_VERSION`?
- The former is needed for this constant to be used in `TestPrometheusIntegration`.
- The latter is needed to avoid ambiguity since we're now using two different docker images, each having different version.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
jirislav: Prometheus reports the `up` metric every 15 seconds, so we expect at least one row, but at most two rows if we filter for the last 15 seconds (edge case). Thanks for asking, I have added a code comment.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/PrometheusServer.java
ebyhr: Try the below code:
```java
private final Closer closer = Closer.create();
...
closer.register(network::close);
```

jirislav: That works, thanks!

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/RequestsRecorder.java
```diff
@@ -0,0 +1,54 @@
...
+import org.testcontainers.containers.output.OutputFrame;
+
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+import java.util.function.Consumer;
+
+public class RequestsRecorder
+        implements Consumer<OutputFrame>
+{
+    private final ObjectMapper objectMapper = new ObjectMapper();
```
chenjian2664: Can we check if it is possible to use `new ObjectMapperProvider().get()` ?

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/RequestsRecorder.java
```diff
@@ -0,0 +1,54 @@
...
+
+    @Override
+    public void accept(OutputFrame outputFrame)
+    {
+        String text = outputFrame.getUtf8String().trim();
+        try {
+            LoggedRequest req = objectMapper.readValue(text, LoggedRequest.class);
+            requests.add(req);
+        }
+        catch (Exception e) {
```
chenjian2664: should catch `IOException`

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusTableHandle.java
```diff
@@ -13,24 +13,30 @@
...
         implements ConnectorTableHandle
 {
     public PrometheusTableHandle
     {
         requireNonNull(schemaName, "schemaName is null");
         requireNonNull(tableName, "tableName is null");
         requireNonNull(predicate, "predicate is null");
+
+        // Throws NullPointerException if collection contains null element
+        var unused = ImmutableList.copyOf(expressions);
```
chenjian2664:
```suggestion
        expressions = ImmutableList.copyOf(expressions);
```

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusSplitManager.java
```diff
@@ -89,37 +90,95 @@ public ConnectorSplitSource getSplits(
...
+                        return labelName + "=~\"" + labelValue + "\"";
+                    }
+                    else {
+                        if (expr.isNegative()) {
+                            return labelName + "!=\"" + labelValue + "\"";
+                        }
+                        return labelName + "=\"" + labelValue + "\"";
+                    }
+                })
+                .collect(Collectors.toList());
```
chenjian2664: use `toImmutableList()`

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusSplitManager.java
```diff
@@ -89,37 +90,95 @@ public ConnectorSplitSource getSplits(
...
-                                queryChunkSizeDuration).toString());
+                                queryChunkSizeDuration,
+                                labelMatchers).toString());
                     }
                     catch (URISyntaxException e) {
                         throw new TrinoException(PROMETHEUS_UNKNOWN_ERROR, "split URI invalid: " + e.getMessage());
                     }
-                }).collect(Collectors.toList());
+                })
+                .collect(Collectors.toList());
```
chenjian2664: use `toImmutableList()`

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusMetadata.java
```diff
@@ -173,8 +176,13 @@ public ColumnMetadata getColumnMetadata(ConnectorSession session, ConnectorTable
...
     {
+        ConnectorExpression expression = constraint.getExpression();
+
+        ConnectorExpressionRewriter<List<LabelFilterExpression>> rewriter = new ConnectorExpressionRewriter<>(LabelFilterExpression.APPLICABLE_EXPRESSION_RULES);
+
         PrometheusTableHandle tableHandle = ((PrometheusTableHandle) handle)
-                .withPredicate(constraint.getSummary());
+                .withPredicate(constraint.getSummary())
+                .withExpressions(rewriter.rewrite(session, expression, constraint.getAssignments()).orElseGet(ImmutableList::of));
         return Optional.of(new ConstraintApplicationResult<>(tableHandle, constraint.getSummary(), constraint.getExpression(), false));
```
chenjian2664: After updating `tableHandle` , it looks like `remainingFilter` and `remainingExpression` still reference the previous state. Is that intentional?

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -0,0 +1,85 @@
...
+        if (!variableName.equals("labels")) {
+            return false;
+        }
+        return isCharOrStringType(connectorExpression.getType());
+    };
+
+    public static String extractLabelName(Captures captures, Capture<Call> captureKey)
+    {
+        // Extract the label name from the call expression, which is actually the key of the labels map.
+        // This method assumes that
```
chenjian2664: Fix the comment ?

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteIn.java
```diff
@@ -0,0 +1,79 @@
...
+    public Optional<List<LabelFilterExpression>> rewrite(Call call, Captures captures, RewriteContext<List<LabelFilterExpression>> context)
+    {
+        String labelName = extractLabelName(captures, LABEL_NAME);
+        List<ConnectorExpression> expressions = captures.get(LABEL_VALUES);
+
+        return Optional.of(ImmutableList.of(new LabelFilterExpression(
+                labelName,
+                expressions.stream()
+                        .map(RewriteHelpers::extractVarchar)
+                        .collect(Collectors.joining("|", "(", ")")),
```
chenjian2664: nit, we can remove the duplicates before building the value

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/RequestsRecorder.java
jirislav: Yes we can, I've changed it. Thanks.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
jirislav: Thanks, there's no assumption anymore since we're checking arguments' list emptiness & their instance types. I have removed this line.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusMetadata.java
jirislav: Yes, this is intentional because we can't push-down arbitrary operation for the labels map (see the `testLabelLikeAndEqAndNeqNegatedPushDown` test to see why - though it's just a single edge-case example).

Yes, Trino will potentially filter something that was already pushed down, but the real value of this PR is that Prometheus doesn't have to return all the timeseries anymore and Trino doesn't have to filter them all. From now, Prometheus will always perform very efficient timeseries filtering based on provided label selectors and Trino will go just through a fraction of data compared to previous version.

In the future, someone might add support for the OR operator, or NOT(x AND y) by spitting the query into multiple Prometheus requests, potentially covering all the logical edge-cases for label filter pushdown. At that point, I think we could safely remove all the filters on label values in the above discussed lines, but I'd argue it's out-of-scope for this PR, as this itself is a major improvement already.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusQueryResponseParse.java
```diff
@@ -53,17 +53,15 @@ public PrometheusQueryResponseParse(InputStream response)
...
-                                    parser.nextToken();
-                                    resultType = parser.getValueAsString();
-                                }
-                                if (parser.currentName().equals("result")) {
-                                    parser.nextToken();
-                                    ArrayNode node = mapper.readTree(parser);
-                                    result = node.toString();
-                                    break;
-                                }
+                            if (parser.currentName().equals("resultType")) {
```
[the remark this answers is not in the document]
jirislav: I removed the condition because it is guaranteed to always evaluate to `true` due to already having this exact condition run on line 49 above.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/PrometheusTableHandle.java
```diff
@@ -40,7 +45,12 @@ public SchemaTableName toSchemaTableName()
...
-        return new PrometheusTableHandle(schemaName, tableName, Optional.of(predicate));
+        return new PrometheusTableHandle(schemaName, tableName, predicate, expressions);
+    }
+
+    public PrometheusTableHandle withExpressions(List<LabelFilterExpression> expressions)
+    {
+        return new PrometheusTableHandle(schemaName, tableName, predicate, expressions);
     }
 
     @Override
```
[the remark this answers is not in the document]
jirislav: That's a good point, thanks, I have added `expressions` to both methods.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteLikeEscapeWithCaseSensitivity.java
```diff
@@ -0,0 +1,89 @@
...
+                labelName,
+                // Only replace LIKE matching characters ('%' and '_') in case they're not escaped.
+                // I didn't manage to get negative lookaheads to work, so we use this dummy workaround.
+                labelValue
+                        .replaceAll(Matcher.quoteReplacement(escapePattern) + "%", "<LiteralLikePercentage>")
+                        .replaceAll(Matcher.quoteReplacement(escapePattern) + "_", "<LiteralLikeUnderscore>")
+                        .replaceAll("%", ".*")
+                        .replaceAll("_", ".")
+                        .replaceAll("<LiteralLikePercentage>", "%")
+                        .replaceAll("<LiteralLikeUnderscore>", "_"),
```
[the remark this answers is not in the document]
jirislav: I consider both `<LiteralLikePercentage>` and `<LiteralLikeUnderscore>` string values as unique enough already. For them to conflict with actual label values, someone would have to literally include them into the predicate literally.

For example, this edge case could misbehave:

```sql
WHERE a LIKE '<LiteralLikePercentage>' --- Input

WHERE a LIKE '%' --- Output
```

This is just to prove that it's wrong to assume it has to do anything with label values, it only acts on the actual literal string expressions provided in raw SQL.

I also don't get what do you mean by `" and "`, there is no such string matching here.

plugin/trino-prometheus/src/main/java/io/trino/plugin/prometheus/expression/RewriteHelpers.java
```diff
@@ -0,0 +1,84 @@
...
+        return type instanceof CharType || type instanceof VarcharType;
+    }
+
+    @SuppressWarnings("DataFlowIssue")
+    public static String extractVarchar(ConnectorExpression expression)
+    {
+        verify(isCharOrStringType(expression.getType()), "Unexpected type of argument: '%s' (expected VARCHAR)", expression.getType());
+
+        Object value = ((Constant) expression).getValue();
+        return ((Slice) value).toStringUtf8();
```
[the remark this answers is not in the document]
jirislav: I don't think it's needed as it is already scoped just to `DataFlowIssue`, but I agree it's not generally good to use `@SupressWarnings`, so I have applied your suggestion, thank you.

plugin/trino-prometheus/src/test/java/io/trino/plugin/prometheus/TestPrometheusIntegration.java
chenjian2664: @jirislav the pushdown needs to check two parts: the plan and the result - that is Trino plan should not contains `ScanFilterProject` if the predicate is pushdowned, ref https://trino.io/docs/current/optimizer/pushdown.html#predicate-pushdown

Another important aspect is the data. If the predicate is pushed down, the result should be exactly the same between the remote system and Trino. The `QueryAssert#isFullyPushedDown` provides a mechanism to verify this

ramcasa:
This is a quite useful feature. Without it querying a decently-sized database is close to impossible. Is there an ETA for merging this?

PR #25958, opened 2025-06-09
Add query-max-write-physical-bytes session property

vvvvh123:
## Description

This pull request adds a new session property: `query_max_write_physical_bytes`, similar to the existing `query_max_scan_physical_bytes`. It allows users to set an upper limit on the total physical bytes written by a query.

Currently my pull request only allows the user to use this as a session property. If needed, I can extend it to support a configuration property as well.

I need to modify the Trino documentation to add details about this new property, but it seems that all the properties [here](https://trino.io/docs/current/admin/properties-query-management.html#query-max-scan-physical-bytes) are config properties with some having corresponding session properties.

This feature was implemented in response to an internal requirement at my company to control resource usage for write-heavy queries. I'm hoping to contribute it to the open source code. 

## Release notes

```markdown
## General
* Add query-max-write-physical-bytes session property ({issue}`25955`)
```


core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
```diff
@@ -403,6 +404,11 @@ public SystemSessionProperties(
                         "Maximum scan physical bytes of a query",
                         queryManagerConfig.getQueryMaxScanPhysicalBytes().orElse(null),
                         false),
+                dataSizeProperty(
```
pajaks, line 407: There should be also property in `memoryManagerConfig` similar to `QUERY_MAX_SCAN_PHYSICAL_BYTES `

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
```diff
@@ -371,4 +382,23 @@ private void enforceScanLimits()
...
+    /**
+     * Enforce query write physical bytes limits
+     */
+    private void enforceWriteLimits()
+    {
+        for (QueryExecution query : queryTracker.getAllQueries()) {
+            if (query.isDone() || query.getState() == FINISHING) {
+                continue;
+            }
+            DataSize writeLimit = getQueryMaxWritePhysicalBytes(query.getSession());
```
pajaks: Use Optional instead of null handling

core/trino-main/src/main/java/io/trino/execution/StateMachine.java
```diff
@@ -108,7 +108,7 @@ public T set(T newState)
 
     /**
      * Tries to change the state.  State will not change if the new state {@code .equals()} the current state,
-     * of if the current state is a terminal state. If the state changed, listeners and waiters will be notified.
+     * or if the current state is a terminal state. If the state changed, listeners and waiters will be notified.
```
marcinsbd, line 111: move this change to separate commit

core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
vvvvh123: Done!

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
vvvvh123: Done

core/trino-main/src/main/java/io/trino/execution/StateMachine.java
vvvvh123: Done!

docs/src/main/sphinx/admin/properties-query-management.md
```diff
@@ -159,6 +159,15 @@ The maximum number of bytes that can be scanned by a query during its execution.
 When this limit is reached, query processing is terminated to prevent excessive
 resource usage.
 
+## `query.max-write-physical-bytes`
```
pajaks: Change naming to query.max-written-data-size (query_max_written_data_size)

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
```diff
@@ -91,6 +95,7 @@ public SqlQueryManager(ClusterMemoryManager memoryManager, Tracer tracer, QueryM
 
         this.maxQueryCpuTime = queryManagerConfig.getQueryMaxCpuTime();
         this.maxQueryScanPhysicalBytes = queryManagerConfig.getQueryMaxScanPhysicalBytes();
+        this.maxQueryWritePhysicalBytes = queryManagerConfig.getQueryMaxWritePhysicalBytes();
```
pajaks: You don't need this as `queryManagerConfig.getQueryMaxWritePhysicalBytes()` is already included in `SystemSessionProperties`

docs/src/main/sphinx/admin/properties-query-management.md
vvvvh123: Sure I can make this change. I just wanted to check if there’s a reason we don’t prefer `query.max-write-physical-bytes` for consistency with the existing `query.max-scan-physical-bytes`.

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
vvvvh123: For `query-max-scan-physical-bytes`, the limit enforced is min(configLimit, sessionLimit). For the write limit, do we want the sessionLimit to override the configLimit, or should we apply the same logic?

pajaks: Yes session property should always override corresponding config property

docs/src/main/sphinx/admin/properties-query-management.md
pajaks: It's closer to statistic name `getPhysicalWrittenDataSize` used for checking limit.

martint: I'll throw in another alternative. It should be `query.max-write-physical-size`. "data-size" is redundant, and doesn't convey the fact that this is about "physical" data written. "bytes" is inappropriate, since the option can take values in units other than bytes. The `query.max-scan-physical-bytes` should eventually be renamed to `query.max-scan-physical-size`.

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
vvvvh123: Hi, done this. Should I also change the logic for query.max-scan-physical-bytes in a separate commit?

raunaqmorarka: Actually for server resource limits, users should not have the ability to set the session property higher than the config property set by the admin.
This should follow the same logic as query-max-scan-physical-bytes

docs/src/main/sphinx/admin/properties-query-management.md
vvvvh123: I've changed the naming to `query.max-write-physical-size`. I can also modify the naming for `query.max-scan-physical-bytes` to `query.max-scan-physical-size` but I guess that would be a breaking change (unless we wish to support both naming formats).

raunaqmorarka: That change should go in a separate PR
You can look at other examples of using `@LegacyConfig` annotation to change existing config names

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
vvvvh123: Reverted to follow the same logic as query-max-scan-physical-bytes

raunaqmorarka (approved):

testing/trino-tests/src/test/java/io/trino/tests/TestQueryManager.java
```diff
@@ -132,4 +143,71 @@ public void testQueryScanExceededSession()
...
+                    "hive.metastore", "file",
+                    "hive.metastore.catalog.dir", hiveDataDir.toFile().getAbsolutePath(),
+                    "fs.hadoop.enabled", "true"));
+
+            Session session = testSessionBuilder()
+                    .setCatalog("hive")
+                    .setSchema("test")
+                    .setClientCapabilities(stream(ClientCapabilities.values())
+                            .map(ClientCapabilities::toString)
+                            .collect(toImmutableSet()))
```
pajaks: Why is it needed?

vvvvh123: thank you, it's not needed

pajaks (approved):

core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
```diff
@@ -506,6 +507,19 @@ public QueryManagerConfig setQueryMaxScanPhysicalBytes(DataSize queryMaxScanPhys
         return this;
     }
 
+    @NotNull
+    public Optional<DataSize> getQueryMaxWritePhysicalSize()
+    {
+        return queryMaxWritePhysicalSize;
+    }
+
+    @Config("query.max-write-physical-size")
```
findinpath, line 516: Can we follow-up with a documentation PR in https://github.com/trinodb/trino/blob/master/docs/src/main/sphinx/admin/properties-query-management.md ?

raunaqmorarka: it's already documented correctly here

PR #25961, opened 2025-06-09
JDBC fetch size

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/client/trino-client/src/main/java/io/trino/client/uri/PropertyName.java
+++ b/client/trino-client/src/main/java/io/trino/client/uri/PropertyName.java
@@ -53,6 +53,7 @@ public enum PropertyName
     KERBEROS_SERVICE_PRINCIPAL_PATTERN("KerberosServicePrincipalPattern"),
     KERBEROS_USE_CANONICAL_HOSTNAME("KerberosUseCanonicalHostname"),
     LOCALE("locale"),
+    FETCH_SIZE("fetchSize"),
     PASSWORD("password"),
     SQL_PATH("path"),
     RESOURCE_ESTIMATES("resourceEstimates"),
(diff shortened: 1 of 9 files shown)


client/trino-client/src/main/java/io/trino/client/uri/PropertyName.java
```diff
@@ -53,6 +53,7 @@ public enum PropertyName
     KERBEROS_SERVICE_PRINCIPAL_PATTERN("KerberosServicePrincipalPattern"),
     KERBEROS_USE_CANONICAL_HOSTNAME("KerberosUseCanonicalHostname"),
     LOCALE("locale"),
+    FETCH_SIZE("fetchSize"),
```
nineinchnick, line 56: This name is confusing. How about `rowPrefetchCount`?

wendigo: It's consistent with the jdbc fetchSize on result set

hqbhoho:
@wendigo Hi, how is the progress of this PR review?

wendigo:
ping @losipiuk

chenjian2664:
What's the purpose to add the fetchSize, since it is not allowed to be updated

hqbhoho:
> What's the purpose to add the fetchSize, since it is not allowed to be updated

@chenjian2664  To reduce high client memory usage caused by long string columns in results, we propose making the rowQueue size in AsyncResultIterator (currently 50,000) a customizable setting.

chenjian2664:
@hqbhoho Does the value able to be updated through this pr approach?

hqbhoho:
@chenjian2664  In this PR, it use fetchSize as maxBufferedRows to new a rowQueue rather than using default `new ArrayBlockingQueue<>(MAX_QUEUED_ROWS)`.  Maybe we can implement `setFetchSize` in `TrinoStatement` not only from uri .

PR #25128, opened 2025-02-24
Remove example HTTP connector from binary artifacts

mosabua:
## Additional context and related issues

Extracted from #25124 as a first test.

## Release notes

(x) Release notes are required, with the following suggested text:

```markdown
tbd
```

PR #25124, opened 2025-02-24
Remove plugins

mosabua:
## Description

Dont review just yet - this is the second part of the linked PR to allow faster merge of that one.

## Additional context and related issues

Part of #22597 

Follow up to #25026 and depends on changes in it.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
tbd
```

Issue #22597, opened 2024-07-05
Improve packaging


mosabua:
The current binary packages provided by the Trino suffer from a number of issues. 

* RPM packaging is barely used and tested, but creates significant work and has a negative impact on the build time.
* Tarball and docker container are large and contain all plugins, typical usage however only facilitates a small subset of the plugins
* The large artifacts result in longer download and installation times, which also affect deployments.
* Occasionally plugins maybe contain dependencies with reported security issues and therefore Trino might be flagged during security analysis at users (rightly or wrongly so).
* Custom package creation is not really supported or documented.

These issues will get worse in the near future with more connectors being merged and also more native binaries for multiple operating systems and processor architectures being required and included. 

This roadmap items collects a number of related work tasks that we want to engage on. Numerous discussions took place prior to filing this issue on slack, in smaller conversations, and at the Trino Contributor Calls and Congregation.

martint:
cc @nineinchnick

bitsondatadev:
@mosabua, 

Just another devex thing we may consider to at least track in an issue, is to build a web app [akin to this Spring Boot one](https://start.spring.io/#!type=gradle-project&language=java&platformVersion=3.3.3&packaging=jar&jvmVersion=17&groupId=com.example&artifactId=demo&name=demo&description=Demo%20project%20for%20Spring%20Boot&packageName=com.example.demo&dependencies=data-jpa,data-jdbc) where you have a nice interface that hits a service to download a tar file with a custom list of plugins.

We could even [make it a rest endpoint on a GitHub action](https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event) that takes in a list of plugins and trino version and returns the tar file with just those plugins. Then we can just have minimal with in-memory (possibly some local filesystem if we want to replicate the DuckDB experience).

mosabua:
@bitsondatadev that is totally also something that's nice to have once we've done the preparatory work in terms of creating a minimal package and other things as proposed in this ticket. I think it's probably within scope here

mosabua:
We cleaned up a few connectors now

#23550 
#23551 
#23588

mosabua:
One thing that came up in @23901 is that some connector and other plugins add global functions to Trino. If we remove such a plugin as part of the packaging clean up .. those functions are no longer available in that packaged version of Trino .. we need to therefore either change implementation of the plugins, ensure they are always there or update docs and explain that they depend on a specific plugin..

[...]

PR #25967, opened 2025-06-10
Remove HTTP server event listener plugin from binaries

mosabua:
## Description

Remove the undocumented and seemingly unused plugin at least from the binaries. 

* Follow up to https://github.com/trinodb/trino/pull/25124
* Part of https://github.com/trinodb/trino/issues/22597
* Related to https://github.com/trinodb/trino/pull/25128

## Release notes

```markdown
## General

* Remove the HTTP server event listener plugin from the tar.gz archive and the Docker container.  ({issue}`25967`)

```

### DIFF
--- a/core/trino-server/src/main/provisio/trino.xml
+++ b/core/trino-server/src/main/provisio/trino.xml
@@ -106,12 +106,6 @@
         </artifact>
     </artifactSet>
 
-    <artifactSet to="plugin/http-server-event-listener">
-        <artifact id="${project.groupId}:trino-http-server-event-listener:zip:${project.version}">
-            <unpack />
-        </artifact>
-    </artifactSet>
-
     <artifactSet to="plugin/hudi">
         <artifact id="${project.groupId}:trino-hudi:zip:${project.version}">
             <unpack />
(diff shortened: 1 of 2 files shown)


wendigo (approved):

mosabua:
Thank you @wendigo 

@raunaqmorarka I left the code since it must be used somewhere .. however it cant be very important since we don't even have docs for it. @losipiuk @wendigo and @lukasz-stec probably have more context but I never got more info about it.

andythsu:
is this the same as https://trino.io/docs/current/admin/event-listeners-http.html? We are using this plugin at the moment. Is it not supported?

wendigo:
@andythsu it is supported, just not included in the tarball

andythsu:
@wendigo I see. Thanks!

mosabua:
Just to be clear .. this is NOT the HTTP event listener .. its the HTTP server event listener plugin that was removed from the binaries by default. 

Compare https://github.com/trinodb/trino/tree/master/plugin/trino-http-event-listener (which is what you are using and are asking about @andythsu ) .. and https://github.com/trinodb/trino/tree/master/plugin/trino-http-server-event-listener (which is not documented apart from the readme that I added in this PR).

mosabua:
I renamed the PR title now .. just in case

PR #25973, opened 2025-06-10
Avoid reading unusually large parquet footers

raunaqmorarka:
## Description

Reading unusually large parquet footers can lead to workers
going into full GC and crashing when decoding the footer in
org.apache.parquet.format.Util#readFileMetaData
This is usually caused by misconfigured parquet writers producing
too many row groups per file. This change adds a guard rail to
fail reads of such files gracefully.

## Release notes

```markdown
## Hive, Delta Lake, Iceberg, Hudi
* Prevent workers from going into full GC or crashing when decoding unusually large parquet footers. ({issue}`25973`)
```

### DIFF
--- a/docs/src/main/sphinx/object-storage/file-formats.md
+++ b/docs/src/main/sphinx/object-storage/file-formats.md
@@ -101,6 +101,12 @@ with Parquet files performed by supported object storage connectors:
     The equivalent catalog session property is
     `parquet_vectorized_decoding_enabled`.
   - `true`
+* - `parquet.max-footer-read-size`
+  - Sets the maximum allowed read size for Parquet file footers. Attempting to
+    read a file with a footer larger than this value will result in an error.
+    This prevents workers from going into full GC or crashing due to poorly
+    configured Parquet writers.
+  - `15MB`
 :::
 
 [](file-compression) is automatically performed and some details can be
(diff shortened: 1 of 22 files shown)


docs/src/main/sphinx/object-storage/file-formats.md
```diff
@@ -101,6 +101,12 @@ with Parquet files performed by supported object storage connectors:
     The equivalent catalog session property is
     `parquet_vectorized_decoding_enabled`.
   - `true`
+* - `parquet.max-footer-read-size`
```
Praveen2112, line 104: Do we need something for ORC footer as well ?

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -80,6 +93,13 @@ public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<
                 metadataIndex);
 
         int completeFooterSize = metadataLength + POST_SCRIPT_SIZE;
+        if (maxFooterReadSize.isPresent() && completeFooterSize > maxFooterReadSize.get().toBytes()) {
+            throw new ParquetCorruptionException(
```
Praveen2112, line 97: nit: Is it possible to assert this behaviour by creating a ParquetReader which has very small value ?

Praveen2112 (approved):
LGTM - Two minor questions

docs/src/main/sphinx/object-storage/file-formats.md
raunaqmorarka: I haven't seen this situation in ORC yet, even in parquet you need really bad configuration to reach this situation

PR #25977, opened 2025-06-11
Support PostGIS extension on non-public schemas in PostgreSQL

### LINKED ISSUE #25972: PostGIS support only works if postgis is installed in the public schema
We've just noticed that, when `postgis` is installed in any schema other than `public`, Trino does not properly interpret `geometry` columns as `geometry`. Instead, the column shows up as a `varchar` and the various Geometry Trino functions are not available for use.

We believe this is because the column type is reported as `myschema.geometry` in the case where the `postgis` extension is installed in `myschema` rather than `public`. This would fail the exact-match logic [on this line](https://github.com/trinodb/trino/blob/b65255c9f05c9784a2c4ba19e7531aa76460503d/plugin/trino-postgresql/src/main/java/io/trino/plugin/postgresql/PostgreSqlClient.java#L592).

Presumably, changing this logic to use some kind of looser regex match against the column type would produce the expected result.

brendanstennett:
Fixes #25972

## Description

This allows supporting of PostgreSQL PostGIS geometry types when the extension is installed on a schema other than public. 

Currently if PostGIS is installed on a schema other than public, the `Geometry` type column will not be recognized or will appear as a `varchar` if `CONVERT_TO_VARCHAR` is set.  This fixes this bug and will allow use of `Geometry` types when this extension is installed on another schema.

### Before change
Screenshot 2025-06-10 at 9 29 54 PM

### After change
Screenshot 2025-06-10 at 9 30 05 PM

## Release notes

```markdown
## PostgreSQL
* Add support for `geometry` types installed in schemas other than `public`. ({issue}`25972`)
```

### DIFF
--- a/plugin/trino-postgresql/src/main/java/io/trino/plugin/postgresql/PostgreSqlClient.java
+++ b/plugin/trino-postgresql/src/main/java/io/trino/plugin/postgresql/PostgreSqlClient.java
@@ -592,9 +592,14 @@ public class PostgreSqlClient
             case "geometry":
                 return Optional.of(geometryColumnMapping());
         }
+
+        // Handling of schema-qualified types
+        // TODO: Find more reliable way to detect these types. The type name can be eg. "schema-name"."vector"
         if (jdbcTypeName.endsWith("\"vector\"")) {
-            // TODO: Find more reliable way to detect vector type. The type name can be "schema-name"."vector"
             return Optional.of(vectorColumnMapping());
+        }
+        if (jdbcTypeName.endsWith("\"geometry\"")) {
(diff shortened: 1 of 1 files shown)


ebyhr (commented):
Please follow commit message guideline. 

https://trino.io/development/process.html#pull-request-and-commit-guidelines-
> Use the [imperative mood, as used in a command or request](https://en.wikipedia.org/wiki/Imperative_mood), in the subject line.

ebyhr:
Could you test this change locally and attach the screenshot in the PR description?

brendanstennett:
@ebyhr Sure thing! Updated commit message and attached screenshots from the local testing I did

PR #25978, opened 2025-06-11
Add support for Databricks 16.4 LTS environment to product tests

chenjian2664:
## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #25979, opened 2025-06-11
Test: Add support for Databricks 164 LTS environment to product tests

ebyhr:
## Description

Run tests with secrets #25978


ebyhr:
@chenjian2664 This is a mirror PR of #25978, as ok-to-test command doesn't work when we update `ci.yml`.

PR #25991, opened 2025-06-11
Add java_version to system.runtime.nodes table

wendigo:
## Release notes

```markdown
## Section
```


martint:
I don't think we should be exposing this. The java version is an implementation detail, just like the version of all the libraries used by Trino. In the future, the intention is to bundle the JDK with Trino so that users don't even have to think about what version they are deploying and running.

PR #26000, opened 2025-06-12
Improve internal communication authentication

wendigo:
By generating and validating request signatures rather then creating and parsing JWT tokens.

This makes it harder to spoof request even if internal signature is captured, as it allows only to call a given URI with a given timestamp validity. This was not the case for JWT token.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
+++ b/core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
@@ -19,57 +19,52 @@ import io.airlift.http.client.HttpRequestFilter;
 import io.airlift.http.client.Request;
 import io.airlift.log.Logger;
 import io.airlift.node.NodeInfo;
-import io.jsonwebtoken.JwtException;
-import io.jsonwebtoken.JwtParser;
+import io.airlift.units.Duration;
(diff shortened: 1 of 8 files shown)


core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
```diff
@@ -90,93 +81,72 @@ public InternalAuthenticationManager(String sharedSecret, String nodeId)
...
                 .build();
         setAuthenticatedIdentity(request, identity);
     }
 
+    private String signature(String nodeId, long requestTimestampMillis, URI uri)
+    {
+        return Base64.getEncoder()
+                .encodeToString(hashing.newHasher()
+                        .putUnencodedChars(nodeId)
+                        .putUnencodedChars(uri.toString())
```
dain: I'm not sure the full URI is maintained perfectly across the HTTP protocol.  I'd think about using just the path and maybe the query string.

wendigo:
@dain PTAL

wendigo:
@dain PTAL, I would like to land this improvement

wendigo:
CleanShot 2025-11-18 at 21 26 48@2x

orange is throughput after this change

PR #26006, opened 2025-06-13
Random improvements

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-tpch/src/main/java/io/trino/plugin/tpch/TpchRecordSet.java
+++ b/plugin/trino-tpch/src/main/java/io/trino/plugin/tpch/TpchRecordSet.java
@@ -191,11 +191,7 @@ public class TpchRecordSet<E extends TpchEntity>
         public long getLong(int field)
         {
             checkState(row != null, "No current row");
-            return getLong(getTpchColumn(field));
-        }
-
-        private long getLong(TpchColumn<E> tpchColumn)
-        {
+            TpchColumn<E> tpchColumn = getTpchColumn(field);
             if (tpchColumn.getType().getBase() == TpchColumnType.Base.DATE) {
                 return tpchColumn.getDate(row);
             }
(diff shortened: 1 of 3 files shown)


core/trino-main/src/main/java/io/trino/server/ServerConfig.java
```diff
@@ -68,6 +69,18 @@ public ServerConfig setIncludeExceptionInResponse(boolean includeExceptionInResp
         return this;
     }
 
+    public boolean isIncludeVersionInStacktrace()
+    {
+        return includeVersionInStacktrace;
+    }
+
+    @Config("version-in-stacktrace")
```
ebyhr: What is the motivation of this config property?

plugin/trino-tpch/src/main/java/io/trino/plugin/tpch/TpchRecordSet.java
```diff
@@ -191,11 +191,7 @@ public boolean getBoolean(int field)
         public long getLong(int field)
         {
             checkState(row != null, "No current row");
-            return getLong(getTpchColumn(field));
-        }
-
-        private long getLong(TpchColumn<E> tpchColumn)
```
raunaqmorarka, line 197: What was wrong with the existing code here ?

core/trino-main/src/main/java/io/trino/server/ServerConfig.java
raunaqmorarka: ya, why not always embed version ?

plugin/trino-tpch/src/main/java/io/trino/plugin/tpch/TpchRecordSet.java
wendigo: It was redundant

core/trino-main/src/main/java/io/trino/server/ServerConfig.java
wendigo: I want to test performance implication of adding a version to the calling frame

raunaqmorarka (approved):

PR #26009, opened 2025-06-14
Minor testing improvements

electrum:
## Release notes

### DIFF
--- a/core/trino-main/src/test/java/io/trino/sql/query/QueryAssertions.java
+++ b/core/trino-main/src/test/java/io/trino/sql/query/QueryAssertions.java
@@ -650,7 +650,7 @@ public class QueryAssertions
         private final boolean ordered;
         private boolean skipTypesCheck;
 
-        private ResultAssert(
+        public ResultAssert(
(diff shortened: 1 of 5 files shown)


wendigo (approved):

core/trino-main/src/test/java/io/trino/sql/query/QueryAssertions.java
```diff
@@ -757,6 +757,21 @@ public ResultAssert containsAll(MaterializedResult expected)
             });
         }
 
+        @CanIgnoreReturnValue
+        public ResultAssert hasColumnNames(String... expectedColumnNames)
```
ebyhr, line 761: This method doesn't exist intentionally https://github.com/trinodb/trino/commit/4d34ed908a75f393274a3c74ff1ae4016f217d1a

electrum: That change seems to be specifically around table functions. I don’t understand why verifying the output column names in the plan is problematic, as that’s the source of the names for the client protocol.

Issue #25999, opened 2025-06-12
Trino worker dies while processing the query from #21443


sajjoseph:
I was checking on this issue - #21443 - in 476 version.

Tried the below query:

```
SELECT *
FROM (
	SELECT concat_ws('', repeat(concat_ws('', repeat('a', 1000)), 500))
) CROSS JOIN UNNEST(sequence(1, 5000));
```

The worker node that worked on the query just died - no indication in the server log about the cause of the crash.

I turned the logging in the node (io.trino=DEBUG).
Following is the output:

```
2025-06-12T11:22:39.473Z        DEBUG   Query-20250612_112234_00000_it8bd-498   io.trino.sql.planner.LogicalPlanner     environment=production,host=worker1 io.trino.sql.planner.optimizations.BeginTableWrite:
Output[columnNames = [_col0, _col1]]
│   Layout: [expr:varchar, field:bigint]
│   _col0 := expr
│   _col1 := field
└─ CrossJoin Unnest[replicate = [expr:varchar], unnest = [expr_0:array(bigint)]]
   │   Layout: [expr:varchar, field:bigint]
   └─ LocalExchange[partitioning = ROUND_ROBIN]
      │   Layout: [expr:varchar, expr_0:array(bigint)]
      └─ Values[]
             Layout: [expr:varchar, expr_0:array(bigint)]
             (varchar 'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa....

2025-06-12T11:22:39.741Z        DEBUG   dispatcher-query-3      io.trino.execution.StageStateMachine    environment=production,host=worker1 Stage 20250612_112234_00000_it8bd.0 is PLANNED
2025-06-12T11:22:39.749Z        DEBUG   dispatcher-query-3      io.trino.execution.QueryStateMachine    environment=production,host=worker1  Query 20250612_112234_00000_it8bd is STARTING
2025-06-12T11:22:39.767Z        DEBUG   stage-scheduler io.trino.execution.scheduler.PipelinedStageExecution    environment=production,host=worker1 Pipelined stage execution 20250612_112234_00000_it8bd.0 is PLANNED
[...]

PR #26013, opened 2025-06-14
Partition spooled pages while encoding data to segments

wendigo:
This prevents dictionary blocks to expand to a single huge segment that most clients won't be able to read due to its size.

This now partitions these kind of pages to smaller ones and allows the operator to output multiple spooled/inlined pages in a single getOutput call.

Fixes https://github.com/trinodb/trino/issues/25999

## Release notes

```markdown
## Client protocol
* Partition large pages to avoid OOM while serializing data in the spooling protocol. ({issue}`25999`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
+++ b/core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
@@ -59,6 +59,7 @@ import static io.trino.operator.OutputSpoolingOperatorFactory.OutputSpoolingOper
 import static io.trino.operator.OutputSpoolingOperatorFactory.OutputSpoolingOperator.State.NEEDS_INPUT;
 import static io.trino.operator.SpoolingController.Mode.INLINE;
 import static io.trino.operator.SpoolingController.Mode.SPOOL;
+import static io.trino.server.protocol.spooling.SpooledMetadataBlockSerde.serialize;
 import static io.trino.server.protocol.spooling.SpoolingSessionProperties.getInitialSegmentSize;
 import static io.trino.server.protocol.spooling.SpoolingSessionProperties.getMaxSegmentSize;
 import static java.util.Objects.requireNonNull;
@@ -148,6 +149,10 @@ public class OutputSpoolingOperatorFactory
 
         private final SpoolingController controller;
         private final ZoneId clientZoneId;
+        private final AtomicLong spooledSegmentsCount = new AtomicLong();
+        private final AtomicLong inlinedSegmentsCount = new AtomicLong();
+        private final long maxSegmentSize;
+        private final long minSegmentSize;
 
         enum State
         {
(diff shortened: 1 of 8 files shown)


core/trino-main/src/main/java/io/trino/operator/SpoolingPagePartitioner.java
```diff
@@ -0,0 +1,113 @@
...
+import static java.lang.Math.clamp;
+
+public class SpoolingPagePartitioner
+{
+    static final double LOWER_BOUND = 0.05; // 5% of the target size
+    static final double UPPER_BOUND = 0.1; // 10% of the target size
+
+    private SpoolingPagePartitioner() {}
+
+    public static List<List<Page>> partition(List<Page> pages, long targetSize)
```
losipiuk, line 35: why operate on multiple pages

wendigo: We buffer multiple pages

raunaqmorarka, line 35: Have you considered using `io.trino.execution.buffer.PageSplitterUtil#splitPage(io.trino.spi.Page, long)` instead ?

wendigo: It's used in that method

losipiuk: I get it - but why not partition page by page. Passing multiple pages makes interface more complex. Let me read deeper

core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
```diff
@@ -148,6 +149,9 @@ static class OutputSpoolingOperator
 
         private final SpoolingController controller;
         private final ZoneId clientZoneId;
+        private final AtomicLong spooledSegmentsCount = new AtomicLong();
+        private final AtomicLong inlinedSegmentsCount = new AtomicLong();
```
losipiuk, line 153: those could be added in separate commit

core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
```diff
@@ -268,51 +274,68 @@ private Page outputBuffer(boolean lastPage)
...
+            for (int i = 0; i < partitions.size(); i++) {
+                List<Page> partition = partitions.get(i);
+                long rows = reduce(partition, Page::getPositionCount);
+                long size = reduce(partition, Page::getSizeInBytes);
+
+                if (i == partitions.size() - 1 && size < SPOOLING_THRESHOLD) {
+                    // If the last partition is small enough, inline it to save the overhead of spooling
+                    spooledMetadataBuilder.addAll(inline(partition));
+                    continue;
+                }
```
losipiuk: this is weird logic. More natural would be to put this remainder back to the list and spool it together with next set of pages. with this logic you can easily get output of shape:
* spooled 
* inline
* spooled
* inline
* ....

wendigo: yes, this is correct and in line with how protocol was designed - the end goal is to avoid spooling small segments which is a cost that doesn't make sense if the inline page fit in the 1MB query result threshold

wendigo: @losipiuk this happens for the last page in the operator pipeline so there won't be more pages to merge with

losipiuk: > @losipiuk this happens for the last page in the operator pipeline so there won't be more pages to merge with 

Let's chat offline - I do not see that is the case in current version of the code.

ebyhr:
@wendigo Could you update the release note entry?

wendigo:
@ebyhr done

PR #26015, opened 2025-06-14
Make queued and buffered rows configurable in cli

wendigo:
This prevents OOM exception when retrieving huge results

Relates to https://github.com/trinodb/trino/pull/26013

## Release notes

```markdown
## CLI
* Allow configuring `--max-buffered-rows` and `--max-queued-rows` in the CLI ({issue}`issuenumber`)
```

### DIFF
--- a/client/trino-cli/src/main/java/io/trino/cli/Console.java
+++ b/client/trino-cli/src/main/java/io/trino/cli/Console.java
@@ -177,7 +177,9 @@ public class Console
         try (QueryRunner queryRunner = new QueryRunner(
                 uri,
                 session,
-                clientOptions.debug)) {
+                clientOptions.debug,
+                clientOptions.maxQueuedRows,
+                clientOptions.maxBufferedRows)) {
             if (hasQuery) {
                 return executeCommand(
                         queryRunner,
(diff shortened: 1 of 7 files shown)


client/trino-cli/src/main/java/io/trino/cli/Console.java
```diff
@@ -177,7 +177,9 @@ public boolean run()
         try (QueryRunner queryRunner = new QueryRunner(
                 uri,
                 session,
-                clientOptions.debug)) {
+                clientOptions.debug,
+                clientOptions.maxQueuedRows,
```
chenjian2664, line 181: Can we have a check for those two configs? i,e at least should greater than 0.

client/trino-cli/src/test/java/io/trino/cli/TestClientOptions.java
```diff
@@ -383,6 +383,8 @@ private boolean isCliSpecificOptions(String name)
             case "editingMode":
```
chenjian2664, line 383: Same here, can we have a unsupported test

chenjian2664 (approved):

losipiuk:
@wendigo this is super technical. Can't we just monitor how big rows are and automatically adjust based on that?

wendigo:
@losipiuk i was thinking about that but there is no easy way of calculating row size hence it's not possible to bound it

PR #26020, opened 2025-06-17
Run case-sensitive property test only in OSS Delta

ebyhr:
## Description

Running with Spark looks sufficient. 

## Release notes


chenjian2664 (commented):
If remove, where is the place we could test Databricks behavior for table properties?

ebyhr:
I don't think it's necessary. Ensuring case-sensitivity with `DELTA_LAKE_OSS` is sufficient in my opinion.

chenjian2664 (approved):

wendigo (approved):

PR #26030, opened 2025-06-18
changes html to get fonts from local css file

nadavtzaysler:
## Description
added css file to fetch all used fonts from fixed path in app
changed dockerfile to take all used fonts from fonts.gstatic and load them into app
changed html to get css from file instead of google api

## Release notes

( X) This is not user-visible or is docs only, and no release notes are required.

```markdown
## Section
```

### DIFF
--- a/core/docker/Dockerfile
+++ b/core/docker/Dockerfile
@@ -67,6 +67,12 @@ COPY --chown=trino:trino trino-server/bin/linux-${ARCH} /usr/lib/trino/bin/linux
 COPY --chown=trino:trino default/etc /etc/trino
 COPY --chown=trino:trino --from=jvmkill /libjvmkill.so /usr/lib/trino/bin
 
+RUN mkdir -p /usr/lib/trino/web/assets/fonts && \
+    curl -L "https://fonts.gstatic.com/s/roboto/v30/KFOmCnqEu92Fr1Mu4mxKKTU1Kg.woff2" -o /usr/lib/trino/web/assets/fonts/Roboto-Regular.woff2 && \
+    curl -L "https://fonts.gstatic.com/s/roboto/v30/KFOlCnqEu92Fr1MmEU9fBBc4AMP6lQ.woff2" -o /usr/lib/trino/web/assets/fonts/Roboto-Medium.woff2 && \
+    curl -L "https://fonts.gstatic.com/s/roboto/v30/KFOlCnqEu92Fr1MmWUlfBBc4AMP6lQ.woff2" -o /usr/lib/trino/web/assets/fonts/Roboto-Bold.woff2 && \
+    chown -R trino:trino /usr/lib/trino/web/assets/fonts
+
 EXPOSE 8080
 USER trino:trino
 CMD ["/usr/lib/trino/bin/run-trino"]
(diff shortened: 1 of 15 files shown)


core/docker/Dockerfile
```diff
@@ -67,6 +67,12 @@ COPY --chown=trino:trino trino-server/bin/linux-${ARCH} /usr/lib/trino/bin/linux
 COPY --chown=trino:trino default/etc /etc/trino
 COPY --chown=trino:trino --from=jvmkill /libjvmkill.so /usr/lib/trino/bin
 
+RUN mkdir -p /usr/lib/trino/web/assets/fonts && \
```
mosabua: Lets not get them with each build .. just download them and check them into the repo

core/trino-web-ui/src/main/resources/webapp/oauth2/failure.html
```diff
@@ -0,0 +1,52 @@
+<!--
```
mosabua: Why are you adding these pages? If we need them like this we should at least have them in a separate commit or even separate PR.

mosabua (changes requested):
Are you going to send a separate PR for the preview UI or is it all sorted out there already?

core/trino-web-ui/src/main/resources/webapp/oauth2/failure.html
nadavtzaysler: those files seem to come from a different pr actually, removed them.

core/docker/Dockerfile
nadavtzaysler: fixed, now all the font that are being currently used are in assets/fonts/fonts.css

nadavtzaysler:
> Are you going to send a separate PR for the preview UI or is it all sorted out there already?
The preview UI does not suffer from this issue, however I will open a new pr for the docs which also becomes slower because of the same issue

core/trino-web-ui/src/main/resources/webapp/oauth2/failure.html
mosabua: Thanks

mosabua:
How does the current approach actually help? You are still not locally embedding the font files .. you are just using a CSS file now to point to them?

nadavtzaysler:
As of now before this pull request trino is trying to fetch the css files from the website fonts.google which causes slowness when this site or internet is not available, this pr is making it so the css files that trino is trying to fetch are already within the repo fixing the slowness without access to fonts.google and improving overall latency of the webserver, at the end its a small fix for a small issue

mosabua:
Thanks for actually adding the fonts.. before you just changed to a different URL from which to load the files and lead them from a separate CSS file .. now the files are actually local. 

Please rebase and test this with different browsers and check on the browser console that no other fonts are requested

mosabua:
Also squash commits and update the commit message to 

``` 
Localize fonts for web ui in the application

- Enable running UI without internet access and loading fonts from Google
```

wendigo:
Superseded by https://github.com/trinodb/trino/pull/26053

nadavtzaysler:
Why was this pr superseded?

nadavtzaysler:
@mosabua Can this pr be reopened,  the pr that was supposed to supersede it doesn't seem more ideal and it's stale while this could've been merged

PR #26032, opened 2025-06-19
Rename variable for clarity in ParquetPageSourceFactory

chenjian2664:
## Description

The variable name ending with builder is misleading—it actually holds a list, not a builder.

## Release notes


ebyhr:
> The variable name ending with builder is misleading—it actually holds a list, not a builder.

I don't think the original name was misleading as the variable is used like a builder.

chenjian2664:
It confuses me when I look at this line:
```       
 ParquetReader parquetReader = parquetReaderProvider.createParquetReader(parquetColumnFieldsBuilder, appendRowNumberColumn);
```
I was trying to get to know why the method accept the builder, and then to follow the code to see where does the parquetColumnFieldsBuilder actual calling build method.

ebyhr:
The new name is misleading in for loop as the value is not finalized there. `Builder` suffix was helpful to clarify it's a variable holding a middle state. Just adding `ImmutableList.copyOf` is enough to me.

ebyhr (approved):

PR #26037, opened 2025-06-19
Minor code cleanup

dain:
## Description
* Decouple DynamicFilterService and SqlQueryExecution
* When updating query info, return existing final info if already set

## Release notes


core/trino-main/src/main/java/io/trino/server/DynamicFilterService.java
```diff
@@ -133,8 +131,8 @@ public void registerQuery(SqlQueryExecution sqlQueryExecution, SubPlan fragmente
         // register query only if it contains dynamic filters
         if (!dynamicFilters.isEmpty()) {
             registerQuery(
-                    sqlQueryExecution.getQueryId(),
-                    sqlQueryExecution.getSession(),
+                    session.getQueryId(),
```
electrum, line 134: This parameter could be removed and done in the method

Issue #24435, opened 2024-12-10
Separate access control to plugin functions from built in functions


homar:
Currently built in functions and [plugin functions](https://github.com/trinodb/trino/blob/90db2f54f25c8b52db56bdfcabea2c1b4fa6dac9/core/trino-spi/src/main/java/io/trino/spi/Plugin.java#L65) all end up in `system.builtin` and access control treats them the same - they are always allowed. 

I believe it would be beneficial to allow more fine grained control  around plugin functions.

That would definitely require changes in file based access control and probably in other places that I am not aware of.

martint:
This is intentional, since legacy plugin functions are not namespaced. To control access to plugin functions, you need to implement them as catalog functions using the newer APIs:  ConnectorMetadata.getFunctions(), etc.

electrum:
I don't think that "legacy" is the right term. Catalog functions is a low level interface. There's no way to do annotation based functions using that API. If we want to deprecate the old way, then we need a clear migration path.

lordicecream:
Duplicate of #19912

mosabua:
See discussion in https://github.com/trinodb/trino/pull/26041

Issue #23278, opened 2024-09-04
File based access control doesn't work for procedures and functions


surendragalwa:
Connector: PostgresSQL

_/etc/trino/access-control/rules.json_

```json5
{
  "catalogs": [
    {
      "group": "admin",
      "catalog": ".*",
      "allow": "all"
    },
  ],
  "functions": [
    {
      "group": "admin",
      "schema": "system",
      "function": ".*",
      "privileges": [
        "EXECUTE"
      ]
    }
  ],
  "procedures": [
    {
      "group": "admin",
      "schema": "system",
      "procedure": ".*",
      "privileges": [
        "EXECUTE"
      ]
    }
  ]
}
```

The rules grant EXECUTE permissions on specified procedures and table functions to admins.

**Procedure-**

On running a passthrough query (a procedure), it fails.

Query
```
CALL catalog_name.system.execute(query => 'SELECT COUNT(*) FROM schema_name.table_name');
```
Error
```
Query 20240902_122014_00036_wktdv failed: Access Denied: Cannot execute procedure catalog_name.system.execute
```

**Table function-**

Table passthrough query which runs query table function -

Query
```
SELECT * FROM TABLE(catalog_name.system.query(query => 'SELECT COUNT(*) FROM catalog_name.schema_name.table_name'));
```

Error
```
Access Denied: Cannot execute function catalog_name.system.query
```

Relates to https://github.com/trinodb/trino/issues/22874, https://github.com/trinodb/trino/issues/17689

surendragalwa:
@hashhar

beauttie:
Just curious as I'm running into a similar issue. Are you on version [431](https://trino.io/docs/current/release/release-431.html#security) for Trino? This release is supposed to support procedure rules in file-based access control.

lordicecream:
@surendragalwa did you get this working?
I am facing the same issue when upgrading from 418 to 464..the procedure rule seems to be ok, but the functions rules are not doing anything.
Even if I don't add the functions block, by default users should not have any access to functions, but everyone can access all functions

AlekHed:
Similiar issiue on 464.  Access Denied: Cannot execute procedure catalog.system.execute

```
"functions": [
    {
      "group": "admin",
      "schema": "system",
      "function": "execute",
      "privileges": [
        "EXECUTE"
      ]
    },
    {
      "catalog": "iceberg",
      "privileges": [
        "EXECUTE", "GRANT_EXECUTE", "OWNERSHIP"
[...]

Issue #19912, opened 2023-11-27
Could you explain why Trino decided to always skip access control on inline and built-in functions..?


okayhooni:
We currently use Apache Ranger for access control on Trino queries, and manage policies to control special privileges to use in-line functions such as some sensitive unhash-like functions. (I know, Ranger is not the officially supported plugin for Trino)

This access control on the in-line functions was working well until `v412` (current version of our production cluster), but it's not working anymore on the latest release of Trino(`v433`) with some breaking changes on the redesigning of access control codes.

I found the reason on the code like below.

- https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/metadata/FunctionResolver.java#L312
```
private static boolean canExecuteFunction(Session session, AccessControl accessControl, CatalogSchemaFunctionName functionName)
    {
        if (isInlineFunction(functionName) || isBuiltinFunctionName(functionName)) {
            return true;
        }
        return accessControl.canExecuteFunction(
                SecurityContext.of(session),
                new QualifiedObjectName(functionName.getCatalogName(), functionName.getSchemaName(), functionName.getFunctionName()));
    }
```

It's easy to fix these code lines on our forked repository of Trino,
but I wonder why Trino decided to remove access control on inline function at all.

hashhar:
@dain

okayhooni:
related PR: https://github.com/trinodb/trino/pull/19160

hashhar:
the reason why in-built functions are always allowed is because they are considered "safe". Can you explain which functions you want to disallow and why?

lordicecream:
@hashhar @dain I have deployed multiple UDFS, that I don't want to expose to everyone, but this issue is allowing all users to access all functions...the functions block is basically useless if this there are no check being done...
Can you please let me know when this is expected to be patched and if there is any workaround?

Trino upgrades are basically blocked for me and my teams across all environments due to this vulnerability

lordicecream:
Yes exactly, @hashhar I have around 5-10 UDFs and if I don't put any function rule block, then they should all be by default not accessible right?

I am seeing that those UDFs are not being controlled and everyone can access all functions including the (UDFS) even when I am not adding the block for functions rule.

Using file based access controller.

Will enable the debug logging and share logs.

lordicecream:
@hashhar even debug logs didn't show anything useful...I am still not able to have control over who can access what function

martint:
[...]

PR #26041, opened 2025-06-20
Optional access control for inbuilt functions

lordicecream:
This is a PR to add a flag that allows users to have to option to enable access control for inbuilt functions.
Currently all inbuilt functions are treated as safe and we are not able to restrict access to specific function as per rbac policy.
This can be problematic as users might wish to have some control over what they wish to restrict over different environments.

By default the inbuilt functions will still not be blocked, this will add a configuration "function-access-control-enabled" which can be marked true in config.properties for users to enable access control over all functions

## Release notes

```markdown
* Fixes : #19912  #24435 #23278 
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/FeaturesConfig.java
+++ b/core/trino-main/src/main/java/io/trino/FeaturesConfig.java
@@ -123,6 +123,8 @@ public class FeaturesConfig
 
     private boolean faultTolerantExecutionExchangeEncryptionEnabled = true;
 
+    private boolean functionAccessControlEnabled;
+
     public enum DataIntegrityVerification
     {
         NONE,
@@ -522,5 +524,18 @@ public class FeaturesConfig
     public void applyFaultTolerantExecutionDefaults()
     {
         exchangeCompressionCodec = LZ4;
+    }
+
+    @Config("function-access-control-enabled")
+    @ConfigDescription("Enable access control on all functions")
+    public FeaturesConfig setFunctionAccessControlEnabled(boolean enabled)
+    {
+        this.functionAccessControlEnabled = enabled;
+        return this;
+    }
+
+    public boolean isFunctionAccessControlEnabled()
+    {
+        return functionAccessControlEnabled;
     }
 }
--- a/core/trino-main/src/main/java/io/trino/metadata/FunctionResolver.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/FunctionResolver.java
@@ -70,18 +70,21 @@ public class FunctionResolver
     private final LanguageFunctionManager languageFunctionManager;
     private final WarningCollector warningCollector;
     private final FunctionBinder functionBinder;
+    private final boolean functionAccessControlEnabled;
 
     public FunctionResolver(
             Metadata metadata,
             TypeManager typeManager,
             LanguageFunctionManager languageFunctionManager,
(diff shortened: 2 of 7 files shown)


martint:
Can you elaborate on why this is a useful feature?

For inline function definitions, this wouldn't solve any problems I can think of. Inline functions are part of the query, just like anything in the SELECT, FROM, WHERE or other clauses, so it doesn't make sense to perform access controls checks -- the user could easily bypass them by changing the name of the function, since the name doesn't mean anything outside of the query in which it's being used.

For built-in functions, what's the scenario where "users might wish to have some control over what they wish to restrict over different environments"? Built-in functions have no side effects . They can't read or write any external data, connect to other systems, etc., so there's really no benefit from restricting access to them.

lordicecream:
> Can you elaborate on why this is a useful feature?
> 
> For inline function definitions, this wouldn't solve any problems I can think of. Inline functions are part of the query, just like anything in the SELECT, FROM, WHERE or other clauses, so it doesn't make sense to perform access controls checks -- the user could easily bypass them by changing the name of the function, since the name doesn't mean anything outside of the query in which it's being used.
> 
> For built-in functions, what's the scenario where "users might wish to have some control over what they wish to restrict over different environments"? Built-in functions have no side effects . They can't read or write any external data, connect to other systems, etc., so there's really no benefit from restricting access to them.

Hi @martint ,

In our use case, we're dealing with extremely large data volumes — around 850 million records ingested daily. Users frequently run aggregations and analytics over months/years of data, across multiple environments configured for different workloads.

While the built-in functions themselves (like `sha256`, `md5`, etc.) are safe and deterministic, we’ve found it important to retain fine-grained control over their usage. In environments with a large and diverse user base, even a few inefficient queries — especially ones involving heavy hashing operations — can lead to significant CPU contention and performance degradation at cluster level.

This isn't about restricting functionality for security reasons, but more about governance and protecting shared resources. Giving administrators the flexibility to control usage of resource-heavy functions ensures better stability, predictability, and performance at scale — especially in multi-tenant environments.

I have edited my PR description, this is only scoped for the inbuilt functions and not inline, that was an error from me

lordicecream:
#19912 #24435 #23278 @ebyhr @dain can you please review

mosabua:
The fact remains that by restricting function use you are not actually solving the problem you state. It is perfectly simple to write normal SQL queries on large data set that will cause the same issue. 

What you actually need to do is figure out a system that allows you to control queries of any shape based on their predicated and actual resource use (and therefore cost). Some of this is already possible in Trino and it would make sense to figure out steps to prevent queries from running with certain estimated. In the Trino Gateway project we are working towards some of those sort of features with rate limiting the amount of queries and setting policies based on query plan.. you can combine that with an external service and then control it as you want .. this is all still early but in my opinion the correct approach.

The idea to just restrict certain part of the overall language is bad for users and also bad for managing the cluster and you wont achieve what you are actually aiming for. I strongly suggest we do NOT proceed with this approach. cc @dain @martint @electrum 

From what I can see this PR also only adds a boolean flag to deactivate function use .. thats not really access control. Also note that for SQL UDFs and Python UDFs they are implemented in a plugin .. so if you want to disable use .. just remove the plugin. There is more info about all this in the plugin documentation. And you could have multiple clusters some with and some without UDFs.. behind Trino Gateway..

PR #26051, opened 2025-06-23
Revert "Cleanup previous snapshot files during materialized view refresh"

mayankvadariya:
This reverts commit 48135782c4c3cfc68032219fb08e0e75fab25067.

The current change may disrupt an existing running query that is accessing a materialized view snapshot, which could be cleaned up by a 'REFRESH MATERIALIZED VIEW' command. This change needs to be reapplied, with the retention period for past snapshots controlled through a configuration setting.

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2178,7 +2178,33 @@ public class IcebergMetadata
                 IcebergSessionProperties.EXPIRE_SNAPSHOTS_MIN_RETENTION);
 
         long expireTimestampMillis = session.getStart().toEpochMilli() - retention.toMillis();
-        executeExpireSnapshots(table, session, expireTimestampMillis);
+        TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), table.io().properties());
+        List<Location> pathsToDelete = new ArrayList<>();
+        // deleteFunction is not accessed from multiple threads unless .executeDeleteWith() is used
+        Consumer<String> deleteFunction = path -> {
+            pathsToDelete.add(Location.of(path));
+            if (pathsToDelete.size() == DELETE_BATCH_SIZE) {
+                try {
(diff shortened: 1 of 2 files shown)


raunaqmorarka (approved):

mayankvadariya:
@raunaqmorarka thanks, please run CI with secrets.

raunaqmorarka:
> @raunaqmorarka thanks, please run CI with secrets.

Why aren't the changes in TestIcebergFileOperations part of the revert commit ?

raunaqmorarka:
/test-with-secrets sha=4ef626791288aa8e0e5171f314caa3222342434e

mayankvadariya:
> > @raunaqmorarka thanks, please run CI with secrets.
> 
> Why aren't the changes in TestIcebergFileOperations part of the revert commit ?

I also wondered the same when I changed this PR to draft but then I realized the test had been removed through https://github.com/trinodb/trino/commit/5210b0e8782281e522406a2154b7044f32bcd8a3#diff-33f2d439b6cac05c1a8528131271f2ff815478c6344b174870c8688370056de1

okayhooni:
Thanks! @mayankvadariya 

After upgrading Trino to version 476, we received a user report that the refresh query for an Iceberg materialized view no longer works and gets stuck at 100%.

Following the content of this PR, we will revert the regression-causing commit to address the issue!

image

chenjian2664:
@raunaqmorarka @mayankvadariya @ebyhr  Can we revisit this now - add it back, since the https://github.com/trinodb/trino/pull/26230 probably fixes the thread safety that in the original https://github.com/trinodb/trino/commit/48135782c4c3cfc68032219fb08e0e75fab25067

see https://github.com/trinodb/trino/pull/26310

raunaqmorarka:
@chenjian2664 There were two reasons for reverting this
1. It was causing REFRESH MV to be slow. We would need to verify that this would no longer be the case with new implementation.
2. It was potentially deleting snapshots that some concurrent table scan on the MV could have been reading. We should always keep at least the previous snapshot to the new one around for running queries. For MVs which update frequently and have long running queries on them, we could expose a `snapshots_retention_threshold` MV property have it default to 4h or something similar.

PR #26053, opened 2025-06-23
Use self-hosted Roboto font in the UI

wendigo:
## Release notes

```markdown
## Section
```


nadavtzaysler:
Why was the pr before this supersede?

wendigo:
@nadavtzaysler it's a simpler implementation

nadavtzaysler:
@wendigo  It's a simple change in the css file, could've been made in the same pr

Also woof2 files are considered better to use in web apps so i believe the implementation before is more ideal

nadavtzaysler:
@wendigo @mosabua I still dont understand why is this simple remove of a few lines in the fonts.css file is requiring a new pr, basically until 5 minutes ago it was just a worse copy of the original pr that will take about 3 to 4 more days to merge while the original was ready

wendigo:
@nadavtzaysler this PR isn't a copy, I've made it from scratch addressing the linked issue

mosabua:
I dont mind which PR is merged - whichever is better.  I think we should just add a co authored by line in the commit message so the changes are attributed to both @nadavtzaysler and @wendigo

wendigo:
Since CLA is not signed I can't add authorship and as said these are two separate PRs addressing the same issue with two different and distinct implementations @mosabua

PR #26054, opened 2025-06-23
Allow custom properties in Iceberg REST catalog

### LINKED ISSUE #24236: Allow passing custom properties to Iceberg REST catalog
### Proposal

Allow passing custom REST catalog properties in the config and session parameters:

```
iceberg.rest-catalog.additional-properties=prop1=val1,prop2=val2
```

```
SET SESSION <catalog_name>.rest-additional-properties=[ARRAY or MAP]['prop1=val1', 'prop2=val2']
```

### Rationale

Iceberg REST specification allows properties to be passed to the REST catalog. Some of these properties are hard-coded and expected by the REST client (e.g., URL, warehouse). Other properties are custom and could be passed to the remote catalog, e.g., `header.X-My-Custom-Header=value` would translate into `X-My-Custom-Header=value` HTTP header in all REST catalog requests. 

Since (1) multiple REST catalogs with different capabilities emerge and (2) even built-in Iceberg properties change over time (see https://github.com/apache/iceberg/commits/main/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java).

Both Spark and Flink have already solved this issue, allowing users to propagate custom properties without any restrictions. E.g.:
```
spark.sql.catalog.test_catalog.header.X-My-Custom-Header=value
```

The ability to pass custom catalog properties is important to ensure that users can always utilize the full power of the underlying REST catalog implementation, regardless of the Trino version used.

rambleraptor:
## Description
Closes #24236

This allows custom properties to be passed through to the Iceberg REST Catalog. This 

The idea is that certain implementations of the Iceberg REST Catalog will expect certain headers being sent. Rather than trying to enumerate all of them across an ever-growing number of implementations, we should allow users to pass in these custom headers.

This adds a new config value + passes it through with the rest of the properties.

#24236 has more discussion around this.

## Release notes

```markdown
## Section
* Adds support for passing custom properties to the Iceberg REST Catalog using `iceberg.rest-catalog.additional-properties`.
```

### DIFF
--- a/docs/src/main/sphinx/object-storage/metastores.md
+++ b/docs/src/main/sphinx/object-storage/metastores.md
@@ -526,6 +526,9 @@ following properties:
 * - `iceberg.rest-catalog.case-insensitive-name-matching.cache-ttl`
   - [Duration](prop-type-duration) for which case-insensitive namespace, table, 
     and view names are cached. Defaults to `1m`.
+* - `iceberg.rest-catalog.additional-properties`
+  - Headers that should be passed over to the REST catalog.
+    Should be provided as a list of strings in the `header=value` format.
   :::
 
 The following example shows a minimal catalog configuration using an Iceberg
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogConfig.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogConfig.java
@@ -22,7 +22,10 @@ import jakarta.validation.constraints.NotNull;
 import org.apache.iceberg.CatalogProperties;
 
 import java.net.URI;
+import java.util.List;
+import java.util.Map;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static java.util.concurrent.TimeUnit.MINUTES;
@@ -54,6 +57,28 @@ public class IcebergRestCatalogConfig
     private boolean sigV4Enabled;
     private boolean caseInsensitiveNameMatching;
     private Duration caseInsensitiveNameMatchingCacheTtl = new Duration(1, MINUTES);
+    private Map<String, String> additionalProperties = Map.of();
+
+    public Map<String, String> getAdditionalProperties()
+    {
+        return additionalProperties;
+    }
+
+    @Config("iceberg.rest-catalog.additional-properties")
+    @ConfigDescription("Additional properties for the REST catalog")
+    public IcebergRestCatalogConfig setAdditionalProperties(List<String> additionalProperties)
+    {
+        try {
+            this.additionalProperties = additionalProperties
+                    .stream()
+                    .collect(Collectors.toUnmodifiableMap(kvs -> kvs.split(":", 2)[0], kvs -> kvs.split(":", 2)[1]));
+        }
+        catch (IndexOutOfBoundsException e) {
+            throw new IllegalArgumentException(String.format("Cannot parse http headers from property iceberg.rest-catalog.additional-properties; value provided was %s, " +
+                    "expected format is \"Header-Name-1: header value 1, Header-Value-2: header value 2, ...\"", String.join(", ", additionalProperties)), e);
+        }
+        return this;
+    }
 
     @NotNull
     public URI getBaseUri()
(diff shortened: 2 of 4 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogConfig.java
```diff
@@ -54,6 +54,20 @@ public enum SessionType
...
     private boolean caseInsensitiveNameMatching;
     private Duration caseInsensitiveNameMatchingCacheTtl = new Duration(1, MINUTES);
+    private Map<String, String> additionalProperties = Map.of();
+
+    public Map<String, String> getAdditionalProperties()
+    {
+        return additionalProperties;
+    }
+
+    @Config("iceberg.rest-catalog.additional-properties")
```
ebyhr, line 67: Please update `TestIcebergRestCatalogConfig` and the documentation.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogConfig.java
```diff
@@ -54,6 +54,20 @@ public enum SessionType
...
+    @ConfigDescription("Additional properties for the REST catalog")
+    public IcebergRestCatalogConfig setAdditionalProperties(Map<String, String> additionalProperties)
```
ebyhr: Does `Map<String, String>` really work? Please take a look at `io.trino.plugin.httpquery.HttpEventListenerConfig#setHttpHeaders`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogConfig.java
```diff
@@ -54,6 +54,20 @@ public enum SessionType
...
+    {
+        this.additionalProperties = Map.copyOf(additionalProperties);
```
ebyhr: `ImmutableMap.copyOf`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoIcebergRestCatalogFactory.java
```diff
@@ -110,6 +111,7 @@ public TrinoIcebergRestCatalogFactory(
                 .expireAfterWrite(restConfig.getCaseInsensitiveNameMatchingCacheTtl().toMillis(), MILLISECONDS)
                 .shareNothingWhenDisabled()
                 .build();
+        this.additionalProperties = restConfig.getAdditionalProperties();
```
ebyhr: `ImmutableMap.copyOf`.

ebyhr (changes requested):
Doing "Request changes" since we haven't yet reached an agreement on allowing arbitrary properties in the Iceberg REST catalog.

ebyhr:
Please capitalize the commit title https://trino.io/development/process.html#pull-request-and-commit-guidelines-. Also, the current title seems a little unclear. We could change it to "Allow custom properties in Iceberg REST catalog" or something.

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -526,6 +526,9 @@ following properties:
 * - `iceberg.rest-catalog.case-insensitive-name-matching.cache-ttl`
   - [Duration](prop-type-duration) for which case-insensitive namespace, table, 
     and view names are cached. Defaults to `1m`.
+* - `iceberg.rest-catalog.additional-properties`
+  - Headers that should be passed over to the REST catalog.
+    Should be provided as a list of strings in the `header=value` format.
```
singhpk234, line 531: why a list of strings and not something like : 
```
iceberg.rest-catalog.header=value
```
like what spark does

spark.sql.catalog.<spark-catalog-name>.header.X-Iceberg-Access-Delegation=vended-credentials 

This would free us from interpretting the `=` and value contain `=` if there are base64 characters ?

singhpk234 (changes requested):
Thank you the change, I think its a feature i am also interested in !

docs/src/main/sphinx/object-storage/metastores.md
ebyhr: @singhpk234 The dynamic config key is unsupported in Trino's configuration framework.

chenjian2664, line 531: The implementation is using `:`, so here is `header:value`?

ebyhr:
I believe your motivation is setting `x-goog-user-project` header for BigLake metastore. We prefer dedicated config properties in this project. #26219 is the alternative approach.

Issue #26058, opened 2025-06-24
Incorrect results on parquet files written by parquet-cpp-arrow version 5.0.0


okayhooni:
After upgrading Trino version from `451` to `476`, we discovered a regression where **Hive tables created on the location of existing Parquet files in S3 return zero rows** when queried.

This pattern with "preloaded-data" worked correctly in Trino version `451` and also works in Apache Spark, but in Trino version `476`, the same setup results in an empty table scan — even though the data files are present and readable.

This regression is particularly impactful for our users, many of whom follow a pattern where:

- They preprocess data using tools such as Pandas
- Upload the resulting Parquet files directly to S3
- Create Hive tables in that S3 location to query that data with Trino

#### Reproduction steps

1. Upload valid Parquet files to an S3 directory.
2. Create a Hive table in anywhere(Spark, Trino v451, or Trino v476) pointing to that location, like:
   ```sql
   CREATE TABLE hive.default.preloaded_parquet_test (
     id BIGINT,
     name VARCHAR
   )
   WITH (
     external_location = 's3://my-bucket/test-data/preloaded/',
     format = 'PARQUET'
   );
3. Query the table:
```sql
SELECT * FROM hive.default.preloaded_parquet_test;
```
→ Returns zero rows

However...the data is read correctly and returned as expected with Spark or Trino v451.

We use AWS Glue Catalog as the metastore.  so, I initially suspected the issue might be related to Trino’s switch to the AWS Glue Data Catalog V2 API, but we confirmed that even Trino 451 works correctly when configured to use Glue V2 API.

### Expected behavior
Trino should discover and read the existing Parquet files in the specified external location and return the data, as it did in version 451 (and as Spark still does).

### Additional context

(additional tests on partitioned tables with "preloaded-data" file location)

Even for partitioned tables, the issue persists:

- Running `MSCK REPAIR TABLE` from Spark to register partitions, or
- Executing Trino’s `CALL system.sync_partition_metadata(...)` procedure

...does not resolve the issue in Trino v476. The table still returns zero rows.
(the data is read correctly and returned as expected with Spark or Trino v451.)

We hope this regression can be addressed in an upcoming patch release.

Let us know if any additional diagnostics or test files are needed to reproduce the issue..!

ebyhr:
Can you find the first bad version with bisect? What happens if you remove the following `/` from `external_location` table property?

okayhooni:
@ebyhr ,

[...]

PR #26064, opened 2025-06-25
 Avoid relying on org.apache.parquet.format.RowGroup.getFile_offset

raunaqmorarka:
## Description
Certain parquet writers like pyarrow can set this offset incorrectly.
That can lead to the parquet reader not idenitfying the row groups
within a split correctly. This code is changed to rely on first page
offset in ColumnChunk instead.

Fixes https://github.com/trinodb/trino/issues/26058
Related to https://github.com/trinodb/trino/pull/24618

Newer versions of pyarrow (tested parquet-cpp-arrow version 20.0.0) don't have this problem, the file showing this problem is using `parquet-cpp-arrow version 5.0.0`

## Release notes

```markdown
## Iceberg, Hive, Hudi, Delta Lake
* Fix incorrect results when reading from parquet files produced by old versions of pyarrow. ({issue}`26058`)
```

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/metadata/ParquetMetadata.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/metadata/ParquetMetadata.java
@@ -111,17 +111,15 @@ public class ParquetMetadata
                 long fileRowCountOffset = fileRowCount;
                 fileRowCount += rowGroup.getNum_rows(); // Update fileRowCount for all row groups
 
-                if (rowGroup.isSetFile_offset()) {
-                    long rowGroupStart = rowGroup.getFile_offset();
-                    boolean splitContainsRowGroup = splitStart <= rowGroupStart && rowGroupStart < splitStart + splitLength;
-                    if (!splitContainsRowGroup) {
-                        continue;
-                    }
-                }
-
                 List<ColumnChunk> columns = rowGroup.getColumns();
                 validateParquet(!columns.isEmpty(), dataSourceId, "No columns in row group: %s", rowGroup);
                 String filePath = columns.get(0).getFile_path();
+                long rowGroupStart = getRowGroupStart(columns, messageType);
+                boolean splitContainsRowGroup = splitStart <= rowGroupStart && rowGroupStart < splitStart + splitLength;
+                if (!splitContainsRowGroup) {
+                    continue;
+                }
+
                 ImmutableList.Builder<ColumnChunkMetadata> columnMetadataBuilder = ImmutableList.builderWithExpectedSize(columns.size());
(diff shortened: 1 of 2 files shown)


lib/trino-parquet/src/main/java/io/trino/parquet/metadata/ParquetMetadata.java
```diff
@@ -145,6 +143,14 @@ public FileMetaData getParquetMetadata()
         return parquetMetadata;
     }
 
+    private static long getRowGroupStart(List<ColumnChunk> columns, MessageType messageType)
+    {
+        // Note: Do not rely on org.apache.parquet.format.RowGroup.getFile_offset or org.apache.parquet.format.ColumnChunk.getFile_offset
+        // because some versions of parquet-cpp-arrow (and potentially other writers) set it incorrectly
+        ColumnChunkMetadata columnChunkMetadata = toColumnChunkMetadata(columns.getFirst(), null, messageType);
```
Praveen2112, line 150: For my understaning but would this result in reading and processing additional data instead of loading the same from `rowGroup` information ?

raunaqmorarka: The metadata is all in memory by this point. The previous code was saving these `toColumnChunkMetadata` calls for all columns which are not within a split.
Now we will do an additional processing for first column in each row group in the file to get the row group start offset. Not ideal, but better than doing conversions for all columns and discarding as we were doing earlier.
Since we can't trust the file offset from the row group, we don't have any alternative for getting this info more cheaply.

lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
```diff
@@ -195,28 +195,24 @@ public static List<RowGroupInfo> getFilteredRowGroups(
     {
         ImmutableList.Builder<RowGroupInfo> rowGroupInfoBuilder = ImmutableList.builder();
         for (BlockMetadata block : parquetMetadata.getBlocks(splitStart, splitLength)) {
-            long blockStart = block.getStartingPos();
-            boolean splitContainsBlock = splitStart <= blockStart && blockStart < splitStart + splitLength;
```
chenjian2664, line 199: Why remove this? Doesn’t it uses "right" position?

raunaqmorarka: Its redundant now because its relying on the same offset as `getBlocks` internally.
Earlier this was needed for the case where the rowgroup file offset was not set.

chenjian2664 (approved):

Issue #26007, opened 2025-06-13
Improve unspilling concurrency


sopel39:
Currently join operator (`io.trino.operator.join.HashBuilderOperator`) is unspilling partition by partition using `io.trino.spiller.FileSingleStreamSpiller`. However, `io.trino.spiller.FileSingleStreamSpiller` is single threaded. In case of `HashBuilderOperator` it should be unspilling each partition in parallel.

sopel39:
cc @osscm

PR #26076, opened 2025-06-26
Improve HashBuilderOperator unspill parallelism

sopel39:
HashBuilderOperator is unspilling sequentially partition by partition. This commit improves join unspilling performance by making FileSingleStreamSpiller unspill single partition in parallel. FileSingleStreamSpiller is enhanced so it can spill to/unspill from multiple files.

This improves unspilling time for join from minutes to seconds on larger machines

## Release notes

```markdown
## General
* Improve performance of spilling join queries. ({issue}`issuenumber`)
```

Fixes https://github.com/trinodb/trino/issues/26007

### DIFF
--- a/core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
+++ b/core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
@@ -16,6 +16,7 @@ package io.trino.spiller;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.AbstractIterator;
 import com.google.common.collect.ImmutableList;
+import com.google.common.collect.Iterators;
 import com.google.common.io.Closer;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
@@ -33,6 +34,7 @@ import io.trino.memory.context.LocalMemoryContext;
 import io.trino.operator.SpillContext;
 import io.trino.spi.Page;
 import io.trino.spi.TrinoException;
+import jakarta.annotation.Nullable;
 
 import javax.crypto.SecretKey;
 
@@ -42,6 +44,7 @@ import java.io.InputStream;
 import java.io.UncheckedIOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
@@ -62,8 +65,14 @@ public class FileSingleStreamSpiller
 {
     @VisibleForTesting
     static final int BUFFER_SIZE = 4 * 1024;
+    static final int DEFAULT_SEGMENT_SIZE = 64 * 1024 * 1024; // 64 MB
 
-    private final FileHolder targetFile;
+    private final int segmentSize;
+
+    private final List<FileHolder> targetFiles;
+    private volatile long currentSegmentBytes;
+    private volatile int currentFileIndex;
+
     private final Closer closer = Closer.create();
     private final PagesSerdeFactory serdeFactory;
     private volatile Optional<SecretKey> encryptionKey;
@@ -84,11 +93,35 @@ public class FileSingleStreamSpiller
             PagesSerdeFactory serdeFactory,
             Optional<SecretKey> encryptionKey,
             ListeningExecutorService executor,
-            Path spillPath,
+            List<Path> spillPaths,
             SpillerStats spillerStats,
             SpillContext spillContext,
             LocalMemoryContext memoryContext,
             Runnable fileSystemErrorHandler)
+    {
+        this(
+                serdeFactory,
+                encryptionKey,
+                executor,
+                spillPaths,
+                spillerStats,
+                spillContext,
+                memoryContext,
+                fileSystemErrorHandler,
+                DEFAULT_SEGMENT_SIZE);
+    }
+
+    @VisibleForTesting
+    public FileSingleStreamSpiller(
+            PagesSerdeFactory serdeFactory,
+            Optional<SecretKey> encryptionKey,
+            ListeningExecutorService executor,
+            List<Path> spillPaths,
+            SpillerStats spillerStats,
+            SpillContext spillContext,
+            LocalMemoryContext memoryContext,
+            Runnable fileSystemErrorHandler,
+            int segmentSize)
     {
         this.serdeFactory = requireNonNull(serdeFactory, "serdeFactory is null");
         this.encryptionKey = requireNonNull(encryptionKey, "encryptionKey is null");
@@ -109,8 +142,15 @@ public class FileSingleStreamSpiller
         // middle of execution when close() is called (note that this applies to both readPages() and writePages() methods).
         this.memoryContext.setBytes(BUFFER_SIZE);
         this.fileSystemErrorHandler = requireNonNull(fileSystemErrorHandler, "filesystemErrorHandler is null");
+        this.segmentSize = segmentSize;
+        requireNonNull(spillPaths, "spillPaths is null");
+        checkState(!spillPaths.isEmpty(), "spillPaths is empty");
         try {
-            this.targetFile = closer.register(new FileHolder(Files.createTempFile(spillPath, SPILL_FILE_PREFIX, SPILL_FILE_SUFFIX)));
+            ImmutableList.Builder<FileHolder> builder = ImmutableList.builder();
+            for (Path path : spillPaths) {
+                builder.add(closer.register(new FileHolder(Files.createTempFile(path, SPILL_FILE_PREFIX, SPILL_FILE_SUFFIX))));
+            }
+            this.targetFiles = builder.build();
         }
         catch (IOException e) {
             this.fileSystemErrorHandler.run();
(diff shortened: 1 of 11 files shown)


sopel39:
cc @osscm

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpillerFactory.java
```diff
@@ -165,14 +170,19 @@ private static void cleanupOldSpillFiles(Path path)
     }
 
     @Override
-    public SingleStreamSpiller create(List<Type> types, SpillContext spillContext, LocalMemoryContext memoryContext)
+    public SingleStreamSpiller create(List<Type> types, SpillContext spillContext, LocalMemoryContext memoryContext, boolean parallelSpill)
     {
         Optional<SecretKey> encryptionKey = spillEncryptionEnabled ? Optional.of(createRandomAesEncryptionKey()) : Optional.empty();
+        ImmutableList.Builder<Path> paths = ImmutableList.builder();
```
pettyjamesm: Nit: this could be `ImmutableList.builderWithExpectedSize()` if moved below `int spillFileCount = ...`

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpillerFactory.java
```diff
@@ -93,6 +95,7 @@ public FileSingleStreamSpillerFactory(BlockEncodingSerde blockEncodingSerde, Spi
                 requireNonNull(blockEncodingSerde, "blockEncodingSerde is null"),
                 spillerStats,
                 featuresConfig.getSpillerSpillPaths(),
+                Math.min(featuresConfig.getSpillerThreads(), taskManagerConfig.getTaskConcurrency()),
```
pettyjamesm, line 98: Seems like we'll want to change the default value for `FeaturesConfig#spillerThreads` since this change will make spilling much more thread hungry.

sopel39: it's already changed to multiplication of spill paths: https://github.com/trinodb/trino/pull/25892/commits/5e7bd69ddf90fcf2257df78e19e032838595379c. Join unspilling works stepwise, so peak unspilling parallelism should remain the same

sopel39:
> @sopel39 - I'm not sure I understand the purpose of segmenting the individual files at 64MB boundaries

Currently, when join hits spilling, it's parallelism will be reduced to single IO stream (build side unspilling) and single CPU thread (reconstructing of lookup source). This PR addresses the first issue (single IO stream). It was observed, that this PR reduces IO part of join unspilling from minutes to seconds.

> Segmenting the files is adding a good bit of complexity to the implementation and it seems unnecessary on a first read through.

Segments are deterministic. When reading segments back it's possible to restore original pages order without any extra index. This allows to read back segments in parallel, which is the purpose of this PR.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +177,176 @@ public long getSpilledPagesInMemorySize()
...
+            }
+
+            int fileIndex = 0;
+            while (true) {
+                boolean found = false;
+                for (int i = 0; i < fileCount; i++) {
+                    int index = (fileIndex + i) % fileCount;
+                    Iterator<List<Page>> segments = iterators.get(index);
+                    if (segments.hasNext()) {
+                        segments.next().forEach(builder::add);
```
pettyjamesm: Probably: `builder.addAll(segments.next())` to allow bulk expansion

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -109,8 +142,15 @@ public FileSingleStreamSpiller(
         // middle of execution when close() is called (note that this applies to both readPages() and writePages() methods).
         this.memoryContext.setBytes(BUFFER_SIZE);
         this.fileSystemErrorHandler = requireNonNull(fileSystemErrorHandler, "filesystemErrorHandler is null");
+        this.segmentSize = segmentSize;
+        requireNonNull(spillPaths, "spillPaths is null");
+        checkState(!spillPaths.isEmpty(), "spillPaths is empty");
```
pettyjamesm: `checkArgument` seems more appropriate here

pettyjamesm (commented):
This still seems massively more complicated than necessary. Wouldn't it be simpler to cut files at the segment boundary and support "pipelined" parallelism instead (i.e.: allow concurrent read-back of up to n files)? It seems like that would give you more control over the read back concurrency and simplify the logic significantly.

sopel39:
> Wouldn't it be simpler to cut files at the segment boundary and support "pipelined" parallelism instead (i.e.: allow concurrent read-back of up to n files)? 

You mean create a new file for each segment? It would create a massive number of files that have to be tracked. Not ideal if you want to keep memory and resources constrained. Also, it wouldn't necessarily make it simpler as you still have to keep track of segment size, but now you also have to keep track of potentially hundreds of thousands of spill files.

dain:
I think this can be simplified using the approach James suggested.  For the spill, you can create a directory (in each target) for the spill and then number the files sequentially.  Each file is 64MB and you form a total order. If file count becomes a problem you could scale up the file size as more files are created, but I don't expect this to be a problem on local disks.

The memory requirement is just the directory name and the max sequence number.  This will also make cleanup on local disks trivial as you can just delete the directories.  You could also parallel delete, but I don't expect that to be a problem on local disks.

I think this might also makes unspill easier to auto tune since you can have parallel loaders and simply monitor the ordered output queue of the pages.  When the queue if not empty, you can reduce readers... you would also want to reduce readers if the time spent waiting for reads is high... anyway a future optimization.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
sopel39: That is a good idea

sopel39:
@dain @pettyjamesm there are multiple ways to tackle this issue as we can see. I don't see one or the other dramatically simpler or harder. I think we all understand the approach I've proposed, so clearly it's not that complex. Unless this approach is fundamentally broken, I would rather keep it. BTW: you are mostly arguing with AI generated code. None of the lines here were written by me, but the concept and review of the final code come from me.

pettyjamesm:
> @dain @pettyjamesm there are multiple ways to tackle this issue as we can see. I don't see one or the other dramatically simpler or harder.

I don't agree that the approaches are fundamentally equivalent in terms of complexity. Tracking the current byte offset within a fixed number of open iterators and switching between them during read-back (while maintaining parallelism) is definitely harder to reason about compared to "reading fully" from N files with each input file being the unit of parallelism. As noted in our offline conversation, there are other upsides to using the file level approach like the ability to delete spill data from disk incrementally and the ability to tune the level of parallelism higher 1 thread per spill path.

sopel39:
> Tracking the current byte offset

It's tracking page size really. Something that has to be done in any approach during spill. In unspill, you suggest to replace page size tracking with tracking of multiple spill files in spill directories (new concept in spill!). I would argue it's even more complex with possible side-effects from OS (what iffs?)

> within a fixed number of open iterators and switching between

iterators won't go away independently of approach. In dir approach they would be nested too (per dir, per file). No complexity saved.

> there are other upsides to using the file level approach like the ability to delete spill data from disk incrementally

It's a marginal difference and not really a contract of `SingleStreamSpiller`. For example, improved join implementation could unspill same partition multiple times (e.g. when unspilling partitions before probe finishing).

> the ability to tune the level of parallelism higher 1 thread per spill path.

We would rather like to reduce parallelism sometimes, which is possible with this solution.

Overall, I still don't see any fundamental flaw in this approach and refuse to refactor PRs (or redo them completely) based on taste :)

dain:
> There are multiple ways to tackle this issue as we can see. I don't see one or the other dramatically simpler or harder. I think we all understand the approach I've proposed, so clearly it's not that complex. Unless this approach is fundamentally broken, I would rather keep it. 

I find the approach quite complex and I think this can and should be done in a much simpler way.  For the spilling code, there are very few maintainers, and we by necessity need to keep this as simple as possible.  Therefore, I do not think we should merge this code, and we should focus on the simpler design.

sopel39:
> I find the approach quite complex and I think this can and should be done in a much simpler way.

I spend last few posts explaining why it's not more complex than other approaches. I think the only alternative here is not to land this or any subsequent PRs that improve spill and join performance. That is also an option. I could also add a feature toggle.

> For the spilling code, there are very few maintainers

In general there are very few active OSS maintainers left. Does that mean that the project should stall? That is an option too.

sopel39:
@pettyjamesm PTAL. I've simplified the approach and used RR to distribute pages across files.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +148,129 @@ public long getSpilledPagesInMemorySize()
...
+            return new AbstractIterator<>()
+            {
+                int fileIndex;
+
+                @Override
+                protected Page computeNext()
+                {
+                    Iterator<Page> iterator = iterators.get(fileIndex);
+                    if (!iterator.hasNext()) {
+                        return endOfData();
```
pettyjamesm, line 170: This looks like it returns `endOfData()` after the first iterator is exhausted, rather than after all iterators are exhausted.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +148,129 @@ public long getSpilledPagesInMemorySize()
...
+                }
+                builder.add(iterator.next());
+                fileIndex = (fileIndex + 1) % fileCount;
+            }
+            return builder.build();
+        }, executor);
     }
 
-    private DataSize writePages(Iterator<Page> pageIterator)
+    private DataSize writePages(Iterator<Page> pages)
```
pettyjamesm, line 233: Does this preserve the performance benefit? Looks like we're still writing out all pages as a single threaded stream over N output streams, but I would think that you would want N threads _and_ N streams to flush to disk as fast as possible (possibly by splitting the `PagesIndex` into sub-ranges and flushing those in parallel.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +148,129 @@ public long getSpilledPagesInMemorySize()
...
+        int fileCount = targetFiles.size();
+
+        try {
+            while (pages.hasNext()) {
+                Page page = pages.next();
                 long pageSizeInBytes = page.getSizeInBytes();
+                Slice serialized = serializer.serialize(page);
+                long serializedPageSize = serialized.length();
+
+                try (SliceOutput out = newSliceOutput(fileIndex)) {
```
pettyjamesm: If this stays single threaded, shouldn’t we prefer to open the output stream only once per file?

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
sopel39: It's fine as we always write pages to files in sequence. Therefore all following iterators will be finished too. However, I've added explicit checkAllIteratorsExhausted too

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
sopel39: Write parallelism is not critical. There are multiple `HashBuilderOperators` instances, each responsible for it's own partition. You will get spill parallelism, when all `HashBuilderOperators` are spilling. We can always improve it later. However, it's a bit niche case as usually all or most `HashBuilderOperators` will be revoking.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
sopel39: Good idea. Changed

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -107,10 +105,16 @@ public FileSingleStreamSpiller(
         // This means we start accounting for the memory before the spiller thread allocates it, and we release the memory reservation
         // before/after the spiller thread allocates that memory -- -- whether before or after depends on whether writePages() is in the
         // middle of execution when close() is called (note that this applies to both readPages() and writePages() methods).
-        this.memoryContext.setBytes(BUFFER_SIZE);
+        this.memoryContext.setBytes((long) SpillFile.BUFFER_SIZE * spillPaths.size());
```
pettyjamesm, line 111: Nit: we should probably `checkArgument` / `requireNonNull` the arguments before setting the memory usage.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +141,140 @@ public long getSpilledPagesInMemorySize()
...
-                long pageSize = serializedPage.length();
-                localSpillContext.updateBytes(pageSize);
-                spillerStats.addToTotalSpilledBytes(pageSize);
-                output.writeBytes(serializedPage);
+                localSpillContext.updateBytes(serializedPageSize);
+                spillerStats.addToTotalSpilledBytes(serializedPageSize);
+
+                fileIndex = (fileIndex + 1) % fileCount;
             }
+
```
pettyjamesm, line 262: Seems like there should be `spillFiles.forEach(SpillFile::closeOutput);` here to flush the buffered data to disk and release the temporary buffers associated with each file.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +141,140 @@ public long getSpilledPagesInMemorySize()
...
+                try (Closer closer = Closer.create()) {
+                    readFilePages(deserializer, file, closer).forEachRemaining(pages::add);
+                }
+                return pages.build();
+            }));
+        }
+
+        // Combine pages from all spill files according to the round-robin order.
+        return Futures.transform(Futures.allAsList(futures), pagesPerFile -> {
+            ImmutableList.Builder<Page> builder = ImmutableList.builder();
```
pettyjamesm: Since we have `List<List<Page>> pagesPerFile` we can compute the total page count cheaply and avoid growing the builder incrementally.

core/trino-main/src/main/java/io/trino/spiller/FileSingleStreamSpiller.java
```diff
@@ -137,61 +141,140 @@ public long getSpilledPagesInMemorySize()
...
+            int fileCount = spillFiles.size();
+
+            List<Iterator<Page>> iterators = new ArrayList<>(fileCount);
+            for (List<Page> pages : pagesPerFile) {
+                iterators.add(pages.iterator());
+            }
+
+            int fileIndex = 0;
+            while (true) {
```
pettyjamesm: Maybe slightly cleaner:
```
while (true) {
  for (int fileIndex = 0; fileIndex < iterators.size(); fileIndex++) {
    Iterator<Page> iterator = iterators.get(fileIndex);
    if (!iterator.hasNext()) {
      checkAllIteratorsExhausted(iterators);
      return builder.build();
    }
    builder.add(iterator.next());
}
```

sopel39: actually, the easiest is
```
            int fileIndex = 0;
            while (iterators.get(fileIndex).hasNext()) {
                builder.add(iterators.get(fileIndex).next());
                fileIndex = (fileIndex + 1) % fileCount;
            }
            checkAllIteratorsExhausted(iterators);
```

sopel39:
@pettyjamesm thanks for review

core/trino-main/src/main/java/io/trino/spiller/SingleStreamSpillerFactory.java
```diff
@@ -21,11 +21,16 @@
 
 public interface SingleStreamSpillerFactory
 {
-    SingleStreamSpiller create(List<Type> types, SpillContext spillContext, LocalMemoryContext memoryContext);
+    default SingleStreamSpiller create(List<Type> types, SpillContext spillContext, LocalMemoryContext memoryContext)
+    {
+        return create(types, spillContext, memoryContext, false);
+    }
+
+    SingleStreamSpiller create(List<Type> types, SpillContext spillContext, LocalMemoryContext memoryContext, boolean parallelSpill);
```
findepi, line 29: `SingleStreamSpiller` maintains data order of data being spilled and read back (FIFO)

is it still the case when it's created as parallelSpill?

if yes -- we should deprecate the old method and use the new one 
if no - ``SingleStreamSpiller`` interface needs update

PR #26085, opened 2025-06-27
Close filesystem streams

wendigo:
## Release notes

```markdown
## Delta lake
* Release native filesystem resources/prevent leaks? ({issue}`issuenumber`)
```


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMergeSink.java
```diff
@@ -709,6 +709,9 @@ public void abort()
         if (cdfPageSink != null) {
             cdfPageSink.abort();
         }
+        if (insertPageSink != null) {
```
chenjian2664: nit: `insertPageSink` never null

chenjian2664 (approved):

losipiuk (approved):

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMergeSink.java
```diff
@@ -709,6 +709,8 @@ public void abort()
         if (cdfPageSink != null) {
             cdfPageSink.abort();
         }
+
+        insertPageSink.abort();
```
homar, line 713: I don't understand why

homar (approved):

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMergeSink.java
wendigo: This leaves parquet writers and associated output streams open

PR #26087, opened 2025-06-27
Add i/o stream leak detection to the native FS

wendigo:
It's enabled by default in the CI

## Release notes

```markdown
## Section
```


raunaqmorarka (approved):
Nice idea

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemModule.java
```diff
@@ -162,6 +165,13 @@ static TrinoFileSystemFactory createFileSystemFactory(
 
         TrinoFileSystemFactory delegate = new SwitchingFileSystemFactory(loader);
         delegate = new TracingFileSystemFactory(tracer, delegate);
+
+        // Enable leak detection if configured or if running in a CI environment
+        boolean trackingDetectionEnabled = getenv("CONTINUOUS_INTEGRATION") != null;
```
electrum, line 170: You can make this the default for the config, allowing it to be overridden in either case.

wendigo: Will follow up

PR #26090, opened 2025-06-27
iceberg: add printing underlying failure cause for page sources

psarna:
## Description

Current implementation hides the details of the underlying exception, and that includes decompression failures.
As a result, Trino will get a "Could not decode the dictionary for X" message, with the actual cause -- e.g. decompression failure -- not being mentioned at all.

With the change applied, hopefully users can see more detailed failures, e.g. that their file is corrupted or that there was a compression algorithm mismatch.

## Release notes

```markdown
## Section
* Error messages when creating page sources now include underlying error cause if present
```


ebyhr:
Could you share stacktraces before & after this change?

PR #26115, opened 2025-07-01
Support ClientId Param in iceberg catalog OAuth2 credentials flow

norton120:
## Description
Iceberg's `OAuth2Util` supports passing a `ClientId` param for client credentials flow - which [Microsoft Entra needs to work](https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-client-creds-grant-flow#first-case-access-token-request-with-a-shared-secret) via `setClientId()`. This flow is ideal for Trino => Catalog auth (machine to machine), but Trino does not expose the `ClientId` property in configs for iceberg.

This PR just adds the optional `client-id` config to an iceberg catalog so that Microsoft Entra (and any other IDP that needs `ClientId`) can work for catalog auth.  

See the underlying [Apache Iceberg code](https://github.com/apache/iceberg/blob/e73344323ddc26f9f45db15d50394e25069821d2/core/src/main/java/org/apache/iceberg/rest/auth/OAuth2Util.java#L80) for details. 

## Release notes

### DIFF
--- a/docs/src/main/sphinx/object-storage/metastores.md
+++ b/docs/src/main/sphinx/object-storage/metastores.md
@@ -504,6 +504,9 @@ following properties:
 * - `iceberg.rest-catalog.oauth2.scope`
   - Scope to be used when communicating with the REST Catalog. Applicable only
     when using `credential`.
+* - `iceberg.rest-catalog.oauth2.client-id`
+  - OAuth2 client ID. This parameter is required by some OAuth2 providers such as
+    Microsoft Entra (Azure Active Directory).
 * - `iceberg.rest-catalog.oauth2.server-uri`
   - The endpoint to retrieve access token from OAuth2 Server.
 * - `iceberg.rest-catalog.oauth2.token-refresh-enabled`
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/OAuth2SecurityConfig.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/OAuth2SecurityConfig.java
@@ -28,6 +28,7 @@ public class OAuth2SecurityConfig
     private String scope;
     private String token;
     private URI serverUri;
+    private String clientId;
     private boolean tokenRefreshEnabled = OAuth2Properties.TOKEN_REFRESH_ENABLED_DEFAULT;
 
     public Optional<String> getCredential()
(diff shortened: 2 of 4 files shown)


norton120:
CLA signed ✍🏻 😎

norton120:
I'm not sure how to fix [this CI run](https://github.com/trinodb/trino/actions/runs/16005581487/job/45151378613?pr=26115) as it references an older failing commit?

wendigo:
@norton120 you need to squash two commits together. commits-check verify that every individual commit compiles fine.

norton120:
This is passing locally but the client connection still isn't working in practice, I want to make sure it has appropriate coverage first

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/OAuth2SecurityConfig.java
```diff
@@ -97,6 +98,19 @@ public OAuth2SecurityConfig setTokenRefreshEnabled(boolean tokenRefreshEnabled)
...
+    }
+
+    @Config("iceberg.rest-catalog.oauth2.client-id")
+    @ConfigDescription("OAuth2 client ID")
+    public OAuth2SecurityConfig setClientId(String clientId)
+    {
+        this.clientId = clientId;
+        return this;
+    }
+
```
ebyhr, line 113: Move these methods above `isTokenRefreshEnabled()` method for consistency with fields.

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -504,6 +504,9 @@ following properties:
 * - `iceberg.rest-catalog.oauth2.scope`
   - Scope to be used when communicating with the REST Catalog. Applicable only
     when using `credential`.
+* - `iceberg.rest-catalog.oauth2.client-id`
```
ebyhr, line 507: Do we really need a new property? `iceberg.rest-catalog.oauth2.credential`'s format is `client-id:secret`.

ebyhr (commented):
The commit title is too long and it must be capitalized. https://trino.io/development/process.html#pull-request-and-commit-guidelines-

docs/src/main/sphinx/object-storage/metastores.md
norton120: Fair point - the spec allows either `client_id` and `client_secret` to be distinct props in the request body or combined in an auth header - so even if Entra requires the former, it probably doesn't make sense for Trino to support both. Closing this PR

PR #26118, opened 2025-07-02
Code clean ups in Memory Connector

chenjian2664:
## Release notes

```markdown
## Section
```


plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemorySplitManager.java
```diff
@@ -98,16 +98,13 @@ private static CompletableFuture<?> whenCompleted(DynamicFilter dynamicFilter)
         return NOT_BLOCKED;
     }
 
-    private static class DelayedSplitSource
+    private record DelayedSplitSource(CompletableFuture<?> delay, ConnectorSplitSource delegate)
```
ebyhr: We usually don't use record for non-model classes.

ebyhr (approved):

PR #26119, opened 2025-07-02
Add new simple announce node manager and DNS node manager

dain:
## Description

This PR adds two new node inventory systems: Announce and DNS.  Announce is the new default and replaces Airlift Discovery. The system used is determined by the `discovery.type` property which currently can be `announce`, `dns` or `airlift-discovery`.

Announce inventory is a trivial announcement service for nodes. Nodes only announce their internal URI every 5 seconds, as opposed to a full discovery announcement.  The announcement server simply keeps a list of known workers that times out after 30 seconds.  The existing internal node manager code handles the rest, reaching out to the node to get state and any other critical information. The announcer client uses the same `discovery.uri` property name as before, so no configuration changes are needed.  The new announcer supports multiple names so announcements can be sent to multiple coordinators.

Secondly, the PR adds a DNS based node manager.  Instead of announcement, this relies on K8s style DNS infrastructure.  In this setup it is expect that there are one or more hostnames that will provide all IP addresses for the workers in the cluster.  It is assumed that all nodes are bound to the same port and mixed HTTP / HTTPS is not allowed (all nodes must be configured the same way).  The hostnames are configured with the `cluster.hosts` property.  DNS does not have an announcement system, so no configuration is needed on workers.

## Release notes

```markdown
## Section
* Add Announce node inventory to replace Airlift Discovery.
* Add DNS node inventory for use in K8s like environments that provide a DNS name for all workers.
* Add `discovery.type` to set node inventory system. The value can be `announce`, `dns` or `airlift-discovery`.
```


core/trino-main/src/test/java/io/trino/node/TestNodeInventoryConfig.java
```diff
@@ -0,0 +1,46 @@
...
+    void testDefaults()
+    {
+        assertRecordedDefaults(recordDefaults(NodeInventoryConfig.class)
+                .setType(ANNOUNCE));
+    }
+
+    @Test
+    void testExplicitPropertyMappings()
+    {
+        Map<String, String> properties = ImmutableMap.of("discovery.type", "announce");
```
electrum: Typo, `announce` should be `airlift-discovery`

electrum: Nit: use builder syntax
```java
Map<String, String> properties = ImmutableMap.<String, String>builder()
        .put("discovery.type", "airlift-discovery")
        .buildOrThrow();

File: core/trino-main/src/main/java/io/trino/node/AnnounceConfig.java [hunk 2]
```diff
@@ -0,0 +1,39 @@
...
+import jakarta.validation.constraints.NotNull;
+
+import java.net.URI;
+import java.util.List;
+
+public class AnnounceConfig
+{
+    private List<URI> coordinatorUris = List.of();
+
+    @NotNull
```
electrum, line 27: Should this be `@NotEmpty`? Is there a purpose to configure `discovery.type=announce` with an empty list?

core/trino-main/src/main/java/io/trino/node/AnnounceConfig.java
```diff
@@ -0,0 +1,39 @@
...
+    public List<@NotNull URI> getCoordinatorUris()
```
electrum, line 28: Does using `@NotNull` here do anything? Can the config system create a list with null elements?

core/trino-main/src/main/java/io/trino/node/AnnounceNodeInventory.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.Set;
+
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+public class AnnounceNodeInventory
+        implements NodeInventory
+{
+    private final Cache<URI, Long> nodes;
+
+    @SuppressModernizer
```
electrum: Use `buildNonEvictableCache()` instead

core/trino-main/src/main/java/io/trino/node/AnnounceNodeInventory.java
```diff
@@ -0,0 +1,49 @@
...
+    public AnnounceNodeInventory()
+    {
+        nodes = CacheBuilder.newBuilder()
+                .expireAfterWrite(30, SECONDS)
+                .build();
+    }
+
+    public void announce(URI node)
+    {
+        nodes.put(node, System.nanoTime());
```
electrum: Is the time here for debugging? The code could be simplified as

```java
private final Set<URI> nodes = newSetFromMap(
        buildNonEvictableCache(CacheBuilder.newBuilder()
                .expireAfterWrite(30, SECONDS)
                .build()).asMap());

public void announce(URI node) { nodes.put(node); }

public Set<URI> getNodes() { return ImmutableSet.copyOf(nodes); }

File: core/trino-main/src/main/java/io/trino/node/AnnounceNodeInventory.java [hunk 6]
```diff
@@ -0,0 +1,49 @@
...
+
+public class AnnounceNodeInventory
+        implements NodeInventory
+{
+    private final Cache<URI, Long> nodes;
+
+    @SuppressModernizer
+    public AnnounceNodeInventory()
+    {
+        nodes = CacheBuilder.newBuilder()
```
electrum: Can move to the initializer

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,182 @@
...
+    {
+        this(currentNode.getInternalUri(), config.getCoordinatorUris(), httpClient);
+    }
+
+    @VisibleForTesting
+    public AnnounceNodeAnnouncer(URI internalUri, Collection<URI> coordinatorUris, HttpClient httpClient)
+    {
+        URI currentUri = internalUri;
+        this.announceUris = coordinatorUris.stream()
+                .filter(uri -> !currentUri.equals(uri))
```
electrum: Nit: can do
```java
.filter(not(currentUri::equals))

File: core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java [hunk 8]
```diff
@@ -0,0 +1,182 @@
...
+                .map(uri -> uriBuilderFrom(uri).appendPath("/v1/announce").build())
+                .distinct()
+                .toList();
+        this.httpClient = requireNonNull(httpClient, "httpClient is null");
+        this.announceExecutor = newSingleThreadScheduledExecutor(daemonThreadsNamed("node-state-poller-%s"));
```
electrum: Can move to initializer

electrum: No need for `-%s` with `newSingleThreadScheduledExecutor` since there is only one thread

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,182 @@
...
+        this.currentHostAnnouncement = createStaticBodyGenerator(currentUri.toString(), UTF_8);
+    }
+
+    @Override
+    public void start()
+    {
+        if (announceUris.isEmpty()) {
+            log.warn("No coordinator URIs configured, skipping node state announcements");
```
electrum: This will log a warning on the coordinator

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,182 @@
...
+
+        announceExecutor.scheduleWithFixedDelay(
+                () -> {
+                    if (state.get() != State.STARTED) {
+                        return;
+                    }
+                    try {
+                        forceAnnounce();
+                    }
+                    catch (Exception e) {
```
electrum: This should be `RuntimeException` since no checked exceptions are thrown

Or use `Throwable` if we don't want to de-schedule on failure

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,182 @@
...
+            Thread.currentThread().interrupt();
+        }
+    }
+
+    @FormatMethod
+    private void logWarning(String format, Object... args)
+    {
+        // log at most once per second per node
+        long now = System.nanoTime();
+        if (now - lastWarningLogged.get() >= 1_000_000_000) {
```
electrum: `SECOND.toNanos(1)`

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,182 @@
...
+
+        checkState(!announceExecutor.isShutdown(), "Announcer has been destroyed");
+        if (!state.compareAndSet(State.NOT_STARTED, State.STARTED)) {
+            log.warn("Announcer is already started, skipping redundant start");
+            return;
+        }
+
+        announceExecutor.scheduleWithFixedDelay(
+                () -> {
+                    if (state.get() != State.STARTED) {
```
electrum: Why is this needed? At this point, it can't be in `NOT_STARTED`, and if it was `STOPPED`, then this can only happen in a tiny window (and doesn't seem to matter)

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,182 @@
...
+                    public void onFailure(Throwable t)
+                    {
+                        logWarning("Error announcing node state to %s: %s", announceUri, t.getMessage());
+                    }
+                }, directExecutor());
+        return Futures.nonCancellationPropagating(responseFuture);
+    }
+
+    @Override
+    public void stop()
```
electrum, line 153: This method is never called. We should probably remove it and just make all of these have `@PreDestroy`

core/trino-main/src/main/java/io/trino/node/AnnounceNodeInventoryModule.java
```diff
@@ -0,0 +1,50 @@
...
+    {
+        boolean coordinator = buildConfigObject(ServerConfig.class).isCoordinator();
+        if (coordinator) {
+            jaxrsBinder(binder).bind(AnnounceNodeResource.class);
+
+            binder.bind(AnnounceNodeInventory.class).in(Scopes.SINGLETON);
+            binder.bind(NodeInventory.class).to(AnnounceNodeInventory.class).in(Scopes.SINGLETON);
+        }
+
+        // both coordinator and worker must announce
```
electrum: This is a bit misleading since coordinators don't announce to themselves

core/trino-main/src/main/java/io/trino/node/CoordinatorNodeManager.java
```diff
@@ -137,10 +140,11 @@ public void stop()
     }
 
     @Override
-    public void refreshNodes()
+    public boolean refreshNodes(boolean forceAndWait)
     {
         // Add new nodes
-        for (URI uri : nodeInventory.getNodes()) {
+        Set<URI> nodes = nodeInventory.getNodes();
```
electrum: Is this just for debugging?

core/trino-main/src/main/java/io/trino/node/NodeManagerModule.java
```diff
@@ -51,6 +54,14 @@ protected void setup(Binder binder)
             binder.bind(InternalNodeManager.class).to(WorkerInternalNodeManager.class).in(Scopes.SINGLETON);
         }
 
-        install(new AirliftNodeInventoryModule(nodeVersion));
+        install(conditionalModule(
```
electrum: This can use the new `SwitchModule`
```java
install(switchModule(
        NodeInventoryConfig.class,
        NodeInventoryConfig::getType,
        type -> switch (type) {
            case AIRLIFT_DISCOVERY -> new AirliftNodeInventoryModule(nodeVersion);
            case ANNOUNCE -> new AnnounceNodeInventoryModule();
        }));

File: core/trino-main/src/test/java/io/trino/node/TestAnnounceConfig.java [hunk 17]
```diff
@@ -0,0 +1,46 @@
...
+    void testDefaults()
+    {
+        assertRecordedDefaults(recordDefaults(AnnounceConfig.class)
+                .setCoordinatorUris(List.of()));
+    }
+
+    @Test
+    void testExplicitPropertyMappings()
+    {
+        Map<String, String> properties = ImmutableMap.of("discovery.uri", "https://example.com:100,https://example.org:200");
```
electrum: Nit: use builder style

core/trino-main/src/test/java/io/trino/node/TestAnnounceNodeInventory.java
```diff
@@ -0,0 +1,186 @@
...
+import java.io.Closeable;
+import java.net.URI;
+import java.util.List;
+import java.util.Set;
+
+import static io.airlift.concurrent.MoreFutures.getFutureValue;
+import static io.airlift.configuration.ConfigBinder.configBinder;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
```
electrum: Static import

core/trino-main/src/test/java/io/trino/node/TestAnnounceNodeInventory.java
```diff
@@ -0,0 +1,186 @@
...
+class TestAnnounceNodeInventory
+{
+    private final HttpClient httpClient;
+
+    TestAnnounceNodeInventory()
+    {
+        httpClient = new JettyHttpClient();
```
electrum: Move to initializer

core/trino-main/src/main/java/io/trino/node/DnsNodeInventory.java
```diff
@@ -0,0 +1,95 @@
...
+                DnsNodeInventory::getAllByName);
+    }
+
+    @VisibleForTesting
+    DnsNodeInventory(boolean https, int port, Set<String> hosts, HostResolver hostResolver)
+    {
+        this.https = https;
+        this.port = port;
+        checkArgument(port > 0 && port < 65536, "invalid port: %s", port);
+        this.hosts = requireNonNull(hosts, "hosts is null")
```
electrum: No need for the check as the `.stream()` will throw

core/trino-main/src/main/java/io/trino/node/DnsNodeInventory.java
```diff
@@ -0,0 +1,95 @@
...
+                .stream()
+                .filter(not(Strings::isNullOrEmpty))
```
electrum, line 58: Is it possible to get a null host from config?

core/trino-main/src/main/java/io/trino/node/DnsNodeInventory.java
```diff
@@ -0,0 +1,95 @@
...
+    public Set<URI> getNodes()
+    {
+        return hosts.stream()
+                .map(hostResolver::getAllByName)
+                .flatMap(Set::stream)
+                .map(address -> {
+                    try {
+                        return new URI(https ? "https" : "http", null, InetAddresses.toUriString(address), port, null, null, null);
+                    }
+                    catch (Exception e) {
```
electrum: `URISyntaxException`

electrum (approved):

core/trino-main/src/main/java/io/trino/node/AnnounceConfig.java
dain: Coordinator does not need an announcement unless there are multiple coordinators.

core/trino-main/src/main/java/io/trino/node/AnnounceConfig.java
dain: It means you cannot have a null element in the list.  It isn't required, but it is reasonable documentation

core/trino-main/src/main/java/io/trino/node/NodeManagerModule.java
dain: Nice! I did not know about that.

core/trino-main/src/main/java/io/trino/node/DnsNodeInventory.java
dain: I don't think so, but I'm being careful with announcement configs.

dain: The real reason is that the docs for `getAllByName` say that null or empty result in the loopback address being returned and I don't want to deal with that

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
dain: I marked this method and the airlift one as `@PreDestroy`.

core/trino-main/src/main/java/io/trino/node/AnnounceConfig.java
electrum: I understand what it does, but seems confusing because it should be impossible. We don't do that anywhere else, but I'm fine if you want to leave it.

core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
```diff
@@ -0,0 +1,174 @@
...
+                            logWarning("Failed to announce node state to %s: %s", announceUri, statusCode);
+                        }
+                    }
+
+                    @Override
+                    public void onFailure(Throwable t)
+                    {
+                        logWarning("Error announcing node state to %s: %s", announceUri, t.getMessage());
+                    }
+                }, directExecutor());
```
electrum, line 147: FYI, this could be
```java
Futures.addCallback(addSuccessCallback(responseFuture, () -> {
    ...
});
Futures.addCallback(addExceptionCallback(responseFuture, t ->
    logWarning("Error announcing node state to %s: %s", announceUri, t.getMessage()));

dain:
I put the followup suggestions in #26125

PR #26125, opened 2025-07-03
Minor cleanup for new node inventory implementations

dain:
## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
+++ b/core/trino-main/src/main/java/io/trino/node/AnnounceNodeAnnouncer.java
@@ -64,7 +64,7 @@ public class AnnounceNodeAnnouncer
     private final boolean coordinator;
 
     @Inject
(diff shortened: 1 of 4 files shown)


core/trino-main/src/main/java/io/trino/node/RemoteNodeState.java
```diff
@@ -185,26 +185,13 @@ else if (!serverInfo.nodeVersion().equals(expectedNodeVersion)) {
...
-                catching,
-                new FutureCallback<>()
-                {
-                    @Override
-                    public void onSuccess(Boolean result)
-                    {
-                        future.compareAndSet(catching, null);
-                        lastUpdateNanos.set(ticker.read());
-                    }
+        addSuccessCallback(catching, _ -> {
```
electrum, line 190: This misses the failure case?

electrum (approved):

core/trino-main/src/main/java/io/trino/node/RemoteNodeState.java
dain: The code above converts the future into a version that can't fail.  I could have made a mistake on that, so please check.

PR #26133, opened 2025-07-04
Remove TestRedshiftConnectorSmokeTest

ebyhr:
## Description

TestRedshiftConnectorTest should be sufficient.

## Release notes


plugin/trino-redshift/pom.xml
```diff
@@ -351,7 +350,6 @@
                                 <!-- locations of GitHub runners which can lead to increased client latency on the -->
                                 <!-- JDBC operations performed on the ephemeral AWS Redshift cluster.  -->
                                 <include>**/TestRedshiftCastPushdown.java</include>
-                                <include>**/TestRedshiftConnectorSmokeTest.java</include>
```
ebyhr, line 354: Ah, I was missing we don't run TestRedshiftConnectorTest on CI. Now I think we shouldn't remove TestRedshiftConnectorSmokeTest.

PR #26135, opened 2025-07-07
Support `DELETE`, `UPDATE`, `MERGE` for Memory connector

chenjian2664:
## Release notes

```markdown
## Memory
* Support sql `DELETE`, `UPDATE`, `MERGE` for Memory connector. ({issue}`issuenumber`)
```


ebyhr:
Could you rebase on master to resolve conflicts?

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -652,6 +677,60 @@ public boolean allowSplittingReadIntoMultipleSubQueries(ConnectorSession session
...
+        updateRowsOnHosts(tableHandle.id(), fragments);
+    }
+
+    @Override
+    public Optional<ConnectorPartitioningHandle> getUpdateLayout(ConnectorSession session, ConnectorTableHandle tableHandle)
+    {
+        return Optional.of(new MemoryPartitioningHandle(createNodeToBucketMap()));
+    }
+
+    private Map<String, Integer> createNodeToBucketMap()
```
losipiuk, line 741: This logic is repeated here and in MemoryNodePartitioningProvider.
This is error prone. Can you structure code so the partitioning rules are reused?

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -652,6 +677,60 @@ public boolean allowSplittingReadIntoMultipleSubQueries(ConnectorSession session
...
+    {
+        Set<Node> nodes = nodeManager.getRequiredWorkerNodes();
```
losipiuk: How will that work if number of workers in the cluster changes. Or if their addresses change.
I think it will start to break in unexpected way.
Can you make you make it failing "nicely". I do not mean it to work but it should faild with readable exception if set of cluster workers changes after memory connector is initially populated with data.

chenjian2664: Yeah, I believe that if the worker nodes change, it could fail because we might end up mapping the same address to an unexpected bucket. Currently, the memory connector doesn’t support failover for its in-memory storage.
What I can do(that I am able to figure out) is store the mapping between the tableHandle and its bucket here, and for subsequent requests, compare them — if there’s any difference, fail the operation. WDYT? I'll put the code here :)

PPL-TBSK:
@chenjian2664 is this PR dead please? or is it in your backlog? thanks

PR #26140, opened 2025-07-07
Use `RunLengthEncodedBlock` for null column appends

chenjian2664:
Replaces manual null appending with `RunLengthEncodedBlock` when adding columns with null values

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryPagesStore.java
+++ b/plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryPagesStore.java
@@ -19,7 +19,7 @@ import com.google.errorprone.annotations.concurrent.GuardedBy;
 import com.google.inject.Inject;
 import io.trino.spi.Page;
 import io.trino.spi.TrinoException;
-import io.trino.spi.block.BlockBuilder;
+import io.trino.spi.block.RunLengthEncodedBlock;
 import io.trino.spi.type.Type;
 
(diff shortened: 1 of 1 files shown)


wendigo (approved):

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryPagesStore.java
```diff
@@ -118,10 +117,7 @@ public synchronized List<Page> getPages(
             }
             // Append missing columns with null values. This situation happens when a new column is added without additional insert.
             for (int j = page.getChannelCount(); j < columnIndexes.length; j++) {
-                Type type = columnTypes.get(j);
-                BlockBuilder builder = type.createBlockBuilder(null, page.getPositionCount());
-                IntStream.range(0, page.getPositionCount()).forEach(_ -> builder.appendNull());
-                page = page.appendColumn(builder.build());
+                page = page.appendColumn(RunLengthEncodedBlock.create(columnTypes.get(i), null, page.getPositionCount()));
```
samarsar, line 120: Wanting to confirm is this the correct variable? Seems like `j` is iterating over column indexes while `i` is iterating over pages.

pettyjamesm: Good catch, fixed this in https://github.com/trinodb/trino/pull/26167

PR #26142, opened 2025-07-07
Recheck that operator is not waiting for memory in hash builder finish

sopel39:
Operator#finish might be called even if operator is blocked (e.g. waiting for memory). Therefore, memory available condition needs to be re-checked every time finish is called.

## Release notes

```markdown
## General
* Reduce number of worker OOMs when running join queries. ({issue}`issuenumber`)
```


sopel39:
cc @osscm

pettyjamesm (approved):
LGTM. Seems like the previous behavior could result in unnecessary OOMs? If so, this PR probably merits a release note mention.

sopel39:
> LGTM. Seems like the previous behavior could result in unnecessary OOMs? If so, this PR probably merits a release note mention.

It's not guaranteed you will hit OOM without this PR. I will suggest some RN

PR #26156, opened 2025-07-09
Update airbase to 277 and airlift to 339

wendigo:
This moves to unshaded guice dependency so that new ASM version can be picked up instead of the shaded one. This makes guice fully functional in the new JDKs.

## Release notes

```markdown
## Section
```


wendigo:
Previously shaded dependency definition was removed and dependency is now banned (`classes` classifier is allowed)

losipiuk (approved):

Issue #17272, opened 2023-04-27
Support copy-on-write mode for Iceberg write


jackye1995:
Today Iceberg writes only support merge-on-read mode. Copy-on-write mode is a frequent ask for users that want better file layout without the need to run compactions frequently.

Technically this could be achieved pretty easily. The CoW implementation is already available for Delta: 

https://github.com/trinodb/trino/blob/master/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMergeSink.java

findepi:
Yes, I think we should support copy-on-write mode for UPDATE and DELETE for case when we update/delete most of the file. Creating big delta files isn't helpful for anyone.
If we support this strategy, we should automatically choose between copy-on-write and merge-on-read within each `IcebergMergeSink` independently.

C-h-e-r-r-y:
@rdblue 

For now these settings can not be set during table creation :(

MotiPoyastro:
This feature is highly requested in my org,
Trino 443 + Iceberg on S3 parquets.
frequent merge into queries becomes a problem on a large scale.
currently dealing with post optimize/compaction for many small files created on a daily basis.
but this is not a desired flow for us.

xkrogen:
One concern called out by @dain ([on Slack](https://trinodb.slack.com/archives/CJ6UC075E/p1729114539093519?thread_ts=1727902704.898979&cid=CJ6UC075E)) that I will persist here:

> CoW has it’s own problems because it breaks the change feed, so you cannot do downstream delta-style transforms (e.g., incremental materialized views)

I don't think this should be a blocker to adding support -- but something to keep in mind e.g. when considering doing a MoR vs CoW decision _automatically_ vs based on explicit user input, since there is some difference in functionality supported.

hashhar:
With https://github.com/trinodb/trino/pull/24031 it should now be possible to specify the default operation mode via the table property as @rdblue mentioned above.

Is it still desirable to provide an "integrated" solution for this in Trino? for most use-cases I think the current default of MoR on Trino is good.

autumnust:
@jackye1995  At this moment on the Iceberg side, we cannot set table properties to control the trino behavior on how it encodes delete files. Meanwhile https://iceberg.apache.org/docs/1.5.2/configuration/#write-properties didn't specify things like `write.delete.mode` is engine specific which is a bit confusing / inaccurate. 

[...]

PR #26161, opened 2025-07-09
Support copy-on-write change for Iceberg Connector

SongChujun:
This PR adds support for COW support for Iceberg Connector.

For the user interface part, we respect iceberg specification https://iceberg.apache.org/docs/latest/configuration/#write-properties add add three table properties ```write_delete_mode```, ```write_update_mode```, ```write_merge_mode```, to control the write behavior of each specific table change operation, node that this is consistent with other popular engines like Spark and Dremio: https://docs.dremio.com/25.x/sonar/query-manage/data-formats/apache-iceberg/table-properties/.

Since all three operations(delete, write, merge) would use the same ```MergeWriterOperator```during table write, the engine is not able to tell which operation is in changing the table from inside ```MergeWriterOperator``` and ```IcebergMeatdata.getMergeRowIdColumnHandle``` to decide either to use COW or MOR to perform the table change. This PR adds an optional field called ```UpdateKind``` in ```IcebertTableHandle``` in order to the record operation being performed on the table. This field is populated by ```metata.getTableHandle``` by adding a new parameter ```UpdateKind```, which is retrieved at ```StatementAnalyzer```. 

As for the COW implementation part, we choose to implement it based on the current MOR implementation framework(MergeProcessOperator+MergeWriterOperator) similar to the COW implementation in Deltalake Connector. The general idea is to rewrite the changed data files based on current data file contents and the delta information from the current change operation. The current data file path is already embedded in the pages for the MOR implementation, and we can read the current data file contents from this path, the delta calculating is also already available for the MOR implementation, however, one caveat is that the data file may also contain several delete files, which should be used to calculate the new version of the data file but is not being embeded in the pages in the current MOR implementation, so we also embed delete files information for the COW implementation.

User facing docs will be updated as when the reviewers are satisfied with the implementation. 

## Description

Fix https://github.com/trinodb/trino/issues/17272

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/Metadata.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/Metadata.java
@@ -54,6 +54,7 @@ import io.trino.spi.connector.TableColumnsMetadata;
 import io.trino.spi.connector.TableFunctionApplicationResult;
 import io.trino.spi.connector.TableScanRedirectApplicationResult;
 import io.trino.spi.connector.TopNApplicationResult;
+import io.trino.spi.connector.UpdateKind;
 import io.trino.spi.connector.WriterScalingOptions;
 import io.trino.spi.expression.ConnectorExpression;
 import io.trino.spi.expression.Constant;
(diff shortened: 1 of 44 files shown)


ebyhr:
The connector intentionally doesn’t support CoW. I recall that some maintainers were opposed to adding support for it. I recommend reaching an agreement before starting the review process.

pettyjamesm:
@ebyhr - I don't see any disagreement about adding support in https://github.com/trinodb/trino/issues/17272. Who should we discuss the possibility of adding CoW support with?

ebyhr:
@pettyjamesm It was an offline discussion. cc: @dain @electrum

wendigo:
How is this exposed to the user?

kekwan:
@ebyhr What were some of the reasons for not supporting CoW?

PR #26162, opened 2025-07-10
Add support for `SET DEFAULT` and `DROP DEFAULT` statements

ebyhr:
## Description

Add support for the following syntax:
```g4
    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
        ALTER COLUMN columnName=identifier SET DEFAULT literal         #setDefaultValue
    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
        ALTER COLUMN columnName=identifier DROP DEFAULT                #dropDefaultValue
```

`DROP DEFAULT` on columns that don't have defaults throws an exception because SQL spec states "The descriptor of C shall include a default value". 

## Release notes

```markdown
## General, Memory
* Add support for `SET DEFAULT` and `DROP DEFAULT` statements. ({issue}`issuenumber`)
```


core/trino-main/src/main/java/io/trino/execution/DropDefaultValueTask.java
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        ColumnMetadata columnMetadata = metadata.getColumnMetadata(session, tableHandle, columnHandle);
+        if (columnMetadata.isHidden()) {
+            throw semanticException(NOT_SUPPORTED, statement, "Cannot modify hidden column");
+        }
+        if (columnMetadata.getDefaultValue().isEmpty()) {
+            throw semanticException(NOT_SUPPORTED, statement, "Column '%s' does not have a default value", column);
+        }
+        metadata.dropDefaultValue(session, tableHandle, columnHandle);
```
Praveen2112, line 112: Like we do for `SetDefaultValueTask` - can we also check if the Connector supports them out ?

core/trino-main/src/main/java/io/trino/execution/SetDefaultValueTask.java
```diff
@@ -0,0 +1,119 @@
...
+                exceptionMessage += ", but a materialized view with that name exists.";
+            }
+            else if (metadata.isView(session, tableName)) {
+                exceptionMessage += ", but a view with that name exists.";
+            }
+            if (!statement.isTableExists()) {
+                throw semanticException(TABLE_NOT_FOUND, statement, "%s", exceptionMessage);
+            }
+            return immediateVoidFuture();
+        }
```
Praveen2112, line 94: Maybe as a follow-up - Do we need to abstract this out for all the DDL operations to be perfomed on a table ?

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -87,6 +87,10 @@ statement
         RENAME COLUMN (IF EXISTS)? from=qualifiedName TO to=identifier #renameColumn
     | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
         DROP COLUMN (IF EXISTS)? column=qualifiedName                  #dropColumn
+    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
+        ALTER COLUMN columnName=identifier SET DEFAULT literal         #setDefaultValue
+    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
+        ALTER COLUMN columnName=identifier DROP DEFAULT                #dropDefaultValue
```
Praveen2112: Do we have any plans on setting the same for field as well ?

Praveen2112 (approved):

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
ebyhr: There is no plan about it as far as I know.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -87,6 +87,10 @@ statement
         RENAME COLUMN (IF EXISTS)? from=qualifiedName TO to=identifier #renameColumn
     | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
         DROP COLUMN (IF EXISTS)? column=qualifiedName                  #dropColumn
+    | ALTER TABLE (IF EXISTS)? tableName=qualifiedName
+        ALTER COLUMN columnName=identifier SET DEFAULT literal         #setDefaultValue
```
kasiafi: Could we accept a `QualifiedName` instead of just an Identifier?

core/trino-main/src/main/java/io/trino/execution/SetDefaultValueTask.java
```diff
@@ -0,0 +1,119 @@
...
+        String columnName = statement.getColumnName().getValue();
+        ColumnHandle columnHandle = metadata.getColumnHandles(session, tableHandle).get(columnName);
+
+        if (columnHandle == null) {
+            throw semanticException(COLUMN_NOT_FOUND, statement, "Column '%s' does not exist", columnName);
+        }
+
+        ColumnMetadata columnMetadata = metadata.getColumnMetadata(session, tableHandle, columnHandle);
+        if (columnMetadata.isHidden()) {
+            throw semanticException(NOT_SUPPORTED, statement, "Cannot modify hidden columnName");
```
kasiafi: `"Cannot modify hidden columnName"` -> `"Cannot modify hidden column"`

core/trino-main/src/test/java/io/trino/execution/TestDropDefaultValueTask.java
```diff
@@ -0,0 +1,177 @@
...
+                .hasMessageContaining("Table '%s' does not exist, but a materialized view with that name exists.", materializedViewName);
+    }
+
+    private ListenableFuture<Void> executeDropDefaultValue(QualifiedName table, Identifier column, boolean tableExists)
+    {
+        return new DropDefaultValueTask(plannerContext.getMetadata(), new AllowAllAccessControl())
+                .execute(new DropDefaultValue(new NodeLocation(1, 1), table, column, tableExists), queryStateMachine, ImmutableList.of(), WarningCollector.NOOP);
+    }
+
+    private static ConnectorTableMetadata simpleTable(QualifiedObjectName tableName)
```
kasiafi: Maybe rename to something like `simpleTableWithDefault`?

core/trino-main/src/test/java/io/trino/execution/TestSetDefaultValueTask.java
```diff
@@ -0,0 +1,177 @@
...
+    @BeforeEach
+    public void setUp()
+    {
+        super.setUp();
+        metadata = new MockMetadataWithDefaultValue(TEST_CATALOG_NAME);
+        plannerContext = plannerContextBuilder().withMetadata(metadata).build();
+    }
+
+    @Test
+    void testSetDefaultValue()
```
kasiafi, line 67: Please test the case where the statement fails because the new default value has wrong type.

core/trino-main/src/main/java/io/trino/execution/DropDefaultValueTask.java
```diff
@@ -0,0 +1,109 @@
...
+        if (columnHandle == null) {
+            throw semanticException(COLUMN_NOT_FOUND, statement, "Column '%s' does not exist", column);
+        }
+
+        ColumnMetadata columnMetadata = metadata.getColumnMetadata(session, tableHandle, columnHandle);
+        if (columnMetadata.isHidden()) {
+            throw semanticException(NOT_SUPPORTED, statement, "Cannot modify hidden column");
+        }
+        if (columnMetadata.getDefaultValue().isEmpty()) {
+            throw semanticException(NOT_SUPPORTED, statement, "Column '%s' does not have a default value", column);
```
kasiafi: If this failure is required by the spec, we should use another error code.
`NOT_SUPPORTED` is for cases where we don't follow the spec yet.

kasiafi (commented):
Reviewed.

By the way, I noticed that when we alter column type, we don't do any generic checks related to the default value. What does the spec say about it?

ebyhr:
The spec doesn't explicitly mention checking default values in `11.19 <alter column data type clause>`, if I understand it correctly.

core/trino-main/src/main/java/io/trino/execution/DropDefaultValueTask.java
```diff
@@ -0,0 +1,115 @@
...
+        if (columnHandle == null) {
+            throw semanticException(COLUMN_NOT_FOUND, statement, "Column '%s' does not exist", columnName);
+        }
+
+        ColumnMetadata columnMetadata = metadata.getColumnMetadata(session, tableHandle, columnHandle);
+        if (columnMetadata.isHidden()) {
+            throw semanticException(NOT_SUPPORTED, statement, "Cannot modify hidden column");
+        }
+        if (columnMetadata.getDefaultValue().isEmpty()) {
+            throw semanticException(GENERIC_USER_ERROR, statement, "Column '%s' does not have a default value", columnName);
```
kasiafi, line 106: You can add a new error code for this case.

martint:
Syntax looks good.

ebyhr:
(Squashed a fixup commit)

PR #26167, opened 2025-07-10
Fix column type index in MemoryPagesStore

pettyjamesm:
## Description
Follows up on https://github.com/trinodb/trino/pull/26140 by addressing [this comment](https://github.com/trinodb/trino/pull/26140#discussion_r2195668539) pointing out that the wrong loop index variable was being used for the RLE blocks attached.

Marking this change as not user visible since the introduced bug has not yet been released.

## Release notes


wendigo:
Is there a way to test that? :)

wendigo (approved):

Issue #25747, opened 2025-05-07
Pinot Connector: pinot.broker-url does not disable service discovery, leading to errors with scaled-down brokers


potlurip:
Certainly, Joe. Below is your complete and updated GitHub issue in Markdown, incorporating the clarified emphasis:

The real issue is that pinot.broker-url is being completely ignored, regardless of the cluster size or broker scale.

⸻

### Description

The `pinot.broker-url` configuration in the Pinot connector is designed to allow Trino to bypass Pinot controller-based broker discovery and instead directly route queries to the specified broker URL. However, we have observed that **Trino completely ignores this configuration**, continues performing service discovery via the controller, and routes queries using the full list of brokers — including those that may have been removed or scaled down.

This results in **intermittent query failures** and **violates the expected behavior** of the `pinot.broker-url` setting.

---

### Expected Behavior

When `pinot.broker-url` is provided in the Trino catalog configuration:

- ✅ Trino should **not contact the Pinot controller** for broker discovery
- ✅ Trino should **only use the broker URL provided**
- ✅ Queries should be routed **exclusively through** the specified broker service
- ✅ The presence or absence of other broker replicas should be **irrelevant**

---

### Actual Behavior

- ❌ Trino continues to perform **controller-based service discovery**
- ❌ The `pinot.broker-url` is **not used at all**, despite being valid
- ❌ Trino attempts to connect to **dead or removed broker instances**
- ❌ Queries intermittently fail with broker unreachable errors

---

### Steps to Reproduce

1. Set up a Pinot cluster with **any number of broker replicas** (e.g., 3)
2. Create a **Kubernetes service endpoint** for the brokers  
   e.g., `pinot-broker.pinot.svc.cluster.local:8000`
3. Verify this service works:
   ```bash
   curl http://pinot-broker.pinot.svc.cluster.local:8000

	•	✅ Returns valid response regardless of number of broker pods behind it

4. Configure Trino’s Pinot connector with the following catalog properties:

connector.name=pinot
pinot.controller-urls=pinot-controller.pinot.svc.cluster.local:9000
pinot.broker-url=pinot-broker.pinot.svc.cluster.local:8000


5. Restart Trino
6. Execute any query using the Pinot connector

⸻

Observed Behavior
	•	Trino logs show it still performs discovery via the controller
	•	It fetches a list of all known brokers — even those no longer running
	•	The provided pinot.broker-url is completely ignored
[...]

PR #26170, opened 2025-07-10
Fix Pinot connector to respect pinot.broker-url when proxy is enabled

shreyas-dev:
## Description
Add configuration validation to prevent conflicting `pinot.broker-url` and `pinot.proxy.enabled` settings in the Pinot connector. 

Previously, users could configure both `pinot.broker-url` and `pinot.proxy.enabled=true` simultaneously, which would result in confusing behavior where the explicit broker URL would silently take precedence over the proxy setting without any indication to the user.

**Changes made:**
- Add `@AssertTrue` validation in `PinotConfig.isValidBrokerProxyConfiguration()` to reject conflicting settings

Related to: https://github.com/trinodb/trino/issues/25747

## Release notes

### DIFF
--- a/plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/client/PinotClient.java
+++ b/plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/client/PinotClient.java
@@ -170,7 +170,7 @@ public class PinotClient
                 asyncReloading(CacheLoader.from(this::getAllTables), executor));
         this.controllerAuthenticationProvider = controllerAuthenticationProvider;
         this.brokerAuthenticationProvider = brokerAuthenticationProvider;
-        brokerHostAndPort = config.getBrokerUrl();
+        this.brokerHostAndPort = config.getBrokerUrl();
     }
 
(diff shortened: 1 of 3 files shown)


plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/client/PinotClient.java
```diff
@@ -170,7 +170,11 @@ public PinotClient(
                 asyncReloading(CacheLoader.from(this::getAllTables), executor));
         this.controllerAuthenticationProvider = controllerAuthenticationProvider;
         this.brokerAuthenticationProvider = brokerAuthenticationProvider;
-        brokerHostAndPort = config.getBrokerUrl();
+        this.brokerHostAndPort = config.getBrokerUrl();
+
+        if (this.brokerHostAndPort.isPresent() && this.proxyEnabled) {
```
wendigo: This should be validated in the config class and rejected instead

chenjian2664:
`TestPinotConfig.testExplicitPropertyMappings:89 Untested properties [pinot.proxy.enabled]` Please fix ci failure, it's related to the change

PR #3434, opened 2020-04-14
Allow using temporary staging path for writing sorted files

raunaqmorarka:
Fixes https://github.com/prestosql/presto/issues/471

PR #17390, opened 2023-05-08
Remove session property overrides for Hive staging location

alexjo2144:
## Description

Users should not have the ability to write files into arbitary locations.

## Additional context and related issues

Extracted first commit into https://github.com/trinodb/trino/pull/17389

## Release notes

( ) This is not user-visible or docs only and no release notes are required.
( ) Release notes are required, please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
# Hive
* Remove session property overrides for `temporary-staging-directory-enabled` and `temporary-staging-directory-path`
```

PR #26172, opened 2025-07-10
Allow using local staging path in iceberg for sorted writes

### LINKED ISSUE #24376: Large insert into sorted Iceberg table fails
For large inserts (in the order of 10s-100s billion rows), we see a couple of different issues

## Failure due to S3 throttling while writing buffer files
S3 throttles Trino when it performs PUT/GET/DELETE operations on the buffer files used for sorting. This leads to long running queries which can fail at any time.

For normal data files, the solution for this is to use the `write.object-storage.enabled` [table property](https://iceberg.apache.org/docs/1.7.0/aws/#object-store-file-layout). This adds a random hash to the prefix, which allows S3 to better distribute resources. Until recently (and in 435 which I am using), this had to be enabled using a system procedure and Trino, during writes, would create `ObjectStoreLocationProvider` instead of `DefaultLocationProvider`.
https://github.com/trinodb/trino/blob/c11e77e08431999ec80c636224889ddcf8d6267d/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java#L618

(This can now be set at the Trino level too, with #20555, but that's unrelated)

This table property does not get used for buffer files, while sorting, however - because the random hash prefix is _before_ `trino-tmp-files`.

The `Location` for `trino-tmp-files` (where buffer files are created) is instantiated here:
https://github.com/trinodb/trino/blob/c11e77e08431999ec80c636224889ddcf8d6267d/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java#L171

A hash is computed, but for `trino-tmp-files` putting the hash before it, here:
https://github.com/trinodb/trino/blob/c11e77e08431999ec80c636224889ddcf8d6267d/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java#L71-L75

And then buffer files simply append to this path here:
https://github.com/trinodb/trino/blob/c11e77e08431999ec80c636224889ddcf8d6267d/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java#L360

**Potential solutions:**
1. Update the logic such that the random hash is _after_ `trino-tmp-files` instead of before, when the `write.object-storage.enabled` table property is enabled
2. Create separate logic for buffer files vs data files, and introduce a new config specific to the buffer files.

On first glance option 2 seems unnecessary, but there is some value to controlling these independently. We use S3 lifecycle rules to expire objects using prefixes (which contain partition information) - if these are coupled (as in option 1), the random hash prefix would also be put before the partition prefixes, and expiring objects like this would no longer be possible. It is possible to define lifecycle rules based on object tags, but tagging objects is not supported in Trino afaik.

## Queries taking very long/timing out due to S3 I/O
This is not quite a _failure_ in the literal sense since we can adjust the query timeout, but an insert running for several (12-13+) hours is not ideal, especially if it's largely due to slow S3 I/O.

We can introduce a config in trino-iceberg, similar to trino-hive's `hive.temporary-staging-directory-enabled` (and `hive.temporary-staging-directory-path`):
https://github.com/trinodb/trino/blob/c11e77e08431999ec80c636224889ddcf8d6267d/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java#L1015

This simply lets Trino use storage local to the Trino worker for these buffer files, completely circumventing S3 and avoiding the I/O bottleneck.

Both these changes can be added, and controlled with configs, to allow a user to choose which option works better for them.

I intend to submit a PR myself to address these, but wanted to start the conversation now and discuss options/solutions.

agoyal-sfdc:
## Description

This change adds support for using temporary staging directory during write operations involving sorted tables. Writes to sorted tables will utilize this path for staging temporary files during sorting operation. When disabled, the target storage will be used for staging while writing sorted tables which can be inefficient when writing to object stores like S3.

Fixes https://github.com/trinodb/trino/issues/24376
Similar to functionality added for Hive in https://github.com/trinodb/trino/pull/3434

## Release notes

```markdown
## Iceberg
* Support using a local staging path for improved performance of writes to sorted tables. This can be enabled using the catalog configuration property `iceberg.sorted-writing.local-staging-path`. ({issue}`24376`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -217,6 +217,15 @@ implementation is used:
   -  Enable [sorted writing](iceberg-sorted-files) to tables with a specified sort order. Equivalent
      session property is `sorted_writing_enabled`.
   -  `true` 
+* - `iceberg.sorted-writing.local-staging-path`
+  -  A local directory that Trino can use for staging writes to sorted tables.
+     The `${USER}` placeholder can be used to use a different
+     location for each user. When this property is not configured, the target 
+     storage will be used for staging while writing to sorted tables which can
+     be inefficient when writing to object stores like S3. When 
+     `fs.hadoop.enabled` is not enabled, using this feature requires setup of 
+     [local file system](/object-storage/file-system-local)
+  -  
 * - `iceberg.allowed-extra-properties`
   -  List of extra properties that are allowed to be set on Iceberg tables.
      Use `*` to allow all properties.
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
@@ -22,8 +22,10 @@ import io.airlift.configuration.LegacyConfig;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import io.airlift.units.ThreadCount;
+import io.trino.filesystem.Location;
 import io.trino.plugin.hive.HiveCompressionOption;
 import jakarta.validation.constraints.AssertFalse;
+import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.DecimalMax;
 import jakarta.validation.constraints.DecimalMin;
 import jakarta.validation.constraints.Max;
@@ -86,6 +88,7 @@ public class IcebergConfig
     private boolean hideMaterializedViewStorageTable = true;
     private Optional<String> materializedViewsStorageSchema = Optional.empty();
     private boolean sortedWritingEnabled = true;
+    private Optional<String> sortedWritingLocalStagingPath = Optional.empty();
     private boolean queryPartitionFilterRequired;
     private Set<String> queryPartitionFilterRequiredSchemas = ImmutableSet.of();
     private int splitManagerThreads = Math.min(Runtime.getRuntime().availableProcessors() * 2, 32);
@@ -451,6 +454,32 @@ public class IcebergConfig
     {
         this.sortedWritingEnabled = sortedWritingEnabled;
         return this;
+    }
+
+    @NotNull
+    public Optional<String> getSortedWritingLocalStagingPath()
+    {
+        return sortedWritingLocalStagingPath;
+    }
+
+    @Config("iceberg.sorted-writing.local-staging-path")
+    @ConfigDescription("Use provided local directory for staging writes to sorted tables. Use ${USER} placeholder to use different location for each user")
+    public IcebergConfig setSortedWritingLocalStagingPath(String sortedWritingLocalStagingPath)
+    {
+        this.sortedWritingLocalStagingPath = Optional.ofNullable(sortedWritingLocalStagingPath);
+        return this;
+    }
+
+    @AssertTrue(message = "iceberg.sorted-writing.local-staging-path must not use any prefix other than file:// or local://")
+    public boolean isSortedWritingLocalStagingPathValid()
+    {
+        if (sortedWritingLocalStagingPath.isEmpty()) {
+            return true;
+        }
+        Optional<String> scheme = Location.of(sortedWritingLocalStagingPath.get()).scheme();
+        return scheme.isEmpty()
+                || scheme.equals(Optional.of("file"))
+                || scheme.equals(Optional.of("local"));
     }
 
     @Config("iceberg.query-partition-filter-required")
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
@@ -149,6 +149,7 @@ public class IcebergPageSink
             List<TrinoSortField> sortOrder,
             DataSize sortingFileWriterBufferSize,
             int sortingFileWriterMaxOpenFiles,
+            Optional<String> sortedWritingLocalStagingPath,
             TypeManager typeManager,
             PageSorter pageSorter)
     {
@@ -171,12 +172,16 @@ public class IcebergPageSink
         this.sortedWritingEnabled = isSortedWritingEnabled(session);
         this.sortingFileWriterBufferSize = requireNonNull(sortingFileWriterBufferSize, "sortingFileWriterBufferSize is null");
         this.sortingFileWriterMaxOpenFiles = sortingFileWriterMaxOpenFiles;
-        this.tempDirectory = Location.of(locationProvider.newDataLocation("trino-tmp-files"));
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.pageSorter = requireNonNull(pageSorter, "pageSorter is null");
         this.columnTypes = getTopLevelColumns(outputSchema, typeManager).stream()
                 .map(IcebergColumnHandle::getType)
                 .collect(toImmutableList());
+
+        this.tempDirectory = sortedWritingLocalStagingPath
+                .map(path -> path.replace("${USER}", session.getIdentity().getUser()))
+                .map(IcebergPageSink::createLocalSchemeIfAbsent)
+                .orElseGet(() -> Location.of(locationProvider.newDataLocation("trino-tmp-files")));
 
         if (sortedWritingEnabled) {
             ImmutableList.Builder<Integer> sortColumnIndexes = ImmutableList.builder();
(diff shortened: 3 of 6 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -388,6 +392,29 @@ private int[] getWriterIndexes(Page page)
...
+        }
+        return tempFilePath;
+    }
+
+    private static Location setSchemeToFileIfAbsent(Location location)
+    {
+        if (location.scheme().isPresent()) {
+            return location;
+        }
+        return Location.of("file:///" + location.path());
```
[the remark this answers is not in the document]
agoyal-sfdc: This is identical to the function in HiveWriterFactory: https://github.com/trinodb/trino/blob/5b81408659fafa778f28bad6ea342f92b5f2e9d1/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveWriterFactory.java#L685-L691

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -87,6 +87,8 @@ public class IcebergConfig
     private boolean hideMaterializedViewStorageTable = true;
     private Optional<String> materializedViewsStorageSchema = Optional.empty();
     private boolean sortedWritingEnabled = true;
+    private boolean temporaryStagingDirectoryEnabled;
+    private String temporaryStagingDirectoryPath = "/tmp/presto-${USER}";
```
raunaqmorarka: presto -> trino

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -217,6 +217,22 @@ implementation is used:
...
+     disabled, the target storage will be used for staging while writing sorted
+     tables which can be inefficient when writing to object stores like S3. Equivalent
+     session property is `sorted_writing_temporary_staging_directory_enabled`.
+  -  `false`
+* - `iceberg.temporary-staging-directory-path`
+  -  Controls the location of temporary staging directory that is used for write
+     operations. The `${USER}` placeholder can be used to use a different
+     location for each user. Equivalent session property is
+     `sorted_writing_temporary_staging_directory_path`.
+  -  `/tmp/presto-${USER}`
```
raunaqmorarka: presto -> trino

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
```diff
@@ -106,6 +107,10 @@ public final class IcebergSessionProperties
     public static final String REMOVE_ORPHAN_FILES_MIN_RETENTION = "remove_orphan_files_min_retention";
     private static final String MERGE_MANIFESTS_ON_WRITE = "merge_manifests_on_write";
     private static final String SORTED_WRITING_ENABLED = "sorted_writing_enabled";
+    private static final String SORTED_WRITING_WRITER_BUFFER_SIZE = "sorted_writing_write_buffer_size";
+    private static final String SORTED_WRITING_WRITER_MAX_OPEN_FILES = "sorted_writing_writer_max_open_files";
+    private static final String SORTED_WRITING_TEMP_STAGING_DIR_ENABLED = "sorted_writing_temporary_staging_directory_enabled";
+    private static final String SORTED_WRITING_TEMP_STAGING_DIR_PATH = "sorted_writing_temporary_staging_directory_path";
```
raunaqmorarka: We don't want to expose any of these as session properties, some of these were removed from hive connector in https://github.com/trinodb/trino/pull/17390/files#diff-7dad18a4dbe9d9a536d45655875f47ffd0d8e3b5d5c6d2f3447246763841fb44L7716

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -566,6 +566,33 @@ public void testFileSortingWithLargerTable()
...
     }
 
+    @Test
+    public void testFileSortingWithLargerTableAndTempDirForBufferFiles()
+    {
+        // Using a larger table forces buffered data to be written to disk
+        Session withSmallRowGroups = Session.builder(getSession())
+                .setCatalogSessionProperty("iceberg", "orc_writer_max_stripe_rows", "200")
+                .setCatalogSessionProperty("iceberg", "parquet_writer_block_size", "20kB")
+                .setCatalogSessionProperty("iceberg", "parquet_writer_batch_size", "200")
```
raunaqmorarka: These are not needed, sorting is per file, not per row-group

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -566,6 +566,33 @@ public void testFileSortingWithLargerTable()
...
+                .setCatalogSessionProperty("iceberg", "sorted_writing_temporary_staging_directory_enabled", "true")
+                .setCatalogSessionProperty("iceberg", "sorted_writing_write_buffer_size", "2kB")
+                .setCatalogSessionProperty("iceberg", "sorted_writing_writer_max_open_files", "5")
```
raunaqmorarka: These should be setup in config properties rather than session

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
raunaqmorarka: You can use small value of `target_max_file_size` to produce multiple files

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -566,6 +566,33 @@ public void testFileSortingWithLargerTable()
         }
     }
 
+    @Test
+    public void testFileSortingWithLargerTableAndTempDirForBufferFiles()
```
raunaqmorarka: testSortedWritingTempStaging

docs/src/main/sphinx/connector/iceberg.md
agoyal-sfdc: @raunaqmorarka I defaulted to the same one as the Hive temp directory. Happy to change too if you prefer

raunaqmorarka: we don't need retain the old naming, please change it here

raunaqmorarka (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -149,6 +149,8 @@ public IcebergPageSink(
             List<TrinoSortField> sortOrder,
             DataSize sortingFileWriterBufferSize,
             int sortingFileWriterMaxOpenFiles,
+            boolean sortedWritingLocalStagingEnabled,
```
wendigo: Instead of having boolean and path we can just have. an Optional<String> path - empty indicates no local staging, present - use local staging

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -171,13 +173,20 @@ public IcebergPageSink(
...
         this.sortingFileWriterBufferSize = requireNonNull(sortingFileWriterBufferSize, "sortingFileWriterBufferSize is null");
         this.sortingFileWriterMaxOpenFiles = sortingFileWriterMaxOpenFiles;
-        this.tempDirectory = Location.of(locationProvider.newDataLocation("trino-tmp-files"));
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.pageSorter = requireNonNull(pageSorter, "pageSorter is null");
         this.columnTypes = getTopLevelColumns(outputSchema, typeManager).stream()
                 .map(IcebergColumnHandle::getType)
                 .collect(toImmutableList());
 
+        if (sortedWritingLocalStagingEnabled) {
```
wendigo: if sortedWritingLocalStagingPath is an Optional this could become:

```
this.tempDirectory = sortedWritingLocalStagingPath
   .map(path -> path.replace("${USER}", session.getIdentity().getUser())
   .map(this::createLocalSchemeIfAbsent)
   .orElseGet(() -> Location.of(locationProvider.newDataLocation("trino-tmp-files"));
```

wendigo (approved):

akroptya:
@agoyal-sfdc Hi. Could you clarify, is iceberg.sorted-writing.local-staging-path requires a shared filesystem folder? Because your link leads to https://trino.io/docs/current/object-storage/file-system-local.html "The local file system must be a local mount point for a shared file system available on all cluster nodes." Thank you.

PR #26174, opened 2025-07-11
Using explicit cast on bigint columns in row level update tests

chenjian2664:
Use explicit casts on bigint columns instead of relying on `skippingTypesCheck`, which is insufficient for some connectors like `Snowflake` and `Oracle`

## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
@@ -5077,13 +5077,13 @@ public abstract class BaseConnectorTest
         try (TestTable table = createTestTableForWrites("test_update", "AS TABLE tpch.tiny.nation", "nationkey,regionkey")) {
             String tableName = table.getName();
             assertUpdate("UPDATE " + tableName + " SET nationkey = 100 + nationkey WHERE regionkey = 2", 5);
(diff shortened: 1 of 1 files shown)


testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -5077,13 +5077,13 @@ public void testRowLevelUpdate()
         try (TestTable table = createTestTableForWrites("test_update", "AS TABLE tpch.tiny.nation", "nationkey,regionkey")) {
             String tableName = table.getName();
             assertUpdate("UPDATE " + tableName + " SET nationkey = 100 + nationkey WHERE regionkey = 2", 5);
-            assertThat(query("SELECT * FROM " + tableName))
+            assertThat(query("SELECT CAST(nationkey AS bigint), name, CAST(regionkey AS bigint), comment FROM " + tableName))
```
ebyhr, line 5080:
```
Update testRowLevelUpdateTest using explicit cast on bigint columns

Use explicit casts on bigint columns instead of relying on `skippingTypesCheck`, which is insufficient for some connectors like `Snowflake` and `Oracle`

f
```

testRowLevelUpdateTest -> testRowLevelUpdate
Also, wrap the body at 72 characters and remove a redundant `f`

ebyhr (approved):

PR #26175, opened 2025-07-11
Handle already closed JDBC connections gracefully

chenjian2664:
Introduces `ConnectionAlreadyClosedException` to signal attempts to
use closed JDBC connections. Updates `LazyConnectionFactory` to throw
this exception when a closed connection is accessed, and updates
`JdbcPageSink` to handle this exception without error

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ConnectionAlreadyClosedException.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ConnectionAlreadyClosedException.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.jdbc;
+
+public class ConnectionAlreadyClosedException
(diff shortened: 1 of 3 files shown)


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSink.java
```diff
@@ -238,6 +238,12 @@ public void abort()
                 connection.close();
             }
         }
+        catch (IllegalStateException e) {
+            if (e.getMessage().equals("Connection is already closed")) {
+                return;
+            }
```
ebyhr: The logic looks unreliable. The message can be changed in the future, the same message can be thrown in JDBC drivers. How about introducing a dedicated exception class?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ConnectionAlreadyClosedException.java
```diff
@@ -0,0 +1,23 @@
+/*
```
ebyhr, line 1:
```
Suppress IllegalStateException when closing closed JDBC connection

The `LazyConnection#delegate` method throws `IllegalStateException` if the connection is already
closed, we should ensure that such exceptions are caught and suppressed during connection close to
avoid unnecessary propagation.
```

The commit title seems stale. 
Also, the commit body is too long. https://trino.io/development/process.html#pull-request-and-commit-guidelines-
> Wrap the body at 72 characters.

ebyhr: The commit title is still wrong to me. `IllegalStateException` seems unrelated to this change.

chenjian2664: yes, after introducing the exception class, the commits is not accurate.
Updated, PTAL

ebyhr (approved):

PR #26180, opened 2025-07-11
Adding retry for on the fly trino_queries table creation for MySql event listener

lordicecream:
adding retry for trino to recreate the trino_queries table for mysql event listener incase the table or schema is dropped on the fly

## Description

While the trino coordinator is running, if the trino_queries table is dropped or the MySql database/schema is dropped and then the database/schema is recreated, then trino is not automatically able to recreate the trino_queries table again, and the query metrics are not pushed anywhere and the trino coordinator needs to be restarted to recreate the trino_queries table again.

This simply adds a retry in case of a tablenotfoundexception, so if due to any reason the database/schema needs to be regenerated, then trino coordinator can perform a retry to recreate the trino_queries table to continue pushing the query metrics.

## Release notes

### DIFF
--- a/plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/MysqlEventListener.java
+++ b/plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/MysqlEventListener.java
@@ -161,7 +161,25 @@ public class MysqlEventListener
                 stats.getCompletedSplits(),
                 context.getRetryPolicy(),
                 createOperatorSummariesJson(metadata.getQueryId(), stats.getOperatorSummaries()));
-        dao.store(entity);
+        try {
+            dao.store(entity);
+        }
+        catch (Exception e) {
+            if (isMissingTableException(e)) {
+                log.warn("Table 'trino_queries' is missing, attempting to create it.");
+                try {
+                    dao.createTable();
+                    dao.store(entity); // Retry once
+                    log.info("Successfully created table and starting to store query event.");
+                }
+                catch (Exception retryEx) {
+                    log.error("Retry failed: Could not create table or store query event (%s): %s", retryEx.getClass().getSimpleName(), retryEx.getMessage());
+                }
+            }
+            else {
+                log.error("Failed to store query event (%s): %s", e.getClass().getSimpleName(), e.getMessage());
+            }
+        }
     }
 
     private Optional<String> createOperatorSummariesJson(String queryId, List<String> summaries)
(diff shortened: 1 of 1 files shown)


plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/MysqlEventListener.java
```diff
@@ -161,7 +161,25 @@ public void queryCompleted(QueryCompletedEvent event)
                 stats.getCompletedSplits(),
                 context.getRetryPolicy(),
                 createOperatorSummariesJson(metadata.getQueryId(), stats.getOperatorSummaries()));
-        dao.store(entity);
+        try {
+            dao.store(entity);
+        }
+        catch (Exception e) {
+            if (isMissingTableException(e)) {
```
chenjian2664, line 168: I’m trying to understand what scenario would hit this—if the table was dropped, shouldn’t the fix be there instead?

lordicecream: This handles cases where the Trino coordinator is running, and the `trino_queries` table is missing — either because it was dropped, or the MySQL instance (including the schema) was recreated.
Without this, Trino requires a coordinator restart to resume logging and to recreate the table.
This change avoids that by retrying table creation once on insert failure, improving resilience without changing default behavior.

lordicecream: Hey @chenjian2664 hope this helped answer your question

chenjian2664: Thanks.
> `trino_queries`  table is missing — either because it was dropped, or the MySQL instance (including the schema) was recreated.

This is what I wanted to bring up—I think this is where the issue lies

> Trino requires a coordinator restart to resume logging and to recreate the table

Assuming the drop action is handled by another system, I was wondering if it might also be responsible for creating the table?

lordicecream: While an external system could recreate the table, its structure is specific to Trino’s internal expectations. Allowing Trino to recreate it ensures schema consistency and avoids errors from mismatches. An end user or external system might not know the exact table definition or how Trino expects the schema to be structured, which increases the risk of incorrect recreation.

lordicecream: Hey @chenjian2664 , can we push this?

lordicecream: Hey @chenjian2664 ,
Was wondering if you had any further inquiries?

danielbelchior:
Hello @lordicecream,

Your PR complements mine: https://github.com/trinodb/trino/pull/26254

It would be great to coordinate and ensure both changes align well. Let me know what you think!

plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/MysqlEventListener.java
chenjian2664: Hi @lordicecream  I think we might need more input. I’d suggest bringing it up in Slack to get more voices on it

lordicecream: Hey @chenjian2664 @ebyhr ,
I believe this would be a good add-on for end user pov.

This is adding fault tolerance for the only things that trino is doing - which is table creation and insertion.

Please let me know your thoughts and if we can take this ahead

Issue #26181, opened 2025-07-11
Feature request: Support HTTP methods other than POST in HTTP Event Listener


danielbelchior:
The current HTTP Event Listener only allows sending requests using the POST method by default. It would be great to have an option to choose between POST and PUT methods for more flexibility in how events are sent.

PR #26182, opened 2025-07-11
Add support for configurable HTTP methods (POST or PUT) in HTTP Event Listener

danielbelchior:
## Description
This PR introduces the ability to configure the HTTP method used by the HTTP Event Listener. Previously, only POST requests were supported by default. Now, users can choose between POST and PUT methods, providing greater flexibility for different use cases.

A new configuration option, http-event-listener.connect-http-method, was added to the HTTP Event Listener configuration file. This option accepts either POST (default) or PUT as values, allowing users to select the HTTP method used for sending event requests.

## Release notes

```markdown
## HTTP Event Listener
* Added support for configuring the HTTP method in the HTTP Event Listener. Users can now choose between POST (default) and PUT for sending event requests by setting the http-event-listener.connect-http-method option. ({issue}`26181`)
```
Closes: #26181

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
@@ -3934,6 +3934,9 @@ public class ExpressionAnalyzer
         if (type instanceof BooleanType) {
             checkArgument(value instanceof Boolean, "Value must be true or false '%s'".formatted(value));
         }
+        if (value == null) {
+            return;
+        }
         if (type instanceof CharType charType) {
             int targetLength = charType.getLength();
             Slice slice = (Slice) value;
--- a/core/trino-main/src/test/java/io/trino/util/TestColumnDefaultOptions.java
+++ b/core/trino-main/src/test/java/io/trino/util/TestColumnDefaultOptions.java
@@ -63,12 +63,6 @@ final class TestColumnDefaultOptions
     }
 
     @Test
-    void testNull()
-    {
-        assertDefaultColumnValue(TINYINT, new NullLiteral(LOCATION));
-    }
-
-    @Test
     void testBoolean()
     {
         assertDefaultColumnValue(BOOLEAN, new BooleanLiteral(LOCATION, "true"));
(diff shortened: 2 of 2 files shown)


plugin/trino-http-event-listener/src/main/java/io/trino/plugin/httpquery/HttpEventListener.java
```diff
@@ -144,7 +147,19 @@ public void splitCompleted(SplitCompletedEvent splitCompletedEvent)
 
     private void sendLog(BodyGenerator eventBodyGenerator, String queryId)
     {
-        Request request = preparePost()
+        Request.Builder requestBuilder;
```
ebyhr: We could use `Request.builder()` and `setMethod(String method)` instead.

plugin/trino-http-event-listener/src/main/java/io/trino/plugin/httpquery/HttpEventListenerConfig.java
```diff
@@ -94,6 +96,20 @@ public HttpEventListenerConfig setIngestUri(String ingestUri)
...
     }
 
+    @Pattern(regexp = "POST|PUT", message = "HTTP method must be either POST or PUT")
+    public String getHttpMethod()
+    {
+        return httpMethod;
+    }
+
+    @ConfigDescription("Specifies the HTTP method to use for the request. Supported values are POST (default) and PUT.")
+    @Config("http-event-listener.connect-http-method")
```
ebyhr: Could you document the new property in `event-listeners-http.md`?

ebyhr (commented):
Please squash commits into one.

plugin/trino-http-event-listener/src/main/java/io/trino/plugin/httpquery/HttpEventListener.java
danielbelchior: yes, well observed

plugin/trino-http-event-listener/src/main/java/io/trino/plugin/httpquery/HttpEventListenerConfig.java
danielbelchior: the documentation has been updated

PR #26027, opened 2025-06-18
Fix reading Variant null value

chenjian2664:
## Description

FIxes https://github.com/trinodb/trino/issues/26016

## Release notes

```markdown
## Delta Lake
* Fix failure when reading tables with `null` on `variant` type. ({issue}`26016`)
```

PR #26184, opened 2025-07-13
Fix reading variant null values in Delta Lake

chenjian2664:
## Description
In cases where a value chunk includes both null and non-null
entries, the chunk must be treated as potentially nullable.

This is the another corner case that missed in https://github.com/trinodb/trino/pull/26027

## Release notes

```markdown
## Delta Lake
* Fix failure when reading `null` values on `json` type columns. ({issue}`26184`)
```


ebyhr:
/test-with-secrets sha=b06be75c90801490ebc04fefaa3ce32f0ab7b6ae

lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
```diff
@@ -299,14 +299,14 @@ private static Optional<Field> constructField(Type type, ColumnIO columnIO, bool
...
-            Field valueField = constructField(VARBINARY, groupColumnIo.getChild(0), false).orElseThrow();
+            PrimitiveField valueField = (PrimitiveField) constructField(VARBINARY, groupColumnIo.getChild(0), false).orElseThrow();
             PrimitiveField metadataField = (PrimitiveField) constructField(VARBINARY, groupColumnIo.getChild(1), false).orElseThrow();
             return Optional.of(new VariantField(
                     type,
                     repetitionLevel,
                     definitionLevel,
                     required,
-                    valueField,
+                    new PrimitiveField(valueField.getType(), false, valueField.getDescriptor(), valueField.getId()),
```
ebyhr, line 309: Does it mean Parquet spec is wrong, or Databricks violates the spec? 
https://github.com/apache/parquet-format/blob/master/VariantEncoding.md#variant-in-parquet

> The value field must be annotated as **_required_** for unshredded Variant values

chenjian2664: Here is the meta of the test parquet file that written by Databricks:
```
parquet meta part-00000-3dae12c4-61bc-4177-bd36-2c936db81e90-c000.snappy.parquet

File path:  part-00000-3dae12c4-61bc-4177-bd36-2c936db81e90-c000.snappy.parquet
Created by: parquet-mr version 1.12.3-databricks-0002 (build 2484a95dbe16a0023e3eb29c201f99ff9ea771ee)
Properties:
                   org.apache.spark.version: 3.5.0
            com.databricks.spark.jobGroupId: 1752418278132_8642402946176337271_3e73399aa411479296f9bc16c62a6681
  org.apache.spark.sql.parquet.row.metadata: {"type":"struct","fields":[{"name":"id","type":"integer","nullable":true,"metadata":{}},{"name":"x","type":"variant","nullable":true,"metadata":{}}]}
             com.databricks.spark.clusterId: 1002-064054-nbosugsx
Schema:
message spark_schema {
  optional int32 id;
  optional group x {
    required binary value;
    required binary metadata;
  }
}
```
I think the Schema is correct

But I don't know why our `NestedColumnReader#readNonNull` can't read correctly  cc @raunaqmorarka

ebyhr (approved):

chenjian2664:
https://github.com/trinodb/trino/issues/26194

PR #26198, opened 2025-07-14
Include additional client failure information in the exception

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
+++ b/client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
@@ -511,6 +511,9 @@ class StatementClientV1
                     state.compareAndSet(State.RUNNING, State.CLIENT_ERROR);
                     throw requestFailedException(taskName, request, response);
                 }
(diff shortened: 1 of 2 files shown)


ebyhr (approved):
> (x) Release notes are required. Please propose a release note for me.

There is no need to mention changes to error messages in the release note in my opinion. This is an implementation detail.

client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
```diff
@@ -511,6 +511,9 @@ private boolean executeRequest(Request request, String taskName, Function<Except
                     state.compareAndSet(State.RUNNING, State.CLIENT_ERROR);
                     throw requestFailedException(taskName, request, response);
                 }
+                cause = new ClientException(format("Expected http code %d but got %d%s", HTTP_OK, response.getStatusCode(), response.getResponseBody()
```
losipiuk, line 514: should you trim response body here?

wendigo: I have no idea. Should I?

PR #26204, opened 2025-07-15
Support multipart upload to Azure

wendigo:
## Release notes

```markdown
## Section
```


wendigo:
This is modeled the same way S3OutputStream is working (except for the fact that there is no multipart abort since the staged blocks are automatically deleted by azure and there is no way to explicitly clean them)

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureOutputStream.java
```diff
@@ -135,32 +129,106 @@ private void ensureOpen()
...
             }
-            finally {
-                memoryContext.close();
+            else {
+                this.buffer = new byte[0];
+                this.initialBufferSize = writeBlockSizeBytes;
+                bufferSize = 0;
             }
+            memoryContext.setBytes(0);
+            waitForPreviousUploadFinish();
```
dain: You might want to have a separate context for the inprogress upload, that gets set to zero when the upload finishes. Then you can just swap current into inprogress after this line.

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureOutputStream.java
```diff
@@ -135,32 +129,106 @@ private void ensureOpen()
...
+        if (finished && !multiStagingStarted) {
+            BlobOutputStream blobOutputStream = blockClient
+                    .getBlobOutputStream(overwrite);
+
+            blobOutputStream.write(buffer, 0, bufferSize);
+            blobOutputStream.close();
+            return;
+        }
+
+        if ((bufferSize == writeBlockSizeBytes) || (finished && (bufferSize > 0))) {
```
dain: I'd invert this and exit early to reduce the extra level of intent here

dain (approved):
Nice, I didn't know they supported multipart upload

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureOutputStream.java
wendigo: This is the same as in the S3OutputStream. I can follow up with the change there

anusudarsan (approved):

chenjian2664:
@wendigo The `Release Notes` section has not been updated yet.

wendigo:
I'll expose it behind the config toggle to have two separate implementations for writes

PR #26210, opened 2025-07-16
Replace `threeten.bp.Duration` with `java.time.Duration`

chenjian2664:
Updated `GcsStorageFactory` to use `java.time.Duration` instead
 of `org.threeten.bp.Duration`

## Release notes

```markdown
## Section
```


ebyhr:
(Fixed build failure)

plugin/trino-bigquery/pom.xml
```diff
@@ -324,11 +324,6 @@
             <version>20250517</version>
         </dependency>
 
-        <dependency>
```
chenjian2664, line 327: ah, I was thought that the dep was kept for some other class usage

PR #26213, opened 2025-07-16
Allow skipping of data deletion in expire_snapshots

tbaeg:
## Description
Allow skipping of file deletion in `expire_snapshots` to allow for more efficient out-of-band deletes.

## Release notes

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -884,7 +884,7 @@ with the `retention_threshold` parameter.
 `expire_snapshots` can be run as follows:
 
 ```sql
-ALTER TABLE test_table EXECUTE expire_snapshots(retention_threshold => '7d');
+ALTER TABLE test_table EXECUTE expire_snapshots(retention_threshold => '7d', delete_files => true);
 ```
 
 The value for `retention_threshold` must be higher than or equal to
@@ -892,6 +892,10 @@ The value for `retention_threshold` must be higher than or equal to
 procedure fails with a similar message: `Retention specified (1.00d) is shorter
 than the minimum retention configured in the system (7.00d)`. The default value
 for this property is `7d`.
+
+The value for `delete_files` can be `true` or `false`. When set to `false` 
+files associated with the expired snapshots will NOT be deleted. The default 
+value for this property is `true`.
 
 (iceberg-remove-orphan-files)=
 ##### remove_orphan_files
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -437,6 +437,7 @@ public class IcebergMetadata
     private static final int OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION = 2;
     private static final int CLEANING_UP_PROCEDURES_MAX_SUPPORTED_TABLE_VERSION = 2;
     private static final String RETENTION_THRESHOLD = "retention_threshold";
+    private static final String DELETE_FILES = "delete_files";
     private static final String UNKNOWN_SNAPSHOT_TOKEN = "UNKNOWN";
     public static final Set<String> UPDATABLE_TABLE_PROPERTIES = ImmutableSet.<String>builder()
             .add(EXTRA_PROPERTIES_PROPERTY)
@@ -1746,12 +1747,13 @@ public class IcebergMetadata
     private Optional<ConnectorTableExecuteHandle> getTableHandleForExpireSnapshots(ConnectorSession session, IcebergTableHandle tableHandle, Map<String, Object> executeProperties)
     {
         Duration retentionThreshold = (Duration) executeProperties.get(RETENTION_THRESHOLD);
+        boolean deleteFiles = (boolean) executeProperties.get(DELETE_FILES);
         Table icebergTable = catalog.loadTable(session, tableHandle.getSchemaTableName());
 
         return Optional.of(new IcebergTableExecuteHandle(
                 tableHandle.getSchemaTableName(),
                 EXPIRE_SNAPSHOTS,
-                new IcebergExpireSnapshotsHandle(retentionThreshold),
+                new IcebergExpireSnapshotsHandle(retentionThreshold, deleteFiles),
                 icebergTable.location(),
                 icebergTable.io().properties()));
     }
@@ -2208,6 +2210,7 @@ public class IcebergMetadata
 
         // ForwardingFileIo handles bulk operations so no separate function implementation is needed
         table.expireSnapshots()
+                .cleanExpiredFiles(expireSnapshotsHandle.deleteFiles())
                 .expireOlderThan(session.getStart().toEpochMilli() - retention.toMillis())
                 .planWith(icebergScanExecutor)
                 .commit();
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/ExpireSnapshotsTableProcedure.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/ExpireSnapshotsTableProcedure.java
@@ -21,6 +21,7 @@ import io.trino.spi.connector.TableProcedureMetadata;
 import static io.trino.plugin.base.session.PropertyMetadataUtil.durationProperty;
 import static io.trino.plugin.iceberg.procedure.IcebergTableProcedureId.EXPIRE_SNAPSHOTS;
 import static io.trino.spi.connector.TableProcedureExecutionMode.coordinatorOnly;
+import static io.trino.spi.session.PropertyMetadata.booleanProperty;
 
 public class ExpireSnapshotsTableProcedure
         implements Provider<TableProcedureMetadata>
@@ -36,6 +37,11 @@ public class ExpireSnapshotsTableProcedure
                                 "retention_threshold",
                                 "Only snapshots older than threshold should be removed",
                                 Duration.valueOf("7d"),
+                                false),
+                        booleanProperty(
+                                "delete_files",
+                                "Delete underlying files associated to the expired snapshot(s)",
+                                true,
                                 false)));
     }
 }
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -6550,6 +6550,44 @@ public abstract class BaseIcebergConnectorTest
     }
 
     @Test
+    public void testExpireSnapshotsKeepFiles()
+            throws Exception
+    {
+        try (TestTable table = newTrinoTable("test_expiring_snapshots_", "(key varchar, value integer)")) {
+            Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES ('one', 1)", 1);
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES ('two', 2)", 1);
+            List<String> initialFiles = getAllDataFilesFromTableDirectory(table.getName());
+
+            assertQuerySucceeds(sessionWithShortRetentionUnlocked, "ALTER TABLE " + table.getName() + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s', delete_files => false)");
+
+            List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(table.getName());
+            assertThat(updatedDataFiles).containsExactlyInAnyOrderElementsOf(initialFiles);
+            assertThat(query("SELECT sum(value), listagg(key, ' ') WITHIN GROUP (ORDER BY key) FROM " + table.getName()))
+                    .matches("VALUES (BIGINT '3', VARCHAR 'one two')");
+        }
+    }
+
+    @Test
+    public void testExpireSnapshotsPartitionedTableKeepFiles()
+            throws Exception
+    {
+        try (TestTable table = newTrinoTable("test_expiring_snapshots_partitioned_table", "(col1 BIGINT, col2 BIGINT) WITH (partitioning = ARRAY['col1'])")) {
+            Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES(1, 100), (1, 101), (1, 102), (2, 200), (2, 201), (3, 300)", 6);
+            assertUpdate("DELETE FROM " + table.getName() + " WHERE col1 = 1", 3);
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES(4, 400)", 1);
+            List<String> initialFiles = getAllDataFilesFromTableDirectory(table.getName());
+
+            assertQuerySucceeds(sessionWithShortRetentionUnlocked, "ALTER TABLE " + table.getName() + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s', delete_files => false)");
+
+            List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(table.getName());
+            assertThat(updatedDataFiles).containsExactlyInAnyOrderElementsOf(initialFiles);
+            assertQuery("SELECT sum(col2) FROM " + table.getName(), "SELECT 1101");
+        }
+    }
+
+    @Test
     public void testExpireSnapshotsOnSnapshot()
     {
         String tableName = "test_expire_snapshots_on_snapshot_" + randomNameSuffix();
@@ -6583,7 +6621,7 @@ public abstract class BaseIcebergConnectorTest
         assertUpdate("INSERT INTO " + tableName + " VALUES ('two', 2)", 1);
 
         assertExplain("EXPLAIN ALTER TABLE " + tableName + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s')",
-                "SimpleTableExecute\\[table = iceberg:schemaTableName:tpch.test_expiring_snapshots.*\\[retentionThreshold=0\\.00s].*");
+                "SimpleTableExecute\\[table = iceberg:schemaTableName:tpch.test_expiring_snapshots.*\\[retentionThreshold=0\\.00s, deleteFiles=true].*");
     }
 
     @Test
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/IcebergExpireSnapshotsHandle.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/IcebergExpireSnapshotsHandle.java
@@ -17,7 +17,7 @@ import io.airlift.units.Duration;
 
 import static java.util.Objects.requireNonNull;
 
-public record IcebergExpireSnapshotsHandle(Duration retentionThreshold)
+public record IcebergExpireSnapshotsHandle(Duration retentionThreshold, boolean deleteFiles)
         implements IcebergProcedureHandle
 {
     public IcebergExpireSnapshotsHandle


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/ExpireSnapshotsTableProcedure.java
```diff
@@ -36,6 +37,11 @@ public TableProcedureMetadata get()
                                 "retention_threshold",
                                 "Only snapshots older than threshold should be removed",
                                 Duration.valueOf("7d"),
+                                false),
+                        booleanProperty(
+                                "delete_files",
+                                "Expire snapshots without deleting the underlying data",
```
ebyhr: The description is confusing since it's about disabled case.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -416,6 +416,7 @@ public class IcebergMetadata
     private static final int OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION = 2;
     private static final int CLEANING_UP_PROCEDURES_MAX_SUPPORTED_TABLE_VERSION = 2;
     private static final String RETENTION_THRESHOLD = "retention_threshold";
+    private static final String DELETE_FILES = "delete_files";
```
ebyhr, line 440: This parameter should be documented.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6550,6 +6550,44 @@ public void testExpireSnapshotsPartitionedTable()
...
+    public void testExpireSnapshotsKeepFiles()
+            throws Exception
+    {
+        String tableName = "test_expiring_snapshots_" + randomNameSuffix();
+        Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
+        assertUpdate("CREATE TABLE " + tableName + " (key varchar, value integer)");
+        assertUpdate("INSERT INTO " + tableName + " VALUES ('one', 1)", 1);
+        assertUpdate("INSERT INTO " + tableName + " VALUES ('two', 2)", 1);
+        assertThat(query("SELECT sum(value), listagg(key, ' ') WITHIN GROUP (ORDER BY key) FROM " + tableName))
+                .matches("VALUES (BIGINT '3', VARCHAR 'one two')");
```
ebyhr: Remove unhelpful assertion.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6550,6 +6550,44 @@ public void testExpireSnapshotsPartitionedTable()
...
+        List<String> initialFiles = getAllDataFilesFromTableDirectory(tableName);
+
+        assertQuerySucceeds(sessionWithShortRetentionUnlocked, "ALTER TABLE " + tableName + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s', delete_files => false)");
+
+        List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(tableName);
+        assertThat(updatedDataFiles).hasSize(initialFiles.size());
```
ebyhr: Please verify the elements using `containsExactlyInAnyOrderElementsOf` instead.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
tbaeg: Added a section in the iceberg documentation.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/ExpireSnapshotsTableProcedure.java
tbaeg: Updated to be more clear.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: Added both.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: Done.

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -881,7 +881,7 @@ with the `retention_threshold` parameter.
 `expire_snapshots` can be run as follows:
 
 ```sql
-ALTER TABLE test_table EXECUTE expire_snapshots(retention_threshold => '7d');
+ALTER TABLE test_table EXECUTE expire_snapshots(retention_threshold => '7d', delete_files => true);
```
raunaqmorarka, line 887: I'd like to understand why this is needed. We don't see this in https://iceberg.apache.org/docs/latest/spark-procedures/#expire_snapshots
What is the "more efficient out-of-band" way of doing deletes ? Is there something can we improve about how we execute deletes in the current implementation to avoid the need to have a flag to disable it ?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
ebyhr: `hasSize` is redundant because `containsExactlyInAnyOrderElementsOf` fails if the size is different.

docs/src/main/sphinx/connector/iceberg.md
tbaeg: > I'd like to understand why this is needed. We don't see this in https://iceberg.apache.org/docs/latest/spark-procedures/#expire_snapshots

So a few quick reasons:

1. To meet data retention policy requirements.
2. Reduce coordinator work load pressure.
3. Flexibility in how table maintenance is performed.
4. `expire_snapshots` execution speed.

The main use case I am targeting is speeding up execution of `expire_snapshots` for a large table(s). 

For a small number of tables, with reasonable amount of data, there isn't a ton of benefit (hence the default is `true`). But I have found at certain scales (i.e. - petabytes of data and or trillions of rows), the call to `expire_snapshots` can take an extremely long time; most of the time taken spent on the deletion of files.

Additionally, depending on how many tables are being curated concurrently, there can be a fair bit of pressure on the coordinator.

With this parameter, we have more control over when we actually delete the underlying data while still benefiting from minimizing metadata for query planning.

> What is the "more efficient out-of-band" way of doing deletes ?

What is "more efficient" depends on underlying storage, but a simple example would be leveraging S3 bucket lifecycle policies. Assuming you had a time based retention policy for data in a table, the lifecycle policy would be a convenient and more efficient way to delete the files.

> Is there something can we improve about how we execute deletes in the current implementation to avoid the need to have a flag to disable it ?

Parallelizing/distributing the deletes would certainly be an improvement. I believe the current implementation is single threaded.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: Total right, fixed.

docs/src/main/sphinx/connector/iceberg.md
raunaqmorarka: > Assuming you had a time based retention policy for data in a table, the lifecycle policy would be a convenient and more efficient way to delete the files.

If we want to rely on lifecycle policy then we need to "soft delete" the files by adding a deleted tag to the files which the lifecycle policy can then act on. Relying on just time based policy has the danger of either deleting live files or keeping unnecessary files around for too long. Its not easy to keep the expiration interval of the table always aligned with a time based lifecycle policy.

> Parallelizing/distributing the deletes would certainly be an improvement. I believe the current implementation is single threaded.

Why don't we use a threadpool in `org.apache.iceberg.ExpireSnapshots#executeDeleteWith` to use more parallelism in deletes ?

tbaeg: > If we want to rely on lifecycle policy then we need to "soft delete" the files by adding a deleted tag to the files which the lifecycle policy can then act on. Relying on just time based policy has the danger of either deleting live files or keeping unnecessary files around for too long. Its not easy to keep the expiration interval of the table always aligned with a time based lifecycle policy.

Yeah, definitely some nuance I intentionally didn't touch on but in principal the point remains. 

> Why don't we use a threadpool in org.apache.iceberg.ExpireSnapshots#executeDeleteWith to use more parallelism in deletes ?

I think we should but the property still adds value for points 1, 2, and 3. I am also happy to throw something together to use a threadpool as well.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6550,6 +6550,41 @@ public void testExpireSnapshotsPartitionedTable()
         assertThat(updatedSnapshots.size()).isLessThan(initialSnapshots.size());
     }
 
+    @Test
+    public void testExpireSnapshotsKeepFiles()
+            throws Exception
+    {
+        String tableName = "test_expiring_snapshots_" + randomNameSuffix();
+        Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
+        assertUpdate("CREATE TABLE " + tableName + " (key varchar, value integer)");
```
chenjian2664: Can we use `newTrinoTable` to create test table?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6550,6 +6550,41 @@ public void testExpireSnapshotsPartitionedTable()
...
+        assertUpdate("INSERT INTO " + tableName + " VALUES ('one', 1)", 1);
+        assertUpdate("INSERT INTO " + tableName + " VALUES ('two', 2)", 1);
+        List<String> initialFiles = getAllDataFilesFromTableDirectory(tableName);
+
+        assertQuerySucceeds(sessionWithShortRetentionUnlocked, "ALTER TABLE " + tableName + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s', delete_files => false)");
+
+        List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(tableName);
+        assertThat(updatedDataFiles).containsExactlyInAnyOrderElementsOf(initialFiles);
```
chenjian2664: Could we add a verification step for the table data here

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: I was basing the test off of other tests but certainly can.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: I'm assuming you'd like a query to verify the table's data is accessible, and not the actual expired table's data files?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: Done.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: Done.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6550,6 +6550,44 @@ public void testExpireSnapshotsPartitionedTable()
...
+            Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES ('one', 1)", 1);
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES ('two', 2)", 1);
+            List<String> initialFiles = getAllDataFilesFromTableDirectory(table.getName());
+
+            assertQuerySucceeds(sessionWithShortRetentionUnlocked, "ALTER TABLE " + table.getName() + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s', delete_files => false)");
+
+            List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(table.getName());
+            assertThat(updatedDataFiles).containsExactlyInAnyOrderElementsOf(initialFiles);
+            assertThat(query("SELECT sum(value), listagg(key, ' ') WITHIN GROUP (ORDER BY key) FROM " + table.getName()))
```
chenjian2664, line 6566: We could use `assertThat(query("SELECT * FROM " + table.getName())) ...`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2184,6 +2186,7 @@ private void executeExpireSnapshots(ConnectorSession session, IcebergTableExecut
 
         // ForwardingFileIo handles bulk operations so no separate function implementation is needed
         table.expireSnapshots()
+                .cleanExpiredFiles(expireSnapshotsHandle.deleteFiles())
```
chenjian2664, line 2213: Will this make the table harder to clean up in the future, since the first maintenance task for removing expired snapshots is always slow?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6551,6 +6551,44 @@ public void testExpireSnapshotsPartitionedTable()
         assertThat(updatedSnapshots.size()).isLessThan(initialSnapshots.size());
     }
 
+    @Test
+    public void testExpireSnapshotsKeepFiles()
```
chenjian2664, line 6553: package priviate

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6551,6 +6551,44 @@ public void testExpireSnapshotsPartitionedTable()
...
+
+            List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(table.getName());
+            assertThat(updatedDataFiles).containsExactlyInAnyOrderElementsOf(initialFiles);
+            assertThat(query("SELECT sum(value), listagg(key, ' ') WITHIN GROUP (ORDER BY key) FROM " + table.getName()))
+                    .matches("VALUES (BIGINT '3', VARCHAR 'one two')");
+        }
+    }
+
+    @Test
+    public void testExpireSnapshotsPartitionedTableKeepFiles()
```
chenjian2664, line 6572: package private

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6551,6 +6551,44 @@ public void testExpireSnapshotsPartitionedTable()
...
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES(1, 100), (1, 101), (1, 102), (2, 200), (2, 201), (3, 300)", 6);
+            assertUpdate("DELETE FROM " + table.getName() + " WHERE col1 = 1", 3);
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES(4, 400)", 1);
+            List<String> initialFiles = getAllDataFilesFromTableDirectory(table.getName());
+
+            assertQuerySucceeds(sessionWithShortRetentionUnlocked, "ALTER TABLE " + table.getName() + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s', delete_files => false)");
+
+            List<String> updatedDataFiles = getAllDataFilesFromTableDirectory(table.getName());
+            assertThat(updatedDataFiles).containsExactlyInAnyOrderElementsOf(initialFiles);
+            assertQuery("SELECT sum(col2) FROM " + table.getName(), "SELECT 1101");
```
chenjian2664, line 6586: Please add a comment why the result is 1101 -- it's not obvious otherwise considering using the `SELECT col2 ..` to check each row

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: I tried to use the same testing assertions as `testExpireSnapshots`. I think we can clean up in a follow up.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
tbaeg: This is the same test as `testExpireSnapshotsPartitionedTable`. It also feels a redundant to list out and explain each sql query that was executed to reach this value. Could this be done in a follow up?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
tbaeg: I'm sure I am following what you mean, could clarify?

raunaqmorarka:
Is expire_snapshots still slow after recent improvements ?
I don't see an option to skip deletes in spark procedure https://iceberg.apache.org/docs/latest/spark-procedures/#expire_snapshots
My preference is we either find ways to make expire_snapshots/remove_orphans faster without skipping deletes.
Or we tag files as "deleted" similar to `org.apache.iceberg.aws.s3.S3FileIOProperties#DELETE_TAGS_PREFIX` and `org.apache.iceberg.aws.s3.S3FileIOProperties#DELETE_ENABLED` and let users configure tag based object lifecycle policies in S3.
Just skipping the deletes altogether and relying completely on time based lifecycle policy on storage layer doesn't sound safe to me as there is no guarantee that it won't delete live files.
Also, I don't think the solution can be tied to just expire_snapshots procedure, it should be generically targeting deletes in iceberg from any operation. I suspect that deletes from remove_orphans are more of a bottleneck than the ones from expire_snapshots.

tbaeg:
> Is expire_snapshots still slow after recent improvements ?

I haven't had a chance to test the improvements as we don't update unless it's targeting a release version.

> I don't see an option to skip deletes in spark procedure https://iceberg.apache.org/docs/latest/spark-procedures/#expire_snapshots My preference is we either find ways to make expire_snapshots/remove_orphans faster without skipping deletes. 

Not sure lack of support in spark is necessarily an indicator against adding this.

> Or we tag files as "deleted" similar to `org.apache.iceberg.aws.s3.S3FileIOProperties#DELETE_TAGS_PREFIX` and `org.apache.iceberg.aws.s3.S3FileIOProperties#DELETE_ENABLED` and let users configure tag based object lifecycle policies in S3. Just skipping the deletes altogether and relying completely on time based lifecycle policy on storage layer doesn't sound safe to me as there is no guarantee that it won't delete live files.

To be clear, I'm not saying time based policy is the end all be all; that was just a simple example to illustrate the general intent. I do think tagging is useful and worthwhile to implement. I can take a look to see what that might look like.

> Also, I don't think the solution can be tied to just expire_snapshots procedure, it should be generically targeting deletes in iceberg from any operation. I suspect that deletes from remove_orphans are more of a bottleneck than the ones from expire_snapshots.

Other than iterative adding features/options to prevent deletion, I'm not sure of a way to target it holistically. I do agree `remove_orphan_files` will be more of a bottleneck than others and I have a few ideas around implementing improvements.

All that said, it sounds like there needs to be a more use cases from the community. Will close this PR for now until there are other(s) that find this useful.

PR #25739, opened 2025-05-06
Extend BaseSystemTable in Iceberg history table

ebyhr:
## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #26217, opened 2025-07-16
Return only published main branch snapshots in history metadata table

maswin:
## Description

History table returns all snapshots right now. But it must only return published snapshots in main branch. 
Modified the implementation to iterate over history entry instead of all snapshots.

Iceberg/Spark reference implementation - https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/HistoryTable.java#L106

## Release notes

(X ) Release notes are required. Please propose a release note for me.

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/HistoryTable.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/HistoryTable.java
@@ -25,11 +25,14 @@ import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.SystemTable;
 import io.trino.spi.predicate.TupleDomain;
 import io.trino.spi.type.TimeZoneKey;
+import org.apache.iceberg.HistoryEntry;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.util.SnapshotUtil;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 import static io.trino.spi.type.BigintType.BIGINT;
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/HistoryTable.java
```diff
@@ -76,8 +79,13 @@ public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, Connect
 
         Set<Long> ancestorIds = ImmutableSet.copyOf(SnapshotUtil.currentAncestorIds(icebergTable));
         TimeZoneKey timeZoneKey = session.getTimeZoneKey();
-        for (Snapshot snapshot : icebergTable.snapshots()) {
-            long snapshotId = snapshot.snapshotId();
+        Map<Long, Snapshot> snapshots = new HashMap<>();
+        for (Snapshot snap : icebergTable.snapshots()) {
```
ebyhr: snap -> snapshot 

https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

ebyhr (commented):
Please add a test.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/HistoryTable.java
maswin: Done

maswin:
> Please add a test to `BaseIcebergSystemTables`.

Added

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergVendingRestCatalogConnectorSmokeTest.java
```diff
@@ -312,6 +312,14 @@ public void testDropTableWithNonExistentTableLocation()
                 .hasMessageMatching("Failed to load table: (.*)");
     }
 
+    @Test
+    @Override
+    public void testCreateOrReplaceTable()
+    {
+        // TODO: Enable this test once tabulario/iceberg-rest image is updated to 1.8.0+
```
ebyhr: I don't think they will release new docker images. Please avoid using "TODO" here.

maswin: Removed

ebyhr:
I think we should go ahead with #25739 instead.

maswin:
> I think we should go ahead with #25739 instead.

The other PR still requires the same set of changes

ebyhr:
Why? The PR makes $history table call `org.apache.iceberg.HistoryTable#convertHistoryEntryFunc` method.

maswin:
> Why? The PR makes $history table call `org.apache.iceberg.HistoryTable#convertHistoryEntryFunc` method.

Ahh ok. Just checked. This change won't be needed there except for the tests part maybe

ebyhr:
I included your tests in #25739 and added you as a co-author. Let me close this PR.

Issue #24518, opened 2024-12-18
Iceberg REST Catalog: Support for vended credentials for GCP


c-thiel:
Currently trino only implements vended credentials for S3 for IRC (https://github.com/trinodb/trino/pull/20186/files#diff-300020eb68920cea152529486b75fddcf41b4fa85c10be3ce893a17d4537881bR105-R106).

I created the same issue for Azure some time back: https://github.com/trinodb/trino/issues/23238

I am hoping to use this issue to track interest for GCP vended credential. It probably also helps users on GCP to notice faster that vended-creds are not implemented yet.

judnich:
Any updates on this? Credential vending is a pretty important part of Iceberg REST Catalog support, so I'm surprised to see Trino missing this feature on a major cloud provider for so long!

deverant:
The change to support the credential vending is actually pretty small, but I am reaching out to folks in Slack about how to handle testing / refreshing of those tokens if the query takes over an hour (normal lifetime of a GCP token).

PR #26219, opened 2025-07-17
Add support for BigLake metastore in Iceberg REST catalog

ebyhr:
## Description

Add `BIGLAKE` to `iceberg.rest-catalog.security` config property for supporting BigLake metastore. 
Also, this PR adds `iceberg.rest-catalog.google-project-id` config property. 

Relates to https://cloud.google.com/bigquery/docs/blms-rest-catalog
Supersedes #26054

## Release notes

```markdown
## Iceberg
* Add support for BigLake metastore in Iceberg REST catalog. ({issue}`26219`)
```

### DIFF
--- a/docs/src/main/sphinx/object-storage/metastores.md
+++ b/docs/src/main/sphinx/object-storage/metastores.md
@@ -488,7 +488,7 @@ following properties:
     `s3://my_bucket/warehouse_location`
 * - `iceberg.rest-catalog.security`
   - The type of security to use (default: `NONE`). Possible values are `NONE`, 
-    `SIGV4` or `OAUTH2`. `OAUTH2` requires either a `token` or `credential`.
+    `SIGV4`, `GOOGLE` or `OAUTH2`. `OAUTH2` requires either a `token` or a `credential`.
 * - `iceberg.rest-catalog.session`
   - Session information included when communicating with the REST Catalog.
     Options are `NONE` or `USER` (default: `NONE`).
@@ -522,6 +522,9 @@ following properties:
   - Enable view endpoints. Defaults to `true`.
 * - `iceberg.rest-catalog.signing-name`
   - AWS SigV4 signing service name. Defaults to `execute-api`.
+* - `iceberg.rest-catalog.google-project-id`
+  - Google Cloud project name. This property must be set when `iceberg.rest-catalog.security` 
+    config property is set to `GOOGLE`. Example: `development-123456`.
 * - `iceberg.rest-catalog.case-insensitive-name-matching`
   - Match namespace, table, and view names case insensitively. Defaults to `false`.
 * - `iceberg.rest-catalog.case-insensitive-name-matching.cache-ttl`
@@ -548,6 +551,22 @@ iceberg.rest-catalog.uri=https://dbc-12345678-9999.cloud.databricks.com/api/2.1/
 iceberg.security=read_only
 iceberg.rest-catalog.security=OAUTH2
 iceberg.rest-catalog.oauth2.token=***
+```
+
+`iceberg.rest-catalog.security` must be `GOOGLE` when connecting to BigLake metastore
+using an Iceberg REST catalog.
+
+```properties
+connector.name=iceberg
+iceberg.catalog.type=rest
+iceberg.unique-table-location=false
+iceberg.rest-catalog.warehouse=gs://example-bucket
+iceberg.rest-catalog.uri=https://biglake.googleapis.com/iceberg/v1beta/restcatalog
+iceberg.rest-catalog.security=GOOGLE
+iceberg.rest-catalog.google-project-id=example-project-id
+iceberg.rest-catalog.view-endpoints-enabled=false
+fs.native-gcs.enable=true
+gcs.json-key-file-path=/path/to/gcs_keyfile.json
 ```
 
 The REST catalog supports [view management](sql-view-management) 
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
@@ -1,0 +1,45 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg.catalog.rest;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.inject.Inject;
+
+import java.util.Map;
+
+import static org.apache.iceberg.gcp.auth.GoogleAuthManager.GCP_CREDENTIALS_PATH_PROPERTY;
+import static org.apache.iceberg.rest.auth.AuthProperties.AUTH_TYPE;
+import static org.apache.iceberg.rest.auth.AuthProperties.AUTH_TYPE_GOOGLE;
+
+public class GoogleAuthProperties
+        implements SecurityProperties
+{
+    private final Map<String, String> properties;
+
+    @Inject
+    public GoogleAuthProperties(GoogleSecurityConfig config)
+    {
+        properties = ImmutableMap.<String, String>builder()
+                .put(AUTH_TYPE, AUTH_TYPE_GOOGLE)
+                .put(GCP_CREDENTIALS_PATH_PROPERTY, config.getJsonKeyFilePath())
+                .put("header.x-goog-user-project", config.getProjectId())
+                .buildOrThrow();
+    }
+
+    @Override
+    public Map<String, String> get()
+    {
+        return properties;
+    }
+}
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
@@ -1,0 +1,55 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg.catalog.rest;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
+import io.airlift.configuration.validation.FileExists;
+import jakarta.validation.constraints.NotNull;
+
+public class GoogleSecurityConfig
+{
+    private String projectId;
+    private String jsonKeyFilePath;
+
+    @NotNull
+    public String getProjectId()
+    {
+        return projectId;
+    }
+
+    @Config("iceberg.rest-catalog.google-project-id")
+    @ConfigDescription("Google project ID")
+    public GoogleSecurityConfig setProjectId(String projectId)
+    {
+        this.projectId = projectId;
+        return this;
+    }
+
+    @NotNull
+    @FileExists
+    public String getJsonKeyFilePath()
+    {
+        return jsonKeyFilePath;
+    }
+
+    // Duplicated from GcsFileSystemConfig to enforce the property in BigLake metastore
+    @Config("gcs.json-key-file-path")
+    @ConfigDescription("JSON key file used to access Google Cloud Storage")
+    public GoogleSecurityConfig setJsonKeyFilePath(String jsonKeyFilePath)
+    {
+        this.jsonKeyFilePath = jsonKeyFilePath;
+        return this;
+    }
+}
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityModule.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityModule.java
@@ -1,0 +1,30 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg.catalog.rest;
+
+import com.google.inject.Binder;
+import io.airlift.configuration.AbstractConfigurationAwareModule;
+
+import static io.airlift.configuration.ConfigBinder.configBinder;
+
+public class GoogleSecurityModule
+        extends AbstractConfigurationAwareModule
+{
+    @Override
+    protected void setup(Binder binder)
+    {
+        configBinder(binder).bindConfig(GoogleSecurityConfig.class);
+        binder.bind(SecurityProperties.class).to(GoogleAuthProperties.class);
+    }
+}
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
@@ -24,10 +24,14 @@ import io.airlift.log.Logger;
 import io.jsonwebtoken.impl.DefaultJwtBuilder;
 import io.jsonwebtoken.jackson.io.JacksonSerializer;
 import io.trino.cache.EvictableCacheBuilder;
+import io.trino.filesystem.Location;
+import io.trino.filesystem.TrinoFileSystem;
 import io.trino.metastore.TableInfo;
 import io.trino.plugin.iceberg.ColumnIdentity;
+import io.trino.plugin.iceberg.IcebergFileSystemFactory;
 import io.trino.plugin.iceberg.IcebergUtil;
 import io.trino.plugin.iceberg.catalog.TrinoCatalog;
+import io.trino.plugin.iceberg.catalog.rest.IcebergRestCatalogConfig.Security;
 import io.trino.plugin.iceberg.catalog.rest.IcebergRestCatalogConfig.SessionType;
 import io.trino.spi.TrinoException;
 import io.trino.spi.catalog.CatalogName;
@@ -71,6 +75,7 @@ import org.apache.iceberg.view.ViewBuilder;
 import org.apache.iceberg.view.ViewRepresentation;
 import org.apache.iceberg.view.ViewVersion;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Date;
 import java.util.Iterator;
@@ -89,6 +94,7 @@ import static io.trino.cache.CacheUtils.uncheckedCacheGet;
 import static io.trino.filesystem.Locations.appendPath;
 import static io.trino.metastore.Table.TABLE_COMMENT;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_CATALOG_ERROR;
+import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_FILESYSTEM_ERROR;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_UNSUPPORTED_VIEW_DIALECT;
 import static io.trino.plugin.iceberg.IcebergSchemaProperties.LOCATION_PROPERTY;
 import static io.trino.plugin.iceberg.IcebergSchemaProperties.SUPPORTED_SCHEMA_PROPERTIES;
@@ -99,6 +105,7 @@ import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
 import static java.util.UUID.randomUUID;
+import static org.apache.iceberg.CatalogUtil.dropTableData;
 import static org.apache.iceberg.view.ViewProperties.COMMENT;
 
 public class TrinoRestCatalog
@@ -109,9 +116,11 @@ public class TrinoRestCatalog
     private static final int PER_QUERY_CACHE_SIZE = 1000;
     private static final String NAMESPACE_SEPARATOR = ".";
 
+    private final IcebergFileSystemFactory fileSystemFactory;
     private final RESTSessionCatalog restSessionCatalog;
     private final CatalogName catalogName;
     private final TypeManager typeManager;
+    private final Security security;
     private final SessionType sessionType;
     private final Map<String, String> credentials;
     private final boolean nestedNamespaceEnabled;
@@ -127,8 +136,10 @@ public class TrinoRestCatalog
             .build();
 
     public TrinoRestCatalog(
+            IcebergFileSystemFactory fileSystemFactory,
             RESTSessionCatalog restSessionCatalog,
             CatalogName catalogName,
+            Security security,
             SessionType sessionType,
             Map<String, String> credentials,
             boolean nestedNamespaceEnabled,
@@ -140,8 +151,10 @@ public class TrinoRestCatalog
             Cache<TableIdentifier, TableIdentifier> remoteTableMappingCache,
             boolean viewEndpointsEnabled)
     {
+        this.fileSystemFactory = requireNonNull(fileSystemFactory, "fileSystemFactory is null");
         this.restSessionCatalog = requireNonNull(restSessionCatalog, "restSessionCatalog is null");
         this.catalogName = requireNonNull(catalogName, "catalogName is null");
+        this.security = requireNonNull(security, "security is null");
         this.sessionType = requireNonNull(sessionType, "sessionType is null");
         this.credentials = ImmutableMap.copyOf(requireNonNull(credentials, "credentials is null"));
         this.nestedNamespaceEnabled = nestedNamespaceEnabled;
@@ -471,6 +484,44 @@ public class TrinoRestCatalog
     @Override
     public void dropTable(ConnectorSession session, SchemaTableName schemaTableName)
     {
+        if (security == Security.GOOGLE) {
+            purgeBigLakeTable(session, schemaTableName);
+        }
+        else {
+            purgeTable(session, schemaTableName);
+        }
+        invalidateTableCache(schemaTableName);
+        invalidateTableMappingCache(schemaTableName);
+    }
+
+    private void purgeBigLakeTable(ConnectorSession session, SchemaTableName schemaTableName)
+    {
+        BaseTable table = loadTable(session, schemaTableName);
+        unregisterTable(session, schemaTableName);
+        try {
+            // Explicitly remove data like TrinoGlueCatalog.dropTable since BigLake doesn't delete its data and metadata
+            dropTableData(table.io(), table.operations().current());
+        }
+        catch (RuntimeException e) {
+            // If the snapshot file is not found, an exception will be thrown by the dropTableData function.
+            // So log the exception and continue with deleting the table location
+            log.warn(e, "Failed to delete table data referenced by metadata");
+        }
+        deleteTableDirectory(fileSystemFactory.create(session.getIdentity(), table.io().properties()), schemaTableName, table.location());
+    }
+
+    private static void deleteTableDirectory(TrinoFileSystem fileSystem, SchemaTableName schemaTableName, String tableLocation)
+    {
+        try {
+            fileSystem.deleteDirectory(Location.of(tableLocation));
+        }
+        catch (IOException e) {
+            throw new TrinoException(ICEBERG_FILESYSTEM_ERROR, format("Failed to delete directory %s of the table %s", tableLocation, schemaTableName), e);
+        }
+    }
+
+    private void purgeTable(ConnectorSession session, SchemaTableName schemaTableName)
+    {
         try {
             if (!restSessionCatalog.purgeTable(convert(session), toRemoteTable(session, schemaTableName, true))) {
                 throw new TrinoException(ICEBERG_CATALOG_ERROR, "Failed to drop table '%s'".formatted(schemaTableName));
@@ -479,8 +530,6 @@ public class TrinoRestCatalog
         catch (RESTException e) {
             throw new TrinoException(ICEBERG_CATALOG_ERROR, "Failed to drop table '%s'".formatted(schemaTableName.getTableName()), e);
         }
-        invalidateTableCache(schemaTableName);
-        invalidateTableMappingCache(schemaTableName);
     }
 
     @Override
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
@@ -229,6 +229,18 @@ public abstract class BaseIcebergConnectorSmokeTest
     }
 
     @Test
+    public void testRecreateTableWithSameName()
+    {
+        String tableName = "test_recreate_table_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " AS SELECT 1 x, 'INDIA' y", 1);
+        assertQuery("SELECT * FROM " + tableName, "VALUES (1, 'INDIA')");
+        assertUpdate("DROP TABLE " + tableName);
+        assertUpdate("CREATE TABLE " + tableName + " AS SELECT 'Trino' data", 1);
+        assertQuery("SELECT * FROM " + tableName, "VALUES ('Trino')");
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
+    @Test
     public void testRegisterTableWithTableLocation()
     {
         String tableName = "test_register_table_with_table_location_" + randomNameSuffix();
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
@@ -1,0 +1,328 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg.catalog.rest;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import io.airlift.json.ObjectMapperProvider;
+import io.trino.filesystem.FileIterator;
+import io.trino.filesystem.Location;
+import io.trino.plugin.iceberg.BaseIcebergConnectorSmokeTest;
+import io.trino.plugin.iceberg.IcebergConfig;
+import io.trino.plugin.iceberg.IcebergConnector;
+import io.trino.plugin.iceberg.IcebergQueryRunner;
+import io.trino.plugin.iceberg.SchemaInitializer;
+import io.trino.plugin.iceberg.catalog.TrinoCatalog;
+import io.trino.plugin.iceberg.catalog.TrinoCatalogFactory;
+import io.trino.spi.connector.SchemaTableName;
+import io.trino.testing.QueryRunner;
+import io.trino.testing.TestingConnectorBehavior;
+import org.apache.iceberg.BaseTable;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Base64;
+import java.util.List;
+
+import static io.trino.plugin.iceberg.IcebergTestUtils.checkParquetFileSorting;
+import static io.trino.testing.SystemEnvironmentUtils.requireEnv;
+import static io.trino.testing.TestingNames.randomNameSuffix;
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assumptions.abort;
+
+final class TestIcebergBigLakeMetastoreConnectorSmokeTest
+        extends BaseIcebergConnectorSmokeTest
+{
+    private static final String SCHEMA = "test_iceberg_biglake_" + randomNameSuffix();
+    private static final String GCP_STORAGE_BUCKET = requireEnv("GCP_STORAGE_BUCKET");
+    private static final byte[] GCS_JSON_KEY_BYTES = Base64.getDecoder().decode(requireEnv("GCP_CREDENTIALS_KEY"));
+    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapperProvider().get();
+
+    public TestIcebergBigLakeMetastoreConnectorSmokeTest()
+    {
+        super(new IcebergConfig().getFileFormat().toIceberg());
+    }
+
+    @Override
+    protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
+    {
+        return switch (connectorBehavior) {
+            case SUPPORTS_CREATE_MATERIALIZED_VIEW,
+                 SUPPORTS_RENAME_MATERIALIZED_VIEW,
+                 SUPPORTS_RENAME_SCHEMA,
+                 SUPPORTS_RENAME_TABLE -> false;
+            default -> super.hasBehavior(connectorBehavior);
+        };
+    }
+
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        Path gcpCredentialsFile = Files.createTempFile("gcp-credentials", ".json");
+        gcpCredentialsFile.toFile().deleteOnExit();
+        Files.write(gcpCredentialsFile, GCS_JSON_KEY_BYTES);
+        String projectId = OBJECT_MAPPER.readTree(GCS_JSON_KEY_BYTES).get("project_id").asText();
+
+        return IcebergQueryRunner.builder(SCHEMA)
+                .addIcebergProperty("iceberg.file-format", format.name())
+                .addIcebergProperty("iceberg.register-table-procedure.enabled", "true")
+                .addIcebergProperty("iceberg.unique-table-location", "false")
+                .addIcebergProperty("iceberg.catalog.type", "rest")
+                .addIcebergProperty("iceberg.rest-catalog.uri", "https://biglake.googleapis.com/iceberg/v1beta/restcatalog")
+                .addIcebergProperty("iceberg.rest-catalog.warehouse", "gs://" + GCP_STORAGE_BUCKET)
+                .addIcebergProperty("iceberg.rest-catalog.security", "GOOGLE")
+                .addIcebergProperty("iceberg.rest-catalog.google-project-id", projectId)
+                .addIcebergProperty("iceberg.rest-catalog.view-endpoints-enabled", "false")
+                .addIcebergProperty("iceberg.writer-sort-buffer-size", "1MB")
+                .addIcebergProperty("iceberg.allowed-extra-properties", "write.metadata.delete-after-commit.enabled,write.metadata.previous-versions-max")
+                .addIcebergProperty("fs.native-gcs.enabled", "true")
+                .addIcebergProperty("gcs.json-key-file-path", gcpCredentialsFile.toString())
+                .setSchemaInitializer(SchemaInitializer.builder()
+                        .withSchemaName(SCHEMA)
+                        .withClonedTpchTables(REQUIRED_TPCH_TABLES)
+                        .withSchemaProperties(ImmutableMap.of("location", "'gs://%s/%s'".formatted(GCP_STORAGE_BUCKET, SCHEMA)))
+                        .build())
+                .build();
+    }
+
+    @AfterAll
+    void cleanup()
+            throws Exception
+    {
+        // Avoid DROP SCHEMA CASCADE since BigLake metastore returns null when loading a table with missing metadata file by testDropTableWithMissingMetadataFile
+        // TODO Use DROP SCHEMA CASCADE once the above issue is fixed
+        fileSystem.deleteDirectory(Location.of(schemaPath()));
+        for (Object tableName : computeActual("SHOW TABLES IN " + SCHEMA).getOnlyColumnAsSet()) {
+            dropTableFromCatalog((String) tableName);
+        }
+        getQueryRunner().execute("DROP SCHEMA " + SCHEMA);
+    }
+
+    @Override
+    protected String getMetadataLocation(String tableName)
+    {
+        TrinoCatalogFactory catalogFactory = ((IcebergConnector) getQueryRunner().getCoordinator().getConnector("iceberg")).getInjector().getInstance(TrinoCatalogFactory.class);
+        TrinoCatalog trinoCatalog = catalogFactory.create(getSession().getIdentity().toConnectorIdentity());
+        BaseTable table = trinoCatalog.loadTable(getSession().toConnectorSession(), new SchemaTableName(getSession().getSchema().orElseThrow(), tableName));
+        return table.operations().current().metadataFileLocation();
+    }
+
+    @Override
+    protected String schemaPath()
+    {
+        return "gs://%s/%s".formatted(GCP_STORAGE_BUCKET, SCHEMA);
+    }
+
+    @Override
+    protected boolean locationExists(String location)
+    {
+        try {
+            return fileSystem.newInputFile(Location.of(location)).exists();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    @Override
+    protected boolean isFileSorted(Location path, String sortColumnName)
+    {
+        return checkParquetFileSorting(fileSystem.newInputFile(path), sortColumnName);
+    }
+
+    @Override
+    protected void deleteDirectory(String location)
+    {
+        try {
+            fileSystem.deleteDirectory(Location.of(location));
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+    }
+
+    @Override
+    protected void dropTableFromCatalog(String tableName)
+    {
+        assertUpdate("CALL system.unregister_table(CURRENT_SCHEMA, '" + tableName + "')");
+    }
+
+    @Test
+    @Override // Override because the location pattern differs
+    public void testShowCreateTable()
+    {
+        assertThat((String) computeScalar("SHOW CREATE TABLE region"))
+                .matches("CREATE TABLE iceberg." + SCHEMA + ".region \\(\n" +
+                        "   regionkey bigint,\n" +
+                        "   name varchar,\n" +
+                        "   comment varchar\n" +
+                        "\\)\n" +
+                        "WITH \\(\n" +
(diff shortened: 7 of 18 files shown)


rambleraptor:
This looks fantastic! Thank you for writing this!

niger-prequel:
Excited to see this get merged

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
```diff
@@ -0,0 +1,47 @@
...
+        implements SecurityProperties
+{
+    private final Map<String, String> properties;
+
+    @Inject
+    public GoogleAuthProperties(GoogleSecurityConfig config, GcsFileSystemConfig gcsConfig)
+    {
+        properties = ImmutableMap.<String, String>builder()
+                .put(AUTH_TYPE, AUTH_TYPE_GOOGLE)
+                .put(GCP_CREDENTIALS_PATH_PROPERTY, requireNonNull(gcsConfig.getJsonKeyFilePath(), "gcs.json-key-file-path config property is required"))
```
regadas: this being required would it still allow auth pass-through scenarios? 🤔

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -488,7 +488,7 @@ following properties:
     `s3://my_bucket/warehouse_location`
 * - `iceberg.rest-catalog.security`
   - The type of security to use (default: `NONE`). Possible values are `NONE`, 
-    `SIGV4` or `OAUTH2`. `OAUTH2` requires either a `token` or `credential`.
+    `SIGV4`, `GOOGLE` or `OAUTH2`. `OAUTH2` requires either a `token` or `credential`.
```
ebyhr: I used `GOOGLE` for consistency with Iceberg: https://github.com/apache/iceberg/blob/059310ead702814e5d95116210b9af77b29f6b94/core/src/main/java/org/apache/iceberg/rest/auth/AuthProperties.java#L31, but `BIGLAKE` might be better. Let me know if you have a better name.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +480,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
...
+        try {
+            // Explicitly remove data like TrinoGlueCatalog.dropTable since BigLake doesn't delete its data and metadata
+            dropTableData(table.io(), table.operations().current());
+        }
+        catch (RuntimeException e) {
+            // If the snapshot file is not found, an exception will be thrown by the dropTableData function.
+            // So log the exception and continue with deleting the table location
+            log.warn(e, "Failed to delete table data referenced by metadata");
+        }
+    }
```
ebyhr: @talatuyarer @rambleraptor Is there any way to make the BigLake metastore delete files forcibly? We'd like to avoid adding this logic on our side.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,369 @@
...
+                        .withClonedTpchTables(REQUIRED_TPCH_TABLES)
+                        .withSchemaProperties(ImmutableMap.of("location", "'gs://%s/%s'".formatted(GCP_STORAGE_BUCKET, SCHEMA)))
+                        .build())
+                .build();
+    }
+
+    @AfterAll
+    void cleanup()
+    {
+        getQueryRunner().execute("DROP SCHEMA IF EXISTS " + SCHEMA + " CASCADE");
```
ebyhr: I will fix this failure tomorrow.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
```diff
@@ -0,0 +1,47 @@
...
+                .put("header.x-goog-user-project", config.getProjectId())
```
findinpath, line 36: Is this property optional?

Addtional context: skimming https://docs.cloud.google.com/biglake/docs/blms-rest-catalog

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
```diff
@@ -0,0 +1,37 @@
...
+    private String projectId;
+
+    @NotNull
+    public String getProjectId()
+    {
+        return projectId;
+    }
+
+    @Config("iceberg.rest-catalog.google-project-id")
+    @ConfigDescription("Google project ID")
```
findinpath, line 33:
```suggestion
    @ConfigDescription("Google project name")
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
findinpath: probably the map building will fail when adding `null` with a less clear exception cause

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityModule.java
```diff
@@ -0,0 +1,32 @@
...
+import static io.airlift.configuration.ConfigBinder.configBinder;
+
+public class GoogleSecurityModule
+        extends AbstractConfigurationAwareModule
+{
+    @Override
+    protected void setup(Binder binder)
+    {
+        configBinder(binder).bindConfig(GoogleSecurityConfig.class);
+        configBinder(binder).bindConfig(GcsFileSystemConfig.class);
```
findinpath: can be removed - it comes through `FileSystemModule`

findinpath: Are we creating a strong binding between this authentication and `fs.native-gcs.enabled` ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
```diff
@@ -0,0 +1,47 @@
...
+import static org.apache.iceberg.rest.auth.AuthProperties.AUTH_TYPE;
+import static org.apache.iceberg.rest.auth.AuthProperties.AUTH_TYPE_GOOGLE;
+
+public class GoogleAuthProperties
+        implements SecurityProperties
+{
+    private final Map<String, String> properties;
+
+    @Inject
+    public GoogleAuthProperties(GoogleSecurityConfig config, GcsFileSystemConfig gcsConfig)
```
findinpath: I understand the need for the project ID , but don't quite support having a dependency towards `GcsFileSystemConfig` in this class.
Pls investigate alternative options.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -411,6 +416,11 @@ public Transaction newCreateTableTransaction(
                 // TODO Replace with createTransaction once S3 Tables supports stage-create option
                 return tableBuilder.create().newTransaction();
             }
+            if (security == Security.GOOGLE) {
```
findinpath:
```suggestion
            if (isBigLake()) {
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +480,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
 
     @Override
     public void dropTable(ConnectorSession session, SchemaTableName schemaTableName)
+    {
+        if (isBigLake()) {
+            dropBigLakeTable(session, schemaTableName);
+        }
+        else {
+            purgeTable(session, schemaTableName);
```
findinpath, line 491: `purgeTable(session, schemaTableName);` seems to be in both code paths.
Is it intentional that you're first need loading the table in `dropBigLakeTable(...)` method before calling `purgeTable(...)`?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
findinpath: > Drop a table and request that files are immediately deleted.

From `org.apache.iceberg.catalog.SessionCatalog#purgeTable` javadoc.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1304,7 +1305,7 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
         Location location = Location.of(transaction.table().location());
         try {
             // S3 Tables internally assigns a unique location for each table
```
findinpath, line 1313: adjust the comment to not be anymore s3 tables specific

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -390,7 +390,7 @@ public void testRegisterTableWithMetadataFile()
     public void testCreateTableWithTrailingSpaceInLocation()
     {
         String tableName = "test_create_table_with_trailing_space_" + randomNameSuffix();
-        String tableLocationWithTrailingSpace = schemaPath() + tableName + " ";
+        String tableLocationWithTrailingSpace = schemaPath() + "/" + tableName + " ";
```
findinpath, line 406: nit: can be a separate commit.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
```diff
@@ -0,0 +1,37 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg.catalog.rest;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
+import jakarta.validation.constraints.NotNull;
```
findinpath, line 19: https://docs.cloud.google.com/biglake/docs/blms-rest-catalog

> This example doesn't use credential vending. To use credential vending, you must use a [catalog in credential vending mode](https://docs.cloud.google.com/biglake/docs/blms-rest-catalog#create_a_catalog) and add the X-Iceberg-Access-Delegation header to Iceberg REST catalog requests with a value of vended-credentials by adding the following line to the SparkSession builder


Is credentials vending thought to be rather a follow-up ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
regadas: `header.x-goog-user-project` is not optional; required for proper auth with catalog

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
ebyhr: This should be a project ID. I'm not sure why name is suggested here.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
ebyhr: Yes, it can be handled in #24518

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
ebyhr: That's intentional since we don't need to load the table when the catalog isn't BigLake metastore.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleAuthProperties.java
ebyhr: > probably the map building will fail when adding null with a less clear exception cause

The bootstrap fails before reaching this line, with meaningful message like: 
```
Error: Invalid configuration property iceberg.rest-catalog.google-project-id: must not be null (for class GoogleSecurityConfig.projectId)
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
talatuyarer: @ebyhr currently we dont have a way. BigLake Metastore does not manage user's files, only metadata.

docs/src/main/sphinx/object-storage/metastores.md
findinpath:
```suggestion
    `SIGV4`, `GOOGLE` or `OAUTH2`. `OAUTH2` requires either a `token` or a `credential`.
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +480,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
...
+        }
+        invalidateTableCache(schemaTableName);
+        invalidateTableMappingCache(schemaTableName);
+    }
+
+    public void dropBigLakeTable(ConnectorSession session, SchemaTableName schemaTableName)
```
findinpath: nit:private - used only within `TrinoRestCatalog`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -411,6 +416,11 @@ public Transaction newCreateTableTransaction(
                 // TODO Replace with createTransaction once S3 Tables supports stage-create option
                 return tableBuilder.create().newTransaction();
             }
+            if (isBigLake()) {
+                // Specifying location results in "Malformed request: The table `location` property can only point to the default location"
+                // even though the generated path is the same as the specific path.
+                return tableBuilder.createTransaction();
+            }
```
findinpath:
```
Caused by: org.apache.iceberg.exceptions.BadRequestException: Malformed request: The table `location` property can only point to the default location: gs://trino-ci-test/test_iceberg_biglake_1z7kegagbw/nation.
```

With `iceberg.unique-table-location` set to `false` the tests in `TestIcebergBigLakeMetastoreConnectorSmokeTest` succeed also without this `if` branch.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,377 @@
...
+        Path gcpCredentialsFile = Files.createTempFile("gcp-credentials", ".json");
+        gcpCredentialsFile.toFile().deleteOnExit();
+        Files.write(gcpCredentialsFile, GCS_JSON_KEY_BYTES);
+        String projectId = OBJECT_MAPPER.readTree(GCS_JSON_KEY_BYTES).get("project_id").asText();
+
+        return IcebergQueryRunner.builder(SCHEMA)
+                .addIcebergProperty("iceberg.file-format", format.name())
+                .addIcebergProperty("iceberg.register-table-procedure.enabled", "true")
+                .addIcebergProperty("iceberg.catalog.type", "rest")
+                .addIcebergProperty("iceberg.rest-catalog.uri", "https://biglake.googleapis.com/iceberg/v1beta/restcatalog")
```
findinpath, line 90: Can we document in https://trino.io/docs/current/object-storage/metastores.html#iceberg-rest-catalog how to connect to BigLake?
It may not be quite straightforward to figure out the minimal configuration setup needed if someone does not dive in the source code of `trinodb/trino`

FWIW, the same comment applies for other Iceberg REST Catalog integrations (e.g. : S3Tables, Unity, ...).

findinpath (approved):
LGTM % https://github.com/trinodb/trino/pull/26219#discussion_r2562798356

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
ebyhr: Added an example to metastores.md. We already have Unity example. There is no other examples, though. 
https://github.com/trinodb/trino/blob/36b1485f3ffb6bc392df8d256a05c42d866e70b0/docs/src/main/sphinx/object-storage/metastores.md?plain=1#L541-L551

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -550,6 +552,22 @@ iceberg.rest-catalog.security=OAUTH2
...
 ```
 
+`iceberg.rest-catalog.security` must be `GOOGLE` when connecting to BigLake metastore
+using an Iceberg REST catalog:
+
+```properties
+connector.name=iceberg
+iceberg.catalog.type=rest
+iceberg.unique-table-location=false
+iceberg.rest-catalog.warehouse=gs://exampl-bucket
```
findinpath: exampl typo

ebyhr:
I guess BigLake metastore created a new metadata file in the background:
```
Error:  Errors: 
Error:    TestIcebergBigLakeMetastoreConnectorSmokeTest>BaseIcebergConnectorSmokeTest.testRegisterTableWithComments:268->AbstractTestQueryFramework.assertUpdate:411->AbstractTestQueryFramework.assertUpdate:416 » 
QueryFailed More than one latest metadata file found at location: gs://trino-ci-test/test_iceberg_biglake_gk2j0xnz67/test_register_table_with_comments_xyo3k42eom/metadata, latest metadata files are [
gs://trino-ci-test/test_iceberg_biglake_gk2j0xnz67/test_register_table_with_comments_xyo3k42eom/metadata/00005-6913e897-0000-2d46-9aaa-14223bce81ce.metadata.json, 
gs://trino-ci-test/test_iceberg_biglake_gk2j0xnz67/test_register_table_with_comments_xyo3k42eom/metadata/00005-69218711-0000-2ea5-81a1-c82add7ff230.metadata.json]
```
https://github.com/trinodb/trino/actions/runs/19693222597/job/56413468126?pr=26219

@talatuyarer Is my understanding correct? Any way to disable the background task?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
```diff
@@ -0,0 +1,55 @@
...
+
+    @NotNull
+    @FileExists
+    public String getJsonKeyFilePath()
+    {
+        return jsonKeyFilePath;
+    }
+
+    // Duplicated from GcsFileSystemConfig to enforce the property in BigLake metastore
+    @Config("gcs.json-key-file-path")
```
chenjian2664, line 48: Is it the only supported authenticated way now, do we support json string?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +475,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
...
+    private void dropBigLakeTable(ConnectorSession session, SchemaTableName schemaTableName)
+    {
+        BaseTable table = loadTable(session, schemaTableName);
+        purgeTable(session, schemaTableName);
+        try {
+            // Explicitly remove data like TrinoGlueCatalog.dropTable since BigLake doesn't delete its data and metadata
+            dropTableData(table.io(), table.operations().current());
+        }
+        catch (RuntimeException e) {
+            // If the snapshot file is not found, an exception will be thrown by the dropTableData function.
```
chenjian2664, line 506: What's the exception here, FileNotFound or what, is it safe to declare that we are success here, no matter what kinda of the exception

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -550,6 +552,22 @@ iceberg.rest-catalog.security=OAUTH2
 iceberg.rest-catalog.oauth2.token=***
 ```
 
+`iceberg.rest-catalog.security` must be `GOOGLE` when connecting to BigLake metastore
```
chenjian2664, line 556: Do we have the place to see each metastore's functionality details, i,e support "rename" or not

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,375 @@
...
+    @AfterAll
+    void cleanup()
+            throws Exception
+    {
+        // Avoid DROP SCHEMA CASCADE since BigLake metastore returns null when loading a table with missing metadata file
+        fileSystem.deleteDirectory(Location.of(schemaPath()));
+        for (Object tableName : computeActual("SHOW TABLES IN " + SCHEMA).getOnlyColumnAsSet()) {
+            dropTableFromMetastore((String) tableName);
+        }
+        getQueryRunner().execute("DROP SCHEMA " + SCHEMA);
```
chenjian2664, line 117: assertUpdate ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
ebyhr: We don't support json string since it's not supported by Iceberg library. We will have to write json string somewhere - it's unsafe.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
ebyhr: We want to suppress any exceptions since the table has been already dropped.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
ebyhr: This is not a test method. No need to assert anything.

docs/src/main/sphinx/object-storage/metastores.md
ebyhr: We can document such limitations here. It can be handled as a follow-up in my opinion.

Noremac201:
> I guess BigLake metastore created a new metadata file in the background:
> 
> ```
> Error:  Errors: 
> Error:    TestIcebergBigLakeMetastoreConnectorSmokeTest>BaseIcebergConnectorSmokeTest.testRegisterTableWithComments:268->AbstractTestQueryFramework.assertUpdate:411->AbstractTestQueryFramework.assertUpdate:416 » 
> QueryFailed More than one latest metadata file found at location: gs://trino-ci-test/test_iceberg_biglake_gk2j0xnz67/test_register_table_with_comments_xyo3k42eom/metadata, latest metadata files are [
> gs://trino-ci-test/test_iceberg_biglake_gk2j0xnz67/test_register_table_with_comments_xyo3k42eom/metadata/00005-6913e897-0000-2d46-9aaa-14223bce81ce.metadata.json, 
> gs://trino-ci-test/test_iceberg_biglake_gk2j0xnz67/test_register_table_with_comments_xyo3k42eom/metadata/00005-69218711-0000-2ea5-81a1-c82add7ff230.metadata.json]
> ```
> 
> https://github.com/trinodb/trino/actions/runs/19693222597/job/56413468126?pr=26219
> 
> @talatuyarer Is my understanding correct? Any way to disable the background task?

Hey! So what happened here is that the last `ADD COMMENT` call returned a 429, but it looks like BigLake metastore leaked the metadata file that was meant to be persisted, and when the retry came from the client side, had two metadata files of version `00005`. 

So BigLake wasn't creating in the background per-se, and this should only rarely happen when the last update before a `RegisterTable` hits a retryable error. 

In any case, that should be fixed, but IMO shouldn't block submission

ebyhr:
@Noremac201 Thank you! To clarify, does "that should be fixed" refer to Trino or BigLake metastore?

> IMO shouldn’t block submission

Unfortunately, we can't merge this PR while we know there are flaky tests.
If resolving the root cause will take time, an alternative would be to temporarily disallow the `register_table` procedure for the BigLake metastore integration.

Noremac201:
Ah I see.

I meant it should be fixed on the BigLake side, ideally we're not leaking any metadata files if the transaction doesn't commit on the server side.

Is it possible to just disable that one test -- `..._with_comments`? The flow of the test with the ~5 updates in succession is triggering the 429, otherwise I'll leave it to @talatuyarer as to whether to disable register table for the time being.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
findinpath: Follow-up https://github.com/apache/iceberg/pull/14713

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1310,8 +1311,8 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
         transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties);
         Location location = Location.of(transaction.table().location());
         try {
-            // S3 Tables internally assigns a unique location for each table
-            if (!isS3Tables(location.toString())) {
+            // S3 Tables and BigLake metastore internally assign a unique location for each table
```
findinpath: I think that the comment is not fully correct with regards to BigLake metastore

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/GoogleSecurityConfig.java
findinpath, line 48: Maybe it is worth mentioning in the docs that this functionality works currently only with the native gcs file system client (`fs.native-gcs.enable=true`)

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -550,6 +552,22 @@ iceberg.rest-catalog.security=OAUTH2
...
+using an Iceberg REST catalog:
+
+```properties
+connector.name=iceberg
+iceberg.catalog.type=rest
+iceberg.unique-table-location=false
```
findinpath, line 562: Maybe it would be worthwhile to have a test like the following

```
    @Test
    public void testRecreateTableWithSameName()
    {
        String tableName = "test_recreate_table_" + randomNameSuffix();
        assertUpdate("CREATE TABLE " + tableName + " AS SELECT 1 x, 'INDIA' y", 1);
        assertQuery("SELECT * FROM " + tableName, "VALUES (1, 'INDIA')");
        assertUpdate("CREATE TABLE " + tableName + " AS SELECT 'Trino' data", 1);
        assertQuery("SELECT * FROM " + tableName, "VALUES ('Trino')");
    }
```

I don't know whether it is relevant for `BaseIcebergConnectorSmokeTest` , but I see rationale in having it done at least for `TestIcebergBigLakeMetastoreConnectorSmokeTest`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,385 @@
...
+                        .withSchemaProperties(ImmutableMap.of("location", "'gs://%s/%s'".formatted(GCP_STORAGE_BUCKET, SCHEMA)))
+                        .build())
+                .build();
+    }
+
+    @AfterAll
+    void cleanup()
+            throws Exception
+    {
+        // Avoid DROP SCHEMA CASCADE since BigLake metastore returns null when loading a table with missing metadata file
```
findinpath: > BigLake metastore returns null when loading a table with missing metadata file

It is rather unclear what this means.
Is the `DROP CASCADE` statement going to fail then (is yes, how?) or will it not remove every file?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,385 @@
...
+    @Test
+    @Override
+    public void testCommentView()
+    {
+        assertThatThrownBy(super::testCommentView)
+                .hasMessageContaining("Server does not support endpoint: POST /v1/{prefix}/namespaces/{namespace}/views");
+    }
+
+    @Test
+    @Override // TODO Enable once flaky issue is fixed
```
findinpath: > // TODO Enable once flaky issue is fixed


It is unclear what flakiness is being made reference towards here.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
findinpath: maybe worth referring towards `testDropTableWithMissingMetadataFile()` method

findinpath: I see the exception `io.trino.spi.TrinoException: Metadata not found in metadata location for table test_iceberg_biglake_lyjynonghf.test_drop_table_with_missing_manifest_list_file_rhwxzmu3ww` while trying to select from the table.

findinpath (approved):
Excited to see this new integration of IRC landing in Trino ❤️

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
findinpath: nit: I see in the codebase that we use abort in such cases.

`abort(....);`

docs/src/main/sphinx/object-storage/metastores.md
chenjian2664: Is the purpose for identifying the "unique" location or just test "drop table"?

ebyhr:
Rebased on master without any changes.

findepi:
is the build failure related?

ebyhr:
No, it's Databricks CI failure #27642

ebyhr:
@wendigo @findepi Could you please review this PR when you have time?

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -522,6 +522,8 @@ following properties:
   - Enable view endpoints. Defaults to `true`.
 * - `iceberg.rest-catalog.signing-name`
   - AWS SigV4 signing service name. Defaults to `execute-api`.
+* - `iceberg.rest-catalog.google-project-id`
+  - Google Cloud project name. Example: `development-123456`.
```
findepi: Worth saying here that's applicable when `iceberg.rest-catalog.security=google`?

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -550,6 +552,23 @@ iceberg.rest-catalog.security=OAUTH2
 iceberg.rest-catalog.oauth2.token=***
 ```
 
+`iceberg.rest-catalog.security` must be `GOOGLE` when connecting to BigLake metastore
+using an Iceberg REST catalog. Only native GCS file system is supported. The
+legacy Hadoop GCS file system is not supported.
```
findepi: interesting, why

docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -488,7 +488,7 @@ following properties:
     `s3://my_bucket/warehouse_location`
 * - `iceberg.rest-catalog.security`
   - The type of security to use (default: `NONE`). Possible values are `NONE`, 
-    `SIGV4` or `OAUTH2`. `OAUTH2` requires either a `token` or `credential`.
+    `SIGV4`, `GOOGLE` or `OAUTH2`. `OAUTH2` requires either a `token` or a `credential`.
```
findepi, line 491: is "GOOGLE" unambiguous enough or already used for the same concept in a different place?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +475,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
...
+            // So log the exception and continue with deleting the table location
+            log.warn(e, "Failed to delete table data referenced by metadata");
+        }
+    }
+
+    public void purgeTable(ConnectorSession session, SchemaTableName schemaTableName)
```
findepi: private

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +475,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
...
+        invalidateTableMappingCache(schemaTableName);
+    }
+
+    private void dropBigLakeTable(ConnectorSession session, SchemaTableName schemaTableName)
+    {
+        BaseTable table = loadTable(session, schemaTableName);
+        purgeTable(session, schemaTableName);
+        try {
+            // Explicitly remove data like TrinoGlueCatalog.dropTable since BigLake doesn't delete its data and metadata
+            dropTableData(table.io(), table.operations().current());
```
findepi, line 503: After `purgeTable` above, this is likely too late to try delete the table from our side. The manifest are gone perhaps?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +475,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
...
+        }
+        catch (RuntimeException e) {
+            // If the snapshot file is not found, an exception will be thrown by the dropTableData function.
+            // So log the exception and continue with deleting the table location
```
findepi, line 507: Where do we "continue with deleting the table location"?
Was it supposed to be followed by a call to `AbstractTrinoCatalog#deleteTableDirectory`?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -470,6 +475,33 @@ public void unregisterTable(ConnectorSession session, SchemaTableName tableName)
 
     @Override
     public void dropTable(ConnectorSession session, SchemaTableName schemaTableName)
+    {
+        if (isBigLake()) {
+            dropBigLakeTable(session, schemaTableName);
+        }
+        else {
+            purgeTable(session, schemaTableName);
+        }
```
findepi: Why these two functions -- doing logically equivalent thing -- have so different names?

let's find more similar names. Maybe just dropBigLakeTable -> pruneBigLakeTable

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2874,6 +2875,10 @@ public TableStatisticsMetadata getStatisticsCollectionMetadataForWrite(Connector
         if (!isExtendedStatisticsEnabled(session) || !isCollectExtendedStatisticsOnWrite(session)) {
             return TableStatisticsMetadata.empty();
         }
+        if (isBigLakeMetastore()) {
+            log.debug("BigLake metastore does not support statistics: %s", tableMetadata.getTableSchema().getTable());
+            return TableStatisticsMetadata.empty();
```
findepi: What happens if we collect stats and ask catalog to persist them? is it they simply drop the information?

Is there a public tracker link for the missing feature that we could link here?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2988,6 +2996,9 @@ public ConnectorTableHandle beginStatisticsCollection(ConnectorSession session,
         if (isS3Tables(icebergTable.location())) {
             throw new TrinoException(NOT_SUPPORTED, "S3 Tables do not support analyze");
         }
+        if (isBigLakeMetastore()) {
+            throw new TrinoException(NOT_SUPPORTED, "BigLake metastore does not support analyze");
```
findepi: BigLake metastore -> BigLake catalog

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+import static io.trino.testing.TestingConnectorSession.SESSION;
+import static io.trino.testing.TestingNames.randomNameSuffix;
+import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.junit.jupiter.api.Assumptions.abort;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@TestInstance(PER_CLASS)
```
findepi: isn't that inherited from the parent class? this should apply to all ATQF subclasses automatically

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+                        .withSchemaProperties(ImmutableMap.of("location", "'gs://%s/%s'".formatted(GCP_STORAGE_BUCKET, SCHEMA)))
+                        .build())
+                .build();
+    }
+
+    @AfterAll
+    void cleanup()
+            throws Exception
+    {
+        // Avoid DROP SCHEMA CASCADE since BigLake metastore returns null when loading a table with missing metadata file by testDropTableWithMissingMetadataFile
```
findepi, line 111: Sounds like a problem we should fix in product code.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+        fileSystem.deleteDirectory(Location.of(schemaPath()));
+        for (Object tableName : computeActual("SHOW TABLES IN " + SCHEMA).getOnlyColumnAsSet()) {
+            dropTableFromCatalog((String) tableName);
+        }
+        getQueryRunner().execute("DROP SCHEMA " + SCHEMA);
+    }
+
+    @Override
+    @BeforeAll
+    public void initFileSystem()
```
findepi: Why overridden?
The default impl should work just fine

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+            GcsStorageFactory storageFactory = new GcsStorageFactory(config, new GcsServiceAccountAuth(authConfig));
+            fileSystem = new GcsFileSystemFactory(config, storageFactory).create(SESSION);
+        }
+        catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    @Override
+    protected String getMetadataLocation(String tableName)
```
findepi, line 121: Why overridden?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+    {
+        assertThatThrownBy(super::testCommentView)
+                .hasMessageContaining("Server does not support endpoint: POST /v1/{prefix}/namespaces/{namespace}/views");
+    }
+
+    @Test
+    @Override // TODO Enable this test once Google fixes a bug. January 2026 release will contain the fix.
+    public void testRegisterTableWithComments()
+    {
+        abort("skipped");
```
findepi: - link to the bug tracker?
- `assertThatThrownBy...hasMessageContaining` would force us to know when the bug is fixed. "automated todo"

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    @Test
+    @Override // TODO Enable once flaky issue is fixed
+    public void testDeleteRowsConcurrently()
+    {
+        abort("skipped");
```
findepi: what issue? 
is this correctness bug, or something else?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    @Test
+    @Override
+    public void testCreateTableWithTrailingSpaceInLocation()
```
findepi, line 250: why overridden?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+        String tableLocationWithoutTrailingSpace = schemaPath() + "/" + tableName;
+        String tableLocationWithTrailingSpace = tableLocationWithoutTrailingSpace + " ";
+
+        assertThat(query(format("CREATE TABLE %s WITH (location = '%s') AS SELECT 1 AS a, 'INDIA' AS b, true AS c", tableName, tableLocationWithTrailingSpace))).failure()
+                        .hasMessage("Failed to create transaction")
+                .hasStackTraceContaining("Malformed request: The table `location` property can only point to the default location");
+    }
+
+    @Test
+    @Override // BigLake metastore requires table location to start with the prefix with the table name
```
findepi, line 262: that's so nice

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+        fileSystem.deleteDirectory(tableLocation);
+        assertThat(fileSystem.listFiles(tableLocation).hasNext())
+                .describedAs("Table location should not exist")
+                .isFalse();
+
+        assertThat(getQueryRunner().tableExists(getSession(), tableName)).isFalse();
+    }
+
+    @Test
+    @Override // BigLake doesn't delete data files when dropping
```
findepi: Isn't that compensated on our side though?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
```diff
@@ -0,0 +1,404 @@
...
+
+        // try to drop table
+        assertUpdate("DROP TABLE " + tableName);
+        assertThat(getQueryRunner().tableExists(getSession(), tableName)).isFalse();
+
+        assertThat(listFiles(tableLocation)).containsExactly(dataFile.location());
+    }
+
+    @Test
+    public void testRecreateTableWithSameName()
```
findepi: is this test BigLake-specific?

docs/src/main/sphinx/object-storage/metastores.md
ebyhr: I couldn't find a better name. Let me rename it to `BIGLAKE`.

docs/src/main/sphinx/object-storage/metastores.md
ebyhr: Since this is a new integration, I don't think we want to increase the usage of legacy code - `hive.gcs.*` properties are deprecated.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
ebyhr: Do you mean purgeBigLakeTable (prune → purge)?

docs/src/main/sphinx/object-storage/metastores.md
talatuyarer: @ebyhr Iceberg Java Core library also use Google for auth type fyi.

ebyhr: @talatuyarer I knew. That is why I used `GOOGLE` here.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
ebyhr: They thew an exception if I remember correctly, but I confirmed it succeeds now. I will fix these code.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
ebyhr: @talatuyarer @Noremac201 Do you have a public link we can access?

talatuyarer: No we dont. But If you want create a bug on Trino's github and assign to me ? I can update Bug after we roll out the fix

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
ebyhr: The base class doesn't have an implementation: 
```java
protected abstract String getMetadataLocation(String tableName);
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
ebyhr: @talatuyarer Thanks! I created #27679. It seems I can't assign you maybe cause you haven't joined `trinodb` organization.

ebyhr: > assertThatThrownBy...hasMessageContaining 

I avoided `assertThatThrownBy` since the behavior isn't deterministic.

docs/src/main/sphinx/object-storage/metastores.md
findepi: is this is GCP-specific auth that happens to be used for their BigLake product, then 'google' or 'gcp' would be a better name. if Iceberg already uses 'google' for this, i'm ok with using 'google' as well here.

docs/src/main/sphinx/object-storage/metastores.md
findepi: Why legacy hadoop support for s3 and gcs wasn't removed yet? I think this tells something about lack of certainty or missing features in trino native filesystems.
However, catalog (this PR) seems orthogonal to filesystem. 
We don't have "The legacy Hadoop GCS file system is not supported." for any other catalog type, so maybe we don't need it for this one. Unless there is a reason to suspect this combination won't work for some specific reason.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
findepi: yes

docs/src/main/sphinx/object-storage/metastores.md
findepi: this is resolved, but no code change. did you intend to push new code version?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findepi: it succeeds as in "does not throw"?
or, it succeeds as in "stats are preserved"?

(is this covered by `BaseIcebergConnectorSmokeTest#testAnalyze` or some other ?)

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
findepi: if it's not deterministic, what are the possible end states?

docs/src/main/sphinx/object-storage/metastores.md
talatuyarer: yes Iceberg uses 'google' keyword. This is for rest client. If someone deploy an Rest Server on GCp they need to use GoogleAuthmanager for token refresh and other scoping. BigLake's Rest Server also uses too.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
ebyhr: The latter. Yes - it's covered by the test.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergBigLakeMetastoreConnectorSmokeTest.java
ebyhr: It's timeout issue. Update the code comment.

goktugkose:
Hello @ebyhr and @talatuyarer,

I tested this PR using the BigLake REST Catalog and encountered an issue that I wanted to double-check with you.

When I enable `iceberg.rest-catalog.case-insensitive-name-matching=true` with the BigLake Metastore and use DataGrip to connect Trino, the catalog loading appears to loop indefinitely, whilst throwing HTTP status 429 errors, and eventually hits the `Iceberg REST Catalog read requests per minute limit` (720 requests). I don’t observe this behavior when case-insensitive name matching is disabled.

Have you experienced something similar, or is this a known limitation with BigLake?

PR #26225, opened 2025-07-17
Add support for multipart uploads in Azure FS

wendigo:
## Release notes

```markdown
## Azure Native FS
* Add `azure.multipart-write-enabled` that enables multipart uploads for large files ({issue}`issuenumber`)
```

### DIFF
--- a/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemConfig.java
+++ b/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemConfig.java
@@ -39,6 +39,7 @@ public class AzureFileSystemConfig
     private DataSize maxSingleUploadSize = DataSize.of(4, Unit.MEGABYTE);
     private Integer maxHttpRequests = 2 * Runtime.getRuntime().availableProcessors();
     private String applicationId = "Trino";
+    private boolean multipartWriteEnabled;
 
     @NotNull
     public AuthType getAuthType()
@@ -144,5 +145,18 @@ public class AzureFileSystemConfig
     {
         this.applicationId = applicationId;
         return this;
+    }
+
+    @Config("azure.multipart-write-enabled")
+    @ConfigDescription("Enable multipart writes for large files")
+    public AzureFileSystemConfig setMultipartWriteEnabled(boolean multipartWriteEnabled)
+    {
+        this.multipartWriteEnabled = multipartWriteEnabled;
+        return this;
+    }
+
+    public boolean isMultipartWriteEnabled()
+    {
+        return multipartWriteEnabled;
     }
 }
--- a/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
+++ b/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
@@ -31,13 +31,18 @@ import jakarta.annotation.PreDestroy;
 import reactor.netty.resources.ConnectionProvider;
 
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static io.airlift.concurrent.Threads.daemonThreadsNamed;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.Executors.newCachedThreadPool;
 
 public class AzureFileSystemFactory
         implements TrinoFileSystemFactory
 {
+    private final ExecutorService uploadExecutor = newCachedThreadPool(daemonThreadsNamed("azure-upload-%s"));
+
     private final AzureAuth auth;
     private final String endpoint;
     private final DataSize readBlockSize;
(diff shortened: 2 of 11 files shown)


wendigo:
Supersedes https://github.com/trinodb/trino/pull/26204

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemConfig.java
```diff
@@ -39,6 +39,7 @@ public enum AuthType
     private DataSize maxSingleUploadSize = DataSize.of(4, Unit.MEGABYTE);
     private Integer maxHttpRequests = 2 * Runtime.getRuntime().availableProcessors();
     private String applicationId = "Trino";
+    private boolean multipartWriteEnabled;
```
raunaqmorarka, line 42: Why not enable by default ?

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
```diff
@@ -108,12 +117,13 @@ public void destroy()
                 // ignored
             }
         }
+        uploadExecutor.shutdown();
```
raunaqmorarka: should we do this first, in case the other stuff can throw ?

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureMultipartOutputStream.java
```diff
@@ -0,0 +1,248 @@
...
+            int writeBlockSizeBytes)
+    {
+        requireNonNull(location, "location is null");
+        requireNonNull(blobClient, "blobClient is null");
+        checkArgument(writeBlockSizeBytes >= 0, "writeBlockSizeBytes is negative");
+
+        this.location = location;
+        this.writeBlockSizeBytes = writeBlockSizeBytes;
+        this.overwrite = overwrite;
+        this.memoryContext = memoryContext.newLocalMemoryContext(AzureOutputStream.class.getSimpleName());
```
raunaqmorarka: AzureOutputStream  -> AzureMultipartOutputStream ?

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureMultipartOutputStream.java
```diff
@@ -0,0 +1,248 @@
...
+            boolean overwrite,
+            AggregatedMemoryContext memoryContext,
+            int writeBlockSizeBytes)
+    {
+        requireNonNull(location, "location is null");
+        requireNonNull(blobClient, "blobClient is null");
+        checkArgument(writeBlockSizeBytes >= 0, "writeBlockSizeBytes is negative");
+
+        this.location = location;
+        this.writeBlockSizeBytes = writeBlockSizeBytes;
```
raunaqmorarka, line 81: `s3.streaming.part-size` default is 32MB. We've seen low values cause throttling in S3 (https://github.com/trinodb/trino/pull/25781).
Is the 4MB default here too low ?

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureMultipartOutputStream.java
```diff
@@ -0,0 +1,248 @@
...
+
+        if (finished) {
+            this.buffer = null;
+        }
+        else {
+            this.buffer = new byte[0];
+            this.initialBufferSize = writeBlockSizeBytes;
+            bufferSize = 0;
+        }
+        memoryContext.setBytes(0);
```
raunaqmorarka: Should memoryContext be zeroed before the upload is done ?
Won't we hold on to the memory at least until upload finishes ?

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemConfig.java
wendigo: @raunaqmorarka i want to make it a default in the future but for now testing and escape hatch is needed

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureMultipartOutputStream.java
wendigo: No. Buffer field is reset before this line

raunaqmorarka: I know we've reset `buffer`. But won't we hold on to `byte[] data` reference until upload of stage finishes ?

wendigo: Done

core/trino-main/src/main/java/io/trino/operator/OperatorContext.java
```diff
@@ -697,6 +697,18 @@ public ListenableFuture<Void> setBytes(long bytes)
             return blocked;
         }
 
+        @Override
+        public ListenableFuture<Void> addBytes(long delta)
```
raunaqmorarka, line 701: you can use this method in `setBytes` to avoid duplication, same for other classes

raunaqmorarka (approved):

core/trino-main/src/main/java/io/trino/operator/OperatorContext.java
wendigo: It's deduplicated in other cases. For this one I would need to introduce more synchronization and I'd like to avoid that.

Issue #26185, opened 2025-07-13
Optimize iceberg manifests fails with INTERNAL ERROR


guyco33:
`OPTIMIZE_MANIFESTS` fails when first partition column is NULL

```sql
CREATE TABLE test_table (
    c1 integer,
    c2 date,
    c3 double)
WITH (
    format = 'PARQUET',
    partitioning = ARRAY['c1', 'month(c2)']);

INSERT INTO test_table VALUES (1,date'2025-07-07', 0.1), (3,null, 0);
INSERT INTO test_table VALUES (null,date'2025-07-13', 0.2), (1,null, 0);
```

```sql
ALTER TABLE test_table EXECUTE OPTIMIZE_MANIFESTS;

java.lang.NullPointerException: Key cannot be null
	at org.apache.iceberg.relocated.com.google.common.base.Preconditions.checkNotNull(Preconditions.java:920)
	at org.apache.iceberg.BaseRewriteManifests.appendEntry(BaseRewriteManifests.java:334)
	at org.apache.iceberg.BaseRewriteManifests.lambda$performRewrite$4(BaseRewriteManifests.java:264)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at org.apache.iceberg.BaseRewriteManifests.lambda$performRewrite$5(BaseRewriteManifests.java:262)
	at org.apache.iceberg.util.Tasks$Builder.runTaskWithRetry(Tasks.java:413)
	at org.apache.iceberg.util.Tasks$Builder$1.run(Tasks.java:315)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1095)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:619)
	at java.base/java.lang.Thread.run(Thread.java:1447)
```

```sql
DELETE FROM test_table WHERE c1 IS NULL;

ALTER TABLE test_table EXECUTE OPTIMIZE_MANIFESTS;
-- succeed
```

PR #26227, opened 2025-07-17
Use first non null partition in Iceberg `optimize_manifests` procedure

chenjian2664:
## Description

Close https://github.com/trinodb/trino/issues/26185

## Release notes

```markdown
## Iceberg
* Fix `optimize_manifests` failure when the partition contains `NULL` value in Iceberg. ({issue}`26185`)
```


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2132,7 +2132,13 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
             // Use the first partition field as the clustering key
             StructLike partition = file.partition();
-            return partition.size() > 1 ? partition.get(0, Object.class) : partition;
+            for (int i = 0; i < partition.size(); i++) {
+                Object partitionValue = partition.get(i, Object.class);
+                if (partitionValue != null) {
+                    return partitionValue;
+                }
+            }
+            return partition;
```
ebyhr: Why do we need to find the first non-null partition instead of changing the value to `Optional`?
```java
return partition.size() > 1 ? Optional.ofNullable(partition.get(0, Object.class)) : partition;
```

chenjian2664: I think yours is better.

PR #23322, opened 2024-09-06
Include Column Lineage Info for All Queries

behrooz-stripe:
## Description
Column lineage info is extremely helpful but is only available for Create View, Table etc. not Select queries. This PR simply expands the support for lineage for other queries as well.

## Additional context and related issues
adds column lineage to updateTarget so it'll be available in `EventListener` without any other plumbing/changes required. Please see `assertLineage` for testing, it'll force all the select queries to go through the same tests and assumptions.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
(x) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #26241, opened 2025-07-18
add lineage support for output columns of all select queries availabl…

behrooz-stripe:
## Description
Column lineage info is extremely helpful but is only available for Create View, Table etc. not Select queries. This PR expands the support for lineage for all select queries as well.

Adds column lineage to `QueryCompletedEvent` so it'll be available in `EventListener`. This PR is a follow up to [this one](https://github.com/trinodb/trino/pull/23322) that went stale. 

## Release notes


behrooz-stripe:
Hey @Praveen2112 could you please provide feedback thank you!

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
```diff
@@ -86,6 +87,7 @@ public class QueryInfo
     private final RetryPolicy retryPolicy;
     private final boolean pruned;
     private final NodeVersion version;
+    private final List<ColumnLineageInfo> selectColumnLineageInfo;
```
Praveen2112: Should it be Optional ? In case of CREATE VIEW/CTAS kind of operations would it be empty ?

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * This class is JSON serializable for storing column lineage information for select queries.
+ */
+public class ColumnLineageInfo
+{
+    private final String name;
+    private final int index;
```
Praveen2112: Any specific reason for maintaining the index ?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -1594,6 +1617,9 @@ protected Scope visitQuery(Query node, Optional<Scope> scope)
                     .build();
 
             analysis.setScope(node, queryScope);
+            if (isTopLevel) {
```
Praveen2112: Any specific reason for computing them in the `StatementAnalyzer` - Can we compute the same in `Analysis` based on `getRootScope().getRelationType().getVisibleFields()` ? `StatementAnalyzer` should capture the lineage details for all columns, but populating it for a specific set of columns should be done in `Analysis` - cc @martint

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
```diff
@@ -0,0 +1,82 @@
...
+    private final Set<ColumnDetail> sourceColumns;
+
+    @JsonCreator
+    public ColumnLineageInfo(@JsonProperty String outputExpression, @JsonProperty int index, @JsonProperty Set<ColumnDetail> sourceColumns)
+    {
+        this.name = outputExpression;
```
Praveen2112: nit: rnn - Can we also have a coherent naming for the field and constructor argument ?

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
```diff
@@ -0,0 +1,82 @@
...
+        this.index = index;
+        this.sourceColumns = sourceColumns;
```
Praveen2112: nit: rnn and ImmutableList#copyOf

Praveen2112 (commented):
I need to check on tests once again

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
behrooz-stripe: I don't think so, if you look at the constructor you'll see `requireNonNull(selectColumnLineageInfo, "selectColumnLineageInfo is null");` so it's guaranteed to be non-null and will be simply empty for other cases (create view etc.). I also think Optional collections are non-ergonomic and cause a lot unnecssary empty checks while the data-structure itself inherently handles it. It also looks like the rest of collections in this class follow the same pattern so I didn't make it Optional. If you have strong preference on it, happy to change it.

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
behrooz-stripe: Yes because when we get visibleFields we can NOT make any assumption about the order of fields:
```
 /**
     * Gets only the visible fields.
     * No assumptions should be made about the order of the fields returned from this method.
     * To obtain the index of a field, call indexOf.
     */
    public Collection<Field> getVisibleFields()
    {
        return visibleFields;
    }
```

behrooz-stripe: This is particularly helpful when we have unnamed output columns.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
behrooz-stripe: I see what you mean, architecturally what you suggest makes more sense `StatementAnalyzer` should invoke lineage analysis and `Analysis` class should do the compute logic. Let me take a stab at it.

behrooz-stripe:
@Praveen2112 thank you for reviewing this and your feedback. I believe I addressed your questions PTAL again.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
behrooz-stripe: I tried using `getRootScope().getRelationType().getVisibleFields()` as you suggested in [this commit](https://github.com/trinodb/trino/pull/26241/commits/58136fd9d1a995baa5e1c4ecb77caccf58fed29e)
but a bunch of tests started failing. It seems like it's because scope was missing. I simply moved the logic to analysis but do pass the scope that is guaranteed to exist from `StatementAnalyzer` to `Analysis`. Here is the stack trace of missing scope.
image

behrooz-stripe:
@Praveen2112 could you please review again? thank you.

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
```diff
@@ -0,0 +1,82 @@
...
+    private final String name;
+    private final int index;
+    private final Set<ColumnDetail> sourceColumns;
+
+    @JsonCreator
+    public ColumnLineageInfo(@JsonProperty String name, @JsonProperty int index, @JsonProperty Set<ColumnDetail> sourceColumns)
+    {
+        this.name = name;
+        this.index = index;
+        this.sourceColumns = Set.copyOf(sourceColumns);
```
Praveen2112, line 37: `requireNonNull` checks and can we use `ImmutableSet#copyOf(sourceColumns)` as it would gurantee us ordering of them

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
```diff
@@ -0,0 +1,82 @@
...
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+import java.util.Set;
+
+/**
+ * This class is JSON serializable for storing column lineage information for select queries.
+ */
+public class ColumnLineageInfo
```
Praveen2112: This can be a record right ?

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
```diff
@@ -193,6 +198,13 @@ public QueryInfo(
         this.retryPolicy = retryPolicy;
         this.pruned = pruned;
         this.version = version;
+        this.selectColumnLineageInfo = selectColumnLineageInfo;
```
Praveen2112: nit: I think we should perform `ImmutableList#copyOf`

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
Praveen2112: What if we capture them as a part of `SelectExpression` - we could resolve the top level based on the query node right ?

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
behrooz-stripe: I added `requireNonNull` but guava and `ImmutableSet` are not available in the SPI module as it's designed to have minimal dependencies, and changing that is a rather big. I am not sure why we'd want to preserve order for source columns here. But if you're really sold on it, I can also change it to a `java.util.LinkedHashSet` that also preserves order but is more costly. For now I leave it as is (`Set.copyOf`) but let me know if you want the linked one or adding dependency. If you check other classes in this package, this seems to be the norm.

behrooz-stripe:
@Praveen2112 addressed your feedback. Please check again, the failing tests seem to be flakey and not related.

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
Praveen2112: > simply empty for other cases (create view etc.)

This is something we would like to avoid - We could have it similar to `Optional<Output> output` bcoz the select lineages are populated only for a specific query operations.

> It also looks like the rest of collections in this class follow the same pattern

The remaing parts like common for all types of queries.

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
```diff
@@ -124,7 +126,9 @@ public QueryInfo(
             @JsonProperty("queryType") Optional<QueryType> queryType,
             @JsonProperty("retryPolicy") RetryPolicy retryPolicy,
             @JsonProperty("pruned") boolean pruned,
-            @JsonProperty("version") NodeVersion version)
+            @JsonProperty("version") NodeVersion version,
+            @JsonProperty("selectColumnLineageInfo") List<ColumnLineageInfo> selectColumnLineageInfo)
```
Praveen2112: nit: Can we move this next to `Optional<Output> ..`

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -269,13 +271,45 @@ public class Analysis
...
+            Field field = outputFields.get(i);
+            String outputColumnName = field.getName().orElse("");
+            ImmutableSet<ColumnDetail> sources = getSourceColumns(field).stream()
+                    .map(SourceColumn::getColumnDetail)
+                    .collect(toImmutableSet());
+            lineageInfo.add(new ColumnLineageInfo(outputColumnName, outputFieldIndices.get(i), sources));
+        }
+        // always sort lineageInfo by index to ensure consistent ordering
+        lineageInfo.sort(Comparator.comparingInt(ColumnLineageInfo::index));
+        this.selectColumnLineageInfo = ImmutableList.copyOf(lineageInfo);
```
Praveen2112: Instead of setting this, can we generate them on demand ? We do have `root` scope and we could get the output fields and populate them if exists. I don't think we need to customize it based on Node type and plan accordinhly.

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
behrooz-stripe: Sure switched to Optional.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
behrooz-stripe: Good idea, this way we don't have to mess with statementanalyzer at all.

behrooz-stripe:
@Praveen2112 all comments are addressed now.

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -983,6 +986,16 @@ public void setOutput(Optional<Output> output)
         this.output.set(output);
     }
 
+    public void setSelectColumnsLineageInfo(List<ColumnLineageInfo> selectOutputColumnsLineage)
```
Praveen2112: nit: Analysis can return it as `Optional<List<ColumnLineageInfo>>` right ? In this case we could directly set the lineage info.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,28 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
         this.queryType = requireNonNull(queryType, "queryType is null");
     }
 
+    public List<ColumnLineageInfo> getSelectColumnLineageInfo()
+    {
+        Scope rootScope = getRootScope();
```
Praveen2112: Additionally can we return an Optional#empty or an empty list of `ColumnIneageInfo` - if `target` is set ?

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,28 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
...
+                .collect(toImmutableList());
+        List<ColumnLineageInfo> lineageInfo = new ArrayList<>();
+        for (int i = 0; i < outputFields.size(); i++) {
+            Field field = outputFields.get(i);
+            String outputColumnName = field.getName().orElse("");
+            ImmutableSet<ColumnDetail> sources = getSourceColumns(field).stream()
+                    .map(SourceColumn::getColumnDetail)
+                    .collect(toImmutableSet());
+            lineageInfo.add(new ColumnLineageInfo(outputColumnName, outputFieldIndices.get(i), sources));
+        }
```
Praveen2112: nit: Can we simplify the for-loop using streams

```
        rootScope.getRelationType().getVisibleFields().stream()
                .map(field -> new ColumnLineageInfo(field.getName().orElse(""), rootScope.getRelationType().indexOf(field), getSourceColumns(field).stream()
                        .map(SourceColumn::getColumnDetail)
                        .collect(toImmutableSet())))
                .sorted(Comparator.comparingInt(ColumnLineageInfo::index))
                .collect(toImmutableList());
```

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
```diff
@@ -0,0 +1,48 @@
...
+import static java.util.Objects.requireNonNull;
+
+/**
+ * This record is JSON serializable for storing column lineage information for select queries.
+ */
+public record ColumnLineageInfo(
+        @JsonProperty String name,
+        /*
+          The index of the column in the select list.
+          This is useful for identifying the position of the column in the result set.
```
Praveen2112: Since we expose it as a `List<ColumnLineageInfo>` - won't this index be a bit redundant ?

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,159 @@ public void testDisallowAggregationFunctionInUnnest()
                 .hasMessage("line 1:46: UNNEST cannot contain aggregations, window functions or grouping operations: [COUNT(t.a)]");
     }
 
+    @Test
+    public void testSelectColumnLineageInfo()
```
Praveen2112: Can we move all the tests to `TestEventListenerBasic` ?

Praveen2112: And also test involving building a row field from two different set of table (via joins)

Praveen2112 (commented):
LGTM. Can we please squash all the commits into one - Plus I have one question on exposing the field number - is it required ?

Praveen2112:
@martint / @electrum / @kasiafi There are a few SPI changes and analysis changes. Can you confirm if the changes are good to go ?

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,159 @@ public void testDisallowAggregationFunctionInUnnest()
...
+
+        // Check third column lineage
+        ColumnLineageInfo colLiteral = lineageInfos.get(2);
+        assertThat(colLiteral.name()).isEqualTo("literal");
+        assertThat(colLiteral.sourceColumns()).isEmpty(); // 'literal' is a literal value
+        assertThat(colLiteral.index()).isEqualTo(2);
+
+        // Check fourth column lineage
+        ColumnLineageInfo colAB = lineageInfos.get(3);
+        assertThat(colAB.name()).isEqualTo(""); // anonymous
```
chenjian2664:
```suggestion
        assertThat(colAB.name()).isEmpty(); // anonymous
```
Same below

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
```diff
@@ -125,6 +128,7 @@ public QueryInfo(
             @JsonProperty("retryPolicy") RetryPolicy retryPolicy,
             @JsonProperty("pruned") boolean pruned,
             @JsonProperty("version") NodeVersion version)
+
```
chenjian2664: remove

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
behrooz-stripe: >Can we move all the tests to TestEventListenerBasic ?

I don't think they belong there. I have a test there as well but these are analysis related logic and I think that's a better home for them. `TestEventListenerBasic` afaik is more about plumbing this info around not calculating it which is an analysis concern. If you have strong preference on this, I can move it though.

core/trino-spi/src/main/java/io/trino/spi/eventlistener/ColumnLineageInfo.java
behrooz-stripe: I had it as a set in the past and also thought explicit index is more apparent to consumers. But I got rid of it as per your suggestion. I did add a comment on the method that the order of the returned list matches the output columns order.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
behrooz-stripe: not really sure what do you mean and how this relates to `target`.

behrooz-stripe:
@Praveen2112 addressed your feedback and also squashed all commits into one.

behrooz-stripe:
Hey @Praveen2112  @martint  @electrum  @kasiafi could we pretty please move this PR forward? happy to address any remaining comment/suggestion you have! Thank you in advance.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,21 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
         this.queryType = requireNonNull(queryType, "queryType is null");
     }
 
+    public Optional<List<ColumnLineageInfo>> getSelectColumnLineageInfo()
```
kasiafi: nit: rename to `getSelectColumnsLineageInfo`

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,21 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
...
+    {
+        Scope rootScope = getRootScope();
+        ImmutableList<ColumnLineageInfo> lineageInfo = rootScope.getRelationType().getVisibleFields().stream()
+                // sort output fields by their index to ensure consistent ordering of lineage info
+                .sorted(Comparator.comparingInt(rootScope.getRelationType()::indexOf))
+                .map(field -> new ColumnLineageInfo(field.getName().orElse(""),
```
kasiafi: Put each argument of the constructor on a new line.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,21 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
         this.queryType = requireNonNull(queryType, "queryType is null");
     }
 
+    public Optional<List<ColumnLineageInfo>> getSelectColumnLineageInfo()
+    {
+        Scope rootScope = getRootScope();
```
kasiafi: You can use the method `getOutputDescriptor()`. It will give you the root `RelationType` directly.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,21 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
...
+        ImmutableList<ColumnLineageInfo> lineageInfo = rootScope.getRelationType().getVisibleFields().stream()
+                // sort output fields by their index to ensure consistent ordering of lineage info
+                .sorted(Comparator.comparingInt(rootScope.getRelationType()::indexOf))
+                .map(field -> new ColumnLineageInfo(field.getName().orElse(""),
+                        getSourceColumns(field)
+                            .stream()
+                            .map(SourceColumn::getColumnDetail)
+                            .collect(toImmutableSet())))
+                .collect(toImmutableList());
+        return lineageInfo.isEmpty() ? Optional.empty() : Optional.of(ImmutableList.copyOf(lineageInfo));
```
kasiafi: No need to do `ImmutableList.copyOf(lineageInfo)`. It is already an ImmutableList.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -276,6 +278,21 @@ public Analysis(@Nullable Statement root, Map<NodeRef<Parameter>, Expression> pa
         this.queryType = requireNonNull(queryType, "queryType is null");
     }
 
+    public Optional<List<ColumnLineageInfo>> getSelectColumnLineageInfo()
+    {
+        Scope rootScope = getRootScope();
+        ImmutableList<ColumnLineageInfo> lineageInfo = rootScope.getRelationType().getVisibleFields().stream()
```
kasiafi: Declare as `List<ColumnLineageInfo>`.

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
kasiafi: `target` is set for UPDATE-type statements (INSERT, UPDATE, DELETE, MERGE, etc.). They have output columns but we don't want to return the lineage for them, so we should return `Optional.empty()`.

On the other hand, there are other statements that go through the Analyzer, and do not set the `target`, for example SET COLUMN TYPE. Usually, they do not produce output columns, but I think we should not rely on it.

For now, we can return `Optional.empty` in two cases:
- no output columns
- target is set

But we should think of a way to explicitly find out if the processed statement is of SELECT type.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,191 @@ public void testDisallowAggregationFunctionInUnnest()
...
     }
 
+    @Test
+    public void testSelectColumnLineageInfo()
+    {
+        String sql = "SELECT a, b + 1 AS b1, 'C' as literal, a + b FROM (VALUES (1, 2)) t(a, b)";
+
+        Analysis analysis = analyze(sql);
+
+        Optional<List<ColumnLineageInfo>> maybeLineageInfo = analysis.getSelectColumnLineageInfo();
```
kasiafi: Rename `maybeLineageInfo` to `optionalLineageInfo`. Here and in the other test cases.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,191 @@ public void testDisallowAggregationFunctionInUnnest()
...
+
+        Optional<List<ColumnLineageInfo>> maybeLineageInfos = analysis.getSelectColumnLineageInfo();
+        assertThat(maybeLineageInfos).isPresent();
+        List<ColumnLineageInfo> lineageInfo = maybeLineageInfos.get();
+        assertThat(lineageInfo.size()).isEqualTo(1);
+
+        // Check the column lineage
+        ColumnLineageInfo colCount = lineageInfo.getFirst();
+        assertThat(colCount.name()).isEmpty(); // anonymous
+        assertThat(colCount.sourceColumns()).isNotEmpty();
```
kasiafi: This assertion is redundant.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,191 @@ public void testDisallowAggregationFunctionInUnnest()
...
+
+        Optional<List<ColumnLineageInfo>> maybeLineageInfos = analysis.getSelectColumnLineageInfo();
+        assertThat(maybeLineageInfos).isPresent();
+        List<ColumnLineageInfo> lineageInfo = maybeLineageInfos.get();
+        assertThat(lineageInfo.size()).isEqualTo(1);
+
+        // Check the column lineage
+        ColumnLineageInfo colA = lineageInfo.getFirst();
+        assertThat(colA.name()).isEqualTo("unionized");
+        assertThat(colA.sourceColumns()).hasSize(3);
```
kasiafi: This assertion is redundant.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,191 @@ public void testDisallowAggregationFunctionInUnnest()
...
+
+        Optional<List<ColumnLineageInfo>> maybeLineageInfo = analysis.getSelectColumnLineageInfo();
+        assertThat(maybeLineageInfo).isPresent();
+        List<ColumnLineageInfo> lineageInfo = maybeLineageInfo.get();
+        assertThat(lineageInfo.size()).isEqualTo(1);
+
+        // Check the column lineage
+        ColumnLineageInfo colA = lineageInfo.getFirst();
+        assertThat(colA.name()).isEqualTo("a");
+        assertThat(colA.sourceColumns()).hasSize(2);
```
kasiafi: Redundant. Here and in the remaining cases.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7430,6 +7432,191 @@ public void testDisallowAggregationFunctionInUnnest()
                 .hasMessage("line 1:46: UNNEST cannot contain aggregations, window functions or grouping operations: [COUNT(t.a)]");
     }
 
+    @Test
```
kasiafi, line 7437: Please test the following scenarios:
- EXPLAIN statement
- INSERT statement (or another update-type statement)
- ADD COLUMN statement or another that does not produce output columns

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
behrooz-stripe: alright, I added the if to return empty when `target` is set. Also the return statement at the end handles the second case you menioned @kasiafi `return lineageInfo.isEmpty() ? Optional.empty() : Optional.of(lineageInfo);` aka when no output columns, there will be no lineage hence `Optional.empty()`

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
behrooz-stripe: Added!

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
behrooz-stripe: @kasiafi actually your request about adding a test for `explain` made me think of adding this check at the top of this function:
```
        //  This single check should handle all cases where we don't want to produce lineage info:
        //  - EXPLAIN ✓
        //  - INSERT/UPDATE/DELETE/MERGE ✓
        //  - ALTER TABLE ADD COLUMN ✓
        //  - SET COLUMN TYPE or any other DDL ✓
        if (!(root instanceof Query)) {
            return Optional.empty();
        }
```
I think this should cover all the cases we do NOT want lineage info and is less brittle than `setUpdateType` where it checks for ` if (queryType != EXPLAIN) {` Let me know what you think!

behrooz-stripe:
@kasiafi I addressed all your feedback, added new tests, merged master and squashed. Ready for you to take another look. Thank you!

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
kasiafi: I ran through different variants of query statements, and indeed they all end up as a `Query`.
We can use this fact and add a condition as you suggested.
However, note that processing them as `Query` is an implementation detail that might change over time, so we should add regression tests.
Please add assertions for the following query shapes (listed based on `SqlBase.g4`):
- WITH SESSION ...
- WITH FUNCTION ...
- WITH named_query AS (...) ...
- TABLE table_name
- VALUES ...
- ( query_in_parentheses )
- query INTERSECT query (or UNION / EXCEPT -- one set operation is enough)
- EXPLAIN ... -- it also produces lineage, right?
- EXPLAIN ANALYZE ... -- should not produce lineage as it isn't processed as `Query`
- SHOW SCHEMAS ... (or another SHOW statement -- they're rewritten into `Query` and should produce lineage)

core/trino-main/src/main/java/io/trino/execution/QueryInfo.java
```diff
@@ -117,6 +119,7 @@ public QueryInfo(
             @JsonProperty("warnings") List<TrinoWarning> warnings,
             @JsonProperty("inputs") Set<Input> inputs,
             @JsonProperty("output") Optional<Output> output,
+            @JsonProperty("selectColumnLineageInfo") Optional<List<ColumnLineageInfo>> selectColumnLineageInfo,
```
kasiafi: nit: rename to `selectColumnsLineageInfo`

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
behrooz-stripe: Alright, I added a test for each and everyone of the scenarios you mentioned. All as expected except Explain. 
Explain does not come as `Query` and comes as [Explain](https://github.com/trinodb/trino/blob/master/core/trino-parser/src/main/java/io/trino/sql/tree/Explain.java#L24); hence gets filtered out by this if statement despite generating an output (`Query Plan`). I think it's sensible to not have `selectColumnsLineageInfo` for Explain queries but if you prefer to include them I can expand this if to allow explain queries go thorough. 
Basically quickly grokking thorough different classes: 
these will match:
  - Regular SELECT queries
  - VALUES
  - TABLE statements
  - UNION/INTERSECT/EXCEPT
  - Subqueries
  - SHOW statements (rewritten to Query internally)

  But it will NOT match:
  - EXPLAIN (extends Statement, not Query)
  - INSERT/UPDATE/DELETE/MERGE (extend Statement, not Query)
  - ALTER TABLE (extends Statement, not Query)

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7432,6 +7434,366 @@ public void testDisallowAggregationFunctionInUnnest()
...
+    {
+        String sql = "SHOW SCHEMAS";
+
+        Analysis analysis = analyze(sql);
+
+        Optional<List<ColumnLineageInfo>> maybeLineageInfo = analysis.getSelectColumnsLineageInfo();
+        assertThat(maybeLineageInfo).isPresent();
+        List<ColumnLineageInfo> lineageInfo = maybeLineageInfo.get();
+        assertThat(lineageInfo).hasSize(1);
+        assertThat(lineageInfo.get(0).name()).isEqualTo("Schema");
```
kasiafi, line 7794: Could we assert the actual lineage?

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -7432,6 +7434,366 @@ public void testDisallowAggregationFunctionInUnnest()
...
+    }
+
+    @Test
+    public void testSelectColumnsLineageInfoWithShow()
+    {
+        String sql = "SHOW SCHEMAS";
+
+        Analysis analysis = analyze(sql);
+
+        Optional<List<ColumnLineageInfo>> maybeLineageInfo = analysis.getSelectColumnsLineageInfo();
```
kasiafi: nit: rename `maybeLineageInfo` to `optionalLineageInfo` everywhere

kasiafi:
@Praveen2112 I have reviewed the Analysis part. Back to you.

kasiafi:
@behrooz-stripe thank you! 
@Praveen2112 is this good to go?

Praveen2112:
Can we reword the commit as `Compute lineage for output columns in all SELECT queries' ?

Praveen2112:
@behrooz-stripe Thanks for working on this !!

behrooz-stripe:
Thanks @Praveen2112 @kasiafi @electrum

chenjian2664:
@behrooz-stripe Thank you for the effort and hard work on this!

PR #26246, opened 2025-07-20
Use AWS v2 SDK for S3 in product tests

electrum:
## Release notes

### DIFF
--- a/testing/trino-product-tests/pom.xml
+++ b/testing/trino-product-tests/pom.xml
@@ -20,16 +20,6 @@
 
     <dependencies>
         <dependency>
-            <groupId>com.amazonaws</groupId>
-            <artifactId>aws-java-sdk-core</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>com.amazonaws</groupId>
-            <artifactId>aws-java-sdk-s3</artifactId>
-        </dependency>
-
-        <dependency>
             <groupId>com.google.guava</groupId>
             <artifactId>guava</artifactId>
         </dependency>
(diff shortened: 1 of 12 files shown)


wendigo:
Test failure is related:

```
tests               | 2025-07-23 01:00:20 INFO: ManageTestResources.onFinish: running checks
tests               | FATAL: io.trino.testng.services.ManageTestResources: 
tests               | 	Test instance field has value that looks like a resource
tests               | 	    Test class: class io.trino.tests.product.deltalake.TestDeltaLakeActiveFilesCache
tests               | 	    Instance: io.trino.tests.product.deltalake.TestDeltaLakeActiveFilesCache@4f2d995e
tests               | 	    Field: private software.amazon.awssdk.services.s3.S3Client io.trino.tests.product.deltalake.TestDeltaLakeActiveFilesCache.s3
tests               | 	    Value: software.amazon.awssdk.services.s3.DefaultS3Client@2919aff3
tests               | 	    Test execution stage: AFTER_CLASS
tests               | 	    Matching rule: isAutoCloseable
tests               | 
tests               | 	Resources must not be allocated in a field initializer or a test constructor,
tests               | 	and must be freed when test class is completed.
tests               | 	Depending which rule has been violated, if the reported class holds on
tests               | 	to resources only conditionally, you can add @ResourcePresence to declare that.
tests               | 
tests               | JVM will be terminated
tests               | (exited)
tests               | (exited)
```

wendigo:
I've fixed remaining issues in the delta-lake tests (unclosed s3 instances)

electrum:
Thanks. Do you know why we didn't have this problem before?

wendigo:
@electrum I guess because previous client wasn't a Closeable?

Issue #26031, opened 2025-06-18
Using Trino without internet connection will cause slowness in the UI


nadavtzaysler:
the Trino web UI is using css files from fonts.google to acquire his fonts, this slows the webserver and can cause security issues and on the more severe cases if there is no internet connection the webserver will take minutes to open, to fix this we need to preload the fonts in the image making and pull them from a custom css who will replace the api

PR #26247, opened 2025-07-20
Use bundled fonts in the documentation

nadavtzaysler:
(X ) This is not user-visible or is docs only, and no release notes are required.


nadavtzaysler:
will solve the same issue in #26031 but for the trino docs

docs/src/main/sphinx/static/trino.css
```diff
@@ -1,3 +1,46 @@
...
+@font-face {
+    font-family: 'Roboto';
+    src: url('./fonts/Roboto-VariableFont_wdth,wght.woff2') format('woff2');
+    font-weight: 100 900;
+    font-stretch: 75% 100%;
+    font-style: normal;
+    font-display: swap;
+  }
+  
+  @font-face {
```
wendigo: Trino docs are not using Open Sans. This is redundant

docs/src/main/sphinx/static/trino.css
```diff
@@ -1,3 +1,46 @@
...
+      font-stretch: 75% 100%;
+      font-style: italic;
+      font-display: swap;
+    }
+  
+  @font-face {
+    font-family: 'Alata';
+    font-style: normal;
+    font-weight: 400;
+    src: url('./fonts/Alata-Regular.woff2') format('woff2');
```
wendigo: you are referencing Alata-Regular.woff2 but checked in file is `alata-v11-lati-regular.woff2`

chenjian2664:
Please rephrase the commit title https://trino.io/development/process.html#pull-request-and-commit-guidelines-

nadavtzaysler:
Seems like the docs are also trying to access github.com which can also be hurtful in a private domain, I will implement a way to check it before trying to acces it, @wendigo would you like it to be in a new pull request or should i add the changes here?

wendigo:
@nadavtzaysler no. There is no point in doing that

wendigo:
I've dropped last commit. Please create separate change for it.

PR #26259, opened 2025-07-22
[Draft] Add support for TIMESTAMP and TIMESTAMP WITH TIME ZONE types in Exasol connector

skyglass:
## Description

Added Exasol Trino connector support for **Timestamp** and **Timestamp With Local Time Zone** JDBC data types

- added correspondent tests for **timestamp** with precision data type
- added correspondent tests for **timestamp with local time zone** with precision data type
- implemented special addRoundTrip method with additional column expression parameter for the predicate to fix the test predicate assumption
- all timestamp with timezone tests use input literal timestamp string, which is interpreted as a JVM timestamp string ("America/Bahia_Banderas") but expected literal strings are interpreted as UTC strings. Therefore the difference between these timestamps is 6 hours (with DST) or 5 (without DST); for some historical values, like 1970 for example, the difference can be 7, which is expected behaviour for timestamp with local time zone

## Release notes

```markdown
## Exasol Connect
* Add support for `timestamp` and `timestamp with time zone` types. 
```

### DIFF
--- a/docs/src/main/sphinx/connector/exasol.md
+++ b/docs/src/main/sphinx/connector/exasol.md
@@ -103,6 +103,12 @@ Trino data type mapping:
 * - `HASHTYPE`
   - `VARBINARY`
   -
+* - `TIMESTAMP(n)`
+  - `TIMESTAMP(n)`
+  -
+* - `TIMESTAMP(n) WITH LOCAL TIME ZONE`
+  - `TIMESTAMP(n) WITH TIME ZONE`
+  -
 :::
 
 No other types are supported.
--- a/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
+++ b/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
@@ -30,6 +30,8 @@ import io.trino.plugin.jdbc.JdbcTableHandle;
 import io.trino.plugin.jdbc.JdbcTypeHandle;
 import io.trino.plugin.jdbc.LongReadFunction;
 import io.trino.plugin.jdbc.LongWriteFunction;
+import io.trino.plugin.jdbc.ObjectReadFunction;
+import io.trino.plugin.jdbc.ObjectWriteFunction;
 import io.trino.plugin.jdbc.QueryBuilder;
 import io.trino.plugin.jdbc.SliceReadFunction;
 import io.trino.plugin.jdbc.SliceWriteFunction;
@@ -43,34 +45,57 @@ import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ColumnPosition;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.type.LongTimestamp;
+import io.trino.spi.type.LongTimestampWithTimeZone;
+import io.trino.spi.type.TimeZoneKey;
+import io.trino.spi.type.TimestampType;
+import io.trino.spi.type.TimestampWithTimeZoneType;
 import io.trino.spi.type.Type;
 
 import java.sql.Connection;
 import java.sql.Date;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Timestamp;
 import java.sql.Types;
+import java.time.Instant;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.util.HexFormat;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Set;
+import java.util.TimeZone;
 import java.util.function.BiFunction;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static io.trino.plugin.jdbc.PredicatePushdownController.FULL_PUSHDOWN;
 import static io.trino.plugin.jdbc.StandardColumnMappings.bigintColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.booleanColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.decimalColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultCharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultVarcharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.doubleColumnMapping;
+import static io.trino.plugin.jdbc.StandardColumnMappings.fromLongTrinoTimestamp;
+import static io.trino.plugin.jdbc.StandardColumnMappings.fromTrinoTimestamp;
 import static io.trino.plugin.jdbc.StandardColumnMappings.integerColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.smallintColumnMapping;
+import static io.trino.plugin.jdbc.StandardColumnMappings.toLongTrinoTimestamp;
+import static io.trino.plugin.jdbc.StandardColumnMappings.toTrinoTimestamp;
 import static io.trino.plugin.jdbc.TypeHandlingJdbcSessionProperties.getUnsupportedTypeHandling;
 import static io.trino.plugin.jdbc.UnsupportedTypeHandling.CONVERT_TO_VARCHAR;
 import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static io.trino.spi.connector.ConnectorMetadata.MODIFYING_ROWS_MESSAGE;
+import static io.trino.spi.type.DateTimeEncoding.packDateTimeWithZone;
+import static io.trino.spi.type.DateTimeEncoding.unpackMillisUtc;
 import static io.trino.spi.type.DateType.DATE;
 import static io.trino.spi.type.DecimalType.createDecimalType;
+import static io.trino.spi.type.TimestampType.createTimestampType;
+import static io.trino.spi.type.TimestampWithTimeZoneType.createTimestampWithTimeZoneType;
+import static io.trino.spi.type.Timestamps.NANOSECONDS_PER_MILLISECOND;
+import static io.trino.spi.type.Timestamps.PICOSECONDS_PER_NANOSECOND;
 import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
@@ -79,10 +104,14 @@ import static java.util.stream.Collectors.joining;
 public class ExasolClient
         extends BaseJdbcClient
 {
+    private static final int EXASOL_TIMESTAMP_WITH_TIMEZONE = 124;
+
     private static final Set<String> INTERNAL_SCHEMAS = ImmutableSet.<String>builder()
             .add("EXA_STATISTICS")
             .add("SYS")
             .build();
+
+    private static final int MAX_EXASOL_TIMESTAMP_PRECISION = 9;
 
     @Inject
     public ExasolClient(
@@ -241,6 +270,10 @@ public class ExasolClient
                 return Optional.of(defaultVarcharColumnMapping(typeHandle.requiredColumnSize(), true));
             case Types.DATE:
                 return Optional.of(dateColumnMapping());
+            case Types.TIMESTAMP:
+                return Optional.of(timestampColumnMapping(typeHandle));
+            case EXASOL_TIMESTAMP_WITH_TIMEZONE:
+                return Optional.of(timestampWithTimeZoneColumnMapping(typeHandle));
         }
 
         if (getUnsupportedTypeHandling(session) == CONVERT_TO_VARCHAR) {
@@ -254,6 +287,257 @@ public class ExasolClient
         return typeHandle.jdbcType() == Types.CHAR
                 && typeHandle.jdbcTypeName().isPresent()
                 && typeHandle.jdbcTypeName().get().equalsIgnoreCase("HASHTYPE");
+    }
+
+    private static ColumnMapping timestampColumnMapping(JdbcTypeHandle typeHandle)
+    {
+        int timestampPrecision = typeHandle.requiredDecimalDigits();
+        TimestampType timestampType = createTimestampType(timestampPrecision);
+        if (timestampType.isShort()) {
+            return ColumnMapping.longMapping(
+                    timestampType,
+                    longTimestampReadFunction(timestampType),
+                    longTimestampWriteFunction(timestampType),
+                    FULL_PUSHDOWN);
+        }
+        return ColumnMapping.objectMapping(
+                timestampType,
+                objectTimestampReadFunction(timestampType),
+                objectTimestampWriteFunction(timestampType),
+                FULL_PUSHDOWN);
+    }
+
+    private static LongReadFunction longTimestampReadFunction(TimestampType timestampType)
+    {
+        return (resultSet, columnIndex) -> {
+            Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+            return toTrinoTimestamp(timestampType, timestamp.toLocalDateTime());
+        };
+    }
+
+    private static ObjectReadFunction objectTimestampReadFunction(TimestampType timestampType)
+    {
+        verifyObjectTimestampPrecision(timestampType);
+        return ObjectReadFunction.of(
+                LongTimestamp.class,
+                (resultSet, columnIndex) -> {
+                    Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+                    return toLongTrinoTimestamp(timestampType, timestamp.toLocalDateTime());
+                });
+    }
+
+    private static void verifyObjectTimestampPrecision(TimestampType timestampType)
+    {
+        int precision = timestampType.getPrecision();
+        checkArgument(precision > TimestampType.MAX_SHORT_PRECISION && precision <= MAX_EXASOL_TIMESTAMP_PRECISION,
+                "Precision is out of range: %s", precision);
+    }
+
+    private static ObjectWriteFunction objectTimestampWriteFunction(TimestampType timestampType)
+    {
+        int precision = timestampType.getPrecision();
+        verifyObjectTimestampPrecision(timestampType);
+
+        return new ObjectWriteFunction() {
+            @Override
+            public Class<?> getJavaType()
+            {
+                return LongTimestamp.class;
+            }
+
+            @Override
+            public void set(PreparedStatement statement, int index, Object value)
+                    throws SQLException
+            {
+                LocalDateTime localDateTime = fromLongTrinoTimestamp((LongTimestamp) value, precision);
+                Timestamp timestamp = Timestamp.valueOf(localDateTime);
+                statement.setObject(index, timestamp);
+            }
+
+            @Override
+            public String getBindExpression()
+            {
+                return getTimestampBindExpression(timestampType);
+            }
+
+            @Override
+            public void setNull(PreparedStatement statement, int index)
+                    throws SQLException
+            {
+                statement.setNull(index, Types.VARCHAR);
+            }
+        };
+    }
+
+    private static LongWriteFunction longTimestampWriteFunction(TimestampType timestampType)
+    {
+        return new LongWriteFunction()
+        {
+            @Override
+            public String getBindExpression()
+            {
+                return getTimestampBindExpression(timestampType);
+            }
+
+            @Override
+            public void set(PreparedStatement statement, int index, long epochMicros)
+                    throws SQLException
+            {
+                LocalDateTime localDateTime = fromTrinoTimestamp(epochMicros);
+                Timestamp timestampValue = Timestamp.valueOf(localDateTime);
+                statement.setObject(index, timestampValue);
+            }
+
+            @Override
+            public void setNull(PreparedStatement statement, int index)
+                    throws SQLException
+            {
+                statement.setNull(index, Types.VARCHAR);
+            }
+        };
+    }
+
+    private static ColumnMapping timestampWithTimeZoneColumnMapping(JdbcTypeHandle typeHandle)
+    {
+        int timestampPrecision = typeHandle.requiredDecimalDigits();
+        TimestampWithTimeZoneType timestampWithTimeZoneType = createTimestampWithTimeZoneType(timestampPrecision);
+
+        if (timestampWithTimeZoneType.isShort()) {
+            return ColumnMapping.longMapping(
+                    timestampWithTimeZoneType,
+                    longTimestampWithTimeZoneReadFunction(),
+                    longTimestampWithTimeZoneWriteFunction(timestampWithTimeZoneType),
+                    FULL_PUSHDOWN);
+        }
+        return ColumnMapping.objectMapping(
+                timestampWithTimeZoneType,
+                objectTimestampWithTimeZoneReadFunction(timestampWithTimeZoneType),
+                objectTimestampWithTimeZoneWriteFunction(timestampWithTimeZoneType),
+                FULL_PUSHDOWN);
+    }
+
+    private static LongReadFunction longTimestampWithTimeZoneReadFunction()
+    {
+        return (resultSet, columnIndex) -> {
+            Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+            return packDateTimeWithZone(timestamp.getTime(), TimeZone.getDefault().getID());
+        };
+    }
+
+    private static ObjectReadFunction objectTimestampWithTimeZoneReadFunction(
+            TimestampWithTimeZoneType timestampType)
+    {
+        verifyObjectTimestampWithTimeZonePrecision(timestampType);
+        return ObjectReadFunction.of(
+                LongTimestampWithTimeZone.class,
+                (resultSet, columnIndex) -> {
+                    Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+
+                    long millisUtc = timestamp.getTime();
+                    long nanosUtc = millisUtc * NANOSECONDS_PER_MILLISECOND + timestamp.getNanos() % NANOSECONDS_PER_MILLISECOND;
+                    int picosOfMilli = (int) ((nanosUtc - millisUtc * NANOSECONDS_PER_MILLISECOND) * PICOSECONDS_PER_NANOSECOND);
+
+                    return LongTimestampWithTimeZone.fromEpochMillisAndFraction(
+                            millisUtc,
+                            picosOfMilli,
+                            TimeZoneKey.getTimeZoneKey(TimeZone.getDefault().getID()));
+                });
+    }
+
+    private static void verifyObjectTimestampWithTimeZonePrecision(TimestampWithTimeZoneType timestampType)
+    {
+        int precision = timestampType.getPrecision();
+        checkArgument(precision > TimestampWithTimeZoneType.MAX_SHORT_PRECISION && precision <= MAX_EXASOL_TIMESTAMP_PRECISION,
+                "Precision is out of range: %s", precision);
+    }
+
+    private static ObjectWriteFunction objectTimestampWithTimeZoneWriteFunction(TimestampWithTimeZoneType timestampType)
+    {
+        verifyObjectTimestampWithTimeZonePrecision(timestampType);
+
+        return new ObjectWriteFunction() {
+            @Override
+            public Class<?> getJavaType()
+            {
+                return LongTimestampWithTimeZone.class;
+            }
+
+            @Override
+            public void set(PreparedStatement statement, int index, Object value)
+                    throws SQLException
+            {
+                LongTimestampWithTimeZone longTimestampWithTimeZone = (LongTimestampWithTimeZone) value;
+                Instant instant = Instant.ofEpochMilli(longTimestampWithTimeZone.getEpochMillis())
+                        .plusNanos(longTimestampWithTimeZone.getPicosOfMilli() / PICOSECONDS_PER_NANOSECOND);
+                Timestamp timestamp = Timestamp.from(instant);
+                statement.setObject(index, timestamp);
+            }
+
+            @Override
+            public String getBindExpression()
+            {
+                return getTimestampWithTimeZoneBindExpression(timestampType);
+            }
+
+            @Override
+            public void setNull(PreparedStatement statement, int index)
+                    throws SQLException
+            {
+                statement.setNull(index, Types.VARCHAR);
+            }
+        };
+    }
+
+    private static LongWriteFunction longTimestampWithTimeZoneWriteFunction(TimestampWithTimeZoneType timestampType)
+    {
+        return new LongWriteFunction()
+        {
+            @Override
+            public String getBindExpression()
+            {
+                return getTimestampWithTimeZoneBindExpression(timestampType);
+            }
+
+            @Override
+            public void set(PreparedStatement statement, int index, long dateTimeWithTimeZone)
+                    throws SQLException
+            {
+                long epochMillis = unpackMillisUtc(dateTimeWithTimeZone);
+                Timestamp timestampValue = Timestamp.from(Instant.ofEpochMilli(epochMillis));
+                statement.setObject(index, timestampValue);
+            }
+
+            @Override
+            public void setNull(PreparedStatement statement, int index)
+                    throws SQLException
+            {
+                statement.setNull(index, Types.VARCHAR);
+            }
+        };
+    }
+
+    private static String getTimestampBindExpression(TimestampType timestampType)
+    {
+        return getTimestampBindExpression(timestampType.getPrecision());
+    }
+
+    private static String getTimestampWithTimeZoneBindExpression(TimestampWithTimeZoneType timestampWithTimeZoneType)
+    {
+        return getTimestampBindExpression(timestampWithTimeZoneType.getPrecision());
+    }
+
+    /**
+     * Returns a {@code TO_TIMESTAMP} bind expression using the appropriate format model
+     * based on the given fractional seconds precision.
+     * See for more details: https://docs.exasol.com/db/latest/sql_references/formatmodels.htm
+     */
+    private static String getTimestampBindExpression(int precision)
+    {
+        //negative precisions are not supported by TimestampType and TimestampWithTimeZoneType
+        if (precision == 0) {
+            return "TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS')";
+        }
+        return format("TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS.FF%d')", precision);
     }
 
     private static ColumnMapping dateColumnMapping()
--- a/plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
+++ b/plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
@@ -40,12 +40,15 @@ import static io.trino.spi.type.CharType.createCharType;
 import static io.trino.spi.type.DateType.DATE;
 import static io.trino.spi.type.DecimalType.createDecimalType;
 import static io.trino.spi.type.DoubleType.DOUBLE;
+import static io.trino.spi.type.TimestampType.createTimestampType;
+import static io.trino.spi.type.TimestampWithTimeZoneType.createTimestampWithTimeZoneType;
 import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static io.trino.spi.type.VarcharType.createVarcharType;
 import static io.trino.testing.TestingNames.randomNameSuffix;
 import static java.lang.String.format;
 import static java.time.ZoneOffset.UTC;
 import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 @Isolated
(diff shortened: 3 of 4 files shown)


plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -28,6 +28,8 @@
 import io.trino.plugin.jdbc.JdbcTypeHandle;
 import io.trino.plugin.jdbc.LongReadFunction;
 import io.trino.plugin.jdbc.LongWriteFunction;
+import io.trino.plugin.jdbc.ObjectReadFunction;
```
ebyhr, line 33: The 1st commit contains redundant commit body. Please remove it.

testing/trino-product-tests/src/main/java/io/trino/tests/product/exasol/TestExasol.java
```diff
@@ -17,6 +17,7 @@
 import org.testng.annotations.Test;
 
 import java.math.BigDecimal;
+import java.sql.Timestamp;
```
ebyhr: > - implemented testing of timestamp columns in the product test

I would drop this commit. 

Product tests are useful for catching class loader issues. `TestExasolTypeMapping` is sufficient in this case. 

https://trino.io/docs/current/develop/tests.html#avoid-product-tests

docs/src/main/sphinx/connector/exasol.md
```diff
@@ -101,6 +101,13 @@ Trino data type mapping:
   * - ``DATE``
     - ``DATE``
     -
+  * - ``TIMESTAMP(n)``
```
ebyhr: > Updated documentation with new TIMESTAMP type mappings

Please squash this commit into the 1st commit. We don't separate a commit for documentation.

docs/src/main/sphinx/connector/exasol.md
```diff
@@ -127,6 +134,34 @@ If no length is specified, the connector uses 2000000.
 Trino's `CHAR(n)` maps to `CHAR(n)` and vice versa if `n` is no greater than 2000.
 Exasol does not support longer values.
 
+(exasol-timestamp-mapping)=
+### Mapping `TIMESTAMP` Types
+
+Exasol `TIMESTAMP(n)` columns are mapped to Trino's `TIMESTAMP(n)` type and vice versa, with the following exceptions:
+
+- **No precision specified**:  
+  If the precision is omitted (i.e., the column is defined as `TIMESTAMP` without `(n)`), Exasol defaults to a precision of 3. This maps to Trino's `TIMESTAMP(3)`.
```
ebyhr: Please wrap at 80 characters. Same for other places.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +282,257 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+                LongTimestampWithTimeZone.class,
+                (resultSet, columnIndex) -> {
+                    Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+                    if (timestamp == null) {
+                        return null;
+                    }
+
+                    long millisUtc = timestamp.getTime();
+                    long nanosUtc = millisUtc * 1_000_000L + timestamp.getNanos() % 1_000_000;
+                    int picosOfMilli = (int) ((nanosUtc - millisUtc * 1_000_000) * 1_000);
```
ebyhr: There are some constants in `Timestamps`.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +282,257 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+            TimestampWithTimeZoneType timestampType)
+    {
+        verifyObjectTimestampWithTimeZonePrecision(timestampType);
+        return ObjectReadFunction.of(
+                LongTimestampWithTimeZone.class,
+                (resultSet, columnIndex) -> {
+                    Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+                    if (timestamp == null) {
+                        return null;
+                    }
```
ebyhr: Why do we need this null handling? `ReadFunction` has `isNull` method.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +282,257 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+
+                long epochMillis = timestampValue.getEpochMillis();
+                int picosOfMilli = timestampValue.getPicosOfMilli();
+                int nanos = picosOfMilli / PICOSECONDS_PER_NANOSECOND;
+
+                Instant instant = Instant.ofEpochMilli(epochMillis).plusNanos(nanos);
+                ZonedDateTime zonedDateTime = instant.atZone(UTC);
+
+                String formatted = TIMESTAMP_NANO_OPTIONAL_FORMATTER.format(zonedDateTime);
+                statement.setString(index, formatted);
```
ebyhr: Could you leave a code comment why using `setString` instead  of other setters such as `setObject`?

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +282,257 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+            }
+        };
+    }
+
+    private static String getTimestampBindExpression(int precision)
+    {
+        if (precision <= 0) {
+            return "TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS')";
+        }
+        else {
```
ebyhr: Remove redundant else. https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#additional-ide-configuration

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,142 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("cast(col_24 AT TIME ZONE 'UTC' AS timestamp(1) with time zone)", "timestamp(1) with local time zone", "TIMESTAMP '2020-09-27 12:34:56.9'", createTimestampWithTimeZoneType(1), "TIMESTAMP '2020-09-27 17:34:56.9 UTC'")
+                .addRoundTrip("cast(col_25 AT TIME ZONE 'UTC' AS timestamp(3) with time zone)", "timestamp(3) with local time zone", "TIMESTAMP '2020-09-27 12:34:56.123'", createTimestampWithTimeZoneType(3), "TIMESTAMP '2020-09-27 17:34:56.123 UTC'")
+                .addRoundTrip("cast(col_26 AT TIME ZONE 'UTC' AS timestamp(3) with time zone)", "timestamp(3) with local time zone", "TIMESTAMP '2020-09-27 12:34:56.999'", createTimestampWithTimeZoneType(3), "TIMESTAMP '2020-09-27 17:34:56.999 UTC'")
+                .addRoundTrip("cast(col_27 AT TIME ZONE 'UTC' AS timestamp(6) with time zone)", "timestamp(6) with local time zone", "TIMESTAMP '2020-09-27 12:34:56.123456'", createTimestampWithTimeZoneType(6), "TIMESTAMP '2020-09-27 17:34:56.123456 UTC'");
+
+        test.execute(getQueryRunner(), session, exasolCreateAndInsert(TEST_SCHEMA + "." + "test_timestamp"));
+    }
+
+    private void testTimestamp(ZoneId sessionZone)
```
ebyhr, line 296: Move this method under `testTimestamp()`.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,142 @@ private void testDate(ZoneId sessionZone)
...
+     * <p>
+     * <b>Important timezone note:</b> The input timestamp string literal is interpreted as a timestamp
+     * in the current JVM time zone ("America/Bahia_Banderas" in this test environment).
+     * The expected timestamp literal is interpreted as a UTC timestamp.
+     * Because of this, the expected literal is offset by 6 hours (during daylight saving time),
+     * 5 hours (outside DST), or 7 hours for some historical dates in 1970 due to timezone rules.
+     * This difference is intentional and verifies correct conversion between zones.
+     * <p>
+     */
+    private void testTimestampWithTimeZone(ZoneId sessionZone)
```
ebyhr: Move this method under `testTimestampWithTimeZone()`.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,142 @@ private void testDate(ZoneId sessionZone)
...
+    @Test
+    void testTimestampWithTimeZone()
+    {
+        //Only JVM Time Zone is currently used for testing timestamp with time zone
+        //Adding test cases for other time zones would require improving the test to forcibly apply session changes for timestamp with timezone
+        testTimestampWithTimeZone(ZoneId.systemDefault());
+    }
+
+    @Test
+    void testUnsupportedTimestampValues()
```
ebyhr: Could you add a similar test with `timestamp with time zone` type?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,142 @@ private void testDate(ZoneId sessionZone)
...
+
+            // Too late
+            assertExasolQueryFails(
+                    "INSERT INTO " + table.getName() + " VALUES (TIMESTAMP '10000-01-01 00:00:00.000000')",
+                    "data exception - invalid character value for cast");
+
+            // Precision > 9
+            assertExasolQueryFails(
+                    "INSERT INTO " + table.getName() + " VALUES (TIMESTAMP '2024-01-01 12:34:56.1234567890')",
+                    "data exception - invalid character value for cast");
```
ebyhr: This should round values instead of failures in my opinion.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,142 @@ private void testDate(ZoneId sessionZone)
...
+        testTimestamp(vilnius);
+        testTimestamp(kathmandu);
+        testTimestamp(TestingSession.DEFAULT_TIME_ZONE_KEY.getZoneId());
+    }
+
+    @Test
+    void testTimestampWithTimeZone()
+    {
+        //Only JVM Time Zone is currently used for testing timestamp with time zone
+        //Adding test cases for other time zones would require improving the test to forcibly apply session changes for timestamp with timezone
```
ebyhr: I don't understand this comment. Can you elaborate on that?

skyglass: Exasol Timestamp with Local Time Zone converts timestamp from session time zone to UTC when saving to database and correspondently converts to session time zone, when retrieving from database.
Currently, the session time zone in the Trino testing environment is "America/Bahia_Bandera"
Current test assumes "America/Bahia_Bandera" time zone and makes correspondent assumptions, which take into account difference of 6 hours (with DST) and 5 hours (without DST) from UTC.

Unfortunately, dynamically changing session time zone at the runtime, doesn't seem to work in the tests right now. Most probably, it is because after changing the session, all current connections must be closed. Session changes are applied only for new fresh connections. The tests seem to not properly refresh the connections after changing the session. This is a testing framework issue, so I decided to study this issue in a separate PR. There is a github issue for that, but I can't share the link in the code, because it is a private repository.

In any case, testing timestamp with local time zone only for one time zone should be enough, but I will try to improve the tests in a separate PR, after fixing the issue with the session.

I will update the comment in the code with this information. Hopefully, it explains the current decision.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
skyglass: It is a default exasol core behaviour. The exception happens when the query is compiled on Exasol Sever. Unfortunately, I can't change it.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: Thank you for noticing!
Exasol actually supports setting Timestamp values with setObject.
I have changed objectTimestampWithTimeZoneWriteFunction and objectTimestampWriteFunction to use setObject with Timestamp.
Now the code is much simpler, thank you!

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,155 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("cast(col_11 AT TIME ZONE 'UTC' AS timestamp(6) with time zone)", "timestamp(6) with local time zone", "TIMESTAMP '1970-01-01 00:13:42.000000'", createTimestampWithTimeZoneType(6), "TIMESTAMP '1970-01-01 07:13:42.000000 UTC'")
+                .addRoundTrip("cast(col_12 AT TIME ZONE 'UTC' AS timestamp(6) with time zone)", "timestamp(6) with local time zone", "TIMESTAMP '2018-04-01 02:13:55.123456'", createTimestampWithTimeZoneType(6), "TIMESTAMP '2018-04-01 08:13:55.123456 UTC'")
+                .addRoundTrip("cast(col_13 AT TIME ZONE 'UTC' AS timestamp(6) with time zone)", "timestamp(6) with local time zone", "TIMESTAMP '2018-03-25 03:17:17.000000'", createTimestampWithTimeZoneType(6), "TIMESTAMP '2018-03-25 09:17:17.000000 UTC'")
+                .addRoundTrip("cast(col_14 AT TIME ZONE 'UTC' AS timestamp(6) with time zone)", "timestamp(6) with local time zone", "TIMESTAMP '1986-01-01 00:13:07.000000'", createTimestampWithTimeZoneType(6), "TIMESTAMP '1986-01-01 07:13:07.000000 UTC'")
+
+                // tests for other precisions (0-5 and some 1's)
```
chenjian2664, line 403: Can we have tests for precision greater than 6?

skyglass: Yes, sure. Thank you for noticing!
I have added the tests for precisions 7, 8 and 9 for timestamp and timestamp with time zone.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -480,11 +481,11 @@ public String getBindExpression()
             }
 
             @Override
-            public void set(PreparedStatement statement, int index, long epochMicros)
+            public void set(PreparedStatement statement, int index, long dateTimeWithTimeZone)
                     throws SQLException
             {
-                LocalDateTime localDateTime = fromTrinoTimestamp(epochMicros);
-                Timestamp timestampValue = Timestamp.valueOf(localDateTime);
+                long epochMillis = unpackMillisUtc(dateTimeWithTimeZone);
```
skyglass, line 505: During final debugging it was discovered that longTimestampWithTimeZoneWriteFunction is never called. Therefore, I had to fix the tests, which fixed the calling of longTimestampWithTimeZoneWriteFunction.
After fixing the test, the bug with longTimestampWithTimeZoneWriteFunction has been discovered and fixed.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -343,42 +344,51 @@ void testTimestampWithTimeZone()
...
-                .addRoundTrip("cast(col_7 AT TIME ZONE 'UTC' AS timestamp with time zone)", "timestamp(3) with local time zone", "TIMESTAMP '1986-01-01 00:13:07.000'", createTimestampWithTimeZoneType(3), "TIMESTAMP '1986-01-01 07:13:07.000 UTC'")
+                .addRoundTrip("timestamp with local time zone", "TIMESTAMP '2019-03-18 10:01:17.123'", createTimestampWithTimeZoneType(3), "TIMESTAMP '2019-03-18 16:01:17.123 UTC'")
+                .addRoundTrip("timestamp(3) with local time zone", "TIMESTAMP '2018-10-28 01:33:17.456'", createTimestampWithTimeZoneType(3), "TIMESTAMP '2018-10-28 07:33:17.456 UTC'")
+                .addRoundTrip("timestamp(3) with local time zone", "TIMESTAMP '2018-10-28 03:33:33.333'", createTimestampWithTimeZoneType(3), "TIMESTAMP '2018-10-28 09:33:33.333 UTC'")
+                .addRoundTrip("timestamp(3) with local time zone", "TIMESTAMP '1970-01-01 00:13:42.000'", createTimestampWithTimeZoneType(3), "TIMESTAMP '1970-01-01 07:13:42.000 UTC'")
+                //'2018-04-01 02:13:55.123' is a DST gap time, which doesn't "exist" in the session time zone
```
skyglass: The issue with "DST gap time" has been discovered by the fixed test, which might point to the incorrect conversion in exasol core db server, or it might be the issue with the testing framework. The test is disabled until further research.

skyglass: In practice, the dst gap times should not happen, unless the timestamp is invalid, so this test case is not relevant and can be safely ignored

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +270,242 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+        return new LongWriteFunction()
+        {
+            @Override
+            public String getBindExpression()
+            {
+                return getTimestampBindExpression(timestampType.getPrecision());
+            }
+
+            @Override
+            public void set(PreparedStatement statement, int index, long dateTimeWithTimeZone)
```
skyglass, line 484: Any updates on the review?
I kept the separate commit to explain the latest changes and the bug, which was revealed by tests and fixed in this commit. I will combine this commit with the main one, as soon as you confirm the last changes.

skyglass:
Any updates on the review?
Please, let me know, if anything is missing or required.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -61,25 +61,25 @@ protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
         return switch (connectorBehavior) {
             // Tests requires write access which is not implemented
             case SUPPORTS_AGGREGATION_PUSHDOWN,
-                 SUPPORTS_JOIN_PUSHDOWN,
-                 SUPPORTS_LIMIT_PUSHDOWN,
-                 SUPPORTS_TOPN_PUSHDOWN -> false;
+                    SUPPORTS_JOIN_PUSHDOWN,
```
chenjian2664: revert

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -61,25 +61,25 @@ protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
...
-                 SUPPORTS_DELETE,
-                 SUPPORTS_INSERT,
-                 SUPPORTS_MAP_TYPE,
-                 SUPPORTS_NEGATIVE_DATE, // min date is 0001-01-01
-                 SUPPORTS_RENAME_COLUMN,
-                 SUPPORTS_RENAME_TABLE,
-                 SUPPORTS_ROW_TYPE,
-                 SUPPORTS_SET_COLUMN_TYPE,
-                 SUPPORTS_UPDATE -> false;
+                    SUPPORTS_ARRAY,
```
chenjian2664: revert

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,180 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("timestamp(6) with local time zone", "TIMESTAMP '2020-09-27 12:34:56.123456'", createTimestampWithTimeZoneType(6), "TIMESTAMP '2020-09-27 17:34:56.123456 UTC'");
+
+        test.execute(getQueryRunner(), session, exasolCreateAndInsert(TEST_SCHEMA + "." + "test_timestamp"));
+    }
+
+    @Test
+    void testUnsupportedTimestampValues()
+    {
+        try (TestTable table = new TestTable(exasolServer::execute, "tpch.test_unsupported_timestamp", "(col TIMESTAMP)")) {
+            // Too early
```
chenjian2664: Could you please provide an explanation of why it is considered ‘too early’? A link to the documentation would be even better

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,180 @@ private void testDate(ZoneId sessionZone)
...
+            assertExasolQueryFails(
+                    "INSERT INTO " + table.getName() + " VALUES (TIMESTAMP '0000-12-31 23:59:59.999999')",
+                    "data exception - invalid date value");
+
+            // Too late
```
chenjian2664: Same as above

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +270,242 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+            }
+        };
+    }
+
+    private static String getTimestampBindExpression(int precision)
+    {
+        if (precision <= 0) {
+            return "TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS')";
+        }
+        return format("TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS.FF%d')", precision);
```
chenjian2664, line 540: Where I can know Exasol 's timestamp type expression format?

skyglass: See for more details: https://docs.exasol.com/db/latest/sql_references/formatmodels.htm
Added correspondent javadocs with the link.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +270,247 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+    }
+
+    /**
+     * Returns a {@code TO_TIMESTAMP} bind expression using the appropriate format model
+     * based on the given fractional seconds precision.
+     * See for more details: https://docs.exasol.com/db/latest/sql_references/formatmodels.htm
+     */
+    private static String getTimestampBindExpression(int precision)
+    {
+        if (precision <= 0) {
```
ebyhr: When does this method take negative precisions?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,182 @@ private void testDate(ZoneId sessionZone)
...
+        test.execute(getQueryRunner(), session, exasolCreateAndInsert(TEST_SCHEMA + "." + "test_timestamp"));
+    }
+
+    @Test
+    // See: https://docs.exasol.com/db/latest/sql_references/data_types/datatypedetails.htm for more details
+    void testUnsupportedTimestampValues()
+    {
+        try (TestTable table = new TestTable(exasolServer::execute, "tpch.test_unsupported_timestamp", "(col TIMESTAMP)")) {
+            // Below minimum supported TIMESTAMP value (must be >= 0001-01-01)
+            assertExasolQueryFails(
```
ebyhr: We should verify the behavior of the connector (=running the similar query on Trino), too.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: Not that I'm aware of any cases of exasol providing negative precisions, but I remember oracle jdbc driver might return "magic" negative precision value in case of some issues. In any case, negative precision would point to some issues and the question is should we throw exception or try to handle the value using default precision. I would prefer the second approach.

Although negative precision metadata values are not possible for exasol jdbc connector (not that I'm aware of or could find documented), I would still keep the handling of the negative precision here.

But if you think that I should remove handling of negative values, so that it doesn't confuse the developers, then I would be fine to remove it.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
skyglass: Not sure, if these values are possible to read with trino exasol connector, because there is no way to insert such values to exasol in the first place. But I have improved the tests to initiate the round trip, like in other tests. I haven't debugged the behaviour, but I think that trino connector is still not involved in these tests, because of the exception, but in any case, the tests look much better now.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
ebyhr: > should we throw exception or try to handle the value using default precision. I would prefer the second approach.

We should throw exceptions for unexpected scenarios.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -275,6 +277,183 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("timestamp", "TIMESTAMP '3030-03-03 12:34:56.123'", createTimestampType(3), "TIMESTAMP '3030-03-03 12:34:56.123'")
+                .addRoundTrip("timestamp(5)", "TIMESTAMP '3030-03-03 12:34:56.12345'", createTimestampType(5), "TIMESTAMP '3030-03-03 12:34:56.12345'")
+                .addRoundTrip("timestamp(6)", "TIMESTAMP '3030-03-03 12:34:56.123456'", createTimestampType(6), "TIMESTAMP '3030-03-03 12:34:56.123456'")
+                .addRoundTrip("timestamp(7)", "TIMESTAMP '3030-03-03 12:34:56.1234567'", createTimestampType(7), "TIMESTAMP '3030-03-03 12:34:56.1234567'")
+                .addRoundTrip("timestamp(8)", "TIMESTAMP '3030-03-03 12:34:56.12345678'", createTimestampType(8), "TIMESTAMP '3030-03-03 12:34:56.12345678'")
+                .addRoundTrip("timestamp(9)", "TIMESTAMP '3030-03-03 12:34:56.123456789'", createTimestampType(9), "TIMESTAMP '3030-03-03 12:34:56.123456789'")
+                .addRoundTrip("timestamp(0)", "TIMESTAMP '2017-07-01'", createTimestampType(0), "TIMESTAMP '2017-07-01'") // summer on northern hemisphere (possible DST)
```
chenjian2664, line 326: Please add timestamp with seconds values test cases

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: > We should throw exceptions for unexpected scenarios.

I added tests for negative precisions, which throw correspondent exception.
No changes are required for the code, the exception is already thrown by Trino.
But I will also remove negative precision check, so that users are not confused by negative precision possibility.

yelianevich:
Hello everyone! Is anything missing on this PR? Looking forward to having it merged!

skyglass:
Thank you very much for your reviews, @ebyhr , @chenjian2664 ! 👍 

Any updates on the PR?
Please, let me know, if anything is missing or required.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +280,212 @@ private void testDate(ZoneId sessionZone)
                 .execute(getQueryRunner(), session, exasolCreateAndInsert(TEST_SCHEMA + "." + "test_date"));
     }
 
+    // See for more details: https://docs.exasol.com/saas/microcontent/Resources/MicroContent/general/hash-data-type.htm
```
chenjian2664: The link seems wrong

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +280,212 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("timestamp(0) with local time zone", "TIMESTAMP '1970-01-01 00:00:03'", createTimestampWithTimeZoneType(0), "TIMESTAMP '1970-01-01 07:00:03 UTC'") // change forward at midnight in JVM
+                .addRoundTrip("timestamp(0) with local time zone", "TIMESTAMP '1983-04-01 00:00:04'", createTimestampWithTimeZoneType(0), "TIMESTAMP '1983-04-01 07:00:04 UTC'") // change forward at midnight in Vilnius
+                .addRoundTrip("timestamp(0) with local time zone", "TIMESTAMP '1983-10-01 00:00:05'", createTimestampWithTimeZoneType(0), "TIMESTAMP '1983-10-01 07:00:05 UTC'") // change backward at midnight in Vilnius
+                .addRoundTrip("timestamp(0) with local time zone", "TIMESTAMP '9999-12-31 00:00:59'", createTimestampWithTimeZoneType(0), "TIMESTAMP '9999-12-31 06:00:59 UTC'"); // max value in Exasol
+
+        test.execute(getQueryRunner(), session, exasolCreateAndInsert(TEST_SCHEMA + "." + "test_timestamp"));
```
chenjian2664: Maybe we could simplify by appending .execute directly instead of keeping the test variable

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +280,212 @@ private void testDate(ZoneId sessionZone)
...
+    }
+
+    /**
+     * <p>
+     * Exasol Timestamp with Local Time Zone interprets timestamp string as JVM time zone.
+     * Currently, the jvm time zone in the Trino testing environment is "America/Bahia_Bandera"
+     * Current test assumes "America/Bahia_Bandera" time zone and makes correspondent assertions,
+     * which take into account difference of 6 hours (with DST) and 5 hours (without DST) from UTC.
+     * String parameter "inputLiteral" in the "addRoundTrip" method
+     * is interpreted as "America/Bahia_Bandera" timestamp string, which must be equal to
```
chenjian2664: Please add test cases that set expectedLiteral to the "America/Bahia_Bandera" timezone to confirm that the value literal matches

docs/src/main/sphinx/connector/exasol.md
```diff
@@ -103,6 +103,12 @@ Trino data type mapping:
 * - `HASHTYPE`
   - `VARBINARY`
   -
+* - ``TIMESTAMP(n)``
+  - ``TIMESTAMP(n)``
+  -
+* - ``TIMESTAMP(n) WITH LOCAL TIME ZONE``
+  - ``TIMESTAMP(n) WITH TIME ZONE``
+  -
```
ebyhr: Please change to single backtick.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
skyglass: Sorry, it was accidentally replaced during merge conflict resolution. Thank you for noticing!

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
skyglass: Unfortunately, the name "timestamp with local time zone" is confusing. This exasol type doesn't store the time zone. It corresponds to UTC Trino "Timestamp With Time Zone".

For example, in **objectTimestampWithTimeZoneReadFunction** and **longTimestampWithTimeZoneReadFunction** the returned trino timestamp with time zone always have hardcoded "UTC" time zone.

This limitation puts some restrictions on the tests. For example, I can't use any other than "UTC" time zone for the expectedLiteral.

Therefore, I can't change test cases to set expectedLiteral as "America/Bahia_Bandera" timezone (which is a JVM time zone for the testing environment)

But I have improved test cases by interpreting inputLiteral and expectedLiteral as "UTC" timezone.

For each round trip, **timestampWithTimeZoneInputLiteral** and **timestampWithTimeZoneExpectedLiteral** pass the same UTC string as parameter.

skyglass: I have updated the javadocs for this method correspondingly.

skyglass (commented):
I decided to move this PR to DRAFT and replace it with the new PR, with more simple, more efficient and less error-prone solution: https://github.com/trinodb/trino/pull/26633

Please, review the new PR, @ebyhr , @chenjian2664 ! 🙏 

After reviewing the implementation, I believe the current approach - mapping Exasol TIMESTAMP WITH LOCAL TIME ZONE to Trino TIMESTAMP WITH TIME ZONE - adds unnecessary complexity and potential pitfalls.

Exasol TIMESTAMP WITH LOCAL TIME ZONE does not actually persist any time zone information. Apart from some minor edge cases, its semantics is very close to a plain Exasol TIMESTAMP.

Mapping to Trino TIMESTAMP WITH TIME ZONE requires hardcoding the JVM time zone, which introduces fragility. For example, values valid in the Exasol DB time zone but invalid in the JVM time zone (e.g., during DST gaps) may cause Trino to throw exceptions, potentially breaking queries.

A simpler mapping to Trino TIMESTAMP avoids these problems, is more efficient, and aligns better with how Exasol handles this type internally.

Given this, I created a new PR proposing the TIMESTAMP WITH LOCAL TIME ZONE → TIMESTAMP mapping, which I believe will result in a cleaner, simpler and more reliable solution: https://github.com/trinodb/trino/pull/26633

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -256,6 +289,257 @@ private boolean isHashType(JdbcTypeHandle typeHandle)
...
+                objectTimestampWithTimeZoneReadFunction(timestampWithTimeZoneType),
+                objectTimestampWithTimeZoneWriteFunction(timestampWithTimeZoneType),
+                FULL_PUSHDOWN);
+    }
+
+    private static LongReadFunction longTimestampWithTimeZoneReadFunction()
+    {
+        return (resultSet, columnIndex) -> {
+            Timestamp timestamp = resultSet.getObject(columnIndex, Timestamp.class);
+            return packDateTimeWithZone(timestamp.getTime(), TimeZone.getDefault().getID());
```
skyglass, line 423: In the previous implementation I used UTC timezone, instead of TimeZone.getDefault(), which is incorrect. Correct time zone here should be TimeZone.getDefault(), which is a JVM time zone. If we want to use UTC time zone here, then timestamp milliseconds must be shifted according to the difference between UTC and JVM time zone. I decided not to go with this implementation, because it would require shifting milliseconds for each read and write. Using JVM time zone here would require less code changes and less computation during runtime.

skyglass (commented):
I decided to move this PR to DRAFT and replace it with the new PR, with more simple, more efficient and less error-prone solution: https://github.com/trinodb/trino/pull/26633

After reviewing the implementation, I believe the current approach - mapping Exasol TIMESTAMP WITH LOCAL TIME ZONE to Trino TIMESTAMP WITH TIME ZONE - adds unnecessary complexity and potential pitfalls.

Exasol TIMESTAMP WITH LOCAL TIME ZONE does not actually persist any time zone information. Apart from some minor edge cases, its semantics is very close to a plain Exasol TIMESTAMP.

Mapping to Trino TIMESTAMP WITH TIME ZONE requires hardcoding the JVM time zone, which introduces fragility. For example, values valid in the Exasol DB time zone but invalid in the JVM time zone (e.g., during DST gaps) may cause Trino to throw exceptions, potentially breaking queries.

A simpler mapping to Trino TIMESTAMP avoids these problems, is more efficient, and aligns better with how Exasol handles this type internally.

Given this, I created a new PR proposing the TIMESTAMP WITH LOCAL TIME ZONE → TIMESTAMP mapping, which I believe will result in a cleaner, simpler and more reliable solution: https://github.com/trinodb/trino/pull/26633

skyglass:
Closed as outdated. See the new PR with improved solution for Exasol TIMESTAMP WITH LOCAL TIME ZONE: https://github.com/trinodb/trino/pull/26633

skyglass:
Closed as outdated. See the new PR with improved solution for Exasol TIMESTAMP WITH LOCAL TIME ZONE: https://github.com/trinodb/trino/pull/26633

chenjian2664:
Does the timestamp part is the same as new pr https://github.com/trinodb/trino/pull/26259 ?

skyglass:
> Does the timestamp part is the same as new pr #26259 ?

The new PR is here: https://github.com/trinodb/trino/pull/26633

The main difference is that the new PR maps Exasol TIMESTAMP WITH LOCAL TIME ZONE to Trino TIMESTAMP.
The new PR is more simple.

This current PR was closed, because it was mapping Exasol TIMESTAMP WITH LOCAL TIME ZONE to Trino TIMESTAMP WITH TIME ZONE, which is not necessary, because Exasol TIMESTAMP WITH LOCAL TIME ZONE doesn't actually store the time zone information.

chenjian2664:
If it just about TIMESTAMP WITH LOCAL TIME ZONE, we could split to support TIMESTAMP first

skyglass:
> Does the timestamp part is the same as new pr #26259 ?

The new PR, which contains only TIMESTAMP changes, has been created:
https://github.com/trinodb/trino/pull/26963

PR #26266, opened 2025-07-23
Convert QueryEntity to record in MySQL event listener

ebyhr:
## Release notes

### DIFF
--- a/plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/RecordAndAnnotatedConstructorMapper.java
+++ b/plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/RecordAndAnnotatedConstructorMapper.java
@@ -1,0 +1,43 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


chenjian2664 (approved):

wendigo (approved):

plugin/trino-mysql-event-listener/src/main/java/io/trino/plugin/eventlistener/mysql/RecordAndAnnotatedConstructorMapper.java
```diff
@@ -0,0 +1,43 @@
...
+import org.jdbi.v3.core.mapper.RowMapper;
+import org.jdbi.v3.core.mapper.RowMapperFactory;
+import org.jdbi.v3.core.mapper.reflect.ConstructorMapper;
+import org.jdbi.v3.core.mapper.reflect.JdbiConstructor;
+
+import java.lang.reflect.Type;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+public class RecordAndAnnotatedConstructorMapper
```
ksobolew, line 26: Why not use `ConstructorMapper` directly?

ebyhr: It requires a row mapper registration per record class, right?

ksobolew: Yes, and JDBI has specialized ways to register `ConstructorMapper` specifically, like `@RegisterConstructorMapper`

PR #26136, opened 2025-07-07
Add support for write operations (INSERT/UPDATE/DELETE/MERGE) with branch

ebyhr:
## Description

Allow specifying a branch name like `{table}@{branch}` in INSERT/UPDATE/DELETE/MERGE statements:
```g4
    | INSERT INTO qualifiedName ('@' branch=identifier)?
       columnAliases? rootQuery                                        #insertInto
    | DELETE FROM qualifiedName ('@' branch=identifier)?
         (WHERE booleanExpression)?                                    #delete
    | UPDATE qualifiedName ('@' branch=identifier)?
        SET updateAssignment (',' updateAssignment)*
        (WHERE where=booleanExpression)?                               #update
    | MERGE INTO
        qualifiedName ('@' branch=identifier)? (AS? alias=identifier)?
        USING relation ON expression mergeCase+                        #merge
```
 
Relates to https://docs.google.com/document/d/1jEF4IkWu-2Gzk5ii2Nb0exuEnAUeo98UbiM3i0xtgWQ

## Release notes

```markdown
## General
* Add support for branching with `INSERT`, `DELETE`, `UPDATE`, and `MERGE` statements. ({issue}`26136`)
```

PR #26271, opened 2025-07-24
Replace start and end version for branching

ebyhr:
## Description

Follow-of of https://github.com/trinodb/trino/pull/26136

The branch name in `FOR VERSION AS OF '{branch}'` is passed as end version. 
We should follow the same behavior. 

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
@@ -587,16 +587,16 @@ class StatementAnalyzer
             // analyze the query that creates the data
             Scope queryScope = analyze(insert.getQuery(), Optional.empty(), false);
 
-            Optional<TableVersion> startVersion = Optional.empty();
+            Optional<TableVersion> endVersion = Optional.empty();
             if (insert.getTable().getBranch().isPresent()) {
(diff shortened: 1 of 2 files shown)


martint:
Can you explain a bit more what this change is trying to achieve? For AS OF VERSION and @, there is no start or end. If it’s a branch, it’s the state of that branch. If it’s a snapshot, it’s the state of try table at that point in time.

findinpath:
> We should follow the behavior in @ syntax.

Please add more details in the description.

ebyhr:
> For AS OF VERSION and @, there is no start or end

[QueryPeriod](https://github.com/trinodb/trino/blob/master/core/trino-parser/src/main/java/io/trino/sql/tree/QueryPeriod.java) and [SPI](https://github.com/trinodb/trino/blob/f574832d2a6fe177efd968bfd9af2cc0ca176c42/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java#L109-L113) have "start" and "end" concept though only "end" is used for now. Connectors don't expect receiving the start version in `ConnectorMetadata.getTableHandle` method (e.g. [Iceberg](https://github.com/trinodb/trino/blob/f574832d2a6fe177efd968bfd9af2cc0ca176c42/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java#L541-L550)), so this PR fixes the wrong parameter order.

PR #26273, opened 2025-07-24
Remove unnecessary usage of SimplePagesHashStrategy

raunaqmorarka:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/PagesIndex.java
+++ b/core/trino-main/src/main/java/io/trino/operator/PagesIndex.java
@@ -50,7 +50,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalDouble;
-import java.util.OptionalInt;
 import java.util.function.Supplier;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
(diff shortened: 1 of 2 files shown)


losipiuk:
I am fine with removing fallback to non-compiled version of pageshashstrategy but the commit message does not state that this commit does that. Please update.

losipiuk (approved):

core/trino-main/src/main/java/io/trino/operator/PagesIndex.java
```diff
@@ -466,22 +463,8 @@ public PagesHashStrategy createPagesHashStrategy(List<Integer> joinChannels)
...
     {
-        try {
-            return joinCompiler.compilePagesHashStrategyFactory(types, joinChannels, outputChannels)
-                    .createPagesHashStrategy(ImmutableList.copyOf(channels));
-        }
-        catch (Exception e) {
-            log.error(e, "Lookup source compile failed for types=%s error=%s", types, e);
-        }
-
-        // if compilation fails, use interpreter
```
pettyjamesm: Removing this path seems like it will break queries where the number of join columns exceeds the JVM class or method size limit (that currently rely on falling back to `SimplePagesHashStrategy`). The other removal site makes sense so long as the generated `PagesHashStrategy` for empty join channels is equivalent, but this one I'm less sure about. `SimplePagesHashStrategy` isn't especially complex, so I'm not sure I see an argument for this being a necessary simplification.

raunaqmorarka: I assumed that the number of join channels it would take to trigger this would be unusual for real queries, and if it did happen the execution would be quite slow and we'd rather want to find out about this and make the compiler work for that case (e.g. like in FlatHashStrategy).
But I'm reverting this for now.

PR #26279, opened 2025-07-25
Use icebergScanExecutor for reading manifests in optimize and expire_snapshots

raunaqmorarka:
## Description
Reduces bottleneck on the default iceberg worker pool

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2029,6 +2029,7 @@ public class IcebergMetadata
         // Set dataSequenceNumber to avoid contention between OPTIMIZE and concurrent writing of equality deletes
         rewriteFiles.dataSequenceNumber(snapshot.sequenceNumber());
         rewriteFiles.validateFromSnapshot(snapshot.snapshotId());
+        rewriteFiles.scanManifestsWith(icebergScanExecutor);
         commitUpdateAndTransaction(rewriteFiles, session, transaction, "optimize");
 
         // TODO (https://github.com/trinodb/trino/issues/15439) this may not exactly be the snapshot we committed, if there is another writer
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2029,6 +2029,7 @@ private void finishOptimize(ConnectorSession session, IcebergTableExecuteHandle
         // Set dataSequenceNumber to avoid contention between OPTIMIZE and concurrent writing of equality deletes
         rewriteFiles.dataSequenceNumber(snapshot.sequenceNumber());
         rewriteFiles.validateFromSnapshot(snapshot.snapshotId());
+        rewriteFiles.scanManifestsWith(icebergScanExecutor);
```
ebyhr, line 2032: `RewriteManifests` for optimize_manifests & `RowDelta` for MERGE also have `scanManifestsWith` method. Did you avoid changing them intentionally?

wendigo: Does it make sense to use thread per task executor (VT) here?

raunaqmorarka: > `RewriteManifests` for optimize_manifests & `RowDelta` for MERGE also have `scanManifestsWith` method. Did you avoid changing them intentionally?

had missed those, updated now, thanks

raunaqmorarka: > Does it make sense to use thread per task executor (VT) here?

That is potentially a good direction. I'm concerned about throttling or some other unforeseen issue from too much parallelism. Maybe a BoundedExecutor per query on top of a shared thread pool could be done.
For now lets see if current changes help.

sopel39:
cc @osscm

PR #26281, opened 2025-07-25
Introduce `DeltaLakeFileSystemFactory` and `VendedCredentialsProvider` in Delta Lake

chenjian2664:
## Release notes

```markdown
## Section
```


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentialsHandle.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        requireNonNull(tableLocation, "tableLocation is null");
+        requireNonNull(vendedCredentials, "vendedCredentials is null");
+
+        if (catalogOwned) {
+            checkArgument(managed, "catalog-owned table must be managed");
+        }
+    }
+
+    public VendedCredentialsHandle withVendedCredentials(VendedCredentials vendedCredentials)
```
ebyhr: This method is unused.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentialsHandle.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return new VendedCredentialsHandle(catalogOwned, managed, tableLocation, Optional.of(vendedCredentials));
+    }
+
+    public static VendedCredentialsHandle empty(boolean catalogOwned, boolean managed, Optional<String> tableId, String tableLocation)
```
ebyhr: `tableId` parameter is unused. Also, the first 3 parameters are static now - `false, false, Optional.empty()`.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentialsHandle.java
chenjian2664: removed

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentialsHandle.java
chenjian2664: updated

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3845,7 +3845,7 @@ public TableStatisticsMetadata getStatisticsCollectionMetadataForWrite(Connector
         if (tableLocation != null) {
             LocatedTableHandle table = getTableHandle(session, tableMetadata.getTable(), Optional.empty(), Optional.empty());
             VendedCredentialsHandle credentialsHandle = Optional.ofNullable(table).map(LocatedTableHandle::toCredentialsHandle)
-                    .orElseGet(() -> VendedCredentialsHandle.empty(false, false, Optional.empty(), tableLocation));
+                    .orElseGet(() -> VendedCredentialsHandle.empty(tableLocation));
```
ebyhr, line 3847: Note that the fixup commit's title is wrong. You can use `git commit --fixup ...` or IDE's feature.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/AbstractDeltaLakePageSink.java
```diff
@@ -118,10 +118,11 @@ public AbstractDeltaLakePageSink(
             List<DeltaLakeColumnHandle> inputColumns,
             List<String> originalPartitionColumns,
             PageIndexerFactory pageIndexerFactory,
-            TrinoFileSystemFactory fileSystemFactory,
+            DeltaLakeFileSystemFactory fileSystemFactory,
```
ebyhr, line 121: I'm wondering how you replaced TrinoFileSystemFactory with DeltaLakeFileSystemFactory. For instance, VacuumProcedure still uses TrinoFileSystemFactory. Is it intentional? 

Also, what happens if the future developers introduce new classes with TrinoFileSystemFactory instead of DeltaLakeFileSystemFactory?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: thanks!

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/AbstractDeltaLakePageSink.java
chenjian2664: > how you replaced TrinoFileSystemFactory with DeltaLakeFileSystemFactory

honesty, I just look at every place that using `TrinoFileSystemFactory` and then replace it to `DeltaLakeFileSystemFactory`. 

> VacuumProcedure still uses TrinoFileSystemFactory. Is it intentional?

It's not intentional, I will update it to use `DeltaLakeFileSystemFactory`

> what happens if the future developers introduce new classes TrinoFileSystemFactory instead of DeltaLakeFileSystemFactory

Good question, I'll investigate on it

chenjian2664:
@ebyhr I just have a try on [ecad7c6](https://github.com/trinodb/trino/pull/26281/commits/ecad7c666fa40ef41676495333fd4c1f1e12337f) for adapting the `FileHiveMetastoreFactory` to avoid use `TrinoFileSystemFacotry` - I am afraid it's a bit overkill 
If we want to avoid user to use `TrinoFileSystemFactory` then we should replace it using `DeltaLakeFileSystemFactory` in all Delta lake modules, see at the `**MetastoreModule` 
I want to follow the Iceberg module's done, just introduce the new interface and use it by default(not remove `TrinoFileSystemFactory`)

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DefaultDeltaLakeFileSystemFactory.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public TrinoFileSystem create(ConnectorSession session, String tableLocation)
+    {
+        return fileSystemFactory.create(session.getIdentity());
+    }
+
+    @Override
+    public TrinoFileSystem create(ConnectorIdentity identity)
+    {
+        throw new UnsupportedOperationException();
```
ebyhr: Could you leave a code comment on why this method throws an exception here?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentials.java
```diff
@@ -0,0 +1,50 @@
...
+        requireNonNull(expireAt, "expireAt is null");
+        credentials = ImmutableMap.copyOf(credentials);
+    }
+
+    public static VendedCredentials empty()
+    {
+        return new VendedCredentials(Optional.empty(), Instant.MAX, ImmutableMap.of());
+    }
+
+    public boolean isFresh()
```
ebyhr: This method is unused.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/LocatedTableHandle.java
```diff
@@ -24,4 +25,6 @@ public interface LocatedTableHandle
     boolean managed();
 
     String location();
+
+    VendedCredentialsHandle toCredentialsHandle();
```
ebyhr, line 29: Do we still need the above managed() and location()? VendedCredentialsHandle contains those information in my understanding.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DefaultDeltaLakeFileSystemFactory.java
chenjian2664: I move the default impl to `DeltaLakeFileSystemFactory` and added doc there

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/LocatedTableHandle.java
chenjian2664: we are not holding the instance of `VendedCredentialsHandle` in `LocatedTableHandle` sub classes, or
do you mean change the method to default? i,e `boolean managed()` to 
```
default boolean managed() 
{ 
    return toCredentialsHandle().managed();
}
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentials.java
```diff
@@ -0,0 +1,42 @@
...
+        requireNonNull(expireAt, "expireAt is null");
+        credentials = ImmutableMap.copyOf(credentials);
+    }
+
+    public static VendedCredentials empty()
+    {
+        return new VendedCredentials(Optional.empty(), Instant.MAX, ImmutableMap.of());
+    }
+
+    private boolean isFresh(Instant now)
```
ebyhr: This method is unused.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1393,6 +1403,12 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
         }
     }
 
+    private VendedCredentialsHandle fetchCredentialsForLocation(String tableLocation)
+    {
+        // TODO: update once we support creating managed table
+        return vendedCredentialsProvider.getFreshCredentials(new VendedCredentialsHandle(false, false, tableLocation, Optional.of(VendedCredentials.empty())));
```
ebyhr: Why do we use `Optional.of(VendedCredentials.empty())` instead of `Optional.empty()`? Could you leave a code comment?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1029,13 +1033,14 @@ private RelationCommentMetadata getRelationCommentMetadata(ConnectorSession sess
             verifyDeltaLakeTable(table);
 
             String tableLocation = HiveMetastoreBackedDeltaLakeMetastore.getTableLocation(table);
+            TrinoFileSystem fileSystem = fileSystemFactory.create(session, convertToDeltaMetastoreTable(table));
```
ebyhr: A local variable for `convertToDeltaMetastoreTable(table)` can be extracted. We have the same call at L1042.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentialsHandle.java
```diff
@@ -0,0 +1,41 @@
...
+        if (catalogOwned) {
+            checkArgument(managed, "catalog-owned table must be managed");
+        }
+    }
+
+    public static VendedCredentialsHandle empty(String tableLocation)
+    {
+        return new VendedCredentialsHandle(false, false, tableLocation, Optional.empty());
+    }
+}
```
ebyhr, line 44: nit: We could another helper method building VendedCredentialsHandle from DeltaMetastoreTable:
```java
    public static VendedCredentialsHandle of(DeltaMetastoreTable table)
    {
        return new VendedCredentialsHandle(table.catalogOwned(), table.managed(), table.location(), table.vendedCredentials());
    }
```

ebyhr (approved):

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: The `Optional.empty` general means we don't support vending, `VendedCredentials.empty()` means we needs to refresh it. 
After looks at the usage of it, I think it could simplify to just use `VendedCredentials.empty()` the behavior could decide on the `vendedCredentialsProvider` as it's bit prone when to use it

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/VendedCredentialsHandle.java
chenjian2664: Nice! It's really reduce a lot of dup codes

PR #26282, opened 2025-07-25
Minor cleanup in RefreshView tests

Praveen2112:
## Description

Each connector might have a different way of handling drops and column rename, we use EXCEPT CORRESPONDING clause to ensure the results of the both the views and tables are same. Additionally fixed error message when dropping a column.

## Release notes

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
@@ -1008,7 +1008,7 @@ public abstract class BaseConnectorTest
 
         try (TestTable table = newTrinoTable("test_table", "(id BIGINT, column_to_dropped BIGINT, column_to_be_renamed BIGINT, column_with_comment BIGINT)", ImmutableList.of("1, 2, 3, 4"));
                 TestView view = new TestView(getQueryRunner()::execute, "test_view", " SELECT * FROM %s".formatted(table.getName()))) {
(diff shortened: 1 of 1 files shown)


chenjian2664:
I didn’t know that syntax until now :)

Praveen2112:
Thanks to @ebyhr for implementing it

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -1035,18 +1035,16 @@ public void testRefreshView()
...
-
-                assertUpdate("ALTER TABLE %s RENAME COLUMN column_with_comment TO renamed_new_column".formatted(table.getName()));
             }
 
             if (hasBehavior(SUPPORTS_DROP_COLUMN)) {
                 assertUpdate("ALTER TABLE %s DROP COLUMN column_to_dropped".formatted(table.getName()));
                 assertQueryFails(
                         "SELECT * FROM " + view.getName(),
-                        ".*is stale or in invalid state: stored view column count \\(5\\) does not match column count derived from the view query analysis \\(4\\)");
+                        ".*is stale or in invalid state: stored view column count \\(6\\) does not match column count derived from the view query analysis \\(5\\)");
```
ebyhr: Is this message pattern consistent regardless of SUPPORTS_COMMENT_ON_COLUMN (the block adds a new column)?

Praveen2112: Good catch !! I have moved to a regex based matching.

PR #26283, opened 2025-07-25
Reuse compiled PagesIndexOrdering

pettyjamesm:
## Description
Avoids calling into code generation logic inside of `PagesIndex#sort(List<Integer>, List<SortOrder>, int startIndex, int endIndex)` by refactoring existing usages to eagerly create and then keep an instance of `PagesIndexOrdering` for reuse. Some scenarios (such as window operators) could repeatedly call into the code generation path which creates cache contention and a severe performance hazard if the entry was evicted from cache between calls.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/function/TableFunctionOperator.java
+++ b/core/trino-main/src/main/java/io/trino/operator/function/TableFunctionOperator.java
@@ -27,6 +27,7 @@ import io.trino.operator.OperatorFactory;
 import io.trino.operator.PageBuffer;
 import io.trino.operator.PagesHashStrategy;
 import io.trino.operator.PagesIndex;
+import io.trino.operator.PagesIndexOrdering;
 import io.trino.operator.WorkProcessor;
 import io.trino.operator.function.RegularTableFunctionPartition.PassThroughColumnSpecification;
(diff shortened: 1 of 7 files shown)


core/trino-main/src/main/java/io/trino/operator/function/TableFunctionOperator.java
```diff
@@ -360,8 +362,8 @@ private static class HashStrategies
         final PagesHashStrategy prePartitionedStrategy;
         final PagesHashStrategy remainingPartitionStrategy;
         final PagesHashStrategy preSortedStrategy;
-        final List<Integer> remainingPartitionAndSortChannels;
-        final List<SortOrder> remainingSortOrders;
+        @Nullable
```
raunaqmorarka: Maybe make it an optional ?

raunaqmorarka (approved):

core/trino-main/src/main/java/io/trino/operator/function/TableFunctionOperator.java
pettyjamesm: Yeah, makes sense in this operator. Done.

PR #26284, opened 2025-07-25
Refresh Catalog Task

CarlosMorales17:
## Description

 This PR introduces the `RefreshCatalogTask` component that enables periodic synchronization of Trino's internal catalog state with external sources of truth, eliminating the need for engine restarts when managing catalogs.

The implementation consists of two main components:
       RefreshCatalogTask - A scheduled task that periodically polls external catalog stores and synchronizes changes with the `CoordinatorDynamicCatalogManager`. It follows the same architectural patterns as the existing `CatalogPruneTask`, ensuring consistency with Trino's design principles.

       Configuration Integration - Adds new configuration properties (`catalog.refresh.enabled`, `catalog.refresh.interval`) that enable dynamic catalog management while maintaining backward compatibility (disabled by default).

The RefreshCatalogTask integrates seamlessly with the existing `CoordinatorDynamicCatalogManager` by calling its `refreshCatalogs()` method, which performs thread-safe catalog updates using the established `catalogsUpdateLock`.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/connector/CoordinatorDynamicCatalogManager.java
+++ b/core/trino-main/src/main/java/io/trino/connector/CoordinatorDynamicCatalogManager.java
@@ -34,6 +34,7 @@ import io.trino.spi.connector.ConnectorName;
 import jakarta.annotation.PreDestroy;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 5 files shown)


hashhar:
are you modifying the catalogstore externally instead of via Trino's provided SQL?

CarlosMorales17:
> are you modifying the catalogstore externally instead of via Trino's provided SQL?
Yes. In a highly distributed environment, modifying catalogs internally would require careful coordination across all deployed instances. As our use case is building atop of the SPI,, we gain the ability to centralize these updates,  ensuring consistency across the entire distributed system.

Issue #26286, opened 2025-07-25
Feature: extend S3 security mapping to other services using IAM


trina242:
From documentation:
> Trino supports flexible security mapping for S3, allowing for separate credentials or IAM roles for specific users or S3 locations. The IAM role for a specific query can be selected from a list of allowed roles by providing it as an extra credential.

Part of this functionality, specifically mapping users to IAM roles, could be useful when communicating with other AWS services: Glue, OpenSearch, Iceberg REST Catalog (Glue) etc.

These days it is especially important when using Glue as a metastore in conjunction with Lake Formation. If Trino is used for querying data location registered with LF, we can rely on restricting access to metadata (Glue) rather than restricting access to the data (S3), as Glue will run resolve every request against Lake Formation first. It gives much more flexibility than managing IAM and bucket policies manually. However, it only makes sense if users are mapped to IAM roles, and for now Trino always uses the same role while communicating with Glue.

My proposal is to decouple generic code (not S3-specific) from S3 Security Mapping to a new library. This way integrating Glue and potentially other services with security mapping will be way easier.

PR #26287, opened 2025-07-25
#26286: Decoupled IAM security mapping from s3 filesystem and added Glue support

trina242:
## Description

This PR decouples generic (not S3-related) security mapping code from S3 filesystem into a new library `trino-iam-aws`, as well as adds security mapping support for Glue metastore.

Related issue: #26286

I'm open to any suggestions related to this implementation, especially when it comes to config parameters - I've been a bit uncomfortable with duplicating so much, but it makes more sense to me than mixing parameters from base class and subclass (like `iam.security-mapping.config-file` and `s3.security-mapping.kms-key-id-credential-name`).

This is my first contribution to Trino, while Java definitely not being my favourite language, so please get easy on me! 😉 

## Release notes


Praveen2112:
But this would create new HTTP Client for each user right ? Won't it throttle the number of threads when multiple users are running the queries ?

trina242:
Hi @Praveen2112, thanks for the feedback! I updated the code and made Glue clients (also STS) share http client. I can't reopen the PR as the branch has been rebased.
New PR: https://github.com/trinodb/trino/pull/26768

PR #26300, opened 2025-07-29
Support creating a branch from other branches

ebyhr:
## Description

Add `FROM other-branch` option to allow creating a new branch from other branches. 
Modifying SPI directly because the method hasn't been released. 

https://docs.google.com/document/d/1jEF4IkWu-2Gzk5ii2Nb0exuEnAUeo98UbiM3i0xtgWQ/edit?tab=t.0#heading=h.dglxb51zw9m2

## Release notes

```markdown
## General
* Add support for creating a branch from specified branches. ({issue}`26300`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
@@ -2741,7 +2741,7 @@ public final class MetadataManager
         CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalogHandle);
         ConnectorMetadata metadata = catalogMetadata.getMetadata(session);
 
-        metadata.createBranch(session.toConnectorSession(catalogHandle), tableHandle.connectorHandle(), branch, fromBranch, properties);
+        metadata.createBranch(session.toConnectorSession(catalogHandle), tableHandle.connectorHandle(), branch, fromBranch, saveMode, properties);
     }
 
     @Override
--- a/core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
+++ b/core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
@@ -1020,11 +1020,11 @@ public class TracingConnectorMetadata
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -120,8 +120,9 @@ statement
     | CALL qualifiedName '(' (callArgument (',' callArgument)*)? ')'   #call
     | CREATE (OR REPLACE)? functionSpecification                       #createFunction
     | DROP FUNCTION (IF EXISTS)? functionDeclaration                   #dropFunction
-    | CREATE (OR REPLACE)? BRANCH (IF NOT EXISTS)? identifier
-        (WITH properties)? IN TABLE qualifiedName                      #createBranch
+    | CREATE (OR REPLACE)? BRANCH (IF NOT EXISTS)? branch=identifier
+        (WITH properties)? IN TABLE qualifiedName
+        (FROM from=identifier)?                                        #createBranch
```
chenjian2664, line 125: What does it actually mean when you replace a branch, but with a `from` branch specified, does the `from` will be ignored?

kasiafi: It is up to the connector, but I expect that `from` should be respected when replacing an existing branch.

kasiafi (approved):
The `from` branch name handling is yet another place where identifier semantics matters. We should try to keep it coherent.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
Praveen2112, line 125: Are we inheriting the properties from the source branch as well ?

Praveen2112 (approved):

ebyhr:
Added another commit to expose `SaveMode` to SPI.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
ebyhr: Not for now, at least. We could add `{ INCLUDING | EXCLUDING } PROPERTIES` option later if needed.

Issue #15911, opened 2023-01-31
Support AWS DocumentDB


iceychris:
First off, thank you all for making Trino. It's a very nice tool!

It seems like right now, there is no support for [AWS DocumentDB](https://aws.amazon.com/documentdb/).
The main outstanding issue for me is that [collation is not supported with DocumentDB](https://community.gravitee.io/t/using-aws-documentdb-option-collation-not-supported-message-in-the-tickets-section/212).

This hunch is confirmed in my initial testing of trying to connect the latest Trino release (406) to DocumentDB yielding this error message when trying to query a collection with something like `select * from mongodb."db"."collection";`:

```
Query 20230131_031421_00000_xxzuh failed: Query failed with error code 303 and error message 'option collation is currently not supported.' on server redacted-read-replica-1.redacted.aws-region.docdb.amazonaws.com:27017
```

[This recent PR](https://github.com/trinodb/trino/pull/15000) seems to introduce collation and I think adding a simple on/off configuration switch for toggling collation might suffice to make this work.

jeff-xu-z:
Just for the benefit for the community, we (Workday) were able to use a custom-built JAR to make Trino 435 working with AWS DocumentDB.

- We used code 435, minus change of the PR (https://github.com/trinodb/trino/pull/15000), to build Trino.
- We installed trino 435 from Trino RPM.
- We used the custom built trino-mongodb-435.jar to replace the installed JAR.

The connector is configured as below. And query runs well through Trino.

`connector.name=mongodb
mongodb.connection-url=mongodb://<user>:<password>@<docdb-hostname>:27017/?replicaSet=rs0
mongodb.case-insensitive-name-matching=true
mongodb.tls.enabled=true
mongodb.tls.truststore-path=<path of global-bundle.pem>
`

sherifkayad:
Would really be great to see the option to support that directly in Trino .. without needing to have a custom build ..

sherifkayad:
I submitted a PR to address the collation issue with AWS DocumentDB via a config flag .. hopefully I am on the right track here ..

ErikAhlund:
Any news on this? How did you make it work @jeff-xu-z ? Any chance you could explain how you built and replaced the .jar file?

jeff-xu-z:
> Any news on this? How did you make it work [@jeff-xu-z](https://github.com/jeff-xu-z) ? Any chance you could explain how you built and replaced the .jar file?

- Follow trino instruction to build.
[...]

PR #26304, opened 2025-07-30
MongoDB: Introduce a config flag to determine whether to skip using collation or not when querying

sherifkayad:
MongoDB: Introduce a config flag to determine whether to skip using collation or not when querying

## Description
MongoDB: Introduce a config flag to determine whether to skip using collation or not when querying. This is mainly to address the issue reported by #15911 

This PR introduces a config flag `mongodb.skip-collation` (defaults to `false`) and based on it, the `MongoSession` decides whether to use a collation or not.

#15911 

## Release notes

(x) Release notes are required, with the following suggested text: `MongoDB: Introduce a config flag to determine whether to skip using collation or not when querying`


ebyhr (commented):
Can we use a localstatck docker container for testing this change?

sherifkayad:
> Can we use a localstatck docker container for testing this change?

@ebyhr I think probably one could do that. Or maybe if you give me an easy way to build a trino docker image locally for the MR I can test against one of my DocumentDB instances in AWS

sherifkayad:
@ebyhr I signed and sent out the CLA yesterday .. however, it seems like the document is still not on file .. any clue how fast that process could be? .. also do you think we can easily build a Trino Docker Image with that change such that I can test it locally?

chenjian2664:
@sherifkayad CLA usually take few days, no worries

ebyhr:
> do you think we can easily build a Trino Docker Image with that change such that I can test it locally?

Please try adding dockerized tests first. We can't ensure future changes if don't add automated tests.

sherifkayad:
> > do you think we can easily build a Trino Docker Image with that change such that I can test it locally?
> 
> Please try adding dockerized tests first. We can't ensure future changes if don't add automated tests.

@ebyhr I believe that DocumentDB is a premium feature of localstack => https://docs.localstack.cloud/aws/services/docdb/ .. I am not sure that a test with localstack would be possible.

ebyhr:
Thanks for the confirmation. Manual testing is fine, then.

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoSession.java
```diff
@@ -528,7 +530,12 @@ public MongoCursor<Document> execute(MongoTableHandle tableHandle, List<MongoCol
 
         MongoCollection<Document> collection = getCollection(tableHandle.remoteTableName());
         Document filter = buildFilter(tableHandle);
-        FindIterable<Document> iterable = collection.find(filter).projection(projection).collation(SIMPLE_COLLATION);
+        FindIterable<Document> iterable = collection.find(filter).projection(projection);
+
+        if (!skipCollation) {
```
ebyhr, line 540: MongoDB doesn't expose whether collation is supported or not, right? I'm asking because we'd like to avoid using configuration properties as much as possible. Ideally, this skip should be determined automatically rather than relying on a flag.

sherifkayad: @ebyhr I tried searching quickly for that and it doesn't seem to be the case that collation is a capability you can simply query for its presence or not .. as far as I have read the only way would be to attempt a try / catch query with collation to see if it's supported or not .. which in my view is not really an elegant solution

sherifkayad: @ebyhr may I ask how we can proceed with that PR here?

sherifkayad:
@ebyhr until now seems like the CLA is not on record .. and again I am not sure how we can proceed here. Can you please tell me what you think?

sherifkayad:
Is there anything we can do to get this one further?

martint:
@cla-bot check

docs/src/main/sphinx/connector/mongodb.md
```diff
@@ -96,6 +97,12 @@ Match database and collection names case insensitively.
 
 This property is optional; the default is `false`.
 
+### `mongodb.skip-collation`
```
chenjian2664, line 100: Is there a way to see what collation the collection uses?

ebyhr: The answer was no according to https://github.com/trinodb/trino/pull/26304#discussion_r2255291734

chenjian2664: I asked ai:

```
How to Find the Collation
Replace yourCollectionName in the command above with the actual name of your collection.

Execute the command in the MongoDB Shell.

The output will be an array containing the collection's information, including its creation options. The collation details will be nested within the options field.

Example Output Structure
For a collection named products:

JavaScript

[
  {
    "name" : "products",
    "type" : "collection",
    "options" : {
      "collation" : {
        "locale" : "en_US", // The main collation setting
        "caseLevel" : false,
        "caseFirst" : "off",
        "strength" : 2,
        // ... other collation parameters
      }
    },
    // ... other collection info
  }
]
```

ebyhr: We can't rely on `options` because the `locale` may be empty even when the MongoDB instance supports collation.

What we actually want to check is whether the MongoDB **_server_** supports collation, which is different from checking whether a specific collection has collation.

PR #26310, opened 2025-07-31
Fix thread safety in Iceberg file deletion logic

chenjian2664:
Replaces `ArrayList` with a thread-safe `BlockingQueue` for collecting file paths to delete during snapshot expiration. This addresses concurrent access issues since file deletions may be executed from multiple threads by default in Iceberg.

## Release notes

```markdown
## Iceberg
* Fix thread safety in Iceberg file deletion logic . ({issue}`issuenumber`)
```


chenjian2664:
ah, seems my master branch is out of date :)

PR #26326, opened 2025-08-01
Improve parallelism in remove_orphan_files procedure

raunaqmorarka:
## Description
Improve parallelism in remove_orphan_files procedure

## Release notes

```markdown
## Iceberg
* Improve performance of remove_orphan_files procedure. ({issue}`26326`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2285,7 +2285,7 @@ public class IcebergMetadata
 
                 validMetadataFileNames.add(fileName(manifest.path()));
                 try (ManifestReader<? extends ContentFile<?>> manifestReader = readerForManifest(table, manifest)) {
-                    for (ContentFile<?> contentFile : manifestReader) {
+                    for (ContentFile<?> contentFile : manifestReader.select(ImmutableList.of("file_path"))) {
                         validDataFileNames.add(fileName(contentFile.location()));
                     }
                 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -25,6 +25,7 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
+import io.airlift.concurrent.MoreFutures;
 import io.airlift.json.JsonCodec;
 import io.airlift.log.Logger;
 import io.airlift.slice.Slice;
@@ -217,6 +218,7 @@ import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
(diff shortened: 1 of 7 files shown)


wendigo (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2369,6 +2375,22 @@ private void scanAndDeleteInvalidFiles(Table table, ConnectorSession session, Sc
...
+            throw new TrinoException(ICEBERG_FILESYSTEM_ERROR, "Failed accessing data for table: " + schemaTableName, e);
+        }
+        finally {
+            deleteFutures.forEach(future -> future.cancel(true));
+        }
+    }
+
+    private void deleteFiles(List<Location> files, SchemaTableName schemaTableName, TrinoFileSystem fileSystem)
+    {
+        log.debug("Deleting files while removing orphan files for table %s [%s]", schemaTableName, files);
```
findinpath, line 2391: Do we want to log all the files (in `DEBUG`) ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2285,7 +2285,7 @@ private void removeOrphanFiles(Table table, ConnectorSession session, SchemaTabl
 
                 validMetadataFileNames.add(fileName(manifest.path()));
                 try (ManifestReader<? extends ContentFile<?>> manifestReader = readerForManifest(table, manifest)) {
-                    for (ContentFile<?> contentFile : manifestReader) {
+                    for (ContentFile<?> contentFile : manifestReader.select(ImmutableList.of("file_path"))) {
```
findinpath, line 2288: nit: we could extract `ImmutableList.of("file_path")` to a static field - it doesn't need to be recreated every time.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2347,18 +2352,19 @@ private static ManifestReader<? extends ContentFile<?>> readerForManifest(Table
...
             while (allFiles.hasNext()) {
                 FileEntry entry = allFiles.next();
                 if (entry.lastModified().isBefore(expiration) && !validFiles.contains(entry.location().fileName())) {
                     filesToDelete.add(entry.location());
                     if (filesToDelete.size() >= DELETE_BATCH_SIZE) {
-                        log.debug("Deleting files while removing orphan files for table %s [%s]", schemaTableName, filesToDelete);
-                        fileSystem.deleteFiles(filesToDelete);
-                        filesToDelete.clear();
+                        List<Location> finalFilesToDelete = filesToDelete;
+                        deleteFutures.add(icebergFileDeleteExecutor.submit(() -> deleteFiles(finalFilesToDelete, schemaTableName, fileSystem)));
```
findinpath, line 2366: It is worth seeing whether in production use cases the object store provider will choose to return `429 Too Many Requests`

Is the filesystem layer being able to cope with backpressure?
I'm looking at https://github.com/trinodb/trino/blob/master/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java and i don't see any `FailSafe` logic built-in.

cc @wendigo @ebyhr

findinpath (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: Overall parallelism is bounded by `iceberg.file-delete-threads`.
In general, either the SDK or the filesystem layer is expected to handle backoff and retry on throttling errors.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: JIT can handle micro-optimizations like that

PR #26328, opened 2025-08-01
Resource Group pattern matching on client tags

jinyangli34:
## Description
Add support for resource group selector pattern matching on client tags

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/StaticSelector.java
+++ b/plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/StaticSelector.java
@@ -20,6 +20,7 @@ import io.trino.spi.resourcegroups.ResourceGroupId;
 import io.trino.spi.resourcegroups.SelectionContext;
 import io.trino.spi.resourcegroups.SelectionCriteria;
 
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
@@ -51,7 +52,7 @@ public class StaticSelector
             Optional<Pattern> originalUserRegex,
             Optional<Pattern> authenticatedUserRegex,
             Optional<Pattern> sourceRegex,
-            Optional<List<String>> clientTags,
+            Optional<List<Pattern>> clientTags,
             Optional<SelectorResourceEstimate> selectorResourceEstimate,
             Optional<String> queryType,
             ResourceGroupIdTemplate group)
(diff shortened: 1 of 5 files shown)


plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/StaticSelector.java
```diff
@@ -90,8 +91,12 @@ public StaticSelector(
                     new BasicMatcher(criteria -> queryTypeValue.equalsIgnoreCase(criteria.getQueryType().orElse("")))))
                 .add(selectorResourceEstimate.map(selectorResourceEstimateValue ->
                     new BasicMatcher(criteria -> selectorResourceEstimateValue.match(criteria.getResourceEstimates()))))
-                .add(clientTags.map(clientTagsValue ->
-                    new BasicMatcher(criteria -> criteria.getTags().containsAll(clientTagsValue))))
+                .add(clientTags.map(tags -> {
```
chenjian2664, line 94: Could this be a breaking change if users previously used client tags containing special characters (like *) that are interpreted in regex

plugin/trino-resource-group-managers/src/test/java/io/trino/plugin/resourcegroups/TestStaticSelector.java
```diff
@@ -212,6 +212,46 @@ public void testClientTags()
         assertThat(selector.match(newSelectionCriteria("A.user", "a source b", ImmutableSet.of("tag1", "tag2", "tag3"), EMPTY_RESOURCE_ESTIMATES)).map(SelectionContext::getResourceGroupId)).isEqualTo(Optional.of(resourceGroupId));
     }
 
+    @Test
+    public void testClientTagsRegex()
```
chenjian2664: package private

plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/StaticSelector.java
jinyangli34: Yes. Adding a release note could help? 
I updated the doc in this PR.

docs/src/main/sphinx/admin/resource-groups.md
```diff
@@ -203,8 +203,8 @@ documentation](https://docs.oracle.com/en/java/javase/24/docs/api/java.base/java
   - `ALTER_TABLE_EXECUTE`: Queries that execute table procedures with [ALTER
     TABLE EXECUTE](alter-table-execute).
 
-- `clientTags` (optional): list of tags. To match, every tag in this list must be in the list of
-  client-provided tags associated with the query.
+- `clientTags` (optional): list of Java regexes. To match, every regex in this list must be
+  matched with at least one of the client-provided tags associated with the query.
```
chenjian2664, line 207: > at least one of 

This is a breaking change, since previously we assumed all tags should match. As far as I know, some companies have already adopted this behavior. I would suggest keeping the logic as `match all` to remain consistent with the previous implementation

jinyangli34: my understanding:
With resource-group tags: [x, y], query tags [a, b, c]

previous:
x **in** (a,b,c) and y **in** (a,b,c)

in this pr:
(x **like** a or x **like** b or x **like** c) and (y **like** a or y **like** b or y **like** c)

It's still `match all` but using regex match logic.

chenjian2664: Okay, looking at the code, it still uses a `match all` semantic.

But what if existing tags contain regex characters, for example, consider `clientTags`: `a*b`. Previously this would only match the literal string "a*b", but now it could also match values like "abb" or "aaaabbb".

PR #26329, opened 2025-08-02
Add ST_Multi Geospatial Function and tests

PeterAustinMoore:
## Description
We have a need to natively convert non-multi types to multi-types, in the same way as the PostGIS function [ST_Multi](https://postgis.net/docs/ST_Multi.html). It is possible todo with a UDF, but is quite inefficient.

No related issues as far as I can tell.

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-geospatial/src/main/java/io/trino/plugin/geospatial/GeoFunctions.java
+++ b/plugin/trino-geospatial/src/main/java/io/trino/plugin/geospatial/GeoFunctions.java
@@ -173,6 +173,9 @@ public final class GeoFunctions
     private static final EnumSet<Type> GEOMETRY_TYPES_FOR_SPHERICAL_GEOGRAPHY = EnumSet.of(
             Type.Point, Type.Polyline, Type.Polygon, Type.MultiPoint);
 
+    private static final EnumSet<GeometryType> COLLECTION_TYPES = EnumSet.of(
+            MULTI_POINT, MULTI_LINE_STRING, MULTI_POLYGON);
+
     private static final EnumSet<GeometryType> VALID_TYPES_FOR_ST_POINTS = EnumSet.of(
             LINE_STRING, POLYGON, POINT, MULTI_POINT, MULTI_LINE_STRING, MULTI_POLYGON, GEOMETRY_COLLECTION);
 
@@ -263,6 +266,22 @@ public final class GeoFunctions
             return null;
         }
         return serialize(createFromEsriGeometry(multipoint, null, true));
+    }
+
+    @SqlNullable
+    @Description("Returns a multi geometry formed from input geometry. If the geometry is already a collection, it is returned unchanged.")
+    @ScalarFunction("ST_Multi")
+    @SqlType(StandardTypes.GEOMETRY)
+    public static Slice stMulti(@SqlType(StandardTypes.GEOMETRY) Slice input)
+    {
+        OGCGeometry geometry = deserialize(input);
+        GeometryType type = GeometryType.getForEsriGeometryType(geometry.geometryType());
+        // We can immediately return the shapes if they are already multi types
+        if(COLLECTION_TYPES.contains(type)) {
+            return serialize(geometry);
+        } else {
+            return serialize(geometry.convertToMulti());
+        }
     }
 
     @Description("Returns a Geometry type Polygon object from Well-Known Text representation (WKT)")
--- a/plugin/trino-geospatial/src/test/java/io/trino/plugin/geospatial/TestGeoFunctions.java
+++ b/plugin/trino-geospatial/src/test/java/io/trino/plugin/geospatial/TestGeoFunctions.java
@@ -2005,6 +2005,24 @@ public class TestGeoFunctions
     }
 
     @Test
+    private void testSTMulti()
+    {
+        testMultiConversion("POINT (0 0)", "MULTIPOINT ((0 0))");
+        testMultiConversion("MULTIPOINT ((1 1))", "MULTIPOINT ((1 1))");
+        testMultiConversion("LINESTRING (1 2, 3 4, 5 6, 7 8)", "MULTILINESTRING ((1 2, 3 4, 5 6, 7 8))");
+        testMultiConversion("MULTILINESTRING ((1 2, 3 4, 5 6, 7 8))", "MULTILINESTRING ((1 2, 3 4, 5 6, 7 8))");
+        testMultiConversion("POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))", "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)))");
+        testMultiConversion("MULTIPOLYGON (((1 1, 1 4, 4 4, 4 1)), ((1 1, 1 4, 4 4, 4 1)))", "MULTIPOLYGON (((1 1, 1 4, 4 4, 4 1)), ((1 1, 1 4, 4 4, 4 1)))");
+    }
+
+    public void testMultiConversion(String wkt, String expected) {
+        assertThat(assertions.expression("ST_AsText(ST_Multi(geometry))")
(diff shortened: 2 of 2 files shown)


wendigo:
@PeterAustinMoore Please squash all commits together

plugin/trino-geospatial/src/test/java/io/trino/plugin/geospatial/TestGeoFunctions.java
```diff
@@ -2004,6 +2004,25 @@ private void assertInvalidMultiPoint(String errorMessage, String... pointWkts)
...
+    {
+        testMultiConversion("POINT (0 0)", "MULTIPOINT ((0 0))");
+        testMultiConversion("MULTIPOINT ((1 1))", "MULTIPOINT ((1 1))");
+        testMultiConversion("LINESTRING (1 2, 3 4, 5 6, 7 8)", "MULTILINESTRING ((1 2, 3 4, 5 6, 7 8))");
+        testMultiConversion("MULTILINESTRING ((1 2, 3 4, 5 6, 7 8))", "MULTILINESTRING ((1 2, 3 4, 5 6, 7 8))");
+        testMultiConversion("POLYGON ((0 0, 1 0, 1 1, 0 1, 0 0))", "MULTIPOLYGON (((0 0, 1 0, 1 1, 0 1, 0 0)))");
+        testMultiConversion("MULTIPOLYGON (((1 1, 4 1, 4 4, 1 4, 1 1)), ((1 1, 4 1, 4 4, 1 4, 1 1)))", "MULTIPOLYGON (((1 1, 4 1, 4 4, 1 4, 1 1)), ((1 1, 4 1, 4 4, 1 4, 1 1)))");
+    }
+
+    public void testMultiConversion(String wkt, String expected)
```
wendigo: private void assertSTMulti

plugin/trino-geospatial/src/main/java/io/trino/plugin/geospatial/GeoFunctions.java
```diff
@@ -265,6 +268,21 @@ public static Slice stMultiPoint(@SqlType("array(" + StandardTypes.GEOMETRY + ")
...
+    @SqlNullable
+    @Description("Returns a multi geometry formed from input geometry. If the geometry is already a collection, it is returned unchanged.")
+    @ScalarFunction("ST_Multi")
+    @SqlType(StandardTypes.GEOMETRY)
+    public static Slice stMulti(@SqlType(StandardTypes.GEOMETRY) Slice input)
+    {
+        OGCGeometry geometry = deserialize(input);
+        GeometryType type = GeometryType.getForEsriGeometryType(geometry.geometryType());
+        // We can immediately return the shapes if they are already multi types
+        if (COLLECTION_TYPES.contains(type)) {
```
wendigo: Instead, add function to the `GeometrySerde` class:

```
public static boolean isMultiCollectionType(Slice shape)
{
   requireNonNull(shape, "shape is null");
   return switch (deserializeType(shape)) {
       case GEOMETRY_COLLECTION, MULTI_POINT, MULTI_LINE_STRING, MULTI_POLYGON -> true;
       default -> false;
    };
}
```

wendigo: Then this function becomes:

```
public static Slice stMulti(@SqlType(StandardTypes.GEOMETRY) Slice input)
{
    if (isMultiCollectionType(input)) {
         return input;
    }
    return serialize(deserialize(input).convertToMulti());
}
```

It doesn't require deserialization when input is already ST Multi collection (it only reads a byte to determine a type)

plugin/trino-geospatial/src/main/java/io/trino/plugin/geospatial/GeoFunctions.java
```diff
@@ -265,6 +268,21 @@ public static Slice stMultiPoint(@SqlType("array(" + StandardTypes.GEOMETRY + ")
         return serialize(createFromEsriGeometry(multipoint, null, true));
     }
 
+    @SqlNullable
+    @Description("Returns a multi geometry formed from input geometry. If the geometry is already a collection, it is returned unchanged.")
```
wendigo: Returns a geometry as multi geometry collection. If the geometry is already a collection, it is returned unchanged

wendigo:
Squash commits and change message to `Implement ST_Multi geospatial function`

wendigo:
@martint ptal for function syntax. It's consistent with the postgis one so lgtm on my end

martint:
The Spatial extension to the SQL specification defines an `ST_ToMultiPolygon` function. Let's rename the function to after that and make sure the semantics match as closely as possible, with the caveats that 
* Trino doesn't have an `ST_MultiPolygon` type, but only a Geometry type
* The specification describes a method on the type instead of a function that takes a value of that type. Trino doesn't support methods, so all spatial functions are currently modeled in that manner.

From the SQL specification:
> 68) For the null-call method ST_ToMultiPolygon():
> Case:
>   a) If SELF is of type ST_MultiPolygon, then return SELF.
>   b) If SELF is of type ST_GeomCollection, then return an ST_MultiPolygon value with the spatial
reference system identifier set to SELF.ST_SRID() and containing the elements of SELF.
>   c) If SELF is of type ST_Polygon, then return an ST_MultiPolygon value with the spatial reference
system identifier set to SELF.ST_SRID() and containing one element, SELF cast to an
ST_Polygon value.
>   d) If SELF is not an empty set, then an exception condition is raised: SQL/MM Spatial exception –
not an empty set.
e) Otherwise, return an empty set of type ST_MultiPolygon with the spatial reference system
identifier set to SELF.ST_SRID().

plugin/trino-geospatial/src/main/java/io/trino/plugin/geospatial/GeoFunctions.java
PeterAustinMoore: For the `requireNonNull(shape, "shape is null");` call - does that mean that `null` geometries will error out?

PeterAustinMoore:
> The Spatial extension to the SQL specification defines an `ST_ToMultiPolygon` function. Let's rename the function to after that and make sure the semantics match as closely as possible, with the caveats that
> 
> * Trino doesn't have an `ST_MultiPolygon` type, but only a Geometry type
> * The specification describes a method on the type instead of a function that takes a value of that type. Trino doesn't support methods, so all spatial functions are currently modeled in that manner.
> 
> From the SQL specification:
> 
> > 68. For the null-call method ST_ToMultiPolygon():
> >     Case:
> >     a) If SELF is of type ST_MultiPolygon, then return SELF.
> >     b) If SELF is of type ST_GeomCollection, then return an ST_MultiPolygon value with the spatial
> >     reference system identifier set to SELF.ST_SRID() and containing the elements of SELF.
> >     c) If SELF is of type ST_Polygon, then return an ST_MultiPolygon value with the spatial reference
> >     system identifier set to SELF.ST_SRID() and containing one element, SELF cast to an
> >     ST_Polygon value.
> >     d) If SELF is not an empty set, then an exception condition is raised: SQL/MM Spatial exception –
> >     not an empty set.
> >     e) Otherwise, return an empty set of type ST_MultiPolygon with the spatial reference system
> >     identifier set to SELF.ST_SRID().

Just checking - this will do more than just convert a polygon to multipolygon, do you still want me to change the name of the function? I was modeling this after ST_Multi in postgis, is that not a correct SQL Spec to follow?

martint:
We don't support multiple spatial reference systems, so you can ignore the bits about SRID. Then, the rules simplify to:

* If it's already a multi polygon return it
* If its a geom collection, return a multi polygon with the elements of the collection
* If it's a polygon, return a multi polygon with a single element

I need to look further into d) and e), as I'm not exactly clear on what the spec means by a "set" in this context, but it's acceptable as a start to raise an error in any other case.

PR #7837, opened 2021-05-05
Add support for case sensitive schema names

Praveen2112:

PR #2350, opened 2019-12-26
Add support for case sensitive identifiers

Praveen2112:
Relates to #17.
Supersedes #1302.

Issue #17, opened 2019-01-21
Add support for case sensitive identifiers


martint:
```
<delimited identifier> ::=
  <double quote> <delimited identifier body> <double quote>

<delimited identifier body> ::=  <delimited identifier part>...
<delimited identifier part> ::=
    <nondoublequote character>
  | <doublequote symbol>

<Unicode delimited identifier> ::=
  U <ampersand> <double quote> <Unicode delimiter body> <double quote>
      <Unicode escape specifier>
<Unicode escape specifier> ::=
  [ UESCAPE <quote> <Unicode escape character> <quote> ]
<Unicode delimiter body> ::=
  <Unicode identifier part>...
<Unicode identifier part> ::=
    <delimited identifier part>
  | <Unicode escape value>
```

```
24) For every <identifier body> IB there is exactly one corresponding case-normal form CNF. CNF is an <identifier body> derived from IB as follows:
Let n be the number of characters in IB. For i ranging from 1 (one) to n, the i-th character Mi of IB is transliterated into the corresponding character 
or characters of CNF as follows:
Case:
   a) If Mi is a lower case character or a title case character for which an equivalent upper case sequence U is de ned by Unicode, then let j be th
       e number of characters in U; the next j characters of CNF are U.
   b) Otherwise, the next character of CNF is Mi.
25) The case-normal form of the <identifier body> of a <regular identifier> is used for purposes such as and including determination of identifier 
      equivalence, representation in the Definition and Information Schemas, and representation in diagnostics areas.

...

27) Two <regular identifier>s are equivalent if the case-normal forms of their <identifier body>s, considered as the repetition of a <character string literal> 
that specifies a <character set specification> of SQL_IDENTIFIER and an implementation-defined collation IDC that is sensitive to case, compare equally 
according to the comparison rules in Subclause 8.2, “<comparison predicate>”.

28) A <regular identifier> and a <delimited identifier> are equivalent if the case-normal form of the <identifier body> of the <regular identifier> and the 
<delimited identifier body> of the <delimited identifier> (with all occurrences of <quote> replaced by <quote symbol> and all occurrences of 
<doublequote symbol> replaced by <double quote>), considered as the repetition of a <character string literal> that specifies a <character set specification>
 of SQL_IDENTIFIER and IDC, compare equally according to the comparison rules in Subclause 8.2, “<comparison predicate>”.


[...]

PR #26339, opened 2025-08-04
Trino Case Sensitivity support : Introduce NameCanonicalizer to the metadata

chrlso:
## Description
Related to issue: [#17](https://github.com/trinodb/trino/issues/17)

Supersedes: [pr #7837](https://github.com/trinodb/trino/pull/7837) and [pr #2350](https://github.com/trinodb/trino/pull/2350)

Adding base connector support for Name canonicalization to enable case sensitivity support at the connector level. This enabled connectors to decide what is the conical form for a database/schema or table identifier when loading table metadata. 

Connectors can use this method to determine how to handle delimited and non delimited identifiers within a given query. 

Currently Trino table metadata handles table and schema names in a case insensitive manner. All table names are downcased meaning the engine cannot distinguish `"name"` from `"Name"` 

This change is a prerequisite to enabling support for case sensitive table and schema names within supported connectors. 

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/AddColumnTask.java
+++ b/core/trino-main/src/main/java/io/trino/execution/AddColumnTask.java
@@ -97,7 +97,7 @@ public class AddColumnTask
     {
         Session session = stateMachine.getSession();
         Map<NodeRef<Parameter>, Expression> parameterLookup = bindParameters(statement, parameters);
-        QualifiedObjectName originalTableName = createQualifiedObjectName(session, statement, statement.getName());
+        QualifiedObjectName originalTableName = createQualifiedObjectName(session, statement, statement.getName(), plannerContext.getMetadata());
         RedirectionAwareTableHandle redirectionAwareTableHandle = plannerContext.getMetadata().getRedirectionAwareTableHandle(session, originalTableName);
         if (redirectionAwareTableHandle.tableHandle().isEmpty()) {
             if (!statement.isTableExists()) {
--- a/core/trino-main/src/main/java/io/trino/execution/CallTask.java
+++ b/core/trino-main/src/main/java/io/trino/execution/CallTask.java
@@ -102,7 +102,7 @@ public class CallTask
         }
 
         Session session = stateMachine.getSession();
-        QualifiedObjectName procedureName = createQualifiedObjectName(session, call, call.getName());
+        QualifiedObjectName procedureName = createQualifiedObjectName(session, call, call.getName(), plannerContext.getMetadata());
         CatalogHandle catalogHandle = getRequiredCatalogHandle(plannerContext.getMetadata(), stateMachine.getSession(), call, procedureName.catalogName());
         Procedure procedure = procedureRegistry.resolve(catalogHandle, procedureName.asSchemaTableName());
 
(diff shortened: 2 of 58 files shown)


ebyhr:
cc: @martint @kasiafi @Praveen2112

chenjian2664:
@chrlso Just check, have you sent the signed cla?

chrlso:
> @chrlso Just check, have you sent the signed cla?

Yes, I submitted last week, waiting on that to be approved

ebyhr:
@cla-bot check

chrlso:
@martint @kasiafi @Praveen2112 @pettyjamesm 

Hi all
I was wondering if there is someone who could help take a look at this PR? This is related to the larger issue around case sensitivity in Trino: https://github.com/trinodb/trino/issues/17 

I also have a draft followup to this to implement the case sensitivity within the Iceberg Rest connector which builds off of these changes: https://github.com/chrlso/trino/pull/3 this isn't finalized yet as I wanted to get some eyes on the initial changes

Thanks
Chris

findepi:
This PR introduces `ConnectorMetadata.canonicalize(ConnectorSession session, String identifier, boolean delimited)`.

I know this is as proposed 6 years ago https://github.com/trinodb/trino/wiki/Delimited-Identifiers#proposal but I would want to rethink this API.

Firstly, per SQL spec, the delimited identifiers do not undergo any canonicalization. If we follow the spec in this regard, we may choose to drop `delimited` parameter of the method and simply call it only for undelimited identifiers.

Furthermore,  it seems incorrect for this method to return original udelimited identifier unmodified. The SQL spec mandates that undelimited identifers `abc`, `AbC` and `ABC` are equivalent, so this function should return same string when invoked with any of them, and it would be incorrect to do something else. Given that the set of possible implementation is narrow, and set of actually needed implementations may just two options, I propose we model canonicalization as an enum.

```
canonicalization(ConnectorSession session) -> LOWER_CASE | UPPER_CASE ( | ... )?
```

findepi:
> model canonicalization as an enum.
> 
> ```
> canonicalization(ConnectorSession session) -> LOWER_CASE | UPPER_CASE ( | ... )?
> ```

i discussed this with @martint and decided to keep `ConnectorMetadata.canonicalize(ConnectorSession session, String identifier` for these reasons

- at least in theory, canonicalization rules can be arbitrary, eg title-casing, or no-op canonicalization.
- we can indeed model them as enums, but it doesn't buy us much, wherever we would store such enum, we can store a reference to ConnectorMetadata (explicit or wrapped in something). this is intermediate state used during query planning only, so nothing that needs to be serializable
- we can, and we should, help ConnectorMetadata implementors write correct canonicalization. they should not need to think whether to pass locale to toUpperCase, and which locale this should be _in general_

We did not discuss whether the function should take `boolean delimited` parameter or simply not be called for delimited identifiers. That's to be followed up upon.

prrvchr:
It seems important to use rules that define how an identifier is composed of its different components.
SQL Server does not allow the use of the catalog when creating a view, for example. Only rules can easily overcome such problems. See [issue#17](https://github.com/trinodb/trino/issues/17#issuecomment-3507544015)

findepi:
> We did not discuss whether the function should take `boolean delimited` parameter or simply not be called for delimited identifiers. That's to be followed up upon.

@martint and I synced on that and decided not to add  `boolean delimited` parameter to `canonicalization` function in `ConnectorMetadata`.
the reasoning is

- the delimited identifiers are treated as-is generally. if a connector has some form of name mangling they can (and may) already do  this in `ConnectorMetadata.getTableHandle`
- if explicit need arises to pass delimited names also thru `canonicalization` function, we can add this later

Issue #26301, opened 2025-07-29
Incorrect user groups for impersonation when using set session authorization


andythsu:
Right now there are two ways to impersonate another user. Option 1 is using `--session-user` and option 2 is using `set session authorization`

For option 1:
if we execute a query using `--session-user` flag
```
~  $ trino --user foo --session-user bar
trino> select 1;
 _col0
-------
     1
(1 row)

Query 20250729_180321_00002_xi3m9, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
0.19 [0 rows, 0B] [0 rows/s, 0B/s]
```
The JSON query detail is 
```
    "user": "bar",
    "originalUser": "bar",
    "setOriginalRoles": [
      
    ],
    "groups": [
      "bar-group"
    ],
    "originalUserGroups": [
      "bar-group"
    ],
    "principal": "bar",
```


For option 2:
if we log in as "foo" and set session authorization as "bar"
```
~  $ trino --user foo
trino> set session authorization "bar";
SET SESSION AUTHORIZATION
trino> select 1;
 _col0
-------
     1
(1 row)

Query 20250729_175156_00001_rjyxe, FINISHED, 1 node
Splits: 1 total, 1 done (100.00%)
0.61 [0 rows, 0B] [0 rows/s, 0B/s]
```

The JSON query detail is 
```
    "user": "bar",
    "originalUser": "foo",
    "setOriginalRoles": [
      
    ],
    "groups": [
      "foo-group"
    ],
    "originalUserGroups": [
      "foo-group"
    ],
    "principal": "foo",
    "enabledRoles": [
      
    ],
    "source": "trino-cli",
```
This output shows the `originalUser` as `foo` and `user` as `bar`. However, the `groups` is `foo-group`, which should have been "bar-group" instead. This also causes the person being impersonated having the same groups as the `originalUser`, and hence returning the wrong results when executing queries.

**Expected Behavior**
Since these look like equivalent ways of impersonating another person, I'm expecting the output should at least be the same. And, I'm expecting the groups that the person currently being impersonated belongs to shows up (should show `bar-group` in this case)

My trino client is (if it matters)
```
~  $ trino --version                                                                                                                                                                                                                                    
Trino CLI 471
```

Chaho12:
I think both is wrong as 

first case the original part is wrong
- user as bar and `originalUser as foo`
- groups as bar and `originalUserGroups as foo-group`

second case only group part is wrong.
[...]

PR #26344, opened 2025-08-05
Replace originalIdentity's groups with identity's groups

andythsu:
## Description

Use identity's groups instead of OriginalIdentity's groups in session. Issue is described [here](https://github.com/trinodb/trino/issues/26301)

This PR specifically resolves Option 2 in the issue

## Release notes

```markdown
## Section
* Fix access control check when access is granted through groups in `SET SESSION AUTHORIZATION`  ({issue}`26344`)
```

### DIFF
--- a/testing/trino-tests/src/test/java/io/trino/security/TestSessionImpersonation.java
+++ b/testing/trino-tests/src/test/java/io/trino/security/TestSessionImpersonation.java
@@ -1,0 +1,157 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.security;
+
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import io.airlift.log.Logging;
+import io.trino.Session;
+import io.trino.jdbc.TrinoConnection;
+import io.trino.plugin.base.security.AllowAllSystemAccessControl;
+import io.trino.plugin.memory.MemoryPlugin;
+import io.trino.server.BasicQueryInfo;
+import io.trino.server.testing.TestingTrinoServer;
+import io.trino.spi.QueryId;
+import io.trino.spi.security.Identity;
+import io.trino.testing.TestingGroupProvider;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.Timeout;
+import org.junit.jupiter.api.parallel.Execution;
+
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.Set;
+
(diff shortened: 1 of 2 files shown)


kokosing:
@homar can you please provide the review here?

kokosing:
@andythsu can you please provide unit tests?

homar:
> @homar can you please provide the review here?

Yep that was clearly a mistake from my side. If you check [SessionRepresentation](https://github.com/trinodb/trino/blob/0a39998b466c5ffd593b60344b6e23105323e1c2/core/trino-main/src/main/java/io/trino/SessionRepresentation.java#L86) it is clear that we need groups from identity here not originalIdentity.
However adding a test would be very beneficial

kokosing:
I believe such test could be added to `TestImpersonation`.

andythsu:
There's also https://github.com/trinodb/trino/blob/master/testing/trino-tests/src/test/java/io/trino/execution/TestUserImpersonationAccessControl.java that although the name suggests "test impersonation", there are no tests in this file that actually touches on the impersonation part. I believe we can merge it with https://github.com/trinodb/trino/blob/master/testing/trino-tests/src/test/java/io/trino/security/TestAccessControl.java. Thoughts? (I can do it in a separate PR)

andythsu:
@kokosing I added a unit test

testing/trino-tests/src/test/java/io/trino/security/TestImpersonation.java
```diff
@@ -138,6 +142,31 @@ public void testImpersonationDisallowedWhenRoleIsNone()
         }
     }
 
+    @Test
+    @Timeout(10)
+    public void testSessionRepresentationReturnsCorrectGroupsDuringImpersonation()
+    {
+        // given
```
kokosing: please remove these comments.

testing/trino-tests/src/test/java/io/trino/security/TestImpersonation.java
```diff
@@ -138,6 +142,31 @@ public void testImpersonationDisallowedWhenRoleIsNone()
...
+        Identity john = Identity.forUser("john").withGroups(johnGroups).build();
+
+        // when
+        Session aliceImpersonationSession = testSessionBuilder()
+                .setOriginalIdentity(alice)
+                .setIdentity(john)
+                .build();
+
+        // then
+        Set<String> originalUserGroups = aliceImpersonationSession.toSessionRepresentation()
```
kokosing: Please instead run the SQL to check the. See https://trino.io/docs/current/functions/session.html and `current_groups()`.

Also please make sure that this test is failing when you revert the change to make sure this test actually provides a true coverage.

andythsu: The `current_groups()` actually shows the right groups. This bug only happens in
- `show schemas from <catalog>`
- `show tables from <catalog>.<schema>`

Other operations will work fine.

It looks like other operations don't call `toSessionRepresentation` but only these two operations will call this method, and hence this error.

kokosing: Ok, please use `current_group()` and also test:
```
show schemas from <catalog>
show tables from <catalog>.<schema>
```

andythsu:
@kokosing I moved my tests to a new file because it requires quite different setup from https://github.com/trinodb/trino/blob/master/testing/trino-tests/src/test/java/io/trino/security/TestImpersonation.java

In addition, I ran these tests without my change and it returned 

image

which is expected

testing/trino-tests/src/test/java/io/trino/security/TestSessionImpersonation.java
```diff
@@ -0,0 +1,144 @@
+package io.trino.security;
```
kokosing, line 14: add copyright

andythsu:
@kokosing everything passed!

kokosing:
I improved release notes a bit. Thank you!

PR #26353, opened 2025-08-06
Hive: Added  support for Twitter Elephant Bird protobufs in sequence files

### LINKED ISSUE #26305: Add support for BytesWritable as values in Hadoop Sequence files
In the Hive connector, sequence files are expected to be of value type `org.apache.hadoop.io.Text`. Adding support for `org.apache.hadoop.io.BytesWritable` would open new possibilities for supporting serialized data structures.

gertjanal:
## Description
A while ago, Twitter introduced a table format to Hive that is based on protocol buffers in a Hadoop sequence file: https://github.com/twitter/elephant-bird/blob/master/hive/src/main/java/com/twitter/elephantbird/hive/serde/ProtobufDeserializer.java
This PR adds that functionality to Trino, so Hive tables created in that format can be queried.

The PR consists of two parts:

* Extended the sequence file reader and writer to support `org.apache.hadoop.io.BytesWritable` instead of only `org.apache.hadoop.io.Text` as values #26305
  See diff https://github.com/trinodb/trino/pull/26353/commits/dcf5e42597fa3bf8858fc1f84968d578c25c2deb
* Added a deserializer for protobufs, using a directory that contains protobuf descriptors, a binary collection of proto files.

More information about the original deserializer can be found here: https://github.com/twitter/elephant-bird/wiki/How-to-use-Elephant-Bird-with-Hive

Tables that can be read have the following scheme:
```
create table users
  row format serde "com.twitter.elephantbird.hive.serde.ProtobufDeserializer"
  with serdeproperties (
    "serialization.class"="com.example.proto.gen.Storage$User")
  stored as
    inputformat "com.twitter.elephantbird.mapred.input.DeprecatedRawMultiInputFormat";
```

Closes #26305

## Release notes

( x ) Release notes are required, with the following suggested text:

```markdown
## Hive connector
* Added support for Twitter Elephantbird protobuf deserialization. ({issue}`26305`)
```

### DIFF
--- a/lib/trino-hive-formats/pom.xml
+++ b/lib/trino-hive-formats/pom.xml
@@ -45,6 +45,11 @@
         </dependency>
 
         <dependency>
+            <groupId>com.google.protobuf</groupId>
+            <artifactId>protobuf-java</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>commons-codec</groupId>
             <artifactId>commons-codec</artifactId>
         </dependency>
@@ -53,6 +58,11 @@
             <groupId>io.airlift</groupId>
             <artifactId>aircompressor-v3</artifactId>
             <optional>true</optional>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>concurrent</artifactId>
         </dependency>
 
         <dependency>
@@ -68,6 +78,11 @@
         <dependency>
             <groupId>io.airlift</groupId>
             <artifactId>units</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-cache</artifactId>
         </dependency>
 
         <dependency>
@@ -221,5 +236,77 @@
             <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
+
+    <build>
+        <pluginManagement>
+            <plugins>
+                <plugin>
+                    <groupId>com.github.os72</groupId>
+                    <artifactId>protoc-jar-maven-plugin</artifactId>
+                    <version>3.11.4</version>
+                    <configuration>
+                        <protocArtifact>com.google.protobuf:protoc:${dep.protobuf.version}</protocArtifact>
+                        <protocVersion>${dep.protobuf.version}</protocVersion>
+                        <addSources>none</addSources>
+                        <inputDirectories>
+                            <include>src/test/resources/protobuf/sources</include>
+                        </inputDirectories>
+                        <outputDirectory>target/generated-test-sources/</outputDirectory>
+                        <outputTargets>
+                            <outputTarget>
+                                <type>java</type>
+                                <addSources>none</addSources>
+                                <outputDirectory>target/generated-test-sources/</outputDirectory>
+                            </outputTarget>
+                            <outputTarget>
+                                <type>descriptor</type>
+                                <addSources>none</addSources>
+                                <outputDirectory>target/test-classes/protobuf/descriptors</outputDirectory>
+                            </outputTarget>
+                        </outputTargets>
+                    </configuration>
+                </plugin>
+            </plugins>
+        </pluginManagement>
+        <plugins>
+            <plugin>
+                <groupId>com.github.os72</groupId>
+                <artifactId>protoc-jar-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>generate-test-sources</id>
+                        <goals>
+                            <goal>run</goal>
+                        </goals>
+                        <phase>generate-test-sources</phase>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.codehaus.mojo</groupId>
+                <artifactId>build-helper-maven-plugin</artifactId>
+                <executions>
+                    <execution>
+                        <id>add-test-sources</id>
+                        <goals>
+                            <goal>add-test-source</goal>
+                        </goals>
+                        <phase>generate-test-sources</phase>
+                        <configuration>
+                            <sources>
+                                <source>${basedir}/target/generated-test-sources</source>
+                            </sources>
+                        </configuration>
+                    </execution>
+                </executions>
+            </plugin>
+        </plugins>
+    </build>
 </project>
--- a/lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
+++ b/lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
@@ -1,0 +1,183 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.hive.formats.line.protobuf;
+
+import com.google.common.collect.ImmutableList;
+import com.google.protobuf.ByteString;
+import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.Descriptors.EnumValueDescriptor;
+import com.google.protobuf.Descriptors.FieldDescriptor;
+import com.google.protobuf.DynamicMessage;
+import io.airlift.slice.Slices;
+import io.trino.hive.formats.line.Column;
+import io.trino.hive.formats.line.LineBuffer;
+import io.trino.hive.formats.line.LineDeserializer;
+import io.trino.spi.PageBuilder;
+import io.trino.spi.block.ArrayBlockBuilder;
+import io.trino.spi.block.BlockBuilder;
+import io.trino.spi.block.RowBlockBuilder;
+import io.trino.spi.type.ArrayType;
+import io.trino.spi.type.BigintType;
+import io.trino.spi.type.BooleanType;
+import io.trino.spi.type.DoubleType;
+import io.trino.spi.type.IntegerType;
+import io.trino.spi.type.RealType;
+import io.trino.spi.type.RowType;
+import io.trino.spi.type.Type;
+import io.trino.spi.type.VarbinaryType;
+import io.trino.spi.type.VarcharType;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Collection;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Stream;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static io.airlift.slice.Slices.utf8Slice;
+import static java.lang.Float.floatToRawIntBits;
+import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElseGet;
+import static java.util.function.Function.identity;
+
+/**
+ * Deserializer based on the Twitter Elephantbird
+ * <a href="https://github.com/twitter/elephant-bird/blob/master/hive/src/main/java/com/twitter/elephantbird/hive/serde/ProtobufDeserializer.java">ProtobufDeserializer.java</a>.
+ * To improve speed, it could be rewritten using {@link com.google.protobuf.CodedInputStream} instead of the {@link DynamicMessage}.
+ */
+public class ProtobufDeserializer
+        implements LineDeserializer
+{
+    private final List<Column> columns;
+    private final Descriptor descriptor;
+    private final Map<String, FieldDescriptor> descriptorFields;
+
+    public ProtobufDeserializer(List<Column> columns, Descriptor descriptor)
+    {
+        this.columns = ImmutableList.copyOf(requireNonNull(columns, "columns"));
+        this.descriptor = requireNonNull(descriptor, "descriptor");
+        this.descriptorFields = descriptor.getFields().stream()
+                .collect(toImmutableMap(FieldDescriptor::getName, identity()));
+    }
+
+    @Override
+    public List<? extends Type> getTypes()
+    {
+        return columns.stream()
+                .map(Column::type)
+                .collect(toImmutableList());
+    }
+
+    @Override
+    public void deserialize(LineBuffer lineBuffer, PageBuilder builder)
+            throws IOException
+    {
+        DynamicMessage message;
+        try (InputStream in = new ByteArrayInputStream(lineBuffer.getBuffer(), 0, lineBuffer.getLength())) {
+            message = DynamicMessage.parseFrom(this.descriptor, in);
+        }
+
+        builder.declarePosition();
+        for (int columnIndex = 0; columnIndex < this.columns.size(); columnIndex++) {
+            Column column = columns.get(columnIndex);
+            BlockBuilder blockBuilder = builder.getBlockBuilder(columnIndex);
+            writeObject(blockBuilder, column.type(), getValue(message, this.descriptorFields, column.name()));
+        }
+    }
+
+    private static Object getValue(DynamicMessage message, Map<String, FieldDescriptor> fieldNameLookup, String fieldName)
+    {
+        FieldDescriptor fieldDescriptor = fieldNameLookup.get(fieldName);
+        if (fieldDescriptor == null) {
+            return null;
+        }
+        return requireNonNullElseGet(message.getField(fieldDescriptor),
+                () -> fieldDescriptor.hasDefaultValue() ? fieldDescriptor.getDefaultValue() : null);
+    }
+
+    private void writeObject(BlockBuilder blockBuilder, Type type, Object value)
+    {
+        if (type instanceof ArrayType t) {
+            ((ArrayBlockBuilder) blockBuilder).buildEntry(elementBuilder -> {
+                // Always create an array, even when the value is null
+                if (value instanceof Collection<?> collection) {
+                    collection.forEach(element -> {
+                        writeObject(elementBuilder, t.getElementType(), element);
+                    });
+                }
+            });
+        }
+        else if (value == null) {
+            blockBuilder.appendNull();
+        }
+        else if (type instanceof BigintType t && value instanceof Long l) {
+            t.writeLong(blockBuilder, l);
+        }
+        else if (type instanceof BooleanType t && value instanceof Boolean b) {
+            t.writeBoolean(blockBuilder, b);
+        }
+        else if (type instanceof DoubleType t && value instanceof Double d) {
+            t.writeDouble(blockBuilder, d);
+        }
+        else if (type instanceof VarcharType && value instanceof EnumValueDescriptor e) {
+            type.writeSlice(blockBuilder, utf8Slice(e.getName()));
+        }
+        else if (type instanceof IntegerType t && value instanceof Integer i) {
+            t.writeInt(blockBuilder, i);
+        }
+        else if (type instanceof RealType t && value instanceof Float f) {
+            t.writeLong(blockBuilder, floatToRawIntBits(f));
+        }
+        else if (type instanceof RowType rowType && value instanceof DynamicMessage rowMessage) {
+            Map<String, FieldDescriptor> fieldNameLookup = collectFieldsToDeserialize(rowMessage);
+            if (fieldNameLookup.isEmpty()) {
+                // The message has no set values nor default values
+                // In the Hive implementation, a struct where all fields are null is returned as null
+                blockBuilder.appendNull();
+            }
+            else {
+                ((RowBlockBuilder) blockBuilder).buildEntry(fieldBuilders -> {
+                    for (int i = 0; i < rowType.getFields().size(); i++) {
+                        RowType.Field rowField = rowType.getFields().get(i);
+                        if (rowField.getName().isPresent()) {
+                            writeObject(fieldBuilders.get(i), rowField.getType(), getValue(rowMessage, fieldNameLookup, rowField.getName().get()));
+                        }
+                        else {
+                            throw new IllegalStateException("Unable to apply value to row field with no name: " + i);
+                        }
+                    }
+                });
+            }
+        }
+        else if (type instanceof VarcharType t && value instanceof String s) {
+            t.writeSlice(blockBuilder, utf8Slice(s));
+        }
+        else if (type instanceof VarbinaryType t && value instanceof ByteString b) {
+            t.writeSlice(blockBuilder, Slices.wrappedBuffer(b.toByteArray()));
+        }
+        else {
+            throw new IllegalStateException("Unimplemented type " + type.getDisplayName());
+        }
+    }
+
+    private static Map<String, FieldDescriptor> collectFieldsToDeserialize(DynamicMessage message)
+    {
+        return Stream.concat(message.getAllFields().keySet().stream(),
+                        message.getDescriptorForType().getFields().stream().filter(FieldDescriptor::hasDefaultValue))
+                .collect(toImmutableMap(FieldDescriptor::getName, identity(), (l, r) -> l));
+    }
+}
--- a/lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
+++ b/lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
@@ -1,0 +1,152 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.hive.formats.line.protobuf;
+
+import com.google.common.base.Splitter;
+import com.google.common.cache.CacheLoader;
+import com.google.common.cache.LoadingCache;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.util.concurrent.ListenableFuture;
+import com.google.common.util.concurrent.ListeningExecutorService;
+import com.google.protobuf.DescriptorProtos.FileDescriptorProto;
+import com.google.protobuf.DescriptorProtos.FileDescriptorSet;
+import com.google.protobuf.Descriptors.Descriptor;
+import com.google.protobuf.Descriptors.DescriptorValidationException;
+import com.google.protobuf.Descriptors.FileDescriptor;
+import io.airlift.units.Duration;
+import io.trino.cache.EvictableCacheBuilder;
+import io.trino.hive.formats.line.Column;
+import io.trino.hive.formats.line.LineDeserializerFactory;
+import io.trino.spi.TrinoException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static com.google.common.util.concurrent.MoreExecutors.listeningDecorator;
+import static io.airlift.concurrent.Threads.daemonThreadsNamed;
+import static io.trino.hive.formats.HiveClassNames.TWITTER_ELEPHANTBIRD_PROTOBUF_SERDE_CLASS;
+import static io.trino.hive.formats.HiveFormatsErrorCode.HIVE_INVALID_METADATA;
+import static io.trino.spi.StandardErrorCode.CONFIGURATION_INVALID;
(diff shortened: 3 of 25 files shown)


lib/trino-hive-formats/src/test/java/io/trino/hive/formats/line/protobuf/examples/DataRecordProtos.java
```diff
@@ -0,0 +1,5986 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+// Generated by the protocol buffer compiler.  DO NOT EDIT!
+// source: datarecord.proto
+
+package io.trino.hive.formats.line.protobuf.examples;
+
+@SuppressWarnings("all")
```
gertjanal: This file was generated by `protoc`. While the `SuppressWarnings` is enough to omit the class in  `mvn checkstyle:check`, the GitHub build fails... What should I do to let the build succeed?

gertjanal: I added the maven plugin to generate the classes during the generate-test-sources phase. That works for now

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
```diff
@@ -0,0 +1,174 @@
...
+                // Always create an array, even when the value is null
+                if (value instanceof Collection<?> collection) {
+                    collection.forEach(element -> {
+                        writeObject(elementBuilder, t.getElementType(), element);
+                    });
+                }
+            });
+        }
+        else if (value == null || (value instanceof DynamicMessage message && message.getAllFields().isEmpty())) {
+            blockBuilder.appendNull();
```
dain: So a present row with where all fields are null is written as null, instead of a struct containing all fields set to null?  That seems like a strange choice.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
```diff
@@ -0,0 +1,174 @@
...
+        }
+        else if (type instanceof BooleanType t && value instanceof Boolean b) {
+            t.writeBoolean(blockBuilder, b);
+        }
+        else if (type instanceof DoubleType t && value instanceof Double d) {
+            t.writeDouble(blockBuilder, d);
+        }
+        else if (value instanceof EnumValueDescriptor e) {
+            type.writeSlice(blockBuilder, utf8Slice(e.getName()));
+        }
```
dain: which Trino types are allowed for an enum value?  I would assume this is limited to VARCHAR and VARBINARY.  You could support CHAR, but you would need to truncate if the name is too long.  You also should truncate if the length exceeds the VARCHAR length.. or you could support only unlimited VARCHAR.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
```diff
@@ -0,0 +1,174 @@
...
+import static java.lang.Float.floatToRawIntBits;
+import static java.util.Objects.requireNonNull;
+import static java.util.Objects.requireNonNullElseGet;
+import static java.util.function.Function.identity;
+
+/**
+ * Deserializer based on the Twitter Elephantbird
+ * <a href="https://github.com/twitter/elephant-bird/blob/master/hive/src/main/java/com/twitter/elephantbird/hive/serde/ProtobufDeserializer.java">ProtobufDeserializer.java</a>
+ */
+public class ProtobufDeserializer
```
dain, line 62: Let's add a developer comment that this should eventually be rewritten using CodedInputStream to make it significantly more efficient... I doubt anyone will do it, but maybe someone will find it an interesting project.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,115 @@
...
+
+    @Override
+    public ProtobufDeserializer create(List<Column> columns, Map<String, String> serdeProperties)
+    {
+        requireNonNull(this.descriptorsDirectory, "No hive.elephantbird.descriptors set in hive configuration");
+
+        String serializationClass = serdeProperties.get("serialization.class");
+        requireNonNull(serializationClass, "serialization.class");
+        checkArgument(serializationClass.contains("$"), "Expected serialization.class to contain {package}${protoname}, but was %s", serializationClass);
+        String[] javaClassNameAndProtoName = serializationClass.split("\\$");
```
dain: Use Guava `Splitter.on('$').limit(2).splitToList(serializationClass)`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,115 @@
...
+        Optional<Descriptor> descriptor = fileDescriptorSets.stream()
+                .map(fileDescriptorSet -> findDescriptor(fileDescriptorSet, javaClassNameAndProtoName[0], javaClassNameAndProtoName[1]))
+                .filter(Objects::nonNull)
+                .findFirst();
+
+        checkArgument(descriptor.isPresent(), "No file descriptor found for serialization class %s in directory %s", serializationClass, descriptorsDirectory);
+        return new ProtobufDeserializer(columns, descriptor.get());
+    }
+
+    Descriptor findDescriptor(FileDescriptorSet fileDescriptorSet, String javaClassName, String protoName)
```
dain: Why not just unroll all of this in the constructor and create a map from class name to the Descriptor?  Then at runtime this is just a lookup.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
```diff
@@ -177,6 +178,8 @@ public class HiveConfig
 
     private int metadataParallelism = 8;
 
+    private Path elephantBirdProtobufDescriptors;
```
dain: I was thinking this would be a storage location, not a location local to the machine.  This would require that the ops team deploy the protobufs on every coordinator and worker.  Also, we should consider periodically refreshing the list so that the messages can be updated at without a cluster restart.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
```diff
@@ -1296,4 +1299,17 @@ public HiveConfig setMetadataParallelism(int metadataParallelism)
...
         return this;
     }
+
+    public Path getElephantBirdProtobufDescriptors()
+    {
+        return elephantBirdProtobufDescriptors;
+    }
+
+    @ConfigDescription("Directory where protobuf descriptors are stored to use for deserializing Twitter Elephant protobufs")
+    @Config("hive.elephantbird.descriptors")
```
dain: I'm not sure using elephantbird here is a great idea.  The format is named sequencefile protobuf, so maybe we just call this `hive.protobuf.descriptors`?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
dain: Alternatively, there could be a table property to specify the exact location of the descriptor file.  This could be done as an override, so if not present we falldback to this directory.

dain (changes requested):
I read throught the protobuf bits and left some comments.

The PR has a bunch of resource group changes, I think you accidently mixed in.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: Only for structs, the value can be a DynamicMessage, which means RowType columns can be null. I noticed this when comparing in the results with Hive.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: In Hive, all types are inferred from the protobuf descriptor when the table is created. For all protobuf enum fields, a string column (or struct field) is created. The name of the enum is used as value, as can be read here: https://github.com/twitter/elephant-bird/blob/3ae48b10bc56b2d66de45739ef7d6aad821c06e0/hive/src/main/java/com/twitter/elephantbird/hive/serde/ProtobufStructObjectInspector.java#L161

I think that is unlimited varchar, although I don't think anyone will create a very large enum value. Is my implementation the variant of the unlimited varchar? Should I change anything?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
gertjanal: Done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
gertjanal: > I was thinking this would be a storage location, not a location local to the machine. This would require that the ops team deploy the protobufs on every coordinator and worker. Also, we should consider periodically refreshing the list so that the messages can be updated at without a cluster restart.

Do you want me to use the `io.trino.filesystem.Location` and `io.trino.filesystem.TrinoFileSystem` classes for this? What kind of configuration would you expect? Is there any example config available that I can use? I was looking for a configuration to enable any storage location, but I can't find it.

I like the idea of the periodic refresh, I'll look into that

gertjanal: > Alternatively, there could be a table property to specify the exact location of the descriptor file. This could be done as an override, so if not present we falldback to this directory.

I have considered that, but it doesn't feel like the right place to put configuration properties that are specific for Trino in the scheme properties in Hive

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: The protobuf in my test is not so complicated, but with protobufs containing a tree of many message types, the list of detected classnames can get very large even though only one is needed for the table (the root message type). Do you know how many times the serializer is created? Is it once {per worker, per table}, or is it created many times per worker? (per task is the right name I guess). When it is created very often, caching might indeed be better...

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: I've looked into that and I hope some day I can be that developer :)

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
gertjanal: > > I was thinking this would be a storage location, not a location local to the machine. This would require that the ops team deploy the protobufs on every coordinator and worker. Also, we should consider periodically refreshing the list so that the messages can be updated at without a cluster restart.
> 
> Do you want me to use the `io.trino.filesystem.Location` and `io.trino.filesystem.TrinoFileSystem` classes for this? What kind of configuration would you expect? Is there any example config available that I can use? I was looking for a configuration to enable any storage location, but I can't find it.
> 
> I like the idea of the periodic refresh, I'll look into that

I have added a scheduled loader for the descriptors :partying_face:

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: Added the comment

lib/trino-hive-formats/pom.xml
```diff
@@ -221,5 +226,77 @@
...
         </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
+
+    <build>
```
dain, line 247: @electrum and @wendigo should review this pom change

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
dain: Add a comment like:

In the Hive implementation, a struct where all fields are null it written as a null

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
dain: I think you should check the Trino type.  Otherwise you can get hard to debug downstream bugs, because we don't validate data when writing.  Also if the type happens to be a BIGINT you'll get an exception here with a bad error message

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
dain: This mesthod gets called for each file reader instance, but the factory is a singleton in Trino.

If there is an issue of this being to burdensom on the servers, I would add a loading cache to this factory.  I would ask for help on the `#core-dev` channel on slack... but I'd sugest the behavior you want is, when a class is asked for, it the class is not present, it loads it from the file system.  I'd also set it up to periodically check if the schema file has changed and refresh (loading caches can do this).

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
```diff
@@ -1296,4 +1301,30 @@ public HiveConfig setMetadataParallelism(int metadataParallelism)
         this.metadataParallelism = metadataParallelism;
         return this;
     }
+
+    public Path getProtobufDescriptors()
+    {
+        return protobufDescriptors;
+    }
+
+    @ConfigDescription("Directory where protobuf descriptors are stored to use for deserializing Twitter Elephant protobufs")
```
dain: Remove `Twitter Elephant`.  We may want to use this for general protobuf deserialization as the descriptor are generic to protobuf.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+            catch (DescriptorValidationException e) {
+                throw new IllegalStateException("Failed to load protobuf fileDescriptor " + proto.getName() + " from " + loader.getDescriptorsDirectory(), e);
+            }
+        }
+        return null;
+    }
+
+    static class FileDescriptorSetsLoaderScheduler
```
dain: I don't think this is thread safe.  I'd stick to a loading cache.

electrum:
Please follow these rules for the commit messages: https://cbea.ms/git-commit/#seven-rules

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: Added

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Converted to loading cache. I used the newly introduced duration on the cache to refresh descriptors. The only downside of this approach is that it is reading all descriptor proto files when refreshing just a single descriptor. I need some time to find a better solution

gertjanal: I added the previously discussed refresh interval to the `cache.refreshAfterWrite()`, so the cache of a descriptor is invalidated after the configured duration. The `refreshAfterWrite()` uses `cache.reload()`. I added a `newFixedThreadPool(1)` to use in this reload, to avoid a concurrent file read when descriptors are invalidated.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveConfig.java
gertjanal: Removed

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Rewritten

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: The hive deserializer converts protobuf enums to strings, so I'm not sure what I should check in Trino here.

gertjanal:
I will squash the commits when the review is complete

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+                    .refreshAfterWrite(updateInterval.toJavaTime())
+                    .maximumSize(maximumSize)
+                    .build(new DescriptorCacheLoader(descriptorsDirectory));
+        }
+    }
+
+    @Override
+    public Set<String> getHiveSerDeClassNames()
+    {
+        return HIVE_SERDE_CLASS_NAMES;
```
electrum: Let's inline this value. It's only used in this one place.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    @Override
+    public ProtobufDeserializer create(List<Column> columns, Map<String, String> serdeProperties)
+    {
+        requireNonNull(this.cache, "No \"hive.protobufs.descriptors\" set in hive configuration");
```
electrum: This should fail with `CONFIGURATION_INVALID` since this is a user visible error

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+
+        String serializationClass = serdeProperties.get("serialization.class");
```
electrum, line 83: What if this property is not set for the table? This should fail with `HIVE_INVALID_METADATA`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+        return new ProtobufDeserializer(columns, cache.getUnchecked(serializationClass));
+    }
+
+    private static class DescriptorCacheLoader
+            extends CacheLoader<String, Descriptor>
+    {
+        private final Path descriptorsDirectory;
+        private final ListeningExecutorService executor = listeningDecorator(newFixedThreadPool(1, new ThreadFactoryBuilder().setNameFormat("protobuf-descriptors-%s").build()));
```
electrum: You can use the `daemonThreadsNamed()` utility method for this

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+
+        DescriptorCacheLoader(Path descriptorsDirectory)
+        {
+            this.descriptorsDirectory = descriptorsDirectory;
```
electrum: Check for null

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        @Override
+        public Descriptor load(String serializationClass)
+                throws IOException
+        {
+            requireNonNull(serializationClass, "serialization.class");
```
electrum: We don't need to check here as the next line will throw. It should be checked before this method as called.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+            checkArgument(serializationClass.matches("^[^$]+\\$[^$]+$"), "Expected serialization.class to contain {package}${protoname}, but was %s", serializationClass);
```
electrum: This should fail with `HIVE_INVALID_METADATA` and should probably be checked before this method is called

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
dain: `type instanceof VarcharType && value instanceof EnumValueDescriptor e`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+
+                                    if (javaClassAndProtoName.getFirst().equals(fileDescriptor.getOptions().getJavaPackage() + "." + fileDescriptor.getOptions().getJavaOuterClassname())) {
+                                        Descriptor descriptor = fileDescriptor.findMessageTypeByName(javaClassAndProtoName.getLast());
+                                        if (descriptor != null) {
+                                            return descriptor;
+                                        }
+                                    }
+                                }
+                                catch (DescriptorValidationException e) {
+                                    throw new IllegalStateException("Failed to load protobuf fileDescriptor " + proto.getName() + " from " + path, e);
```
electrum: Use `TrinoException` with `INVALID_CONFIGURATION`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+            throw new IllegalArgumentException("No file descriptor found for serialization class " + serializationClass);
```
electrum: Use `HIVE_INVALID_METADATA`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        @Override
+        public ListenableFuture<Descriptor> reload(String serializationClass, Descriptor oldValue)
+        {
+            requireNonNull(serializationClass);
+            requireNonNull(oldValue);
+
+            SettableFuture<Descriptor> future = SettableFuture.create();
```
electrum: This shouldn't be needed as `executor` is a `ListeningExecutorService` which returns `ListenableFuture`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
```diff
@@ -0,0 +1,154 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.util.concurrent.MoreExecutors.listeningDecorator;
+import static com.google.protobuf.DescriptorProtos.FileDescriptorSet.parseFrom;
+import static com.google.protobuf.Descriptors.FileDescriptor.buildFrom;
+import static io.trino.cache.EvictableCacheBuilder.newBuilder;
```
dain: This methods names are very generic name and a reader won't really know where they are coming from without an IDE, so don't static import it. Instead use the class+method name, so EvictableCacheBuilder.newBuilder

electrum:
Please drop the third commit when updating and rebasing.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
electrum: This can use `newSingleThreadExecutor()`

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializer.java
gertjanal: Thanks, I misunderstood your comment

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Removed

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Moved

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Found it, thanks

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: ok

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/line/protobuf/ProtobufDeserializerFactory.java
gertjanal: Done

dain (approved):

gertjanal:
@coderabbitai full review

PR #25445, opened 2025-03-28
Add support for lock-free commit in Iceberg HMS catalog

ebyhr:
## Description

Fixes #22182

Request from Iceberg community: https://apache-iceberg.slack.com/archives/C025PH0G1D4/p1743095891453949
Iceberg documentation: https://iceberg.apache.org/docs/nightly/configuration/#hadoop-configuration

## Release notes

```markdown
## Iceberg
* Add support for lock-free commit in Hive catalog. ({issue}`issuenumber`)
```

PR #26358, opened 2025-08-07
Don't Pass expected_parameter_key/value with Hive Locking Enabled

laithalzyoud:
## Description

This PR https://github.com/trinodb/trino/pull/25445 introduced lock-free commits in iceberg HMS catalog, however if locking is used then there is no need to pass `{"expected_parameter_key": "metadata_location",
"expected_parameter_value", metadataLocation}` to the environmentContext. This leads to unnecessary operation on the metastore side to verify that the metadata location is still the same before commiting while Hive locking is already being used to ensure the commit is atomic, example SQL query:
```
UPDATE "TABLE_PARAMS"
SET "PARAM_VALUE" = 'P2' -- The new path
WHERE "TBL_ID" = 123
  AND "PARAM_KEY" = 'metadata_location'
  AND "PARAM_VALUE" = 'P1'; -- The expected old path
``` 
I'm using HMS 3.1.2 and this was a **breaking change** where commits were not succeeding (probably due to a bug in HMS) that caused transactions to become stuck in the DB leading to timeouts and failed commits and eventually the MySQL DB used by the HMS wasn't able to process other transactions timely leading to timeout across all other HMS operations until the underlying DB was restarted to clear the stuck transactions and the Trino version was rolled back. 

## Release notes

```markdown
## Section
```


nineinchnick:
@cla-bot check

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
```diff
@@ -113,10 +113,13 @@ private void commitTableUpdate(Table table, TableMetadata metadata, BiFunction<T
 
             Table updatedTable = tableUpdateFunction.apply(table, newMetadataLocation);
 
+            // it's redundant to pass metadata_location to environmentContext if Hive locking is enabled
+            Map<String, String> environmentContext = lockingEnabled ? ImmutableMap.of() : environmentContext(metadataLocation);
```
nineinchnick, line 117: I don't know anything about this code, but is there a test we can update or add a new one?

ebyhr (dismissed):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
```diff
@@ -113,10 +113,13 @@ private void commitTableUpdate(Table table, TableMetadata metadata, BiFunction<T
 
             Table updatedTable = tableUpdateFunction.apply(table, newMetadataLocation);
 
+            // Passing environment context causes redundant operations if Hive locking is enabled
```
chenjian2664, line 116: Could you give more details about the consequences that what is "redundant operations" in the comments?

laithalzyoud:
Hey @nineinchnick @chenjian2664! I added some explanation in the related issue https://github.com/trinodb/trino/issues/26360, basically this code was originally introduced in https://github.com/trinodb/trino/pull/25445 to allow committing to Iceberg tables without using Hive locks by passing the latest metadata file location to Hive, if this is passed to Hive then it will go through [extra operations](https://github.com/apache/hive/blob/master/standalone-metastore/metastore-server/src/main/java/org/apache/hadoop/hive/metastore/HiveAlterHandler.java#L181-L210) to confirm that the metadata location didn't change since the write/update operation started. In the case when Hive locking is enabled, there is no need to do this extra check on Hive side as Hive locking already guarantees the atomicity of the transaction, therefore it's only needed when Hive locking is disabled.

nineinchnick:
@laithalzyoud the #25445 added some tests. Can they be updated to cover this change? Thanks for adding the explanation in the issue, but it would be ideal to have a test that would also explain it.

laithalzyoud:
Thanks for merging @ebyhr!

PR #26359, opened 2025-08-07
Don't set max_commit_retry Iceberg table property by default

ebyhr:
## Description

Setting the table property to Iceberg's default value is redundant. 

## Release notes


chenjian2664 (approved):

wendigo (approved):

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -998,8 +998,7 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
     for row level deletes.
 * - `max_commit_retry`
   - Number of times to retry a commit before failing. Defaults to the value of 
-    the `iceberg.max-commit-retry` catalog configuration property, which 
-    defaults to `4`.
+    the `iceberg.max-commit-retry` catalog configuration property.
```
raunaqmorarka: Document should still mention the iceberg default

raunaqmorarka (approved):

docs/src/main/sphinx/connector/iceberg.md
ebyhr: I removed it because I was worried about the upstream change. I will revert anyway since the value looks stable.

PR #25015, opened 2025-02-13
Add arrow encoding to spooling

wendigo:
Just a working prototype without all of the types covered

TODO:

- [ ] add support for interval type (encode/decode)
- [ ] add support for date time type (encode/decode)
- [ ] add support for timestamp type (encode/decode)
- [ ] add support for timestamp with tz type (encode/decode)
- [ ] add support for time type (encode/decode)
- [ ] add support for time with tz type (encode/decode)
- [ ] coalesce small pages to bigger arrow batches (configurable?)
- [ ] optimize decoding in the client
- [ ] test every type with NULL, map key, map value, list

Issue #24891, opened 2025-02-03
trino to arrow type conversion for spooling protocol


dysn:
in order to support arrow ipc format for spooling, we need to be able to convert trino internal data format to arrow

arrow types to not directly map to all trino types. in general, we can handle mismatches by some combination of

1. perform a lossy conversion. eg. truncate trino picosecond timestamps to arrow nanosecond timestamps
2. use an arrow extension type. arrow extension types allow a user to define a type and how it should map to underlying arrow types. eg. create a 12 byte integer for picosecond timestamps
3. fail queries that produce types that cant be converted

the following types are known to have issues

1. time and timestamps. trino can represent picosecond precision times, arrow is limited to nanoseconds. 
2. timestamp and time with timezone. trino can represent timestamps with a value specific timezone. arrow (like parquet) can only represent timestamps with a file scoped timezone
3. hyperloglog
4. user defined types

summarizing a slack discussion on this topic

@dain suggested that there may be some dependency on client protocol 
"The difficult with Arrow is everyone tries to mix together “what should we do for Trino (java) clients” with “What should we do for Python”. For python, my opinion is you map everything to standard arrow types.  For everything else, my opinion is you use a custom protocol as you can support all types, and be much more efficient."

@wendigo suggested that we can ignore user defined types for now 

"I think that we can narrow the Arrow-support scope only to the Trino-owned types, without the support for custom types at the moment (I’ve never seen a single external type usage)"

@dain suggested that we could support user defined types similarly to how arrow supports extension types
" think we need to extend types so that they can describe how they are mapped to low level primitive types.  Then if we are in an situation where we don’t understand how to serialize a type (e.g., compatible arrow), the type gives us direction."

wendigo:
I agree that we should start small:

- in first iteration we should support only Trino types supported by Arrow (without adding extension ones)
- in the second iteration we can add extension types for the types outside of the Arrow type system,
- in the third iteration we can add support for custom Trino types (this is rather want than must have)

dysn:
that makes sense to me. 

for the first iteration, do you think we should just return an error when incompatible types are used? 

if you want, i think i can get the first and the second iterations done

Issue #24586, opened 2024-12-26
support apache arrow client


dysn:
Arrow is primarily an in memory columnar data format, but defines ways for applications to exchange such data efficiently. Many common data processing/analysis tools support working directly on arrow data efficiently (eg polars, duckdb)

At our company, we are seeing a pattern emerge where data scientists and analysts will first generate modest amounts of data (100s of mbs to 10s of gbs) with heavy trino queries, and then load that into something like polars or duckdb for further analysis. We would like to better support this pattern. One current pain point is that users must convert the row oriented trino data into arrow data. Arrow provides utilities for this, but that is not so efficient.

There are a few ways trino could return arrow data to users. One would be by implementing the arrow flight (and higher level arrow flight sql) grpc protocol. This seems like it could be a pretty big lift, and likely would require some extra up front design. It may not be possible to support an acceptable subset of trinos features on the existing arrow flight protocol. In particular, fetching data in arrow flight is a grpc endpoint, which doesnt fit well with trino's spooling protocol

An easier way might be to implement a java adbc connector for trino. ADBC is just a java interface and doesnt require any specific protocol. There are two ways we could implement and adbc connector. We could modify the direct protocol to return data in the arrow wire format, or we could enable trino to spool data in the arrow ipc format. Of course, these aren't mutually exclusive.

Is an arrow based trino client desirable? If so, how should that be implemented?

nineinchnick:
Duplicate of #18038, you could comment there if you have a different and/or more specific use case.

dysn:
i think this is not exactly a duplicate, but i can move discussion there if you prefer

this ticket is more generally about having some kind of arrow based client for trino, whereas 18038 is specifically about arrow flight. its also not clear to me if 18038 is about creating an arrow flight connector for trino so it can fetch data from an arrow flight server, or implementing the arrow flight endpoints in trino so it can serve as an arrow flight server

nineinchnick:
Thanks, I see the distinction now. It might be actually a much smaller effort to use Arrow instead of JSON as the transport format, than implementing a client compatible with Arrow Flight SQL.

I think the biggest issue with Arrow is how to handle Trino data types that don't map to Arrow types. This would require special logic on the client side - we should enumerate such types and see if that would not invalidate the whole point of using Arrow, that is, if it would be simpler to convert all data on the client side.

dysn:
[...]

Issue #18038, opened 2023-06-25
support API for Arrow Flight SQL 


erikerlandson:
The desired use case is data scientists or other users working with arrow-based data, such as parquet, pandas, polars, etc, and who would like to get that data into or out of Trino, without having to hack through intermediate steps or intermediate ascii sql row representations.

> Flight SQL means database servers can implement a standard interface that is designed around Apache Arrow and columnar data from the start. Just like how Arrow provides a standard in-memory format, Flight SQL saves developers from having to design and implement an entirely new wire protocol. As mentioned, Flight already implements features like encryption on the wire and authentication of requests, which databases do not need to re-implement.

https://arrow.apache.org/blog/2022/02/16/introducing-arrow-flight-sql/

erikerlandson:
cc @caldeirav @MichaelTiemannOSC

jwilliams-ocient:
Heavy ➕. 

The team at [Ocient](https://ocient.com) is also looking to allow Trino to connect to a FlightSQL data source. The connector implementation seems like it would be pretty straightforward if I'm following https://trino.io/docs/current/develop/example-jdbc.html ([source](https://github.com/trinodb/trino/tree/master/plugin/trino-example-jdbc)) correctly.

OmarSultan85:
> Heavy ➕.
> 
> The team at [Ocient](https://ocient.com) is also looking to allow Trino to connect to a FlightSQL data source. The connector implementation seems like it would be pretty straightforward if I'm following https://trino.io/docs/current/develop/example-jdbc.html ([source](https://github.com/trinodb/trino/tree/master/plugin/trino-example-jdbc)) correctly.

Hey Jwilliams,

Just wondering how far long are you in this track? We are looking to implement a custom connector to allow for querying data from a FlightSQL Data Source. But the documentation in the example jdbc driver doesn't seem to provide enough information on how to actually  query data and run the queries.

Were you able to reach anything?

YuriyGavrilov:
+1

YuriyGavrilov:
just to leave this here: https://www.vldb.org/pvldb/vol10/p1022-muehleisen.pdf

YuriyGavrilov:
some good example of Flight SQL server:
- https://github.com/voltrondata/flight-sql-server-example 
- https://www.kamu.dev/blog/2023-09-datafusion-flightsql/ https://github.com/kamu-data/kamu-cli

georgewfisher:
+1

[...]

PR #26365, opened 2025-08-07
[WIP] Add Arrow format support for spooling protocol

jonasbrami:
This PR continues work from [PR #25015](https://github.com/trinodb/trino/pull/25015) by @wendigo and @dysn, implementing Arrow format support for Trino's spooling protocol. Since the original branch (`serafin/arrow-encoding`) was removed after the PR was closed, this implementation was recreated by manually applying original patch on current trino master.
Related issues are: 
- https://github.com/trinodb/trino/issues/24891
- https://github.com/trinodb/trino/issues/18038
- https://github.com/trinodb/trino/issues/24586

The goal of this PR is to increase the retrieval speed of large query output results by supporting arrow format for the spooling segment. 
Currently JSON is being used to transmit data over the wire, it is very expensive in time and space to serialize and deserialize JSON, especially when the deserialization occurs in a "slow" language like Python. Additionally, Arrow's columnar format enables zero copy, built-in parallelism and vectorized operations across columns, while JSON's row-based text format requires sequential parsing that can be very challenging to parallelize.

This PR comes together with another [( PR in aiotrino)](https://github.com/mvanderlee/aiotrino/pull/7), the Asyncio Python Trino client. 

When benchmarking JSON vs ARROW, we observed end-to-end:

- 50x speed up of our Arrow-compatible Aiotrino versus Trino Java client
- 400x speed up of our Arrow-compatible Aiotrino version versus pure Python clients

image

More details regarding the benchmark available [here](https://github.com/jonasbrami/aiotrino)

This PR demonstrates a working proof of concept for Arrow serialization in Trino's spooling segments. While functional for evaluation and testing, additional work is needed before production readiness.

**Implementation Details:**

**Type Mapping:**

The following Trino types are supported with their Arrow equivalents:

| Trino Type | Arrow Type | Notes |
|------------|------------|-------|
| `BOOLEAN` | `Bool` | |
| `TINYINT` | `Int(8, signed)` | |
| `SMALLINT` | `Int(16, signed)` | |
| `INTEGER` | `Int(32, signed)` | |
| `BIGINT` | `Int(64, signed)` | |
| `REAL` | `FloatingPoint(SINGLE)` | |
| `DOUBLE` | `FloatingPoint(DOUBLE)` | |
| `VARCHAR` | `Utf8` | |
| `CHAR` | `Utf8` | Fixed-length semantics lost, converted to variable-length |
| `VARBINARY` | `Binary` | |
| `DATE` | `Date(DAY)` | |
| `TIME(p)` | `Time(unit, bitWidth)` | **PRECISION LOSS**: p=12 (picosecond) → nanosecond precision |
| `TIME(p) WITH TIME ZONE` | `Time(unit, bitWidth)` | **TIMEZONE LOSS**: Original timezone info lost, normalized to UTC offset |
| `TIMESTAMP(p)` | `Timestamp(unit, timezone=null)` | **PRECISION LOSS**: p=12 (picosecond) → nanosecond precision |
| `TIMESTAMP(p) WITH TIME ZONE` | `Timestamp(unit, timezone='UTC')` | **TIMEZONE LOSS**: Original timezone lost, all values normalized to UTC |
| `DECIMAL(p,s)` | `Decimal(precision, scale, 128)` | 128-bit decimal |
| `UUID` | `FixedSizeBinary(16)` | With Arrow extension metadata `arrow.uuid` |
| `ARRAY<T>` | `List` | With nested element type |
| `MAP<K,V>` | `Map(false)` | With struct entries containing key/value fields |
| `ROW(...)` | `Struct` | With nested field types |
| `INTERVAL DAY TO SECOND` | `Interval(DAY_TIME)` | |
| `INTERVAL YEAR TO MONTH` | `Interval(YEAR_MONTH)` | |
| `JSON` | `Utf8` |  |
| `UNKNOWN` | `Null` | |

- **Testing**: Arrow spooling functionality is verified by `TestArrowSpooledDistributedQueries` (extends `AbstractSpooledQueryDataDistributedQueries`, overrides timezone-related tests to expect UTC normalization) and `TestArrowEncodingUtils` for round-trip conversion validation between Trino pages and Arrow vectors.
- The rationale is to stay as close to native Arrow as possible, prioritizing simplicity, Arrow client compatibility, and deserialization performance over perfect type fidelity. For example, we could have preserved Timestamp with Timezone by using an Arrow struct containing both timestamp and timezone fields, but we chose to normalize to UTC instead to minimize the deserialization costs. This approach allows clients to use any Arrow-compatible library of their choice to directly consume the results without custom handling.

**Progress from original PR #25015 TODO list:**
- [x] add support for interval type (encode/decode) ✅ **Added**: `IntervalDayWriter.java`, `IntervalYearMonthWriter.java`
- [x] add support for date time type (encode/decode) ✅ **Added**: `DateWriter.java` 
- [x] add support for timestamp type (encode/decode) ✅ **Improved**: Basic support existed, extended with full precision variants (`TimestampSecWriter.java`, `TimestampMilliWriter.java`, `TimestampMicroWriter.java`, improved `TimestampNanoWriter.java`)
- [x] add support for timestamp with tz type (encode/decode) ✅ **Added**: `TimestampWithTimeZone*Writer.java` (all precision variants)
- [x] add support for time type (encode/decode) ✅ **Added**: `TimeSecWriter.java`, `TimeMilliWriter.java`, `TimeMicroWriter.java`, `TimeNanoWriter.java`
- [x] add support for time with tz type (encode/decode) ✅ **Added**: `TimeWithTimeZone*Writer.java` (all precision variants)
- [ ] coalesce small pages to bigger arrow batches (configurable?) ⏳ **Remaining work**
- [x] optimize decoding in the client ✅ **Done in aiotrino fork** (Python client optimization)
- [x] test every type with NULL, map key, map value, list ✅ **Added but to be improved**: `TestArrowSpooledDistributedQueries.java`, `TestArrowEncodingUtils.java`

**Additional implementations since PR #25015:**
- [x] Nested Complex types: Array (`ArrayWriter.java`), Map (`MapWriter.java`), Row (`RowWriter.java`)
- [x] Special types: UUID (`UuidWriter.java`), Decimal (`DecimalWriter.java`), JSON, Unknown/Null
- [x] Client-side decoding optimization: Implemented in aiotrino Python fork
- [x] Configuration system: Arrow-specific memory allocation and concurrency controls

**Outstanding work for production readiness:**
- Java implementation refinements and code review feedback
- Commit structure cleanup and proper messaging
- Additional test coverage and edge case handling
- Detailed performance analysis to isolate the sources of speedup in Arrow-enabled Aiotrino:
    - Arrow vs JSON deserialization overhead
    - Columnar parallelization benefits
    - Asynchronous segment retrieval impact

**Additional Configuration**: Arrow spooling requires specific configuration properties:
- `protocol.spooling.encoding.arrow.enabled=true` and `protocol.spooling.encoding.arrow+zstd.enabled=true` to enable Arrow encoding
- `protocol.spooling.arrow.max-concurrent-serialization=5` (default) to control parallelism 
- `protocol.spooling.arrow.max-allocation=200MB` (default) for memory management
- JVM flags: `--enable-native-access=ALL-UNNAMED --add-opens=java.base/java.nio=ALL-UNNAMED`

The new `protocol.spooling.arrow.max-concurrent-serialization` and `protocol.spooling.arrow.max-allocation` parameters are necessary to control the amount of off-heap/native memory used during serialization by Trino. Arrow Java uses `Unsafe` off-heap memory allocation that is not managed by the JVM. 

## Release notes

```markdown
## Section
```


aschreiber1:
Just saying this would be amazing if this got finished! It would make a HUGE difference for performance

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
```diff
@@ -0,0 +1,64 @@
...
+    protected ArrowBuf doCompress(BufferAllocator allocator, ArrowBuf uncompressedBuffer)
+    {
+        throw new UnsupportedOperationException("Compression is not supported client-side");
+    }
+
+    @Override
+    protected ArrowBuf doDecompress(BufferAllocator allocator, ArrowBuf compressedBuffer)
+    {
+        int decompressedLength = toIntExact(readUncompressedLength(compressedBuffer));
+        ArrowBuf uncompressedBuffer = allocator.buffer(decompressedLength);
```
RoeyoOgen, line 41: allocator.buffer(decompressedLength) allocates native memory before verifying decompression success — if decompressZstd throws, uncompressedBuffer will leak - maybe clean-up in a finally.

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
```diff
@@ -0,0 +1,64 @@
...
+    @Override
+    protected ArrowBuf doCompress(BufferAllocator allocator, ArrowBuf uncompressedBuffer)
+    {
+        throw new UnsupportedOperationException("Compression is not supported client-side");
+    }
+
+    @Override
+    protected ArrowBuf doDecompress(BufferAllocator allocator, ArrowBuf compressedBuffer)
+    {
+        int decompressedLength = toIntExact(readUncompressedLength(compressedBuffer));
```
RoeyoOgen, line 40:
```suggestion
long maybeDecompressedLength = readUncompressedLength(compressedBuffer);
    if (maybeDecompressedLength <= 0 || maybeDecompressedLength > Integer.MAX_VALUE) {
        throw new IllegalStateException("Unsupported or missing uncompressed length in zstd frame: " + maybeDecompressedLength);
    }
    int decompressedLength = toIntExact(maybeDecompressedLength);
```
explicit handling and clearer error message

core/trino-main/src/main/java/io/trino/server/protocol/spooling/encoding/ArrowCompressionFactory.java
```diff
@@ -0,0 +1,41 @@
...
+    public CompressionCodec createCodec(CompressionUtil.CodecType codecType)
+    {
+        if (requireNonNull(codecType) == CompressionUtil.CodecType.ZSTD) {
+            return new ZstdCompressionCodec();
+        }
+        throw new IllegalArgumentException("Compression type not supported: " + codecType);
+    }
+
+    @Override
+    public CompressionCodec createCodec(CompressionUtil.CodecType codecType, int compressionLevel)
```
RoeyoOgen, line 34: compressionLevel is ignored

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
```diff
@@ -0,0 +1,64 @@
...
+        ArrowBuf uncompressedBuffer = allocator.buffer(decompressedLength);
+
+        ByteBuffer inputBuffer = compressedBuffer.nioBuffer(SIZE_OF_UNCOMPRESSED_LENGTH, toIntExact(compressedBuffer.writerIndex() - SIZE_OF_UNCOMPRESSED_LENGTH));
+        ByteBuffer outputBuffer = uncompressedBuffer.nioBuffer(0, decompressedLength);
+
+        long uncompressedSize = decompressZstd(inputBuffer, outputBuffer);
+
+        if (uncompressedSize != decompressedLength) {
+            uncompressedBuffer.close();
+            throw new RuntimeException(
```
RoeyoOgen, line 50: not IOException?

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
```diff
@@ -0,0 +1,64 @@
...
+                    "Expected != actual decompressed length: "
+                            + decompressedLength
+                            + " != "
+                            + uncompressedSize);
```
RoeyoOgen, line 54:
```suggestion
                    String.format("zstd decompressed size mismatch: expected=%d actual=%d", decompressedLength, uncompressedSize)
```

chenjian2664:
@jonasbrami Before adjusting all the “TODO”s in https://github.com/trinodb/trino/pull/25015, could you start a discussion with the community first? The changes look quite large

jonasbrami:
> @jonasbrami Before adjusting all the “TODO”s in #25015, could you start a discussion with the community first? The changes look quite large

Hi @chenjian2664 , is the Trino slack a good channel for discussing this ?

chenjian2664:
@jonasbrami I think it's ok to raise discussion via slack in `dev` or `core-dev`

adamchen:
This PR would be super useful for some projects that I'm working on - sad to see it's closed currently.

Reading through this, is the outstanding question around type mappings? If so, I'd be happy to dedicate some time to help move that forward. Let me know what I can do to help!

Chaho12:
We should get this moving and not closed

nickBes:
This is a great initiative. How can it be revived?

RoeyoOgen:
@jonasbrami 
Looks like there is lots of interest

Wanna pick it up again?

adamchen:
As I said above, if you need anything extra, happy to help @jonasbrami !

jonasbrami:
> @jonasbrami Looks like there is lots of interest
> 
> Wanna pick it up again?

Hey
Yes sure! Maybe we should discuss the direction we want this to take in the next Trino contributor call ?

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
jonasbrami: The allocated buffer is already released on failure via the existing try { ... } catch (Throwable t) { uncompressedBuffer.close(); throw t; }. A finally would be incorrect here, since on the success path the buffer is returned and its ownership transferred to the caller.

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
jonasbrami: Added: the uncompressed length is now read as a long and validated (rejecting <= 0 and > Integer.MAX_VALUE) with a clear message before toIntExact.

core/trino-main/src/main/java/io/trino/server/protocol/spooling/encoding/ArrowCompressionFactory.java
jonasbrami: Trino spooling exposes no configurable compression level, so it is intentionally ignored -- consistent with the json+zstd encoding, which also compresses at zstd default level. The two overloads are now deduplicated (the two-arg one delegates to createCodec(codecType)) and the comment explains why.

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
jonasbrami: A checked IOException is not possible here: Arrow CompressionCodec.decompress/doDecompress declare no checked exceptions (hence the unchecked throw). Changed RuntimeException to IllegalStateException, a more specific unchecked exception for the corrupt-frame case.

client/trino-client/src/main/java/io/trino/client/spooling/encoding/AirliftZstdCompressionCodec.java
jonasbrami: Done -- now uses format("zstd decompressed size mismatch: expected=%d actual=%d", decompressedLength, uncompressedSize).

lidavidm:
Just to chime in - I'm one of the developers behind Arrow, Arrow Java, and Arrow Flight - please feel free to ping me if there are questions or concerns around Arrow, I'd love to see this happen!

koszti:
> Just to chime in - I'm one of the developers behind Arrow, Arrow Java, and Arrow Flight - please feel free to ping me if there are questions or concerns around Arrow, I'd love to see this happen!

thanks @lidavidm . I think one of the main blockers of making arrow available in trino is this issue: https://github.com/apache/arrow-java/issues/511 - trino doesn't want to depend on the deprecated `sun.misc.Unsafe` API.

lidavidm:
That's long been a thorn in my side too, but I'm not sure there's an easy way out:
- We might be able to replace those APIs and use normal on-heap memory, but we'd need to verify this doesn't cause a performance regression for current users
- If we could rely on Java 25+, we could use FFM, but of course that's not a realistic requirement
- If there is a nontrivial performance difference, I'm not sure we'd have the capacity to maintain effectively two forks of arrow-java

That said I was recently made aware that someone had already made a fork that uses on-heap memory so possibly we could integrate it back into upstream

PR #26368, opened 2025-08-07
Cleanup ExchangeContext

pettyjamesm:
## Description
Simple refactor that drops an unnecessary interface indirection and makes `ExchangeContext` a record class.

## Release notes


pettyjamesm:
Closing, minor cleanup isn't worth an SPI break cycle

PR #26371, opened 2025-08-07
Handle long running queries in DirectTrinoClient

mwd410:
## Description

When a query being executed by DirectTrinoClient takes longer than the query.client.timeout, the query is then canceled. This client is used in some cases when executing queries from within the coordinator

The `QueryTracker` uses this timeout in its `private boolean isAbandoned(T query)` method to check it vs the last heartbeat. @lukasz-stec recommended i periodically record a heartbeat while waiting for this future to avoid falling into that case.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
+++ b/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
@@ -21,6 +21,7 @@ import io.trino.dispatcher.DispatchManager;
 import io.trino.dispatcher.DispatchQuery;
 import io.trino.exchange.DirectExchangeInput;
 import io.trino.execution.QueryManager;
+import io.trino.execution.QueryManagerConfig;
 import io.trino.execution.QueryState;
 import io.trino.execution.buffer.PageDeserializer;
 import io.trino.memory.context.SimpleLocalMemoryContext;
@@ -40,6 +41,8 @@ import java.net.URI;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
 
 import static io.airlift.concurrent.MoreFutures.whenAnyComplete;
 import static io.trino.SystemSessionProperties.getRetryPolicy;
@@ -56,13 +59,20 @@ public class DirectTrinoClient
     private final QueryManager queryManager;
     private final DirectExchangeClientSupplier directExchangeClientSupplier;
     private final BlockEncodingSerde blockEncodingSerde;
+    private final long heartBeatIntervalMillis;
 
-    public DirectTrinoClient(DispatchManager dispatchManager, QueryManager queryManager, DirectExchangeClientSupplier directExchangeClientSupplier, BlockEncodingSerde blockEncodingSerde)
+    public DirectTrinoClient(
+            DispatchManager dispatchManager,
+            QueryManager queryManager,
+            QueryManagerConfig queryManagerConfig,
+            DirectExchangeClientSupplier directExchangeClientSupplier,
+            BlockEncodingSerde blockEncodingSerde)
     {
         this.dispatchManager = requireNonNull(dispatchManager, "dispatchManager is null");
         this.queryManager = requireNonNull(queryManager, "queryManager is null");
         this.directExchangeClientSupplier = requireNonNull(directExchangeClientSupplier, "directExchangeClientSupplier is null");
         this.blockEncodingSerde = requireNonNull(blockEncodingSerde, "blockEncodingSerde is null");
+        this.heartBeatIntervalMillis = requireNonNull(queryManagerConfig, "queryManagerConfig is null").getClientTimeout().toMillis() / 2;
     }
 
     public DispatchQuery execute(SessionContext sessionContext, @Language("SQL") String sql, QueryResultsListener queryResultsListener)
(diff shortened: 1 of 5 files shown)


core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
```diff
@@ -103,7 +113,27 @@ public DispatchQuery execute(SessionContext sessionContext, @Language("SQL") Str
                     Page page = pageDeserializer.deserialize(serializedPage);
                     queryResultsListener.consumeOutputPage(page);
                 }
-                getQueryFuture(whenAnyComplete(ImmutableList.of(queryManager.getStateChange(queryId, state), exchangeClient.isBlocked())));
+
+                ListenableFuture<Object> anyCompleteFuture = whenAnyComplete(ImmutableList.of(
+                        queryManager.getStateChange(queryId, state),
+                        exchangeClient.isBlocked()));
+                for (boolean anyComplete = false; !anyComplete; dispatchQuery.recordHeartbeat()) {
```
lukasz-stec: nit: I think waiting for the `future.isDone()` instead of a separate variable would make the code a bit simpler
```
                while (!future.isDone()){
                    try {
                        future.get(heartBeatIntervalMillis, TimeUnit.MILLISECONDS);
                    }
                    catch (TimeoutException e) {
                        // continue waiting until the query state changes or the exchange client is blocked.
                        // we need to periodically record the heartbeat to prevent the query from being purged
                        dispatchQuery.recordHeartbeat();
                    }
...
```

lukasz-stec (commented):
Looks good. Please add a test with a small `query.client.timeout` and a query running longer

core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
```diff
@@ -21,6 +21,7 @@
 import io.trino.dispatcher.DispatchQuery;
```
kokosing, line 21: > When a query being executed by DirectTrinoClient takes longer than the query.client.timeout,
the query is then purged.

It is not "purged". The engine marks this query as abandoned by client and cancels it.

kokosing (commented):
Is it possible to test it it works? For example you can have a query that takes 5 seconds to complete and `query.client.timeout` set to 1 second, and see that you are able to run such query.

core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
mwd410: agreed. good call

core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
mwd410: i will adjust comment to `canceled` instead of `purged`

testing/trino-tests/src/test/java/io/trino/client/direct/TestDirectTrinoClient.java
```diff
@@ -0,0 +1,70 @@
...
+        queryRunner.execute("CREATE TABLE blackhole.test_table (col1 VARCHAR, col2 INTEGER)" +
+                "WITH (" +
+                "   split_count = 1, " +
+                "   pages_per_split = 1, " +
+                "   rows_per_page = 1, " +
+                "   page_processing_delay = '10s'" +
+                ")");
+    }
+
+    @Test
```
kokosing, line 58: Add timeout

testing/trino-tests/src/test/java/io/trino/client/direct/TestDirectTrinoClient.java
```diff
@@ -0,0 +1,70 @@
...
+    public void testDirectTrinoClientLongQuery()
+    {
+        String sql = "SELECT * FROM blackhole.test_table";
+        try {
```
kokosing: there is no need to try-catch, any exception will fail the test anyway

testing/trino-tests/src/test/java/io/trino/client/direct/TestDirectTrinoClient.java
```diff
@@ -0,0 +1,70 @@
...
+    public void setup()
+            throws Exception
+    {
+        queryRunner = new StandaloneQueryRunner(
+                TEST_SESSION,
+                builder -> builder.overrideProperties(ImmutableMap.of(
+                        "query.client.timeout", "1s")));
+        queryRunner.installPlugin(new BlackHolePlugin());
+        queryRunner.createCatalog("blackhole", "blackhole");
+        queryRunner.execute("CREATE TABLE blackhole.test_table (col1 VARCHAR, col2 INTEGER)" +
```
kokosing: call it `slow_test_table`

testing/trino-tests/src/test/java/io/trino/client/direct/TestDirectTrinoClient.java
```diff
@@ -0,0 +1,60 @@
...
+                builder -> builder.overrideProperties(ImmutableMap.of(
+                        "query.client.timeout", "1s")));
+        queryRunner.installPlugin(new BlackHolePlugin());
+        queryRunner.createCatalog("blackhole", "blackhole");
+        queryRunner.execute("CREATE TABLE blackhole.slow_test_table (col1 VARCHAR, col2 INTEGER)" +
+                "WITH (" +
+                "   split_count = 1, " +
+                "   pages_per_split = 1, " +
+                "   rows_per_page = 1, " +
+                "   page_processing_delay = '10s'" +
```
lukasz-stec: nit: this could be lower e.g. 3s

lukasz-stec (approved):
lgtm

kokosing:
Ping me when it is green

kokosing (approved):

kokosing:
Thanks!

core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
```diff
@@ -103,7 +113,27 @@ public DispatchQuery execute(SessionContext sessionContext, @Language("SQL") Str
...
+                        queryManager.getStateChange(queryId, state),
+                        exchangeClient.isBlocked()));
+                while (!anyCompleteFuture.isDone()) {
+                    try {
+                        anyCompleteFuture.get(heartBeatIntervalMillis, TimeUnit.MILLISECONDS);
+                    }
+                    catch (TimeoutException e) {
+                        // continue waiting until the query state changes or the exchange client is blocked.
+                        // we need to periodically record the heartbeat to prevent the query from being canceled
+                        dispatchQuery.recordHeartbeat();
```
chenjian2664, line 127: Nice! I’ve seen abandoned queries before, especially when the cluster was under high pressure — it have been the actual cause back then

PR #26386, opened 2025-08-12
Add node identifier to spooled segments

wendigo:
This prevents collisions when spooling on large number of nodes

## Release notes

```markdown
## Section
```


plugin/trino-spooling-filesystem/src/main/java/io/trino/spooling/filesystem/ExpiringClock.java
```diff
@@ -0,0 +1,51 @@
...
+import static java.util.Objects.requireNonNull;
+
+public class ExpiringClock
+        extends Clock
+{
+    private final long ttlMillis;
+
+    public ExpiringClock(Duration ttl)
+    {
+        this.ttlMillis = requireNonNull(ttl, "ttl is null").toMillis();
```
raunaqmorarka: requireNonNull is unnecessary here

PR #26390, opened 2025-08-12
Record error message in S3FileSystem#deleteObjects

raunaqmorarka:
## Description
Record error message in S3FileSystem#deleteObjects

## Release notes

```markdown
## Section
```


lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -203,7 +203,7 @@ private void deleteObjects(Collection<Location> locations)
                 try {
                     DeleteObjectsResponse response = client.deleteObjects(request);
                     for (S3Error error : response.errors()) {
-                        failures.put("s3://%s/%s".formatted(bucket, error.key()), error.code());
+                        failures.put("s3://%s/%s".formatted(bucket, error.key()), "%s (%s)".formatted(error.message(), error.code()));
```
wendigo: S3Error has a toString implementation. Can we use it instead?

wendigo: `failures.put("s3://%s/%s".formatted(bucket, error.key()), error.toString())`

raunaqmorarka: If we only use that, then we will not get the full path that we get from `"s3://%s/%s".formatted(bucket, error.key())`.
If we use it in addition to that, then it will print redundant info.

PR #26404, opened 2025-08-13
Check block isNull array on copy

pettyjamesm:
## Description
Opportunistically discards block `isNull` arrays during copy when the resulting block contains no null positions.

## Release notes

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
@@ -238,7 +238,8 @@ public final class ArrayBlock
         IntArrayList valuesPositions = new IntArrayList();
         for (int i = 0; i < length; i++) {
             int position = positions[offset + i];
-            if (newValueIsNull != null && isNull(position)) {
+            checkReadablePosition(this, position);
+            if (valueIsNull != null && valueIsNull[position + arrayOffset]) {
                 hasNull = true;
                 newValueIsNull[i] = true;
                 newOffsets[i + 1] = newOffsets[i];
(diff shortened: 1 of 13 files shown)


core/trino-spi/src/main/java/io/trino/spi/block/BlockUtil.java
```diff
@@ -108,6 +108,41 @@ static int calculateBlockResetBytes(int currentBytes)
...
+        }
+        // If no nulls encountered, return null as the result
+        if (firstNull == length) {
+            return null;
+        }
+        boolean[] result = new boolean[length];
+        int offset = 0;
+        for (; offset < firstNull; offset++) {
+            result[offset] = false;
+        }
```
chenjian2664: The initial value of boolean array is false by default, we can skip it

core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
```diff
@@ -238,7 +238,8 @@ public ArrayBlock copyPositions(int[] positions, int offset, int length)
         IntArrayList valuesPositions = new IntArrayList();
         for (int i = 0; i < length; i++) {
             int position = positions[offset + i];
-            if (newValueIsNull != null && isNull(position)) {
+            checkReadablePosition(this, position);
+            if (valueIsNull != null && valueIsNull[position + arrayOffset]) {
                 hasNull = true;
```
chenjian2664: What is the improvement?

pettyjamesm: - Unconditionally checks isReadablePosition (previously, this was only called as part of isNull when newValueIsNull != null)
- Inlines accessing valueIsNull instead of going through the isNull method (which wasn’t quite equivalent after inlining due to the checkReadablePosition call, and in any case was dependent on a JIT inlining decision being made)

core/trino-spi/src/main/java/io/trino/spi/block/BlockUtil.java
pettyjamesm: I thought this would be relevant for the JIT to elide a `memset` over the initial array during allocation, but after microbenchmarking it appears not to be and `System.arraycopy` performs better than the hand-rolled loops- so I've updated in the latest revision.

dain (approved):

PR #26405, opened 2025-08-13
Introduce startOffset to RowBlock to reduce possibly redundant memory…

SongChujun:
… allocation of rowIsNull array

## Release notes

```markdown
## Section
```


core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
```diff
@@ -101,31 +105,36 @@ private RowBlock(int positionCount, @Nullable boolean[] rowIsNull, Block[] field
...
-            if (positionCount != fieldBlocks[i].getPositionCount()) {
-                throw new IllegalArgumentException("Expected field %s to have %s positions but has %s positions".formatted(i, positionCount, fieldBlocks[i].getPositionCount()));
+        int firstFieldBlockPositionCount = fieldBlocks[0].getPositionCount();
+        for (int i = 1; i < fieldBlocks.length; i++) {
+            if (firstFieldBlockPositionCount != fieldBlocks[i].getPositionCount()) {
+                throw new IllegalArgumentException(format("length of field blocks differ: field 0: %s, block %s: %s", firstFieldBlockPositionCount, i, fieldBlocks[i].getPositionCount()));
             }
         }
 
+        if (startOffset + positionCount > fieldBlocks[0].getPositionCount()) {
```
pettyjamesm: Nit: `firstFieldBlockPositionCount` here too

core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
```diff
@@ -232,35 +241,32 @@ public RowBlock copyPositions(int[] positions, int offset, int length)
         checkArrayRange(positions, offset, length);
 
         Block[] newBlocks = new Block[fieldBlocks.length];
+
+        int[] rawPositions = new int[positions.length];
```
pettyjamesm: This loop can be skipped if `startOffset == 0`

core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
```diff
@@ -282,14 +288,14 @@ public RowBlock copyRegion(int positionOffset, int length)
 
         Block[] newBlocks = new Block[fieldBlocks.length];
         for (int i = 0; i < fieldBlocks.length; i++) {
-            newBlocks[i] = fieldBlocks[i].copyRegion(positionOffset, length);
+            newBlocks[i] = fieldBlocks[i].copyRegion(startOffset + positionOffset, length);
         }
 
-        boolean[] newRowIsNull = rowIsNull == null ? null : compactArray(rowIsNull, positionOffset, length);
+        boolean[] newRowIsNull = rowIsNull == null ? null : compactArray(rowIsNull, positionOffset + startOffset, length);
         if (newRowIsNull == rowIsNull && arraySame(newBlocks, fieldBlocks)) {
```
pettyjamesm: Technically we should check `startOffset == 0` here too, although the other checks guarantee that it will always be safe since `newRowIsNull == rowIsNull` would never match unless that was the case.

core/trino-spi/src/main/java/io/trino/spi/block/RowBlockEncoding.java
```diff
@@ -43,10 +44,10 @@ public void writeBlock(BlockEncodingSerde blockEncodingSerde, SliceOutput sliceO
...
-        sliceOutput.appendInt(rawFieldBlocks.length);
-        for (Block rawFieldBlock : rawFieldBlocks) {
-            blockEncodingSerde.writeBlock(sliceOutput, rawFieldBlock);
+        List<Block> fieldBlocks = rowBlock.getFieldBlocks();
+        sliceOutput.appendInt(fieldBlocks.size());
+        for (Block fieldBlock : fieldBlocks) {
+            blockEncodingSerde.writeBlock(sliceOutput, fieldBlock);
         }
 
         EncoderUtil.encodeNullsAsBits(sliceOutput, block);
```
pettyjamesm: After https://github.com/trinodb/trino/pull/26437 merged, you'll need to propagate the offset here.

core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
```diff
@@ -101,31 +105,36 @@ private RowBlock(int positionCount, @Nullable boolean[] rowIsNull, Block[] field
...
-                throw new IllegalArgumentException("Expected field %s to have %s positions but has %s positions".formatted(i, positionCount, fieldBlocks[i].getPositionCount()));
+        int firstFieldBlockPositionCount = fieldBlocks[0].getPositionCount();
+        for (int i = 1; i < fieldBlocks.length; i++) {
+            if (firstFieldBlockPositionCount != fieldBlocks[i].getPositionCount()) {
+                throw new IllegalArgumentException(format("length of field blocks differ: field 0: %s, block %s: %s", firstFieldBlockPositionCount, i, fieldBlocks[i].getPositionCount()));
             }
         }
 
+        if (startOffset + positionCount > fieldBlocks[0].getPositionCount()) {
+            throw new IllegalArgumentException("sum of startOffset and positionCount is greater than fieldBlocks positionCount");
```
dain: I think we can improve this error message.  The check is really, we expected the field blocks to be at least a certain size.  I don't think we need to mention our internal variables

core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
```diff
@@ -101,31 +105,36 @@ private RowBlock(int positionCount, @Nullable boolean[] rowIsNull, Block[] field
...
         this.rowIsNull = positionCount == 0 ? null : rowIsNull;
         this.fieldBlocks = fieldBlocks;
-        this.fieldBlocksList = List.of(fieldBlocks);
     }
 
-    Block[] getRawFieldBlocks()
+    public List<Block> getFieldBlocks()
     {
-        return fieldBlocks;
+        return Arrays.stream(fieldBlocks).map(block -> block.getRegion(startOffset, positionCount)).toList();
```
dain, line 141: This makes `getFieldBlocks` significantly more expensive, so we should review the users.  I did a quick skim and most uses are in expensive parts of the code like file writing, but on usage that stood out is `RowBlockBuilder`.  It has 4 methods that use this method to append a single row to the builder.  We should update all of those to access the raw blocks (which it can because it is in the same package).  For the other usages, we should look for any code that is getting blocks to access a single field because this is now a `getRegion` call.  Also double check the compilers to see if they call this directly.

Finally add some java doc here saying that this is calling get region on the underlying block.

@pettyjamesm Do you think it is worth adding a boolean field to know that we already have perfectly sized blocks, so we don't need to do the region call?

pettyjamesm: > @pettyjamesm Do you think it is worth adding a boolean field to know that we already have perfectly sized blocks, so we don't need to do the region call?

Maybe just a private boolean method since the check for perfectly sized is just `startOffset == 0`

core/trino-main/src/main/java/io/trino/operator/output/RowPositionsAppender.java
```diff
@@ -79,9 +78,17 @@ public void append(IntArrayList positions, ValueBlock block)
...
+        int[] adjustedPositions = new int[positions.size()];
+        for (int i = 0; i < positions.size(); i++) {
+            adjustedPositions[i] = startOffset + positions.getInt(i);
+        }
+
+        Block[] rawFieldBlocks = sourceRowBlock.getRawFieldBlocks();
 
         for (int i = 0; i < fieldAppenders.length; i++) {
-            fieldAppenders[i].append(positions, sourceRowBlock.getFieldBlock(i));
+            fieldAppenders[i].append(IntArrayList.wrap(adjustedPositions), rawFieldBlocks[i]);
```
pettyjamesm: Lift `IntArrayList.wrap(adjustedPositions)` out of the loop

core/trino-spi/src/main/java/io/trino/spi/block/RowBlockBuilder.java
```diff
@@ -265,15 +270,22 @@ public void appendPositions(ValueBlock block, int[] positions, int offset, int l
         RowBlock rowBlock = (RowBlock) block;
         ensureCapacity(positionCount + length);
 
-        List<Block> fieldBlocks = rowBlock.getFieldBlocks();
+        Block[] rawFieldBlocks = rowBlock.getRawFieldBlocks();
+        int startOffset = rowBlock.getOffsetBase();
+
+        int[] adjustedPositions = new int[length];
```
pettyjamesm: If `startOffset == 0`, `adjustedPositions` is unnecessary

core/trino-main/src/main/java/io/trino/operator/output/RowPositionsAppender.java
```diff
@@ -79,9 +78,17 @@ public void append(IntArrayList positions, ValueBlock block)
         }
         ensureCapacity(positions.size());
         RowBlock sourceRowBlock = (RowBlock) block;
+        int startOffset = sourceRowBlock.getOffsetBase();
+
+        int[] adjustedPositions = new int[positions.size()];
```
pettyjamesm: If `startOffset == 0`, `adjustedPositions` is unnecessary

core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
pettyjamesm: Looks like I was slightly mistaken, the check needs to be `startOffset == 0 && positionCount == <field block>.getPositionCount()`, so a `boolean` makes sense but would assume that all field blocks have the same length which I'm not sure is a necessary constraint. I would err on the side of flexibility on that point, but don't feel particularly strongly.

The only usage I can find in compilers is [AccumulatorCompiler](https://github.com/trinodb/trino/blob/76808e820a971b5a718da7c68d6c9846767ee5d4/core/trino-main/src/main/java/io/trino/operator/aggregation/AccumulatorCompiler.java#L718) which shouldn't be significantly more expensive. Other code generation sites of interest all appear to be going through `SqlRow` which should be safe, but this search is expensive so if you know of a better way than searching for the strings `"getFieldBlock"`, `"getFieldBlocks"`, and `"getRawFieldBlocks"`  then you might double check my work on that.

SongChujun: The assumption that all all field blocks have the same length is guaranteed in the RowBlock constructor 
```
int firstFieldBlockPositionCount = fieldBlocks[0].getPositionCount();
for (int i = 1; i < fieldBlocks.length; i++) {
    if (firstFieldBlockPositionCount != fieldBlocks[i].getPositionCount()) {
        throw new IllegalArgumentException(format("length of field blocks differ: field 0: %s, block %s: %s", firstFieldBlockPositionCount, i, fieldBlocks[i].getPositionCount()));
    }
}
``` 

So this assumption is valid. I am also on the side of flexibility at this point. But also open to other thoughts.

core/trino-spi/src/main/java/io/trino/spi/block/RowBlock.java
```diff
@@ -101,33 +105,63 @@ private RowBlock(int positionCount, @Nullable boolean[] rowIsNull, Block[] field
...
     }
 
-    Block[] getRawFieldBlocks()
+    /**
+     * Returns the field blocks for this row block's visible region.
+     * <p>
+     * If this row block has a non-zero {@code startOffset}, each returned field block is a
+     * region view created via {@link Block#getRegion(int, int)} with this block's
+     * {@code startOffset} and {@code positionCount}. When {@code startOffset} is zero,
+     * the underlying field blocks are returned as-is.
```
dain: This comment should be updated to state the position count.  Or maybe just say if the underlying blocks are fully visible...

core/trino-spi/src/main/java/io/trino/spi/block/RowBlockBuilder.java
```diff
@@ -162,15 +163,17 @@ public void appendRange(ValueBlock block, int offset, int length)
         RowBlock rowBlock = (RowBlock) block;
         ensureCapacity(positionCount + length);
 
-        List<Block> fieldBlocks = rowBlock.getFieldBlocks();
+        Block[] rawFieldBlocks = rowBlock.getRawFieldBlocks();
+        int startOffset = rowBlock.getOffsetBase();
+
```
dain, line 169: nit: be consistent throughout this code with the ordering and formatting.  The above code has the fields in a the opposite order and no space.

PR #26411, opened 2025-08-14
Add configuration to disable managed table creation in Hive connector

OmerRaifler:
## Description

This PR introduces a new Hive connector configuration property:

`hive.managed-table-creates-enabled` (default: true)

When set to false, this property prevents creation of managed (internal) tables in Hive.

There is an existing configuration, `hive.non-managed-table-creates-enabled`, which allows controlling creation of non-managed (external) tables. However, there is currently no way to disable creation of managed tables.

In our environment, managed tables have historically caused locks in the Hive Metastore database. These locks impacted other workloads and required manual intervention. We prefer to enforce external tables only, avoiding managed tables entirely.

This change allows administrators to explicitly block managed table creation via configuration, improving stability and preventing accidental use.

## Release notes

```markdown
## Hive Connector
* Add configuration property `hive.managed-table-creates-enabled` to control whether creation of managed (internal) tables is allowed.
```


chenjian2664:
It's strange that support a toggle don't allow create managed table to me, could you share more context about what problem you are resolving?

OmerRaifler:
[Link to Slack thread](https://trinodb.slack.com/archives/CP1MUNEUX/p1755170719371759)

OmerRaifler:
I’ve decided to close this PR after looking deeper into the implications. The change could break existing functionality for users who rely on the  `hive.create.as.external.legacy=true` property.  

With that property enabled, if a `CREATE TABLE` statement doesn’t include an `external_location`,  
the Hive Metastore translates the table into an **EXTERNAL** table with the properties:  

- `translated_to_external=true`  
- `external.table.purge=true`  

With my change, instead of that behavior, the statement would fail with an exception:  

"Cannot create managed Hive table"

because no `external_location` was provided.  

In our environment we *do* use `hive.create.as.external.legacy=true`, and we’re okay with the current behavior.  
It would have been nice to restrict managed table creation by explicitly looking for  
`CREATE MANAGED TABLE …`, but that turns out to be more complicated.  

- When `hive.create.as.external.legacy=true`:  
  - Managed tables are only created if the **MANAGED** keyword is explicitly specified.  
  - If `external_location` is null, the table is translated to **EXTERNAL** as described above.  

- When `hive.create.as.external.legacy=false`:  
  - Managed tables are created whenever `external_location` is null.  

Given that, enforcing a restriction without breaking the legacy property’s semantics doesn’t look straightforward.

PR #22879, opened 2024-07-30
Load event listeners on coordinator only

sopel39:
Instantiating event listeners on workers doesn't make sense even if event listener config is present. This makes deployment easier and less error prone.

PR #26425, opened 2025-08-16
Add ability to load event listeners on workers

anton-kutuzov:
## Description
After these [changes](https://github.com/trinodb/trino/pull/22879), the [query log plugin](https://github.com/rchukh/trino-querylog) stopped working on the workers, and as a result we no longer have logs of splits from them.
I added the ability to run event listeners on the worker if it is necessary.

## Release notes

```markdown
## Section
```


raunaqmorarka:
While it is a genuine bug that `io.trino.spi.eventlistener.EventListener#splitCompleted` is not working anymore, I'm wondering what's the use case for it. We already get detailed source stage metrics through `io.trino.spi.eventlistener.QueryStatistics#getOperatorSummaries` (see `io.trino.operator.OperatorStats#OperatorStats`). These also contain histograms for scheduled and CPU time which are useful for assessing skews in processing of splits.
The collection of individual split events feels overkill to me, especially with data lake connectors where we're taking about thousands of splits per query.
I would rather we remove support for `io.trino.spi.eventlistener.EventListener#splitCompleted` altogether and rely on the information already present in `io.trino.spi.eventlistener.EventListener#queryCompleted` instead. Since the coordinator generates splits and and knows when they are finished, it can provide additional metrics to EventListener, if needed, rather than having all workers send events.

anton-kutuzov:
We are using the query log plugin to collect all information about splits. After that, we store the logs in a ClickHouse database and use this data to create different metrics in Redash. For example, right now I need information about splits because I want to see how much time is spent on calculating column statistics and how many splits we generate.

raunaqmorarka:
> We are using the query log plugin to collect all information about splits. After that, we store the logs in a ClickHouse database and use this data to create different metrics in Redash. For example, right now I need information about splits because I want to see how much time is spent on calculating column statistics and how many splits we generate.

> how much time is spent on calculating column statistics 

That information is not part of splits processing at all. That is time spent in `io.trino.spi.connector.ConnectorMetadata#getTableStatistics` calls. This can be tracked through https://trino.io/docs/current/admin/opentelemetry.html or through JMX metrics within some connectors. Typically we track `io.trino.execution.QueryStats#planningTime`, `io.trino.execution.QueryStats#planningCpuTime` and the most time consuming optimizer rules in `io.trino.execution.QueryStats#optimizerRulesSummaries`. A high planning time with low planning CPU time along with the specific rules which took time are usually enough to infer a high statistics fetch time. In any case, splits processing on workers has nothing to do with that, this something that happens entirely in the coordinator.

> how many splits we generate

That is already available in the operator summaries that are part of QueryCompletedEvent. The number of drivers in scan operator is equal to the number of splits generated for that scan.

anton-kutuzov:
> That information is not part of splits processing at all. That is time spent in io.trino.spi.connector.ConnectorMetadata#getTableStatistics calls. This can be tracked through https://trino.io/docs/current/admin/opentelemetry.html or through JMX metrics within some connectors. Typically we track io.trino.execution.QueryStats#planningTime, io.trino.execution.QueryStats#planningCpuTime and the most time consuming optimizer rules in io.trino.execution.QueryStats#optimizerRulesSummaries. A high planning time with low planning CPU time along with the specific rules which took time are usually enough to infer a high statistics fetch time. In any case, splits processing on workers has nothing to do with that, this something that happens entirely in the coordinator.

I need information about how much time is spent on calculating statistics for columns. As I understand, `io.trino.spi.connector.ConnectorMetadata#getTableStatistics` is responsible for retrieving statistics from the metastore(in planning time), but I need to measure how much time is spent on calculation when the `hive.collect-column-statistics-on-write` property is enabled and and we are inserting into a table.

raunaqmorarka:
> measure how much time is spent on calculation when the hive.collect-column-statistics-on-write property is enabled and and we are inserting into a table.

Splits processing covers only the part of the query which reads data from the source tables. Column statistics computation for write takes places in the table writer stage, which is not captured by split completed events.
You'll need to look at `statisticsWallTime` and `statisticsCpuTime` fields in operator summary for TableWriterOperator, this should be in `io.trino.spi.eventlistener.QueryStatistics#getOperatorSummaries`.
There is also time spent in inserting the computed statistics into the metastore, this is recorded in finishingTime of the query or can be tracked through tracing `io.trino.plugin.hive.HiveMetadata#finishInsert` in open telemetry traces.

anton-kutuzov:
> Splits processing covers only the part of the query which reads data from the source tables. Column statistics computation for write takes places in the table writer stage, which is not captured by split completed events.
You'll need to look at statisticsWallTime and statisticsCpuTime fields in operator summary for TableWriterOperator, this should be in io.trino.spi.eventlistener.QueryStatistics#getOperatorSummaries.

Ok, thank you) Do I understand correctly that I don’t need information from the workers, because the statisticsWallTime and statisticsCpuTime values are logged on the coordinator in the queryCompleted method?

And the last thing: why we are looking at split statistics from workers for monitoring purposes—specifically, to see how much CPU and memory are used by a user or a query on a specific(each) worker? Could you help us find out where we can get this information?

raunaqmorarka:
> Do I understand correctly that I don’t need information from the workers, because the statisticsWallTime and statisticsCpuTime values are logged on the coordinator in the queryCompleted method?

yes

> how much CPU and memory are used by a user or a query on a specific(each) worker?

You can get query level CPU and memory usage from `io.trino.spi.eventlistener.QueryStatistics#cpuTime` and `io.trino.spi.eventlistener.QueryStatistics#peakUserMemoryBytes` which are part of QueryCompletedEvent.
You also have `io.trino.spi.eventlistener.QueryStatistics#peakTaskUserMemory` which gives peak memory used per worker.
You can also use task statistics from `io.trino.spi.eventlistener.QueryStatistics#taskStatistics` to get per worker cpu and memory usage distributions.
I'm not sure it is useful to further drill down to individual worker level stats, but if you really want to, then that's available in `io.trino.execution.StageInfo#tasks` which can be pulled through REST endpoint `io.trino.server.QueryResource#getQueryInfo` the same way as the Trino web UI.
The spilt completed events were only tracking the resource usage from table scan, they don't record the overall resource usage from the rest of the query on the worker, so they're unsuitable for what you want to do anyway.

anton-kutuzov:
> You can get query level CPU and memory usage from io.trino.spi.eventlistener.QueryStatistics#cpuTime and io.trino.spi.eventlistener.QueryStatistics#peakUserMemoryBytes which are part of QueryCompletedEvent.
You also have io.trino.spi.eventlistener.QueryStatistics#peakTaskUserMemory which gives peak memory used per worker.

We are already using it for monitoring, and it has helped us identify solutions for noisy neighbors. 

> I'm not sure it is useful to further drill down to individual worker level stats, but if you really want to, then that's available in io.trino.execution.StageInfo#tasks which can be pulled through REST endpoint io.trino.server.QueryResource#getQueryInfo the same way as the Trino web UI.

We want have the ability to find queries that put a heavy load on specific workers, because from time to time errors with unavailable workers occur (using JMX metrics, we can see that one worker has a higher load than others), and we want to understand the root cause. 

Thank you for your help we will think how to monitor.

Seems that I need to close PR.

raunaqmorarka:
I've created https://github.com/trinodb/trino/pull/26436 to officially drop support for SplitCompletedEvent to clarify the current situation.

PR #26427, opened 2025-08-17
Implement streamRelationColumns in BlackHole

chenjian2664:
## Release notes

```markdown
## Section
```


plugin/trino-blackhole/src/main/java/io/trino/plugin/blackhole/BlackHoleMetadata.java
```diff
@@ -185,9 +188,29 @@ public Map<SchemaTableName, List<ColumnMetadata>> listTableColumns(ConnectorSess
     @Override
     public Iterator<TableColumnsMetadata> streamTableColumns(ConnectorSession session, SchemaTablePrefix prefix)
     {
-        return tables.values().stream()
+        throw new UnsupportedOperationException("The deprecated streamTableColumns is not supported because streamRelationColumns is implemented instead");
+    }
+
+    @Override
+    public Iterator<RelationColumnsMetadata> streamRelationColumns(ConnectorSession session, Optional<String> schemaName, UnaryOperator<Set<SchemaTableName>> relationFilter)
```
ebyhr: This line is too long. Please put parameters on separate lines. Also, `synchronized` looks missing.

chenjian2664: Why need `synchronized` here?

plugin/trino-blackhole/src/main/java/io/trino/plugin/blackhole/BlackHoleMetadata.java
```diff
@@ -185,9 +188,32 @@ public Map<SchemaTableName, List<ColumnMetadata>> listTableColumns(ConnectorSess
...
                 .filter(table -> prefix.matches(table.toSchemaTableName()))
-                .map(handle -> TableColumnsMetadata.forTable(handle.toSchemaTableName(), handle.toTableMetadata().getColumns()))
+                .map(BlackHoleTableHandle::toTableMetadata)
+                .forEach(columnsMetadata -> {
+                    SchemaTableName name = columnsMetadata.getTable();
+                    relationColumns.put(name, RelationColumnsMetadata.forTable(name, columnsMetadata.getColumns()));
+                });
+
+        for (Map.Entry<SchemaTableName, ConnectorViewDefinition> entry : getViews(session, schemaName).entrySet()) {
+            relationColumns.put(entry.getKey(), RelationColumnsMetadata.forView(entry.getKey(), entry.getValue().getColumns()));
```
ebyhr, line 212: The current `BlackHoleMetadata.getViews` method ignores `schemaName` parameter. We shouldn't put columns on tables which exist in other schemas though they will be filtered in the engine.

plugin/trino-blackhole/src/main/java/io/trino/plugin/blackhole/BlackHoleMetadata.java
ebyhr: Not specific here. For instance, `BlackHoleMetadata.renameTable` method contains `tables.remove` & `tables.put` calls. Streaming columns may miss tables in the middle of renaming. No need to fix as it's pre-existing though.

ebyhr (approved):
Looks good except for a view comment.

plugin/trino-blackhole/src/main/java/io/trino/plugin/blackhole/BlackHoleMetadata.java
chenjian2664: ah, I think we should fix in `getViews` method instead of here?

ebyhr: Yeah, let's fix `getViews` method.

chenjian2664:
Thank you @ebyhr for clean ups of the code

PR #26429, opened 2025-08-18
Avoid using newFastAppend for writes in iceberg REFRESH MV

raunaqmorarka:
## Description
Avoid using newFastAppend for writes in iceberg REFRESH MV.
Increase parallelism for scanning manifests for merging during iceberg writes. 

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -1533,6 +1533,7 @@ public class IcebergMetadata
             cleanExtraOutputFiles(session, writtenFiles.build());
         }
 
+        appendFiles.scanManifestsWith(icebergScanExecutor);
         commitUpdateAndTransaction(appendFiles, session, transaction, "insert");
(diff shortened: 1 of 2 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1533,6 +1533,7 @@ public Optional<ConnectorOutputMetadata> finishInsert(
             cleanExtraOutputFiles(session, writtenFiles.build());
         }
 
+        appendFiles.scanManifestsWith(icebergScanExecutor);
```
chenjian2664, line 1536: What's the purpose to add this, for release memory pressure?
If we don't specify the threadpool here, Iceberg uses `Math.max(2, Runtime.getRuntime().availableProcessors())` this may make the process slower, since the process is less

raunaqmorarka: The executor here is coming from `iceberg.split-manager-threads`, which is bigger by default.
We want to avoid relying on internal threadpools of iceberg as the configuration of that is harder to tune and can be used for any operation.
We're consolidating all the reads of manifests into one threadpool which should keep memory usage controlled under high concurrency while still allowing more parallelism for individual queries when concurrency is lower.

PR #26432, opened 2025-08-18
Increase S3 delete batch size to 1000

raunaqmorarka:
## Description
Matches batch size used in legacy FS io.trino.hdfs.s3.TrinoS3FileSystem#DELETE_BATCH_SIZE
1000 is the maximum batch size supported by S3 client 
Reduces potential of getting throttled by S3

## Release notes

```markdown
## Iceberg
* Reduce query failures from deletes on S3 failing due to throttling errors. ({issue}`26432`)
```

### DIFF
--- a/lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
+++ b/lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
@@ -1214,7 +1214,7 @@ public abstract class AbstractTestTrinoFileSystem
             throws IOException
     {
         try (Closer closer = Closer.create()) {
-            List<TempBlob> blobs = randomBlobs(closer);
+            List<TempBlob> blobs = randomBlobs(closer, 100);
 
             List<Location> sortedLocations = blobs.stream()
                         .map(TempBlob::location)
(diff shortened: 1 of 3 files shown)


lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -858,6 +858,25 @@ void testDeleteFiles()
...
     }
 
+    @Test
+    void testDeleteManyFiles()
+            throws IOException
+    {
+        try (Closer closer = Closer.create()) {
+            // create a large number of files to test batch deletion over multiple batches
+            // io.trino.filesystem.s3.S3FileSystem.DELETE_BATCH_SIZE is 1000, so we create more files than that
+            List<TempBlob> blobs = randomBlobs(closer, 1100);
```
wendigo: I'm a little concerned about the cost of that test. It will call API 1100 times per build.

raunaqmorarka: right, i moved the test to TestS3FileSystemMinIo to avoid hitting S3 APIs

raunaqmorarka: using bigger batch size than 1000 is enough to fail test on minio as well

wendigo: @raunaqmorarka thanks!

PR #26435, opened 2025-08-18
Allow providing column handles for table execute function

vlad-lyutenko:
## Description

By this PR we want to give ability to Connectors to provide column handles for table execute function.
It could be useful for iceberg, delta lake connector

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/Metadata.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/Metadata.java
@@ -444,6 +444,8 @@ public interface Metadata
      */
     ColumnHandle getMergeRowIdColumnHandle(Session session, TableHandle tableHandle);
 
+    Optional<List<ColumnHandle>> getColumnHandlesForExecute(Session session, TableExecuteHandle tableExecuteHandle, TableHandle tableHandle);
+
     /**
      * Get the physical layout for updated or deleted rows of a MERGE operation.
      */
--- a/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
@@ -1250,6 +1250,14 @@ public final class MetadataManager
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


martint:
What's the motivation for this change? How will the connectors use this?

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -444,6 +444,8 @@ Optional<ConnectorOutputMetadata> finishRefreshMaterializedView(
      */
     ColumnHandle getMergeRowIdColumnHandle(Session session, TableHandle tableHandle);
 
+    Optional<List<ColumnHandle>> getColumnHandlesForExecute(Session session, TableExecuteHandle tableExecuteHandle, TableHandle tableHandle);
```
martint, line 447: What does the result of this method represent?

chenjian2664:
> What's the motivation for this change

Allow connector to provide additional column info for the table execute function.
This is especially useful when running `optimize` on iceberg with [row lineage](https://iceberg.apache.org/spec/#row-lineage-assignment) support, since the hidden column `_row_id` cannot be read without this change

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
ebyhr: @vlad-lyutenko Could you add javadoc explaining the above?

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -1249,6 +1249,14 @@ public ColumnHandle getMergeRowIdColumnHandle(Session session, TableHandle table
         return metadata.getMergeRowIdColumnHandle(session.toConnectorSession(catalogHandle), tableHandle.connectorHandle());
     }
 
+    @Override
```
ebyhr, line 1252: Is it possible to test this change with mock connector or something?

PR #26436, opened 2025-08-18
Remove code for collecting SplitCompletedEvent on workers

raunaqmorarka:
## Description
After https://github.com/trinodb/trino/pull/22879 SplitCompletedEvent are no longer being populated.
We already get detailed source stage metrics through `io.trino.spi.eventlistener.QueryStatistics#getOperatorSummaries`. These also contain histograms for scheduled and CPU time which are useful for assessing skews in processing of splits.
The collection of individual split events seems overkill and unnecessary to maintain at this point, especially with data lake connectors where we're taking about thousands of splits per query.
This PR removes support for `io.trino.spi.eventlistener.EventListener#splitCompleted` altogether and makes it clear that users need to rely on the information already present in `io.trino.spi.eventlistener.EventListener#queryCompleted` instead. Since the coordinator generates splits and and knows when they are finished, it can provide additional metrics to EventListener in future, if needed, rather than having all workers send events.

Supercedes https://github.com/trinodb/trino/pull/26425

## Release notes

```markdown
## Event listener
* Remove support for `io.trino.spi.eventlistener.EventListener#splitCompleted`. ({issue}`26436`)
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/eventlistener/EventListener.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/eventlistener/EventListener.java
@@ -19,6 +19,7 @@ public interface EventListener
 
     default void queryCompleted(QueryCompletedEvent queryCompletedEvent) {}
 
+    @Deprecated(forRemoval = true)
     default void splitCompleted(SplitCompletedEvent splitCompletedEvent)
     {
         throw new UnsupportedOperationException();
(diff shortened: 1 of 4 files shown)


raunaqmorarka:
> This is backward-incompatible change but since these events were not published for over a year now and no one until now complained, I'm fine with removing the leftovers and recommending to use the QueryCompletedEvent as a source of this information instead.

JFYI we're not breaking SPI compatiblity yet to make the transition easier. Its marked for removal and will be fully deleted in a future release.

wendigo:
@raunaqmorarka :shipit:

PR #26437, opened 2025-08-18
Improve BlockEncoding null bits packing

pettyjamesm:
## Description
Improve block serialization logic to prefer using primitive arrays over virtual calls when checking positions for nulls.

Benchmarks: [BenchmarkBlockSerde.serialize*](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/ed61da9bd57b3bc742bf0a55920bf78b/raw/f8237b39edca6dbc77122909ab921d2d62c7f5b8/null_packing_baseline.json,https://gist.githubusercontent.com/pettyjamesm/ed61da9bd57b3bc742bf0a55920bf78b/raw/83526f49bd05cf957663443375b79aa0283fe4d3/null_packing_improved_v3.json)

## Release notes

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
@@ -70,7 +70,10 @@ final class EncoderUtil
 
     /**
      * Append null values for the block as a stream of bits.
+     *
+     * @deprecated Use {@link EncoderUtil#encodeNullsAsBits(SliceOutput, boolean[], int, int)} instead
      */
+    @Deprecated(forRemoval = true)
     @SuppressWarnings({"NarrowingCompoundAssignment", "ImplicitNumericConversion"})
     public static void encodeNullsAsBits(SliceOutput sliceOutput, Block block)
     {
--- a/core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
@@ -29,6 +29,49 @@ final class EncoderUtil
      * Append null values for the block as a stream of bits.
      */
     @SuppressWarnings({"NarrowingCompoundAssignment", "ImplicitNumericConversion"})
+    public static void encodeNullsAsBits(SliceOutput sliceOutput, @Nullable boolean[] isNull, int offset, int length)
+    {
+        sliceOutput.writeBoolean(isNull != null);
+        if (isNull == null) {
+            return;
+        }
+        // inlined from Objects.checkFromIndexSize
+        if ((length | offset) < 0 || length > isNull.length - offset) {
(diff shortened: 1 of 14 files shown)


wendigo:
It this PR about reducing the network overhead for nulls or does it improve performance due to autovectorization as well?

pettyjamesm:
> It this PR about reducing the network overhead for nulls or does it improve performance due to autovectorization as well?

This PR doesn't change the serialized representation (and therefore has no impact on network overhead), the performance comes from better JIT code generation for the serialization loops, although not necessarily autovectorization specifically.

wendigo:
@pettyjamesm since you've shared benchmark results, can you contribute benchmark itself too?

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -21,10 +21,52 @@
...
+        sliceOutput.writeBoolean(isNull != null);
+        if (isNull == null) {
+            return;
+        }
+        checkFromIndexSize(offset, length, isNull.length);
+        byte[] packedIsNull = new byte[(length + 7) / 8];
+        int currentByte = 0;
+        for (int position = 0; position < (length & ~0b111); position += 8) {
+            byte value = 0;
+            value |= (isNull[position + offset] ? 1 : 0) << 7;
```
wendigo, line 46: Is it faster than:

```
            value |= block.isNull(position) ? 0b1000_0000 : 0;
            value |= block.isNull(position + 1) ? 0b0100_0000 : 0;
            value |= block.isNull(position + 2) ? 0b0010_0000 : 0;
            value |= block.isNull(position + 3) ? 0b0001_0000 : 0;
            value |= block.isNull(position + 4) ? 0b0000_1000 : 0;
            value |= block.isNull(position + 5) ? 0b0000_0100 : 0;
            value |= block.isNull(position + 6) ? 0b0000_0010 : 0;
            value |= block.isNull(position + 7) ? 0b0000_0001 : 0;
```

pettyjamesm: Yes, the new version performs better (and more consistently) than the previous implementation across all null rates in the benchmark including when branches are (nearly) perfectly predicted (null rate 0.01 and 0.99). The performance of the new version is nearly flat (with some variation in performance coming from the logic performing null-supression of the values array) where as we see a typical "M"-shaped curve with the previous version where performance relies on the branch prediction rate- which suggests the new version is branchless although unfortunately I don't have a JVM build with hdsdis available to show the generated assembly to prove it.

pettyjamesm:
> @pettyjamesm since you've shared benchmark results, can you contribute benchmark itself too?

The benchmark is already checked in as `BenchmarkBlockSerDe`. The results posted above only include the serialization benchmarks since the deserialization paths are unaffected by this change.

dain (approved):
Looks good

PR #23909, opened 2024-10-25
Improve performance when listing columns in Iceberg

shohamyamin:
# Improve Performance for Retrieving Tables Metadata from Iceberg Catalog

## Overview

This PR addresses a performance issue encountered when querying column metadata from the Iceberg catalog in Trino (Issue #[23468](https://github.com/trinodb/trino/issues/23468)). The primary concern is that requests to retrieve table metadata are executed sequentially for each table, which significantly impacts query performance when dealing with a large number of tables.

## Description

The issue manifests when executing the following query on the Iceberg catalog:

```sql
SELECT * FROM iceberg.information_schema.columns;
```

In environments with a large number of tables, the query’s response time increases considerably due to the sequential execution of catalog requests.

### Changes Made

To improve the performance, this PR introduces parallelism in the metadata retrieval process. The old code executed catalog requests sequentially for each table, which resulted in longer execution times. The updated code utilizes `CompletableFuture` to handle requests asynchronously, thereby reducing the overall execution time.

The updated implementation leverages `CompletableFuture` to process tables concurrently, significantly reducing the latency in retrieving column metadata.

## Additional Context and Related Issues

Fixes #23468

### Test Setup & Observations:
 I checked the code changes and get faster query in over x4 faster then without the proposed change
- Trino Version: 463 and trino with the code in master branch with the proposed change lets called it trino 464-SNAPSHOT
- Nessie Version: 0.94.4 (In memory)
- REST Catalog based on JDBC (e.g., Tabular REST Catalog)
- Testing environment: Docker Compose with MinIO for Iceberg storage and Jaeger for tracing.
- Hardware: Intel Core i9 processor, 32GB RAM

### Performance Comparison

As you can see in this table we get around 70% reduce in time for retrieving the columns

| **Number of Tables** | **Trino Version** | **Catalog Setup** | **Query Execution Time (s)** | **Performance Increase (%)** |
|-----------------------|-------------------|-------------------|-------------------------------|------------------------------|
| **10**                | Trino 463         | REST              | 0.691                         |                              |
|                       |                   | Nessie            | 1.07                          |                              |
|                       | Trino 464-SNAPSHOT         | REST              | 0.677                         |  2.02%    |
|                       |                   | Nessie            | 0.982                         | 8.23%       |
|-----------------------|-------------------|-------------------|-------------------------------|------------------------------|
| **100**               | Trino 463         | REST              | 1.24                          |                              |
|                       |                   | Nessie            | 1.89                          |                              |
[...]

PR #26438, opened 2025-08-18
Improve remove_orphan_files

raunaqmorarka:
## Description
Avoid separate scan and delete operations for "data" and "metadata" folders to reduce filesystem calls.
Scan manifest files for valid file paths in parallel.

## Release notes

```markdown
## Iceberg
* Improve performance of remove_orphan_files. ({issue}`26438`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2401,11 +2401,14 @@ public class IcebergMetadata
             }
 
             deleteFutures.forEach(MoreFutures::getFutureValue);
+            // All futures completed normally
+            deleteFutures.clear();
         }
         catch (IOException | UncheckedIOException e) {
             throw new TrinoException(ICEBERG_FILESYSTEM_ERROR, "Failed removing orphan files for table: " + schemaTableName, e);
         }
         finally {
+            // Ensure any futures still running are canceled in case of failure
             deleteFutures.forEach(future -> future.cancel(true));
         }
     }


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2302,17 +2303,19 @@ private void removeOrphanFiles(Table table, ConnectorSession session, SchemaTabl
                 }
 
                 validFileNames.add(fileName(manifest.path()));
-                try (ManifestReader<? extends ContentFile<?>> manifestReader = readerForManifest(manifest, table)) {
-                    for (ContentFile<?> contentFile : manifestReader.select(ImmutableList.of("file_path"))) {
-                        validFileNames.add(fileName(contentFile.location()));
+                manifestScanFutures.add(icebergScanExecutor.submit(() -> {
```
grantatspothero, line 2307: Looks like `icebergScanExecutor` is intended for any iceberg scan operation: https://github.com/trinodb/trino/pull/23909

Seems like a reasonable usage of that executor 👍

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2326,7 +2329,13 @@ private void removeOrphanFiles(Table table, ConnectorSession session, SchemaTabl
 
         validFileNames.add("version-hint.text");
 
-        scanAndDeleteInvalidFiles(table, session, schemaTableName, expiration, validFileNames.build(), fileIoProperties);
+        try {
```
grantatspothero, line 2333: small improvement, use util which wraps executor completion service: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/util/Executors.java#L41-L62 

Benefit is the executor completion service will fail the remaining futures immediately if a single future fails, while this code waits on the futures in order. Gives quicker failure feedback (and wastes less resources)

raunaqmorarka: I would have liked to use that, the problem is it forces serial execution of reading manifest lists upfront. In the current code we get to interleave reading from manifest lists on the main thread with reading the manifest files in the threadpool.

PR #26452, opened 2025-08-20
Add support for INTERVAL types in Exasol connector

skyglass:
## Description

Added Exasol Trino connector support for **INTERVAL YEAR TO MONTH** and **INTERVAL DAY TO SECOND** JDBC data types

- added correspondent tests for **interval year(y) to month** with precision data type
- added correspondent tests for **interval day(d) to second(s)** with precision data type

## Release notes

```markdown
## Exasol Connect
* Add support for `interval year to month` and `interval day to second` types. 
```

### DIFF
--- a/docs/src/main/sphinx/connector/exasol.md
+++ b/docs/src/main/sphinx/connector/exasol.md
@@ -103,6 +103,12 @@ Trino data type mapping:
 * - `HASHTYPE`
   - `VARBINARY`
   -
+* - `INTERVAL YEAR(y) TO MONTH`
+  - `BIGINT`
+  -
+* - `INTERVAL DAY(d) TO SECOND(s)`
+  - `BIGINT`
+  -
 :::
 
 No other types are supported.
--- a/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
+++ b/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
@@ -17,6 +17,8 @@ import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
 import io.airlift.slice.Slices;
 import io.trino.plugin.base.mapping.IdentifierMapping;
+import io.trino.plugin.exasol.type.interval.ExasolIntervalDayTimeParser;
+import io.trino.plugin.exasol.type.interval.ExasolIntervalYearMonthParser;
 import io.trino.plugin.jdbc.BaseJdbcClient;
 import io.trino.plugin.jdbc.BaseJdbcConfig;
 import io.trino.plugin.jdbc.ColumnMapping;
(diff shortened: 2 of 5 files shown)


plugin/trino-exasol/pom.xml
```diff
@@ -45,11 +45,21 @@
             <artifactId>trino-base-jdbc</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-client</artifactId>
+        </dependency>
```
ebyhr: No need to depend on trino-client in my opinion. See PostgreSqlClient implementation.

skyglass: The problem is that I need IntervalDayTime and IntervalYearMonth classes from the package trino-client. 
I could also implement the logic myself, but it would lead to the duplication of the code.
Maybe we should move IntervalDayTime and IntervalYearMonth to package "io.trino.spi.type" of "trino-spi" module. These classes encapsulate the logic for the correspondent types, so I think they should belong there.
I can do this refactoring first, in the separate PR.

ebyhr: Is there alternative way to get those values with Exasol JDBC driver? e.g. ResultSet.getInt or custom type
I'm not a fan of the current string parsing logic. The format might be different between Trino and Exasol.

skyglass: > Is there alternative way to get those values with Exasol JDBC driver?

Unfortunately, exasol jdbc driver can't help here. It throws exception, if you try to get int or long. We have to implement parsing logic by ourselves. Luckily, it is more or less standard format, so the one that trino uses in trino-client is reliable and could even serve as a prototype for new Interval trino types, which don't exist yet (but could exist in many other databases, so they would just need to use new trino interval types, without reimplementing the parsing logic)

> I'm not a fan of the current string parsing logic. The format might be different between Trino and Exasol.

Yes, there are edge-cases, for which the format might be a little different, but there are only few of them and they all covered by tests. For example, trino doesn't recognize. "+" at the beginning for positive intervals, so "+" must be removed. Also, the number of fractional digits must always be 3, so in case there are less than 3 fractional digits, then we must add one or two "0" digits at the end of the string. In case the number of fractional digits is more than 3, then trino rounds up to the nearest millisecond. In order to implement this adapter between exasol and trino, I implemented normalizeIntervalYearMonthString and normalizeIntervalDaySecondString methods. Combination of these methods with native trino IntervalDayMonth and IntervalDayTime classes allows to safely convert between exasol interval string and trino bigint and vice versa. The correspondence of exasol interval string to bigint is one to one, which prevents any other edge cases. In case, trino will create new interval types it would be easy to switch to these types instead of using bigint (number of months for interval day to month and number of milliseconds for interval day to second). In any case, using current bigint type is ok for now as it provides all required capabilities of range indexes. Using raw trino varchar type might not be very efficient. For example, the sorting of intervals might be broken, if we use range indexes by raw string.

skyglass: I have decided to remove **trino-client** dependency and implement the interval parsing logic myself by adding new classes **ExasolIntervalDayTimeParser** and **ExasolIntervalYearMonthParser**.

This solution will not require **trino-client** dependency or refactoring **trino-client** and **trino-spi** modules, so I decided to go with it.

Also I moved string normalization to these new classes, so all parsing and formatting logic for exasol interval types is encapsulated in these classes.

skyglass: Any updates on the review?

skyglass:
Thank you very much for your review, @ebyhr ! 👍 

Any updates on the PR?
Please, let me know, if anything is missing or required.

findepi:
FYI, there are some upcoming changes to Interval handling in Trino
- https://github.com/trinodb/trino/pull/27539

PR #26459, opened 2025-08-21
Perform disjoint/contains operation on Polygon instead of Envelope

Praveen2112:
## Description

This would fix issue related to inaccuracies when converting from Geometry to BingTile.

Benchmark also shows an improvement in average time per operation

Before:

Benchmark                                         Mode  Cnt    Score    Error  Units
BenchmarkGeometryToBingTiles.envelopeToBingTiles  avgt   10    0.007 ±  0.001  ms/op
BenchmarkGeometryToBingTiles.geometryToBingTiles  avgt   10  172.830 ±  5.088  ms/op

After:

Benchmark                                         Mode  Cnt    Score    Error  Units
BenchmarkGeometryToBingTiles.envelopeToBingTiles  avgt   10    0.007 ±  0.001  ms/op
BenchmarkGeometryToBingTiles.geometryToBingTiles  avgt   10  104.172 ±  0.749  ms/op

## Release notes

```markdown
## GeoFunctions
*Fixed an issue where geometry_to_bing_tiles did not cover the full geometry area, ensuring complete tile coverage.  ({issue}`issuenumber`)
```


wendigo (approved):

plugin/trino-geospatial/src/test/resources/large_polygon.txt
martint, line 1: Where is this used?

Praveen2112: This is used in `testLargeGeometryToBingTiles` where we iterate over these entries for validation. We have also added an additional validation to ensure that the polygon created from tile covers the input polygon.

raunaqmorarka (approved):

PR #26460, opened 2025-08-21
Move interval classes from trino-client to trino-spi

skyglass:
## Description

Moved **trino-client** classes  **IntervalYearMonth** and **IntervalDatyTime** to **trino-spi** module (package **io.trino.type.interval**)

- this refactoring is required for **trino-exasol** module, which doesn't see **trino-client** module but requires access to its IntervalYearMonth and IntervalDaySecond classes to implement INTERVAL type
- adding **trino-client** dependency to **trino-exasol** will increase coupling and therefore is not desirable architectural decision (in addition to other challenges, like cascading scope changes of related modules)
- see related PR for more details: https://github.com/trinodb/trino/pull/26452

## Release notes

```markdown
## Trino Client Refactoring
* Move IntervalYearMonth and IntervalDayTime from trino-client to trino-spi
```


skyglass:
Closed, as it might break dependencies

Issue #11838, opened 2022-04-07
No iceberg table column comment


cutiechi:
There are no comments for iceberg table fields in the Hive Metastore

findepi:
IMO there is no need to write any table schema to metastore.
The table schema, including comments, is within table metadata in table directory. This is the source of truth.

cutiechi:
> IMO there is no need to write any table schema to metastore. The table schema, including comments, is within table metadata in table directory. This is the source of truth.

But i want to list cols without hdfs or s3

MichaelTiemannOSC:
Suggesting @erikerlandson watch this.

PPL-TBSK:
best way to list all comments for all tables? is the best way to create custom script for that?
There is the iceberg.information_schema.columns table and I would love to get comment information to all of the columns in that table.

Issue #11161, opened 2022-02-23
Do not update the metastore when setting Iceberg table/column comment


findinpath:
https://github.com/trinodb/trino/pull/11143#discussion_r812786798

PR #26467, opened 2025-08-22
Add column comments on Iceberg Hive catalog

jinyangli34:
## Release notes

```markdown
## Section
```


wendigo (approved):

ebyhr:
Let me close this PR because it's not stored intentionally. See https://github.com/trinodb/trino/issues/11838 & https://github.com/trinodb/trino/issues/11161 for more details.

jinyangli34:
> I don't think we want to support this. What is the motivation of this change?

When listing/filtering tables on Hive Metastore, we can't use the `comment` information for tables created by Trino. while tables created by Iceberg library/Spark have the column comments in HMS.


> The metastore has the max length limitation (e.g. 256 in mariadb) for column comments. This change silently stores truncated column comments.

Do you think we can have a cluster config / table property to control the expose extra information on HMS? 
Besides comment, there are other fields like `default-partition-spec`, `current-schema` missing on Trino Iceberg tables.

PR #26469, opened 2025-08-22
Attach authenticateIdentity groups to canImpersonate check

mwd410:
## Description

Pretty self-explanatory - for AccessControls that have Role -> Group assignments, not including groups on this impersonation check can fail unexpectedly.

This test feels a little naive, though i am not sure how else to test it. Hopefully someone can steer me in the right direction there.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
+++ b/core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
@@ -218,6 +218,7 @@ public class HttpRequestSessionContextFactory
                 // load enabled roles for authenticated identity, so impersonation permissions can be assigned to roles
                 authenticatedIdentity = Identity.from(authenticatedIdentity)
                         .withEnabledRoles(metadata.listEnabledRoles(authenticatedIdentity))
+                        .withGroups(groupProvider.getGroups(authenticatedIdentity.getUser()))
                         .build();
                 accessControl.checkCanImpersonateUser(authenticatedIdentity, originalIdentity.getUser());
             }
--- a/core/trino-main/src/test/java/io/trino/server/TestHttpRequestSessionContextFactory.java
+++ b/core/trino-main/src/test/java/io/trino/server/TestHttpRequestSessionContextFactory.java
@@ -16,19 +16,27 @@ package io.trino.server;
 import com.google.common.collect.ImmutableListMultimap;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import io.airlift.configuration.secrets.SecretsResolver;
 import io.airlift.jaxrs.testing.GuavaMultivaluedMap;
 import io.trino.client.ProtocolHeaders;
+import io.trino.security.AccessControl;
 import io.trino.security.AllowAllAccessControl;
 import io.trino.server.protocol.PreparedStatementEncoder;
 import io.trino.server.protocol.spooling.QueryDataEncoder;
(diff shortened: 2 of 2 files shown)


core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
```diff
@@ -218,6 +218,7 @@ public Identity extractAuthorizedIdentity(Optional<Identity> optionalAuthenticat
                 // load enabled roles for authenticated identity, so impersonation permissions can be assigned to roles
                 authenticatedIdentity = Identity.from(authenticatedIdentity)
```
kokosing, line 219: I was under impression that this method copies groups.

ksobolew: I don't think `authenticatedIdentity` has any groups. Authenticators should not deal with groups.

mwd410: It also copies roles, but we need to add them anyway

kokosing:
I think the test is not very important here. First of all we need to be sure that we are fixing the issue here. @ksobolew are you sure about that?

ksobolew:
Well, I haven't tested this manually - I assume @mwd410 did :) But it looks legit.

core/trino-main/src/test/java/io/trino/server/TestHttpRequestSessionContextFactory.java
```diff
@@ -189,6 +203,53 @@ public void testInternalExtraCredentialName()
                 .hasMessage("Invalid extra credential name: internal$abc");
     }
 
+    @Test
+    public void testImpersonation()
```
chenjian2664, line 207: package private https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations

core/trino-main/src/test/java/io/trino/server/TestHttpRequestSessionContextFactory.java
```diff
@@ -198,12 +259,19 @@ public void testInternalExtraCredentialName()
     }
 
     private static HttpRequestSessionContextFactory sessionContextFactory(ProtocolHeaders headers)
+    {
+        return sessionContextFactory(headers, ImmutableSet::of, new AllowAllAccessControl());
+    }
+
+    private static HttpRequestSessionContextFactory sessionContextFactory(ProtocolHeaders headers,
```
chenjian2664, line 266: Start a new line for the parameters

core/trino-main/src/test/java/io/trino/server/TestHttpRequestSessionContextFactory.java
```diff
@@ -189,6 +203,53 @@ public void testInternalExtraCredentialName()
...
+        TestingAccessControlManager accessControl = new TestingAccessControlManager(
+                createTestTransactionManager(),
+                emptyEventListenerManager(),
+                new SecretsResolver(ImmutableMap.of()));
+
+        accessControl.loadSystemAccessControl("allow-all", ImmutableMap.of());
+
+        HttpRequestSessionContextFactory sessionContextFactory = sessionContextFactory(TRINO_HEADERS, groupProvider, accessControl);
+
+        HashSet<String> impersonatedUsers = new HashSet<>();
```
chenjian2664, line 223: Use `Set` or consider use `ImmutableSet.Builder`

PR #26470, opened 2025-08-23
Update release artifact download urls

martint:
Going forward, binary artifacts will be available under GitHub releases
## Release notes

### DIFF
--- a/docs/src/main/sphinx/ext/download.py
+++ b/docs/src/main/sphinx/ext/download.py
@@ -17,80 +17,78 @@ from docutils import nodes, utils
 # noinspection PyUnresolvedReferences
 from sphinx.errors import SphinxError
 
-GROUP_ID = 'io.trino'
 ARTIFACTS = {
-    'server': ('trino-server', 'tar.gz', None),
-    'server-core': ('trino-server-core', 'tar.gz', None),
-    'cli': ('trino-cli', 'jar', 'executable'),
-    'jdbc': ('trino-jdbc', 'jar', None),
+    'server': ('trino-server', 'tar.gz'),
+    'server-core': ('trino-server-core', 'tar.gz'),
+    'cli': ('trino-cli', None),
+    'jdbc': ('trino-jdbc', 'jar'),
     #plugins
-    'ai-functions':  ('trino-ai-functions', 'zip', None),
-    'bigquery':  ('trino-bigquery', 'zip', None),
-    'blackhole':  ('trino-blackhole', 'zip', None),
-    'cassandra':  ('trino-cassandra', 'zip', None),
-    'clickhouse':  ('trino-clickhouse', 'zip', None),
-    'delta-lake':  ('trino-delta-lake', 'zip', None),
-    'druid':  ('trino-druid', 'zip', None),
-    'duckdb':  ('trino-duckdb', 'zip', None),
-    'elasticsearch':  ('trino-elasticsearch', 'zip', None),
-    'example-http':  ('trino-example-http', 'zip', None),
-    'exasol':  ('trino-exasol', 'zip', None),
-    'exchange-filesystem':  ('trino-exchange-filesystem', 'zip', None),
-    'exchange-hdfs':  ('trino-exchange-hdfs', 'zip', None),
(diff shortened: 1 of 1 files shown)


docs/src/main/sphinx/ext/download.py
```diff
@@ -17,80 +17,78 @@
...
 
-GROUP_ID = 'io.trino'
 ARTIFACTS = {
-    'server': ('trino-server', 'tar.gz', None),
-    'server-core': ('trino-server-core', 'tar.gz', None),
-    'cli': ('trino-cli', 'jar', 'executable'),
-    'jdbc': ('trino-jdbc', 'jar', None),
+    'server': ('trino-server', 'tar.gz'),
+    'server-core': ('trino-server-core', 'tar.gz'),
+    'cli': ('trino-cli', None),
```
mosabua, line 23: Is this going to work? The filename should be like https://repo1.maven.org/maven2/io/trino/trino-cli/476/trino-cli-476-executable.jar

mosabua (changes requested):

mosabua:
Please confirm about expected filenames and then also review and comment in https://github.com/trinodb/trino.io/pull/811 since that assumes identical filenames to what we had in Central

docs/src/main/sphinx/ext/download.py
martint: No, it can be anything we want. We’ll publish those artifacts to github, so we’ll change the name. No need for the “-executable” qualifier

mosabua: Ok .. so the filename will be just trino-cli-476.jar ? If so I have to adjust my PR for the website .. which is fine by my .. just might be confusing for readers of the book or older docs for what they are used to.  Just let me know so I can adjust the other PR

mosabua (approved):
I am good with either approach for the CLI filename .. the rest if fine anyway.. so I am approving.

docs/src/main/sphinx/ext/download.py
martint: No, it will be just “trino-cli-476”. It’s just a binary executable

mosabua: Uff .. I think thats actually a bad idea. It is not a pure executable .. it still requires a JVM on the path .. and with the .jar extension it can be recognized and run .. without the extension that is harder to figure out and might not work so well.

mosabua: Kind of feels like a unrelated change that we should not pull in unless we have to.

mosabua: Wdyt @wendigo @electrum ? I know the file works with the name without extension if Java is on the path, but if not it might be a bad experience. Have any of you played with that in the past?

martint: The extension doesn’t actually do anything. The file behaves like a script with a shebang line.

wendigo: It helps if you are running on Windows (I'm not sure whether we support it or not). We could add a check in the cli itself, to check for required flags if it's run via `java -jar`).

mosabua: I guess either way should be fine. The instructions in the docs already talk about renaming it so it doesnt matter I guess - see https://trino.io/docs/current/client/cli.html

We just have to adjust the example for Windows in that same doc since it uses the old -executable.jar

docs/src/main/sphinx/ext/download.py
```diff
@@ -17,80 +17,78 @@
...
+    return artifact + '-' + version + extension
 
 
-def maven_download(group, artifact, version, packaging, classifier):
-    base = 'https://repo1.maven.org/maven2/'
-    group_path = group.replace('.', '/')
-    filename = maven_filename(artifact, version, packaging, classifier)
-    return base + '/'.join((group_path, artifact, version, filename))
+def download_url(artifact, version, extension):
+    base = 'https://github.com/trinodb/trino/releases/download/'
```
mosabua:
```suggestion
    base = 'https://github.com/trinodb/trino/releases/download'
```

to avoid double /

PR #26478, opened 2025-08-25
Reuse IdentityCacheMapping for BigQuery

ssheikin:
## Release notes

```markdown
## Section
```


lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/authentication/cache/IdentityCacheMapping.java
```diff
@@ -11,7 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.trino.plugin.jdbc;
+package io.trino.plugin.base.authentication.cache;
```
Praveen2112: Why do we need this under `authentication` ?

kokosing: Identity is part of authentication layer

kokosing (approved):

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/authentication/cache/IdentityCacheMapping.java
Praveen2112: Yes - But this cache mapping is more of a derivative based on `Identity` right ?

chenjian2664 (approved):
the description seems broken

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/authentication/cache/IdentityCacheMapping.java
ssheikin: @Praveen2112 repackaged to `cache.identity` and dropped unrelated kerberos repackage. PTAL.

Praveen2112 (approved):

PR #26483, opened 2025-08-25
fixed typo in documentation for kafka-event-listener example "kafka.config.resources"

maxgruber19:
## Description
There is a little typo in the kafka-event-listener configuration example which leads to errors when copy-pasting the config into trino deployments leading to trino failing with following exception

```
Configuration property 'kafka.config.resources' was not used
```

Got my trino up and running and producing to kafka after setting it to the property key listed in the documentation below

## Release notes


maxgruber19:
Just kicked off a CLA :-)

PR #26490, opened 2025-08-26
Replace createTmestampType to createTimestampWithTimeZoneType in TestLongTimestampWithTimeZoneType

### LINKED ISSUE #6448: Document type conversions supported by and type inference of PreparedStatement#setObject
`java.sql.PreparedStatement#setObject` has two variants:

- plan `setObject` 
  - -> document what is supported here & how the type is inferred
- `setObject` with type id
  - -> document what is supported here

karolhor:
## Description
Fixes #6448

## Release notes


core/trino-main/src/test/java/io/trino/type/TestLongTimestampWithTimeZoneType.java
```diff
@@ -114,58 +114,78 @@ public void testNextValue()
...
-                Arguments.of(6, Long.MIN_VALUE, Long.MAX_VALUE, 1L));
+                Arguments.of(4, Long.MIN_VALUE, Long.MAX_VALUE, 100_000_000),
+                Arguments.of(5, Long.MIN_VALUE, Long.MAX_VALUE, 10_000_000),
+                Arguments.of(6, Long.MIN_VALUE, Long.MAX_VALUE, 1_000_000),
+                Arguments.of(7, Long.MIN_VALUE, Long.MAX_VALUE, 100_000),
+                Arguments.of(8, Long.MIN_VALUE, Long.MAX_VALUE, 10_000),
+                Arguments.of(9, Long.MIN_VALUE, Long.MAX_VALUE, 1_000),
+                Arguments.of(10, Long.MIN_VALUE, Long.MAX_VALUE, 100),
+                Arguments.of(11, Long.MIN_VALUE, Long.MAX_VALUE, 10),
+                Arguments.of(12, Long.MIN_VALUE, Long.MAX_VALUE, 1));
```
ebyhr: The parameter name (maxValue) and the actual value don't match.

Please change `long` to `LongTimestampWithTimeZone` here instead of test methods.

core/trino-main/src/test/java/io/trino/type/TestLongTimestampWithTimeZoneType.java
```diff
@@ -114,58 +114,78 @@ public void testNextValue()
 
     @ParameterizedTest
     @MethodSource("testPreviousNextValueEveryPrecisionDataProvider")
-    public void testPreviousValueEveryPrecision(int precision, long minValue, long maxValue, long step)
+    public void testPreviousValueEveryPrecision(int precision, long minValue, long maxValue, int step)
```
chenjian2664: change to `int` seems not necessary here

core/trino-main/src/test/java/io/trino/type/TestLongTimestampWithTimeZoneType.java
```diff
@@ -114,58 +114,78 @@ public void testNextValue()
...
     {
-        Type type = createTimestampType(precision);
-
-        assertThat(type.getPreviousValue(minValue))
+        long middleRangeEpochMillis = 123_456_789_000_000L;
+        int oneMsInPicos = 1_000_000_000;
```
chenjian2664: rename to `picosOfMilli`

core/trino-main/src/test/java/io/trino/type/TestLongTimestampWithTimeZoneType.java
ebyhr: This change makes sense because the 2nd parameter of `fromEpochMillisAndFraction` is `int`.

core/trino-main/src/test/java/io/trino/type/TestLongTimestampWithTimeZoneType.java
karolhor: Thanks. I modified the MethodSource: testPreviousNextValueEveryPrecisionDataProvider

core/trino-main/src/test/java/io/trino/type/TestLongTimestampWithTimeZoneType.java
```diff
@@ -57,6 +57,8 @@ public static ValueBlock createTestBlock()
         return blockBuilder.buildValueBlock();
     }
 
+    private static final int PICOS_OF_MILLI = 1_000_000_000;
```
chenjian2664: nit: there exists the constant in `Timestamps#PICOSECONDS_PER_MILLISECOND`, we could static import here

karolhor: I didn't know there is already a const for that. thanks

karolhor:
@ebyhr @chenjian2664 is there anything more I can improve in this ticket or it's ok now?

karolhor:
Hi, is there something I can do with this PR?

PR #26507, opened 2025-08-28
Set Calendar type used for temporal types in ORC footer

marcinsbd:
## Release notes

```markdown
## Section
```


lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +88,45 @@ public void testWriteHugeChunk()
         testWriteOutput(columnNameBuilder.build(), data);
     }
 
+    @Test
+    public void testWritingFooter()
```
Praveen2112: nit: `testFooter`

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +88,45 @@ public void testWriteHugeChunk()
...
+        testWritingFooter(DATE, dates, CalendarKind.PROLEPTIC_GREGORIAN);
+
+        List<SqlTimestamp> timestamps = ImmutableList.of("2023-04-11T05:16:12.123", "2021-04-11T05:16:12.123", "1999-04-11T05:16:12.123").stream()
+                .map(LocalDateTime::parse)
+                .map(ldt -> SqlTimestamp.fromSeconds(TIMESTAMP_MILLIS.getPrecision(), ldt.toEpochSecond(UTC), ldt.getNano()))
+                .collect(toImmutableList());
+        testWritingFooter(TIMESTAMP_MILLIS, timestamps, CalendarKind.PROLEPTIC_GREGORIAN);
+    }
+
+    private static void testWritingFooter(Type type, List<?> values, CalendarKind expectedCalendarKind)
```
Praveen2112: nit: assertFooter ?

lib/trino-orc/src/main/java/io/trino/orc/metadata/Footer.java
```diff
@@ -93,6 +96,11 @@ public Optional<Integer> getWriterId()
         return writerId;
     }
 
+    public Optional<CalendarKind> getCalendar()
```
martint: What's the purpose of this? It seems unused in Trino. Since Trino only uses prolectic gregorian, it should be hardcoded to write that.

Separately, since Trino only supports prolectic gregorian, it should throw an error or do the appropriate conversions when it encounters a different calendar while reading a file.

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +88,45 @@ public void testWriteHugeChunk()
...
+    public void testWritingFooter()
+    {
+        List<String> strings = ImmutableList.of("aaa1", "qwerty", "asdf", "zxcvb", "1234");
+        testWritingFooter(VARCHAR, strings, CalendarKind.UNKNOWN_CALENDAR);
+
+        List<SqlDate> dates = ImmutableList.of("2020-01-01", "2021-02-02", "2022-03-03", "2023-04-04", "2024-05-05").stream()
+                .map(LocalDate::parse)
+                .map(LocalDate::toEpochDay)
+                .map(Math::toIntExact)
+                .map(SqlDate::new)
```
ebyhr: We could simplify like this:
```java
.map(text -> sqlDateOf(LocalDate.parse(text)))
```

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +88,45 @@ public void testWriteHugeChunk()
...
+                .collect(toImmutableList());
+        testWritingFooter(DATE, dates, CalendarKind.PROLEPTIC_GREGORIAN);
+
+        List<SqlTimestamp> timestamps = ImmutableList.of("2023-04-11T05:16:12.123", "2021-04-11T05:16:12.123", "1999-04-11T05:16:12.123").stream()
+                .map(LocalDateTime::parse)
+                .map(ldt -> SqlTimestamp.fromSeconds(TIMESTAMP_MILLIS.getPrecision(), ldt.toEpochSecond(UTC), ldt.getNano()))
```
ebyhr: We could simplify like this:
```java
.map(text -> sqlTimestampOf(TIMESTAMP_MILLIS.getPrecision(), LocalDateTime.parse(text)))
```

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +88,45 @@ public void testWriteHugeChunk()
...
+        try (TempFile tempFile = new TempFile()) {
+            OrcTester.writeOrcColumnTrino(tempFile.getFile(), CompressionKind.NONE, type, values.iterator(), new OrcWriterStats());
+
+            OrcDataSource orcDataSource = new FileOrcDataSource(tempFile.getFile(), READER_OPTIONS);
+            Footer footer = OrcReader.createOrcReader(orcDataSource, READER_OPTIONS)
+                    .orElseThrow(() -> new RuntimeException("File is empty"))
+                    .getFooter();
+
+            assertThat(footer.getCalendar().isPresent()).isTrue();
+            assertThat(footer.getCalendar().get()).isEqualTo(expectedCalendarKind);
```
ebyhr: AssertJ provides `contains()` method for optional types:
```java
assertThat(footer.getCalendar()).contains(expectedCalendarKind);
```

lib/trino-orc/src/main/java/io/trino/orc/metadata/Footer.java
Praveen2112: > What's the purpose of this? 

This is to ensure that other engines like Hive can be forced to read the data in proleptic gregorian. 

As a follow up we are thinking on do a conversion when we encounter a different calendar when reading a file.

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
marcinsbd: thanks,  TIL DateTimeTestingUtils

lib/trino-orc/src/main/java/io/trino/orc/metadata/Footer.java
marcinsbd: The Hive will treat temporal data in Proleptic Gregorian calendar if there is an explicit entry in metadata. If there is no entry the Hive read the temporal as in case for Hybrid (Julian-Gregorian) calendar.

lib/trino-orc/src/main/java/io/trino/orc/OrcWriter.java
```diff
@@ -529,7 +531,8 @@ private List<OrcDataOutput> bufferFileFooter()
                 orcTypes,
                 fileStats,
                 userMetadata,
-                Optional.empty()); // writer id will be set by MetadataWriter
+                Optional.empty(), // writer id will be set by MetadataWriter
+                containsTemporalType(orcTypes) ? Optional.of(PROLEPTIC_GREGORIAN) : Optional.empty());
```
Praveen2112: Any specific reason on why we write the calendar only if temporal types are available ? I guess for files written by Trino it should always be `PROLEPTIC_GREGORIAN`

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +89,40 @@ public void testWriteHugeChunk()
         testWriteOutput(columnNameBuilder.build(), data);
     }
 
+    @Test
+    public void testFooter()
+    {
+        List<String> strings = ImmutableList.of("aaa1", "qwerty", "asdf", "zxcvb", "1234");
+        assertFooter(VARCHAR, strings, CalendarKind.UNKNOWN_CALENDAR);
```
Praveen2112: I think it should set always

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +89,40 @@ public void testWriteHugeChunk()
...
+    }
+
+    private static void assertFooter(Type type, List<?> values, CalendarKind expectedCalendarKind)
+    {
+        try (TempFile tempFile = new TempFile()) {
+            OrcTester.writeOrcColumnTrino(tempFile.getFile(), CompressionKind.NONE, type, values.iterator(), new OrcWriterStats());
+
+            OrcDataSource orcDataSource = new FileOrcDataSource(tempFile.getFile(), READER_OPTIONS);
+            Footer footer = OrcReader.createOrcReader(orcDataSource, READER_OPTIONS)
+                    .orElseThrow(() -> new RuntimeException("File is empty"))
```
Praveen2112: nit: we could apply an assertion on `Optional<Footer>` directly right ?

lib/trino-orc/src/main/java/io/trino/orc/metadata/Footer.java
Praveen2112: @martint Making it `Optional` is to ensure that older version of ORC files so as to be backward compatible.

lib/trino-orc/src/main/java/io/trino/orc/OrcWriter.java
```diff
@@ -544,6 +547,11 @@ private List<OrcDataOutput> bufferFileFooter()
         return outputData;
     }
 
+    private boolean containsTemporalType(ColumnMetadata<OrcType> orcTypes)
+    {
+        return orcTypes.stream().map(OrcType::getOrcTypeKind).anyMatch(TEMPORAL_TYPES::contains);
+    }
```
ebyhr: This method seems to return false incorrectly if a temporal type exists in nested fields. However, we might be able to remove it as Praveen mentioned above.

lib/trino-orc/src/main/java/io/trino/orc/OrcWriter.java
marcinsbd: no reason, changed

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
marcinsbd: done

lib/trino-orc/src/main/java/io/trino/orc/metadata/CalendarKind.java
```diff
@@ -0,0 +1,21 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.orc.metadata;
+
+public enum CalendarKind
```
ebyhr, line 16: Why do we add our own enum instead of reusing `org.apache.orc.OrcProto.CalendarKind`?

marcinsbd: 1. We have already created object Footer which is a double for OrcProto.Footer , I suspect the reason was not to force the other modules like trino-iceberg to be dependent on  org.apache.orc:orc-format when uses trino-orc. Just to encapsulate all orc's releated dependecies within/behind trino-orc.
2. trino-iceberg uses Footer, without our doubled enum in trino-orc , the trino-iceberg needs to add the org.apache.orc:orc-format to its dependency

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +89,38 @@ public void testWriteHugeChunk()
         testWriteOutput(columnNameBuilder.build(), data);
     }
 
+    @Test
+    public void testFooter()
```
chenjian2664: upgrade name to "testFooterCalendarTypeForTemporalTypes" or something similar

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
```diff
@@ -150,7 +153,8 @@ public Footer readFooter(HiveWriterVersion hiveWriterVersion, InputStream inputS
                 toType(footer.getTypesList()),
                 toColumnStatistics(hiveWriterVersion, footer.getStatisticsList(), false),
                 toUserMetadata(footer.getMetadataList()),
-                Optional.of(footer.getWriter()));
+                Optional.of(footer.getWriter()),
+                Optional.of(toTrinoOrcCalendarKind(footer.getCalendar())));
```
chenjian2664: Should we check if it is `PROLEPTIC_GREGORIAN` when reading?
we can throw exception for `JULIAN_GREGORIAN` since we don't support, and what if it is `UNKNOWN_CALENDAR`?

lib/trino-orc/src/main/java/io/trino/orc/metadata/Footer.java
martint: If `Optional.empty()` means Julian calendar, then the reader should return that instead. I don't see the benefit of signaling that the calendar type is missing.

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
martint: Yes, until such a time when we support Julian, we should throw an error -- otherwise, the reader will produce incorrect values.

lib/trino-orc/src/main/java/io/trino/orc/metadata/Footer.java
marcinsbd: This  ( `Optional.empty()` means  Hybrid (Julian-Gregorian) calendar is a Hive logic.
However,  imho Trino should treat empty or Unknown type as Proleptic Gregorian Type  in order to be backward compatible ( as Trino were writing Orc files without specified Calendar type before).
We could drop Optional and treat empty as Unknown Type.
In this way we will have:
Hive: 
unknown type, hybird(julian_gregorian) type -> read hybird(julian_gregorian)
proleptic_gregorian type -> read proleptic_gregorian

Trino:
unknown type, proleptic_gregorian type  ->  read proleptic_gregorian

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
marcinsbd: Just to make sure we are on the same page, we are talking about the dates before _Sept 16th, 1582_. 
Hybrid (Julian-Gregorian) keeps in Julian dates up to Sept 15th, 1582. The later dates are in Gregorian so even the calendar type is Julian-Gregorian, the later dates will be read correctly by Trino. 
The question is if:
1. should we throw exception just by judging by calendar type from the ORC's footer (in case  Hybrid -JulianGregorian)
2. should we throw exception when  the calendar type set on Hybrid(JulianGregorian) and the file contains dates before Sep 16th 1582  - when we approach the legacy dates during reading (not sure if feasible to implement) 
3. leave it as it's working now, so don't not throw exception at all  but only add to the doc a limitation warning about the fact that we don't support dates before Sept 16th  1582 when Julian Calendar specified, and they will be read incorrectly

Personally, I'm for 3.
The aim of the PR was only add info about the calendar at the write part and leave the read part as it was.

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
marcinsbd: since it's added always, i changed to `testCalendarEntryInFooter`

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
marcinsbd: @martint @Praveen2112 @ebyhr wdyt ?

chenjian2664 (approved):
Everything looks good. The only point left is the decision around reading, which I expect should not significantly impact the PR’s structure

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
raunaqmorarka: I'm strongly against 1. Changing the behaviour of the reader to now starting throwing exceptions for all dates if we don't recognize calendar type could be highly disruptive to existing workloads.
2 would be ideal, but I don't recommend spending time on that. Over time workloads are moving away from Hive and into iceberg/delta, so we shouldn't spend too much time agonizing over adjusting to Hive intricacies.
You can file a separate issue about the reader part, I don't see why that should block the writer part in this PR.

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
```diff
@@ -78,6 +89,38 @@ public void testWriteHugeChunk()
         testWriteOutput(columnNameBuilder.build(), data);
     }
 
+    @Test
+    public void testCalendarEntryInFooter()
```
raunaqmorarka, line 93: Is it feasible to provide a product test to verify that Hive reads the values as expected ?

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
martint: I see. I was under the impression that it was for all dates, not just those before 1582. In that case, I agree with @raunaqmorarka. Option 2 would be ideal, but not necessary for this PR -- let's file an issue about that and mark it with the "correctness" label.

Let's go with 3 and add a note to the documentation indicating that dates before 1582 will produce incorrect results with that calendar.

chenjian2664: @marcinsbd Is it possible to have a test to verify the case 3, the calendar type is hybird but we can read correctly read the time after Sept 15th, 1582

marcinsbd:
I'm planning to continue work on the issue this week.

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
marcinsbd: As we agreed:
- issue https://github.com/trinodb/trino/issues/26865
- test added

marcinsbd: @mosabua 
Which place in the doc could be the best place to put info about the limitation of ORC reader for legacy dates and timestamp when Calendar in the Orc file is set to Hybrid. or maybe the issue is just enough  (https://github.com/trinodb/trino/issues/26865)

lib/trino-orc/src/test/java/io/trino/orc/TestOrcWriter.java
marcinsbd: done

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
mosabua: I think probably in https://trino.io/docs/current/object-storage/file-formats.html#orc-format-configuration-properties somehow

marcinsbd: @mosabua https://github.com/trinodb/trino/pull/26874

marcinsbd: @martint @raunaqmorarka @chenjian2664 @ebyhr @Praveen2112 @mosabua 
Thanks for the review.
As we agreed:
- tests added
- issue submitted https://github.com/trinodb/trino/issues/26865
-  doc entry https://github.com/trinodb/trino/pull/26874

Is there anything else from your side?

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveOnOrcLegacyDateCompatibility.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test(groups = {HIVE4, PROFILE_SPECIFIC_TESTS})
+    public void testReadLegacyDateFromOrcWrittenByTrino()
+    {
+        String hiveTableName = "test_hive_orc_legacy_date_compatibility_%s".formatted(randomNameSuffix());
+        String trinoTableName = format("%s.%s.%s", TRINO_CATALOG, SCHEMA, hiveTableName);
+
+        try {
+            onTrino().executeQuery("CREATE TABLE %s (date_col date, t timestamp) WITH (format = 'ORC')".formatted(trinoTableName));
+            onTrino().executeQuery(("INSERT INTO %s VALUES " +
```
chenjian2664: nit: Remove redundant `(` and `)`, we could use string concat since it just the one param

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveOnOrcLegacyDateCompatibility.java
```diff
@@ -0,0 +1,103 @@
...
+
+    @Test(groups = {HIVE4, PROFILE_SPECIFIC_TESTS})
+    public void testReadLegacyDateFromOrcWrittenByHive()
+    {
+        String hiveTableName = "test_hive_orc_legacy_date_compatibility_%s".formatted(randomNameSuffix());
+        String trinoTableName = format("%s.%s.%s", TRINO_CATALOG, SCHEMA, hiveTableName);
+
+        try {
+            onHive().executeQuery("CREATE TABLE %s (date_col date, t timestamp) STORED AS ORC".formatted(hiveTableName));
+            onHive().executeQuery(("INSERT INTO %s VALUES " +
```
chenjian2664: Same above

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveOnOrcLegacyDateCompatibility.java
```diff
@@ -0,0 +1,103 @@
...
+            QueryAssert.Row[] expectedRowsHive = {
+                    row(Date.valueOf("0002-01-01"), Timestamp.valueOf("0002-01-01 00:00:00.123")),
+                    row(Date.valueOf("1500-01-01"), Timestamp.valueOf("1500-01-01 00:00:00.123")),
+                    row(Date.valueOf("1582-10-04"), Timestamp.valueOf("1582-10-04 00:00:00.123")),
+                    row(Date.valueOf("1582-11-04"), Timestamp.valueOf("1582-11-04 00:00:00.123")),
+                    row(Date.valueOf("2000-02-29"), Timestamp.valueOf("2000-02-29 00:00:00.123"))};
+            assertThat(onHive().executeQuery("SELECT date_col, t FROM " + hiveTableName)).containsOnly(expectedRowsHive);
+
+            QueryAssert.Row[] expectedRowsTrino = {
+                    row(Date.valueOf("0001-12-30"), Timestamp.valueOf("0001-12-30 00:00:00.123")),
```
chenjian2664, line 95: nit: Add a link to the "limitation" doc link or the issue link, if it is not able to "correct" read

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveOnOrcLegacyDateCompatibility.java
marcinsbd: We need them for formatted to be call on merged regular strings literal across multiple lines.
I refactored it to text block

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveOnOrcLegacyDateCompatibility.java
marcinsbd: I refactored it to text block

lib/trino-orc/src/main/java/io/trino/orc/metadata/OrcMetadataReader.java
raunaqmorarka: lgtm, please resolve the build failures

marcinsbd: done

PR #26512, opened 2025-08-28
Add support for HASHTYPE data type in Exasol connector

skyglass:
## Description

Added Exasol Trino connector support for **HASHTYPE** with mapping to CHAR or VARBINARY, depending on the config

- added correspondent tests for mapping **HASHTYPE** to **VARBINARY**
- added correspondent tests for mapping **HASHTYPE** to **CHAR**
- added correspondent tests for unsupported **HASHTYPE** values

## Release notes

```markdown
## Exasol Connect
* Add support for `hashtype` data type
```


docs/src/main/sphinx/connector/exasol.md
```diff
@@ -127,6 +131,32 @@ If no length is specified, the connector uses 2000000.
...
+- **Input format**:  
+  HASHTYPE values must be supplied as strings:
+    - Hexadecimal characters only.
+    - Dashes (`-`) are allowed in the string but ignored.
+    - Curly braces (`{}`) around the string are allowed.
+
+- **Validation**:
+    - The input string must match the expected size (`2n` hex characters).
+    - If the string length or characters are invalid, an exception is thrown.
+    - Any non-hex characters (other than `-` or `{}` / `}`) make the value invalid.
```
ebyhr: I don't think we want to document it in the connector. Exasol doens't have the official documentation?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/config/TestExasolConfig.java
```diff
@@ -0,0 +1,60 @@
...
+import com.google.common.collect.ImmutableMap;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static io.airlift.configuration.testing.ConfigAssertions.assertFullMapping;
+import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
+import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+
+public class TestExasolConfig
```
ebyhr: https://trino.io/docs/current/develop/tests.html
> Test classes should be defined as package-private and final.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/config/TestExasolConfig.java
```diff
@@ -0,0 +1,60 @@
...
+{
+    @Test
+    public void testDefaults()
```
ebyhr: https://trino.io/docs/current/develop/tests.html
> Test methods should be defined as package-private.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/config/ExasolSessionProperties.java
```diff
@@ -0,0 +1,54 @@
...
+        implements SessionPropertiesProvider
+{
+    public static final String MAP_HASHTYPE_AS_CHAR = "map_hashtype_as_char";
+
+    private final List<PropertyMetadata<?>> sessionProperties;
+
+    @Inject
+    public ExasolSessionProperties(ExasolConfig exasolConfig)
+    {
+        sessionProperties = ImmutableList.of(
```
ebyhr: Use builder() so we can minimize the diff when adding a new entry to the bottom.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -262,6 +270,74 @@ private static LongWriteFunction dateWriteFunctionUsingSqlDate()
...
+        // Get the number of characters in the hex string.
+        int len = hex.length();
+
+        // Each byte is represented by 2 hex characters, so allocate half as many bytes.
+        byte[] data = new byte[len / 2];
+
+        // Process two characters (1 byte) at a time.
+        for (int i = 0; i < len; i += 2) {
+            // Convert the first hex character (the "high 4 bits") into a number 0–15.
+            int hi = Character.digit(hex.charAt(i), 16);
```
ebyhr: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-abbreviations

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/config/ExasolConfig.java
```diff
@@ -0,0 +1,35 @@
...
+    private boolean mapHashTypeAsChar;
+
+    public boolean isMapHashTypeAsChar()
+    {
+        return mapHashTypeAsChar;
+    }
+
+    @Config("exasol.legacy.map-hashtype-as-char")
+    @ConfigDescription("Map Exasol HashType as char instead of varbinary")
+    public ExasolConfig setMapHashTypeAsChar(boolean mapHashTypeAsChar)
```
Praveen2112: Any specific reason for mapping it to char ?

docs/src/main/sphinx/connector/exasol.md
Praveen2112: I think exasol has documentation around here - https://docs.exasol.com/db/latest/sql_references/data_types/datatypedetails.htm?Highlight=hashtype

Praveen2112 (commented):
Any specific reason behind mapping it to `char` instead of `varbinary` ?

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -15,22 +15,9 @@
...
-import io.trino.plugin.jdbc.JdbcJoinCondition;
-import io.trino.plugin.jdbc.JdbcOutputTableHandle;
-import io.trino.plugin.jdbc.JdbcTableHandle;
-import io.trino.plugin.jdbc.JdbcTypeHandle;
-import io.trino.plugin.jdbc.LongReadFunction;
-import io.trino.plugin.jdbc.LongWriteFunction;
-import io.trino.plugin.jdbc.QueryBuilder;
-import io.trino.plugin.jdbc.WriteFunction;
-import io.trino.plugin.jdbc.WriteMapping;
+import io.trino.plugin.jdbc.*;
```
Praveen2112: Can we revert this changes ? It is recomended to use specific imports instead of '*'

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/config/ExasolConfig.java
skyglass: Mapping to char is the current implementation, which we still have to support for backward compatibility.
This PR adds support for mapping to varbinary, which from now on will be the default behaviour.
But for backward compatibility, if the customer decides that he would like to continue mapping to char, he could use this property to make it possible.
But if this property is not set, then the default behaviour would be mapping to varbinary.
There is currently at least one customer, who uses trino exasol connector in production, so we have to deal with backward compatibility when changing things.

docs/src/main/sphinx/connector/exasol.md
skyglass: Yes, there is an official documentation: https://docs.exasol.com/saas/microcontent/Resources/MicroContent/general/hash-data-type.htm

I have added this link instead of the removed text.

skyglass:
> Now we support mapping it to varbinary ? Can we use this mapping by default ?

Yes, this is a default mapping. You don't need to set any properties.
See t**estHashtypeAsVarBinaryMapping** in T**estExasolTypeMapping** for more details. I don't set any properties there, so the default behaviour of mapping to varbinary is applied.

skyglass:
Thank you very much for the review comments, @ebyhr , @Praveen2112 ! 👍 
I have resolved the conversations and added code changes, based on the review comments.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -262,6 +286,74 @@ private static LongWriteFunction dateWriteFunctionUsingSqlDate()
         });
     }
 
+    private static ColumnMapping hashTypeVarBinaryColumnMapping()
```
Praveen2112: nit: hashTypeVarbinaryColumnMapping

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -262,6 +286,74 @@ private static LongWriteFunction dateWriteFunctionUsingSqlDate()
...
+    {
+        return SliceWriteFunction.of(Types.CHAR, (statement, index, slice) -> {
+            byte[] bytes = slice.getBytes();
+            String hex = byteArrayToHexString(bytes);
+            statement.setString(index, hex);
+        });
+    }
+
+    // Converts a hexadecimal string (e.g. "0A1B2C") into a byte array.
+    private static byte[] hexStringToByteArray(String hexString)
```
Praveen2112: nit: Would BaseEncoding.base16().decode(hexString) works here ?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,10 +280,143 @@ private void testDate(ZoneId sessionZone)
...
+    }
+
+    private void testHashtypeAsVarBinaryMapping()
+    {
+        SqlDataTypeTest.create()
+                // Null
+                .addRoundTrip("hashtype", "NULL", VARBINARY, "from_hex(NULL)")
+
+                // Explicit 16-byte
+                .addRoundTrip("hashtype(16 byte)", "'550e8400-e29b-11d4-a716-446655440000'", VARBINARY, "from_hex('550e8400e29b11d4a716446655440000')")
```
Praveen2112, line 290: What if the row to be inserted is less than precision ?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestingExasolServer.java
```diff
@@ -47,7 +47,7 @@ public class TestingExasolServer
 
     public TestingExasolServer()
     {
-        container = new ExasolContainer<>("8.32.0")
+        container = new ExasolContainer<>("8.34.0")
```
Praveen2112: Any specific reason for updating the version ? Was 8.32.0 is EoLed ?

Praveen2112: We might need to update in the docs as well and it could be a prep commit

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -262,6 +286,74 @@ private static LongWriteFunction dateWriteFunctionUsingSqlDate()
...
+    {
+        // Get the number of characters in the hex string.
+        int stringLength = hexString.length();
+
+        // Each byte is represented by 2 hex characters, so allocate half as many bytes.
+        byte[] byteArray = new byte[stringLength / 2];
+
+        // Process two characters (1 byte) at a time.
+        for (int charIndex = 0; charIndex < stringLength; charIndex += 2) {
```
chenjian2664: Did you implement the logic from scratch, or could you provide a reference/link if available?

skyglass: Thank you for pointing that out!
I used the following solution: https://stackoverflow.com/questions/140131/convert-a-string-representation-of-a-hex-dump-to-a-byte-array-using-java

But I have just realized, reading this article again, that since Java 17 there is already a standard function for that. 
So I will use **HexFormat.of().parseHex(hex)** and **HexFormat.of().formatHex(bytes)** instead, from **java.util** package, and will remove all this code, which implements the logic from scratch.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/config/ExasolConfig.java
Praveen2112: But this backward compatible change comes with a maintenance cost right ? Bunch of additional properties and session properties to be maintained. 

> if the customer decides that he would like to continue mapping to char

I'm curious on the benefits for this usage ? Any benefits are we seeing in this case ? At some point in the future if we are removing this config property then customer has to rewrite their query.

skyglass: > But this backward compatible change comes with a maintenance cost right ? Bunch of additional properties and session properties to be maintained.
> 
> I'm curious on the benefits for this usage ? Any benefits are we seeing in this case ? At some point in the future if we are removing this config property then customer has to rewrite their query.

We currently have customers running the Exasol connector in production, so we need to take backward compatibility into account.

**Existing behavior:** mapping HASHTYPE → CHAR.

**New behavior:** this PR introduces mapping HASHTYPE → VARBINARY, which becomes the default one

**Compatibility option:** for customers who rely on the old CHAR behavior, we provide a config property. This way, existing workloads are not broken when upgrading.

So effectively, the property adds flexibility to the upgrade process:

**New deployments** will get the correct default (VARBINARY).

**Existing deployments** can continue with the old behavior until they’re ready to migrate.

Yes, this adds a small maintenance cost (an extra property to maintain), but it’s the safest way to transition without forcing existing production customers to rewrite queries immediately. Once customers have migrated, we can consider deprecating and eventually removing the property.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestingExasolServer.java
skyglass: > Any specific reason for updating the version ? Was 8.32.0 is EoLed ?
> We might need to update in the docs as well and it could be a prep commit

The version is updated only for tests, no need to update the docs.
Ideally, the test should run for both **8.32.0** and **8.34.0**, but if we have to choose only one version, then it is better to use the latest version to be sure that there is no regression.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
skyglass: There is a similar test case and many other unsupported test cases in the `testUnsupportedHashType()`
In any case, I added to `testUnsupportedHashType()` one more test case: `Too short for declared size (expecting 16 bytes = 32 hex chars, got 31)`

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: > nit: Would BaseEncoding.base16().decode(hexString) works here ?

If **hexString** is a Java String, then yes. 
If **hexString** is a byte array, then no.
In Java API **BaseEncoding.base16().decode()**, requires String or CharSequence, so you would have to convert byte array to String like this:
**BaseEncoding.base16().decode(HexFormat.of().formatHex(bytes));**

In any case, this method is removed and replaced by **HexFormat.of().parseHex(hex)** from **java.util** package, which is available since Java 17

skyglass:
Thank you very much for the review, @Praveen2112 ! 👍
I have resolved the conversations and added code changes, based on the review comments.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -206,6 +212,13 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
         if (mapping.isPresent()) {
             return mapping;
         }
+        if (isHashTypeMapping(typeHandle)) {
+            Boolean useToCharMapping = session.getProperty(MAP_HASHTYPE_AS_CHAR, Boolean.class);
```
ebyhr: Add a helper method to ExasolSessionProperties, and make `MAP_HASHTYPE_AS_CHAR` private.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -237,6 +250,18 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
         return Optional.empty();
     }
 
+    private boolean isHashTypeMapping(JdbcTypeHandle typeHandle)
```
ebyhr: Rename to isHashType. JdbcTypeHandle is not a mapping.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -262,6 +287,32 @@ private static LongWriteFunction dateWriteFunctionUsingSqlDate()
         });
     }
 
+    private static ColumnMapping hashTypeVarbinaryColumnMapping()
+    {
+        return ColumnMapping.sliceMapping(
+                VarbinaryType.VARBINARY,
```
ebyhr: Static import `VARBINARY`.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/config/ExasolSessionProperties.java
```diff
@@ -0,0 +1,50 @@
...
+        implements SessionPropertiesProvider
+{
+    public static final String MAP_HASHTYPE_AS_CHAR = "map_hashtype_as_char";
+
+    private final List<PropertyMetadata<?>> sessionProperties;
+
+    @Inject
+    public ExasolSessionProperties(ExasolConfig exasolConfig)
+    {
+        // Use builder to minimize diff when adding new entries later
```
ebyhr: Remove an unhelpful comment.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,10 +280,150 @@ private void testDate(ZoneId sessionZone)
     @Test
     void testHashtype()
     {
+        testHashtypeAsVarbinaryMapping();
+        testHashTypeAsCharMapping();
```
ebyhr: Add `@Test` to those 2 test methods instead.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,10 +280,150 @@ private void testDate(ZoneId sessionZone)
...
+            String exasolType,
+            String inputLiteral,
+            String expectedValue,
+            String expectedException)
+    {
+        assertThatThrownBy(() ->
+                SqlDataTypeTest.create()
+                        .addRoundTrip(exasolType, inputLiteral, VARBINARY, expectedValue)
+                        .execute(getQueryRunner(),
+                                exasolCreateAndInsert(TEST_SCHEMA + "." + "test_unsupported_hashtype")))
```
ebyhr: Unwrap.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestingExasolServer.java
ebyhr: Please update `Requirements` section of `exasol.md`. We avoid mentioning untested versions in our documentation.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/config/ExasolConfig.java
ebyhr: I don't think we want to add this property. This isn't the usual way we handle type mappings in connectors.

chenjian2664: @skyglass we have the `jdbc-types-mapped-to-varchar` to allow types mapping to varchar, does it resolve your problem?

skyglass: Actually, Clickhouse used the similar approach by providing the custom property "map_string_as_varchar": 
[ClickHouseSessionProperties.java](https://github.com/trinodb/trino/blob/master/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseSessionProperties.java)

But after discussion with the product owner we decided to only keep mapping to varbinary, so I will remove correspondent classes and properties related to mapping hashtype to varchar.

skyglass:
Thank you very much for the review, @ebyhr ! 👍
I have resolved the conversations and added code changes, based on the review comments.

docs/src/main/sphinx/connector/exasol.md
```diff
@@ -16,7 +16,8 @@ The Exasol connector allows querying an [Exasol](https://www.exasol.com/) databa
 
 To connect to Exasol, you need:
 
-* Exasol database version 7.1 or higher.
+* Exasol database 8.34.0 or higher. 
+Older versions (7.1 or higher) may work, but are not regularly tested.
```
Praveen2112: I think this change should be a part of commit where we update the version.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClientModule.java
```diff
@@ -43,6 +46,8 @@ protected void setup(Binder binder)
     {
         binder.bind(JdbcClient.class).annotatedWith(ForBaseJdbc.class).to(ExasolClient.class).in(Scopes.SINGLETON);
         configBinder(binder).bindConfig(JdbcStatisticsConfig.class);
+        configBinder(binder).bindConfig(ExasolConfig.class);
```
Praveen2112: Do we have this config and session properties ?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +280,114 @@ private void testDate(ZoneId sessionZone)
...
+            String inputLiteral,
+            String expectedValue,
+            String expectedException)
+    {
+        assertThatThrownBy(() ->
+                SqlDataTypeTest.create()
+                        .addRoundTrip(exasolType, inputLiteral, VARBINARY, expectedValue)
+                        .execute(getQueryRunner(), exasolCreateAndInsert(TEST_SCHEMA + "." + "test_unsupported_hashtype")))
+                .cause()
+                .hasMessageContaining(expectedException);
```
Praveen2112: If we are going to validate the insert behaviour on Exasol engine, isn't using `SqlDataTypeTest` is a bit heavy ? We could directly use 
`assertThatThrownBy(() -> exasolServer.execute("CREATE TABLE " + TEST_SCHEMA + ".unsupported_hash (col hashtype)"))` to assert failure during creation and apply a similar assertion via `TestTable` for inserting data.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestingExasolServer.java
Praveen2112, line 1: I would personally swap the commit order -

skyglass:
> Now we support mapping it to varbinary ? Can we use this mapping by default ?

Yes, I removed support for mapping to char, now mapping to varbinary is used by default.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClientModule.java
skyglass: Sorry, forgot to remove them, will update the PR now...

docs/src/main/sphinx/connector/exasol.md
skyglass: OK, I will create a separate PR for version update change and will remove version update for now.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestingExasolServer.java
skyglass: I have reverted exasol db version update and will do it in a separate PR.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
skyglass: The tests have been rewritten to directly run exasol sql queries on exasol server. Also, after refactoring, there are now two types of tests: testUnsupportedHashTypeDefinition and testUnsupportedHashTypeValue.

skyglass:
Thank you very much for the review, @Praveen2112 ! 👍
I have resolved the conversations and added code changes, based on the review comments.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +281,115 @@ private void testDate(ZoneId sessionZone)
...
+        }
+    }
+
+    private void testUnsupportedHashTypeValue(
+            String exasolType,
+            String inputLiteral,
+            String expectedException)
+    {
+        try {
+            exasolServer.execute("CREATE TABLE %s.test_unsupported_hashtype (col %s)".formatted(TEST_SCHEMA, exasolType));
```
Praveen2112: Can we use test table here ? As the create statement is valid

skyglass: I tried to rewrite it with TestTable, but it doesn't look good to me. It hides the actual sql query and also the created testTable is never used. I think that providing the explicit sql query here is better for readability of the test and understanding why the sql query throws the exception. But I improved the tests a little bit by removing try finally block and dropping of the table in **testUnsupportedHashTypeDefinition** (the exception happens during table creation, so the table is never created, so "DROP TABLE" is not needed there)

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +281,110 @@ private void testDate(ZoneId sessionZone)
...
+                expectedException);
+    }
+
+    private void testUnsupportedHashTypeValue(
+            String exasolType,
+            String inputLiteral,
+            String expectedException)
+    {
+        try {
+            exasolServer.execute("CREATE TABLE %s.test_unsupported_hashtype (col %s)".formatted(TEST_SCHEMA, exasolType));
```
ebyhr: Please add random suffix to the table name. The static table name will be troublesome if we use this method from several tests.

skyglass:
Thank you very much for your approval @Praveen2112 ! 👍 
Could you, please, reapprove again?
I made a small change based on your latest comment.

Also, the PR says "Merging is blocked", even after 1 approval.
How many approvals do I need to be able to merge?

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -262,6 +277,32 @@ private static LongWriteFunction dateWriteFunctionUsingSqlDate()
         });
     }
 
+    private static ColumnMapping hashTypeVarbinaryColumnMapping()
```
ebyhr: Rename to `hashTypeColumnMapping`

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +281,110 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("hashtype(1 byte)", "'FF'", VARBINARY, "from_hex('FF')")
+
+                // Boundary: maximum size (1024 byte) - use repeated pattern
+                .addRoundTrip("hashtype(1024 byte)", "'" + "AA".repeat(1024) + "'", VARBINARY, "from_hex('" + "AA".repeat(1024) + "')")
+
+                .execute(getQueryRunner(), exasolCreateAndInsert(TEST_SCHEMA + "." + "test_hashtype_as_varbinary_mapping"));
+    }
+
+    @Test
+    // See for more details: https://docs.exasol.com/saas/microcontent/Resources/MicroContent/general/hash-data-type.htm
```
ebyhr, line 310: Move this comment above `@Test`. Otherwise, auto-format changes the indentation.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +281,110 @@ private void testDate(ZoneId sessionZone)
...
+    private void testUnsupportedHashTypeDefinition(
+            String exasolType,
+            String expectedException)
+    {
+        assertExasolSqlQueryFails(
+                "CREATE TABLE %s.test_unsupported_hashtype (col %s)".formatted(TEST_SCHEMA, exasolType),
+                expectedException);
+    }
+
+    private void testUnsupportedHashTypeValue(
```
Praveen2112: This is specific to insert operation right ?

skyglass: Yes, in this test, the creation of the table should be successful, but insert operation should fail. 
I will rename it to **testUnsupportedHashTypeInsertValue**

skyglass (commented):
The last review suggestions have been resolved.
Thank you, @ebyhr , @Praveen2112 ! 👍 

If the build is green and there is 1 approval, is it enough to merge the PR. Or there should be more approvals?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +282,112 @@ private void testDate(ZoneId sessionZone)
...
+    {
+        String tableName = "test_unsupported_hashtype_" + randomNameSuffix();
+        try {
+            exasolServer.execute("CREATE TABLE %s.%s (col %s)".formatted(TEST_SCHEMA, tableName, exasolType));
+            assertExasolSqlQueryFails(
+                    "INSERT INTO %s.%s VALUES (%s)".formatted(TEST_SCHEMA, tableName, inputLiteral),
+                    expectedException);
+        }
+        finally {
+            exasolServer.execute("DROP TABLE %s.%s".formatted(TEST_SCHEMA, tableName));
```
ebyhr: This statement may throw an exception if `CREATE TABLE` fails for some reason. Please add `IF EXISTS` option.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -279,9 +282,112 @@ private void testDate(ZoneId sessionZone)
     void testHashtype()
     {
         SqlDataTypeTest.create()
-                .addRoundTrip("hashtype", "NULL", createCharType(32), "CAST(NULL AS char(32))")
-                .addRoundTrip("hashtype", "'550e8400-e29b-11d4-a716-446655440000'", createCharType(32), "CAST('550e8400e29b11d4a716446655440000' AS char(32))")
-                .execute(getQueryRunner(), exasolCreateAndInsert(TEST_SCHEMA + "." + "test_hashtype"));
+                // Null
```
ebyhr, line 286: https://docs.exasol.com/saas/microcontent/Resources/MicroContent/general/hash-data-type.htm says: 
> Curly brackets surrounding the input hash string are allowed.

Please add the test case about it.

skyglass (commented):
The last review comments have been resolved.
Thank you, @ebyhr ! 👍

ebyhr (approved):

PR #26519, opened 2025-08-30
Support RSA private key authentication in Snowflake connector

### LINKED ISSUE #26514: Snowflake Connector: Add support for `privateKey` authentication property
The Trino Snowflake connector currently does not support using the `privateKey` property that the Snowflake JDBC driver exposes for key-pair authentication. According to the Snowflake documentation, this property allows passing a private key directly in the connection properties:
[https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-configure#privatekey-property-in-connection-properties](https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-configure?utm_source=chatgpt.com#privatekey-property-in-connection-properties)

Right now, Trino only supports passing key-related parameters through the JDBC URL. That approach requires providing a file path and password, which can introduce security risks and add operational overhead. Managing file paths is particularly difficult in containerized or cloud-native environments.

Supporting the `privateKey` property directly would be both safer and more convenient. It removes the need to reference external files and instead allows secure configuration as part of the catalog itself. This makes key-pair authentication much easier to adopt while aligning with best practices for passwordless connections.

In our use case, we are also leveraging dynamic catalog creation. Having support for the `privateKey` property at the catalog level makes this workflow much simpler and avoids the complexity of file-based approaches, which are harder to manage dynamically.

parthshyara:
## Description
* Adds support for key-pair authentication in the Snowflake connector through the `snowflake.connection-private-key` property, which accepts a Base64-encoded PKCS#8 RSA private key.
* Adds an optional `snowflake.connection-private-key-password` property so encrypted private keys are supported.
* The key material is passed through to the Snowflake JDBC driver (`private_key_base64` and `private_key_pwd`), which parses and decrypts it. Both encrypted and unencrypted keys are supported without any key handling in the connector.
* Users can configure key-pair authentication directly in the catalog instead of relying on file-based approaches.

* Fixes #26514
* Provides a more secure and convenient alternative to using JDBC URL parameters with file paths and passwords.
* Particularly useful for dynamic catalog creation, where file system dependencies are harder to manage.

## Release notes

(&#x2713;) Release notes are required, with the following suggested text:

```markdown
## Snowflake connector
* Add support for key-pair authentication with the `snowflake.connection-private-key` and `snowflake.connection-private-key-password` catalog properties. (#26514)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -765,6 +765,8 @@ jobs:
           SNOWFLAKE_URL: ${{ vars.SNOWFLAKE_URL }}
           SNOWFLAKE_USER: ${{ vars.SNOWFLAKE_USER }}
           SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
+          SNOWFLAKE_PRIVATE_KEY: ${{ secrets.SNOWFLAKE_PRIVATE_KEY }}
+          SNOWFLAKE_PRIVATE_KEY_PASSPHRASE: ${{ secrets.SNOWFLAKE_PRIVATE_KEY_PASSPHRASE }}
           SNOWFLAKE_DATABASE: ${{ vars.SNOWFLAKE_DATABASE }}
           SNOWFLAKE_ROLE: ${{ vars.SNOWFLAKE_ROLE }}
           SNOWFLAKE_WAREHOUSE: ${{ vars.SNOWFLAKE_WAREHOUSE }}
@@ -774,6 +776,8 @@ jobs:
             -Dsnowflake.test.server.url="${SNOWFLAKE_URL}" \
             -Dsnowflake.test.server.user="${SNOWFLAKE_USER}" \
             -Dsnowflake.test.server.password="${SNOWFLAKE_PASSWORD}" \
+            -Dsnowflake.test.server.private-key="${SNOWFLAKE_PRIVATE_KEY}" \
+            -Dsnowflake.test.server.private-key-passphrase="${SNOWFLAKE_PRIVATE_KEY_PASSPHRASE}" \
             -Dsnowflake.test.server.database="${SNOWFLAKE_DATABASE}" \
             -Dsnowflake.test.server.role="${SNOWFLAKE_ROLE}" \
             -Dsnowflake.test.server.warehouse="${SNOWFLAKE_WAREHOUSE}"
--- a/docs/src/main/sphinx/connector/snowflake.md
+++ b/docs/src/main/sphinx/connector/snowflake.md
@@ -26,12 +26,19 @@ connection properties as appropriate for your setup:
 connector.name=snowflake
 connection-url=jdbc:snowflake://<account>.snowflakecomputing.com
 connection-user=root
-connection-password=secret
+snowflake.connection-private-key=private-key-content
 snowflake.account=account
 snowflake.database=database
 snowflake.role=role
 snowflake.warehouse=warehouse
 ```
+
+:::{note}
+The `snowflake.connection-private-key` property accepts a Base64-encoded RSA private key
+in PKCS#8 format. If the private key is encrypted, also set
+`snowflake.connection-private-key.passphrase` to the passphrase used to decrypt it.
+Alternatively, you can use `connection-password` for password-based authentication.
+:::
 
 The Snowflake connector uses Apache Arrow as the serialization format when
 reading from Snowflake. Add the following required, additional JVM argument
--- a/plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
+++ b/plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
@@ -72,6 +72,8 @@ public class SnowflakeClientModule
         snowflakeConfig.getDatabase().ifPresent(database -> properties.setProperty("db", database));
         snowflakeConfig.getRole().ifPresent(role -> properties.setProperty("role", role));
         snowflakeConfig.getWarehouse().ifPresent(warehouse -> properties.setProperty("warehouse", warehouse));
+        snowflakeConfig.getPrivateKey().ifPresent(privateKey -> properties.setProperty("private_key_base64", privateKey));
+        snowflakeConfig.getPrivateKeyPassphrase().ifPresent(passphrase -> properties.setProperty("private_key_pwd", passphrase));
 
         setOutputProperties(properties);
 
--- a/plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
+++ b/plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
@@ -14,7 +14,9 @@
 package io.trino.plugin.snowflake;
 
 import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
 import io.airlift.configuration.ConfigSecuritySensitive;
+import jakarta.validation.constraints.AssertTrue;
 
 import java.util.Optional;
 
@@ -25,6 +27,8 @@ public class SnowflakeConfig
     private String role;
     private String warehouse;
     private String httpProxy;
+    private String privateKey;
+    private String privateKeyPassphrase;
 
     public Optional<String> getAccount()
     {
@@ -85,5 +89,39 @@ public class SnowflakeConfig
     {
         this.httpProxy = httpProxy;
         return this;
+    }
+
+    public Optional<String> getPrivateKey()
+    {
+        return Optional.ofNullable(privateKey);
+    }
+
+    @Config("snowflake.connection-private-key")
+    @ConfigDescription("Base64-encoded PKCS#8 RSA private key for key-pair authentication")
+    @ConfigSecuritySensitive
+    public SnowflakeConfig setPrivateKey(String privateKey)
+    {
+        this.privateKey = privateKey;
+        return this;
+    }
+
+    public Optional<String> getPrivateKeyPassphrase()
+    {
+        return Optional.ofNullable(privateKeyPassphrase);
+    }
+
+    @Config("snowflake.connection-private-key.passphrase")
+    @ConfigDescription("Passphrase to decrypt an encrypted private key")
+    @ConfigSecuritySensitive
+    public SnowflakeConfig setPrivateKeyPassphrase(String privateKeyPassphrase)
+    {
+        this.privateKeyPassphrase = privateKeyPassphrase;
+        return this;
+    }
+
+    @AssertTrue(message = "snowflake.connection-private-key.passphrase is only allowed when snowflake.connection-private-key is set")
+    public boolean isPrivateKeyPassphraseAllowed()
+    {
+        return privateKey != null || privateKeyPassphrase == null;
     }
 }
--- a/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/SnowflakeQueryRunner.java
+++ b/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/SnowflakeQueryRunner.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import static io.airlift.testing.Closeables.closeAllSuppress;
 import static io.trino.plugin.tpch.TpchMetadata.TINY_SCHEMA_NAME;
 import static io.trino.testing.QueryAssertions.copyTpchTables;
+import static io.trino.testing.TestingProperties.requiredNonEmptySystemProperty;
 import static io.trino.testing.TestingSession.testSessionBuilder;
 import static java.util.Objects.requireNonNull;
 
@@ -48,7 +49,6 @@ public final class SnowflakeQueryRunner
         Builder builder = new Builder()
                 .addConnectorProperty("connection-url", TestingSnowflakeServer.TEST_URL)
                 .addConnectorProperty("connection-user", TestingSnowflakeServer.TEST_USER)
-                .addConnectorProperty("connection-password", TestingSnowflakeServer.TEST_PASSWORD)
                 .addConnectorProperty("snowflake.database", TestingSnowflakeServer.TEST_DATABASE)
                 .addConnectorProperty("snowflake.warehouse", TestingSnowflakeServer.TEST_WAREHOUSE);
         TestingSnowflakeServer.TEST_ROLE.ifPresent(role -> builder.addConnectorProperty("snowflake.role", role));
@@ -111,6 +111,7 @@ public final class SnowflakeQueryRunner
             throws Exception
     {
         DistributedQueryRunner queryRunner = builder()
+                .addConnectorProperty("connection-password", requiredNonEmptySystemProperty("snowflake.test.server.password"))
                 .addCoordinatorProperty("http-server.http.port", "8080")
                 .build();
 
--- a/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakePlugin.java
+++ b/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakePlugin.java
@@ -30,4 +30,33 @@ public class TestSnowflakePlugin
         ConnectorFactory factory = getOnlyElement(plugin.getConnectorFactories());
         factory.create("test", ImmutableMap.of("connection-url", "jdbc:snowflake://test"), new TestingConnectorContext()).shutdown();
     }
+
+    @Test
+    public void testCreateConnectorWithPrivateKey()
+    {
+        Plugin plugin = new SnowflakePlugin();
+        ConnectorFactory factory = getOnlyElement(plugin.getConnectorFactories());
+        factory.create(
+                        "test",
+                        ImmutableMap.of(
+                                "connection-url", "jdbc:snowflake://test",
+                                "snowflake.connection-private-key", "dummy-private-key"),
+                        new TestingConnectorContext())
+                .shutdown();
+    }
+
+    @Test
+    public void testCreateConnectorWithEncryptedPrivateKey()
+    {
+        Plugin plugin = new SnowflakePlugin();
+        ConnectorFactory factory = getOnlyElement(plugin.getConnectorFactories());
+        factory.create(
+                        "test",
+                        ImmutableMap.of(
+                                "connection-url", "jdbc:snowflake://test",
+                                "snowflake.connection-private-key", "dummy-private-key",
+                                "snowflake.connection-private-key.passphrase", "dummy-passphrase"),
+                        new TestingConnectorContext())
+                .shutdown();
+    }
 }
--- a/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java
+++ b/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.snowflake;
+
+import io.trino.plugin.jdbc.BaseJdbcConnectorSmokeTest;
+import io.trino.testing.QueryRunner;
+import io.trino.testing.TestingConnectorBehavior;
+
+import static io.trino.testing.TestingProperties.requiredNonEmptySystemProperty;
+
+final class TestSnowflakeWithPrivateKeyConnectorSmokeTest
+        extends BaseJdbcConnectorSmokeTest
+{
+    @Override
+    protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
+    {
+        return switch (connectorBehavior) {
+            case SUPPORTS_ADD_COLUMN_WITH_COMMENT,
+                 SUPPORTS_ADD_COLUMN_WITH_POSITION,
+                 SUPPORTS_ARRAY,
+                 SUPPORTS_COMMENT_ON_COLUMN,
+                 SUPPORTS_CREATE_TABLE_WITH_COLUMN_COMMENT,
+                 SUPPORTS_MAP_TYPE,
+                 SUPPORTS_MERGE,
+                 SUPPORTS_PREDICATE_EXPRESSION_PUSHDOWN,
+                 SUPPORTS_PREDICATE_PUSHDOWN_WITH_VARCHAR_EQUALITY,
+                 SUPPORTS_PREDICATE_PUSHDOWN_WITH_VARCHAR_INEQUALITY,
+                 SUPPORTS_ROW_LEVEL_UPDATE,
+                 SUPPORTS_ROW_TYPE,
+                 SUPPORTS_SET_COLUMN_TYPE -> false;
+            default -> super.hasBehavior(connectorBehavior);
+        };
+    }
+
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return SnowflakeQueryRunner.builder()
+                .addConnectorProperty("snowflake.connection-private-key", requiredNonEmptySystemProperty("snowflake.test.server.private-key"))
+                .addConnectorProperty("snowflake.connection-private-key.passphrase", requiredNonEmptySystemProperty("snowflake.test.server.private-key-passphrase"))
+                .setInitialTables(REQUIRED_TPCH_TABLES)
+                .build();
+    }
+}
--- a/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestingSnowflakeServer.java
+++ b/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestingSnowflakeServer.java
@@ -22,6 +22,7 @@ import java.sql.Statement;
 import java.util.Optional;
 import java.util.Properties;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static io.trino.plugin.snowflake.SnowflakeClientModule.setOutputProperties;
 import static io.trino.testing.TestingProperties.requiredNonEmptySystemProperty;
 
@@ -31,7 +32,9 @@ public final class TestingSnowflakeServer
 
     public static final String TEST_URL = requiredNonEmptySystemProperty("snowflake.test.server.url");
     public static final String TEST_USER = requiredNonEmptySystemProperty("snowflake.test.server.user");
-    public static final String TEST_PASSWORD = requiredNonEmptySystemProperty("snowflake.test.server.password");
+    public static final Optional<String> TEST_PASSWORD = Optional.ofNullable(System.getProperty("snowflake.test.server.password"));
+    public static final Optional<String> TEST_PRIVATE_KEY = Optional.ofNullable(System.getProperty("snowflake.test.server.private-key"));
+    public static final Optional<String> TEST_PRIVATE_KEY_PASSPHRASE = Optional.ofNullable(System.getProperty("snowflake.test.server.private-key-passphrase"));
     public static final String TEST_DATABASE = requiredNonEmptySystemProperty("snowflake.test.server.database");
     public static final String TEST_WAREHOUSE = requiredNonEmptySystemProperty("snowflake.test.server.warehouse");
     public static final Optional<String> TEST_ROLE = Optional.ofNullable(System.getProperty("snowflake.test.server.role"));
@@ -57,7 +60,16 @@ public final class TestingSnowflakeServer
     {
         Properties properties = new Properties();
         properties.setProperty("user", TEST_USER);
-        properties.setProperty("password", TEST_PASSWORD);
+        checkArgument(
+                TEST_PRIVATE_KEY.isPresent() || TEST_PASSWORD.isPresent(),
+                "At least one of the private key or password must be set");
+        if (TEST_PRIVATE_KEY.isPresent()) {
+            properties.setProperty("private_key_base64", TEST_PRIVATE_KEY.get());
+            TEST_PRIVATE_KEY_PASSPHRASE.ifPresent(passphrase -> properties.setProperty("private_key_pwd", passphrase));
+        }
+        else {
+            properties.setProperty("password", TEST_PASSWORD.get());
+        }
         properties.setProperty("db", TEST_DATABASE);
         properties.setProperty("schema", TEST_SCHEMA);
         properties.setProperty("warehouse", TEST_WAREHOUSE);
--- a/plugin/trino-snowflake/pom.xml
+++ b/plugin/trino-snowflake/pom.xml
@@ -47,6 +47,11 @@
         </dependency>
 
         <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>net.snowflake</groupId>
             <artifactId>snowflake-jdbc</artifactId>
         </dependency>
@@ -230,6 +235,7 @@
                             <excludes>
                                 <exclude>**/TestSnowflakeConnectorTest.java</exclude>
                                 <exclude>**/TestSnowflakeTypeMapping.java</exclude>
+                                <exclude>**/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java</exclude>
                             </excludes>
                         </configuration>
                     </plugin>
@@ -252,6 +258,7 @@
                             <includes>
                                 <include>**/TestSnowflakeConnectorTest.java</include>
                                 <include>**/TestSnowflakeTypeMapping.java</include>
+                                <include>**/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java</include>
                             </includes>
                         </configuration>
                     </plugin>
--- a/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeConfig.java
+++ b/plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeConfig.java
@@ -14,6 +14,7 @@
 package io.trino.plugin.snowflake;
 
 import com.google.common.collect.ImmutableMap;
+import jakarta.validation.constraints.AssertTrue;
 import org.junit.jupiter.api.Test;
 
 import java.util.Map;
@@ -21,6 +22,7 @@ import java.util.Map;
 import static io.airlift.configuration.testing.ConfigAssertions.assertFullMapping;
 import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
 import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+import static io.airlift.testing.ValidationAssertions.assertFailsValidation;
 
 public class TestSnowflakeConfig
 {
@@ -32,7 +34,9 @@ public class TestSnowflakeConfig
                 .setDatabase(null)
                 .setRole(null)
                 .setWarehouse(null)
-                .setHttpProxy(null));
+                .setHttpProxy(null)
+                .setPrivateKey(null)
+                .setPrivateKeyPassphrase(null));
     }
 
     @Test
@@ -44,6 +48,8 @@ public class TestSnowflakeConfig
                 .put("snowflake.role", "MYROLE")
                 .put("snowflake.warehouse", "MYWAREHOUSE")
                 .put("snowflake.http-proxy", "MYPROXY")
+                .put("snowflake.connection-private-key", "MYPRIVATEKEY")
+                .put("snowflake.connection-private-key.passphrase", "MYPASSPHRASE")
                 .buildOrThrow();
 
         SnowflakeConfig expected = new SnowflakeConfig()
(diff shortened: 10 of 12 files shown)


docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -689,7 +689,9 @@ properties:
 * - `iceberg.snowflake-catalog.user`
   - Snowflake user (required).
 * - `iceberg.snowflake-catalog.password`
-  - Snowflake password (required).
+  - Snowflake password (this or private-key required).
+* - `iceberg.snowflake-catalog.private-key`
+  - Private key content in PEM format (this or password required).
```
chenjian2664: Does it means Exactly one of the password or private key must be specified?
If so, please refer the https://github.com/trinodb/trino/blob/4f50ae66de7af866c8ad354499a2503c93027be5/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3SecurityMappingConfig.java#L143
add the verification here

parthshyara: Makes sense. Added.

parthshyara: @chenjian2664 Upon further analysis and testing, I figured that iceberg-jdbc connector does not support arbitrary types as jdbc property values. The `privateKey` need to be an instance of `java.security.PrivateKey`. Since the iceberg-jdbc's `JDBCClientPool` only supports Map<String, String> , we cannot implement this change for icerberg-snowflake as of now. Hence, removed all the changes related to that.

chenjian2664: I don't get you, which stops us doing this? is the implementation or the is that we shouldn't have such check?

parthshyara: > we cannot implement this change

This comment was regarding the task as a whole. We cannot implement `privateKey` as a JVM property for iceberg-snowflake. Hence, your comment regarding adding the check is no longer valid. Because all the changes to this connector are now removed.

parthshyara: @chenjian2664 Please do take a look and let me know if there are concerns. Additionally, for snowflake cloud tests, we need to add `SNOWFLAKE_PRIVATE_KEY` github secret and relevant changes in the cloud snowflake account used for CI testing.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
```diff
@@ -116,4 +120,17 @@ protected static void setOutputProperties(Properties properties)
         // Don't treat decimals as bigints as they may overflow
         properties.setProperty("JDBC_TREAT_DECIMAL_AS_INT", "FALSE");
     }
+
+    protected static PrivateKey parsePrivateKey(String privateKey)
+    {
+        try {
+            byte[] keyBytes = java.util.Base64.getDecoder().decode(privateKey);
```
chenjian2664: static import

docs/src/main/sphinx/connector/snowflake.md
```diff
@@ -27,12 +27,18 @@ connector.name=snowflake
 connection-url=jdbc:snowflake://<account>.snowflakecomputing.com
 connection-user=root
 connection-password=secret
+snowflake.private-key=unencrypted-private-key-content
```
chenjian2664: Does this two needs to be provided when setup?

parthshyara: Yes, one of the two needs to be provided, not both.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
parthshyara: Fixed.

docs/src/main/sphinx/connector/snowflake.md
chenjian2664: then please remove password from the example configurations, mention it is deprecated in the note section

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestingSnowflakeServer.java
```diff
@@ -57,7 +59,8 @@ private static Properties getProperties()
     {
         Properties properties = new Properties();
         properties.setProperty("user", TEST_USER);
-        properties.setProperty("password", TEST_PASSWORD);
+        TEST_PRIVATE_KEY.ifPresent(privateKeyString -> properties.put("privateKey", SnowflakeClientModule.parsePrivateKey(privateKeyString)));
+        TEST_PASSWORD.ifPresent(password -> properties.setProperty("password", password));
```
chenjian2664: add checkArgument either the `TEST_PRIVATE_KEY` or `TEST_PASSWORD` should provide in the test

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/EnvMultinodeSnowflake.java
```diff
@@ -48,6 +48,7 @@ public void extendEnvironment(Environment.Builder builder)
                         .withEnv("SNOWFLAKE_URL", requireEnv("SNOWFLAKE_URL"))
                         .withEnv("SNOWFLAKE_USER", requireEnv("SNOWFLAKE_USER"))
                         .withEnv("SNOWFLAKE_PASSWORD", requireEnv("SNOWFLAKE_PASSWORD"))
+                        .withEnv("SNOWFLAKE_PRIVATE_KEY", requireEnv("SNOWFLAKE_PRIVATE_KEY"))
```
chenjian2664: I would recommend to add another product env somehow like "EnvMultinodeSnowflakeWithPrivateKey", and it's for the properties that you added in ci.yml

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/SnowflakeQueryRunner.java
```diff
@@ -41,9 +41,13 @@ public static Builder builder()
...
                 .addConnectorProperty("connection-url", TestingSnowflakeServer.TEST_URL)
                 .addConnectorProperty("connection-user", TestingSnowflakeServer.TEST_USER)
-                .addConnectorProperty("connection-password", TestingSnowflakeServer.TEST_PASSWORD)
                 .addConnectorProperty("snowflake.database", TestingSnowflakeServer.TEST_DATABASE)
                 .addConnectorProperty("snowflake.warehouse", TestingSnowflakeServer.TEST_WAREHOUSE);
+
+        TestingSnowflakeServer.TEST_PRIVATE_KEY
+                .ifPresent(privateKey -> builder.addConnectorProperty("snowflake.private-key", privateKey));
+        TestingSnowflakeServer.TEST_PASSWORD
+                .ifPresent(password -> builder.addConnectorProperty("connection-password", password));
```
chenjian2664: add checkArgument either the TEST_PRIVATE_KEY or TEST_PASSWORD should provide in the test

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
chenjian2664, line 95: if `privateKey` is already provided above, does this condition still be true?

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestingSnowflakeServer.java
parthshyara: Done.

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/EnvMultinodeSnowflake.java
parthshyara: Done!

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/SnowflakeQueryRunner.java
parthshyara: Done!

docs/src/main/sphinx/connector/snowflake.md
parthshyara: Done!

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
parthshyara: Yes, because this is formed from the `httpProxy` connection string itself. This should be unrelated to the snowflake credentials.

docs/src/main/sphinx/connector/snowflake.md
```diff
@@ -26,13 +26,18 @@ connection properties as appropriate for your setup:
...
 snowflake.account=account
 snowflake.database=database
 snowflake.role=role
 snowflake.warehouse=warehouse
 ```
 
+:::{note}
+Use private-key for RSA based authentication.
+Password based authentication (`connection-password=secret`) is deprecated by Snowflake.
```
findepi: We should document `connection-password` as long as it works, and we don't need to replicate Snowflake's deprecation notices here. It's not our job.

It's good that the example above is updated to reflect recommended configuration.

findepi:
> add privateKey as a JVM property for Snowflake

The private key is handled as a normal connector configuration property, and not via JVM system properties, right?
if so, please update the PR title

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/SnowflakeQueryRunner.java
```diff
@@ -41,9 +42,18 @@ public static Builder builder()
...
                 .addConnectorProperty("connection-url", TestingSnowflakeServer.TEST_URL)
                 .addConnectorProperty("connection-user", TestingSnowflakeServer.TEST_USER)
-                .addConnectorProperty("connection-password", TestingSnowflakeServer.TEST_PASSWORD)
                 .addConnectorProperty("snowflake.database", TestingSnowflakeServer.TEST_DATABASE)
                 .addConnectorProperty("snowflake.warehouse", TestingSnowflakeServer.TEST_WAREHOUSE);
+
+        checkArgument(
+                (TestingSnowflakeServer.TEST_PRIVATE_KEY.isPresent() ? 1 : 0)
+                        + (TestingSnowflakeServer.TEST_PASSWORD.isPresent() ? 1 : 0)
+                        == 1,
```
chenjian2664:
```suggestion
TestingSnowflakeServer.TEST_PRIVATE_KEY.isPresent() ^ TestingSnowflakeServer.TEST_PASSWORD.isPresent()
```

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestingSnowflakeServer.java
```diff
@@ -57,7 +60,13 @@ private static Properties getProperties()
     {
         Properties properties = new Properties();
         properties.setProperty("user", TEST_USER);
-        properties.setProperty("password", TEST_PASSWORD);
+        checkArgument(
+                (TestingSnowflakeServer.TEST_PRIVATE_KEY.isPresent() ? 1 : 0)
```
chenjian2664: ditto

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
```diff
@@ -116,4 +121,17 @@ protected static void setOutputProperties(Properties properties)
         // Don't treat decimals as bigints as they may overflow
         properties.setProperty("JDBC_TREAT_DECIMAL_AS_INT", "FALSE");
     }
+
+    protected static PrivateKey parsePrivateKey(String privateKey)
+    {
+        try {
+            byte[] keyBytes = Base64.getDecoder().decode(privateKey);
+            PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(keyBytes);
+            KeyFactory keyFactory = KeyFactory.getInstance("RSA");
```
chenjian2664: Is it the only supported algo that used for generating private key, please document it

parthshyara: Done!

docs/src/main/sphinx/connector/snowflake.md
parthshyara: Makes sense. Done.

docs/src/main/sphinx/connector/snowflake.md
```diff
@@ -26,13 +26,18 @@ connection properties as appropriate for your setup:
...
-connection-password=secret
+snowflake.private-key=unencrypted-private-key-content
 snowflake.account=account
 snowflake.database=database
 snowflake.role=role
 snowflake.warehouse=warehouse
 ```
 
+:::{note}
+The `snowflake.private-key` property accepts a Base64-encoded private key in PKCS8 format. Only RSA algorithm is supported for private key authentication.
```
chenjian2664: This line is too long

parthshyara: Please check again.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
```diff
@@ -86,4 +87,17 @@ public SnowflakeConfig setHttpProxy(String httpProxy)
         this.httpProxy = httpProxy;
         return this;
     }
+
+    public Optional<String> getPrivateKey()
```
chenjian2664, line 94: The implementation logic looks good to me, but I think it is better if we could add the configuration into a subclass of `CredentialConfig`, i,e `SnowflakeCredentialConfig`, adding such dedicate credential config class allowing extending easily in the future, and we could do the validation within such class.

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/EnvMultinodeSnowflakeWithPrivateKey.java
```diff
@@ -0,0 +1,61 @@
+/*
```
ebyhr: Why do we add a product test instead of integration test? https://trino.io/docs/current/develop/tests.html#avoid-product-tests

parthshyara: The product test was added for parity with existing testing behaviour of Snowflake Suite. Snowflake credentials testing does require a snowflake server too. In that case, this would make sense to be added.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
parthshyara: In that case, we need to extend CredentialProviderModule and implement `getConnectorFactories` for Snowflake separately. Another approach is we need find all subclasses of CredentialConfig and bind them in `CredentialProviderModule`.
If we want to make this config snowflake specific, we can do that and bind in SnowflakeClientModule but then it wont need to extend `CredentialConfig` since those are not being directly used in snowflake module and passed via CredentialProvider.

Thoughts?

.github/workflows/ci.yml
```diff
@@ -515,7 +514,8 @@ jobs:
             - { modules: plugin/trino-resource-group-managers }
             - { modules: plugin/trino-singlestore }
             - { modules: plugin/trino-snowflake }
-            - { modules: plugin/trino-snowflake, profile: cloud-tests }
+            - { modules: plugin/trino-snowflake, profile: cloud-tests, auth: private-key }
+            - { modules: plugin/trino-snowflake, profile: cloud-tests, auth: password }
```
ebyhr: There is no need to separate a profile. Please merge them into one.

parthshyara: Both the profiles require different set of env variables. I tried to follow existing pattern of one maven test per CI step. Do you recommend combining both env vars into one and run two separate maven commands in the same CI step?

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
chenjian2664: What about adding implementation of `CredentialPropertiesProvider` for it?

parthshyara: That makes sense. But the `DriverConnectionFactory.Builder` takes in a `credentialProvider` provider in the initialization which is used to initializes `DefaultCredentialPropertiesProvider`. If we create snowflake specific credential properties provider, we need to inject `credentialProvider` there and this will lead to redundancy, since we are already injecting it in the `getConnectionFactory` of `SnowflakeClientModule`.

According to me, we have two options to gaurantee one source for credentials using credentialPropertiesProvider.
1. Pass `EmptyCredentialsProvider` in the `DriverConnectionFactory.Builder.init`
2. Create another constructor for `DriverConnectionFactory.Builder` with credentialPropertiesProvider.

What do you suggest?

parthshyara: @chenjian2664 Please suggest.

hpopuri2:
> ## Description
> * Adds support for the privateKey authentication property in the Snowflake connector.
> * The Snowflake JDBC driver already supports this property, but Trino previously had no way to pass it through.


@parthshyara  I tried hardcoding the  private key as a JDBC URL parameter(successful in creating a catalog), but it’s failing while querying because the Snowflake driver expects a java.security.PrivateKey object rather than a string. It looks like passing a local file path via private_key_file is mandatory for this to work in Trino.

parthshyara:
@hpopuri2 You are right. If you want to pass the credentials in the URL param, file is mandatory. But if you set this as a connection property, the snowflake driver accepts it and is able to authenticate. Here's the [relevant documentation](https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-configure?utm_source=chatgpt.com#privatekey-property-in-connection-properties) for the same.

parthshyara:
@chenjian2664 @ebyhr Can you please review the PR?

parthshyara:
Tracked on #28562

gdoron:
@findepi @ebyhr with snowflake already deprecated password authentication, and **just about to block it for good**, we really need to have this feature.
I'm not sure how other companies do not ring the bells on this.
Do you want me to pick up this PR from where the author stopped engaging?

[Reference announcement ](https://docs.snowflake.com/en/user-guide/security-mfa-rollout)

JGustavo0:
> @findepi @ebyhr with snowflake already deprecated password authentication, and **just about to block it for good**, we really need to have this feature. I'm not sure how other companies do not ring the bells on this. Do you want me to pick up this PR from where the author stopped engaging?
> 
> [Reference announcement ](https://docs.snowflake.com/en/user-guide/security-mfa-rollout)

Totally agree, thanks for your push here.

parthshyara:
FWIW, I did push multiple times on this PR for review after promptly resolving the comments and then re-opening another #28562 after this getting closed due to inactivity. I also pinged on the trino slack channel for review. Didn't get any response. Ultimately, built a plugin internally and got unblocked. The PRs were well tested then and verified working. Happy to contribute again on this.

cc: @chenjian2664 @ebyhr @findepi

gdoron:
@wendigo @dain @findepi @byhr this is becoming very urgent to fix, even beside the real security issue.
Everyone's connectors to snowflake are going to break in a month. Can we please get your attention to this?

gdoron:
I want to hope the maintainers for some reason are not getting the notifications, as I know can happen on old issues.
FYI, I've opened a new issue: https://github.com/trinodb/trino/issues/30310
If that would go quiet as well, we would complete our phasing out trino usage in my company.

.github/workflows/ci.yml
ebyhr: > Do you recommend combining both env vars into one and run two separate maven commands in the same CI step?

Yes and no. No need to separate a maven command. See Iceberg `cloud-tests` profile for example.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
ebyhr:
```suggestion
    @Config("snowflake.connection-private-key")
```

ebyhr, line 100: Add `@ConfigDescription` annotation.

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/EnvMultinodeSnowflakeWithPrivateKey.java
ebyhr: Disagreed, please remove it. We want to avoid new product tests as much as possible.

.github/workflows/ci.yml
```diff
@@ -675,16 +675,16 @@ jobs:
         env:
           SNOWFLAKE_URL: ${{ vars.SNOWFLAKE_URL }}
           SNOWFLAKE_USER: ${{ vars.SNOWFLAKE_USER }}
-          SNOWFLAKE_PASSWORD: ${{ secrets.SNOWFLAKE_PASSWORD }}
```
ebyhr, line 680: Why did you remove SNOWFLAKE_PASSWORD?

parthshyara: Ack. I was rebasing and cleaning up profiles. I am working on adding it similar to Iceberg tests.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
```diff
@@ -116,4 +121,17 @@ protected static void setOutputProperties(Properties properties)
...
+            return keyFactory.generatePrivate(keySpec);
+        }
+        catch (Exception e) {
+            throw new TrinoException(NOT_SUPPORTED, "Unable to parse provided private key", e);
```
ebyhr: NOT_SUPPORTED error code should be used when we plan to support it in the future.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
```diff
@@ -116,4 +121,17 @@ protected static void setOutputProperties(Properties properties)
         // Don't treat decimals as bigints as they may overflow
         properties.setProperty("JDBC_TREAT_DECIMAL_AS_INT", "FALSE");
     }
+
+    protected static PrivateKey parsePrivateKey(String privateKey)
```
ebyhr: Please verify an invalid PrivateKey in `TestSnowflakePlugin`.

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/SnowflakeQueryRunner.java
```diff
@@ -44,11 +45,24 @@ private SnowflakeQueryRunner() {}
...
+        return baseBuilder()
+                .addConnectorProperty("connection-password", requiredNonEmptySystemProperty("snowflake.test.server.password"));
+    }
+
+    public static Builder builderWithPrivateKey()
+    {
+        // Key-pair (RSA private key) authentication
+        return baseBuilder()
+                .addConnectorProperty("snowflake.connection-private-key", requiredNonEmptySystemProperty("snowflake.test.server.private-key"));
+    }
```
ebyhr: Remove `builderWithPrivateKey` method and move each `addConnectorProperty` to the caller side.

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeConnectorSmokeTestWithPrivateKey.java
```diff
@@ -0,0 +1,52 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.snowflake;
+
+import io.trino.plugin.jdbc.BaseJdbcConnectorSmokeTest;
+import io.trino.testing.QueryRunner;
+import io.trino.testing.TestingConnectorBehavior;
+
+public class TestSnowflakeConnectorSmokeTestWithPrivateKey
```
ebyhr: Rename to TestSnowflakeWithPrivateKeyConnectorSmokeTest. See ensureTestNamingConvention.

ebyhr:
/test-with-secrets sha=681f5076136b8126577ebc761a1c62000d199ec5

ebyhr:
```
TestSnowflakeWithPrivateKeyConnectorSmokeTest.: plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java#L51
System property snowflake.test.server.private-key must be non-empty
```

Please share the expected key on Slack DM. https://trino.io/slack.html

parthshyara:
@ebyhr A new RSA key-pair needs to be created and added to the integration test's snowflake user. And then the relevant private key needs to be added to `SNOWFLAKE_PRIVATE_KEY` Github Secret. I dont have access to either. Reference: https://github.com/trinodb/trino/pull/26519#discussion_r2333581972.

It will be more secure if you can generate the pair. I can share the verified commands I used during testing.

```
# Generate private key
openssl genrsa 2048 | openssl pkcs8 -topk8 -inform PEM -out rsa_key.p8 -nocrypt

# Generate public key
openssl rsa -in rsa_key.p8 -pubout -out rsa_key.pub

# Extract public key content (remove headers and format as single line)
cat rsa_key.pub | grep -v "BEGIN\|END" | tr -d '\n'
```

wendigo:
I've asked @dprophet for help

ebyhr:
@parthshyara Please squash commits into one.

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
```diff
@@ -116,4 +122,17 @@ protected static void setOutputProperties(Properties properties)
...
+    {
+        try {
+            byte[] keyBytes = Base64.getDecoder().decode(privateKey);
+            PKCS8EncodedKeySpec keySpec = new PKCS8EncodedKeySpec(keyBytes);
+            KeyFactory keyFactory = KeyFactory.getInstance("RSA");
+            return keyFactory.generatePrivate(keySpec);
+        }
+        catch (Exception e) {
+            throw new TrinoException(GENERIC_USER_ERROR, "Unable to parse provided private key", e);
```
dprophet: This is a really broad exception. Too broad.

The realistic failures are IllegalArgumentException (bad Base64), NoSuchAlgorithmException, and InvalidKeySpecException. Trino generally avoids catching bare Exception;
prefer:
  catch (GeneralSecurityException | IllegalArgumentException e)

This also stops it from masking unexpected RuntimeExceptions as user errors.

Also, isnt GENERIC_USER_ERROR usually for user queries? Isnt CONFIGURATION_INVALID a better error?

dprophet:
https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-parameters#private_key_pwd

The driver does support encrypted keys. Only unencrypted PKCS#8 keys are handled?

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeClientModule.java
parthshyara: Done, made both changes. On the error code I went with `CONFIGURATION_INVALID` like you suggested. One note is that since it's an `INTERNAL_ERROR`, a bad key passed through a dynamic `CREATE CATALOG` will surface as an internal error rather than a user error. I made the change anyway since that matches how Trino handles other invalid config values as per my exploration.

parthshyara:
@dprophet Those parameters are likely a new addition on snowflake's side. When I first created this PR, only file-path based approach was supporting password (as mentioned in the [original issue here](https://github.com/trinodb/trino/issues/26514)). There's also [this doc](https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-configure#privatekey-property-in-connection-properties) which mentions that we need to decrypt the key ourselves using the Bouncy Castle Crypto APIs. That also mentions the deprecated fields and is now outdated which I had followed for this PR.

I suggest we merge this to unblock. We can then do a fast-follow to generate a separate PR to support the new fields and maybe replace this one. I can work on getting that PR up sometime this weekend. If you feel different, I can re-work this PR itself sooner. Snowflake will start blocking password based auth starting August, so sooner the better. Let me know your thoughts.

dprophet:
@wendigo Assistance was provided, as discussed

parthshyara:
@ebyhr @dprophet @wendigo Update: given the urgency, I found some time today to get the fast-follow PR up: #30512. It drops the in-connector parsing and passes the key to the driver via `private_key_base64`, plus an optional `snowflake.connection-private-key-password` for encrypted keys (`private_key_pwd`), so both encrypted and unencrypted keys are supported.

Since it supersedes this one, I'd suggest we close this PR and merge #30512 instead so we can land it before the August cutoff.

ebyhr:
@parthshyara We don't need a new PR. It loses the review comments history in this PR. Please simply update this PR.

ebyhr:
/test-with-secrets sha=640709f7a209109887ffcfc042e900fa836f6abf

parthshyara:
@ebyhr Done. Can you please trigger the tests again?

parthshyara:
For testing the encrypted flows, the `-nocrypt` needs to be removed from the key generation command shared above. We also need to add `SNOWFLAKE_PRIVATE_KEY_PASSWORD` to test the encrypted key based flow. If omitted, the test will continue as the non-encrypted flow.
cc: @dprophet @ebyhr 

Here's the command for brevity.
`openssl genrsa 2048 | openssl pkcs8 -topk8 -inform PEM -out rsa_key.p8 `

ebyhr:
/test-with-secrets sha=f9acf47d435f00d5fc0cf967de64575667396e6b

docs/src/main/sphinx/connector/snowflake.md
```diff
@@ -26,13 +26,20 @@ connection properties as appropriate for your setup:
 connector.name=snowflake
 connection-url=jdbc:snowflake://<account>.snowflakecomputing.com
 connection-user=root
-connection-password=secret
+snowflake.connection-private-key=private-key-content
```
dprophet, line 29: You might want to document both the username/password configurations, private key, and private encrypted private key.

Note that Password support should be a programmatic access token not a user password since Snowflake is deprecating password based logins.

Might also want to add a note that the private key is a base64 encoded WITHOUT headers.

parthshyara: @dprophet The details are already mentioned in the notes. The current note is as per a recommendation from a [previous comment](https://github.com/trinodb/trino/pull/26519#discussion_r2607274860) on this PR.

ebyhr:
```
[ERROR]   TestSnowflakeWithPrivateKeyConnectorSmokeTest.createQueryRunner:51 » IllegalArgument System property snowflake.test.server.private-key must be non-empty
```

parthshyara:
Hi @ebyhr, can you please generate a key-pair and update the Github Secrets and Snowflake Cloud Account user auth? Details mentioned here.
https://github.com/trinodb/trino/pull/26519#issuecomment-5084126674
https://github.com/trinodb/trino/pull/26519#issuecomment-5126251118

andrewa-habitat (approved):

ebyhr:
@parthshyara I don't have the permission to access the Snowflake environment. Please coordinate with @dprophet.

gdoron:
👋 thanks everyone for working on this.
Is there anything still blocking the PR? can we get this merged and released?

ebyhr:
@parthshyara Please fix check-commit-messages failure: 
```
  f9acf47d435f Support RSA private key authentication in Snowflake connector
    subject: 61 characters
```

ebyhr:
/test-with-secrets sha=ff21f6efc66275fdd42806a97370a60a16a9f14c

ebyhr:
@parthshyara Could you rebase on master branch to fix `.mvn/settings-ci.xml` file not found failure?

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java
```diff
@@ -0,0 +1,57 @@
...
+                 SUPPORTS_COMMENT_ON_COLUMN,
+                 SUPPORTS_CREATE_TABLE_WITH_COLUMN_COMMENT,
+                 SUPPORTS_MAP_TYPE,
+                 SUPPORTS_PREDICATE_PUSHDOWN_WITH_VARCHAR_EQUALITY,
+                 SUPPORTS_PREDICATE_PUSHDOWN_WITH_VARCHAR_INEQUALITY,
+                 SUPPORTS_PREDICATE_EXPRESSION_PUSHDOWN,
+                 SUPPORTS_ROW_TYPE,
+                 SUPPORTS_SET_COLUMN_TYPE,
+                 SUPPORTS_MERGE,
+                 SUPPORTS_ROW_LEVEL_UPDATE -> false;
```
ebyhr: Sort entries alphabetically.

testing/trino-testing-services/src/main/java/io/trino/testing/TestingProperties.java
```diff
@@ -70,4 +71,9 @@ public static String requiredNonEmptySystemProperty(String propertyName)
         checkArgument(!isNullOrEmpty(value), "System property %s must be non-empty", propertyName);
         return value;
     }
+
+    public static Optional<String> optionalSystemProperty(String propertyName)
+    {
+        return Optional.ofNullable(System.getProperty(propertyName));
+    }
```
ebyhr: I don't think this method is helpful. Inline please.

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java
```diff
@@ -0,0 +1,57 @@
...
+ */
+package io.trino.plugin.snowflake;
+
+import io.trino.plugin.jdbc.BaseJdbcConnectorSmokeTest;
+import io.trino.testing.QueryRunner;
+import io.trino.testing.TestingConnectorBehavior;
+
+import static io.trino.testing.TestingProperties.requiredNonEmptySystemProperty;
+
+public class TestSnowflakeWithPrivateKeyConnectorSmokeTest
```
ebyhr: https://trino.io/docs/current/develop/tests.html
```suggestion
final class TestSnowflakeWithPrivateKeyConnectorSmokeTest
```

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakePlugin.java
```diff
@@ -30,4 +30,31 @@ public void testCreateConnector()
...
+    public void testCreateConnectorWithPrivateKey()
+    {
+        Plugin plugin = new SnowflakePlugin();
+        ConnectorFactory factory = getOnlyElement(plugin.getConnectorFactories());
+        factory.create(
+                "test",
+                ImmutableMap.of(
+                        "connection-url", "jdbc:snowflake://test",
+                        "snowflake.connection-private-key", "dummy-private-key"),
+                new TestingConnectorContext()).shutdown();
```
ebyhr: Move `shutdown()` to the next line. Same for testCreateConnectorWithEncryptedPrivateKey.

plugin/trino-snowflake/src/test/java/io/trino/plugin/snowflake/TestSnowflakeWithPrivateKeyConnectorSmokeTest.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        SnowflakeQueryRunner.Builder builder = SnowflakeQueryRunner.builder()
+                .addConnectorProperty("snowflake.connection-private-key", requiredNonEmptySystemProperty("snowflake.test.server.private-key"));
+        TestingSnowflakeServer.TEST_PRIVATE_KEY_PASSWORD.ifPresent(password -> builder.addConnectorProperty("snowflake.connection-private-key-password", password));
+        return builder
+                .setInitialTables(REQUIRED_TPCH_TABLES)
+                .build();
```
ebyhr: No need to handle `TEST_PRIVATE_KEY_PASSWORD` as optional: 
```java
        return SnowflakeQueryRunner.builder()
                .addConnectorProperty("snowflake.connection-private-key", requiredNonEmptySystemProperty("snowflake.test.server.private-key"))
                .addConnectorProperty("snowflake.connection-private-key-password", requiredNonEmptySystemProperty("snowflake.test.server.private-key-password"))
                .setInitialTables(REQUIRED_TPCH_TABLES)
                .build();
```

plugin/trino-snowflake/src/main/java/io/trino/plugin/snowflake/SnowflakeConfig.java
```diff
@@ -86,4 +89,32 @@ public SnowflakeConfig setHttpProxy(String httpProxy)
...
+        this.privateKey = privateKey;
+        return this;
+    }
+
+    public Optional<String> getPrivateKeyPassword()
+    {
+        return Optional.ofNullable(privateKeyPassword);
+    }
+
+    @Config("snowflake.connection-private-key-password")
```
ebyhr: `snowflake.connection-private-key.passphrase`

ebyhr: Please ensure the passphrase is allowed only when private-key property exists.

ebyhr (commented):
Looks good except for comments.

ebyhr:
/test-with-secrets sha=cd7d0b004151c02186ca66028b625c98da1e78c1

ebyhr:
The added test isn't running because of https://github.com/trinodb/trino/issues/30823. We need to fix the timeout first.

gdoron:
@parthshyara @ebyhr
Hi mates,
Are you still on this? do you need help with this?
While I don't write Java, I'll happily contribute my God-given AI tokens to do whatever needed to push this to the finish line.

Please let me know if there is **anything** I can do to help.

parthshyara:
I haved shared the private and public keys for the snowflake test with @wendigo @ebyhr @dprophet over slack. Once they update the same and re-run the tests, we can merge this. cc: @vlad-lyutenko who has opened a similar PR for snowflake iceberg connector.

ebyhr:
Note that just rerunning CI is insufficient. We need to fix #30823 first.

PR #26525, opened 2025-09-01
Avoid fetching of statistics from DeterminePartitionCount in small cluster

raunaqmorarka:
## Description
When the max possible tasks count is close to the min partition count, it is worth paying the cost of fetching statistics to determine optimal partitions count

## Release notes

```markdown
## General
* Improve performance of simple queries in clusters with low nodes count. ({issue}`issuenumber`)
```


core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestDeterminePartitionCount.java
```diff
@@ -241,6 +241,33 @@ SELECT count(column_a) FROM table_with_stats_a group by column_b
                                                                 node(TableScanNode.class))))))));
     }
 
+    @Test
```
losipiuk, line 244: nit: you can add test to check if that is not triggered if we have FTE enabled.

raunaqmorarka:
/test-with-secrets sha=7ec454205c90180e60e225bcfc929e0f847da5c5

PR #26531, opened 2025-09-02
Simplify compiled page projection

raunaqmorarka:
## Description
Avoid constructing new PageProjectionWork object per page

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/gen/PageFunctionCompiler.java
+++ b/core/trino-main/src/main/java/io/trino/sql/gen/PageFunctionCompiler.java
@@ -61,7 +61,7 @@ import org.objectweb.asm.MethodTooLargeException;
 import org.weakref.jmx.Managed;
 import org.weakref.jmx.Nested;
 
-import java.lang.invoke.MethodHandle;
+import java.lang.reflect.Constructor;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -72,6 +72,7 @@ import java.util.function.Supplier;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Throwables.throwIfInstanceOf;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.airlift.bytecode.Access.FINAL;
 import static io.airlift.bytecode.Access.PRIVATE;
 import static io.airlift.bytecode.Access.PUBLIC;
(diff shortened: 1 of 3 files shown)


wendigo:
Do you have benchmarks?

raunaqmorarka:
> Do you have benchmarks?

I'm not expected much gain, its mainly to simplify the code

wendigo:
```
Error:    TestArrayFunctions.testArrayConstructor:62 » QueryFailed Failed to execute query; there may be too many columns used or expressions are too complex

```

raunaqmorarka:
> ```
> Error:    TestArrayFunctions.testArrayConstructor:62 » QueryFailed Failed to execute query; there may be too many columns used or expressions are too complex
> ```

Seems the max parameters allowed came down slightly just because of using a method parameter instead of class field to pass an argument. Fixed that now by avoiding that change

core/trino-main/src/main/java/io/trino/sql/gen/PageProjectionWork.java
```diff
@@ -13,9 +13,13 @@
...
+import io.trino.operator.project.SelectedPositions;
 import io.trino.spi.block.Block;
+import io.trino.spi.block.BlockBuilder;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.SourcePage;
 
 public interface PageProjectionWork
 {
-    Block process();
+    Block process(BlockBuilder builder, ConnectorSession session, SourcePage page, SelectedPositions selectedPositions);
```
wendigo: nit: I'd move ConnectorSession as a first argument

wendigo (approved):

core/trino-main/src/main/java/io/trino/sql/gen/PageFunctionCompiler.java
```diff
@@ -232,89 +240,96 @@ private ClassDefinition definePageProjectWorkClass(RowExpression projection, Cal
...
 
         Scope scope = method.getScope();
         Variable thisVariable = method.getThis();
         BytecodeBlock body = method.getBody();
 
-        Variable from = scope.declareVariable("from", body, thisVariable.getField(selectedPositions).invoke("getOffset", int.class));
-        Variable to = scope.declareVariable("to", body, add(thisVariable.getField(selectedPositions).invoke("getOffset", int.class), thisVariable.getField(selectedPositions).invoke("size", int.class)));
+        for (int i = 0; i < inputChannels.size(); i++) {
+            int channel = inputChannels.get(i);
+            body.append(thisVariable.setField(blockFields.get(i), page.invoke("getBlock", Block.class, constantInt(channel))));
```
lukasz-stec, line 301: Looks like those fields are never cleared, so between executions, the last used blocks will be referenced here and thus non-eligible for GC

raunaqmorarka: right, i was thinking that the PageProjection instance itself would go out of scope once the operator is done.
but on 2nd thought, operators can be blocked, how about we add a `clear` method to PageProjectionWork which nullifies the blocks ?

raunaqmorarka: https://github.com/trinodb/trino/pull/26546

Issue #26533, opened 2025-09-02
Improving contributor experience


dolfinus:
After working on #25704 for some time, and as a follow up for #26528, I want to leave a few notes about working on PRs as a new contributor.

For 200 lines of new code, there are over 40 comments in #25704, mostly related:
* Airlift validators, assertions
* Explicit null checks
* Code formatting (e.g. indentations, imports order)

It looks that there are a lot of rules for writing Trino code which maintainers are aware of, but new contributors don't. So review can take a really long time - maintainers manually check if these rules are applied for every commit (due to rebase policy), then contributors have to address these comments, and this feedback loop takes hours or days. It can be reduces just to few minutes of most of these checks can be run locally on contributor side.

My proposals are:
* Document Airlift configuration - how validators are applied, how to test them, what annotations could be used (`@NotEmpty`. `@AssertTrue` and so on). Currently Airlift has almost no documentation https://github.com/airlift/airlift, and adding new config is done using try-and-error method or copy-pasting of existing config options.
* If there are linters that contributor have to run before commiting changes, let's document them in https://trino.io/development/process. Or maybe add a pre-commit hooks which will run linters automatically before commit is even made.
* Add more checks to a linter, or add multiple linters, to check for rules required in this repo. Like adding explicit `requireNotNull` in every constructor, or making Java compiler warnings an actual error if it appers in new code. For now, even if `./mvnw verify` is passed, this doesn't guarantee that the code will be accepted in PR.
* Make test output less noisy. Even if linter or Java compiler shows a warning, some tests output (e.g. `io.trino.plugin.openlineage.TestOpenLineageEventListenerMarquezIntegration`) fills up the entire console window, and its very easy to track that there was some warning.
* Add a code formatter, or at least document it. Half of comments in PR above are indentation issues, which can be easily automated, so contributors don't have to do this manually. Also errors due to order of `import` statements are very annoying.
* Document format rules for config options and their description - `kebab-case`, no ending dots in docstrings, and so on.
* Document how to run tests for one specific project only - for example, `trino-openlineage` depends on multiple other Maven projects within a repo, which should be build and installed to local Maven repo first, and then contributor can run tests:
  ```bash
  ./mvnw -pl :trino-openlineage -am install -DskipTests
  ./mvnw -pl :trino-openlineage test
  ```
  This command will run tests for `trino-openlineage` and all its dependencies, which is slow:
  ```bash
  ./mvnw -pl :trino-openlineage -am test
  ```

[...]

PR #26536, opened 2025-09-02
Make trino-openlineage tests output less verbose

dolfinus:
## Description

Make `TestOpenLineageEventListenerMarquezIntegration` output less verbose by showing Marquez response body only on failed assertion.

Related to #26533

## Release notes

### DIFF
--- a/plugin/trino-openlineage/src/test/java/io/trino/plugin/openlineage/TestOpenLineageEventListenerMarquezIntegration.java
+++ b/plugin/trino-openlineage/src/test/java/io/trino/plugin/openlineage/TestOpenLineageEventListenerMarquezIntegration.java
@@ -13,7 +13,6 @@
  */
 package io.trino.plugin.openlineage;
 
-import io.airlift.log.Logger;
 import io.trino.SessionRepresentation;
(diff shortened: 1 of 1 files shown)


plugin/trino-openlineage/src/test/java/io/trino/plugin/openlineage/TestOpenLineageEventListenerMarquezIntegration.java
```diff
@@ -191,10 +190,13 @@ private void checkJobRegistration(HttpClient client, String expectedJobName)
 
             HttpResponse<String> responseJob = client.send(requestJob, HttpResponse.BodyHandlers.ofString());
 
-            logger.info(responseJob.body());
```
Praveen2112, line 194: It looks like `logger` is not used - Can we remove that variable ?

dolfinus: Yes. Thank you for fix

wendigo (approved):

Issue #25555, opened 2025-04-11
[Improvement] Use real dataset location in OpenLineage integration


dolfinus:
Currently OpenLineage returns dataset with fields like:
```
namespace: trino://localhost:8080
name: mycatalog.myschema.mytable
```

But Trino catalog is just alias for real dataset location. For example, if this is a table in Postgres, dataset should point to its real location:
```
namespace: postgres://localhost:5432
name: mydb.myschema.mytable
```

Or by linking catalog with real database location using `SYMLINK`:
```
namespace: trino://localhost:8080
name: mycatalog.myschema.mytable
symlinks:
  - type: LOCATION
    namespace: postgres://localhost:5432
    name: mydb.myschema.mytable
```

wendigo:
This is not possible to implement currently as the engine doesn't know the database connector is connecting to (this is part of the JDBC url that is a configuration detail of the catalog).

dolfinus:
As I see, [CatalogStore.getCatalogs](https://javadoc.io/static/io.trino/trino-spi/474/trino.spi/io/trino/spi/catalog/CatalogStore.html#getCatalogs()) allows to iterate over all catalogs, find specific one by name, and fetch all properties, including JDBC URL. Although, without `connectorName` attribute it will be hard to distinguish between jdbc catalog and other types. And also this is O(n) operation which can be replaced with O(1) with proper API.

Without information about real dataset location, companies using multiple tools to access the same table (Trino, Spark, Airflow operators and so on) will see one lineage tree for Trino:
```
Run1:
  namespace: trino://localhost:8080
  name: queryId
  inputs:
    - Dataset1:
        namespace: trino://localhost:8080
        name: pgcatalog.pgschema.pgtable
    - Dataset2:
        namespace: trino://localhost:8080
        name: mysqlcatalog.mysqlschema.mysqltable
```
and completely different tree for all other tools:
```
Run1:
  namespace: spark://localhost
  name: my-spark-session
  inputs:
    - Dataset1:
        namespace: postgres://localhost:5432
        name: pgdb.pgschema.pgtable
    - Dataset2:
        namespace: mysql://localhost:3306
        name: mysqldb.mysqlschema.mysqltable
Run2:
  namespace: airflow://localhost:8081
  name: somedag.sometask
  inputs:
    - Dataset2:
        namespace: mysql://localhost:3306
        name: mysqldb.mysqlschema.mysqltable
Run3:
  namespace: flink://localhost:8088
  name: somejob
[...]

PR #26542, opened 2025-09-03
WIP: Real dataset location

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
@@ -20,6 +20,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import io.trino.plugin.jdbc.expression.ParameterizedExpression;
 import io.trino.spi.connector.ColumnHandle;
+import io.trino.spi.connector.ConnectorTableLocation;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.expression.Constant;
 import io.trino.spi.predicate.TupleDomain;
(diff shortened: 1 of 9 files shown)


dolfinus:
For JDBC-based connectors this is fine, but there are other connectors with different connection url format. Hive uses `thrift://...`, Kafka uses list of `host:port` of brokers, Loki uses `http://`. In most cases these can be represented in string form, but formats are different for each connector. Also, Google Sheets don't have connection URL at all, so nothing to parse.

To handle these connection URLs differently, there should be at least `TableInfo.connectorName` field. `TableInfo.catalog` field contains catalog name, not connector name.

wendigo:
We can add a connector name to `TableInfo`. The location should be connector agnostic while being part of `TableInfo` (SPI implementations shouldn't leak there) so `Optional<String>` would need to do

dolfinus:
Sounds reasonable.

One more thing though - `TableInfo.location` can make users think that this represents something like `CREATE TABLE iceberg.mytable (...) LOCATION 's3://bucket/warehouse/mytable'`. So I suggest using `connectionURL`/`connectorAddress` naming instead.

wendigo:
@dolfinus that wouldn't make sense in the context of the lake connectors where you want s3://bucket/table/location

dolfinus:
I mean that these should be separated options. Table location for lake connectors is one of table properties called `location`, and Catalog connector URL is a different thing which should not be named `TableInfo.location`.

chenjian2664:
@dolfinus I was expecting the location to be within the table’s domain. I don't understand why we need to involve the catalog perspective

dolfinus:
@chenjian2664 Have you checked the issue #25555?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
```diff
@@ -38,6 +39,7 @@
 
 public final class JdbcTableHandle
         extends BaseJdbcConnectorTableHandle
+        implements ConnectorTableLocation
```
Praveen2112, line 42: Shouldn't we have a SPI specific API in `ConnectorMetadata` which would resolve location based on `ConnectorTableHandle` ?

Issue #15439, opened 2022-12-16
Extends Iceberg table stats API to allow publish data and stats atomically


findepi:
This is a proxy issue for https://github.com/apache/iceberg/issues/6442

See also TODOs in the code being added in https://github.com/trinodb/trino/pull/15441

findepi:
cc @rdblue

findepi:
thank you @raunaqmorarka !

PR #26543, opened 2025-09-03
Update data and statistics in single transaction in iceberg

raunaqmorarka:
## Description
Previous implementation was making two commits which were resulting in two metadata json files per update to iceberg table

fixes https://github.com/trinodb/trino/issues/15439

## Release notes

```markdown
## Iceberg
* Reduce amount of metadata generated in writes to iceberg tables. ({issue}`15439`)
```


ebyhr (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1529,42 +1528,27 @@ public Optional<ConnectorOutputMetadata> finishInsert(
...
-
-                commitTransaction(transaction, "update statistics on insert");
-            }
-            catch (Exception e) {
-                // Write was committed, so at this point we cannot fail the query
-                // TODO (https://github.com/trinodb/trino/issues/15439): it would be good to publish data and stats atomically
-                log.error(e, "Failed to save table statistics");
-            }
-            transaction = null;
+            long newSnapshotId = icebergTable.currentSnapshot().snapshotId();
```
findepi, line 1537: Is this a new snapshot id that's about to be committed?
if so, then we're good. However, as we're pulling it from "current snapshot", it might be unclear to the reader. Worth a code comment?

raunaqmorarka: yes, this snapshot id is the one generated after the `commitUpdate(appendFiles, session, "insert");`
i was hoping the variable name is clear enough
the nice thing here is that doing this any other way fails many tests

findepi: the variable name is clear, the rhs expression _looks_ like not matching. hence ask for a comment

maybe it's because i am not familiar enough with iceberg api. i was convinced (clearly wrong) snapshot id doesn't exist before it's committed

PR #26548, opened 2025-09-03
Expose /v1/integrations/trinoGateway/metrics

andythsu:
## Release notes

```markdown
## Section
```


core/trino-main/src/main/java/io/trino/memory/LocalMemoryManager.java
```diff
@@ -61,7 +61,11 @@ private void validateHeapHeadroom(NodeMemoryConfig config, long availableMemory)
 
     public MemoryInfo getInfo()
     {
-        return new MemoryInfo(OPERATING_SYSTEM_MX_BEAN.getAvailableProcessors(), memoryPool.getInfo());
+        double systemCpuLoad = 0.0;
+        if (OPERATING_SYSTEM_MX_BEAN instanceof com.sun.management.OperatingSystemMXBean) {
+            systemCpuLoad = ((com.sun.management.OperatingSystemMXBean) OPERATING_SYSTEM_MX_BEAN).getCpuLoad();
+        }
```
andythsu: looking for suggestions here. This could lead to false metrics if `OPERATING_SYSTEM_MX_BEAN instanceof com.sun.management.OperatingSystemMXBean` is always false. The `systemCpuLoad` will always be `0.0`, and trino-gateway will think there's no load on this cluster

wendigo:
Why another endpoint if public `v1/status` contains CPU load and available memory?

wendigo:
```
{
  "nodeId": "presto-master",
  "nodeVersion": {
    "version": "477-test-1-7-g515c8c0-dirty"
  },
  "environment": "test",
  "coordinator": true,
  "uptime": "8.73s",
  "externalAddress": "presto-master.docker.cluster",
  "internalAddress": "presto-master.docker.cluster",
  "memoryInfo": {
    "availableProcessors": 16,
    "pool": {
      "maxBytes": 1503238554,
      "reservedBytes": 0,
      "reservedRevocableBytes": 0,
      "queryMemoryReservations": {

      },
      "queryMemoryAllocations": {

      },
      "queryMemoryRevocableReservations": {

      },
      "taskMemoryReservations": {

      },
      "taskMemoryRevocableReservations": {

      },
      "freeBytes": 1503238554
    }
  },
  "processors": 16,
  "processCpuLoad": 0.000903614457831325,
  "systemCpuLoad": 0.142035287566504,
  "heapUsed": 244729616,
  "heapAvailable": 2147483648,
  "nonHeapUsed": 306181120
}
```

andythsu:
@wendigo as far as I understand `/v1/status` only retrieves the status "per node". It doesn't give the entire cluster's metrics. This PR aggregates the metrics so it returns the metrics "per cluster"

wendigo:
@andythsu how this PR addresses this need?

andythsu:
@wendigo all nodes are available on the coordinator node. We can easily sum all node's free bytes up

```
        long totalFreeBytes = clusterMemoryManager.getAllNodesMemoryInfo()
                .values()
                .stream()
                .flatMap(Optional::stream)
                .map(MemoryInfo::getPool)
                .mapToLong(MemoryPoolInfo::getFreeBytes)
                .sum();
```

core/trino-main/src/main/java/io/trino/server/GatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+
+    @Inject
+    public TrinoGatewayResource(ClusterMemoryManager clusterMemoryManager)
+    {
+        this.clusterMemoryManager = clusterMemoryManager;
+    }
+
+    @GET
+    @Path("metrics")
+    @ResourceSecurity(MANAGEMENT_READ)
```
andythsu: I'm not sure if we want to use `MANAGEMENT_READ` because this endpoint will be hit by a dummy user from gateway. Most likely it'll be a role account. We may not want this role account to have management privileges.

Since this endpoint is for integrations, maybe we can have something like `user A needs to have <integration> privilege in order to call this endpoint`. We can easily use this access type for future integrations as well. Not sure if any of the existing access types can achieve such thing.

RoeyoOgen: In my opinion the 'MANAGEMENT_READ' is totally fine :)

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+import io.trino.server.security.ResourceSecurity;
+import io.trino.spi.memory.MemoryPoolInfo;
+import jakarta.ws.rs.GET;
+import jakarta.ws.rs.Path;
+
+import java.util.Optional;
+
+import static io.trino.server.security.ResourceSecurity.AccessType.MANAGEMENT_READ;
+
+@Path("/v1/integrations/trinoGateway")
```
RoeyoOgen: maybe use kebab-case or snake_case for paths ?

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+    {
+        this.clusterMemoryManager = clusterMemoryManager;
+    }
+
+    @GET
+    @Path("metrics")
+    @ResourceSecurity(MANAGEMENT_READ)
+    public ClusterMetrics clusterMetrics()
+    {
+        long totalFreeBytes = clusterMemoryManager.getAllNodesMemoryInfo()
```
RoeyoOgen: You're calling clusterMemoryManager.getAllNodesMemoryInfo() twice, which results in unnecessary overhead and potential inconsistency if the state changes between calls.
PLS cache the result in a variable

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+                .values()
+                .stream()
+                .flatMap(Optional::stream)
+                .map(MemoryInfo::getPool)
+                .mapToLong(MemoryPoolInfo::getFreeBytes)
+                .sum();
+        double totalSystemLoad = clusterMemoryManager.getAllNodesMemoryInfo()
```
RoeyoOgen:
```suggestion
        double aggregatedSystemLoad = clusterMemoryManager.getAllNodesMemoryInfo()
```

core/trino-main/src/main/java/io/trino/server/GatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+    public ClusterMetrics clusterMetrics()
+    {
+        long totalFreeBytes = clusterMemoryManager.getAllNodesMemoryInfo()
+                .values()
+                .stream()
+                .flatMap(Optional::stream)
+                .map(MemoryInfo::getPool)
```
RoeyoOgen, line 51: can getPool() be null? if so: wrap with a null check to avoid potential NullPointerException

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+                .values()
+                .stream()
+                .flatMap(Optional::stream)
+                .mapToDouble(MemoryInfo::getSystemCpuLoad)
+                .sum();
+        return new ClusterMetrics(totalFreeBytes, totalSystemLoad);
+    }
+
+    public record ClusterMetrics(long totalFreeBytes, double totalSystemLoad)
```
RoeyoOgen: perhaps add documentation to clarify what each field means

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,62 @@
...
+    @Inject
+    public TrinoGatewayResource(ClusterMemoryManager clusterMemoryManager)
+    {
+        this.clusterMemoryManager = clusterMemoryManager;
+    }
+
+    @GET
+    @Path("metrics")
+    @ResourceSecurity(MANAGEMENT_READ)
+    public ClusterMetrics clusterMetrics()
```
RoeyoOgen: rename to getClusterMetrics (as it is actionable)

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
wendigo: Actually we should use a path that is more "generic" as this endpoint could be used for other user cases as well. These kind of endpoints over time become a de facto part of the protocol so it's design and shape should be considered carefully.

andythsu: other endpoints in trino codebase use camel case

core/trino-main/src/main/java/io/trino/server/GatewayResource.java
andythsu: no, `pool` won't be null. https://github.com/trinodb/trino/blob/77af5adef1b435328568a3ba0aa8e0459f19f130/core/trino-main/src/main/java/io/trino/memory/MemoryInfo.java#L34

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
andythsu: Sure

core/trino-main/src/main/java/io/trino/memory/LocalMemoryManager.java
Chaho12: We could log a warning
```suggestion
        } else {
            // Log a warning, or throw if this metric is critical
            log.warn("Could not retrieve system CPU load: OperatingSystemMXBean is not an instance of com.sun.management.OperatingSystemMXBean");
        }
```

dain: We could consider making this a JVM requirement we verify in `TrinoSystemRequirements`.  @electrum what do you think?

dain: Alternatively, we could make this `Optional` everywhere, and then the client of this endpoint would need to decide what to do when the statistic is not available.

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
dain: @wendigo there was a discussion on [slack](https://trinodb.slack.com/archives/CP1MUNEUX/p1756844431344589) about making a special endpoint for the gateway like we have for the UI.  The implication being that changes to the endpoint would be decided by the gateway team, and could be backwards incompatible if they decided, which they likely will not as they need compatiblity, but it would be their choice.

core/trino-main/src/main/java/io/trino/server/GatewayResource.java
dain: We added management read for just this kind of thing.  I allows the caller to read data that some may find sensitive, like the IPs of workers.  Honestly I would have make this stuff public, but people get concerned.

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,69 @@
...
+        return new ClusterMetrics(totalFreeBytes, aggregatedSystemLoad);
+    }
+
+    /**
+     * Returns the aggregated metrics from each node in Trino cluster
+     * @param totalFreeBytes the sum of free memory from each node in the cluster
+     * @param aggregatedSystemLoad the sum of system load from each node in the cluster
+     */
+    public record ClusterMetrics(long totalFreeBytes, double aggregatedSystemLoad)
+    {}
```
dain: This can be on one line

dain (commented):
This all seems fine to me, but @wendigo should signoff.

core/trino-main/src/main/java/io/trino/memory/LocalMemoryManager.java
andythsu: I think it's better to make it a jvm requirement if possible. It's also how it's used here: https://github.com/trinodb/trino/blob/17e7b42e59f97a38669ef8753b86e1eb14deb889/core/trino-main/src/main/java/io/trino/server/StatusResource.java#L62

electrum: That seems fine to me. We already require `com.sun.management.UnixOperatingSystemMXBean` and Trino has other implicit requirements on OpenJDK.

andythsu:
@wendigo Could you take another look at this PR?

core/trino-main/src/main/java/io/trino/node/DnsNodeInventoryModule.java
```diff
@@ -30,7 +30,6 @@ protected void setup(Binder binder)
         boolean coordinator = buildConfigObject(ServerConfig.class).isCoordinator();
         if (coordinator) {
             jaxrsBinder(binder).bind(AnnounceNodeResource.class);
-
```
wendigo, line 33: undo

core/trino-main/src/main/java/io/trino/server/CoordinatorModule.java
```diff
@@ -242,6 +242,9 @@ protected void setup(Binder binder)
 
         newExporter(binder).export(ClusterMemoryManager.class).withGeneratedName();
 
+        // metrics used by Trino Gateway
+        jaxrsBinder(binder).bind(TrinoGatewayResource.class);
```
wendigo: drop comment. rename class to GatewayResource.class

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
wendigo: let's make the endpont call just `gateway`

core/trino-main/src/main/java/io/trino/server/TrinoGatewayResource.java
```diff
@@ -0,0 +1,69 @@
...
+
+@Path("/v1/integrations/trinoGateway")
+public class TrinoGatewayResource
+{
+    private final ClusterMemoryManager clusterMemoryManager;
+
+    @Inject
+    public TrinoGatewayResource(ClusterMemoryManager clusterMemoryManager)
+    {
+        this.clusterMemoryManager = clusterMemoryManager;
```
wendigo: `requireNonNull(clusterMemoryManager, "clusterMemoryManager is null")`

core/trino-main/src/main/java/io/trino/server/GatewayResource.java
```diff
@@ -0,0 +1,69 @@
...
+                .sum();
+        return new ClusterMetrics(totalFreeBytes, aggregatedSystemLoad);
+    }
+
+    /**
+     * Returns the aggregated metrics from each node in Trino cluster
+     *
+     * @param totalFreeBytes the sum of free memory from each node in the cluster
+     * @param aggregatedSystemLoad the sum of system load from each node in the cluster
+     */
```
wendigo, line 65: record definition doesn't return anything so it should be rather "represents"

core/trino-main/src/main/java/io/trino/memory/LocalMemoryManager.java
```diff
@@ -63,7 +65,7 @@ private void validateHeapHeadroom(NodeMemoryConfig config, long availableMemory)
 
     public MemoryInfo getInfo()
     {
-        return new MemoryInfo(AVAILABLE_PROCESSORS.get(), memoryPool.getInfo());
+        return new MemoryInfo(AVAILABLE_PROCESSORS.get(), ((UnixOperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean()).getCpuLoad(), memoryPool.getInfo());
```
wendigo: I think that we should memoize with timeout system load (for 5 seconds)?

andythsu: Is there a reason why we should memoize this? I'm not an expert on how this library is implemented, but I'd guess it's trivial to retrieve this information. Also, I don't anticipate this endpoint being spammed in a short amount of time since this is mainly used by Gateway. So unless the user has 10k Gateways running at the same time and each of them calls this endpoint every 500ms, in which case I'd argue they should cache it on the gateway side, not trino, it should be a low-effort retrieval. WDYT?

core/trino-main/src/main/java/io/trino/node/DnsNodeInventoryModule.java
andythsu: hmm.. not sure what changed here because it looks fine in my local

andythsu:
@wendigo this is ready for another review!

core/trino-main/src/main/java/io/trino/server/GatewayResource.java
```diff
@@ -0,0 +1,70 @@
...
+                .flatMap(Optional::stream)
+                .map(MemoryInfo::getPool)
+                .mapToLong(MemoryPoolInfo::getFreeBytes)
+                .sum();
+        double aggregatedSystemLoad = memoryInfo
+                .values()
+                .stream()
+                .flatMap(Optional::stream)
+                .mapToDouble(MemoryInfo::getSystemCpuLoad)
+                .sum();
```
chenjian2664, line 59: What does this `sum` mean to user?

andythsu: it should aggregate all cpu load of workers  in a trino cluster and return the sum. It's not calculated perfectly but I'm asusming if the load is high the sum should be high

core/trino-main/src/main/java/io/trino/memory/LocalMemoryManager.java
wendigo: I'd memoize it regardless just to be on a safe side

wendigo (approved):

wendigo:
@andythsu I applied some changes, take a look and if you are ok with it, I can merge this PR

andythsu:
@wendigo is this the only change you made?
```
    private static final Supplier<Double> SYSTEM_CPU_LOAD = Suppliers
            .memoizeWithExpiration(() -> clamp(((UnixOperatingSystemMXBean) ManagementFactory.getOperatingSystemMXBean()).getCpuLoad(), 0.0, 1.0), 5, TimeUnit.SECONDS);
```
if so, I'm happy with this change. TIL `clamp` is cool

wendigo:
@andythsu https://github.com/trinodb/trino/compare/58537883e8797e3a95987aa4b6a93810f5203c85..aae18d323dba7f8e522cd5c80bf8833a5c535bd4

andythsu:
@wendigo thanks! huh weird I couldn't find that new commit. The diff LGTM!

ebyhr:
The PR title and description are stale - the actual endpoint is different. Please update it.

PR #26552, opened 2025-09-04
[BUG] All column chunks in row group should have same starting position

sopel39:
## Release notes

```markdown
## Section
```


raunaqmorarka:
> All column chunks in row group should have same starting position

These are byte positions in the file, every column chunk defines starting position of only its own pages and every column chunk is written in a continuous stream one after the other in a row group.

PR #25174, opened 2025-02-27
Feature/support column properties with jdbc connector

hqbhoho:
## Description
Add column properties for base jdbc connector to support auto_increment, default value and other column attribute. 

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Mysql
*  Add support create table with auto increment column
```

PR #26561, opened 2025-09-05
Support column properties with jdbc connector

hqbhoho:
## Description
ReOpening https://github.com/trinodb/trino/pull/25174 since it went stale.

Add column properties for base jdbc connector to support auto_increment  and other column attribute. 
When running SHOW CREATE TABLE include column properties present for these column attribute.

## Release notes

```markdown
## Mysql
*  Add support for creating tables with `auto_increment` column property.
```

### DIFF
--- a/docs/src/main/sphinx/connector/mysql.md
+++ b/docs/src/main/sphinx/connector/mysql.md
@@ -105,37 +105,6 @@ creates a catalog named `sales` using the configured connector.
 The connector supports {doc}`/admin/fault-tolerant-execution` of query
 processing. Read and write operations are both supported with any retry policy.
 
-## Table properties
-
-Table property usage example:
-
-```
-CREATE TABLE person (
-  id INT NOT NULL,
-  name VARCHAR,
-  age INT,
-  birthday DATE 
-)
-WITH (
-  primary_key = ARRAY['id']
-);
-```
-
-The following are supported MySQL table properties: 
-
-:::{list-table}
-:widths: 30, 10, 60
-:header-rows: 1
-
-* - Property name
-  - Required
-  - Description
-* - `primary_key`
-  - No
-  - The primary key of the table, can choose multi columns as the table primary key.
-    All key columns must be defined as `NOT NULL`.
-:::
-
 (mysql-type-mapping)=
 ## Type mapping
 
@@ -398,6 +367,37 @@ following features:
 
 (mysql-schema-and-table-management)=
 ### Schema and table management
+
+#### Table properties
+
+Table property usage example:
+
+```
+CREATE TABLE person (
+  id INT NOT NULL,
+  name VARCHAR,
+  age INT,
+  birthday DATE 
+)
+WITH (
+  primary_key = ARRAY['id']
+);
+```
+
+The following are supported MySQL table properties:
+
+:::{list-table}
+:widths: 30, 10, 60
+:header-rows: 1
+
+* - Property name
+  - Required
+  - Description
+* - `primary_key`
+  - No
+  - The primary key of the table, can choose multi columns as the table primary key.
+    All key columns must be defined as `NOT NULL`.
+:::
 
 #### Column properties
 


hqbhoho:
@ebyhr  @chenjian2664   Could you help review it? Thanks

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
```diff
@@ -40,11 +40,12 @@ public final class JdbcColumnHandle
     private final Type columnType;
     private final boolean nullable;
     private final Optional<String> comment;
+    private final boolean autoIncrement;
```
Praveen2112: Should we have this `autoIncrement` as a part of all JDBC connector ?`JdbcColumnHandle` is something which should be applicable to most of the JDBC based connectors - `autoIncrement` looks like something which would be specific to certain types of JDBC connector. Can we directly populate them in `ColumnMetadata` ?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -334,6 +334,7 @@ public List<JdbcColumnHandle> getColumns(ConnectorSession session, SchemaTableNa
                 Optional<ColumnMapping> columnMapping = toColumnMapping(session, connection, typeHandle);
                 log.debug("Mapping data type of '%s' column '%s': %s mapped to %s", schemaTableName, columnName, typeHandle, columnMapping);
                 boolean nullable = (resultSet.getInt("NULLABLE") != columnNoNulls);
+                boolean autoIncrement = "YES".equals(resultSet.getString("IS_AUTOINCREMENT"));
```
Praveen2112: Should we have something specific to MySQL client.

plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlMetadata.java
```diff
@@ -0,0 +1,100 @@
...
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static io.trino.plugin.jdbc.JdbcMetadata.getColumns;
+
+public class MySqlMetadata
```
Praveen2112: I think this should be handled as a part of `MySqlClient`

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
Praveen2112: How about Map<String, Object> as columnProperties.

hqbhoho: Thank you for your feedback, It sounds good, I will try later!

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
hqbhoho: done

plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlMetadata.java
hqbhoho: done

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -336,13 +336,16 @@ public List<JdbcColumnHandle> getColumns(ConnectorSession session, SchemaTableNa
                 boolean nullable = (resultSet.getInt("NULLABLE") != columnNoNulls);
                 // Note: some databases (e.g. SQL Server) do not return column remarks/comment here.
                 Optional<String> comment = Optional.ofNullable(emptyToNull(resultSet.getString("REMARKS")));
+                // Note: the default implementation returns an empty map, and each connector needs to provide its own implementation.
```
chenjian2664: The comments seem more appropriate at the method level. Could you move them there?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
```diff
@@ -169,6 +203,7 @@ public static final class Builder
         private Type columnType;
         private boolean nullable = true;
         private Optional<String> comment = Optional.empty();
+        private Map<String, Object> columnProperties = emptyMap();
```
chenjian2664: ImmutableMap ?

plugin/trino-mysql/src/test/java/io/trino/plugin/mysql/BaseMySqlConnectorTest.java
```diff
@@ -355,6 +355,130 @@ private void verifyTableDefinitionWithUnsupportedKey(String tableDefinition, Str
         }
     }
 
+    @Test
+    public void testCreateTableWithAutoIncrementColumn()
+    {
+        verifyCreateTableDefinition(
+                "(a tinyint NOT NULL WITH (auto_increment = true), b bigint, c bigint) WITH (primary_key = ARRAY['a'])",
```
chenjian2664, line 362: Does it allow specifying auto_increment as false?

plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlClient.java
```diff
@@ -487,15 +492,30 @@ protected List<String> createTableSqls(RemoteTableName remoteTableName, List<Str
...
             columnDefinitions.add("PRIMARY KEY (" + primaryKeys.stream().map(this::quoted).collect(joining(", ")) + ")");
         }
         return ImmutableList.of(format("CREATE TABLE %s (%s) COMMENT %s", quoted(remoteTableName), join(", ", columnDefinitions.build()), mysqlVarcharLiteral(tableMetadata.getComment().orElse(NO_COMMENT))));
     }
 
-    private static void verifyPrimaryKey(List<String> primaryKeys, List<ColumnMetadata> columns)
+    private static void verifyColumnMetadata(List<String> primaryKeys, List<ColumnMetadata> columns)
     {
+        Set<String> autoIncrementColumnNames = columns.stream()
+                .filter(column -> (boolean) column.getProperties().getOrDefault(AUTO_INCREMENT, false))
```
chenjian2664: I was thinking whether `column.getProperties().containsKey` could be used. I’m uncertain about the scenario where the property exists but its value is `false`

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
```diff
@@ -149,7 +165,25 @@ public long getRetainedSizeInBytes()
...
-                + jdbcTypeHandle.getRetainedSizeInBytes();
+                + jdbcTypeHandle.getRetainedSizeInBytes()
+                + estimatedSizeOf(columnProperties, SizeOf::estimatedSizeOf, JdbcColumnHandle::estimatedObjectSizeOf);
+    }
+
+    /**
+     * Returns the estimated size of a property value.
+     * <p>
+     * Supported types are the same as those accepted by
+     * {@link io.trino.metadata.PropertyUtil#toExpression}.
```
chenjian2664: Why accept the same as `PropertyUtil#toExpression` ?

hqbhoho:
> Wondering the `auto_increment` can be false?

Thank you for your feedback!

Currently, the `MySQLClient#getColumnProperties` method ignores the "IS_AUTOINCREMENT" property unless its value is explicitly "YES". Allowing the creation of tables in Trino using WITH (auto_increment = false) is unnecessary, as this configuration is meaningless in this context.

To address this, I will implement a validation check to prevent such table creation attempts and add corresponding test cases to ensure the functionality works as expected

plugin/trino-mysql/src/test/java/io/trino/plugin/mysql/BaseMySqlConnectorTest.java
hqbhoho: You are right, we need add a check to prevent the creation of tables in Trino using WITH (auto_increment = false).It is meaningless.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
hqbhoho: `SHOW CREATE TABLE`  will use `PropertyUtil#toExpression` parse property value to `Expression`. So here, I want to restrict the types of property values. I should put this check in `JdbcColumnHandle#Constructor ` and then `estimatedObjectSizeOf` only needs to support the legal types. Or do you have any better suggestions

docs/src/main/sphinx/connector/mysql.md
```diff
@@ -395,6 +365,72 @@ following features:
...
+:header-rows: 1
+
+* - Property name
+  - Required
+  - Description
+* - `auto_increment`
+  - No
+  - Auto generate a unique identity for new rows. There can be only one auto increment column
+    and must be defined as the first key. Only applies to integer types (`TINYINT`,
+    `SMALLINT`, `INTEGER`, `BIGINT`) column. The value must be set to `true`.
```
chenjian2664: Adding a description for the behavior when the value is set to `false`

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
```diff
@@ -294,4 +294,9 @@ default List<JdbcColumnHandle> getPrimaryKeys(ConnectorSession session, RemoteTa
     {
         return List.of();
     }
+
+    default ColumnMetadata toColumnMetadata(JdbcColumnHandle columnHandle)
```
chenjian2664, line 298: Since the `getColumnProperties` will hold column properties, is the change still necessary?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
```diff
@@ -149,7 +163,44 @@ public long getRetainedSizeInBytes()
...
+            case List<?> list -> estimatedSizeOf(list, JdbcColumnHandle::estimatedObjectSizeOf);
+            case Map<?, ?> map -> estimatedSizeOf(map, JdbcColumnHandle::estimatedObjectSizeOf, JdbcColumnHandle::estimatedObjectSizeOf);
+            case null -> 0L;
+            default -> throw new TrinoException(INVALID_COLUMN_PROPERTY, "Unsupported property value type: " + value.getClass().getName());
+        };
+    }
+
+    private static Map<String, Object> verifyColumnProperties(Map<String, Object> columnProperties)
+    {
+        if (columnProperties != null) {
```
chenjian2664: Is the columnProperties nullable, when?

plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlColumnProperties.java
```diff
@@ -0,0 +1,49 @@
...
+                .add(booleanProperty(
+                        AUTO_INCREMENT,
+                        "Auto generate a unique identity for new rows",
+                        null,
+                        false))
+                .build();
+    }
+
+    @Override
+    public List<PropertyMetadata<?>> getColumnProperties()
```
chenjian2664, line 45: Where is the usage of the method?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
hqbhoho: I think directly passing `columnProperties` from `JdbcColumnHandle` to `ColumnMetadata` might not be appropriate. Could there be a scenario where some attributes stored in `JdbcColumnHandle` are not intended to be passed to `ColumnMetadata`? Adding a `toColumnMetadata` method on the client side would allow clients to define custom conversion logic.

In the current scenario, I want to add `extraInfo` to `ColumnMetadata` based on `columnProperties`.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
hqbhoho: No, it's not nullable when use builder() to new , but constructor method is public. I will add a check.

plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlColumnProperties.java
hqbhoho: In `MySqlClientModule` will inject `ColumnPropertiesProvider` with  `bindColumnPropertiesProvider(binder, MySqlColumnProperties.class);`
`ColumnPropertyManager#getProperties` will be used when execute `create table`,`add column` or `show table table` and so on.

docs/src/main/sphinx/connector/mysql.md
hqbhoho: done

Issue #21046, opened 2024-03-13
Question: Is there a way to get MetaData in a Plugin for row based access control?


roshipro:
I am trying to implement a security layer for my database at row level access control in a plugin. For the responsible method the column names of the tables are not given, is there an easy way to get them?

Issue #1480, opened 2019-09-10
Row-level filtering and masking


martint:
Presto should support row-level filtering and data masking for more granular access controls to data.

The general approach would be for `SystemAccessControl`/`AccessControl` to provide filters/projections on the base table given an authorization identifier (user/role) that the query planner injects into the plan.

Some questions that need to be considered:
* Are the filters/masks provided as an expression or a reference to a function (e.g., a table function)?
* If an expression, is it provided as SQL text or IR?
* In the case of SQL text,
    * What's the "environment" under which expressions get evaluated (path, catalog, schema, authorization identifier)?
    * What happens if the SQL has free variables? Do they get bound to the enclosing scope if possible? Or is it treated as an error? (this would require some complex changes to the analyzer)
    * What happens if the SQL has semantic errors (references to missing columns, type mismatch, missing functions, etc)? It might be hard/impossible to report a user-friendly error.

There are some potential benefits of modeling the filter/masks as references to (polymorphic) table functions:
* The invocation semantics of functions in SQL is well-defined, especially around how the session and authorization contexts behave
* It can potentially support opaque implementations (where the engine is not exposed to the filtering/masking logic) or transparent implementations by leveraging SQL-defined functions and related machinery.

Relates to https://github.com/prestosql/presto/issues/18

l0yu:
> Presto should support row-level filtering and data masking for more granular access controls to data.
> 
> The general approach would be for `SystemAccessControl`/`AccessControl` to provide filters/projections on the base table given an authorization identifier (user/role) that the query planner injects into the plan.
> 
> Some questions that need to be considered:
> 
>     * Are the filters/masks provided as an expression or a reference to a function (e.g., a table function)?
> 
>     * If an expression, is it provided as SQL text or IR?
> 
>     * In the case of SQL text,
>       
>       * What's the "environment" under which expressions get evaluated (path, catalog, schema, authorization identifier)?
>       * What happens if the SQL has free variables? Do they get bound to the enclosing scope if possible? Or is it treated as an error? (this would require some complex changes to the analyzer)
>       * What happens if the SQL has semantic errors (references to missing columns, type mismatch, missing functions, etc)? It might be hard/impossible to report a user-friendly error.
> 
> 
> There are some potential benefits of modeling the filter/masks as references to (polymorphic) table functions:
> 
>     * The invocation semantics of functions in SQL is well-defined, especially around how the session and authorization contexts behave
> 
[...]

PR #26582, opened 2025-09-08
Enable getRowFilter to be column aware

huw0:
### Summary
Enable getRowFilter to be column aware.

### Detail

Presently the access control interfaces provide `getRowFilter` with the security context, catalog name and schema name.

This works well when adding the row filter based on a restricted list of rules (`FileBasedAccessControl`) or where the metadata is known elsewhere (OPA or Ranger).

However this does not allow for filters to be dynamically added based on the columns.

For example - _whenever a table has a `security_group` column, add a filter to enable row-level security against the user's current groups._

To enable this use case, this PR modifies the `getRowFilter` so that the `List<ColumnSchema>` is also passed in as a parameter, matching `getColumnMasks`.

### Implementation Notes

* **Backwards Compatibility** - everything in `trino-main` calls the new method. However the old method is deprecated but called by the default implementation of the new method within the interfaces. Hopefully this preserves backwards compatibility?
* **Testing** - where is best to add coverage of this functionality?
* **FileBasedAccessControl** - I propose a follow on PR to make use of this functionality within `FileBasedAccessControl` and associated documentation.

#1480 - original implementation of row filtering
#21046 - a question around this functionality

## Release notes

```markdown
## SPI
* Enable row filtering to be column aware ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/security/AccessControl.java
+++ b/core/trino-main/src/main/java/io/trino/security/AccessControl.java
@@ -641,7 +641,7 @@ public interface AccessControl
      */
     void checkCanFastForwardBranch(SecurityContext context, QualifiedObjectName tableName, String sourceBranchName, String targetBranchName);
 
-    default List<ViewExpression> getRowFilters(SecurityContext context, QualifiedObjectName tableName)
+    default List<ViewExpression> getRowFilters(SecurityContext context, QualifiedObjectName tableName, List<ColumnSchema> columns)
     {
         return ImmutableList.of();
     }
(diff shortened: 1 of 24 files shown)


huw0:
PR amended from draft to ready. The failing check is due to a Github actions timeout, rather than an actual failure.

I want to draw attention to my question in the original summary around adding a test and where this would be best placed?
There are currently some tests for row filtering within `TestEventListenerBasic` or `TestAccessControl`? Probably the latter is the most sensible.

martint:
The expectation is that the access control knows about the schema of the tables that are being secured. It is **not** the responsibility of the engine to tell the access control plugin what the schema (columns) looks like.

Therefore, in the example you provided:

> For example - whenever a table has a security_group column, add a filter to enable row-level security against the user's current groups.

the access control implementation should obtain the table metadata to see if the table has a security_group column and return the necessary filters.

PR #26584, opened 2025-09-08
Change the visibility of shared variables in TimeSharingTaskHandle from protected to private

SongChujun:
## Description

Keeping the shared variable here to have protected access makes it possible that its values get mutated outside of this class without proper synchronization

## Release notes


ebyhr:
Could you shorten the commit title? https://trino.io/development/process#pull-request-and-commit-guidelines-
> Limit the subject line to 50 characters.

PR #26576, opened 2025-09-08
Add document for `mongodb.implicit-row-field-prefix` in Mongodb

chenjian2664:
## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

PR #26585, opened 2025-09-08
Fix reading mixed-type array in Mongodb

chenjian2664:
## Description

Currently, reading mixed-type array type in Mongo fails with message:
```
ClassCastException on nested arrays: ArrayList cannot be cast to BSON Document
```

Reproduce by below test:
```
    void testMongoMixedTypeArrayType()
    {
        String schema = getSession().getSchema().orElseThrow();
        String table = "test_mixed_array_" + randomNameSuffix();
        MongoDatabase db = client.getDatabase(schema);

        db.createCollection(table);
        db.getCollection(table)
                .insertOne(new Document("mixed_array_col", ImmutableList.of(1, "two", 3.0, new Document("nested_arr", ImmutableList.of(4, 5)))));

        assertThat(query("SHOW COLUMNS FROM " + table))
                .skippingTypesCheck()
                .matches("VALUES " +
                         "('mixed_array_col', 'row(_pos1 bigint, _pos2 varchar, _pos3 double, _pos4 row(nested_arr array(bigint)))', '', '')");

        assertThat(query("SELECT mixed_array_col._pos1, mixed_array_col._pos2, mixed_array_col._pos3 FROM " + table))
                .matches("VALUES (BIGINT '1', VARCHAR 'two', DOUBLE '3.0')");
        assertThat(query("SELECT mixed_array_col._pos4.nested_arr[1], mixed_array_col._pos4.nested_arr[2] FROM " + table))
                .matches("VALUES (BIGINT '4', BIGINT '5')");

        assertUpdate("DROP TABLE " + table);
    }
```

Revert https://github.com/trinodb/trino/pull/26576 since we are going to hide the configuration `implicit-row-field-prefix`

## Release notes

```markdown
## MongoDB
* Fix failure when reading array type with different element types. ({issue}`26585`)
```

### DIFF
--- a/plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoPageSource.java
+++ b/plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoPageSource.java
@@ -66,6 +66,7 @@ import static io.trino.plugin.mongodb.MongoSession.COLLECTION_NAME;
 import static io.trino.plugin.mongodb.MongoSession.DATABASE_NAME;
 import static io.trino.plugin.mongodb.MongoSession.ID;
 import static io.trino.plugin.mongodb.ObjectIdType.OBJECT_ID;
+import static io.trino.plugin.mongodb.TypeUtils.getImplicitRowFieldIndex;
 import static io.trino.plugin.mongodb.TypeUtils.isJsonType;
 import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static io.trino.spi.type.BigintType.BIGINT;
@@ -103,11 +104,13 @@ public class MongoPageSource
     private boolean finished;
 
     private final PageBuilder pageBuilder;
+    private final String implicitPrefix;
 
     public MongoPageSource(
             MongoSession mongoSession,
             MongoTableHandle tableHandle,
-            List<MongoColumnHandle> columns)
+            List<MongoColumnHandle> columns,
+            String implicitPrefix)
     {
         this.columns = ImmutableList.copyOf(requireNonNull(columns, "columns is null"));
         this.columnTypes = columns.stream().map(MongoColumnHandle::type).collect(toList());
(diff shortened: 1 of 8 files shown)


plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoPageSource.java
```diff
@@ -383,7 +387,7 @@ else if (type instanceof RowType rowType) {
         output.appendNull();
     }
 
-    private static Object getColumnValue(Document document, MongoColumnHandle mongoColumnHandle)
+    private static Object getColumnValue(Document document, MongoColumnHandle mongoColumnHandle, String implicitPrefix)
```
ebyhr: nit: We could change to non-static method so we can use `implicitPrefix` in fields.

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoPageSource.java
```diff
@@ -392,17 +396,27 @@ private static Object getColumnValue(Document document, MongoColumnHandle mongoC
...
+            }
+            else {
+                checkArgument(value instanceof List<?>, "Unsupported dereference of %s in %s", value.getClass(), mongoColumnHandle);
+                List<?> arrayValue = (List<?>) value;
+                int arrayPosition = getImplicitRowFieldIndex(dereferenceName, implicitPrefix) - 1;
+                checkArgument(arrayPosition >= 0, "Invalid array position %s in %s", dereferenceName, mongoColumnHandle);
+                if (arrayPosition >= arrayValue.size()) {
+                    return null;
+                }
+                value = arrayValue.get(arrayPosition);
```
ebyhr, line 416: I'm a little wondering whether comparison with `mongodb.implicit-row-field-prefix` is safe or not.

The behavior isn't "implicit" after the initial schema inference. It physically stores _pos* fields in _schema collection. What if the config value is changed when _schema already contains the table definition with the older config value? 

If we don't want to handle such an edge case, we could hide the property (revert the recent documentation & add `@ConfigHidden`) not to encourage changing this property.

chenjian2664: Yes, I think you’re right. Comparing the prefix directly from the configuration is not always safe

ebyhr (approved):

PR #26596, opened 2025-09-09
Improve error messages for OpenID Metadata discovery

hashhar:
## Description

Improve error messages for OpenID Metadata discovery

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/security/oauth2/OidcDiscovery.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/oauth2/OidcDiscovery.java
@@ -93,9 +93,9 @@ public class OidcDiscovery
         if (statusCode != OK.code()) {
             // stop on any client errors other than REQUEST_TIMEOUT and TOO_MANY_REQUESTS
             if (statusCode < 400 || statusCode >= 500 || statusCode == REQUEST_TIMEOUT.code() || statusCode == TOO_MANY_REQUESTS.code()) {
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/server/security/oauth2/OidcDiscovery.java
```diff
@@ -93,9 +93,9 @@ private OAuth2ServerConfig parseConfigurationResponse(HTTPResponse response)
         if (statusCode != OK.code()) {
             // stop on any client errors other than REQUEST_TIMEOUT and TOO_MANY_REQUESTS
             if (statusCode < 400 || statusCode >= 500 || statusCode == REQUEST_TIMEOUT.code() || statusCode == TOO_MANY_REQUESTS.code()) {
-                throw new RuntimeException("Invalid response from OpenID Metadata endpoint: " + statusCode);
+                throw new RuntimeException("Invalid response from OpenID Metadata endpoint " + response.getLocation() + ": " + statusCode);
             }
-            throw new IllegalStateException(format("Invalid response from OpenID Metadata endpoint. Expected response code to be %s, but was %s", OK.code(), statusCode));
+            throw new IllegalStateException(format("Invalid response from OpenID Metadata endpoint " + response.getLocation() + ". Expected response code to be %s, but was %s", OK.code(), statusCode));
```
chenjian2664, line 98: It looks like both string concatenation and formatting are used here. Would it be better to unify them?

PR #26599, opened 2025-09-10
Update supported Exasol DB version to 8.34.0

skyglass:
## Description

Update supported Exasol DB version to 8.34.0
Earlier versions ≥ 7.1 are also supported, but not regularly tested.

## Release notes


ebyhr:
@skyglass You don't need to rebase on master unless there are conflicts.

PR #26603, opened 2025-09-10
Add support for JOIN pushdown in Exasol Connector

skyglass:
## Description

* Add support for JOIN pushdown in Exasol Connector

* Implement toWriteMapping in ExasolClient for DECIMAL type
* Implement basic convertPredicate in ExasolClient

### DIFF
--- a/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
+++ b/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
@@ -16,6 +16,7 @@ package io.trino.plugin.exasol;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
 import io.airlift.slice.Slices;
+import io.trino.plugin.base.expression.ConnectorExpressionRewriter;
 import io.trino.plugin.base.mapping.IdentifierMapping;
 import io.trino.plugin.jdbc.BaseJdbcClient;
 import io.trino.plugin.jdbc.BaseJdbcConfig;
@@ -35,6 +36,9 @@ import io.trino.plugin.jdbc.SliceReadFunction;
 import io.trino.plugin.jdbc.SliceWriteFunction;
 import io.trino.plugin.jdbc.WriteFunction;
 import io.trino.plugin.jdbc.WriteMapping;
+import io.trino.plugin.jdbc.expression.JdbcConnectorExpressionRewriterBuilder;
+import io.trino.plugin.jdbc.expression.ParameterizedExpression;
+import io.trino.plugin.jdbc.expression.RewriteIn;
 import io.trino.plugin.jdbc.logging.RemoteQueryModifier;
 import io.trino.spi.TrinoException;
 import io.trino.spi.connector.AggregateFunction;
@@ -43,6 +47,8 @@ import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ColumnPosition;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.expression.ConnectorExpression;
+import io.trino.spi.type.DecimalType;
 import io.trino.spi.type.Type;
 
 import java.sql.Connection;
@@ -65,6 +71,8 @@ import static io.trino.plugin.jdbc.StandardColumnMappings.defaultCharColumnMappi
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultVarcharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.doubleColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.integerColumnMapping;
+import static io.trino.plugin.jdbc.StandardColumnMappings.longDecimalWriteFunction;
+import static io.trino.plugin.jdbc.StandardColumnMappings.shortDecimalWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.smallintColumnMapping;
 import static io.trino.plugin.jdbc.TypeHandlingJdbcSessionProperties.getUnsupportedTypeHandling;
 import static io.trino.plugin.jdbc.UnsupportedTypeHandling.CONVERT_TO_VARCHAR;
@@ -84,6 +92,7 @@ public class ExasolClient
             .add("EXA_STATISTICS")
             .add("SYS")
             .build();
+    private final ConnectorExpressionRewriter<ParameterizedExpression> connectorExpressionRewriter;
 
     @Inject
     public ExasolClient(
@@ -94,6 +103,25 @@ public class ExasolClient
             RemoteQueryModifier queryModifier)
     {
         super("\"", connectionFactory, queryBuilder, config.getJdbcTypesMappedToVarchar(), identifierMapping, queryModifier, false);
+        // Basic implementation required to enable JOIN pushdown support
+        // It is covered by "testJoinpushdown" integration tests.
+        // More detailed test case scenarios are covered by Unit tests in "TestExasolClient"
+        this.connectorExpressionRewriter = JdbcConnectorExpressionRewriterBuilder.newBuilder()
+                .addStandardRules(this::quoted)
+                .add(new RewriteIn())
+                .map("$equal(left, right)").to("left = right")
+                .map("$not_equal(left, right)").to("left <> right")
+                // Exasol doesn't support "IS NOT DISTINCT FROM" expression,
+                // so "$identical(left, right)" is rewritten with equivalent "(left = right OR (left IS NULL AND right IS NULL))" expression
+                .map("$identical(left, right)").to("(left = right OR (left IS NULL AND right IS NULL))")
+                .map("$less_than(left, right)").to("left < right")
+                .map("$less_than_or_equal(left, right)").to("left <= right")
+                .map("$greater_than(left, right)").to("left > right")
+                .map("$greater_than_or_equal(left, right)").to("left >= right")
+                .map("$not($is_null(value))").to("value IS NOT NULL")
+                .map("$not(value: boolean)").to("NOT value")
+                .map("$is_null(value)").to("value IS NULL")
+                .build();
     }
 
     @Override
@@ -196,10 +224,15 @@ public class ExasolClient
     }
 
     @Override
+    public Optional<ParameterizedExpression> convertPredicate(ConnectorSession session, ConnectorExpression expression, Map<String, ColumnHandle> assignments)
+    {
+        return connectorExpressionRewriter.rewrite(session, expression, assignments);
+    }
+
+    @Override
     protected boolean isSupportedJoinCondition(ConnectorSession session, JdbcJoinCondition joinCondition)
     {
-        // Deactivated because test 'testJoinPushdown()' requires write access which is not implemented for Exasol
-        return false;
+        return true;
     }
 
     @Override
@@ -311,7 +344,15 @@ public class ExasolClient
     @Override
     public WriteMapping toWriteMapping(ConnectorSession session, Type type)
     {
-        throw new TrinoException(NOT_SUPPORTED, "This connector does not support writing");
+        if (type instanceof DecimalType decimalType) {
+            String dataType = "decimal(%s, %s)".formatted(decimalType.getPrecision(), decimalType.getScale());
+            if (decimalType.isShort()) {
+                return WriteMapping.longMapping(dataType, shortDecimalWriteFunction(decimalType));
+            }
+            return WriteMapping.objectMapping(dataType, longDecimalWriteFunction(decimalType));
+        }
+
+        throw new TrinoException(NOT_SUPPORTED, "Unsupported column type: " + type.getDisplayName());
     }
 
     @Override
--- a/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
+++ b/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
@@ -16,6 +16,7 @@ package io.trino.plugin.exasol;
 import com.google.common.collect.ImmutableSet;
 import com.google.inject.Inject;
 import io.airlift.slice.Slices;
+import io.trino.plugin.base.expression.ConnectorExpressionRewriter;
 import io.trino.plugin.base.mapping.IdentifierMapping;
 import io.trino.plugin.jdbc.BaseJdbcClient;
 import io.trino.plugin.jdbc.BaseJdbcConfig;
@@ -38,6 +39,9 @@ import io.trino.plugin.jdbc.SliceReadFunction;
 import io.trino.plugin.jdbc.SliceWriteFunction;
 import io.trino.plugin.jdbc.WriteFunction;
 import io.trino.plugin.jdbc.WriteMapping;
+import io.trino.plugin.jdbc.expression.JdbcConnectorExpressionRewriterBuilder;
+import io.trino.plugin.jdbc.expression.ParameterizedExpression;
+import io.trino.plugin.jdbc.expression.RewriteIn;
 import io.trino.plugin.jdbc.logging.RemoteQueryModifier;
 import io.trino.spi.TrinoException;
 import io.trino.spi.connector.AggregateFunction;
@@ -46,6 +50,8 @@ import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ColumnPosition;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.expression.ConnectorExpression;
+import io.trino.spi.type.DecimalType;
 import io.trino.spi.type.LongTimestamp;
 import io.trino.spi.type.TimestampType;
 import io.trino.spi.type.Type;
@@ -78,6 +84,8 @@ import static io.trino.plugin.jdbc.StandardColumnMappings.doubleColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.fromLongTrinoTimestamp;
 import static io.trino.plugin.jdbc.StandardColumnMappings.fromTrinoTimestamp;
 import static io.trino.plugin.jdbc.StandardColumnMappings.integerColumnMapping;
+import static io.trino.plugin.jdbc.StandardColumnMappings.longDecimalWriteFunction;
+import static io.trino.plugin.jdbc.StandardColumnMappings.shortDecimalWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.smallintColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.toLongTrinoTimestamp;
 import static io.trino.plugin.jdbc.StandardColumnMappings.toTrinoTimestamp;
@@ -101,6 +109,7 @@ public class ExasolClient
             .add("EXA_STATISTICS")
             .add("SYS")
             .build();
+    private final ConnectorExpressionRewriter<ParameterizedExpression> connectorExpressionRewriter;
 
     private static final int MAX_EXASOL_TIMESTAMP_PRECISION = 9;
 
@@ -113,6 +122,25 @@ public class ExasolClient
             RemoteQueryModifier queryModifier)
     {
         super("\"", connectionFactory, queryBuilder, config.getJdbcTypesMappedToVarchar(), identifierMapping, queryModifier, false);
+        // Basic implementation required to enable JOIN pushdown support
+        // It is covered by "testJoinpushdown" integration tests.
+        // More detailed test case scenarios are covered by Unit tests in "TestExasolClient"
+        this.connectorExpressionRewriter = JdbcConnectorExpressionRewriterBuilder.newBuilder()
+                .addStandardRules(this::quoted)
+                .add(new RewriteIn())
+                .map("$equal(left, right)").to("left = right")
+                .map("$not_equal(left, right)").to("left <> right")
+                // Exasol doesn't support "IS NOT DISTINCT FROM" expression,
+                // so "$identical(left, right)" is rewritten with equivalent "(left = right OR (left IS NULL AND right IS NULL))" expression
+                .map("$identical(left, right)").to("(left = right OR (left IS NULL AND right IS NULL))")
+                .map("$less_than(left, right)").to("left < right")
+                .map("$less_than_or_equal(left, right)").to("left <= right")
+                .map("$greater_than(left, right)").to("left > right")
+                .map("$greater_than_or_equal(left, right)").to("left >= right")
+                .map("$not($is_null(value))").to("value IS NOT NULL")
+                .map("$not(value: boolean)").to("NOT value")
+                .map("$is_null(value)").to("value IS NULL")
+                .build();
     }
 
     @Override
@@ -221,10 +249,15 @@ public class ExasolClient
     }
 
     @Override
+    public Optional<ParameterizedExpression> convertPredicate(ConnectorSession session, ConnectorExpression expression, Map<String, ColumnHandle> assignments)
+    {
+        return connectorExpressionRewriter.rewrite(session, expression, assignments);
+    }
+
+    @Override
     protected boolean isSupportedJoinCondition(ConnectorSession session, JdbcJoinCondition joinCondition)
     {
-        // Deactivated because test 'testJoinPushdown()' requires write access which is not implemented for Exasol
-        return false;
+        return true;
     }
 
     @Override
@@ -471,7 +504,15 @@ public class ExasolClient
     @Override
     public WriteMapping toWriteMapping(ConnectorSession session, Type type)
     {
-        throw new TrinoException(NOT_SUPPORTED, "This connector does not support writing");
+        if (type instanceof DecimalType decimalType) {
+            String dataType = "decimal(%s, %s)".formatted(decimalType.getPrecision(), decimalType.getScale());
+            if (decimalType.isShort()) {
+                return WriteMapping.longMapping(dataType, shortDecimalWriteFunction(decimalType));
+            }
+            return WriteMapping.objectMapping(dataType, longDecimalWriteFunction(decimalType));
+        }
+
+        throw new TrinoException(NOT_SUPPORTED, "Unsupported column type: " + type.getDisplayName());
     }
 
     @Override
--- a/plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
+++ b/plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
@@ -13,7 +13,9 @@
  */
 package io.trino.plugin.exasol;
 
+import io.trino.Session;
 import io.trino.plugin.jdbc.BaseJdbcConnectorTest;
+import io.trino.plugin.jdbc.JoinOperator;
 import io.trino.sql.planner.plan.AggregationNode;
 import io.trino.sql.planner.plan.ProjectNode;
 import io.trino.testing.MaterializedResult;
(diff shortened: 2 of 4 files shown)


plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolClientToWriteMapping.java
```diff
@@ -0,0 +1,108 @@
...
+import java.math.BigDecimal;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
+
+import static com.google.common.reflect.Reflection.newProxy;
+import static io.trino.spi.type.DecimalType.createDecimalType;
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestExasolClientToWriteMapping
```
ebyhr: We should add integration tests instead of unit tests.

ebyhr (dismissed):

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolClientToWriteMapping.java
skyglass: The integration tests **testToWriteMappingForDecimalType** have been implemented in **TestExasolConnectorTest**

**toWriteMapping** and **convertPredicate** are triggered for LEFT JOIN queries when **assertJoinConditionallyPushedDown** check is done in integration tests.

Different column types are used to provide different values for **toWriteMapping** parameters.
Unfortunately, there is no way to check the business-logic in more details for these integration tests. These tests only make sure that the checks are successful and the **assertJoinConditionallyPushedDown** and **expectJoinPushdownOnEmptyProjection** are successfully supported for LEFT JOIN queries with different column types. By triggering **toWriteMapping** and **convertPredicate** these integration tests indirectly test these methods.

For more fine-grained scenarios and assertions, please, check **TestExasolToWriteMapping** unit tests.

**toWriteMapping** is also triggered for INSERT queries, but because Exasol Connector is read-only and doesn't support INSERT queries, we don't need to cover **toWriteMapping** with the tests for INSERT queries.

Basic implementations of **toWriteMapping** and **convertPredicate** are prerequisites for enabling the upcoming **JOIN_PUSHDOWN** feature.

In the next PR I will enable **testJoinPushdown()** in **BaseJdbcConnectorTest**, which requires basic implementations of **toWriteMapping** and **convertPredicate** to make **testJoinPushdown()** work and to enable **JOIN_PUSHDOWN** feature.

skyglass (commented):
Thank you very much for your review, @ebyhr ! 👍 

I have implemented integration tests, as requested.
In order to make the integration tests work, I also had to add basic implementation of **convertPredicate** method, which is also a prerequisite for the upcoing JOIN_PUSHDOWN feature.

Unfortunately, I couldn't implement **convertPredicate** in a separate PR, because the implemented integration tests require both **toWriteMapping** and **convertPredicate** methods.


The integration tests **testToWriteMappingForDecimalType** have been implemented in T**estExasolConnectorTest**

**toWriteMapping** and **convertPredicate** are triggered for LEFT JOIN queries when **assertJoinConditionallyPushedDown** check is done in integration tests.

Different column types are used to provide different values for **toWriteMapping** parameters.
Unfortunately, there is no way to check the business-logic in more details for these integration tests. These tests only make sure that the checks are successful and the **assertJoinConditionallyPushedDown** and **expectJoinPushdownOnEmptyProjection** are successfully supported for LEFT JOIN queries with different column types. By triggering **toWriteMapping** and **convertPredicate** these integration tests indirectly test these methods.

For more fine-grained assertions, please, check **TestExasolToWriteMapping** unit tests.

**toWriteMapping** is also triggered for INSERT queries, but because Exasol Connector is read-only and doesn't support INSERT queries, we don't need to cover **toWriteMapping** with the tests for INSERT queries.

Basic implementations of **toWriteMapping** and **convertPredicate** are prerequisites for enabling the upcoming JOIN_PUSHDOWN feature.

In the next PR I will enable **testJoinPushdown()** in **BaseJdbcConnectorTest**, which requires basic implementations of **toWriteMapping** and **convertPredicate** to make **testJoinPushdown()** work and to enable JOIN_PUSHDOWN feature.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -394,6 +411,33 @@ public void testExecuteProcedureWithNamedArgument()
         }
     }
 
+    @ParameterizedTest
+    @CsvSource(textBlock = """
+            16, 6, 123456.123456
+            36, 12, 123456789012345612345678.901234567890
+            19, 0, 1
+            19, 0, 1234567890123456789
+            """)
```
ebyhr: https://trino.io/docs/current/develop/tests.html#avoid-data-providers-and-parametric-tests

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestToWriteMapping.java
```diff
@@ -0,0 +1,119 @@
...
+import java.math.BigDecimal;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Types;
+
+import static com.google.common.reflect.Reflection.newProxy;
+import static io.trino.spi.type.DecimalType.createDecimalType;
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestToWriteMapping
```
ebyhr: I don't think test class is helpful. Remove.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestConvertPredicate.java
```diff
@@ -0,0 +1,243 @@
...
+import static io.airlift.slice.Slices.utf8Slice;
+import static io.trino.SessionTestUtils.TEST_SESSION;
+import static io.trino.spi.type.BigintType.BIGINT;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static io.trino.spi.type.VarcharType.createVarcharType;
+import static io.trino.sql.ir.IrExpressions.not;
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestConvertPredicate
```
ebyhr: Rename to `TestExasolClient`

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -308,9 +343,18 @@ private static SliceWriteFunction hashTypeWriteFunction()
     }
 
     @Override
+    // Basic implementation of "toWriteMapping" is a prerequisite for enabling the JOIN_PUSHDOWN support
```
ebyhr: Remove unhelpful comment.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -194,11 +222,18 @@ protected void renameTable(ConnectorSession session, Connection connection, Stri
...
+    {
+        return connectorExpressionRewriter.rewrite(session, expression, assignments);
+    }
+
     @Override
     protected boolean isSupportedJoinCondition(ConnectorSession session, JdbcJoinCondition joinCondition)
     {
-        // Deactivated because test 'testJoinPushdown()' requires write access which is not implemented for Exasol
-        return false;
+        // Enabled to add support for JOIN_PUSHDOWN feature
```
ebyhr: Remove unhelpful comment.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -194,11 +222,18 @@ protected void renameTable(ConnectorSession session, Connection connection, Stri
         throw new TrinoException(NOT_SUPPORTED, "This connector does not support renaming tables");
     }
 
+    @Override
+    // Basic implementation of "convertPredicate" is a prerequisite for enabling the JOIN_PUSHDOWN support
```
ebyhr: Remove unhelpful comment.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -60,35 +65,47 @@ protected QueryRunner createQueryRunner()
...
+                     SUPPORTS_UPDATE -> false;
 
             default -> super.hasBehavior(connectorBehavior);
         };
     }
 
     @Override
+    // Overridden, because read-only Exasol connector does not support CREATE TABLE and INSERT statements
+    // Overridden to fix `testJoinpushdown`, `testCaseSensitiveTopNPushdown`
+    // and `testNullSensitiveTopNPushdown` tests in `BaseJdbcConnectorTest`
```
ebyhr: Remove unhelpful comment. The existing comment in the method is sufficient.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -93,6 +102,25 @@ public ExasolClient(
             RemoteQueryModifier queryModifier)
     {
         super("\"", connectionFactory, queryBuilder, config.getJdbcTypesMappedToVarchar(), identifierMapping, queryModifier, false);
+        // Basic implementation required to enable JOIN_PUSHDOWN support
```
ebyhr: JOIN_PUSHDOWN → JOIN pushdown

skyglass (commented):
Thank you very much for your review, @ebyhr ! 👍 
The changes have been applied.

skyglass:
@ebyhr Any updates on this PR?

pj-spoelders:
Hi @ebyhr ,
Can you take another look at this PR please? 
A mutual customer of ours is looking to use this optimization.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -93,6 +102,25 @@ public ExasolClient(
...
+                // so "$identical(left, right)" is rewritten with equivalent "(left = right OR (left IS NULL AND right IS NULL))" expression
+                .map("$identical(left, right)").to("(left = right OR (left IS NULL AND right IS NULL))")
+                .map("$less_than(left, right)").to("left < right")
+                .map("$less_than_or_equal(left, right)").to("left <= right")
+                .map("$greater_than(left, right)").to("left > right")
+                .map("$greater_than_or_equal(left, right)").to("left >= right")
+                .map("$not($is_null(value))").to("value IS NOT NULL")
+                .map("$not(value: boolean)").to("NOT value")
+                .map("$is_null(value)").to("value IS NULL")
+                .build();
```
chenjian2664, line 143: This seems to me should be added as predicate pushdown

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -310,7 +343,15 @@ private static SliceWriteFunction hashTypeWriteFunction()
     @Override
     public WriteMapping toWriteMapping(ConnectorSession session, Type type)
     {
-        throw new TrinoException(NOT_SUPPORTED, "This connector does not support writing");
+        if (type instanceof DecimalType decimalType) {
```
chenjian2664, line 507: Does the connector support write now?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -60,24 +63,30 @@ protected QueryRunner createQueryRunner()
     protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
     {
         return switch (connectorBehavior) {
+            case SUPPORTS_JOIN_PUSHDOWN,
```
chenjian2664: `SUPPORTS_JOIN_PUSHDOWN_WITH_DISTINCT_FROM` and others will derive from `SUPPORTS_JOIN_PUSHDOWN`, so add `SUPPORTS_JOIN_PUSHDOWN` is enough?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -60,24 +63,30 @@ protected QueryRunner createQueryRunner()
...
-                 SUPPORTS_ROW_TYPE,
-                 SUPPORTS_NEGATIVE_DATE, // min date is 0001-01-01
-                 SUPPORTS_RENAME_COLUMN,
-                 SUPPORTS_RENAME_TABLE,
-                 SUPPORTS_SET_COLUMN_TYPE,
-                 SUPPORTS_AGGREGATION_PUSHDOWN,
-                 SUPPORTS_PREDICATE_EXPRESSION_PUSHDOWN,
-                 SUPPORTS_UPDATE,
-                 SUPPORTS_MERGE -> false;
+                     SUPPORTS_ARRAY,
```
chenjian2664: revert

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: This logic is intentionally added as part of the join pushdown work, even though it looks like predicate pushdown.

In Trino, join pushdown depends on predicate pushdown: join conditions are represented internally as expressions that must be rewritten into connector-specific SQL before a join can be pushed down.

Without basic predicate/expression rewriting (e.g. =, <, IS NULL, NOT, IDENTICAL), the connector is unable to serialize join conditions into Exasol SQL, which makes join pushdown impossible.

This implementation is therefore a minimum prerequisite for join pushdown, not a complete predicate pushdown feature. It provides only the expression rewrites required for the current join pushdown integration tests to pass.

A full predicate pushdown implementation (including filters beyond join conditions) is out of scope for this change and can be addressed separately.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: No, the Exasol connector does not support writing.

However, toWriteMapping is not used exclusively for write operations. In the JDBC framework, it is also invoked when binding values to prepared statement parameters, including those generated during join pushdown.

One of the join pushdown integration tests exercises this code path and fails without a minimal toWriteMapping implementation, because values must be converted to JDBC-compatible representations before being bound.

The added logic is therefore not enabling write support. It is a minimal implementation required for value conversion in prepared statements, specifically to support join pushdown. All actual write operations remain explicitly unsupported and continue to fail with NOT_SUPPORTED.

A full write implementation is out of scope for this connector and this change.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
skyglass: As far as I know, setting SUPPORTS_JOIN_PUSHDOWN to true does not automatically enable the more specific join pushdown capability flags.

In BaseJdbcConnectorTest, each flag (for example SUPPORTS_JOIN_PUSHDOWN_WITH_DISTINCT_FROM, SUPPORTS_JOIN_PUSHDOWN_WITH_NON_EQUI_JOIN, etc.) is evaluated independently to decide whether the corresponding test cases should run.

If any of these flags is set to false, the related join pushdown tests are skipped, which directly reduces test coverage for the JOIN PUSHDOWN feature.

Therefore, if the connector supports these specific join semantics, the corresponding flags must be explicitly set to true. Setting only SUPPORTS_JOIN_PUSHDOWN is not sufficient to enable the full join pushdown test suite.

chenjian2664: I understand that we needs to enable those flags for testing, isn't the `SUPPORTS_JOIN_PUSHDOWN_WITH_DISTINCT_FROM` default value depends on `SUPPORTS_JOIN_PUSHDOWN` - https://github.com/trinodb/trino/blob/master/testing/trino-testing/src/main/java/io/trino/testing/TestingConnectorBehavior.java#L62-L67
is anything I am missing?

skyglass (commented):
@chenjian2664, thank you very much for your feedback and clarification! 👍 
I actually didn't know that SUPPORT_JOIN_PUSHDOWN automatically enables all sub flags, but I checked the TestPostgreSqlConnectorTest and it only has SUPPORT_JOIN_PUSHDOWN flag, without any sub-flags.

Therefore, I have removed all sub-flags and kept only SUPPORT_JOIN_PUSHDOWN flag, as you suggested.

The PR was updated, please review again.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
```diff
@@ -87,7 +92,16 @@ protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
     protected TestTable newTrinoTable(String namePrefix, @Language("SQL") String tableDefinition, List<String> rowsToInsert)
     {
         // Use Exasol executor because the connector does not support creating tables
-        return new TestTable(exasolServer.getSqlExecutor(), TEST_SCHEMA + "." + namePrefix, tableDefinition, rowsToInsert);
+        return new TestTable(exasolServer.getSqlExecutor(), TEST_SCHEMA + "." + namePrefix,
```
chenjian2664: It feels like we're heading in the wrong direction and continuing down it. For `newTrinoTable`, what we should actually do is use the Trino query runner instead of the Exasol query runner, and somehow use new `TestTable` instead.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -94,6 +103,25 @@ public ExasolClient(
             RemoteQueryModifier queryModifier)
     {
         super("\"", connectionFactory, queryBuilder, config.getJdbcTypesMappedToVarchar(), identifierMapping, queryModifier, false);
+        // Basic implementation required to enable JOIN pushdown support
+        // It is covered by "testJoinpushdown" integration tests.
+        // More detailed test case scenarios are covered by Unit tests in "TestExasolClient"
+        this.connectorExpressionRewriter = JdbcConnectorExpressionRewriterBuilder.newBuilder()
+                .addStandardRules(this::quoted)
+                .add(new RewriteIn())
+                .map("$equal(left, right)").to("left = right")
```
chenjian2664, line 131: It seems that current predicate covering test is insufficient, the connector is declared support predicate pushdown but without any support of those expressions rewrite

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -195,11 +223,16 @@ protected void renameTable(ConnectorSession session, Connection connection, Stri
         throw new TrinoException(NOT_SUPPORTED, "This connector does not support renaming tables");
     }
 
+    @Override
+    public Optional<ParameterizedExpression> convertPredicate(ConnectorSession session, ConnectorExpression expression, Map<String, ColumnHandle> assignments)
+    {
+        return connectorExpressionRewriter.rewrite(session, expression, assignments);
```
chenjian2664, line 254: I think this one should be part of the predicate pushdown support instead of supporting join, could you split the change into another commit?

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: This PR declares "basic" support for join pushdown. 
But it is impossible for join pushdown to work without enabling "basic" predicate pushdown. 
So when we declare support for "basic" join pushdown, it automatically means that we declare support for "basic" predicate pushdown.

TestExasolClient, which is part of this PR, provides sufficient test coverage for the new code in ExasolClient.
In addition there is a thorough integration test coverage in `testJoinPushdown` method of `BaseJdbConnectorTest`.
I didn't have to write integration tests for join pushdown feature, because `testJoinPushdown` in BaseJdbcConnectorTest is already there and can be simply enabled by setting SUPPORT_JOIN_PUSHDOWN flag to true (see TestExasolConnectorTest)

Of course, it is easy to think of the integration tests, which would fail, because join pushdown feature must be further developed to support more column types and predicate expressions, but this PR doesn't mean to provide full support of join pushdown. 
Still, even this basic join pushdown support, enabled by the new code in this PR, successfully passes all complex integration testing scenarios in `testJoinPushdown` method of `BaseJdbConnectorTest`.

We would be more than happy to continue with extending predicate, join and aggregation push down support, as soon as this PR is merged, so we are not blocked from further development.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java
skyglass: Unfortunately, we can't use trino query runner, because exasol connector is read-only and doesn't support writing, including creation of the tables.

But we still need to insert test data for integration tests and this workaround allows to insert test data directly to Exasol database, omitting trino query runner.

For join pushdown feature it doesn't really matter how the test data are inserted.
Although I agree that inserting data through query runner, would improve statistics and performance, so we should eventually start thinking about supporting write queries.
In any case, without supporting join pushdown, performance would be slow anyway, so it is still better to have join pushdown, even with read-only support.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
skyglass: Unfortunately, it is not possible to split this PR without breaking testJoinPushdown tests in BaseJdbcConnectorTest.

Current changes are as minimal as possible, to make the tests pass. If I remove at least one line of the code, some test would fail. In other words, the changes in this PR are transactional. All these changes must be done in one commit.

As a developer, I received the task to enable join pushdown support, which is what was requested by customers. 

How to enable join pushdown support?
The first step is to set SUPPORTS_JOIN_PUSHDOWN flag to true in TestExasolConnectorTest.
This would make tests in testJoinPushdown method of BaseJdbcConnectorTest fail, proving that even basic join pushdown is currently not supported.

Now we need to fix these tests by implementing basic join pushdown support.
How to implement basic join pushdown support?
The most obvisous way is to fix testJoinPushdown tests in BaseJdbcConnectorTest.
If these tests work, then basic join pushdown support is ready. 
I don't see any reason to claim the opposite, unless we don't trust the current join pushdown tests in BaseJdbcConnectorTest.

So I started implementation line by line, until all tests were passed.

* If I split the PR in any way, with true SUPPORTS_JOIN_PUSHDOWN flag, then testing pipeline will fail.
* If I split the PR in any way, with false SUPPORTS_JOIN_PUSHDOWN flag, then we would have a code, which is not covered by integration tests.

skyglass (commented):
Thank you very much for your review, @chenjian2664 !
I have answered your comments, hopefully it resolves the current PR.
Our customers would be very happy if we could continue further support of join and aggregate pushdown, but, unfortunately, this unresolved PR is blocking us from further development.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
chenjian2664: > TestExasolClient, which is part of this PR, provides sufficient test coverage for the new code in ExasolClient.

It's not, we should verify the pushdowned result as well
thus my suggestion is that refer the https://github.com/trinodb/trino/pull/28025 to add a simple predicate test in this connector and verify the predicate support works correctly first.

chenjian2664: >  but this PR doesn't mean to provide full support of join pushdown.

Could you elaborate on this? which part is not supported in this pr

skyglass: > Could you elaborate on this? which part is not supported in this pr

Thank you for the question, @chenjian2664 - happy to clarify.

The current JOIN_PUSHDOWN support in this PR is intentionally limited and primarily driven by the existing test coverage in `BaseJdbcConnectorTest`, specifically the `testJoinPushdown` tests. These tests only exercise a restricted set of column types and predicate expressions, and the implementation in this PR is scoped to make those tests pass.

At the moment, JOIN_PUSHDOWN does not work for certain types such as TIMESTAMP or INTERVAL, simply because these types are not yet supported by the Exasol Trino connector. Supporting them is a prerequisite for enabling JOIN_PUSHDOWN on join predicates involving those types. There are already open PRs addressing this:

TIMESTAMP support: https://github.com/trinodb/trino/pull/26963
INTERVAL support: https://github.com/trinodb/trino/pull/26452

As a result, this PR provides a minimal, initial implementation of JOIN_PUSHDOWN that is sufficient to satisfy the current `BaseJdbcConnectorTest` coverage, but not broader use cases.

In terms of expression rewriter support, this PR does not enable JOIN_PUSHDOWN for more complex predicate expressions with operators like add, multiply, divide, negate, like, ilike, etc. 
If predicate expressions in JOIN contain such operators, then JOIN_PUSHDOWN will currently not apply.

There is already a follow-up PR that significantly extends expression rewriter support as part of AGGREGATE_PUSHDOWN:
https://github.com/trinodb/trino/pull/26678

The AGGREGATE_PUSHOWN PR is currently in DRAFT because it is blocked by the current PR. 
In addition to aggregate pushdown support, it also expands expression rewriter support that can be reused by JOIN_PUSHDOWN.

While working on JOIN_PUSHDOWN, I followed a test-driven approach:
* Enabled SUPPORT_JOIN_PUSHDOWN flag, which caused `testJoinPushdown` to fail.
* Identified that a `ConnectorExpressionRewriter<ParameterizedExpression>` implementation was missing.
* Added a minimal implementation, extending it expression by expression until all tests passed.
* Intentionally kept the implementation minimal to limit the scope of this PR.

When working on AGGREGATE_PUSHDOWN, I then extended the `ConnectorExpressionRewriter<ParameterizedExpression>` much more comprehensively, as shown in the draft PR above.

Overall, this PR should be seen as a minimal, incremental step that enables basic JOIN_PUSHDOWN support within the limits of currently supported column types and current expression rewriter capabilities, covered by existing tests, with broader support planned and already partially implemented in the follow-up PRs.

skyglass: > TestExasolClient, which is part of this PR, provides sufficient test coverage for the new code in ExasolClient.
> 
> It's not, we should verify the pushdowned result as well

Pushdowned results are already verified by `TestExasolConnectorTest`.

`TestExasolConnectorTest` extends `BaseJdbcConnectorTest`, which verifies the pushdown results with `testJoinPushdown` tests. These tests are enabled by SUPPORT_JOIN_PUSHDOWN flag.

The `isFullyPushdown()` method in `BaseJdbcConnectorTest `validates correctness via `hasCorrectResultsRegardlessOfPushdown()` from `QueryAssertions`, which ensures that the results are correct: https://github.com/trinodb/trino/blob/master/core/trino-main/src/test/java/io/trino/sql/query/QueryAssertions.java

Therefore, existing `testJoinPushdown` tests already cover result correctness in addition to pushdown behavior.

skyglass: > thus my suggestion is that refer the https://github.com/trinodb/trino/pull/28025 to add a simple predicate test in this connector and verify the predicate support works correctly first.

Actually, predicate pushdown support already exists in the Exasol connector and is covered by existing integration tests:
https://github.com/trinodb/trino/blob/master/plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolConnectorTest.java

But predicate pushdown doesn't require expression rewriter, because predicates in WHERE don't require translation of predicate expressions from Exasol to Trino.

Therefore, predicate pushdown implementation doesn't require implementation of `ConnectorExpressionRewriter<ParameterizedExpression>`, which is why expression rewriter had to be introduced from the scratch for JOIN_PUSHDOWN.

skyglass:
Any updates on this PR?

skyglass:
Any updates on the PR?

skyglass:
Any updates on the PR?

PR #26614, opened 2025-09-11
Increase the timeout in the CI checks from 10 to 20 minutes

gertjanal:
## Description
Checks fail occasionally, lets see if increasing the timeout fixes it.

## Release notes


ebyhr:
I don't think we want to increase the timeout. We should fix the root cause. https://trinodb.slack.com/archives/C07ABNN828M/p1757025316621759

Issue #26323, opened 2025-08-01
OOME in Trino coordinator when performing Iceberg OPTIMIZE_MANIFESTS operations


szisiu:
Recently I've upgraded from Trino v471 to Trino 476. Infrastructure consists of Trino servers utilizing the Apache Iceberg connector. 
My setup uses a Hive Metastore (HMS) 4 (backend by MySQL) as the Iceberg metastore, and the Iceberg tables themselves (data and metadata files) are stored on Google Cloud Storage (GCS). 

Coordinator has 64GB and workers have 32GB RAM.

The biggest table size I have is ~30GB - maybe 100 of them. I have ~3k tables in total.

My settings for Iceberg catalog:

```
connector.name=iceberg
iceberg.catalog.type=hive_metastore
iceberg.expire-snapshots.min-retention=60s
iceberg.remove-orphan-files.min-retention=60s
iceberg.metadata-cache.enabled=true
iceberg.unique-table-location=false
iceberg.bucket-execution=false
iceberg.allowed-extra-properties=*
hive.metastore.uri=${uri}
hive.metastore.thrift.delete-files-on-drop=true
hive.metastore.thrift.assume-canonical-partition-keys=true
fs.native-gcs.enabled=true
gcs.json-key-file-path=${path}
gcs.client.max-retries=7
gcs.client.max-retry-time=100s
gcs.client.min-backoff-delay=1s
gcs.client.max-backoff-delay=32s
gcs.client.backoff-scale-factor=2
fs.cache.enabled=false # on trino v471 it was enabled
```

Settings for trino workers (Spot/Preemptive VMs):

```
coordinator=false
discovery.uri=${trino_coordinator_host}
http-server.http.port=${trino_port}
shutdown.grace-period=10s
# Catalog management
catalog.management=static
catalog.config-dir=${trino_config_dir}/catalog
# Resource management
query.max-cpu-time=7d
query.max-memory-per-node=${trino_memory_per_node_mb}MB
query.max-memory=32GB
query.max-total-memory=64GB
memory.heap-headroom-per-node=${trino_memory_per_node_mb}MB
# Query management
query.max-planning-time=1m
query.max-execution-time=1h
query.max-run-time=1h
retry-policy=NONE
# Query optimizer
optimizer.dictionary-aggregation=true
# Distributed sort
distributed-sort=true
# Logging
log.format=JSON
```

JVM settings:

```
-server
-Xmx51G # 64G*0.8
[...]

PR #26616, opened 2025-09-11
Limit number of manifest files written by OPTIMIZE_MANIFESTS

raunaqmorarka:
## Description
Avoids coordinator OOM and too many small manifests

Fixes https://github.com/trinodb/trino/issues/26323

## Release notes

```markdown
## Iceberg
* Fix coordinator crashes from running OPTIMIZE_MANIFESTS on partitioned tables. ({issue}`26323`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -176,7 +176,6 @@ import org.apache.iceberg.SnapshotUpdate;
 import org.apache.iceberg.SortField;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.StatisticsFile;
-import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.TableProperties;
 import org.apache.iceberg.TableScan;
@@ -200,6 +199,7 @@ import org.apache.iceberg.types.Types.IntegerType;
 import org.apache.iceberg.types.Types.NestedField;
 import org.apache.iceberg.types.Types.StringType;
 import org.apache.iceberg.types.Types.StructType;
+import org.apache.iceberg.util.StructLikeWrapper;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
@@ -219,6 +219,7 @@ import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2116,17 +2117,23 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
             return;
         }
-        if (manifests.size() == 1 && manifests.getFirst().length() < icebergTable.operations().current().propertyAsLong(MANIFEST_TARGET_SIZE_BYTES, MANIFEST_TARGET_SIZE_BYTES_DEFAULT)) {
+        long manifestTargetSizeBytes = icebergTable.operations().current().propertyAsLong(MANIFEST_TARGET_SIZE_BYTES, MANIFEST_TARGET_SIZE_BYTES_DEFAULT);
+        if (manifests.size() == 1 && manifests.getFirst().length() < manifestTargetSizeBytes) {
             return;
         }
+        long totalManifestsSize = manifests.stream().mapToLong(ManifestFile::length).sum();
+        // Having too many open manifest writers can potentially cause OOM on the coordinator
+        long targetManifestClusters = Math.min(((totalManifestsSize + manifestTargetSizeBytes - 1) / manifestTargetSizeBytes), 100);
```
oskar-szwajkowski, line 2126: would it make sense to make this `100` configurable?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2116,17 +2117,23 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
 
         beginTransaction(icebergTable);
         RewriteManifests rewriteManifests = transaction.rewriteManifests();
+        Types.StructType structType = icebergTable.spec().partitionType();
         rewriteManifests
                 .clusterBy(file -> {
```
oskar-szwajkowski, line 2132: can inside of this lambda + lines 2124-2126 be put in separate method?
so that all comments can be put in javadoc and its logic is in single place

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: That's mainly there as a safety net. It should be rare to reach that number. We've seen coordinator OOM with large number of open manifest writers here, so don't want to expose it as a configureable to users.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: separate method seems unnecessary here as we don't repeat it anywhere and its a small amount of code

PR #26628, opened 2025-09-12
Allow specifying list of allowed retry policies

losipiuk:
I most cases it is better if single cluster only runs queries with specific retry policy matching cluster configuration.
This PR adds new configuration property `allowed-retry-policies` so this can be enforced.

## Release notes

```markdown
## General
* Add `allowed-retry-policies` configuration property to specify which retry policies can be selected by user
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
+++ b/core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
@@ -794,6 +794,11 @@ public final class SystemSessionProperties
                         "Retry policy",
                         RetryPolicy.class,
                         queryManagerConfig.getRetryPolicy(),
+                        value -> {
+                            if (!queryManagerConfig.getAllowedRetryPolicies().contains(value)) {
+                                throw new TrinoException(INVALID_SESSION_PROPERTY, format("Retry policy %s not allowed. Must be one of %s", value, queryManagerConfig.getAllowedRetryPolicies()));
+                            }
+                        },
                         true),
                 integerProperty(
                         QUERY_RETRY_ATTEMPTS,
--- a/core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
+++ b/core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
@@ -13,6 +13,7 @@
  */
 package io.trino.execution;
 
+import com.google.common.collect.ImmutableSet;
 import io.airlift.configuration.Config;
 import io.airlift.configuration.ConfigDescription;
 import io.airlift.configuration.DefunctConfig;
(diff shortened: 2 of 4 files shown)


core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
```diff
@@ -794,6 +794,12 @@ public SystemSessionProperties(
                         "Retry policy",
                         RetryPolicy.class,
                         queryManagerConfig.getRetryPolicy(),
+                        value -> {
+                            if (queryManagerConfig.getAllowedRetryPolicies().contains(value)) {
```
wendigo: can we just do:

```
if (!queryManagerConfig.getAllowedRetryPolicies().contains(value)) {
    throw ...
}
```

core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
```diff
@@ -637,6 +638,19 @@ public boolean isRetryPolicyAllowed()
         return allowedRetryPolicies.contains(retryPolicy);
     }
 
+    public boolean isUseOfExchangeForQueryRetryPolicyAllowed()
+    {
+        return useOfExchangeForQueryRetryPolicyAllowed;
+    }
+
+    @Config("use-of-exchange-for-query-retry-policy-allowed")
```
martint: What’s the use case for this? Every option we add makes the system harder to configure and reason about.

Also, the name of that option is a mouthful. We need a better name.

losipiuk: The default is permissive, and makes sense for most users, so it does not add any extra complexity, unless you are in situation that you actually want to change it.

But the usecase where you actually want to limit the exchange use, just for task retries may be too niche. I will drop it.

core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
```diff
@@ -613,6 +619,24 @@ public QueryManagerConfig setRetryPolicy(RetryPolicy retryPolicy)
         return this;
     }
 
+    public Set<RetryPolicy> getAllowedRetryPolicies()
+    {
+        return allowedRetryPolicies;
+    }
+
+    @Config("allowed-retry-policies")
```
raunaqmorarka: Please add ConfigDescription and update documentation

losipiuk:
@martint please drop the "CR" when you are around.

core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
raunaqmorarka: Can you add intended use case to the PR description ?

raunaqmorarka (approved):

core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
```diff
@@ -613,6 +619,25 @@ public QueryManagerConfig setRetryPolicy(RetryPolicy retryPolicy)
...
```
martint: retry-policy.allowed

core/trino-main/src/main/java/io/trino/execution/QueryManagerConfig.java
```diff
@@ -613,6 +619,25 @@ public QueryManagerConfig setRetryPolicy(RetryPolicy retryPolicy)
...
+
+    @Config("retry-policy.allowed")
+    @ConfigDescription("Retry policies that are allowed to be used")
+    public QueryManagerConfig setAllowedRetryPolicies(Set<RetryPolicy> allowedRetryPolicies)
+    {
+        this.allowedRetryPolicies = ImmutableSet.copyOf(allowedRetryPolicies);
+        return this;
+    }
+
+    @AssertTrue(message = "Selected retry-policy not present in retry-policy.allowed list")
```
martint: "Selected retry policy not present in..."

PR #26629, opened 2025-09-12
Support configurable ExchangeManager

### LINKED ISSUE #26611: Feature Request: Make exchange-manager.properties file location configurable via ENV variable or config property
### Problem

Trino always expects `exchange-manager.properties` to be located at `/etc/trino/exchange-manager.properties`.  
This path is hardcoded in `ExchangeManagerRegistry.java` and cannot be configured.

In Kubernetes and container setups this is limiting because:
- Config files could be injected dynamically (Vault Agent, Helm, Secrets).
- `/etc/trino` may be cleared or overwritten.

---

### Proposed Solution

Allow a custom location for the `exchange-manager.properties` file, for example:

1. **Environment variable**  
   `TRINO_EXCHANGE_MANAGER_CONFIG=/path/to/exchange-manager.properties`

2. **Config property in `config.properties`**  
   `exchange-manager.config-file=/path/to/exchange-manager.properties`

If nothing is set, keep the current default `/etc/trino/exchange-manager.properties` for backward compatibility.

---

### Why

- Easier integration with Vault Agent and Kubernetes secrets.  
- No need to overwrite `/etc/trino` inside containers.  
- Consistent with access control, which already supports custom file locations via `security.config-file`.  
- Removes the limitation of a hardcoded path in `ExchangeManagerRegistry.java`.

yangshangqing95:
## Description

This PR adds support for configuring the location of the Exchange Manager configuration file.  
Users can now specify a custom path for the Exchange Manager config file, which allows for more flexible deployment and easier configuration management.

---

Closes #26611

https://github.com/trinodb/trino/issues/26611

---

## Release notes

```markdown
## General
* Add support for specifying the location of the Exchange Manager configuration file. ({issue}`26611`)

### DIFF
--- a/core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
+++ b/core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
@@ -1,0 +1,37 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.exchange;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.validation.FileExists;
+
+import java.io.File;
+import java.util.Optional;
+
+public class ExchangeManagerConfig
+{
+    private Optional<File> exchangeManagerConfigFile = Optional.empty();
+
+    public Optional<@FileExists File> getExchangeManagerConfigFile()
+    {
+        return exchangeManagerConfigFile;
+    }
+
+    @Config("exchange-manager.config-file")
+    public ExchangeManagerConfig setExchangeManagerConfigFile(File exchangeManagerConfigFile)
+    {
+        this.exchangeManagerConfigFile = Optional.ofNullable(exchangeManagerConfigFile);
+        return this;
+    }
+}
--- a/core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerRegistry.java
+++ b/core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerRegistry.java
@@ -29,6 +29,7 @@ import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 
 import static com.google.common.base.Preconditions.checkArgument;
@@ -52,16 +53,19 @@ public class ExchangeManagerRegistry
 
     private volatile ExchangeManager exchangeManager;
     private final SecretsResolver secretsResolver;
+    private final Optional<File> configFile;
 
     @Inject
     public ExchangeManagerRegistry(
             OpenTelemetry openTelemetry,
             Tracer tracer,
-            SecretsResolver secretsResolver)
+            SecretsResolver secretsResolver,
+            ExchangeManagerConfig config)
     {
         this.openTelemetry = requireNonNull(openTelemetry, "openTelemetry is null");
         this.tracer = requireNonNull(tracer, "tracer is null");
         this.secretsResolver = requireNonNull(secretsResolver, "secretsResolver is null");
+        this.configFile = config.getExchangeManagerConfigFile();
     }
 
     public void addExchangeManagerFactory(ExchangeManagerFactory factory)
@@ -74,13 +78,14 @@ public class ExchangeManagerRegistry
 
     public void loadExchangeManager()
     {
-        if (!CONFIG_FILE.exists()) {
+        File configFile = this.configFile.orElse(CONFIG_FILE);
+        if (!configFile.exists()) {
             return;
         }
 
-        Map<String, String> properties = loadProperties(CONFIG_FILE);
+        Map<String, String> properties = loadProperties(configFile);
         String name = properties.remove(EXCHANGE_MANAGER_NAME_PROPERTY);
-        checkArgument(!isNullOrEmpty(name), "Exchange manager configuration %s does not contain %s", CONFIG_FILE, EXCHANGE_MANAGER_NAME_PROPERTY);
+        checkArgument(!isNullOrEmpty(name), "Exchange manager configuration %s does not contain %s", configFile, EXCHANGE_MANAGER_NAME_PROPERTY);
 
         loadExchangeManager(name, properties);
     }
--- a/core/trino-main/src/test/java/io/trino/exchange/TestExchangeConfig.java
+++ b/core/trino-main/src/test/java/io/trino/exchange/TestExchangeConfig.java
@@ -1,0 +1,50 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.exchange;
+
+import com.google.common.collect.ImmutableMap;
+import org.junit.jupiter.api.Test;
+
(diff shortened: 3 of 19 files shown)

### CODE REVIEW

**Comment by wendigo** [comment]
Personally, I don't think that we want change like that. It doesn't solve any particular issue, at least none that is highlighted in that PR. How does it make deployment more flexible?

**Review by wendigo — DISMISSED** [review]
.

**Comment by losipiuk** [comment]
> Personally, I don't think that we want change like that. It doesn't solve any particular issue, at least none that is highlighted in that PR. How does it make deployment more flexible?

I agree.

**Comment by yangshangqing95** [comment]
> Personally, I don't think that we want change like that. It doesn't solve any particular issue, at least none that is highlighted in that PR. How does it make deployment more flexible?

Hi @wendigo @losipiuk Thanks for reviewing the PR!

Let me clarify the motivation — the goal is to make deployments in containerized/Kubernetes environments more flexible.

issue link: https://github.com/trinodb/trino/issues/26611

Today ExchangeManagerRegistry always looks for
`/etc/exchange-manager.properties`, which creates a few practical issues:

In Kubernetes, configuration is usually mounted from Secrets or Vault Agent sidecars, and the target path is not always /etc/trino (sometimes it’s a read-only volume or gets replaced by the image build).

Helm charts and operators often template configs into arbitrary paths; to use exchange manager we currently have to copy/patch them into /etc/trino, which breaks immutability and complicates upgrades.

**Other Trino subsystems (e.g. access-control.config-files, event-listener.config-files) already allow a custom location; but this component still relies on a hardcoded path.**

By adding an optional property
`exchange-manager.config-file=/path/to/exchange-manager.properties`
we keep the existing default for backward compatibility, but remove the need to rebuild the image or override /etc/trino when running on K8s or integrating with Vault Agent.

So the change doesn’t affect default users, but gives operators more flexibility and aligns this component with the rest of the configuration system.

Would that address the concern about “no particular issue”? Happy to update the PR description.

**Comment by yangshangqing95** [comment]
Hi @ebyhr Documentation has been added to reflect the new exchange-manager.config-file option, please help review, thank you!

File: core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerRegistry.java [hunk 1]
```diff
@@ -74,13 +77,14 @@ public void addExchangeManagerFactory(ExchangeManagerFactory factory)
 
     public void loadExchangeManager()
     {
-        if (!CONFIG_FILE.exists()) {
+        File configFile = (this.configFile != null) ? this.configFile : CONFIG_FILE;
```
losipiuk: can you just use `CONFIG_FILE` as default in config and simplify logic here?

yangshangqing95: Yes, in the initial version I set `CONFIG_FILE` as the default in `ExchangeManagerConfig`, but to stay consistent with other modules and respect previous implementations, I moved it back to `ExchangeManagerRegistry`. 

For example, in the EventListener module, `CONFIG_FILE` is also a separate static variable in EventListenerManager:
`private static final File CONFIG_FILE = new File("etc/event-listener.properties");`

Other modules follow a similar pattern. If you think this simplification is necessary, I’m happy to make the change.

losipiuk: Nah - maybe it is for a reason. Let's keep it this way for now (unless someone else will think this is necessary). Maybe we will do bulk refactor if needed.

yangshangqing95: Thx. Glad to start contributing to our community.

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public File getExchangeManagerConfigFile()
+    {
+        return exchangeManagerConfigFile;
+    }
+
+    @Config("exchange-manager.config-file")
+    public ExchangeManagerConfig setExchangeManagerConfigFile(String exchangeManagerConfigFile)
+    {
+        if (!Strings.isNullOrEmpty(exchangeManagerConfigFile)) {
```
ebyhr: Static import `isNullOrEmpty`.

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
```diff
@@ -0,0 +1,38 @@
...
+import com.google.common.base.Strings;
+import io.airlift.configuration.Config;
+
+import java.io.File;
+
+public class ExchangeManagerConfig
+{
+    private File exchangeManagerConfigFile;
+
+    public File getExchangeManagerConfigFile()
```
ebyhr: Please consider changing the return type to `Optional<@FileExists File>`. We shouldn't return NULL as much as possible.

core/trino-main/src/test/java/io/trino/exchange/TestExchangeConfig.java
```diff
@@ -0,0 +1,50 @@
...
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Map;
+
+import static io.airlift.configuration.testing.ConfigAssertions.assertFullMapping;
+import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
+import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+
+public class TestExchangeConfig
```
ebyhr:
```suggestion
final class TestExchangeConfig
```

https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
* Test classes should be defined as package-private and final.

core/trino-main/src/test/java/io/trino/exchange/TestExchangeConfig.java
```diff
@@ -0,0 +1,50 @@
...
+{
+    @Test
+    public void testDefaults()
+    {
+        assertRecordedDefaults(recordDefaults(ExchangeManagerConfig.class)
+                .setExchangeManagerConfigFile(null));
+    }
+
+    @Test
+    public void testExplicitPropertyMappings()
```
ebyhr:
```suggestion
    void testDefaults()
    {
        assertRecordedDefaults(recordDefaults(ExchangeManagerConfig.class)
                .setExchangeManagerConfigFile(null));
    }

    @Test
    void testExplicitPropertyMappings()
```

https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
* Test methods should be defined as package-private.

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
yangshangqing95: okay

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
yangshangqing95: done

core/trino-main/src/test/java/io/trino/exchange/TestExchangeConfig.java
yangshangqing95: done

core/trino-main/src/test/java/io/trino/exchange/TestExchangeConfig.java
yangshangqing95: done

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
yangshangqing95: done

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
```diff
@@ -0,0 +1,41 @@
...
+    public Optional<@FileExists File> getExchangeManagerConfigFile()
+    {
+        return exchangeManagerConfigFile;
+    }
+
+    @Config("exchange-manager.config-file")
+    public ExchangeManagerConfig setExchangeManagerConfigFile(String exchangeManagerConfigFile)
+    {
+        if (!isNullOrEmpty(exchangeManagerConfigFile)) {
+            this.exchangeManagerConfigFile = Optional.of(new File(exchangeManagerConfigFile));
```
ebyhr: Could you change the parameter to `File`?
```java
    public ExchangeManagerConfig setExchangeManagerConfigFile(File exchangeManagerConfigFile)
    {
        this.exchangeManagerConfigFile = Optional.ofNullable(exchangeManagerConfigFile);
```

ebyhr (approved):

core/trino-main/src/main/java/io/trino/exchange/ExchangeManagerConfig.java
yangshangqing95: sure, submitted

wendigo:
@cla-bot check

losipiuk:
rebsed

martint:
@cla-bot check

PR #26631, opened 2025-09-12
Fix copyWithAppendedNull for RowBlock with startOffset

SongChujun:
## Description
Fix a bug in #26405 where the ```rowIsNull``` is not properly adjusted with startOffset when calculating ```newRowIsNull```.

This change is not user visible for the purposes of release notes because the earlier bug has not yet been released.

## Release notes


pettyjamesm (approved):
LGTM

PR #26632, opened 2025-09-14
Partition-Aware Fan-Out for Iceberg OPTIMIZE

itamarwe:
# Enable Partition-Aware Fan-Out for Iceberg OPTIMIZE in Trino

## 🎯 Summary

This PR implements **partition-aware fan-out** for `ALTER TABLE ... EXECUTE OPTIMIZE` in Trino, enabling partition-based parallelism that scales with the number of partitions while avoiding the small file problem on the writer side.

## 🚀 Problem Statement

Currently, when running `ALTER TABLE ... EXECUTE OPTIMIZE` on partitioned Iceberg tables:

- **All readers start from the same few partitions**, causing contention
- **Poor parallelism utilization** - downstream writers become a bottleneck
- **Scalability issues** - performance doesn't improve with more workers

image

## ✅ Solution

Enable partition-aware scan scheduling for OPTIMIZE by:

1. **Activating table partitioning** during OPTIMIZE execution
2. **Leveraging existing Trino engine support** for partition-aware scheduling

### Core Implementation

**Key Changes**:
- Modified `beginOptimize()` method to include partitioning information
- Uses existing `getTablePartitioning()` method (no code duplication)
- Activates partitioning immediately for OPTIMIZE (bypasses normal activation rules)
- Works with all Iceberg partition types (identity, bucket, time-based)

### Test Coverage

**File**: `plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java`

Added comprehensive tests for:
- Identity partitioning (`partitioning = ARRAY['region']`)
- Bucket partitioning (`partitioning = ARRAY['bucket(id, 4)']`)
- Verification that OPTIMIZE works correctly with partition-aware fan-out

### Before (Traditional OPTIMIZE)
```
Worker 1: [file1.parquet, file5.parquet, file9.parquet, ...]  ← All from same partitions
Worker 2: [file2.parquet, file6.parquet, file10.parquet, ...] ← All from same partitions
Worker 3: [file3.parquet, file7.parquet, file11.parquet, ...] ← All from same partitions
```

### After (Partition-Aware Fan-Out)
```
Worker 1: [region='US' files, bucket=0 files, day=2023-01-15 files, ...]  ← Partition bucket 1
Worker 2: [region='EU' files, bucket=1 files, day=2023-01-16 files, ...]  ← Partition bucket 2
Worker 3: [region='ASIA' files, bucket=2 files, day=2023-01-17 files, ...] ← Partition bucket 3
```

### Execution Flow

1. **Query Planning**: `beginOptimize()` creates table handle with activated partitioning
2. **Engine Decision**: Trino's `DetermineTableScanNodePartitioning` rule enables partition-aware scheduling
3. **Split Generation**: Each split includes partition values for proper distribution
4. **Split Distribution**: Readers are assigned to partition buckets instead of competing for the same data
5. **Parallel Processing**: Each worker processes its partition bucket independently

### Configuration
To enable full benefits, users should set:
```properties
optimizer.use-table-scan-node-partitioning = true
```

### Migration

No migration required. This change is:
- **Backward compatible**: Existing OPTIMIZE operations continue to work
- **Automatic**: Partition-aware fan-out is enabled automatically when beneficial
- **Configurable**: Can be controlled via existing Trino configuration properties

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
@@ -1,0 +1,100 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+import org.junit.jupiter.api.Test;
+
+import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
+import static io.trino.testing.TestingNames.randomNameSuffix;
+
+public class TestOptimizePartitioning
+        extends BaseIcebergConnectorTest
+{
+    public TestOptimizePartitioning()
+    {
+        super(PARQUET);
+    }
+
+    @Override
+    protected boolean supportsIcebergFileStatistics(String typeName)
+    {
+        return true;
+    }
+
+    @Override
+    protected boolean supportsRowGroupStatistics(String typeName)
+    {
+        return !(typeName.equalsIgnoreCase("varbinary") ||
+                typeName.equalsIgnoreCase("time") ||
+                typeName.equalsIgnoreCase("time(6)") ||
+                typeName.equalsIgnoreCase("timestamp(3) with time zone") ||
+                typeName.equalsIgnoreCase("timestamp(6) with time zone"));
+    }
+
+    @Override
+    protected boolean isFileSorted(String path, String sortColumnName)
+    {
+        return false;
+    }
+
+    @Test
+    public void testOptimizeWithIdentityPartitioning()
+    {
+        String tableName = "test_optimize_identity_" + randomNameSuffix();
+
+        // Create a table with identity partitioning
+        assertUpdate("CREATE TABLE " + tableName + " (id bigint, name varchar, region varchar) " +
+                "WITH (partitioning = ARRAY['region'])");
+
+        // Insert some data
+        assertUpdate("INSERT INTO " + tableName + " VALUES " +
+                "(1, 'Alice', 'US'), " +
+                "(2, 'Bob', 'US'), " +
+                "(3, 'Charlie', 'EU'), " +
+                "(4, 'David', 'EU')", 4);
+
+        // Run OPTIMIZE - this should now use partition-aware fan-out
+        assertQuerySucceeds("ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+
+        // Verify data is still correct after optimization
+        assertQuery("SELECT count(*) FROM " + tableName, "VALUES (4)");
+
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
+    @Test
+    public void testOptimizeWithBucketPartitioning()
+    {
+        String tableName = "test_optimize_bucket_" + randomNameSuffix();
+
+        // Create a table with bucket partitioning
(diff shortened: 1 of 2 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
```diff
@@ -0,0 +1,100 @@
...
+        assertUpdate("INSERT INTO " + tableName + " VALUES " +
+                "(1, 'Alice', 'US'), " +
+                "(2, 'Bob', 'US'), " +
+                "(3, 'Charlie', 'EU'), " +
+                "(4, 'David', 'EU')", 4);
+
+        // Run OPTIMIZE - this should now use partition-aware fan-out
+        assertQuerySucceeds("ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+
+        // Verify data is still correct after optimization
```
RoeyoOgen: i see this verifies data correctness, but how does this confirm that the optimisation actually used the new partition-aware fan-out logic?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
```diff
@@ -0,0 +1,100 @@
...
+        assertQuery("SELECT count(*) FROM " + tableName, "VALUES (4)");
+
+        assertUpdate("DROP TABLE " + tableName);
```
RoeyoOgen: If an assertion fails before the DROP TABLE statement, the may test leaves orphan tables behind, polluting the catalog and causing cascading failures.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
```diff
@@ -0,0 +1,100 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
```
RoeyoOgen: general:
The test code does not test for multiple partitions being optimized concurrently. perhaps add a test with many partitions

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
itamarwe: It's mostly a regression test at the moment. I'll try to add tests that verify the fan-out.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestOptimizePartitioning.java
itamarwe: Good catch. I'll fix.

MikeWallis42:
@itamarwe can you please keep this one alive, we could really benefit from having this PR go in. Many thanks!

itamarwe:
Sure thing. I'm waiting for a review :)

MikeWallis42:
From what I can see there are 2 CI checks failing and a missing Contributor License Agreement, without those they may not even look at the PR.

itamarwe:
I sent the CLA a month ago and still haven't received a response.
I'll fix the CI issues today/tomorrow.

martint:
@itamarwe, it seems like we never got the email with the CLA. Would you mind resubmitting?

itamarwe:
I’ve just sent it again now.

martint:
@cla-bot check

itamarwe:
@raunaqmorarka any chance you could review?

raunaqmorarka:
@itamarwe Apologies for the delay here, I looked into this and came up with an alternative approach to handling this issue at https://github.com/trinodb/trino/pull/27104
The problem with the approach here is that it is hard-coding the use of table partitioning for the scan in OPTIMIZE, but this should still be subject to the conditions in `io.trino.plugin.iceberg.IcebergMetadata#applyPartitioning` and `io.trino.sql.planner.iterative.rule.DetermineTableScanNodePartitioning`.
This approach is also retaining an unnecessary remote exchange between the scan and execute nodes.
PTAL at above PR and let me know if that resolves the underlying problem for your case.

raunaqmorarka:
superseded by https://github.com/trinodb/trino/pull/27104

PR #26537, opened 2025-09-02
Revert "Remove PagesResponseWriter as a provider"

wendigo:
This reverts commit 46c28bf608c6add93cec0a09eb5539bc94181369.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

PR #26644, opened 2025-09-16
Do not throw EOFException when writing pages

wendigo:
This restores previous behavior that was removed in the 46c28bf608c6add93cec0a09eb5539bc94181369

Supersedes https://github.com/trinodb/trino/pull/26537

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/PagesInputStreamFactory.java
+++ b/core/trino-main/src/main/java/io/trino/server/PagesInputStreamFactory.java
@@ -20,10 +20,13 @@ import io.airlift.slice.SliceOutput;
 import io.trino.FeaturesConfig;
 import io.trino.FeaturesConfig.DataIntegrityVerification;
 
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/server/PagesInputStreamFactory.java
```diff
@@ -42,17 +45,24 @@ public PagesInputStreamFactory(FeaturesConfig featuresConfig)
...
+            header.writeInt(serializedPages.size());
+            header.flush();
 
-        stream.close();
+            for (Slice page : serializedPages) {
+                page.getInput().transferTo(stream);
+                stream.flush();
+            }
+        }
+        catch (IOException | UncheckedIOException e) {
```
losipiuk, line 60: can you wrap only relevant part of code (for loop)

PR #18390, opened 2023-07-24
Add support for `GROUP BY AUTO` aggregation

ebyhr:
## Description

This syntax allows omitting column positions or names after `GROUP BY`. 
For instance, `SELECT name, count(1) FROM GROUP BY AUTO ` would be translated to `SELECT name, count(1) FROM GROUP BY name`

References in other database/query engines
* BigQuery https://cloud.google.com/bigquery/docs/reference/standard-sql/query-syntax#group_by_all
* Snowflake https://docs.snowflake.com/en/sql-reference/constructs/group-by#label-group-by-all-columns
* Databricks https://docs.databricks.com/sql/language-manual/sql-ref-syntax-qry-select-groupby.html
* ClickHouse https://clickhouse.com/docs/en/sql-reference/statements/select/group-by#group-by-all
* Databend https://databend.rs/doc/sql-commands/query-syntax/query-group-by#group-by-all-columns

## Release notes

(x) Release notes are required, with the following suggested text:

```markdown
# General
* Add support for `GROUP BY AUTO` aggregation that implicitly groups all non-aggregated columns. ({issue}`18390`)
```

PR #26646, opened 2025-09-16
Document GROUP BY AUTO syntax

ebyhr:
## Description

Follow-up of #18390

Screenshot 2025-10-04 at 5 38 55

## Release notes

### DIFF
--- a/docs/src/main/sphinx/sql/select.md
+++ b/docs/src/main/sphinx/sql/select.md
@@ -379,6 +379,32 @@ Complex grouping operations are often equivalent to a `UNION ALL` of simple
 does not apply, however, when the source of data for the aggregation
 is non-deterministic.
 
+### AUTO
+
+When `AUTO` is specified, the Trino engine automatically determines the grouping
+columns instead of requiring them to be listed explicitly. In this mode, any
+column in the `SELECT` list that is not part of an aggregate function is
+implicitly treated as a grouping column.
+
+This example query calculates the total account balance per market segment.
+The `AUTO` clause derives `mktsegment` as the grouping key, since it is not used
+in any aggregate function (i.e., `sum`).
+
+```sql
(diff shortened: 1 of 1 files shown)


docs/src/main/sphinx/sql/select.md
```diff
@@ -379,6 +379,30 @@ Complex grouping operations are often equivalent to a `UNION ALL` of simple
 does not apply, however, when the source of data for the aggregation
 is non-deterministic.
 
+### AUTO
+
+`AUTO` syntax implicitly groups on every column not included in an aggregation
```
martint: This needs better phrasing. Maybe something like:

> When `AUTO` is specified, the Trino engine automatically determines the grouping columns instead of requiring them to be listed explicitly. In this mode, any column in the `SELECT` list that is not part of an aggregate function is implicitly treated as a grouping column.

docs/src/main/sphinx/sql/select.md
```diff
@@ -379,6 +379,30 @@ Complex grouping operations are often equivalent to a `UNION ALL` of simple
...
 is non-deterministic.
 
+### AUTO
+
+`AUTO` syntax implicitly groups on every column not included in an aggregation 
+function.
+
+This query sums the account balance (`acctbal`) for each market segment 
+(`mktsegment`), with `AUTO` automatically determining the grouping 
+keys:
```
martint: This paragraph describes the query, but does not explain the conclusion that AUTO derives `mktsegment` as the grouping key since it's not part of any aggregate function. Maybe rephrase to:

> This example query calculates the total account balance per market segment. The `AUTO` clause derives `mktsegment` as the grouping key, since it is not used in any aggregate function (i.e., `sum`).

martint (changes requested):

martint (approved):

PR #26657, opened 2025-09-18
Verify that Trino has access to UnixOperatingSystemMXBean

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
+++ b/core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
@@ -29,8 +29,6 @@ import java.nio.charset.StandardCharsets;
 import java.time.Year;
 import java.util.List;
 import java.util.Locale;
-import java.util.OptionalLong;
-import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.lang.String.format;
@@ -55,9 +53,17 @@ final class TrinoSystemRequirements
         verifyOsArchitecture();
         verifyByteOrder();
         verifyUsingG1Gc();
+        verifyUnixOperatingMBeans();
         verifyFileDescriptor();
         verifySlice();
         verifyUtf8();
+    }
+
+    private static void verifyUnixOperatingMBeans()
+    {
+        if (!(ManagementFactory.getOperatingSystemMXBean() instanceof UnixOperatingSystemMXBean)) {
+            failRequirement("Trino requires access to UnixOperatingSystemMXBean");
+        }
     }
 
     private static void verify64BitJvm()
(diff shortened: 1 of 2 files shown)


wendigo:
Context https://github.com/trinodb/trino/pull/26548#discussion_r2358171725

core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
```diff
@@ -60,6 +62,13 @@ private static void verifyJvmRequirements()
         verifyUtf8();
     }
 
+    private static void verifyManagementBeans()
+    {
+        if (!(ManagementFactory.getOperatingSystemMXBean() instanceof OperatingSystemMXBean)) {
+            failRequirement("Trino requires access to the com.sun.management MBeans");
```
raunaqmorarka: If this access was not there earlier, was it failing startup or queries, or just missing JMX stats ?

wendigo: It would actually return empty CPU load statistics (which we don't want).

core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
```diff
@@ -60,6 +62,13 @@ private static void verifyJvmRequirements()
         verifyUtf8();
     }
 
+    private static void verifyManagementBeans()
+    {
+        if (!(ManagementFactory.getOperatingSystemMXBean() instanceof OperatingSystemMXBean)) {
```
raunaqmorarka: `io.trino.server.TrinoSystemRequirements#verifyFileDescriptor` is already making stronger guarantee that this is instance of `UnixOperatingSystemMXBean`.
Also, the return type here is already `OperatingSystemMXBean`, so not sure what the instanceof check will do

wendigo: @raunaqmorarka not quite. There are two classes in two different packages. Com.sun provides method that we use

wendigo: `java.lang.management.OperatingSystemMXBean` vs `com.sun.management.OperatingSystemMXBean` (`extends java.lang.management.OperatingSystemMXBean`) :)

wendigo: I've put even stronger check (which makes sense since we verify for linux/macosx anyway)

raunaqmorarka: ah, didn't realize there are JVM specific extensions for basic system metrics

wendigo: Yes. That's super confusing (same interface name, different packages)

losipiuk (approved):

PR #26661, opened 2025-09-18
Output metrics about remove_orphan_files execution

raunaqmorarka:
## Description
```
alter table lineitem execute remove_orphan_files(retention_threshold => '0d');

            metric_name         | metric_value
    ----------------------------+--------------
     processed_manifests_count  |            2
     valid_files_count          |           98
     scanned_files_count        |           97
     deleted_files_count        |            0
```

## Release notes

```markdown
## Iceberg
* Produce metrics about execution in the output of remove_orphan_files procedure. ({issue}`issuenumber`)
```


core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -121,7 +121,7 @@ Optional<TableExecuteHandle> getTableHandleForExecute(
 
     void finishTableExecute(Session session, TableExecuteHandle handle, Collection<Slice> fragments, List<Object> tableExecuteState);
 
-    void executeTableExecute(Session session, TableExecuteHandle handle);
+    Map<String, Long> executeTableExecute(Session session, TableExecuteHandle handle);
```
wendigo, line 126: Why Map<String, Long> instead of something more generic?

wendigo: Like some Connector specific handle that has a method to get a Page and list of symbols ?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -160,9 +160,10 @@ default void finishTableExecute(ConnectorSession session, ConnectorTableExecuteH
     }
 
     /**
-     * Execute a {@link TableProcedureExecutionMode#coordinatorOnly() coordinator-only} table procedure.
+     * Execute a {@link TableProcedureExecutionMode#coordinatorOnly() coordinator-only} table procedure
+     * and return procedure execution metrics that will be populated in the query output.
      */
-    default void executeTableExecute(ConnectorSession session, ConnectorTableExecuteHandle tableExecuteHandle)
+    default Map<String, Long> executeTableExecute(ConnectorSession session, ConnectorTableExecuteHandle tableExecuteHandle)
```
Praveen2112, line 166: Are we planning to limit ourselves to `Long` or - can be metrics be additional type like double ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2277,8 +2291,11 @@ private void scanAndDeleteInvalidFiles(Table table, ConnectorSession session, Sc
             // Ensure any futures still running are canceled in case of failure
             deleteFutures.forEach(future -> future.cancel(true));
         }
+        return new ScanAndDeleteResult(deletedFilesCount, scannedFilesCount);
     }
 
+    private record ScanAndDeleteResult(long deletedFilesCount, long scannedFilesCount) {}
```
Praveen2112: nit: Maybe swap arguments to be in sync with the record name

testing/trino-testing/src/main/java/io/trino/testing/QueryAssertions.java
```diff
@@ -336,7 +336,7 @@ private static void assertDistributedQuery(
         List<MaterializedRow> actualRows = actualResults.getMaterializedRows();
         List<MaterializedRow> expectedRows = expectedResults.getMaterializedRows();
 
-        if (compareUpdate) {
+        if (compareUpdate && !actualResults.getUpdateType().equals(Optional.of("ALTER TABLE EXECUTE"))) {
```
Praveen2112, line 339: Instead of skipping here, can we introduce something `assertAlter` or a dedicated method - which would set `compareUpdate` to false ?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
Praveen2112: Or should we return them in `Row` type ?

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
raunaqmorarka: This method is running directly inside SimpleTableExecuteOperator, so we don't need anything complex.
There isn't a need for anything more generic here at the moment, the use case of all the iceberg procedures where we want to add this is satisfied by this.

Praveen2112: What if we could expose additional metrics in double or other type ?

wendigo: Sure, but it could run for other cases as well. I can image that other table procedures will output some kind of an information

raunaqmorarka: We've had one BIGINT field all these years and failed to even populate that properly.
So no, I don't think there are many use cases waiting for a more generic framework.
I've gone through all the iceberg procedures and everything worth showing in the output is an integer. See Outputs in https://iceberg.apache.org/docs/latest/spark-procedures/
There is nothing stopping anyone from using something more generic like `io.trino.spi.metrics.Metric` in the future. At the moment, I find it to be unnecessary.

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -907,12 +907,39 @@ time is recommended to keep size of a table's data directory under control.
...
+
+:::{list-table} Output
+:widths: 40, 60
+:header-rows: 1
+
+* - Property name
+  - Description
+* - `processed_manifests_count`
+  - The count of manifest files read by remove_orphan_files.
+* - `valid_files_count`
```
chenjian2664: I think the meaning of `valid` here may be unclear to users. It doesn’t seem very meaningful to expose it directly, could we add more details to clarify what it represents/contains

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -907,12 +907,39 @@ time is recommended to keep size of a table's data directory under control.
 ALTER TABLE test_table EXECUTE remove_orphan_files(retention_threshold => '7d');
 ```
 
+```text
+        metric_name         | metric_value
+----------------------------+--------------
+ processed_manifests_count  |            2
```
chenjian2664, line 922: I noticed the output in Spark looks different from this: https://iceberg.apache.org/docs/latest/spark-procedures/#output_7.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
raunaqmorarka: Its limited to `long` for now, as that's what the current procedures need

testing/trino-testing/src/main/java/io/trino/testing/QueryAssertions.java
raunaqmorarka: There should be no need for a special `assertAlter` method. We should be able to assert a more detailed output with assertUpdate itself.
This branch is written with the assumption that an update query can only produce a row count output and nothing else. Since ALTER commands did not output anything before, this branch was never executed before.
Now with the more detailed output, we need to skip over this branch to allow the assertion on detailed results.

docs/src/main/sphinx/connector/iceberg.md
raunaqmorarka: Yes, the spark output is printing deleted file names. I don't want to do that since that is potentially a large list. The main use case here is that we want to provide a summary of what the procedure did. Right now you get no clue what/if the procedure did anything, unless you run metadata queries before and after and search for differences to the table.

docs/src/main/sphinx/connector/iceberg.md
raunaqmorarka: I've renamed it to `active_files_count` and clarified that these are files belonging to snapshots that have not been expired yet.

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
sopel39: I could imagine something like `OPTIMIZE` could produce something like input file size histogram.
We could just use `io.trino.spi.metrics.Metrics` here.
However, I think `Map<String, Long>` is also OK for now

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -1323,6 +1323,9 @@ protected Scope visitTableExecute(TableExecute node, Optional<Scope> scope)
             analysis.setUpdateType("ALTER TABLE EXECUTE");
             analysis.setUpdateTarget(executeHandle.catalogHandle().getVersion(), tableName, Optional.of(table), Optional.empty());
 
+            if (!procedureMetadata.getExecutionMode().isReadsData()) {
```
sopel39, line 1326: do we need this special case?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -1323,6 +1323,9 @@ protected Scope visitTableExecute(TableExecute node, Optional<Scope> scope)
...
+                return createAndAssignScope(node, scope, Field.newUnqualified("metric_name", VARCHAR), Field.newUnqualified("metric_value", BIGINT));
```
sopel39, line 1327: This will display `metric_name` `metric_value` even if there are not rows?

I think it would be more natural to have each metric as a separate titled column, but IMO current solution is OK for now

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
sopel39: > Or should we return them in Row type ?

In the future, this should probably be `Page` where channels correspond to metric names.

testing/trino-testing/src/main/java/io/trino/testing/QueryAssertions.java
sopel39: Maybe add `boolean compareExecute` boolean flag and special `assertExecute` method?

chenjian2664 (approved):

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
raunaqmorarka: Yes, we display `rows` today even when there are no rows.
Separate column per metric is harder to read in the output and likely takes more code changes (every procedure would have to declare its columns for the analyzer somehow)

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
raunaqmorarka: Yes, the code path for OPTIMIZE is different from all the other procedures, so this condition is needed to make sure the changes don't apply to OPTIMIZE. That has to be tackled separately

PR #26665, opened 2025-09-18
Add sql scalar function to transform value in a row with a single expression

gertjanal:
## Description
Updating an item in an array is possible with the function `transform(ARRAY[5, 6], x -> x + 1)`. This PR adds a new function to update the value for a field in a row. It can be useful when updating multiple rows in an array with a single expression, or when updating an inner value of a nested row:

For a single row:
```sql
SELECT transform(
    CAST(ROW('hello', 'world') AS ROW(greeting varchar, planet varchar))",
    'greeting', '', greeting -> concat(greeting, ' or goodbye'));
-- ROW('hello or goodbye', 'world')
```

For a nested row:
```sql
SELECT transform(
    CAST(ROW(ROW('hello'), 'world') as ROW(greeting ROW(text varchar), planet varchar)),
    'greeting',
    CAST(ROW('') as ROW(text varchar)),
    greeting -> transform(
        greeting, 'text', '', text -> concat(text, ' or goodbye')));
-- ROW(ROW('hello or goodbye'), 'world')
```

```sql
SELECT transform(ARRAY[
    cast(ROW('hello', 'world') AS ROW(greeting varchar, planet varchar)),
    cast(ROW('hi', 'mars') AS ROW(greeting varchar, planet varchar))],
    data -> transform(data, 'greeting', '', greeting -> concat(greeting, ' or goodbye')));
-- ARRAY[ROW('hello or goodbye', 'world'), ROW('hi or goodbye', 'mars')]
```

This PR is a draft as I'd like the opinion of developers:

* Is this a function we want?
* Can the `V` dummy value be removed and be inferred? I couldn't get it to work without the dummy value; only `.argumentType(functionType(UNKNOWN.getTypeSignature(), new TypeSignature("V")))` seemed to match the signature, but I couldn't get `UNKNOWN` to be updated to the actual value type.
* Is the reading magic with the `rawIndex` correct? Would it still work for larger data structures?

## Release notes

( x ) Release notes are required. Please propose a release note for me.

```markdown
## Functions
* Add sql scalar function to transform value in a row with a single expression
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/RowTransformFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/RowTransformFunction.java
@@ -28,6 +28,7 @@ import io.trino.spi.type.RowType;
 import io.trino.spi.type.RowType.Field;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.TypeSignature;
+import io.trino.spi.type.TypeSignatureParameter;
 import io.trino.sql.gen.lambda.UnaryFunctionInterface;
 
 import java.lang.invoke.MethodHandle;
--- a/core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
@@ -1036,6 +1036,10 @@ public class SignatureBinder
 
             TypeSignature constraintTypeSignature = applyBoundVariables(formalTypeSignature, bindings);
 
+            if (actualType == UNKNOWN) {
+                return SolverReturnStatus.UNCHANGED_SATISFIED;
+            }
+
             return satisfiesCoercion(relationshipType, actualType, constraintTypeSignature) ? SolverReturnStatus.UNCHANGED_SATISFIED : SolverReturnStatus.UNCHANGED_NOT_SATISFIED;
         }
     }
--- a/core/trino-main/src/main/java/io/trino/sql/ir/IrUtils.java
+++ b/core/trino-main/src/main/java/io/trino/sql/ir/IrUtils.java
@@ -21,6 +21,8 @@ import io.trino.spi.type.Type;
 import io.trino.sql.planner.DeterminismEvaluator;
 import io.trino.sql.planner.Symbol;
 import io.trino.sql.planner.SymbolsExtractor;
+import io.trino.type.FunctionType;
+import io.trino.type.UnknownType;
 
 import java.util.Arrays;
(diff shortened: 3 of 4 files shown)


core/trino-main/src/main/java/io/trino/operator/scalar/RowTransformFunction.java
```diff
@@ -0,0 +1,123 @@
...
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, String.format("Field with name %s not found in row", fieldName));
+        }
+        if (match.getType().getClass() != valueType.getClass()) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, String.format("Incompatible function types: field is of type %s but lambda returns %s", match.getType(), valueType));
+        }
+
+        Block[] blocks = new Block[fields.size()];
+        for (int i = 0; i < fields.size(); i++) {
+            if (i != fieldIndex) {
+                blocks[i] = sqlRow.getRawFieldBlock(i).getSingleValueBlock(sqlRow.getRawIndex());
```
gertjanal, line 115: This seems to give the right result, but is it correct to get the blocks this way?

core/trino-spi/src/main/java/io/trino/spi/type/ArrayType.java
```diff
@@ -117,6 +117,11 @@ public TypeOperatorDeclaration getTypeOperatorDeclaration(TypeOperators typeOper
         return operatorDeclaration;
     }
 
+    public static ArrayType arrayType(Type elementType)
```
gertjanal, line 120: Added to be consistent with `RowType.rowType`

mosabua:
Added stale-ignore label to prevent closing again since @gertjanal will continue on this.  Would be great to get a review from @martint

findepi:
Cool feature!

> ```sql
> SELECT transform(
>     CAST(ROW('hello', 'world') AS ROW(greeting varchar, planet varchar))",
>     'greeting', '', greeting -> concat(greeting, ' or goodbye'));
> ```


It seems counter-intuitive that field name `greeting` shows up in apostrophes. Field name is part of the type, not a value. When referencing a field name we don't need apostrophes (double quotes can be used), so it would be nice to be able to update also without apostrophes. Or is the intention to be able to dynamically select the field to be updated?

BTW some languages (JavaScript, Rust, Kotlin, C#) have a dedicated syntax for constructing a new struct out of existing.
It would be nice to have something like this for SQL `ROW` too.

PR #26669, opened 2025-09-19
Add support for exchange spooling on Alluxio in FTE

hqbhoho:
## Description
Support exchange spooling on Alluxio

## Release notes

```markdown
## General
* Support exchange spooling on Alluxio when fault-tolerant execution is enabled.
```


plugin/trino-exchange-filesystem/src/main/java/io/trino/plugin/exchange/filesystem/alluxio/ExchangeAlluxioConfig.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @Config("exchange.alluxio.block-size")
+    @ConfigDescription("Block size for Alluxio storage")
+    public ExchangeAlluxioConfig setAlluxioStorageBlockSize(DataSize alluxioStorageBlockSize)
+    {
+        this.alluxioStorageBlockSize = alluxioStorageBlockSize;
+        return this;
+    }
+
+    public Optional<@FileExists String> getAlluxioSiteConfPath()
```
losipiuk, line 49: Why do we need this one? We do not have it for `io.trino.filesystem.alluxio`.
If we need to be able to change some specific Alluxio config parameters it is better to expose those as specific getters/setters in `ExchangeAlluxioConfig`

hqbhoho: Thank you for your feedback!

In [Alluxio filesystem doc](https://trino.io/docs/current/object-storage/file-system-alluxio.html#alluxio-client-configuration), the Alluxio cluster connection is configured in the `alluxio-site.properties` properties file. The same config file must be located in `/opt/alluxio/conf` on all Trino cluster nodes. 

Since the Alluxio client automatically loads configurations from `/opt/alluxio/conf/alluxio-site.properties`, a custom configuration is likely necessary for exchange spooling, as its requirements may differ from those of the Alluxio filesystem used for the catalog. So if Alluixo is used for exchange spooling, should we avoid loading the configuration from `/opt/alluxio/conf/alluxio-site.properties` which is only used for catalog?

This configuration is similar to `hdfs.config.resources` which would be used if HDFS serves as the exchange storage.

​​Due to the vast number of configuration settings available in Alluxio, It seems that using configuration files would be more convenient.

losipiuk: Yeah. It makes sense. Thanks

plugin/trino-exchange-filesystem/src/main/java/io/trino/plugin/exchange/filesystem/alluxio/AlluxioFileSystemExchangeStorage.java
```diff
@@ -0,0 +1,376 @@
...
+            catch (IOException | RuntimeException e) {
+                return immediateFailedFuture(e);
+            }
+            return immediateVoidFuture();
+        }
+
+        @Override
+        public long getRetainedSize()
+        {
+            return INSTANCE_SIZE;
```
losipiuk, line 373: Should you take output stream into account? There is some buffering there probably

hqbhoho:
> This looks good. I am a bit concerned about using syncronous APIs for implementing asynchronous SPI but let's see how it works in practice.
> 
> Would you be fine adding some minimal documentation to `fault-tolerant-execution.md`?

Currently, I use Alluxio as exchange storage by HDFS client,  yet the P99 query latency decreased by 1/3 compared to using HDFS.

plugin/trino-exchange-filesystem/src/main/java/io/trino/plugin/exchange/filesystem/alluxio/AlluxioFileSystemExchangeStorage.java
hqbhoho: The `AlluxioFileOutStream` is not an instance of `BufferedOutputStream`. I will check the source code.

hqbhoho: In source code , data will send chunk by chunk. Whether need add the chunk size? They may use direct buffer.
```
public static final PropertyKey USER_STREAMING_WRITER_CHUNK_SIZE_BYTES =
      dataSizeBuilder(Name.USER_STREAMING_WRITER_CHUNK_SIZE_BYTES)
          .setAlias(Name.USER_NETWORK_WRITER_CHUNK_SIZE_BYTES)
          .setDefaultValue("1MB")
          .setDescription("When a client writes to a remote worker, the maximum chunk size.")
          .setConsistencyCheckLevel(ConsistencyCheckLevel.WARN)
          .setScope(Scope.CLIENT)
          .build();        
```

losipiuk (approved):

losipiuk:
Thanks

raunaqmorarka:
Now that this is supported natively, do we still need `exchange.hdfs.skip-directory-scheme-validation` ?
Since using that requires dropping jars into Trino, that's a more error prone way of doing the same thing

hqbhoho:
> Now that this is supported natively, do we still need `exchange.hdfs.skip-directory-scheme-validation` ? Since using that requires dropping jars into Trino, that's a more error prone way of doing the same thing

The configuration  `exchange.hdfs.skip-directory-scheme-validation` is used to ​​enable the use of Hadoop-Compatible File Systems with the HDFS client​​.
Not only for Alluxio, it enables users with more choices. In that case, I think it should be retained.

Issue #26598, opened 2025-09-10
`EXPLAIN ANALYZE` with `ALTER TABLE ... EXECUTE optimize` on unpartition tables shows exception message


ebyhr:
```
trino:tpch> explain analyze alter table region execute optimize;
Query 20250910_031025_00025_xqb6h failed: Unexpected ExplainAnalyzeNode found in plan; probably connector was not able to handle provided WHERE expression
java.lang.IllegalStateException: Unexpected ExplainAnalyzeNode found in plan; probably connector was not able to handle provided WHERE expression
	at io.trino.sql.planner.sanity.TableExecuteStructureValidator.lambda$validate$1(TableExecuteStructureValidator.java:56)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:421)
	at io.trino.sql.planner.sanity.TableExecuteStructureValidator.validate(TableExecuteStructureValidator.java:54)
	at io.trino.sql.planner.sanity.PlanSanityChecker.lambda$validate$0(PlanSanityChecker.java:107)
	at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:421)
	at io.trino.sql.planner.sanity.PlanSanityChecker.validate(PlanSanityChecker.java:107)
	at io.trino.sql.planner.sanity.PlanSanityChecker.validateFinalPlan(PlanSanityChecker.java:78)
	at io.trino.sql.planner.LogicalPlanner.plan(LogicalPlanner.java:274)
	at io.trino.sql.planner.LogicalPlanner.plan(LogicalPlanner.java:238)
	at io.trino.sql.planner.LogicalPlanner.plan(LogicalPlanner.java:233)
	at io.trino.execution.SqlQueryExecution.doPlanQuery(SqlQueryExecution.java:496)
	at io.trino.execution.SqlQueryExecution.planQuery(SqlQueryExecution.java:476)
	at io.trino.execution.SqlQueryExecution.start(SqlQueryExecution.java:414)
	at io.trino.execution.SqlQueryManager.createQuery(SqlQueryManager.java:284)
	at io.trino.dispatcher.LocalDispatchQuery.startExecution(LocalDispatchQuery.java:150)
	at io.trino.dispatcher.LocalDispatchQuery.lambda$waitForMinimumWorkers$1(LocalDispatchQuery.java:134)
	at io.airlift.concurrent.MoreFutures.lambda$addSuccessCallback$12(MoreFutures.java:570)
[...]

PR #26670, opened 2025-09-19
Fix failures for EXPLAIN and EXPLAIN ANALYZE on Iceberg OPTIMIZE queries

yangshangqing95:
## Description

This PR fixes failures when running EXPLAIN and EXPLAIN ANALYZE on Iceberg OPTIMIZE statements:

1. Fixed server-side failure in `EXPLAIN`; `EXPLAIN` prints an error on the server side, which is only ignored because of the `IGNORE_STATS_CALCULATOR_FAILURES` session property. When this property is set to false, it can still cause the query to fail.
   - Exception example:
     ```
     java.lang.IllegalArgumentException: Unexpected scanned files recording set
     	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:141)
     	at io.trino.plugin.iceberg.IcebergMetadata.getTableStatistics(IcebergMetadata.java:3546)
     	at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorMetadata.getTableStatistics(ClassLoaderSafeConnectorMetadata.java:356)
     	at io.trino.tracing.TracingConnectorMetadata.getTableStatistics(TracingConnectorMetadata.java:333)
     	at io.trino.metadata.MetadataManager.getTableStatistics(MetadataManager.java:492)
     	at io.trino.tracing.TracingMetadata.getTableStatistics(TracingMetadata.java:318)
     	at io.trino.cost.CachingTableStatsProvider.getTableStatisticsInternal(CachingTableStatsProvider.java:53)
     	at java.base/java.util.Map.computeIfAbsent(Map.java:1067)
        .......
     ```

2. Fixed failure in `EXPLAIN ANALYZE OPTIMIZE`.  
   - Related issue: [#26598](https://github.com/trinodb/trino/issues/26598)

This change ensures both `EXPLAIN` and `EXPLAIN ANALYZE OPTIMIZE` queries complete without errors.

This would also close https://github.com/trinodb/trino/issues/26598

## Release notes

```markdown
## Iceberg
* Fix failures when running EXPLAIN and EXPLAIN ANALYZE on OPTIMIZE statements. (#26598)
```

### DIFF
--- a/core/trino-main/src/test/java/io/trino/sql/planner/sanity/TestTableExecuteStructureValidator.java
+++ b/core/trino-main/src/test/java/io/trino/sql/planner/sanity/TestTableExecuteStructureValidator.java
@@ -1,0 +1,243 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.sql.planner.sanity;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import io.trino.connector.CatalogHandle;
+import io.trino.execution.warnings.WarningCollector;
+import io.trino.metadata.TableExecuteHandle;
+import io.trino.metadata.TableHandle;
+import io.trino.plugin.tpch.TpchTableHandle;
+import io.trino.plugin.tpch.TpchTransactionHandle;
+import io.trino.spi.connector.ColumnHandle;
+import io.trino.spi.connector.SchemaTableName;
+import io.trino.spi.connector.WriterScalingOptions;
+import io.trino.spi.predicate.TupleDomain;
+import io.trino.sql.ir.Constant;
+import io.trino.sql.ir.Expression;
+import io.trino.sql.planner.OrderingScheme;
+import io.trino.sql.planner.Partitioning;
+import io.trino.sql.planner.PartitioningScheme;
+import io.trino.sql.planner.PlanNodeIdAllocator;
+import io.trino.sql.planner.Symbol;
+import io.trino.sql.planner.assertions.BasePlanTest;
+import io.trino.sql.planner.plan.Assignments;
+import io.trino.sql.planner.plan.ExchangeNode;
+import io.trino.sql.planner.plan.ExplainAnalyzeNode;
+import io.trino.sql.planner.plan.FilterNode;
+import io.trino.sql.planner.plan.OutputNode;
+import io.trino.sql.planner.plan.PlanNode;
+import io.trino.sql.planner.plan.ProjectNode;
+import io.trino.sql.planner.plan.StatisticAggregations;
+import io.trino.sql.planner.plan.StatisticAggregationsDescriptor;
+import io.trino.sql.planner.plan.TableExecuteNode;
+import io.trino.sql.planner.plan.TableFinishNode;
+import io.trino.sql.planner.plan.TableScanNode;
+import io.trino.sql.planner.plan.TableWriterNode;
+import io.trino.testing.TestingTableExecuteHandle;
+import io.trino.testing.TestingTransactionHandle;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import static io.trino.spi.type.BigintType.BIGINT;
+import static io.trino.sql.planner.SystemPartitioningHandle.SINGLE_DISTRIBUTION;
+import static io.trino.sql.planner.TestingPlannerContext.PLANNER_CONTEXT;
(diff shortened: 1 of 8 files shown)


raunaqmorarka:
@cla-bot check

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka, line 3585: I think the fix can be to either return empty statistics here instead of the `checkArgument`, 
```
if (originalHandle.isRecordScannedFiles()) {
    // Skip
    return TableStatistics.empty();
}
```
or just remove the checks altogether and include `originalHandle.isRecordScannedFiles()` in the `cacheKey` below.

The test for the fix would be to successfully run EXPLAIN and EXPLAIN ANALYZE on OPTIMIZE query with `ignore_stats_calculator_failures` session property set to `false`.

yangshangqing95: Thanks for the suggestion! I agree this is the simplest fix — I actually considered that approach earlier as well.

However, I decided to introduce the `disableTableStatisticsCache` option because it provides a more explicit and controllable behavior. It allows users (and tests) to completely turn off caching when diagnosing issues related to stale or inconsistent statistics.

This approach keeps the cache key logic clean and avoids coupling it with isRecordScannedFiles(), while also giving us better flexibility to handle similar scenarios in the future — especially if we later support outputting statistics for operations like OPTIMIZE or other similar behaviors in Iceberg.

That said, I’m open to adjusting the naming or scope of the option if you think that would make the intent clearer.

raunaqmorarka: We don't want to introduce the amount of complexity added here for solving this case.
Iceberg doesn't have a problem of stale statistics because we don't cache statistics across queries and the metadata file cached for planning are immutable, so can never provide stale information.
If we want to output statistics for OPTIMIZE, then my second suggestion of "remove the checks altogether and include originalHandle.isRecordScannedFiles() in the cacheKey below" is the way to do it.

yangshangqing95: Hi @raunaqmorarka If we are just trying to solve the current issue, I agree with your suggestion. The changes have already been pushed.

yangshangqing95:
@cla-bot check

chenjian2664:
@cla-bot check

yangshangqing95:
@cla-bot check

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9321,6 +9350,13 @@ protected Session withoutSmallFileThreshold(Session session)
                 .build();
     }
 
+    protected Session withIgnoreStatsCalculatorFailuresFalse(Session session)
```
raunaqmorarka: withIgnoreStatsCalculatorFailuresFalse -> withoutIgnoreStatsCalculatorFailures

core/trino-main/src/test/java/io/trino/sql/planner/sanity/TestTableExecuteStructureValidator.java
```diff
@@ -0,0 +1,243 @@
...
+
+import static io.trino.spi.type.BigintType.BIGINT;
+import static io.trino.sql.planner.SystemPartitioningHandle.SINGLE_DISTRIBUTION;
+import static io.trino.sql.planner.TestingPlannerContext.PLANNER_CONTEXT;
+import static io.trino.sql.planner.plan.ExchangeNode.Scope.LOCAL;
+import static io.trino.sql.planner.plan.ExchangeNode.Type.REPARTITION;
+import static io.trino.testing.TestingHandles.TEST_CATALOG_HANDLE;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+final class TestTableExecuteStructureValidator
```
raunaqmorarka: I don't think these tests are needed, please remove this class

raunaqmorarka (commented):
minor comments around tests, lgtm otherwise

core/trino-main/src/test/java/io/trino/sql/planner/sanity/TestTableExecuteStructureValidator.java
yangshangqing95: removed

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
yangshangqing95: done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9324,6 +9350,13 @@ protected Session withoutSmallFileThreshold(Session session)
                 .build();
     }
 
+    protected Session withoutIgnoreStatsCalculatorFailures(Session session)
```
raunaqmorarka: private

yangshangqing95:
> Please rebase to latest master
sure, rebased

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
yangshangqing95: done

raunaqmorarka (approved):

PR #26672, opened 2025-09-19
Improve distribution of splits with filesystem caching

raunaqmorarka:
## Description
Allows multiple splits from a file to be scheduled on different nodes to reduce chances of a hotspot in the cluster

## Release notes

```markdown
## Iceberg, Hive, Delta Lake
* Improve uniformity of load distribution among workers when filesystem caching is enabled. ({issue}`26672`)
```

### DIFF
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CachingHostAddressProvider.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CachingHostAddressProvider.java
@@ -22,5 +22,12 @@ public interface CachingHostAddressProvider
     /**
      * Returns a lists of hosts which are preferred to cache the split with the given path.
      */
-    List<HostAddress> getHosts(String splitPath, List<HostAddress> defaultAddresses);
+    List<HostAddress> getHosts(String splitKey, List<HostAddress> defaultAddresses);
+
+    // Include offset and length in key to allow different splits belonging to the same file to be cached on different nodes
+    // This can help with avoiding hotspots when multiple splits are created from a file
+    static String getSplitKey(String splitPath, long offset, long length)
+    {
+        return splitPath + ":" + offset + ":" + length;
+    }
 }
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/ConsistentHashingHostAddressProvider.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/ConsistentHashingHostAddressProvider.java
@@ -60,9 +60,9 @@ public class ConsistentHashingHostAddressProvider
     }
 
     @Override
-    public List<HostAddress> getHosts(String splitPath, List<HostAddress> defaultAddresses)
+    public List<HostAddress> getHosts(String splitKey, List<HostAddress> defaultAddresses)
     {
-        return consistentHashRing.locate(splitPath, replicationFactor)
+        return consistentHashRing.locate(splitKey, replicationFactor)
                 .stream()
                 .map(TrinoNode::getHostAndPort)
                 .sorted(hostAddressComparator)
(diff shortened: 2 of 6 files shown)


lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CachingHostAddressProvider.java
```diff
@@ -22,5 +22,12 @@ public interface CachingHostAddressProvider
     /**
      * Returns a lists of hosts which are preferred to cache the split with the given path.
      */
-    List<HostAddress> getHosts(String splitPath, List<HostAddress> defaultAddresses);
+    List<HostAddress> getHosts(String splitKey, List<HostAddress> defaultAddresses);
+
+    // Include offset and length in key to allow different splits belonging to the same file to be cached on different nodes
+    // This can help with avoiding hotspots when multiple splits are created from a file
+    static String getSplitKey(String splitPath, long offset, long length)
```
jkylling, line 29: How deterministic are the offsets and lengths for different queries? E.g., if we run `SELECT x, y FROM t`, and then `SELECT x FROM t`, are we likely to get a cache miss or a cache hit with the second query? With the existing mechanism I believe we would get a cache hit.

If it's not deterministic, one option could be to do something like:
```java
static String getSplitKey(String splitPath, long offset)
{
    return splitPath + ":" + offset / resolution;
}
```
where `resolution` is `cachePageSize`, `trinoObjectStoreReadSize`, or some multiple of these.

raunaqmorarka: In delta and iceberg, the splits are deterministic across queries.
In hive, it depends on `hive.max-initial-splits` and `hive.max-initial-split-size`, I tweaked the code a bit there to account for that.

jkylling: In which scenarios do we read different lengths for the same offset?

To run through what will happen when reading a Parquet file:
By default we read data in 32 (?) MB chunks from object storage, and cache data in 1 MB chunks to local disk. With this change, 1 MB chunks which are at the boundary between two columns of the Parquet file we read will be cached on two nodes instead of one. This reduces the cache efficiency a bit, but it is a very reasonable trade off to get better load distribution.

jkylling: To try and answer my own question:
- (location, offset) always determines length uniquely for Delta: https://github.com/trinodb/trino/blob/4b67d9fd50dcff35d03e2a2a58bf2771e0897938/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeSplitManager.java#L318-L321
- For Iceberg this also seems to be the case, but is guaranteed by some code in Apache Iceberg.

Maybe we don't need the length in the key calculation?

jkylling (approved):

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CachingHostAddressProvider.java
raunaqmorarka: By default we read upto 8MB chunks, might be lower depending on physical size of queried columns and how far apart in the file they are.
The existing code is sending all the splits belonging to a file to the same node. This has the potential of creating hot spots in the cluster and reducing parallelism. Now we are including split start and end in the hash key to allow splits generated from the same file to be scheduled on different nodes. The only potential downside is that today file footer reads of different splits from same file can reuse the cache when they all land on the same node, but now they may land on different nodes and read and cache footers independently. But I think this is going to be offset by the improvement in parallelism and reduced hotspots.
Different length for the same offset can happen when for example the iceberg table property for split size is doubled. A split is uniquely identified by path, offset and length. Its easier to reason about behaviour with both offset and length included. We don't expect configurations around split sizes to be constantly changing between queries.

sopel39:
lgtm

PR #26681, opened 2025-09-22
Add `gcs.auth-type` configuration

chenjian2664:
This makes it easier to add new GCS authentication types in the future
also deprecated `gcs.use-access-token`

## Release notes

```markdown
## Section
* Replace `gcs.use-access-token` with `gcs.auth-type` configuration. ({issue}`26681`)
```

### DIFF
--- a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
+++ b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
@@ -21,6 +21,7 @@ import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import io.airlift.units.MinDuration;
 import jakarta.annotation.Nullable;
+import jakarta.validation.constraints.AssertFalse;
 import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.Min;
 import jakarta.validation.constraints.NotNull;
@@ -33,6 +34,12 @@ import static io.airlift.units.DataSize.Unit.MEGABYTE;
 
 public class GcsFileSystemConfig
 {
+    public enum AuthType
+    {
+        ACCESS_TOKEN,
+        SERVICE_ACCOUNT;
+    }
+
     private DataSize readBlockSize = DataSize.of(2, MEGABYTE);
     private DataSize writeBlockSize = DataSize.of(16, MEGABYTE);
     private int pageSize = 100;
@@ -41,7 +48,8 @@ public class GcsFileSystemConfig
     private String projectId;
     private Optional<String> endpoint = Optional.empty();
 
-    private boolean useGcsAccessToken;
+    private Optional<Boolean> useGcsAccessToken = Optional.empty();
+    private Optional<AuthType> authType = Optional.empty();
     private String jsonKey;
     private String jsonKeyFilePath;
     private int maxRetries = 20;
@@ -134,15 +142,33 @@ public class GcsFileSystemConfig
         return this;
     }
 
-    public boolean isUseGcsAccessToken()
+    @NotNull
+    public AuthType getAuthType()
     {
-        return useGcsAccessToken;
+        if (useGcsAccessToken.isPresent() && useGcsAccessToken.get()) {
+            return AuthType.ACCESS_TOKEN;
+        }
+        return authType.orElse(AuthType.SERVICE_ACCOUNT);
     }
 
+    @Config("gcs.auth-type")
+    public GcsFileSystemConfig setAuthType(AuthType authType)
+    {
+        this.authType = Optional.of(authType);
+        return this;
+    }
+
+    @Deprecated
+    public boolean isUseGcsAccessToken()
+    {
+        return useGcsAccessToken.orElse(false);
+    }
+
+    @Deprecated
     @Config("gcs.use-access-token")
     public GcsFileSystemConfig setUseGcsAccessToken(boolean useGcsAccessToken)
     {
-        this.useGcsAccessToken = useGcsAccessToken;
+        this.useGcsAccessToken = Optional.of(useGcsAccessToken);
         return this;
     }
 
@@ -268,13 +294,19 @@ public class GcsFileSystemConfig
         return minBackoffDelay.compareTo(maxBackoffDelay) <= 0;
     }
 
-    @AssertTrue(message = "Either gcs.use-access-token or gcs.json-key or gcs.json-key-file-path must be set")
+    @AssertTrue(message = "Either gcs.auth-type or gcs.json-key or gcs.json-key-file-path must be set")
     public boolean isAuthMethodValid()
     {
-        if (useGcsAccessToken) {
+        if (getAuthType() == AuthType.ACCESS_TOKEN) {
             return jsonKey == null && jsonKeyFilePath == null;
         }
 
         return (jsonKey == null) ^ (jsonKeyFilePath == null);
+    }
+
+    @AssertFalse(message = "Cannot set both gcs.use-access-token and gcs.auth-type")
+    public boolean isAuthTypeAndGcsAccessTokenConfigured()
+    {
+        return authType.isPresent() && useGcsAccessToken.isPresent();
     }
 }
--- a/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
+++ b/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
@@ -16,6 +16,8 @@ package io.trino.filesystem.gcs;
 import com.google.common.collect.ImmutableMap;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
+import io.trino.filesystem.gcs.GcsFileSystemConfig.AuthType;
+import jakarta.validation.constraints.AssertFalse;
 import jakarta.validation.constraints.AssertTrue;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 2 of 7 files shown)


wendigo:
I've restored compatibility mode for `gcs.use-access-token`

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -30,13 +31,26 @@
...
         ACCESS_TOKEN,
-        DEFAULT,
+        DEFAULT;
+
+        // Ensures backward compatibility with gcs.use-access-token config
+        // TRUE is mapped to ACCESS_TOKEN
+        // FALSE is mapped to DEFAULT
+        public static AuthType fromString(String value)
+        {
+            return switch (value.toUpperCase(ENGLISH)) {
```
losipiuk: hacky. I also wanted backward compatibility here but I think deprecating `setUseAccessToken` but still keeping it. And add validation that `setUseAccessToken` is not used together with `setAuthType` would be cleaner

marcinsbd: 1. +1 for the deprecation ( @ Deprecated and  even @ ConfigHidden) and mutual exclusiveness of those properties
2. should we add the ticket with next release after this one introducing new property for removing the deprecated property

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -157,17 +145,37 @@ public GcsFileSystemConfig setEndpoint(Optional<String> endpoint)
...
+    public boolean isUseGcsAccessToken()
+    {
+        return useGcsAccessToken;
+    }
+
+    @Deprecated
+    @Config("gcs.use-access-token")
+    public GcsFileSystemConfig setUseGcsAccessToken(boolean useGcsAccessToken)
+    {
+        checkArgument(authType == null, "Cannot set both gcs.use-access-token and gcs.auth-type");
```
wendigo: What if the authType is set after user-access-token? This will override the value without clear indication that this shouldn't be called. I think that this approach is not superior to the one I've already proposed.

losipiuk: This is not how validation works. 
It is run after all setters are run and we could model it in such way that you know that both setters were called - and that should result in failure.

wendigo: So the setUseGcsAccessToken should be optional in that case

losipiuk: Yeah - implementation wise changing `useGcsAccessToken` from `boolean` to `Optional<Boolean>` seems natural if we want to go this path.

chenjian2664: In `TestGcsFileSystemConfig#testDefaults` we need use set method explicitly, do you have idea how to check those two property?

losipiuk: testing defaults should be fine. you just assert on getters. 
As for testing explicit setting is a bit more tricky. But recently we added `assertFullMapping(Map<String, String> properties, T expected, Set<String> skipped)` - so you can test scenarios when one setter is used, and then when the other is used.

chenjian2664: @wendigo @losipiuk I'm not sure if this is okay, but please let me know if you have any advice on it

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
chenjian2664: @marcinsbd Agree with on " add the ticket with next release after this one introducing new property", I am going to add a draft pr for it, and wait current release finalized

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
losipiuk: I sent out commit with some suggestions. But `testDefaults` is still messy - can be fixed with adding `skipped` param to `assertRecordedDefaults` too.

chenjian2664: sent https://github.com/airlift/airlift/pull/1584

chenjian2664:
@wendigo @losipiuk PTAL, seems we can progress without changing airlift

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -33,6 +33,12 @@
 
 public class GcsFileSystemConfig
 {
+    public enum AuthType
+    {
+        ACCESS_TOKEN,
+        DEFAULT;
```
martint: Is there an authentication scheme named "DEFAULT" for GCS? Or did you pick that value as a way to represent the default scheme to use when nothing else is chosen? If so, let's use the actual name of the authentication scheme.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -268,13 +293,19 @@ public boolean isRetryDelayValid()
...
-        if (useGcsAccessToken) {
+        if (getAuthType() == AuthType.ACCESS_TOKEN) {
             return jsonKey == null && jsonKeyFilePath == null;
         }
 
         return (jsonKey == null) ^ (jsonKeyFilePath == null);
     }
+
+    @AssertTrue(message = "Cannot set both gcs.use-access-token and gcs.auth-type")
+    public boolean isAuthTypeAndUseGcsAccessTokenMutuallyExclusive()
```
martint: The name of this method reads funny. It sounds like a question whose answer should always be true. Rename it to `areAuthTypeAndGcsAccessTokenConfigured` and use AssertFalse

wendigo: This by convention needs to start with "is"

martint: No it doesn't -- unless the validation framework is not able to use the method without that prefix.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
chenjian2664: renamed to `SERVICE_ACCOUNT`

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -33,6 +34,12 @@
 
 public class GcsFileSystemConfig
 {
+    public enum AuthType
+    {
+        ACCESS_TOKEN,
+        SERVICE_ACCOUNT;
```
losipiuk, line 40: please update doc according to this chagne

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
```diff
@@ -125,5 +164,37 @@ public void testValidation()
                 "retryDelayValid",
                 "gcs.client.min-backoff-delay must be less than or equal to gcs.client.max-backoff-delay",
                 AssertTrue.class);
+
+        assertFailsValidation(
```
losipiuk, line 169: this fails now. Is it because of renamed "assert" method?

chenjian2664: Yes, fixed

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
chenjian2664: I think @wendigo is right, the validation method needs to start with "is", otherwise the test in `TestGcsFileSystemConfig` will fail due to validation, see the change https://github.com/trinodb/trino/pull/26681/commits/5512447feb53a6ff97b35be8b53bdc1f94390e6c

martint: Ok, then let's rename it to `isAuthTypeAndGcsAccessTokenConfigured`

losipiuk (commented):
LGTM - please squash

chenjian2664:
@wendigo PTAL

chenjian2664:
@losipiuk

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
chenjian2664: Added https://github.com/trinodb/trino/pull/26941 for removal of the configuration

ebyhr:
I'm not sure why we separated PR for `gcs.use-access-token` removal. It's marked as `@Deprecated`, but it's actually unused. This approach doesn't make sense to me.

PR #26573, opened 2025-09-07
Upgrade Oracle docker image for testing to oracle-free 23

takezoe:
## Description
Upgrade Oracle docker image for testing from oracle-xe 11 to oracle-free 23.

This will also allow us to run Oracle connector test cases on Apple Silicon Mac as oracle-free has arm64 image while oracle-xe doesn't.

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #26688, opened 2025-09-22
Add support for TopN pushdown for Oracle

### LINKED ISSUE #26566: [Bug] TopN pushdown doesn't work on Oracle Connector
## Description

TopN pushdown doesn't work using LIMIT and ORDER BY on Oracle Connector 
Trino version: 476

The following query  do TopN pushdown :
```
SELECT a.my_record_id, A.anno
FROM oracle_repo.myschema.PROVA2 a
LIMIT 10
```
but this one dont .
```
SELECT a.my_record_id, A.anno
FROM oracle_repo.myschema.PROVA2 a
ORDER BY a.MY_record_id ASC
LIMIT 10
```

## Expected Result:  

`ORDER BY ... LIMIT` TopN pushdown : should be [supported ](https://trino.io/docs/476/connector/oracle.html#pushdown) on Oracle Connector 

## Current Result:
```
                                                                         Query Plan                                                                          
-------------------------------------------------------------------------------------------------------------------------------------------------------------
 Trino version: 476                                                                                                                                          
 Fragment 0 [SINGLE]                                                                                                                                         
     Output layout: [my_record_id, anno]                                                                                                                     
     Output partitioning: SINGLE []                                                                                                                          
     Output[columnNames = [my_record_id, anno]]                                                                                                              
          Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                            
          Estimates: {rows: 10 (340B), cpu: 0, memory: 0B, network: 0B}                                                                                       
          TopN[count = 10, orderBy = [my_record_id ASC NULLS LAST]]                                                                                            
             Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                         
             Estimates: {rows: 10 (340B), cpu: ?, memory: ?, network: ?}                                                                                      
             LocalExchange[partitioning = SINGLE]                                                                                                              
                Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                      
                Estimates: {rows: ? (?), cpu: 0, memory: 0B, network: 0B}                                                                                     
                RemoteSource[sourceFragmentIds = [1]]                                                                                                          
                  Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                   
                                                                                                                                                             
 Fragment 1 [SOURCE]                                                                                                                                         
     Output layout: [my_record_id, anno]                                                                                                                     
     Output partitioning: SINGLE []                                                                                                                          
     TopNPartial[count = 10, orderBy = [my_record_id ASC NULLS LAST]]                                                                                        
          Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                            
          LocalExchange[partitioning = SINGLE]                                                                                                                 
             Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                         
             Estimates: {rows: ? (?), cpu: 0, memory: 0B, network: 0B}                                                                                        
             TopNPartial[count = 10, orderBy = [my_record_id ASC NULLS LAST]]                                                                                  
                Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                      
                LocalExchange[partitioning = ROUND_ROBIN]                                                                                                      
                   Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                   
                   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                  
                   TopNPartial[count = 10, orderBy = [my_record_id ASC NULLS LAST]]                                                                            
                      Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                                
                      TableScan[table = oracle_repo:myschema.prova2 myschema.PROVA2 columns=[MY_RECORD_ID:decimal(19,0):NUMBER, ANNO:decimal(38,8):NUMBER]] 
                        Layout: [my_record_id:decimal(19,0), anno:decimal(38,8)]                                                                             
                        Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                            
                        my_record_id := MY_RECORD_ID:decimal(19,0):NUMBER                                                                                    
                        anno := ANNO:decimal(38,8):NUMBER 
```

codluca:
Add support for TopN pushdown for Oracle

Fixes https://github.com/trinodb/trino/issues/26566

## Description

The newest FETCH FIRST number ROWS ONLY syntax is used (supported in Oracle version >=12).

Trino no longer supports older versions of Oracle (i.e. v11), which support only WHERE ROWNUM <= number syntax.

~We applied the FETCH FIRST syntax to the LIMIT expression, which was converted before to "SELECT * FROM (%s  ORDER BY %s) WHERE ROWNUM <= %s".~

Oracle sets NULL values last by default. Sometimes, however, the default does not work, so we used explicit flag for nulls.

## Release notes

```markdown
## Section
* Fix some things. ({issue}`26566`)
```

### DIFF
--- a/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
+++ b/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
@@ -33,6 +33,7 @@ import io.trino.plugin.jdbc.DoubleWriteFunction;
 import io.trino.plugin.jdbc.JdbcColumnHandle;
 import io.trino.plugin.jdbc.JdbcExpression;
 import io.trino.plugin.jdbc.JdbcJoinCondition;
+import io.trino.plugin.jdbc.JdbcSortItem;
 import io.trino.plugin.jdbc.JdbcTableHandle;
 import io.trino.plugin.jdbc.JdbcTypeHandle;
 import io.trino.plugin.jdbc.LongReadFunction;
@@ -103,9 +104,11 @@ import java.util.Set;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.function.BiFunction;
+import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Strings.emptyToNull;
 import static com.google.common.base.Verify.verify;
 import static io.airlift.slice.Slices.utf8Slice;
@@ -169,6 +172,7 @@ import static java.lang.String.join;
 import static java.sql.Types.FLOAT;
 import static java.util.Locale.ENGLISH;
 import static java.util.concurrent.TimeUnit.DAYS;
+import static java.util.stream.Collectors.joining;
 
 public class OracleClient
         extends BaseJdbcClient
@@ -226,6 +230,17 @@ public class OracleClient
             .put(DATE, WriteMapping.longMapping("date", trinoDateToOracleDateWriteFunction()))
             .put(TIMESTAMP_TZ_MILLIS, WriteMapping.longMapping("timestamp(3) with time zone", oracleTimestampWithTimeZoneWriteFunction()))
             .buildOrThrow();
+
+    private static final Set<String> ORACLE_COLLATABLE_TYPES = ImmutableSet.<String>builder()
+            .add("char")
+            .add("nchar")
+            .add("varchar")
+            .add("varchar2")
+            .add("nvarchar")
+            .add("nvarchar2")
+            .add("clob")
+            .add("nclob")
+            .build();
 
     private final boolean synonymsEnabled;
     private final ConnectorExpressionRewriter<ParameterizedExpression> connectorExpressionRewriter;
@@ -699,6 +714,71 @@ public class OracleClient
 
     @Override
     public boolean isLimitGuaranteed(ConnectorSession session)
+    {
+        return true;
+    }
+
+    @Override
+    public boolean supportsTopN(ConnectorSession session, JdbcTableHandle handle, List<JdbcSortItem> sortOrder)
+    {
+        for (JdbcSortItem sortItem : sortOrder) {
+            if (isCollatableInTrino(sortItem.column()) && !isCollatableInOracle(sortItem.column())) {
+                // If it is a Trino text type but its Oracle type is NOT collatable,
+                // we cannot guarantee correct sorting and must prevent pushdown.
+                return false;
+            }
+            // Non-textual types (numbers, dates, etc.) are safe for TopN pushdown
+            // as their sorting is same in Trino and Oracle
+        }
+        return true;
+    }
+
+    @Override
+    protected Optional<TopNFunction> topNFunction()
+    {
+        return Optional.of((query, sortItems, limit) -> {
+            String orderBy = sortItems.stream()
+                    .flatMap(sortItem -> {
+                        String collation = "";
+                        if (isCollatableInTrino(sortItem.column())) {
+                            checkState(
+                                    isCollatableInOracle(sortItem.column()),
+                                    "Column '%s' is collatable in Trino but not in Oracle. Check database configuration.",
+                                    sortItem.column().getColumnName());
+                            // Trino encodes all text as UTF-8, and sorts text as Unicode codepoints.
+                            // In Oracle BINARY collation provides unsigned byte-by-byte comparison.
+                            // If the Oracle DB uses UTF8 or AL32UTF8 charset, then BINARY collation will match Trino's sorting.
+                            // If the Oracle DB uses a non UTF8 charset, like WE8ISO8859P1 or WE8MSWIN1252,
+                            // then Oracle will not sort the same way as in Trino.
+                            collation = "COLLATE BINARY";
+                        }
+                        String ordering = sortItem.sortOrder().isAscending() ? "ASC" : "DESC";
+                        String nullsHandling = switch (sortItem.sortOrder()) {
+                            // In Oracle both ASC and DESC imply NULLS LAST, but we'll be explicit
+                            case ASC_NULLS_FIRST, DESC_NULLS_FIRST -> "NULLS FIRST";
+                            case ASC_NULLS_LAST, DESC_NULLS_LAST -> "NULLS LAST";
+                        };
+                        return Stream.of(format("%s %s %s %s", quoted(sortItem.column().getColumnName()), collation, ordering, nullsHandling));
+                    })
+                    .collect(joining(", "));
+            return format("%s ORDER BY %s FETCH FIRST %s ROWS ONLY", query, orderBy, limit);
+        });
+    }
+
+    private boolean isCollatableInTrino(JdbcColumnHandle column)
+    {
+        return column.getColumnType() instanceof CharType || column.getColumnType() instanceof VarcharType;
+    }
+
+    private boolean isCollatableInOracle(JdbcColumnHandle column)
+    {
+        String jdbcTypeName = column.getJdbcTypeHandle().jdbcTypeName()
+                .orElseThrow(() -> new TrinoException(JDBC_ERROR, "Type name is missing: " + column.getJdbcTypeHandle()));
+        return ORACLE_COLLATABLE_TYPES.contains(jdbcTypeName.toLowerCase(ENGLISH));
+    }
+
+    @Override
+    public boolean isTopNGuaranteed(ConnectorSession session)
     {
         return true;
     }
--- a/plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
+++ b/plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
@@ -62,7 +62,6 @@ public abstract class BaseOracleConnectorTest
                  SUPPORTS_RENAME_TABLE_ACROSS_SCHEMAS,
                  SUPPORTS_ROW_TYPE,
                  SUPPORTS_SET_COLUMN_TYPE,
-                 SUPPORTS_TOPN_PUSHDOWN,
                  SUPPORTS_MERGE,
                  SUPPORTS_ROW_LEVEL_UPDATE -> false;
             default -> super.hasBehavior(connectorBehavior);


plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
```diff
@@ -1119,7 +1119,7 @@ public void testNullSensitiveTopNPushdown()
         }
 
         try (TestTable testTable = newTrinoTable(
-                "test_null_sensitive_topn_pushdown",
+                "null_sens_topn",
```
martint: Unrelated change?

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
```diff
@@ -1183,7 +1183,7 @@ public void testCaseSensitiveTopNPushdown()
         PlanMatchPattern topNOverTableScan = project(node(TopNNode.class, anyTree(node(TableScanNode.class))));
 
         try (TestTable testTable = newTrinoTable(
-                "test_case_sensitive_topn_pushdown",
+                "case_sens_tp",
```
martint: Unrelated change?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +632,38 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+                        String columnSorting = format("%s %s", quoted(sortItem.column().getColumnName()), ordering);
+
+                        return switch (sortItem.sortOrder()) {
+                            // In Oracle both ASC and DESC imply NULLS LAST, but we'll be explicit
+                            case ASC_NULLS_LAST, DESC_NULLS_LAST -> Stream.of(format("%s NULLS LAST", columnSorting));
+                            case ASC_NULLS_FIRST, DESC_NULLS_FIRST -> Stream.of(format("%s NULLS FIRST", columnSorting));
+                        };
+                    })
+                    .collect(joining(", "));
+            return format("SELECT * FROM (%s  ORDER BY %s) WHERE ROWNUM <= %s", query, orderBy, limit);
```
martint: Do the ordering semantics in Oracle match those of Trino for all data types? In particular, for string-like data types due to collation rules.

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
codluca: There is a 30 characters limit for oracle tables.
The long name + the random suffix exceeded the limit.
The TopN tests are activated now for Oracle, so there is this issue with the max table name.

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
codluca: As above

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: I don't know the answer to this question.
These is no special treatment in any of the other connectors where topNFunction() is implemented.

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
martint: Ok, in that case let's rename it to `null_sensitive_topn` (avoid abbreviations)

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
martint: Similarly, let's rename this one to `case_sensitive_topn`

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
martint: Take a look at the postgresql plugin, which does:

```
if (isCollatable(sortItem.column())) {
    collation = "COLLATE \"C\"";
}
```

In particular, for strings that include accented words, mixed lower-case/upper case, unicode, how does Oracle sort them, and does it match how Trino sorts them? If we can't find that from Oracle documentation, it should be possible to run some tests in both systems and compare the results.

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
codluca: Fixed

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
codluca: Fixed

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: From my investigation, Trino's string comparison is byte-value-based.
PostgreSQL achieves byte-value-based collation by adding COLLATE "C" to the string columns. It's default collation is linguistic.
Oracle achieves byte-value-based collation by adding COLLATE BINARY to the string columns. It's default collation is BINARY, but it's safer to always enforce it.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,59 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+                        String ordering = sortItem.sortOrder().isAscending() ? "ASC" : "DESC";
+                        String nullsHandling = switch (sortItem.sortOrder()) {
+                            // In Oracle both ASC and DESC imply NULLS LAST, but we'll be explicit
+                            case ASC_NULLS_FIRST, DESC_NULLS_FIRST -> "NULLS FIRST";
+                            case ASC_NULLS_LAST, DESC_NULLS_LAST -> "NULLS LAST";
+                        };
+                        return Stream.of(format("%s %s %s %s", quoted(sortItem.column().getColumnName()), collation, ordering, nullsHandling));
+                    })
+                    .collect(joining(", "));
+            return format("SELECT * FROM (%s  ORDER BY %s) WHERE ROWNUM <= %s", query, orderBy, limit);
```
findepi: Nesting ORDER BY in a subquery followed by ROWNUM search requires some code-level explanation of the correctness. Note that generally in SQL, such nested ORDER BY can be freely ignored by a database.
It would be simpler if the connector just used the FETCH syntax.

findepi:
> Older versions of Oracle (i.e. v11) support only WHERE ROWNUM <= number syntax, not FETCH FIRST number ROWS ONLY.

according a google search, Oracle 11 is EOL and we generally do not support old versions of 3rd party products that are already sunset.

it should be fine to move on, especially when it simplifies connector code as it is in this case

codluca:
> > Older versions of Oracle (i.e. v11) support only WHERE ROWNUM <= number syntax, not FETCH FIRST number ROWS ONLY.
> 
> according a google search, Oracle 11 is EOL and we generally do not support old versions of 3rd party products that are already sunset.
> 
> it should be fine to move on, especially when it simplifies connector code as it is in this case

Should the upgrade be done as part of this task?
This implies modifying applyLimit() also.
And upgrading the container image and, possibly, some unit tests.

TestingOracleServer and TestJdbcVendorCompatibility use gvenzl/oracle-xe:**11.2.0.2-full**.
TestDbResourceGroupsOracleFlywayMigration uses gvenzl/oracle-xe:18.4.0-slim
Should we upgrade to gvenzl/oracle-xe:18.4.0-full or gvenzl/oracle-xe:18.4.0-slim?

ebyhr:
Could you rebase on master branch? We've already upgraded the version in https://github.com/trinodb/trino/pull/26573 thanks to @takezoe

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/BaseJdbcConnectorTest.java
```diff
@@ -1183,7 +1183,7 @@ public void testCaseSensitiveTopNPushdown()
         PlanMatchPattern topNOverTableScan = project(node(TopNNode.class, anyTree(node(TableScanNode.class))));
 
         try (TestTable testTable = newTrinoTable(
-                "test_case_sensitive_topn_pushdown",
+                "case_sensitive_topn",
```
ebyhr: We don't need to shorten these names anymore, right?

codluca: Correct. I restored them.

codluca:
> Could you rebase on master branch? We've already upgraded the version in #26573 thanks to @takezoe

Rebased. Thanks

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,59 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+                        String ordering = sortItem.sortOrder().isAscending() ? "ASC" : "DESC";
+                        String nullsHandling = switch (sortItem.sortOrder()) {
+                            // In Oracle both ASC and DESC imply NULLS LAST, but we'll be explicit
+                            case ASC_NULLS_FIRST, DESC_NULLS_FIRST -> "NULLS FIRST";
+                            case ASC_NULLS_LAST, DESC_NULLS_LAST -> "NULLS LAST";
+                        };
+                        return Stream.of(format("%s %s %s %s", quoted(sortItem.column().getColumnName()), collation, ordering, nullsHandling));
+                    })
+                    .collect(joining(", "));
+            return format("%s ORDER BY %s FETCH FIRST %s ROWS ONLY", query, orderBy, limit);
```
chenjian2664, line 764: Please update the pr description to reflect the changes

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -620,7 +634,7 @@ private static Optional<JdbcTypeHandle> toTypeHandle(DecimalType decimalType)
     @Override
     protected Optional<BiFunction<String, Long, String>> limitFunction()
     {
-        return Optional.of((sql, limit) -> format("SELECT * FROM (%s) WHERE ROWNUM <= %s", sql, limit));
+        return Optional.of((sql, limit) -> format("%s FETCH FIRST %s ROWS ONLY", sql, limit));
```
chenjian2664, line 712: unrelated change?

codluca: As support for oracle 11 was dropped, I figured it's better to also update the limitFunction() to use the FETCH FIRST %s ROWS ONLY syntax.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Updated the PR description.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
findepi: This is a good change, but not strictly related to TopN.
Would you mind running this as a separate PR (preferred) or at least a preparatory commit within this PR?

codluca: > This is a good change, but not strictly related to TopN. Would you mind running this as a separate PR (preferred) or at least a preparatory commit within this PR?

Should I open another issue and open the PR for the new issue?

I removed the update to the limitFunction() from this PR.

findepi: no need for new issue. you can create a PR without one

codluca: Here is the new PR: https://github.com/trinodb/trino/pull/26926

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,59 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+    @Override
+    protected Optional<TopNFunction> topNFunction()
+    {
+        return Optional.of((query, sortItems, limit) -> {
+            String orderBy = sortItems.stream()
+                    .flatMap(sortItem -> {
+                        String collation = "";
+                        if (isCollatable(sortItem.column())) {
+                            // In Oracle BINARY collation provides byte-based comparison
+                            collation = "COLLATE BINARY";
```
findepi: Treating the bytes as signed or unsigned?

It's worth a code comment explaining that Trino string codepoint-based collation is equivalent to sorting (unsigned) bytes ... in UTF-8 encoding.
When we sent `COLLATE BINARY` to oracle, are we assuming the data is compared byte-wise on Utf-8 encoding, or can it be something else?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,59 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+        }
+
+        // non-textual types don't have the concept of collation
+        return false;
+    }
+
+    private boolean isCollatable(String jdbcTypeName)
+    {
+        return ORACLE_COLLATABLE_TYPES.contains(jdbcTypeName);
+    }
```
findepi: inline

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,59 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+            return format("%s ORDER BY %s FETCH FIRST %s ROWS ONLY", query, orderBy, limit);
+        });
+    }
+
+    private boolean isCollatable(JdbcColumnHandle column)
+    {
+        if (column.getColumnType() instanceof CharType || column.getColumnType() instanceof VarcharType) {
+            String jdbcTypeName = column.getJdbcTypeHandle().jdbcTypeName()
+                    .orElseThrow(() -> new TrinoException(JDBC_ERROR, "Type name is missing: " + column.getJdbcTypeHandle()));
+            return isCollatable(jdbcTypeName);
```
findepi: When column is mapped to trino char/varchar, but not coming from a collatable type, we cannot expect Oracle to sort the data the way trino would.
when such column is involved, we should abstain from pushing the TopN

findepi: Is there any oracle type T mapped to trino type char or varchar that is none of:  char, nchar, varchar, varchar2, nvarchar, nvarchar2, clob, nclob?
can there be such in the future?

if answer to both questions is 'no', then ORACLE_COLLATABLE_TYPES set can be removed
if answer to any is 'yes', then for a column of such type T the code is probably going not to behave correctly

codluca: Added the check in supportsTopN() method.
If at least one Trino text sort column is not Oracle collatable, the pushdown is not made.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Done

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Added this comment:

> Used COLLATE BINARY for ORDER BY for string types, in order to enforce a comparison similar to Trino.
> 
> Trino encodes all text as UTF-8, and sorts text as Unicode codepoints.
> 
> In Oracle BINARY collation provides unsigned byte-by-byte comparison.
> 
> If the Oracle DB uses UTF8 or AL32UTF8 charset, then BINARY collation will match Trino's sorting.
> 
> If the Oracle DB uses a non UTF8 charset, like LATIN1, then characters outside ASCII range 
> will not sort the same way as in Trino.

We have a limitation here, which seems similar to what exists for PostgreSQL connector.
PostgreSQL COLLATE "C" and Oracle COLLATE BINARY will do a byte-for-byte comparison. If the DB encoding is UTF, then this will match Trino's UTF-8 codepoint comparison.

findepi: thanks. that is a good comment.
Is this why PostgreSQL connector uses COLLATE C? 

> If the Oracle DB uses a non UTF8 charset, like LATIN1, then characters outside ASCII range
will not sort the same way as in Trino.

this sounds like a potential correctness issue. thanks for calling it out! cc @ebyhr @martint 
I am not an Oracle DBA. Is it a known figure how often Oracle DBs uses non-UTF8 charset? Is it a theoretical problem or something to address? Can this be addressed automatically?


from https://docs.oracle.com/en/database/oracle/oracle-database/19/sqlrf/Data-Types.html#GUID-A3C0D836-BADB-44E5-A5D4-265BA5968483

> NVARCHAR2(size) | Variable-length Unicode character string having maximum length size characters. You must specify size for NVARCHAR2. The number of bytes can be up to two times size for AL16UTF16 encoding and three times size for UTF8 encoding.
> -- | --

it sounds like Oracle assumes each codepoint is at most 3 bytes in UTF-8 encoding, whereas actual UTF-8 can produce 4 bytes for a single codepoint.
What does Oracle do when string contains codepoints > U+FFFF (outside Basic Multilingual Plane, like emojis)?

codluca: We could add a boolean property to enable/disable the topN pushdown. The pushdown could be enabled by default, and the user could disable it in case the character encoding in not UTF8.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: > Is there any oracle type T mapped to trino type char or varchar that is none of: char, nchar, varchar, varchar2, nvarchar, nvarchar2, clob, nclob? can there be such in the future?
> 
> if answer to both questions is 'no', then ORACLE_COLLATABLE_TYPES set can be removed if answer to any is 'yes', then for a column of such type T the code is probably going not to behave correctly

Oracle has the LONG type, which has some restrictions regarding ORDER BY. This type is deprecated and is not mentioned in the connector type mapping.
From the type mapping it seems that there is a bijection between collatable types in Trino and Oracle.
To be on the safe side, I'd keep the ORACLE_COLLATABLE_TYPES check. But I can remove it, if needed.

For PostgreSQL, it has MONEY mapped to VARCHAR, so the COLLATABLE_TYPES check is needed.
It might have the issue mentioned in your comment.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,71 @@ public boolean isLimitGuaranteed(ConnectorSession session)
         return true;
     }
 
+    @Override
+    public boolean supportsTopN(ConnectorSession session, JdbcTableHandle handle, List<JdbcSortItem> sortOrder)
+    {
+        for (JdbcSortItem sortItem : sortOrder) {
+            if (isCollatableInTrino(sortItem.column())) {
+                if (!isCollatableInOracle(sortItem.column())) {
```
findepi: nit: can be one line

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,71 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+
+    @Override
+    protected Optional<TopNFunction> topNFunction()
+    {
+        return Optional.of((query, sortItems, limit) -> {
+            String orderBy = sortItems.stream()
+                    .flatMap(sortItem -> {
+                        String collation = "";
+                        if (isCollatableInTrino(sortItem.column())) {
+                            if (isCollatableInOracle(sortItem.column())) {
```
findepi: the second `if` can be turned into `checkState`

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,71 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+        for (JdbcSortItem sortItem : sortOrder) {
+            if (isCollatableInTrino(sortItem.column())) {
+                if (!isCollatableInOracle(sortItem.column())) {
+                    // If it is a Trino text type but its Oracle type is NOT collatable,
+                    // we cannot guarantee correct sorting and must prevent pushdown.
+                    return false;
+                }
+            }
+            // Non-textual types (numbers, dates, etc.) are always safe for TopN pushdown,
+            // as their sorting is typically binary and language-independent.
```
findepi:
```suggestion
            // Non-textual types (numbers, dates, etc.) are safe for TopN pushdown
            // as their sorting is same in Trino and Oracle
```

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +643,71 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+        return Optional.of((query, sortItems, limit) -> {
+            String orderBy = sortItems.stream()
+                    .flatMap(sortItem -> {
+                        String collation = "";
+                        if (isCollatableInTrino(sortItem.column())) {
+                            if (isCollatableInOracle(sortItem.column())) {
+                                // Trino encodes all text as UTF-8, and sorts text as Unicode codepoints.
+                                // In Oracle BINARY collation provides unsigned byte-by-byte comparison.
+                                // If the Oracle DB uses UTF8 or AL32UTF8 charset, then BINARY collation will match Trino's sorting.
+                                // If the Oracle DB uses a non UTF8 charset, like LATIN1, then characters outside ASCII range
```
findepi: not sure whether LATIN1 is a good example.

the following is true for every character c representable in latin1 encoding: 
```
unsigned byte value with which c is encoded in latin1 = unicode code point of c
```

obviously this is not true for any other encoding, so maybe just change example in the comment

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
findepi: thanks. please also try filling in the Release notes section

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
findepi: continued in https://github.com/trinodb/trino/pull/26688#discussion_r2427015915

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
findepi: Is it possible to inspect oracle server side setting to know whether pushdown is safe to do?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Changed

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Changed

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Used checkState

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
codluca: Replaced with WE8ISO8859P1 or WE8MSWIN1252

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +644,71 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+                        if (isCollatableInTrino(sortItem.column())) {
+                            checkState(
+                                    isCollatableInOracle(sortItem.column()),
+                                    "Column '%s' is collatable in Trino but not in Oracle. Check database configuration.",
+                                    sortItem.column().getColumnName());
+                            // Trino encodes all text as UTF-8, and sorts text as Unicode codepoints.
+                            // In Oracle BINARY collation provides unsigned byte-by-byte comparison.
+                            // If the Oracle DB uses UTF8 or AL32UTF8 charset, then BINARY collation will match Trino's sorting.
+                            // If the Oracle DB uses a non UTF8 charset, like WE8ISO8859P1 or WE8MSWIN1252,
+                            // then Oracle will not sort the same way as in Trino.
```
findepi, line 752: @kokosing please help me understand whether this is a very unlikely edge case that we can leave to be addressed later, or a pretty common thing, requiring this new feature to be opt-in

also, do we need a kill-switch?

findepi (approved):

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
kokosing: Collations are common.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -629,6 +644,71 @@ public boolean isLimitGuaranteed(ConnectorSession session)
...
+    }
+
+    private boolean isCollatableInTrino(JdbcColumnHandle column)
+    {
+        return column.getColumnType() instanceof CharType || column.getColumnType() instanceof VarcharType;
+    }
+
+    private boolean isCollatableInOracle(JdbcColumnHandle column)
+    {
+        String jdbcTypeName = column.getJdbcTypeHandle().jdbcTypeName()
```
kokosing, line 775: We also have `io.trino.plugin.jdbc.JdbcTypeHandle#caseSensitivity`. Can we use it to determine types with the same collation as in Trino?

codluca: CaseSensitivity enum has 2 values: CASE_SENSITIVE and CASE_INSENSITIVE.
It doesn't seem suited for this.

There could be a property to enable or disable the pushdown:
i.e. oracle.topn.pushdown.enabled
If it's ok, what should be the default value (true or false)?

kokosing: > It doesn't seem suited for this.

What needs to be changed to support this?

> There could be a property to enable or disable the pushdown:

We already have a property for unsafe pushdowns, maybe we could reuse that? @findepi WDYT?

Issue #26464, opened 2025-08-22
Internal error if pattern variables are not uppercase


OrangieLou:
Hi,

I experience an internal error with no further error message if the pattern variable in the `AFTER MATCH` subclause of a `MATCH_RECOGNIZE` query is lowercase instead of uppercase.

Minimal working example:
```sql
SELECT mr.a, mr.b, mr.c
FROM (SELECT 1 AS a, 6 AS b UNION SELECT 2 AS a, 4 AS b UNION SELECT 3 AS a, 8 AS b)
MATCH_RECOGNIZE (
    ORDER BY a
    MEASURES CLASSIFIER() AS c
    ALL ROWS PER MATCH
    AFTER MATCH SKIP TO FIRST Y
    PATTERN (x+ y z?)
    DEFINE
        x AS x.b > 5,
        y AS y.b < x.b
) AS mr;

```

While this query runs fine, switching `AFTER MATCH SKIP TO FIRST Y` to `AFTER MATCH SKIP TO FIRST y` (lowercase `y`) the new query fails. 

```
Query FAILED, 2 nodes
Splits: 131 total, 3 done (2.29%)
0.02 [2 rows, 0B] [100 rows/s, 0B/s]

Query failed: Internal error
```

This feels unexpected since uppercase/lowercase should not matter and it does not in all the other places.

I tried this in a minimal setup running trino 476 in docker (2 workers, 1 coordinator, 1 ubuntu container with the trino cli).

Cheers!

martint:
cc @kasiafi

codluca:
Created pull request: https://github.com/trinodb/trino/pull/26691

PR #26691, opened 2025-09-23
Fix internal error if pattern variables are not uppercase

codluca:
Fix internal error if pattern variables are not uppercase

Fixes https://github.com/trinodb/trino/issues/26464

## Description

Replaced using directly skipTo.identifier.value, and used analysis::getResolvedLabel method, which is used in other places in RelationPlanner.planPatternRecognitionComponents method.

## Release notes

```markdown
## General
* Fix internal planner error if query pattern variables are not uppercase ({issue}`26464`)
```


wendigo:
Thanks @codluca !

PR #26696, opened 2025-09-23
Warn if the spooling is enabled but client lacks support

wendigo:
Warning can be enabled on the server-side by setting configuration property
`protocol.spooling.unsupported-warning.enabled` to `true` (default is `false`)
and disabled via session property `spooling_unsupported_warning` set to `false`
(`true` by default).

## Release notes

```markdown
## Spooling protocol
* Generate query-level warning if server supports spooling but client does not ({issue}`issuenumber`)
```


martint:
Why is this a warning condition? It's perfectly valid for a client not to support the parallel protocol even if the server does.

wendigo:
@martint this is to incentivise old clients to upgrade to benefit from spooling. It's behind the configuration property so it can be disabled.

PR #26697, opened 2025-09-23
Make Trino headers case insensitive

jzaralim:
## Description

From [RFC 2616](https://datatracker.ietf.org/doc/html/rfc2616#section-4.2):

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
+++ b/core/trino-main/src/main/java/io/trino/server/HttpRequestSessionContextFactory.java
@@ -292,7 +292,10 @@ public class HttpRequestSessionContextFactory
 
     private static List<String> splitHttpHeader(MultivaluedMap<String, String> headers, String name)
     {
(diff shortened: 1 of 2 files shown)


wendigo:
@jzaralim can you share background for this change? I believe that this is handled by Jetty already

wendigo:
From the Jakarta's `HttpHeaders.getRequestHeaders`: 

CleanShot 2025-09-23 at 18 34 26@2x

jzaralim:
I ran into this issue while working on an MCP tool. Airlift's MCP library uses `HttpServletRequest`, and `HttpServletRequest.getHeaderNames()` is not case sensitive. If this doesn't belong here, I could make the same change on my end.

wendigo:
@jzaralim These are two different layers. Trino endpoints are jakarta which is case insensitive. I don't know about Airlift but if there is an issue with MCP it belongs there, not here

Issue #26038, opened 2025-06-19
Inefficient queries involving "OR" operator on different columns


cccs-eric:
We have witnessed this problem using the Iceberg plugin.  The returned rows from those queries are as expected, but a lot more files (splits) are being scanned for nothing since Iceberg metrics are not being leveraged.  We believe this is caused by a bug in `TupleDomain.columnWiseUnion()`.

# How To Reproduce
It is very easy to reproduce the issue by watching the number of splits in the UI while running queries on the following table:
```sql
-- I created that table using SparkSQL
CREATE TABLE
    my_catalog.my_namespace.trino_or_bug TBLPROPERTIES (
        'write.target-file-size-bytes' = '524288', -- Make small files to amplify the problem.
        'write.metadata.metrics.default' = 'none', -- We don't want metrics on every column to show what a full scan does.
        'write.metadata.metrics.column.str_col_1' = 'truncate(16)', -- Iceberg metric default
        'write.metadata.metrics.column.str_col_2' = 'truncate(16)', -- Iceberg metric default
        'write.metadata.metrics.column.int_col_1' = 'truncate(16)', -- Iceberg metric default
        'write.metadata.metrics.column.int_col_2' = 'truncate(16)', -- Iceberg metric default
        'write.metadata.metrics.column.ts_col_1' = 'truncate(16)',  -- Iceberg metric default
        'write.metadata.metrics.column.ts_col_2' = 'truncate(16)'   -- Iceberg metric default
    ) AS
SELECT
    'str_1' AS str_col_1,
    'str_2' AS str_col_2,
    'str_3' AS str_col_3,
    1 AS int_col_1,
    2 AS int_col_2,
    3 AS int_col_3,
    TO_TIMESTAMP('2025-06-01 00:00:00') AS ts_col_1,
    TO_TIMESTAMP('2025-06-02 00:00:00') AS ts_col_2,
    TO_TIMESTAMP('2025-06-03 00:00:00') AS ts_col_3
FROM
    range(100 * 1000 * 1000)
```
We have static values for each column, and we have 3 different data types to show the problem is not type-dependent.  The 3rd column of each type do not have metrics to force Trino into a full table scan when those columns are queried.

While monitoring the Trino UI, run the following query:
```sql
select count(*) from my_catalog.my_namespace.trino_or_bug
where int_col_1 = -1;
```
The number of completed splits should be very small (18 for me).

Do the same for `int_col_2`:
```sql
[...]

PR #26698, opened 2025-09-23
Add full connector expression to split factory constraint

dain:
## Description

The constraint passed to connector split factory does not contain the ConnectorExpression, which limits the ability to perform split time filtering.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/planner/SplitSourceFactory.java
+++ b/core/trino-main/src/main/java/io/trino/sql/planner/SplitSourceFactory.java
@@ -78,8 +78,10 @@ import io.trino.sql.planner.plan.WindowNode;
 
 import java.util.List;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.Optional;
 
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static io.trino.spi.connector.DynamicFilter.EMPTY;
 import static io.trino.sql.ir.Booleans.TRUE;
@@ -179,6 +181,9 @@ public class SplitSourceFactory
             // we are interested only in functional predicate here, so we set the summary to ALL.
             Constraint constraint = new Constraint(
                     TupleDomain.all(),
+                    ConnectorExpressionTranslator.translateConjuncts(session, nonDynamicFilter).connectorExpression(),
+                    assignments.entrySet().stream()
+                            .collect(toImmutableMap(entry -> entry.getKey().name(), Entry::getValue)),
                     evaluator::isCandidate,
                     evaluator.getArguments());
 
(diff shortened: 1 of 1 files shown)


electrum (approved):

core/trino-main/src/main/java/io/trino/sql/planner/SplitSourceFactory.java
```diff
@@ -173,11 +175,17 @@ private SplitSource createSplitSource(TableHandle table, Map<Symbol, ColumnHandl
...
+            Expression nonDynamicFilter = filterConjuncts(filterPredicate.orElse(TRUE), expression -> !DynamicFilters.isDynamicFilter(expression));
+            LayoutConstraintEvaluator evaluator = new LayoutConstraintEvaluator(plannerContext, session, assignments, nonDynamicFilter);
+
+            // we are interested only in functional predicate here, so we set the summary to ALL.
+            Constraint constraint = new Constraint(
+                    TupleDomain.all(),
+                    ConnectorExpressionTranslator.translateConjuncts(session, nonDynamicFilter).connectorExpression(),
+                    assignments.entrySet().stream()
+                            .collect(toImmutableMap(entry -> entry.getKey().name(), Entry::getValue)),
+                    evaluator::isCandidate,
```
raunaqmorarka, line 187: I think this is differing from the contract specified in `io.trino.spi.connector.Constraint#predicate`
```
    * A predicate that can be used to filter data. If present, it is equivalent to, or stricter than, {@link #getSummary()} and different from, and should be AND-ed with, {@link #getExpression()}
```
`io.trino.spi.connector.Constraint#getExpression`
```
an expression predicate which is different from, and should be AND-ed with, {@link #getSummary} or {@link #predicate} (if present).
```
It looks like what we're doing now is providing the same predicate through `evaluator::isCandidate` and ConnectorExpression.
It will help to change the javadocs and clarify that these can be the same thing

raunaqmorarka: The iceberg connector is using `io.trino.spi.connector.Constraint#predicate` today to filter splits based on partition values when there is a non-TupleDomain filter. I think it would be interesting to be able to use non-TupleDomain predicates to prune splits based on file min/max as well and not just partition values. e.g. https://github.com/trinodb/trino/issues/26038
But that requires translating the ConnectorExpression to iceberg expression.

dain: I didn't touch the Constraint because I don't really understand the implications of the docs.  I'm not really sure why these are marked as related in the javadoc.  I would expect that you use one or the other.  I'd be happy to update the docs, if you have suggested text.

As for the iceberg improvement, this is exactally why this is useful.  I have a custom connector and I can filter splits quickly if I know the connector expression.  In my case I have no benefit with apply filter at the table level, so for me this is way simpler than stuffing predicates into a table handle just to unpack here.

Issue #26257, opened 2025-07-22
Invalid query result when `IS NOT DISTINCT FROM` used as a JOIN condition


PavelkoSemen:
Hello everyone!
Has anyone encountered such behavior in Trino: queries using NOT DISTINCT FROM return 0 rows after a JOIN.If in the query we leave only the part of the condition for K3, where both values are equal to NULL, then one row is returned.
If we use K1 and K2, one row is also returned.
However, when all three keys (K1, K2, K3) are specified in the condition, the query returns 0 rows.

```sql 
CREATE  TABLE ice.test.ys_stage1
(
  k1 int,
  k2 int,
  k3 int,
  a1 int,
  a2 int,
  a3 int
);

CREATE  TABLE ice.test.ys_merge1
(
  k1 int,
  k2 int,
  k3 int,
  a1 int,
  a2 int,
  a3 int
);

INSERT INTO ice.test.ys_stage1 VALUES(0,0, NULL,9,9,9);
INSERT INTO ice.test.ys_merge1 VALUES(0,0, NULL,9,9,9);

# 0 row
SELECT
  count(*)
FROM ice.test.ys_stage1 t
  INNER JOIN ice.test.ys_merge1 AS s
    ON 1=1
    AND s.k1 IS NOT DISTINCT FROM t.k1
    AND s.k2 IS NOT DISTINCT FROM t.k2
    AND s.k3 IS NOT DISTINCT FROM t.k3;    

 # 1 row
SELECT
  count(*)
FROM ice.test.ys_stage1 t
  INNER JOIN ice.test.ys_merge1 AS s
    ON 1=1
    AND ((s.k1 = t.k1) OR (s.k1 IS NULL AND t.k1 IS NULL) )
    AND ((s.k2 = t.k2) OR (s.k2 IS NULL AND t.k2 IS NULL) )
    AND ((s.k3 = t.k3) OR (s.k3 IS NULL AND t.k3 IS NULL) );
```

wendigo:
cc @kasiafi

kasiafi:
The issue is related to dynamic filters. Investigating further.

kasiafi:
This seems to be related to
- how we represent dynamic filter domains with `null` values
- how we optimize the resulting tuple domain

When we collect a dynamic filter, and there are only `null` values, the resulting domain is `NONE`. In the example above, we collect the values of the `ice.test.ys_merge1.k3` column and get a `NONE` domain.
[...]

PR #26702, opened 2025-09-24
Fix 'IS NOT DISTINCT FROM' not working in JOIN when nulls are used

codluca:
Fix 'IS NOT DISTINCT FROM' not working in JOIN when nulls are the only value.

Fixes issue: https://github.com/trinodb/trino/issues/26257

## Description

The changes try to enable Domain with only null value to be used.

Modify JoinDomainBuilder to take into consideration the nulls from the value blocks.

## Release notes

```markdown
## Section
* Fix some things. ({issue}`26257`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
+++ b/core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
@@ -99,6 +99,14 @@ public class JoinDomainBuilder
 
     private long retainedSizeInBytes = INSTANCE_SIZE;
 
+    /**
+     * Indicates whether null values are allowed in the join domain.
+     * This is set to true if any null values are observed in the input blocks
+     * during domain building, and is used to determine whether the resulting
+     * domain should include nulls.
+     */
+    private boolean nullAllowed;
+
     public JoinDomainBuilder(
             Type type,
             int maxDistinctValues,
@@ -160,6 +168,9 @@ public class JoinDomainBuilder
 
     public void add(Block block)
     {
+        if (block.hasNull()) {
+            nullAllowed = true;
+        }
         if (collectDistinctValues) {
             switch (block) {
                 case ValueBlock valueBlock -> {
@@ -290,24 +301,22 @@ public class JoinDomainBuilder
                     }
                 }
             }
-            // Inner and right join doesn't match rows with null key column values.
-            return Domain.create(ValueSet.copyOf(type, values.build()), false);
+            return Domain.create(ValueSet.copyOf(type, values.build()), nullAllowed);
         }
         if (collectMinMax) {
             if (minValue == null) {
                 // all values were null
-                return Domain.none(type);
+                return nullAllowed ? Domain.onlyNull(type) : Domain.none(type);
             }
             Object min = readNativeValue(type, minValue, 0);
             Object max = readNativeValue(type, maxValue, 0);
-            return Domain.create(ValueSet.ofRanges(range(type, min, true, max, true)), false);
+            return Domain.create(ValueSet.ofRanges(range(type, min, true, max, true)), nullAllowed);
         }
         return Domain.all(type);
     }
 
     private void add(ValueBlock block, int position)
     {
-        // Inner and right join doesn't match rows with null key column values.
         if (block.isNull(position)) {
             return;
         }
--- a/core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
+++ b/core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
@@ -296,20 +296,21 @@ public final class DynamicFilters
                 return domain;
             }
             if (domain.isNone()) {
-                // Dynamic filter collection skips nulls
-                // In case of IS NOT DISTINCT FROM, an empty Domain should still allow null
-                if (nullAllowed) {
-                    return Domain.onlyNull(domain.getType());
-                }
                 return domain;
+            }
+            if (domain.isOnlyNull()) {
+                if (nullAllowed) {
+                    return domain;
+                }
+                return Domain.none(domain.getType());
             }
             Range span = domain.getValues().getRanges().getSpan();
             return switch (operator) {
                 case EQUAL -> {
                     if (nullAllowed) {
-                        yield Domain.create(domain.getValues(), true);
+                        yield domain;
                     }
-                    yield domain;
+                    yield Domain.create(domain.getValues(), false);
                 }
                 case LESS_THAN -> {
                     Range range = Range.lessThan(span.getType(), span.getHighBoundedValue());
--- a/testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
+++ b/testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
@@ -1,0 +1,115 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.tests;
+
+import com.google.common.collect.ImmutableMap;
+import io.trino.plugin.memory.MemoryPlugin;
+import io.trino.sql.query.QueryAssertions;
+import io.trino.testing.QueryRunner;
+import io.trino.testing.StandaloneQueryRunner;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+import static io.trino.testing.TestingNames.randomNameSuffix;
+import static io.trino.testing.TestingSession.testSessionBuilder;
+import static java.lang.String.format;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@TestInstance(PER_CLASS)
+@Execution(ExecutionMode.SAME_THREAD)
+class TestJoinIsNotDistinct
+{
+    private static final String LOCAL_CATALOG = "local";
+    private static final String DEFAULT_SCHEMA = "default";
+
+    private final QueryRunner queryRunner;
+
+    private final QueryAssertions assertions;
+
+    TestJoinIsNotDistinct()
+    {
+        queryRunner = new StandaloneQueryRunner(testSessionBuilder()
+                .setCatalog(LOCAL_CATALOG)
+                .setSchema(DEFAULT_SCHEMA)
+                .build());
+        queryRunner.installPlugin(new MemoryPlugin());
+        queryRunner.createCatalog(LOCAL_CATALOG, "memory", ImmutableMap.of());
+
+        assertions = new QueryAssertions(queryRunner);
+    }
+
+    @AfterAll
+    void teardown()
+    {
+        assertions.close();
+    }
+
+    @Test
+    void testJoinWithIsNotDistinctFromOnNulls()
+    {
+        String tableName1 = "test_tab_" + randomNameSuffix();
+        String tableName2 = "test_tab_" + randomNameSuffix();
+        queryRunner.execute(format("CREATE TABLE %s (k1 INT, k2 INT)", tableName1));
+        queryRunner.execute(format("CREATE TABLE %s (k1 INT, k2 INT)", tableName2));
+
+        queryRunner.execute(format("INSERT INTO %s VALUES (1, NULL), (NULL, NULL), (NULL, 2), (2, 2)", tableName1));
+        queryRunner.execute(format("INSERT INTO %s VALUES (1, NULL), (NULL, NULL), (3, NULL), (2, 2)", tableName2));
+
+        assertThat(assertions.query(format("""
+                SELECT *
+                FROM %s t
+                INNER JOIN %s AS s
+                    ON  s.k1 IS NOT DISTINCT FROM t.k1
+                    AND s.k2 IS NOT DISTINCT FROM t.k2
+                """, tableName1, tableName2)))
+                .matches("VALUES (1, CAST(NULL AS INTEGER), 1, CAST(NULL AS INTEGER))," +
+                         " (CAST(NULL AS INTEGER), CAST(NULL AS INTEGER), CAST(NULL AS INTEGER), CAST(NULL AS INTEGER))," +
+                         " (2, 2, 2, 2)");
+    }
+
+    @Test
+    void testJoinWithIsNotDistinctFromOnNullsOnDerivedTables()
+    {
+        assertThat(assertions.query("""
+                SELECT *
+                FROM (SELECT 1 AS k1, NULL AS k2) t
+                INNER JOIN (SELECT 1 AS k1, NULL AS k2) AS s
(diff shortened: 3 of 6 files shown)


codluca:
@raunaqmorarka The pull request is ready.
There is a failed check (ci / test plugin/trino-pinot), but it is something random, not related to the code changes.

testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
```diff
@@ -0,0 +1,155 @@
...
+import io.trino.testing.QueryRunner;
+import io.trino.tests.tpch.TpchQueryRunner;
+import org.junit.jupiter.api.Test;
+
+import static io.trino.testing.TestingNames.randomNameSuffix;
+import static io.trino.testing.TestingSession.testSessionBuilder;
+import static java.lang.String.format;
+
+public class TestJoinIsNotDistinct
+        extends AbstractTestQueryFramework
```
martint: Use QueryAssertions instead of AbstractTestQueryFramework. Take a look at the tests under `io.trino.sql.query`.

codluca: If on master (without the fix) I add this test to io.trino.sql.query.TestJoin, it will succeed:
```
    @Test
    public void testJoinWithIsNotDistinctFromOnNullsInMemory()
    {
        assertThat(assertions.query(
                """
                WITH
                  a AS (SELECT k1, k2 FROM (VALUES (1, NULL)) AS t(k1, k2)),
                  b AS (SELECT k1, k2 FROM (VALUES (1, NULL)) AS t(k1, k2))
                SELECT *
                FROM a
                INNER JOIN b ON a.k1 IS NOT DISTINCT FROM b.k1
                            AND a.k2 IS NOT DISTINCT FROM b.k2
                """))
                .matches("VALUES (1, NULL, 1, NULL)");
    }
```
The code doesn't follow the same path, using the JoinDomainBuilder, if the temporary tables are used. The Domain is build differently.
I added the extra test from below (testJoinWithIsNotDistinctFromOnNullsInMemory) just for the future, as the test was working before the fix.

martint: You should be able to use temporary tables with QueryAssertions, as well. You just need to initialize QueryAssertions with the memory plugin and populate the tables for the test.

codluca: Replaced with QueryAssertions

YuriyGavrilov:
@wendigo could you please review it?

YuriyGavrilov:
@martint hi, it seam to be ready to merge?  thx forward

wendigo:
@YuriyGavrilov I don't have enough context here to decide whether fix is correct or not so I defer decision to @martint or @findepi

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
```diff
@@ -160,6 +168,9 @@ public boolean isCollecting()
 
     public void add(Block block)
     {
+        if (block.hasNull()) {
+            nullsAllowed = true;
```
findepi: is this applied only to join conditions that use IS NOT DISTINCT FROM, or also to those using `=` or `ON (...)`?

codluca: This test does not reach the code above:
* = inside ON
* = inside WHERE
```
assertThat(assertions.query(format("SELECT *" +
        " FROM %s t" +
        " INNER JOIN %s AS s" +
        "    ON ((s.k1 IS NULL AND t.k1 IS NULL) OR s.k1=t.k1)" +
        "    AND ((s.k2 IS NULL AND t.k2 IS NULL) OR s.k2=t.k2)", tableName1, tableName2)))
        .matches("VALUES (1, CAST(NULL AS INTEGER), 1, CAST(NULL AS INTEGER))");

assertThat(assertions.query(format("SELECT *" +
        " FROM %s t" +
        " INNER JOIN %s AS s ON 1=1" +
        "    WHERE ((s.k1 IS NULL AND t.k1 IS NULL) OR s.k1=t.k1)" +
        "    AND ((s.k2 IS NULL AND t.k2 IS NULL) OR s.k2=t.k2)", tableName1, tableName2)))
        .matches("VALUES (1, CAST(NULL AS INTEGER), 1, CAST(NULL AS INTEGER))");
```
This test reaches the code above:
* IS NOT DISTINCT FROM inside ON
* IS NOT DISTINCT FROM inside WHERE
```
assertThat(assertions.query(format("SELECT *" +
        " FROM %s t" +
        " INNER JOIN %s AS s" +
        "    ON s.k1 IS NOT DISTINCT FROM t.k1" +
        "    AND s.k2 IS NOT DISTINCT FROM t.k2", tableName1, tableName2)))
        .matches("VALUES (1, CAST(NULL AS INTEGER), 1, CAST(NULL AS INTEGER))");

assertThat(assertions.query(format("SELECT *" +
        " FROM %s t" +
        " INNER JOIN %s AS s ON 1=1" +
        "    WHERE s.k1 IS NOT DISTINCT FROM t.k1" +
        "    AND s.k2 IS NOT DISTINCT FROM t.k2", tableName1, tableName2)))
        .matches("VALUES (1, CAST(NULL AS INTEGER), 1, CAST(NULL AS INTEGER))");
```

findepi: > This test does not reach the code above:
> 
> * = inside ON

That's surprising to me. It looked to me this is general backbone for creating dynamic filters.
We definitely produce DF for `=` conditions.

cc @sopel39  @raunaqmorarka in case you know what i am missing.

YuriyGavrilov:
Help somebody with this

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
chenjian2664: cc @kasiafi

findepi: @sopel39 can you PTAL?

YuriyGavrilov:
> This pull request has gone a while without any activity. Ask for help on #core-dev on Trino slack.

This is not stale, just need someone who can review it

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
```diff
@@ -290,8 +301,7 @@ public Domain build()
                     }
                 }
             }
-            // Inner and right join doesn't match rows with null key column values.
-            return Domain.create(ValueSet.copyOf(type, values.build()), false);
+            return Domain.create(ValueSet.copyOf(type, values.build()), nullAllowed);
```
kasiafi, line 304: We should also consider the `nullAllowed` information when creating the Domain in the `collectMinMax` mode (below).

core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
```diff
@@ -295,19 +295,18 @@ public Domain applyComparison(Domain domain)
...
-            if (domain.isNone()) {
-                // Dynamic filter collection skips nulls
+            if (domain.getValues().isNone()) {
                 // In case of IS NOT DISTINCT FROM, an empty Domain should still allow null
                 if (nullAllowed) {
                     return Domain.onlyNull(domain.getType());
                 }
-                return domain;
+                return Domain.none(domain.getType());
             }
```
kasiafi, line 306: Refactor as:
```
if (domain.isNone()) {
    return domain;
}
if (domain.isOnlyNull()) {
    if (nullAllowed) {
        return domain;
    }
    return Domain.none(domain.getType());
}
```

core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
```diff
@@ -295,19 +295,18 @@ public Domain applyComparison(Domain domain)
...
             Range span = domain.getValues().getRanges().getSpan();
             return switch (operator) {
                 case EQUAL -> {
-                    if (nullAllowed) {
-                        yield Domain.create(domain.getValues(), true);
+                    if (nullAllowed != domain.isNullAllowed()) {
+                        yield Domain.create(domain.getValues(), nullAllowed);
                     }
                     yield domain;
                 }
```
kasiafi, line 314: Refactor as:
```
case EQUAL -> {
    if (nullAllowed) {
        yield domain;
    }
    yield Domain.create(domain.getValues(), false);
}
```

kasiafi (commented):
LGTM but @raunaqmorarka PTAL

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
codluca: Ok

core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
codluca: Changed

core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
codluca: Changed

wendigo:
Test failures are related

core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
codluca: There are some tests failing now.
For instance:
```
Error:  Errors: 
Error:    TestFaultTolerantExecutionDynamicFiltering
>AbstractTestCoordinatorDynamicFiltering.testIsNotDistinctFromJoinWithSelectiveBuildSide:244
->AbstractTestCoordinatorDynamicFiltering.assertQueryDynamicFilters:467
->AbstractTestCoordinatorDynamicFiltering.assertQueryDynamicFilters:472
->AbstractTestCoordinatorDynamicFiltering.assertQueryDynamicFilters:485
->AbstractTestQueryFramework.computeActual:319 » QueryFailed 
```
```
Suppressed: java.lang.Exception: SQL: SELECT * FROM store_sales JOIN tpcds.tiny.store ON store_sales.ss_sold_date_sk IS NOT DISTINCT FROM store.s_closed_date_sk AND store.s_closed_date_sk < 0
		at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:595)
		... 11 more
Caused by: org.opentest4j.AssertionFailedError: 
expected: {TestingColumnHandle{OptionalInt[0], name=ss_sold_date_sk, type=Optional[bigint]}=[NULL]}
 but was: NONE
```

It does not fail with the previous code.
I'll paste it here for reference, maybe it helps:
```java
        public Domain applyComparison(Domain domain)
        {
            if (domain.isAll()) {
                return domain;
            }
            if (domain.getValues().isNone()) {
                // In case of IS NOT DISTINCT FROM, an empty Domain should still allow null
                if (nullAllowed) {
                    return Domain.onlyNull(domain.getType());
                }
                return Domain.none(domain.getType());
            }
            Range span = domain.getValues().getRanges().getSpan();
            return switch (operator) {
                case EQUAL -> {
                    if (nullAllowed != domain.isNullAllowed()) {
                        yield Domain.create(domain.getValues(), nullAllowed);
                    }
                    yield domain;
                }
```

codluca:
> Test failures are related

I'm not sure how to proceed. I don't know if there is an issue with the fixes, or with the tests.
All the failing tests seem to fail for the same reason.

kasiafi:
>I'm not sure how to proceed. I don't know if there is an issue with the fixes, or with the tests.
All the failing tests seem to fail for the same reason.

It seems that the failing tests changed their behavior accordingly to the change, and should be updated. It would be best to both:
- update the test result accordingly to the change
- add another modified version of the test case which returns the same result

kasiafi:
Also, please address this comment: https://github.com/trinodb/trino/pull/26702#discussion_r2728259438

core/trino-main/src/main/java/io/trino/sql/DynamicFilters.java
codluca: Fixed unit tests

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
codluca: I'm assuming you ask to add something like this:
```java
case LESS_THAN -> {
    Range range = Range.lessThan(span.getType(), span.getHighBoundedValue());
    yield Domain.create(ValueSet.ofRanges(range), this.nullAllowed && domain.isNullAllowed());
...
```

However, DynamicFilters.Descriptor has this:
checkArgument(!nullAllowed || operator == EQUAL, "nullAllowed should be true only with EQUAL operator");

codluca:
> Also, please address this comment: [#26702 (comment)](https://github.com/trinodb/trino/pull/26702#discussion_r2728259438)

I answered the comment. Please take a look.

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
kasiafi: I mean this code: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java#L296:L304
Both return statements ignore the `nullAllowed` info.

kasiafi:
> I answered the comment. Please take a look.

I mean a different piece of code. Please see the link I pasted.

core/trino-main/src/main/java/io/trino/operator/JoinDomainBuilder.java
codluca: I already fixed that. I forgot to reply.

codluca:
> > I answered the comment. Please take a look.
> 
> I mean a different piece of code. Please see the link I pasted.

I already fixed that. I forgot to reply.

testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
```diff
@@ -0,0 +1,135 @@
...
+        queryRunner.execute(format("CREATE TABLE %s (k1 INT, k2 INT)", tableName1));
+        queryRunner.execute(format("CREATE TABLE %s (k1 INT, k2 INT)", tableName2));
+
+        queryRunner.execute(format("INSERT INTO %s VALUES (1, NULL)", tableName1));
+        queryRunner.execute(format("INSERT INTO %s VALUES (1, NULL)", tableName2));
+        assertThat(assertions.query(format("SELECT *" +
+                " FROM %s t" +
+                " INNER JOIN %s AS s" +
+                "    ON s.k1 IS NOT DISTINCT FROM t.k1" +
+                "    AND s.k2 IS NOT DISTINCT FROM t.k2", tableName1, tableName2)))
```
martint: Use text blocks instead of string concatenation (here and in all the tests below)

testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
```diff
@@ -0,0 +1,135 @@
...
+                .matches("VALUES (1, CAST(NULL AS INTEGER), 1, CAST(NULL AS INTEGER))");
+
+        queryRunner.execute(format("INSERT INTO %s VALUES (NULL, NULL)", tableName1));
+        queryRunner.execute(format("INSERT INTO %s VALUES (NULL, NULL)", tableName2));
```
martint: Insert all the necessary values ahead of time and then run the test queries. Otherwise, the tests build on each other and it's harder to follow and maintain them.

testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
codluca: Fixed

testing/trino-tests/src/test/java/io/trino/tests/TestJoinIsNotDistinct.java
codluca: Fixed

kasiafi (approved):

PR #26703, opened 2025-09-24
Validate out of range DateTime value in clickhouse connector

krvikash:
## Description
Validate out of range DateTime value in clickhouse connector

## Release notes

### DIFF
--- a/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
+++ b/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
@@ -636,6 +636,7 @@ public class ClickHouseClient
             return mapping;
         }
 
+        ClickHouseVersion version = getClickHouseServerVersion(session);
         ClickHouseColumn column = ClickHouseColumn.of("", jdbcTypeName);
         ClickHouseDataType columnDataType = column.getDataType();
         switch (columnDataType) {
@@ -724,7 +725,7 @@ public class ClickHouseClient
                         DISABLE_PUSHDOWN));
 
             case Types.DATE:
-                return Optional.of(dateColumnMappingUsingLocalDate(getClickHouseServerVersion(session)));
+                return Optional.of(dateColumnMappingUsingLocalDate(version));
 
             case Types.TIMESTAMP:
                 if (columnDataType == ClickHouseDataType.DateTime) {
@@ -733,7 +734,7 @@ public class ClickHouseClient
                     return Optional.of(ColumnMapping.longMapping(
                             TIMESTAMP_SECONDS,
                             timestampReadFunction(TIMESTAMP_SECONDS),
-                            timestampSecondsWriteFunction(getClickHouseServerVersion(session))));
+                            timestampSecondsWriteFunction(version)));
                 }
                 // TODO (https://github.com/trinodb/trino/issues/10537) Add support for Datetime64 type
                 return Optional.of(timestampColumnMapping(TIMESTAMP_MILLIS));
@@ -745,7 +746,7 @@ public class ClickHouseClient
                     return Optional.of(ColumnMapping.longMapping(
                             TIMESTAMP_TZ_SECONDS,
                             shortTimestampWithTimeZoneReadFunction(),
-                            shortTimestampWithTimeZoneWriteFunction(column.getTimeZone())));
+                            shortTimestampWithTimeZoneWriteFunction(version, column.getTimeZone())));
                 }
         }
 
@@ -926,12 +927,15 @@ public class ClickHouseClient
         };
     }
 
-    private static LongWriteFunction shortTimestampWithTimeZoneWriteFunction(TimeZone columnTimeZone)
+    private static LongWriteFunction shortTimestampWithTimeZoneWriteFunction(ClickHouseVersion version, TimeZone columnTimeZone)
     {
         return (statement, index, value) -> {
             long millisUtc = unpackMillisUtc(value);
             // Clickhouse JDBC driver inserts datetime as string value as yyyy-MM-dd HH:mm:ss and zone from the Column metadata would be used.
-            statement.setObject(index, Instant.ofEpochMilli(millisUtc).atZone(columnTimeZone.toZoneId()));
+            Instant instant = Instant.ofEpochMilli(millisUtc);
+            // ClickHouse stores incorrect results when the values are out of supported range.
+            DATETIME.validate(version, instant.atZone(UTC).toLocalDateTime());
+            statement.setObject(index, instant.atZone(columnTimeZone.toZoneId()));
         };
     }
 
--- a/plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
+++ b/plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
@@ -59,37 +59,35 @@ import static io.trino.spi.type.VarcharType.createUnboundedVarcharType;
 import static io.trino.testing.TestingSession.testSessionBuilder;
 import static io.trino.type.IpAddressType.IPADDRESS;
 import static java.lang.String.format;
-import static java.time.ZoneOffset.UTC;
 import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 
 @TestInstance(PER_CLASS)
 public abstract class BaseClickHouseTypeMapping
         extends AbstractTestQueryFramework
 {
-    private final ZoneId jvmZone = ZoneId.systemDefault();
-
+    private static final ZoneId UTC = ZoneId.of("UTC");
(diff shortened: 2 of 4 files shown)


plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -1018,6 +1018,30 @@ public void testUnsupportedTimestamp(String unsupportedTimestamp)
         }
     }
 
+    @Test
+    void testUnsupportedDateTimeWithTimeZone()
+    {
+        for (ZoneId zoneId : timezones()) {
+            String inputType = DATETIME_TYPE_FACTORY.apply(zoneId);
+            testUnsupportedDateTimeWithTimeZone(inputType, "1969-12-31 23:59:59 UTC", "1969-12-31 23:59:59"); // min - 1 second
+            testUnsupportedDateTimeWithTimeZone(inputType, "2106-02-07 06:28:16 UTC", "2106-02-07 06:28:16"); // max + 1 second
```
chenjian2664: Why `2106-02-07 06:28:16` could you put a reference or comment for it

chenjian2664: Is it possible to extract a static variable in somewhere and used in here
And I see there is difference between different version about the max supported value, do we take that into account in the test?

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -1018,6 +1018,30 @@ public void testUnsupportedTimestamp(String unsupportedTimestamp)
...
+            testUnsupportedDateTimeWithTimeZone(inputType, "1970-01-01 00:00:00 Asia/Kathmandu", "1969-12-31 18:30:00");
+            testUnsupportedDateTimeWithTimeZone(inputType, "1970-01-01 00:13:42 Asia/Kathmandu", "1969-12-31 18:43:42");
+        }
+    }
+
+    private void testUnsupportedDateTimeWithTimeZone(String inputType, String unsupportedTimestampWithTz, String unsupportedTimestampUtc)
+    {
+        String minSupportedTimestamp = "1970-01-01 00:00:00";
+        String maxSupportedTimestamp = "2106-02-07 06:28:15";
```
chenjian2664: Does the max value fixed?

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -1018,6 +1018,30 @@ public void testUnsupportedTimestamp(String unsupportedTimestamp)
...
+
+        try (TestTable table = new TestTable(onRemoteDatabase(), "tpch.test_unsupported_timestamp_with_tz", "(dt %s) ENGINE=Log".formatted(inputType))) {
+            assertQueryFails(
+                    "INSERT INTO %s VALUES (TIMESTAMP '%s')".formatted(table.getName(), unsupportedTimestampWithTz),
+                    "Timestamp must be between %s and %s in ClickHouse: %s".formatted(minSupportedTimestamp, maxSupportedTimestamp, unsupportedTimestampUtc));
```
chenjian2664: nit: we could inline the min and max supported value

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
krvikash: yes, https://clickhouse.com/docs/sql-reference/data-types/datetime

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
krvikash: > And I see there is difference between different version about the max supported value, do we take that into account in the test?

Currently, the clickhouse server that we create runs on a version above 21.4. That's why the test case verifies the latest min and max value.

https://github.com/trinodb/trino/blob/68504f73969960d949800aa17a79e59d657ceb4f/plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/TestingClickHouseServer.java#L39-L52

krvikash: Versions less than 21.4 are EOL https://docs.altinity.com/altinitystablebuilds/#altinity-stable-builds-life-cycle-table.

krvikash:
Thanks @chenjian2664, Addressed your comments.

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -1039,39 +1061,38 @@ public void testClickHouseDateTimeWithTimeZone()
 
     private SqlDataTypeTest dateTimeWithTimeZoneTest(Function<ZoneId, String> inputTypeFactory)
     {
-        ZoneId utc = ZoneId.of("UTC");
```
anusudarsan, line 1043: this refactoring could be moved to the first commit

anusudarsan (approved):

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
```diff
@@ -926,12 +927,15 @@ private static LongReadFunction shortTimestampWithTimeZoneReadFunction()
         };
```
chenjian2664, line 927: I think we should validate the reading as well

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
krvikash: However, I have added the test coverage for altinity server as well for type mapping.

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
krvikash: When an out-of-range value is inserted into a ClickHouse table, ClickHouse stores it incorrectly by mapping it to a value within the supported range.

Example:
`1912-01-01 00:00:00` → stored as `1970-01-01 00:00:00`
`3102-01-01 00:00:00` → stored as `2027-10-17 11:03:28`

In Trino, the value that gets read appears to be within the supported range, making it impossible to tell whether `1970-01-01 00:00:00` was originally `1912-01-01 00:00:00` or actually `1970-01-01 00:00:00`.

krvikash:
Thanks @anusudarsan @chenjian2664, AC.

krvikash:
(rebased with master)

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/TestAltinityClickHouseTypeMapping.java
```diff
@@ -0,0 +1,40 @@
...
+ */
+package io.trino.plugin.clickhouse;
+
+import io.trino.testing.QueryRunner;
+import org.junit.jupiter.api.Test;
+
+import static io.trino.plugin.clickhouse.TestingClickHouseServer.ALTINITY_DEFAULT_IMAGE;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class TestAltinityClickHouseTypeMapping
```
ssheikin: Test classes should be defined as package-private and final.
Test methods should be defined as package-private.

krvikash: Since `testUnsupportedPoint` is declared public in the base class, it cannot be made package-private in this subclass.

Need refactoring of Base class. I think that can be a follow up.

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
marcinsbd: >`3102-01-01 00:00:00` → stored as `2027-10-17 11:03:28`

Just curious: If the given date above MAX it's stored not as MAX (2106-02-07 06:28:15) but as 2027-10-17 11:03:28 ?

krvikash: Yes, I have checked with some different values as well and it is inserting different values for those cases.

raunaqmorarka (approved):

PR #26705, opened 2025-09-24
Add buttons to toggle between old and preview UI

koszti:
## Description

This PR introduces a button that allows users to switch between the current UI and the preview UI. There are feature parity between the two interfaces and it makes the preview UI more accessible to a wider audience.

Changes
* Added a button to switch between the current and preview UIs.
* The button is only visible when `web-ui.preview.enabled` is set to `true`.
* Updated the default value of `web-ui.preview.enabled` to `true`, so more users will see the toggle by default.
* The current UI remains the default interface.

Documentation will need to be updated at: https://trino.io/docs/current/admin/preview-web-interface.html

## Screenshots

image

image

## Release notes

```markdown
## Section
* Set web-ui.preview.enabled property to default to true. ({issue}`26705`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/ui/ClusterResource.java
+++ b/core/trino-main/src/main/java/io/trino/server/ui/ClusterResource.java
@@ -36,19 +36,21 @@ public class ClusterResource
     private final NodeVersion version;
     private final String environment;
     private final long startTime = System.nanoTime();
+    private final boolean previewUiEnabled;
 
     @Inject
-    public ClusterResource(NodeVersion nodeVersion, NodeInfo nodeInfo)
+    public ClusterResource(NodeVersion nodeVersion, NodeInfo nodeInfo, WebUiConfig webUiConfig)
     {
         this.version = requireNonNull(nodeVersion, "nodeVersion is null");
         this.environment = nodeInfo.getEnvironment();
+        this.previewUiEnabled = webUiConfig.isPreviewEnabled();
     }
 
     @GET
     @Produces(APPLICATION_JSON)
     public ClusterInfo getInfo()
     {
-        return new ClusterInfo(version, environment, nanosSince(startTime));
+        return new ClusterInfo(version, environment, nanosSince(startTime), previewUiEnabled);
     }
 
     @Immutable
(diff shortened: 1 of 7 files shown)


wendigo:
@koszti one question regarding preview UI. Is it possible for the query sub pages (query json, stages, live plan etc) to be routable ? Right now URI stays the same for every tab which if you refresh page, will navigate to the first tab

wendigo:
trino-main failures are related @koszti

core/trino-main/src/main/java/io/trino/server/ui/ClusterResource.java
```diff
@@ -36,19 +36,21 @@ public class ClusterResource
...
     private final long startTime = System.nanoTime();
+    private final WebUiConfig webUiConfig;
 
     @Inject
-    public ClusterResource(NodeVersion nodeVersion, NodeInfo nodeInfo)
+    public ClusterResource(NodeVersion nodeVersion, NodeInfo nodeInfo, WebUiConfig webUiConfig)
     {
         this.version = requireNonNull(nodeVersion, "nodeVersion is null");
         this.environment = nodeInfo.getEnvironment();
+        this.webUiConfig = requireNonNull(webUiConfig, "webUiConfig is null");
```
wendigo: instead of assigning config itself, just assign value of `isPreviewEnabled`
so:

```
this.previewUiEnabled = webUiConfig.isPreviewEnabled();
```

rnn check is not required as it will throw NPE immediately in the constructor due to a chained call.

core/trino-main/src/main/java/io/trino/server/ui/ClusterResource.java
```diff
@@ -57,12 +59,14 @@ public static class ClusterInfo
         private final NodeVersion nodeVersion;
         private final String environment;
         private final Duration uptime;
+        private final boolean previewEnabled;
 
-        public ClusterInfo(NodeVersion nodeVersion, String environment, Duration uptime)
+        public ClusterInfo(NodeVersion nodeVersion, String environment, Duration uptime, boolean previewEnabled)
```
wendigo: previewUiEnabled

koszti:
> @koszti one question regarding preview UI. Is it possible for the query sub pages (query json, stages, live plan etc) to be routable ? Right now URI stays the same for every tab which if you refresh page, will navigate to the first tab

@wendigo, implemented by a separated PR: https://github.com/trinodb/trino/pull/26709

PR #26718, opened 2025-09-26
Require implementing shutdown method on a Connector 

wendigo:
This ensures that at least every Connector will call LifeCycleManager.stop()
which is important because every Bootstrap creates one LifeCycleManager
which registers itself with the shutdown hook creating a shutdown Thread.

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-jmx/src/main/java/io/trino/plugin/jmx/JmxConnector.java
+++ b/plugin/trino-jmx/src/main/java/io/trino/plugin/jmx/JmxConnector.java
@@ -14,7 +14,7 @@
 package io.trino.plugin.jmx;
 
 import com.google.inject.Inject;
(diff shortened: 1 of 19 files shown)


plugin/trino-jmx/src/main/java/io/trino/plugin/jmx/JmxConnector.java
```diff
@@ -29,22 +30,22 @@ public class JmxConnector
 {
     private static final Logger log = Logger.get(JmxConnector.class);
```
raunaqmorarka: seems to be unused now

ebyhr:
@wendigo Could you update "Release notes" section in PR description?

PR #26721, opened 2025-09-26
Add meaningful names to bootstraps

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/connector/FileCatalogStoreFactory.java
+++ b/core/trino-main/src/main/java/io/trino/connector/FileCatalogStoreFactory.java
@@ -32,7 +32,7 @@ public class FileCatalogStoreFactory
     @Override
     public CatalogStore create(Map<String, String> config)
     {
-        Bootstrap app = new Bootstrap(new FileCatalogStoreModule());
+        Bootstrap app = new Bootstrap("io.trino.bootstrap.store." + getName(), new FileCatalogStoreModule());
 
         Injector injector = app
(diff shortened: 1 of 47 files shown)


core/trino-main/src/main/java/io/trino/connector/FileCatalogStoreFactory.java
```diff
@@ -32,7 +32,7 @@ public String getName()
     @Override
     public CatalogStore create(Map<String, String> config)
     {
-        Bootstrap app = new Bootstrap(new FileCatalogStoreModule());
+        Bootstrap app = new Bootstrap("trino.store." + getName(), new FileCatalogStoreModule());
```
losipiuk: As https://github.com/airlift/airlift/pull/1573/files does not imply common prefix for string used for setup logging please ensure it here. Make sure it is obvious it is about bootstrap logging.
`Bootstrap.` seems like a good prefix for me

wendigo: io.trino.bootstrap. ?

wendigo: @losipiuk wdyt?

losipiuk: ok - but you did only change a few

losipiuk (approved):

PR #26725, opened 2025-09-26
Improve performance of PagesIndexComparator for strings

raunaqmorarka:
## Description

Made changes to prefer adapting ordering operator from the
more efficient version of `DefaultOrderingOperators#comparisonOperator`
that avoids unnecessary copies of Slice for comparison

```
BenchmarkPagesIndexOrdering.benchmarkQuickSort
(numberOfChannels)  (typeName)  Mode  Cnt  Before score    After score    Unit
                 1      BIGINT  avgt   20  29.536 ± 0.696  28.299 ± 0.344 ms/op
                 1     VARCHAR  avgt   20  66.897 ± 1.914  53.538 ± 0.635 ms/op
                10      BIGINT  avgt   20  36.965 ± 0.270  36.577 ± 0.605 ms/op
                10     VARCHAR  avgt   20  79.887 ± 3.077  60.413 ± 0.678 ms/op
```

E.g. from current code
```
io.airlift.slice.Slice.slice(Slice.java:1042)
io.trino.spi.block.VariableWidthBlock.getSlice(VariableWidthBlock.java:160)
io.trino.spi.type.VarcharType.getSlice(VarcharType.java:188)
io.trino.$gen.PagesIndexComparator_20250923_193215_67.compareTo(Unknown Source)
io.trino.operator.PagesIndexOrdering.quickSort(PagesIndexOrdering.java:88)
io.trino.operator.PagesIndexOrdering.quickSort(PagesIndexOrdering.java:134)
io.trino.operator.PagesIndexOrdering.sort(PagesIndexOrdering.java:37)
io.trino.operator.PagesIndex.sort(PagesIndex.java:424)
io.trino.operator.WindowOperator.sortPagesIndexIfNecessary(WindowOperator.java:905)
```
Now this will use `io.trino.spi.type.AbstractVariableWidthType.DefaultOrderingOperators#comparisonOperator(io.trino.spi.block.VariableWidthBlock, int, io.trino.spi.block.VariableWidthBlock, int)` instead which avoids Slice copies

## Release notes

```markdown
## General
* Improve performance of queries with ORDER BY. ({issue}`26725`)
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
@@ -43,6 +43,7 @@ import static io.trino.spi.function.InvocationConvention.InvocationArgumentConve
 import static io.trino.spi.function.InvocationConvention.InvocationArgumentConvention.FLAT;
 import static io.trino.spi.function.InvocationConvention.InvocationArgumentConvention.NEVER_NULL;
 import static io.trino.spi.function.InvocationConvention.InvocationArgumentConvention.NULL_FLAG;
+import static io.trino.spi.function.InvocationConvention.InvocationArgumentConvention.VALUE_BLOCK_POSITION_NOT_NULL;
 import static io.trino.spi.function.InvocationConvention.InvocationReturnConvention.BLOCK_BUILDER;
 import static io.trino.spi.function.InvocationConvention.InvocationReturnConvention.FAIL_ON_NULL;
 import static io.trino.spi.function.InvocationConvention.simpleConvention;
@@ -496,7 +497,12 @@ public class TypeOperators
             SortOrder sortOrder = operatorConvention.sortOrder().orElseThrow(() -> new IllegalArgumentException("Operator convention does not contain a sort order"));
             OperatorType comparisonType = operatorConvention.operatorType();
             if (operatorConvention.callingConvention().getArgumentConventions().equals(List.of(BLOCK_POSITION, BLOCK_POSITION))) {
-                OperatorConvention comparisonOperator = new OperatorConvention(operatorConvention.type(), comparisonType, Optional.empty(), simpleConvention(FAIL_ON_NULL, BLOCK_POSITION, BLOCK_POSITION));
+                OperatorConvention comparisonOperator = new OperatorConvention(
+                        operatorConvention.type(),
+                        comparisonType,
+                        Optional.empty(),
+                        // null positions are handled separately in adaptBlockPositionComparisonToOrdering and not used in the comparison
+                        simpleConvention(FAIL_ON_NULL, BLOCK_POSITION_NOT_NULL, BLOCK_POSITION_NOT_NULL));
                 MethodHandle comparisonInvoker = adaptOperator(comparisonOperator);
                 return adaptBlockPositionComparisonToOrdering(sortOrder, comparisonInvoker);
             }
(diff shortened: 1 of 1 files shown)


core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
```diff
@@ -541,6 +542,9 @@ private static int getScore(OperatorMethodHandle operatorMethodHandle)
             else if (argument == BLOCK_POSITION) {
```
wendigo: can we `|| argument == VALUE_BLOCK_POSITION_NOT_NULL` ?

core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
```diff
@@ -538,7 +544,7 @@ private static int getScore(OperatorMethodHandle operatorMethodHandle)
             if (argument == NULL_FLAG || argument == FLAT) {
                 score += 100;
             }
-            else if (argument == BLOCK_POSITION) {
+            else if (argument == BLOCK_POSITION || argument == VALUE_BLOCK_POSITION_NOT_NULL) {
```
martint, line 547: Why is this needed? Have you checked if it has any impact on other usages not related to ordering?

raunaqmorarka: This is needed to ensure that we prioritize adapting from the `VALUE_BLOCK_POSITION_NOT_NULL` version of a given operator. Without this, we will end up adapting from one of the implementations containing `NEVER_NULL` arguments which leads to inefficient adaptation.
This code is common to every type and every operator. It seems logical to me that if the VALUE_BLOCK_POSITION_NOT_NULL implementation exists, that should at least get priority over the NEVER_NULL implementation. I don't see a case where both BLOCK_POSITION and VALUE_BLOCK_POSITION_NOT_NULL exist.
However, I don't have a full understanding of this. Maybe @dain can chime in here

findepi (approved):

martint (approved):

core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
martint: Sounds good. Potential conflicts between BLOCK_POSITION and VALUE_BLOCK_POSITION_NOT_NULL should be detected elsewhere by the code that loads all the function definitions and implementations.

core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
```diff
@@ -538,7 +544,7 @@ private static int getScore(OperatorMethodHandle operatorMethodHandle)
             if (argument == NULL_FLAG || argument == FLAT) {
```
dain, line 544: BTW, I think it is a bug we have `FLAT` twice.  I think all three of these clauses are supposed to be a single if/else block.  Maybe we should convert to a switch expression `score += switch(argument) {...`

core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
dain: I expect we are missing some other argument types.

dain (approved):
This looks good to me.

core/trino-spi/src/main/java/io/trino/spi/type/TypeOperators.java
raunaqmorarka: That does look suspicious, but I'm unsure about what the correct implementation looks like. Could you tackle that part ?

PR #26350, opened 2025-08-06
Add Trino 477 release notes

ebyhr:
## Description

Assemble the release notes for the upcoming Trino 477 release.

## Additional context and related issues

See https://github.com/trinodb/trino/pulls?q=is%3Apr+is%3Aclosed+milestone%3A477

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

## Verification for each pull request

Format: PR/issue number, ✅ / ❌ rn ✅ / ❌ docs
✅ rn - release note added and verified, or assessed to be not necessary, set to ❌ rn before completion
✅ docs - need for docs assessed and merged, or assessed to be not necessary, set to ❌ docs before completion

## 06 Jun 2025
* #25892 ✅ rn ✅ docs
* #25948 ✅ rn ✅ docs
* #25951 ✅ rn ✅ docs
* #25927 ✅ rn ✅ docs
* #25946 ✅ rn ✅ docs
* #25944 ✅ rn ✅ docs

## 08 Jun 2025
* #25945 ✅ rn ✅ docs
* #25949 ✅ rn ✅ docs

## 09 Jun 2025
* #25959 ✅ rn ✅ docs
* #25954 ✅ rn ✅ docs
* #25962 ✅ rn ✅ docs
* #25964 ✅ rn ✅ docs
* #25965 ✅ rn ✅ docs
* #25956 ✅ rn ✅ docs

## 10 Jun 2025
* #25971 ✅ rn ✅ docs

## 11 Jun 2025
* #25977 ✅ rn ✅ docs
* #25973 ✅ rn ✅ docs
* #25889 ✅ rn ✅ docs
* #25983 ✅ rn ✅ docs
* #25988 ✅ rn ✅ docs

## 12 Jun 2025
* #25981 ✅ rn ✅ docs
* #25967 ✅ rn ✅ docs
* #25985 ✅ rn ✅ docs
* #25987 ✅ rn ✅ docs
* #25980 ✅ rn ✅ docs

## 13 Jun 2025
* #26001 ✅ rn ✅ docs
* #26003 ✅ rn ✅ docs
* #25996 ✅ rn ✅ docs
* #25989 ✅ rn ✅ docs
* #26004 ✅ rn ✅ docs
* #26005 ✅ rn ✅ docs
* #26006 ✅ rn ✅ docs
* #25811 ✅ rn ✅ docs

## 14 Jun 2025
* #26002 ✅ rn ✅ docs
* #26012 ✅ rn ✅ docs

[...]

PR #26726, opened 2025-09-26
Add Trino 478 release note

ebyhr:
## Release notes


wendigo:
@ebyhr I've added two more entries

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,100 @@
+# Release 478 (dd Oct 2025)
+
+## General
+
+* {{breaking}} Move startup logs to `io.trino.bootstrap` package. ({issue}`26721`)
+* Run Trino on JDK 25.0.0 (build 36). ({issue}`26693`)
```
martint: This should go in the Docker image section

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,100 @@
+# Release 478 (dd Oct 2025)
+
+## General
+
+* {{breaking}} Move startup logs to `io.trino.bootstrap` package. ({issue}`26721`)
```
martint: This is an internal implementation detail, so remove it

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,100 @@
...
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+
+## Hudi connector
+
+## Iceberg connector
+
+* Fix failure due to column count mismatch when executing `add_files_from_table`
+  procedure. ({issue}`26774`)
+* Support using a local staging path for improved performance of writes to sorted
```
martint: Improve performance when writing sorted tables and the xxx config option is set.

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Blackhole connector
+
+## Cassandra connector
+
+## ClickHouse connector
+
+## Delta Lake connector
+
+* Fix failure when reading `NULL` map on `json` type. ({issue}`26700`)
```
martint: Json doesn’t have a concept of maps, so what what’s a null map?

ebyhr: cc: @wendigo @chenjian2664

chenjian2664: it fixed the issue with reading map columns where the value type is json and the value is null
How about "Fix reading map type when value type is json and value is `null`"

ebyhr: @chenjian2664 Please rephrase to clarify whether it was a failure or a correctness issue. It should start with either "Fix failure when ..." or "Fix incorrect results when …"

chenjian2664: Thanks.
It is a failure, so should be "Fix failure when reading map type with value type is json and value is `null`"

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
+# Release 478 (dd Oct 2025)
+
+## General
+
+* Add support for column lineage in `UNNEST` clause. ({issue}`16946`)
+* Add `allowed-retry-policies` configuration property to specify which retry
+  policies can be selected by user. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Allow to drop catalog that failed to load correctly. ({issue}`26918`)
```
wendigo: to drop `uninitialized catalog that failed to load` maybe ?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+* Improve performance of queries with `ORDER BY`. ({issue}`26725`)
+* Improve performance of `MERGE` statement involving `NOT MATCHED` case. ({issue}`26759`)
+* Fix potential incorrect results when reading `row` type. ({issue}`26806`)
```
wendigo, line 14: should be ROW uppercased?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+* Make `metadata.catalogs` table return all catalogs. ({issue}`26918`)
```
wendigo: what's the difference from previous state? maybe add something like `including uninitialized ones` ?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Security
+
+* Propagate `queryId` to [Open Policy Agent](/security/opa-access-control)
+  authorizer. ({issue}`26851`)
+
+## Web UI
+
+* Add support for filtering queries by `X-Trino-Trace-Token` value in [](/admin/preview-web-interface). ({issue}`26447`)
+* Improve rendering performance of large query JSON code in [](/admin/preview-web-interface). ({issue}`26807`)
```
wendigo: drop `code` just `query JSON in admin/preview-web-interface`

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+* Fix query plan default viewport in [](/admin/preview-web-interface). ({issue}`26749`)
```
wendigo: `Fix rendering of a large query plans in ...`

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+* Fix query splits timeline in [](/admin/preview-web-interface) to prevent
```
wendigo: `Fix rendering of splits timeline for queued queries`

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+  blank screen ({issue}`26920`)
+
+## JDBC driver
+
+* Add support for zstd, brotli, and gzip compressions. ({issue}`26875`)
```
wendigo: It was backed off. Drop

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Docker image
+
+* Run Trino on JDK 25.0.0 (build 36). ({issue}`26693`)
```
wendigo, line 34: FYI: This will change to 25.0.1 soon

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+
+## CLI
+
+* Add support for zstd, brotli, and gzip compressions. ({issue}`26875`)
```
wendigo: Drop

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Blackhole connector
+
+## Cassandra connector
+
+## ClickHouse connector
+
+## Delta Lake connector
+
+* Fix failure when reading `NULL` map on `json` type. ({issue}`26700`)
```
wendigo: `JSON` type ?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Google Sheets connector
+
+* Fix potential query failure when `gsheets.delegated-user-email` config property
+  is used. ({issue}`26501`)
+
+## Hive connector
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+* Fix ORC writer to ensure that dates and timestamps older than 1582 are read
```
wendigo: gregorian dates ?

docs/src/main/sphinx/release/release-478.md
ebyhr: No, we use lowercase for type names.

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,139 @@
...
+
+## Hudi connector
+
+* Deprecate `gcs.use-access-token` in favor of `gcs.auth-type` config property. ({issue}`26681`)
+
+## Iceberg connector
+
+* Improve performance when writing sorted tables and the `iceberg.sorted-writing.local-staging-path`
+  config option is set. ({issue}`24376`)
+* Produce metrics about execution in the output of `remove_orphan_files` procedure. ({issue}`26661`)
```
wendigo: Return execution metrics while running `remove_orphan_files` procedure

ebyhr:
@wendigo Please feel free to edit this PR directly.

wendigo:
@ebyhr i didn't want to do that to avoid conflicts

docs/src/main/sphinx/release/release-478.md
ebyhr: cc: @homar

homar: `including uninitialized ones` sounds good, should I make a change ?

docs/src/main/sphinx/release/release-478.md
ebyhr: The past release note uses actual dates instead of "gregorian" keyword. I replaced 1582 with 1582-10-15. 

cc: @raunaqmorarka @marcinsbd

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
+# Release 478 (dd Oct 2025)
```
mosabua: Reminder to update .. probably 22

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+
+## General
+
+* Add support for column lineage in `UNNEST` clause. ({issue}`16946`)
```
mosabua:
```suggestion
* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Add `allowed-retry-policies` configuration property to specify which retry
```
mosabua: Can we link to docs for this ?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  policies can be selected by user. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
```
mosabua:
```suggestion
* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
```

docs/src/main/sphinx/release/release-478.md
mosabua: Yeah .. also 

to drop an uninitialized catalog

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Allow to drop catalog that failed to load correctly. ({issue}`26918`)
+* Improve performance of queries with `ORDER BY`. ({issue}`26725`)
```
mosabua: either "with an ORDER BY clause" or "using an ORDER BY clause"

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Improve performance of `MERGE` statement involving `NOT MATCHED` case. ({issue}`26759`)
```
mosabua:
```suggestion
* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Improve performance of spilling join queries. ({issue}`26076`)
+* Fix potential incorrect results when reading `row` type. ({issue}`26806`)
+* Make `metadata.catalogs` table return all catalogs. ({issue}`26918`)
```
mosabua: Return all catalogs, including uninitialized ones, for queries from `metadata.catalogs`.

Or 

Track all catalogs, including uninitialized ones, in the metadata.catalogs system table

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Ensure that queries with and without `EXPLAIN ANALYZE` are planned in the same
```
mosabua:
```suggestion
* Ensure that queries with and without `EXPLAIN ANALYZE` are planned identically.
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  way. ({issue}`26938`)
+
+## Security
+
+* Propagate `queryId` to [Open Policy Agent](/security/opa-access-control)
```
mosabua:
```suggestion
* Propagate `queryId` to the [Open Policy Agent](/security/opa-access-control)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  authorizer. ({issue}`26851`)
+
+## Web UI
+
+* Add support for filtering queries by `X-Trino-Trace-Token` value in [](/admin/preview-web-interface). ({issue}`26447`)
```
mosabua:
```suggestion
* Add support for filtering queries by `X-Trino-Trace-Token` value in the [](/admin/preview-web-interface). ({issue}`26447`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Make `metadata.catalogs` table return all catalogs. ({issue}`26918`)
+* Ensure that queries with and without `EXPLAIN ANALYZE` are planned in the same
+  way. ({issue}`26938`)
+
+## Security
+
+* Propagate `queryId` to [Open Policy Agent](/security/opa-access-control)
+  authorizer. ({issue}`26851`)
+
+## Web UI
```
mosabua, line 24: Also ... not sure by @martint did not want changes for preview UI in the release notes in the past .. however I think we should have them

docs/src/main/sphinx/release/release-478.md
mosabua: Same for others

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+
+* Add support for filtering queries by `X-Trino-Trace-Token` value in [](/admin/preview-web-interface). ({issue}`26447`)
+* Improve rendering performance of large query JSON in [](/admin/preview-web-interface). ({issue}`26807`)
+* Fix rendering of a large query plans in [](/admin/preview-web-interface). ({issue}`26749`)
```
mosabua:
```suggestion
* Fix rendering of large query plans in [](/admin/preview-web-interface). ({issue}`26749`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Fix rendering of splits timeline for queued queries in [](/admin/preview-web-interface) to prevent
```
mosabua:
```suggestion
* Fix rendering of the splits timeline for queued queries in [](/admin/preview-web-interface) to prevent
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  blank screen ({issue}`26920`)
+
+## JDBC driver
```
mosabua: Remove this and all other empty ones

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+## Blackhole connector
+
+## Cassandra connector
+
+## ClickHouse connector
+
+## Delta Lake connector
+
+* Fix failure when reading `NULL` map on `json` type. ({issue}`26700`)
+* Deprecate `gcs.use-access-token` in favor of `gcs.auth-type` config property. ({issue}`26681`)
```
mosabua:
```suggestion
* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+
+## Elasticsearch connector
+
+## Exasol connector
+
+## Faker connector
+
+## Google Sheets connector
+
+* Fix potential query failure when `gsheets.delegated-user-email` config property
```
mosabua:
```suggestion
* Fix potential query failure when the `gsheets.delegated-user-email` configuration property
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  is used. ({issue}`26501`)
+
+## Hive connector
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+* Deprecate `gcs.use-access-token` in favor of `gcs.auth-type` config property. ({issue}`26681`)
```
mosabua: Same as other .. personally I think we should avoid "in favor"

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Improve performance of queries using complex predicates on `$path` column. ({issue}`27000`)
+* Fix ORC writer to ensure that dates and timestamps older than `1582-10-15` are
```
mosabua: Prevent writing invalid dates and timestamps before `1582-10-15` by the ORC writer.

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  read correctly by Apache Hive. ({issue}`26507`)
+
+## Hudi connector
+
+* Deprecate `gcs.use-access-token` in favor of `gcs.auth-type` config property. ({issue}`26681`)
+
+## Iceberg connector
+
+* Improve performance when writing sorted tables and the `iceberg.sorted-writing.local-staging-path`
```
mosabua:
```suggestion
* Improve performance when writing sorted tables and `iceberg.sorted-writing.local-staging-path`
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+  config option is set. ({issue}`24376`)
```
mosabua:
```suggestion
 is set. ({issue}`24376`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Return execution metrics while running `remove_orphan_files` procedure. ({issue}`26661`)
```
mosabua:
```suggestion
* Return execution metrics while running the `remove_orphan_files` procedure. ({issue}`26661`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,140 @@
...
+* Deprecate `gcs.use-access-token` in favor of `gcs.auth-type` config property. ({issue}`26681`)
+* Fix failure due to column count mismatch when executing `add_files_from_table`
```
mosabua:
```suggestion
* Fix failure due to column count mismatch when executing the `add_files_from_table`
```

mosabua (changes requested):
A couple of changes and final clean up needed

docs/src/main/sphinx/release/release-478.md
marcinsbd: The current note is more about the outcome 
```
Fix ORC writer to ensure that dates and timestamps older than `1582-10-15` are read correctly by Apache Hive. ({issue}`26507`)
```
but what we really implemented is more like the foloowing:
```
ORC writer adds the Calendar type (proleptic Gregorian) used for writing dates and timestamps to the ORC footer
```

docs/src/main/sphinx/release/release-478.md
ebyhr: (Ignoring as an optional comment)

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,79 @@
+# Release 478 (22 Oct 2025)
+
+## General
+
+* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
+* Add `retry-policy.allowed` configuration property to specify which retry
+  policies can be selected by user. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
+* Allow to drop an uninitialized catalog that failed to load. ({issue}`26918`)
```
colebow:
```suggestion
* Allow dropping an uninitialized catalog that failed to load. ({issue}`26918`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,79 @@
+# Release 478 (22 Oct 2025)
+
+## General
+
+* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
+* Add `retry-policy.allowed` configuration property to specify which retry
+  policies can be selected by user. ({issue}`26628`)
```
colebow:
```suggestion
  policies can be selected by the user. ({issue}`26628`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,79 @@
...
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
+* Allow to drop an uninitialized catalog that failed to load. ({issue}`26918`)
+* Improve performance of queries with an `ORDER BY` clause. ({issue}`26725`)
+* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
+* Improve performance of spilling join queries. ({issue}`26076`)
```
colebow: Should we say something more explicit like "queries with joins which spill to disk"?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,79 @@
...
+
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+
+## Iceberg connector
+
+* Improve performance when writing sorted tables and `iceberg.sorted-writing.local-staging-path`
+ is set. ({issue}`24376`)
+* Return execution metrics while running the `remove_orphan_files` procedure. ({issue}`26661`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+* Collect NDV stats on all columns when replacing tables. ({issue}`26983`)
```
colebow: I'm not sure even a power Trino user would 100% understand the NDV acronym. [A link to Iceberg docs](https://iceberg.apache.org/puffin-spec/#apache-datasketches-theta-v1-blob-type) might be good.

ebyhr: NDV isn't specific to Iceberg. I rephrased as "distinct values count" to avoid an abbreviation.

findepi:
FYI, i changed the target branch of this PR
- because of https://github.com/trinodb/trino/issues/27068
- it now targets source of https://github.com/trinodb/trino/pull/27070 PR (the fix for that issue)
- it will auto-udpate back to `master` when that PR is merged

colebow:
Very funny (and unhelpful) that Copilot is randomly suggesting different PR numbers.

findepi:
@colebow  thanks for noticing!

BTW Whenever you see a bot's comment, like these copilot's, that you verified to be unhelpful, please resolve them (ideally along with thumb down reaction, but at least resolve)

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
+# Release 478 (22 Oct 2025)
+
+## General
+
+* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
+* Add `retry-policy.allowed` configuration property to specify which retry
+  policies can be selected by the user. ({issue}`26628`)
```
findepi, line 7:
```suggestion
* Add `retry-policy.allowed` configuration property to specify which query retry
  policies can be selected by the user. ({issue}`26628`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
```
findepi, line 9:
```suggestion
* Add the `/v1/integrations/gateway` endpoint that exposes additional cluster metrics for Trino Gateway. ({issue}`26548`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Allow dropping an uninitialized catalog that failed to load. ({issue}`26918`)
```
findepi, line 10: do we want to quantify this with "When using dynamic catalogs" or similar?

```
* When using dynamic catalogs, allow dropping ...
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Improve performance of queries with an `ORDER BY` clause. ({issue}`26725`)
```
findepi, line 11:
```suggestion
* Improve performance of queries with an `ORDER BY` clause using `varchar` or `varbinary` types. ({issue}`26725`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
+* Improve performance of queries with joins which spill to disk. ({issue}`26076`)
+* Fix potential incorrect results when reading `row` type. ({issue}`26806`)
+* Return all catalogs, including uninitialized ones, for queries from `metadata.catalogs`. ({issue}`26918`)
+* Ensure that queries with and without `EXPLAIN ANALYZE` are planned identically. ({issue}`26938`)
```
findepi, line 16: ❤️ 

this is a bug fix, so it's best to start with "Fix"

```suggestion
* Fix `EXPLAIN ANALYZE` planning so that it executes with the same plan as would be used to execute the query being analyzed. ({issue}`26938`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
+* Add `retry-policy.allowed` configuration property to specify which retry
+  policies can be selected by the user. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
+* Allow dropping an uninitialized catalog that failed to load. ({issue}`26918`)
+* Improve performance of queries with an `ORDER BY` clause. ({issue}`26725`)
+* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
+* Improve performance of queries with joins which spill to disk. ({issue}`26076`)
+* Fix potential incorrect results when reading `row` type. ({issue}`26806`)
```
findepi, line 14: Do we put Fix-es first?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Return all catalogs, including uninitialized ones, for queries from `metadata.catalogs`. ({issue}`26918`)
+* Ensure that queries with and without `EXPLAIN ANALYZE` are planned identically. ({issue}`26938`)
+* In row pattern matching, restrict logical navigations to current match in running semantics. ({issue}`26981`)
```
findepi, line 17: this is also a bug fix, maybe

```suggestion
* Fix row pattern matching logical navigations in running semantics to be always constraint to current match. Previously, a logical navigation function such as `FIRST` could return position outside of the current match. ({issue}`26981`)
```

docs/src/main/sphinx/release/release-478.md
findepi: The docs say "The Preview Web UI is not suitable for production usage, and only available for testing and evaluation purposes."
Thus, any changes in that component are irrelevant from end user / trino operator standpoint. Release notes are for end users' and operators' consumption.


Let's remove all entries pertaining the preview web ui

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Fix rendering of the splits timeline for queued queries in the [](/admin/preview-web-interface) to prevent
+  blank screen ({issue}`26920`)
+
+## Docker image
+
+* Run Trino on JDK 25.0.0 (build 36). ({issue}`26693`)
+
+## Delta Lake connector
+
+* Fix failure when reading `map` type with value type is `json` and value is `NULL`. ({issue}`26700`)
```
findepi, line 38:
```suggestion
* Fix failure when reading `map` type with `json` value type when a value is `NULL`. ({issue}`26700`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+
+* Fix potential query failure when the `gsheets.delegated-user-email` configuration property
+  is used. ({issue}`26501`)
+
+## Hive connector
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+* Improve performance of queries using complex predicates on `$path` column. ({issue}`27000`)
+* Prevent writing invalid dates and timestamps before `1582-10-15` by the ORC writer. ({issue}`26507`)
```
findepi, line 51: this is a bug fix

```suggestion
* Fix writing invalid dates and timestamps before `1582-10-15` when writing ORC files by setting calendar type in the file footer. Previously, Trino did not set the calendar type in the file footer, and values before `1582-10-15` could be read incorrectly by other query engines such as Apache Hive.  ({issue}`26507`)
```

findepi: @marcinsbd @raunaqmorarka does this apply to Iceberg connector too?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Hudi connector
+
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+
+## Iceberg connector
+
+* Improve performance when writing sorted tables and `iceberg.sorted-writing.local-staging-path`
+ is set. ({issue}`24376`)
+* Return execution metrics while running the `remove_orphan_files` procedure. ({issue}`26661`)
```
findepi, line 61: it's not "procedure". The code calls this "table procedure"
But the docs calls it "command" https://trino.io/docs/current/connector/iceberg.html#remove-orphan-files

```suggestion
* Return execution metrics while running the `remove_orphan_files` command. ({issue}`26661`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Fix failure when executing `SHOW CREATE SCHEMA` on a schema with unsupported properties
+  with REST, Glue or Nessie catalog. ({issue}`24744`)
+
+## Kafka connector
+
+* Fix failure when filtering partitions by timestamp offset. ({issue}`26787`)
+
+## SPI
+
+* Require `shutdown` to be implemented by the `Connector`. ({issue}`26718`)
```
findepi, line 77: `Connector` is the name of the interface connectors implement.
The interface does not implement `shutdown`. Quite contrary, it lost its default implementation.


```suggestion
* Remove default implementation from `Connector.shutdown()`. ({issue}`26718`)
```

docs/src/main/sphinx/release/release-478.md
findepi, line 9: BTW I am not convinced we should have added that.
Trino and Trino Gateway are both projects we manage, so we _can_ have coupling between them. That does not mean we _should_ or _want_ such coupling. Everything is a trade off, of course. If this endpoint exposes information previously unavailable, and also interesting to Trino Gateway only, we perhaps want to have it. But it doesn't look like this is the case here. The endpoint exposes cluster utilization metrics (memory, load), which are generally useful for observability and also for any other query routing logic. We should expose them in gateway-agnostic manner, rather than build a public-but-do-not-use-kind-of-internal-yet-public API.

cc @andythsu @wendigo @martint

wendigo: We've discussed that already and settled on the decision that it's ok to have a client specific endpoint (like we have for the UI). We can revisit this decision but if it blocks the release we should just revert the change and move on

docs/src/main/sphinx/release/release-478.md
ebyhr: Martin suggested the following order https://github.com/trinodb/trino/pull/26350#discussion_r2294232330

> Entries should be sorted according to the following order:
> * new features / improvements
> * performance improvements
> * bug fixes

docs/src/main/sphinx/release/release-478.md
losipiuk, line 17: The nomenclature here is technical. Does it match what we us in docs?

docs/src/main/sphinx/release/release-478.md
losipiuk: Why are we treating this as user visible? Because we support attaching external plugins to Trino run from official docker image?

findepi:
> * it will auto-udpate back to `master` when that PR is merged

it's how i saw it working many times, but this time it did not work. 
sorry

docs/src/main/sphinx/release/release-478.md
findepi: > ok to have a client specific endpoint (like we have for the UI)

UI is not the best example, because it's maintained within this repo. The UI endpoints can change as the UI needs evolve.

This new endpoint is different. It's meant for the Trino Gateway, so in theory it should not be used by any other software. Any other query routing software will needs its own endpoint, which is quite undesirable. All this while the information exposed is no way trino gateway specific and could be exposed in a more reusable manner.

> We can revisit this decision but if it blocks the release we should just revert the change and move on

Agreed!

findepi:
unfortunately it is not possible to reopen this PR, even after i recreated the base branch
i copied current state of this PR to a new one

- https://github.com/trinodb/trino/pull/27089

i will go over recent comments and apply them as separate fixup there.

docs/src/main/sphinx/release/release-478.md
findepi: -> https://github.com/trinodb/trino/pull/27089/files#r2459915160

docs/src/main/sphinx/release/release-478.md
findepi: -> https://github.com/trinodb/trino/pull/27089/files#r2459921217

docs/src/main/sphinx/release/release-478.md
findepi: thanks @ebyhr for the pointer!
i see this is also documented at https://trino.io/development/process#release-note-guidelines-

docs/src/main/sphinx/release/release-478.md
andythsu: I agree with you that it's undesirable to have an endpoint for each service and it should be in a reusable manner

IMO, the coupling between Trino Gateway and Trino is inevitable. If Trino changes, Trino Gateway **will** need to change. For example, in [this PR](https://github.com/trinodb/trino-gateway/pull/774) because Trino 477 removed some of the jmx stats, which caused Trino Gateway to fail, we had to pin Trino version to 476 and look for workarounds.

For this endpoint specifically, yes the data it returns can be useful for other software as well and should be gateway-agnostic. But I think the point is that **each software should deserve its own dedicated endpoint so that they are not affected by unintentional changes**. Other software owners are welcome, of course, to use this endpoint, but it is their responsibility to make sure their software doesn't break at **all times**, even after we roll out new changes to this endpoint. It's another way of saying there's no SLA on this endpoint if they use it, whereas if we create a generic endpoint, there's implied SLA

At the end of the day, if other query routing logic software needs similar data to be returned, they will likely end up with similar code in their endpoint. Yes this results in duplicate code. But we can look into how to extract common logic to a helper function. At that point, maybe we will find some discrepancies in these seemingly-similar looking endpoints.

findepi: > But I think the point is that **each software should deserve its own dedicated endpoint so that they are not affected by unintentional changes**.

I appreciate the openness but I doubt that would be want we would actually want to maintain in the project.

BTW trino already exports tons of metrics in standard format (that's why jmx became obsolete), so i don't quite yet understand why we need dedicated endpoints just for selected metrics useful for selected use case

docs/src/main/sphinx/release/release-478.md
raunaqmorarka: I don't agree with that reasoning at all. We're shipping the preview web UI because we want users to use that and retire the old UI. So it doesn't make sense to hide away the developments happening on that front. Also, the UI being new or not having all the features of the old one doesn't make it "not suitable for production usage". It's easy to toggle back and forth between new and old UI with a button after https://github.com/trinodb/trino/pull/26705 and `web-ui.preview.enabled` is enabled by default.

findepi: -> https://github.com/trinodb/trino/pull/27089/commits/29ca14fa03a211e76872c2ef26055cb501d83b23#r2464779544

PR #26729, opened 2025-09-27
Let building page from resultSet async

chenjian2664:
## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
@@ -51,7 +51,7 @@ public final class JdbcPageSource
         implements ConnectorPageSource
 {
     private static final Logger log = Logger.get(JdbcPageSource.class);
-    private static final CompletableFuture<ResultSet> UNINITIALIZED_RESULT_SET_FUTURE = CompletableFuture.completedFuture(null);
+    private static final CompletableFuture<Page> UNINITIALIZED_RESULT_SET_FUTURE = CompletableFuture.completedFuture(null);
 
     private final List<JdbcColumnHandle> columnHandles;
     private final ReadFunction[] readFunctions;
@@ -67,7 +67,7 @@ public final class JdbcPageSource
     private final PreparedStatement statement;
     private final AtomicLong readTimeNanos = new AtomicLong(0);
     private final PageBuilder pageBuilder;
-    private CompletableFuture<ResultSet> resultSetFuture;
+    private CompletableFuture<Page> resultSetFuture;
     @Nullable
     private ResultSet resultSet;
     private boolean finished;
@@ -155,13 +155,13 @@ public final class JdbcPageSource
     @Override
     public SourcePage getNextSourcePage()
     {
-        verify(pageBuilder.isEmpty(), "Expected pageBuilder to be empty");
         if (finished) {
             return null;
         }
         try {
             if (resultSetFuture == UNINITIALIZED_RESULT_SET_FUTURE && resultSet == null) {
                 checkState(!closed, "page source is closed");
+                // execute query async and then build first page from resultSet
                 resultSetFuture = supplyAsync(() -> {
                     long start = nanoTime();
                     try {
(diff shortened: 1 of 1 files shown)


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
```diff
@@ -133,7 +135,22 @@ else if (javaType == Slice.class) {
...
+                }
+                catch (SQLException e) {
+                    throw handleSqlException(e);
+                }
+                finally {
+                    readTimeNanos.addAndGet(nanoTime() - start);
+                }
+            }, executor).thenAccept(resultSet -> {
+                this.resultSet = requireNonNull(resultSet, "resultSet is null");
+                buildPageFromResultSet();
```
martint: Which thread does this work? If it's not the driver thread, then this means the connector is "cheating" and using more CPU than it's allocated by the engine.

chenjian2664: It uses the executor that installed in the `JdbcModule`, I think it is not the driver thread.

Checked the executor was introduced for preparing `resultSet` in a async way, is it a way also "cheating"
cc @raunaqmorarka 

Would it make sense to run `buildPageFromResultSet` on the current thread instead of using the thread pool? @martint

raunaqmorarka: It's not "cheating" when we're talking about mostly just waiting on result of a network call. We should go back to using main thread as soon as the `resultSet` is ready to be read from. I've seen JDBC and BigQuery splits sit on the task worker threads for several minutes at a time doing nothing but waiting on result of processing in another system, while other splits in Trino remain blocked from running.
The fair scheduling in the engine depends on the connectors being good citizens by not taking too long to produce a Page, and signalling through `isBlocked` that they're not ready to produce a Page yet.

chenjian2664: Thanks, switch consuming resultSet part to use main thread.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
```diff
@@ -45,13 +45,14 @@
 import static io.trino.plugin.jdbc.JdbcErrorCode.JDBC_ERROR;
 import static java.lang.System.nanoTime;
 import static java.util.Objects.requireNonNull;
+import static java.util.concurrent.CompletableFuture.runAsync;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
+import static org.weakref.jmx.$internal.guava.util.concurrent.MoreExecutors.directExecutor;
```
martint: Wrong import

chenjian2664:
@raunaqmorarka PTAL

chenjian2664:
gentle reminder @raunaqmorarka @martint

chenjian2664:
Any thoughts? @raunaqmorarka

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
```diff
@@ -133,7 +132,22 @@ else if (javaType == Slice.class) {
             pageBuilder = new PageBuilder(columnHandles.stream()
                     .map(JdbcColumnHandle::getColumnType)
                     .collect(toImmutableList()));
-            resultSetFuture = UNINITIALIZED_RESULT_SET_FUTURE;
```
findepi, line 136: it was clearly intentional to delay loading until the page source is first used
triggering query immediately is beneficial only if there is significant delay between construction and first use. why would be so?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
```diff
@@ -133,7 +132,22 @@ else if (javaType == Slice.class) {
...
+                    return statement.executeQuery();
+                }
+                catch (SQLException e) {
+                    throw handleSqlException(e);
+                }
+                finally {
+                    readTimeNanos.addAndGet(nanoTime() - start);
+                }
+            }, executor).thenAcceptAsync(resultSet -> {
+                this.resultSet = requireNonNull(resultSet, "resultSet is null");
```
findepi: `this` reference escapes during construction. This is generally not a good idea

findepi:
This PR lacks Description. What problem are we trying to solve, why do the change at all?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcPageSource.java
chenjian2664: Thanks!

findepi:
is the build failure important?

chenjian2664:
I don't know, can't reproduce on my local.
Rebased master.

sajjoseph:
Is there a memory leak possible in this change? I couldn't pinpoint where it could be happening, but following is what I observed. When I used this PR to see how the query performance changes, I noticed the memory allocation for a query (`query.max-memory-per-node`) on workers is getting exhausted and the query fails.

PR #26738, opened 2025-09-29
Add toString implementations

raunaqmorarka:
## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/tracking/TrackingInputFile.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/tracking/TrackingInputFile.java
@@ -76,4 +76,10 @@ public class TrackingInputFile
     {
         return delegate.location();
     }
+
+    @Override
+    public String toString()
+    {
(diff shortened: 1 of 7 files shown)


lib/trino-filesystem/src/main/java/io/trino/filesystem/tracking/TrackingInputFile.java
```diff
@@ -76,4 +76,10 @@ public Location location()
     {
         return delegate.location();
     }
+
+    @Override
+    public String toString()
```
wendigo, line 81: what about tracing input file?

wendigo: It has:

```
@Override
    public String toString()
    {
        return location().toString();
    }
```

Can you change it for consistency to call delegate instead?

raunaqmorarka: done

PR #26741, opened 2025-09-29
Add fast path for single partition in PagePartitioner

pettyjamesm:
## Description
Adds special case handling for single partition output in `PagePartitioner` such that partition calculations can be skipped when all rows will necessarily go to a single output partition. This improves performance for small queries or when running on small clusters to avoid the work of computing output page partitions.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/output/PagePartitioner.java
+++ b/core/trino-main/src/main/java/io/trino/operator/output/PagePartitioner.java
@@ -44,6 +44,7 @@ import java.util.OptionalInt;
 import java.util.function.IntUnaryOperator;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Verify.verify;
 import static io.trino.execution.buffer.PageSplitterUtil.splitAndSerializePage;
 import static io.trino.spi.block.PageBuilderStatus.DEFAULT_MAX_PAGE_SIZE_IN_BYTES;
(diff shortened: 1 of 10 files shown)


core/trino-main/src/main/java/io/trino/operator/output/PagePartitioner.java
```diff
@@ -138,7 +139,16 @@ public void partitionPage(Page page, OperatorContext operatorContext)
         }
 
         int outputPositionCount = replicatesAnyRow && !hasAnyRowBeenReplicated ? page.getPositionCount() + positionsAppenders.length - 1 : page.getPositionCount();
-        if (page.getPositionCount() < partitionFunction.partitionCount() * COLUMNAR_STRATEGY_COEFFICIENT) {
+        if (positionsAppenders.length == 1) {
+            // single output partition, skip partition calculation and append the entire page to the output partition
+            checkState(partitionFunction.partitionCount() == 1, "partitionFunction must be single partition");
```
lukasz-stec, line 144: How about doing this if in the `LocalExecutionPlanner` and using `TaskOutputOperator` instead of `PartitionedOutputOperator` there?

pettyjamesm: Yeah, I looked at that too and it seems to yield the same benefits in tests locally. The biggest difference is that PatitionedOutputOperator uses PagePartitioner to buffer pages before producing output, which prevents small pages from propagating across remote exchanges (where we know there will be a local exchange that further shreds the input page into smaller ones).

That should be fixable by adding a PageBuilder to TaskOutputOperator, but now the target buffer size needs to be different between “normal” and “partitioned to one partition” TaskOutputOperator instances (or buffering disabled on one, enabled on the other). PagePartitioner instances are already shared between drivers to avoid small pages at the split boundary which would be much more complex to reimplement for TaskOutputOperator, but I’m not sure how important that is in practice.

It made sense to me to special case this in PartitionedOutputOperator instead, but if a different line of reasoning makes you think the opposite approach is better I’m happy to switch implementations. Both approaches seem workable to me.

lukasz-stec: I reasoned that it would be easier in the planner, and we don't have to have the special case here, but either way looks good.
By the way, the Page partitioner sharing was targeted at a case with many partitions, such as FTE.

core/trino-main/src/test/java/io/trino/operator/output/TestPagePartitioner.java
```diff
@@ -595,22 +595,33 @@ private void testOutputEqualsInput(Type type, PartitioningMode mode1, Partitioni
...
-        pagePartitioner.close();
+        multiPartitionPartitioner.close();
 
-        List<Object> partitioned = readChannel(outputBuffer.getEnqueuedDeserialized(), 1, type);
-        assertThat(partitioned).containsExactlyInAnyOrderElementsOf(expected); // output of the PagePartitioner can be reordered
+        List<Object> multiPartitionedOutput = readChannel(outputBuffer.getEnqueuedDeserialized(), 1, type);
+        assertThat(multiPartitionedOutput).containsExactlyInAnyOrderElementsOf(expected); // output of the PagePartitioner can be reordered
         outputBuffer.clear();
+
+        // Test single partition output matches input
```
losipiuk, line 750: Are we exercising RLE/Dictionary specific code you added here?
If your new code is covered by existing tests didn't we loose some coverage for old code paths?

losipiuk (approved):
LGTM - asked some questions.

core/trino-main/src/test/java/io/trino/operator/output/TestPagePartitioner.java
pettyjamesm: The RLE / Dictionary specific code for `appendRange` is only called on single partitions, so no coverage was lost- I went back and added single partition testing for all the RLE / Dictionary scenarios already in `PagePartitioner` in the latest revision.

PR #26744, opened 2025-09-29
fix(iceberg): fix table creation without namespace location

### LINKED ISSUE #26742: [iceberg] non-staged table creation fails on non-s3-tables REST catalog
Hi all,

with a REST catalog which does not return a namespace location but at the same time is not s3tables, it's not possible to create a table.

The reason is that trino:

1. uses the existence of the namespace property location as a switch to decide if a table should be created as staged or not
2. then uses the `--table-s3` suffix as a switch to decide if it should check that a table location is empty after creation
3. any non s3tables catalog that doesn't return a namespace property location now fails on table creation since the conditions for non-staged creation & exemption from the is-empty-check are not the same

This early return makes the location become empty:

```java
    public String defaultTableLocation(ConnectorSession session, SchemaTableName schemaTableName)
    {
        String tableName = createLocationForTable(schemaTableName.getTableName());

        Map<String, Object> properties = loadNamespaceMetadata(session, schemaTableName.getSchemaName());
        String databaseLocation = (String) properties.get(IcebergSchemaProperties.LOCATION_PROPERTY);
        if (databaseLocation == null) {
            // Iceberg REST catalog doesn't require location property.
            // S3 Tables doesn't return the property.
            return null;
        }
        ...
```

This if decides on staged / non-staged create:

```java
    public Transaction newCreateTableTransaction(
            ...
    )
    {
            ...
            if (location.isEmpty()) {
                // TODO Replace with createTransaction once S3 Tables supports stage-create option
                return tableBuilder.create().newTransaction();
            }
            return tableBuilder.withLocation(location.get()).createTransaction();
        ...
    }
``` 

The non-staged create ends up writing to the table location which then leads this check to fail:

```java
if (!isS3Tables(location.toString())) {
                TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), transaction.table().io().properties());
                if (!replace && fileSystem.listFiles(location).hasNext()) {
                    throw new TrinoException(ICEBERG_FILESYSTEM_ERROR, format("" +
                            "Cannot create a table on a non-empty location: %s, set 'iceberg.unique-table-location=true' in your Iceberg catalog properties " +
                            "to use unique table locations for every table.", location));
                }
            }
```

### Potential solutions:

1. add a property which lets us disable the location-is-empty check
2. remember that `location.isEmpty()` was true and replace the `isS3Tables` condition with that

As an aside, would it also be acceptable to add a property to disable Trino's attempts at setting table locations entirely?

twuebi:
## Description

This should fix #26742 where it's impossible to create a table on non-s3tables REST catalogs that don't return a namespace location.

There are two separate conditions being used to determine if a table should be non-staged and if its location has to be empty. If the table is being created as non-staged, then the location won't be empty. 

## Release notes

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -1281,10 +1281,13 @@ public class IcebergMetadata
                     .orElseGet(() -> catalog.defaultTableLocation(session, tableMetadata.getTable()));
         }
         transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties);
+
         Location location = Location.of(transaction.table().location());
         try {
             // S3 Tables internally assigns a unique location for each table
-            if (!isS3Tables(location.toString())) {
+            // we create a non-staged table if tableLocation.isEmpty(), in that case, the table location will not be
+            // empty
+            if (!isS3Tables(location.toString()) && !(tableLocation != null && tableLocation.isEmpty())) {
                 TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), transaction.table().io().properties());
                 if (!replace && fileSystem.listFiles(location).hasNext()) {
                     throw new TrinoException(ICEBERG_FILESYSTEM_ERROR, format("" +
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -1281,7 +1281,6 @@ public class IcebergMetadata
                     .orElseGet(() -> catalog.defaultTableLocation(session, tableMetadata.getTable()));
         }
         transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties);
-
         Location location = Location.of(transaction.table().location());
         try {
             // S3 Tables internally assigns a unique location for each table


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1301,18 +1281,21 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
                     .orElseGet(() -> catalog.defaultTableLocation(session, tableMetadata.getTable()));
         }
         transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties);
+
```
wendigo: undo

twuebi: undone

ebyhr (commented):
Is it possible to add a test?

twuebi:
Sure, could you point me in the correct direction @ebyhr?

I likely simply didn't see it, where is this behavior currently being tested?

I checked `plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest` but didn't see

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1304,15 +1284,17 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
         Location location = Location.of(transaction.table().location());
         try {
             // S3 Tables internally assigns a unique location for each table
-            if (!isS3Tables(location.toString())) {
+            // we create a non-staged table if tableLocation.isEmpty(), in that case, the table location will not be
+            // empty
+            if (!isS3Tables(location.toString()) && !(tableLocation != null && tableLocation.isEmpty())) {
```
marcinsbd, line 1289: what about non-zero length whitespaced string :  " " or  "   " (more than one space) ?
should we also add  tableLocation.isBlank() e.g.

`if (!isS3Tables(location.toString()) && !(tableLocation != null && (tableLocation.isEmpty() || tableLocation.isBlank()))) {`

twuebi: Can add those, wondering if location should reject such invalid location strings?

twuebi: TBH, this all looks a bit hacky, ideally, we'd know if a table's creation was staged or not and only verify the location's emptiness based on that

chenjian2664: I may be missing some context here when I am reading the comment, but I'm not fully clear on why the empty check was added. Would it be possible to add a brief explanation to the code comment?

twuebi: @marcinsbd can probably answer better than me, I thought it was meant to treat empty string and null in the same way. I can still see the argument for rejecting invalid locations like empty string, maybe there's a robustness argument to be made in favor of treating "" == null.

marcinsbd: Probably more readable will be in form of:
`(tableLocation == null || !tableLocation.isEmpty())`

marcinsbd: (!isS3Tables(location.toString()) && !(tableLocation != null && tableLocation.isEmpty()))
TRUE: when ( not s3 Tables ) AND  ( tableLocation==NULL   OR  tableLocation is not Empty (so valid path?) )

marcinsbd: Having in mind the code
```
    public static Location of(String location)
    {
        requireNonNull(location, "location is null");
        checkArgument(!location.isEmpty(), "location is empty");
        checkArgument(!location.isBlank(), "location is blank");
```
just wanted to add the isBlank condition to also cover the non-zero length whitespaces string
to be in the same case as  "" (empty string)

i.e.
`!(tableLocation != null && tableLocation.isBlank()) becomes: tableLocation == null || !tableLocation.isBlank()`

after simplifying
`if (!isS3Tables(location.toString()) && (tableLocation == null || !tableLocation.isBlank())) `

marcinsbd: Although I can be wrong about it.
Nevertheless , I suggest adding the tests which probably will clarify the case.

twuebi: @marcinsbd, could you point me in the direction where this behavior is currently tested? It's not clear to me how to proceed with adding tests here, I checked `plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest` in September.

huangjc7:
Could you please provide support for this version as soon as possible? I desperately need it.
@wendigo @ebyhr  @chenjian2664  @marcinsbd

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
cyclane: Just chiming in with what I mentioned in #26742 - surely the check should just be?

```java
if (!isS3Tables(location.toString()) && tableLocation != null)
```

Since the reason it's `isEmpty()` in [some of the existing code](https://github.com/trinodb/trino/blob/3201583f47097298ab9b4832e9573688473c070e/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java#L423-L426) is because by then it is wrapped in `Optional.ofNullable(...)`, but here it is not.

cyclane: Also, just to try help unblock this PR, I had a go at implementing a test [here](https://github.com/cyclane/trino/commit/2b0c4d07e2524d5abf7e1f3294dcbccf2797185e).

@ebyhr @marcinsbd is this roughly what you were looking for?

cyclane:
@twuebi have you had a chance to look at this again? If not, I'm happy to open a new PR myself if that would be easier?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
marcinsbd: agree with @cyclane on the condition:
`if (!isS3Tables(location.toString()) && tableLocation != null)`
test also looks ok

cyclane: @marcinsbd I've put the condition and test into a separate PR since this one seems to be going stale again: #29640

Issue #26724, opened 2025-09-26
Preview UI feedback


wendigo:
- [ ] Broken live plan when plan is complex (gets rendered out of box)
- [ ] Slow rendering of query json when it's large

wendigo:
cc @koszti

koszti:
wrt broken live plan. can you please share an sql that’s using `tpch` data so we can reproduce the issue?

wendigo:
TPCDS Q01 is enough:

```
WITH
  customer_total_return AS (
   SELECT
     "sr_customer_sk" "ctr_customer_sk"
   , "sr_store_sk" "ctr_store_sk"
   , "sum"("sr_return_amt") "ctr_total_return"
   FROM
     tpcds.sf10.store_returns
   , tpcds.sf10.date_dim
   WHERE ("sr_returned_date_sk" = "d_date_sk")
      AND ("d_year" = 2000)
   GROUP BY "sr_customer_sk", "sr_store_sk"
)
SELECT "c_customer_id"
FROM
  customer_total_return ctr1
, tpcds.sf10.store
, tpcds.sf10.customer
WHERE ("ctr1"."ctr_total_return" > (
      SELECT ("avg"("ctr_total_return") * DECIMAL '1.2')
      FROM
        customer_total_return ctr2
      WHERE ("ctr1"."ctr_store_sk" = "ctr2"."ctr_store_sk")
   ))
   AND ("s_store_sk" = "ctr1"."ctr_store_sk")
   AND ("s_state" = 'TN')
   AND ("ctr1"."ctr_customer_sk" = "c_customer_sk")
ORDER BY "c_customer_id" ASC
LIMIT 100

```

wendigo:
Plan will be rendered off box: 

<img width="1446" height="1574" alt="Image" src="https://github.com/user-attachments/assets/02538e76-c48f-41a0-adf9-23c0bb872bd4" />

wendigo:
When you scroll it, it's there: 

<img width="1396" height="1568" alt="Image" src="https://github.com/user-attachments/assets/239ea2ba-d88b-410e-8786-9ac1f9a0ca0d" />

wendigo:
Same query causes JSON tab to hung for a bit

koszti:
@wendigo, I see two possible approaches:

1. Dynamically position the first stage so it’s always visible.
[...]

PR #26749, opened 2025-09-29
Fix query plan default viewport in preview UI

koszti:
## Description

Fix an issue with query live plan in preview UI when complex queries are not visible by default.
Partially addressing https://github.com/trinodb/trino/issues/26724

The query plan is rendered correctly but the default viewport often falls outside the visible area. This PR addresses the issue by making [fitView ](https://reactflow.dev/api-reference/react-flow#fitview) the default ensuring the entire flow is zoomed and positioned to fit within the window.

In addition it introduces an `Origin` button that dynamically centers on the first node, serving both as a convenient reset and as a default zoom-in option if the full flow initially appears too small.

image

## Release notes

```markdown
## Section
* Fix query plan default viewport in preview UI. ({issue}`26749`)
```

### DIFF
--- a/core/trino-web-ui/src/main/resources/webapp-preview/src/components/QueryLivePlan.tsx
+++ b/core/trino-web-ui/src/main/resources/webapp-preview/src/components/QueryLivePlan.tsx
@@ -13,12 +13,12 @@
  */
 import { useParams } from 'react-router-dom'
 import { useEffect, useRef, useState } from 'react'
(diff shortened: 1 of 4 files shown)


wendigo:
@koszti just one nit - can `ORIGIN` center the first stage?

koszti:
> @koszti just one nit - can `ORIGIN` center the first stage?

done that, updated the PR accordingly

PR #26759, opened 2025-09-30
Improve performance of Merge for MarkDistinct

hqbhoho:
## Description
Currently, `uniqueIdSymbol` is come from target table.In `INSERT` scenarios, `unique_id`  is always null, this causes the entire dataset to be processed by a single task, severely limiting parallelism and performance.
 ```
 // Mark distinct combinations of the unique_id value and the case_number
Symbol isDistinctSymbol = symbolAllocator.newSymbol("is_distinct", BOOLEAN);
MarkDistinctNode markDistinctNode = new MarkDistinctNode(idAllocator.getNextId(), project, isDistinctSymbol, ImmutableList.of(uniqueIdSymbol, caseNumberSymbol));

```
## Release notes

```markdown
## Section
```


core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
```diff
@@ -774,8 +774,16 @@ public MergeWriterNode plan(Merge merge)
         RelationPlan source = new RelationPlanner(analysis, symbolAllocator, idAllocator, lambdaDeclarationToSymbolMap, plannerContext, outerContext, session, recursiveSubqueries)
                 .process(merge.getSource());
 
+        // Assign a unique id to every source table row
+        Symbol sourceUniqueIdSymbol = symbolAllocator.newSymbol("source_unique_id", BIGINT);
+        RelationPlan sourcePlanWithUniqueId = new RelationPlan(
+                new AssignUniqueId(idAllocator.getNextId(), source.getRoot(), sourceUniqueIdSymbol),
```
kasiafi, line 779: Is there guarantee that the unique IDs produced by two different operators (one for target and one for source) are mutually distinct? Even if it's implemented this way, I would not assume this is part of the contract for AssignUniqueId. We should use some additional mechanism to avoid clashes.

core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
```diff
@@ -897,7 +906,12 @@ public MergeWriterNode plan(Merge merge)
             Symbol symbol = planWithPresentColumn.getFieldMappings().get(fieldIndex);
             projectionAssignmentsBuilder.putIdentity(symbol);
         }
-        projectionAssignmentsBuilder.putIdentity(uniqueIdSymbol);
+
+        // define a unique id come from target or source, if join match will use target, otherwise use source
```
kasiafi: Please explain in the comment why using the unique ID from source side does not break semantics, and that it is done to improve performance of MarkDistinct.

kasiafi: Also, after this change, the `isNull` check is redundant:
```
Logical.and(
      not(metadata, isDistinctSymbol.toSymbolReference()),
      not(metadata, new IsNull(uniqueIdSymbol.toSymbolReference()))),
```

kasiafi (commented):
Please make sure we have tests for all the involved paths:
- all source rows unmatched
- a mix of matched and unmatched rows
Same cases with failure from duplicate match

hqbhoho:
@kasiafi Thanks for your feedback.
In the following example：
```
trino> select * from iceberg.tmp.test_merge_01;
 id | name
----+------
(0 rows)

trino> select * from iceberg.tmp.test_merge_02;
 id | name
----+------
  1 |    1
  2 |    2
(2 rows)

trino> EXPLAIN MERGE INTO iceberg.tmp.test_merge_01 a
    -> using iceberg.tmp.test_merge_02 b
    -> on a.id = b.id
    -> WHEN MATCHED
    ->         THEN UPDATE SET name = b.name
    ->     WHEN NOT MATCHED
    ->         THEN INSERT (id ,name)
    ->               VALUES(b.id, b.name);
```
The logical plan (Trino version 443) is:
```
Fragment 0 [COORDINATOR_ONLY]                                                                                                                                                                                     >
     Output layout: [rows]                                                                                                                                                                                         >
     Output partitioning: SINGLE []                                                                                                                                                                                >
     Output[columnNames = [rows]]                                                                                                                                                                                  >
     │   Layout: [rows:bigint]                                                                                                                                                                                     >
     │   Estimates: {rows: ? (?), cpu: 0, memory: 0B, network: 0B}                                                                                                                                                 >
     └─ TableCommit[target = iceberg:tmp.test_merge_01$data@1242211021876735512]                                                                                                                                   >
        │   Layout: [rows:bigint]                                                                                                                                                                                  >
        └─ LocalExchange[partitioning = SINGLE]                                                                                                                                                                    >
           │   Layout: [partialrows:bigint, fragment:varbinary]                                                                                                                                                    >
           │   Estimates: {rows: ? (?), cpu: 0, memory: 0B, network: 0B}                                                                                                                                           >
           └─ RemoteSource[sourceFragmentIds = [1]]                                                                                                                                                                >
                  Layout: [partialrows:bigint, fragment:varbinary]                                                                                                                                                 >
                                                                                                                                                                                                                   >
 Fragment 1 [MERGE [update = iceberg:INSTANCE]]                                                                                                                                                                    >
     Output layout: [partialrows, fragment]                                                                                                                                                                        >
     Output partitioning: SINGLE []                                                                                                                                                                                >
     MergeWriter[table = iceberg:tmp.test_merge_01$data@1242211021876735512]                                                                                                                                       >
     │   Layout: [partialrows:bigint, fragment:varbinary]                                                                                                                                                          >
     └─ LocalExchange[partitioning = MERGE [update = iceberg:INSTANCE], arguments = [SymbolReference[type=tinyint, name=operation], SymbolReference[type=row(_file varchar, _pos bigint, partition_spec_id integer,>
        │   Layout: [id:integer, name:integer, operation:tinyint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), insert_from_update:tinyint]                            >
        │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                              >
        └─ RemoteSource[sourceFragmentIds = [2]]                                                                                                                                                                   >
               Layout: [id:integer, name:integer, operation:tinyint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), insert_from_update:tinyint]                         >
                                                                                                                                                                                                                   >
 Fragment 2 [HASH]                                                                                                                                                                                                 >
     Output layout: [id, name, operation, field, insert_from_update]                                                                                                                                               >
     Output partitioning: MERGE [update = iceberg:INSTANCE] [operation, field]                                                                                                                                     >
     MergeProcessor[]                                                                                                                                                                                              >
     │   Layout: [id:integer, name:integer, operation:tinyint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), insert_from_update:tinyint]                               >
     │   target: iceberg:tmp.test_merge_01$data@1242211021876735512                                                                                                                                                >
     │   merge row column: merge_row                                                                                                                                                                               >
     │   row id column: field                                                                                                                                                                                      >
     │   redistribution columns: []                                                                                                                                                                                >
     │   data columns: [id, name]                                                                                                                                                                                  >
     └─ Filter[filterPredicate = (CASE WHEN ((NOT is_distinct) AND (NOT (unique_id IS NULL))) THEN CAST(system.builtin.fail(integer '119', varchar 'One MERGE target table row matched more than one source row') A>
        │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), merge_row:row(id integer, name integer, boolean, tinyint, integer), case_number:in>
        │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                              >
        └─ MarkDistinct[distinct = [unique_id:bigint, case_number:integer], marker = is_distinct]                                                                                                                  >
           │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), merge_row:row(id integer, name integer, boolean, tinyint, integer), case_number>
           └─ LocalExchange[partitioning = HASH, arguments = [SymbolReference[type=bigint, name=unique_id], SymbolReference[type=integer, name=case_number]]]                                                      >
              │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), merge_row:row(id integer, name integer, boolean, tinyint, integer), case_num>
              │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                        >
              └─ RemoteSource[sourceFragmentIds = [3]]                                                                                                                                                             >
                     Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), merge_row:row(id integer, name integer, boolean, tinyint, integer), case_>
                                                                                                                                                                                                                   >
 Fragment 3 [HASH]                                                                                                                                                                                                 >
     Output layout: [unique_id, field, merge_row, case_number]                                                                                                                                                     >
     Output partitioning: HASH [unique_id, case_number]                                                                                                                                                            >
     Project[]                                                                                                                                                                                                     >
     │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), merge_row:row(id integer, name integer, boolean, tinyint, integer), case_number:integ>
     │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                                 >
     │   case_number := merge_row[integer '5']                                                                                                                                                                     >
     └─ Project[]                                                                                                                                                                                                  >
        │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), merge_row:row(id integer, name integer, boolean, tinyint, integer)]               >
        │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                              >
        │   merge_row := (CASE WHEN present THEN ROW (id, name_1, (NOT (present IS NULL)), tinyint '3', integer '0') WHEN (present IS NULL) THEN ROW (id_0, name_1, (NOT (present IS NULL)), tinyint '1', integer '>
        └─ RightJoin[criteria = (id = id_0), distribution = PARTITIONED]                                                                                                                                           >
           │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), id:integer, present:boolean, id_0:integer, name_1:integer]                     >
           │   Estimates: {rows: ? (?), cpu: ?, memory: ?, network: 0B}                                                                                                                                            >
           │   Distribution: PARTITIONED                                                                                                                                                                           >
           │   dynamicFilterAssignments = {id_0 -> #df_727}                                                                                                                                                        >
           ├─ RemoteSource[sourceFragmentIds = [4]]                                                                                                                                                                >
           │      Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), id:integer, present:boolean]                                                >
           └─ LocalExchange[partitioning = HASH, arguments = [SymbolReference[type=integer, name=id_0]]]                                                                                                           >
              │   Layout: [id_0:integer, name_1:integer]                                                                                                                                                           >
              │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                        >
              └─ RemoteSource[sourceFragmentIds = [5]]                                                                                                                                                             >
                     Layout: [id_0:integer, name_1:integer]                                                                                                                                                        >
                                                                                                                                                                                                                   >
 Fragment 4 [SOURCE]                                                                                                                                                                                               >
     Output layout: [unique_id, field, id, present]                                                                                                                                                                >
     Output partitioning: HASH [id]                                                                                                                                                                                >
     Project[]                                                                                                                                                                                                     >
     │   Layout: [unique_id:bigint, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), id:integer, present:boolean]                                                         >
     │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                                 >
     │   present := boolean 'true'                                                                                                                                                                                 >
     └─ AssignUniqueId[]                                                                                                                                                                                           >
        │   Layout: [id:integer, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar), unique_id:bigint]                                                                       >
        │   Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                              >
        └─ ScanFilter[table = iceberg:tmp.test_merge_01$data@1242211021876735512, dynamicFilters = {id = #df_727}]                                                                                                 >
               Layout: [id:integer, field:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar)]                                                                                      >
               Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}/{rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                            >
               field := -2147483647:$row_id:row(_file varchar, _pos bigint, partition_spec_id integer, partition_data varchar)                                                                                     >
               id := 1:id:integer                                                                                                                                                                                  >
                                                                                                                                                                                                                   >
 Fragment 5 [SOURCE]                                                                                                                                                                                               >
     Output layout: [id_0, name_1]                                                                                                                                                                                 >
     Output partitioning: HASH [id_0]                                                                                                                                                                              >
     TableScan[table = iceberg:tmp.test_merge_02$data@3640267961122064680]                                                                                                                                         >
         Layout: [id_0:integer, name_1:integer]                                                                                                                                                                    >
         Estimates: {rows: ? (?), cpu: ?, memory: 0B, network: 0B}                                                                                                                                                 >
         name_1 := 2:name:integer                                                                                                                                                                                  >
         id_0 := 1:id:integer
```
In Fragment 2,  `MarkDistinct` will base on`unique_id` and `case_number`, when inserting new rows into the target table, a Right Join will not match any rows, so the `unique_id` of the newly joined data should all be `null`.This would cause all newly inserted data to be processed within a single task during `MarkDistinct`.
```
 MarkDistinct[distinct = [unique_id:bigint, case_number:integer], marker = is_distinct] 
```
My current implementation is such that if a Right Join does not match any rows, it uses the `unique_id` generated by the original table instead of `null` and newly inserted data to be processed within  multiple tasks during `MarkDistinct`. Therefore, this approach would not break the existing semantics.

I'm not sure if this implementation approach is feasible, or whether there might be a better way to implement it.If this solution is feasible, I will proceed with the work.

kasiafi:
Hi @hqbhoho 
in the example you mention, it's indeed correct. In case when there are no matches (which happens when the target table is empty), the coalesce will return the unique id from source for each row.

The problematic case is when unique ids from target get mixed with unique ids from source, and they accidentally clash -- so we have a false duplicate. Unique ids from target nad source get mixed when there is partial match: some source rows match to target rows, and some are unmatched.

To disambiguate the target and source unique ids, we propose a change to `AssignUniqueId` operator. We will allow to use additional mask to make sure that different AssignUniqueId operations produce non-overlapping results. @losipiuk is working on it.

hqbhoho:
@kasiafi 
In current implementation, A `MarkDistinct` node adds an is_distinct column which is true if no other row has the same unique id and WHEN clause number, and false otherwise. 

So It will check rows which is matched, in other words, the same match condition cannot be used for an `UPDATE` operation more than twice.
But rows which is not matched will not check because of `unique_id` is aways `null`, so we can insert two same rows if the row not in target table, Is this correct? 

I think after right join, the  `unique_id` should not `null` whether matched or not. Currently, the `unique_id` come from the target table，so after right join it could be `null` when not matched.  

Only `NOT MATCHED` will use source table `unique_id`, so there will be no problem because  additional `case_nunber` field will ensure uniqueness.

core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
losipiuk: I added the commit to address that - but if that is not needed feel free to drop :)
@kasiafi can you PTAL?

kasiafi:
>Only NOT MATCHED will use source table unique_id, so there will be no problem because additional case_nunber field will ensure uniqueness.

@hqbhoho thanks for the explanation. you're right. The case number is also considered, and MERGE cases cannot overlap between the matched and unmatched rows. Your solution is thus correct. 
We can drop @losipiuk change for now, and revisit it when there's need to disambiguate between different AssignUniqueId operations.

Feel free to proceed with your work.

core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
kasiafi: @losipiuk as @hqbhoho [explained](https://github.com/trinodb/trino/pull/26759#issuecomment-3385245375), we don't need this disambiguation here, as it's guaranteed by considering the merge branch.

hqbhoho:
@kasiafi  PTAL when you have time, thanks!

kasiafi:
@hqbhoho great job, thanks!

chenjian2664:
cool stuff!

PR #26766, opened 2025-09-30
Random improvements here and there

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/dispatcher/QueuedStatementResource.java
+++ b/core/trino-main/src/main/java/io/trino/dispatcher/QueuedStatementResource.java
@@ -100,6 +100,7 @@ import static jakarta.ws.rs.core.MediaType.APPLICATION_JSON;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.Executors.newSingleThreadScheduledExecutor;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
 @Path("/v1/statement")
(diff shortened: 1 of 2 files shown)


core/trino-main/src/main/java/io/trino/dispatcher/QueuedStatementResource.java
```diff
@@ -344,9 +345,9 @@ public long getLastToken()
             return lastToken.get();
         }
 
-        public boolean tryAbandonSubmissionWithTimeout(Duration querySubmissionTimeout)
+        public boolean tryAbandonSubmissionWithTimeout(long querySubmissionTimeoutNanos)
```
martint, line 358: Duration provides better type safety for this method. Is the instantiation of Duration in the body of this method a real problem?

wendigo: I was running Trino under high QPS load (600 QPS) and this stood out in the allocation profile (one of the first non-jdk classes along with DataSize)

martint: Did it cause actual GC problems (higher CPU usage by GC, GC pressure, etc)? Short-lived allocations are essentially "free", so the fact that they show up in the allocation profiling may be misleading.

wendigo: These could be promoted to old gen too (since we use these to track query progression)

martint: This line allocates it, uses it and throws it away:

```
Duration.nanosSince(initTime).compareTo(querySubmissionTimeout) >= 0
```

wendigo: Yeah, this one does, but generally I've been looking at other places where we keep Durations for a longer period of time

PR #26767, opened 2025-09-30
Improve QueryId and TaskId

pettyjamesm:
## Description
Avoids constantly re-parsing `QueryId` and `StageId` values from `TaskId` instances by caching the result of those parses directly. This is particularly relevant for `TaskId#getQueryId()` which happens frequently inside of `MemoryPool` when tracking query level memory and relatively more expensive operations inflate the amount of time that each memory update operation needs to hold the lock.

Also improves `QueryId.parseDottedId` by switching to a concrete implementation that avoids using a regular expression to split the input. 

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/TaskId.java
+++ b/core/trino-main/src/main/java/io/trino/execution/TaskId.java
@@ -39,12 +39,17 @@ public class TaskId
     }
 
     private final String fullId;
+    private final StageId stageId;
+    private final int partitionId;
+    private final int attemptId;
 
     public TaskId(StageId stageId, int partitionId, int attemptId)
     {
-        requireNonNull(stageId, "stageId is null");
+        this.stageId = requireNonNull(stageId, "stageId is null");
         checkArgument(partitionId >= 0, "partitionId is negative: %s", partitionId);
         checkArgument(attemptId >= 0, "attemptId is negative: %s", attemptId);
+        this.partitionId = partitionId;
+        this.attemptId = attemptId;
 
         // There is a strange JDK bug related to the CompactStrings implementation in JDK20+ which causes some fullId values
         // to get corrupted when this particular line is JIT-optimized. Changing implicit concatenation to a String.join call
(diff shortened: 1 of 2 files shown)


wendigo:
🚀

core/trino-spi/src/main/java/io/trino/spi/QueryId.java
```diff
@@ -97,23 +98,42 @@ public static String validateId(String id)
...
 
-        List<String> ids = List.of(id.split("\\."));
-        checkArgument(ids.size() == expectedParts, "Invalid %s %s", name, id);
-
-        for (String part : ids) {
+        String[] parts = new String[expectedParts];
+        int startOffset = 0;
+        for (int i = 0; i < parts.length - 1; i++) {
+            int dotIndex = id.indexOf('.', startOffset);
+            if (dotIndex < 0) {
```
wendigo, line 108: `verify(dotIndex > 0, "Invalid %s %s", name, id");`

core/trino-spi/src/main/java/io/trino/spi/QueryId.java
```diff
@@ -97,23 +98,42 @@ public static String validateId(String id)
...
+                throw new IllegalArgumentException(format("Invalid %s %s", name, id));
+            }
+            String part = id.substring(startOffset, dotIndex);
```
wendigo, line 111: we could check if startOffset - dotIndex > 0 to check whether part is empty, we can then simplify validateId to drop is null and is empty checks (move it to constructor)

then this method becomes:

```
    public static List<String> parseDottedId(String id, int expectedParts, String name)
    {
        requireNonNull(id, "id is null");
        checkArgument(expectedParts > 1, "expectedParts must be at least 2");
        requireNonNull(name, "name is null");

        String[] parts = new String[expectedParts];
        int startOffset = 0;
        for (int i = 0; i < parts.length - 1; i++) {
            int dotIndex = id.indexOf('.', startOffset);
            checkArgument(dotIndex - startOffset > 0, "Invalid %s %s", name, id);
            parts[i] = validateId(id.substring(startOffset, dotIndex));
            startOffset = dotIndex + 1;
        }
        checkArgument(id.indexOf('.', startOffset) == -1, "Invalid %s %s", name, id);
        parts[parts.length - 1] = validateId(id.substring(startOffset));
        return List.of(parts);
    }
```
`isValid` returns validated String so we can use it directly in the assignment

wendigo: We need to make `checkArgument` accept vararg too:

```
    @FormatMethod
    private static void checkArgument(boolean condition, @FormatString String message, Object... argument)
    {
        if (!condition) {
            throw new IllegalArgumentException(format(message, argument));
        }
    }
```

wendigo:
I've added a couple of changes on top of your PR here: https://github.com/trinodb/trino/pull/26769

pettyjamesm:
Closing in favor of https://github.com/trinodb/trino/pull/26769

PR #26777, opened 2025-10-01
Planner cleanup

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/cost/CachingCostProvider.java
+++ b/core/trino-main/src/main/java/io/trino/cost/CachingCostProvider.java
@@ -38,7 +38,7 @@ public class CachingCostProvider
     private final Optional<Memo> memo;
     private final Session session;
 
-    private final Map<PlanNode, PlanCostEstimate> cache = new IdentityHashMap<>();
+    private final Map<PlanNode, PlanCostEstimate> cache = new IdentityHashMap<>(0);
 
     public CachingCostProvider(CostCalculator costCalculator, StatsProvider statsProvider, Session session)
     {


lib/trino-matching/src/test/java/io/trino/matching/TestMatcher.java
```diff
@@ -140,7 +139,12 @@ public void capturingMatchesInATypesafeManner()
 
         ProjectNode tree = new ProjectNode(new FilterNode(new ScanNode("orders"), null));
 
+        System.out.println(pattern);
```
raunaqmorarka: remove ?

lib/trino-matching/src/test/java/io/trino/matching/TestMatcher.java
```diff
@@ -140,7 +139,12 @@ public void capturingMatchesInATypesafeManner()
...
+
         Match match = assertMatch(pattern, tree);
+
+        System.out.println(match);
```
raunaqmorarka: remove ?

wendigo: oh right :)

lib/trino-matching/src/main/java/io/trino/matching/Pattern.java
```diff
@@ -103,32 +103,33 @@ public Optional<Pattern<?>> previous()
...
     public <C> boolean matches(Object object, C context)
     {
-        return match(object, context)
-                .findFirst()
-                .isPresent();
+        return !match(object, context).isEmpty();
     }
 
-    public final Stream<Match> match(Object object)
+    public final List<Match> match(Object object)
```
martint: Using stream is intentional to be able to allow lazy matching

wendigo: Do we ever do?

martint: Yes, see OrPattern, for instance:

```
        return patterns.stream()
                .flatMap(pattern -> pattern.accept(object, captures, context));
```

wendigo: Do we have more than two terms in the or patterns?

martint: It's possible. This is a framework, so that's a supported use case

wendigo: Dropping controversial commit. I'll reimplement this without streams but using lazy API

wendigo:
Ping @raunaqmorarka

PR #26790, opened 2025-10-02
Add dependabot for maven dependencies

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -14,3 +14,20 @@ updates:
       security-updates:
         applies-to: security-updates
         dependency-type: production
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    assignees:
+      - wendigo
+    open-pull-requests-limit: 10
+    groups:
+      dependency-updates:
+        applies-to: version-updates
+        update-types:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -14,3 +14,20 @@ updates:
       security-updates:
         applies-to: security-updates
         dependency-type: production
+  - package-ecosystem: "maven"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
raunaqmorarka, line 20: Why does it have to be daily ?
There is nothing so urgent about dependency upgrades that makes it necessary to do every day.
Can we make it weekly to reduce the amount of PRs this will generate ?

.github/dependabot.yml
```diff
@@ -14,3 +14,20 @@ updates:
...
+    assignees:
+      - wendigo
+    open-pull-requests-limit: 10
```
raunaqmorarka, line 23: Why do we need 10 open PRs about dependency upgrade ?
Can't we have it limited to 2 or 3 ?

wendigo: Let's see in a week if the amount of PRs is too high

PR #26795, opened 2025-10-02
Convert StageId and TaskId to a record

wendigo:
QueryId was already converted

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/TaskId.java
+++ b/core/trino-main/src/main/java/io/trino/execution/TaskId.java
@@ -70,7 +70,7 @@ public class TaskId
 
     public QueryId getQueryId()
     {
-        return stageId.getQueryId();
+        return stageId.queryId();
     }
 
     public StageId getStageId()
--- a/core/trino-main/src/main/java/io/trino/execution/TaskId.java
+++ b/core/trino-main/src/main/java/io/trino/execution/TaskId.java
@@ -25,67 +25,35 @@ import static io.airlift.slice.SizeOf.estimatedSizeOf;
 import static io.airlift.slice.SizeOf.instanceSize;
 import static io.trino.spi.QueryId.parseDottedId;
 import static java.lang.Integer.parseInt;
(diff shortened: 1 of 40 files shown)


wendigo:
Important because of https://github.com/openjdk/jdk/blob/dfd383224dbc2e41c9f44b1acd09ffb179cc38f3/src/hotspot/share/ci/ciField.cpp#L237

Also less lines :)

losipiuk (approved):

core/trino-main/src/main/java/io/trino/execution/TaskId.java
```diff
@@ -25,67 +25,35 @@
 import static io.airlift.slice.SizeOf.instanceSize;
 import static io.trino.spi.QueryId.parseDottedId;
 import static java.lang.Integer.parseInt;
-import static java.lang.String.join;
 import static java.util.Objects.requireNonNull;
 
-public class TaskId
+public record TaskId(String fullId, StageId stageId, int partitionId, int attemptId)
```
martint, line 30: This  now exposes a bunch of private fields, so it should not be a record.

wendigo: Except for `fullId` all of these fields were accessed through the corresponding getters:

```
    public StageId getStageId()
    {
        return stageId;
    }

    public int getPartitionId()
    {
        return partitionId;
    }

    public int getAttemptId()
    {
        return attemptId;
    }
```

I don't see how accessing `fullId()` which you can anyway get through `toString()` breaks encapsulation. All of these fields were accessible externally through the previously exposed API

martint: It does break encapsulation because it allows TaskId to be constructed with stageId/partitionId/attemptId that don't agree with fullId

wendigo: https://github.com/trinodb/trino/pull/26799

PR #26797, opened 2025-10-02
Create and drop catalog tasks utilize Metadata class

homar:
## Description
Notify system security when Catalog is Created/Dropped

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
@@ -35,6 +35,7 @@ import io.trino.spi.ErrorCode;
 import io.trino.spi.QueryId;
 import io.trino.spi.RefreshType;
 import io.trino.spi.TrinoException;
+import io.trino.spi.catalog.CatalogName;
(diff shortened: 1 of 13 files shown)


kokosing:
Thanks!

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -793,6 +798,29 @@ public List<RelationCommentMetadata> listRelationComments(Session session, Strin
...
+            systemSecurityMetadata.catalogCreated(session, catalog);
+        }
+    }
+
+    @Override
+    public void dropCatalog(Session session, CatalogName catalog, boolean cascade)
+    {
+        Optional<CatalogMetadata> catalogMetadata = Optional.empty();
+        if (catalogManager.getCatalog(catalog).isPresent()) {
+            catalogMetadata = Optional.of(getCatalogMetadataForWrite(session, catalog.toString()));
```
findepi, line 816: getCatalogMetadataForWrite registers catalog for cleanup, but the cleanup may fail.

- https://github.com/trinodb/trino/pull/27025

PR #26799, opened 2025-10-02
Validate whether TaskId was constructed properly

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/TaskId.java
+++ b/core/trino-main/src/main/java/io/trino/execution/TaskId.java
@@ -44,11 +44,12 @@ public record TaskId(String fullId, StageId stageId, int partitionId, int attemp
         requireNonNull(stageId, "stageId is null");
         checkArgument(partitionId >= 0, "partitionId is negative: %s", partitionId);
         checkArgument(attemptId >= 0, "attemptId is negative: %s", attemptId);
+        checkArgument(fullId.equals(constructFullId(stageId, partitionId, attemptId)), "fullId does not match provided components");
     }
 
     public TaskId(StageId stageId, int partitionId, int attemptId)
     {
-        this("%s.%s.%s".formatted(stageId.toString(), String.valueOf(partitionId), String.valueOf(attemptId)), stageId, partitionId, attemptId);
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/execution/TaskId.java
```diff
@@ -44,11 +44,12 @@ public static TaskId valueOf(String fullId)
         requireNonNull(stageId, "stageId is null");
         checkArgument(partitionId >= 0, "partitionId is negative: %s", partitionId);
         checkArgument(attemptId >= 0, "attemptId is negative: %s", attemptId);
+        checkArgument(fullId.equals(constructFullId(stageId, partitionId, attemptId)), "fullId does not match provided components");
```
martint, line 47: The problem is that this is now a runtime check. Before, it was physically impossible to construct an invalid TaskId object. Either the fullId should be provided during construction or the individual parts, but not both.

wendigo: I think that benefits of a record class outweights the cost of runtime check in that case. But it's just my opinion.

martint: What are the benefits aside from a terser implementation?

wendigo: https://github.com/openjdk/jdk/blob/dfd383224dbc2e41c9f44b1acd09ffb179cc38f3/src/hotspot/share/ci/ciField.cpp#L237 record classes are treated as effectively final (which isn't the case for ordinary final class with final fields). This allows a bunch of optimizations on top of that.

martint: That's temporary benefit. The JVM will continue to get better.

Also, it's only a theoretical benefit for this specific case. Have you benchmarked it to see if it actually matters?

Finally, the runtime check may outweigh all those benefits, performance-wise.

wendigo: Yeah, but I guess we will wait for Valhalla couple of releases more :)

wendigo:
Replaced by https://github.com/trinodb/trino/pull/26801

PR #26800, opened 2025-10-02
Improve BlockBuilder#appendRange null handling

pettyjamesm:
## Description
Adds a fast-path through `BlockBuilder#appendRange` to use `System.arraycopy` for the null mask (when present) once both null and non-null values have been found.

Benchmark results: [jmh.morethan.io](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/b7ece413528b0786109518e4aedf5c67/raw/c83d1a20a14242251b2db6daba0a6d1e1f55c012/baseline_blockbuilder.json,https://gist.githubusercontent.com/pettyjamesm/b7ece413528b0786109518e4aedf5c67/raw/c83d1a20a14242251b2db6daba0a6d1e1f55c012/improved_blockbuilder.json)

## Release notes

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlockBuilder.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlockBuilder.java
@@ -180,12 +180,15 @@ public class ArrayBlockBuilder
         boolean[] rawValueIsNull = arrayBlock.getRawValueIsNull();
         if (rawValueIsNull != null) {
             for (int i = 0; i < length; i++) {
-                if (rawValueIsNull[rawOffsetBase + offset + i]) {
-                    valueIsNull[positionCount + i] = true;
-                    hasNullValue = true;
+                boolean isNull = rawValueIsNull[rawOffsetBase + offset + i];
+                hasNullValue |= isNull;
+                hasNonNullValue |= !isNull;
+                if (hasNullValue & hasNonNullValue) {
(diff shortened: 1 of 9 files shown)


martint:
What's the performance impact of this change?

pettyjamesm:
> What's the performance impact of this change?

Haven’t had a chance to micro-benchmark it, but the intuition is that when nulls and non-nulls are intermixed the previous branching approach will pay mis-predict penalties frequently (no nulls should be already mostly handled by the block.mayHaveNull() check, but even if not the branches will predict reliably and be mostly equivalent). The new strategy will switch directly to array copies once one null and one non-null has been observed which should always be faster than per-boolean looping.

This improvement will stack on top of https://github.com/trinodb/trino/pull/26741 (for single node executions) once that lands, but will still improve page merging without those changes landing yet. Not sure if that’s going to show up in the benchmarks you ran.

pettyjamesm:
Benchmark results: [jmh.morethan.io](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/b7ece413528b0786109518e4aedf5c67/raw/c83d1a20a14242251b2db6daba0a6d1e1f55c012/baseline_blockbuilder.json,https://gist.githubusercontent.com/pettyjamesm/b7ece413528b0786109518e4aedf5c67/raw/c83d1a20a14242251b2db6daba0a6d1e1f55c012/improved_blockbuilder.json)

**TL;DR**: slight regression with 100% nulls and 0% nulls (although in practice most blocks without nulls have no `isNull` mask and would not regress) with significant lifts in all cases that mix nulls and non-nulls.

Issue #26751, opened 2025-09-30
Lakehouse connector throws exception when reading system tables


ebyhr:
```sql
TABLE lakehouse.iceberg."region$files";
```

```
java.lang.UnsupportedOperationException: Unsupported table handle class io.trino.connector.system.SystemTableHandle
	at io.trino.plugin.lakehouse.LakehousePageSourceProviderFactory.forHandle(LakehousePageSourceProviderFactory.java:66)
	at io.trino.plugin.lakehouse.LakehousePageSourceProviderFactory.lambda$createPageSourceProvider$0(LakehousePageSourceProviderFactory.java:56)
	at io.trino.connector.system.SystemPageSourceProvider.createPageSource(SystemPageSourceProvider.java:92)
	at io.trino.split.PageSourceManager$PageSourceProviderInstance.createPageSource(PageSourceManager.java:79)
	at io.trino.operator.TableScanOperator.getOutput(TableScanOperator.java:259)
	at io.trino.operator.Driver.processInternal(Driver.java:403)
	at io.trino.operator.Driver.lambda$process$0(Driver.java:306)
	at io.trino.operator.Driver.tryWithLock(Driver.java:709)
	at io.trino.operator.Driver.process(Driver.java:298)
	at io.trino.operator.Driver.processForDuration(Driver.java:269)
	at io.trino.execution.SqlTaskExecution$DriverSplitRunner.processFor(SqlTaskExecution.java:851)
	at io.trino.execution.executor.dedicated.SplitProcessor.run(SplitProcessor.java:77)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.lambda$run$0(TaskEntry.java:205)
	at io.trino.$gen.Trino_testversion____20250930_024623_1.run(Unknown Source)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.run(TaskEntry.java:206)
	at io.trino.execution.executor.scheduler.FairScheduler.runTask(FairScheduler.java:177)
	at io.trino.execution.executor.scheduler.FairScheduler.lambda$submit$0(FairScheduler.java:164)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:128)
[...]

PR #26830, opened 2025-10-03
Fix failure when reading Iceberg `$files` table in Lakehouse connector

codluca:
Fix support for system tables for Lakehouse.
The $files and $history system tables for Iceberg tables are not handled properly.

Fixes https://github.com/trinodb/trino/issues/26751

## Description

For system tables, the handle will be an instance of io.trino.connector.system.SystemTableHandle.
Some Iceberg system tables ($files and $history) use special splits (FilesTableSplit).

io.trino.connector.system.SystemTableHandle is part of trino-main, which is a test dependency for trino-lakehouse.
The dependency to trino-main is not changed.

## Release notes

```markdown
## Lakehouse
* Fix failure when reading Iceberg `$files` table. ({issue}`26751`)
```

### DIFF
--- a/plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
+++ b/plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
@@ -22,8 +22,10 @@ import io.trino.plugin.hudi.HudiPageSourceProvider;
 import io.trino.plugin.hudi.HudiTableHandle;
 import io.trino.plugin.iceberg.IcebergPageSourceProviderFactory;
 import io.trino.plugin.iceberg.IcebergTableHandle;
+import io.trino.plugin.iceberg.system.files.FilesTableSplit;
 import io.trino.spi.connector.ConnectorPageSourceProvider;
 import io.trino.spi.connector.ConnectorPageSourceProviderFactory;
+import io.trino.spi.connector.ConnectorSplit;
 import io.trino.spi.connector.ConnectorTableHandle;
 
 import static java.util.Objects.requireNonNull;
@@ -53,17 +55,21 @@ public class LakehousePageSourceProviderFactory
     public ConnectorPageSourceProvider createPageSourceProvider()
     {
         return (transaction, session, split, table, columns, dynamicFilter) ->
-                forHandle(table).createPageSource(transaction, session, split, table, columns, dynamicFilter);
+                forHandle(split, table).createPageSource(transaction, session, split, table, columns, dynamicFilter);
     }
 
-    private ConnectorPageSourceProvider forHandle(ConnectorTableHandle handle)
+    private ConnectorPageSourceProvider forHandle(ConnectorSplit split, ConnectorTableHandle handle)
     {
+        if (split instanceof FilesTableSplit) {
+            return icebergPageSourceProviderFactory.createPageSourceProvider();
+        }
+
         return switch (handle) {
             case HiveTableHandle _ -> hivePageSourceProvider;
             case IcebergTableHandle _ -> icebergPageSourceProviderFactory.createPageSourceProvider();
             case DeltaLakeTableHandle _ -> deltaLakePageSourceProvider;
             case HudiTableHandle _ -> hudiPageSourceProvider;
-            default -> throw new UnsupportedOperationException("Unsupported table handle " + handle.getClass());
+            default -> throw new UnsupportedOperationException("Unsupported table handle " + handle.getClass() + " with split " + split.getClass());
         };
     }
 }
--- a/plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
+++ b/plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
@@ -15,6 +15,11 @@ package io.trino.plugin.lakehouse;
 
 import org.junit.jupiter.api.Test;
 
+import static io.trino.plugin.deltalake.DeltaLakeTableType.DATA;
+import static io.trino.plugin.deltalake.DeltaLakeTableType.HISTORY;
+import static io.trino.plugin.deltalake.DeltaLakeTableType.PARTITIONS;
+import static io.trino.plugin.deltalake.DeltaLakeTableType.PROPERTIES;
+import static io.trino.plugin.deltalake.DeltaLakeTableType.TRANSACTIONS;
 import static io.trino.plugin.lakehouse.TableType.DELTA;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
@@ -58,5 +63,28 @@ public class TestLakehouseDeltaConnectorSmokeTest
                    location = \\E's3://test-bucket-.*/tpch/region.*'\\Q,
                    type = 'DELTA'
                 )\\E""");
+    }
+
+    @Test
+    void testSelectMetadataTable()
+    {
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$history\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$transactions\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$properties\"")).matches("VALUES (CAST(3 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$partitions\"")).matches("VALUES (CAST(0 AS BIGINT))");
+
+        // This test should get updated if a new system table is added
+        assertThat(io.trino.plugin.deltalake.DeltaLakeTableType.values())
+                .containsExactly(
+                        DATA,
+                        HISTORY,
+                        TRANSACTIONS,
+                        PROPERTIES,
+                        PARTITIONS);
+
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$files\""))
+                .failure().hasMessageMatching(".* Table .* does not exist");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$timeline\""))
+                .failure().hasMessageMatching(".* Table .* does not exist");
     }
 }
--- a/plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseIcebergConnectorSmokeTest.java
+++ b/plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseIcebergConnectorSmokeTest.java
@@ -15,6 +15,19 @@ package io.trino.plugin.lakehouse;
 
 import org.junit.jupiter.api.Test;
 
+import static io.trino.plugin.iceberg.TableType.ALL_ENTRIES;
+import static io.trino.plugin.iceberg.TableType.ALL_MANIFESTS;
+import static io.trino.plugin.iceberg.TableType.DATA;
+import static io.trino.plugin.iceberg.TableType.ENTRIES;
+import static io.trino.plugin.iceberg.TableType.FILES;
+import static io.trino.plugin.iceberg.TableType.HISTORY;
+import static io.trino.plugin.iceberg.TableType.MANIFESTS;
+import static io.trino.plugin.iceberg.TableType.MATERIALIZED_VIEW_STORAGE;
+import static io.trino.plugin.iceberg.TableType.METADATA_LOG_ENTRIES;
+import static io.trino.plugin.iceberg.TableType.PARTITIONS;
+import static io.trino.plugin.iceberg.TableType.PROPERTIES;
+import static io.trino.plugin.iceberg.TableType.REFS;
+import static io.trino.plugin.iceberg.TableType.SNAPSHOTS;
 import static io.trino.plugin.lakehouse.TableType.ICEBERG;
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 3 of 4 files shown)


plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
```diff
@@ -59,4 +59,24 @@ public void testShowCreateTable()
                    type = 'DELTA'
                 )\\E""");
     }
+
+    @Test
+    public void testSelectDeltaMetaTables()
```
ebyhr: https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
* Test methods should be defined as package-private.

This test can be renamed to `testSelectMetadataTable`.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
```diff
@@ -59,4 +59,24 @@ public void testShowCreateTable()
...
+    {
+        computeActual(
+                """
+                CREATE TABLE delta_table
```
ebyhr: Please avoid a static table name. Use `TestTable` with try-with-resources instead. Or please consider using the existing table.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
```diff
@@ -59,4 +59,24 @@ public void testShowCreateTable()
...
+                WITH (
+                    type = 'DELTA'
+                )
+                AS SELECT * FROM tpch.tiny.region
+                """);
+
+        assertQuery("SELECT count(*) FROM \"delta_table$history\"", "VALUES 1");
```
ebyhr: `assertQuery` method isn't recommend. Please use `assertThat(query(...))` instead.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
```diff
@@ -59,4 +59,24 @@ public void testShowCreateTable()
                    type = 'DELTA'
                 )\\E""");
     }
+
+    @Test
+    public void testSelectDeltaMetaTables()
+    {
+        computeActual(
```
ebyhr: We usually use `assertUpdate` for DDL.

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
```diff
@@ -53,17 +55,24 @@ public LakehousePageSourceProviderFactory(
...
         return switch (handle) {
             case HiveTableHandle _ -> hivePageSourceProvider;
             case IcebergTableHandle _ -> icebergPageSourceProviderFactory.createPageSourceProvider();
             case DeltaLakeTableHandle _ -> deltaLakePageSourceProvider;
             case HudiTableHandle _ -> hudiPageSourceProvider;
-            default -> throw new UnsupportedOperationException("Unsupported table handle " + handle.getClass());
+            default -> {
+                // For system tables, the handle will be an instance of io.trino.connector.system.SystemTableHandle
+                // Some Iceberg system tables ($files and $history) use special splits (FilesTableSplit)
+                if (split instanceof FilesTableSplit) {
```
chenjian2664: What's the `handle` in this case? it should be the `FilesTable` right?

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
codluca: "Test classes should be defined as package-private and final."
"Test methods should be defined as package-private."
Should I change the class to be package private and final, and the other methods to be package private?
Or just the new methods?

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
codluca: No, it's of io.trino.connector.system.SystemTableHandle type.
And the split is of io.trino.plugin.iceberg.system.files.FilesTableSplit type.

io.trino.connector.system.SystemTableHandle is from trino-main module, which is only a test dependency. That's why I didn't add a new switch case for SystemTableHandle.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
ebyhr: You can change existing visibility, but I recommend handling in a separate PR.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
```diff
@@ -59,4 +59,18 @@ public void testShowCreateTable()
...
+    {
+        assertThat((Long) computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$history\"")).isEqualTo(1L);
+        assertThat((Long) computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$transactions\"")).isEqualTo(1L);
+        assertThat((Long) computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$properties\"")).isEqualTo(3L);
+        assertThat((Long) computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$partitions\"")).isEqualTo(0L);
+
+        assertThatThrownBy(() -> computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$files\""))
+                .hasMessageMatching(".* Table .* does not exist");
+        assertThatThrownBy(() -> computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$timeline\""))
+                .hasMessageMatching(".* Table .* does not exist");
```
ebyhr: Please use `assertThat(query(...)).failure()` so we can ensure that the exception class is `TrinoException`.

codluca: I'm not able to get the assertion to work. I can't find an example either.
TrinoException is wrapped in a QueryFailedException, which has a cause of FailureException.
The TrinoException is stored as FailureInfo in FailureException.

```
        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .failure().isInstanceOf(TrinoException.class).hasMessage(".* Table .* does not exist");
- fails with ->
java.lang.AssertionError: 
Expecting actual throwable to be an instance of:
  io.trino.spi.TrinoException
but was:
  io.trino.testing.QueryFailedException: line 1:22: Table 'lakehouse.tpch."region$history"' does not exist
```
```
        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .failure().hasCauseInstanceOf(TrinoException.class).hasMessage(".* Table .* does not exist");
- fails with ->
java.lang.AssertionError: 
Expecting a throwable with cause being an instance of:
  io.trino.spi.TrinoException
but was an instance of:
  io.trino.client.FailureException
Throwable that failed the check:
io.trino.testing.QueryFailedException: line 1:22: Table 'lakehouse.tpch."region$history"' does not exist
```
```
        assertThatThrownBy(() -> computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .isInstanceOf(TrinoException.class).hasMessageMatching(".* Table .* does not exist";
- fails with ->
-

Comment (codluca) [reply plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java]: I'm not able to get the assertion to work. I can't find an example either.
TrinoException is wrapped in a QueryFailedException, which has a cause of FailureException.
The TrinoException is stored as FailureInfo in FailureException.

```
        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .failure().isInstanceOf(TrinoException.class).hasMessage(".* Table .* does not exist");
- fails with ->
java.lang.AssertionError: 
Expecting actual throwable to be an instance of:
  io.trino.spi.TrinoException
but was:
  io.trino.testing.QueryFailedException: line 1:22: Table 'lakehouse.tpch."region$history"' does not exist
```
```
        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .failure().hasCauseInstanceOf(TrinoException.class).hasMessage(".* Table .* does not exist");
- fails with ->
java.lang.AssertionError: 
Expecting a throwable with cause being an instance of:
  io.trino.spi.TrinoException
but was an instance of:
  io.trino.client.FailureException
Throwable that failed the check:
io.trino.testing.QueryFailedException: line 1:22: Table 'lakehouse.tpch."region$history"' does not exist

        assertThatThrownBy(() -> computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .isInstanceOf(TrinoException.class).hasMessageMatching(".* Table .* does not exist";
```
       assertThatThrownBy(() -> computeScalar("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .isInstanceOf(TrinoException.class).hasMessageMatching(".* Table .* does not exist";
- fails like the first example above
```

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
codluca: Ok, fixed

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
codluca: Used existing table

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
codluca: Changed to use existing table

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
codluca: Changed to using assertThat(query(...))

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
chenjian2664: Thanks.
In this case we could add a case branch instead of putting it in the default, and raise nice error message to mention which kind of split we are not support yet with the SystemTableHandle

codluca: io.trino.connector.system.SystemTableHandle is from trino-main module, which is only a test dependency.
What should be the new scope of the trino-main module?

chenjian2664: Make that scope to compile should be fine in this case

ebyhr: Disagree. Please don't include trino-main as a dependency. The module isn't supposed to be depended on by connectors.

electrum: It wouldn't work anyway, as the class is in a different class loader.

chenjian2664: Thanks @ebyhr @electrum , just got to know

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseDeltaConnectorSmokeTest.java
chenjian2664: the `failure` already check the exception is or caused by the `TrinoException`, just remove `isInstanceOf` from your above example

codluca: The code already uses the recommendations:
```
    @Test
    void testSelectMetadataTable()
    {
        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$history\""))
                .failure().hasMessageMatching(".* Table .* does not exist");
        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$files\""))
                .failure().hasMessageMatching(".* Table .* does not exist");
    }
```

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseIcebergConnectorSmokeTest.java
```diff
@@ -44,4 +44,23 @@ public void testShowCreateTable()
...
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$snapshots\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$all_manifests\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$manifests\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$partitions\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$files\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$all_entries\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$entries\"")).matches("VALUES (CAST(1 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$properties\"")).matches("VALUES (CAST(6 AS BIGINT))");
+        assertThat(query("SELECT count(*) FROM lakehouse.tpch.\"region$refs\"")).matches("VALUES (CAST(1 AS BIGINT))");
```
ebyhr, line 74: We'll forget to update this test when we add a new metadata table. Could you add another assertion with `TableType`? 

https://github.com/trinodb/trino/blob/3545c68bc7ce0ef805a68a09efc3e94865ca9e94/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/glue/TestIcebergGlueCatalogAccessOperations.java#L501-L503

codluca: Done

Issue #23795, opened 2024-10-15
Cast operator `::`


findepi:
Add support for `x::type` cast operator as an alternative syntax for `CAST(x AS type)`.

- this is pure awesomeness from SQL readability perspective
- at this point it's very unlikely that future SQL spec will have something conflicting, given that all PostgreSQL derivates and e.g. Snowflake, DuckDB support  `x::type` syntax

vVv-AA:
Hi @findepi 

Does the implementation require anything other than adding the operator to grammar, and handling the change in AstBuilder? (Excluding tests and documentation)

findepi:
@vVv-AA implementation-wise I think it's exactly as you said, but @martint may know better.

martint:
The SQL spec reserves double colon for static method invocations, so that's incompatible with this proposed cast syntax.

```
<static method invocation> ::= <path-resolved user-defined type name> <double colon> <method name> [ <SQL argument list> ]
```

vVv-AA:
@martint If I understood correctly, this would make the grammar ambiguous? 


Since many dbs are not following the spec and having :: do casting instead, would it somehow be possible to let push the behavior down to the DB in a generic way.

PR #26841, opened 2025-10-04
Add support for direct method invocation

ebyhr:
## Description

This is a proposal to add support for chained function calls, similar to [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-reference#chained-function-calls), [DuckDB](https://duckdb.org/docs/stable/sql/functions/overview.html#function-chaining-via-the-dot-operator). 
It improves the readability of deeply nested expressions. Also, it eliminates the need to move the cursor backward when writing functions.

```sql
SELECT
  REPLACE(
    REPLACE(
      REPLACE(
        REPLACE(
          REPLACE('one two three four five', 'one', '1'),
          'two', '2'),
        'three', '3'),
      'four', '4'),
    'five', '5');
```
↓
```sql
SELECT
  ('one two three four five')
  .REPLACE('one', '1')
  .REPLACE('two', '2')
  .REPLACE('three', '3')
  .REPLACE('four', '4')
  .REPLACE('five', '5');
```

The corresponding SQL syntax:

**6.17 \<method invocation\>**
```
<method invocation> ::=
    <direct invocation> 
  | <generalized invocation> 

<direct invocation> ::=
  <value expression primary>  <period>  <method name>  [ <SQL argument list>  ]

<generalized invocation> ::=
  <left paren>  <value expression primary>  AS <data type>  <right paren> 
      <period>  <method name>  [ <SQL argument list>  ]

<method selection> ::=
  <routine invocation> 

<constructor method selection> ::=
  <routine invocation> 
```

**6.18 \<static method invocation\>**
```
<static method invocation> ::=
  <path-resolved user-defined type name>  <double colon>  <method name> 
      [ <SQL argument list>  ]

<static method selection> ::=
  <routine invocation> 
```

## Release notes

```markdown
## General
* Add support for chained function calls. ({issue}`26841`)
```

### DIFF
--- a/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
+++ b/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
@@ -601,10 +601,8 @@ primaryExpression
         (ON OVERFLOW listAggOverflowBehavior)? ')'
         (WITHIN GROUP '(' orderBy ')')
         filter? over?                                                                     #listagg
-    | processingMode? qualifiedName '(' (label=identifier '.')? ASTERISK ')'
-        filter? over?                                                                     #functionCall
-    | processingMode? qualifiedName '(' (setQuantifier? expression (',' expression)*)?
-        orderBy? ')' filter? (nullTreatment? over)?                                       #functionCall
+    | function                                                                            #functions
+    | left=primaryExpression '.' right=function                                           #directInvocation
     | identifier over                                                                     #measure
     | identifier '->' expression                                                          #lambda
     | '(' (identifier (',' identifier)*)? ')' '->' expression                             #lambda
@@ -667,6 +665,13 @@ primaryExpression
         )?
         (RETURNING type (FORMAT jsonRepresentation)?)?
      ')'                                                                                  #jsonArray
+    ;
+
+function
+    : processingMode? qualifiedName '(' (label=identifier '.')? ASTERISK ')'
+        filter? over?                                                                     #functionCall
+    | processingMode? qualifiedName '(' (setQuantifier? expression (',' expression)*)?
+        orderBy? ')' filter? (nullTreatment? over)?                                       #functionCall
     ;
 
 literal
--- a/core/trino-main/src/test/java/io/trino/sql/query/TestDirectInvocation.java
+++ b/core/trino-main/src/test/java/io/trino/sql/query/TestDirectInvocation.java
@@ -1,0 +1,67 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.sql.query;
+
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.TestInstance;
+import org.junit.jupiter.api.parallel.Execution;
+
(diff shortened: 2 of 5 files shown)


kasiafi:
For the expression `foo.bar(baz)` how do we know if it's:
- function `foo.bar` applied to argument `baz`
- function `bar` applied to arguments `foo`, `baz`?

For an expression like `a.b.c.d()`, how do we know where the argument ends and the function name starts?

findepi:
> This is a proposal to add support for chained function calls, similar to [BigQuery](https://cloud.google.com/bigquery/docs/reference/standard-sql/functions-reference#chained-function-calls), [DuckDB](https://duckdb.org/docs/stable/sql/functions/overview.html#function-chaining-via-the-dot-operator).

Are these the only ones which support this syntax sugar?

> It improves the readability of deeply nested expressions.

So do "lateral column aliases".
Does the proposed syntax solve a problem that is hard to solve without it?

Coming from object-oriented languages, I very much like the idea of SQL values having methods, but I am not convinced `val.f(...)` method would always be desired to mean `f(val, ...)`. 
The `replace` case looks nice (but is also likely a logic error -- one usually wants to do a multi-replace in one shot so that the replacement values from the first pattern as not considered when looking for the second pattern).
Other case where one could want to chain calls is `date_add` and/or other date/time functions, but there the value being operated on doesn't come as the first one.

ebyhr:
> For the expression `foo.bar(baz)`

A function require `()`, so the expression is always `foo.bar` function with `baz` parameter in my opinion. 
"function `bar` applied to arguments `foo`, `baz`" should be `foo().bar(baz)`. Am I missing something?

> Are these the only ones which support this syntax sugar?

Yes, as far as I confirmed. 

> So do "lateral column aliases".
> Does the proposed syntax solve a problem that is hard to solve without it?

I don't think LCA and chained function calls actually solve any "problems". Those syntaxes exists to improve user experience. 
For non-first values, we could consider introducing a special symbol or keyword if truely needed.

kasiafi:
>For the expression foo.bar(baz)

Function names are `QualifiedNames`. Column names are also `QualifiedNames`. With the proposed chained calling syntax, we can chain the column name (if the function operates on a column) with the function name, and create a multi-part qualified name, which is ambiguous to the parser.

`foo.bar(baz)` can be function `bar(baz)` applied to column `foo`. But it can also be function `foo.bar(baz)` in the traditional calling convention. In the latter case, `foo` is the schema name.

In `a.b.c.d()` there are even more possibilities. 
Column `a`, function `b.c.d()`
Column `a.b`, function `c.d()`
Column `a.b.c`, function `d()`

In fact, field names can contain even more than 3 parts.

wendigo:
But if we use other syntax (ie :: instead of a dot) it shouldn't be ambiguous anymore, right?

kasiafi:
>But if we use other syntax (ie :: instead of a dot) it shouldn't be ambiguous anymore, right?

Yes. 
Regarding `::` however, I recall that it's reserved for function calls on types https://github.com/trinodb/trino/issues/23795#issuecomment-2423326758, so this would be probably also ambiguous.

ebyhr:
Can we resolve the ambiguity by requiring `()` when we want to use a column name at the beginning of the chain?

That's the BigQuery's syntax:
```sql
SELECT (name).lower() FROM tpch.region;
```

The below query throws "Function not found: name.lower":
```sql
SELECT name.lower() FROM tpch.region;
```

kasiafi:
>Can we resolve the ambiguity by requiring () when we want to use a column name at the beginning of the chain?

We could require `()`. I think that after the proposed grammar change it's already required. Not sure how intuitive using parenthesis in this context is to the users. Column names / qualified names are the only primary expressions that require parenthesis for disambiguation. There's a chance of getting unexpected output when a user misses parenthesis and the column name overlaps with function prefix. But this does not seem a likely scenario.

martint:
There's a whole section in the SQL spec for instance method invocations. We don't need to reinvent any wheels.

However, as @findepi pointed out above, we can't blindly interpret every function as a method invocation.

ebyhr:
> There's a whole section in the SQL spec for instance method invocations. 

@martint Do you mean `6.17 <method invocation>` P.292 in ISO/IEC 9075-2? 

> we can't blindly interpret every function as a method invocation.

What is your recommendation?

PR #26883, opened 2025-10-08
Cleanup OutputBufferMemoryManager and peak memory calculations

pettyjamesm:
## Description
**OutputBufferMemoryManager**
- Switch the relatively more expensive `AtomicLong#updateAndGet`  for `AtomicLong#accumulateAndGet` operation
- Check `currentBufferedBytes > bufferedBytes.get()` before calling `bufferedBytes.accumulateAndGet(currentBufferedBytes, Math::max)` since a volatile read is cheaper than a volatile write and updating the maximum is not the common case.
- Refactor `getUtilization` logic to avoid unnecessary volatile reads
- Make the class final

**Peak Memory Calculations**
- Updates various places where peak memory is calculated using `AtomicLong` to check whether the value is greater than peak before calling into `AtomicLong::accumulateAndGet` so to avoid unnecessary contention on volatile updates

## Release notes

## Summary by Sourcery

Improve performance and simplify synchronization in OutputBufferMemoryManager by replacing atomic operations, reducing volatile writes, refactoring utilization tracking, and marking the class as final

Bug Fixes:
- Introduce an IllegalStateException on negative buffer adjustments instead of using updateAndGet’s argument check

Enhancements:
- Replace AtomicLong#updateAndGet with AtomicLong#addAndGet and explicit negative-value check to reduce overhead
- Guard peakMemoryUsage updates behind a conditional volatile read to avoid unnecessary writes
- Refactor getUtilization and recordBufferUtilization to accept the currentBufferedBytes parameter and eliminate redundant volatile reads
- Mark OutputBufferMemoryManager as a final class to prevent subclassing

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/TableWriterOperator.java
+++ b/core/trino-main/src/main/java/io/trino/operator/TableWriterOperator.java
@@ -414,7 +414,9 @@ public class TableWriterOperator
     {
         long pageSinkMemoryUsage = pageSink.getMemoryUsage();
         pageSinkMemoryContext.setBytes(pageSinkMemoryUsage);
-        pageSinkPeakMemoryUsage.accumulateAndGet(pageSinkMemoryUsage, Math::max);
+        if (pageSinkMemoryUsage > pageSinkPeakMemoryUsage.get()) {
+            pageSinkPeakMemoryUsage.accumulateAndGet(pageSinkMemoryUsage, Math::max);
+        }
     }
 
     @VisibleForTesting


core/trino-main/src/main/java/io/trino/execution/buffer/OutputBufferMemoryManager.java
```diff
@@ -121,23 +121,26 @@ public void updateMemoryUsage(long bytesAdded)
                     this.bufferBlockedFuture = null;
                 }
             }
-            recordBufferUtilization();
+            recordBufferUtilization(currentBufferedBytes);
+        }
+        // Reduce contention by reading first and only updating if the new value might become the maximum (uncommon)
+        if (currentBufferedBytes > peakMemoryUsage.get()) {
+            peakMemoryUsage.accumulateAndGet(currentBufferedBytes, Math::max);
```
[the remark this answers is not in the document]
pettyjamesm: This would be true for some operations, but `Math::max` is structurally safe for this pattern of check before update.

core/trino-main/src/main/java/io/trino/execution/buffer/OutputBufferMemoryManager.java
```diff
@@ -99,9 +99,11 @@ public void updateMemoryUsage(long bytesAdded)
...
             }
 
-            currentBufferedBytes = bufferedBytes.updateAndGet(bytes -> {
-                long result = bytes + bytesAdded;
-                checkArgument(result >= 0, "bufferedBytes (%s) plus delta (%s) would be negative", bytes, bytesAdded);
+            currentBufferedBytes = bufferedBytes.accumulateAndGet(bytesAdded, (bufferedBytes, delta) -> {
+                long result = bufferedBytes + delta;
+                if (result < 0) {
+                    throw new IllegalArgumentException("bufferedBytes (%s) plus delta (%s) is negative".formatted(bufferedBytes, delta));
+                }
```
raunaqmorarka: Why did we need to change the `checkArgument` to manually throwing IllegalArgumentException ?

pettyjamesm: Minor cleanup. This code path is high traffic and therefore high contention (updates occur on each worker thread adding data and consumer pulling data out) so any extra time spent here adds up to extra contention quite quickly.

raunaqmorarka: I don't follow why this version is better. checkArgument is
```
public static void checkArgument(
      boolean expression, String errorMessageTemplate, long p1, long p2) {
    if (!expression) {
      throw new IllegalArgumentException(lenientFormat(errorMessageTemplate, p1, p2));
    }
  }
```
Is JIT unable to inline this or something else about this is worse ?

pettyjamesm: It should, but relying on that is more fragile than doing the check directly and doesn’t add much visual noise so being direct seems like a worthwhile trade off here.

raunaqmorarka: We do things like checkArgument, verify, checkState etc. in many perf sensitive parts of the code base on the assumption that JIT is smart enough to optimize these cases.
If we're inlining by hand here, then it potentially opens the door to doing the same in many other parts of the code base. I don't think we'd want to do that unless there is a reason to believe that the JIT is not good enough here. If this is a special case, then it should at least be justified by some JMH benchmark.
cc: @martint @findepi

pettyjamesm: Sure, fair enough- removed that portion of the first commit.

PR #26884, opened 2025-10-08
Disallow dropping sort column explicitly in Iceberg

ebyhr:
## Release notes

## Summary by Sourcery

Disallow explicitly dropping sort fields in Iceberg tables and adjust the test to verify the updated exception message.

Enhancements:
- Prevent dropping sort columns by throwing a NOT_SUPPORTED exception when attempting to drop a sort field in IcebergMetadata

Tests:
- Update test assertion to expect the new error message 'Cannot drop sort field: comment' when dropping a sort column


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -1631,7 +1631,7 @@ public void testDroppingSortColumn()
                 "WITH (sorted_by = ARRAY['comment']) AS SELECT * FROM nation WITH NO DATA")) {
             assertUpdate(withSmallRowGroups, "INSERT INTO " + table.getName() + " SELECT * FROM nation", 25);
             assertThat(query("ALTER TABLE " + table.getName() + " DROP COLUMN comment"))
-                    .failure().hasMessageContaining("Cannot find source column for sort field");
+                    .failure().hasMessageContaining("Cannot drop sort field: comment");
```
findepi, line 1634: ❤️

PR #26891, opened 2025-10-09
Cleanup JDK provider in tests

wendigo:
We only use Temurin for both testing and while producing docker images so there is no point in keeping more complicated implementation if it's not used.

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Switch JDK provisioning to use Temurin exclusively and remove legacy JDK provider logic

New Features:
- Add TemurinJdkProvider for downloading and providing Temurin JDKs in test environments

Enhancements:
- Simplify Docker build script to read .temurin-release and construct Temurin download URLs dynamically
- Refactor test launcher to replace --trino-jdk-version and distribution paths with --trino-jdk-release and use TemurinJdkProvider

Tests:
- Update JDK provider naming test to expect temurin as the canonical provider name

Chores:
- Remove legacy core/jdk distribution directories and obsolete JDK provider classes


wendigo:
This makes it easier to forward test new JDK version (and upgrade these too)

PR #26900, opened 2025-10-09
Add per catalog metastore metrics to QueryStats

lukasz-stec:
## Description
Slow metastore can be a root cause of slow analysis or planning. This adds explicit metrics to the `QueryStats` with remote metastore call stats made for a given query.

This is what this looks like in the query.json:
```
{
  "queryStats": {
    ...
    "catalogMetadataMetrics": {
      "hive": {
        "metastore.getDatabase.time.distribution": {
          "@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
          "total": 1,
          "min": 281684250,
          "max": 281684250,
          "p01": 281684250,
          "p05": 281684250,
          "p10": 281684250,
          "p25": 281684250,
          "p50": 281684250,
          "p75": 281684250,
          "p90": 281684250,
          "p95": 281684250,
          "p99": 281684250
        },
        "metastore.getTable.time.distribution": {
          "@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
          "total": 1,
          "min": 22330833,
          "max": 22330833,
          "p01": 22330833,
          "p05": 22330833,
          "p10": 22330833,
          "p25": 22330833,
          "p50": 22330833,
          "p75": 22330833,
          "p90": 22330833,
          "p95": 22330833,
          "p99": 22330833
        },
        "metastore.all.time.distribution": {
          "@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
          "total": 2,
          "min": 22330833,
          "max": 281684250,
          "p01": 22330833,
          "p05": 22330833,
          "p10": 22330833,
          "p25": 22330833,
          "p50": 281684250,
          "p75": 281684250,
          "p90": 281684250,
          "p95": 281684250,
          "p99": 281684250
        },
        "metastore.getDatabase.time.total": {
          "@class": "io.trino.plugin.base.metrics.LongCount",
          "total": 281684250
        },
        "metastore.all.time.total": {
          "@class": "io.trino.plugin.base.metrics.LongCount",
          "total": 304015083
        },
        "metastore.getTable.time.total": {
          "@class": "io.trino.plugin.base.metrics.LongCount",
          "total": 22330833
        }
      }
    },
    ...
  }
}
```

## Release notes

( X) Release notes are required, with the following suggested text:

```markdown
## SPI
* Replace `double getPercentile(double percentile)` with `double[] getPercentiles(double... percentiles)` in `io.trino.spi.metrics.Distribution`

## Event listener
* Add `catalogMetadataMetrics` to `io.trino.spi.eventlistener.QueryStatistics` for tracking time taken by the query during metadata operations.
```


lukasz-stec:
There are related CI failures. Moving to draft until I fix it

lukasz-stec:
@findepi @raunaqmorarka This is ready for review. There is one CI failure, but it is unrelated.

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -386,6 +389,30 @@ static QueryStateMachine beginWithTicker(
         return queryStateMachine;
     }
```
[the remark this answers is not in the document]
lukasz-stec: No exception should be thrown during metrics collection. If it is, it is a bug and should be brought forward.

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,499 @@
...
+{
+    private final HiveMetastore delegate;
+    private final MetastoreApiCallStats allApiCallsStats = new MetastoreApiCallStats();
+    private final Map<String, MetastoreApiCallStats> apiCallStats = new ConcurrentHashMap<>();
+    private final Ticker ticker = Ticker.systemTicker();
+
+    public MeasuredHiveMetastore(HiveMetastore delegate)
+    {
+        this.delegate = requireNonNull(delegate, "delegate is null");
+    }
```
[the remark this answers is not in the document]
lukasz-stec: not needed for now

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9214,6 +9222,52 @@ public void testCreateTableWithDataLocationButObjectStoreLayoutDisabled()
                 "Data location can only be set when object store layout is enabled");
     }
 
+    @Test
+    public void testCatalogMetadataMetrics()
```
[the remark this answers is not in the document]
lukasz-stec: Only iceberg and hive expose the metrics, and we don't have generic tests that use both. Adding it is overkill.

findepi:
[test (plugin/trino-lakehouse)](https://github.com/trinodb/trino/actions/runs/18406467859/job/52447532072?pr=26900#logs) this job hanged it contains a couple errors like this

```
Treating the bytes as signed or unsigned?

It's worth a code comment explaining that Trino string codepoint-based collation is equivalent to sorting (unsigned) bytes ... in UTF-8 encoding.
When we sent `COLLATE BINARY` to oracle, are we assuming the data is compared byte-wise on Utf-8 encoding, or can it be something else?
```

and then
```
2025-10-10T06:42:14.693-0600	WARN	TestHangMonitor	io.trino.testing.services.junit.LogTestDurationListener	No test started or completed in 8.00m. Running tests:
	TestLakehouseConnectorTest running for 9.33m
	JUnit Jupiter running for 9.41m
	TestLakehouseDeltaConnectorSmokeTest running for 9.41m
	TestLakehouseHiveConnectorSmokeTest running for 9.26m
	TestLakehouseIcebergConnectorSmokeTest running for 9.38m
	TestLakehouseFileConnectorSmokeTest running for 9.41m.
```

are they related?

lukasz-stec:
> [test (plugin/trino-lakehouse)](https://github.com/trinodb/trino/actions/runs/18406467859/job/52447532072?pr=26900#logs) this job hanged it contains a couple errors like this
...
> 
> are they related?

Yes, the impl for the `io.trino.plugin.lakehouse.LakehouseMetadata#getMetrics` was broken. I fixed it and added proper tests

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -386,6 +389,30 @@ static QueryStateMachine beginWithTicker(
...
 
+    private void collectCatalogMetadataMetrics()
+    {
+        // collect the metrics only once. This avoid issue with transaction being removed
```
findepi:
```suggestion
        // collect the metrics only once. This avoids issue with transaction being removed
```

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -386,6 +389,30 @@ static QueryStateMachine beginWithTicker(
...
+        // after the check but before the metrics collection
+        if (catalogMetadataMetricsCollected.compareAndSet(false, true)) {
+            if (session.getTransactionId().filter(transactionManager::transactionExists).isEmpty()) {
```
findepi, line 395: In theory this is racy. A check in TM for a transaction on this line does not guarantee the TM knows about the transaction during `listActiveCatalogs`. Perhaps, the current query is timed out and cleanup happens asyncly.

Perhaps instead we could just validate transaction only once?

```java
List<CatalogInfo> activeCatalogs;
try {
    activeCatalogs = metadata.listActiveCatalogs(session);
} catch (NotInTransactionException e) {
    // explanation
    return;
}
```

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -1229,6 +1257,7 @@ private boolean transitionToFailed(Throwable throwable, boolean log)
     {
         queryStateTimer.endQuery();
 
+        collectCatalogMetadataMetrics();
```
findepi: esp in case of query failure we could want to be defensive, and suppress exceptions coming from here.

the easiest way would be to put the metrics collection code inside `cleanupQuery`.
are there any ordering requirements between `metadata.cleanupQuery` and `metadata.getMetrics`?

core/trino-main/src/test/java/io/trino/execution/TestQueryStateMachine.java
```diff
@@ -842,10 +826,23 @@ public QueryStateMachineBuilder withAddPreparedStatements(Map<String, String> pr
...
-            if (metadata == null) {
-                metadata = createTestMetadataManager();
-            }
             TransactionManager transactionManager = createTestTransactionManager();
+            Metadata metadata = TestMetadataManager.builder()
+                    .withTransactionManager(transactionManager)
+                    .build();
+            if (beforeQueryCleanup.isPresent()) {
+                Runnable beforeQueryCleanupAction = beforeQueryCleanup.get();
+                metadata = new TracingMetadata(noopTracer(), metadata)
```
findepi, line 836: `// Using TracingMetadata in lieu of "Forwarding Metadata" which currently does not exist`

testing/trino-tests/src/test/java/io/trino/connector/system/metadata/TestSystemMetadataConnector.java
```diff
@@ -416,7 +416,8 @@ private void assertMetadataCalls(@Language("SQL") String actualSql, @Language("S
 
         actualMetadataCallsCount = actualMetadataCallsCount.stream()
                 // Every query involves beginQuery and cleanupQuery, so ignore them.
```
findepi, line 418: and getMetrics

(`Add Metadata.getMetrics` commit)

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,499 @@
...
+        return new MeasuredMetastoreFactory(metastoreFactory);
+    }
+
+    @Override
+    public Metrics getMetrics()
+    {
+        ImmutableMap.Builder<String, Metric<?>> metrics = ImmutableMap.builder();
+        allApiCallsStats.put(metrics, "metastore.all");
+        for (Map.Entry<String, MetastoreApiCallStats> callStats : apiCallStats.entrySet()) {
+            callStats.getValue().put(metrics, "metastore.%s".formatted(callStats.getKey()));
```
findepi: `put` -> `storeTo`

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,499 @@
...
+            totalTimeNanos += nanos;
+        }
+
+        public synchronized void addFailure()
+        {
+            totalFailures++;
+        }
+
+        public synchronized void put(ImmutableMap.Builder<String, Metric<?>> metrics, String prefix)
+        {
```
findepi, line 467: nit

```java
 DistributionSnapshot distributionSnapshot;
            long totalTimeNanos;
            long totalFailures;
            synchronized (this) {
                // DistributionSnapshot does not retain reference to the histogram 
                distributionSnapshot = new DistributionSnapshot(new TDigestHistogram(timeNanosDistribution));
                totalTimeNanos = this.totalTimeNanos;
                totalFailures = this.totalFailures;
            }
```

perhaps this is not needed because collecting metrics will generally not coincide with mestore comments
feel free to dismiss, but then add explanatory comment

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,499 @@
...
+            metrics.put(prefix + ".time.distribution", new DistributionSnapshot(new TDigestHistogram(timeNanosDistribution)));
+            metrics.put(prefix + ".time.total", new LongCount(totalTimeNanos));
+
+            if (totalFailures > 0) {
+                metrics.put(prefix + ".failures", new LongCount(totalFailures));
```
findepi, line 474: i assume this is to save space? OTOH it's nice if easier to consume metric that's always present vs one that shows up rarely.
worth code comment

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,499 @@
...
+            }
+        }
+    }
+
+    public static class MeasuredMetastoreFactory
+            implements HiveMetastoreFactory
```
findepi, line 480: This class should probably implement all optional methods.  Add a test.

E.g. hasBuiltInCaching is not implemented.

Add similar test for MeasuredHiveMetastore completeness.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
```diff
@@ -9398,6 +9403,39 @@ public void testFlushMetadataDisabled()
...
+    @Test
+    public void testCatalogMetadataMetricsWithOptimizerTimeoutExceeded()
+    {
+        String query = "SELECT count(*) FROM region r, nation n WHERE r.regionkey = n.regionkey";
+        try {
+            Session smallOptimizerTimeout = TestingSession.testSessionBuilder(getSession())
+                    .setSystemProperty(ITERATIVE_OPTIMIZER_TIMEOUT, succinctDuration(1, MILLISECONDS).toString())
+                    .build();
+            MaterializedResultWithPlan result = getQueryRunner().executeWithPlan(smallOptimizerTimeout, query);
+            fail(format("Expected query to fail: %s [QueryId: %s]", query, result.queryId()));
```
findepi: Is the query really guaranteed to fail?

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
lukasz-stec: That won't work, unfortunately, as we need the transaction also to access `ConnectorMetadata` that keeps the metrics.
I moved the invocation of the `collectCatalogMetadataMetrics()` to the places where only a single thread can reach before the transaction is committed or aborted.

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
lukasz-stec: There should be no exception thrown here except for a bug. IMO, it is better to fail the query in that case, but since this is connector-specific, I understand the urge to be defensive here. Will add try catch,
There is no ordering requirement with `cleanupQuery` directly but `cleanupQuery` can be executed by multiple threads, so it is not a good place for the `collectCatalogMetadataMetrics` unless I move it after the
```
QueryState oldState = queryState.trySet(FAILED);
        if (oldState.isDone()) {
            if (log) {
                QUERY_STATE_LOG.debug(throwable, "Failure after query %s finished", queryId);
            }
            return false;
        }
```

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: For machines, it does not matter much if the metric is always there. For humans, the usual reader of the query json, omitting 0 failures here and other not-used call metrics in general, reduces the cognitive load significantly.
Comment added

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: I'm sorry, I don't get it. What is this change about?

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: It definitely should. Will do

lukasz-stec: done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
lukasz-stec: No, it is not guaranteed to fail. Do you have an idea how to test the optimizer failure deterministically?
Maybe I  will drop this test and add something to the `TestQueryStateMachine`

lukasz-stec: `TestQueryStateMachine` does not work as there are no connectors there.
I updated the connector tests with the `MockConnector` that makes sure the planning fails on getTableStatistics. Let's see if it works

lukasz-stec:
@findepi @raunaqmorarka There were issues with collecting the metadata metrics concurrently with updating the final query info. The query info could be triggered asynchronously, and thus, missing the collected metrics. To fix this, I added a metadata metrics collection before every query stats collection. This also makes the metadata metrics available before the query is done, which is an additional benefit.

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
findepi: reducing synchronization scope

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
findepi: do we need this `if`?
i realized there is already a `catch (NotInTransactionExcept` below

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -386,6 +389,35 @@ static QueryStateMachine beginWithTicker(
...
+                Metrics metrics = metadata.getMetrics(session, activeCatalog.catalogName());
+                if (!metrics.getMetrics().isEmpty()) {
+                    catalogMetadataMetrics.put(activeCatalog.catalogName(), metrics);
+                }
+            }
+
+            this.catalogMetadataMetrics.set(catalogMetadataMetrics.buildOrThrow());
+        }
+        catch (NotInTransactionException e) {
+            // ignore
```
findepi: Explain why this should be ignored, ie what are the legit cases where this can happen

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
lukasz-stec: I just added the `catch NotInTransactionExcept` yesterday. It was not there before. Now, the `if` is not strictly necessary, but since this method may be called without a transaction running, and it is not exceptional flow (e.g.,`START TRANSACTION` statement does not itself have a transaction, or async `updateQueryInfo` that can happen after the query is finished), I thought I would leave it in place.
I'm fine dropping it as well.

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
lukasz-stec: comment added

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: I will leave it as is, as the `ImmutableMap.Builder.put` is pretty simple, so the synchronization scope would not be reduced much + there should not be a lot of contention here anyway, as in normal operation, the metrics updates and the metrics collections happen at different times.
I added a comment about `DistributionSnapshot`

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -1236,14 +1276,14 @@ private boolean transitionToFailed(Throwable throwable, boolean log)
         failureCause.compareAndSet(null, toFailure(throwable));
 
         cleanupQueryQuietly();
-
```
raunaqmorarka, line 1251: unrelated change ?

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -1167,6 +1206,7 @@ public boolean transitionToFinishing()
             transitionToFailed(e);
             return true;
         }
+        collectCatalogMetadataMetrics();
```
raunaqmorarka, line 1221: We make metadata calls during FINISHING state as well, so we should collect metrics at the end of that

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -147,6 +148,11 @@ Optional<TableExecuteHandle> getTableHandleForExecute(
 
     Optional<Object> getInfo(Session session, TableHandle handle);
 
+    /**
+     * Return connector-specific, metadata operations metrics for the given session.
```
raunaqmorarka, line 152: connector -> catalog

core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
```diff
@@ -252,6 +253,15 @@ public Optional<Object> getInfo(ConnectorSession session, ConnectorTableHandle t
         }
     }
 
+    @Override
+    public Metrics getMetrics(ConnectorSession session)
+    {
+        Span span = startSpan("getMetrics");
```
raunaqmorarka: Do we really want to have Span for get metrics collection call ?
If we're calling this every time QueryStats is constructed, the traces are going to show this Span a lot

core/trino-main/src/main/java/io/trino/tracing/TracingMetadata.java
```diff
@@ -284,6 +285,15 @@ public Optional<Object> getInfo(Session session, TableHandle handle)
         }
     }
 
+    @Override
+    public Metrics getMetrics(Session session, String catalogName)
```
raunaqmorarka, line 289: Same comment as TracingConnectorMetadata

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -248,6 +249,14 @@ default Optional<Object> getInfo(ConnectorSession session, ConnectorTableHandle
         return Optional.empty();
     }
 
+    /**
+     * Return connector-specific, metadata operations metrics for the given session.
```
raunaqmorarka, line 254: connector -> catalog

testing/trino-tests/src/test/java/io/trino/connector/informationschema/TestInformationSchemaConnector.java
```diff
@@ -410,6 +412,8 @@ private void assertMetadataCalls(@Language("SQL") String actualSql, @Language("S
             assertQuery(actualSql, expectedSql);
         });
 
+        // Every query involves some number of ConnectorMetadata.getMetrics calls, so ignore them.
+        actualMetadataCallsCount = Multisets.filter(actualMetadataCallsCount, e -> !Objects.equals(e, "ConnectorMetadata.getMetrics"));
```
raunaqmorarka: This is probably a sign that these calls are not worth tracking in telemetry spans

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,507 @@
...
+            timeNanosDistribution.add(nanos);
+            totalTimeNanos += nanos;
+        }
+
+        public synchronized void addFailure()
+        {
+            totalFailures++;
+        }
+
+        public synchronized void storeTo(ImmutableMap.Builder<String, Metric<?>> metrics, String prefix)
```
raunaqmorarka, line 466: storeTo -> collectMetrics

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,507 @@
...
+        {
+            // DistributionSnapshot does not retain reference to the histogram
+            metrics.put(prefix + ".time.distribution", new DistributionSnapshot(new TDigestHistogram(timeNanosDistribution)));
```
raunaqmorarka: I don't like the fact that we're recording TDigestHistograms here. We've seen before that using this on coordinator adds significantly to CPU and memory usage and also bloats queryinfo jsons. We've spent a lot of time on optimizing the existing usage of these and they're still often the topmost things in JFR profiles for coordinator.
I'm skeptical that recording a detailed distribution of time taken is even useful for us. I suggest dropping the distribution in favor of recording just total time, number of invocations and failures (e.g. io.trino.spi.eventlistener.QueryPlanOptimizerStatistics).

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,507 @@
...
+    }
+
+    @ThreadSafe
+    static class MetastoreApiCallStats
+    {
+        private final TDigest timeNanosDistribution = new TDigest();
+        private long totalTimeNanos;
+        private long totalFailures;
+
+        public synchronized void addTime(long nanos)
```
raunaqmorarka: addTime -> record

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
```diff
@@ -0,0 +1,507 @@
...
+        {
+            timeNanosDistribution.add(nanos);
+            totalTimeNanos += nanos;
+        }
+
+        public synchronized void addFailure()
```
raunaqmorarka: recordFailure

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
raunaqmorarka: Maybe keep the distribution only in `metastore.all` if you really feel convinced about the value of having it

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/CachingHiveMetastoreModule.java
```diff
@@ -65,7 +66,7 @@ public static HiveMetastoreFactory createHiveMetastore(
             SharedHiveMetastoreCache sharedHiveMetastoreCache)
     {
         // cross TX metastore cache is enabled wrapper with caching metastore
-        return sharedHiveMetastoreCache.createCachingHiveMetastoreFactory(metastoreFactory);
+        return sharedHiveMetastoreCache.createCachingHiveMetastoreFactory(new MeasuredMetastoreFactory(metastoreFactory));
```
raunaqmorarka, line 69: Will cached metastore calls show up in the metrics ?
Also, please check that the metrics tracking is working for Glue metastore as well, the way that is setup is a but different.

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -251,6 +254,26 @@ public Optional<Object> getInfo(ConnectorSession session, ConnectorTableHandle t
...
+            metrics.put("iceberg." + metricName, metric);
+        });
+        deltaMetadata.getMetrics(session).getMetrics().forEach((metricName, metric) -> {
+            metrics.put("delta." + metricName, metric);
+        });
+        hudiMetadata.getMetrics(session).getMetrics().forEach((metricName, metric) -> {
+            metrics.put("hudi." + metricName, metric);
+        });
+        return new Metrics(metrics
+                .buildOrThrow());
```
raunaqmorarka: one line

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/TrinoHiveCatalog.java
```diff
@@ -880,6 +881,12 @@ public Optional<CatalogSchemaTableName> redirectTable(ConnectorSession session,
         return Optional.empty();
     }
 
+    @Override
+    public Metrics getMetrics()
+    {
+        return metastore.getMetrics();
```
raunaqmorarka, line 887: I think we would at least want tracking in REST and Glue catalogs as well

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2328,6 +2328,12 @@ public Optional<Object> getInfo(ConnectorSession session, ConnectorTableHandle t
                 totalDeleteFiles));
     }
 
+    @Override
+    public io.trino.spi.metrics.Metrics getMetrics(ConnectorSession session)
+    {
+        return catalog.getMetrics();
```
raunaqmorarka, line 2351: This approach is only catching the metadata calls made to the metastore, which is relatively few operations in iceberg. The bulk of metadata fetching in iceberg is happening through the use of iceberg APIs in this class. You'll have to register InMemoryMetricsReporter using `org.apache.iceberg.Scan#metricsReporter` and collect those metrics like in IcebergSplitSource.
That will still not catch the metadata calls in commit operations, but I don't see a way in iceberg APIs to register metrics reporter for commit operations. Maybe something to improve in upstream iceberg.
fyi @ebyhr @findepi

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
lukasz-stec: yes, reverted

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
lukasz-stec: We remove the transaction right after this call, so there is no way to collect the metrics anymore.

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
lukasz-stec: Well, the metrics are connector-specific, but we return them for the provided catalog.

core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
lukasz-stec: There should be only a few of those calls made + since this is connector specific, we don't really know what is there. I think it is ok to keep it consistent with the rest of the class.

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: `storeTo` was suggested by @findepi. Im going to keep it to avoid infinite loop of comments here.

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: We are not recording `TDigestHistograms` here, but rather a `DistributionSnapshot` that only contains a limited list of percentiles. The percentiles could be useful to see problems with non-constant latency patterns.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/CachingHiveMetastoreModule.java
lukasz-stec: No, cached calls are not measured.
I will check the glue

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
lukasz-stec: Yes, for now, I'm only adding support to metastore metrics. This could be later extended.

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
raunaqmorarka: This is definitely better than the merging of TDigests we're doing elsewhere, which is the most expensive step, but this is still way more than we need.
DistributionSnapshot is recording 9 percentiles in addition to min, max and total, I am not convinced we've ever used that level of detail.
I suggest changing DistributionSnapshot and Distribution to leverage `io.airlift.stats.TDigest#valuesAt(double...)` to compute all the quantiles in one go, and to cut down the percentiles recorded to `10, 25, 50, 90`.

core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
lukasz-stec: btw I tested this and those spans are not part of the query trace but only as a part of a standalone `Metadata.getMetrics` trace. Seems like a bug.

Image

raunaqmorarka: I would still suggest skipping tracing for this API. It should always only collect in-memory objects and not make remote calls. There's no real value in tracing this. You can add a code comment explaining why it is not consistent with the rest of the code. We know that the biggest problem when working with traces today is the sheer amount of things in it.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/CachingHiveMetastoreModule.java
lukasz-stec: It works for

[glue_20251016_184141_00003_txnrc.json](https://github.com/user-attachments/files/22956940/glue_20251016_184141_00003_txnrc.json)

 glue
```
"activeCatalogs": [
{
"name": "hive",
"version": "95607a88f6655db4de4dbfdc0c255acf246148f0550ddf891c903b8aaea7f7b6",
"connectorName": "hive",
"properties": {
"hive.metastore": "glue",
"hive.security": "allow-all",
"hive.metastore.glue.default-warehouse-dir": "local:///glue",
"hive.max-partitions-per-scan": "1000",
"fs.hadoop.enabled": "true",
"hive.max-partitions-for-eager-load": "1000"
}
}
],
```

```
"catalogMetadataMetrics": {
"hive": {
"metastore.getPartitionNamesByFilter.time.distribution": {
"@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
"total": 1,
"min": 1600526792,
"max": 1600526792,
"p01": 1600526792,
"p05": 1600526792,
"p10": 1600526792,
"p25": 1600526792,
"p50": 1600526792,
"p75": 1600526792,
"p90": 1600526792,
"p95": 1600526792,
"p99": 1600526792
},
"metastore.getTable.time.total": {
"@class": "io.trino.plugin.base.metrics.LongCount",
"total": 210967292
},
"metastore.all.time.distribution": {
"@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
"total": 15,
"min": 169969959,
"max": 1600526792,
"p01": 169969959,
"p05": 169969959,
"p10": 176701709,
"p25": 194937458,
"p50": 206969875,
"p75": 214762042,
"p90": 362961542,
"p95": 1600526792,
"p99": 1600526792
},
"metastore.all.time.total": {
"@class": "io.trino.plugin.base.metrics.LongCount",
"total": 4566443044
},
"metastore.getTable.time.distribution": {
"@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
"total": 1,
"min": 210967292,
"max": 210967292,
"p01": 210967292,
"p05": 210967292,
"p10": 210967292,
"p25": 210967292,
"p50": 210967292,
"p75": 210967292,
"p90": 210967292,
"p95": 210967292,
"p99": 210967292
},
"metastore.getPartitionNamesByFilter.time.total": {
"@class": "io.trino.plugin.base.metrics.LongCount",
"total": 1600526792
},
"metastore.getPartitionsByNames.time.total": {
"@class": "io.trino.plugin.base.metrics.LongCount",
"total": 2754948960
},
"metastore.getPartitionsByNames.time.distribution": {
"@class": "io.trino.plugin.base.metrics.DistributionSnapshot",
"total": 13,
"min": 169969959,
"max": 362961542,
"p01": 169969959,
"p05": 169969959,
"p10": 176701709,
"p25": 194937458,
"p50": 203406792,
"p75": 210784958,
"p90": 219027333,
"p95": 362961542,
"p99": 362961542
}
}
},
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/TrinoHiveCatalog.java
lukasz-stec: The goal of this PR is to add Hive metastore metrics, as this is often the source of perf problems.
Support for REST and glue catalogs can be added as a follow-up.

testing/trino-tests/src/test/java/io/trino/connector/informationschema/TestInformationSchemaConnector.java
lukasz-stec: tracking removed

core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
lukasz-stec: ok, tracing removed

core/trino-main/src/main/java/io/trino/tracing/TracingMetadata.java
lukasz-stec: tracing removed

lukasz-stec (commented):
comments addressed

lib/trino-metastore/src/main/java/io/trino/metastore/MeasuredHiveMetastore.java
lukasz-stec: Are you suggesting changing the `DistributionSnapshot` globally or just for this case?

raunaqmorarka: globally

lukasz-stec: I ran a benchmark for the snapshot creation, and going with a single `valuesAt` call makes the performance significantly better, so I replaced the old approach with this one. Now, it is fast enough (50 - 100 ns per snapshot creation) to keep the old percentiles list 
```
Before
Benchmark                                    (valuesPerHistogram)  Mode  Cnt    Score    Error  Units
BenchmarkHistogram.benchmarkTDigestSnapshot                    10  avgt   10  152.097 ± 22.783  ns/op
BenchmarkHistogram.benchmarkTDigestSnapshot                   100  avgt   10  350.355 ± 13.313  ns/op
BenchmarkHistogram.benchmarkTDigestSnapshot                  1000  avgt   10  413.406 ±  6.479  ns/op

After
Benchmark                                    (valuesPerHistogram)  Mode  Cnt    Score    Error  Units
BenchmarkHistogram.benchmarkTDigestSnapshot                    10  avgt   10   52.963 ± 23.871  ns/op
BenchmarkHistogram.benchmarkTDigestSnapshot                   100  avgt   10  106.029 ±  0.304  ns/op
BenchmarkHistogram.benchmarkTDigestSnapshot                  1000  avgt   10  109.722 ±  0.327  ns/op
```

lukasz-stec:
rebased on master to resolve conflicts

core/trino-spi/src/main/java/io/trino/spi/metrics/Distribution.java
```diff
@@ -23,4 +23,6 @@ public interface Distribution<T>
     double getMax();
 
     double getPercentile(double percentile);
```
raunaqmorarka: Can you get rid of `double getPercentile(double percentile)` ? This is just used in tests now

raunaqmorarka (commented):
lgtm % minor comment

core/trino-spi/src/main/java/io/trino/spi/metrics/Distribution.java
lukasz-stec: done

client/trino-jdbc/src/test/java/io/trino/jdbc/TestTrinoDatabaseMetaData.java
```diff
@@ -2204,8 +2204,9 @@ private void assertMetadataCalls(
         });
 
         actualMetadataCallsCount = actualMetadataCallsCount.stream()
-                // Every query involves beginQuery and cleanupQuery, so ignore them.
-                .filter(method -> !"ConnectorMetadata.beginQuery".equals(method) && !"ConnectorMetadata.cleanupQuery".equals(method))
+                // Every query involves beginQuery, cleanupQuery and getMetrics, so ignore them.
+                .filter(method -> !"ConnectorMetadata.beginQuery".equals(method) && !"ConnectorMetadata.cleanupQuery".equals(method)
+                        && !"ConnectorMetadata.getMetrics".equals(method))
```
raunaqmorarka: This is unnecessary now, right ?

lukasz-stec: Yeah, it is a side effect of using tracing to detect metadata calls in tests. I dropped those changes.

lukasz-stec:
@findepi @raunaqmorarka The build is finally green. We should take the chance...
image

lukasz-stec:
Rebased on the master to resolve conflicts

PR #26912, opened 2025-10-10
Lazily build MongoClient in MongoClientModule

ebyhr:
## Description

The existing `MongoClientModule. defaultConfigurator` looks up DNS record when SRV is used.
Connectors shouldn't try accessing remote datasources during the initialization. 

## Release notes

### DIFF
--- a/plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoClientModule.java
+++ b/plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoClientModule.java
@@ -13,6 +13,8 @@
  */
 package io.trino.plugin.mongodb;
 
+import com.google.common.base.Supplier;
+import com.google.common.base.Suppliers;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Binder;
 import com.google.inject.Provides;
@@ -76,10 +78,12 @@ public class MongoClientModule
     @Provides
     public static MongoSession createMongoSession(TypeManager typeManager, MongoClientConfig config, Set<MongoClientSettingConfigurator> configurators, OpenTelemetry openTelemetry)
     {
-        MongoClientSettings.Builder options = MongoClientSettings.builder();
-        configurators.forEach(configurator -> configurator.configure(options));
-        options.addCommandListener(MongoTelemetry.builder(openTelemetry).build().newCommandListener());
-        MongoClient client = MongoClients.create(options.build());
+        Supplier<MongoClient> client = Suppliers.memoize(() -> {
+            MongoClientSettings.Builder options = MongoClientSettings.builder();
+            configurators.forEach(configurator -> configurator.configure(options));
+            options.addCommandListener(MongoTelemetry.builder(openTelemetry).build().newCommandListener());
+            return MongoClients.create(options.build());
+        });
 
         return new MongoSession(
                 typeManager,
--- a/plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoSession.java
+++ b/plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoSession.java
@@ -14,6 +14,7 @@
 package io.trino.plugin.mongodb;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Supplier;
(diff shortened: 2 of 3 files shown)


plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoClientModule.java
```diff
@@ -68,22 +60,17 @@ public void setup(Binder binder)
...
 
     @Singleton
     @Provides
-    public static MongoSession createMongoSession(TypeManager typeManager, MongoClientConfig config, Set<MongoClientSettingConfigurator> configurators, OpenTelemetry openTelemetry)
+    public static MongoSession createMongoSession(TypeManager typeManager, MongoClientFactory clientFactory, MongoClientConfig config)
     {
-        MongoClientSettings.Builder options = MongoClientSettings.builder();
-        configurators.forEach(configurator -> configurator.configure(options));
-        options.addCommandListener(MongoTelemetry.builder(openTelemetry).build().newCommandListener());
-        MongoClient client = MongoClients.create(options.build());
```
findepi, line 82: Would it be enough to wrap this initl block with this?

```java
client = Suppliers.memoize(() -> {
  ...
});
```

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoClientModule.java
```diff
@@ -68,22 +60,17 @@ public void setup(Binder binder)
...
-
         return new MongoSession(
                 typeManager,
-                client,
+                clientFactory,
```
findepi: `MongoClient` used to be singleton, and it doesn't seemed to be cached now. intentional?

findepi:
> The existing `MongoClientModule.createMongoSession` looks up DNS record when SRV is used.
> Connectors shouldn't try accessing remote datasources during the initialization.

❤️ 
cc @homar

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoPageSink.java
```diff
@@ -109,7 +108,6 @@ public MongoPageSink(
     @Override
     public CompletableFuture<?> appendPage(Page page)
     {
-        MongoCollection<Document> collection = mongoSession.getCollection(remoteTableName);
```
chenjian2664, line 112: unrelated to the change?

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoClientModule.java
ebyhr: It was intentional, but `Suppliers.memoize` looks better. Updated.

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoSession.java
```diff
@@ -200,12 +201,12 @@ public MongoSession(TypeManager typeManager, MongoClient client, MongoClientConf
     @Override
     public void close()
     {
-        client.close();
+        getClient().close();
```
findepi, line 205: This requires a comment that this is lazy-loaded shared client instance.
Otherwise it look like we're creating a new client just to close it.

Alternatively, we could inject supplier without memoization and explicitly cache the client on a field in here.

plugin/trino-mongodb/src/test/java/io/trino/plugin/mongodb/TestMongoPlugin.java
```diff
@@ -44,4 +44,17 @@ public void testCreateConnector()
 
         connector.shutdown();
     }
+
+    @Test
+    void testMongoDbSrv()
```
findepi: - the test above is `public`
- `testUnreachable....`?

plugin/trino-mongodb/src/main/java/io/trino/plugin/mongodb/MongoSession.java
ebyhr: Let me leave a comment instead of injecting a supplier.

Issue #23596, opened 2024-09-28
Flaky test `TestAlluxioFileSystem` and `TestAlluxioFileSystemExchangeManager`


ebyhr:
https://github.com/trinodb/trino/actions/runs/11073808274/job/30771180097
```
Error:  io.trino.filesystem.alluxio.TestAlluxioFileSystem -- Time elapsed: 90.72 s <<< ERROR!
java.util.concurrent.ExecutionException: org.testcontainers.containers.ContainerLaunchException: Container startup failed for image alluxio/alluxio:2.9.5
	at java.base/java.util.concurrent.CompletableFuture.wrapInExecutionException(CompletableFuture.java:345)
	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:440)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2117)
	at org.testcontainers.containers.GenericContainer.start(GenericContainer.java:327)
	at org.testcontainers.junit.jupiter.TestcontainersExtension$StoreAdapter.start(TestcontainersExtension.java:276)
	at org.testcontainers.junit.jupiter.TestcontainersExtension$StoreAdapter.access$200(TestcontainersExtension.java:263)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.lambda$null$4(TestcontainersExtension.java:83)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.lambda$startContainers$5(TestcontainersExtension.java:83)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1597)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.startContainers(TestcontainersExtension.java:83)
	at org.testcontainers.junit.jupiter.TestcontainersExtension.beforeAll(TestcontainersExtension.java:57)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1458)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2034)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:189)
Caused by: org.testcontainers.containers.ContainerLaunchException: Container startup failed for image alluxio/alluxio:2.9.5
[...]

PR #26918, opened 2025-10-10
Dynamic catalogs improvements

homar:
## Description
There are several things in this PR:
1. Making `metadata.catalogs` return all catalogs (included those that didn't load correctly), together with the their status 
2. Adding a test documenting behaviour of catalogs when there is an issue with loading them
3. Allowing dropping a catalog that was not loaded correctly

## Release notes

```markdown
## Section
* Allow to drop catalog that failed to load correctly
* Make metadata.catalogs return all catalogs 
```

## Summary by Sourcery

Track and surface the load state of catalogs by replacing the boolean loaded flag with a CatalogStatus enum, exposing a new state column in system.metadata.catalogs, and adding end-to-end tests for healthy and failed dynamic catalogs.

New Features:
- Introduce a CatalogStatus enum (LOADED, FAILED_TO_LOAD) to represent catalog load state
- Add a state column to the system.metadata.catalogs table to report each catalog’s load status
- Add TestDynamicCatalogs to validate dynamic catalog creation, loading failures, and drops

Enhancements:
- Replace boolean loaded flags with CatalogStatus across Catalog, CatalogInfo, CatalogConnector, transaction manager, and metadata components
- Update TestSystemMetadataCatalogTable to assert on the new state column
- Refine TestConfiguredFeatures to use unified row-based assertions for connector listing

Tests:
- Updated TestSystemMetadataCatalogTable and TestConfiguredFeatures for catalog state changes
- Added comprehensive tests in TestDynamicCatalogs for both healthy and unhealthy catalog scenarios

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
@@ -155,6 +155,7 @@ import static io.airlift.concurrent.MoreFutures.toListenableFuture;
 import static io.trino.SystemSessionProperties.getRetryPolicy;
 import static io.trino.metadata.CatalogMetadata.SecurityManagement.CONNECTOR;
 import static io.trino.metadata.CatalogMetadata.SecurityManagement.SYSTEM;
+import static io.trino.metadata.CatalogStatus.OPERATIONAL;
 import static io.trino.metadata.GlobalFunctionCatalog.BUILTIN_SCHEMA;
 import static io.trino.metadata.GlobalFunctionCatalog.isBuiltinFunctionName;
 import static io.trino.metadata.LanguageFunctionManager.isTrinoSqlLanguageFunction;
@@ -812,7 +813,9 @@ public final class MetadataManager
     public void dropCatalog(Session session, CatalogName catalog, boolean cascade)
     {
         Optional<CatalogMetadata> catalogMetadata = Optional.empty();
-        if (catalogManager.getCatalog(catalog).isPresent()) {
+        // there is a potential race condition here, TODO: https://github.com/trinodb/trino/issues/26927
+        Optional<Catalog> optionalCatalog = catalogManager.getCatalog(catalog);
+        if (optionalCatalog.isPresent() && optionalCatalog.get().getCatalogStatus() == OPERATIONAL) {
             catalogMetadata = Optional.of(getCatalogMetadataForWrite(session, catalog.toString()));
         }
         catalogManager.dropCatalog(catalog, cascade);
--- a/testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
+++ b/testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
@@ -16,7 +16,6 @@ package io.trino.connector;
 import com.google.common.collect.ImmutableMap;
 import com.google.inject.Key;
 import io.trino.Session;
-import io.trino.metadata.CatalogManager;
 import io.trino.plugin.memory.MemoryPlugin;
 import io.trino.spi.catalog.CatalogName;
 import io.trino.spi.catalog.CatalogProperties;
@@ -84,16 +83,16 @@ public class TestDynamicCatalogs
         H2QueryRunner h2QueryRunner = new H2QueryRunner();
         CatalogStore catalogStore = queryRunner.getCoordinator().getInstance(new Key<>() {});
         ConnectorServicesProvider connectorServicesProvider = queryRunner.getCoordinator().getInstance(new Key<>() {});
-        CatalogManager catalogManager = queryRunner.getCoordinator().getInstance(new Key<>() {});
         CatalogProperties catalogProperties = catalogStore.createCatalogProperties(new CatalogName(catalogName), new ConnectorName("memory"), ImmutableMap.of("invalid", "128MB"));
         catalogStore.addOrReplaceCatalog(catalogProperties);
         connectorServicesProvider.loadInitialCatalogs();
 
         assertQuery(queryRunner, session, "SHOW CATALOGS", h2QueryRunner, "VALUES 'healthy_catalog', '" + catalogName + "', 'system'", false, false);
         assertQueryFails(queryRunner, session, "CREATE TABLE %s.default.test_table (age INT)".formatted(catalogName), ".*Catalog '%s' failed to initialize and is disabled.*".formatted(catalogName));
+        assertQueryFails(queryRunner, session, "SELECT * FROM %s.default.test_table".formatted(catalogName), ".*Catalog '%s' failed to initialize and is disabled.*".formatted(catalogName));
         assertQueryFails(queryRunner, session, "CREATE CATALOG %s USING memory WITH (\"memory.max-data-per-node\" = '128MB')".formatted(catalogName), ".*Catalog '%s' already exists.*".formatted(catalogName));
-        assertQueryFails(queryRunner, session, "DROP CATALOG " + catalogName, ".*Catalog '%s' failed to initialize and is disabled.*".formatted(catalogName));
 
-        catalogManager.dropCatalog(new CatalogName(catalogName), false);
+        assertUpdate(queryRunner, session, "DROP CATALOG " + catalogName, OptionalLong.empty(), Optional.empty());
+        assertQuery(queryRunner, session, "SHOW CATALOGS", h2QueryRunner, "VALUES 'healthy_catalog', 'system'", false, false);
     }
 }


core/trino-main/src/main/java/io/trino/connector/CatalogConnector.java
```diff
@@ -54,8 +54,7 @@ public CatalogConnector(
                 connectorName,
                 catalogConnector,
                 informationSchemaConnector,
-                systemConnector,
-                true);
+                systemConnector);
```
findepi: > Revert "metadata.catalogs should only list loaded catalogs"
>
> This reverts commit https://github.com/trinodb/trino/commit/a7d9d9136b8b8d2659a05bd9ab73fe421d715752.

Please include rationale for the revert

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -0,0 +1,87 @@
...
+{
+    private CatalogStore catalogStore;
+    private ConnectorServicesProvider connectorServicesProvider;
+
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        Session session = testSessionBuilder().build();
+        QueryRunner queryRunner = DistributedQueryRunner.builder(session)
```
findepi:
```suggestion
        QueryRunner queryRunner = DistributedQueryRunner.builder(testSession())
```

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -0,0 +1,87 @@
...
+                .build();
+        queryRunner.installPlugin(new MemoryPlugin());
+        queryRunner.createCatalog("healthy_catalog", "memory", ImmutableMap.of("memory.max-data-per-node", "128MB"));
+        catalogStore = queryRunner.getCoordinator().getInstance(new Key<>() {});
+        connectorServicesProvider = queryRunner.getCoordinator().getInstance(new Key<>() {});
+        return queryRunner;
+    }
+
+    @Test
+    public void testHealthyCatalog()
```
findepi: testNewHealthyCatalog

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -0,0 +1,87 @@
...
+        assertUpdate("CREATE TABLE %s.default.test_table (age INT)".formatted(catalogName));
+        assertUpdate("INSERT INTO %s.default.test_table VALUES (10)".formatted(catalogName), 1);
+        assertQuery("SELECT * FROM %s.default.test_table".formatted(catalogName), "VALUES (10)");
+
+        assertUpdate("DROP CATALOG " + catalogName);
+        assertQuery("SHOW CATALOGS", "VALUES 'healthy_catalog', 'system'");
+    }
+
+    @Test
+    public void testUnhealthyCatalog()
```
findepi:
```suggestion
    public void testNewUnhealthyCatalog()
```

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -0,0 +1,87 @@
...
+    {
+        String catalogName = "new_catalog" + randomNameSuffix();
+        // simulate loading an unhealthy catalog during a startup
+        CatalogProperties catalogProperties = catalogStore.createCatalogProperties(new CatalogName(catalogName), new ConnectorName("memory"), ImmutableMap.of("invalid", "128MB"));
+        catalogStore.addOrReplaceCatalog(catalogProperties);
+        connectorServicesProvider.loadInitialCatalogs();
+
+        assertQuery("SHOW CATALOGS", "VALUES 'healthy_catalog', '" + catalogName + "', 'system'");
+        assertQueryFails("CREATE TABLE %s.default.test_table (age INT)".formatted(catalogName), ".*Catalog '%s' failed to initialize and is disabled.*".formatted(catalogName));
```
findepi: assert what's the error for a SELECT statement over such catalog

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -0,0 +1,87 @@
...
+        assertQueryFails("CREATE CATALOG %s USING memory WITH (\"memory.max-data-per-node\" = '128MB')".formatted(catalogName), ".*Catalog '%s' already exists.*".formatted(catalogName));
+        assertUpdate("DROP CATALOG " + catalogName);
+    }
```
findepi: assert that it was successfully dropped

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -0,0 +1,87 @@
...
+
+    @Test
+    public void testUnhealthyCatalog()
+    {
+        String catalogName = "new_catalog" + randomNameSuffix();
+        // simulate loading an unhealthy catalog during a startup
+        CatalogProperties catalogProperties = catalogStore.createCatalogProperties(new CatalogName(catalogName), new ConnectorName("memory"), ImmutableMap.of("invalid", "128MB"));
+        catalogStore.addOrReplaceCatalog(catalogProperties);
+        connectorServicesProvider.loadInitialCatalogs();
```
findepi, line 88: The function has the code in there

```
if (state == State.INITIALIZED) {
                return;
            }
```

does this line do anything?

core/trino-main/src/main/java/io/trino/metadata/CatalogStatus.java
```diff
@@ -0,0 +1,20 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.metadata;
+
+public enum CatalogStatus
+{
+    LOADED,
+    FAILED_TO_LOAD,
```
findepi: "loaded catalogs" inuititively means those loaded into the system, aka. defined

maybe we can call these options: OPERATIONAL / FAILING

testing/trino-tests/src/test/java/io/trino/connector/system/TestSystemMetadataCatalogTable.java
```diff
@@ -0,0 +1,88 @@
...
+        assertQuery("SELECT * FROM system.metadata.catalogs", "VALUES" +
+                "('healthy_catalog', 'healthy_catalog', 'memory', 'LOADED'), " +
+                "('brain', 'brain', 'memory', 'LOADED'), " +
+                "('system', 'system', 'system', 'LOADED')");
+
+        assertUpdate("DROP CATALOG brain");
+    }
+
+    @Test
+    public void testCatalogTableShowsCorrectStatusWhenCatalogIsNotLoadedCorrectly()
```
findepi: `testCatalogNotLoadedCorrectly` ?

testing/trino-tests/src/test/java/io/trino/connector/system/TestSystemMetadataCatalogTable.java
```diff
@@ -0,0 +1,88 @@
...
+        queryRunner.installPlugin(new MemoryPlugin());
+        queryRunner.createCatalog("healthy_catalog", "memory", ImmutableMap.of("memory.max-data-per-node", "128MB"));
+        catalogStore = queryRunner.getCoordinator().getInstance(new Key<>() {});
+        connectorServicesProvider = queryRunner.getCoordinator().getInstance(new Key<>() {});
+        catalogManager = queryRunner.getCoordinator().getInstance(new Key<>() {});
+        return queryRunner;
+    }
+
+    @Test
+    public void testCatalogTableShowsCorrectStatusWhenCatalogsAreLoadedCorrectly()
```
findepi: `testNewCatalog`  / `testNewCatalogStatus`?

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -812,7 +813,8 @@ public void createCatalog(Session session, CatalogName catalog, ConnectorName co
     public void dropCatalog(Session session, CatalogName catalog, boolean cascade)
     {
         Optional<CatalogMetadata> catalogMetadata = Optional.empty();
-        if (catalogManager.getCatalog(catalog).isPresent()) {
+        Optional<Catalog> optionalCatalog = catalogManager.getCatalog(catalog);
+        if (optionalCatalog.isPresent() && optionalCatalog.get().getCatalogStatus() == LOADED) {
```
findepi: What we do here:

1. we first read from `catalogManager`
   1.1. we conditionally re-read something in `getCatalogMetadataForWrite`
3. we then issue unconditional drop in `catalogManager`
4. depending on the status if the (1) read, `systemSecurityMetadata.catalogDropped` is called

What we want to guarantee is that:
_if a catalog ceased to exist, systemSecurityMetadata.catalogDropped has been called_

Which may not be the case when **e.g.** 

- (pre-existing) catalog did not exist and was created between (1) and (2) 
- (new) catalog did exist but unhealthy and was fixed between (1) and (2)


It's pre-existing problem. Let's file a follow-up issue and link TODO comment here.

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
```diff
@@ -84,9 +81,7 @@ public void testUnhealthyCatalog()
 
         assertQuery("SHOW CATALOGS", "VALUES 'healthy_catalog', '" + catalogName + "', 'system'");
         assertQueryFails("CREATE TABLE %s.default.test_table (age INT)".formatted(catalogName), ".*Catalog '%s' failed to initialize and is disabled.*".formatted(catalogName));
-        assertQueryFails("DROP CATALOG " + catalogName, ".*Catalog '%s' failed to initialize and is disabled.*".formatted(catalogName));
         assertQueryFails("CREATE CATALOG %s USING memory WITH (\"memory.max-data-per-node\" = '128MB')".formatted(catalogName), ".*Catalog '%s' already exists.*".formatted(catalogName));
-
-        catalogManager.dropCatalog(new CatalogName(catalogName), false);
+        assertUpdate("DROP CATALOG " + catalogName);
```
findepi: this assert the DROP did not fail
assert also it's indeed gone.

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
homar: Actually when you use queryRunner then `loadInitialCatalogs` is not called at all. So in fact this is the first invocation of this method. Normally it is called from `io.trino.server.Server` class but for queryRunners the base class is `TestServer` not Server

core/trino-main/src/main/java/io/trino/metadata/CatalogStatus.java
homar: Sure I can try. cc @kokosing

testing/trino-tests/src/test/java/io/trino/connector/TestDynamicCatalogs.java
findepi: Sounds like an omission in testing framework. i don't see yet a reason for `loadInitialCatalogs` not be called in tests. Maybe i am overlooking something though

Also, given how loadInitialCatalogs has guards, does it make sense to call it in two test methods? Or will the second call do nothing? (Maybe the test class should have no QR shared between test methods, and each test method should create own thrown-away instance?)

homar: I agree it is an omission. Not sure though I want to fix it now, I am not sure what problems could that cause. 
I will go with your suggestion and have QR per test

homar: @findepi so I went with QR per test, unfortunately that made `getQueryRunner` method obsolete and so I didn't have to extend `AbstractTestQueryFramework` anymore. Which resulted in much uglier assertions :(

findepi:
CI
- https://github.com/trinodb/trino/issues/23596
- https://github.com/trinodb/trino/issues/26943 (new)

PR #26919, opened 2025-10-10
Add support for vectorized null suppression for block serde

SongChujun:
## Description

This PR add support for vectorized null suppression for block serde using Java SIMD API.

### detect SIMD support from the CPU.

This functionality is essential to prevent regression from happening, even though Java vector API is platform-agonitic, it only provides guarantess for correcntess, but no guarantee for performance improbement. If the JVM is running on a older CPU without decent SIMD support, the Java vector API may fall back to emulated execution, which can be slower compared to the extsing scalar path, instead of real SIMD execution. So we want a extra layer gating make sure that such fall back would not happen.

Currently, we add support for Intel and AMD CPUs, we may extend to support Graviton later if experiments can show speed up on Graviton machines.

### add vectorized path for null suppression in block serde

Add vectorized path for null suppression for byte/short/integer/long.

Microbenchmark results are given below.

[Microbenchmark](https://jmh.morethan.io/?gists=612261f3588b4435623b6241700c729d,20c86e3753f9e940a60f7de3ce226a05) on Intel CPU with avx512F support.

[Microbenchmark](https://jmh.morethan.io/?gists=c9295a839b5b6fe64eed351062f435ce,75f2ec3560a2ee6d058096e59da8afe2) on AMD zen4 CPU with avx512F support.

The reason that the speed up is not the potential maximum speed up(16x for Int, 8x for Long for AVX512) is 
1. We only optimize null supression part, which only accounts for a part of the latency for serialize pages that benchmark is testing(~55% for compressInts)
image

2. The vectorized null suppression has a much higher L1d cache miss rate, 54.14% for vectorized and 6.56% for scalar.  Such cache misses can easily wipe out SIMD’s performance gains.

## Tests:
- Add TestEncoderUtil to verify scalar and vectorized compression outputs match

## Next steps
1. Currently, the gating on using vectorized code path is on machines with avx512 support only, we may want to extend this to avx2 support. This could open the opporunnties for AMD zen3 CPUs which only have avx2 support.
2. Extend SIMD support to AWS Gravition machines
4. Add SIMD support for other operations.

## Release notes

```markdown
## General
* Improve performance of data exchanges by using SIMD instructions on x86 CPUs that support the required extensions. This can be disabled by setting `experimental.blockserde-vectorized-null-suppression-strategy=NONE`  ({issue}`26919`)
```

## Summary by Sourcery

Add vectorized SIMD-based null suppression for block serialization using the Java Vector API with dynamic CPU feature detection

New Features:
- Add compressBytesWithNulls, compressShortsWithNulls, compressIntsWithNulls, compressLongsWithNulls methods in EncoderUtil with vectorized and scalar paths
- Introduce SimdSupport SPI interface and platform-specific implementations (Intel, AMD, Graviton) with SimdSupportManager for runtime detection
- Refactor block encodings to delegate null filtering to new EncoderUtil methods
- Add SimdInitializer for eager SIMD detection at server startup

Enhancements:
- Expose EncoderUtil publicly and centralize null suppression logic
- Reduce benchmark row count and initialize SIMD support in benchmarks and tests
- Update module-info to require jdk.incubator.vector

Build:
- Add SimdInitializer binding in Guice and include incubator vector module

Tests:
- Add TestEncoderUtil to verify scalar and vectorized compression outputs match

## Summary by Sourcery

Add vectorized null suppression for block serialization using the Java Vector API with dynamic CPU feature detection, refactor block encodings to leverage the new SIMD path, and update benchmarks and tests to validate correctness.

New Features:
- Introduce SimdSupport SPI and runtime detection (SimdSupportManager) with Intel, AMD, and Graviton implementations
- Add vectorized null suppression methods in EncoderUtil for byte, short, int, and long arrays
- Add SimdInitializer to eagerly detect and install SIMD support at server startup

Enhancements:
- Refactor existing block encodings to delegate null filtering to the new EncoderUtil.compress*WithNulls methods
- Reduce row count in BenchmarkBlockSerde to match real workloads and ensure SIMD initialization

Tests:
- Add TestEncoderUtil to verify scalar and vectorized compression outputs match
- Initialize SIMD support in existing SPI block encoding tests

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/BlockEncodingManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/BlockEncodingManager.java
@@ -13,6 +13,9 @@
  */
 package io.trino.metadata;
 
+import com.google.inject.Inject;
+import io.trino.simd.BlockEncodingSimdSupport;
+import io.trino.simd.BlockEncodingSimdSupport.SimdSupport;
 import io.trino.spi.block.ArrayBlockEncoding;
 import io.trino.spi.block.Block;
 import io.trino.spi.block.BlockEncoding;
@@ -41,14 +44,16 @@ public final class BlockEncodingManager
     // for serialization
     private final Map<Class<? extends Block>, BlockEncoding> blockEncodingNamesByClass = new ConcurrentHashMap<>();
 
-    public BlockEncodingManager()
+    @Inject
+    public BlockEncodingManager(BlockEncodingSimdSupport blockEncodingSimdSupport)
     {
         // add the built-in BlockEncodings
+        SimdSupport simdSupport = blockEncodingSimdSupport.getSimdSupport();
         addBlockEncoding(new VariableWidthBlockEncoding());
-        addBlockEncoding(new ByteArrayBlockEncoding());
-        addBlockEncoding(new ShortArrayBlockEncoding());
-        addBlockEncoding(new IntArrayBlockEncoding());
-        addBlockEncoding(new LongArrayBlockEncoding());
+        addBlockEncoding(new ByteArrayBlockEncoding(simdSupport.expandAndCompressByte()));
+        addBlockEncoding(new ShortArrayBlockEncoding(simdSupport.expandAndCompressShort()));
+        addBlockEncoding(new IntArrayBlockEncoding(simdSupport.expandAndCompressInt()));
+        addBlockEncoding(new LongArrayBlockEncoding(simdSupport.expandAndCompressLong()));
         addBlockEncoding(new Fixed12BlockEncoding());
         addBlockEncoding(new Int128ArrayBlockEncoding());
         addBlockEncoding(new DictionaryBlockEncoding());
--- a/core/trino-main/src/test/java/io/trino/server/remotetask/TestHttpRemoteTask.java
+++ b/core/trino-main/src/test/java/io/trino/server/remotetask/TestHttpRemoteTask.java
@@ -61,6 +61,7 @@ import io.trino.server.DynamicFilterService;
 import io.trino.server.FailTaskRequest;
 import io.trino.server.HttpRemoteTaskFactory;
 import io.trino.server.TaskUpdateRequest;
+import io.trino.simd.BlockEncodingSimdSupport;
 import io.trino.spi.ErrorCode;
 import io.trino.spi.QueryId;
 import io.trino.spi.block.Block;
@@ -676,6 +677,7 @@ public class TestHttpRemoteTask
                         jsonCodecBinder(binder).bindJsonCodec(TaskUpdateRequest.class);
                         jsonCodecBinder(binder).bindJsonCodec(FailTaskRequest.class);
 
+                        binder.bind(BlockEncodingSimdSupport.class).toInstance(new BlockEncodingSimdSupport(true));
                         binder.bind(TypeManager.class).toInstance(TESTING_TYPE_MANAGER);
                         binder.bind(BlockEncodingManager.class).in(SINGLETON);
                         binder.bind(BlockEncodingSerde.class).to(InternalBlockEncodingSerde.class).in(SINGLETON);
--- a/core/trino-spi/pom.xml
+++ b/core/trino-spi/pom.xml
@@ -263,6 +263,34 @@
                                     <old>method void io.trino.spi.eventlistener.QueryStatistics::&lt;init&gt;(java.time.Duration, java.time.Duration, java.time.Duration, java.time.Duration, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, long, long, long, long, long, long, long, long, long, long, long, long, long, long, double, double, java.util.List&lt;io.trino.spi.eventlistener.StageGcStatistics&gt;, int, boolean, java.util.List&lt;io.trino.spi.eventlistener.StageCpuDistribution&gt;, java.util.List&lt;io.trino.spi.eventlistener.StageOutputBufferUtilization&gt;, java.util.List&lt;io.trino.spi.eventlistener.StageOutputBufferMetrics&gt;, java.util.List&lt;io.trino.spi.eventlistener.StageTaskStatistics&gt;, java.util.List&lt;io.trino.spi.eventlistener.DynamicFilterDomainStatistics&gt;, java.util.function.Supplier&lt;java.util.List&lt;java.lang.String&gt;&gt;, java.util.List&lt;io.trino.spi.eventlistener.QueryPlanOptimizerStatistics&gt;, java.util.Map&lt;java.lang.String, io.trino.spi.metrics.Metrics&gt;, java.util.Optional&lt;java.lang.String&gt;)</old>
                                     <new>method void io.trino.spi.eventlistener.QueryStatistics::&lt;init&gt;(java.time.Duration, java.time.Duration, java.time.Duration, java.time.Duration, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, java.util.Optional&lt;java.time.Duration&gt;, long, long, long, long, long, long, long, long, long, long, long, long, long, long, double, double, java.util.List&lt;io.trino.spi.eventlistener.StageGcStatistics&gt;, int, boolean, java.util.List&lt;io.trino.spi.eventlistener.StageCpuDistribution&gt;, java.util.List&lt;io.trino.spi.eventlistener.StageOutputBufferUtilization&gt;, java.util.List&lt;io.trino.spi.eventlistener.StageOutputBufferMetrics&gt;, java.util.List&lt;io.trino.spi.eventlistener.StageTaskStatistics&gt;, java.util.List&lt;io.trino.spi.eventlistener.DynamicFilterDomainStatistics&gt;, java.util.function.Supplier&lt;java.util.List&lt;java.lang.String&gt;&gt;, java.util.List&lt;io.trino.spi.eventlistener.QueryPlanOptimizerStatistics&gt;, java.util.Map&lt;java.lang.String, io.trino.spi.metrics.Metrics&gt;, java.util.Optional&lt;java.lang.String&gt;)</new>
                                 </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.numberOfParametersChanged</code>
+                                    <old>method void io.trino.spi.block.ByteArrayBlockEncoding::&lt;init&gt;()</old>
+                                    <new>method void io.trino.spi.block.ByteArrayBlockEncoding::&lt;init&gt;(boolean)</new>
+                                    <justification>ByteArrayBlockEncoding need to accept a parameter to enable SIMD support</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.numberOfParametersChanged</code>
+                                    <old>method void io.trino.spi.block.IntArrayBlockEncoding::&lt;init&gt;()</old>
+                                    <new>method void io.trino.spi.block.IntArrayBlockEncoding::&lt;init&gt;(boolean)</new>
+                                    <justification>IntArrayBlockEncoding need to accept a parameter to enable SIMD support</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.numberOfParametersChanged</code>
+                                    <old>method void io.trino.spi.block.LongArrayBlockEncoding::&lt;init&gt;()</old>
+                                    <new>method void io.trino.spi.block.LongArrayBlockEncoding::&lt;init&gt;(boolean)</new>
+                                    <justification>LongArrayBlockEncoding need to accept a parameter to enable SIMD support</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.numberOfParametersChanged</code>
+                                    <old>method void io.trino.spi.block.ShortArrayBlockEncoding::&lt;init&gt;()</old>
+                                    <new>method void io.trino.spi.block.ShortArrayBlockEncoding::&lt;init&gt;(boolean)</new>
+                                    <justification>ShortArrayBlockEncoding need to accept a parameter to enable SIMD support</justification>
+                                </item>
                             </differences>
                         </revapi.differences>
                     </analysisConfiguration>
--- a/core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
@@ -16,13 +16,26 @@ package io.trino.spi.block;
 import io.airlift.slice.SliceInput;
 import io.airlift.slice.SliceOutput;
 import jakarta.annotation.Nullable;
+import jdk.incubator.vector.ByteVector;
+import jdk.incubator.vector.IntVector;
+import jdk.incubator.vector.LongVector;
+import jdk.incubator.vector.ShortVector;
+import jdk.incubator.vector.VectorMask;
+import jdk.incubator.vector.VectorSpecies;
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Optional;
 
+import static java.util.Objects.checkFromIndexSize;
+
 final class EncoderUtil
 {
+    private static final VectorSpecies<Long> LONG_SPECIES = LongVector.SPECIES_PREFERRED;
+    private static final VectorSpecies<Integer> INT_SPECIES = IntVector.SPECIES_PREFERRED;
+    private static final VectorSpecies<Byte> BYTE_SPECIES = ByteVector.SPECIES_PREFERRED;
+    private static final VectorSpecies<Short> SHORT_SPECIES = ShortVector.SPECIES_PREFERRED;
+
     private EncoderUtil() {}
 
     /**
@@ -119,5 +132,149 @@ final class EncoderUtil
         catch (IOException e) {
             throw new UncheckedIOException(e);
         }
+    }
+
+    static void compactBytesWithNullsVectorized(SliceOutput sliceOutput, byte[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        byte[] compacted = new byte[length];
+        int valuesIndex = 0;
+        int compactedIndex = 0;
+        for (; valuesIndex < BYTE_SPECIES.loopBound(length); valuesIndex += BYTE_SPECIES.length()) {
+            VectorMask<Byte> mask = BYTE_SPECIES.loadMask(isNull, valuesIndex + offset).not();
+            ByteVector.fromArray(BYTE_SPECIES, values, valuesIndex + offset)
+                    .compress(mask)
+                    .intoArray(compacted, compactedIndex);
+            compactedIndex += mask.trueCount();
+        }
+        for (; valuesIndex < length; valuesIndex++) {
+            compacted[compactedIndex] = values[valuesIndex + offset];
+            compactedIndex += isNull[valuesIndex + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeBytes(compacted, 0, compactedIndex);
+    }
+
+    static void compactBytesWithNullsScalar(SliceOutput sliceOutput, byte[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        byte[] compacted = new byte[length];
+        int compactedIndex = 0;
+        for (int i = 0; i < length; i++) {
+            compacted[compactedIndex] = values[i + offset];
+            compactedIndex += isNull[i + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeBytes(compacted, 0, compactedIndex);
+    }
+
+    static void compactShortsWithNullsVectorized(SliceOutput sliceOutput, short[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        short[] compacted = new short[length];
+        int valuesIndex = 0;
+        int compactedIndex = 0;
+        for (; valuesIndex < SHORT_SPECIES.loopBound(length); valuesIndex += SHORT_SPECIES.length()) {
+            VectorMask<Short> mask = SHORT_SPECIES.loadMask(isNull, valuesIndex + offset).not();
+            ShortVector.fromArray(SHORT_SPECIES, values, valuesIndex + offset)
+                    .compress(mask)
+                    .intoArray(compacted, compactedIndex);
+            compactedIndex += mask.trueCount();
+        }
+        for (; valuesIndex < length; valuesIndex++) {
+            compacted[compactedIndex] = values[valuesIndex + offset];
+            compactedIndex += isNull[valuesIndex + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeShorts(compacted, 0, compactedIndex);
+    }
+
+    static void compactShortsWithNullsScalar(SliceOutput sliceOutput, short[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        short[] compacted = new short[length];
+        int compactedIndex = 0;
+        for (int i = 0; i < length; i++) {
+            compacted[compactedIndex] = values[i + offset];
+            compactedIndex += isNull[i + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeShorts(compacted, 0, compactedIndex);
+    }
+
+    static void compactIntsWithNullsVectorized(SliceOutput sliceOutput, int[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        int[] compacted = new int[length];
+        int valuesIndex = 0;
+        int compactedIndex = 0;
+        for (; valuesIndex < INT_SPECIES.loopBound(length); valuesIndex += INT_SPECIES.length()) {
+            VectorMask<Integer> mask = INT_SPECIES.loadMask(isNull, valuesIndex + offset).not();
+            IntVector.fromArray(INT_SPECIES, values, valuesIndex + offset)
+                    .compress(mask)
+                    .intoArray(compacted, compactedIndex);
+            compactedIndex += mask.trueCount();
+        }
+        for (; valuesIndex < length; valuesIndex++) {
+            compacted[compactedIndex] = values[valuesIndex + offset];
+            compactedIndex += isNull[valuesIndex + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeInts(compacted, 0, compactedIndex);
+    }
+
+    static void compactIntsWithNullsScalar(SliceOutput sliceOutput, int[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        int[] compacted = new int[length];
+        int compactedIndex = 0;
+        for (int i = 0; i < length; i++) {
+            compacted[compactedIndex] = values[i + offset];
+            compactedIndex += isNull[i + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeInts(compacted, 0, compactedIndex);
+    }
+
+    static void compactLongsWithNullsVectorized(SliceOutput sliceOutput, long[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        long[] compacted = new long[length];
+        int valuesIndex = 0;
+        int compactedIndex = 0;
+        for (; valuesIndex < LONG_SPECIES.loopBound(length); valuesIndex += LONG_SPECIES.length()) {
+            VectorMask<Long> mask = LONG_SPECIES.loadMask(isNull, valuesIndex + offset).not();
+            LongVector.fromArray(LONG_SPECIES, values, valuesIndex + offset)
+                    .compress(mask)
+                    .intoArray(compacted, compactedIndex);
+            compactedIndex += mask.trueCount();
+        }
+        for (; valuesIndex < length; valuesIndex++) {
+            compacted[compactedIndex] = values[valuesIndex + offset];
+            compactedIndex += isNull[valuesIndex + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeLongs(compacted, 0, compactedIndex);
+    }
+
+    static void compactLongsWithNullsScalar(SliceOutput sliceOutput, long[] values, boolean[] isNull, int offset, int length)
+    {
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        long[] compacted = new long[length];
+        int compactedIndex = 0;
+        for (int i = 0; i < length; i++) {
+            compacted[compactedIndex] = values[i + offset];
+            compactedIndex += isNull[i + offset] ? 0 : 1;
+        }
+        sliceOutput.writeInt(compactedIndex);
+        sliceOutput.writeLongs(compacted, 0, compactedIndex);
     }
 }
--- a/core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
@@ -17,16 +17,24 @@ import io.airlift.slice.SliceInput;
 import io.airlift.slice.SliceOutput;
 import jakarta.annotation.Nullable;
 
+import static io.trino.spi.block.EncoderUtil.compactIntsWithNullsScalar;
+import static io.trino.spi.block.EncoderUtil.compactIntsWithNullsVectorized;
 import static io.trino.spi.block.EncoderUtil.decodeNullBits;
 import static io.trino.spi.block.EncoderUtil.encodeNullsAsBits;
 import static io.trino.spi.block.EncoderUtil.retrieveNullBits;
 import static java.lang.System.arraycopy;
-import static java.util.Objects.checkFromIndexSize;
 
 public class IntArrayBlockEncoding
         implements BlockEncoding
 {
     public static final String NAME = "INT_ARRAY";
+
+    private final boolean enableVectorizedNullSuppression;
+
+    public IntArrayBlockEncoding(boolean enableVectorizedNullSuppression)
+    {
+        this.enableVectorizedNullSuppression = enableVectorizedNullSuppression;
+    }
 
     @Override
     public String getName()
@@ -51,7 +59,6 @@ public class IntArrayBlockEncoding
         @Nullable
         boolean[] isNull = intArrayBlock.getRawValueIsNull();
         int[] rawValues = intArrayBlock.getRawValues();
-        checkFromIndexSize(rawOffset, positionCount, rawValues.length);
 
         encodeNullsAsBits(sliceOutput, isNull, rawOffset, positionCount);
 
@@ -59,15 +66,12 @@ public class IntArrayBlockEncoding
             sliceOutput.writeInts(rawValues, rawOffset, positionCount);
         }
         else {
-            int[] valuesWithoutNull = new int[positionCount];
-            int nonNullPositionCount = 0;
-            for (int i = 0; i < positionCount; i++) {
-                valuesWithoutNull[nonNullPositionCount] = rawValues[i + rawOffset];
-                nonNullPositionCount += isNull[i + rawOffset] ? 0 : 1;
+            if (enableVectorizedNullSuppression) {
+                compactIntsWithNullsVectorized(sliceOutput, rawValues, isNull, rawOffset, positionCount);
             }
-
-            sliceOutput.writeInt(nonNullPositionCount);
-            sliceOutput.writeInts(valuesWithoutNull, 0, nonNullPositionCount);
+            else {
+                compactIntsWithNullsScalar(sliceOutput, rawValues, isNull, rawOffset, positionCount);
+            }
         }
     }
 
--- a/core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
+++ b/core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
@@ -1,0 +1,229 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.spi.block;
+
+import io.airlift.slice.DynamicSliceOutput;
+import org.junit.jupiter.api.Test;
+
+import java.util.Random;
+
+import static io.trino.spi.block.EncoderUtil.compactBytesWithNullsScalar;
+import static io.trino.spi.block.EncoderUtil.compactBytesWithNullsVectorized;
+import static io.trino.spi.block.EncoderUtil.compactIntsWithNullsScalar;
+import static io.trino.spi.block.EncoderUtil.compactIntsWithNullsVectorized;
+import static io.trino.spi.block.EncoderUtil.compactLongsWithNullsScalar;
+import static io.trino.spi.block.EncoderUtil.compactLongsWithNullsVectorized;
+import static io.trino.spi.block.EncoderUtil.compactShortsWithNullsScalar;
+import static io.trino.spi.block.EncoderUtil.compactShortsWithNullsVectorized;
+import static org.assertj.core.api.Assertions.assertThat;
+
+final class TestEncoderUtil
+{
+    private static final int[] TEST_LENGTHS = {0, 3, 255, 257, 512, 530, 1024, 2048, 8192};
+    private static final int[] TEST_OFFSETS = {0, 2, 256};
+    private static final long RANDOM_SEED = 42;
+
+    @Test
+    void testBytesScalarEqualsVector()
+    {
+        for (int length : TEST_LENGTHS) {
+            for (int offset : TEST_OFFSETS) {
+                byte[] values = randomBytes(offset + length);
+                for (boolean[] isNull : getIsNullArray(offset + length)) {
+                    byte[] scalar = compressBytesScalar(values, isNull, offset, length);
+                    byte[] vector = compressBytesVectorized(values, isNull, offset, length);
+                    assertThat(vector).as("bytes: scalar and vector outputs differ").isEqualTo(scalar);
+                }
+            }
+        }
+    }
+
+    @Test
+    void testShortsScalarEqualsVector()
+    {
+        for (int length : TEST_LENGTHS) {
+            for (int offset : TEST_OFFSETS) {
+                short[] values = randomShorts(offset + length);
+                for (boolean[] isNull : getIsNullArray(offset + length)) {
+                    byte[] scalar = compressShortsScalar(values, isNull, offset, length);
+                    byte[] vector = compressShortsVectorized(values, isNull, offset, length);
+                    assertThat(vector).as("shorts: scalar and vector outputs differ").isEqualTo(scalar);
+                }
+            }
+        }
+    }
+
+    @Test
+    void testIntsScalarEqualsVector()
+    {
+        for (int length : TEST_LENGTHS) {
+            for (int offset : TEST_OFFSETS) {
+                int[] values = randomInts(offset + length);
+                for (boolean[] isNull : getIsNullArray(offset + length)) {
+                    byte[] scalar = compressIntsScalar(values, isNull, offset, length);
+                    byte[] vector = compressIntsVectorized(values, isNull, offset, length);
+                    assertThat(vector).as("ints: scalar and vector outputs differ").isEqualTo(scalar);
+                }
+            }
+        }
+    }
+
+    @Test
+    void testLongsScalarEqualsVector()
+    {
+        for (int length : TEST_LENGTHS) {
+            for (int offset : TEST_OFFSETS) {
+                long[] values = randomLongs(offset + length);
+                for (boolean[] isNull : getIsNullArray(offset + length)) {
+                    byte[] scalar = compressLongsScalar(values, isNull, offset, length);
+                    byte[] vector = compressLongsVectorized(values, isNull, offset, length);
+                    assertThat(vector).as("longs: scalar and vector outputs differ").isEqualTo(scalar);
+                }
+            }
+        }
+    }
+
+    static boolean[][] getIsNullArray(int length)
+    {
+        return new boolean[][] {
+                all(false, length),
+                all(true, length),
+                alternating(length),
+                randomBooleans(length)};
+    }
+
+    static byte[] compressBytesScalar(byte[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Byte.BYTES + 4));
+        compactBytesWithNullsScalar(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] compressBytesVectorized(byte[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Byte.BYTES + 4));
+        compactBytesWithNullsVectorized(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] compressShortsScalar(short[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Short.BYTES + 4));
+        compactShortsWithNullsScalar(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] compressShortsVectorized(short[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Short.BYTES + 4));
+        compactShortsWithNullsVectorized(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] compressIntsScalar(int[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Integer.BYTES + 4));
+        compactIntsWithNullsScalar(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] compressIntsVectorized(int[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Integer.BYTES + 4));
+        compactIntsWithNullsVectorized(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    static byte[] compressLongsScalar(long[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Long.BYTES + 4));
+        compactLongsWithNullsScalar(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] compressLongsVectorized(long[] values, boolean[] isNull, int offset, int length)
+    {
+        DynamicSliceOutput out = new DynamicSliceOutput(length * (Long.BYTES + 4));
+        compactLongsWithNullsVectorized(out, values, isNull, offset, length);
+        return out.slice().getBytes();
+    }
+
+    private static byte[] randomBytes(int size)
+    {
+        byte[] data = new byte[size];
+        Random r = new Random(RANDOM_SEED);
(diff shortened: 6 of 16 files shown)


core/trino-spi/src/main/java/io/trino/spi/simd/SimdUtils.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /**
+     * Returns true if the Linux host appears to be an AWS Graviton instance.
+     * Heuristics:
+     * - /proc/cpuinfo contains "graviton" (explicit hint on many kernels)
+     * - Otherwise, if the platform vendor indicates AWS/EC2 and CPU info suggests Neoverse cores,
+     *   treat as Graviton (covers Graviton2/3 where the model may be Neoverse N1/V1/V2).
+     */
+    public static boolean isLinuxGraviton()
+    {
```
wendigo: Instead of a custom detection logic we should use first what https://github.com/oshi/oshi supports and only then fallback to custom parsing

SongChujun: If we want to add SIMD detection logic in SPI, I am less sure if we want to use some third party library since  SPI should have very mimimal dependency, and the detection logic is rather simple that we are able to maintain.

wendigo: First of all, I don't think it should live in the SPI in the first place. These kind of optimizations in 99% use cases will live in the trino-main module which depends already on oshi. Having your own implementation, rather than relying on the external one - has its cost that we don't want to have.

wendigo: The only reason why this is needed in the SPI is the fact that BlockEncoding implementation live there but tbh concrete implementations shouldn't be a part of the trino-spi. BlockEncoding's can be a part of the plugin but built-in ones should be just moved to the trino-main.

SongChujun: I agree we indeed need to think about the place where such logic should be placed, we need to guarantee it is accessible from lib(the case for vectorizedDecoding for parquet reader), SPI(tehe code for BlockEncoding, though it is detabale since we may move that to trino-main), trino-main, plugin etc.

pettyjamesm:
Discussed this offline, from here we want to:
1. Add a configuration property to explicitly disable vectorized block encoding logic (in case a performance regression occurs on some specific hardware in the wild)
2. Put the operations that are either vectorized or scalar behind an interface, defined in `trino-spi`, e.g.: `interface BlockEncodingSupport`. The SPI defined implementation will be purely scalar, but the implementation bound from `trino-main` will be either vectorized or scalar depending on the configuration setting and the hardware support detection (which can be based on oshi at that point). @wendigo, does that strategy work for you?

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -14,14 +14,25 @@
 package io.trino.util;
 
 import com.google.common.base.StandardSystemProperty;
+import org.weakref.jmx.$internal.guava.collect.ImmutableSet;
```
pettyjamesm: Wrong import path

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -14,14 +14,25 @@
...
+import oshi.hardware.CentralProcessor.ProcessorIdentifier;
+
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 
 import static java.lang.Math.min;
+import static java.util.Locale.ENGLISH;
+import static org.weakref.jmx.$internal.guava.collect.ImmutableSet.toImmutableSet;
```
pettyjamesm: Wrong import path

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -15,16 +15,38 @@
...
+    private static final VectorSpecies<Long> LONG_SPECIES = LongVector.SPECIES_PREFERRED;
+    private static final VectorSpecies<Integer> INT_SPECIES = IntVector.SPECIES_PREFERRED;
+    private static final VectorSpecies<Byte> BYTE_SPECIES = ByteVector.SPECIES_PREFERRED;
+    private static final VectorSpecies<Short> SHORT_SPECIES = ShortVector.SPECIES_PREFERRED;
+    private static SimdSupport simd = SimdSupport.NONE;
+    private static final int SIMD_OPTIMIZATION_LENGTH_THRESHOLD = 1024;
+
     private EncoderUtil() {}
 
+    public static void setSimdSupport(SimdSupport simdSupport)
```
pettyjamesm: Using a static initializer here is undesirable, let's defer the decision about whether to SIMD each operation to the caller and just make the static methods available for both scalar and vectorized encoding.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -15,16 +15,38 @@
...
 
 import java.io.IOException;
 import java.io.UncheckedIOException;
 import java.util.Optional;
 
-final class EncoderUtil
+import static java.util.Objects.checkFromIndexSize;
+import static java.util.Objects.requireNonNull;
+
+public final class EncoderUtil
```
pettyjamesm: Not sure this class needs to become public if we remove the `setSimdSupport` method, maybe not.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -120,4 +188,200 @@ public static byte[] retrieveNullBits(SliceInput sliceInput, int positionCount)
             throw new UncheckedIOException(e);
         }
     }
+
+    // Bytes
+    public static void compressBytesWithNulls(SliceOutput sliceOutput, byte[] values, boolean[] isNull, int offset, int length)
+    {
+        if (simd.enableBlockSerdeVectorizedNullSuppression() && simd.supportByteGeneric() && simd.supportByteCompress() && length >= SIMD_OPTIMIZATION_LENGTH_THRESHOLD) {
```
pettyjamesm: Let's reorganize this logic by:
1. Moving `simd.enableBlockSerdeVectorizedNullSuppression() && simd.supportByteGeneric()` (and equivalent methods for other primitive types) into the caller context and have it either call the vectorized or non-vectorized method directly
2. Move the check for `length >= SIMD_OPTIMIZATION_LENGTH_THRESHOLD` into the beginning of the vectorized method (to call directly into the scalar implementation)

core/trino-spi/src/main/java/io/trino/spi/SimdSupport.java
```diff
@@ -0,0 +1,127 @@
...
+ * limitations under the License.
+ */
+package io.trino.spi;
+
+/**
+ * Generic SIMD capability interface exposed by SPI.
+ * Implementations should answer true only when the operation is supported
+ * without emulation on the current CPU/JVM.
+ */
+public interface SimdSupport
```
pettyjamesm: Instead of making this an interface with each CPU architecture defined separately, let's do CPU specific parsing and just emit a simple record class with fields for supported operations, e.g.:

```
public record SimdSupport(
   boolean nativeExpandAndCompressByte,
   boolean nativeExpandAndCompressShort,
   ...
)
```

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
```diff
@@ -0,0 +1,224 @@
...
+import java.util.Random;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Execution(ExecutionMode.SAME_THREAD)
+public class TestEncoderUtil
+{
+    private static final int[] TEST_LENGTHS = {0, 3, 255, 257, 512, 530, 1024, 2048, 8192};
+    private static final int[] TEST_OFFSETS = {0, 2, 256};
+    private static final long seed = 42;
```
pettyjamesm: Nit: `RANDOM_SEED` (capitalize constants)

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
```diff
@@ -0,0 +1,224 @@
...
+    private static boolean[] alternating(int size)
+    {
+        boolean[] out = new boolean[size];
+        for (int i = 0; i < size; i++) {
+            out[i] = (i % 2) == 0;
+        }
+        return out;
+    }
+
+    private static boolean[] randomBools(int size)
```
pettyjamesm: Nit: avoid abbreviations, so `randomBooleans`

core/trino-spi/src/test/java/io/trino/spi/block/TestingBlockEncodingSerde.java
```diff
@@ -47,10 +47,10 @@ public TestingBlockEncodingSerde(Function<TypeId, Type> types)
         this.types = requireNonNull(types, "types is null");
         // add the built-in BlockEncodings
         addBlockEncoding(new VariableWidthBlockEncoding());
-        addBlockEncoding(new ByteArrayBlockEncoding());
-        addBlockEncoding(new ShortArrayBlockEncoding());
-        addBlockEncoding(new IntArrayBlockEncoding());
-        addBlockEncoding(new LongArrayBlockEncoding());
+        addBlockEncoding(new ByteArrayBlockEncoding(false));
```
pettyjamesm: Let's prefer the vectorized implementation by default, even though the testing hardware may not support the specific instruction sets we're interested in. We'll rely on the JVM fallback implementations under the assumption that those should match the semantics of vectorization (even though they'll perform worse) and give us better coverage of the relevant codepaths. (also add an inline comment in the code to note this assumption).

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
```diff
@@ -133,6 +134,12 @@ public enum DataIntegrityVerification
         /**/;
     }
 
+    public enum BlockSerdeVectorizedNullSuppressionStrategy
```
pettyjamesm: Seems like we could get away with a boolean here since we're not going to allow force-enabling beyond auto-detect.

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -52,4 +63,74 @@ public static int getAvailablePhysicalProcessorCount()
...
+    {
+        flag = flag.toLowerCase(ENGLISH).replace("_", "").trim();
+
+        // Skip stray keys that may sneak in if the colon wasn’t found.
+        if (flag.equals("flags") || flag.equals("features")) {
+            return "";
+        }
+
+        // Simple ARM/Graviton aliasing (Linux often reports "asimd" for NEON).
+        if (isArm) {
```
pettyjamesm: Seems like we could handle this in the caller side by just searching for both `"asimd"` and `"neon"` in the CPU flags

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,97 @@
...
+import oshi.hardware.CentralProcessor.ProcessorIdentifier;
+
+import java.util.Set;
+
+import static io.trino.util.MachineInfo.normalizeFlag;
+import static io.trino.util.MachineInfo.readCpuFlags;
+import static java.util.Locale.ENGLISH;
+
+@Singleton
+public final class SimdSupportManager
```
pettyjamesm: I think we can get rid of this class and do the following:
1. Move the logic from `detectSimd()` into a static method on `SimdSupport` (returning a `SimdSupport` instance).
2. Switch `ServerMainModule` to bind: `binder.bind(SimdSupport.class).toInstance(SimdSupport.detectSupport())`;
3. Take both `FeaturesConfig` and `SimdSupport` as injection parameters in `BlockEncodingManager`.

pettyjamesm: Also, maybe just add a constant to `SimdSupport` for `DETECTED` so that we have a static singleton instance that shows the result of auto-detection for debugging and avoiding accidentally detecting multiple times.

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+        }
+
+        return SimdSupport.NONE;
+    }
+
+    private SimdSupport detectX86SimdSupport()
+    {
+        Set<String> flags = readCpuFlags(false);
+
+        boolean[] has = new boolean[X86Isa.values().length];
```
pettyjamesm: EnumSet<X86Isa> x86Flags = new EnumSet<>(x86Isa.class);

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+        if (vendor.contains("intel") || vendor.contains("amd")) {
+            return detectX86SimdSupport();
+        }
+
+        return SimdSupport.NONE;
+    }
+
+    private SimdSupport detectX86SimdSupport()
+    {
+        Set<String> flags = readCpuFlags(false);
```
pettyjamesm: Remove `isArm`

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+        boolean[] has = new boolean[X86Isa.values().length];
+
+        if (!flags.isEmpty()) {
+            for (X86Isa isa : X86Isa.values()) {
+                String token = normalizeFlag(isa.name(), false);
```
pettyjamesm: Add field to `X86Isa` to avoid depending on the logic in `normalizeFlag` to establish the search string.

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
pettyjamesm: Resolved base on offline conversation: detection must happen from `trino-main` and not `trino-spi` so we can't move the logic into `SimdSupport`.

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+import java.util.Set;
+
+import static io.trino.FeaturesConfig.BlockSerdeVectorizedNullSuppressionStrategy.AUTO;
+import static io.trino.util.MachineInfo.normalizeFlag;
+import static io.trino.util.MachineInfo.readCpuFlags;
+import static java.util.Locale.ENGLISH;
+
+@Singleton
+public final class SimdSupportManager
```
pettyjamesm: Rename: `BlockEncodingSimdSupport`

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+{
+    public static final int MINIMUM_SIMD_LENGTH = 512;
+    private final SimdSupport simdSupport;
+    private final FeaturesConfig featuresConfig;
+
+    public static final SimdSupportManager TESTING_SIMD_SUPPORT_MANAGER = new SimdSupportManager(new FeaturesConfig().setBlockSerdeVectorizedNullSuppressionStrategy(AUTO));
+
+    private enum X86Isa {
+        AVX2,
```
pettyjamesm: Remove

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+
+    @Inject
+    public SimdSupportManager(
+            FeaturesConfig featuresConfig)
+    {
+        this.featuresConfig = featuresConfig;
+        simdSupport = detectSimd();
+    }
+
+    private SimdSupport detectSimd()
```
pettyjamesm: Nit: static

core/trino-main/src/main/java/io/trino/simd/SimdSupportManager.java
```diff
@@ -0,0 +1,110 @@
...
+        AVX512F,
+        AVX512VBMI2
+    }
+
+    @Inject
+    public SimdSupportManager(
+            FeaturesConfig featuresConfig)
+    {
+        this.featuresConfig = featuresConfig;
+        simdSupport = detectSimd();
```
pettyjamesm: Consider short-circuit when `featuresConfig.getBlockSerdeVectorizedNullSuppressionStrategy() == NONE`

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
SongChujun: discussed offline that we can use enum here because the semantics is richer than true/false

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -52,4 +63,67 @@ public static int getAvailablePhysicalProcessorCount()
...
+            Set<String> tokens = Arrays.stream(body.trim().split("\\s+"))
+                    .map(token -> normalizeFlag(token))
+                    .filter(token -> !token.isEmpty())
+                    .collect(toImmutableSet());
+
+            if (tokens.isEmpty()) {
+                continue;
+            }
+
+            if (intersection == null) {
```
pettyjamesm, line 105: What's going on with the intersection logic here? Is line of flags the hardware support for an individual core and we're looking for only flags advertised by all cores? If so, let's add an inline comment to that effect to make this easier to read.

SongChujun: sure will do

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,96 @@
...
+        Set<String> flags = readCpuFlags();
+        EnumSet<X86Isa> x86Flags = EnumSet.noneOf(X86Isa.class);
+
+        if (!flags.isEmpty()) {
+            for (X86Isa isa : X86Isa.values()) {
+                if (flags.contains(isa.name())) {
+                    x86Flags.add(isa);
+                }
+            }
+        }
```
dain, line 99: I'd just use the same technique we used in `io.trino.parquet.reader.ColumnReaderFactory#isVectorizedDecodingSupported`.  My memory is that from testeing they found that ARM Neon was slower, so basincally we just disable vector instructions unless `PREFERRED_BIT_WIDTH >= 256`.

core/trino-spi/pom.xml
```diff
@@ -319,6 +319,34 @@
                                     <old>method long[] io.trino.spi.PageSorter::sort(java.util.List&lt;io.trino.spi.type.Type&gt;, java.util.List&lt;io.trino.spi.Page&gt;, java.util.List&lt;java.lang.Integer&gt;, java.util.List&lt;io.trino.spi.connector.SortOrder&gt;, int)</old>
                                     <new>method java.util.Iterator&lt;io.trino.spi.Page&gt; io.trino.spi.PageSorter::sort(java.util.List&lt;io.trino.spi.type.Type&gt;, java.util.List&lt;io.trino.spi.Page&gt;, java.util.List&lt;java.lang.Integer&gt;, java.util.List&lt;io.trino.spi.connector.SortOrder&gt;, int)</new>
                                 </item>
+                                <item>
```
dain, line 266: Ideally, I'd like to do this change without messing with the SPI.  My experience witht he vector stuff is that it either works or doesn't, so maybe we can just have a global kill switch.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,96 @@
...
+        if (vendor.contains("intel") || vendor.contains("amd")) {
+            return detectX86SimdSupport();
+        }
+
+        return SimdSupport.NONE;
+    }
+
+    private static SimdSupport detectX86SimdSupport()
+    {
+        enum X86Isa {
```
pettyjamesm: Nit: These are "instruction sets" but not "instruction set architectures". Maybe rename to `Avx512InstructionSet`.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,96 @@
...
+import static io.trino.util.MachineInfo.readCpuFlags;
+import static java.util.Locale.ENGLISH;
+
+@Singleton
+public final class BlockEncodingSimdSupport
+{
+    public static final int MINIMUM_SIMD_LENGTH = 512;
+    private final SimdSupport simdSupport;
+
+    public static final BlockEncodingSimdSupport TESTING_BLOCK_ENCODING_SIMD_SUPPORT = new BlockEncodingSimdSupport(new FeaturesConfig().setBlockSerdeVectorizedNullSuppressionStrategy(AUTO));
```
pettyjamesm: Minor: this can be simplified by splitting the constructor into two:
1. the current constructor, annotated with `@Inject` that takes `FeaturesConfig` which calls the second constructor
2. a constructor that takes only `boolean enableAutoDetectedSimdSupport`

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,96 @@
...
+
+    @Inject
+    public BlockEncodingSimdSupport(
+            FeaturesConfig featuresConfig)
+    {
+        simdSupport = detectSimd(featuresConfig);
+    }
+
+    private static SimdSupport detectSimd(FeaturesConfig featuresConfig)
```
pettyjamesm: 1. With the constructor changes above and can remove the `FeaturesConfig` argument altogether, just perform the auto-detection
2. Cache the result of auto-detection as a `private static final SimdSupport AUTO_DETECTED_SUPPORT = detectSimdSupport();`
3. Switch between `SimdSupport.NONE` and `AUTO_DETECTED_SUPPORT` in the constructor (which saves repeated parsing and checks in test environments).

core/trino-main/src/test/java/io/trino/server/remotetask/TestHttpRemoteTask.java
```diff
@@ -676,6 +679,9 @@ public void configure(Binder binder)
                         jsonCodecBinder(binder).bindJsonCodec(TaskUpdateRequest.class);
                         jsonCodecBinder(binder).bindJsonCodec(FailTaskRequest.class);
 
+                        configBinder(binder).bindConfig(FeaturesConfig.class);
+
+                        binder.bind(BlockEncodingSimdSupport.class).in(SINGLETON);
```
pettyjamesm: With the constructor simplification, we can `binder.bind(BlockEncodingSimdSupport.class).toInstance(new BlockEncodingSimdSupport(true));`

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
pettyjamesm: Discussed offline, for this purpose we need specifically to detect AVX512F (for `VPCOMPRESSD` / `VPCOMPRESSQ` instruction support for `int` and `long` types) and AVX512VBMI2 (for `VPCOMPRESSB` / `VPCOMPRESSW` instruction support over `byte` and `short` types).

1. AVX512F is fairly widely supported starting in Intel Xeon Skylake CPUs and Zen 4 CPUs
2. AVX512 VBMI2 support starts in Intel Icelake and AMD Zen 4 CPUs

The real thing we would _like_ to check here is whether `Vector<T>#compress(VectorMask<T>)` is supported natively in hardware or emulated by the JVM- because the emulated support is so much slower than the simple scalar code that exists, but since we don't have the ability to detect that directly from the JDK vector API we have to assume that native support exists whenever the CPU advertises it

core/trino-spi/pom.xml
pettyjamesm: That's more or less what you're looking at here. Detection for what the hardware supports has to happen from within `trino-main` and then get passed as a `boolean` parameter to the block encoding constructors. Hardware detection is combined with a kill switch to disable the feature if needed- but the block encoder constructors need to change either way to accept the result of config + hardware detection.

pettyjamesm:
Changes and approach overall look good to me at this point, with the following items gating final approval / merge:
1. Cleanup the commit history so that we don't introduce `SimdSupport` under `trino-spi` in the first commit only to move it in the second commit back to `trino-main`.
2. Get additional maintainer review / approval since this is a significant extension to the current vector API usage in Trino. cc: @raunaqmorarka / @dain

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
raunaqmorarka: I don't think it should be an Enum if there are only two possible values.
In the parquet reader also we have an auto detection logic that may decide to skip Vector API even if the flag is enabled, that didn't justify turning it into an Enum

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
```diff
@@ -366,6 +373,19 @@ public FeaturesConfig setExchangeCompressionCodec(CompressionCodec exchangeCompr
         return this;
     }
 
+    @Config("experimental.blockserde-vectorized-null-suppression-strategy")
```
raunaqmorarka: `exchange.experimental.vectorized-serde.enabled`
blockserde...null-suppression implies a very specific thing, we may want to use Vector API for more than just that
having a more generic name keeps the door open to adding more usage under same flag and is easier for users to grasp

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
```diff
@@ -366,6 +373,19 @@ public FeaturesConfig setExchangeCompressionCodec(CompressionCodec exchangeCompr
...
+    @ConfigDescription("Strategy used for vectorized null suppression in block serde")
```
raunaqmorarka: Enable using Java Vector API for faster serialization and deserialization of exchange data

core/trino-main/src/test/java/io/trino/execution/buffer/BenchmarkBlockSerde.java
```diff
@@ -81,7 +81,7 @@ public class BenchmarkBlockSerde
 {
     private static final DecimalType LONG_DECIMAL_TYPE = createDecimalType(30, 5);
 
-    public static final int ROWS = 10_000_000;
+    public static final int ROWS = 8192;
```
raunaqmorarka: I don't follow why this is better.
The actual size of pages produced in benchmark setup is decided by `io.trino.spi.PageBuilder#isFull` and that is unchanged. But we have now reduced the total number of rows generated as input for the benchmark, I think this will cut input of the benchmark to one 8K Page. Wouldn't we want to run the JMH for at least a few dozen pages ?

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    private static SimdSupport detectSimd()
+    {
+        ProcessorIdentifier id = MachineInfo.getProcessorInfo();
+
+        String vendor = id.getVendor().toLowerCase(ENGLISH);
+
+        if (vendor.contains("intel") || vendor.contains("amd")) {
```
raunaqmorarka, line 76: Lack of ARM support is a bummer. Graviton machines do so much better in the benchmarks that its heavily favored in production deployments at the moment and an incremental gain from better vectorization on intel wouldn't move the needle in the other direction.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,117 @@
...
+                    x86Flags.add(instructionSet);
+                }
+            }
+        }
+
+        return new SimdSupport(
+                (ByteVector.SPECIES_PREFERRED.vectorBitSize() >= MINIMUM_SIMD_LENGTH) && x86Flags.contains(X86SimdInstructionSet.avx512vbmi2),
+                (ShortVector.SPECIES_PREFERRED.vectorBitSize() >= MINIMUM_SIMD_LENGTH) && x86Flags.contains(X86SimdInstructionSet.avx512vbmi2),
+                (IntVector.SPECIES_PREFERRED.vectorBitSize() >= MINIMUM_SIMD_LENGTH) && x86Flags.contains(X86SimdInstructionSet.avx512f),
+                (LongVector.SPECIES_PREFERRED.vectorBitSize() >= MINIMUM_SIMD_LENGTH) && x86Flags.contains(X86SimdInstructionSet.avx512f));
```
raunaqmorarka: It seems overkill to add a separate flag per byte species.
I think we should simplify this to check only for `AVX512VBMI2` support as that is an extension on `AVX512F`, and fall back to scalar if AVX512VBMI2 support is not found.
I would also do the vector bit size in a similar way as `io.trino.parquet.reader.ColumnReaderFactory#getVectorBitSize` to ensure that we don't get failures when the JVM flag for enabling vector preview is not enabled.

core/trino-spi/pom.xml
raunaqmorarka: Instead of boolean parameter to existing BlockEncoding classes, couldn't you create new Vector...BlockEncoding classes and register them as
```
if (vectorEnabled) {
     addBlockEncoding(new ByteArrayBlockEncoding());
}
else {
     addBlockEncoding(new VectorByteArrayBlockEncoding());
}
```

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
```diff
@@ -0,0 +1,224 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+import java.util.Random;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Execution(ExecutionMode.SAME_THREAD)
+public class TestEncoderUtil
```
raunaqmorarka: public -> final

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
```diff
@@ -0,0 +1,224 @@
...
+{
+    private static final int[] TEST_LENGTHS = {0, 3, 255, 257, 512, 530, 1024, 2048, 8192};
+    private static final int[] TEST_OFFSETS = {0, 2, 256};
+    private static final long RANDOM_SEED = 42;
+
+    @Test
+    public void testBytesScalarEqualsVector()
```
raunaqmorarka: please drop `public` on the test methods per https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations
`Test classes should be defined as package-private and final.`

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -120,4 +134,156 @@ public static byte[] retrieveNullBits(SliceInput sliceInput, int positionCount)
             throw new UncheckedIOException(e);
         }
     }
+
+    static void compressBytesWithNullsVectorized(SliceOutput sliceOutput, byte[] values, boolean[] isNull, int offset, int length)
```
raunaqmorarka: Naming these methods with "compress" prefix is a bit confusing since this is not actual compression, it's just compacting the values list to remove nulls.
Using "write" or "compact" as the prefix would be more accurate.

core/trino-main/src/main/java/io/trino/metadata/BlockEncodingManager.java
```diff
@@ -41,14 +45,19 @@ public final class BlockEncodingManager
     // for serialization
     private final Map<Class<? extends Block>, BlockEncoding> blockEncodingNamesByClass = new ConcurrentHashMap<>();
 
-    public BlockEncodingManager()
+    public static final BlockEncodingManager TESTING_BLOCK_ENCODING_MANAGER = new BlockEncodingManager(TESTING_BLOCK_ENCODING_SIMD_SUPPORT);
```
piotrrzysko: This is used only in tests and in `PlanTester`. Can we avoid mixing production and test code?

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,119 @@
...
+    {
+        public static final SimdSupport NONE = new SimdSupport(false, false, false, false);
+        public static final SimdSupport ALL = new SimdSupport(true, true, true, true);
+    }
+
+    public static final int MINIMUM_SIMD_LENGTH = 512;
+    private final SimdSupport simdSupport;
+    private static final SimdSupport AUTO_DETECTED_SUPPORT = detectSimd();
+
+    public static final BlockEncodingSimdSupport TESTING_BLOCK_ENCODING_SIMD_SUPPORT = new BlockEncodingSimdSupport(true);
```
piotrrzysko: Same comment as for `BlockEncodingManager.TESTING_BLOCK_ENCODING_MANAGER`.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: We do plan to add support for ARM, we want to use this PR as a starting PR, and gradually add support for other CPUs.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: The thing is even if ```AVX512VBMI2``` support is not found, Integers and Longs are still able to benefit from SIMD optimization, it's just Bytes and Shorts are not able to be vectorized without ```AVX512VBMI2```, so we want to add support for Byte, Short, Integer, Long seperately.

Will add a similar function like ```io.trino.parquet.reader.ColumnReaderFactory#getVectorBitSize``` to ensure that we don't get failures when the JVM flag for enabling vector preview is not enabled

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
SongChujun: Sure, will change the value

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
raunaqmorarka: I understand that presence of AVX512F is useful to optimize some cases. My point was about whether that level of complexity is actually warranted for this optimization. The more platform/CPU specific code paths we have, the harder testing and future maintenance becomes. Unless the gains are drastic, I'd rather the enabling conditions be more straightforward.

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
SongChujun: For the comment on adding ```VectorByteArrayBlockEncoding```, the thing is we are only optimizing a part of code for ```ByteArrayBlockEncoding``` now, if we create a another class called ```VectorByteArrayBlockEncoding```, it will share a lot of code with ```ByteArrayBlockEncoding``` which casues code duplicates issues. So we want to keep ```ByteArrayBlockEncoding``` now.

In the future if we changes ByteArrayBlockEncoding entirely to use SIMD, we can create a separate class though

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
SongChujun: will do

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
SongChujun: Will change the prefix to use compact

core/trino-main/src/main/java/io/trino/metadata/BlockEncodingManager.java
SongChujun: The similar practice is also used for ```TESTING_TYPE_MANAGER``` in ```InternalTypeManager```, so I just follow the practice there.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: Same as above comment

core/trino-main/src/test/java/io/trino/execution/buffer/BenchmarkBlockSerde.java
SongChujun: The thing is when we profile BenchmarkBlockSerde, we find a siginficant amount of time of the benchmark is spent on ```createValues``` call in setup command because it is trying to create a large array, the results are in https://github.com/trinodb/trino/pull/26919#issue-3504719945, which makes the benchmark not being able to correctly reflect the time spend in block serde part. So we want to reduce this value to be a lower value to make sure large array creation doesn't take a significant amount of time.

raunaqmorarka: Anything that happens in the setup of a JMH benchmark does not count towards the benchmarked method's JMH profile result. As for the profile of the JMH benchmark itself, you should ignore everything coming from the setup method in that.
You can change `io.trino.execution.buffer.BenchmarkDataGenerator#createValues` to provide values iteratively instead of allocating a large array.
I think cutting the size of the input to the benchmark to make the setup shorter makes the JMH result less realistic.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
pettyjamesm: In our case, we have a very large fleet that supports AVX512F but not AVX512VBMI2- so this distinction does seem to be worthwhile compared to lumping them together. Whether the gains are significant is a bit subjective, but ~50% improvement for serialization at 50% null-rate seems like a nice win here that justifies the improvement. This will potentially also stack when deserialization is implemented using the vector API too- but that work isn't in this PR yet and will depend on the same extension detection logic.

pettyjamesm: Small note here @raunaqmorarka:

> I would also do the vector bit size in a similar way as io.trino.parquet.reader.ColumnReaderFactory#getVectorBitSize to ensure that we don't get failures when the JVM flag for enabling vector preview is not enabled.

That won't work on this approach. This change is going to make a hard requirement to set that flag, because otherwise the code in `EncoderUtil` that has both vectorized and scalar implementations will fail during classloading. That seems acceptable to me because we probably will want to keep adding extensions that rely on vectorization- but wanted to call that out explicitly.

SongChujun: Thanks for the clarification, good call out!

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
```diff
@@ -94,6 +94,7 @@ public class FeaturesConfig
      * default value is overwritten for fault tolerant execution in {@link #applyFaultTolerantExecutionDefaults()}}
      */
     private CompressionCodec exchangeCompressionCodec = NONE;
+    private boolean blockSerdeVectorizedNullSuppressionEnabled = true;
```
raunaqmorarka: exchangeVectorizedSerdeEnabled

core/trino-main/src/main/java/io/trino/metadata/BlockEncodingManager.java
```diff
@@ -41,14 +45,19 @@ public final class BlockEncodingManager
...
+
+    @Inject
+    public BlockEncodingManager(
+            BlockEncodingSimdSupport blockEncodingSimdSupport)
```
raunaqmorarka: same line

core/trino-spi/pom.xml
```diff
@@ -259,10 +259,74 @@
                                     <justification>Revapi now detects new API changes to vararg args</justification>
                                 </item>
                                 <item>
+                                    <ignore>true</ignore>
```
raunaqmorarka: This section looks unrelated to your changes

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,118 @@
...
+        if (!flags.isEmpty()) {
+            for (X86SimdInstructionSet instructionSet : X86SimdInstructionSet.values()) {
+                if (flags.contains(instructionSet.name())) {
+                    x86Flags.add(instructionSet);
+                }
+            }
+        }
+
+        return new SimdSupport(
+                (ByteVector.SPECIES_PREFERRED.vectorBitSize() >= MINIMUM_SIMD_LENGTH) && x86Flags.contains(X86SimdInstructionSet.avx512vbmi2),
```
raunaqmorarka: Why is the calculation `ByteVector.SPECIES_PREFERRED.vectorBitSize()` per Vector type ?
Don't they all return the same value as `VectorShape.preferredShape().vectorBitSize()` ?
I assume we're looking for bit width of SIMD register on the CPU here, so that should be available from one method call.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,118 @@
...
+            boolean expandAndCompressByte,
+            boolean expandAndCompressShort,
+            boolean expandAndCompressInt,
+            boolean expandAndCompressLong)
+    {
+        public static final SimdSupport NONE = new SimdSupport(false, false, false, false);
+        public static final SimdSupport ALL = new SimdSupport(true, true, true, true);
+    }
+
+    public static final int MINIMUM_SIMD_LENGTH = 512;
```
raunaqmorarka: Please add a code comment explaining the rationale for 512 bit width here.
My observation in the past was that even graviton 3 machines will return `vectorBitSize` of 256 and there is a clear gain at that bit width as well, anything lower was definitely not worth it.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,118 @@
...
+public final class BlockEncodingSimdSupport
+{
+    public record SimdSupport(
+            boolean expandAndCompressByte,
+            boolean expandAndCompressShort,
+            boolean expandAndCompressInt,
+            boolean expandAndCompressLong)
+    {
+        public static final SimdSupport NONE = new SimdSupport(false, false, false, false);
+        public static final SimdSupport ALL = new SimdSupport(true, true, true, true);
```
raunaqmorarka: Do we use `ALL` ?

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,118 @@
...
+
+    public static final int MINIMUM_SIMD_LENGTH = 512;
+    private final SimdSupport simdSupport;
+    private static final SimdSupport AUTO_DETECTED_SUPPORT = detectSimd();
+
+    public static final BlockEncodingSimdSupport TESTING_BLOCK_ENCODING_SIMD_SUPPORT = new BlockEncodingSimdSupport(true);
+
+    @Inject
+    public BlockEncodingSimdSupport(
+            FeaturesConfig featuresConfig)
```
raunaqmorarka: same line

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,118 @@
...
+    {
+        this(featuresConfig.isBlockSerdeVectorizedNullSuppressionEnabled());
+    }
+
+    public BlockEncodingSimdSupport(
+            boolean enableAutoDetectedSimdSupport)
```
raunaqmorarka: same line

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -27,6 +27,12 @@ public class IntArrayBlockEncoding
         implements BlockEncoding
 {
     public static final String NAME = "INT_ARRAY";
+    private final boolean enableVectorizedNullSuppression;
```
raunaqmorarka, line 32: newline between static and non-static member variable declarations

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
raunaqmorarka: Can't you create a separate VectorEncoderUtil class for the code using vector packages and avoid that problem ?

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
```diff
@@ -0,0 +1,224 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+import java.util.Random;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Execution(ExecutionMode.SAME_THREAD)
+public final class TestEncoderUtil
```
raunaqmorarka: drop public

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
```diff
@@ -0,0 +1,224 @@
...
+import io.airlift.slice.DynamicSliceOutput;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.parallel.Execution;
+import org.junit.jupiter.api.parallel.ExecutionMode;
+
+import java.util.Random;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Execution(ExecutionMode.SAME_THREAD)
```
raunaqmorarka: Why "SAME_THREAD" ? I don't see why the tests can't run in parallel

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -120,4 +134,156 @@ public static byte[] retrieveNullBits(SliceInput sliceInput, int positionCount)
...
     }
+
+    static void compactBytesWithNullsVectorized(SliceOutput sliceOutput, byte[] values, boolean[] isNull, int offset, int length)
+    {
+        requireNonNull(sliceOutput, "sliceOutput is null");
+        checkFromIndexSize(offset, length, values.length);
+        checkFromIndexSize(offset, length, isNull.length);
+        byte[] compressed = new byte[length];
+        int valuesIndex = 0;
+        int compressedIndex = 0;
```
raunaqmorarka: Please rename all the "compressed" variables to "compacted" or similar

core/trino-main/src/main/java/io/trino/metadata/BlockEncodingManager.java
raunaqmorarka: Yes, but why was it necessary to repeat that here ?
I don't think we should follow that practice elsewhere, please move this to test only classes

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
pettyjamesm: That might work, but we'd need to verify it- I forget whether calling a static method in a branch not taken will still result in the class loading and failure even with the flag disabled. I suppose the question here is whether we _need_ to do that. Keeping the implementations together has a small benefit to be able to compare the results side-by-side, and maybe the real question is whether we want to definitely keep the vector extension API as fully optional or whether we are comfortable saying that we will in fact use it throughout the engine where appropriate? I can see making that case either way.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
SongChujun: The original implementation used compressed as the variable name so I just kept it, I can change all occurance of compressed to compacted.

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
SongChujun: Will remove the annonation.

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
SongChujun: sure

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: No at this moment, just put it for parity with ```NONE```, I can remove it.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: I can put in in the same line, but curious on which case should we put the parameters in the same line and on which case one parameter per line?

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: will do

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
SongChujun: will do

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: Just want to be conservative since I haven't tested the 256 bit width case. But since you have data point suggesting that 256 will also bring clear gain. I can lower the threshold to 256.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: True, I can unify on it.

core/trino-spi/pom.xml
SongChujun: This appears when I was rebasing against master branch, will investigate.

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
SongChujun: will change

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
raunaqmorarka: I think we should hold off on making addition of the vector module compulsory until this is still in "incubating" phase. Some large, slow moving organisations can be skittish about having that as a requirement to JVM args even if there is a flag to disable its usage internally.
We are still free to add usage of vector API wherever we want, the code in parquet reader around vector usage was also purposely structured in a way that addition of the vector module would not become compulsory.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
raunaqmorarka: If the list of parameters is short, it should be on same line, if it is long then each argument should be on new line

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
SongChujun: Will use @Execution(ExecutionMode.CONCURRENT)

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -0,0 +1,112 @@
...
+        if (!flags.isEmpty()) {
+            for (X86SimdInstructionSet instructionSet : X86SimdInstructionSet.values()) {
+                if (flags.contains(instructionSet.name())) {
+                    x86Flags.add(instructionSet);
+                }
+            }
+        }
+
+        return new SimdSupport(
+                (PREFERRED_BIT_WIDTH >= MINIMUM_SIMD_LENGTH) && x86Flags.contains(X86SimdInstructionSet.avx512vbmi2),
```
raunaqmorarka: May be a bit less verbose to just return `SimdSupport.NONE` when `PREFERRED_BIT_WIDTH < MINIMUM_SIMD_LENGTH`

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -120,4 +134,156 @@ public static byte[] retrieveNullBits(SliceInput sliceInput, int positionCount)
             throw new UncheckedIOException(e);
         }
     }
+
+    static void compactBytesWithNullsVectorized(SliceOutput sliceOutput, byte[] values, boolean[] isNull, int offset, int length)
+    {
+        requireNonNull(sliceOutput, "sliceOutput is null");
```
raunaqmorarka: I don't think the `requireNonNull` here are necessary.
The previous code wasn't doing it and even if it is null, we will still get a clean exception message from the point where it is used. So we can drop it.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -120,4 +134,156 @@ public static byte[] retrieveNullBits(SliceInput sliceInput, int positionCount)
...
+        checkFromIndexSize(offset, length, values.length);
```
raunaqmorarka, line 139: This check is redundant with `checkFromIndexSize(rawOffset, positionCount, rawValues.length)` already happening in BlockEncoding implementations. We can drop these checks.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -120,4 +134,156 @@ public static byte[] retrieveNullBits(SliceInput sliceInput, int positionCount)
...
+        checkFromIndexSize(offset, length, isNull.length);
```
raunaqmorarka, line 140: This is redundant with the checks happening at `io.trino.spi.block.EncoderUtil#encodeNullsAsBits(io.airlift.slice.SliceOutput, boolean[], int, int)`, so we can drop this as well

core/trino-spi/src/test/java/io/trino/spi/block/TestEncoderUtil.java
raunaqmorarka: you don't need to specify `@Execution(ExecutionMode.CONCURRENT)` unless you also specify `@TestInstance(PER_CLASS)`.
per method concurrency is the default

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -59,15 +66,12 @@ public void writeBlock(BlockEncodingSerde blockEncodingSerde, SliceOutput sliceO
             sliceOutput.writeInts(rawValues, rawOffset, positionCount);
         }
         else {
-            int[] valuesWithoutNull = new int[positionCount];
-            int nonNullPositionCount = 0;
-            for (int i = 0; i < positionCount; i++) {
-                valuesWithoutNull[nonNullPositionCount] = rawValues[i + rawOffset];
-                nonNullPositionCount += isNull[i + rawOffset] ? 0 : 1;
+            if (enableVectorizedNullSuppression) {
+                EncoderUtil.compactIntsWithNullsVectorized(sliceOutput, rawValues, isNull, rawOffset, positionCount);
```
raunaqmorarka: Please static import all the EncoderUtil method calls

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
pettyjamesm: I’m not sure that’s the right strategy here. Even though this check should never fail as a result of earlier checks in the block encoding implementations, the JIT needs to “see” the bounds checks happening within the current method scope after inlining. Otherwise, it’s going to have to intersperse bounds checks within the loop body which will hurt performance. If everything gets inlined into the block encoder, then we’ll be ok- but if the compact methods aren’t inlined then this check is load bearing.

Honestly, we could go the other way on this and inline the implementation of checkFromIndexSize into these methods so that we’re not depending on that method getting inlined into this method.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
pettyjamesm: Same reasoning here about ensuring the JIT has enough information to omit bounds checking in the critical path through the loop body.

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
SongChujun: will do

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
raunaqmorarka: Do we still need the `checkFromIndexSize(rawOffset, positionCount, rawValues.length)` in `io.trino.spi.block.BlockEncoding#writeBlock` then ?
That looks unnecessary in the current code

pettyjamesm: I don’t think we necessarily do because the “no nulls” path should have the bounds checking handled inside of SliceOutput (so both paths should have coverage). Having a redundant check there shouldn’t be a performance problem and might give us a little more robustness to future refactors, but I’m fine with removing those too. Up to you @raunaqmorarka

raunaqmorarka (approved):

PR #26922, opened 2025-10-12
Deprecate io.trino.spi.type.Type#appendTo

raunaqmorarka:
## Description
Migrates all existing usages to `io.trino.spi.block.BlockBuilder#append`
This simplifies the Type API and the building of blocks from block positions

## Release notes

```markdown
## SPI
* Deprecate `io.trino.spi.type.Type#appendTo`. ({issue}`26922`)
```

## Summary by Sourcery

Deprecate the SPI Type.appendTo method and migrate all call sites to BlockBuilder.append for streamlined block construction.

Enhancements:
- Mark Type.appendTo as deprecated in the SPI and remove its concrete implementations from built-in types
- Replace all calls to Type.appendTo with BlockBuilder.append using block.getUnderlyingValueBlock() and getUnderlyingValuePosition()
- Simplify code paths and generated methods by removing appendTo parameters and unused type fields in PagesHashStrategy, PagesIndex, spatial index, and related classes
- Update benchmarks and test code to use the new append approach and adjust BenchmarkMapCopy annotations

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/type/Type.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/type/Type.java
@@ -183,9 +183,13 @@ public interface Type
     void writeObject(BlockBuilder blockBuilder, Object value);
 
     /**
-     * Append the value at {@code position} in {@code block} to {@code blockBuilder}.
+     * @deprecated Use {@link BlockBuilder#append}
      */
-    void appendTo(Block block, int position, BlockBuilder blockBuilder);
+    @Deprecated(forRemoval = true)
+    default void appendTo(Block block, int position, BlockBuilder blockBuilder)
+    {
+        blockBuilder.append(block.getUnderlyingValueBlock(), block.getUnderlyingValuePosition(position));
+    }
 
     /**
      * Return the range of possible values for this type, if available.
--- a/core/trino-main/src/main/java/io/trino/type/FunctionType.java
+++ b/core/trino-main/src/main/java/io/trino/type/FunctionType.java
@@ -188,12 +188,6 @@ public class FunctionType
     }
 
     @Override
-    public void appendTo(Block block, int position, BlockBuilder blockBuilder)
-    {
-        throw new UnsupportedOperationException();
-    }
-
-    @Override
     public BlockBuilder createBlockBuilder(BlockBuilderStatus blockBuilderStatus, int expectedEntries, int expectedBytesPerEntry)
     {
         throw new UnsupportedOperationException();
--- a/core/trino-main/src/main/java/io/trino/type/IpAddressType.java
+++ b/core/trino-main/src/main/java/io/trino/type/IpAddressType.java
@@ -127,25 +127,6 @@ public class IpAddressType
     }
 
     @Override
-    public void appendTo(Block block, int position, BlockBuilder blockBuilder)
-    {
-        appendTo(
-                (Int128ArrayBlock) block.getUnderlyingValueBlock(),
-                block.getUnderlyingValuePosition(position),
-                (Int128ArrayBlockBuilder) blockBuilder);
-    }
-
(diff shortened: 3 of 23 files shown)


wendigo:
@raunaqmorarka I suppose that it doesn't affect performance

raunaqmorarka:
> @raunaqmorarka I suppose that it doesn't affect performance

I don't expect it to make a significant difference

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayConcatUtils.java
```diff
@@ -40,8 +42,9 @@ public static Block appendElement(Type elementType, Block block, long value)
     public static Block appendElement(Type elementType, Block block, boolean value)
     {
         BlockBuilder blockBuilder = elementType.createBlockBuilder(null, block.getPositionCount() + 1);
+        ValueBlock valueBlock = block.getUnderlyingValueBlock();
         for (int i = 0; i < block.getPositionCount(); i++) {
-            elementType.appendTo(block, i, blockBuilder);
+            blockBuilder.append(valueBlock, block.getUnderlyingValuePosition(i));
```
dain, line 47: For all of these methods that append position at a time, there is an opportunity to optimize.  We should move `BlockUtil.appendRawBlockRange` to a location usable by these methods.  I'm not suggesting we put this on BlockBuilder right now, but just that we have it somewhere accessible to the main module (for now).  We could also do something in the SPI, but I'm not sure what we would do exactaly.

core/trino-spi/src/main/java/io/trino/spi/type/Type.java
```diff
@@ -183,9 +183,13 @@ default Object getObject(ConnectorSession ignored, Block block, int position)
...
 
     /**
-     * Append the value at {@code position} in {@code block} to {@code blockBuilder}.
+     * @deprecated Use {@link BlockBuilder#append}
      */
-    void appendTo(Block block, int position, BlockBuilder blockBuilder);
+    @Deprecated(forRemoval = true)
+    default void appendTo(Block block, int position, BlockBuilder blockBuilder)
+    {
+        throw new UnsupportedOperationException();
```
dain: I'd have an implementation here:
```
blockBuilder.append(block.getUnderlyingValueBlock(), block.getUnderlyingValuePosition(position));
```

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayConcatUtils.java
raunaqmorarka: Yes, I had noticed that but thought it would be better to do that in a separate PR.
Is it okay to just make `io.trino.spi.block.BlockUtil#appendRawBlockRange` public, or do you think exposing that in SPI has some risk ?

core/trino-spi/src/main/java/io/trino/spi/type/Type.java
raunaqmorarka: done

raunaqmorarka:
> My only other concern, which I did not actually see, is that we still are caching compiled code using the types so we get specialization, even if they type is not actually used in the compiled code... so they cache keys are unchanged.

I think we still want the code to be specialized by Block/BlockBuilder class. Letting the code stay specialized by Type is easier than trying to dispatch based on Block/BlockBuilder class. It is more specific than what we really need, but doesn't hurt us much. I think if we had to do it, we'd need a `Class<? extends Block> io.trino.spi.type.Type#getBlockClass` or something similar to make this easier.

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayConcatUtils.java
raunaqmorarka: Created https://github.com/trinodb/trino/pull/26964

Issue #7103, opened 2021-03-02
ClickHouse connector


findepi:
- [ ] Create connector for ClickHouse #4500
- [ ] Map ClickHouse UUID to Trino UUID #7097 
- [ ] Map ClickHouse IPv4 and IPv6 to Trino IPADDRESS #7098
- [ ] Improve String, FixedString mapping in ClickHouse connector #7102
- [ ] Document ClickHouse connector type mapping #7099
- [ ] More predicate pushdown in ClickHouse connector #7100
- [ ] Fix disabled tests in ClickHouse connector #7101

findepi:
cc @wgzhao

PChou:
Is this connector planned to be implemented based on jdbc?

findepi:
Yes, see https://github.com/trinodb/trino/pull/4909

PChou:
@findepi perfect! That's what we really want!

wgzhao:
@findepi  please add  #7130 into TODO List

wgzhao:
* [x] Support for create table with MergeTree engine in in ClickHouse connector PR #7135

hashhar:
Aggregation pushdown now merged in https://github.com/trinodb/trino/pull/7434.

wgzhao:
* [x] Map ClickHouse UUID to Trino UUID #7915

sondn98:
Should we have a fault-tolerant support for this connector?

serg-kovalev:
> Should we have a fault-tolerant support for this connector?

Are there any plans to have a fault-tolerant support for ClickHouse connector?

PR #26957, opened 2025-10-14
Add MAP type support to the ClickHouse connector

alexjo2144:
## Description

Add `MAP` type support to the ClickHouse connector.
https://clickhouse.com/docs/sql-reference/data-types/map

ClickHouse maps are multi-value, but when I tried that out I only got one value in the client. 

I'll add docs if this looks good to reviewers. 

https://github.com/trinodb/trino/issues/7103

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Add MAP type support to the ClickHouse connector by introducing read and write mappings for MapType, refactoring column mapping logic, and adding relevant tests.

New Features:
- Add support for ClickHouse MAP type in connector, including column mapping and write mapping.

Enhancements:
- Consolidate column mapping logic into a dedicated toColumnMapping method.
- Inject TypeOperators into ClickHouseClient to enable MapType creation.
- Refine decimal and timestamp mapping to use optional decimalDigits and columnSize parameters.

Tests:
- Enable SUPPORTS_MAP_TYPE for ClickHouse connector tests.
- Add round-trip insert tests for MapType with integer and varchar value types.

### DIFF
--- a/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
+++ b/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
@@ -48,6 +48,7 @@ import io.trino.plugin.jdbc.LongReadFunction;
 import io.trino.plugin.jdbc.LongWriteFunction;
 import io.trino.plugin.jdbc.ObjectReadFunction;
 import io.trino.plugin.jdbc.ObjectWriteFunction;
+import io.trino.plugin.jdbc.PreparedQuery;
 import io.trino.plugin.jdbc.QueryBuilder;
 import io.trino.plugin.jdbc.RemoteTableName;
 import io.trino.plugin.jdbc.SliceWriteFunction;
@@ -75,6 +76,8 @@ import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ColumnPosition;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.connector.JoinStatistics;
+import io.trino.spi.connector.JoinType;
 import io.trino.spi.expression.ConnectorExpression;
 import io.trino.spi.expression.Variable;
 import io.trino.spi.type.CharType;
@@ -252,6 +255,13 @@ public class ClickHouseClient
                 .map("$not($is_null(value))").to("value IS NOT NULL")
                 .map("$not(value: boolean)").to("NOT value")
                 .map("$is_null(value)").to("value IS NULL")
+                .map("$equal(left, right)").to("left = right")
+                .map("$not_equal(left, right)").to("left <> right")
+                .map("$less_than(left, right)").to("left < right")
+                .map("$less_than_or_equal(left, right)").to("left <= right")
+                .map("$greater_than(left, right)").to("left > right")
+                .map("$greater_than_or_equal(left, right)").to("left >= right")
+                .map("$operator$subscript(map, value)").to("map[value]")
                 .build();
         this.aggregateFunctionRewriter = new AggregateFunctionRewriter<>(
                 this.connectorExpressionRewriter,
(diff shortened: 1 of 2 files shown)


plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/TestClickHouseConnectorTest.java
```diff
@@ -665,6 +665,24 @@ public void testInsertIntoNotNullColumn()
...
+
+    private void testMapRoundTrip(String valueType, String value)
+    {
+        try (TestTable table = newTrinoTable("test_insert_map_", "(col map(INTEGER, %s) NOT NULL)".formatted(valueType))) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES map(ARRAY[1], ARRAY[%s])".formatted(value), 1);
+            assertThat(query("SELECT col[1] FROM " + table.getName()))
+                    .matches("VALUES " + value);
+        }
+    }
+
```
ebyhr, line 689: Could you please update `BaseClickHouseTypeMapping` as well? We use `***TypeMapping` class to test type mapping in JDBC-based connectors.

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
```diff
@@ -677,11 +701,13 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
                 return Optional.of(varbinaryColumnMapping());
             case UUID:
                 return Optional.of(uuidColumnMapping());
+            case Map:
```
ebyhr, line 746: Could you update `Type mapping` section in clickhouse.md?

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/TestClickHouseConnectorTest.java
```diff
@@ -665,6 +665,24 @@ public void testInsertIntoNotNullColumn()
...
+    public void testInsertMap()
+    {
+        // TODO: Add more types here
+        testMapRoundTrip("INTEGER", "2");
+        testMapRoundTrip("VARCHAR", "CAST('foobar' AS VARCHAR)");
+    }
+
+    private void testMapRoundTrip(String valueType, String value)
+    {
+        try (TestTable table = newTrinoTable("test_insert_map_", "(col map(INTEGER, %s) NOT NULL)".formatted(valueType))) {
```
alexjo2144: This fails without the `NOT NULL` right now, gotta figure that out or at least improve the error message

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/TestClickHouseConnectorTest.java
alexjo-dd: Yep! Added tests for all the primitives at least. Got all of them working except Timestamp w/ TZ, not exactly sure why but the precision for a timestamp in a Map is coming back as 29 but it should be 0.

Going to add some tests for nested maps tomorrow.

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
```diff
@@ -630,14 +644,27 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
     {
         String jdbcTypeName = typeHandle.jdbcTypeName()
                 .orElseThrow(() -> new TrinoException(JDBC_ERROR, "Type name is missing: " + typeHandle));
-
```
krvikash, line 633: unrelated change.

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
```diff
@@ -946,26 +974,30 @@ private ColumnMapping ipAddressColumnMapping(String clickhouseType)
...
-                        arraycopy(address, 0, bytes, 12, 4);
-                    }
-                    else if (address.length == 16) {
-                        bytes = address;
-                    }
-                    else {
-                        throw new TrinoException(GENERIC_INTERNAL_ERROR, "Invalid InetAddress length: " + address.length);
-                    }
-
+                    byte[] bytes = parseIpAddressBytes(address);
```
krvikash, line 1016: Is this refactoring needed for the current PR?

alexjo-dd: In order to avoid copying the code in that method, yeah. The representation is a little different when an ip address is the column's root level type or if it's in a Map for some reason, but either way we get a byte[] which needs to be converted into the Trino representation.

alexjo2144:
Just a couple small tweaks in the last push, getting support for maps of maps working and fixing a few of the round trip type tests.

PR #26959, opened 2025-10-15
Stop running product tests with Databricks 11.3 LTS

ebyhr:
## Description

Databricks 11.3 LTS is going to reach EOL on Oct 19. https://docs.databricks.com/aws/en/release-notes/runtime/

## Release notes

## Summary by Sourcery

Remove end-of-life Databricks 11.3 LTS from live product tests and CI, clean up obsolete version checks, and add lightweight plugin unit support for reading Databricks 11.3 data

Enhancements:
- Drop DatabricksVersion constant and conditional branches for runtime version 11.3 in product tests
- Omit all suite definitions, CI jobs, and test groups targeting Databricks 11.3 LTS
- Add a static plugin resource and a TestDeltaLakeBasic unit test to verify reading region data from Databricks 11.3 LTS

CI:
- Remove suite-delta-lake-databricks113 job and related environment variables from GitHub Actions CI


plugin/trino-delta-lake/src/test/resources/databricks113/region/README.md
```diff
@@ -0,0 +1,13 @@
+Data generated using Databricks 11.3:
+
+```sql
+CREATE TABLE default.region (regionkey bigint, name string, comment string)
```
chenjian2664, line 4: do we still needs those files?

ebyhr: Yes because tables created by Databricks 11.3 should be readable even after the EOL.

PR #26963, opened 2025-10-15
Add support for TIMESTAMP type in exasol connector

skyglass:
## Description

Added Exasol Trino connector support for **Timestamp** JDBC data type

- Exasol TIMESTAMP type is mapped to Trino TIMESTAMP type
- added correspondent tests for **timestamp** with precision data type
- added tests for DST gap and DST overlap times in "America/Bahia_Bandera" JVM time zone

## Release notes

```markdown
## Exasol Connector
* Add support for `timestamp` type. 
```

### DIFF
--- a/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
+++ b/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
@@ -30,6 +30,8 @@ import io.trino.plugin.jdbc.JdbcTableHandle;
 import io.trino.plugin.jdbc.JdbcTypeHandle;
 import io.trino.plugin.jdbc.LongReadFunction;
 import io.trino.plugin.jdbc.LongWriteFunction;
+import io.trino.plugin.jdbc.ObjectReadFunction;
+import io.trino.plugin.jdbc.ObjectWriteFunction;
 import io.trino.plugin.jdbc.QueryBuilder;
 import io.trino.plugin.jdbc.RemoteTableName;
 import io.trino.plugin.jdbc.SliceReadFunction;
@@ -44,12 +46,18 @@ import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ColumnPosition;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorTableMetadata;
+import io.trino.spi.type.LongTimestamp;
+import io.trino.spi.type.TimestampType;
 import io.trino.spi.type.Type;
 
 import java.sql.Connection;
 import java.sql.Date;
+import java.sql.PreparedStatement;
+import java.sql.SQLException;
+import java.sql.Timestamp;
 import java.sql.Types;
 import java.time.LocalDate;
+import java.time.LocalDateTime;
 import java.util.HexFormat;
 import java.util.List;
 import java.util.Map;
@@ -59,20 +67,28 @@ import java.util.Set;
 import java.util.function.BiFunction;
 import java.util.function.Consumer;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static io.trino.plugin.jdbc.PredicatePushdownController.FULL_PUSHDOWN;
 import static io.trino.plugin.jdbc.StandardColumnMappings.bigintColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.booleanColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.decimalColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultCharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultVarcharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.doubleColumnMapping;
+import static io.trino.plugin.jdbc.StandardColumnMappings.fromLongTrinoTimestamp;
+import static io.trino.plugin.jdbc.StandardColumnMappings.fromTrinoTimestamp;
 import static io.trino.plugin.jdbc.StandardColumnMappings.integerColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.smallintColumnMapping;
+import static io.trino.plugin.jdbc.StandardColumnMappings.toLongTrinoTimestamp;
+import static io.trino.plugin.jdbc.StandardColumnMappings.toTrinoTimestamp;
 import static io.trino.plugin.jdbc.TypeHandlingJdbcSessionProperties.getUnsupportedTypeHandling;
 import static io.trino.plugin.jdbc.UnsupportedTypeHandling.CONVERT_TO_VARCHAR;
 import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static io.trino.spi.connector.ConnectorMetadata.MODIFYING_ROWS_MESSAGE;
 import static io.trino.spi.type.DateType.DATE;
 import static io.trino.spi.type.DecimalType.createDecimalType;
+import static io.trino.spi.type.TimestampType.MAX_SHORT_PRECISION;
+import static io.trino.spi.type.TimestampType.createTimestampType;
 import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
@@ -85,6 +101,8 @@ public class ExasolClient
             .add("EXA_STATISTICS")
             .add("SYS")
             .build();
+
+    private static final int MAX_EXASOL_TIMESTAMP_PRECISION = 9;
 
     @Inject
     public ExasolClient(
@@ -243,7 +261,10 @@ public class ExasolClient
                 // https://docs.exasol.com/db/latest/sql/select.htm#UsageNotes
                 yield Optional.of(defaultVarcharColumnMapping(typeHandle.requiredColumnSize(), true));
             }
+            // DATE and TIMESTAMP types are described here in more details:
+            // https://docs.exasol.com/db/latest/sql_references/data_types/datatypedetails.htm
             case Types.DATE -> Optional.of(dateColumnMapping());
+            case Types.TIMESTAMP -> Optional.of(timestampColumnMapping(typeHandle));
             default -> {
                 if (getUnsupportedTypeHandling(session) == CONVERT_TO_VARCHAR) {
                     yield mapToUnboundedVarchar(typeHandle);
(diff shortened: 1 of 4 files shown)


plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -307,6 +293,128 @@ private static SliceWriteFunction hashTypeWriteFunction()
...
+                objectTimestampReadFunction(timestampType),
+                objectTimestampWriteFunction(timestampType),
+                FULL_PUSHDOWN);
+    }
+
+    private static LongReadFunction longTimestampReadFunction(TimestampType timestampType)
+    {
+        return (resultSet, columnIndex) -> {
+            Timestamp timestamp = resultSet.getTimestamp(columnIndex);
+            return toTrinoTimestamp(timestampType, timestamp.toLocalDateTime());
```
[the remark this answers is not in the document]
skyglass: Null check is done by trino earlier.
The test confirms that NULL value is handled correctly:
.addRoundTrip("timestamp", "NULL", createTimestampType(3), "CAST(NULL AS TIMESTAMP)")

skyglass:
> * Instead of hardcoding MAX_EXASOL_TIMESTAMP_PRECISION = 9 in ExasolClient, consider reusing the max precision constant from TimestampType (e.g. TimestampType.MAX_PRECISION) or centralizing it to avoid duplication.

MAX_EXASOL_TIMESTAMP_PRECISION is 9, butTimestampType.MAX_PRECISION is 12, so we have to create a separate constant.
We still need to check for precision <= MAX_EXASOL_TIMESTAMP_PRECISION, because there shouldn't be any precision more than 9. It would mean some programmatic error, because all data comes from Exasol and Exasol doesn't support TIMESTAMP columns with precision > 9

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -307,6 +328,128 @@ private static SliceWriteFunction hashTypeWriteFunction()
...
+                    FULL_PUSHDOWN);
+        }
+        return ColumnMapping.objectMapping(
+                timestampType,
+                objectTimestampReadFunction(timestampType),
+                objectTimestampWriteFunction(timestampType),
+                FULL_PUSHDOWN);
+    }
+
+    private static LongReadFunction longTimestampReadFunction(TimestampType timestampType)
```
chenjian2664, line 353: Please verify the timestamp precision is in the expected range

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -307,6 +328,128 @@ private static SliceWriteFunction hashTypeWriteFunction()
...
+    {
+        return (resultSet, columnIndex) -> {
+            Timestamp timestamp = resultSet.getTimestamp(columnIndex);
+            return toTrinoTimestamp(timestampType, timestamp.toLocalDateTime());
+        };
+    }
+
+    private static LongWriteFunction longTimestampWriteFunction(TimestampType timestampType)
+    {
+        return new LongWriteFunction()
```
chenjian2664, line 366: same here

chenjian2664 (commented):
Looks good

chenjian2664:
@ebyhr

skyglass:
> * In both LongWriteFunction and ObjectWriteFunction setNull methods you’re calling statement.setNull(index, Types.VARCHAR), but since this is a TIMESTAMP column you should use Types.TIMESTAMP to avoid unintended type conversions.
> 

> * In testUnsupportedInsertValue you hardcode the table name to test_unsupported_hashtype; switch to a per-test randomNameSuffix (like in testUnsupportedDefinition) to prevent conflicts when running tests in parallel.


Thank you very much for your review findings! 👍 
These changes have been applied in the updated PR

skyglass:
> Looks good

Thank you very much for your review, @chenjian2664 ! 👍 
The requested changes have been applied, including review findings from **sourcery-ai**

chenjian2664 (approved):

chenjian2664:
@ebyhr @Praveen2112

skyglass:
@ebyhr @Praveen2112  any updates on the PR?
All review findings have been resolved, can we merge it now?

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +279,112 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("timestamp(9)", "TIMESTAMP '2013-03-11 17:30:15.123456789'", createTimestampType(9), "TIMESTAMP '2013-03-11 17:30:15.123456789'")
+                .addRoundTrip("timestamp(1)", "TIMESTAMP '2016-08-19 19:28:05.0'", createTimestampType(1), "TIMESTAMP '2016-08-19 19:28:05.0'")
+                .addRoundTrip("timestamp(2)", "TIMESTAMP '2016-08-19 19:28:05.01'", createTimestampType(2), "TIMESTAMP '2016-08-19 19:28:05.01'")
+                .addRoundTrip("timestamp", "TIMESTAMP '3030-03-03 12:34:56.123'", createTimestampType(3), "TIMESTAMP '3030-03-03 12:34:56.123'")
+                .addRoundTrip("timestamp(4)", "TIMESTAMP '3030-03-03 12:34:56.1234'", createTimestampType(4), "TIMESTAMP '3030-03-03 12:34:56.1234'")
+                .addRoundTrip("timestamp(5)", "TIMESTAMP '3030-03-03 12:34:56.12345'", createTimestampType(5), "TIMESTAMP '3030-03-03 12:34:56.12345'")
+                .addRoundTrip("timestamp(9)", "TIMESTAMP '3030-03-03 12:34:56.123456789'", createTimestampType(9), "TIMESTAMP '3030-03-03 12:34:56.123456789'")
```
chenjian2664: This line seems duplicate with line 326

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +279,112 @@ private void testDate(ZoneId sessionZone)
...
+                .addRoundTrip("timestamp", "TIMESTAMP '2018-10-28 01:13:55.12345'", createTimestampType(3), "TIMESTAMP '2018-10-28 01:13:55.123'")
+                .addRoundTrip("timestamp", "TIMESTAMP '2018-10-28 03:33:33.333'", createTimestampType(3), "TIMESTAMP '2018-10-28 03:33:33.333'")
+                .addRoundTrip("timestamp", "TIMESTAMP '1970-01-01 00:13:42.000'", createTimestampType(3), "TIMESTAMP '1970-01-01 00:13:42.000'")
+                .addRoundTrip("timestamp", "TIMESTAMP '2020-09-27 12:34:56.999'", createTimestampType(3), "TIMESTAMP '2020-09-27 12:34:56.999'")
+                .addRoundTrip("timestamp", "TIMESTAMP '2018-03-25 03:17:17.000'", createTimestampType(3), "TIMESTAMP '2018-03-25 03:17:17.000'")
+                .addRoundTrip("timestamp", "TIMESTAMP '1986-01-01 00:13:07.000'", createTimestampType(3), "TIMESTAMP '1986-01-01 00:13:07.000'")
+                .addRoundTrip("timestamp(6)", "TIMESTAMP '2013-03-11 17:30:15.123456'", createTimestampType(6), "TIMESTAMP '2013-03-11 17:30:15.123456'")
+                .addRoundTrip("timestamp(6)", "TIMESTAMP '2013-03-11 17:30:15.123456789'", createTimestampType(6), "TIMESTAMP '2013-03-11 17:30:15.123456'")
```
chenjian2664: shouldn't the cast behavior is round? https://trino.io/docs/current/language/types.html#timestamp-p 
```
Casting to lower precision causes the value to be rounded, and not truncated. Casting to higher precision appends zeros for the additional digits.
```

skyglass: For the Exasol connector, the expected behavior for TIMESTAMP precision is truncation, not rounding. While Trino casts to lower precision round values, this applies only to SQL casts inside Trino, not connector round-trip semantics.

Unfortunately, Exasol’s documentation does not explicitly state whether values beyond the declared precision are truncated or rounded. However, in practice, values read via the JDBC driver are truncated to the defined precision, and the connector preserves this behavior. This truncation behavior is demonstrated by the current round-trip test.

chenjian2664: Please move the cases out of this test method, add a coercion test instead.
you can refer the `TestPostgreSqlTypeMapping#testTimestampCoercion` as the example

skyglass: Thank you!
I have created separate  TestExasolTypeMapping#testTimestampCoercion and moved the cases with coercion to this new test method. I also added some additional test cases with coercion.

skyglass (commented):
@chenjian2664 , Thank you very much for your feedback! 👍 

I have created separate  TestExasolTypeMapping#testTimestampCoercion and moved the cases with coercion to this new test method. 
I also added some additional test cases with coercion.

chenjian2664:
@coderabbitai full review

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +279,154 @@ private void testDate(ZoneId sessionZone)
...
+
+                //DST ambiguity (overlap) time for "America/Bahia_Banderas" time zone
+                .addRoundTrip("timestamp(3)", "TIMESTAMP '2018-10-28 01:13:55.123'", createTimestampType(3), "TIMESTAMP '2018-10-28 01:13:55.123'")
+                .addRoundTrip("timestamp(6)", "TIMESTAMP '2018-10-28 01:13:55.123456'", createTimestampType(6), "TIMESTAMP '2018-10-28 01:13:55.123456'")
+
+                // Invalid DST gap time in "America/Bahia_Banderas" JVM time zone.
+                // The value '2018-04-01 02:13:55.123' is invalid in JVM time zone, because this local time
+                // never occurs: the clock jumps from 01:59 to 03:00 during DST
+                .addRoundTrip("timestamp(3)", resolveInvalidDstGapValue("2018-04-01 02:13:55.123"), createTimestampType(3), "TIMESTAMP '2018-04-01 03:13:55.123'")
+                .addRoundTrip("timestamp(3)", resolveInvalidDstGapValue("2018-04-01 02:13:55.123"), createTimestampType(3), "TIMESTAMP '2018-04-01 03:13:55.123'")
```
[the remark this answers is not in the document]
chenjian2664: cc @skyglass

chenjian2664:
```
The @Test annotation should come after the Javadoc comment, not before it. The conventional ordering is: Javadoc → annotations → method signature.
``` 
FYI

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -278,6 +279,154 @@ private void testDate(ZoneId sessionZone)
...
+        testTimestamp(TestingSession.DEFAULT_TIME_ZONE_KEY.getZoneId());
+    }
+
+    private void testTimestamp(ZoneId sessionZone)
+    {
+        Session session = Session.builder(getSession())
+                .setTimeZoneKey(TimeZoneKey.getTimeZoneKey(sessionZone.getId()))
+                .build();
+
+        SqlDataTypeTest.create()
```
Praveen2112, line 302: Can we add additional test cases with historic dates like before the gregorian julian calendar switch ?

skyglass: Added tests for historical dates before and after gregorian julian calendar switch and dates for julian-gregorian calendar cut-over.

skyglass (commented):
Thank you very much for your review, @chenjian2664 , @Praveen2112 , @ebyhr !
The PR has been updated and your review findings have been applied.

skyglass:
Any updates on the PR?

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -311,6 +332,140 @@ private static SliceWriteFunction hashTypeWriteFunction()
...
+                objectTimestampReadFunction(timestampType),
+                objectTimestampWriteFunction(timestampType),
+                FULL_PUSHDOWN);
+    }
+
+    private static LongReadFunction longTimestampReadFunction(TimestampType timestampType)
+    {
+        verifyLongTimestampPrecision(timestampType);
+        return (resultSet, columnIndex) -> {
+            Timestamp timestamp = resultSet.getTimestamp(columnIndex);
```
ebyhr, line 358: Let's leave a code comment explaining why we don't use `resultSet.getObject(columnIndex, LocalDateTime.class)`. Same for `objectTimestampReadFunction`.

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -311,6 +332,140 @@ private static SliceWriteFunction hashTypeWriteFunction()
...
+                "Precision is out of range: %s",
+                precision);
+    }
+
+    /**
+     * Returns a {@code TO_TIMESTAMP} bind expression using the appropriate format model
+     * based on the given fractional seconds precision.
+     * See for more details: <a href="https://docs.exasol.com/db/latest/sql_references/formatmodels.htm">Date/time format models</a>
+     */
+    private static String getTimestampBindExpression(int precision)
```
ebyhr:
```suggestion
    private static String timestampBindExpression(int precision)
```

https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-get-in-method-names-unless-an-object-must-be-a-java-bean

plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
```diff
@@ -311,6 +332,140 @@ private static SliceWriteFunction hashTypeWriteFunction()
...
+    {
+        checkArgument(precision >= 0, "Precision is negative: %s", precision);
+        if (precision == 0) {
+            return "TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS')";
+        }
+        return format("TO_TIMESTAMP(?, 'YYYY-MM-DD HH24:MI:SS.FF%d')", precision);
```
ebyhr, line 468: Could you add a new test verifying predicates with negative timestamp doesn't cause correctness issue? 

Inserting `TIMESTAMP '2013-03-11'` on Exasol, and run `SELECT ... WHERE dt = TIMESTAMP '-2013-03-11'` on Trino.

plugin/trino-exasol/src/test/java/io/trino/plugin/exasol/TestExasolTypeMapping.java
```diff
@@ -451,4 +624,54 @@ private SqlExecutor onRemoteDatabase()
...
+    }
+
+    private void testUnsupportedInsertValue(
+            String exasolType,
+            String inputLiteral,
+            String expectedException)
+    {
+        try (TestTable table = new TestTable(
+                onRemoteDatabase(),
+                TEST_SCHEMA + ".test_unsupported_hashtype_" + randomNameSuffix(),
```
ebyhr: Please fix the table name.

ebyhr (approved):

skyglass (commented):
Thank you very much for the review, @ebyhr ! 👍 

I’ve addressed the requested changes:

  - Renamed getTimestampBindExpression to timestampBindExpression.
  - Added comments explaining why getTimestamp(...).toLocalDateTime() is used instead of getObject(..., LocalDateTime.class).
  - Added a regression test for predicate pushdown with a negative Trino timestamp literal.
  - Added a comment explaining why the test stores a positive Exasol timestamp and filters with the corresponding negative Trino
    timestamp.

  - Fixed the unsupported timestamp test table prefix.
  - Renamed testUnsupportedInsertValue to testUnsupportedTimestampInsertValue.

PR #26964, opened 2025-10-15
Use bulk BlockBuilder APIs for appending range of blocks

raunaqmorarka:
## Description
Moves `appendRawBlockRange` to BlockBuilder to make it public.
Used it in Array operators

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Expose bulk block appending in BlockBuilder and update array/map builders and related functions to use the new API for more concise block range appends.

New Features:
- Expose a public bulk appendRawBlockRange API on BlockBuilder

Enhancements:
- Replace manual loops over ValueBlock with appendRawBlockRange across array/map writers and concat/flatten functions
- Remove redundant appendRawBlockRange implementation from BlockUtil and eliminate ValueBlock imports

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayConcatFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayConcatFunction.java
@@ -18,7 +18,6 @@ import io.trino.metadata.SqlScalarFunction;
 import io.trino.spi.TrinoException;
 import io.trino.spi.block.Block;
 import io.trino.spi.block.BufferedArrayValueBuilder;
-import io.trino.spi.block.ValueBlock;
 import io.trino.spi.function.BoundSignature;
 import io.trino.spi.function.FunctionMetadata;
 import io.trino.spi.function.Signature;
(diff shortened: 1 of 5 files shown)


core/trino-spi/src/main/java/io/trino/spi/block/BlockBuilder.java
```diff
@@ -89,4 +89,13 @@ default BlockBuilder newBlockBuilderLike(BlockBuilderStatus blockBuilderStatus)
     {
         return newBlockBuilderLike(calculateBlockResetSize(getPositionCount()), blockBuilderStatus);
     }
+
+    static void appendRawBlockRange(Block rawBlock, int offset, int length, BlockBuilder blockBuilder)
```
dain: In the other PR, I mentioned that I wasn't sure if we want to make this public.  My reson for this is because I can't think of a good name for this funcetion.  My thought is that this should just be a default method on the interface and not a static method, but we already used the name appendRange for value blocks, which should be preferred as they are way faster.  Since ValueBlock extends from Block we don't wan to have the same name as it creates overloading confustion in performance sensitive code.  I guess we could go with `appendRawRange`, but I don't really like that.  Maybe someone else has an idea here.  Regardless this method should have javadoc explaining that the other methods are preferred (unless your code will just have the same switch statement)

raunaqmorarka: I updated this to be `default void appendBlockRange(Block rawBlock, int offset, int length)` and added javadoc.
PTAL

PR #26970, opened 2025-10-15
Avoid null pointer exceptions from empty tables without snapshot

raunaqmorarka:
## Description
It is possible for optimize_manifests to fail
```
{"type":"java.lang.NullPointerException","message":"Cannot invoke \"org.apache.iceberg.Snapshot.allManifests(org.apache.iceberg.io.FileIO)\" because \"snapshot\" is null","suppressed":[],"stack":["io.trino.plugin.iceberg.IcebergMetadata.loadAllManifestsFromSnapshot(IcebergMetadata.java:4183)","io.trino.plugin.iceberg.IcebergMetadata.executeOptimizeManifests
```

## Release notes

```markdown
## Iceberg
* Fix failure in `optimize_manifests` for empty tables. ({issue}`26970`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -3423,6 +3423,9 @@ public class IcebergMetadata
 
     private static List<ManifestFile> loadAllManifestsFromSnapshot(Table icebergTable, Snapshot snapshot)
     {
+        if (snapshot == null) {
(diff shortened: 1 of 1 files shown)


grantatspothero (approved):
Thank you for fixing this.

wendigo:
Can we add a test?

raunaqmorarka:
> Can we add a test?

We already have a test for an empty table in `io.trino.plugin.iceberg.procedure.TestIcebergOptimizeManifestsProcedure#testEmptyManifest`. However, CREATE through Trino always creates a snapshot, I couldn't find a straightforward way of simulating this scenario.

PR #26971, opened 2025-10-16
Test optimize_manifests without snapshot in Iceberg

ebyhr:
## Description

Follow-up of #26970

## Release notes

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/procedure/TestIcebergOptimizeManifestsProcedure.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/procedure/TestIcebergOptimizeManifestsProcedure.java
@@ -13,22 +13,33 @@
  */
 package io.trino.plugin.iceberg.procedure;
 
+import com.google.common.collect.ImmutableMap;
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/procedure/TestIcebergOptimizeManifestsProcedure.java
```diff
@@ -217,6 +230,25 @@ void testNotRewriteSingleManifest()
...
+                        tableName,
+                        new Schema(Types.NestedField.required(1, "x", Types.LongType.get())),
+                        PartitionSpec.unpartitioned(),
+                        SortOrder.unsorted(),
+                        Optional.ofNullable(catalog.defaultTableLocation(SESSION, tableName)),
+                        ImmutableMap.of())
+                .commitTransaction();
+        assertThat(catalog.loadTable(SESSION, tableName).currentSnapshot()).isNull();
+
+        assertUpdate("ALTER TABLE " + tableName + " EXECUTE optimize_manifests");
```
chenjian2664, line 249: Does it make sense to have a query after it? even it returns empty result

PR #26988, opened 2025-10-16
Make query JSON more compact for UI

wendigo:
DataSize are serialized to exact bytes to ensure that aggregated values are exact and errors do not accumulate. For the human ingestion it is better to format to most succinct string.

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Enable succinct, human-readable DataSize formatting in UI JSON responses for QueryInfo.

Enhancements:
- Introduce DataSizeSerializer to conditionally output concise human-readable data sizes
- Bind DataSizeSerializer as the JSON serializer for DataSize in ServerMainModule
- Update UiQueryResource to build a custom JsonCodec<QueryInfo> with succinct DataSize formatting and return JSON

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/ui/UiQueryResource.java
+++ b/core/trino-main/src/main/java/io/trino/server/ui/UiQueryResource.java
@@ -13,11 +13,18 @@
  */
 package io.trino.server.ui;
 
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.cfg.ContextAttributes;
 import com.google.common.collect.ImmutableList;
 import com.google.inject.Inject;
+import io.airlift.json.JsonCodec;
+import io.airlift.json.JsonCodecFactory;
 import io.trino.dispatcher.DispatchManager;
 import io.trino.execution.QueryInfo;
 import io.trino.execution.QueryState;
+import io.trino.operator.OperatorInfo;
+import io.trino.plugin.base.metrics.TDigestHistogram;
 import io.trino.security.AccessControl;
 import io.trino.server.BasicQueryInfo;
 import io.trino.server.DisableHttpCache;
@@ -26,6 +33,7 @@ import io.trino.server.HttpRequestSessionContextFactory;
 import io.trino.server.security.ResourceSecurity;
 import io.trino.spi.QueryId;
 import io.trino.spi.TrinoException;
+import io.trino.spi.metrics.Metric;
 import io.trino.spi.security.AccessDeniedException;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.ws.rs.ForbiddenException;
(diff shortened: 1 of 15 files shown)


wendigo:
I've added digest pruning so that query json will be smaller :)

martint:
The query json is not meant to be for human consumption.

wendigo:
@martint but it is used by humans :) (at least by some of us)

core/trino-web-ui/src/main/resources/webapp/src/components/QueryList.jsx
```diff
@@ -188,12 +188,12 @@ export class QueryListItem extends React.Component {
                             </div>
                             <div className="col-xs-4 text-right">
                                 <a
-                                    href={'/ui/api/query/' + query.queryId + '?pretty'}
+                                    href={'/ui/api/query/' + query.queryId + '?download=true'}
```
raunaqmorarka: Why are we changing the UI ?
We should be able to see the JSON in the browser instead of having to always download it

raunaqmorarka (commented):
lgtm % UI changes

core/trino-web-ui/src/main/resources/webapp/src/components/QueryList.jsx
wendigo: It's the query list (small icon there). The "JSON" tab works as before.

wendigo: CleanShot 2025-10-16 at 23 16 45@2x

raunaqmorarka:
> The query json is not meant to be for human consumption.

We use it all the time to debug performance issues. The fields removed here don't make the json any less consume-able by machines and bring down the size of the json pulled by the UI or any other user.

wendigo:
UI changes dropped (will revisit later with better proposal)

raunaqmorarka (approved):

martint:
It's not just removing the t-digests. It's changing the encoding of data sizes to use the succinct representation, which is lossy.

raunaqmorarka:
> It's not just removing the t-digests. It's changing the encoding of data sizes to use the succinct representation, which is lossy.

The succinct representation should only be used on the final DataSize values that go out from the coordinator REST endpoint. I think we want even the UI to show succinct representation as that is easier to read.

wendigo:
@martint it is lossy but only applies to the query json download. Not to every other serialization path. These are final values, not ones before aggregation.

For query json order of magnitude matters. Not down to a single byte values. Eventually we could make this configurable using server config but I doubt that this would be used.

core/trino-main/src/main/java/io/trino/server/ui/UiQueryResource.java
```diff
@@ -144,4 +157,30 @@ private Response failQuery(QueryId queryId, TrinoException queryException, HttpS
...
+            ObjectMapper mapper = objectMapper
+                    .copy()
+                    .setDefaultAttributes(attrs);
+            // Don't serialize TDigestHistogram.digest which isn't useful and human readable
+            mapper.configOverride(TDigestHistogram.class).setIgnorals(forIgnoredProperties("digest"));
+
+            // Do not output @class property for metric types
+            mapper.addMixIn(Metric.class, DropTypeInfo.class);
+            // Do not output @type property for OperatorInfo
+            mapper.addMixIn(OperatorInfo.class, DropTypeInfo.class);
```
martint, line 187: This is brittle, and will get out of sync as the QueryInfo continues to evolve.

martint:
The UI should be the one to decide how to render the data sizes. For example, if it's trying to draw a chart, it may need to normalize to a different unit that the most succinct one for one metric.

> but only applies to the query json download.

That's not what I see in the code. It seems to apply to all requests to `/ui/api/query`

wendigo:
@martint only to the `getQueryInfo` where codec is used

core/trino-main/src/main/java/io/trino/server/ui/UiQueryResource.java
martint: If we want a dump in a human-consumable format, we should consider introducing an explicit API for that instead of monkey-patching the representation produced by the programmatic API used by the UI.

martint:
Yes, sorry, that's what I meant: `/ui/api/query/<id>`

wendigo:
I narrowed it down only to the `Query JSON` tab in the UI

Issue #26755, opened 2025-09-30
`SELECT DISTINCT` with Delimited Alias in `ORDER BY` Errors (Case Sensitivity)


jimsimon-wk:
A query using `SELECT DISTINCT` fails when attempting to order by a column alias that is a delimited identifier (e.g., `"x"`), even though the unquoted `ORDER BY` expression (`X`) should resolve to it via case-insensitive matching.

This specific issue—the failure to resolve the case-insensitive `ORDER BY` expression to the alias in a `DISTINCT` context—was previously addressed and fixed in PR [#4233](https://github.com/trinodb/trino/pull/4233).

I suspect this to be a side effect of the modifications to `Identifier#getCanonicalValue` introduced by PR [#6550](https://github.com/trinodb/trino/pull/6550), which appears to have broken the case-insensitive matching for delimited aliases in the `ORDER BY` clause.

The following minimal, reproducible example demonstrates the failure:

```
SELECT DISTINCT a as "x" FROM (VALUES 2, 1, 2) t(a) ORDER BY X
```

**Expected Behavior**
The query should execute resolving the identifier in the `ORDER BY` scope to reference the alias in the `SELECT`, regardless of case.

**Actual Behavior**
```
line 1:1: For SELECT DISTINCT, ORDER BY expressions must appear in select list
```

Here's patch for a reproducing test

```patch
diff --git a/core/trino-main/src/test/java/io/trino/sql/query/TestDistinctWithOrderBy.java b/core/trino-main/src/test/java/io/trino/sql/query/TestDistinctWithOrderBy.java
index 68b7447f8fd..867142211fc 100644
--- a/core/trino-main/src/test/java/io/trino/sql/query/TestDistinctWithOrderBy.java
+++ b/core/trino-main/src/test/java/io/trino/sql/query/TestDistinctWithOrderBy.java
@@ -72,6 +72,9 @@ public class TestDistinctWithOrderBy
         assertThat(assertions.query("SELECT DISTINCT a, b a FROM (VALUES (2, 10), (1, 20), (2, 10)) T(a, b) ORDER BY T.a"))
                 .ordered()
                 .matches("VALUES (1, 20), (2, 10)");
+
+        assertThat(assertions.query("SELECT DISTINCT a as \"x\" FROM (VALUES 2, 1, 2) t(a) ORDER BY X"))
+            .matches("VALUES 1, 2");
     }
 
     @Test
```

findepi:
cc @Praveen2112

[...]

PR #27004, opened 2025-10-17
Fix SELECT DISTINCT with delimited alias in ORDER BY bug

codluca:
Fix SELECT DISTINCT with delimited alias in ORDER BY bug

Fixes https://github.com/trinodb/trino/issues/26755

## Description

The canonical value of a non-delimited Identifier was the upper-case value. The canonical value of a delimited Identifier was simply the value. Thus, statements like
SELECT DISTINCT a as x FROM (VALUES 2, 1, 2) t(a) ORDER BY "x"
would fail, as the canonical value of the identifier from SELECT DISTINCT would be "X",
while the canonical value of the identifier from ORDER BY would be "x", and the identifiers would not match.

## Release notes

```markdown
## Section
* Fix some things. ({issue}`26755`)
```

## Summary by Sourcery

Fix SELECT DISTINCT alias matching in ORDER BY to handle delimited and mixed-case identifiers by adding ignore-case canonicalization support and updating alias comparison accordingly.

Bug Fixes:
- Fix failure when ordering by a delimited SELECT DISTINCT alias due to mismatched canonical values.

Enhancements:
- Add ignore-case mode in CanonicalizationAware for identifier equality and hashing.
- Update Identifier to support case-insensitive canonical value retrieval.
- Adjust StatementAnalyzer to use ignore-case canonicalization when resolving DISTINCT aliases.

Tests:
- Add test cases for SELECT DISTINCT with various combinations of delimited and mixed-case aliases in ORDER BY.

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/CanonicalizationAware.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/CanonicalizationAware.java
@@ -25,20 +25,27 @@ import static java.util.Objects.requireNonNull;
 public class CanonicalizationAware<T extends Node>
 {
     private final T node;
+    private final boolean ignoreCase;
 
     // Updates to this field are thread-safe despite benign data race due to:
     // 1. idempotent hash computation
     // 2. atomic updates to int fields per JMM
     private int hashCode;
 
-    private CanonicalizationAware(T node)
+    private CanonicalizationAware(T node, boolean ignoreCase)
     {
         this.node = requireNonNull(node, "node is null");
+        this.ignoreCase = ignoreCase;
     }
 
     public static <T extends Node> CanonicalizationAware<T> canonicalizationAwareKey(T node)
     {
-        return new CanonicalizationAware<T>(node);
+        return new CanonicalizationAware<T>(node, false);
+    }
+
+    public static <T extends Node> CanonicalizationAware<T> canonicalizationAwareKey(T node, boolean ignoreCase)
+    {
+        return new CanonicalizationAware<T>(node, ignoreCase);
     }
 
     public T getNode()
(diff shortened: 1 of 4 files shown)


core/trino-main/src/main/java/io/trino/sql/analyzer/CanonicalizationAware.java
```diff
@@ -91,6 +100,15 @@ public static Boolean canonicalizationAwareComparison(Node left, Node right)
         return null;
     }
 
+    public static Boolean canonicalizationAwareIgnoreCaseComparison(Node left, Node right)
```
Praveen2112, line 103: If we are ignoring case comparison - Say if we have a `ORDER BY x, "X"` - How would it match ?

core/trino-main/src/main/java/io/trino/sql/analyzer/CanonicalizationAware.java
```diff
@@ -25,20 +25,27 @@
...
     private final T node;
+    private final boolean ignoreCase;
 
     // Updates to this field are thread-safe despite benign data race due to:
     // 1. idempotent hash computation
     // 2. atomic updates to int fields per JMM
     private int hashCode;
 
-    private CanonicalizationAware(T node)
+    private CanonicalizationAware(T node, boolean ignoreCase)
```
martint, line 35: This isn't right. The whole purpose of this class is to compare identifiers taking into account SQL canonicalization rules. I.e., delimited identifiers are kept as is, while non-delimited identifiers are canonicalized (to upper-case, per the SQL standard) before comparison.

martint (changes requested):

core/trino-main/src/main/java/io/trino/sql/analyzer/CanonicalizationAware.java
codluca: Indeed, if the adopted convention is similar to Oracle (non-delimited identifiers converted to uppercase, delimited identifiers kept as they are), then the changes in this PR are not ok.

The changes in this PR try to follow the Trino documentation https://trino.io/docs/current/language/reserved.html#language-identifiers
"Identifiers are not treated as case sensitive."
I read the "Identifiers" as meaning all identifiers, non-delimited and delimited.

The SELECT (without DISTINCT) follows the convention in the online documentation, by ignoring the case of all identifiers.
The changes here try to bring the SELECT DISTINCT to be similar to SELECT.

I modified the CanonicalizationAware, instead of removing its usage from SELECT DISTINCT, because the Identifier's equals and hashCode consider the case sensitive value.

I will copy this comment on the issue https://github.com/trinodb/trino/issues/26755, as the issue is more visible, and this PR seems to be wrong.

PR #27005, opened 2025-10-17
Ban AWS SDK v1 usage in new code

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-tests/pom.xml
+++ b/testing/trino-tests/pom.xml
@@ -440,6 +440,20 @@
     <build>
         <plugins>
             <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-enforcer-plugin</artifactId>
+                <configuration>
+                    <rules>
+                        <bannedDependencies>
+                            <includes combine.children="append">
+                                <!-- Legacy HDFS file system uses AWS S3 SDK v1 -->
+                                <include>com.amazonaws:*:*</include>
+                            </includes>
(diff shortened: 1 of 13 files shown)


testing/trino-tests/pom.xml
```diff
@@ -178,12 +178,6 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-hdfs</artifactId>
```
findepi, line 183: This might have been used dependency

```
TestQueryManager.testQueryWriteExceededSession

Configuration errors:

1) Error: An exception was caught and reported. Message: HDFS should be on the classpath for tests
  at ConfigurationFactory.registerConfigurationClasses(ConfigurationFactory.java:187)
```

wendigo: Yeah, we need to migrate our hive/iceberg tests that are using local filesystem to some alternative (like native local fs)

findepi: does this need to be part of this pr? or can we just ban AWSK SDK v1 in new code without changing the tests?

wendigo: @findepi independent

findepi: so why not split into two PRs? one doing what the title says: "Ban AWS SDK v1 usage in new code" and what's already approved here.

Issue #26981, opened 2025-10-16
`match_recognize`: Logical navigation function `FIRST` yields unexpected output


OrangieLou:
Hi everyone,

the query 
```sql
WITH t (rid, price) AS (
  SELECT * FROM (VALUES 
    (1, 1009),
    (2, 1019),
    (3, 1029),
    (4, 1039),
    (5, 1049),
    (6, 1059),
    (7, 1069))
)
SELECT rid, first_price, first_price1
FROM t MATCH_RECOGNIZE(
  ORDER BY rid
  MEASURES
    FIRST(price)    AS first_price,
    FIRST(price, 1) AS first_price1
  ALL ROWS PER MATCH
  PATTERN (A B+)
  DEFINE B AS B.price < 1050
);
```
yields the following output in Trino:
```
 rid | first_price | first_price1
-----+-------------+--------------
   1 |        1009 |         1019
   2 |        1009 |         1019
   3 |        1009 |         1019
   4 |        1009 |         1019
   5 |        1009 |         1019
(5 rows)
```

The same query in Oracle
```sql
WITH t (rid, price) AS (
    SELECT 1, 1009 FROM dual UNION ALL
    SELECT 2, 1019 FROM dual UNION ALL
    SELECT 3, 1029 FROM dual UNION ALL
    SELECT 4, 1039 FROM dual UNION ALL
    SELECT 5, 1049 FROM dual UNION ALL
    SELECT 6, 1059 FROM dual UNION ALL
    SELECT 7, 1069 FROM dual
)
SELECT rid, first_price, first_price1
FROM t MATCH_RECOGNIZE(
  ORDER BY rid
  MEASURES
    FIRST(price)    AS first_price,
    FIRST(price, 1) AS first_price1
  ALL ROWS PER MATCH
  PATTERN (A B+)
  DEFINE B AS B.price < 1050
);
```
yields a different output:
```
       RID FIRST_PRICE FIRST_PRICE1
---------- ----------- ------------
	 1	  1009
	 2	  1009	       1019
	 3	  1009	       1019
	 4	  1009	       1019
	 5	  1009	       1019
```

The output Trino produces is the output Oracle produces when using `FINAL FIRST(price, 1)`.

Is this an intended difference to Oracle? I did understand that unless the `FINAL` keyword is used only the part of the match up to the current row is known and therefore will yield `NULL` values if one refers to future rows of a match as with the offset of 1. Is my notion wrong here?

I work with Trino 476 in Docker.

[...]

PR #27006, opened 2025-10-17
Restrict logical navigations to the running match in running semantics

kasiafi:
## Description
Before this change, the logical navigation function `FIRST` could return position out of bounds of the current match in the `running` semantics. Now it is restricted to the current match.

Fixes: https://github.com/trinodb/trino/issues/26981

## Release notes

```markdown
## General
* In row pattern matching, restrict logical navigations to current match in running semantics. ({issue}`26981`)
```

## Summary by Sourcery

Restrict logical navigation functions to the current match bounds in running semantics and update tests accordingly.

Bug Fixes:
- Prevent FIRST and LAST navigations from returning positions outside the current match in running semantics

Tests:
- Adjust expected results for FIRST navigation to null when the match is not yet complete in running semantics

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/window/pattern/LogicalIndexNavigation.java
+++ b/core/trino-main/src/main/java/io/trino/operator/window/pattern/LogicalIndexNavigation.java
@@ -76,26 +76,20 @@ public class LogicalIndexNavigation
         checkArgument(currentRow >= patternStart && currentRow < patternStart + matchedLabels.length(), "current row is out of bounds of the match");
 
         int relativePosition;
+        int patternEndInclusive = running ? currentRow - patternStart : matchedLabels.length() - 1;
         if (last) {
-            int start;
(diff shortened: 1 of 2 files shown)


core/trino-main/src/main/java/io/trino/operator/window/pattern/LogicalIndexNavigation.java
```diff
@@ -110,11 +104,11 @@ private int findLastAndBackwards(int searchStart, ArrayView matchedLabels)
     }
 
     // FIRST(A.price, 3): find the first occurrence of label "A" and go 3 occurrences forward
-    private int findFirstAndForward(ArrayView matchedLabels)
+    private int findFirstAndForward(ArrayView matchedLabels, int patternEndInclusive)
```
Praveen2112: nit : Can we make method signature similar to `findLastAndBackwards`

PR #16725, opened 2023-03-25
Add support for `PRIMARY KEY` column constraint in engine

ebyhr:
## Description

Adds engine support for declaring a `PRIMARY KEY` table constraint in `CREATE TABLE`:

```sql
CREATE TABLE orders (
    orderkey bigint NOT NULL,
    custkey bigint,
    PRIMARY KEY (orderkey))
```

### Syntax and semantics

- The primary key is modeled as a table-level constraint (`PRIMARY KEY (a, b)`), not a column property, so multi-column keys are supported. Column-level `PRIMARY KEY` sugar is not supported.
- `PRIMARY` becomes a reserved word, as in the SQL standard. Queries using unquoted `primary` as an identifier need to quote it after this change (see release notes).
- The engine validates that at most one primary key constraint is specified, the column list is non-empty and free of duplicates, all listed columns exist, their types are comparable, and each column is declared `NOT NULL`. Violations fail with the new `INVALID_PRIMARY_KEY` error code.
- Requiring an explicit `NOT NULL` declaration is stricter than the standard, which makes primary key columns not nullable implicitly (SQL-92 4.13 and 11.7). This can be relaxed later without a breaking change.
- The declaration-only semantics deviate from the standard, where a primary key always enforces uniqueness. The standard's `NOT ENFORCED` constraint characteristic is explicitly disallowed for unique and primary key constraints ("If `<unique constraint definition>` is specified, then `<constraint characteristics>` shall not specify a `<constraint enforcement>`"), so there is no standard syntax for an informational primary key. This matches the position of warehouse engines that treat primary keys as informational (DB2 `NOT ENFORCED`, Snowflake, Azure Synapse `NONCLUSTERED NOT ENFORCED`, Databricks `RELY`). Note the consequence: `SHOW CREATE TABLE` prints standard primary key syntax for a constraint Trino does not enforce.
- `SHOW CREATE TABLE` renders the primary key when the connector returns one in table metadata.
- `CREATE TABLE LIKE` does not copy the primary key, matching the standard, whose `<like options>` cover only identity and defaults (`INCLUDING`/`EXCLUDING IDENTITY`/`DEFAULTS`); copying constraints (as in PostgreSQL `INCLUDING CONSTRAINTS`) is a vendor extension. `CREATE TABLE AS` cannot declare a primary key, matching the standard's `<as subquery clause>`, which allows only a column name list.

### SPI

[...]

PR #27026, opened 2025-10-21
Support Iceberg Identifier fields

### LINKED ISSUE #23563: Support Iceberg Identifier fields
[The Iceberg spec mentions ](https://iceberg.apache.org/spec/#identifier-field-ids)

> Identifier Field IDs
> A schema can optionally track the set of primitive fields that identify rows in a table, using the property identifier-field-ids (see JSON encoding in Appendix C).
> 
> Two rows are the "same"---that is, the rows represent the same entity---if the identifier fields are equal. However, uniqueness of rows by this identifier is not guaranteed or required by Iceberg and it is the responsibility of processing engines or data providers to enforce.
> 
> Identifier fields may be nested in structs but cannot be nested within maps or lists. Float, double, and optional fields cannot be used as identifier fields and a nested field cannot be used as an identifier field if it is nested in an optional struct, to avoid null values in identifiers.

Support defining Identifier Field IDs when creating tables and showing that these are identifier fields when displaying with "show create table"

As above no expectation of enforcing these is required in Trino but the creation, storage and access to this metadata would be helpful to those creating automated processes to merge tables.

7hong:
## Description

fixed https://github.com/trinodb/trino/issues/23563

Supports specifying the `identifier_fields` attribute when creating an iceberg table, and can also modify or remove the `identifier_fields` attribute in the Alter table statement.

## Release notes

( x) Release notes are required, with the following suggested text:

```markdown
## Section
* Add support for the `identifier_fields` property in the iceberg table. ({issue}`issuenumber`)
```

## Summary by Sourcery

Add support for the identifier_fields table property in Iceberg tables, allowing users to specify, modify, and remove identifier fields via CREATE TABLE WITH and ALTER TABLE, enforcing validation for field existence and non-nullability, and propagating identifier_fields through metadata operations, schema creation (including materialized views), and SHOW CREATE TABLE.

New Features:
- Allow specifying the identifier_fields property in Iceberg CREATE TABLE and ALTER TABLE statements
- Support clearing identifier_fields via an empty array in ALTER TABLE
- Reflect identifier_fields in SHOW CREATE TABLE output

Enhancements:
- Extend schemaFromMetadata, IcebergMetadata, and IcebergUtil to handle identifier_fields
- Validate identifier_fields for existence and non-nullability before applying
- Include identifier_fields support in materialized view creation path

Documentation:
- Document the identifier_fields property in the Iceberg connector documentation

Tests:
- Add unit tests for creating tables with single and multiple identifier_fields and updating them
- Add error-case tests for invalid and optional identifier fields
- Extend connector smoke tests to cover identifier_fields


ebyhr (changes requested):
I don't think we want to support this field in Iceberg connector because uniqueness isn't enforced. This is intentionally unsupported if I remember correctly. 

What's the benefit of supporting this property?

7hong:
> I don't think we want to support this field in Iceberg connector because uniqueness isn't enforced. This is intentionally unsupported if I remember correctly.
> 
> What's the benefit of supporting this property?

Indeed, there is no need to enforce uniqueness in Trino, nor in the Iceberg specification. In my scenario, I use trino to create and manage the iceberg table. In addition to trino, we also use streaming writing such as flink. In this case, identifier-field-ids is necessary.

lozbrown:
> I don't think we want to support this field in Iceberg connector because uniqueness isn't enforced. This is intentionally unsupported if I remember correctly. 
> 
> What's the benefit of supporting this property? 

This is useful place to store the details of identifier fields such that such that any tooling can use these details to build a the necessary merge SQL statement, otherwise we have to store these additional details somewhere else to build these merges.
Various other downstream apps use these fields this way.

@ebyhr if we document clearly that the iceberg connector will not enforce b uniqueness what's downside of merging this useful functionality that has kindly been contributed. 

Although it does feel like this paves the way for an "upsert into ..." Style command which although not standard SQL exists in several engines and is essentially an auto merge statement.

findepi:
> Supports specifying the `identifier_fields` attribute when creating an iceberg table, and can also modify or remove the `identifier_fields` attribute in the Alter table statement.

from the iceberg spec

> A schema can optionally track the set of primitive fields that identify rows in a table, using the property identifier-field-ids (see JSON encoding in Appendix C).
> 
> Two rows are the "same"---that is, the rows represent the same entity---if the identifier fields are equal. However, uniqueness of rows by this identifier is not guaranteed or required by Iceberg and it is the responsibility of processing engines or data providers to enforce.

Does this PR address the "responsibility of processing engines or data providers to enforce" part?

lozbrown:
@findepi would your concern be placated if the documentation were updated to clearly state that trino will not enforce the uniqueness?

@7hong would you be happy to enhance the docs as such? 

I understand quite what a large undertaking enforceing uniqueness and also what an expensive operation that would be for a large dataset. However I feel the functionality this PR provides it's useful as it stands and would really like to see it merged.

raunaqmorarka:
@lozbrown what's the behaviour implemented by other engines here ?
This is probably not a thing that Trino alone is dealing with, so if there is consensus in other iceberg supporting engines to support this without guaranteeing uniqueness, then we could potentially do the same here.

lozbrown:
@raunaqmorarka 
> @lozbrown what's the behavior implemented by other engines here ? 

In all the examples I'm aware of its optional, but essentially they support upsert, which is an equivalent of a merge but with SQL syntax similar to "Insert into..." these are not possible if
- the identifier fields are not defined on the table 
- the identifier fields are not supplied with the data to be upserted
- I believe in some cases its also necessary to supply any partition field or the source field on the partition field when using iceberg partition transforms

Some mention of it here:
https://iceberg.apache.org/docs/latest/spark-ddl/#alter-table-set-identifier-fields
https://iceberg.apache.org/docs/latest/flink-writes/#upsert

Whilst I'd love to see an implementation of upsert in trino along with a separate authorization  for that function (such that a user can be permissioned for upsert but not insert (and possibly not update)) I feel that's a big implementation that is a way off.

On the other hand a full implementation of this part is here, documented, has tests, ready to merge. This would allow me to continue to use Trino for all DDL changes. 

Ps, for my part I'm looking to implement Flink and also have an end user API for uploading files that uses upserts, for this it forms a merge in spark, the storage of identifier fields in both cases is a headache

lozbrown:
@raunaqmorarka @findepi @electrum @grantatspothero @findinpath 

has there been a ruling here, I'm keen to keep any momentum going here. 

BTW, spark is an example of an app that allows setting these in DDL commands but does not do anything to enforce.

raunaqmorarka:
> @raunaqmorarka @findepi @electrum @grantatspothero @findinpath
> 
> has there been a ruling here, I'm keen to keep any momentum going here.
> 
> BTW, spark is an example of an app that allows setting these in DDL commands but does not do anything to enforce.

So to clarify, if an iceberg table with identity identifiers exists in Spark, then you can run an INSERT or a MERGE INTO that would potentially break the uniqueness contract ?
I don't have a problem with that behaviour if we're not deviating from what the rest of the iceberg ecosystem is doing.

lozbrown:
> So to clarify, if an iceberg table with identity identifiers exists in Spark, then you can run an INSERT or a MERGE INTO that would potentially break the uniqueness contract ?

Correct

raunaqmorarka:
@martint could you chime in on this ?
The ask is to support "identity fields" table property in iceberg, for specifying primary keys in the table, but without any engine enforcement of the constraints.
https://iceberg.apache.org/spec/#identifier-field-ids
I see that there are connectors in Trino which allow specifying "primary_key" table properties, so its not a new idea.

lozbrown:
> @martint could you chime in on this ?
> The ask is to support "identity fields" table property in iceberg, for specifying primary keys in the table, but without any engine enforcement of the constraints.
> https://iceberg.apache.org/spec/#identifier-field-ids
> I see that there are connectors in Trino which allow specifying "primary_key" table properties, so its not a new idea.
> 
@martint please can you give your opinion here?

findepi:
> I see that there are connectors in Trino which allow specifying "primary_key" table properties, so its not a new idea.

In case of RDBMSes, we rely on separate engine to enforce uniqueness. 
We can cheaply define any constraints we want, and we won't violate any.

With Iceberg this is different. At least per spec, engine is supposed to enforce uniqueness defined in the table level.

raunaqmorarka:
> At least per spec, engine is supposed to enforce uniqueness defined in the table level.

I think a user can already configure "identifier-field-ids" from outside of Trino, and Trino has no awareness of this property today. So we're already in violation of enforcement requirement. Given that Apache Spark is not enforcing this as well, I'm inclined to match that. 
If we really want to be pedantic about following the spec, then our current behaviour is already a bug and we should throw an exception in iceberg connector insert/merge when identifier-field-ids are defined.

martint:
I'm ok with supporting this in the Iceberg connector. However, if identifier fields are a essentially a primary key, I would rather we introduced support for the PRIMARY KEY clause in the language and map it to the underlying properties as necessary.

raunaqmorarka:
> I'm ok with supporting this in the Iceberg connector. However, if identifier fields are a essentially a primary key, I would rather we introduced support for the PRIMARY KEY clause in the language and map it to the underlying properties as necessary.

@martint do you have a view on whether the iceberg connector should block inserts on such tables until we implement constraint enforcement, or its okay to leave this up to users ?

martint:
I don’t see being able to enforce constraints as a possibility for any engine until Iceberg supports some form of indexing, but I’m not familiar with what’s planned in Iceberg’s roadmap.

I’m ok with not enforcing it in Trino, either. However, we should document the limitations and potential for data corruption and incorrect results very clearly and explicitly.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -848,6 +857,23 @@ public static Schema schemaFromMetadata(List<ColumnMetadata> columns)
         return new Schema(icebergSchema.asStructType().fields());
```
raunaqmorarka: This should be replaced by `schemaFromMetadata(columns, ImmutableList.of());`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -848,6 +857,23 @@ public static Schema schemaFromMetadata(List<ColumnMetadata> columns)
...
     }
 
+    public static Schema schemaFromMetadata(List<ColumnMetadata> columns, List<String> identifierFields)
+    {
+        Schema schema = schemaFromMetadata(columns);
```
raunaqmorarka: Inline the logic of that method here and do the identifierFields look-ups in the existing for loop

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2491,6 +2494,13 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
             }
         }
 
+        if (properties.containsKey(IDENTIFIER_FIELDS_PROPERTY)) {
+            @SuppressWarnings("unchecked")
+            List<String> identifierFields = (List<String>) properties.get(IDENTIFIER_FIELDS_PROPERTY)
```
raunaqmorarka, line 2535: throw an exception if there are duplicates

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1485,6 +1485,72 @@ public void testMetadataDeleteAfterCommitEnabled()
...
+            icebergTable = loadTable(table.getName());
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .contains("identifier_fields = ARRAY['name']");
+            identifierFieldNames = icebergTable.schema().identifierFieldNames();
+            assertThat(identifierFieldNames).containsExactlyInAnyOrder("name");
+        }
+    }
+
+    @Test
+    public void testCreateTableWithMutilIdentifierFields()
```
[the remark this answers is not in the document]
raunaqmorarka: please fix typo

raunaqmorarka:
@martint do we want to hold this change until the engine has support for PRIMARY KEY ?
I think that will require a separate discussion about the syntax and SPI changes

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/glue/GlueConverter.java
ebyhr, line 295: > Wrap storage descriptor errors in TrinoExceptions

Revert unrelated changes and fix the commit title.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2526,6 +2529,17 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
...
+                    .orElse(Collections.emptyList());
+            Set<String> uniqueFields = new HashSet<>(identifierFields);
+            if (uniqueFields.size() != identifierFields.size()) {
+                throw new IllegalArgumentException("Duplicate identifier fields detected: " + identifierFields);
```
ebyhr: Throw TrinoException instead.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableProperties.java
```diff
@@ -70,6 +70,7 @@ public class IcebergTableProperties
     public static final String OBJECT_STORE_LAYOUT_ENABLED_PROPERTY = "object_store_layout_enabled";
     public static final String DATA_LOCATION_PROPERTY = "data_location";
     public static final String EXTRA_PROPERTIES_PROPERTY = "extra_properties";
+    public static final String IDENTIFIER_FIELDS_PROPERTY = "identifier_fields";
```
ebyhr, line 73: Move this line above EXTRA_PROPERTIES_PROPERTY.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -894,6 +894,24 @@ public void testIcebergTablesSystemTable()
         }
     }
 
+    @Test
+    public void testCreateTableWithIdentifierFields()
+    {
+        if (!hasBehavior(SUPPORTS_CREATE_TABLE)) {
+            return;
+        }
```
ebyhr: Why do we need this condition?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -894,6 +894,24 @@ public void testIcebergTablesSystemTable()
...
+        try (TestTable table = newTrinoTable(
+                "test_identifier_fields",
+                " (id BIGINT NOT NULL, name VARCHAR NOT NULL) WITH (identifier_fields = ARRAY['id'])")) {
```
ebyhr, line 902: Remove a redundant leading space.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1500,6 +1500,72 @@ public void testMetadataDeleteAfterCommitEnabled()
...
+                    .failure()
+                    .hasMessage("Identifier field 'area' cannot be optional");
+
+            assertUpdate("ALTER TABLE " + table.getName() + " SET PROPERTIES identifier_fields = ARRAY[]");
+            icebergTable.refresh();
+            identifierFieldNames = icebergTable.schema().identifierFieldNames();
+            assertThat(identifierFieldNames).isEmpty();
+        }
+    }
+
```
ebyhr, line 1578: Please add more test cases:
* Identifier field on nested fields
* CREATE TABLE with identifier field on nullable column

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1060,6 +1060,10 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
     Defaults to `false`. 
 * - `data_location`
   - Optionally specifies the file system location URI for the table's data files
+* - `identifier_fields`
+  - Optionally specifies table `identifier-field-ids`. If a table is `identifier-field-ids` by
+    columns `c1` and `c2`, the identifier_fields property is `identifier_fields =
+    ARRAY['c1', 'c2']`.
```
ebyhr: We should mention this property doesn't enforce uniqueness.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2573,6 +2587,30 @@ private static void updatePartitioning(Table icebergTable, Transaction transacti
...
+                    throw new TrinoException(NOT_SUPPORTED, "Identifier field '" + identifierField + "' cannot be optional");
+                }
+            }
+        }
+        updateSchema.setIdentifierFields(identifierFields);
+        try {
+            updateSchema.commit();
+        }
+        catch (Exception e) {
+            throw new TrinoException(ICEBERG_COMMIT_ERROR, "Failed to set new identifierFields value", e);
```
ebyhr:
```suggestion
            throw new TrinoException(ICEBERG_COMMIT_ERROR, "Failed to set new identifier fields", e);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -329,6 +332,11 @@ public static Map<String, Object> getIcebergTableProperties(BaseTable icebergTab
 
         compressionCodec.ifPresent(hiveCompressionCodec -> properties.put(COMPRESSION_CODEC, HiveCompressionOption.valueOf(hiveCompressionCodec.name())));
 
+        ImmutableList<String> identifierFieldNames = icebergTable.schema().identifierFieldNames().stream().collect(toImmutableList());
```
ebyhr:
```suggestion
        List<String> identifierFieldNames = icebergTable.schema().identifierFieldNames().stream().collect(toImmutableList());
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1500,6 +1500,72 @@ public void testMetadataDeleteAfterCommitEnabled()
...
+    public void testCreateTableWithIdentifierFields()
+    {
+        try (TestTable table = newTrinoTable("test_identifier_fields",
+                "(id BIGINT NOT NULL, name VARCHAR NOT NULL) WITH (identifier_fields = ARRAY['id'])")) {
+
+            BaseTable icebergTable = loadTable(table.getName());
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .contains("identifier_fields = ARRAY['id']");
+            Set<String> identifierFieldNames = icebergTable.schema().identifierFieldNames();
+            assertThat(identifierFieldNames).containsExactlyInAnyOrder("id");
```
ebyhr: `InAnyOrder` is redundant for single entry. Same for the following assertion.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1500,6 +1500,72 @@ public void testMetadataDeleteAfterCommitEnabled()
...
+    }
+
+    @Test
+    public void testCreateTableWithMultiIdentifierFields()
+    {
+        try (TestTable table = newTrinoTable("test_identifier_fields",
+                "(id BIGINT NOT NULL, name VARCHAR NOT NULL, area INT NOT NULL) WITH (identifier_fields = ARRAY['id','name'])")) {
+            BaseTable icebergTable = loadTable(table.getName());
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .containsAnyOf("identifier_fields = ARRAY['id','name']", "identifier_fields = ARRAY['name','id']");
```
ebyhr: The order is deterministic as far as I ran this with `@RepeatedTest(100)`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2573,6 +2587,30 @@ private static void updatePartitioning(Table icebergTable, Transaction transacti
         }
     }
 
+    private void updateIdentifierFields(Table icebergTable, Transaction transaction, List<String> identifierFields)
```
ebyhr: This method can be static.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2573,6 +2587,30 @@ private static void updatePartitioning(Table icebergTable, Transaction transacti
...
+        UpdateSchema updateSchema = transaction.updateSchema();
+        if (!identifierFields.isEmpty()) {
+            Schema schema = icebergTable.schema();
+            for (String identifierField : identifierFields) {
+                NestedField field = schema.findField(identifierField);
+                if (field == null) {
+                    throw new TrinoException(COLUMN_NOT_FOUND, "Field '" + identifierField + "' does not exist");
+                }
+                if (field.isOptional()) {
+                    throw new TrinoException(NOT_SUPPORTED, "Identifier field '" + identifierField + "' cannot be optional");
```
ebyhr: It's an invalid operation and there is no plan to support it. We should throw a different error code.

ebyhr: Alos, "optional" is internal term in Iceberg. I recommend changing to "Identifier field must be NOT NULL" or something.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
ebyhr: There is a bug around nested field handling:
```sql
trino:tpch> create table test(x row(f int) not null) with (identifier_fields = array['x.f']);
CREATE TABLE
trino:tpch> show create table test;
                                                                      Create Table
---------------------------------------------------------------------------------------------------------------------------------------------------------
 CREATE TABLE iceberg.tpch.test (
    x ROW(f integer) NOT NULL
 )
 WITH (
    format = 'PARQUET',
    format_version = 2,
    location = 'file:/var/folders/9s/_zwn4r_n2_9bp0krllp1pl3c0000gp/T/iceberg_query_runner639451971921028017/tpch/test-512309486ce24def9f6d976cd199b52f'
 )
```

ebyhr (commented):
> Wrap storage descriptor errors in TrinoExceptions

Please fix commit title.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
7hong: @ebyhr It seems that setting the internal attribute of row type to NOT NULL is not supported in trino, so it cannot be set to identifier_fields. I am going to list this situation as not supported. Do you think it is okay?

lozbrown: https://github.com/trinodb/trino/issues/27404

ebyhr: @lozbrown The issue is unrelated for new tables.

ebyhr: @7hong Blocking CREATE TABLE is fine, but what if the existing table contain identifier fields on nested fields (= created by other engines)? SHOW CREATE TABLE result should be reusable.

7hong: @ebyhr  Yes, this situation should be reusable. I will add unit test to verify this situation.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2526,6 +2530,17 @@ public void setTableProperties(ConnectorSession session, ConnectorTableHandle ta
             }
         }
 
+        if (properties.containsKey(IDENTIFIER_FIELDS_PROPERTY)) {
+            @SuppressWarnings("unchecked")
+            List<String> identifierFields = (List<String>) properties.get(IDENTIFIER_FIELDS_PROPERTY)
+                    .orElse(Collections.emptyList());
```
chenjian2664: use `ImmutableList.of()`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2573,6 +2588,36 @@ private static void updatePartitioning(Table icebergTable, Transaction transacti
...
+        if (!identifierFields.isEmpty()) {
+            Schema schema = icebergTable.schema();
+            for (String identifierField : identifierFields) {
+                NestedField field = schema.findField(identifierField);
+                if (field == null) {
+                    throw new TrinoException(COLUMN_NOT_FOUND, "Field '" + identifierField + "' does not exist");
+                }
+                // Currently, it is not possible to set the inner field of a nested field to NOT NULL in trino
+                boolean isTopLevel = schema.columns().stream()
+                        .anyMatch(nf -> nf.fieldId() == field.fieldId());
```
chenjian2664: nf -> `nestedField`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2573,6 +2588,36 @@ private static void updatePartitioning(Table icebergTable, Transaction transacti
...
 
+    private static void updateIdentifierFields(Table icebergTable, Transaction transaction, List<String> identifierFields)
+    {
+        UpdateSchema updateSchema = transaction.updateSchema();
+        if (!identifierFields.isEmpty()) {
+            Schema schema = icebergTable.schema();
+            for (String identifierField : identifierFields) {
+                NestedField field = schema.findField(identifierField);
+                if (field == null) {
+                    throw new TrinoException(COLUMN_NOT_FOUND, "Field '" + identifierField + "' does not exist");
```
chenjian2664: The exception should be `ICEBERG_INVALID_IDENTIFIER_FIELDS` since we are update here , should be user error right

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -842,10 +852,16 @@ public static Schema schemaFromMetadata(List<ColumnMetadata> columns)
                         .withDoc(column.getComment())
                         .build();
                 icebergColumns.add(field);
+                if (identifierFields.contains(column.getName())) {
```
chenjian2664, line 855: List.contains is insufficient, can we consider using Set

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1500,6 +1502,127 @@ public void testMetadataDeleteAfterCommitEnabled()
...
+
+    @Test
+    public void testCreateTableWithMultiIdentifierFields()
+    {
+        try (TestTable table = newTrinoTable("test_identifier_fields",
+                "(id BIGINT NOT NULL, name VARCHAR NOT NULL, area INT NOT NULL) WITH (identifier_fields = ARRAY['name','id'])")) {
+            BaseTable icebergTable = loadTable(table.getName());
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .contains("identifier_fields = ARRAY['id','name']");
+            List<String> identifierFieldNames = icebergTable.schema().identifierFieldNames().stream().sorted().collect(Collectors.toList());
```
chenjian2664: use `toImmutableList()`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1500,6 +1502,127 @@ public void testMetadataDeleteAfterCommitEnabled()
...
+            assertThat(identifierFieldNames).containsExactly("id", "name");
+
+            assertUpdate("ALTER TABLE " + table.getName() + " SET PROPERTIES identifier_fields = ARRAY['name','area']");
+            icebergTable = loadTable(table.getName());
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .contains("identifier_fields = ARRAY['area','name']");
+            identifierFieldNames = icebergTable.schema().identifierFieldNames().stream().sorted().collect(Collectors.toList());
```
chenjian2664: ditto

raunaqmorarka:
@7hong this needs to wait for https://github.com/trinodb/trino/pull/16725 to land, so that this can be done through PRIMARY KEY support in the engine rather than a connector table property

lozbrown:
> @7hong this needs to wait for https://github.com/trinodb/trino/pull/16725 to land, so that this can be done through PRIMARY KEY support in the engine rather than a connector table property

It's there any chance this will actually happen any time soon, that PR is 3 years old by a maintainer.

lozbrown:
It's there any good of reviving this any time soon..

We're having to update identifiers in spark in order to use the identifiers downstream because their l this is not supported in trino.

The aim of the original issue was just to support them as a table property without having to implement full enforcement of uniqueness. 

@raunaqmorarka @chenjian2664 @ebyhr ?

PR #27029, opened 2025-10-21
Send view security on view creation to OPA authorizer

sbernauer:
## Description

When creating views you can define two security modes: Invoker and definer, see [the docs](https://trino.io/docs/current/sql/create-view.html#security)

We only want to allow views with invoker security using OPA authorization.
For that to work we need to pass the information of the view security to OPA, what exactly is what this PR does.
Going forward other authorizers (e.g. file based or Ranger) can also profit from this information.

First commit does the SPI change.
Second commit uses the new attribute in the OPA authorizer.

## Release notes

```markdown
# Open Policy Agent access control
* Add support for OPA `CreateView` authorization based on the view security mode

# SPI
* Add additional view security argument to `AccessControl.checkCanCreateView`
```

This is a breaking change for third-party plugins that implement the `AccessControl` interfaces. I guess we need to also mention that?

## Summary by Sourcery

Extend view creation authorization to carry and enforce the SQL SECURITY mode by introducing a new ViewSecurity SPI parameter, propagating it through the engine and plugins, and integrating it into the OPA policy checks

New Features:
- Introduce ViewSecurity enum to represent INVOKER and DEFINER modes
- Extend AccessControl SPI checkCanCreateView to accept an Optional<ViewSecurity> parameter
- Capture and propagate view security mode in CreateViewTask and StatementAnalyzer

Enhancements:
- Update OPA authorizer to include view security in policy queries and enforce invoker-security for CreateView
- Propagate the new view security parameter through all forwarding and connector access control implementations
- Make file-based and Ranger system access controls aware of the view security mode

Tests:
- Add OPA plugin tests to verify CreateView requests with and without security mode

### DIFF
--- a/plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
+++ b/plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
@@ -468,7 +468,7 @@ public sealed class OpaAccessControl
     @Override
     public void checkCanCreateView(SystemSecurityContext context, CatalogSchemaTableName view, Optional<ViewSecurity> security)
     {
-        checkTableOperation(context, "CreateView", view, AccessDeniedException::denyCreateView);
+        checkViewOperation(context, "CreateView", view, security, AccessDeniedException::denyCreateView);
     }
 
     @Override
@@ -778,6 +778,17 @@ public sealed class OpaAccessControl
                 actionName,
                 () -> deny.accept(table.toString(), columns),
                 OpaQueryInputResource.builder().table(new TrinoTable(table).withColumns(columns)).build());
+    }
+
+    private void checkViewOperation(SystemSecurityContext context, String actionName, CatalogSchemaTableName view, Optional<ViewSecurity> security, Consumer<String> deny)
+    {
+        OpaQueryInputResource.Builder opaQueryInputResourceBuilder = OpaQueryInputResource.builder().table(new TrinoTable(view));
+        security.ifPresent(opaQueryInputResourceBuilder::security);
+        opaHighLevelClient.queryAndEnforce(
+                buildQueryContext(context),
+                actionName,
+                () -> deny.accept(view.toString()),
+                opaQueryInputResourceBuilder.build());
     }
 
     private <T> void enforcePermissionManagementOperation(Consumer<T> deny, T arg)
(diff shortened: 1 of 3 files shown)


core/trino-main/src/main/java/io/trino/security/AccessControl.java
```diff
@@ -313,7 +314,7 @@ public interface AccessControl
      *
      * @throws AccessDeniedException if not allowed
      */
-    void checkCanCreateView(SecurityContext context, QualifiedObjectName viewName);
+    void checkCanCreateView(SecurityContext context, QualifiedObjectName viewName, Optional<ViewSecurity> security);
```
kokosing, line 317: It is not needed. This method is simply not called for INVOKER.

sbernauer: Uhh interesting, thanks for the pointer!
Did I get it right that users can always create views (with invoker), and there is no way to prohibit that?

kokosing: Oh good point. I was thinking about `io.trino.security.ViewAccessControl#checkCanSelectFromColumns` that is checking this during the query of the view.

You are right, we are checking this for view creation disregarding security mode.

kokosing: I don't think this should be optional.

Also why we want access control grant/revoke access to create a view depending if it is INVOKER or DEFINER?

sbernauer: Trino admins want to authorize based on rules in Open policy agent, they want to control what people do have access to what.
If suddenly people in Trino can grant other people access to data sources which they shouldn't have access to, this breaks the "OPA is the source of truth for authorization" approach.
With this PR Trino admins have the ability (but are not forced) to prohibit views with `SECURITY DEFINER`, and be safe that e.g. legal rules in OPA are actually *always* enforced.

I modeled it as Optional, as that is how it is modeled in [`CreateView`](https://github.com/trinodb/trino/blob/254f381944a729c92565055342f03ff383b8a306/core/trino-parser/src/main/java/io/trino/sql/tree/CreateView.java#L37). We would need to explicitly default to `INVOKER` in `CreateViewTask`, is that what you are recommending? What if the default changes? Or (finally) becomes configurable?

PR #27035, opened 2025-10-21
Use LongAdder over AtomicLong when capturing metrics

wendigo:
## Release notes

```markdown
## Section
```

## Summary by Sourcery

Improve concurrent statistics collection by switching from AtomicLong/AtomicInteger to LongAdder across several stats classes and parallelizing driver stats aggregation in QueryManagerStats.

Enhancements:
- Replace atomic counters with LongAdder in QueryManagerStats, RuleStats, and OptimizerStats for improved concurrency
- Parallelize the stream in QueryManagerStats.updateDriverStats to leverage LongAdder’s performance benefits

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/planner/optimizations/OptimizerStats.java
+++ b/core/trino-main/src/main/java/io/trino/sql/planner/optimizations/OptimizerStats.java
@@ -18,11 +18,11 @@ import org.weakref.jmx.Managed;
 import org.weakref.jmx.Nested;
 
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
 
 public class OptimizerStats
 {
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/sql/planner/iterative/RuleStats.java
```diff
@@ -18,40 +18,40 @@
...
 
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
 
 public class RuleStats
 {
-    private final AtomicLong invocations = new AtomicLong();
-    private final AtomicLong hits = new AtomicLong();
+    private final LongAdder invocations = new LongAdder();
```
findepi, line 25: are these contended?

wendigo: Saw that on the JFR so I believe so

PR #27038, opened 2025-10-22
Allow reads to kafka-based tables using manually overridden offsets

jordepic:
## Description

Prior to this change, Trino's Kafka connector reads from dynamically loaded offsets fetched at query time. This is not conducive to snapshot reads, which can only be achieved by specifying a start and end offset per topic partition to read from.

In this change, we enable doing so by setting
kafka.override-topic-partition-offsets to true in the kafka catalog config file.

From there, we can use the session level setting,
kafka_topic_partition_offset_overrides, in order to specify manual overrides for a set of topic partitions. If specified, these will override the existing offsets that the connector loads at runtime.

## Release notes

Add support for reading user-specified offsets from the kafka connector. See [kafka docs](docs/src/main/sphinx/connector/kafka.md) for details.

## Summary by Sourcery

Enable snapshot-style reads from Kafka by allowing users to specify manual start and end offsets per topic partition via a new catalog flag and session property

New Features:
- Add config property 'kafka.override-topic-partition-offsets' to require explicit offset overrides for Kafka reads
- Introduce session property 'kafka_topic_partition_offset_overrides' to specify custom offset ranges per topic-partition

Enhancements:
- Introduce TopicPartitionOffsetProvider interface with Current and Overridden implementations to centralize offset retrieval
- Refactor KafkaSplitManager and KafkaFilterManager to use unified KafkaPartitionInputs record

Documentation:
- Document the new offset override config and session property in the Kafka connector documentation

Tests:
- Add unit tests for OverriddenTopicPartitionOffsetProvider and update KafkaConfig tests to cover the new catalog setting

### DIFF
--- a/docs/src/main/sphinx/connector/kafka.md
+++ b/docs/src/main/sphinx/connector/kafka.md
@@ -90,7 +90,7 @@ org.apache.kafka=WARN
 The following configuration properties are available:
 
 | Property name                                         | Description                                                                                                                                                                                                                 |
-| ----------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+|-------------------------------------------------------| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | `kafka.default-schema`                                | Default schema name for tables.                                                                                                                                                                                             |
 | `kafka.nodes`                                         | List of nodes in the Kafka cluster.                                                                                                                                                                                         |
 | `kafka.buffer-size`                                   | Kafka read buffer size.                                                                                                                                                                                                     |
@@ -109,6 +109,7 @@ The following configuration properties are available:
 | `kafka.ssl.key.password`                              | Password for the private key in the keystore file.                                                                                                                                                                          |
 | `kafka.ssl.endpoint-identification-algorithm`         | Endpoint identification algorithm used by clients to validate server host name; defaults to `https`.                                                                                                                        |
 | `kafka.config.resources`                              | A comma-separated list of Kafka client configuration files. These files must exist on the machines running Trino. Only specify this if absolutely necessary to access Kafka. Example: `/etc/kafka-configuration.properties` |
+| `kafka.override-topic-partition-offsets`              | Whether to allow overriding offsets to read for a particular kafka partition. Offsets are set with session parameter: `kafka.kafka_topic_partition_offset_overrides`.                                                         |
 
 In addition, you must configure {ref}`table schema and schema registry usage
 <kafka-table-schema-registry>` with the relevant properties.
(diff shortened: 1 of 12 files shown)


docs/src/main/sphinx/connector/kafka.md
```diff
@@ -218,6 +219,17 @@ Kafka uses `https` as default. Use `disabled` to disable server host name valida
...
+### `kafka.override-topic-partition-offsets`
+
+If enabled, a user may override the offsets to read for a particular kafka topic-partition to achieve snapshot read semantics.
+
+This property is optional; default is `false`.
+
+The name of the session property to use is `kafka_topic_partition_offset_overrides`.
+
+To read offsets 0 to 10 (exclusive) on partition 0 and 5 to 20 (exclusive) on partition 1 of topic x:
+`SET SESSION kafka.kafka_topic_partition_offset_overrides = 'x-0=0-10,x-1=5-20';`
```
dngurdl, line 231: P1: I think it'd be awesome to support reading the offset overrides from a file (as well as inlined as string)

plugin/trino-kafka/src/test/java/io/trino/plugin/kafka/TestOverriddenTopicPartitionOffsetProvider.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static io.trino.plugin.kafka.KafkaSessionProperties.KAFKA_TOPIC_PARTITION_OFFSET_OVERRIDES;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestOverriddenTopicPartitionOffsetProvider
+{
+    @Test
+    public void testCanParseOverriddenOffsets()
+    {
+        KafkaPartitionInputs inputs = OverriddenTopicPartitionOffsetProvider.parseOverriddenInputs("x-0=0-10, y-1=5-20 ");
```
dngurdl, line 38: Consider adding additional test cases around new config syntax parse error ("x-0?1-10").

docs/src/main/sphinx/connector/kafka.md
jordepic: Thanks for the comment!

Going to push back a little bit.  Since the configuration file is for a long running cluster, and I expect the set of offsets that users want to pull from to change fairly recently, this would lock them in for the duration of it until they changed the session variable anyways.

plugin/trino-kafka/src/test/java/io/trino/plugin/kafka/TestOverriddenTopicPartitionOffsetProvider.java
jordepic: Done :)

martint:
@cla-bot check

PR #27052, opened 2025-10-22
Support retrieving clustering information of Delta Lake tables.

yangshangqing95:
## Description
Support retrieving clustering information from Delta Lake tables, controlled by session and configuration settings.
By default, retrieving clustering information is disabled (false).
This serves as a foundation for future integration with the Delta Lake Liquid Clustering feature.
The work to fully support Delta Lake’s Liquid Clustering capability is already planned and in progress.

About Delta Lake Liquid Clustering: https://delta.io/blog/liquid-clustering/

## Release notes

```markdown
## Delta Lake
* Add support for retrieving clustering information from Delta Lake tables.
```

## Summary by Sourcery

Enable optional retrieval of clustering information for Delta Lake tables and expose it as a new table property to support future Liquid Clustering integration.

New Features:
- Introduce delta.show-clustered-columns session and catalog property to control retrieval of clustered columns
- Add clustered_by table property and include it in SHOW CREATE TABLE when enabled
- Implement clustering metadata extraction via ClusteringMetadataUtil reading Delta Lake transaction logs

Enhancements:
- Propagate clustering columns through DeltaLakeTableHandle, metadata, and properties handling
- Cache clustered columns in TableSnapshot and integrate with TransactionLogAccess

Documentation:
- Document the new delta.show-clustered-columns property in connector documentation

Tests:
- Add unit and integration tests for clustering metadata utilities, configuration mapping, session properties, and SHOW CREATE TABLE behavior


ebyhr:
Is this PR preparatory work for future performance improvements, with no immediate benefit?

yangshangqing95:
> Is this PR preparatory work for future performance improvements, with no immediate benefit?

Hi @ebyhr 
From a performance perspective, this PR doesn’t introduce any immediate improvements. But functionally, it allows users to view clustering column information in the `SHOW CREATE TABLE` output for Delta tables.

ebyhr:
Could you please share the final solution (= how to improve performance eventually)?

yangshangqing95:
Sure @ebyhr 
The long-term goal is to leverage Delta Lake’s **Liquid Clustering** to improve query performance once full support is added. Here, full support refers to enabling Trino to handle reading, writing with Liquid Clustering, and performing optimize operations to reorganize Liquid Clustering, an so on.

`Liquid Clustering` is a flexible data layout mechanism that organizes data based on clustering keys, instead of traditional directory-based partitions. Unlike static partitions (e.g., /country=US/), it stores clustering information in metadata rather than the filesystem, allowing the **data layout to evolve dynamically** as new data is written. Also, **Liquid Clustering is the best practice officially recommended by Delta Lake.** You can find more information here: https://delta.io/blog/liquid-clustering/

In simple terms, compared with partitions, Liquid Clustering provides:

- Flexible organization – data is grouped by clustering key ranges instead of fixed partition directories.

- Dynamic adjustment – clustering keys can be updated at any time; new files automatically adhere to the updated clustering layout. Additionally, existing data can be reorganized through some operations, giving the table structure much greater evolution flexibility than static partitions.

- Lower metadata overhead – no need to maintain thousands of partition directories.

- Better query pruning – queries filtering on clustering keys can skip large data ranges even when those fields aren’t partitions. In addition, Liquid Clustering can use `nested fields` (e.g., person.age) as clustering keys. Delta Lake tracks the full field path and its value range in metadata, so even struct subfields can benefit from clustering-based pruning. 

Compared with Parquet column statistics, which only store per-file min/max values without any global organization, Liquid Clustering provides a higher-level layout strategy. It ensures that values of clustering keys are physically localized across files, making Parquet’s per-file statistics far more effective for data skipping and reducing the number of files scanned.

Once Trino integrates with this metadata, it will be able to perform more accurate file pruning and data skipping, enabling more flexible data organization and clustering, thereby significantly improving the performance of queries filtering on clustering keys.

ebyhr:
Thanks for explaining the details. I already know about liquid clustering. I wanted to know the actual follow-up plan (especially read part) for the Delta Lake connector. 

> Once Trino integrates with this metadata, it will be able to perform more accurate file pruning and data skipping, enabling more flexible data organization and clustering, thereby significantly improving the performance of queries filtering on clustering keys.

The connector already uses stats from the transaction logs. Are you planning to read different metadata in the future? If so, could you elaborate on that?

chenjian2664:
@yangshangqing95 Would you mind sharing the plan for supporting the write path of liquid clustering? I suspect the read path won’t change much, since Trino already supports pruning with statistics, so reading clustering information doesn’t seem to provide additional benefit in my opinion

yangshangqing95:
Haha, please ignore my long-winded message above.
Hi @ebyhr @chenjian2664 
1. Regarding the **read performance improvements**, in my actual use cases, many Delta Lake tables choose a **clustering field** that is a **scalar nested inside a struct field**, where the root field itself is a large struct. Delta Lake can record **subfield statistics** in its stats, but currently **Trino** cannot perform **subfield predicate pushdown** during filtering. In such scenarios, we are unable to leverage Delta Lake’s stats for effective pruning — this is a problem we’re currently facing. Once we can obtain information about the clustering column, implementing pushdown for such clustered fields shouldn’t be difficult, and I can follow up on that later.

2. Regarding **where stats are stored**, yes — according to the Delta Lake protocol, they are still recorded in the **AddFile entry**, and that hasn’t changed.

3. Regarding **write-side support for Liquid Clustering**, some key points are:

   1. Support collecting statistics for **specified columns and fields**.
   2. Implement the ability to **compute clustering values** for fields (e.g., values based on **Hilbert curves**).
   3. Before writing files, introduce **clustering operators** to handle these tasks (extraction, computation, classification, sorting, etc.). If clustering is enabled, these operators should be applied. Optionally, intermediate results can be spilled to disk — fortunately, we already have this capability and can reuse it.

Coming back to this PR itself — It’s mostly about test data and test code — I hope to cover enough cases. My idea is to break down the larger system into smaller modules or features, which makes it easier to review and identify issues during testing.

Open to any discussions or suggestions.

yangshangqing95:
Hi @chenjian2664, any advice on this? Looking forward to proceeding ~

chenjian2664:
@yangshangqing95 I thought liquid clustering mainly affects the writing side, the reading part shouldn't need too many changes if we already support pruning scan files using field stats. Please help educate me: does liquid clustering write a different form of stats compared to non-clustered tables (for both primitive and struct columns)?

If we already support reading complex type stats, do we actually need the clustering info (since we don't write it)? Why would that be needed?

yangshangqing95:
> @yangshangqing95 I thought liquid clustering mainly affects the writing side, the reading part shouldn't need too many changes if we already support pruning scan files using field stats. Please help educate me: does liquid clustering write a different form of stats compared to non-clustered tables (for both primitive and struct columns)?
> 
> If we already support reading complex type stats, do we actually need the clustering info (since we don't write it)? Why would that be needed?

Hi @chenjian2664 Good questions, what I mentioned earlier was too general. just share some ideas, let's make it more clear with a simple example.
There is a table with two struct columns
| column   | type |
|--------|------|
| person   |   struct<age int, name string>   |
| address   |   struct<city string, zip int>   |

I can set the cluster key for this table in Databricks by using `delta.dataSkippingStatsColumns`, like `'delta.dataSkippingStatsColumns'= 'address.zip, person.age'`
image
refer to https://docs.databricks.com/aws/en/delta/data-skipping
Afterwards, we write three files.
| file   | person.age scope | record count |
|--------|------|------|
| file 1   |   0 ~ 20   | 1 |
| file 2  |   22 ~ 40   | 3 |
| file 3  |   60 ~ 70   | 4 |

Then, during a query, when we use person.age as a filter condition, Databricks can leverage the clustered column statistics it maintains internally to perform pruning.
for example: filter with condition `person.age < 20`
Here is the query details in databricks:
image
Clearly, two files has been pruned and only **one record** was read.

But when the same query executed in Trino, things are different. Due to the subfield can not be pushed down to the tables-scan stage, the statistics of clustered fields can not be used for pruning. As a result, there would be a full table scan.
image
**All files and all eight records were scanned.**

This is also an issue I encounter in practice. I found that for the same query, Trino takes significantly longer to execute, and this performance degradation becomes more pronounced as the data volume grows. Of course, in practice, the actual table structure and data volume are much more complex than in the example above.

Now back to your question
> does liquid clustering write a different form of stats compared to non-clustered tables?

No, these stats will still stored in the addfile entry, for both primitive and struct columns.

> If we already support reading complex type stats, do we actually need the clustering info (since we don't write it)? Why would that be needed?

We support read complex type stats but now we're not actually using them. As we don't know if a subfield is a clustered key and don't know if the stats of the subfield has been recorded.

My thought is that we can extract the clustered key separately for filtering, elevating it to the level of a partition, then allows us to make full use of all statistical information for more thorough pruning.

And the first step is to identify which field/column is the clustered key.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeTableProperties.java
```diff
@@ -39,6 +39,7 @@ public class DeltaLakeTableProperties
 {
     public static final String LOCATION_PROPERTY = "location";
     public static final String PARTITIONED_BY_PROPERTY = "partitioned_by";
+    public static final String CLUSTER_BY_PROPERTY = "clustered_by";
```
ebyhr: We shouldn't expose this property until we support creating a table with this property. Otherwise, SHOW CREATE TABLE result gets non-reusable (exception should happen). The current implementation ignores the property in CREATE TABLE - this is also no go.

plugin/trino-delta-lake/src/test/resources/databricks172/liquid_clustering_with_operations/README.md
```diff
@@ -0,0 +1,2 @@
+Data generated using Databricks 17.2
+Only the transaction log is included here, to test retrieving clustered information across all operation types.
```
ebyhr, line 2: Please document CREATE TABLE statements for future readers.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/clustering/ClusteringMetadataUtil.java
```diff
@@ -0,0 +1,296 @@
...
+                return ImmutableList.of();
+            }
+        }
+        else if (clusteredKey.equals(NEW_CLUSTERING_PARAMETER_KEY)) {
+            clusteredColumns = Arrays.stream(clusteredValue.split(","))
+                    .map(String::trim)
+                    .collect(toImmutableList());
+        }
+        else {
+            LOG.error(String.format("Unknown clustering key: %s", clusteredKey));
```
ebyhr: Avoid using String.format in logger.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/DeltaLakeTableFeatures.java
```diff
@@ -40,6 +40,7 @@ public final class DeltaLakeTableFeatures
     public static final String VARIANT_TYPE_FEATURE_NAME = "variantType";
     public static final String VARIANT_TYPE_PREVIEW_FEATURE_NAME = "variantType-preview";
     public static final String V2_CHECKPOINT_FEATURE_NAME = "v2Checkpoint";
+    public static final String CLUSTERED_TABLES_FEATURE_NAME = "clustering";
```
ebyhr: Move this constant under CHECK_CONSTRAINTS_FEATURE_NAME for ordering by alphabetically.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -649,6 +652,11 @@ private static long getTargetVersion(TrinoFileSystem fileSystem, String tableLoc
         return snapshotId;
     }
 
+    public static int getTemporalTimeTravelLinearSearchMaxSize()
+    {
+        return TEMPORAL_TIME_TRAVEL_LINEAR_SEARCH_MAX_SIZE;
+    }
```
ebyhr: Make TEMPORAL_TIME_TRAVEL_LINEAR_SEARCH_MAX_SIZE public, and remove this method.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/OperationTest.java
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.deltalake.clustering;
+
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+public class OperationTest
```
ebyhr: Please follow https://trino.io/docs/current/develop/tests.html#conventions-and-recommendations

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/TestClusteringMetadataUtil.java
```diff
@@ -0,0 +1,822 @@
...
+    {
+        String tableLocation = "databricks172/liquid_clustering_with_operations/liquid_clustering_with_auto_optimize";
+        TableSnapshot tableSnapshot = new TableSnapShotBuilderForTestBuilder()
+                .withTableLocation(getResourceLocation(tableLocation))
+                .build();
+
+        Optional<List<String>> result = ClusteringMetadataUtil.getLatestClusteredColumns(FILE_SYSTEM, tableSnapshot);
+
+        assertThat(result.isPresent()).isTrue();
+        assertThat(ImmutableList.of()).isEqualTo(result.get());
```
ebyhr: AssertJ provides contains method for optional type.
```java
assertThat(result).contains(ImmutableList.of());
```

Also, the order of parameter is wrong. `assertThat()` should take actual, not expected.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/TestClusteringMetadataUtil.java
```diff
@@ -0,0 +1,822 @@
...
+    {
+        CommitInfoEntry commitInfoEntry = new CommitInfoEntryForTestBuilder()
+                .withOperation("WRITE")
+                .withOperationParameters(ImmutableMap.of("clusterBy", "[]"))
+                .build();
+
+        Optional<CommitInfoEntry> commitInfo = Optional.of(commitInfoEntry);
+        List<String> result = ClusteringMetadataUtil.extractClusteredColumns(commitInfo);
+
+        assertThat(result.isEmpty()).isTrue();
```
ebyhr:
```java
assertThat(result).isEmpty();
```

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/TestClusteringMetadataUtil.java
```diff
@@ -0,0 +1,822 @@
...
+    @Test
+    void testGetClusteredColumnListWithClusterBy()
+    {
+        String clusteredKey = "clusterBy";
+        String clusteredValue = "[\"col1\", \"col2\"]";
+
+        List<String> result = getClusteredColumnList(clusteredKey, clusteredValue);
+
+        assertThat(2).isEqualTo(result.size());
+        assertThat(result.containsAll(ImmutableList.of("col1", "col2"))).isTrue();
```
ebyhr: Please use helper methods in AssertJ as much as possible:
```java
assertThat(result).containsExactly("col1", "col2");
```

ebyhr (changes requested):

plugin/trino-delta-lake/src/test/resources/databricks172/liquid_clustering_with_operations/liquid_clustering_with_auto_optimize/_delta_log/00000000000000000000.json
```diff
@@ -0,0 +1 @@
+{"commitInfo":{"timestamp":1760712838006,"userId":"user1","userName":"user1","operation":"OPTIMIZE","operationParameters":{"clusterBy":"[]","zOrderBy":"[]","batchId":"0","predicate":"[]","auto":true},"notebook":{"notebookId":"xxxxxx"},"clusterId":"cluster","readVersion":9,"isolationLevel":"SnapshotIsolation","isBlindAppend":false,"operationMetrics":{"numRemovedFiles":"2","numRemovedBytes":"5398","p25FileSize":"3647","numDeletionVectorsRemoved":"1","minFileSize":"3647","p75FileSize":"3647","p50FileSize":"3647","numAddedBytes":"3647","numAddedFiles":"1","maxFileSize":"3647"},"tags":{"delta.rowTracking.preserved":"true"},"engineInfo":"Databricks-Runtime/17.2.x-photon-scala2.13","txnId":"xxxxxx"}}
```
findinpath, line 1: pls add README.md on how the data was created - see the other resource directories in the module as point of reference.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeTableProperties.java
yangshangqing95: make sense, removed

plugin/trino-delta-lake/src/test/resources/databricks172/liquid_clustering_with_operations/README.md
yangshangqing95: documented

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/clustering/ClusteringMetadataUtil.java
yangshangqing95: fixed

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/DeltaLakeTableFeatures.java
yangshangqing95: done

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
yangshangqing95: done

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/OperationTest.java
yangshangqing95: modified

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/TestClusteringMetadataUtil.java
yangshangqing95: fixed

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/TestClusteringMetadataUtil.java
yangshangqing95: fixed

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/clustering/TestClusteringMetadataUtil.java
yangshangqing95: fixed

plugin/trino-delta-lake/src/test/resources/databricks172/liquid_clustering_with_operations/liquid_clustering_with_auto_optimize/_delta_log/00000000000000000000.json
yangshangqing95: done

raunaqmorarka:
> 
> column	type
> person	struct<age int, name string>
> address	struct<city string, zip int>
> I can set the cluster key for this table in Databricks by using `delta.dataSkippingStatsColumns`, like `'delta.dataSkippingStatsColumns'= 'address.zip, person.age'` image refer to https://docs.databricks.com/aws/en/delta/data-skipping Afterwards, we write three files.
> 
> file	person.age scope	record count
> file 1	0 ~ 20	1
> file 2	22 ~ 40	3
> file 3	60 ~ 70	4
> Then, during a query, when we use person.age as a filter condition, Databricks can leverage the clustered column statistics it maintains internally to perform pruning. for example: filter with condition `person.age < 20` Here is the query details in databricks: image Clearly, two files has been pruned and only **one record** was read.
> 
> But when the same query executed in Trino, things are different. Due to the subfield can not be pushed down to the tables-scan stage, the statistics of clustered fields can not be used for pruning. As a result, there would be a full table scan. image 
> **All files and all eight records were scanned.**
> 
> This is also an issue I encounter in practice. I found that for the same query, Trino takes significantly longer to execute, and this performance degradation becomes more pronounced as the data volume grows. Of course, in practice, the actual table structure and data volume are much more complex than in the example above.

In that example above, does the parquet file have min/max statistics on the struct subfields ?
Even if the delta connector is not pruning splits based on delta file statistics, the parquet reader should still be able to prune parquet row groups based on statistics in parquet file footer.
There was a PR in the past to support splits pruning based on delta statistics on ROW subfields https://github.com/trinodb/trino/pull/17464. This work is independent of whether clustering exists or anything done by the writer.
fyi @krvikash

yangshangqing95:
> > column	type
> > person	struct<age int, name string>
> > address	struct<city string, zip int>
> > I can set the cluster key for this table in Databricks by using `delta.dataSkippingStatsColumns`, like `'delta.dataSkippingStatsColumns'= 'address.zip, person.age'` image refer to https://docs.databricks.com/aws/en/delta/data-skipping Afterwards, we write three files.
> > file	person.age scope	record count
> > file 1	0 ~ 20	1
> > file 2	22 ~ 40	3
> > file 3	60 ~ 70	4
> > Then, during a query, when we use person.age as a filter condition, Databricks can leverage the clustered column statistics it maintains internally to perform pruning. for example: filter with condition `person.age < 20` Here is the query details in databricks: image Clearly, two files has been pruned and only **one record** was read.
> > But when the same query executed in Trino, things are different. Due to the subfield can not be pushed down to the tables-scan stage, the statistics of clustered fields can not be used for pruning. As a result, there would be a full table scan. image
> > **All files and all eight records were scanned.**
> > This is also an issue I encounter in practice. I found that for the same query, Trino takes significantly longer to execute, and this performance degradation becomes more pronounced as the data volume grows. Of course, in practice, the actual table structure and data volume are much more complex than in the example above.
> 
> In that example above, does the parquet file have min/max statistics on the struct subfields ? Even if the delta connector is not pruning splits based on delta file statistics, the parquet reader should still be able to prune parquet row groups based on statistics in parquet file footer. There was a PR in the past to support splits pruning based on delta statistics on ROW subfields #17464. This work is independent of whether clustering exists or anything done by the writer. fyi @krvikash

Hi @raunaqmorarka I don’t think there’s any relationship between them.
1. Without pushing down the subfield, the Parquet reader cannot access the filter conditions for pruning.
2. For Delta Lake clustered key pushdown, there’s no need to read the Parquet footer for pruning, it would introduce unnecessary I/O overhead.
3. In my opinion, the PR #17464 you mentioned introduces a kind of “virtual” column here, which might not be a general or universal approach.
```
for (RowType.Field field : rowType.getFields()) {
    projectColumns(name + "#" + field.getName().orElseThrow(), field.getType(), projectedColumnNames);
}
```

yangshangqing95:
Hi @ebyhr @findinpath I've fixed/updated all above comments, any new thoughts on this? I’m very much looking forward to starting the next phase of work, thank you all!

PR #27054, opened 2025-10-22
Introduce ConnectorContextModule

ksobolew:
## Description

This reduces code duplication - a lot - and also frees the plugin writers from guessing which objects are actually needed by the plugin. The answer this module gives is: all of them, don't worry about it.

> +201 −342

Seemingly not much, but there are also quality-of-development improvements in that (I believe).

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Consolidate connector context bindings into a new ConnectorContextModule, simplify TypeDeserializerModule, and refactor connector factories and tests to use the new module, reducing code duplication.

Enhancements:
- Add ConnectorContextModule to centralize binding of ConnectorContext and CatalogName in Guice modules
- Refactor connector factories to replace repetitive manual binder lambdas with ConnectorContextModule.from
- Simplify TypeDeserializerModule by removing its TypeManager dependency

Tests:
- Update connector tests to use ConnectorContextModule for context and catalog name bindings


lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/ConnectorContextModule.java
```diff
@@ -0,0 +1,57 @@
...
+    private ConnectorContextModule() {}
+
+    public static Module from(String catalogName, ConnectorContext context)
+    {
+        return combine(
+                ConnectorContextModule.from(context),
+                new CatalogNameModule(catalogName));
+    }
+
+    public static Module from(ConnectorContext context)
```
wendigo: can we instead of a lambda here have `ConnectorContextModule` implement module and have a configure method?

lib/trino-hdfs/src/main/java/io/trino/filesystem/hdfs/HdfsFileSystemManager.java
```diff
@@ -61,9 +61,9 @@ public HdfsFileSystemManager(
         modules.add(new HdfsModule());
         modules.add(new HdfsAuthenticationModule());
         modules.add(new HiveCosModule());
+        modules.add(new CatalogNameModule(catalogName));
         modules.add(binder -> {
             binder.bind(OpenTelemetry.class).toInstance(openTelemetry);
-            binder.bind(CatalogName.class).toInstance(new CatalogName(catalogName));
```
chenjian2664, line 66: this seems not relates to the change?

ksobolew: Not strictly related, yes. This is a follow-up cleanup.

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/ConnectorContextModule.java
ksobolew: I actually removed `CatalogName` from this module because it already has its own. So now we can make it implement `Module` and use a simple constructor, which blends well to existing code.

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/ConnectorContextModule.java
```diff
@@ -0,0 +1,54 @@
...
+import io.trino.spi.type.TypeManager;
+
+public class ConnectorContextModule
+        implements Module
+{
+    private final ConnectorContext context;
+
+    public ConnectorContextModule(ConnectorContext context)
+    {
+        this.context = context;
```
kokosing: rnn

plugin/trino-ai-functions/src/main/java/io/trino/plugin/ai/functions/AiConnectorFactory.java
```diff
@@ -42,8 +42,8 @@ public Connector create(String catalogName, Map<String, String> config, Connecto
         Bootstrap app = new Bootstrap(
                 "io.trino.bootstrap.catalog." + catalogName,
                 new AiModule(),
+                new ConnectorContextModule(context),
```
kokosing: maybe we should pass connector name as constructor parameter and bind it there too? It seems to be popular too.

ksobolew: I did that in the initial iteration, but then removed it. There's already `ConnectorNameModule`, we should use that one.

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/ConnectorContextModule.java
ksobolew: Fixed, thanks

plugin/trino-ai-functions/src/main/java/io/trino/plugin/ai/functions/AiConnectorFactory.java
kokosing: I think now you could maybe remove `ConnectorNameModule`

ssheikin: if possible - it would be nice

ksobolew: Ok, I'll do it

ksobolew: Done. It replaces the last commit with a different one.

ksobolew:
Oops, broke Hive 🤦 It's initialized using reflection so I missed a signature change.

ksobolew:
`TestDirectTrinoClient.testDirectTrinoClientLongQuery` seems to be slightly flaky...

ksobolew:
> `TestDirectTrinoClient.testDirectTrinoClientLongQuery` seems to be slightly flaky...

trinodb/trino#27082

kokosing:
Thanks!

PR #27058, opened 2025-10-23
Remove the query field in the URI before reconstructing to avoid duplicated query field in new URI

AndyJiang99:
## Description
When querying a non-standard userinfo endpoint (not the Microsoft one found in https://login.microsoftonline.com/<TENANT_ID>/v2.0/.well-known/openid-configuration), and instead, using UserInfo endpoint querying method (https://learn.microsoft.com/en-us/entra/identity-platform/userinfo) , the Trino code identifies the URL with the query params, however, during the original parsing, rather than starting a URL builder from scratch, it's taking the original URL and re-appending the query string to itself, resulting in duplicate query param.
E.g. `https://graph.microsoft.com/v1.0/me?select=id,userPrincipalName,displayName,email,onPremisesSamAccountName&select=id,userPrincipalName,displayName,email,onPremisesSamAccountName` instead of the expected `https://graph.microsoft.com/v1.0/me?select=id,userPrincipalName,displayName,email,onPremisesSamAccountName`. 

487552600-c3a5e755-7f02-467b-8e76-c1ab1e2ba626

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Bug Fixes:
- Strip the original query field in UriBuilder to avoid appending duplicate query parameters

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/security/oauth2/NimbusAirliftHttpClient.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/oauth2/NimbusAirliftHttpClient.java
@@ -79,7 +79,7 @@ public class NimbusAirliftHttpClient
                 .setMethod(method.name())
                 .setFollowRedirects(httpRequest.getFollowRedirects());
 
-        UriBuilder url = UriBuilder.fromUri(httpRequest.getURI());
+        UriBuilder url = UriBuilder.fromUri(httpRequest.getURI()).replaceQuery(null);
         if (method.equals(GET) || method.equals(DELETE)) {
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/server/security/oauth2/NimbusAirliftHttpClient.java
```diff
@@ -79,7 +79,7 @@ public <T> T execute(com.nimbusds.oauth2.sdk.Request nimbusRequest, Parser<T> pa
                 .setMethod(method.name())
```
kokosing, line 79: can you please provide a test that reproduces the initial issue?

kokosing:
@lukasz-walkiewicz can you please take a look?

PR #27059, opened 2025-10-23
Fix flush_metadata_cache failure when metastore impersonation is enabled

ebyhr:
## Description

Fix flush_metadata_cache failure when metastore impersonation is enabled

## Release notes

```markdown
## Hive
* Fix failure when executing `flush_metadata_cache` procedure with impersonation enabled. ({issue}`27059`)
```

## Summary by Sourcery

Enable flush_metadata_cache to function correctly when metastore impersonation is enabled by propagating an impersonationEnabled flag through Hive connector factories and modules, and updating the cache flush logic to use session-specific caching.

New Features:
- Introduce an impersonationEnabled flag in TestingHivePlugin, TestingHiveConnectorFactory, HiveConnectorFactory, and HiveMetastoreFactory to support metastore impersonation.
- Propagate the impersonationEnabled parameter through HiveMetastoreModule and connector factory constructors.

Enhancements:
- Update FlushMetadataCacheProcedure to recognize ImpersonationCachingHiveMetastoreFactory and create a session-scoped CachingHiveMetastore for flushing.
- Default connectors and modules now explicitly specify impersonationEnabled=false when impersonation isn’t configured.

Tests:
- Add TestThriftMetastoreImpersonation to verify flush_metadata_cache works under metastore impersonation.

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
@@ -21,6 +21,7 @@ import io.trino.metastore.HiveMetastore;
 import io.trino.metastore.HiveMetastoreFactory;
 import io.trino.metastore.Table;
 import io.trino.metastore.cache.CachingHiveMetastore;
+import io.trino.metastore.cache.SharedHiveMetastoreCache.ImpersonationCachingHiveMetastoreFactory;
 import io.trino.plugin.hive.HiveErrorCode;
 import io.trino.plugin.hive.fs.DirectoryLister;
 import io.trino.plugin.hive.metastore.glue.GlueCache;
@@ -134,6 +135,18 @@ public class FlushMetadataCacheProcedure
 
     private void doFlushMetadataCache(ConnectorSession session, Optional<String> schemaName, Optional<String> tableName, List<String> partitionColumns, List<String> partitionValues)
     {
+        if (hiveMetadataFactory instanceof ImpersonationCachingHiveMetastoreFactory impersonationCachingHiveMetastoreFactory) {
+            checkState(cachingHiveMetastore.isEmpty(), "CachingHiveMetastore should not be set when using ImpersonationCachingHiveMetastoreFactory");
+            Optional<CachingHiveMetastore> impersonationCachingHiveMetastore = Optional.of((CachingHiveMetastore) impersonationCachingHiveMetastoreFactory.createMetastore(Optional.of(session.getIdentity())));
+            doFlushMetadataCache(session, impersonationCachingHiveMetastore, schemaName, tableName, partitionColumns, partitionValues);
+        }
+        else {
+            doFlushMetadataCache(session, cachingHiveMetastore, schemaName, tableName, partitionColumns, partitionValues);
+        }
+    }
+
+    private void doFlushMetadataCache(ConnectorSession session, Optional<CachingHiveMetastore> cachingHiveMetastore, Optional<String> schemaName, Optional<String> tableName, List<String> partitionColumns, List<String> partitionValues)
+    {
         if (cachingHiveMetastore.isEmpty() && glueCache.isEmpty()) {
             // TODO this currently does not work. CachingHiveMetastore is always bound for metastores other than Glue, even when caching is disabled,
             //  so for consistency we do not discern between GlueCache NOOP and real.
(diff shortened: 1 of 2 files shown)


plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
```diff
@@ -134,7 +135,7 @@ public void flushMetadataCache(
 
     private void doFlushMetadataCache(ConnectorSession session, Optional<String> schemaName, Optional<String> tableName, List<String> partitionColumns, List<String> partitionValues)
     {
-        if (cachingHiveMetastore.isEmpty() && glueCache.isEmpty()) {
+        if (!(hiveMetadataFactory instanceof ImpersonationCachingHiveMetastoreFactory) && cachingHiveMetastore.isEmpty() && glueCache.isEmpty()) {
```
ssheikin: Shouldn't flush procedure also work for other types of caching, not only `ImpersonationCaching`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
```diff
@@ -144,6 +145,11 @@ private void doFlushMetadataCache(ConnectorSession session, Optional<String> sch
                 partitionColumns.size() == partitionValues.size(),
                 "Parameters partition_column and partition_value should have same length");
 
+        Optional<CachingHiveMetastore> cachingHiveMetastore = this.cachingHiveMetastore;
+        if (hiveMetadataFactory instanceof ImpersonationCachingHiveMetastoreFactory impersonationCachingHiveMetastoreFactory) {
+            cachingHiveMetastore = Optional.of((CachingHiveMetastore) impersonationCachingHiveMetastoreFactory.createMetastore(Optional.of(session.getIdentity())));
```
chenjian2664: When the `hiveMetadataFactory` is a `ImpersonationCachingHiveMetastoreFactory` does that means the `cachingHiveMetastore` always not exists, otherwise we may miss flushing the cache.

chenjian2664: maybe we could move the logic at the beginning of the method, after determine the metastore we can keep the remaining logic.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
ebyhr: Sorry, I’m not sure which caching you’re referring to. Could you elaborate?

ebyhr:
CI hit #27082

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
ebyhr: Added the following check: 
```java
            checkState(cachingHiveMetastore.isEmpty(), "CachingHiveMetastore should not be set when using ImpersonationCachingHiveMetastoreFactory");
```

ebyhr:
Rebased on master to resolve conflicts.

lib/trino-metastore/src/main/java/io/trino/metastore/HiveMetastoreFactory.java
```diff
@@ -35,25 +35,27 @@ default boolean hasBuiltInCaching()
...
     }
 
     class StaticHiveMetastoreFactory
             implements HiveMetastoreFactory
     {
         private final HiveMetastore metastore;
+        private final boolean impersonationEnabled;
 
-        private StaticHiveMetastoreFactory(HiveMetastore metastore)
+        private StaticHiveMetastoreFactory(HiveMetastore metastore, boolean impersonationEnabled)
```
chenjian2664, line 49: It's actual always `false` in current code

ebyhr: No, it can be true when it's called from HiveQueryRunner.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
kokosing: Can we simply remove this condition?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
```diff
@@ -144,6 +145,11 @@ private void doFlushMetadataCache(ConnectorSession session, Optional<String> sch
                 partitionColumns.size() == partitionValues.size(),
                 "Parameters partition_column and partition_value should have same length");
 
+        Optional<CachingHiveMetastore> cachingHiveMetastore = this.cachingHiveMetastore;
```
kokosing: Maybe extract a method from the below code and pass `cachingHiveMetastore` as parameter? Then if a condition above is needed I woudl:

```
if (impersonationEnabled...) {
   /// flush impersonation caches}
}
else {
  // flush regular caches
}
```

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
ebyhr: Do you mean the entire condition? The 1st condition (`ImpersonationCachingHiveMetastoreFactory`) is removed by the fixup commit.

ebyhr:
Just squashed fixup commits. 

> Don't we need io.trino.plugin.hive.metastore.cache.TestCachingHiveMetastore but with the impersonation?

Ideally we need, it should have existed even before this PR though. I think that would be overkill for this PR.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/procedure/FlushMetadataCacheProcedure.java
ssheikin: lgtm now

kokosing (approved):

kokosing:
Thank you!

PR #27063, opened 2025-10-23
Remove unused ConnectorSplit.getSplitInfo method

chenjian2664:
## Release notes

```markdown
## SPI
* Remove the deprecated `ConnectorSplit.getSplitInfo` method. ({issue}`27063`)
```

## Summary by Sourcery

Remove the deprecated getSplitInfo method and related unused imports from ConnectorSplit interface

Enhancements:
- Remove deprecated getSplitInfo method from ConnectorSplit interface

Chores:
- Remove unused imports of JsonIgnore and Map in ConnectorSplit

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplit.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplit.java
@@ -13,12 +13,10 @@
  */
 package io.trino.spi.connector;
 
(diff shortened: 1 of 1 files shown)


ebyhr:
> (x) This is not user-visible or is docs only, and no release notes are required.

I think we should mention this change in a release note because it's a change in SPI. A build will fail if third party developers were implementing this method in their connector split class.

Issue #27068, opened 2025-10-23
Docs-only PRs cannot be merged


findepi:
https://github.com/trinodb/trino/pull/27036 introduced a `build-success` job in `ci` workflow and later it was made required for a PR to be merged.
This works as intended for regular PRs.
It does not work for docs only PRs. For these, we don't trigger `ci` workflow at all, so `build-success`  is not reported and such a PR cannot be merged. This blocks https://github.com/trinodb/trino/pull/26726, so I am marking this a release blocker. 

cc @martint @ebyhr

findepi:
Solving this for docs only PRs is simple -- just adding `build-success`  do docs workflow would suffice.
However, this won't work for PRs that change both docs and code. For such, `docs` and `ci` workflows are both triggered, and we don't want to different workflows to report status for the same ``build-success``  expected job.

Issue #26943, opened 2025-10-14
`setup` action sometimes fails in `Clearing up disk usage` step


findepi:
https://github.com/trinodb/trino/actions/runs/18470171202/job/52621935606?pr=26918

```
Prepare all required actions
Getting action download info
Download action repository 'actions/setup-java@v4' (SHA:c5195efecf7bdfc987ee8bae7a71cb8b11521c00)
Download action repository 'actions/cache@v4' (SHA:0057852bfaa89a56745cba8c7296529d2fc39830)
Run ./.github/actions/setup
Run expected=
Run .github/bin/git-fetch-base-ref.sh
GITHUB_BASE_REF is not set or is master, not fetching it
Run ./.github/bin/free-disk-space.sh
Disk space usage before cleaning:
Filesystem     1K-blocks     Used Available Use% Mounted on
/dev/root       75085112 52344880  22723848  70% /
Clearing up disk usage
  Reading package lists...
  Building dependency tree...
  Reading state information...
  Package 'aspnetcore-runtime-6.0' is not installed, so not removed
  Package 'aspnetcore-targeting-pack-6.0' is not installed, so not removed
  Package 'firefox-dbg' is not installed, so not removed
  Package 'firefox-dev' is not installed, so not removed
  Package 'firefox-geckodriver' is not installed, so not removed
  Package 'firefox-mozsymbols' is not installed, so not removed
  Package 'firefox-esr' is not installed, so not removed
  Package 'firefox-esr-l10n-all' is not installed, so not removed
  Package 'firefox-l10n-all' is not installed, so not removed
  Removing llvm-17 (1:17.0.6-9ubuntu1) ...
  Removing libprotobuf-lite32t64:amd64 (3.21.12-8.2ubuntu0.2) ...
  Removing libtdb1:amd64 (1.4.10-1build1) ...
  Removing libtimedate-perl (2.3300-2) ...
  Removing liburi-perl (5.27-1) ...
  Removing libz3-dev:amd64 (4.8.12-3.1build1) ...
  Removing libz3-4:amd64 (4.8.12-3.1build1) ...
  Removing llvm-16 (1:16.0.6-23ubuntu4) ...
  Removing llvm-16-linker-tools (1:16.0.6-23ubuntu4) ...
  Removing llvm-16-runtime (1:16.0.6-23ubuntu4) ...
  Removing llvm-16-tools (1:16.0.6-23ubuntu4) ...
[...]

PR #27070, opened 2025-10-23
Remove separate docs GitHub workflow

findepi:
Before the change, we had two workflows: `ci` and `docs`.

- for code changes we run `ci`
- for docs changes we run `docs`
- for code and docs changes we run both, but the docs PR was configured
  to do nothing

`ci` has `build-success` status marker and it's required job for a PR to
be merged.

We could add similar job to `docs` workflow.

- it can also be called `build-success`. This would work for docs-only
  PRs, but not for code & docs PRs.
- it can be called `docs-build-success` and also become required, but
  that doesn't solve original problem at all.

Solving this requires that

- the workflows to be truly mutually exclusive (for now they are
  logically mutually exclusive but not really) OR
- there is only one workflow, with `docs` workflow merging into `ci`
  workflow.

This change implements the latter option, as it's conceptually simpler
and so should be better from long term maintainability perspective.

- The `docs` workflow is removed.
- The `ci` workflow is run on all PRs.
- The `docs-checks` is already covered by `maven-checks` job in `ci`
  workflow.

The removed `docs` workflow had one more job: `test-docs`. This was
supposed to run Java tests that check docs changes. This is currently
lost. However, it wasn't working reliably anyway. It would run on
docs-only changes, but would not run on PRs mixing docs changes with any
other code changes, unless those changes would trigger these tests to
run (https://github.com/trinodb/trino/issues/27076). This is left to be fixed later.

- fixes https://github.com/trinodb/trino/issues/27068

## Summary by Sourcery

Merge the separate docs GitHub workflow into the main CI workflow by introducing path-based filtering, reintroducing a final build-success job, and removing docs.yml.

Enhancements:
- Update build-matrix-from-impacted.py to run all modules when the impacted list is empty

Build:
- Update docs/pom.xml to run code generation in the generate-sources phase instead of validate
- Modify cleanup.yml to cancel only the ci.yml workflow

CI:
- Merge docs.yml into ci.yml and remove the standalone docs workflow
- Introduce a path-filters job to classify docs vs non-docs changes and conditionally skip CI jobs for docs-only PRs
- Reinstate the build-success job as the final aggregator of all CI job results

Tests:
- Refactor TestCiWorkflow to use a BUILD_SUCCESS constant and add a test ensuring the build-success job is last in ci.yml

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -47,6 +47,19 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  path-filters:
+    runs-on: ubuntu-latest
+    outputs:
+      docs: ${{ steps.filter.outputs.docs }}
+      non_docs: ${{ steps.filter.outputs.non_docs }}
+    steps:
+      - uses: dorny/paths-filter@de90cc6fb38fc0963ad72b210f1f284cd68cea36 # v3
+        id: filter
+        with:
+          filters: |
+            docs: 'docs/**'
+            non_docs: '!docs/**'
+
   maven-checks:
     runs-on: ubuntu-latest
     name: maven-checks ${{ matrix.java-version }}
@@ -86,6 +99,8 @@ jobs:
         run: rm -rf ~/.m2/repository/io/trino/trino-*
 
   artifact-checks:
+    needs: path-filters
+    if: needs.path-filters.outputs.non_docs == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 45
     steps:
@@ -119,8 +134,9 @@ jobs:
         run: core/docker/build.sh
 
   check-commits-dispatcher:
+    needs: path-filters
+    if: github.event_name == 'pull_request' && needs.path-filters.outputs.non_docs == 'true'
     runs-on: ubuntu-latest
-    if: github.event_name == 'pull_request'
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
@@ -172,6 +188,8 @@ jobs:
           base_ref: ${{ github.event.pull_request.base.ref }}
 
   error-prone-checks:
+    needs: path-filters
+    if: needs.path-filters.outputs.non_docs == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 45
     steps:
@@ -200,6 +218,8 @@ jobs:
             -pl '!:trino-docs,!:trino-server'
 
   test-jdbc-compatibility:
+    needs: path-filters
+    if: needs.path-filters.outputs.non_docs == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
@@ -240,6 +260,8 @@ jobs:
           upload-heap-dump: ${{ env.SECRETS_PRESENT == '' && github.event_name == 'pull_request' && github.event.pull_request.head.repo.full_name != github.repository }}
 
   hive-tests:
+    needs: path-filters
+    if: needs.path-filters.outputs.non_docs == 'true'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
@@ -296,6 +318,8 @@ jobs:
           github_token: ${{ secrets.GITHUB_TOKEN }}
 
   test-other-modules:
+    needs: path-filters
+    if: needs.path-filters.outputs.non_docs == 'true'
     runs-on: ubuntu-latest
     timeout-minutes: 60
     steps:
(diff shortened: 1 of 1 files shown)


.github/bin/build-matrix-from-impacted.py
```diff
@@ -98,7 +98,7 @@ def build(matrix_file, impacted_file, output_file):
 def check_modules(modules, impacted):
     if isinstance(modules, str):
         modules = [modules]
-    if impacted and not any(module in impacted for module in modules):
```
findepi, line 101: this was needed so that master branch builds run all the tests. now they will run none

findepi:
The workflow changes were tested here

- https://github.com/trinodb/trino/pull/27066
- https://github.com/trinodb/trino/pull/27067
- https://github.com/trinodb/trino/pull/27071
- https://github.com/trinodb/trino/pull/27073
- https://github.com/trinodb/trino/pull/27074

testing/trino-tests/src/test/java/io/trino/tests/ci/TestCiWorkflow.java
```diff
@@ -137,6 +139,20 @@ public void testBuildSuccessDependencies()
...
+    public void testBuildSuccessIsLast()
+            throws Exception
+    {
+        String buildSuccessJobName = BUILD_SUCCESS;
+        Yaml yaml = new Yaml();
+        Map<?, ?> workflow = yaml.load(new StringReader(Files.readString(findRepositoryRoot().resolve(CI_YML_REPO_PATH))));
+        Map<String, ?> jobs = getMap(workflow, "jobs");
+        // This assumes the `jobs` map preserves source order
+        assertThat(getLast(jobs.keySet()))
+                .describedAs("The %s job is logically last and depends on all others, we want it to be last in the source file".formatted(buildSuccessJobName))
```
electrum: `describedAs()` takes a format string

testing/trino-tests/src/test/java/io/trino/tests/ci/TestCiWorkflow.java
```diff
@@ -101,7 +103,7 @@ public void testUploadTestResultsCondition()
     public void testBuildSuccessDependencies()
             throws Exception
     {
-        String buildSuccessJobName = "build-success";
+        String buildSuccessJobName = BUILD_SUCCESS;
```
electrum: Inline these variables. They don't serve a purpose anymore and are longer than the constant.

testing/trino-tests/src/test/java/io/trino/tests/ci/TestCiWorkflow.java
```diff
@@ -137,6 +139,20 @@ public void testBuildSuccessDependencies()
                 .isEqualTo(expectedRunDefinition.toString());
     }
 
+    @Test
+    public void testBuildSuccessIsLast()
+            throws Exception
+    {
+        String buildSuccessJobName = BUILD_SUCCESS;
```
electrum: Inline this

docs/pom.xml
```diff
@@ -51,7 +51,8 @@
                         <goals>
                             <goal>generate-thrift-idl</goal>
                         </goals>
-                        <phase>validate</phase>
+                        <!-- TODO should this be `test` phase? the generated file seems to be used below just for a test with `diff` -->
```
electrum: This is also included by `thrift.md` so that the IDL is in the documentation, so `generate-sources` seems like the correct phase. Remove the TODO.

docs/pom.xml
```diff
@@ -83,7 +84,8 @@
                         <goals>
                             <goal>java</goal>
                         </goals>
-                        <phase>validate</phase>
+                        <!-- TODO should this be `test` phase? -->
```
electrum: Make both of these `test` and remove the TODO since they can be considered tests (you could think of them as tests or validation but they're definitely not source generation).

docs/pom.xml
electrum: The phases are a bit weird since this isn't a Java module. We could also change the Sphinx execution to run in `prepare-package` so that it's strictly before `package` (even though that's guaranteed by the plugin declaration order in the POM).

electrum (approved):
Will build the docs module if the dependent Java module changes? Specifically, if you add another reserved word for the parser, does the `ReservedIdentifiers` check still run?

testing/trino-tests/src/test/java/io/trino/tests/ci/TestCiWorkflow.java
findepi: i kept them _intentionally_ to indicate the test method operates on single thing across its flow.
But can also inline 🤷‍♂️

docs/pom.xml
findepi: ofc they are test, that's why i put this todo comment. thanks for confirming the test phase will work for this

findepi:
> Will build the docs module if the dependent Java module changes? Specifically, if you add another reserved word for the parser, does the `ReservedIdentifiers` check still run?

Yes

- a build against current master https://github.com/trinodb/trino/pull/27085
- a build against this PR https://github.com/trinodb/trino/pull/27086

Before and after the changes, this is covered by `ci` workflow, the `maven-checks` job.
The `docs` workflow (the one being removed here) does _nothing_ for a PR that has _any_ non-docs changes.

findepi:
CI
- https://github.com/trinodb/trino/issues/26943
- https://github.com/trinodb/trino/issues/23596

i am going to squash fixups now

PR #27080, opened 2025-10-23
Extend TestSchedulingUtils with top level union

losipiuk:
Add some missing test coverage

## Summary by Sourcery

Tests:
- Add test for canStream over a union operator with multiple child sources

### DIFF
--- a/core/trino-main/src/test/java/io/trino/execution/scheduler/TestSchedulingUtils.java
+++ b/core/trino-main/src/test/java/io/trino/execution/scheduler/TestSchedulingUtils.java
@@ -73,6 +73,34 @@ public class TestSchedulingUtils
     }
 
     @Test
+    public void testCanStreamUnion()
+    {
+        /*
+                  parent(union)
+                   /    \    \
+     -------------------------------------- stage boundary
+                  a     b     c
+         */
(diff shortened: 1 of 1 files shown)


core/trino-main/src/test/java/io/trino/execution/scheduler/TestSchedulingUtils.java
```diff
@@ -72,6 +72,34 @@ public void testCanStreamNoJoin()
...
+        SubPlan bSubPlan = valuesSubPlan("b");
+        RemoteSourceNode remoteSourceB = remoteSource("b");
+
+        SubPlan cSubPlan = valuesSubPlan("c");
+        RemoteSourceNode remoteSourceC = remoteSource("c");
+
+        SubPlan parentSubPlan = createSubPlan(
+                "parent",
+                union("union", ImmutableList.of(remoteSourceA, remoteSourceB, remoteSourceC)),
+                ImmutableList.of(aSubPlan, bSubPlan));
```
chenjian2664: Why `cSubPlan` isn't a child of "parent"?

losipiuk: Yeah - mistake - thanks

PR #27088, opened 2025-10-24
Minor cleanup on free-disk-space.sh

Praveen2112:
## Description

This PR attempts to speed up the `free-disk-space.sh` which takes more than 15mins to execute. We first remove`man-db` as updating it takes some time - Even official runners from Nov 10th would have them removed - https://github.com/actions/runner-images/issues/13213

Additionally we try to use `rsync` to drop the files instead of `rm -rf`.

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Enhancements:
- Add logging of disk usage before cleaning in free-disk-space script

## Summary by Sourcery

Optimize disk cleanup script by excluding man-db, introducing rsync-based deletion for directories, and logging pre-cleanup disk usage.

Enhancements:
- Exclude man-db from the list of Ubuntu package removals to speed up updates
- Add delete_directories_with_rsync function to use rsync for faster purging of toolchain directories
- Replace bulk rm -rf directory removal with rsync-based deletion
- Log disk usage before running cleanup steps to provide pre-cleanup metrics

### DIFF
--- a/.github/bin/free-disk-space.sh
+++ b/.github/bin/free-disk-space.sh
@@ -23,16 +23,29 @@ function free_up_disk_space_ubuntu()
     sudo apt-get autoclean -y
 
     echo  "Removing toolchains"
-    sudo rm -rf \
-      /usr/local/graalvm \
-      /usr/local/lib/android/ \
-      /usr/share/dotnet/ \
-      /opt/ghc/ \
-      /usr/local/share/boost/ \
-      "${AGENT_TOOLSDIRECTORY}"
+    directories_to_be_removed=(
+      "/usr/local/graalvm/"
+      "/usr/local/lib/android/"
+      "/usr/share/dotnet/"
+      "/opt/ghc/"
+      "/usr/local/share/boost/"
+      "${AGENT_TOOLSDIRECTORY}")
 
-      echo "Prune docker images"
-      sudo docker system prune --all -f
+    delete_directories_with_rsync "${directories_to_be_removed[@]}"
+
+    echo "Prune docker images"
+    sudo docker system prune --all -f
+}
+
+function delete_directories_with_rsync()
+{
+    sudo mkdir /tmp/empty
+    for dir in "$@"; do
+        echo "Deleting contents of $dir using rsync"
+        sudo rsync --delete -a /tmp/empty/ "$dir"
+        sudo rmdir "$dir"
+    done
+    sudo rmdir /tmp/empty
 }
 
 echo "Disk space usage before cleaning:"


.github/bin/free-disk-space.sh
```diff
@@ -22,16 +23,27 @@ function free_up_disk_space_ubuntu()
...
+
+    echo "Prune docker images"
+    sudo docker system prune --all -f
+}
+
+function delete_directories_with_rsync()
+{
+    for dir in "$@"; do
+        echo "Deleting contents of $dir using rsync"
+        sudo rsync --delete -a --exclude='*' ./ "$dir"
```
findepi: what's going on here?

.github/bin/free-disk-space.sh
```diff
@@ -22,16 +23,29 @@ function free_up_disk_space_ubuntu()
...
+    echo "Prune docker images"
+    sudo docker system prune --all -f
+}
+
+function delete_directories_with_rsync()
+{
+    sudo mkdir /tmp/empty
+    for dir in "$@"; do
+        echo "Deleting contents of $dir using rsync"
+        sudo rsync --delete -a /tmp/empty/ "$dir"
```
findepi, line 45: why would we want to use rync when all we want is rm?

Praveen2112: If the number of files in a directory is large, them `rm -rf` might take sometime, which leads to increase in the execution time crossing the limit i.e 15mins

Praveen2112: This `rsync --delete -a /tmp/empty/ "$dir"` is more of mimicking `rm -rf`

.github/bin/free-disk-space.sh
Praveen2112: Moving to this thread - https://github.com/trinodb/trino/pull/27088#discussion_r2461001146

.github/bin/free-disk-space.sh
findepi: Why would rsync be faster than rm? is it because it does different system calls, it runs things in threads (good for networks), or?

chenjian2664: I asked the ai and give context "work with data heavy systems", it answer:
```
Because you work with data-heavy systems (databases, lakes, large file trees) you’ll likely encounter directories with very many files and subdirectories. In such cases:
	•	Using rm -rf triggers lots of individual filesystem operations (unlink, rmdir) in what may be a sub-optimal order; the overhead (and metadata cost) becomes dominant.
	•	Using rsync -a --delete (with an empty source) shifts the pattern: rsync walks the destination tree, determines all items present, then issues deletions in a more efficient pattern (for many files) which tends to reduce metadata churn and avoid worst-case filesystem behaviour.
	•	Thus for very large file trees, rsync may “appear” much faster in wall-time, even though conceptually you’re still deleting everything.
	•	Given your context (large datasets, perhaps S3 or other object stores, etc) you may adapt: for local filesystem large-scale deletion this trick is a practical tool.
```

findepi: my ai insist there is no system call to delete files in bulk... either can be wrong (or both)

Praveen2112: We don't delete them in bulk - In short we found `rm -rf` on android libs were taking some time more than 15 mins which caused the failure so was exploring `rsync ` which is to used to sync directories and we tweak the arguments a bit so as to support delete - From the `rmdir` executed successfully we could assert the directory is empty which means the files are deleted - The disk usage was reduced from 75% to 40% like always

PR #27089, opened 2025-10-24
Add Trino 478 release notes

findepi:
Release notes for the upcoming Trino 478 release.

- supersedes https://github.com/trinodb/trino/pull/26726 per https://github.com/trinodb/trino/pull/26726#issuecomment-3442629763

## Summary by Sourcery

Add documentation for Trino 478 release notes and include it in the release notes index

Documentation:
- Add release-478.md with comprehensive notes covering general improvements, security, Web UI, Docker image, and connector updates
- Include release-478 entry in the main release notes table of contents

## Summary by Sourcery

Add the release notes for Trino 478 and register the new documentation in the release notes index.

Documentation:
- Create release-478.md with detailed notes on general improvements, security updates, Docker image changes, and connector enhancements and fixes
- Include the new release-478 entry in the main release notes table of contents

### DIFF
--- a/docs/src/main/sphinx/release/release-478.md
+++ b/docs/src/main/sphinx/release/release-478.md
@@ -1,0 +1,72 @@
+# Release 478 (29 Oct 2025)
+
+## General
+
+* Include lineage information for columns used in `UNNEST` expressions. ({issue}`16946`)
+* Add support for limiting which retry policies a user can select. This can be configured using
+  the `retry-policy.allowed` option. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Allow dropping catalogs that failed to load correctly. ({issue}`26918`)
+* Improve performance of queries with an `ORDER BY` clause using `varchar` or `varbinary` types. ({issue}`26725`)
+* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
+* Improve performance of queries involving `JOIN` when the join spills to disk. ({issue}`26076`)
+* Fix potential incorrect results when query uses `row` type. ({issue}`26806`)
+* Include catalogs that failed to load in the `metadata.catalogs` table. ({issue}`26918`)
+* Fix `EXPLAIN ANALYZE` planning so that it executes with the same plan as would be used to execute the query
+  being analyzed. ({issue}`26938`)
+* Fix incorrect results when using logical navigation function `FIRST` in row pattern recognition. ({issue}`26981`)
+
+## Security
+
+* Propagate `queryId` to the [Open Policy Agent](/security/opa-access-control)
+  authorizer. ({issue}`26851`)
+
+## Docker image
+
+* Run Trino on JDK 25.0.0 (build 36). ({issue}`26693`)
+
+## Delta Lake connector
+
+* Fix failure when reading a `map(..., json)` column when the map item value is `NULL`. ({issue}`26700`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+
+## Google Sheets connector
+
+* Fix potential query failure when the `gsheets.delegated-user-email` configuration property
+  is used. ({issue}`26501`)
+
+## Hive connector
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+* Improve performance of queries using complex predicates on `$path` column. ({issue}`27000`)
+* Fix writing ORC files to ensure that dates and timestamps before `1582-10-15` are read correctly by Apache Hive. ({issue}`26507`)
+
+## Hudi connector
+
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+
+## Iceberg connector
+
+* Improve performance when writing sorted tables and `iceberg.sorted-writing.local-staging-path`
+  is set. ({issue}`24376`)
+* Return execution metrics while running the `remove_orphan_files` command. ({issue}`26661`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+* Collect distinct values count on all columns when replacing tables. ({issue}`26983`)
+* Fix failure due to column count mismatch when executing the `add_files_from_table`
+  procedure. ({issue}`26774`)
+* Fix failure when executing `optimize_manifests` on tables without a snapshot. ({issue}`26970`)
+* Fix incorrect results when reading Avro files migrated from Hive. ({issue}`26863`)
+* Fix failure when executing `SHOW CREATE SCHEMA` on a schema with unsupported properties
+  with REST, Glue or Nessie catalog. ({issue}`24744`)
+* Fix failure when running `EXPLAIN` or `EXPLAIN ANALYZE` on `OPTIMIZE` command. ({issue}`26598`)
+
+## Kafka connector
+
+* Fix failure when filtering partitions by timestamp offset. ({issue}`26787`)
+
+## SPI
+
+* Remove default implementation from `Connector.shutdown()`. ({issue}`26718`)
+* Remove the deprecated `ConnectorSplit.getSplitInfo` method. ({issue}`27063`)
+* Deprecate `io.trino.spi.type.Type#appendTo` method. ({issue}`26922`)
--- a/docs/src/main/sphinx/release.md
+++ b/docs/src/main/sphinx/release.md
@@ -6,6 +6,7 @@
 ```{toctree}
 :maxdepth: 1
 
+release/release-478
 release/release-477
 release/release-476
 release/release-475


docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
+* Allow dropping an uninitialized catalog that failed to load. ({issue}`26918`)
+* Improve performance of queries with an `ORDER BY` clause. ({issue}`26725`)
+* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
+* Improve performance of queries with joins which spill to disk. ({issue}`26076`)
+* Fix potential incorrect results when reading `row` type. ({issue}`26806`)
+* Return all catalogs, including uninitialized ones, for queries from `metadata.catalogs`. ({issue}`26918`)
+* Ensure that queries with and without `EXPLAIN ANALYZE` are planned identically. ({issue}`26938`)
+* In row pattern matching, restrict logical navigations to current match in running semantics. ({issue}`26981`)
```
findepi: from @losipiuk (https://github.com/trinodb/trino/pull/26726#discussion_r2459416279):

The nomenclature here is technical. Does it match what we us in docs?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
+# Release 478 (22 Oct 2025)
+
+## General
+
+* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
+* Add `retry-policy.allowed` configuration property to specify which retry
+  policies can be selected by the user. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint for integration with Trino Gateway. ({issue}`26548`)
```
findepi: from https://github.com/trinodb/trino/pull/26726#discussion_r2459196190

@findepi 

> BTW I am not convinced we should have added that.
> Trino and Trino Gateway are both projects we manage, so we can have coupling between them. That does not mean we should or want such coupling. Everything is a trade off, of course. If this endpoint exposes information previously unavailable, and also interesting to Trino Gateway only, we perhaps want to have it. But it doesn't look like this is the case here. The endpoint exposes cluster utilization metrics (memory, load), which are generally useful for observability and also for any other query routing logic. We should expose them in gateway-agnostic manner, rather than build a public-but-do-not-use-kind-of-internal-yet-public API.
> 
> cc @andythsu @wendigo @martint

@wendigo 

> We've discussed that already and settled on the decision that it's ok to have a client specific endpoint (like we have for the UI). We can revisit this decision but if it blocks the release we should just revert the change and move on

@findepi 

> > ok to have a client specific endpoint (like we have for the UI)
> 
> UI is not the best example, because it's maintained within this repo. The UI endpoints can change as the UI needs evolve.
> 
> This new endpoint is different. It's meant for the Trino Gateway, so in theory it should not be used by any other software. Any other query routing software will needs its own endpoint, which is quite undesirable. All this while the information exposed is no way trino gateway specific and could be exposed in a more reusable manner.
> 
> > We can revisit this decision but if it blocks the release we should just revert the change and move on
> 
> Agreed!
> 
>

findepi:
draft until https://github.com/trinodb/trino/pull/27089/files#r2459921217

i cannot 'request changes' on a PR that I opened, treat draft status as meaning just that

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+* Allow dropping an uninitialized catalog that failed to load. ({issue}`26918`)
```
findepi: do we want to quantify this with "When using dynamic catalogs" or similar?

```
* When using dynamic catalogs, allow dropping ...
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,78 @@
...
+
+* Fix potential query failure when the `gsheets.delegated-user-email` configuration property
+  is used. ({issue}`26501`)
+
+## Hive connector
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+* Improve performance of queries using complex predicates on `$path` column. ({issue}`27000`)
+* Prevent writing invalid dates and timestamps before `1582-10-15` by the ORC writer. ({issue}`26507`)
```
findepi: @marcinsbd @raunaqmorarka does this apply to Iceberg connector too?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
+# Release 478 (22 Oct 2025)
+
+## General
+
+* Add support for column lineage in `UNNEST` clauses. ({issue}`16946`)
```
martint:
```
* Include lineage information for columns used in `UNNEST` expressions. ({issue}`16946`)
```

(it was already provided in the PR description)

docs/src/main/sphinx/release/release-478.md
martint: What's an "uninitialized catalog"?

docs/src/main/sphinx/release/release-478.md
Praveen2112: Yes

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+* Add `retry-policy.allowed` configuration property to specify which query retry
```
martint:
```
* Add support for limiting which retry policies a user can select. This can be configured using the `retry-policy.allowed` option. ({issue}`26628`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+  policies can be selected by the user. ({issue}`26628`)
+* Add support for loading plugins from multiple directories. ({issue}`26855`)
+* Add the `/v1/integrations/gateway` endpoint that exposes additional cluster metrics for Trino Gateway.
```
martint: This doesn't need to be mentioned in release notes. It's a plumbing API and not "user visible"

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+  ({issue}`26548`)
+* Allow dropping an uninitialized catalog that failed to load. ({issue}`26918`)
+* Improve performance of queries with an `ORDER BY` clause using `varchar` or `varbinary` types. ({issue}`26725`)
+* Improve performance of `MERGE` statements involving a `NOT MATCHED` case. ({issue}`26759`)
+* Improve performance of queries with joins which spill to disk. ({issue}`26076`)
```
martint:
```
* Improve performance of queries involving `JOIN` when spilling is enabled. ({issue}`26076`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+* Fix potential incorrect results when reading `row` type. ({issue}`26806`)
```
martint: What does "when reading row type" mean? Read from where? Or is it generally when processing values of row type?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+* Return all catalogs, including uninitialized ones, for queries from `metadata.catalogs`. ({issue}`26918`)
```
martint:
```
* Include catalogs that failed to load in the `metadata.catalogs` table. ({issue}`26918`)
```

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+* Fix `EXPLAIN ANALYZE` planning so that it executes with the same plan as would be used to execute the query
+  being analyzed. ({issue}`26938`)
+* Fix row pattern matching logical navigations in running semantics to be always constraint to current match.
```
martint: This is too technical. What's the effect of "returning a position outside the current match"? What's the user-visible behavior?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+  Previously, a logical navigation function such as `FIRST` could return position outside the current match.
+  ({issue}`26981`)
+* Return correct "Table does not exist" error when querying a non-existent table in an existing schema within
```
martint: This doesn't need to be documented.

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+  `system` catalog. ({issue}`27047`)
+* Avoid logging an exception when listing columns from tables from a set of catalogs that includes the `system`
```
martint: This doesn't need to be documented.

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+* Propagate `queryId` to the [Open Policy Agent](/security/opa-access-control)
+  authorizer. ({issue}`26851`)
+
+## Docker image
+
+* Run Trino on JDK 25.0.0 (build 36). ({issue}`26693`)
+
+## Delta Lake connector
+
+* Fix failure when reading `map` type with `json` value type when a value is `NULL`. ({issue}`26700`)
```
martint: This is still not clear to me. Does "map type" refer to Delta Lake's type? Does "json value type" refer to Trino's type?

docs/src/main/sphinx/release/release-478.md
```diff
@@ -0,0 +1,80 @@
...
+
+* Fix potential query failure when the `gsheets.delegated-user-email` configuration property
+  is used. ({issue}`26501`)
+
+## Hive connector
+
+* Add support for reading encrypted Parquet files. ({issue}`24517`, {issue}`9383`)
+* Deprecate the `gcs.use-access-token` configuration property. Use `gcs.auth-type` instead. ({issue}`26681`)
+* Improve performance of queries using complex predicates on `$path` column. ({issue}`27000`)
+* Fix writing invalid dates and timestamps before `1582-10-15` when writing ORC files by setting calendar type
```
martint: The "by setting calendar type ..." part is unnecessary. If someone wants to know the implementation details, they can look at the PR.

So, simplify it to:

```
* Fix writing invalid dates and timestamps before `1582-10-15` when writing ORC data.
```

docs/src/main/sphinx/release/release-478.md
findepi: one that failed to load correctly 

cc @kokosing @homar

docs/src/main/sphinx/release/release-478.md
findepi: cc @pettyjamesm @wendigo

docs/src/main/sphinx/release/release-478.md
findepi: for end users for sure. but we also document things relevant for operators. anyway removing.

docs/src/main/sphinx/release/release-478.md
findepi: fwiw, originally PR had this

> Fix reading `null` map with `VARIANT` as value type.

i believe "json value type" refers to Trino's type

docs/src/main/sphinx/release/release-478.md
findepi: i think the important part is that trino was reading that data back correctly (or as-if correctly), and some other systems were not. did you omit this part intentionally, @martint ?

docs/src/main/sphinx/release/release-478.md
pettyjamesm: The issue wasn’t with “reading” exactly. It was a bug with RowBlockBuilder handling when starting offset != 0, so most situations were not affected but any situation involving block builder could be affected (issue was found with duplicate rows coming through a MERGE operation)

findepi:
```
Fix potential incorrect results when query uses `row` type.
```
?

docs/src/main/sphinx/release/release-478.md
martint: Fair enough. We can extend it with:

```
* Fix writing invalid dates and timestamps before `1582-10-15` in ORC data, which could affect correctness when reading them in other query engines such as Apache Hive.
```

docs/src/main/sphinx/release/release-478.md
martint: What about "map type"?

docs/src/main/sphinx/release/release-478.md
martint: cc @kasiafi

docs/src/main/sphinx/release/release-478.md
martint: Then the "that failed to load" part is redundant. I would just say:

```
* Allow dropping catalogs that failed to load correctly.
```

docs/src/main/sphinx/release/release-478.md
ebyhr: The map type also refers to Trino type in my understanding. The PR contains a below test:  
```sql
MAP(ARRAY['key1'], ARRAY[NULL])
```
The column type is `map(varchar, json)`.

ebyhr:
Added newer entries since Oct 23.

docs/src/main/sphinx/release/release-478.md
raunaqmorarka: Can we stick with the wording of the original release note:
```
Fix ORC files written by Trino to ensure that dates and timestamps before `1582-10-15` are read correctly by Apache Hive.
```
I find that to be far more clear, concise and faithful to what is actually implemented than all the edits since.

docs/src/main/sphinx/release/release-478.md
raunaqmorarka: "when spilling is enabled" is not an accurate description, performance is improved only when the join actually spills to disk.
So this should rather be 
```
Improve performance of queries involving `JOIN` when the join spills to disk
```
 or something along those lines.

docs/src/main/sphinx/release/release-478.md
raunaqmorarka: Not really, we may be writing the calendar type metadata in orc files regardless of the table format being used, but that does not mean the information is actually relevant in iceberg. Iceberg is mostly non-committal on this issue, see https://lists.apache.org/thread/hdzqzpg68wjphr9t4vcrnfjkcfczztyz
The best summarisation of the calendar situation in iceberg is
```
Iceberg uses ISO8601 calendar. For dates prior to the
Julien/Gregorian calendar, implementations are encouraged to use
proleptic-gregorian but this is left unspecified by the specification.
```
So I wouldn't make any claim about iceberg for this change's release note

marcinsbd: The idea of the change was to make the Trino transparent and explicit about which calendar is used when writing the date and timestamps to ORC files.
I'd suggest focusing on the current behaviour rather than potential outcomes (e.g. whether 3rd party tools will or will not respect the metadata entries).
A clear statement could be something like:
"ORC writer sets the Calendar type (proleptic Gregorian) used for writing dates and timestamps in the ORC footer ({issue}`26507`"
This change resolves the issue with Hive misinterpreting dates/timestamps before `1582-10-15`.
The issue was specifically targeted Hive only; Iceberg was not covered as port of this work.

docs/src/main/sphinx/release/release-478.md
```diff
@@ -25,14 +25,6 @@
 * Propagate `queryId` to the [Open Policy Agent](/security/opa-access-control)
   authorizer. ({issue}`26851`)
 
-## Web UI
-
-* Add support for filtering queries by `X-Trino-Trace-Token` value in the [](/admin/preview-web-interface). ({issue}`26447`)
-* Improve rendering performance of large query JSON in the [](/admin/preview-web-interface). ({issue}`26807`)
-* Fix rendering of large query plans in the [](/admin/preview-web-interface). ({issue}`26749`)
-* Fix rendering of the splits timeline for queued queries in the [](/admin/preview-web-interface) to prevent
-  blank screen ({issue}`26920`)
```
findepi: from https://github.com/trinodb/trino/pull/26726#discussion_r2452760888

@mosabua 

> Also ... not sure by @martint did not want changes for preview UI in the release notes in the past .. however I think we should have them

@findepi 

> The docs say "The Preview Web UI is not suitable for production usage, and only available for testing and evaluation purposes."
> Thus, any changes in that component are irrelevant from end user / trino operator standpoint. Release notes are for end users' and operators' consumption.
> 
> 
> Let's remove all entries pertaining the preview web ui

@raunaqmorarka 

> I don't agree with that reasoning at all. We're shipping the preview web UI because we want users to use that and retire the old UI. So it doesn't make sense to hide away the developments happening on that front. Also, the UI being new or not having all the features of the old one doesn't make it "not suitable for production usage". It's easy to toggle back and forth between new and old UI with a button after https://github.com/trinodb/trino/pull/26705 and `web-ui.preview.enabled` is enabled by default.

docs/src/main/sphinx/release/release-478.md
findepi: > I'd suggest focusing on the current behaviour rather than potential outcomes (e.g. whether 3rd party tools will or will not respect the metadata entries).

Release notes is something that end users read.
They may not understand consequences even if we describe code changes factually correct. 
If this change had no visible change for Trino and also no visible change for any other system we know of, we would not merge it.

Removing this note from Iceberg section per your feedback.

docs/src/main/sphinx/release/release-478.md
findepi: i am committing the Martin's wording now, but can also change to Raunaq's wording. Both work for me.
Leaving the conversation open. Up to @martint or anyone else to weight in.

docs/src/main/sphinx/release/release-478.md
findepi: Applied the latter version.

docs/src/main/sphinx/release/release-478.md
findepi: resolving, cont'd at https://github.com/trinodb/trino/pull/27089#discussion_r2461329302

docs/src/main/sphinx/release/release-478.md
martint: "written by Trino" is redundant, but that wording is good. Here's a simplified version. Let's go with this:

```
Fix writing ORC files to ensure that dates and timestamps before `1582-10-15` are read correctly by Apache Hive.
```

docs/src/main/sphinx/release/release-478.md
martint: Thanks @ebyhr. Then I'd suggest something like:

```
Fix failure when reading a `map(..., json)` column when the map item value is `NULL`.
```

docs/src/main/sphinx/release/release-478.md
findepi: i synced with @kasiafi  and changed the wording to

```
* Fix incorrect results when using logical navigation function `FIRST` in row pattern recognition. ({issue}`26981`)
```

findepi:
What we have so far is good. @martint approved offline.
I squashed and will merge when the build passes.

findepi:
there are two notesworthy PRs merged in the meantime  so
- https://github.com/trinodb/trino/pull/27145

PR #27093, opened 2025-10-24
Checkout repo before running path-filter job

Praveen2112:
## Description
The path-filter CI job when run on master requires the repo to be checked out. 

## Release notes

```markdown
## Section
```

## Summary by Sourcery

CI:
- Insert actions/checkout@v5 as the first step in the path-filter job with a conditional ref expression for repository_dispatch events

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,6 +53,12 @@ jobs:
       docs: ${{ steps.filter.outputs.docs }}
       non_docs: ${{ steps.filter.outputs.non_docs }}
     steps:
+      - uses: actions/checkout@v5
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -53,6 +53,12 @@ jobs:
       docs: ${{ steps.filter.outputs.docs }}
       non_docs: ${{ steps.filter.outputs.non_docs }}
     steps:
+      - uses: actions/checkout@v5
```
findepi, line 56: This won't work correctly.
It will detect no changes, and then jobs won't run.

see

- https://github.com/trinodb/trino/pull/27094

PR #27094, opened 2025-10-24
Fix `ci` workflow on `master`

findepi:
The ci workflow got broken in https://github.com/trinodb/trino/commit/3698755cc25d6ae37f2345134878b43453cd247c (https://github.com/trinodb/trino/pull/27070)
because `dorny/paths-filter` action works out of the box on PR runs
only.

It can be made to work on branch push runs too (requires
actions/checkout), but then it's meaningless -- there won't be any
chnages anyway. Or actually can be misleading, as it will compare the
branch against master, so may work surprisingly when the workflow is
enabled on any other branch.


findepi:
A branch `push` build -- https://github.com/trinodb/trino/actions/runs/18786453830/job/53605831163

PR #27095, opened 2025-10-24
Change to use Chainguard MinIO container

mosabua:
## Description

Change to use the MinIO container image from Chainguard Containers. Only "latest" is available so I am trying with that.

This PR is just an initial test for now. 

MinIO no longer publishes updated to their container images and even the latest one has CVEs reported against it.

## Release notes

### DIFF
--- a/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
+++ b/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
@@ -39,7 +39,7 @@ public class Minio
 
     private static final String MINIO_ACCESS_KEY = "minio-access-key";
     private static final String MINIO_SECRET_KEY = "minio-secret-key";
-    private static final String MINIO_RELEASE = "RELEASE.2025-01-20T14-49-07Z";
+    private static final String MINIO_RELEASE = "latest";
 
     private static final int MINIO_PORT = 9080; // minio uses 9000 by default, which conflicts with hadoop
     private static final int MINIO_CONSOLE_PORT = 9001;
@@ -69,7 +69,7 @@ public class Minio
 
     private DockerContainer createMinioContainer()
     {
-        DockerContainer container = new DockerContainer("minio/minio:" + MINIO_RELEASE, MINIO_CONTAINER_NAME)
+        DockerContainer container = new DockerContainer("cgr.dev/chainguard/minio:" + MINIO_RELEASE, MINIO_CONTAINER_NAME)
                 .withEnv(ImmutableMap.<String, String>builder()
                         .put("MINIO_ACCESS_KEY", MINIO_ACCESS_KEY)
                         .put("MINIO_SECRET_KEY", MINIO_SECRET_KEY)
--- a/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/SpoolingMinio.java
+++ b/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/SpoolingMinio.java
@@ -47,7 +47,7 @@ public class SpoolingMinio
     private static final String MINIO_SPOOLING_CONTAINER_NAME = "spooling-minio";
     private static final String MINIO_ACCESS_KEY = "minio-access-key";
     private static final String MINIO_SECRET_KEY = "minio-secret-key";
-    private static final String MINIO_RELEASE = "RELEASE.2025-01-20T14-49-07Z";
+    private static final String MINIO_RELEASE = "latest";
 
     private static final int MINIO_PORT = 9080; // minio uses 9000 by default, which conflicts with hadoop
     private static final int MINIO_CONSOLE_PORT = 9001;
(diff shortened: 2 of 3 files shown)


testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
```diff
@@ -39,7 +39,7 @@ public class Minio
 
     private static final String MINIO_ACCESS_KEY = "minio-access-key";
     private static final String MINIO_SECRET_KEY = "minio-secret-key";
-    private static final String MINIO_RELEASE = "RELEASE.2025-01-20T14-49-07Z";
+    private static final String MINIO_RELEASE = "latest";
```
findepi, line 42: >Change to use the MinIO container image from Chainguard Containers. Only "latest" is available so I am trying with that.

we need reproducible builds and deterministic tests. My "latest" won't be the same as your "latest".

how do we achieve that?

findepi: do we need to copy the container into ghcr trino org and tag ourselves?

electrum: It's possible to pull the image using the hash:
```shell
docker pull cgr.dev/chainguard/minio@sha256:66bd82c8fe5e75868ae7d0b2e102d9a0dcf971b270a41bd060a9e6a643476ff8
```
Maybe this also works with Testcontainers?

electrum: Confirmed, that syntax works fine. We should see if we can centralize the version instead of copying it in multiple places.

electrum:
The Minio container seems to be returning 403 errors. I see these in the `minio` container logs:
```
API: ListObjectsV2(bucket=test-bucket)
Time: 19:37:23 UTC 10/24/2025
DeploymentID: a950e157-b70b-463f-9ce4-82c331e7c000
RequestID: 18718486BF8418FE
RemoteHost: 172.18.0.4
Host: minio:9080
UserAgent: aws-sdk-java/2.36.1 md/io#sync md/http#Apache ua/2.1 api/S3#2.36.x os/Linux#6.11.0-1018-azure lang/java#25 md/OpenJDK_64-Bit_Server_VM#25+36-LTS md/vendor#Eclipse_Adoptium md/en_US md/kotlin/2.2.20-release-333 app/Trino m/D,N,N,C,e
Error: file access denied (cmd.StorageErr)
       7: internal/logger/logonce.go:118:logger.(*logOnceType).logOnceIf()
       6: internal/logger/logonce.go:149:logger.LogOnceIf()
       5: cmd/logging.go:116:cmd.internalLogOnceIf()
       4: cmd/metacache-walk.go:184:cmd.(*xlStorage).WalkDir.func3()
       3: cmd/metacache-walk.go:406:cmd.(*xlStorage).WalkDir()
       2: cmd/metacache-walk.go:419:cmd.(*xlStorageDiskIDCheck).WalkDir()
       1: cmd/metacache-set.go:1050:cmd.listPathRaw.func3()
```

Which results in this exception on the Hive metastore side:
```
Caused by: io.trino.hive.thrift.metastore.MetaException: Got exception: java.nio.file.AccessDeniedException s3a://test-bucket/test_hive_orc_legacy_date_compatibility_5l6d06aifd: getFileStatus on s3a://test-bucket/test_hive_orc_legacy_date_compatibility_5l6d06aifd: com.amazonaws.services.s3.model.AmazonS3Exception: Forbidden (Service: Amazon S3; Status Code: 403; Error Code: 403 Forbidden; Request ID: 18718486C7A41AC8; S3 Extended Request ID: 4b51e88a8a967ce09da1fc1d19f7299b638740a99d000967a4bf556e1518031f; Proxy: null), S3 Extended Request ID: 4b51e88a8a967ce09da1fc1d19f7299b638740a99d000967a4bf556e1518031f:403 Forbidden
```

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
mosabua: The digests and all will continue to be avaiable so we can either tag ourselves with something like a date value or we can use hashes.

electrum: We can use a single constant of the form
```java
MINIO_IMAGE = "cgr.dev/chainguard/minio@sha256:66bd82c8fe5e75868ae7d0b2e102d9a0dcf971b270a41bd060a9e6a643476ff8";

mosabua: That seems fine with me as well. Should I refactor the code to use a constant along those lines?

mosabua:
@imjasonh @xnox @amouat can you look at the output from @electrum and check out what is going on there? I have a feeling it has something do to with different user setup or permissions in our Chainguard container.

electrum:
Note that the log lines above are from the GHA logs for the failed jobs in this PR. I stripped out the prefix from each line to make them more readable.

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
electrum: Yeah, let’s do that if it’s possible without messing up the module structure. I haven’t looked at the dependencies between these locations.

xnox:
Chainguard build of minio defaults to running unprivileged.
The previous minio builds defaulted to running as root.

Thus to have compatibility, one would need to either ensure data / volume mounts use the same unprivileged uid shift inside the container,
or run as root at launch with `-u0` / `--user 0` in docker run option terms - adapted as appropriate for your container runtime.

wendigo:
I've applied some fixes in the https://github.com/trinodb/trino/pull/27131

Makes it usable in our CI :)

Issue #26984, opened 2025-10-16
Enhance 'unauthenticated' auth type for GCS storage


szisiu:
This is related to 
- https://github.com/trinodb/trino/issues/25810  (https://github.com/trinodb/trino/pull/25811/files)
- https://github.com/trinodb/trino/issues/26854

## What your current code does (v477)

In [GcsServiceAccountAuth](https://github.com/trinodb/trino/blob/master/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsServiceAccountAuth.java), if neither jsonKey nor jsonKeyFilePath is supplied and ADC cannot be resolved, the catch block returns null, so we simply never call builder.setCredentials(...). That means the Storage client is built with credentials == null. In practice, the google-cloud-storage library will then send requests without an Authorization header.

## Potential downsides of just “null”

- Silent failure: If you accidentally run against real GCS (not eg a fake server) and no credentials are available, you won’t see an early, explicit error; you’ll just get 401/403 when making calls.
- Repeated ADC attempts: In some library versions, constructing clients or performing certain operations may retry ADC resolution internally, adding overhead or noisy logs (to be verified in the current version).
- Ambiguity: It’s not obvious to a future reader that “null” was intentionally used to mean “unauthenticated mode.”

## Proposal

- Prefer explicit `NoCredentials` for clarity. The GCS library provides `NoCredentials.getInstance()`. This documents intent and prevents future code paths from trying to refresh tokens.
- We could add  `NoCredentials` to [GcsServiceAccountAuth](https://github.com/trinodb/trino/blob/master/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsServiceAccountAuth.java#L65)
- We could add  a new `GcsNoAuth` auth type that is configurable with `UNAUTHENTICATED` (in [GcsFileSystemConfig](https://github.com/trinodb/trino/blob/master/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java#L37)) 

WDYT?

ebyhr:
cc: @chenjian2664

szisiu:
Also it is not possible (at least I cannot figure it out) to use GCS without credentials [Trino 477 Release notes](https://trino.io/docs/current/release/release-477.html#iceberg-connector).

[...]

Issue #26772, opened 2025-09-30
Trino 477 cannot use default auth in GKE environment


jiayeZhu:
I use helm to deploy trino to my GKE cluster. playing with Iceberg table storing on GCS. works fine with version 476.

Today I upgraded to 477 and  it's complaining this:
```
Invalid configuration property with prefix '' Either gcs.use-access-token or gcs.json-key or gcs.json-key-file-path must be set 
```
the pod is running with default service account, having no problem with version 476. only change i made is update the tag from 476 to 477.

jiayeZhu:
helm chart version 1.41.0

thomas-bright:
I'm encountering the same error, but for a specific service account that uses Workload Identity instead of using json credentials. 

Was using Trino 476 with a Iceberg table stored in GCS, accessed with a specific service account with Workload Identity, and that worked fine.

Upgraded to Trino 477 using Helm Chart v1.41.0, and now am also getting this error. 

Even while explicitly setting `gcs.use-access-token=false` in the catalog properties, the same error is given.

brokenjacobs:
Just to add, if you specify gcs.use-access-token=true then all your queries involving GCS fail.

jetlag11235:
also impacted by this issue.  it may be caused by [this PR](https://github.com/trinodb/trino/pull/25811).

the [documentation](https://trino.io/docs/474/object-storage/file-system-gcs.html#authentication) suggests that one of `use-access-token`, `json-key`, or `json-key-file-path` must be used; however, we have been relying on a seemingly undocumented fourth option to use gke workload identity bindings to automatically inject credentials into the pod.

we suspect we could use 477 by adopting one of the documented approaches above, but preferred our existing solution because it seems more secure (no secrets to manage).

@patricklucas : curious if you have any thoughts on this.

patricklucas:
Sorry for the delay, but thanks for the ping @jetlag11235.

Finally got a chance to look at this. I was surprised that my change could cause the issue described here, since it's very close to my own use-case that I tested against—namely, letting the GCP SDK do auth automatically, be that by discovering GOOGLE_APPLICATION_CREDENTIALS in the environment or by hitting the metadata URL as in this case. (I assume that's where the injected credentials come from).

[...]

PR #26941, opened 2025-10-14
Remove deprecated `gcs.use-access-token` from gcs filesystem

chenjian2664:
## Description

Now we recommend to use the `gcs.auth-type` instead

## Additional context and related issues

- https://github.com/trinodb/trino/pull/26681

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
## Hive, Delta Lake, Iceberg, Hudi
* {{breaking}} Remove the deprecated `gcs.use-access-token` config property. ({issue}`26941`)
```

Issue #26854, opened 2025-10-06
Emulating Google Cloud Storage as object storage for Iceberg


szisiu:
Our development and test environments use MinIO as an S3-compatible object store for Iceberg. Since our production environment uses GCS, this requires us to install heavy Amazon S3 libraries for the Hive Metastore (HMS) in our development setup. This adds unnecessary dependencies, increases build size, and creates a divergence between our development and production environments.

We could replace MinIO with a lightweight, GCS-compatible emulator to better align our development environment with production. The recommended solution is [fsouza/fake-gcs-server](https://github.com/fsouza/fake-gcs-server), which can be easily integrated into our existing containerized setup. 

The new Trino native GCS file system (`fs.native-gcs.enabled=true`) does not provide a mechanism to disable authentication, making it impossible to use with local, unauthenticated GCS emulators like `fake-gcs-server`.

Even when `gcs.endpoint` is configured to point to a local emulator, the underlying Google Cloud client library still attempts to perform a real OAuth2 token exchange with the public Google endpoint (`https://oauth2.googleapis.com/token`). This fails because a dummy credential file must be provided (to avoid an "ADC not found" error), but this dummy credential is invalid for the real authentication service.

There should be a way to use the recommended native client for local testing without it attempting to contact external authentication services.

### Environment

  * **Trino Version:** latest
  * **Connector:** Iceberg (`iceberg.catalog.type=hive_metastore`)
  * **Storage Emulator:** `fsouza/fake-gcs-server`
  * **Metastore:** Hive Metastore 4.1
  * **Setup:** Docker Compose

### Configuration

1.  **`docker-compose.yml` template:**

    ```yaml
    services:
      gcs-emulator:
        image: fsouza/fake-gcs-server
        container_name: gcs-emulator
        ports:
          - "4443:4443"
        command: -scheme http -public-host gcs-emulator:4443

      mysql-db:
        image: mysql:8.0
        container_name: mysql-db
        environment:
          - MYSQL_ROOT_PASSWORD=secret
          - MYSQL_DATABASE=metastore
          - MYSQL_USER=hive
          - MYSQL_PASSWORD=hive

      hive-metastore:
        image: apache/hive:4.0.0
        container_name: hive-metastore
        depends_on: [mysql-db]
        ports: ["9083:9083"]
        volumes:
          - ./hive/conf/metastore-site.xml:/opt/hive/conf/metastore-site.xml

      trino:
        image: trinodb/trino:latest
        container_name: trino
[...]

PR #27106, opened 2025-10-27
Add `APPLICATION_DEFAULT` authentication type for GCS

chenjian2664:
Add a dedicated `APPLICATION_DEFAULT ` type to explicitly support unauthenticated access.

In this PR
	• Upgraded the `GcsFileSystemConfig` validation method to make it consistent with the current implementation when the auth type is `SERVICE_ACCOUNT` (default auth type).  ⚠️  This change also removes the fallback behavior when credentials are not provided. Users must now supply credentials explicitly.
	• Added support for the new `APPLICATION_DEFAULT ` auth type, allowing users to explicitly access Gcs without providing any credentials.

## Description

Fixes https://github.com/trinodb/trino/issues/26984

## Release notes

```markdown
## Iceberg, Delta Lake, Hudi
* Add `APPLICATION_DEFAULT` authentication type for GCS. ({issue}`26984`)
* Remove unauthentication access support in `SERVICE_ACCOUNT`  ({issue}`26984`) TBD
```


ebyhr:
@chenjian2664 Is this a release blocker of 478?

docs/src/main/sphinx/object-storage/file-system-gcs.md
```diff
@@ -75,10 +75,12 @@ Cloud Storage:
   - Authentication type to use for Google Cloud Storage access. Default to `SERVICE_ACCOUNT`.
   Supported values are:
     * `SERVICE_ACCOUNT`: loads credentials from the environment. Either `gcs.json-key` or
-      `gcs.json-key-file-path` can be set in addition to override the default
-      credentials provider.
+      `gcs.json-key-file-path` can be set.
     * `ACCESS_TOKEN`: usage of client-provided OAuth 2.0 token to access Google
       Cloud Storage.
+    * `NO_CREDENTIALS_AUTH`: allows accessing Google Cloud Storage without providing
```
ebyhr:
```suggestion
    * `NO_CREDENTIALS`: allows accessing Google Cloud Storage without providing
```

ebyhr:
CI failure is related to this change: 
```
Error:  io.trino.filesystem.gcs.TestGcsStorageFactory.testDefaultCredentials -- Time elapsed: 0.609 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 
[if credentials are not explicitly configured, should have same behavior as the GCS client] 
expected: null
 but was: NoCredentials{requestMetadata=null, temporaryAccess=null}
	at io.trino.filesystem.gcs.TestGcsStorageFactory.testDefaultCredentials(TestGcsStorageFactory.java:44)
```

ebyhr:
> Add NO_CREDENTIALS_AUTH authentication type for GCS

Please replace NO_CREDENTIALS_AUTH with NO_CREDENTIALS in the commit title.

docs/src/main/sphinx/object-storage/file-system-gcs.md
```diff
@@ -75,10 +75,12 @@ Cloud Storage:
...
   Supported values are:
     * `SERVICE_ACCOUNT`: loads credentials from the environment. Either `gcs.json-key` or
-      `gcs.json-key-file-path` can be set in addition to override the default
-      credentials provider.
+      `gcs.json-key-file-path` can be set.
     * `ACCESS_TOKEN`: usage of client-provided OAuth 2.0 token to access Google
       Cloud Storage.
+    * `NO_CREDENTIALS`: allows accessing Google Cloud Storage without providing
+      explicit credentials. Attempts to acquire application default credentials from the
+      environment, and falls back to no-credentials access if none are available.
```
Praveen2112: Shouldn't this be behaviour by default ? Why do we need to implement them as a dedicated enum or configure them out ?

chenjian2664: Previously, the `NO_CREDENTIALS` behavior was implicitly covered by the `SERVICE_ACCOUNT` auth type, but in practice it was never reachable because the `SERVICE_ACCOUNT` validator requires either a `json-key` or `json-key-file-path` to be configured. By introducing a dedicated `NO_CREDENTIALS` type, we can now support this scenario explicitly and ensure the validation logic stays consistent, context see https://github.com/trinodb/trino/issues/26984

Praveen2112: I meant if we didn't set the auth type to any value - it should exhibit this behaviour right ?

chenjian2664: Previously the default is `SERVICE_ACCOUNT`, so I don't want to change it in this pr - just want to introduce a new type.
WDYT @wendigo

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/NoCredentialsAuth.java
```diff
@@ -0,0 +1,36 @@
...
+{
+    @Override
+    public void setAuth(StorageOptions.Builder builder, ConnectorIdentity identity)
+    {
+        try {
+            builder.setCredentials(GoogleCredentials.getApplicationDefault());
+        }
+        catch (IOException e) {
+            builder.setCredentials(NoCredentials.getInstance());
+        }
```
szisiu, line 40: Can we just leave `builder.setCredentials(NoCredentials.getInstance());`?
The `NoCredentialsAuth` class single responsibility is to provide the UNAUTHENTICATED access only, no?

IMO, we do not need `builder.setCredentials(GoogleCredentials.getApplicationDefault())` here because it breaks the purpose and it is somehow misleading. I'd leave it in [GcsServiceAccountAuth](https://github.com/trinodb/trino/pull/27106/files#diff-0a10de45766835578f6e53acf7c91c94cf85858a6ef42e2f36badbc49dd50616) or extract it to a dedicated eg `GcsAdcAuth` class (potential breaking change).

See my original (but I think outdated) proposal/idea: https://github.com/trinodb/trino/issues/26854#issuecomment-3374313658

docs/src/main/sphinx/object-storage/file-system-gcs.md
```diff
@@ -75,10 +75,12 @@ Cloud Storage:
   - Authentication type to use for Google Cloud Storage access. Default to `SERVICE_ACCOUNT`.
   Supported values are:
     * `SERVICE_ACCOUNT`: loads credentials from the environment. Either `gcs.json-key` or
-      `gcs.json-key-file-path` can be set in addition to override the default
-      credentials provider.
+      `gcs.json-key-file-path` can be set.
     * `ACCESS_TOKEN`: usage of client-provided OAuth 2.0 token to access Google
       Cloud Storage.
+    * `NO_CREDENTIALS`: allows accessing Google Cloud Storage without providing
```
martint: This is misleading. "No credentials" would seem to imply "unauthenticated", but that is not the case based on the description.

It's also somewhat confusing that both SERVICE_ACCOUNT and NO_CREDENTIALS "acquire application default credentials from the environment". So, what's the difference?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/NoCredentialsAuth.java
chenjian2664: Make sense, I complicated the problem, switched

docs/src/main/sphinx/object-storage/file-system-gcs.md
chenjian2664: Agreed, the name is confusing for users. I've removed the fallback logic

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsServiceAccountAuth.java
```diff
@@ -56,18 +58,15 @@ else if (jsonKeyFilePath != null) {
...
     public void setAuth(StorageOptions.Builder builder, ConnectorIdentity identity)
     {
-        GoogleCredentials credentials = jsonGoogleCredential.orElseGet(() -> {
+        Credentials credentials = jsonGoogleCredential.orElseGet(() -> {
             try {
                 return GoogleCredentials.getApplicationDefault();
             }
             catch (IOException e) {
-                // This is consistent with the GCP SDK when no credentials are available in the environment
-                return null;
```
szisiu, line 65: I'd leave it as-is now.

szisiu: @chenjian2664 wdyt?

patricklucas:
Just for clarity, can you confirm that this change should fix #26772? That is, it should again be allowed to not specify any GCP auth config options to let the GCS client library follow its default behavior of discovering credentials in the environment, e.g. from the GOOGLE_APPLICATION_CREDENTIALS env var or from the GCE metadata URL (which isn't the same thing as passing `NoCredentials.getInstance()`)?

szisiu (approved):

szisiu:
@patricklucas I think it should be good now.

Without any settings `gcs.auth-type=SERVICE_ACCOUNT` according to https://github.com/trinodb/trino/pull/27106/files#diff-0cf9ded0011b1e78e8a5113b9fcde8e5efe20520b99b28378e482c49466a429aL151 (GcsFileSystemConfig.java)

Then `GcsFileSystemConfig#isAuthMethodValid` will accept the config (ie return true).

Lastly https://github.com/trinodb/trino/blob/master/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsServiceAccountAuth.java is untouched by this PR so hte logic should execute as you described.
Also this test should verify it: https://github.com/trinodb/trino/pull/27106/files#diff-0dfdbc389f7bd5fedf4e09a388deaabf6ba994653c1995b675503d14c7ad3fbaR29

image

Please, someone, double check it so we could be 100% sure.

patricklucas:
@szisiu thanks for confirming. I'll comment on #26772 and see if one of the affected users can try it out.

However, the test you linked did not catch this before because it does not exercise the config validation code, which is where the issue arose (it creates the config object and uses it without triggering validation).

It might be worth updating that test to call whatever the opposite of `assertFailsValidation` on the `config` object after creation.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -37,7 +37,8 @@ public class GcsFileSystemConfig
     public enum AuthType
     {
         ACCESS_TOKEN,
-        SERVICE_ACCOUNT;
+        SERVICE_ACCOUNT,
+        UNAUTHENTICATED;
```
wendigo: I don't like the name here. It's actually authenticated by the credentials are provided by the environment.

wendigo: Maybe ASSUMED?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -297,7 +298,12 @@ public boolean isRetryDelayValid()
     @AssertTrue(message = "Either gcs.auth-type or gcs.json-key or gcs.json-key-file-path must be set")
     public boolean isAuthMethodValid()
     {
-        if (getAuthType() == AuthType.ACCESS_TOKEN) {
+        // allow not specify any of configuration for SERVICE_ACCOUNT auth
+        if (getAuthType() == AuthType.SERVICE_ACCOUNT && jsonKey == null && jsonKeyFilePath == null) {
```
wendigo: We could bound separate config based on a type to have these checked for other then ASSUMED methods

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemModule.java
```diff
@@ -36,6 +36,7 @@ protected void setup(Binder binder)
                 type -> switch (type) {
                     case ACCESS_TOKEN -> _ -> binder.bind(GcsAuth.class).to(GcsAccessTokenAuth.class).in(Scopes.SINGLETON);
                     case SERVICE_ACCOUNT -> _ -> binder.bind(GcsAuth.class).to(GcsServiceAccountAuth.class).in(Scopes.SINGLETON);
+                    case UNAUTHENTICATED -> _ -> binder.bind(GcsAuth.class).to(UnauthenticatedAuth.class).in(Scopes.SINGLETON);
```
wendigo: AssumedAuth

wendigo (approved):

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
Praveen2112: How about `DEFAULT` - Like we have in `AzureFileSystemConfig` ?

Praveen2112: I don't think `DEFAULT` would help us as it should be set to `getApplicationDefault` instead of `NoCredentials`

Praveen2112: I guess the same argument goes for `ASSUMED` as well - We don't load the credentials provided by the environment - It is more on passing an empty credentials

docs/src/main/sphinx/object-storage/file-system-gcs.md
```diff
@@ -79,6 +79,8 @@ Cloud Storage:
       credentials provider.
     * `ACCESS_TOKEN`: usage of client-provided OAuth 2.0 token to access Google
       Cloud Storage.
+    * `UNAUTHENTICATED`: allows access to Google Cloud Storage without explicitly providing
```
Praveen2112: This is mostly used for local emulation right ?

Praveen2112:
Sorry for joining the party late, instead of introducing a new `UNAUTHENTICATED` enum, can we introduce a `DEFAULT` which would try to load the default credentials - if it doesn't exist it could load `NoCredentials`  something like 
```
            try {
                return GoogleCredentials.getApplicationDefault();
            }
            catch (IOException e) {
                // This is consistent with the GCP SDK when no credentials are available in the environment
                return NoCredentials.getInstance();
            }
```

And we could remove the logic on loading the `getApplicationDefault` from `SERVICE_ACCOUNT`. It could be a breaking change but it would make the code a bit easier to understand and interpret. WDYT ?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
patricklucas: IIUC no, UNAUTHENTICATED delegates to `NoCredentials` which is actually no credentials. Authenticating from the environment using a service account key is handled by SERVICE_ACCOUNT without setting `jsonKey`/`jsonKeyFilePath`.

szisiu:
If we want a clear separation of concerns and explicit auth options we should have: 

- ACCESS_TOKEN - Client-provided OAuth 2.0 token
- SERVICE_ACCOUNT - Explicit service account with JSON key (file or inline) - Trino's default
- APPLICATION_DEFAULT - Google Application Default Credentials (ADC) from environment
- UNAUTHENTICATED - No credentials (for emulators, public buckets, etc.)

with tests and clear documentation for each auth type.

But this is a breaking change due to the fact that current SERVICE_ACCOUNT implementation needs to be split into new SERVICE_ACCOUNT logic (w/o ADC)  and APPLICATION_DEFAULT.
If we needs to preserve bacward compatibility we could introduce  the DEFAULT option which is the current SERVICE_ACCOUNT. 

Otherwise, changes in this PR looks good from my perspective.

szisiu:
We're trying to overload existing auth types instead of having explicit, single-responsibility authentication methods.

I believe we need a clear decision from the maintainers on the architectural direction before we can finalize this.

The primary issue is whether we're willing to make a breaking change for better architecture.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
chenjian2664: Do you mean move jsonKey and jsonFilePath into a new `GcsServiceAccountAuthConfig` ?

docs/src/main/sphinx/object-storage/file-system-gcs.md
chenjian2664: I think so, also maybe could be used to access public bucket

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
chenjian2664: updated

chenjian2664:
> And we could remove the logic on loading the getApplicationDefault from SERVICE_ACCOUNT. It could be a breaking change but it would make the code a bit easier to understand and interpret. WDYT ?

this is the initial implementation of the pr, it is an indeed breaking change, if we add it in this pr, then for the `SERVICE_ACCOUNT` will need the config the either jsonkey or json file path, it is back to situation same as before.

I think it is good to spilt it to have better semantic, and I prefer for adding it at the next release and let this release as a "fix" for the case of the `SERVICE_ACCOUNT` but without any credentials.

Praveen2112:
> If we want a clear separation of concerns and explicit auth options we should have:

> ACCESS_TOKEN - Client-provided OAuth 2.0 token
SERVICE_ACCOUNT - Explicit service account with JSON key (file or inline) - Trino's default
APPLICATION_DEFAULT - Google Application Default Credentials (ADC) from environment
UNAUTHENTICATED - No credentials (for emulators, public buckets, etc.)

@szisiu Instead of having a dedicated option for `UNAUTHENTICATED` which seems to be an overkill - Can we handle it as `DEFAULT` which uses `Google Application Default Credentials (ADC) from environment` if available or else loads `NoCredentials` - Would emulator or public buckets access be restricted when we use `Google Application Default Credentials` ?


> I think it is good to spilt it to have better semantic, and I prefer for adding it at the next release and let this release as a "fix" for the case of the SERVICE_ACCOUNT but without any credentials.

But if it is a breaking change - would this release or next release make any significant difference ? Introducing `ASSUMED`/`DEFAULT`  authentication mechanism as a quick fix for this release would be broken - as it we set `NoCredentials` it is not exactly assumed but passing an empty set of credentials - Instead we could try 
```
            try {
                return GoogleCredentials.getApplicationDefault();
            }
            catch (IOException e) {
                // This is consistent with the GCP SDK when no credentials are available in the environment
                return NoCredentials.getInstance();
            }
```
which wouldn't break anything. 

>  if we add it in this pr, then for the SERVICE_ACCOUNT will need the config the either jsonkey or json file path, it is back to situation same as before.

The default authentication mode would be `DEFAULT` - which doesn't require any `jsonkey` or `json file path` and relies on GoogleADC and if the user need to provide service account credentials then they could update the config property to `SERVICE_ACCOUNT`

If we are planning to introduce a breaking change - it would be nice to have in a single release rather than tacking them across releases - I guess the changes here are simple one

szisiu:
> If we are planning to introduce a breaking change - it would be nice to have in a single release

I agree.

szisiu:
My subjective opinion for this PR now is to do the following:

I would advise against using `ASSUMED` for this context. 
It’s unclear what is being “assumed”: a role? default credentials? anonymous access? And increase documentation burden explaining what it does. While Google does not use the verb “assume” prominently, multiple features are semantically similar and users often analogize them to AWS’s AssumeRole, eg Service Account Impersonation or Workload Identity Federation.

Prefer not to silently fallback inside GcsServiceAccountAuth. Instead, provide a clearly named composite mode (`AUTO`). Current approach is compact and convienient but complicates tests, hidden fallback harms predictability and security.

To sum up, I still believe we should have the following auth options:
- `ACCESS_TOKEN` - Client-provided OAuth 2.0 token
- `SERVICE_ACCOUNT` - Explicit service account with JSON key (file or inline). If key is not provided then throw config error eagerly.
- `APPLICATION_DEFAULT` - Google Application Default Credentials resolution chain. Maps directly to well-known Google concept. Clarify in docs that `APPLICATION_DEFAULT` encompasses env var `GOOGLE_APPLICATION_CREDENTIALS`, metadata server, gcloud user credentials, etc.
- `NONE` or `ANONYMOUS` - No credentials. Warn that `ANONYMOUS` only works for publicly readable buckets/objects or emulators. Requests omit Authorization header. Industry‑standard wording; immediately clear that no credential material is presented. Cover setting storage credentials to null or NoCredentials - both lead to anonymous requests.
- `AUTO` - Trino's default. Internally tries (in documented order) `SERVICE_ACCOUNT` then `APPLICATION_DEFAULT` then `NONE`, but make that an explicit opt-in.

Each mode is mutually exclusive and does exactly one resolution path.
No implicit fallback between them.

It covers all cases and issues we discussed here so far.

yannick-bright:
I've built the code in this branch and can confirm it's now running in our setup (using Workload Identity) without any problem.
Thanks for the effort :-)

chenjian2664:
@szisu Sorry for the delay here. I think `APPLICATION_DEFAULT` is good and simple.
Regarding the naming of `ASSUMED`, would it be okay to revert to the original `NO_CREDENTIALS`? Since we are using the Google method which refers to it as `NoCredentials`. cc @wendigo @praveen
Personally, I don't think `AUTO` is a good choice because it hides the authentication details, similar to what `SERVICE_ACCOUNT` does, and we should avoid such behavior unless it’s the consensus across the industry.

chenjian2664:
@Praveen2112 
> which wouldn't break anything.

The breaking change here is whether we want to move the application default credentials behavior out of `SERVICE_ACCOUNT` - and now we are going to move it, so it is.
I think adding a fallback is acceptable, rather than introducing an assumed credentials mechanism. I'm considering adding the fallback. However, I don't agree with renaming the `ASSUMED` it to `DEFAULT`,  that would impact more users because we previously treated `SERVICE_ACCOUNT` as the implicit default. Renaming and setting it default introduces unnecessary disruption.

My proposal is:
* Keep the name(`ASSUMED`) to `APPLICATION_DEFAULT` and with the fallback behavior and document it clearly.
* Preserve `SERVICE_ACCOUNT` as the default type, but remove the fallback.

With this design:
* If users want to run without explicit credentials, they should use `APPLICATION_DEFAULT`.
* If they use `SERVICE_ACCOUNT`, they must provide the `json key` or `json file` path. We intentionally do not auto-read environment ADCs under `SERVICE_ACCOUNT`, so it's impossible to implicitly use them, and that's by design
	
The consequence with this is we can't re-introduce the previous `SERVICE_ACCOUNT` mechanism, that is with fallbacks `json key` -> `application default` -> `null`, unless we introduce a new type maybe like `AUTO` @szisiu mention 

@wendigo @Praveen2112 @szisiu What do you think of it?

szisiu:
@chenjian2664 it is fine with me, so we have:

- `ACCESS_TOKEN`  - w/o any fallback
- `SERVICE_ACCOUNT` - w/o any fallback
- `APPLICATION_DEFAULT` - with `null` fallback
- `NO_CREDENTIALS` - w/o any fallback
- `AUTO` ? - maybe it could be useful to achieve backward compatibility; it does not hide  the authentication details as it is a composite type

chenjian2664:
@szisiu Actually, I'm thinking it should be:
* `ACCESS_TOKEN` - without any fallback
* `SERVICE_ACCOUNT` - without any fallback, still keep as the default
* `APPLICATION_DEFAULT` - with `NoCredentials` as a fallback
* `AUTO`, maybe, but I'll hold my opinion on this one and not reflect it in this PR.

szisiu:
@chenjian2664 ok

Praveen2112:
> My proposal is:
Keep the name(ASSUMED) to APPLICATION_DEFAULT and with the fallback behavior and document it clearly.
Preserve SERVICE_ACCOUNT as the default type, but remove the fallback.

This is fine for me. 

> ACCESS_TOKEN - w/o any fallback
SERVICE_ACCOUNT - w/o any fallback
APPLICATION_DEFAULT - with null fallback
NO_CREDENTIALS - w/o any fallback
AUTO ? - maybe it could be useful to achieve backward compatibility; it does not hide the authentication details as it is a composite type

I would rather stick on to three options - `ACCESS_TOKEN`, `SERVICE_ACCOUNT` and `APPLICATION_DEFAULT` - which tries to load credentials from the system, if it fails then we could fallback to `NO_CREDENTIALS`. 

@szisiu WDYT ?

`APPLICATION_DEFAULT` deviates a bit with the approach we have in other system like `AzureAuth` where we name it as `DEFAULT` - I would like to have others opinion for the same.

szisiu:
@Praveen2112 sgtm

martint:
> DEFAULT - I would like to have others opinion for the same.

A name other than `DEFAULT` is preferrable. Otherwise the name of the option can imply that that's the option that's configured by default, which may not be the case. It be also be very confusing if the default were, for instance, SERVICE_ACCOUNT but there's an option named DEFAULT that's not the default.

chenjian2664:
PTAL @szisiu @Praveen2112 @wendigo

szisiu (approved):

szisiu:
lgtm

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
```diff
@@ -297,6 +297,11 @@ public boolean isRetryDelayValid()
     @AssertTrue(message = "Either gcs.auth-type or gcs.json-key or gcs.json-key-file-path must be set")
     public boolean isAuthMethodValid()
     {
+        // allow not specify any of configuration for SERVICE_ACCOUNT auth
+        if (getAuthType() == AuthType.SERVICE_ACCOUNT && jsonKey == null && jsonKeyFilePath == null) {
```
Praveen2112: Do we need this change for `SERVICE_ACCOUNT` - IIUC we will prohibiting this configuration right ?

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
```diff
@@ -198,6 +198,22 @@ public void testValidation()
                 "authTypeAndGcsAccessTokenConfigured",
                 "Cannot set both gcs.use-access-token and gcs.auth-type",
                 AssertFalse.class);
+
+        assertFailsValidation(
+                new GcsFileSystemConfig()
+                        .setUseGcsAccessToken(true)
+                        .setAuthType(AuthType.APPLICATION_DEFAULT),
+                "authTypeAndGcsAccessTokenConfigured",
+                "Cannot set both gcs.use-access-token and gcs.auth-type",
```
Praveen2112, line 179: The error message can be improved right ? And we could implement a dedicated config like we do for `SERVICE_ACCOUNT` ?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystemConfig.java
Praveen2112, line 1: Can we have it as a preparatory commit i.e not as a last commit

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
chenjian2664: Yes, but we are planning to remove the `gcs.use-access-token` in https://github.com/trinodb/trino/pull/26941

chenjian2664:
comments AC @Praveen2112

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsServiceAccountAuthConfig.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.common.collect.ImmutableSet;
+import jakarta.validation.constraints.AssertTrue;
+import org.junit.jupiter.api.Test;
+
+import static io.airlift.configuration.testing.ConfigAssertions.assertFullMapping;
+import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
+import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+import static io.airlift.testing.ValidationAssertions.assertFailsValidation;
+
+public class TestGcsServiceAccountAuthConfig
```
Praveen2112: nit: final class right ? As per test guidelines

Praveen2112 (approved):
LGTM. Minor change - LMK once it is applied - I'll merge this one

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsServiceAccountAuthConfig.java
chenjian2664: AC

PR #27118, opened 2025-10-27
Avoid unnecessary building of TaskStats in SqlTaskManager

raunaqmorarka:
## Description
The information required in removeOldTasks and failAbandonedTasks can be obtained cheaply without contructing TaskStats repeatedly

JFR profiles of coordinator often show significant CPU time spent in `io.trino.operator.TaskContext#getTaskStats`
Screenshot 2025-10-27 at 11 12 06 PM

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Avoid constructing TaskStats repeatedly during task cleanup by querying end time and heartbeat directly and caching the end timestamp in the state machine

Enhancements:
- Simplify removeOldTasks and failAbandonedTasks in SqlTaskManager to use SqlTask.getTaskState, getTaskEndTime, and lastHeartbeat instead of building TaskInfo/TaskStats
- Track execution end time in TaskStateMachine with an AtomicReference and expose it via a new getEndTime method
- Add SqlTask methods getTaskEndTime and lastHeartbeat for direct access to task timestamps

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/SqlTask.java
+++ b/core/trino-main/src/main/java/io/trino/execution/SqlTask.java
@@ -252,6 +252,12 @@ public class SqlTask
         return taskStateMachine.getCreatedTime();
     }
 
+    @Nullable
+    public Instant getTaskEndTime()
+    {
+        return taskStateMachine.getEndTime();
+    }
+
     public TaskId getTaskId()
     {
         return taskStateMachine.getTaskId();
@@ -265,6 +271,11 @@ public class SqlTask
     public void recordHeartbeat()
     {
         lastHeartbeat.set(Instant.now());
+    }
+
+    public Instant lastHeartbeat()
+    {
+        return lastHeartbeat.get();
     }
 
     public TaskInfo getTaskInfo()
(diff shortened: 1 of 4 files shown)


wendigo (approved):
Beautiful change!

core/trino-main/src/main/java/io/trino/execution/SqlTask.java
```diff
@@ -252,6 +252,11 @@ public Instant getTaskCreatedTime()
         return taskStateMachine.getCreatedTime();
     }
 
+    public Instant getTaskEndTime()
```
pettyjamesm, line 256: Nit: `@Nullable`

core/trino-main/src/main/java/io/trino/execution/TaskStateMachine.java
```diff
@@ -61,20 +63,36 @@ public class TaskStateMachine
...
 
     public Instant getCreatedTime()
     {
         return createdTime;
     }
 
+    @Nullable
+    public Instant getEndTime()
+    {
+        if (getState().isDone()) {
```
pettyjamesm, line 89: This is definitely cheaper, but do we have another alternative to duplicating the `taskEndTime` field between here and `TaskContext` to avoid two _slightly_ different timestamps in two places for the "same thing"?

raunaqmorarka: I've removed the tracking of task end time in TaskContext and used this field there as well now

PR #27134, opened 2025-10-28
Use virtual threads for loading plugins and creating connectors

wendigo:
This work is mostly I/O bound (reading JARs) so it's a good fit for virtual threads

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Enhancements:
- Replace BoundedExecutor backed by newCachedThreadPool with ThreadPerTaskExecutor using virtual threads for the startup executor


wendigo:
@raunaqmorarka it's roughly the same but there are less threads remaining after startup (startup executor isn't closed as you can still create new catalogs dynamically which is a waste because you keep a thread pool running that isn't used most of the time)

chenjian2664 (approved):

findepi:
> This work is mostly I/O bound (reading JARs)

i would think there is considerable time spent on decompressing the JARs and then loading the classes.

> waste because you keep a thread pool running that isn't used most of the time)

most thread pools expire their threads after some time (like 1 minute). was this one different?

PR #27137, opened 2025-10-28
Add dynamic schema support based on protobuf descriptor

gertjanal:
## Description
Since Trino 477, the Hive connector has support for reading sequence files containing protobuf bytes; https://github.com/trinodb/trino/pull/26353. With that PR, records could be read using the protobuf deserializer.

The code was created for Hive tables that were created using Hive Twitter Elephantbird: https://github.com/twitter/elephant-bird/wiki/How-to-use-Elephant-Bird-with-Hive. In Hive, the tables use the `serialization.class` protobuf file to determine the columns for the table as well, which is named "Dynamic Schemas". An example for such a table in Hive is:

```
create table users
  row format serde "com.twitter.elephantbird.hive.serde.ProtobufDeserializer"
  with serdeproperties (
    "serialization.class"="com.example.proto.gen.Storage$User")
  stored as
    inputformat "com.twitter.elephantbird.mapred.input.DeprecatedRawMultiInputFormat";
```

When using `show create table X;` in Hive, all resolved columns are shown. When the protobuf file is updated with new fields, the columns are automatically added (in Hive). However, in Trino the Hive Metastore is asked for the table columns and I found out that the columns returned by the Hive Metastore (via Trino) were the columns that were determined upon creation of the original table. Later added fields to the protobuf were shown when querying Hive directly, but Trino received the original version via Hive Metastore. Hence this PR.

This PR adds the dynamic schema support to Trino. When the storage format SEQUENCEFILE_PROTOBUF is detected, the columns are derived from the protobuf fields, as is done in the Hive implementation.

## Release notes

## Summary by Sourcery

Enable dynamic schema support for Hive SEQUENCEFILE_PROTOBUF tables by deriving column definitions from Protobuf descriptors instead of relying on static metastore schemas.

New Features:
- Derive table and partition column schemas at runtime for SEQUENCEFILE_PROTOBUF storage by reading Protobuf descriptor fields
- Introduce DynamicSchemaHiveMetastore wrapper that fetches and caches dynamic schemas from Protobuf descriptors
- Add DynamicSchemaLoader utility to map Protobuf FieldDescriptors to Trino Column definitions

Enhancements:
- Refactor ProtobufDeserializerFactory to expose descriptor retrieval via getDescriptor and introduce ProtobufDeserializerFactoryLoader for dependency injection

Build:
- Add protobuf-java dependency and configure protoc-jar-maven-plugin to generate test Protobuf descriptors

Tests:
- Add TestDynamicSchemaHiveMetastore to validate dynamic schema extraction and include Protobuf proto test fixtures

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadataFactory.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadataFactory.java
@@ -19,11 +19,14 @@ import io.airlift.concurrent.BoundedExecutor;
 import io.airlift.json.JsonCodec;
 import io.airlift.units.Duration;
 import io.trino.filesystem.TrinoFileSystemFactory;
+import io.trino.hive.formats.line.protobuf.ProtobufDeserializerFactory;
 import io.trino.metastore.HiveMetastore;
 import io.trino.metastore.HiveMetastoreFactory;
 import io.trino.plugin.hive.fs.DirectoryLister;
 import io.trino.plugin.hive.fs.TransactionScopeCachingDirectoryListerFactory;
 import io.trino.plugin.hive.metastore.SemiTransactionalHiveMetastore;
+import io.trino.plugin.hive.metastore.dynamic.DynamicSchemaHiveMetastore;
+import io.trino.plugin.hive.metastore.dynamic.ProtobufDeserializerFactoryLoader;
 import io.trino.plugin.hive.security.AccessControlMetadataFactory;
 import io.trino.plugin.hive.security.UsingSystemSecurity;
 import io.trino.plugin.hive.statistics.MetastoreHiveStatisticsProvider;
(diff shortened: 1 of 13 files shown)


plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadataFactory.java
```diff
@@ -230,7 +238,10 @@ public HiveMetadataFactory(
     @Override
     public TransactionalMetadata create(ConnectorIdentity identity, boolean autoCommit)
     {
-        HiveMetastore hiveMetastore = createPerTransactionCache(metastoreFactory.createMetastore(Optional.of(identity)), perTransactionCacheMaximumSize);
+        HiveMetastore hiveMetastore = new DynamicSchemaHiveMetastore(
+                createPerTransactionCache(metastoreFactory.createMetastore(Optional.of(identity)), perTransactionCacheMaximumSize),
+                protobufDeserializerFactory,
+                Duration.valueOf("1h")); // TODO make configurable
```
gertjanal, line 245: I will make a proper config property of this TODO when reviewers agree with the rest of the code

dain:
I think this feature is too complex based on the number of user that are interested in using sequence files containing protocol buffers. We have spend the last few years attempting to simplify the Hive connector and especiall they metadata. IMO, I suggest you convert to a modern format.  I understand there a significant effort to rewrite legacy data, but you will have to do it sometime, so why not now?

gertjanal:
:sob:

PR #27140, opened 2025-10-29
Allow time travel reads to Hudi Tables

jordepic:
This commit handles simple "FOR VERSION AS OF X" syntax in
the Hudi connector order to ensure that we can read previous
table state.

Previously, all reads to Hudi tables would occur at the latest
commit timestamp. While a user could technically filter down
the data using a predicate on the _hoodie_commit_time column,
there is no functionality that pushes this down into the Hudi
API to minimize file reads.

Timestamps are provided as strings..

## Release notes

Added time travel reads in Hudi connector.

### DIFF
--- a/docs/src/main/sphinx/connector/hudi.md
+++ b/docs/src/main/sphinx/connector/hudi.md
@@ -170,6 +170,21 @@ GROUP BY dt;
 (1 rows)
 ```
 
+### Time travel queries
+
+The connector offers the ability to query historical data. This allows viewing
+the state of the table at a previous snapshot.
+
+The historical data of the table can be retrieved by specifying the last valid
+commit timestamp:
+```sql
+SELECT * FROM  hudi.default.table_name FOR VERSION AS OF '20251027183851494';
+```
+Note: Hudi stores its table versions as a yyyyMMddHHmmssSSS representation of
+the commit time in the table's time zone, though any string can be used. Hudi
+will perform a string comparison between the passed in string and the commit
+time.
+
 ### Schema and table management
 
 Hudi supports [two types of tables](https://hudi.apache.org/docs/table_types)
--- a/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
+++ b/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
@@ -15,6 +15,7 @@ package io.trino.plugin.hudi;
 
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
+import io.airlift.slice.Slice;
 import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystemFactory;
 import io.trino.metastore.Column;
@@ -33,6 +34,7 @@ import io.trino.spi.connector.ConnectorTableMetadata;
 import io.trino.spi.connector.ConnectorTableVersion;
 import io.trino.spi.connector.Constraint;
 import io.trino.spi.connector.ConstraintApplicationResult;
+import io.trino.spi.connector.PointerType;
 import io.trino.spi.connector.RelationColumnsMetadata;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.SchemaTablePrefix;
@@ -104,10 +106,6 @@ public class HudiMetadata
     @Override
     public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName tableName, Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
     {
-        if (startVersion.isPresent() || endVersion.isPresent()) {
-            throw new TrinoException(NOT_SUPPORTED, "This connector does not support versioned tables");
-        }
-
         if (isHiveSystemSchema(tableName.getSchemaName())) {
             return null;
         }
@@ -115,6 +113,7 @@ public class HudiMetadata
         if (table.isEmpty()) {
             return null;
         }
+        Optional<String> readVersion = getReadVersion(startVersion, endVersion);
         if (!isHudiTable(table.get())) {
             throw new TrinoException(UNSUPPORTED_TABLE_TYPE, format("Not a Hudi table: %s", tableName));
         }
@@ -130,7 +129,30 @@ public class HudiMetadata
                 COPY_ON_WRITE,
                 getPartitionKeyColumnHandles(table.get(), typeManager),
                 TupleDomain.all(),
-                TupleDomain.all());
+                TupleDomain.all(),
+                readVersion);
+    }
+
+    private static Optional<String> getReadVersion(Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
+    {
+        if (startVersion.isPresent()) {
+            throw new TrinoException(NOT_SUPPORTED, "Read table with start version is not supported");
+        }
+
+        if (endVersion.isEmpty()) {
+            return Optional.empty();
+        }
+
+        ConnectorTableVersion version = endVersion.get();
+        if (version.getPointerType() == PointerType.TEMPORAL) {
+            throw new TrinoException(NOT_SUPPORTED, "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead");
+        }
+
+        if (version.getVersion() instanceof Slice slice) {
+            return Optional.of(slice.toStringUtf8());
+        }
+
+        throw new TrinoException(NOT_SUPPORTED, "Provided read version must be a string");
     }
 
     @Override
--- a/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
+++ b/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
@@ -24,6 +24,7 @@ import io.trino.spi.predicate.TupleDomain;
 import org.apache.hudi.common.model.HoodieTableType;
 
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 
 import static io.trino.spi.connector.SchemaTableName.schemaTableName;
@@ -41,6 +42,7 @@ public class HudiTableHandle
     private final Set<HiveColumnHandle> constraintColumns;
     private final TupleDomain<HiveColumnHandle> partitionPredicates;
     private final TupleDomain<HiveColumnHandle> regularPredicates;
+    private final Optional<String> readVersion;
 
     @JsonCreator
     public HudiTableHandle(
@@ -50,9 +52,10 @@ public class HudiTableHandle
             @JsonProperty("tableType") HoodieTableType tableType,
             @JsonProperty("partitionColumns") List<HiveColumnHandle> partitionColumns,
             @JsonProperty("partitionPredicates") TupleDomain<HiveColumnHandle> partitionPredicates,
-            @JsonProperty("regularPredicates") TupleDomain<HiveColumnHandle> regularPredicates)
+            @JsonProperty("regularPredicates") TupleDomain<HiveColumnHandle> regularPredicates,
+            @JsonProperty("readVersion") Optional<String> readVersion)
     {
-        this(schemaName, tableName, basePath, tableType, partitionColumns, ImmutableSet.of(), partitionPredicates, regularPredicates);
+        this(schemaName, tableName, basePath, tableType, partitionColumns, ImmutableSet.of(), partitionPredicates, regularPredicates, readVersion);
     }
 
     public HudiTableHandle(
@@ -63,7 +66,8 @@ public class HudiTableHandle
             List<HiveColumnHandle> partitionColumns,
             Set<HiveColumnHandle> constraintColumns,
             TupleDomain<HiveColumnHandle> partitionPredicates,
-            TupleDomain<HiveColumnHandle> regularPredicates)
+            TupleDomain<HiveColumnHandle> regularPredicates,
+            Optional<String> readVersion)
     {
         this.schemaName = requireNonNull(schemaName, "schemaName is null");
         this.tableName = requireNonNull(tableName, "tableName is null");
@@ -73,6 +77,7 @@ public class HudiTableHandle
         this.constraintColumns = requireNonNull(constraintColumns, "constraintColumns is null");
         this.partitionPredicates = requireNonNull(partitionPredicates, "partitionPredicates is null");
         this.regularPredicates = requireNonNull(regularPredicates, "regularPredicates is null");
+        this.readVersion = requireNonNull(readVersion, "readVersion is null");
     }
 
     @JsonProperty
@@ -124,6 +129,12 @@ public class HudiTableHandle
         return regularPredicates;
     }
 
+    @JsonProperty
+    public Optional<String> getReadVersion()
+    {
+        return readVersion;
+    }
+
     public SchemaTableName getSchemaTableName()
     {
         return schemaTableName(schemaName, tableName);
@@ -142,7 +153,8 @@ public class HudiTableHandle
                 partitionColumns,
                 constraintColumns,
                 partitionPredicates.intersect(partitionTupleDomain),
-                regularPredicates.intersect(regularTupleDomain));
+                regularPredicates.intersect(regularTupleDomain),
+                readVersion);
     }
 
     @Override
--- a/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
+++ b/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
@@ -35,11 +35,13 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.airlift.units.DataSize.Unit.MEGABYTE;
 import static java.lang.Math.max;
 import static java.lang.Math.min;
+import static java.util.Objects.requireNonNull;
 import static org.apache.hudi.common.table.view.HoodieTableFileSystemView.fileListingBasedFileSystemView;
 
 public class HudiReadOptimizedDirectoryLister
@@ -51,6 +53,7 @@ public class HudiReadOptimizedDirectoryLister
     private final HoodieTableFileSystemView fileSystemView;
     private final List<Column> partitionColumns;
     private final Map<String, HudiPartitionInfo> allPartitionInfoMap;
+    private final HudiTableHandle tableHandle;
 
     public HudiReadOptimizedDirectoryLister(
             HudiTableHandle tableHandle,
@@ -76,13 +79,17 @@ public class HudiReadOptimizedDirectoryLister
                                 tableHandle.getPartitionPredicates(),
                                 hiveTable,
                                 hiveMetastore)));
+        this.tableHandle = requireNonNull(tableHandle, "tableHandle is null");
     }
 
     @Override
     public List<HudiFileStatus> listStatus(HudiPartitionInfo partitionInfo)
     {
         LOG.debug("List partition: partitionInfo=%s", partitionInfo);
-        return fileSystemView.getLatestBaseFiles(partitionInfo.getRelativePartitionPath())
+        Stream<HoodieBaseFile> baseFileStream = tableHandle.getReadVersion().isEmpty()
+                ? fileSystemView.getLatestBaseFiles(partitionInfo.getRelativePartitionPath())
+                : fileSystemView.getLatestBaseFilesBeforeOrOn(partitionInfo.getRelativePartitionPath(), tableHandle.getReadVersion().get());
+        return baseFileStream
                 .map(HudiReadOptimizedDirectoryLister::getStoragePathInfo)
                 .map(fileEntry -> new HudiFileStatus(
                         Location.of(fileEntry.getPath().toString()),
--- a/plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
+++ b/plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
@@ -25,13 +25,15 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class TestHudiConnectorTest
         extends BaseConnectorTest
 {
+    private static final long COMMIT_TIMESTAMP = 20251027183851494L;
+
     @Override
     protected QueryRunner createQueryRunner()
             throws Exception
     {
         return HudiQueryRunner.builder()
                 .addConnectorProperty("hudi.columns-to-hide", COLUMNS_TO_HIDE)
-                .setDataLoader(new TpchHudiTablesInitializer(REQUIRED_TPCH_TABLES))
+                .setDataLoader(new TpchHudiTablesInitializer(REQUIRED_TPCH_TABLES, String.valueOf(COMMIT_TIMESTAMP)))
                 .build();
     }
 
@@ -87,5 +89,25 @@ public class TestHudiConnectorTest
     {
         assertThat(computeActual("SHOW SCHEMAS").getOnlyColumnAsSet()).doesNotContain("sys");
         assertQueryFails("SHOW TABLES IN hudi.sys", ".*Schema 'sys' does not exist");
+    }
+
+    @Override
+    protected void verifyVersionedQueryFailurePermissible(Exception e)
+    {
+        assertThat(e)
+                .hasMessageMatching("Read table with start version is not supported|" +
+                        "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead|" +
+                        "Provided read version must be a string");
+    }
+
+    @Test
+    public void testSelectTableUsingTargetIdVersion()
+    {
+        String expectedValues = "VALUES 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24";
+        assertThat(query("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation")).matches(expectedValues);
+        assertThat(query("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR VERSION AS OF '" + COMMIT_TIMESTAMP + "'")).matches(expectedValues);
+        assertThat(query("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR VERSION AS OF '" + (COMMIT_TIMESTAMP - 1) + "'")).returnsEmptyResult();
+        assertQueryFails("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR VERSION AS OF 0", "Provided read version must be a string");
+        assertQueryFails("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR TIMESTAMP AS OF TIMESTAMP '2025-10-29 15:00:00'", "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead");
     }
 }
--- a/plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
+++ b/plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
@@ -113,10 +113,12 @@ public class TpchHudiTablesInitializer
     private static final HdfsContext CONTEXT = new HdfsContext(SESSION);
 
     private final List<TpchTable<?>> tpchTables;
+    private final String commitTimestamp;
 
-    public TpchHudiTablesInitializer(List<TpchTable<?>> tpchTables)
+    public TpchHudiTablesInitializer(List<TpchTable<?>> tpchTables, String commitTimestamp)
     {
         this.tpchTables = requireNonNull(tpchTables, "tpchTables is null");
+        this.commitTimestamp = requireNonNull(commitTimestamp, "commitTimestamp is null");
     }
 
     @Override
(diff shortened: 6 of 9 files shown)


docs/src/main/sphinx/connector/hudi.md
```diff
@@ -101,6 +101,10 @@ Additionally, following configuration properties can be set depending on the use
     failing the query. This skips data that may be expected to be part of the
     table.
   - `false`
+* - `hudi.time_travel_read_timestamp`
+  - Read data from the table as of a certain commit timestamp. If left as an
+    empty string, Trino will read from the most recent commit timestamp.
+  - `''`
```
ebyhr: Why do we use a session property instead of a time travel syntax (`FOR TIMESTAMP AS OF`)?

jordepic: Great call.  Changed to fix.

docs/src/main/sphinx/connector/hudi.md
```diff
@@ -46,61 +46,61 @@ Additionally, following configuration properties can be set depending on the use
 :header-rows: 1
 
 * - Property name
-  - Description
-  - Default
+- Description
```
ebyhr: Please revert unrelated changes.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -93,6 +94,28 @@ public HudiMetadata(HiveMetastore metastore, TrinoFileSystemFactory fileSystemFa
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
     }
 
+    static long getReadTimestamp(Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
```
ebyhr: Move above `getSystemTable` method.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiSessionProperties.java
```diff
@@ -218,4 +212,10 @@ public static boolean isIgnoreAbsentPartitions(ConnectorSession session)
     {
         return session.getProperty(IGNORE_ABSENT_PARTITIONS, Boolean.class);
     }
+
+    @Override
```
ebyhr: Please revert unrelated changes.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiMetadata.java
```diff
@@ -0,0 +1,63 @@
+/*
```
ebyhr: Please update `TestHudiConnectorTest` for testing time travel with query. 

There is no need to add `TestHudiMetadata` in this PR. It's not so helpful.

docs/src/main/sphinx/connector/hudi.md
jordepic: Done.  Sorry, weird IntelliJ formatting.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: Done.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiMetadata.java
jordepic: Done.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiSessionProperties.java
jordepic: Done.

jordepic:
@ebyhr I signed my CLA but nobody has reviewed it in a week.  Would it be possible to get anybody to take a look?

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
```diff
@@ -87,4 +94,39 @@ public void testHideHiveSysSchema()
...
+    {
+        assertThat(HudiMetadata.getReadTimestamp(Optional.empty(), Optional.empty())).isEqualTo(Long.MAX_VALUE);
+    }
+
+    @Test
+    public void usesProvidedTimestampWhenEndVersionSpecified()
+    {
+        ConnectorTableVersion endVersion = new ConnectorTableVersion(PointerType.TARGET_ID, IpAddressType.IPADDRESS, 1L);
+        assertThat(HudiMetadata.getReadTimestamp(Optional.empty(), Optional.of(endVersion))).isEqualTo(1L);
+    }
```
ebyhr, line 112: Please add query based tests instead. You can refer to Iceberg and Delta Lake connectors.

docs/src/main/sphinx/connector/hudi.md
```diff
@@ -170,6 +170,16 @@ GROUP BY dt;
 (1 rows)
 ```
 
+### Time travel queries
+
+The connector offers the ability to query historical data. This allows viewing the state of the table at a previous snapshot.
```
ebyhr: Wrap at 80 characters.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
jordepic: Done, thank you.

ebyhr:
@cla-bot check

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
```diff
@@ -76,13 +78,23 @@ public HudiReadOptimizedDirectoryLister(
...
                                 hiveMetastore)));
+        this.tableHandle = tableHandle;
+    }
+
+    private static StoragePathInfo getStoragePathInfo(HoodieBaseFile baseFile)
+    {
+        if (baseFile.getBootstrapBaseFile().isPresent()) {
+            return baseFile.getBootstrapBaseFile().get().getPathInfo();
+        }
+        return baseFile.getPathInfo();
```
ebyhr: Revert unrelated change.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
```diff
@@ -76,13 +78,23 @@ public HudiReadOptimizedDirectoryLister(
                                 tableHandle.getPartitionPredicates(),
                                 hiveTable,
                                 hiveMetastore)));
+        this.tableHandle = tableHandle;
```
ebyhr: Add `requireNonNull`.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
                 COPY_ON_WRITE,
                 getPartitionKeyColumnHandles(table.get(), typeManager),
                 TupleDomain.all(),
-                TupleDomain.all());
+                TupleDomain.all(),
+                getReadTimestamp(startVersion, endVersion));
+    }
+
+    static long getReadTimestamp(Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
```
ebyhr: This method can be private.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
...
+
+        long readTimestamp = MAX_VERSION;
+        if (endVersion.isPresent()) {
+            if (endVersion.get().getVersion() instanceof Long versionNumber && String.valueOf(versionNumber).length() == VERSION_LENGTH) {
+                readTimestamp = versionNumber;
+            }
+            else {
+                throw new TrinoException(NOT_SUPPORTED, "Provided read timestamp must be a number in format yyyyMMddHHmmssSSS of the table's time zone");
+            }
+        }
```
ebyhr: Hudi uses String for timestamp. Why do we want to use numeric type here?

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
```diff
@@ -50,9 +51,10 @@ public HudiTableHandle(
             @JsonProperty("tableType") HoodieTableType tableType,
             @JsonProperty("partitionColumns") List<HiveColumnHandle> partitionColumns,
             @JsonProperty("partitionPredicates") TupleDomain<HiveColumnHandle> partitionPredicates,
-            @JsonProperty("regularPredicates") TupleDomain<HiveColumnHandle> regularPredicates)
+            @JsonProperty("regularPredicates") TupleDomain<HiveColumnHandle> regularPredicates,
+            @JsonProperty("readAsOf") long readTimestamp)
```
ebyhr: Why do we use a different JsonProperty value from the parameter name?

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
```diff
@@ -113,10 +113,12 @@ public class TpchHudiTablesInitializer
     private static final HdfsContext CONTEXT = new HdfsContext(SESSION);
 
     private final List<TpchTable<?>> tpchTables;
+    private final String commitTimestamp;
 
-    public TpchHudiTablesInitializer(List<TpchTable<?>> tpchTables)
+    public TpchHudiTablesInitializer(List<TpchTable<?>> tpchTables, String commitTimestamp)
     {
         this.tpchTables = requireNonNull(tpchTables, "tpchTables is null");
+        this.commitTimestamp = commitTimestamp;
```
ebyhr: Add `requireNonNull`.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
...
                 getPartitionKeyColumnHandles(table.get(), typeManager),
                 TupleDomain.all(),
-                TupleDomain.all());
+                TupleDomain.all(),
+                getReadTimestamp(startVersion, endVersion));
+    }
+
+    static long getReadTimestamp(Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
+    {
+        if (startVersion.isPresent()) {
```
chenjian2664, line 138: I think we should move this check earlier, similar to how it’s done in Delta Lake

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
...
+            throw new TrinoException(NOT_SUPPORTED, "Read table with start version is not supported");
+        }
+
+        if (endVersion.isPresent() && endVersion.get().getPointerType().equals(PointerType.TEMPORAL)) {
+            throw new TrinoException(NOT_SUPPORTED, "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead");
+        }
+
+        long readTimestamp = MAX_VERSION;
```
chenjian2664: "readTimestamp" -> "readVersion"

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
```diff
@@ -76,13 +78,23 @@ public HudiReadOptimizedDirectoryLister(
...
+        }
+        return baseFile.getPathInfo();
     }
 
     @Override
     public List<HudiFileStatus> listStatus(HudiPartitionInfo partitionInfo)
     {
         LOG.debug("List partition: partitionInfo=%s", partitionInfo);
-        return fileSystemView.getLatestBaseFiles(partitionInfo.getRelativePartitionPath())
+        Stream<HoodieBaseFile> baseFileStream = fileSystemView.getLatestBaseFilesBeforeOrOn(partitionInfo.getRelativePartitionPath(), String.valueOf(tableHandle.getReadTimestamp()));
```
chenjian2664: Just to confirm: does "reading latest table version" basically mean "reading the latest table with MAX_VERSION", since previously we don't pass version, now if the `endVersion` not exists, we will pass MAX_VERSION to here

jordepic: I think so, but to be safe I'll just use an optional.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
jordepic: Done.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/query/HudiReadOptimizedDirectoryLister.java
jordepic: Done

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: Done

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: Done, agreed this is better.  Didn't want to let users specify something random, but the HUDI api allows that anwyays.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
jordepic: Done.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
jordepic: Done.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: Done.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: Done.

docs/src/main/sphinx/connector/hudi.md
jordepic: Done.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: I moved it slightly earlier.  One of the tests (testing a specific error occurs if we do a time travel read on a non existent table) fails otherwise.

Going to leave like this for now.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
```diff
@@ -88,4 +90,22 @@ public void testHideHiveSysSchema()
...
+                .hasMessageMatching("Read table with start version is not supported|" +
+                        "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead|" +
+                        "Provided read version must be a string");
+    }
+
+    @Test
+    public void testSelectTableUsingTargetIdVersion()
+    {
+        assertQuery("SELECT nationkey FROM hudi.tests.nation", "VALUES 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24");
+        assertThat(query("SELECT nationkey FROM hudi.tests.nation FOR VERSION AS OF '" + (COMMIT_TIMESTAMP - 1) + "'")).returnsEmptyResult();
```
voonhous: Current test covers "before", add another assert to test for "on":

```
String expectedData = "VALUES 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24";

assertQuery("SELECT nationkey FROM hudi.tests.nation FOR VERSION AS OF '" + COMMIT_TIMESTAMP + "'", expectedData);
```

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
...
+
+        Optional<String> readVersion = Optional.empty();
+        if (endVersion.isPresent()) {
+            if (endVersion.get().getVersion() instanceof Slice slice) {
+                readVersion = Optional.of(slice.toStringUtf8());
+            }
+            else {
+                throw new TrinoException(NOT_SUPPORTED, "Provided read version must be a string");
+            }
+        }
```
voonhous, line 153: Nit: Flatten the logic to make it more readable (stylistic):

```java
private static Optional<String> getReadVersion(Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
{
    if (startVersion.isPresent()) {
        throw new TrinoException(NOT_SUPPORTED, "Read table with start version is not supported");
    }

    if (endVersion.isEmpty()) {
        return Optional.empty();
    }

    ConnectorTableVersion version = endVersion.get();
    if (version.getPointerType().equals(PointerType.TEMPORAL)) {
        throw new TrinoException(NOT_SUPPORTED, "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead");
    }

    if (version.getVersion() instanceof Slice slice) {
        return Optional.of(slice.toStringUtf8());
    }

    throw new TrinoException(NOT_SUPPORTED, "Provided read version must be a string");
}
```

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
```diff
@@ -124,6 +129,12 @@ public TupleDomain<HiveColumnHandle> getRegularPredicates()
         return regularPredicates;
     }
 
+    @JsonProperty
+    public Optional<String> getReadVersion()
+    {
+        return readVersion;
```
voonhous, line 135: nit: `readVersion` is not a standard hudi term. Is it possible we standardise the nomenclature? I think `asOfInstant` would be a good replacement.


Instead of:

```java
tableHandle.getReadVersion().get()
```


We can use:
```java
tableHandle.getAsOfInstant().get()
```

What do you think?

jordepic: I don't see "as of instant" in the official hudi docs in there either.  AS OF is in there, but I honestly don't think that naming is really going to improve perception of things here either.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
ebyhr: It's not done.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
...
+        if (startVersion.isPresent()) {
+            throw new TrinoException(NOT_SUPPORTED, "Read table with start version is not supported");
+        }
+
+        if (endVersion.isEmpty()) {
+            return Optional.empty();
+        }
+
+        ConnectorTableVersion version = endVersion.get();
+        if (version.getPointerType().equals(PointerType.TEMPORAL)) {
```
ebyhr:
```suggestion
        if (version.getPointerType() == PointerType.TEMPORAL) {
```

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
jordepic: Done!

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
voonhous:
```java
org.apache.hudi.common.config.HoodieCommonConfig#TIMESTAMP_AS_OF
```

ebyhr: There is no need to strictly follow Hudi's name convention here. The current name is fine.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
```diff
@@ -88,4 +90,24 @@ public void testHideHiveSysSchema()
...
+                .hasMessageMatching("Read table with start version is not supported|" +
+                        "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead|" +
+                        "Provided read version must be a string");
+    }
+
+    @Test
+    public void testSelectTableUsingTargetIdVersion()
+    {
+        String expectedValues = "VALUES 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24";
+        assertQuery("SELECT nationkey FROM hudi.tests.nation", expectedValues);
```
ebyhr: Use `assertThat(query(...)).matches(...)` instead. Same for L108.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
```diff
@@ -88,4 +90,24 @@ public void testHideHiveSysSchema()
...
+        assertQuery("SELECT nationkey FROM hudi.tests.nation FOR VERSION AS OF '" + (COMMIT_TIMESTAMP) + "'", expectedValues);
```
ebyhr:
```suggestion
        assertQuery("SELECT nationkey FROM hudi.tests.nation FOR VERSION AS OF '" + COMMIT_TIMESTAMP + "'", expectedValues);
```

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiTableHandle.java
voonhous: Sure, just calling it out as i thought we are all on the same page of fixing stylistic changes. We'll just have to be wary about the different terms then.

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
chenjian2664: What's the test is that? is that just the error message or exception type not match?

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
```diff
@@ -130,7 +129,30 @@ public HudiTableHandle getTableHandle(ConnectorSession session, SchemaTableName
...
+            throw new TrinoException(NOT_SUPPORTED, "Read table with start version is not supported");
+        }
+
+        if (endVersion.isEmpty()) {
+            return Optional.empty();
+        }
+
+        ConnectorTableVersion version = endVersion.get();
+        if (version.getPointerType() == PointerType.TEMPORAL) {
+            throw new TrinoException(NOT_SUPPORTED, "Cannot read 'TIMESTAMP' of Hudi table, use 'VERSION' instead");
```
chenjian2664, line 148:
```suggestion
            throw new TrinoException(NOT_SUPPORTED, "This connector does not support reading tables with TIMESTAMP AS OF");
```

chenjian2664, line 148: Also, is this limitation(reading with temporal) coming from Hudi, or is it something that could be supported in the future?
If it's the latter, could you file an issue and add a TODO here with a link to it?

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
```diff
@@ -88,4 +90,24 @@ public void testHideHiveSysSchema()
...
+    }
+
+    @Test
+    public void testSelectTableUsingTargetIdVersion()
+    {
+        String expectedValues = "VALUES 0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 23, 24";
+        assertThat(query("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation")).matches(expectedValues);
+        assertThat(query("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR VERSION AS OF '" + COMMIT_TIMESTAMP + "'")).matches(expectedValues);
+        assertThat(query("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR VERSION AS OF '" + (COMMIT_TIMESTAMP - 1) + "'")).returnsEmptyResult();
+        assertQueryFails("SELECT CAST(nationkey AS INT) FROM hudi.tests.nation FOR VERSION AS OF 0", "Provided read version must be a string");
```
chenjian2664, line 110: Could you also add a case that is negative number/string?

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
```diff
@@ -25,13 +25,15 @@
 public class TestHudiConnectorTest
         extends BaseConnectorTest
 {
+    private static final long COMMIT_TIMESTAMP = 20251027183851494L;
```
chenjian2664, line 28: Is the naming aligned with the Hudi glossary? using "commit timestamp" sounds more like a temporal read, it actually more like a "commit version"

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
jordepic: Hudi just takes in an arbitrary string.  I think it would be a bit misleading to a user to take in a timestamp, as it may leave some questions as to how that gets converted to the string.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
jordepic: What do you expect to happen in that case?  We're doing lexicographic string comparisons, so it wouldn't make much sense :(

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
jordepic: We had a similar discussion to this with @voonhous , and @ebyhr seemed to imply that keeping consistent with hudi vernacular wasn't too crucial.  It's worth noting that the commit version is logically equal to the timestamp of the last commit

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
chenjian2664: We have two optionals in Trino to do time travel, use "FOR VERSION " or "FOR TIMESTAMP".
The hudi time travel seems https://hudi.apache.org/docs/sql_queries/#time-travel-query does NOT distinguish version or timestamp, so I think it is our limitation, I recommend to file an issue to track it, we could also convert temporal to a Hudi accepted string to support it in future.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/TestHudiConnectorTest.java
chenjian2664: Since it is obviously a "wrong" case, I am wondering Hudi support it or not, if not I think it is better to add a check inadvance on our side, that said no matter what's the decision is, we should add a test case to track the behavior first, if some logic changed on Hudi side, that won't surprise us too much.

voonhous:
@ebyhr What is left unaddressed here? What's blocking the merge?

plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiMetadata.java
chenjian2664: reminder

PR #27150, opened 2025-10-29
Add connection properties for Druid connector

yangshangqing95:
## Description
This PR adds support for configuring additional Druid connector properties, specifically timeout.
The motivation is to provide client-side safeguards: poorly written SQL queries can trigger full table scans in Druid, putting significant pressure on the Druid cluster. By allowing these properties to be set on the client side, we can limit query execution time and control time zone interpretation, mitigating potential cluster impact.

`druid.execution-timeout` allows users to limit the maximum execution time of a query.

## Release notes

```markdown
## Druid
* Add support for configuring `druid.execution-timeout` config property. ({issue}`27150`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/druid.md
+++ b/docs/src/main/sphinx/connector/druid.md
@@ -53,6 +53,21 @@ determine the user credentials for the connection, often a service user. You can
 use {doc}`secrets </security/secrets>` to avoid actual values in the catalog
 properties files.
 
+Additionally, the following configuration properties can be set depending
+on the use-case:
+
+:::{list-table} Druid Configuration Properties
+:widths: 30, 35, 35
+:header-rows: 1
+
+* - Property Name
+  - Description
+  - Default
+* - `druid.execution-timeout`
+  - Query timeout for Druid queries, beyond which unfinished queries will be cancelled. Example: `10s`, `1m`, and `1000ms`.
+  - `None`
+:::
+
 ```{include} jdbc-authentication.fragment
 ```
 
--- a/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
+++ b/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
@@ -1,0 +1,38 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.druid;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
+import io.airlift.units.Duration;
+
+import java.util.Optional;
+
+public class DruidConfig
+{
+    private Duration executionTimeout;
+
+    public Optional<Duration> getExecutionTimeout()
+    {
+        return Optional.ofNullable(executionTimeout);
+    }
+
+    @Config("druid.execution-timeout")
+    @ConfigDescription("Maximum time to wait for a Druid query to complete before timing out")
+    public DruidConfig setExecutionTimeout(Duration executionTimeout)
+    {
+        this.executionTimeout = executionTimeout;
+        return this;
+    }
+}
--- a/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
+++ b/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
@@ -32,6 +32,7 @@ import org.apache.calcite.avatica.remote.Driver;
 import java.util.Properties;
 
 import static com.google.inject.multibindings.Multibinder.newSetBinder;
+import static io.airlift.configuration.ConfigBinder.configBinder;
 
 public class DruidJdbcClientModule
         implements Module
@@ -40,19 +41,28 @@ public class DruidJdbcClientModule
     public void configure(Binder binder)
     {
         binder.bind(JdbcClient.class).annotatedWith(ForBaseJdbc.class).to(DruidJdbcClient.class).in(Scopes.SINGLETON);
+        configBinder(binder).bindConfig(DruidConfig.class);
         newSetBinder(binder, ConnectorTableFunction.class).addBinding().toProvider(Query.class).in(Scopes.SINGLETON);
     }
 
     @Provides
     @Singleton
     @ForBaseJdbc
-    public static ConnectionFactory createConnectionFactory(BaseJdbcConfig config, CredentialProvider credentialProvider, OpenTelemetry openTelemetry)
+    public static ConnectionFactory createConnectionFactory(BaseJdbcConfig config, CredentialProvider credentialProvider, DruidConfig druidConfig, OpenTelemetry openTelemetry)
     {
         Properties connectionProperties = new Properties();
 
         // Druid can be configured to approximate `count_distinct()` function calls.
         // This behavior could be confusing - we can force Druid to never approximate.
         connectionProperties.setProperty("useApproximateCountDistinct", "false");
+        // According to Druid documentation, Druid does not support passing connection context parameters from the
+        // JDBC connection string to Druid. These context parameters must be passed using a Properties object instead
+        // https://druid.apache.org/docs/latest/api-reference/sql-jdbc/
+        if (druidConfig.getExecutionTimeout().isPresent()) {
+            // Query timeout in millis, beyond which unfinished queries will be cancelled
+            // https://druid.apache.org/docs/latest/querying/query-context-reference/
+            connectionProperties.setProperty("timeout", String.valueOf(druidConfig.getExecutionTimeout().get().toMillis()));
+        }
         return DriverConnectionFactory.builder(new Driver(), config.getConnectionUrl(), credentialProvider)
                 .setOpenTelemetry(openTelemetry)
                 .setConnectionProperties(connectionProperties)
--- a/plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConfig.java
+++ b/plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConfig.java
@@ -1,0 +1,48 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.druid;
+
+import com.google.common.collect.ImmutableMap;
+import io.airlift.units.Duration;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static io.airlift.configuration.testing.ConfigAssertions.assertFullMapping;
+import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
+import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+final class TestDruidConfig
+{
+    @Test
+    void testDefaults()
+    {
+        assertRecordedDefaults(recordDefaults(DruidConfig.class)
+                .setExecutionTimeout(null));
+    }
+
+    @Test
+    void testExplicitPropertyMappings()
+    {
+        Map<String, String> properties = ImmutableMap.<String, String>builder()
+                .put("druid.execution-timeout", "10s")
+                .buildOrThrow();
+
+        DruidConfig expected = new DruidConfig()
+                .setExecutionTimeout(new Duration(10, SECONDS));
+
+        assertFullMapping(properties, expected);
+    }
+}
--- a/plugin/trino-druid/pom.xml
+++ b/plugin/trino-druid/pom.xml
@@ -27,6 +27,16 @@
         </dependency>
 
         <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>configuration</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>units</artifactId>
+        </dependency>
+
+        <dependency>
             <groupId>io.trino</groupId>
             <artifactId>trino-base-jdbc</artifactId>
         </dependency>
@@ -105,6 +115,12 @@
 
         <dependency>
             <groupId>io.airlift</groupId>
+            <artifactId>configuration-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
             <artifactId>junit-extensions</artifactId>
             <scope>test</scope>
         </dependency>
@@ -124,12 +140,6 @@
         <dependency>
             <groupId>io.airlift</groupId>
             <artifactId>testing</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.airlift</groupId>
-            <artifactId>units</artifactId>
             <scope>test</scope>
         </dependency>
 


plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
```diff
@@ -38,16 +42,30 @@ public class DruidJdbcClientModule
...
+
+    private static Properties getConnectionProperties(DruidConfig druidConfig)
+    {
+        Properties connectionProperties = new Properties();
+        if (druidConfig.getExecutionTimeout() != 0) {
+            connectionProperties.setProperty("timeout", String.valueOf(druidConfig.getExecutionTimeout()));
+        }
+        if (!Strings.isNullOrEmpty(druidConfig.getSqlTimeZone())) {
+            connectionProperties.setProperty("sqlTimeZone", druidConfig.getSqlTimeZone());
+        }
```
ebyhr: We can already set these properties via the `connection-url` without this PR, right?

yangshangqing95: It seems that we can set properties in the URL separated by semicolons, but in my actual tests, this method didn’t work. If there are other ways to configure the settings, please let me know, I can run further tests.
My test case:
`jdbc:avatica:remote:url=xxxxx/avatica/;timeout=15`

yangshangqing95: And Druid document says that the context parameters CAN NOT be set in the URL
image
They must be set in the code.

yangshangqing95: Hi @ebyhr , any update on this one? thx

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,47 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.druid;
+
+import io.airlift.configuration.Config;
+
+public class DruidConfig
+{
+    private long executionTimeout; // milliseconds
```
ebyhr: Use airlift duration class.

yangshangqing95: done

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+package io.trino.plugin.druid;
+
+import io.airlift.configuration.Config;
+import io.airlift.units.Duration;
+
+public class DruidConfig
+{
+    private Duration executionTimeout; // milliseconds
```
ebyhr: Remove the obsolete code comment.

yangshangqing95: done

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConfig.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static io.airlift.configuration.testing.ConfigAssertions.assertFullMapping;
+import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
+import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+class TestDruidConfig
```
chenjian2664: final

yangshangqing95: done

yangshangqing95:
Hi @chenjian2664 @ebyhr, is this good to be merged? thx

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,27 @@ connection-user=root
...
+
+:::{list-table} Druid configuration properties
+:widths: 30, 55, 15
+:header-rows: 1
+
+* - Property name
+  - Description
+  - Default
+* - `druid.execution-timeout`
+  - Query timeout for druid queries, beyond which unfinished queries will be cancelled.
```
chenjian2664: druid -> Druid

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
```diff
@@ -38,16 +42,30 @@ public class DruidJdbcClientModule
...
                 .build();
     }
+
+    private static Properties getConnectionProperties(DruidConfig druidConfig)
+    {
+        Properties connectionProperties = new Properties();
+        if (druidConfig.getExecutionTimeout() != null) {
+            connectionProperties.setProperty("timeout", String.valueOf(druidConfig.getExecutionTimeout().toMillis()));
+        }
+        if (!Strings.isNullOrEmpty(druidConfig.getSqlTimeZone())) {
```
chenjian2664: static import

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,48 @@
...
+        this.executionTimeout = executionTimeout;
+        return this;
+    }
+
+    public Duration getExecutionTimeout()
+    {
+        return executionTimeout;
+    }
+
+    @Config("druid.sql-timezone")
```
chenjian2664: What's the default value if we don't specify it in Druid?
Can we add a test to confirm that this property does not affect the correctness of time-related reading in Druid?

docs/src/main/sphinx/connector/druid.md
yangshangqing95: done

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
yangshangqing95: done

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
yangshangqing95: Hi @chenjian2664 
The default value is Druid is UTC.
image
As for the testing, this is a native Druid SQL parameter. Trino does not interpret or apply this value by itself, we only pass it through to Druid. From the Trino connector’s perspective, this is equivalent to other pass-through configuration properties used by JDBC-style connectors.
Therefore, we should not add Trino-side tests to validate Druid’s timezone correctness. Such validation belongs to Druid itself rather than the Trino connector.

chenjian2664: Thanks. I am not clear why we needs this property, this is more likely a choice that should control by Druid itself, Trino also offering similar [functionility](https://trino.io/docs/current/sql/set-time-zone.html)

yangshangqing95: Yes, Trino supports setting the time zone.
The key point is that we want to support configuring the runtime; the time zone was just added incidentally.
It can be removed. @chenjian2664

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,23 @@ connection-user=root
...
+
+:::{list-table} Druid configuration properties
+:widths: 30, 55, 15
+:header-rows: 1
+
+* - Property name
+  - Description
+  - Default
+* - `druid.execution-timeout`
+  - Query timeout for Druid queries, beyond which unfinished queries will be cancelled.
```
chenjian2664: The cancellation is happening on server side automatically or it affects the client behavior, i,e sending cancellation + stop waiting results?

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,23 @@ connection-user=root
...
+:widths: 30, 55, 15
+:header-rows: 1
+
+* - Property name
+  - Description
+  - Default
+* - `druid.execution-timeout`
+  - Query timeout for Druid queries, beyond which unfinished queries will be cancelled. 
+    0 timeout means no timeout (up to the server-side maximum query timeout,
+    `druid.server.http.maxQueryTimeout`)
```
chenjian2664: Does the value has min/max constraint on Druid side?
we could set to "0ms" if it is same as the default behavior so that we could validate the in configuration class make that not null

docs/src/main/sphinx/connector/druid.md
yangshangqing95: server side

docs/src/main/sphinx/connector/druid.md
yangshangqing95: > 1. Does the value has min/max constraint on Druid side?
Depends on the Druid server side configuration, we don't need to consider about this.
> 2. we could set to "0ms" if it is same as the default behavior so that we could validate the in configuration class make that not null
Updated
Hi @chenjian2664 pls help review, thx

chenjian2664:
The CI failure seems tied to this change. Could we be missing something about the default behavior?

yangshangqing95:
> The CI failure seems tied to this change. Could we be missing something about the default behavior?

Hi @chenjian2664 The PR is quite simple and should be no missing part.
And the failed tests is not related to the PR itself.
image
Could there be some change in the CI part and should rebase master?

ebyhr:
@yangshangqing95 The change in DruidJdbcClientModule affects the entire Druid connector. I believe the CI failure is related to this PR change.

yangshangqing95:
Hi @ebyhr @chenjian2664 
It turns out the issue was caused by the default value of the timeout property being set to 0, which behaves differently from what’s described in the Druid official documentation. This might be related to the older Druid version used in CI.
I’ve reverted the default value of the timeout configuration back to null, and the CI is now passing.
image

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,21 @@ connection-user=root
 connection-password=secret
 ```
 
+Additionally, the following configuration properties can be set depending on the use-case:
```
ebyhr: Wrap at 80 characters.

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,34 @@
...
+package io.trino.plugin.druid;
+
+import io.airlift.configuration.Config;
+import io.airlift.units.Duration;
+
+public class DruidConfig
+{
+    private Duration executionTimeout;
+
+    @Config("druid.execution-timeout")
```
ebyhr, line 31: Could you add `@ConfigDescription`?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,34 @@
...
+    public DruidConfig setExecutionTimeout(Duration executionTimeout)
+    {
+        this.executionTimeout = executionTimeout;
+        return this;
+    }
+
+    public Duration getExecutionTimeout()
+    {
+        return executionTimeout;
+    }
```
ebyhr: Move this method above `setExecutionTimeout`. Most config classes use getter -> setter order.

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
```diff
@@ -38,16 +41,27 @@ public class DruidJdbcClientModule
...
-    public static ConnectionFactory createConnectionFactory(BaseJdbcConfig config, CredentialProvider credentialProvider, OpenTelemetry openTelemetry)
+    public static ConnectionFactory createConnectionFactory(BaseJdbcConfig config, CredentialProvider credentialProvider, DruidConfig druidConfig, OpenTelemetry openTelemetry)
     {
         return DriverConnectionFactory.builder(new Driver(), config.getConnectionUrl(), credentialProvider)
+                .setConnectionProperties(getConnectionProperties(druidConfig))
                 .setOpenTelemetry(openTelemetry)
                 .build();
     }
+
+    private static Properties getConnectionProperties(DruidConfig druidConfig)
```
ebyhr: nit: There is no need to extract a method. The method is simple and single-usage.

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClientModule.java
```diff
@@ -38,16 +41,27 @@ public class DruidJdbcClientModule
...
+    {
+        Properties connectionProperties = new Properties();
+        if (druidConfig.getExecutionTimeout() != null) {
+            connectionProperties.setProperty("timeout", String.valueOf(druidConfig.getExecutionTimeout().toMillis()));
```
ebyhr: Could you add a link to the Druid documentation (otherwise, the unit is unclear), and leave a code comment explaining why setting via URL doesn't work?

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConfig.java
```diff
@@ -0,0 +1,46 @@
...
+import static io.airlift.configuration.testing.ConfigAssertions.assertRecordedDefaults;
+import static io.airlift.configuration.testing.ConfigAssertions.recordDefaults;
+import static java.util.concurrent.TimeUnit.SECONDS;
+
+final class TestDruidConfig
+{
+    @Test
+    void testDefaults()
+    {
+        assertRecordedDefaults(recordDefaults(DruidConfig.class).setExecutionTimeout(null));
```
ebyhr:
```suggestion
        assertRecordedDefaults(recordDefaults(DruidConfig.class)
                .setExecutionTimeout(null));
```

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidConfig.java
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @Test
+    void testExplicitPropertyMappings()
+    {
+        Map<String, String> properties = ImmutableMap.<String, String>builder()
+                .put("druid.execution-timeout", "10s")
+                .buildOrThrow();
+
+        DruidConfig expected = new DruidConfig().setExecutionTimeout(new Duration(10, SECONDS));
```
ebyhr:
```suggestion
        DruidConfig expected = new DruidConfig()
                .setExecutionTimeout(new Duration(10, SECONDS));
```

ebyhr (commented):
Looks good except for comments.

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,21 @@ connection-user=root
...
+:::{list-table} Druid configuration properties
+:widths: 30, 55, 15
+:header-rows: 1
+
+* - Property name
+  - Description
+  - Default
+* - `druid.execution-timeout`
+  - Query timeout for Druid queries, beyond which unfinished queries will be cancelled.
+    example: `10s`, `1m`, `1000ms`.
```
ebyhr: Please use `prop-type-duration` so we can remove the example.

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,34 @@
...
+    private Duration executionTimeout;
+
+    @Config("druid.execution-timeout")
+    public DruidConfig setExecutionTimeout(Duration executionTimeout)
+    {
+        this.executionTimeout = executionTimeout;
+        return this;
+    }
+
+    public Duration getExecutionTimeout()
```
ebyhr: Add `@MinDuration` and `@NotNull` annotations.

ebyhr:
> Add connection properties for Druid connector

The commit title is too general. Could you change to "Support setting execution timeout in Druid" or something?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
yangshangqing95: Hi @ebyhr
I tend to think this should be an optional configuration. If nothing is set, it will fall back to the default configuration on the Druid server side, so I didn’t add @MinDuration or @NotNull.
Everything else has been updated except for this.
Thx

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -11,20 +11,26 @@
...
+import io.airlift.configuration.ConfigDescription;
+import io.airlift.units.Duration;
 
-public record DynamicFilterAlias(String alias)
+public class DruidConfig
 {
-    public DynamicFilterAlias
+    private Duration executionTimeout;
+
+    public Duration getExecutionTimeout()
```
ebyhr: Please change the return type to `Optional<Duration>`.

yangshangqing95: @ebyhr Updated, thx

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,38 @@
...
+        return executionTimeout == null ? Optional.empty() : Optional.of(executionTimeout);
+    }
+
+    @Config("druid.execution-timeout")
+    @ConfigDescription("Maximum time to wait for a Druid query to complete before timing out")
+    public DruidConfig setExecutionTimeout(Duration executionTimeout)
+    {
+        this.executionTimeout = executionTimeout;
+        return this;
+    }
```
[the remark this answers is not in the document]
ebyhr: This feedback makes sense to me.

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,17 @@ connection-user=root
...
+Additionally, the following configuration properties can be set depending 
+on the use-case:
+
+## `druid.execution-timeout`
+
+- **Type:** [](prop-type-duration)
+- **Examples:** `10s`, `1m`, `1000ms`
+- **Default value:** `None`
+
+Query timeout for Druid queries, beyond which unfinished queries will be cancelled.
```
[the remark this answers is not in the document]
ebyhr: This feedback makes sense to me.

ebyhr:
@yangshangqing95 Could you fix review comments by copilot?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,38 @@
...
+    private Duration executionTimeout;
+
+    public Optional<Duration> getExecutionTimeout()
+    {
+        return executionTimeout == null ? Optional.empty() : Optional.of(executionTimeout);
+    }
+
+    @Config("druid.execution-timeout")
+    @ConfigDescription("Maximum time to wait for a Druid query to complete before timing out")
+    public DruidConfig setExecutionTimeout(Duration executionTimeout)
```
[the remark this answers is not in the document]
yangshangqing95: PR description updated

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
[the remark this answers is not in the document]
yangshangqing95: This does not make sense, Duration can not be negative. @ebyhr 

public Duration(double value, TimeUnit unit) {
        if (Double.isInfinite(value)) {
            throw new IllegalArgumentException("value is infinite: " + value);
        } else {
            Preconditions.checkArgument(!Double.isNaN(value), "value is not a number");
            **if (value < (double)0.0F) {
                throw new IllegalArgumentException("value is negative: " + value);**
            } else {
                Objects.requireNonNull(unit, "unit is null");
                this.value = value;
                this.unit = unit;
            }
        }
    }

docs/src/main/sphinx/connector/druid.md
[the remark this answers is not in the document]
yangshangqing95: updated

yangshangqing95:
> @yangshangqing95 Could you fix review comments by copilot?

Hi @ebyhr updated, pls review, thx

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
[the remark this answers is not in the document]
chenjian2664: @yangshangqing95 For "0ms" it is indeed seems not needed, but it isn't harm if add - at least we could see the info directly instead of having to know the details of `Duration`.

yangshangqing95:
Hi @chenjian2664 got a notification, can this be merged? thx

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,17 @@ connection-user=root
...
+Additionally, the following configuration properties can be set depending 
+on the use-case:
+
+### `druid.execution-timeout`
+
+- **Type:** [](prop-type-duration)
+- **Examples:** `10s`, `1m`, `1000ms`
+- **Default value:** `None`
+
+Query timeout for Druid queries, beyond which unfinished queries will be cancelled.
```
ebyhr: Could you please use `:::{list-table}` and share the screenshot? The current style requires redundant Type,  Examples, Default value prefixes for every new properties.

ebyhr (approved):
Looks good except for a documentation.

docs/src/main/sphinx/connector/druid.md
yangshangqing95: Hi @ebyhr it used to be in a list, I remember I modified the doc to the non-list style by some previous review comments.
But never mind I changed it to the list type.
Would this be OK?
image

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,23 @@ connection-user=root
...
 
+Additionally, the following configuration properties can be set depending 
+on the use-case:
+
+:::{list-table} Druid Execution Configuration Properties
+:widths: 30, 30, 33, 33
+:header-rows: 1
+
+* - Property Name
+  - Type
```
ebyhr: We usually don't add `Type` column to the config table. Please refer to other connectors.

yangshangqing95: Hi @ebyhr updated. The type was added according to previous comments as well.
image

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+import java.util.Optional;
+
+public class DruidConfig
+{
+    private Duration executionTimeout;
+
+    public Optional<Duration> getExecutionTimeout()
+    {
+        return executionTimeout == null ? Optional.empty() : Optional.of(executionTimeout);
```
chenjian2664: nit: `Optional.ofNullable`

chenjian2664 (commented):
LGTM
@ebyhr would you like to take a final look

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,21 @@ connection-user=root
 connection-password=secret
 ```
 
+Additionally, the following configuration properties can be set depending 
+on the use-case:
+
+:::{list-table} Druid Execution Configuration Properties
+:widths: 30, 33, 33
```
ebyhr: The total number should be 100.

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,21 @@ connection-user=root
 connection-password=secret
 ```
 
+Additionally, the following configuration properties can be set depending 
+on the use-case:
+
+:::{list-table} Druid Execution Configuration Properties
```
ebyhr:
```suggestion
:::{list-table} Druid Configuration Properties
```

docs/src/main/sphinx/connector/druid.md
```diff
@@ -45,6 +45,21 @@ connection-user=root
...
+:header-rows: 1
+
+* - Property Name
+  - Description
+  - Default
+* - `druid.execution-timeout`
+  - Query timeout for Druid queries, beyond which unfinished queries will be cancelled. Example: `10s`, `1m`, and `1000ms`.
+  - `None`
+:::
+
```
ebyhr, line 70: The new documentation has been added in the middle of `connection-user` and `connection-password`. Please consider moving it above L71.

yangshangqing95: image
how about here?

docs/src/main/sphinx/connector/druid.md
yangshangqing95: done

docs/src/main/sphinx/connector/druid.md
yangshangqing95: done

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidConfig.java
yangshangqing95: done

PR #27151, opened 2025-10-29
Some minor tweaks and cleanups

wendigo:
Mostly around:

- reducing (un)boxing of Integer/Longs:
  - by using OptionalInt instead of Optional<Integer>
  - using Comparator.comparingLong/comparingInt instead of boxing
  - using Int2ObjectMap when keys are primitive ints
- using presized Immutable*.Builders if size is known
- using LongAdder when writes are frequent but reads are not

## Release notes

```markdown
## Section
```

## Summary by Sourcery

Refactor codebase to reduce boxing and Stream overhead, improve performance, and standardize on primitive-based APIs

Enhancements:
- Replace Optional<Integer> and boxing-heavy comparators with OptionalInt and primitive comparators to reduce allocation
- Migrate Stream-based matching and lookup APIs to Iterable or Guava Iterables, eliminating unnecessary Stream usage
- Use LongAdder instead of AtomicLong for concurrent counters and Int2ObjectMap for primitive-keyed maps to reduce overhead
- Presize ImmutableList/Map builders when size is known and leverage records and sealed interfaces for cleaner implementations

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/executor/scheduler/TaskControl.java
+++ b/core/trino-main/src/main/java/io/trino/execution/executor/scheduler/TaskControl.java
@@ -18,7 +18,7 @@ import com.google.errorprone.annotations.ThreadSafe;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 
 import java.util.Objects;
-import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.LongAdder;
 import java.util.concurrent.locks.Condition;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
@@ -53,10 +53,10 @@ final class TaskControl
     private State state;
 
     private volatile long periodStart;
-    private final AtomicLong startNanos = new AtomicLong();
-    private final AtomicLong scheduledNanos = new AtomicLong();
-    private final AtomicLong blockedNanos = new AtomicLong();
-    private final AtomicLong waitNanos = new AtomicLong();
+    private final LongAdder startNanos = new LongAdder();
+    private final LongAdder scheduledNanos = new LongAdder();
+    private final LongAdder blockedNanos = new LongAdder();
+    private final LongAdder waitNanos = new LongAdder();
     private final int hashCode;
     private volatile Thread thread;
 
@@ -275,10 +275,10 @@ final class TaskControl
         long now = ticker.read();
         long elapsed = now - periodStart;
         switch (state) {
-            case RUNNING -> scheduledNanos.addAndGet(elapsed);
-            case BLOCKED -> blockedNanos.addAndGet(elapsed);
-            case NEW -> startNanos.addAndGet(elapsed);
-            case WAITING -> waitNanos.addAndGet(elapsed);
+            case RUNNING -> scheduledNanos.add(elapsed);
+            case BLOCKED -> blockedNanos.add(elapsed);
+            case NEW -> startNanos.add(elapsed);
+            case WAITING -> waitNanos.add(elapsed);
             case INTERRUPTED, FINISHED -> {}
         }
         periodStart = now;
@@ -307,22 +307,22 @@ final class TaskControl
 
     public long getStartNanos()
     {
-        return startNanos.get();
+        return startNanos.longValue();
     }
 
     public long getWaitNanos()
     {
-        return waitNanos.get();
+        return waitNanos.longValue();
     }
 
     public long getScheduledNanos()
     {
-        return scheduledNanos.get();
+        return scheduledNanos.longValue();
     }
 
     public long getBlockedNanos()
     {
-        return blockedNanos.get();
+        return blockedNanos.longValue();
     }
 
     @Override


core/trino-main/src/main/java/io/trino/sql/planner/iterative/Memo.java
```diff
@@ -195,10 +196,14 @@ private void decrementReferenceCounts(PlanNode fromNode, int fromGroup)
...
-                .map(GroupReference::getGroupId)
-                .collect(Collectors.toSet());
+        ImmutableSet.Builder<Integer> groupIds = ImmutableSet.builderWithExpectedSize(node.getSources().size());
+        for (PlanNode source : node.getSources()) {
+            if (!(source instanceof GroupReference groupReference)) {
+                throw new IllegalArgumentException("Expected %s to be a group reference".formatted(source));
+            }
+            groupIds.add(groupReference.getGroupId());
+        }
+        return groupIds.build();
```
martint, line 206: Does this actually matter? I find the updated code harder to read.

core/trino-main/src/main/java/io/trino/sql/planner/plan/Assignments.java
```diff
@@ -197,12 +208,7 @@ public Builder putAll(Map<Symbol, ? extends Expression> assignments)
         public Builder put(Symbol symbol, Expression expression)
         {
             checkArgument(symbol.type().equals(expression.type()), "Types don't match: %s vs %s, for %s and %s", symbol.type(), expression.type(), symbol, expression);
-
-            Expression replaced = assignments.put(symbol, expression);
-            if (replaced != null && !replaced.equals(expression)) {
-                assignments.put(symbol, replaced); // restore previous value
-                throw new IllegalStateException("Symbol %s already has assignment %s, while adding %s".formatted(symbol.name(), replaced, expression));
-            }
```
martint, line 205: Where did this go?

core/trino-main/src/main/java/io/trino/sql/planner/iterative/Lookup.java
```diff
@@ -50,18 +54,39 @@ default PlanNode resolve(PlanNode node)
      */
     static Lookup noLookup()
     {
-        return node -> {
-            throw new UnsupportedOperationException();
-        };
```
martint, line 58: I'm pretty sure the JVM treats this as a constant already, so this change is unnecessary.

core/trino-main/src/main/java/io/trino/sql/planner/plan/Assignments.java
wendigo: I reverted this. Duplicates are allowed but only if the inserted element is the same.

lib/trino-matching/src/main/java/io/trino/matching/CapturePattern.java
```diff
@@ -34,9 +34,9 @@ public Capture<T> capture()
     }
 
     @Override
-    public <C> Stream<Match> accept(Object object, Captures captures, C context)
+    public <C> Iterable<Match> accept(Object object, Captures captures, C context)
```
raunaqmorarka: What's the benefit of using Iterable instead of Stream ?
I thought Stream is usually preferred in modern java code

core/trino-main/src/main/java/io/trino/sql/planner/iterative/Lookup.java
```diff
@@ -45,7 +44,7 @@ default PlanNode resolve(PlanNode node)
      *
      * @throws IllegalArgumentException if the node is not a GroupReference
      */
-    Stream<PlanNode> resolveGroup(PlanNode node);
+    Iterable<PlanNode> resolveGroup(PlanNode node);
```
raunaqmorarka: Why does it matter ?

core/trino-main/src/main/java/io/trino/sql/planner/iterative/Lookup.java
raunaqmorarka: +1

core/trino-main/src/main/java/io/trino/sql/planner/plan/Assignments.java
```diff
@@ -197,7 +211,6 @@ public Builder putAll(Map<Symbol, ? extends Expression> assignments)
         public Builder put(Symbol symbol, Expression expression)
         {
             checkArgument(symbol.type().equals(expression.type()), "Types don't match: %s vs %s, for %s and %s", symbol.type(), expression.type(), symbol, expression);
-
```
raunaqmorarka, line 200: unnecessary change

core/trino-main/src/main/java/io/trino/sql/planner/plan/Assignments.java
```diff
@@ -234,7 +247,12 @@ public Builder putIdentity(Symbol symbol)
 
         public Assignments build()
         {
-            return new Assignments(assignments);
+            try {
```
raunaqmorarka: why the try/catch ?

core/trino-main/src/main/java/io/trino/execution/scheduler/NodeScheduler.java
```diff
@@ -101,10 +104,11 @@ public static List<InternalNode> selectExactNodes(NodeMap nodeMap, List<HostAddr
...
         for (HostAddress host : hosts) {
-            nodeMap.getNodesByHostAndPort().get(host).stream()
-                    .filter(node -> includeCoordinator || !coordinatorIds.contains(node.getNodeIdentifier()))
-                    .forEach(chosen::add);
-
+            for (InternalNode node : nodeMap.getNodesByHostAndPort().get(host)) {
+                if (includeCoordinator || !coordinatorIds.contains(node.getNodeIdentifier())) {
+                    chosen.add(node);
+                }
+            }
```
raunaqmorarka: why is this change needed ?

core/trino-main/src/main/java/io/trino/execution/scheduler/NodeScheduler.java
```diff
@@ -116,9 +120,11 @@ public static List<InternalNode> selectExactNodes(NodeMap nodeMap, List<HostAddr
...
                 }
 
-                nodeMap.getNodesByHost().get(address).stream()
-                        .filter(node -> includeCoordinator || !coordinatorIds.contains(node.getNodeIdentifier()))
-                        .forEach(chosen::add);
+                for (InternalNode node : nodeMap.getNodesByHost().get(address)) {
+                    if (includeCoordinator || !coordinatorIds.contains(node.getNodeIdentifier())) {
+                        chosen.add(node);
+                    }
+                }
```
raunaqmorarka: why is this change needed ?

core/trino-main/src/main/java/io/trino/execution/scheduler/NodeScheduler.java
```diff
@@ -63,11 +66,14 @@ public NodeSelector createNodeSelector(Session session)
         return nodeSelectorFactory.createNodeSelector(requireNonNull(session, "session is null"));
     }
 
-    public static List<InternalNode> getAllNodes(NodeMap nodeMap, boolean includeCoordinator)
+    public static Set<InternalNode> getAllNodes(NodeMap nodeMap, boolean includeCoordinator)
```
raunaqmorarka: NodeScheduler changes look like they belong to a separate PR, its not related to the planner commits up to this point

raunaqmorarka (commented):
Please try to avoid stacking 20+ unrelated commits in one PR and try to organize PRs by some common theme

core/trino-main/src/main/java/io/trino/sql/planner/plan/Assignments.java
wendigo: Leftover

core/trino-main/src/main/java/io/trino/sql/planner/plan/Assignments.java
wendigo: Dropped this commit

core/trino-main/src/main/java/io/trino/execution/scheduler/NodeScheduler.java
wendigo: Dropped this commit

core/trino-main/src/main/java/io/trino/execution/scheduler/NodeScheduler.java
wendigo: Dropped this commit

lib/trino-matching/src/main/java/io/trino/matching/CapturePattern.java
wendigo: We convert stream to the iterator in the usage site anyway so I don't think it is a right abstraction here. I'm not fan of streams either :)

core/trino-main/src/main/java/io/trino/sql/planner/iterative/Memo.java
wendigo: Presized builders are not growing when you add new elements above the initial capacity

core/trino-main/src/main/java/io/trino/sql/planner/iterative/Lookup.java
wendigo: Dropped commit

core/trino-main/src/main/java/io/trino/execution/scheduler/NodeScheduler.java
wendigo: Dropped.

wendigo:
I've dropped controversial API changes around `Rule<T>` and `Pattern<T>`

PR #27171, opened 2025-10-30
Require JDK 25 to run Trino

wendigo:
## Release notes

```markdown
## General
* Require JDK 25 to build and run Trino ({issue}`issuenumber`)
```


core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
```diff
@@ -106,7 +106,7 @@ else if ("Mac OS X".equals(osName)) {
 
     private static void verifyJavaVersion()
     {
-        Version required = Version.parse("24.0.1");
+        Version required = Version.parse("25.0.1");
```
findepi: Is there something specific about 25.0.1, that we don't want to allow "any 25"?

wendigo: We've been following this pattern for some time - upgrade to X, wait for X.0.1, require X.0.1 and switch language level to X. There is nothing specific for 25.0.1 required.

PR #27175, opened 2025-10-31
Planning: Rule to push projection through join into TableScan

bvolpato:
> [!WARNING]
> This is still work in progress, working on some test cases.

## Description

A new rule `PushProjectionThroughJoinIntoTableScan` was introduced to push projections that appear above a join down to the table scans on either side of the join. This optimization is particularly beneficial for cross-connector joins where the join itself cannot be pushed down to the connectors.

The rule applies when:
- All projection expressions are deterministic.
- Each projection expression references columns from only one side of the join.
- For outer joins, projections on the non-preserved side are not pushed to maintain correctness.

This transformation creates new `ProjectNode`s on each side of the join, potentially followed by the original `ProjectNode` if some expressions could not be pushed down. The rule also ensures that symbols required by the join criteria and filter, as well as the original project's output symbols, are preserved through identity projections on the respective sides.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/test/java/io/trino/sql/planner/iterative/rule/TestPushProjectionThroughJoinIntoTableScan.java
+++ b/core/trino-main/src/test/java/io/trino/sql/planner/iterative/rule/TestPushProjectionThroughJoinIntoTableScan.java
@@ -23,8 +23,6 @@ import io.trino.plugin.tpch.TpchColumnHandle;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.function.OperatorType;
-import io.trino.spi.type.RowType;
-import io.trino.spi.type.Type;
 import io.trino.sql.ir.Call;
 import io.trino.sql.ir.Reference;
 import io.trino.sql.planner.Symbol;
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/PushProjectionThroughJoinIntoTableScan.java
```diff
@@ -0,0 +1,246 @@
...
+ *         TableScan(a, ...)
+ *       Project(y := g(b), b)  -- pushed down
+ *         TableScan(b, ...)
+ * </pre>
+ *
+ * <p>The rule only applies when: - All projection expressions are deterministic - Each projection
+ * expression references columns from only one side of the join - For outer joins, projections on
+ * the non-preserved side are not pushed (to maintain correctness)
+ */
+public class PushProjectionThroughJoinIntoTableScan
```
Praveen2112, line 71: This is more of a push projection through join right ? TableScan would be a bit redundant here

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/PushProjectionThroughJoinIntoTableScan.java
```diff
@@ -0,0 +1,246 @@
...
+ *       TableScan(a, ...)
+ *       TableScan(b, ...)
+ *
+ * Into:
+ *   Project(x, y)  -- identity projections
+ *     Join(a = b)
+ *       Project(x := f(a), a)  -- pushed down
+ *         TableScan(a, ...)
+ *       Project(y := g(b), b)  -- pushed down
+ *         TableScan(b, ...)
```
Praveen2112, line 64: But if the `Join` filters the data then these projection would be redundant right ?

PR #27179, opened 2025-10-31
Some more cleanups related to ConnectorContextModule

ksobolew:
## Description

Two kinds of commits:

1. Remove duplicate bindings that are already provided by `ConnectorContextModule`, like #27178; this should have been part of the original PR, tbh
2. Start using `ConnectorContextModule` in a plugin; these were missed because they do the bindings inside plugin-specific modules instead of using a common pattern of a module binding them "manually" 

`ConnectorContextModule` was introduced in #27054.

## Release notes

## Summary by Sourcery

Standardize context injection by using ConnectorContextModule across connectors and eliminate duplicate manual bindings in plugin modules

Enhancements:
- Remove redundant manual bindings of TypeManager, NodeManager, and Node from plugin modules
- Inject ConnectorContextModule in all connector factories (memory, faker, tpcds, tpch, cassandra) for consistent context setup
- Simplify plugin-specific modules by relying on ConnectorContextModule for context dependencies

### DIFF
--- a/plugin/trino-faker/src/main/java/io/trino/plugin/faker/FakerConnectorFactory.java
+++ b/plugin/trino-faker/src/main/java/io/trino/plugin/faker/FakerConnectorFactory.java
@@ -16,6 +16,7 @@ package io.trino.plugin.faker;
 
 import com.google.inject.Injector;
 import io.airlift.bootstrap.Bootstrap;
+import io.trino.plugin.base.ConnectorContextModule;
 import io.trino.spi.connector.Connector;
 import io.trino.spi.connector.ConnectorContext;
 import io.trino.spi.connector.ConnectorFactory;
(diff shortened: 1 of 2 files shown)


ebyhr (approved):

ebyhr:
CI hit #27183

plugin/trino-faker/src/main/java/io/trino/plugin/faker/FakerConnectorFactory.java
```diff
@@ -48,7 +49,10 @@ public Connector create(String catalogName, Map<String, String> requiredConfig,
         requireNonNull(context, "context is null");
         checkStrictSpiVersionMatch(context, this);
 
-        Bootstrap app = new Bootstrap("io.trino.bootstrap.catalog." + catalogName, new FakerModule(context.getTypeManager()));
+        Bootstrap app = new Bootstrap(
+                "io.trino.bootstrap.catalog." + catalogName,
+                new ConnectorContextModule(catalogName, context),
```
chenjian2664, line 54: The `ConnectorContextModule` binds more than just `TypeManager` :) , but we just want to use `ConnectorContextModule` right?

ksobolew: Yes, we want to use it whenever at least one object from `ConnectorContext` is passes to the connector.

chenjian2664: Thanks

PR #27205, opened 2025-11-04
Do not modify iterated collections in MemoryMetadata#renameSchema

ksobolew:
## Description

When a schema is renamed, we need to update all the internal collections to remove tables in that schema and insert replacements with the new schema name. But the way it was implemented we did the replacement by removing the current entry and inserting a new entry *while iterating* each of the collections. And because these collections are not concurrent, this is very error-prone.

I guess renaming schema is not a very common operation :)

## Release notes

```markdown
## Section
* Fix concurrent modification exception in `RENAME SCHEMA` of the `memory` connector. ({issue}`issuenumber`)
```

## Summary by Sourcery

Prevent modifying iterated collections in renameSchema by safely updating schema-qualified metadata maps for tables, views, functions, and table IDs

Bug Fixes:
- Avoid ConcurrentModificationException by not altering maps during iteration in renameSchema

Enhancements:
- Refactor tableIds, views, and functions updates to use iterator.remove with batch putAll
- Simplify tables map update using Map.replaceAll to update schema names

### DIFF
--- a/plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
+++ b/plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
@@ -165,32 +165,40 @@ public class MemoryMetadata
         }
         schemas.add(target);
 
-        for (Map.Entry<SchemaTableName, Long> table : tableIds.entrySet()) {
+        Map<SchemaTableName, Long> newTableIds = new HashMap<>();
+        for (Iterator<Map.Entry<SchemaTableName, Long>> iterator = tableIds.entrySet().iterator(); iterator.hasNext(); ) {
+            Map.Entry<SchemaTableName, Long> table = iterator.next();
             if (table.getKey().getSchemaName().equals(source)) {
-                tableIds.remove(table.getKey());
-                tableIds.put(new SchemaTableName(target, table.getKey().getTableName()), table.getValue());
+                iterator.remove();
+                newTableIds.put(new SchemaTableName(target, table.getKey().getTableName()), table.getValue());
             }
         }
+        tableIds.putAll(newTableIds);
 
-        for (TableInfo table : tables.values()) {
-            if (table.schemaName().equals(source)) {
-                tables.put(table.id(), new TableInfo(table.id(), target, table.tableName(), table.columns(), false, table.dataFragments(), table.comment()));
-            }
-        }
+        tables.replaceAll((tableId, table) ->
+                table.schemaName().equals(source)
+                        ? new TableInfo(tableId, target, table.tableName(), table.columns(), table.truncated(), table.dataFragments(), table.comment())
+                        : table);
 
-        for (Map.Entry<SchemaTableName, ConnectorViewDefinition> view : views.entrySet()) {
+        Map<SchemaTableName, ConnectorViewDefinition> newViews = new HashMap<>();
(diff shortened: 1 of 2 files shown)


plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -165,32 +165,40 @@ public synchronized void renameSchema(ConnectorSession session, String source, S
         }
```
kokosing, line 165: Is it possible to have a test to reproduce issues?

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -165,32 +165,40 @@ public synchronized void renameSchema(ConnectorSession session, String source, S
...
             }
         }
+        tableIds.putAll(newTableIds);
 
-        for (TableInfo table : tables.values()) {
-            if (table.schemaName().equals(source)) {
-                tables.put(table.id(), new TableInfo(table.id(), target, table.tableName(), table.columns(), false, table.dataFragments(), table.comment()));
-            }
-        }
+        tables.replaceAll((tableId, table) ->
```
kokosing, line 178: do you need to replace if you remove above?

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -165,32 +165,40 @@ public synchronized void renameSchema(ConnectorSession session, String source, S
...
+        for (Iterator<Map.Entry<SchemaFunctionName, Map<String, LanguageFunction>>> iterator = functions.entrySet().iterator(); iterator.hasNext(); ) {
+            Map.Entry<SchemaFunctionName, Map<String, LanguageFunction>> function = iterator.next();
             if (function.getKey().getSchemaName().equals(source)) {
-                functions.remove(function.getKey());
-                functions.put(new SchemaFunctionName(target, function.getKey().getFunctionName()), function.getValue());
+                iterator.remove();
+                newFunctions.put(new SchemaFunctionName(target, function.getKey().getFunctionName()), function.getValue());
             }
         }
+        functions.putAll(newFunctions);
```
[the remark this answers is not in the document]
ksobolew: I assume that the RENAME SCHEMA operation already checked that the target schema does not exist

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
ksobolew: This collection is special - the key is the same for the original and replaced entries, so we can do the replacement in-place. The original loop simply overwrote the entry.

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
ksobolew: That's a good question...

ksobolew: Added test. It needs to have at least several tables in the schema (here: 10) to catch the `ConcurrentModificationException`.

kokosing:
Thank you!

PR #27213, opened 2025-11-05
Fix nullAllowed of Domain for identical operator in UtcConstraintExtractor

hqbhoho:
## Description
```
if (functionName.equals(IDENTICAL_OPERATOR_FUNCTION_NAME)) {
 return Optional.of(Domain.create(ValueSet.ofRanges(Range.range(type, startOfDate, true, startOfNextDate, false)), true));
}
```
Since `IDENTICAL_OPERATOR_FUNCTION_NAME` resolves to a range query (>=startOfDate AND <startOfNextDate), which is designed to match a specific date interval, the `nullAllowed` parameter would be set to `false`.

## Release notes

```markdown
## Iceberg/Delta Lake/MySQL/PostgreSQL
* Fix incorrect results due to incorrect pushdown of `IS NOT DISTINCT FROM`. ({issue}`27213`)
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/expression/StandardFunctions.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/expression/StandardFunctions.java
@@ -52,6 +52,9 @@ public final class StandardFunctions
     public static final FunctionName LESS_THAN_OR_EQUAL_OPERATOR_FUNCTION_NAME = new FunctionName("$less_than_or_equal");
     public static final FunctionName GREATER_THAN_OPERATOR_FUNCTION_NAME = new FunctionName("$greater_than");
     public static final FunctionName GREATER_THAN_OR_EQUAL_OPERATOR_FUNCTION_NAME = new FunctionName("$greater_than_or_equal");
+    /**
+     * $identical function is equivalent to the SQL operator "IS NOT DISTINCT FROM".
+     */
     public static final FunctionName IDENTICAL_OPERATOR_FUNCTION_NAME = new FunctionName("$identical");
 
     /**


lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/filter/UtcConstraintExtractor.java
```diff
@@ -248,7 +248,7 @@ private static Optional<Domain> createDomain(FunctionName functionName, Type typ
             return Optional.of(Domain.create(ValueSet.ofRanges(Range.greaterThanOrEqual(type, startOfDate)), false));
         }
         if (functionName.equals(IDENTICAL_OPERATOR_FUNCTION_NAME)) {
-            return Optional.of(Domain.create(ValueSet.ofRanges(Range.range(type, startOfDate, true, startOfNextDate, false)), true));
+            return Optional.of(Domain.create(ValueSet.ofRanges(Range.range(type, startOfDate, true, startOfNextDate, false)), false));
```
findepi, line 251: Would be good to add a jdoc to IDENTICAL_OPERATOR_FUNCTION_NAME documenting semantics (like SQL's NOT DISTINCT FROM, i assume)

hqbhoho: Yes,`$identical` is equal to `IS NOT DISTINCT FROM` for SQL.
```
IDENTICAL(IDENTICAL_OPERATOR_FUNCTION_NAME, "IS NOT DISTINCT FROM"),
```
 https://github.com/trinodb/trino/blob/3f97a0f867df966de09c93004af5161cec7716db/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/expression/ComparisonOperator.java#L41C71-L41C72

findepi: it would be great to document it where `io.trino.spi.expression.StandardFunctions#IDENTICAL_OPERATOR_FUNCTION_NAME` is defined.

hqbhoho: I have added a jdoc like:
```$identical function is equivalent to the SQL operator "IS NOT DISTINCT FROM".```

**Comment by findepi** [comment]
Merged, thanks!

**Comment by chenjian2664** [comment]
@findepi @hqbhoho I think this PR probably deserves a release note, could you check if we should add one?

**Comment by hqbhoho** [comment]
> @findepi @hqbhoho I think this PR probably deserves a release note, could you check if we should add one?

@chenjian2664 @findepi PTAL, I have added a release note.

**Comment by findepi** [comment]
I don't think it should cause incorrect query results.
I thought it is a small perf improvement in predicate pushdown, allowing us to filter better. 

Please clarify.

Also, if this did indeed cause incorrect query results, would you mind adding a regression test that would show that correctness issue in plain sight (if applied before the fix was applied)?

**Comment by hqbhoho** [comment]
@findepi  I have added a test to show  incorrect query results, https://github.com/trinodb/trino/pull/27242

**Comment by findepi** [comment]
@chenjian2664 i suggest following RN

```markdown
* Fix incorrect results due to incorrect pushdown of `IS [NOT] DISTINCT FROM`.  
```

this applies to Iceberg, Delta, PostgreSQL, MySQL

hqbhoho:
> @chenjian2664 i suggest following RN
> 
> ```
> * Fix incorrect results due to incorrect pushdown of `IS [NOT] DISTINCT FROM`.  
> ```
> 
> this applies to Iceberg, Delta, PostgreSQL, MySQL

I think maybe only `IS NOT DISTINCT FROM` will query incorrect results,`IS DISTINCT FROM` is equal to `$not($identical())` will not pushdown in `UtcConstraintExtractor`.

chenjian2664:
Thanks both!

PR #27242, opened 2025-11-07
Add regression test for IS NOT DISTINCT FROM pushdown correctness

hqbhoho:
## Description
Add regression test for IS NOT DISTINCT FROM pushdown in Iceberg/Delta Lake

- fix merged in https://github.com/trinodb/trino/pull/27213

### DIFF
--- a/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
+++ b/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
@@ -815,6 +815,13 @@ public class TestDeltaLakeConnectorTest
                 .isFullyPushedDown()
                 .returnsEmptyResult();
 
+        assertThat(query("SELECT * FROM " + tableName + " WHERE year(part) IS DISTINCT FROM 2006"))
+                .isNotFullyPushedDown(FilterNode.class);
+
+        assertThat(query("SELECT * FROM " + tableName + " WHERE year(part) IS NOT DISTINCT FROM 2006"))
+                .isFullyPushedDown()
+                .returnsEmptyResult();
+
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -3286,6 +3286,10 @@ else if (format == AVRO) {
         assertThat(query("SELECT * FROM test_year_transform_timestamptz WHERE day_of_week(d) = 2 AND b % 7 = 3"))
                 .matches("VALUES (TIMESTAMP '2015-09-15 14:21:02.345678 UTC', 10)");
 
+        // test IS NOT DISTICNT FROM pushwown
+        assertThat(query("SELECT * FROM test_year_transform_timestamptz WHERE year(d) IS NOT DISTINCT FROM 2015"))
+                .matches("VALUES (TIMESTAMP '2015-05-15 13:05:01.234567 UTC', 9), (TIMESTAMP '2015-09-15 14:21:02.345678 UTC', 10), (NULL, 101)");
```
findepi:
```suggestion
                assertThat(query("SELECT * FROM test_year_transform_timestamptz WHERE year(d) IS DISTINCT FROM 2015"))
                .isNotFullyPushedDown(FilterNode.class);

        assertThat(query("SELECT * FROM test_year_transform_timestamptz WHERE year(d) IS NOT DISTINCT FROM 2015"))
                .isFullyPushedDown();
```

findepi:
@hqbhoho thank you for standing me corrected on the nature of the bug fixed in https://github.com/trinodb/trino/pull/27213.
we definitely should merge the regression test. Please revert the temporary changes from the PR so it's mergeable

chenjian2664 (approved):

PR #27246, opened 2025-11-07
Correcting access-control.properties path for user's guidance

Nagaashokreddy:
Updating access-control.properties path for user's guidance

The path that is mentioned in the official trino documentation is not working, after looking into the code for this ranger plugin, it seems that the access-control.properties file should be in /etc/trino folder for the plugin to pick it up.

## Release notes

( x) This is not user-visible or is docs only, and no release notes are required.

```markdown
## Section
```


Nagaashokreddy:
This document has to be updated as well: https://trino.io/docs/current//security/ranger-access-control.html

ebyhr:
The path depends on deployment. For instance, `etc/access-control.properties` is correct for tarball, while `etc/trino/access-control.properties` is correct for Docker. 

The current documentation uses the former style. Let me reject this change because only changing Ranger example is confusing.

PR #27248, opened 2025-11-07
Enable cost-based join distribution selection for BROADCAST and PARTITIONED joinDistributionType

SongChujun:
```DetermineJoinDistributionType```‎ will try to place the smaller table on build side of a join only if the ```joinDistributionType``` is automatic. If the joinDistributionType is broadcast or partitioned . It will not try to place the smaller table on the build side of a join, but just use synthetic order specified by the query itself, which may generate a suboptimal plan where the larger table is placed on the RHS of the join.

Even if the ```joinDistributionType```  is broadcast or partitioned , we can still try to place the smaller table on the build side of the join instead of using the synthetic order.

## Release notes

```markdown
## Section
```


SongChujun:
Discussed with @martint offline that such a change removes the ability to override the join order and type chosen by the optimizer when such a choice is incorrect. So we don't want to introduce this optimization.

PR #27250, opened 2025-11-09
Refactor: Optimize regex usage by precompiling Patterns

### LINKED ISSUE #26914: Improve performance: eliminate non-precompiled Regex string patterns
There are still a few places in the "project production" java files where precompiled Regexps are not used.
For example, some calls to the String#replaceAll and String#replaceFirst methods.
Trino code examples:
```
name.replaceAll("[^A-Za-z0-9_$]", "_");
metadataLocation.replaceFirst("/metadata/[^/]*$", "");
```

We can avoid that by explicitly using a Pattern object.
More info:
https://stackoverflow.com/a/6262439 

The performance increase will probably be minimal, but... it's still something ;)

HarshMehta112:
Fixed #26914

Replace inline regex in String#replaceAll and String#replaceFirst calls with precompiled Pattern objects to avoid recompiling the regex on every call.

Slight performance improvement and cleaner, reusable code.

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/CountingMockConnector.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/CountingMockConnector.java
@@ -29,6 +29,7 @@ import io.trino.tracing.TracingConnectorMetadata;
 
 import java.util.Optional;
 import java.util.Set;
+import java.util.regex.Pattern;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 
@@ -43,6 +44,7 @@ import static java.util.stream.Collectors.joining;
 public class CountingMockConnector
         implements AutoCloseable
 {
+    private static final Pattern TRINO_PREFIX = Pattern.compile("^trino\\.");
     private final Object lock = new Object();
 
     private final Set<String> tablesTestSchema1 = IntStream.range(0, 1000)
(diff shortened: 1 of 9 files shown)


chenjian2664:
Please update commit message https://trino.io/development/process.html#pull-request-and-commit-guidelines

chenjian2664 (approved):

testing/trino-testing/src/main/java/io/trino/testing/CountingMockConnector.java
```diff
@@ -110,7 +112,7 @@ public Multiset<String> runTracing(Runnable runnable)
                         String attributes = span.getAttributes().asMap().entrySet().stream()
                                 .map(entry -> entry(entry.getKey().getKey(), entry.getValue()))
                                 .filter(entry -> !entry.getKey().equals("trino.catalog"))
-                                .map(entry -> "%s=%s".formatted(entry.getKey().replaceFirst("^trino\\.", ""), entry.getValue()))
+                                .map(entry -> "%s=%s".formatted(TRINO_PREFIX.matcher(entry.getKey()).replaceFirst(""), entry.getValue()))
```
findepi, line 115: Inline is actually more readable. For tests readability is more important than slight performance gains.

Issue #26567, opened 2025-09-06
Flaky TestQueuesDb.testUpdateSoftMemoryLimit


ebyhr:
https://github.com/trinodb/trino/actions/runs/17484719252/job/49661484113
```
Error:  io.trino.execution.resourcegroups.db.TestQueuesDb.testUpdateSoftMemoryLimit -- Time elapsed: 5.139 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 

expected: 4509715660L
 but was: 2254857830L
	at io.trino.execution.resourcegroups.db.TestQueuesDb.testUpdateSoftMemoryLimit(TestQueuesDb.java:398)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1394)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1970)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
```

ebyhr:
https://github.com/trinodb/trino/actions/runs/19351619729/job/55364480791?pr=27259

PR #27259, opened 2025-11-10
Fix deadlock failure when listing objects in SQL Server connector

### LINKED ISSUE #10846: SQL Server metadata listing may fail with "Transaction was deadlocked on lock resources with another process" in case of concurrent table creation/deletion

ebyhr:
## Description

Fixes #10846

## Release notes

```markdown
## SQL Server
* Fix potential failure when listing tables and columns. ({issue}`10846`)
```

### DIFF
--- a/plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
+++ b/plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
@@ -77,6 +77,8 @@ import io.trino.spi.connector.ConnectorTableMetadata;
 import io.trino.spi.connector.JoinCondition;
 import io.trino.spi.connector.JoinStatistics;
 import io.trino.spi.connector.JoinType;
+import io.trino.spi.connector.RelationColumnsMetadata;
+import io.trino.spi.connector.RelationCommentMetadata;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.TableNotFoundException;
 import io.trino.spi.expression.ConnectorExpression;
@@ -121,6 +123,7 @@ import java.time.format.DateTimeFormatter;
 import java.time.format.DateTimeFormatterBuilder;
 import java.util.Collection;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -426,6 +429,24 @@ public class SqlServerClient
     protected boolean isTableLockNeeded(ConnectorSession session)
     {
         return isBulkCopyForWrite(session) && isBulkCopyForWriteLockDestinationTable(session);
+    }
+
+    @Override
+    public List<RelationCommentMetadata> getAllTableComments(ConnectorSession session, Optional<String> schema)
+    {
+        return retryOnDeadlock(() -> super.getAllTableComments(session, schema), "error when getting all table comments for '%s'".formatted(schema));
+    }
+
+    @Override
+    public Iterator<RelationColumnsMetadata> getAllTableColumns(ConnectorSession session, Optional<String> schema)
+    {
+        return retryOnDeadlock(() -> super.getAllTableColumns(session, schema), "error when getting all table columns for '%s'".formatted(schema));
+    }
+
+    @Override
+    public List<JdbcColumnHandle> getColumns(ConnectorSession session, SchemaTableName schemaTableName, RemoteTableName remoteTableName)
+    {
+        return retryOnDeadlock(() -> super.getColumns(session, schemaTableName, remoteTableName), "error when getting columns for table '%s'".formatted(remoteTableName));
     }
 
     @Override
--- a/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
+++ b/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
@@ -276,28 +276,4 @@ public class TestSqlServerConnectorTest
                 .add("close]bracket")
                 .build();
     }
-
-    @Test
-    @Override
-    public void testSelectInformationSchemaTables()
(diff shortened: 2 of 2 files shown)


plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
```diff
@@ -1238,6 +1239,15 @@ public Connection getConnection(ConnectorSession session, JdbcOutputTableHandle
         return connection;
     }
 
+    @Override
+    public Connection openConnectionForListOperations(ConnectorSession session)
+            throws SQLException
+    {
+        Connection connection = connectionFactory.openConnection(session);
+        connection.setTransactionIsolation(TRANSACTION_READ_UNCOMMITTED);
```
findinpath: What is the downside of using `TRANSACTION_READ_UNCOMMITTED` for table listing?

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
```diff
@@ -277,27 +278,55 @@ private List<String> testTableNameTestData()
...
+
+    @RepeatedTest(100)
     @Override
     public void testSelectInformationSchemaTables()
     {
-        // Isolate this test to avoid problem described in https://github.com/trinodb/trino/issues/10846
-        executeExclusively(super::testSelectInformationSchemaTables);
+        super.testSelectInformationSchemaTables();
+        try {
+            super.testUpdateRowConcurrently();
```
findinpath: copy-paste issue or intentional?

findinpath:
> Use TRANSACTION_READ_UNCOMMITTED for listing objects in SQL Server
> This resolve "Transaction was deadlocked on lock resources with another process" when listing objects in SQL Server connector.

It would be worthwhile to explain what is the rationale for going with the TRANSACTION_READ_UNCOMMITTED isolation level.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
```diff
@@ -277,27 +278,55 @@ private List<String> testTableNameTestData()
...
     }
 
-    @Test
+    @RepeatedTest(100)
+    @Override
+    public void testShowTablesLike()
+    {
+        super.testShowTablesLike();
+        try {
+            super.testUpdateRowConcurrently();
```
findinpath: This test creates `test_update_row...` tables which will likely left unaffected the `SHOW TABLES LIKE 'or%'` query from `testShowTablesLike`.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
ebyhr: Intentional. The purpose is to test concurrent modification.

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
ebyhr: The downside is potential risk of dirty reads. For instance, the connector might see a table that will be rolled back (doesn't really exist).

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
ebyhr: That is fine because the condition isn't pushed-down to SQL Server. I confirmed the current test can reproduce a deadlock issue.

ebyhr:
CI hit #26567

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
findepi: I don't think "read uncommitted" is desired.
the sql server's snapshot  isolation was supposed to avoid deadlocks.
Besides, listing operations should be retried on retryable exceptions (as should be all other forms of network communication with external services), and those retries would likely solve the problem at hand

ebyhr: I was missing the existing `retryOnDeadlock` method. Updated.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
```diff
@@ -277,27 +278,55 @@ private List<String> testTableNameTestData()
                 .build();
     }
 
-    @Test
+    @RepeatedTest(100)
```
findepi: I understand this is only for development, and not meant for final merge?

ebyhr: Right, I just dropped the tmp commit.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
```diff
@@ -276,28 +276,4 @@ private List<String> testTableNameTestData()
...
-        executeExclusively(super::testSelectInformationSchemaColumns);
-    }
-
-    @Test
-    @Override
-    public void testBulkColumnListingOptions()
-    {
-        // Isolate this test to avoid problem described in https://github.com/trinodb/trino/issues/10846
-        executeExclusively(super::testBulkColumnListingOptions);
-    }
```
findepi, line 302: ❤️

PR #27263, opened 2025-11-10
Harden the retry policy for running Redshift SQL queries

findinpath:
## Description

Add additional logic to distinguish whether running a query on AWS Redshift  that ends in an exception is actually recoverable. 

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestingRedshiftServer.java
+++ b/plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestingRedshiftServer.java
@@ -39,7 +39,7 @@ public final class TestingRedshiftServer
     public static void executeInRedshiftWithRetry(String sql)
     {
         Failsafe.with(RetryPolicy.builder()
-                        .handleIf(e -> e.getMessage().matches(".* concurrent transaction .*"))
+                        .handleIf(TestingRedshiftServer::isExceptionRecoverable)
                         .withDelay(Duration.ofSeconds(10))
                         .withMaxRetries(3)
                         .build())
@@ -61,5 +61,14 @@ public final class TestingRedshiftServer
             throws E
     {
         return Jdbi.create(JDBC_URL, JDBC_USER, JDBC_PASSWORD).withHandle(callback);
+    }
+
+    private static boolean isExceptionRecoverable(Throwable exception)
+    {
+        return exception != null && (
+                exception.getMessage().matches(".* concurrent transaction .*")
+                        || exception.getMessage().matches(".*deadlock detected.*")
+                        || exception.getMessage().matches(".*could not open relation with OID.*")
+                        || exception.getMessage().matches(".*The connection attempt failed.*"));
     }
 }


plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestingRedshiftServer.java
```diff
@@ -19,8 +19,10 @@
 import org.jdbi.v3.core.HandleConsumer;
 import org.jdbi.v3.core.Jdbi;
 
+import java.net.SocketTimeoutException;
 import java.time.Duration;
 
+import static com.google.common.base.Throwables.getCausalChain;
```
anusudarsan: can you try to enable the TestRedshiftConnectorTest in CI (at least temporarily)

ebyhr:
/test-with-secrets sha=4dac5a091f9054c385dc6c8709d1c18d948f02df

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestingRedshiftServer.java
```diff
@@ -62,4 +64,15 @@ public static <T, E extends Exception> T executeWithRedshift(HandleCallback<T, E
...
+
+    private static boolean isExceptionRecoverable(Throwable exception)
+    {
+        return exception != null && (
+                exception.getMessage().matches(".* concurrent transaction .*")
+                        || exception.getMessage().matches(".*deadlock detected.*")
+                        || exception.getMessage().matches(".*could not open relation with OID.*")
+                        || exception.getMessage().matches(".*The connection attempt failed.*")
+                        || getCausalChain(exception).stream()
+                        .anyMatch(e -> e instanceof SocketTimeoutException));
```
ebyhr: If a SocketTimeoutException occurs, we don't know whether the remote query succeeded or failed, right? COPY command could lead to duplicate rows? 

https://github.com/trinodb/trino/blob/471434fafed75f8625de30e7d6cea49b7e2f82d5/plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/RedshiftQueryRunner.java#L181-L186

findinpath: Correct. I'll leave this out.

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestingRedshiftServer.java
```diff
@@ -62,4 +62,13 @@ public static <T, E extends Exception> T executeWithRedshift(HandleCallback<T, E
     {
         return Jdbi.create(JDBC_URL, JDBC_USER, JDBC_PASSWORD).withHandle(callback);
     }
+
+    private static boolean isExceptionRecoverable(Throwable exception)
+    {
+        return exception != null && (
+                exception.getMessage().matches(".* concurrent transaction .*")
+                        || exception.getMessage().matches(".*deadlock detected.*")
+                        || exception.getMessage().matches(".*could not open relation with OID.*")
```
findepi, line 71: I don't think this method should be ever invoked with `exception`  being `null`.
However, `exception.getMessage()` can legally be `null`.

Also, details such as  "deadlock detected" can often be buried in causal chain and not necessarily be surfaced as top level exception's message.

PR #27272, opened 2025-11-11
Use Block#copyPositions to avoid BlockBuilder#append per position

raunaqmorarka:
## Description
```
Benchmark                     Mode  Cnt   Score   Error  Units
BenchmarkArraySort.arraySort  avgt   20  23.173 ± 0.806  ns/op

Benchmark                     Mode  Cnt   Score   Error  Units
BenchmarkArraySort.arraySort  avgt   20  18.068 ± 0.456  ns/op
```

## Release notes

```markdown
## General
* Improve performance of array_sort function. ({issue}`27272`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/aggregation/DistinctWindowAccumulator.java
+++ b/core/trino-main/src/main/java/io/trino/operator/aggregation/DistinctWindowAccumulator.java
@@ -30,12 +30,14 @@ import io.trino.spi.block.ValueBlock;
 import io.trino.spi.function.WindowAccumulator;
 import io.trino.spi.function.WindowIndex;
 import io.trino.spi.type.Type;
+import it.unimi.dsi.fastutil.ints.IntArrayList;
 
 import java.util.List;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 import static io.trino.spi.type.BooleanType.BOOLEAN;
+import static java.lang.Math.min;
 import static java.util.Objects.requireNonNull;
 
 public class DistinctWindowAccumulator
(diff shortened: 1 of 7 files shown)


core/trino-main/src/main/java/io/trino/operator/aggregation/DistinctWindowAccumulator.java
```diff
@@ -141,18 +142,19 @@ private void indexCurrentPage(Page page)
             }
         }
         else {
-            PageBuilder filteredPageBuilder = new PageBuilder(argumentTypes);
+            IntArrayList selectedPositions = new IntArrayList();
```
pettyjamesm: Seems like we could pre-size this collection if we knew the groupId count before and after inserting the current page into the `hash` (requires adding a public accessor for `MarkDistinctHash#nextDistinctId`)

core/trino-main/src/main/java/io/trino/operator/aggregation/DistinctWindowAccumulator.java
```diff
@@ -141,18 +142,19 @@ private void indexCurrentPage(Page page)
...
-                    Block block = page.getBlock(channel);
-                    filteredPageBuilder.getBlockBuilder(channel).append(block.getUnderlyingValueBlock(), block.getUnderlyingValuePosition(position));
-                }
-                filteredPageBuilder.declarePosition();
             }
-            pagesIndex.addPage(filteredPageBuilder.build());
+
+            Block[] filteredBlocks = new Block[argumentChannels.size()];
+            for (int channel = 0; channel < argumentChannels.size(); channel++) {
+                Block block = page.getBlock(channel);
```
dain: I'd inline this

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayExceptFunction.java
```diff
@@ -63,14 +62,12 @@ public static Block except(
         for (int i = 0; i < rightPositionCount; i++) {
             set.add(rightArray, i);
         }
-        BlockBuilder distinctElementBlockBuilder = type.createBlockBuilder(null, leftPositionCount);
-        ValueBlock leftValueBlock = leftArray.getUnderlyingValueBlock();
+        IntArrayList distinctPositions = new IntArrayList();
```
dain: The default capacity for fastutils IntArrayList is 10.  You might want to consider defaulting this to be larger.  Maybe `min(32, leftPositionCount)`?  I just made up 32.  I'd use a min like this anyway, as there is no need to over allocate for tiny arrays.

dain (approved):

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayExceptFunction.java
raunaqmorarka: done

core/trino-main/src/main/java/io/trino/operator/aggregation/DistinctWindowAccumulator.java
raunaqmorarka: I'm unclear on whether that gives us overall distinct count or the distinct count in the input Page, I think we'd need the latter.
I'm going with the simpler pre-sizing of `min(128, positionCount)` to avoid having to expose that info. Can be revisited later if needed.

pettyjamesm: It gives you the overall distinct count, but you can get the distinct count for the page by subtracting the total count after adding the page from the distinct count before adding it. I've opened https://github.com/trinodb/trino/pull/27289 to do that.

PR #10264, opened 2021-12-10
Optimize null checking in DictionaryBlock

skrzypo987:

PR #27275, opened 2025-11-11
Simplify JSON encoding

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
@@ -354,9 +354,11 @@ public final class ArrayBlock
     @Override
     public boolean isNull(int position)
     {
+        if (!mayHaveNull()) {
+            return false;
+        }
         checkReadablePosition(this, position);
-        boolean[] valueIsNull = this.valueIsNull;
-        return valueIsNull != null && valueIsNull[position + arrayOffset];
+        return valueIsNull[position + arrayOffset];
     }
 
     @Override
--- a/core/trino-spi/src/main/java/io/trino/spi/block/Int128ArrayBlock.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/Int128ArrayBlock.java
@@ -155,8 +155,11 @@ public final class Int128ArrayBlock
     @Override
     public boolean isNull(int position)
     {
+        if (!mayHaveNull()) {
+            return false;
+        }
         checkReadablePosition(this, position);
-        return valueIsNull != null && valueIsNull[position + positionOffset];
+        return valueIsNull[position + positionOffset];
     }
 
     @Override
--- a/core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlock.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlock.java
@@ -158,8 +158,11 @@ public final class ByteArrayBlock
     @Override
     public boolean isNull(int position)
     {
+        if (!mayHaveNull()) {
+            return false;
+        }
         checkReadablePosition(this, position);
-        return valueIsNull != null && valueIsNull[position + arrayOffset];
+        return valueIsNull[position + arrayOffset];
     }
 
     @Override
--- a/core/trino-spi/src/main/java/io/trino/spi/block/DictionaryBlock.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/DictionaryBlock.java
@@ -312,7 +312,7 @@ public final class DictionaryBlock
     @Override
     public boolean isNull(int position)
     {
-        if (!mayHaveNull) {
+        if (!mayHaveNull()) {
             return false;
         }
         checkValidPosition(position, positionCount);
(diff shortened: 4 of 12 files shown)


core/trino-spi/src/main/java/io/trino/spi/block/Int128ArrayBlock.java
```diff
@@ -155,8 +155,11 @@ public boolean hasNull()
     @Override
     public boolean isNull(int position)
     {
+        if (valueIsNull == null) {
```
pettyjamesm: What's the motivation for lifting this up as its own condition here? Seems like the previous implementation handled this already, and the `checkReadablePosition` call should arguably happen with or without the `valueIsNull` array being present to catch incorrect logic in more scenarios.

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -130,7 +130,12 @@ public static void writePagesToJsonGenerator(Consumer<TrinoException> throwableC
                 for (int position = 0; position < page.getPositionCount(); position++) {
                     generator.writeStartArray();
                     for (int column = 0; column < typeEncoders.length; column++) {
-                        typeEncoders[column].encode(generator, blocks[column], position);
+                        if (blocks[column].isNull(position)) {
```
raunaqmorarka: Pulling out the null check is adding a megamorphic call site here.
The interface here should rather be `void encode(JsonGenerator generator, Block block)` to allow specialized loops for each block type.

core/trino-spi/src/main/java/io/trino/spi/block/Int128ArrayBlock.java
raunaqmorarka: @pettyjamesm please see the contract for `io.trino.spi.block.Block#isNull`
```
/**
     * Is the specified position null?
     *
     * @throws IllegalArgumentException if this position is not valid. The method may return false
     * without throwing exception when there are no nulls in the block, even if the position is invalid
     */
```
We made similar change in https://github.com/trinodb/trino/pull/10264

core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
```diff
@@ -354,9 +354,11 @@ public long getEstimatedDataSizeForStats(int position)
     @Override
     public boolean isNull(int position)
     {
```
raunaqmorarka, line 356: The current change is fine, but in general to avoid this problem
```
checkReadablePosition is called at least twice for non-null blocks,
when block position is checked for nullness and value is retrieved.
```
the nicer solution is to use `mayHaveNull` to avoid the `isNull` checks in the caller

core/trino-spi/src/main/java/io/trino/spi/block/Int128ArrayBlock.java
wendigo: Yeah i think this is wasteful if we check for position validity twice. Even in Slice we gave unsafe getters without bound checks (that works under the assumption that caller knows the bounds and moves only in that range)

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
wendigo: I can't have specialized loops per type as I'm encoding JSON by rows, not by columns

raunaqmorarka: okay, you're still better off keeping the existing code structure then

core/trino-spi/src/main/java/io/trino/spi/block/ArrayBlock.java
wendigo: Done

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -124,13 +124,20 @@ public static void writePagesToJsonGenerator(Consumer<TrinoException> throwableC
 
             for (Page page : pages) {
                 Block[] blocks = new Block[sourcePageChannels.length];
+                boolean[] mayHaveNull = new boolean[sourcePageChannels.length];
                 for (int i = 0; i < sourcePageChannels.length; i++) {
                     blocks[i] = page.getBlock(sourcePageChannels[i]);
+                    mayHaveNull[i] = blocks[i].mayHaveNull();
```
raunaqmorarka: Thinking some more about it, I think this `mayHaveNull` optimization may not be useful anymore after the improvement to `isNull` checks in the last commit.
I would reorder that to be the first commit and drop this commit entirely.

core/trino-spi/src/main/java/io/trino/spi/block/Int128ArrayBlock.java
pettyjamesm: I'm not opposed to making this change, but might point out that we _could_ consider making this assertions so that we don't have the runtime performance hit in production while still getting early detection of logical issues in tests. I know that's against the style guide we've used historically, but maybe worth reconsidering inside of these access paths that we use in performance sensitive loops.

Issue #27276, opened 2025-11-11
feature request: Functionality for filesystem caching to cache only selective tables.


RatulDawar:
This feature can improve performace in cases where there are some central widely used tables in most org. 
For e.g. : Like central dimension tables containing product data will be required by multiple other tables such as orders table, pricing table etc. 

This can help in cases where organisational context of table level query frequency, update frequency etc needs to be taken into account for caching. 
Based on principle that 80% of queries access 20% of data.  This can improve the cache hit frequency with much less disk requirements.

PR #27277, opened 2025-11-11
[WIP] feature: Make filesystem cache table aware

RatulDawar:
Related Issue : https://github.com/trinodb/trino/issues/27276 

Earlier there was no provision for selective table caching, all the queried files from all the tables competed for LRU eviction, this PR introduces ability to cache specific tables by passing the table names in description. 
Setting this property for a particular table will led to Filesystem caching being enabled for that table itself, users can then check their S3 API logs to optimize their API call cost / performance as they will have business context around what tables contribute most to the API cost. 
Example - A `fact` table can be a highly queried fast changing table, while a `dim` table will be equally highly queried slowly changing table, so it will be more beneficial to cache the `dim` table, rather than the `fact` table. 

Introduces `fs.cache.include-tables` configuration property that accepts a list of table patterns to include in caching
   - Supports wildcards: `schema.table`, `schema.*`, or `*` to cache all tables
   - Default value is `['*']` (cache all tables) to maintain backward compatibility

Testing 
- [ ]  Tested locally only using minio. Will further test on production workload too.

### DIFF
--- a/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
+++ b/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
@@ -13,8 +13,11 @@
  */
 package io.trino.filesystem.manager;
 
+import com.google.common.collect.ImmutableList;
 import io.airlift.configuration.Config;
 import io.airlift.configuration.ConfigDescription;
+
+import java.util.List;
 
 import static java.lang.System.getenv;
 
@@ -27,6 +30,7 @@ public class FileSystemConfig
     private boolean nativeGcsEnabled;
     private boolean nativeLocalEnabled;
     private boolean cacheEnabled;
+    private List<String> cacheIncludeTables = ImmutableList.of();
 
     // Enable leak detection if configured or if running in a CI environment
     private boolean trackingEnabled = getenv("CONTINUOUS_INTEGRATION") != null;
@@ -112,6 +116,19 @@ public class FileSystemConfig
     public FileSystemConfig setCacheEnabled(boolean enabled)
     {
         this.cacheEnabled = enabled;
+        return this;
+    }
+
+    public List<String> getCacheIncludeTables()
+    {
+        return cacheIncludeTables;
+    }
+
+    @Config("fs.cache.include-tables")
+    @ConfigDescription("List of tables to include in file system cache (schema.table format, supports wildcards like schema.* or *)")
+    public FileSystemConfig setCacheIncludeTables(List<String> tables)
+    {
+        this.cacheIncludeTables = ImmutableList.copyOf(tables);
         return this;
     }
 
--- a/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
+++ b/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
@@ -125,7 +125,7 @@ public class FileSystemConfig
     }
 
     @Config("fs.cache.include-tables")
-    @ConfigDescription("List of tables to include in file system cache (schema.table format, supports wildcards like schema.* or *)")
+    @ConfigDescription("List of tables to include in file system cache, use * to cache listings for all tables in all schemas")
     public FileSystemConfig setCacheIncludeTables(List<String> tables)
     {
         this.cacheIncludeTables = ImmutableList.copyOf(tables);
--- a/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
+++ b/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
@@ -16,6 +16,7 @@ package io.trino.filesystem.manager;
 import com.google.common.collect.ImmutableList;
 import io.airlift.configuration.Config;
 import io.airlift.configuration.ConfigDescription;
+import jakarta.validation.constraints.NotEmpty;
 
 import java.util.List;
 
@@ -30,7 +31,7 @@ public class FileSystemConfig
     private boolean nativeGcsEnabled;
     private boolean nativeLocalEnabled;
     private boolean cacheEnabled;
-    private List<String> cacheIncludeTables = ImmutableList.of();
+    private List<String> cacheIncludeTables = ImmutableList.of("*");
 
     // Enable leak detection if configured or if running in a CI environment
     private boolean trackingEnabled = getenv("CONTINUOUS_INTEGRATION") != null;
@@ -119,6 +120,7 @@ public class FileSystemConfig
         return this;
     }
 
+    @NotEmpty
     public List<String> getCacheIncludeTables()
     {
         return cacheIncludeTables;
--- a/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
+++ b/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
@@ -16,6 +16,7 @@ package io.trino.filesystem.manager;
 import com.google.common.collect.ImmutableList;
 import io.airlift.configuration.Config;
 import io.airlift.configuration.ConfigDescription;
+import jakarta.validation.constraints.AssertTrue;
 import jakarta.validation.constraints.NotEmpty;
 
 import java.util.List;
@@ -127,11 +128,17 @@ public class FileSystemConfig
     }
 
     @Config("fs.cache.include-tables")
-    @ConfigDescription("List of tables to include in file system cache, use * to cache listings for all tables in all schemas")
+    @ConfigDescription("List of tables to include in file system cache, use * to cache all tables in all schemas")
     public FileSystemConfig setCacheIncludeTables(List<String> tables)
     {
         this.cacheIncludeTables = ImmutableList.copyOf(tables);
         return this;
+    }
+
+    @AssertTrue(message = "fs.cache.enabled must be true when fs.cache.include-tables is explicitly configured")
+    public boolean isCacheIncludeTablesConfigValid()
+    {
+        return cacheEnabled || cacheIncludeTables.equals(ImmutableList.of("*"));
     }
 
     public boolean isTrackingEnabled()
--- a/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
+++ b/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
@@ -40,13 +40,7 @@ public class TableCachingPredicate
     public TableCachingPredicate(List<String> includeTables)
     {
         requireNonNull(includeTables, "includeTables is null");
-
-        if (includeTables.isEmpty()) {
-            this.predicate = _ -> true;
-        }
-        else {
-            this.predicate = matches(includeTables);
-        }
+        this.predicate = buildPredicate(includeTables);
     }
 
     public boolean test(SchemaTableName table)
@@ -54,7 +48,7 @@ public class TableCachingPredicate
         return predicate.test(table);
     }
 
-    private static Predicate<SchemaTableName> matches(List<String> tables)
+    private static Predicate<SchemaTableName> buildPredicate(List<String> tables)
     {
         return tables.stream()
                 .map(TableCachingPredicate::parseTableName)
--- a/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
+++ b/lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
@@ -13,6 +13,7 @@
  */
 package io.trino.filesystem.manager;
 
+import com.google.inject.Inject;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.SchemaTablePrefix;
 
@@ -29,6 +30,12 @@ import static java.util.Objects.requireNonNull;
 public class TableCachingPredicate
 {
     private final Predicate<SchemaTableName> predicate;
+
+    @Inject
+    public TableCachingPredicate(FileSystemConfig config)
+    {
+        this(config.getCacheIncludeTables());
+    }
 
     public TableCachingPredicate(List<String> includeTables)
     {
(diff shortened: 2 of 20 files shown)


lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
```diff
@@ -115,6 +119,19 @@ public FileSystemConfig setCacheEnabled(boolean enabled)
         return this;
     }
 
+    public List<String> getCacheIncludeTables()
+    {
+        return cacheIncludeTables;
+    }
+
+    @Config("fs.cache.include-tables")
+    @ConfigDescription("List of tables to include in file system cache (schema.table format, supports wildcards like schema.* or *)")
```
RotRotAl: The wording is a bit off,
```suggestion
    @ConfigDescription("List of tables to include in file system cache, 
use * to cache listings for all tables in all schemas
```

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.List;
+import java.util.function.Predicate;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Predicate to determine if a table should be cached based on configured include list.
+ * Supports wildcards: schema.table, schema.*, or *
```
RotRotAl: use * to cache listings for all tables in all schemas

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CacheFileSystem.java
```diff
@@ -35,30 +35,37 @@ public final class CacheFileSystem
     private final TrinoFileSystem delegate;
     private final TrinoFileSystemCache cache;
     private final CacheKeyProvider keyProvider;
+    private final boolean cachingEnabled;
```
RotRotAl, line 38: is it necessary?

RotRotAl (changes requested):
Hi please fix your code (5 test shave failed), and submit signed cla as the bot commented.

RotRotAl:
@RatulDawar have you tried adding this feature to hive connector as well? it will be very helpful

ratuldawar11:
> Hi please fix your code (5 test shave failed), and submit signed cla as the bot commented.

Yes have already submitted cla 3-4 days back. Also still working on this PR. It's a WIP.

ratuldawar11:
> @RatulDawar have you tried adding this feature to hive connector as well? it will be very helpful

Sure will add that too in the same PR.

RotRotAl:
Hi is there something new?

RatulDawar:
> Hi is there something new?

Hey resolving the comments in fews hours, was waiting for CLA to be signed, it just got signed today.

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CacheFileSystem.java
RatulDawar: I wanted to keep the cache file system table context free like it currently is.

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemConfig.java
```diff
@@ -115,6 +119,19 @@ public FileSystemConfig setCacheEnabled(boolean enabled)
         return this;
     }
 
+    public List<String> getCacheIncludeTables()
+    {
+        return cacheIncludeTables;
+    }
+
+    @Config("fs.cache.include-tables")
+    @ConfigDescription("List of tables to include in file system cache, use * to cache listings for all tables in all schemas")
```
findepi: if `*` means "cache all" and that's the old behavior, please initialize `cacheIncludeTables` to `["*"]` 

caching configure with `fs.cache.include-tables=` is an empty list and should be rejected as invalid configuration (add a bean validator)

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        this(config.getCacheIncludeTables());
+    }
+
+    public TableCachingPredicate(List<String> includeTables)
+    {
+        requireNonNull(includeTables, "includeTables is null");
+
+        if (includeTables.isEmpty()) {
+            this.predicate = _ -> true;
```
findepi: remove.

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+            this.predicate = matches(includeTables);
+        }
+    }
+
+    public boolean test(SchemaTableName table)
+    {
+        return predicate.test(table);
+    }
+
+    private static Predicate<SchemaTableName> matches(List<String> tables)
```
findepi: matches -> buildPredicate

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return tables.stream()
+                .map(TableCachingPredicate::parseTableName)
+                .map(prefix -> (Predicate<SchemaTableName>) prefix::matches)
+                .reduce(Predicate::or)
```
findepi, line 56: Looks fancy, but it would be better to just collect parsed table names on a field
and implement predicate with explicit for loop over them.

also, if any entry is `*`, then other elements still should be parsed (validated), but don't need to be remembered

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystemFactory.java
```diff
@@ -24,4 +24,14 @@ default TrinoFileSystem create(ConnectorSession session)
...
+
+    default TrinoFileSystem create(ConnectorSession session, boolean cachingEnabled)
+    {
+        return create(session);
+    }
+
+    default TrinoFileSystem create(ConnectorIdentity identity, boolean cachingEnabled)
+    {
+        return create(identity);
+    }
```
findepi, line 46: i generally like the idea that the FS calling code can knowingly bypass caching layer. Today this is handled by connector's CacheKeyProvider having knowledge of other components internals, eg
https://github.com/trinodb/trino/blob/99b456abb751198d763277c6b062d8d1764987b8/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/cache/DeltaLakeCacheKeyProvider.java#L52-L54
https://github.com/trinodb/trino/blob/2cb497d1f8e43d0fb7fd771179b9de48edcb48c4/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/cache/IcebergCacheKeyProvider.java#L28-L31

I am not sure these new functions are the best way to do it.
at least for FileMetastore i would try to simply inject the uncached underlying FS via Guice, or maybe have it configurable for FileMetastore separately (eg read-only use-cases on static data sets are OK to cache)

cc @electrum @raunaqmorarka

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
```diff
@@ -68,4 +69,30 @@ public TrinoFileSystem create(ConnectorIdentity identity, Map<String, String> fi
...
+                    .withGroups(identity.getGroups())
+                    .withPrincipal(identity.getPrincipal())
+                    .withEnabledSystemRoles(identity.getEnabledSystemRoles())
+                    .withConnectorRole(identity.getConnectorRole())
+                    .withExtraCredentials(ImmutableMap.<String, String>builder()
+                            .put(EXTRA_CREDENTIALS_ACCESS_KEY_PROPERTY, fileIoProperties.get(VENDED_S3_ACCESS_KEY))
+                            .put(EXTRA_CREDENTIALS_SECRET_KEY_PROPERTY, fileIoProperties.get(VENDED_S3_SECRET_KEY))
+                            .put(EXTRA_CREDENTIALS_SESSION_TOKEN_PROPERTY, fileIoProperties.get(VENDED_S3_SESSION_TOKEN))
+                            .buildOrThrow())
+                    .build();
```
findepi, line 79: That looks like a copy of the method above. Avoid copying (duplicating) this logic.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileSystemFactory.java
```diff
@@ -14,11 +14,25 @@
...
+
+    /**
+     * Create a file system with explicit caching control.
+     *
+     * @param session the connector session
+     * @param fileIoProperties file I/O properties
+     * @param cachingEnabled whether file system caching should be enabled
+     * @return a TrinoFileSystem instance
+     */
+    default TrinoFileSystem create(ConnectorSession session, Map<String, String> fileIoProperties, boolean cachingEnabled)
```
findepi, line 34: the problem is that `cachingEnabled = true` does not imply there will be any caching
it depends what the backing `TrinoFileSystemFactory` is. i.e. caching needs to be allowed on the FS layer as well

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Predicate to determine if a table should be cached based on configured include list.
+ * Supports wildcards: schema.table, schema.*, or use * to cache listings for all tables in all schemas
+ */
+public class TableCachingPredicate
+{
+    private final Predicate<SchemaTableName> predicate;
+
+    @Inject
+    public TableCachingPredicate(FileSystemConfig config)
```
findepi, line 35: This looks like part of trino-filesystem-manager caching layer, but it's used only in Iceberg.

For FS caching it would be more natural to cache based on locations, but i can see how user-unfriendly this might be and that cache control by  table names is reasonable.

This currently is a connector-specific cache control and if implemented as such, it should be configured as such. see `delta.fs.cache.disable-transaction-log-caching` for example of pre-existing connector-specific cache control.

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystemFactory.java
findepi: if we want to go this direction with new methods in TFSF, please open a separate PR adding those methods.
We can add usage of this methods in FileMetastore (controlled by configuration). This will be a simpler prep change & will ensure we have agreement on the API.

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/TableCachingPredicate.java
RatulDawar: @findepi  I think we can extended it for all importers of trino-file-system manager as this doesn't require any connector specific handling. Adding this specific to connectos seems to add more complexity than implementing it connector agonistic.

RatulDawar: Let me confirm if this is possible or not

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystemFactory.java
RatulDawar: @findepi I think these API's make this bypassing more robust both the Guice binding and api's with caching flag seems better approches than the current cache key provider with context internal file paths. 
I would prefer API changes given that it is more flexible for possible future requirements. Open to discussion on this... !

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileSystemFactory.java
RatulDawar: Seems fair, I think then it's make it virtually impossible to use apis with caching flags. 
Should we go guice dependencies ? That seems fine over handling this in cachekeyprovider.

RoeyoOgen:
@RatulDawar any update?

RatulDawar:
> @RatulDawar any update?

I have addressed all the comments need to start a discussion on how the API's should be formed here, will add my pointers and do that in this week itself, will again start working on this actively

RoeyoOgen:
@RatulDawar lmk on slack when a review is needed :)

RatulDawar:
> @RatulDawar lmk on slack when a review is needed :)

Tagging you in the relevant slack thread

fishlinghu:
Sorry I am a bit late to the discussion. 

Instead of putting the inclusion logic directly into the TrinoFileSystem or Connector, have we consider some plugin-based approach by defining a simple SPI (e.g., CacheParticipationPolicy)?

The default implementation can be the ConfigBasedCachePolicy (which does exactly what this PR proposes—reads from fs.cache.include-tables).

However, having the interface would allow us to implement dynamic policies later, such as:
- Fetching rules from a dynamic config source (ZooKeeper/Etcd/File) to change caching without restarts.
- Integrating with external cost-monitoring services.

This makes the feature extensible without over-complicating the initial implementation.

RatulDawar:
> Sorry I am a bit late to the discussion. 
> 
> 
> 
> Instead of putting the inclusion logic directly into the TrinoFileSystem or Connector, have we consider some plugin-based approach by defining a simple SPI (e.g., CacheParticipationPolicy)?
> 
> 
> 
> The default implementation can be the ConfigBasedCachePolicy (which does exactly what this PR proposes—reads from fs.cache.include-tables).
> 
> 
> 
> However, having the interface would allow us to implement dynamic policies later, such as:
> 
> - Fetching rules from a dynamic config source (ZooKeeper/Etcd/File) to change caching without restarts.
> 
> - Integrating with external cost-monitoring services.
> 
> 
> 
> This makes the feature extensible without over-complicating the initial implementation.

Feels a bit overkill to me

RatulDawar:
Hi @electrum I see that you have reopened this PR, if we are still interested in this I can take this forward. Had already resolved the comments here and on slack too.

PR #26354, opened 2025-08-07
Check if HNS is enabled for Azure storage account using Get Account Info

charlesjmorgan:
## Description
According to [this comment](https://github.com/Azure/azure-sdk-for-java/issues/38912#issuecomment-3160847541) from an Azure engineer, the version 12.31.0 of the azure-storage-blob dependency now supports OAuth for the [Get Account Info endpoint](https://learn.microsoft.com/en-us/rest/api/storageservices/get-account-information?tabs=shared-access-signatures#authorization). This PR updates the check for HNS on a storage account to use the Get Account Info endpoint, rather than the "hack" introduced in  #20897.

## Additional context and related issues
#20897
https://github.com/Azure/azure-sdk-for-java/issues/38912#issuecomment-3160847541

## Release notes

(X) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

Issue #23537, opened 2024-09-24
Flaky TestAzureFileSystemOAuthGen2Flat.testPreSignedUris


ebyhr:
https://github.com/trinodb/trino/actions/runs/11005419985/job/30558354477?pr=22632
```
Error:  io.trino.filesystem.azure.TestAzureFileSystemOAuthGen2Flat.testPreSignedUris -- Time elapsed: 3.457 s <<< ERROR!
java.io.IOException: 
Failed to retrieve, got response code: 403, body: ﻿<?xml version="1.0" encoding="utf-8"?><Error><Code>AuthenticationFailed</Code><Message>Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:f7ae768d-801e-0022-5c26-0ecbd1000000
Time:2024-09-24T02:09:00.5820573Z</Message><AuthenticationErrorDetail>Signature not valid in the specified time frame: Start [Tue, 24 Sep 2024 02:08:57 GMT] - Expiry [Tue, 24 Sep 2024 02:09:00 GMT] - Current [Tue, 24 Sep 2024 02:09:00 GMT]</AuthenticationErrorDetail></Error>
	at io.trino.filesystem.AbstractTestTrinoFileSystem.retrieveUri(AbstractTestTrinoFileSystem.java:1040)
	at io.trino.filesystem.AbstractTestTrinoFileSystem.testPreSignedUris(AbstractTestTrinoFileSystem.java:1018)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at java.base/java.util.concurrent.RecursiveAction.exec(RecursiveAction.java:194)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1458)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2034)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:189)
```

wendigo:
TTL was too short and it seems that test executed pretty slowly. I'll monitor it

ebyhr:
https://github.com/trinodb/trino/actions/runs/12667317946/job/35300517106

wendigo:
I haven't seen it for a while. Closing.

ebyhr:
[...]

PR #27278, opened 2025-11-11
Revert "Check if HNS is enabled for Azure storage account using Get Account Info" commit

findinpath:
## Description

https://github.com/trinodb/trino/pull/26354

After applying the above mentioned changes, it has been discovered that the hierarchical namespace check is sometimes failing. This happens when the Azure role "Storage Account Contributor" is missing on the storage account for which we're checking whether the hierarchical namespace is enabled.
Fallback to the original method of checking whether the root blob "/" exists because this method does not require additional permissions.

```
   "message":"Checking whether hierarchical namespace is enabled for the location abfs://xxx/.... failed",
      "cause":{
         "type":"com.azure.storage.blob.models.BlobStorageException",
         "message":"If you are using a StorageSharedKeyCredential, and the server returned an error message that says 'Signature did not match', you can compare the string to sign with the one generated by the SDK. To log the string to sign, pass in the context key value pair 'Azure-Storage-Log-String-To-Sign': true to the appropriate method call.\nIf you are using a SAS token, and the server returned an error message that says 'Signature did not match', you can compare the string to sign with the one generated by the SDK. To log the string to sign, pass in the context key value pair 'Azure-Storage-Log-String-To-Sign': true to the appropriate generateSas method call.\nPlease remember to disable 'Azure-Storage-Log-String-To-Sign' before going to production as this string can potentially contain PII.\nIf you are using a StorageSharedKeyCredential, and the server returned an error message that says 'Signature did not match', you can compare the string to sign with the one generated by the SDK. To log the string to sign, pass in the context key value pair 'Azure-Storage-Log-String-To-Sign': true to the appropriate method call.\nIf you are using a SAS token, and the server returned an error message that says 'Signature did not match', you can compare the string to sign with the one generated by the SDK. To log the string to sign, pass in the context key value pair 'Azure-Storage-Log-String-To-Sign': true to the appropriate generateSas method call.\nPlease remember to disable 'Azure-Storage-Log-String-To-Sign' before going to production as this string can potentially contain PII.\nStatus code 403, \"•<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>AuthorizationPermissionMismatch</Code><Message>This request is not authorized to perform this operation using this permission.\nRequestId:9dd14fea-f01e-0067-38ce-520758000000\nTime:2025-11-11T05:48:27.2395973Z</Message></Error>\"",
```
NOTE the `Status code 403` in the above message.

## Release notes

```markdown
## File System
* Harden hierarchical namespace check in Azure with root blob fallback check. ({issue}`issuenumber`)
```

### DIFF
--- a/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
+++ b/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
@@ -26,6 +26,7 @@ import com.azure.storage.blob.models.ListBlobsOptions;
 import com.azure.storage.blob.models.UserDelegationKey;
 import com.azure.storage.blob.sas.BlobSasPermission;
 import com.azure.storage.blob.sas.BlobServiceSasSignatureValues;
+import com.azure.storage.blob.specialized.BlockBlobClient;
 import com.azure.storage.common.sas.SasProtocol;
 import com.azure.storage.file.datalake.DataLakeDirectoryClient;
 import com.azure.storage.file.datalake.DataLakeFileClient;
@@ -602,10 +603,10 @@ public class AzureFileSystem
             throws IOException
     {
         try {
-            return createBlobContainerClient(location, Optional.empty())
-                    .getServiceClient()
-                    .getAccountInfo()
-                    .isHierarchicalNamespaceEnabled();
+            BlockBlobClient blockBlobClient = createBlobContainerClient(location, Optional.empty())
+                    .getBlobClient("/")
+                    .getBlockBlobClient();
+            return blockBlobClient.exists();
         }
         catch (RuntimeException e) {
             throw new IOException("Checking whether hierarchical namespace is enabled for the location %s failed".formatted(location), e);


lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
```diff
@@ -601,17 +605,45 @@ private Set<Location> listBlobDirectories(AzureLocation location)
...
+                catch (BlobStorageException e) {
+                    if (e.getStatusCode() == HTTP_FORBIDDEN) {
+                        isStorageAccountContributorRoleAssigned = false;
+                        return isHierarchicalNamespaceEnabledFallback(blobContainerClient);
+                    }
+                    throw e;
+                }
+            }
+            else {
+                return isHierarchicalNamespaceEnabledFallback(blobContainerClient);
```
wendigo: drop this else. not needed

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
```diff
@@ -601,17 +605,45 @@ private Set<Location> listBlobDirectories(AzureLocation location)
...
 
+    private static boolean isHierarchicalNamespaceEnabled(BlobContainerClient blobContainerClient)
+    {
+        return blobContainerClient
+                .getServiceClient()
+                .getAccountInfo()
+                .isHierarchicalNamespaceEnabled();
+    }
+
+    private static Boolean isHierarchicalNamespaceEnabledFallback(BlobContainerClient blobContainerClient)
```
wendigo: primitive boolean

ebyhr:
/test-with-secrets sha=60f2fa791d4216f46971102fc78aaafd46c71e26

ebyhr:
CI hit #23537

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
```diff
@@ -601,17 +605,43 @@ private Set<Location> listBlobDirectories(AzureLocation location)
     private boolean isHierarchicalNamespaceEnabled(AzureLocation location)
             throws IOException
     {
+        BlobContainerClient blobContainerClient = createBlobContainerClient(location, Optional.empty());
         try {
-            return createBlobContainerClient(location, Optional.empty())
-                    .getServiceClient()
-                    .getAccountInfo()
-                    .isHierarchicalNamespaceEnabled();
+            if (isStorageAccountContributorRoleAssigned) {
```
findinpath: I was skimming through the codebase and am not quite sure whether it makes sense to introduce any state here.

If a service principal does not have the "Storage Account Contributor" role for a storage account, it does not necessarily mean that it does not have the role for ANY storage account.

I'm reverting to the "fallback" implementation.

ebyhr:
/test-with-secrets sha=8feccda089633070b4924c0671f810186cf9fa3d

chenjian2664:
question about the release section "File System", it should be "General" ?

ebyhr:
@chenjian2664 We don't use "File System" section. Please put the entry in all affected connectors - Hive, Delta Lake, Iceberg, and Hudi.

chenjian2664:
@ebyhr Thanks!

Issue #27232, opened 2025-11-06
Flaky `TestHiveParquetEncryption.testEncryptedDictionaryPruningTwoColumns`


ebyhr:
https://github.com/trinodb/trino/actions/runs/19131497146/job/54673195949
```
2025-11-06T10:03:25.7150105Z [ERROR] io.trino.plugin.hive.crypto.TestHiveParquetEncryption.testEncryptedDictionaryPruningTwoColumns -- Time elapsed: 0.224 s <<< FAILURE!
2025-11-06T10:03:25.7151597Z java.lang.AssertionError: 
2025-11-06T10:03:25.7151949Z 
2025-11-06T10:03:25.7152162Z Expecting actual:
2025-11-06T10:03:25.7152629Z   0L
2025-11-06T10:03:25.7153363Z to be greater than:
2025-11-06T10:03:25.7153819Z   0L
2025-11-06T10:03:25.7154039Z 
2025-11-06T10:03:25.7155055Z 	at io.trino.plugin.hive.crypto.TestHiveParquetEncryption.writeTwoColumnEncryptedDictionaryFile(TestHiveParquetEncryption.java:294)
2025-11-06T10:03:25.7157133Z 	at io.trino.plugin.hive.crypto.TestHiveParquetEncryption.testEncryptedDictionaryPruningTwoColumns(TestHiveParquetEncryption.java:157)
```

cc: @sopel39 @raunaqmorarka

sopel39:
I will take a look. Please disable the test for the moment if it fails constantly

PR #27287, opened 2025-11-12
Remove flaky check.

sopel39:
dictionary_page_offset might not be reliable, even when dictionary encoding is present.

## Release notes

```markdown
## Section
```

Fixes https://github.com/trinodb/trino/issues/27232

### DIFF
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/crypto/TestHiveParquetEncryption.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/crypto/TestHiveParquetEncryption.java
@@ -291,8 +291,6 @@ public class TestHiveParquetEncryption
                     .filter(column -> column.getPath().equals(ID_PATH))
                     .findFirst().orElseThrow();
 
-            assertThat(ageChunk.getDictionaryPageOffset()).isGreaterThan(0);
-            assertThat(idChunk.getDictionaryPageOffset()).isGreaterThan(0);
             assertThat(ageChunk.getEncodings()).anyMatch(org.apache.parquet.column.Encoding::usesDictionary);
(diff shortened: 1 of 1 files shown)


sopel39:
https://stackoverflow.com/questions/55225108/why-is-dictionary-page-offset-0-for-plain-dictionary-encoding

plugin/trino-hive/src/test/java/io/trino/plugin/hive/crypto/TestHiveParquetEncryption.java
```diff
@@ -291,8 +291,6 @@ private static void writeTwoColumnEncryptedDictionaryFile(java.nio.file.Path pat
                     .filter(column -> column.getPath().equals(ID_PATH))
                     .findFirst().orElseThrow();
 
-            assertThat(ageChunk.getDictionaryPageOffset()).isGreaterThan(0);
-            assertThat(idChunk.getDictionaryPageOffset()).isGreaterThan(0);
```
raunaqmorarka, line 295: I'm not sure this will address the problem
If a column uses dictionary encoding, then it must write a dictionary page and that has to be a non-zero offset from beginning of file (because parquet file has a compulsory header of few bytes).
So if this was failing sometimes, then the next check for use of dictionary encoding should also fail, as the dictionary page offset can be zero only if there is no dictionary page at all.

sopel39: I cannot really reproduce the issue locally.

Take a look at https://stackoverflow.com/a/55226688.

> If a column uses dictionary encoding, then it must write a dictionary page and that has to be a non-zero offset from beginning of file (because parquet file has a compulsory header of few bytes).

Assuming that example is correct (where parquet-mr puts `0` even when dictionaries are present), does it mean we do not support dictionary offset correctly?

Issue #22306, opened 2024-06-06
ResultSetMetaData does not provide catalog, schema and table name for a column


prrvchr:
Hi all,

I am trying to continue integrating the Trino driver into [jdbcDriverOOo](https://prrvchr.github.io/jdbcDriverOOo/) which is a JDBC driver for LibreOffice Base.

I can create tables in Base if their name is given in lowercase characters.

I can display the contents of a table or view, but am unable to add, modify or delete any record since I do not have access to the name of the schema catalog and the table in the `ResultSetMetaData`.

Would it be possible to have access to such data?

prrvchr:
Hi mosabua,

> Why do you need a separate additional driver?

Because LibreOffice's native JDBC driver is not working properly, it is unable to edit (create, delete edit) a forward only `ResultSet`. It is written in C which makes it very slow (it seems that the C/Java conversion used requires many calls) and it is completely abandoned by the developers of LibreOffice.
jdbcDriverOOo is a type 4 driver written only in Java.

On the other hand if jdbcDriverOOo has access to the name of the catalog, the schema and the table then **any Trino `ResultSet` will become editable in Base**.

mosabua:
But a ResultSet is just that .. it is the result of a query that ran in Trino (or part of it and the query is still running). That result could come from one or more tables, from one or more schemas, and one of more catalogs. 

It is not some sort of handle to the underlying table or whatever.

If the JDBC drivers together allow you to send queries to Trino .. those queries can be insert, update, delete and so on. Any data you can see is also the result of queries .. just straight select queries from the tables or views.

On a side note .. happy to chat on Trino slack or even on a call some time as well. If you get this working it might be interesting to add to the website and maybe even do a Trino Community Broadcast interview and demo with you.

prrvchr:
I'm talking about methods:
- [java.sql.ResultSetMetaData.getCatalogName(int)](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F/java/sql/ResultSetMetaData.html#getCatalogName(int))
- [java.sql.ResultSetMetaData.getSchemaName(int)](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F/java/sql/ResultSetMetaData.html#getSchemaName(int))
- [java.sql.ResultSetMetaData.getTableName(int)](https://docs.oracle.com/javase%2F7%2Fdocs%2Fapi%2F/java/sql/ResultSetMetaData.html#getTableName(int))

[...]

PR #27321, opened 2025-11-14
Improvement of ResultSet metadata

prrvchr:
## Description

These changes add the following to the metadata of a ResultSet:

- the catalog name (if not provided, the default value is an empty string)
- the schema name (if not provided, the default value is an empty string)
- the table name (if not provided, the default value is an empty string)
- the column label (if it exists, otherwise it's the column name)
- is the column auto increment.
- is the column read only.

They are intended to address issue #22306

Furthermore, this allows a JDBC connector to manage a default value for each column in a table.

## Release notes

```markdown
## Section
* Fix some things. ({issue}`22306`)
```


client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoDatabaseMetaData.java
```diff
@@ -1429,10 +1429,10 @@ public ResultSet getClientInfoProperties()
         ClientTypeSignature varchar = new ClientTypeSignature("varchar", ImmutableList.of(ClientTypeSignatureParameter.ofLong(VARCHAR_UNBOUNDED_LENGTH)));
 
         ImmutableList.Builder<Column> columns = ImmutableList.builder();
-        columns.add(new Column("NAME", "varchar", varchar));
-        columns.add(new Column("MAX_LEN", "integer", new ClientTypeSignature("integer")));
-        columns.add(new Column("DEFAULT_VALUE", "varchar", varchar));
-        columns.add(new Column("DESCRIPTION", "varchar", varchar));
+        columns.add(new Column("", "", "", "NAME", "NAME", "varchar", varchar));
```
findepi: `""` is not the best way to model "no catalog"
Maybe that should be `Optional<String>`?

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -1010,9 +1010,9 @@ public void outputTaskFailed(TaskId taskId, Throwable failure)
         outputManager.outputTaskFailed(taskId, failure);
     }
 
-    public void setColumns(List<String> columnNames, List<Type> columnTypes)
+    public void setColumns(List<String> catalogNames, List<String> schemaNames, List<String> tableNames, List<String> columnNames, List<String> columnLabels, List<Type> columnTypes)
```
findepi: `List<info about each column>` would be preferred over 6 separate lists

client/trino-jdbc/src/test/java/io/trino/jdbc/TestTrinoDatabaseMetaData.java
```diff
@@ -284,7 +284,7 @@ public void testGetCatalogs()
 
                 ResultSetMetaData metadata = rs.getMetaData();
                 assertThat(metadata.getColumnCount()).isEqualTo(1);
-                assertThat(metadata.getColumnLabel(1)).isEqualTo("TABLE_CAT");
+                assertThat(metadata.getColumnName(1)).isEqualTo("TABLE_CAT");
```
findepi, line 293: in `TestTrinoDatabaseMetaData`, should i expect test coverage for the new info being presented to users? catalog, schema, ...

client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoDatabaseMetaData.java
prrvchr: > `""` is not the best way to model "no catalog" Maybe that should be `Optional<String>`?

I agree. And perhaps we should consider:

- Creating a new constructor in Column:
  ```
      public Column(String name, String type, ClientTypeSignature typeSignature)
      {
          this(Optional.empty(), Optional.empty(), Optional.empty(), name, name, type, typeSignature);
      }
  ```

  However, it seems we can only have one constructor with the `@JsonCreator` annotation, and I don't know how to find the calls to this constructor that uses and requires this annotation.

- Replacing the necessary additional information (i.e., catalog name, schema name, and table name) with the `io.trino.metadata.QualifiedObjectName` class. This information is actually extracted from this class. However, does this class support serialization and passing data via JSON? I'm not sure.

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
prrvchr: > `List<info about each column>` would be preferred over 6 separate lists

Creating a new record that groups all the information from a column is certainly preferable.

However, for the `ResultSet` to be easily updatable and insertable, the following information should also be known for each column:
- Whether the column is auto-incrementing (ie: `ResultSetMetaData.isAutoIncrement(int column)`.
- Whether the column is definitively writable (ie: `ResultSetMetaData.isDefinitelyWritable(int column)`.

If it is too complicated to integrate this additional information into the `ResultSetMetaData`, it must at least be available through the method `DatabaseMetaData.getColumns(String catalog, String schema, String table, String column)`.

In any case, to go further we must first answer the first question: do we replace the catalog, the schema and the table name with the `io.trino.metadata.QualifiedObjectName` class if possible of course?

client/trino-jdbc/src/test/java/io/trino/jdbc/TestTrinoDatabaseMetaData.java
prrvchr: Yes, indeed, it would be good to add these tests.

client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoDatabaseMetaData.java
findepi: > * it seems we can only have one constructor with the `@JsonCreator` annotation

correct

> and I don't know how to find the calls to this constructor that uses and requires this annotation.

`io.trino.client` classes are used in Trino protocol, between trino server and client such as JDBC


> * Replacing the necessary additional information (i.e., catalog name, schema name, and table name) with the `io.trino.metadata.QualifiedObjectName` class.

this class is not in `io.trino.client` and I don't think it's a good idea to move it there.
adding fields to `io.trino.client.Column`, as you did, looks preferred to me

client/trino-jdbc/src/test/java/io/trino/jdbc/TestTrinoDatabaseMetaData.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
prrvchr: A new class, [ColumnInfo](https://github.com/trinodb/trino/pull/27321/files#diff-42ebca3cc5090bbd77f9d9e4f0605b0b2769692919d04254d194bc7c4e63bbba) (or more accurately, record), has been added to handle this.

This new record allows you to store the catalog, schema, table, and label.

client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoDatabaseMetaData.java
prrvchr: Now the Column class has new constructors and the constructor used by JSON has been augmented only with parameters of type `Optional<String>`.

This ensures backward compatibility because JSON still manages to use the same constructor even without the new parameters. Apparently, it passes an `Optional.empty()` to the constructor if no parameters are present.

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.execution;
+
+import java.util.Optional;
+
+public record ColumnInfo(Optional<String> catalog, Optional<String> schema, Optional<String> table, String name, String label)
```
findepi: Can you please add some code comment explaining what is a label?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,39 @@
...
+{
+    public ColumnInfo(String name)
+    {
+        this(Optional.empty(), Optional.empty(), Optional.empty(), name, name);
```
findepi: This looks like the label could want to be optional too?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,39 @@
...
+
+    public ColumnInfo(String catalog, String schema, String table, String name)
+    {
+        this(Optional.of(catalog), Optional.of(schema), Optional.of(table), name, name);
+    }
+
+    public ColumnInfo(String catalog, String schema, String table, String name, String label)
+    {
+        this(Optional.of(catalog), Optional.of(schema), Optional.of(table), name, label);
+    }
```
findepi, line 157: We avoid multiple overloaded constructors unless there is a good reason to have them.

core/trino-main/src/main/java/io/trino/execution/QueryExecution.java
```diff
@@ -99,22 +99,27 @@ interface QueryExecutionFactory<T extends QueryExecution>
...
     class QueryOutputInfo
     {
-        private final List<String> columnNames;
+        private final List<ColumnInfo> columnInfos;
         private final List<Type> columnTypes;
         private final Queue<ExchangeInput> inputsQueue;
         private final boolean noMoreInputs;
 
-        public QueryOutputInfo(List<String> columnNames, List<Type> columnTypes, Queue<ExchangeInput> inputsQueue, boolean noMoreInputs)
+        public QueryOutputInfo(List<ColumnInfo> columnInfos, List<Type> columnTypes, Queue<ExchangeInput> inputsQueue, boolean noMoreInputs)
```
findepi, line 115: This now looks much better than 6 separate lists, thank you!

Should Type be part of ColumnInfo?

core/trino-main/src/main/java/io/trino/execution/QueryExecution.java
```diff
@@ -99,22 +99,27 @@ interface QueryExecutionFactory<T extends QueryExecution>
...
 
+        public List<ColumnInfo> getColumnInfos()
+        {
+            return columnInfos;
+        }
+
         public List<String> getColumnNames()
         {
-            return columnNames;
+            return columnInfos.stream().map(ColumnInfo::name).toList();
```
findepi: use `toImmutableList`. this prevents data copy when the returned value goes thru `ImmutableList.copyOf` at some later point (which very likely is the case)

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Are we to understand that you prefer to see only one ctor, or that you prefer that ctor not call each other more than once?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Ok, done.

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Ok done.

core/trino-main/src/main/java/io/trino/execution/QueryExecution.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/execution/QueryExecution.java
prrvchr: You're right, that's much better. But I prefer to proceed this way; it allows for gradual validation without completely breaking everything.
I tried searching for the type, but no, it seems there's a plan specifically for types.

prrvchr:
@findepi

Perhaps it would be a good idea to immediately add the missing information that will allow LibreOffice Base to efficiently update ResultSets, namely:

- ResultSetMetaData.isAutoIncrement(int column)

- ResultSetMetaData.isReadOnly(int column)

This will allow Base to recognize auto-incremented columns and allow me to know which ones have a calculated value.

prrvchr:
@findepi 
~~Here's the latest complete version. I hope it passes the tests. Thanks in advance for your corrections.~~

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
findepi: > Are we to understand that you prefer to see only one ctor,

this



> or that you prefer that ctor not call each other more than once?

calling other ctor is 👍

findepi:
> Perhaps it would be a good idea to immediately add the missing information that will allow LibreOffice Base to efficiently update ResultSets, namely:
> 
> * ResultSetMetaData.isAutoIncrement(int column)
> * ResultSetMetaData.isReadOnly(int column)

I am not ready for this topic. Would prefer for that to be separate PR.

prrvchr:
@findepi, Okay, I'm reverted and will propose a PR just for that.

prrvchr:
@findepi I've done everything you asked. I see that this PR isn't progressing anymore. What's the problem?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.execution;
+
+import java.util.Optional;
+
+// XXX: ColumnInfo allows the grouping of all the naming elements necessary to deliver complete ResultSetMetaData (see issue#22306).
```
findepi: Did you mean to address this TODO comment?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,26 @@
...
+public record ColumnInfo(Optional<String> catalog, Optional<String> schema, Optional<String> table, String name, Optional<String> label)
+{
+    // XXX: This constructor is used by the PlanBuilder.output() method and allows you to create ColumnInfo using only a column name.
```
findepi: Did you mean to address this TODO comment?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,26 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.execution;
+
+import java.util.Optional;
+
+// XXX: ColumnInfo allows the grouping of all the naming elements necessary to deliver complete ResultSetMetaData (see issue#22306).
+public record ColumnInfo(Optional<String> catalog, Optional<String> schema, Optional<String> table, String name, Optional<String> label)
```
findepi: Java record doesn't validate non-nullness of its components automatically. This needs to be done explicitly in each class.
see here for example
 https://github.com/trinodb/trino/blob/53cee48ae8b30368e8d8669786f000f44ed1f47c/core/trino-main/src/main/java/io/trino/json/ir/IrAbsMethod.java#L27

core/trino-main/src/main/java/io/trino/execution/QueryExecution.java
```diff
@@ -99,22 +100,27 @@ interface QueryExecutionFactory<T extends QueryExecution>
...
 
+        public List<ColumnInfo> getColumnInfos()
+        {
+            return columnInfos;
+        }
+
         public List<String> getColumnNames()
         {
-            return columnNames;
+            return columnInfos.stream().map(ColumnInfo::name).collect(toImmutableList());
```
findepi: This is executed every time `QueryInfo` is constructed.
This might be not a big deal, but i am not sure.
Can you move column name list extraction to `QueryOutputInfo` constructor?

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
```diff
@@ -75,10 +76,20 @@ public final class ProtocolUtil
...
+    {
+        return createColumn(Optional.empty(), Optional.empty(), Optional.empty(), name, Optional.of(name), type, supportsParametricDateTime);
+    }
+
+    public static Column createColumn(ColumnInfo columnInfo, Type type, boolean supportsParametricDateTime)
+    {
+        return createColumn(columnInfo.catalog(), columnInfo.schema(), columnInfo.table(), columnInfo.name(), columnInfo.label(), type, supportsParametricDateTime);
+    }
+
+    public static Column createColumn(Optional<String> catalog, Optional<String> schema, Optional<String> table, String name, Optional<String> label, Type type, boolean supportsParametricDateTime)
```
findepi: `private`

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
```diff
@@ -75,10 +76,20 @@ public final class ProtocolUtil
     private ProtocolUtil() {}
 
     public static Column createColumn(String name, Type type, boolean supportsParametricDateTime)
+    {
```
findepi, line 83: remove this one. it's not supposed to be used anymore

core/trino-main/src/main/java/io/trino/sql/planner/plan/OutputNode.java
```diff
@@ -18,36 +18,38 @@
...
     private final PlanNode source;
-    private final List<String> columnNames;
+    private final List<ColumnInfo> columnInfos;
     private final List<Symbol> outputs; // column name = symbol
 
     @JsonCreator
     public OutputNode(@JsonProperty("id") PlanNodeId id,
             @JsonProperty("source") PlanNode source,
-            @JsonProperty("columns") List<String> columnNames,
+            @JsonProperty("columns") List<ColumnInfo> columnInfos,
```
findepi: Would be nice, if we could keep the JSON plan representation intact. WDYT?

cc @raunaqmorarka @losipiuk

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
```diff
@@ -342,7 +343,7 @@ public PlanNode planStatement(Analysis analysis, Statement statement)
                 statement instanceof RefreshMaterializedView && analysis.isSkipMaterializedViewRefresh()) {
             Symbol symbol = symbolAllocator.newSymbol("rows", BIGINT);
             PlanNode source = new ValuesNode(idAllocator.getNextId(), ImmutableList.of(symbol), ImmutableList.of(new Row(ImmutableList.of(new Constant(BIGINT, 0L)))));
-            return new OutputNode(idAllocator.getNextId(), source, ImmutableList.of("rows"), ImmutableList.of(symbol));
+            return new OutputNode(idAllocator.getNextId(), source, ImmutableList.of(new ColumnInfo(Optional.of("catalogs"), Optional.of("schemas"), Optional.of("tables"), "rows", Optional.of("labels"))), ImmutableList.of(symbol));
```
findepi: "catalogs", "schemas", "tables", "labels" -- these all look like fake values.
did you mean to write Optional.empty here?

findepi: btw, it would be better not to have to change PlanNodes and LogicalPlanner, if possible.
The full column info (catalog, schema, table, name, labels) for the client protocol may come from analysis and doesn't have to be part of the plan.

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
```diff
@@ -899,21 +900,30 @@ private RelationPlan createMergePlan(Analysis analysis, Merge node)
...
         ImmutableList.Builder<Symbol> outputs = ImmutableList.builder();
-        ImmutableList.Builder<String> names = ImmutableList.builder();
+        ImmutableList.Builder<ColumnInfo> columns = ImmutableList.builder();
 
         int columnNumber = 0;
         RelationType outputDescriptor = analysis.getOutputDescriptor();
         for (Field field : outputDescriptor.getVisibleFields()) {
             String name = field.getName().orElse("_col" + columnNumber);
-            names.add(name);
+            String label = field.getOriginColumnName().orElse(name);
```
findepi: remove `.orElse(name)`, keep label optional here.
i am aware that somewhere else we also do this logic (`label.or(name)`), and especially because of that we should do same logic here.

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Your right, done.

core/trino-main/src/main/java/io/trino/execution/QueryExecution.java
prrvchr: Done.

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
prrvchr: Done.

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
prrvchr: I find it used in method [newDecoder(List<TypedColumn> types)](https://github.com/trinodb/trino/blob/fc170a76953941d77cddd879307d439ab9892fc7/core/trino-main/src/test/java/io/trino/server/protocol/TestJsonEncodingUtils.java#L575) in `TestJsonEncodingUtils` class?

core/trino-main/src/main/java/io/trino/sql/planner/plan/OutputNode.java
prrvchr: ???

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
prrvchr: Done.

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
prrvchr: Done.

core/trino-main/src/main/java/io/trino/sql/planner/plan/OutputNode.java
losipiuk: >Would be nice, if we could keep the JSON plan representation intact. WDYT?

Yeah - seems more natural to me to extend the information sent to the client with metadata based on analysis result on coordiantor.

prrvchr: @losipiuk Okay, this means that OutputNode only knows the list of column names.
But then how to extend the information sent to the client?

prrvchr: @findepi @losipiuk Here is a version that keep the representation of the JSON plan intact.

core/trino-main/src/main/java/io/trino/connector/informationschema/InformationSchemaTable.java
```diff
@@ -36,7 +36,9 @@ public enum InformationSchemaTable
             .column("column_name", createUnboundedVarcharType())
             .column("ordinal_position", BIGINT)
             .column("column_default", createUnboundedVarcharType())
+            .column("is_autoincrement", createUnboundedVarcharType())
             .column("is_nullable", createUnboundedVarcharType())
+            .column("is_readonly", createUnboundedVarcharType())
```
findepi, line 41: @martint PTAL

core/trino-spi/src/main/java/module-info.java
```diff
@@ -18,6 +18,7 @@
     requires jakarta.annotation;
     requires transitive slice;
     requires jdk.incubator.vector;
+    requires java.sql;
```
findepi: what is this needed for?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -305,7 +305,10 @@ protected List<JdbcColumnHandle> getColumns(ConnectorSession session, Connection
             Type type = toColumnMapping(session, connection, jdbcTypeHandle)
                     .orElseThrow(() -> new UnsupportedOperationException(format("Unsupported type: %s of column: %s", jdbcTypeHandle, name)))
                     .getType();
-            columns.add(new JdbcColumnHandle(name, jdbcTypeHandle, type));
+            boolean autoIncrement = metadata.isAutoIncrement(column);
+            boolean nullable = metadata.isNullable(column) != columnNoNulls;
+            boolean readOnly = metadata.isReadOnly(column);
```
findepi: `getColumns(ConnectorSession session, Connection connection, ResultSetMetaData metadata)` is used to describe query output. It should be fair to assume no autoincrements and read-only for these.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -335,14 +338,18 @@ public List<JdbcColumnHandle> getColumns(ConnectorSession session, SchemaTableNa
                 Optional<ColumnMapping> columnMapping = toColumnMapping(session, connection, typeHandle);
                 log.debug("Mapping data type of '%s' column '%s': %s mapped to %s", schemaTableName, columnName, typeHandle, columnMapping);
                 boolean nullable = (resultSet.getInt("NULLABLE") != columnNoNulls);
+                boolean autoIncrement = getBoolean(resultSet, "IS_AUTOINCREMENT", false);
+                boolean readOnly = getBoolean(resultSet, "IS_GENERATEDCOLUMN", false);
```
findepi: I like how you consistently ordered the triple (auto-increment, nullable, read-only), let's keep it ordered everywhere. 

BTW i would prefer the ordering: type, nullable, read-only, then auto-increment.
type and nullable are quite related. 
auto-increment matters only if column is not read-only.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -557,6 +569,16 @@ protected static Optional<Integer> getInteger(ResultSet resultSet, String column
...
     }
 
+    protected static boolean getBoolean(ResultSet resultSet, String columnLabel, boolean defaultValue)
+            throws SQLException
+    {
+        String value = resultSet.getString(columnLabel);
+        if (resultSet.wasNull()) {
+            return defaultValue;
+        }
+        return value.equals("YES");
```
findepi: This is not generic "get boolean from ResutlSet" method. It needs a different name.
Or, maybe we just let the caller do `"YES".equalls(rs.getString(...))` and not have this method.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
```diff
@@ -38,13 +38,20 @@ public final class JdbcColumnHandle
...
     private final Optional<String> comment;
 
     // All and only required fields
     public JdbcColumnHandle(String columnName, JdbcTypeHandle jdbcTypeHandle, Type columnType)
     {
-        this(columnName, jdbcTypeHandle, columnType, true, Optional.empty());
+        this(columnName, jdbcTypeHandle, columnType, false, true, false, Optional.empty());
+    }
+
+    public JdbcColumnHandle(String columnName, JdbcTypeHandle jdbcTypeHandle, Type columnType, boolean autoIncrement, boolean nullable, boolean readOnly)
```
findepi: `JdbcColumnHandle` intentionally has two constructors (minimal and full). 
Remove this one, it has only one call site anyway.

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -356,7 +356,7 @@ public synchronized MemoryOutputTableHandle beginCreateTable(ConnectorSession se
         for (int i = 0; i < tableMetadata.getColumns().size(); i++) {
             ColumnMetadata column = tableMetadata.getColumns().get(i);
             MemoryColumnHandle handle = new MemoryColumnHandle(i, column.getName(), column.getType());
-            columns.add(new ColumnInfo(handle, column.getDefaultValue(), column.isNullable(), Optional.ofNullable(column.getComment())));
+            columns.add(new ColumnInfo(handle, column.getDefaultValue(), column.isAutoIncrement(), column.isNullable(), column.isReadOnly(), Optional.ofNullable(column.getComment())));
```
findepi: Memory connector doesn't support auto-increments, nor read-only columns.
And it doesn't have to -- the engine will never ask it to create such columns.

It should be fine to remove changes here.

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
findepi: Yes, it's used only in one test. The test should be updated. 
it's better to update the test than to have 3 overloads of `createColumn` method here.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
prrvchr: My sorting rule is as follows:
- For reading the `ResultSet`: the order of the column index (even if the reading is done by column name).
- For passing parameters, the three Boolean properties are sorted alphabetically.

But of course, all this can be changed if necessary.

core/trino-spi/src/main/java/module-info.java
prrvchr: That's IntelliJ. Removed

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
prrvchr: Ok done

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
prrvchr: Ok done

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcColumnHandle.java
prrvchr: Ok done

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
prrvchr: Ok done

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
prrvchr: Thank you for this insightful comment. It's done

prrvchr:
Certain connectors pose problems when accessing the metadata of the columns present in SQL select command.
And it would seem that this problem arises when accessing column metadata relating to a view.

To access the column metadata the `io.trino.metadata.Metadata` interface is used:
- first the `getTableHandle()` method to obtain a `TableHandle`
- then the `getTableMetadata()` method to obtain a map (ie: key: column name, value: column metadata)

Some connectors provide the `TableHandler` but throw an exception when trying to access the metadata.
For these connectors, it was necessary to avoid returning a `TableHandler` (ie: `return null;`) in order to prevent the exception from being thrown (see: [DeltaLakeMetadata](https://github.com/prrvchr/trino/blob/01300f436271cf8b2d3fafb8f7661ea682020c35/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L713) and [MemoryMetadata](https://github.com/prrvchr/trino/blob/01300f436271cf8b2d3fafb8f7661ea682020c35/plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java#L219)).
I'm not sure this is the right way to do it and for the iceberg connectors I haven't found a solution.

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,158 @@
...
+    private final boolean autoIncrement;
+    private final boolean nullable;
+    private final boolean readOnly;
+
+    public ColumnInfo(String name)
+    {
+        this(Optional.empty(), name, Optional.empty(), false, true, true);
+    }
+
+    public ColumnInfo(Optional<CatalogSchemaTableName> schemaTable, String name, Optional<String> label, boolean autoIncrement, boolean nullable, boolean readOnly)
```
findepi: This class has a builder, so let's channel all usage via it.
Make this constructor private.

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
```diff
@@ -0,0 +1,158 @@
...
+public class ColumnInfo
+{
+    private final Optional<CatalogSchemaTableName> schemaTable;
+    private final String name;
+    private final Optional<String> label;
+    private final boolean autoIncrement;
+    private final boolean nullable;
+    private final boolean readOnly;
+
+    public ColumnInfo(String name)
```
findepi, line 33: remove?

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -1042,9 +1048,75 @@ public void outputTaskFailed(TaskId taskId, Throwable failure)
...
+    {
+        Optional<TableHandle> tableHandle;
+        try {
+            tableHandle = metadata.getTableHandle(session, tableName);
+        }
+        catch (Throwable e) {
+            // FIXME: Some connector throw exception here
+            tableHandle = Optional.empty();
+            QUERY_STATE_LOG.error(e, "QueryStateMachine was unable to retrieve the TableHandle for table : " + tableName.toString());
+            // throw e;
```
findepi: Was this supposed to be addressed?

core/trino-main/src/main/java/io/trino/execution/QueryStateMachine.java
```diff
@@ -1042,9 +1048,75 @@ public void outputTaskFailed(TaskId taskId, Throwable failure)
...
+    {
+        if (tableHandle.isPresent()) {
+            try {
+                TableMetadata tableMetadata = metadata.getTableMetadata(session, tableHandle.get());
+                return tableMetadata.columns().stream().collect(Collectors.toMap(ColumnMetadata::getName, e -> e));
+            }
+            catch (Throwable e) {
+                // FIXME: Some connector throw exception here
+                QUERY_STATE_LOG.error(e, "QueryStateMachine was unable to retrieve the TableMetadata for table : " + tableName.toString());
+                // throw e;
```
findepi: Was this supposed to be addressed?

core/trino-main/src/main/java/io/trino/execution/SqlQueryExecution.java
```diff
@@ -528,9 +530,12 @@ private void planDistribution(PlanRoot plan, CachingTableStatsProvider tableStat
 
         // record output field
         PlanFragment rootFragment = plan.getRoot().getFragment();
-        stateMachine.setColumns(
-                ((OutputNode) rootFragment.getRoot()).getColumnNames(),
-                rootFragment.getTypes());
+        List<String> columnNames = ((OutputNode) rootFragment.getRoot()).getColumnNames();
+        List<Type> columnTypes = rootFragment.getTypes();
+
+        final boolean isQuery = analysis.isQuery();
```
findepi: Per project code style, we generally don't declare local variables as `final`

https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#additional-ide-configuration

client/trino-client/src/main/java/io/trino/client/Column.java
```diff
@@ -18,33 +18,121 @@
...
+
+    public Column(String catalog, String schema, String table, String name, String label, String type, ClientTypeSignature typeSignature)
+    {
+        this(Optional.of(catalog), Optional.of(schema), Optional.of(table), name, Optional.of(label), false, false, true, true, type, typeSignature);
+    }
+
+    public Column(String catalog, String schema, String table, String name, String label, boolean autoIncrement, boolean nullable, boolean readOnly, String type, ClientTypeSignature typeSignature)
+    {
+        this(Optional.of(catalog), Optional.of(schema), Optional.of(table), name, Optional.of(label), autoIncrement, false, nullable, readOnly, type, typeSignature);
+    }
```
findepi: Does this class really need 3 additional constructors?

core/trino-main/src/main/java/io/trino/sql/analyzer/Analysis.java
```diff
@@ -797,6 +797,28 @@ public ColumnHandle getColumn(Field field)
         return columns.get(field);
     }
 
+    public boolean isQuery()
+    {
+        return root instanceof Query;
```
findepi, line 866: @martint can you please review changes in Analysis?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -708,7 +709,7 @@ public LocatedTableHandle getTableHandle(
             return null;
         }
         Optional<Table> metastoreTable = metastore.getRawMetastoreTable(tableName.getSchemaName(), tableName.getTableName());
-        if (metastoreTable.isEmpty()) {
+        if (metastoreTable.isEmpty() || isSomeKindOfAView(metastoreTable.get())) {
```
findepi: looks unrelated. revert

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/metastore/TestDeltaLakeMetastoreAccessOperations.java
```diff
@@ -341,6 +341,10 @@ public void testSelectVersionedWithoutMetadataInMetastore()
         assertMetastoreInvocations("SELECT * FROM test_select_versioned_without_cache FOR VERSION AS OF 0",
                 ImmutableMultiset.<MetastoreMethod>builder()
                         .add(GET_TABLE)
+                        .build(),
+                ImmutableMultiset.<MetastoreMethod>builder()
+                        .add(GET_TABLE)
+                        .add(REPLACE_TABLE)
```
findepi, line 599: I don't understand why this change impacts Delta connector at all. It shouldn't. Is this because of the DeltaLakeMetadata change that will be reverted?

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemoryMetadata.java
```diff
@@ -216,7 +216,7 @@ public synchronized ConnectorTableHandle getTableHandle(ConnectorSession session
         }
 
         Long id = tableIds.get(schemaTableName);
-        if (id == null) {
+        if (id == null || !tables.containsKey(id)) {
```
findepi: Why this change?

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerJdbcConnectionAccesses.java
```diff
@@ -88,17 +88,17 @@ private static ConnectionCountingConnectionFactory getConnectionCountingConnecti
     @Test
     public void testJdbcConnectionCreations()
     {
-        assertJdbcConnections("SELECT * FROM nation LIMIT 1", 5, Optional.empty());
-        assertJdbcConnections("SELECT * FROM nation ORDER BY nationkey LIMIT 1", 5, Optional.empty());
-        assertJdbcConnections("SELECT * FROM nation WHERE nationkey = 1", 5, Optional.empty());
+        assertJdbcConnections("SELECT * FROM nation LIMIT 1", 6, Optional.empty());
```
findepi, line 91: That looks like regression. Where is it coming form?

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/execution/ColumnInfo.java
prrvchr: It is used by test...

PR #27323, opened 2025-11-15
docs(elasticsearch): update example timestamp

bvolpato:
## Description

Updating the example timestamp in the doc to a more current date in the Elasticsearch connector documentation.

This sounds like a weird PR, but a coworker noticed that Google was assuming the doc was created in 1987 (thanks @skedida!), maybe because it is a too specific timestamp in a structured format:

image

To avoid that quirk, I'm bumping the timestamp in the example to a more reasonable one this year (unaware if there's any reason to keep it in 1987).

## Release notes


chenjian2664:
ah, looks like Google treats that page as the creation or supported start date 😄 
first time notice about it lol

mosabua:
I wonder if its also wrong in OpenSearch docs btw @bvolpato

bvolpato:
> I wonder if its also wrong in OpenSearch docs btw @bvolpato

Thanks! I've submitted another PR for OpenSearch https://github.com/trinodb/trino/pull/27589

PR #27328, opened 2025-11-16
Fix preview UI to render long single line SQLs correctly

koszti:
## Description

The `CodeBlock` component in the preview UI doesn't render word-wrapped long lines with the correct height, which results in an incorrect layout. This PR fixes the issue by calculating the height properly.

Before:
image

After (sensitive data is masked):
image

## Release notes

```markdown
## Section
* Fix preview UI to render single line sqls. ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-web-ui/src/main/resources/webapp-preview/src/components/CodeBlock.tsx
+++ b/core/trino-web-ui/src/main/resources/webapp-preview/src/components/CodeBlock.tsx
@@ -57,13 +57,28 @@ export const CodeBlock = (props: ICodeBlockProps) => {
             return
         }
 
-        const lineHeightOption = editor.getOption(monaco.editor.EditorOption.lineHeight)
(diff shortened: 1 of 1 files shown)


wendigo:
@koszti I've also noticed recently, while on the plane, that query doesn't render offline as some of the components are fetched over network. Can we make these bundled so that preview UI works offline too?

koszti:
> @koszti I've also noticed recently, while on the plane, that query doesn't render offline as some of the components are fetched over network. Can we make these bundled so that preview UI works offline too?

oh good to know thanks. i’ll send PR

koszti:
@wendigo not fully related to this PR but I tested the offline mode and was not able to reproduce. Can you share any further info ideally a network error from you browser's dev tools? I've done this:

1. git pull latest trino from master
2. `./mvnw clean install -DskipTests`
3. disconnecting ethernet cable and wifi
4. running `WebUiPreviewQueryRunner`
5. opening http://localhost:8080/ui/preview

wendigo:
@koszti this happens for query json only

PR #27337, opened 2025-11-17
Remove unnecessary use of PageBuilder in PositionDeleteWriter

raunaqmorarka:
## Description
We can use pre-sized arrays instead of going through PageBuilder/BlockBuilder

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/PositionDeleteWriter.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/PositionDeleteWriter.java
@@ -24,8 +24,8 @@ import io.trino.plugin.iceberg.IcebergFileWriterFactory;
 import io.trino.plugin.iceberg.MetricsWrapper;
 import io.trino.plugin.iceberg.PartitionData;
 import io.trino.spi.Page;
-import io.trino.spi.PageBuilder;
 import io.trino.spi.block.Block;
+import io.trino.spi.block.LongArrayBlock;
 import io.trino.spi.block.RunLengthEncodedBlock;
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/PositionDeleteWriter.java
```diff
@@ -112,26 +111,69 @@ public void abort()
 
     private void writeDeletes(ImmutableLongBitmapDataProvider rowsToDelete)
     {
-        PageBuilder pageBuilder = new PageBuilder(List.of(BIGINT));
-
+        PositionsList deletedPositions = new PositionsList(4 * 1024);
```
sopel39, line 114: why `4 * 1024`

raunaqmorarka: It's small enough that oversizing won't be an issue and large enough to still benefit from batching processing in writer

pettyjamesm:
I would tend to agree with Copilot about the array re-use during `PositionsList#reset()` any reason we can't do that here?

raunaqmorarka:
> I would tend to agree with Copilot about the array re-use during `PositionsList#reset()` any reason we can't do that here?

That array becomes part of Page that is sent into the writer. Its safer to not make an assumption about what the writer does internally with that Page and make a fresh allocation for the next Page.

pettyjamesm:
> Its safer to not make an assumption about what the writer does internally with that Page and make a fresh allocation for the next Page.

Fair enough, lgtm

PR #27340, opened 2025-11-17
Require JDK Vector API at runtime

pettyjamesm:
## Description
Makes the JDK Vector API extensions ([JEP 508](https://openjdk.org/jeps/508)) required by Trino at runtime.

Before this change, code in the parquet reader would attempt to discover whether the API was enabled via reflection and gracefully fall back to scalar code when not found at runtime. This worked, but required any SIMD implementations to be kept fully isolated such that SIMD related code would only be loaded at runtime by the JVM if enabled which is onerous to maintain as we add more SIMD logic throughout the engine. Requiring this API to be enabled at runtime allows us to write code that need not use reflection tricks and separate classes to detect and isolate SIMD implementations from scalar code. 

- https://github.com/trinodb/trino/pull/26919 Adding more Vector API usages for block serialization
- [Launcher code](https://github.com/airlift/launcher/blob/9ad873a99e5c5931cb0f654ebb016d0df0d701d3/src/main/go/args/java_static_config.go#L8) already enables the vector API by default

## Release notes

```markdown
## General
* Require the JDK Vector API to be enabled at runtime ({issue}`27340`)
```

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/reader/ColumnReaderFactory.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/reader/ColumnReaderFactory.java
@@ -32,6 +32,7 @@ import io.trino.spi.type.TimestampType;
 import io.trino.spi.type.TimestampWithTimeZoneType;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.VarcharType;
+import jdk.incubator.vector.VectorShape;
 import org.apache.parquet.schema.LogicalTypeAnnotation;
 import org.apache.parquet.schema.LogicalTypeAnnotation.DateLogicalTypeAnnotation;
 import org.apache.parquet.schema.LogicalTypeAnnotation.DecimalLogicalTypeAnnotation;
@@ -84,7 +85,7 @@ import static org.apache.parquet.schema.PrimitiveType.PrimitiveTypeName.INT96;
 
 public final class ColumnReaderFactory
 {
-    private static final int PREFERRED_BIT_WIDTH = getVectorBitSize();
+    private static final int PREFERRED_BIT_WIDTH = VectorShape.preferredShape().vectorBitSize();
 
     private final DateTimeZone timeZone;
     private final boolean vectorizedDecodingEnabled;
@@ -371,17 +372,5 @@ public final class ColumnReaderFactory
         // Performance gains with vectorized decoding are validated only when the hardware platform provides at least 256 bit width registers
         // Graviton 2 machines return false here, whereas x86 and Graviton 3 machines return true
         return PREFERRED_BIT_WIDTH >= 256;
-    }
-
-    // get VectorShape bit size via reflection to avoid requiring the preview feature is enabled
-    private static int getVectorBitSize()
-    {
-        try {
-            Class<?> clazz = Class.forName("jdk.incubator.vector.VectorShape");
-            return (int) clazz.getMethod("vectorBitSize").invoke(clazz.getMethod("preferredShape").invoke(null));
-        }
-        catch (Throwable e) {
-            return -1;
-        }
     }
 }


core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
```diff
@@ -129,6 +130,20 @@ private static void verifyUsingG1Gc()
         }
     }
 
+    private static void verifyVectorApiEnabled()
+    {
+        try {
+            Class<?> clazz = Class.forName("jdk.incubator.vector.VectorShape");
+            int preferredBits = (int) clazz.getMethod("vectorBitSize").invoke(clazz.getMethod("preferredShape").invoke(null));
```
martint: Any reason for this specific choice of class and method to verify the presence of the Vector API?

docs/src/main/sphinx/installation/deployment.md
```diff
@@ -35,6 +35,10 @@ Trino requires a 64-bit version of Java 24, with a minimum required version of
 such as Java 8, Java 11, Java 17, Java 21 or Java 23 do not work.
 Newer versions such as Java 25 are not supported -- they may work, but are not tested.
 
+Trino requires enabling the [JDK Vector API extensions](https://openjdk.org/jeps/489) at runtime. Code using
```
martint: I don't think this is needed. The launcher enables it automatically, so there's nothing for users to do.

docs/src/main/sphinx/installation/deployment.md
```diff
@@ -136,6 +140,8 @@ The following provides a good starting point for creating `etc/jvm.config`:
 -XX:ReservedCodeCacheSize=512M
 -XX:PerMethodRecompilationCutoff=10000
 -XX:PerBytecodeRecompilationCutoff=10000
+# Enable Incubating Vector API (https://openjdk.org/jeps/489)
+--add-modules=jdk.incubator.vector
```
martint: This is already enabled by the launcher

pettyjamesm: Will remove

docs/src/main/sphinx/installation/deployment.md
pettyjamesm: Shall I rephrase this to make it clear the launcher already sets that flag, or just remove the section altogether? I've rephrased this section to make it more clear that the launcher sets the flag, but am happy to remove it altogether if you'd prefer.

core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
pettyjamesm: I took it from the existing detection logic that the parquet reader was using. Successfully loading any class from within `jdk.incubator.vector` should be sufficient as far as I know, but this was what we were already using to do the check there.

docs/src/main/sphinx/installation/deployment.md
martint: I would remove it altogether. It's a detail users shouldn't be concerned with -- we handle everything that's needed under the covers.

pettyjamesm: Ok, removed.

core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
martint: I think it would be cleaner to check whether `jdk.incubator.vector` is present in `ModuleLayer.boot().modules()`.

docs/src/main/sphinx/installation/deployment.md
raunaqmorarka: I think it is good idea to mention it because we should showcase that we're actually using some cool new Java features. Also, developers launch Trino process without going through trino launcher and the information about adding the vector module as a requirement will avoid some confusion.

martint: Those are separate issues.
* We should write blog posts, etc. to showcase all the cool under-the-hood things we do
* We should document the need for that in the developer section

However, the end user deployment guide should not talk about things that are implementation details.

core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
pettyjamesm: Updated to use that API instead, much cleaner!

docs/src/main/sphinx/installation/deployment.md
pettyjamesm: Which developer section should I add a mention of this to?

martint: Somewhere in here: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md, which is linked from https://trino.io/docs/current/develop.html

wendigo: If launcher is used, vector is linked by default: https://github.com/airlift/launcher/blob/master/src/main/go/args/java_static_config.go#L8

core/trino-main/src/main/java/io/trino/server/TrinoSystemRequirements.java
```diff
@@ -129,6 +130,13 @@ private static void verifyUsingG1Gc()
         }
     }
 
+    private static void verifyVectorApiEnabled()
+    {
+        if (ModuleLayer.boot().findModule("jdk.incubator.vector").isEmpty()) {
+            failRequirement("Trino requires the JDK vector API to be enabled");
```
wendigo: I'd phrase it as:

`Trino requires Vector API to be enabled/linked at runtime`

pettyjamesm: Reworded.

findepi:
> Before this change, code in the parquet reader would attempt to discover whether the API was enabled via reflection and gracefully fall back to scalar code when not found at runtime.

Do you know why it was done so?

pettyjamesm:
> Do you know why it was done so?

Presumably because we wanted to fallback gracefully when the API was not enabled (it's an incubating API, after all)

wendigo:
@findepi I think that we assumed at that time, that we don't want to require vector API as we are unsure of its stability. Since then, it proved stable and the only reason this is incubating is its reliance on the Valhalla.

martint:
It was from the days where we required Java 17 as the minimum version, and before we started tracking the latest JVM version. There was no canonical version to target. So, it was really an optional feature at the time.

docs/src/main/sphinx/installation/deployment.md
pettyjamesm: Will follow up with docs in a subsequent PR. The relevant bits to include in `DEVELOPMENT.md` might merit some additional discussion that isn't necessary to block this PR on.

pettyjamesm: Opened https://github.com/trinodb/trino/pull/27342

PR #27153, opened 2025-10-30
Add Trino 479 release notes

chenjian2664:
### The entry are not added yet
https://github.com/trinodb/trino/pull/27341
https://github.com/trinodb/trino/pull/27340
https://github.com/trinodb/trino/pull/27133

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

Format: PR/issue number, ✅ / ❌ rn ✅ / ❌ docs
✅ rn - release note added and verified, or assessed to be not necessary, set to ❌ rn before completion
✅ docs - need for docs assessed and merged, or assessed to be not necessary, set to ❌ docs before completion

## 30 Oct 2025
* #26521 ✅ rn ✅docs
* #27156 ✅ rn ✅ docs
* #27154 ✅ rn ✅ docs
* #27161 ✅ rn ✅ docs
* #27157 ✅ rn ✅ docs
* #27147 ✅ rn ✅ docs
* #27164 ✅ rn ✅ docs
* #27168 ✅ rn ✅ docs
* #27165 ✅ rn ✅ docs
* #27117 ✅ rn ✅ docs

## 31 Oct 2025
* #27167 ✅ rn ✅ docs
* #27166 ✅ rn ✅ docs
* #27169 ✅ rn ✅ docs
* #27177 ✅ rn ✅ docs 
* #27176 ✅ rn ✅ docs
* #27030 ✅ rn ✅ docs
* #27178 ✅ rn ✅ docs

## 03 Nov 2025
* #27179  ✅ rn ✅ docs
* #27185  ✅ rn ✅ docs
* #27160  ✅ rn ✅ docs
* #27190  ✅ rn ✅ docs
* #27191   ✅ rn ✅ docs - covered in `21506`
* #27189  ✅ rn ✅ docs
* #27080  ✅ rn ✅ docs
* #27051  ✅ rn ✅ docs

## 04 Nov 2025
* #27055  ✅ rn ✅ docs
* #26746  ✅rn ✅ docs
* #27201  ✅ rn✅ docs
* #27202 ✅ rn ✅ docs
* #27192 ✅ rn ✅ docs
* #27203 ✅ rn ✅ docs

## 05 Nov 2025
* #27211 ✅ rn ✅ docs
* #27208 ✅ rn ✅ docs
* #26629 ✅  rn  ✅docs
* #27024 ✅ rn ✅ docs
* #27214 ✅ rn ✅ docs
* #27193 ✅ rn ✅ docs
* #27205 ✅ rn ✅ docs
[...]

PR #27341, opened 2025-11-17
Apply workaround for antlr/antlr4#4901

wendigo:
## Release notes

```markdown

PF: this should not be mentioned in release notes IMO

## General
* Workaround query corruption issue that happens in a highly concurrent scenario on certain ARM processors ({issue}`issuenumber`)
```


findepi:
cc @lukasz-stec @martint

findepi:
re: RNs
this should not be mentioned in release notes IMO.

we don't have evidence of this affecting trino. we merged this change to be on the safe side.
based on the knowledge of the bug, i actually did try to find places where trino could work incorrectly, and did not find any actual candidate place.

chenjian2664:
@findepi Thanks, will add a ❌ for this entry in #27153

PR #27343, opened 2025-11-17
Remove unused MemoryPoolInfo.queryMemoryRevocableReservations field

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/memory/ClusterMemoryPool.java
+++ b/core/trino-main/src/main/java/io/trino/memory/ClusterMemoryPool.java
@@ -154,14 +154,14 @@ public class ClusterMemoryPool
         Map<String, MemoryAllocation> mergedAllocations = new HashMap<>();
 
         for (MemoryAllocation allocation : left) {
-            mergedAllocations.put(allocation.getTag(), allocation);
+            mergedAllocations.put(allocation.tag(), allocation);
         }
 
         for (MemoryAllocation allocation : right) {
             mergedAllocations.merge(
-                    allocation.getTag(),
+                    allocation.tag(),
                     allocation,
-                    (a, b) -> new MemoryAllocation(a.getTag(), a.getAllocation() + b.getAllocation()));
+                    (a, b) -> new MemoryAllocation(a.tag(), a.allocation() + b.allocation()));
         }
 
         return new ArrayList<>(mergedAllocations.values());
--- a/core/trino-spi/pom.xml
+++ b/core/trino-spi/pom.xml
@@ -721,6 +721,49 @@
                                     <old>method java.util.Map&lt;io.trino.spi.QueryId, java.lang.Long&gt; io.trino.spi.memory.MemoryPoolInfo::getQueryMemoryRevocableReservations()</old>
                                     <justification>Removal of unused field</justification>
                                 </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.annotation.removed</code>
+                                    <old>parameter void io.trino.spi.memory.MemoryAllocation::&lt;init&gt;(===java.lang.String===, long)</old>
+                                    <new>parameter void io.trino.spi.memory.MemoryAllocation::&lt;init&gt;(===java.lang.String===, long)</new>
+                                    <annotation>@com.fasterxml.jackson.annotation.JsonProperty("tag")</annotation>
+                                    <justification>Class converted to a record</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.annotation.removed</code>
+                                    <old>parameter void io.trino.spi.memory.MemoryAllocation::&lt;init&gt;(java.lang.String, ===long===)</old>
+                                    <new>parameter void io.trino.spi.memory.MemoryAllocation::&lt;init&gt;(java.lang.String, ===long===)</new>
+                                    <annotation>@com.fasterxml.jackson.annotation.JsonProperty("allocation")</annotation>
+                                    <justification>Class converted to a record</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.annotation.removed</code>
+                                    <old>method void io.trino.spi.memory.MemoryAllocation::&lt;init&gt;(java.lang.String, long)</old>
+                                    <new>method void io.trino.spi.memory.MemoryAllocation::&lt;init&gt;(java.lang.String, long)</new>
+                                    <annotation>@com.fasterxml.jackson.annotation.JsonCreator</annotation>
+                                    <justification>Class converted to a record</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.removed</code>
+                                    <old>method long io.trino.spi.memory.MemoryAllocation::getAllocation()</old>
+                                    <justification>Class converted to a record</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.removed</code>
+                                    <old>method java.lang.String io.trino.spi.memory.MemoryAllocation::getTag()</old>
+                                    <justification>Class converted to a record</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.class.kindChanged</code>
+                                    <old>class io.trino.spi.memory.MemoryAllocation</old>
+                                    <new>class io.trino.spi.memory.MemoryAllocation</new>
+                                    <justification>Class converted to a record</justification>
+                                </item>
                             </differences>
                         </revapi.differences>
                     </analysisConfiguration>
--- a/core/trino-spi/src/main/java/io/trino/spi/memory/MemoryAllocation.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/memory/MemoryAllocation.java
@@ -13,34 +13,12 @@
  */
 package io.trino.spi.memory;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
-
 import static java.util.Objects.requireNonNull;
 
-public final class MemoryAllocation
+public record MemoryAllocation(String tag, long allocation)
 {
-    private final String tag;
-    private final long allocation;
-
-    @JsonCreator
-    public MemoryAllocation(
-            @JsonProperty("tag") String tag,
-            @JsonProperty("allocation") long allocation)
+    public MemoryAllocation
     {
-        this.tag = requireNonNull(tag, "tag is null");
-        this.allocation = allocation;
-    }
-
-    @JsonProperty
-    public String getTag()
-    {
-        return tag;
-    }
-
-    @JsonProperty
-    public long getAllocation()
-    {
-        return allocation;
+        requireNonNull(tag, "tag is null");
     }
 }


core/trino-main/src/main/java/io/trino/execution/QueryTracker.java
```diff
@@ -55,7 +54,7 @@ public class QueryTracker<T extends TrackedQuery>
     private final int maxQueryHistory;
     private final Duration minQueryExpireAge;
 
-    private final ConcurrentMap<QueryId, T> queries = new ConcurrentHashMap<>();
+    private final ConcurrentMap<QueryId, T> queries = new ConcurrentHashMap<>(0);
```
raunaqmorarka: Seems unrelated change.
Also, why bother ? I don't think we will have many instances of this class and having more than 16 queries tracked is not unusual

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
```diff
@@ -341,33 +336,30 @@ public QueryTracker<QueryExecution> getQueryTracker()
      */
     private void enforceMemoryLimits()
     {
-        List<QueryExecution> runningQueries = queryTracker.getAllQueries().stream()
-                .filter(query -> query.getState() == RUNNING)
-                .collect(toImmutableList());
-        memoryManager.process(runningQueries, this::getQueries);
+        memoryManager.process(queryTracker::getAllQueries, queryTracker::tryGetQuery);
```
raunaqmorarka: Why is the `.filter(query -> query.getState() == RUNNING)` condition moved from here into memoryManager ?

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
```diff
@@ -174,18 +174,25 @@ public synchronized void addChangeListener(Consumer<MemoryPoolInfo> listener)
         changeListeners.add(listener);
     }
 
-    public synchronized void process(Iterable<QueryExecution> runningQueries, Supplier<List<BasicQueryInfo>> allQueryInfoSupplier)
+    public synchronized void process(Supplier<Stream<QueryExecution>> allQueries, Function<QueryId, Optional<QueryExecution>> executionInfoSupplier)
```
raunaqmorarka: Given that stream itself is lazy, why is there a need to use a Supplier to create the Stream ?

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryLeakDetector.java
```diff
@@ -46,47 +44,47 @@ public class ClusterMemoryLeakDetector
     private Set<QueryId> leakedQueries;
 
     /**
-     * @param queryInfoSupplier All queries that the coordinator knows about.
      * @param queryMemoryReservations The memory reservations of queries in the cluster memory pool.
      */
```
raunaqmorarka, line 51: `@param executionInfoSupplier ...`

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryPool.java
```diff
@@ -52,23 +55,23 @@ public class ClusterMemoryPool
 
     // Does not include queries with zero memory usage
     @GuardedBy("this")
-    private final Map<QueryId, Long> queryMemoryReservations = new HashMap<>();
+    private final Object2LongArrayMap<QueryId> queryMemoryReservations = new Object2LongArrayMap<>();
```
raunaqmorarka: Conversion from Map to another structure is unrelated to using lazy streams, this should go in separate commit

raunaqmorarka: Why is `Object2LongArrayMap` the better choice here instead of `Object2LongMap` ?
Please add a code comment explaining that, otherwise someone else is just as likely to pick another structure

core/trino-main/src/main/java/io/trino/security/AccessControlUtil.java
```diff
@@ -34,22 +35,23 @@ public static void checkCanViewQueryOwnedBy(Identity identity, Identity queryOwn
         accessControl.checkCanViewQueryOwnedBy(identity, queryOwner);
     }
 
-    public static List<BasicQueryInfo> filterQueries(Identity identity, List<BasicQueryInfo> queries, AccessControl accessControl)
+    public static Stream<BasicQueryInfo> filterQueries(Supplier<Stream<BasicQueryInfo>> queries, Identity identity, AccessControl accessControl)
```
raunaqmorarka: Why not just Stream, instead of Supplier of Stream

raunaqmorarka: Why change the order of the parameters ?

core/trino-main/src/main/java/io/trino/server/ui/ClusterStatsResource.java
```diff
@@ -77,7 +79,9 @@ public ClusterStats getClusterStats()
         long totalInputBytes = dispatchManager.getStats().getConsumedInputBytes().getTotalCount();
         long totalCpuTimeSecs = dispatchManager.getStats().getConsumedCpuTimeSecs().getTotalCount();
 
-        for (BasicQueryInfo query : dispatchManager.getQueries()) {
+        Iterator<BasicQueryInfo> iterator = dispatchManager.getQueries().iterator();
+        while (iterator.hasNext()) {
+            BasicQueryInfo query = iterator.next();
```
raunaqmorarka: forEach ?

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
```diff
@@ -174,18 +174,25 @@ public synchronized void addChangeListener(Consumer<MemoryPoolInfo> listener)
...
         boolean outOfMemory = isClusterOutOfMemory();
 
         boolean queryKilled = false;
         long totalUserMemoryBytes = 0L;
         long totalMemoryBytes = 0L;
-        for (QueryExecution query : runningQueries) {
+        int queriesCount = 0;
+        Iterator<QueryExecution> iterator = allQueries.get().iterator();
+        while (iterator.hasNext()) {
+            QueryExecution query = iterator.next();
```
raunaqmorarka: forEach on the stream ?

testing/trino-tests/src/test/java/io/trino/memory/TestMemoryManager.java
```diff
@@ -173,7 +175,9 @@ private void waitForQueryToBeKilled(QueryRunner queryRunner)
     {
         while (true) {
             boolean hasRunningQuery = false;
-            for (BasicQueryInfo info : queryRunner.getCoordinator().getQueryManager().getQueries()) {
+            Iterator<BasicQueryInfo> iterator = queryRunner.getCoordinator().getQueryManager().getQueries().iterator();
+            while (iterator.hasNext()) {
+                BasicQueryInfo info = iterator.next();
```
raunaqmorarka: forEach ?

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
wendigo: In other similar places filtering happens inside the enforcement logic

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
wendigo: You can't consume stream twice (it's like an iterator). In this case we need to read it twice

core/trino-main/src/main/java/io/trino/security/AccessControlUtil.java
wendigo: Stream is read twice in that method

wendigo: Seems more logical (filter what how)

core/trino-main/src/main/java/io/trino/server/ui/ClusterStatsResource.java
wendigo: I'll check but some usages update non final counters etc

martint:
The PR does exactly what the title says, but what the motivation for this change?

wendigo:
@martint we process the queries lists in the streaming fashion. This PR avoids unnecessary copies and reduces memory allocations when working with long query lists.

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryLeakDetector.java
```diff
@@ -47,47 +44,47 @@ public class ClusterMemoryLeakDetector
     private Set<QueryId> leakedQueries;
 
     /**
-     * @param queryInfoSupplier All queries that the coordinator knows about.
```
raunaqmorarka, line 49: This `param` should only be renamed instead of removed in this commit, right ?

martint:
Where do we see those unnecessary allocations? Is that a problem in practice? A cluster will usually not be dealing with so many queries that that should be visible.

BTW, my concern with using Stream in a general purpose API like this is that it forces a style of usage that we don't want to thrust upon on every caller. It may also introduce hidden costs if you need to do something with the results twice, especially if the underlying stream has complex filters/transformations applied to it (or force the caller to materialize into a temporary collection).

wendigo:
@martint I was looking at these while running high concurrency/small queries benchmarks.

In query enforcement code (scan, write, memory, cpu limits) we are processing queries like this, effectively always copying a list of queries every 1 s (x4)

```
        List<QueryExecution> runningQueries = queryTracker.getAllQueries().stream()
                .filter(query -> query.getState() == RUNNING)
                .collect(toImmutableList());
```

and then processing queries one-by-one. The only place where we work on the stream twice is the memory leak detector where we run callOomKiller once again over a stream but this happens if we are out of memory (so not every time).

martint:
QueryTracker.getAllQueries() doesn't copy the queries:

```java
public Collection<T> getAllQueries()
{
    return unmodifiableCollection(queries.values());
}
```

So, the copying is confined to the method you posted above. It sounds like we just need a better interaction between that method and ClusterQueryManager.

wendigo:
@martint what's wrong with this change if the elements are processed one by one and we only do filtering/mapping?

martint:
It's too invasive (making SqlQueryTracker return Stream), when we're trying to solve how SqlQueryManager processes the queries and feeds them to the ClusterQueryManager.

From an abstraction perspective, there's no inherent "streaminess" or "laziness" to SqlQueryTracker. A Stream is also more constraining. It only allows a simple one-pass computations, and forces callers to adopt a style of processing that may not be appropriate in all situations. Streams are more appropriate for representing computation pipelines, not a data model/data structure.

If the caller needs a Stream (for the purpose of filtering/transforming), it can just call stream() on the collection it returns.

wendigo:
@martint I'm not talking about some future applications but what we have right now. I disagree that "streams are representing computational pipelines". In this case, we are exposing an ability to traverse list of queries, rather than a container for these queries (given that this is a snapshot in time since the list changes frequently). From the consumer perspective, this states an intent - "walk through this structure" rather than "stash it/keep it somewhere". Stream having its "walk once" semantics better reflects the fact that this is heavily mutated structure and you can only peek into its current content, right here, right now.

martint:
@dain, can you comment on the original intent and abstraction provided by QueryTracker from when you refactored that code a few years ago (if you recall)?

dain:
> @dain, can you comment on the original intent and abstraction provided by QueryTracker from when you refactored that code a few years ago (if you recall)?

Quite simply to have a master list of queries, so we could handle all of the complex timeout logic in one, easy to understand, place.  In the old days, we had, abandonded query detection, timeout, and history management spread throughtout the even more complex query manager system, and we had tons of bugs (memory leaks).  I pulled this part out so that we could test independently and have a better chance of not getting more bugs.

For this PR, I'm not sure of the value of this change.  It basically forces a `.stream()` call on every caller, and that is it.  It doesn't change the underlying behavior of the collection, which is a live, unmodifiable, view over a concurrent map, So it takes away flexibility from the users.  That said if I want a snapshot then, I can call `.toList()`... obviously with this change you no longer get a live collection, but I'm not sure I'd every want that either.

Is there some bad behavior users get into with the original design?

wendigo:
@dain no, just optimizing for memory usage by reducing number of unnecessary memory copies/materialization

dain:
@wendigo I don't see how this changes "memory copies/materialization"

wendigo:
@dain:

```
List<QueryExecution> runningQueries = queryTracker.getAllQueries().stream()
                .filter(query -> query.getState() == RUNNING)
                .collect(toImmutableList());
```

this copies running queries to a new list

dain:
I assume you are talking about:
```
    private void enforceMemoryLimits()
    {
        List<QueryExecution> runningQueries = queryTracker.getAllQueries().stream()
                .filter(query -> query.getState() == RUNNING)
                .collect(toImmutableList());
        memoryManager.process(runningQueries, this::getQueries);
    }
```
Which has been updated to:
```
    private void enforceMemoryLimits()
    {
        memoryManager.process(queryTracker::getAllQueries, queryTracker::tryGetQuery);
    }
```
But you could have done the same thing with:
```
    private void enforceMemoryLimits()
    {
        memoryManager.process(queryTracker.getAllQueries()::stream, queryTracker::tryGetQuery);
    }
```
 
So my point is I don't see how changing  `getAllQueries` to return Stream instead of Collection improves memory copies/materialization.  I do see how changing `memoryManager.process` to take `Supplier<Stream<QueryExecution>>` does do this.

wendigo:
@dain agreed but then on every usage site I'm doing ::stream so what's the point?

dain:
> @dain agreed but then on every usage site I'm doing ::stream so what's the point?

There are lots of cases where I have collections an only interact with them as streams, but I don't change them to be `Supplier<Stream<T>>`.  I'm not saying that is can't be done, or that there aren't cases that it should be done.  I am saying, I don't think you have made your caase, and I don't see the need for that specific part of this PR.  I don't feel super strongly about this, as this is an internal API and we can just switch back so 🤷

wendigo:
@dain yeah as an internal API it's easy to change it back when usage patter changes

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryLeakDetector.java
```diff
@@ -49,16 +50,16 @@ public class ClusterMemoryLeakDetector
      * @param queryInfoSupplier All queries that the coordinator knows about.
      * @param queryMemoryReservations The memory reservations of queries in the cluster memory pool.
      */
-    void checkForMemoryLeaks(Supplier<List<BasicQueryInfo>> queryInfoSupplier, Map<QueryId, Long> queryMemoryReservations)
+    void checkForMemoryLeaks(Supplier<List<BasicQueryInfo>> queryInfoSupplier, Object2LongMap<QueryId> queryMemoryReservations)
```
findepi: This hurts eyes. we should use fastutils for very big maps only.
Maps keyed by QueryId  are unlikely big.

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
```diff
@@ -226,20 +229,21 @@ public synchronized void process(Iterable<QueryExecution> runningQueries, Suppli
...
-        updateMemoryPool(Iterables.size(runningQueries));
+        updateMemoryPool(queriesCount);
         updateNodes();
     }
 
-    private synchronized void callOomKiller(Iterable<QueryExecution> runningQueries)
+    private synchronized void callOomKiller(Collection<QueryExecution> allQueries, Function<QueryId, Optional<QueryExecution>> queryExecutionSupplier)
     {
-        List<RunningQueryInfo> runningQueryInfos = Streams.stream(runningQueries)
+        List<RunningQueryInfo> runningQueryInfos = allQueries
```
chenjian2664: filtering by state -> RUNNING ?

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryLeakDetector.java
wendigo: it's not about big but more about boxing primitive long

findepi: Boxing here doesn't matter.
Boxing matters where applied to large number if objects (millions+). There is no way this map, keyed by query id, will have millions of entries.

Prefer performance over readability when performance gains are meaningful and readability is not totally lost.
Prefer readability over illusory performance gains.

findepi:
This PR does too much IMO, changing places where improvement can be expected, but also places where it rather cannot.
Example here: https://github.com/trinodb/trino/pull/27343#discussion_r2599221741 but it looks like a pattern rather than one-off.

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryLeakDetector.java
dain: That would be my guess also.  If we have even 1k queries running currently, there are likely much bigger problems.  Is this an issue you see in production?  If so can you explain what it looks like.

core/trino-main/src/main/java/io/trino/connector/system/QuerySystemTable.java
```diff
@@ -104,12 +104,8 @@ public ConnectorTableMetadata getTableMetadata()
     public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, ConnectorSession session, TupleDomain<Integer> constraint)
     {
         checkState(dispatchManager.isPresent(), "Query system table can return results only on coordinator");
-
-        List<BasicQueryInfo> queries = dispatchManager.get().getQueries();
-        queries = filterQueries(((FullConnectorSession) session).getSession().getIdentity(), queries, accessControl);
-
         Builder table = InMemoryRecordSet.builder(QUERY_TABLE);
-        for (BasicQueryInfo queryInfo : queries) {
+        for (BasicQueryInfo queryInfo : filterQueries(dispatchManager.get().getQueries(), ((FullConnectorSession) session).getSession().getIdentity(), accessControl)) {
```
dain: For this one, does it actually stop a list copy, or is this just a simpler inlineing of the variables.  I ask because it looks like getQueries still returns a list, and filterQueries still returns a list.  Am I missing something or is this not the copy you are eliminating?

dain (commented):
I'm not sure I buy the value of the first commit, but I also don't think it is a big deal either way.  I believe there are other places using flat structures in this code so it isn't that weird.

I have one question on the last commit about which copies we are saving.

I have no objections to merging this.  Seems fine to me, but others have commented, so I'll leave it up to them.

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
```diff
@@ -344,7 +344,7 @@ private void enforceMemoryLimits()
         List<QueryExecution> runningQueries = queryTracker.getAllQueries().stream()
                 .filter(query -> query.getState() == RUNNING)
                 .collect(toImmutableList());
-        memoryManager.process(runningQueries, this::getQueries);
+        memoryManager.process(runningQueries, queryTracker::tryGetQuery);
```
findepi: Can this affect concurrency subtly?

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
```diff
@@ -174,18 +171,20 @@ public synchronized void addChangeListener(Consumer<MemoryPoolInfo> listener)
         changeListeners.add(listener);
     }
 
-    public synchronized void process(Iterable<QueryExecution> runningQueries, Supplier<List<BasicQueryInfo>> allQueryInfoSupplier)
+    public synchronized void process(Collection<QueryExecution> runningQueries, Function<QueryId, Optional<QueryExecution>> executionInfoSupplier)
```
findepi: The collection being passed here is a List.

I think we generally prefer List interface (as more readable) over the more generic Collection, unless of course we benefit from more flexibility

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
```diff
@@ -226,20 +225,21 @@ public synchronized void process(Iterable<QueryExecution> runningQueries, Suppli
 
         if (!lowMemoryKillers.isEmpty() && outOfMemory && !queryKilled) {
             if (isLastKillTargetGone()) {
-                callOomKiller(runningQueries);
+                callOomKiller(runningQueries, executionInfoSupplier);
```
findepi: It appears to me that this might be changing what the state of the world the OOM killer may observe.

Before the changes, the OOM killer logic would run only after all other checks above. In the OOM situation, the `resourceOvercommit && outOfMemory` check above could select some query to be killed. Then `queryKilled==true` and OOM killer isn't invoked.

After the changes, we're not snapshotting the state of queries anymore and that state can be changing concurrently. This could potentially lead to a different final outcome.

findepi (commented):
> `Avoid copying list of tracked queries`

This commit affects coordinator's concurrent code that tracks distributed state across the cluster. This is high risk change to me. The commit lacks clear motivation why we're doing this, or what are the expected benefits.

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
wendigo: @findepi oom killer and other enforcement bits are running on schedule (every 1 second) so it's not guaranteed that these limits will not be crossed anyway (this can happen in a 1 second window between the last call and the next one)

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
wendigo: yes, but doesn't matter if query finished concurrently

core/trino-main/src/main/java/io/trino/memory/ClusterMemoryManager.java
findepi: i already knew that part.

core/trino-main/src/main/java/io/trino/execution/SqlQueryManager.java
findepi: Depends on the exact effects of this subtle concurrency changes. What are the new interleavings possible now that we copy/snapshot less and thus semantics of sequential code do change. My question about OOM killer (https://github.com/trinodb/trino/pull/27343#discussion_r2601518751) exemplary concern.

findepi:
This PR lacks description that motivates the change.
It also changes concurrent code in ways that could affect what the code does.
It is normal for a change to break something, the risk of breakage is inherent and I understand that. However, if something breaks because of a change, it would be good to understand _why_ we did the change in the first place.

Issue #23747, opened 2024-10-10
Materialized view should only check permissions of underlying query for refresh


dain:
From a security perspective a materialized view is just a table, and should be treated as such.  The only time we should be checking if the definer has permissions to run the view attached to the query is during refresh.

This was found in the discussion: https://github.com/trinodb/trino/discussions/23387

lozbrown:
@dain should refresh materialized view be a separate permission in the API/SPI, refreshing a view of a big table can be a significant load on the platform and we may want to prevent users doing that too frequently?

lozbrown:
@dain Thinking about this more, whilst I agree for select an MV should be treated as a table, should the end user really need permission to the underlying table just to refresh?

Imagine this scenario

You create an MV over a sensitive_table_a for a report that non_priveliged_userb uses on a regular basis. If the MV aggregates it otherwise does not include the sensitive data by refreshing non_priveliged_userb is not gaining any additional access to data? 

By this motion only select on the MV should be necessary for the refresh and only the creator needs access to the underlying MV. 

However refreshing an MV can be potentially expensive and may have been created (instead of a view) to prevent users from doing the expensive query underneath too frequently, hence a separate refresh permission would be better so that refreshes may only occur when the grace period runs out or someone with the permission refreshes

dain:
> [@dain](https://github.com/dain) should refresh materialized view be a separate permission in the API/SPI

It is.  [Here](https://github.com/trinodb/trino/blob/master/core/trino-spi/src/main/java/io/trino/spi/security/SystemAccessControl.java#L615) is the check.



> should the end user really need permission to the underlying table just to refresh?

I think you misunderstood me.  During refresh we check if the **"definer"** (a.k.a., the owner) of the materialized view still has permission to access the underlying data of the view.  The refresh is executed using the permissions of the definer not the "invoker" of the command.

codyzwief:
I have attempted to take a stab at this in https://github.com/trinodb/trino/pull/25872

Issue #15326, opened 2022-12-07
Define grace/staleness period for materialized views


findepi:
Per @martint comment https://github.com/trinodb/trino/pull/15104#issuecomment-1339816761

> I'm thinking of syntax like this:
> 
> ```
> CREATE MATERIALIZED VIEW <name>
>     (GRACE PERIOD <interval>)?
>     (WHEN STALE (FAIL | REFRESH))?
>     AS (<query>)
>     (WITH NO DATA)?
> ```
> 
> * A materialized view can be in either `usable` or `unusable` state.
> * `GRACE PERIOD` indicates the maximum allowed time since creation or last refresh before the MV is considered unusable. If omitted, the MV will always be considered usable (i.e., infinite grace period)
> * `WHEN STALE (FAIL | REFRESH)` indicates what to do once the MV is considered `unusable`. If `FAIL`, querying the MV will result in a failure indicating the MV is stale. If `REFRESH`, the query will transparently produce the correct results as if the MV was up to date (either by refreshing the MV before the query runs or by running the underlying query directly as if it were a regular view.
> * `WITH NO DATA` causes the MV to be created but not populated in the beginning. One open question is whether the MV is considered unusable until the first refresh occurs (explicitly via `REFRESH MATERIALIZED VIEW` or implicitly if `WHEN STALE REFRESH` is specified), or whether it's considered within the grace period. In the latter case, the `WHEN STALE` clause would only make sense when a grace period is specified explicitly (since otherwise, it's infinite), so we could constrain it syntactically just for that scenario.
> 
> We don't need to implement all those features up front, but I'd like to have a plausible plan for what the syntax would look like when we get to it.

findepi:
> * `GRACE PERIOD` indicates the maximum allowed time since creation or last refresh before the MV is considered unusable. If omitted, the MV will always be considered usable (i.e., infinite grace period)

I like this syntax.

Let's acknowledge that we need a backward incompatible change to introduce it:
Today we have no `GRACE PERIOD`, and this currently implies zero grace period.
What we want is to default to infinite grace period.

I think this is the change we should start from.

> WHEN STALE (FAIL | REFRESH) indicates what to do once the MV is considered unusable. If FAIL, querying the MV will result in a failure indicating the MV is stale. If REFRESH, the query will transparently produce the correct results as if the MV was up to date (either by refreshing the MV before the query runs or by running the underlying query directly as if it were a regular view.

[...]

PR #27356, opened 2025-11-18
Add `WHEN STALE` syntax for `CREATE MATERIALIZED VIEW`

piotrrzysko:
## Description

This PR introduces a new optional clause for `CREATE MATERIALIZED VIEW` that allows users to control the behavior of materialized views when they are stale. 

I will updated the documentation once the proposed syntax and behavior are approved.

### New syntax
``` sql
CREATE MATERIALIZED VIEW mv_name
WHEN STALE { FAIL | INLINE }
AS SELECT ...
```

### Behavior
* **FAIL** - If the MV is stale, queries referencing it will fail. Analysis of the underlying query is never performed while querying the MV.
* **INLINE** (default) - Preserves the current behavior. Even if the MV is stale, it is expanded like a logical view, and the underlying query is analyzed.

### Motivation
* Avoid analyzing the underlying query every time, which can be costly when it references many tables.
* Allow using fresh MVs even when some data sources are temporarily unavailable.

* The proposed clause was originally discussed in https://github.com/trinodb/trino/issues/15326
* https://github.com/trinodb/trino/issues/23747

## Release notes


core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -2293,10 +2295,13 @@ protected Scope visitTable(Table table, Optional<Scope> scope)
                     checkStorageTableNotRedirected(storageTableName);
                     TableHandle tableHandle = metadata.getTableHandle(session, storageTableName)
                             .orElseThrow(() -> semanticException(INVALID_VIEW, table, "Storage table '%s' does not exist", storageTableName));
-                    return createScopeForMaterializedView(table, name, scope, materializedViewDefinition, Optional.of(tableHandle));
+                    return createScopeForMaterializedView(table, name, scope, materializedViewDefinition, Optional.of(tableHandle), useLogicalViewSemantics);
```
piotrrzysko: I'm a bit concerned about this line: the new clause affects the path for fresh MVs (`WHEN STALE FAIL` skips analysis, while `WHEN STALE INLINE` runs the analysis), even though the syntax suggests it should only define behavior for stale MVs.

Is analysis of the underlying query even necessary for fresh MVs, regardless of the `WHEN STALE` clause? I understand that it can detect schema changes in the base tables or changes to the MV definer’s access permissions - but is that the intended behavior?

Praveen2112:
What is our plan on changing this behaviour for existing or new MVs ?

core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
```diff
@@ -156,12 +158,21 @@ Analysis executeInternal(
                     return Duration.ofMillis(milliseconds);
                 });
 
+        Optional<WhenStaleBehavior> whenStale = statement.getWhenStaleBehavior()
+                .map(_ -> {
+                    if (!plannerContext.getMetadata().getConnectorCapabilities(session, catalogHandle).contains(MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR)) {
+                        throw semanticException(NOT_SUPPORTED, statement, "Catalog '%s' does not support WHEN STALE", catalogName);
+                    }
```
Praveen2112, line 165: Should catalogs need to handle anything in order to support it ? If it is some sort of MV property (like security mode in case of views) - It should be supported for all connectors supporting MVs right ? Or we could fail during the creation or alter phase ?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -2293,10 +2295,13 @@ protected Scope visitTable(Table table, Optional<Scope> scope)
...
+                }
+                else if (!useLogicalViewSemantics) {
+                    throw semanticException(VIEW_IS_STALE, table, "Materialized view '%s' is stale", name);
                 }
                 // This is a stale materialized view and should be expanded like a logical view
-                return createScopeForMaterializedView(table, name, scope, materializedViewDefinition, Optional.empty());
+                return createScopeForMaterializedView(table, name, scope, materializedViewDefinition, Optional.empty(), useLogicalViewSemantics);
```
Praveen2112: Instead of passing a `useLogicalViewSemantics` can we create a scope for the storage table ?

Praveen2112:
Apart from the benefit on skipping the analysis what are the other benefits would I get if a run a query as `WHEN STALE FAIL` ? Now when I create a MV with `WHEN STALE` - I won't be able to use it unless I run a `REFRESH MV`.  In case of pipeline which uses this MV might be affected if a MV went stale ? Is this expected ? Should we add this as an extension of `WHEN FRESH SKIP ANALYSIS` or something similar ?

martint:
For context, https://github.com/trinodb/trino/issues/15326

martint:
Also, https://github.com/trinodb/trino/discussions/23387#discussioncomment-10640385 and https://github.com/trinodb/trino/issues/23747

piotrrzysko:
@Praveen2112 

> What is our plan on changing this behaviour for existing or new MVs ?

This change is meant to be backward-compatible. The new clause is optional. If it is not specified, or if `WHEN STALE INLINE` is used, the MV will behave as it does currently.

> Apart from the benefit on skipping the analysis what are the other benefits would I get if a run a query as WHEN STALE FAIL ? Now when I create a MV with WHEN STALE - I won't be able to use it unless I run a REFRESH MV. In case of pipeline which uses this MV might be affected if a MV went stale ? Is this expected ? 

Skipping the analysis is not the goal in itself. The goal is to achieve better availability when the source tables are not available and to get closer to the semantics mentioned in the issues that Martin and I linked in this PR - specifically, that within the grace period, the MV should be treated as if it were a regular storage table. And today, as far as I know, the only blocker for this is the analysis performed every time the MV is queried.

> Should we add this as an extension of WHEN FRESH SKIP ANALYSIS or something similar ?

This is more or less what I wanted to express in this comment: https://github.com/trinodb/trino/pull/27356#discussion_r2537145553. I’m happy to discuss what the syntax should look like to achieve the goals described above. Personally, `WHEN FRESH SKIP ANALYSIS` feels too low-level - as mentioned in my comment, I'd consider always skipping analysis when the MV is fresh.

piotrrzysko:
@martint just a gentle reminder about the PR when you have a moment

findepi:
@piotrrzysko can you please rebase? there is a conflict.

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -562,6 +562,7 @@ private Query showCreateMaterializedView(ShowCreate node)
                     false,
                     false,
                     Optional.empty(), // TODO support GRACE PERIOD
```
findepi, line 564: did we miss doing this?

findepi (commented):
> `Add WHEN STALE FAIL/INLINE syntax`

lgtm

core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
findepi: I don't understand why we need a connector capability for this.
What's the default behavior of WHEN STALE going to be?

findepi: oh, i get it -- storing the clause requires a capability.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -5818,11 +5818,11 @@ public void testAnalyzeMaterializedViewWithAccessControl()
         TestingAccessControlManager accessControlManager = new TestingAccessControlManager(transactionManager, emptyEventListenerManager(), new SecretsResolver(ImmutableMap.of()));
         accessControlManager.setSystemAccessControls(List.of(AllowAllSystemAccessControl.INSTANCE));
 
-        analyze("SELECT * FROM fresh_materialized_view");
+        analyze(CLIENT_SESSION, "SELECT * FROM fresh_materialized_view", accessControlManager);
```
findepi, line 5822: > Fix testAnalyzeMaterializedViewWithAccessControl

can you perhaps move to separate PR so we just merge this?

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
piotrrzysko: Yeah, it looks like _someone_ (https://github.com/trinodb/trino/commit/6d5215c7911b16736e11569ebd7b20f0e53fe973) forgot about it 😀.

findepi: do you want to create a separate PR to have this addressed?

piotrrzysko: Sure!

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -1598,21 +1598,207 @@ public void testMaterializedViewGracePeriod()
...
+            assertUpdate("INSERT INTO " + baseTable + " (regionkey, name) VALUES (42, 'foo new region')", 1);
+            String updatedResults = initialResults + " UNION ALL VALUES (42, 'foo new region')";
+
+            // The materialization is stale now
+            assertThat(query(defaultSession, "TABLE " + viewName))
+                    .hasPlan(readFromStorageTable)
+                    .matches(initialResults);
+            assertThat(query(futureSession, "TABLE " + viewName))
+                    .hasPlan(readFromBaseTables)
+                    .matches(updatedResults);
```
findepi: The MV has `GRACE PERIOD INTERVAL '1' HOUR`.
The GRACE is defined earlier in the MV definition and should take precedence. I.e. we should keep querying the MV storage when it's known to be stale but grace period did not elapse.

findepi (approved):

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
findepi: moved to https://github.com/trinodb/trino/pull/27502#discussion_r2584189063

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
findepi: > I understand that it can detect schema changes in the base tables or changes to the MV definer’s access permissions - but is that the intended behavior?

i think that was intended behavior...




> Is analysis of the underlying query even necessary for fresh MVs, regardless of the `WHEN STALE` clause?

... however i think  it should be skipped. this would be significant benefit when MVs are used in low latency scenarios. 
in fact, for an MV with 24h grace period that was refreshed with last day, we should skip this line during a SELECT query, because it doesn't matter
https://github.com/trinodb/trino/blob/0ce5929a237c92a0e1d50270251213e6b9900908/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java#L3906

let's file an issue for this and follow-up separately

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -1479,9 +1479,6 @@ protected Scope visitCreateMaterializedView(CreateMaterializedView node, Optiona
                 throw semanticException(NOT_SUPPORTED, node, "'CREATE OR REPLACE' and 'IF NOT EXISTS' clauses can not be used together");
             }
             node.getGracePeriod().ifPresent(gracePeriod -> analyzeExpression(gracePeriod, Scope.create()));
-            if (node.getWhenStaleBehavior().isPresent()) {
-                throw new TrinoException(NOT_SUPPORTED, "WHEN STALE is not supported yet");
-            }
```
findepi: Not sure whether `Add when stale behavior in ConnectorMaterializedViewDefinition` commit should be removing those lines.

piotrrzysko: This is needed to test whether `MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR` is respected, which happens during execution: https://github.com/trinodb/trino/blob/d9e7986feff0d5e9a0ae099475d4133a6e54276c/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java#L163-L165

One line below, I throw `NOT_SUPPORTED` even if a connector claims to support WHEN STALE: 
https://github.com/trinodb/trino/blob/d9e7986feff0d5e9a0ae099475d4133a6e54276c/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java#L166

PR #27359, opened 2025-11-18
Add iceberg.default-new-tables-gc.enabled config property

sopel39:
`iceberg.default-new-tables-gc.enabled` will set `gc.enabled` flag for new tables created by Trino.

## Release notes

```markdown
## Iceberg
* Honor and allow to set default value for `gc.enabled` flag ({issue}`issuenumber`)
```


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -567,6 +569,19 @@ public IcebergConfig setAllowedExtraProperties(List<String> allowedExtraProperti
         return this;
     }
 
+    public boolean isDefaultNewTablesGcEnabled()
+    {
+        return defaultNewTablesGcEnabled;
+    }
+
+    @Config("iceberg.default-new-tables-gc.enabled")
```
raunaqmorarka, line 577: `iceberg.default-new-tables.gc.enabled`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -955,10 +957,15 @@ public static Map<String, String> createTableProperties(ConnectorTableMetadata t
             propertiesBuilder.put(TABLE_COMMENT, tableMetadata.getComment().get());
         }
 
-        Map<String, String> baseProperties = propertiesBuilder.buildOrThrow();
         Map<String, String> extraProperties = IcebergTableProperties.getExtraProperties(tableMetadata.getProperties()).orElseGet(ImmutableMap::of);
+        verifyExtraProperties(propertiesBuilder.buildOrThrow().keySet(), extraProperties, allowedExtraProperties);
+
+        // If user doesn't set gc.enabled, we need to set it to defaultNewTablesGcEnabled value
+        if (!extraProperties.containsKey(GC_ENABLED) && defaultNewTablesGcEnabled != GC_ENABLED_DEFAULT) {
+            propertiesBuilder.put(GC_ENABLED, Boolean.toString(defaultNewTablesGcEnabled));
```
raunaqmorarka, line 965: One thing to note about relying on GC_ENABLED property of iceberg is that it will also disallow expire_snapshots.
Is that something desirable ?
I can imagine that some users just want protection on DROP, but with this they can't run the regular maintenance operation of expiring old snapshots

sopel39: I guess they could enable GC for `expire_snapshots`. Users often have side spark anyway. Unless there is another property that would just limit DROP, then I think it's reasonable to expect data admins to know how to handle this situation.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -219,7 +219,7 @@ protected IcebergQueryRunner.Builder createQueryRunnerBuilder()
                 .setIcebergProperties(ImmutableMap.<String, String>builder()
                         .put("iceberg.file-format", format.name())
                         // Only allow some extra properties. Add "sorted_by" so that we can test that the property is disallowed by the connector explicitly.
-                        .put("iceberg.allowed-extra-properties", "extra.property.one,extra.property.two,extra.property.three,sorted_by")
+                        .put("iceberg.allowed-extra-properties", "extra.property.one,extra.property.two,extra.property.three,sorted_by,gc.enabled")
```
ebyhr, line 222: We use `BaseIcebergConnectorSmokeTest` or `BaseTrinoCatalogTest` when we verify catalog behavior. I recommend adding at least one positive test case to `BaseIcebergConnectorSmokeTest`.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5129,6 +5129,96 @@ public void testSplitPruningForFilterOnNonPartitionColumn()
         }
     }
 
+    @Test
+    public void testDefaultGcEnabledTablePropertyOnCreateAndCtas() throws IOException
```
ebyhr, line 5133: nit: Put `throws IOException` on a new line. Same for other tests.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
ebyhr, line 577: Could you document this property in `iceberg.md`?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
raunaqmorarka: There is a check in iceberg library `org.apache.iceberg.RemoveSnapshots#RemoveSnapshots`
```
ValidationException.check(
        PropertyUtil.propertyAsBoolean(base.properties(), GC_ENABLED, GC_ENABLED_DEFAULT),
        "Cannot expire snapshots: GC is disabled (deleting files may corrupt other tables)");
```
Using Spark or something else won't help with that.
I'm not sure how admin deals with this, adding and removing property before/after every expire_snapshots is cumbersome and removes the DROP protection for the duration of the command. Also, most ppl have automated maintenance services, and we wouldn't want ppl to have to go and modify that.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -95,6 +96,7 @@ public class IcebergConfig
     private int planningThreads = Math.min(Runtime.getRuntime().availableProcessors(), 16);
     private int fileDeleteThreads = Runtime.getRuntime().availableProcessors() * 2;
     private List<String> allowedExtraProperties = ImmutableList.of();
+    private boolean defaultNewTablesGcEnabled = GC_ENABLED_DEFAULT;
```
findinpath, line 99: The common approach is to restrain referencing in the config files values from dependent libraries because this may trigger a silent change in the default configuration of the connector.
Please consider using rather `true` instead

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -567,6 +569,19 @@ public IcebergConfig setAllowedExtraProperties(List<String> allowedExtraProperti
         return this;
     }
 
+    public boolean isDefaultNewTablesGcEnabled()
+    {
+        return defaultNewTablesGcEnabled;
+    }
+
+    @Config("iceberg.default-new-tables-gc.enabled")
+    @ConfigDescription("Default value for Iceberg property gc.enabled when creating new tables")
```
findinpath, line 578: Let's add as well corresponding documentation in `iceberg.md` as well.
I noticed that the documentation for `gc.enabled` (introduced by https://github.com/apache/iceberg/pull/1796 )  table property is unfortunately missing as well on `apache/iceberg` - As a reference point you could use https://github.com/apache/iceberg/pull/9231


> Property to disable garbage collection operations such as expiring snapshots or removing orphan files

It seems that in the meantime the table property has as well repercussions on whether to delete dropped table's files as well.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/glue/TrinoGlueCatalog.java
```diff
@@ -707,7 +710,9 @@ public void dropTable(ConnectorSession session, SchemaTableName schemaTableName)
             // So log the exception and continue with deleting the table location
             LOG.warn(e, "Failed to delete table data referenced by metadata");
         }
-        deleteTableDirectory(fileSystemFactory.create(session), schemaTableName, table.location());
+        if (propertyAsBoolean(table.properties(), GC_ENABLED, GC_ENABLED_DEFAULT)) {
```
findinpath, line 713: Pls replace `GC_ENABLED_DEFAULT` with `true` - we don't have control on the value of `GC_ENABLED_DEFAULT` when the iceberg library gets updated.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/TrinoHiveCatalog.java
```diff
@@ -436,7 +439,9 @@ public void dropTable(ConnectorSession session, SchemaTableName schemaTableName)
             // So log the exception and continue with deleting the table location
             log.warn(e, "Failed to delete table data referenced by metadata");
         }
-        deleteTableDirectory(fileSystemFactory.create(session), schemaTableName, metastoreTable.getStorage().getLocation());
+        if (propertyAsBoolean(metadata.properties(), GC_ENABLED, GC_ENABLED_DEFAULT)) {
```
findinpath, line 442:
```suggestion
        if (propertyAsBoolean(metadata.properties(), GC_ENABLED, true)) {
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/jdbc/TrinoJdbcCatalog.java
```diff
@@ -348,7 +351,9 @@ public void dropTable(ConnectorSession session, SchemaTableName schemaTableName)
             // So log the exception and continue with deleting the table location
             LOG.warn(e, "Failed to delete table data referenced by metadata");
         }
-        deleteTableDirectory(fileSystemFactory.create(session), schemaTableName, table.location());
+        if (propertyAsBoolean(table.properties(), GC_ENABLED, GC_ENABLED_DEFAULT)) {
```
findinpath, line 354:
```suggestion
        if (propertyAsBoolean(table.properties(), GC_ENABLED, true)) {
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/AbstractTrinoCatalog.java
```diff
@@ -312,7 +313,7 @@ protected Location createMaterializedViewStorage(
         Schema schema = schemaFromMetadata(columns);
         PartitionSpec partitionSpec = parsePartitionFields(schema, getPartitioning(materializedViewProperties));
         SortOrder sortOrder = parseSortFields(schema, getSortOrder(materializedViewProperties));
-        Map<String, String> properties = createTableProperties(new ConnectorTableMetadata(storageTableName, columns, materializedViewProperties, Optional.empty()), _ -> false);
+        Map<String, String> properties = createTableProperties(new ConnectorTableMetadata(storageTableName, columns, materializedViewProperties, Optional.empty()), _ -> false, GC_ENABLED_DEFAULT);
```
findinpath, line 316:
```suggestion
        Map<String, String> properties = createTableProperties(new ConnectorTableMetadata(storageTableName, columns, materializedViewProperties, Optional.empty()), _ -> false, true);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/AbstractTrinoCatalog.java
```diff
@@ -350,7 +351,7 @@ protected SchemaTableName createMaterializedViewStorageTable(
         ConnectorTableMetadata tableMetadata = new ConnectorTableMetadata(storageTable, columns, materializedViewProperties, Optional.empty());
         String tableLocation = getTableLocation(tableMetadata.getProperties())
                 .orElseGet(() -> defaultTableLocation(session, tableMetadata.getTable()));
-        Transaction transaction = IcebergUtil.newCreateTableTransaction(this, tableMetadata, session, false, tableLocation, _ -> false);
+        Transaction transaction = IcebergUtil.newCreateTableTransaction(this, tableMetadata, session, false, tableLocation, _ -> false, GC_ENABLED_DEFAULT);
```
findinpath, line 354:
```suggestion
        Transaction transaction = IcebergUtil.newCreateTableTransaction(this, tableMetadata, session, false, tableLocation, _ -> false, true);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -869,7 +871,7 @@ public static List<ViewColumn> viewColumnsFromSchema(TypeManager typeManager, Sc
                 .toList();
     }
 
-    public static Transaction newCreateTableTransaction(TrinoCatalog catalog, ConnectorTableMetadata tableMetadata, ConnectorSession session, boolean replace, String tableLocation, Predicate<String> allowedExtraProperties)
+    public static Transaction newCreateTableTransaction(TrinoCatalog catalog, ConnectorTableMetadata tableMetadata, ConnectorSession session, boolean replace, String tableLocation, Predicate<String> allowedExtraProperties, boolean defaultNewTablesGcEnabled)
```
findinpath, line 874: nit: Please separate this long list of parameters on new lines for increased readability.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/glue/TestTrinoGlueCatalog.java
```diff
@@ -152,7 +153,8 @@ public void testNonLowercaseGlueDatabase()
                     newDirectExecutorService(),
                     directExecutor(),
                     newDirectExecutorService(),
-                    newDirectExecutorService());
+                    newDirectExecutorService(),
+                    GC_ENABLED_DEFAULT);
```
findinpath, line 157:
```suggestion
                    true);
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/nessie/TestTrinoNessieCatalog.java
```diff
@@ -210,7 +211,8 @@ public void testNonLowercaseNamespace()
                     newDirectExecutorService(),
                     directExecutor(),
                     newDirectExecutorService(),
-                    newDirectExecutorService());
+                    newDirectExecutorService(),
+                    GC_ENABLED_DEFAULT);
```
findinpath, line 215:
```suggestion
                    true);
```

findinpath:
> `iceberg.default-new-tables-gc.enabled` will set `gc.enabled` flag for new tables created by Trino.

Why should the Trino Iceberg connector be proactive in setting the `gc.enabled` field on the tables created?


Assuming that we find a good rationale on the question above, I suggest rather to handle the implementation of the config property rather differently.
Take `hive.max-splits-per-second` as a proof of concept 


```

    @Nullable
    public Boolean getDefaultGarbageCollection()
    {
        return defaultGarbageCollection;
    }

    @Config("iceberg.default-garbage-collection")
    @ConfigDescription("Default value for Iceberg property gc.enabled when creating new tables")
    public IcebergConfig setDefaultGarbageCollection(Boolean defaultGarbageCollection)
    {
        this.defaultGarbageCollection = defaultGarbageCollection;
        return this;
    }
```

`iceberg.object-store-layout.enabled` - although applies for new tables, does not have `new-table` token in the property name.

`iceberg.default-garbage-collection` can be `true`, `false` or not specified.

sopel39:
> Why should the Trino Iceberg connector be proactive in setting the gc.enabled field on the tables created?

It's not proactive. It only sets it when the config toggle is different than `GC_ENABLED`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5129,6 +5129,96 @@ public void testSplitPruningForFilterOnNonPartitionColumn()
...
+        String ctasTable = "test_gc_enabled_ctas_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + ctasTable + " AS SELECT 1 x", 1);
+        String ctasTableLocation = getTableLocation(ctasTable);
+        String ctasMetadataLocation = getLatestMetadataLocation(fileSystem, ctasTableLocation);
+        TableMetadata ctasMetadata = TableMetadataParser.read(FILE_IO_FACTORY.create(fileSystem), ctasMetadataLocation);
+        assertThat(ctasMetadata.properties().getOrDefault("gc.enabled", "")).isEqualTo("");
+        assertThat(ctasMetadata.properties().getOrDefault("gc.enabled", "true")).isEqualTo("true");
+
+        // collect data files and verify they are removed on drop
+        var ctasDataFiles = getAllDataFilesFromTableDirectory(ctasTable);
```
chenjian2664, line 5145: Can we avoid to use `var` ?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5129,6 +5129,96 @@ public void testSplitPruningForFilterOnNonPartitionColumn()
...
+    @Test
+    public void testDefaultGcEnabledTablePropertyOnCreateAndCtas() throws IOException
+    {
+        // CREATE TABLE AS SELECT
+        String ctasTable = "test_gc_enabled_ctas_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + ctasTable + " AS SELECT 1 x", 1);
+        String ctasTableLocation = getTableLocation(ctasTable);
+        String ctasMetadataLocation = getLatestMetadataLocation(fileSystem, ctasTableLocation);
+        TableMetadata ctasMetadata = TableMetadataParser.read(FILE_IO_FACTORY.create(fileSystem), ctasMetadataLocation);
+        assertThat(ctasMetadata.properties().getOrDefault("gc.enabled", "")).isEqualTo("");
```
chenjian2664, line 5141: use `isEmpty()`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
chenjian2664: It looks like a breaking change to `expire_snapshots`?

But if other engine disable the gc flag, it's also a problem to Trino, user not able to expire snapshots due a unsupported or exposed property

sopel39:
After discussion with @raunaqmorarka and @ebyhr we figured that `gc.enabled` might not be the right approach, because it prevents both data delete on table drop and maintainable ops (e.g. expire snapshots).

Alternatives are:
1. Trino specific property that prevents data delete on table drop in Trino only, but allows maintenance operations.
2. Iceberg native property that prevents data delete on table drop across all engines, but allows maintenance operations.

PR #18333, opened 2023-07-19
Support `CASCADE` option for `DROP SCHEMA` statement in Delta Lake

ebyhr:
## Release notes

(x) Release notes are required, with the following suggested text:

```markdown
# Delta Lake
* Support `CASCADE` option for `DROP SCHEMA` statement. ({issue}`18333`)
```

Fixes #18333

PR #27361, opened 2025-11-19
Ignore non-existing table when dropping Iceberg schema with cascade

ebyhr:
## Description

The `getTableHandle` method returns null if the table is dropped concurrently during the loop. 

## Release notes

```markdown
## Iceberg
* Fix potential failure when dropping a schema with cascade. ({issue}`27361`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -1171,7 +1171,11 @@ public class IcebergMetadata
                 dropView(session, viewName);
             }
             for (SchemaTableName tableName : listTables(session, Optional.of(schemaName))) {
-                dropTable(session, getTableHandle(session, tableName, Optional.empty(), Optional.empty()));
+                ConnectorTableHandle tableHandle = getTableHandle(session, tableName, Optional.empty(), Optional.empty());
+                if (tableHandle != null) {
+                    // getTableHandle method returns null if the table is dropped concurrently
+                    dropTable(session, tableHandle);
+                }
             }
         }
         catalog.dropNamespace(session, schemaName);


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1171,7 +1171,10 @@ public void dropSchema(ConnectorSession session, String schemaName, boolean casc
                 dropView(session, viewName);
             }
             for (SchemaTableName tableName : listTables(session, Optional.of(schemaName))) {
-                dropTable(session, getTableHandle(session, tableName, Optional.empty(), Optional.empty()));
+                ConnectorTableHandle tableHandle = getTableHandle(session, tableName, Optional.empty(), Optional.empty());
+                if (tableHandle != null) {
```
findepi, line 1175: Worth code comment

raunaqmorarka: +1

chenjian2664: Why view not have such issue?

Math-ias: ~Reposting to main PR thread~. I think views don't have that issue because it's a quick update to remove them from the metastore. Files connected to an iceberg table will take longer to delete and that increases risk of the table disappearing underneath.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1171,7 +1171,10 @@ public void dropSchema(ConnectorSession session, String schemaName, boolean casc
...
+                    dropTable(session, tableHandle);
```
Math-ias, line 1177: So to understand the motivation of the change correctly ... Looks like `dropTable` is potentially blocking (for a while) the thread?

<details>
<summary>following method calls in github</summary>
<ul>
<li><a href="https://github.com/trinodb/trino/blob/5f5f506a22158b4a8328d5824d6fd06a904cb939/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/TrinoHiveCatalog.java#L432">here in the dropTable implementation of TrinoHiveCatalog</a></li>
<li><a href="https://github.com/apache/iceberg/blob/6e873baa7db22429fa231d552a696248f64ea1f7/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L127-L144">to here in iceberg's CatalogUtils.dropTableData</a></li>
<li><a href="https://github.com/apache/iceberg/blob/6e873baa7db22429fa231d552a696248f64ea1f7/core/src/main/java/org/apache/iceberg/CatalogUtil.java#L165">to here in one of iceberg's `deleteFiles` implementations</a> looks like it's reading manifest files to delete the files they reference recursively</li>
<li><a href="https://github.com/apache/iceberg/blob/6e873baa7db22429fa231d552a696248f64ea1f7/core/src/main/java/org/apache/iceberg/util/Tasks.java#L357-L359">to iceberg `Tasks` here to see how those tasks get run</a></li>
<li><a href="https://github.com/apache/iceberg/blob/6e873baa7db22429fa231d552a696248f64ea1f7/core/src/main/java/org/apache/iceberg/util/Tasks.java#L474">to this implementation of `waitFor` in iceberg</a> that shows a <code>while (true)</code> loop running, and sleeping 10 milliseconds to wait for every task to get done. That's blocking.</li>
</details>

Therefore tables we've listed from metadata can get dropped from beneath this thread as a `DROP SCHEMA schema_name CASCADE` is completing ...

I'm guessing the risk of dropping a table taking so long is mostly when there's externally managed files. The delta lake connector should be comparable. [In your implementation of CASCADE there in the delta lake connector you use an additional `try` on dropTable even after we've checked for tableHandle existence](https://github.com/trinodb/trino/pull/18333/files#diff-79f367e47040dda32ac81ecec46b27cfd7cc822858bc8c31fe435a15a47446dcR795). Is that worth bringing back to this PR as well for extra safety?

ebyhr: (I didn't read this message carefully before merging the PR). Each Iceberg catalog maintains its own table metadata cache, so the first `TrinoCatalog.loadTable` call from `IcebergMetadata.getTableHandle` populates the cache, and the second `TrinoCatalog.loadTable` in `TrinoCatalog.dropTable` will retrieve the object from the cache. We can still add the catch block, though.

PR #27362, opened 2025-11-19
Add `retain_last` and `clean_expired_metadata` options to Iceberg `expire_snapshots` table procedure

### LINKED ISSUE #27357: Iceberg expire_snapshot parameters support
There is only one parameter available for _expire_snapshots_, which is _retention_threshold_.
In our use case, we have over 200+ schema changes, all the columns are commented and our metadata file is getting huge.
Trino is the only tool we use for all of our needs. 
Is it safe and possible to add support for more parameters for _expire_snapshots_? i.e.: _clean_expired_metadata_

Thanks

ebyhr:
## Description

Fixes #27357

## Release notes

```markdown
## Iceberg
* Add `retain_last` and `clean_expired_metadata` options to `expire_snapshots` command. ({issue}`27357`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -906,6 +906,10 @@ The command accepts an optional `retain_last` parameter to specify the minimum
 number of ancestor snapshots to preserve (defaults to 1), regardless of the
 `retention_threshold` value.
 
+The command accepts an optional `clean_expired_metadata` parameter (defaults to false).
+When true, cleans up metadata such as partition specs and schemas that are no
+longer referenced by snapshots.
+
 (iceberg-remove-orphan-files)=
 ##### remove_orphan_files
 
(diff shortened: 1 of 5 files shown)


docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -906,6 +906,10 @@ The command accepts an optional `retain_last` parameter to specify the minimum
 number of ancestor snapshots to preserve (defaults to 1), regardless of the
 `retention_threshold` value.
 
+The command accepts an optional `clean_expired_metadata` parameter (defaults to false).
+When true, cleans up metadata such as partition specs and schemas that are no
+longer referenced by snapshots.
+
```
raunaqmorarka, line 912: Why don't we just do this always by default ?
I believe the only reason this is not default in upstream iceberg is that its a new feature and they wanted to be conservative about enabling it https://lists.apache.org/thread/5xpwcnh6b8br9r255n9ovso66rog0bkw
I think we could retain the flag but make it enabled by default.
If doesn't result in issues, then probably we can remove the flag after a while as well

ebyhr: I was a little concerned about the additional overhead of reading all the manifests: https://github.com/apache/iceberg/blob/b6b89268de26b119817db9ce20b540baa349c541/core/src/main/java/org/apache/iceberg/RemoveSnapshots.java#L228-L237

I'll turn it on by default if there are no objections from other reviewers.

raunaqmorarka: That could indeed be a problem.
Without clean_expired_metadata, I guess expire snapshots would read only manifests belonging to old snapshots + the oldest surviving one right ?

raunaqmorarka (approved):

docs/src/main/sphinx/connector/iceberg.md
findepi: It's safer to introduce this off by default, test things out and enable after same time (IF we build confidence it's safe to do so)

PR #27364, opened 2025-11-19
Add ILIKE operator for case-insensitive pattern matching

### LINKED ISSUE #2491: Implement ILIKE predicate
Comparing case-insensitive strings can be source for confusion.
we must use lower() on both sides of the query like shown:
https://stackoverflow.com/questions/42850329/how-to-configure-presto-searches-to-be-case-insensitive

Some DBs like PostgreSQL extended to support also ILIKE 

> The key word ILIKE can be used instead of LIKE to make the match case-insensitive according to the active locale. This is not in the SQL standard but is a PostgreSQL extension.
> 
> The operator ~~ is equivalent to LIKE, and ~~ * corresponds to ILIKE. There are also !~~ and !~~* operators that represent NOT LIKE and NOT ILIKE, respectively. All of these operators are PostgreSQL-specific.

https://www.postgresql.org/docs/current/functions-matching.html

This is very useful. I hope this can be added to Presto.

bvolpato:
## Description

This PR implements the ILIKE operator for case-insensitive pattern matching, addressing #2491. ILIKE follows PostgreSQL semantics and works like LIKE but ignores case differences in both the value and pattern.

The implementation provides a native way to perform case-insensitive pattern matching without requiring manual `lower()` calls on both sides of the comparison.

Rather than creating a separate ILIKE AST node, the implementation extends the existing `LikePredicate` with a `caseInsensitive` boolean flag. This minimizes code duplication and leverages the existing pattern matching infrastructure.

**Compatibility:**
ILIKE is natively supported by PostgreSQL, Snowflake, DuckDB, Databricks, and Redshift. This implementation works with all Trino connectors - pushdown is enabled for PostgreSQL, while other connectors evaluate ILIKE in Trino's engine.

## Release notes

```markdown
## General
* Add `ILIKE` operator for case-insensitive pattern matching. ({issue}`2491`)
```

- Fixes #2491

### DIFF
--- a/core/trino-grammar/src/test/java/io/trino/grammar/sql/TestSqlKeywords.java
+++ b/core/trino-grammar/src/test/java/io/trino/grammar/sql/TestSqlKeywords.java
@@ -169,6 +169,7 @@ public class TestSqlKeywords
                         "LEFT",
                         "LEVEL",
                         "LIKE",
+                        "ILIKE",
                         "LIMIT",
                         "LISTAGG",
                         "LOCAL",
--- a/core/trino-main/src/main/java/io/trino/type/LikeFunctions.java
+++ b/core/trino-main/src/main/java/io/trino/type/LikeFunctions.java
@@ -60,7 +60,7 @@ public final class LikeFunctions
     @SqlType(LikePatternType.NAME)
     public static LikePattern likePattern(@SqlType("varchar") Slice pattern)
     {
-        return LikePattern.compile(pattern.toStringUtf8(), Optional.empty(), false);
+        return LikePattern.compile(pattern.toStringUtf8(), Optional.empty(), false, true);
     }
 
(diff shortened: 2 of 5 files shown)


findepi:
@martint can you please weight in https://github.com/trinodb/trino/issues/2491 to clear syntax review?

martint:
Since this can be easily implemented as a function, that’s the preferred approach instead of adding dedicated syntax that’s not covered by the SQL specification.

findepi:
@bvolpato please see above comment from @martint 
can you please update the PR to implement this functionality as string functions with following signatures?

```
ilike(value, patter) -> boolean
ilike(value, patter, scape) -> boolean
```

jasonxwlin:
@bvolpato Does ILIKE match implementation perform better than `lower(col) like '<pattern>'` from performance-wise perspective?

PR #27369, opened 2025-11-19
Improve RepeatFunction

raunaqmorarka:
## Description
Improve performance of RepeatFunction by using RunLengthEncodedBlock directly
instead of going through BlockBuilder

Expand the max size limits on the function

## Release notes

```markdown
## General
* Improve performance of `repeat` function. ({issue}`27369`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/RepeatFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/RepeatFunction.java
@@ -13,6 +13,7 @@
  */
 package io.trino.operator.scalar;
 
+import io.airlift.units.DataSize;
 import io.trino.spi.block.Block;
 import io.trino.spi.block.RunLengthEncodedBlock;
 import io.trino.spi.function.Description;
@@ -23,6 +24,7 @@ import io.trino.spi.function.TypeParameter;
 import io.trino.spi.type.StandardTypes;
 import io.trino.spi.type.Type;
 
+import static io.airlift.units.DataSize.Unit.MEGABYTE;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
 import static io.trino.type.UnknownType.UNKNOWN;
 import static io.trino.util.Failures.checkCondition;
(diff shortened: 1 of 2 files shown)


chenjian2664 (approved):

core/trino-main/src/main/java/io/trino/operator/scalar/RepeatFunction.java
```diff
@@ -109,36 +89,34 @@ public static Block repeat(
...
         }
-        return blockBuilder.build();
+        return RunLengthEncodedBlock.create(type, element, toIntExact(count));
     }
 
-    private static BlockBuilder createBlockBuilder(Type type, long count)
+    private static void checkCountConditions(long count)
     {
-        checkCondition(count <= MAX_RESULT_ENTRIES, INVALID_FUNCTION_ARGUMENT, "count argument of repeat function must be less than or equal to 10000");
+        checkCondition(count <= MAX_RESULT_ENTRIES, INVALID_FUNCTION_ARGUMENT, "count argument of repeat function must be less than or equal to " + MAX_RESULT_ENTRIES);
```
pettyjamesm: Avoid string concatenation here and use the overload that takes MAX_RESULT_ENTRIES as a formatting argument to avoid the overhead when the condition is satisfied

core/trino-main/src/main/java/io/trino/operator/scalar/RepeatFunction.java
```diff
@@ -109,36 +89,34 @@ public static Block repeat(
...
 
-    private static void checkMaxSize(long bytes, long count)
+    private static void checkMaxSize(Block block)
     {
         checkCondition(
-                bytes <= (MAX_SIZE_IN_BYTES + count) / count,
+                block.getSizeInBytes() <= MAX_SIZE_IN_BYTES,
                 INVALID_FUNCTION_ARGUMENT,
-                "result of repeat function must not take more than 1000000 bytes");
+                format("result of repeat function must not take more than %s bytes", MAX_SIZE_IN_BYTES));
```
pettyjamesm: No need to format this eagerly, use the formatting overload instead

findepi:
> Expand the max size limits on the function

Why so? I thought that the limit is there to prevent users from creating  large Pages that will are problematic for the execution.

raunaqmorarka:
> > Expand the max size limits on the function
> 
> Why so? I thought that the limit is there to prevent users from creating large Pages that will are problematic for the execution.

We're using RleBlock now instead of a BlockBuilder, so we don't incur high memory usage like in the previous code.
The new 4mb/100K entries limit is in line with `io.trino.operator.scalar.BlockSet#MAX_FUNCTION_MEMORY`, `io.trino.plugin.deltalake.DeltaLakePageSourceProvider#MAX_ROW_ID_POSITIONS` etc. and there as a safety net.

PR #27373, opened 2025-11-19
Do not buffer response during HttpPageBufferClient execution

wendigo:
HttpClient's executeAsync always buffers the entire response before handing it over to the ResponseHandler. This is inefficent as the decoding could be streaming and that doesn't require large amount of memory while reading results.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/HttpPageBufferClient.java
+++ b/core/trino-main/src/main/java/io/trino/operator/HttpPageBufferClient.java
@@ -64,7 +64,7 @@ import static com.google.common.base.Strings.isNullOrEmpty;
 import static com.google.common.base.Strings.nullToEmpty;
 import static com.google.common.net.HttpHeaders.CONTENT_TYPE;
 import static com.google.common.util.concurrent.MoreExecutors.listeningDecorator;
-import static io.airlift.concurrent.Threads.virtualThreadsNamed;
+import static io.airlift.concurrent.Threads.daemonThreadsNamed;
 import static io.airlift.http.client.HttpStatus.NO_CONTENT;
 import static io.airlift.http.client.HttpStatus.familyForStatusCode;
 import static io.airlift.http.client.Request.Builder.prepareDelete;
@@ -95,7 +95,7 @@ import static io.trino.util.Failures.WORKER_NODE_ERROR;
 import static java.lang.String.format;
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
-import static java.util.concurrent.Executors.newThreadPerTaskExecutor;
+import static java.util.concurrent.Executors.newCachedThreadPool;
 import static java.util.concurrent.TimeUnit.NANOSECONDS;
 import static java.util.concurrent.TimeUnit.SECONDS;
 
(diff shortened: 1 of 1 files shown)


core/trino-main/src/test/java/io/trino/operator/TestDirectExchangeClient.java
```diff
@@ -94,6 +95,8 @@
 
 @TestInstance(PER_CLASS)
 @Execution(CONCURRENT)
+// TODO: reenable once https://github.com/airlift/airlift/pull/1646 lands
+@Disabled
```
findepi: https://github.com/airlift/airlift/pull/1646 is closed. wrong link?

findepi: We cannot disable this test.

core/trino-main/src/test/java/io/trino/operator/TestHttpPageBufferClient.java
```diff
@@ -76,6 +77,7 @@
 
 @TestInstance(PER_CLASS)
 @Execution(CONCURRENT)
+@Disabled
```
findepi: We cannot disable this test.

findepi:
@wendigo and I talked about this and here is the plan

1. ~[@wendigo]~ `ThrowingLimitingInputStream`  from https://github.com/airlift/airlift/pull/1659 will become separate PR so that can focus on InputStream implementation correctness and test coverage - https://github.com/airlift/airlift/pull/1709
1. we gonna apply this to enforce content length as configured also in http client synchronous API. today it seems not enforced - https://github.com/airlift/airlift/pull/1709
1. [@findepi] we gonna see whether we can improve airlift http client synchronous API so that i can provide cancellation - https://github.com/airlift/airlift/pull/1742
1. if all above goes well, we will be able to use airlift http client synchronous API pretty much like this PR proposes
   we should also deprecate airlift http client async API because "executeAsync always buffers the entire response before handing it over to the ResponseHandler" (it actually needs 2x the response size memory and copies the data twice before letting the handler consume it)
   - or maybe we can improve it to the point we don't want to remove it anymore
      - https://github.com/airlift/airlift/pull/1740 + TODOs added in that PR
      - considerations in https://github.com/trinodb/trino/issues/27980
      - https://github.com/jetty/jetty.project/issues/14373

raunaqmorarka:
> we gonna apply this to enforce content length as configured also in http client synchronous API. today it seems not enforced

The issue from user perspective is that queries with large values/rows have started to fail, which is a regression from before. Imposing the limit more strongly would still cause query failures for the user. So we should likely raise the default by enough to restore successful run of now failing queries. We should also document `http-client.max-content-length` prominently if we expect that users need to tune this higher to allow queries on large values/rows.

findepi:
> The issue from user perspective is that queries with large values/rows have started to fail, which is a regression from before.

Agreed it's a regression from user perspective.

If, however, something as some point works because safety is off, it cannot be construed as a non-negligible argument to keep safety off at all times.

I started discussion on slack yesterday with hope to get more clarity on what the `http-client.max-content-length` behavior should be:
- https://trinodb.slack.com/archives/C07ABNN828M/p1767117434862999

findepi:
(rebased)

wendigo:
```
2026-01-26T05:31:48.213-0600	INFO	page-buffer-client-callback-3	stderr	Exception in thread "page-buffer-client-callback-3"
2026-01-26T05:31:48.213-0600	INFO	page-buffer-client-callback-3	stderr	java.lang.NullPointerException: Cannot invoke "String.contains(java.lang.CharSequence)" because the return value of "java.lang.Throwable.getMessage()" is null
```

findepi:
The `iceberg/sf1000_parquet_unpart` benchmark run didn't show significant diffrences.
The two `iceberg/sf1000_parquet_part` runs shows significant higher whole benchmark run, and also regressions for selected queries (but different in each run).

image

image

I will ask someone smarter to confirm i am reading results the right way.

findepi:
The switch from virtual threads to system mostly eliminated change in overall benchmark run time, but there still are changes on individual queries

image

findepi:
I run benchmarks again, this time taking same baseline master commit and a PR with no change at all:
- https://github.com/trinodb/trino/pull/28051

This produced results pretty close to the ones with system threads (https://github.com/trinodb/trino/pull/27373#issuecomment-3820108446), and somewhat better than those with virtual threads (https://github.com/trinodb/trino/pull/27373#issuecomment-3803956829, https://github.com/trinodb/trino/pull/27373#issuecomment-3805322319).

PR #27374, opened 2025-11-19
Throw correctly from S3 createExclusive with larger data

findepi:
Before the change, the `S3OutputStream` configured to write exclusively
(using S3 conditional writes), would detect object existence and throw
`FileAlreadyExistsException` only in the case of single upload, but not
when multi-part upload is used.

### DIFF
--- a/lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
+++ b/lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
@@ -17,6 +17,7 @@ import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Ordering;
 import com.google.common.io.Closer;
 import io.airlift.slice.Slice;
+import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import io.trino.filesystem.encryption.EncryptionEnforcingFileSystem;
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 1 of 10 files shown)


raunaqmorarka (approved):

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -623,6 +623,12 @@ void testOutputFile()
                 assertThatThrownBy(() -> outputFile.createExclusive(new byte[0]))
                         .isInstanceOf(FileAlreadyExistsException.class)
                         .hasMessageContaining(tempBlob.location().toString());
+
+                // likely larger than any internal buffering or potential multipart threshold a filesystem may have
+                byte[] twentyMegabytes = new byte[20 * MEGABYTE];
```
losipiuk: Is it really big enough or should be more than 32

findepi: the S3-related test classes all seem to do `.setStreamingPartSize(DataSize.valueOf("5.5MB"))`
so this value is actually technically OK, but it's not very apparent.

how can i improve?

losipiuk: Maybe you can put this 20 as a constant and then add a test in S3 related  test classes which assert that configured value is lower than the constant defined here. Not sure if that is extremly beneficial

findepi: 💡

PR #15760, opened 2023-01-18
Optimize null representation in encoded VariableBlockWidthBlock

radek-starburst:
## Description

Currently, when block `VariableWidthBlock` is encoded it is writing an array of offsets for all positions regardless of the fact whether a position is null or not. Instead we could save the lengths only for non-null positions and compute offsets from an array of lengths and the array of nullability (array that determines whether position is `null` or not).

This change should be tested by `io.trino.spi.block.TestVariableWidthBlockEncoding`.

The difference was tested on query with different value of `X` to have a control on null frequency
```
with cs as (select *,  case when rand(100) < X then null else '0' end nullek from catalog_sales cs)
SELECT  count_if(cs.nullek is null), count(*), cs.nullek FROM cs
RIGHT JOIN call_center cc ON cc.cc_call_center_sk =  cs.cs_call_center_sk
GROUP BY 3
ORDER BY 1, 2;
```

## Results (cumulative size of exchanged GB via network) 

```
           No nullw   50% of nulls    99% of nulls   
baseline   28GB      27GB            26GB                    
change     28GB      24GB            21 GB
```  

Let's wait for benchmarks results.

## Release notes

(*) This is not user-visible or docs only and no release notes are required.
( ) Release notes are required, please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #27377, opened 2025-11-19
Improve VariableWidthBlock encoding

pettyjamesm:
## Description
Avoids converting between offsets and lengths when serializing and deserializing VariableWidthBlock instances, which enables a fast-path conversion for blocks without nulls present. When nulls are present, the compaction and expansion of offsets still outperforms the length to offset conversion.

Also removes an additional `int` per serialized block by referencing the `offsets` array to determine the slice length instead of sending it as a separate field.

### Benchmarks 
[jmh.morethan.io](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/b952c6b57c29e00823f03ca838a4b1ce/raw/7924fa875fa3de5bc0edb1505294c13e0902783a/baseline.json,https://gist.githubusercontent.com/pettyjamesm/b952c6b57c29e00823f03ca838a4b1ce/raw/32692c8d7e21e54c891052b0caa0feff993b4818/improved_v2.json) - Significant improvements
```
# Before
Benchmark                                   (nullChance)  (offset)  Mode  Cnt  Score   Error  Units
BenchmarkBlockSerde.deserializeSliceDirect             0      true  avgt   10  1.139 ± 0.134  ns/op
BenchmarkBlockSerde.deserializeSliceDirect             0     false  avgt   10  1.097 ± 0.056  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .01      true  avgt   10  1.476 ± 0.026  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .01     false  avgt   10  2.066 ± 0.081  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .10      true  avgt   10  2.196 ± 0.038  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .10     false  avgt   10  2.228 ± 0.197  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .50      true  avgt   10  1.251 ± 0.013  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .50     false  avgt   10  1.234 ± 0.021  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .90      true  avgt   10  1.434 ± 0.023  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .90     false  avgt   10  1.624 ± 0.132  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .99      true  avgt   10  1.423 ± 0.069  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .99     false  avgt   10  1.443 ± 0.011  ns/op
BenchmarkBlockSerde.serializeSliceDirect               0      true  avgt   10  0.659 ± 0.010  ns/op
BenchmarkBlockSerde.serializeSliceDirect               0     false  avgt   10  0.645 ± 0.001  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .01      true  avgt   10  1.151 ± 0.039  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .01     false  avgt   10  1.162 ± 0.017  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .10      true  avgt   10  1.104 ± 0.040  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .10     false  avgt   10  1.174 ± 0.008  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .50      true  avgt   10  1.002 ± 0.014  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .50     false  avgt   10  0.962 ± 0.012  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .90      true  avgt   10  0.869 ± 0.001  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .90     false  avgt   10  0.901 ± 0.059  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .99      true  avgt   10  0.868 ± 0.012  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .99     false  avgt   10  0.796 ± 0.011  ns/op


# After
Benchmark                                   (nullChance)  (offset)  Mode  Cnt  Score   Error  Units
BenchmarkBlockSerde.deserializeSliceDirect             0      true  avgt   10  0.648 ± 0.013  ns/op
BenchmarkBlockSerde.deserializeSliceDirect             0     false  avgt   10  0.665 ± 0.013  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .01      true  avgt   10  1.302 ± 0.071  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .01     false  avgt   10  1.302 ± 0.017  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .10      true  avgt   10  1.267 ± 0.030  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .10     false  avgt   10  1.282 ± 0.074  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .50      true  avgt   10  0.973 ± 0.002  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .50     false  avgt   10  0.973 ± 0.017  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .90      true  avgt   10  0.743 ± 0.008  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .90     false  avgt   10  0.760 ± 0.058  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .99      true  avgt   10  0.688 ± 0.001  ns/op
BenchmarkBlockSerde.deserializeSliceDirect           .99     false  avgt   10  0.699 ± 0.021  ns/op
BenchmarkBlockSerde.serializeSliceDirect               0      true  avgt   10  0.538 ± 0.010  ns/op
BenchmarkBlockSerde.serializeSliceDirect               0     false  avgt   10  0.399 ± 0.033  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .01      true  avgt   10  1.035 ± 0.041  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .01     false  avgt   10  1.034 ± 0.013  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .10      true  avgt   10  1.003 ± 0.005  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .10     false  avgt   10  1.007 ± 0.004  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .50      true  avgt   10  0.909 ± 0.071  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .50     false  avgt   10  0.858 ± 0.014  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .90      true  avgt   10  0.754 ± 0.004  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .90     false  avgt   10  0.755 ± 0.001  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .99      true  avgt   10  0.725 ± 0.040  ns/op
BenchmarkBlockSerde.serializeSliceDirect             .99     false  avgt   10  0.718 ± 0.019  ns/op

```

## Release notes

```markdown
## Section
* Improve performance of data exchanges involving variable width data ({issue}`27377`)
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
@@ -60,9 +60,7 @@ public class VariableWidthBlockEncoding
         int startingOffset = rawOffsets[arrayBaseOffset];
         int totalLength = rawOffsets[positionCount + arrayBaseOffset] - startingOffset;
 
-        sliceOutput
-                .appendInt(totalLength)
-                .writeBytes(variableWidthBlock.getRawSlice(), startingOffset, totalLength);
+        sliceOutput.writeBytes(variableWidthBlock.getRawSlice(), startingOffset, totalLength);
     }
 
     @Override
@@ -73,8 +71,8 @@ public class VariableWidthBlockEncoding
 
         int[] offsets = readOffsetsWithNullsCompacted(sliceInput, valueIsNull, positionCount);
 
-        int blockSize = sliceInput.readInt();
-        Slice slice = Slices.allocate(blockSize);
+        int sliceSize = offsets[offsets.length - 1];
+        Slice slice = Slices.allocate(sliceSize);
         sliceInput.readBytes(slice);
 
         return new VariableWidthBlock(0, positionCount, slice, offsets, valueIsNull);
--- a/core/trino-main/src/test/java/io/trino/execution/buffer/TestPagesSerde.java
+++ b/core/trino-main/src/test/java/io/trino/execution/buffer/TestPagesSerde.java
@@ -211,14 +211,14 @@ public class TestPagesSerde
         // empty page
         Page page = new Page(builder.build());
         int pageSize = serializedSize(ImmutableList.of(VARCHAR), page);
-        assertThat(pageSize).isEqualTo(48);
+        assertThat(pageSize).isEqualTo(44);
 
         // page with one value
         VARCHAR.writeString(builder, "alice");
         pageSize = 44; // Now we have moved to the normal block implementation so the page size overhead is 44
         page = new Page(builder.build());
         int firstValueSize = serializedSize(ImmutableList.of(VARCHAR), page) - pageSize;
-        assertThat(firstValueSize).isEqualTo(8 + 5); // ending offsets + nonNullsCount + "alice"
+        assertThat(firstValueSize).isEqualTo(4 + 5); // ending offset + nonNullsCount + "alice"
 
         // page with two values
         VARCHAR.writeString(builder, "bob");


core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -277,4 +278,81 @@ static void compactLongsWithNullsScalar(SliceOutput sliceOutput, long[] values,
         sliceOutput.writeInt(compactedIndex);
         sliceOutput.writeLongs(compacted, 0, compactedIndex);
     }
+
+    static void writeOffsetsWithNullsCompacted(SliceOutput sliceOutput, int[] rawOffsets, @Nullable boolean[] valueIsNull, int baseOffset, int positionCount)
+    {
+        // inlined from Objects.checkFromIndexSize (this is necessary for performance)
```
raunaqmorarka: I don't follow why. We've been using `checkFromIndexSize` in lots of performance sensitive code paths.
This isn't even inside a hit loop, so I don't get what is the benefit of manual inlining for this.
Do JMH benchmarks actually show a significant difference from manually inlining this ?

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -19,12 +19,10 @@
...
 import jakarta.annotation.Nullable;
 
-import java.util.Arrays;
-
 import static io.trino.spi.block.EncoderUtil.decodeNullBits;
 import static io.trino.spi.block.EncoderUtil.encodeNullsAsBits;
-import static java.lang.String.format;
-import static java.util.Objects.checkFromIndexSize;
+import static io.trino.spi.block.EncoderUtil.readOffsetsWithNullsCompacted;
+import static io.trino.spi.block.EncoderUtil.writeOffsetsWithNullsCompacted;
```
raunaqmorarka: Why do these methods belong in a utils class ?
These seem specific to VariableWidthBlockEncoding and have only one usage, so seems more appropriate to leave it in this class

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -277,4 +278,81 @@ static void compactLongsWithNullsScalar(SliceOutput sliceOutput, long[] values,
...
+        }
+    }
+
+    static int[] readOffsetsWithNullsCompacted(SliceInput sliceInput, @Nullable boolean[] valueIsNull, int positionCount)
+    {
+        int offsetsLength = positionCount + 1;
+        int nonNullOffsets = sliceInput.readInt();
+        if (nonNullOffsets > offsetsLength) {
+            throw new IllegalArgumentException(format("nonNullOffsets must be <= offsetsLength, found: %s > %s", nonNullOffsets, offsetsLength));
+        }
```
raunaqmorarka: Seems we're manually inlining checkArgument
I have similar concern about it as above

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
pettyjamesm: It did, but that was halfway through iterating on the code. It no longer makes a difference so I've switched this back to using `checkFromIndexSize`.

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: I was setting the stage to come back and add a vectorized implementation (offsets are easier to vectorize) and wanted to put it with the rest of the vectorized encoder logic, but I'll move this back into `VariableWidthBlockEncoding` for now.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
pettyjamesm: This code lives in `trino-spi`, so Guava's `checkArgument` method isn't available to call.

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
pettyjamesm: For context, it mattered not because the cost of calling the method was high- but because the JIT needs to see the bounds checks in the same compilation scope as the loop body. The issue before wasn't a failure to inline exactly, it was the way that `offsetsLength` vs `positionCount + 1` were used in different places.

testing/trino-faulttolerant-tests/src/test/java/io/trino/faulttolerant/BaseFaultTolerantExecutionTest.java
```diff
@@ -37,7 +37,7 @@ public void testTableWritePreferredWritePartitioningSkewMitigation()
         @Language("SQL") String createTableSql =
                 """
                 CREATE TABLE test_table_writer_skew_mitigation WITH (%s = ARRAY['returnflag']) AS
-                SELECT orderkey, partkey, suppkey, linenumber, quantity, extendedprice, discount, tax, linestatus, shipdate, commitdate, receiptdate, shipinstruct, shipmode, comment, returnflag
```
SongChujun, line 40: Why do we want to delete column ```comment```?

pettyjamesm: The test is highly dependent on the serialized page size, so adding 4 extra bytes per serialized page breaks the test. I'm duplicating the workaround from https://github.com/trinodb/trino/pull/15760 but this is indeed a hack.

SongChujun: I understand, hacky part is from the test setup itself though.

pettyjamesm: I'm actually going to avoid changing the serialized size- we don't need to add the extra 4 bytes anyway since we can just send the non-null ending offsets and not send the starting offset of `0` in the serialized representation.

pettyjamesm: ...and it didn't work. Reintroducing the hack.

SongChujun: I actually think this is not a good fix to the spurious test, since we will probably always change block serde, another code change may fail on the case where ```comment``` is deleted, but pass on the case where ```comment``` is added back. I think we should change the assertion here instead https://github.com/trinodb/trino/blob/master/testing/trino-faulttolerant-tests/src/test/java/io/trino/faulttolerant/BaseFaultTolerantExecutionTest.java#L62C20-L62C39, the two numbers just happen to be equal and is not guaranteed to be equal on all cases.

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -117,28 +80,68 @@ public Block readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput sliceIn
...
+                    .appendInt(nonNullEndingOffsetCount)
+                    .writeInts(nonNullEndingOffsets, 0, nonNullEndingOffsetCount);
+        }
+    }
+
+    private static int[] readOffsetsWithNullsCompacted(SliceInput sliceInput, @Nullable boolean[] valueIsNull, int positionCount)
     {
-        if (lengthIndex < 0 || lengthIndex > offsets.length) {
-            throw new IllegalArgumentException(format("Invalid lengthIndex %s for offsets %s", lengthIndex, offsets.length));
+        if (valueIsNull != null && valueIsNull.length != positionCount) {
```
raunaqmorarka, line 120: Can we use `io.trino.spi.connector.Preconditions#checkArgument` ?
I see that checkArgument is privately defined at many places in trino-spi, it would be good to just move Preconditions to `io.trino.spi` package and use that everywhere in spi

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -117,28 +80,68 @@ public Block readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput sliceIn
...
+
+        int startingOffset = rawOffsets[baseOffset];
+        if (valueIsNull == null && startingOffset == 0) {
+            // No translation of offsets required, write the range of raw offsets directly to the output
+            sliceOutput
+                    .appendInt(positionCount)
+                    .writeInts(rawOffsets, baseOffset + 1, positionCount);
+        }
+        else if (valueIsNull == null) {
+            // Subtract starting offset from each offset value to translate them to start from zero, no null suppression required
```
raunaqmorarka: Is it possible to reuse `io.trino.spi.block.BlockUtil#compactOffsets` ?

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -117,28 +80,68 @@ public Block readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput sliceIn
...
+            // Translate offsets and suppress null values from the output
+            int[] nonNullEndingOffsets = new int[positionCount];
+            int nonNullEndingOffsetCount = 0;
+            for (int i = 0; i < positionCount; i++) {
+                nonNullEndingOffsets[nonNullEndingOffsetCount] = rawOffsets[i + baseOffset + 1] - startingOffset;
+                nonNullEndingOffsetCount += valueIsNull[i + baseOffset] ? 0 : 1;
+            }
+            sliceOutput
+                    .appendInt(nonNullEndingOffsetCount)
+                    .writeInts(nonNullEndingOffsets, 0, nonNullEndingOffsetCount);
```
raunaqmorarka: Can we pull out this part from the if/else branches ?
```
sliceOutput
                    .appendInt(newOffsetsCount)
                    .writeInts(newOffsets, 0, newOffsetsCount);
```

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -117,28 +80,68 @@ public Block readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput sliceIn
...
+    }
+
+    private static int[] readOffsetsWithNullsCompacted(SliceInput sliceInput, @Nullable boolean[] valueIsNull, int positionCount)
     {
-        if (lengthIndex < 0 || lengthIndex > offsets.length) {
-            throw new IllegalArgumentException(format("Invalid lengthIndex %s for offsets %s", lengthIndex, offsets.length));
+        if (valueIsNull != null && valueIsNull.length != positionCount) {
+            throw new IllegalArgumentException(format("valueIsNull length must match positionCount, found %s <> %s", valueIsNull.length, positionCount));
+        }
+        int nonNullEndingOffsets = sliceInput.readInt();
```
raunaqmorarka: The `Ending` part in `nonNullEndingOffsets` seems unnecessary. You could mention it once in a comment somewhere, but I think it is understood that VariableWdithBlock and related things work only on "end" offsets.

raunaqmorarka: nonNullOffsetsCount ?

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -117,28 +80,68 @@ public Block readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput sliceIn
...
-            lengthIndex += isNull ? 0 : 1;
-            currentOffset += isNull ? 0 : length;
-            offsets[i] = currentOffset;
+            return offsets;
         }
-        if (lengthIndex != offsets.length) {
-            throw new IllegalArgumentException(format("Failed to consume all length entries, found %s <> %s", lengthIndex, offsets.length));
+        // Shift the offsets from the end of the offsets array downwards, repeating the previous offset when nulls are encountered
+        // until no more nulls are present
+        int position = 0;
```
raunaqmorarka: move it into for loop initialization ?

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
```diff
@@ -117,28 +80,68 @@ public Block readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput sliceIn
...
+        int copyFrom = compactIndex - 1;
```
raunaqmorarka: copyFrom  -> readIndex

losipiuk (approved):

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: Not quite, `compactOffsets` leaves the initial starting offset value of `0` in the resulting array whereas this branch omits it. I think we're better off with slight duplication to emphasize that we're working only with the ending offset for each position here.

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: Renamed

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: Moved to loop initializer, this was from an earlier version that had a post-loop that validated the remainder of the `valueIsNull` array based on the final position which was then removed for performance.

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: Renamed

pettyjamesm: Renamed to remove "ending" part of the offsets variable names.

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: Only partially. I've unified this part for the final two branch paths, but still need to retain a the initial fast path portion separately for performance.

core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockEncoding.java
pettyjamesm: I'd rather leave that as a separate PR, making the usage site adjustments would make this PR quite large and the validation checks aren't too obnoxious in this one particular scenario to justify it by itself.

PR #22944, opened 2024-08-05
Add non-secret claims in Identity

raj-manvar:
## Description
This is for handling https://github.com/trinodb/trino/issues/22943 and adds the JWT claims as extraCredentials to the Identity class

## Release notes

( ) Release notes are required. Please propose a release note for me.

Issue #16539, opened 2023-03-14
Allow to disable impersonation


kokosing:
This is a feature request to introduce a config property to disallow using any form of impersonation like:
 - SECURITY DEFINER
 - user for mask and filters
 - SET SESSION AUTHORIZATION
 
That way it would be possible to provide groups during authentication. See https://github.com/trinodb/trino/pull/15669

PR #15669, opened 2023-01-11
Deprecate getting groups in Oauth2 authentication

kokosing:
Deprecate getting groups in Oauth2 authentication

This feature is not working properly and it is not possible to fix it.
The problem is widely understood impersonation:
 - view SECURITY DEFINER
 - sessionUser JDBC parameter
 - identity from view expression for column masking and row filtering

Using Oauth2 is able to provide groups only during authentication.
However it is unable to provide groups for any impersonated user.

## Release notes

- Configuration property ``http-server.authentication.oauth2.groups-field`` got deprecated and it is about to be removed. In order to discourage the usage of it, it was renamed to ``deprecated.http-server.authentication.oauth2.groups-field``.

PR #27387, opened 2025-11-20
Add method to provide Identity from Authenticator's in SPI

tbaeg:
## Description
Introduce additional methods to the SPI authenticators to directly provide an `Identity`.

### Problem
Some authentication methods can also have authorization information (i.e. - `Identity.groups` population using `JWT claims`.. also see [comments](https://github.com/trinodb/trino/pull/22944#issuecomment-2483047829)) but there is no way to provide that information to the `Identity` given the current implementation (only accepts a `Principal`).

### Solution
Allow implementing classes to provide an `Identity` directly. This allows significantly more flexibility in shaping what properties an `Identity` has and better decoupling from main.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
@@ -21,6 +21,7 @@ import jakarta.ws.rs.container.ContainerRequestContext;
 import java.security.Principal;
 import java.security.cert.X509Certificate;
 import java.util.List;
+import java.util.Optional;
 
 import static io.trino.server.security.UserMapping.createUserMapping;
 import static java.util.Objects.requireNonNull;
@@ -55,11 +56,15 @@ public class CertificateAuthenticator
         }
 
         try {
+            Optional<Identity> authenticatedIdentity = authenticatorManager.getAuthenticator().createAuthenticatedIdentity(certificates);
+            if (authenticatedIdentity.isPresent()) {
+                return userMapping.mapIdentity(authenticatedIdentity.get());
+            }
             Principal principal = authenticatorManager.getAuthenticator().authenticate(certificates);
-            String authenticatedUser = userMapping.mapUser(principal.toString());
-            return Identity.forUser(authenticatedUser)
+            Identity principalIdentity = Identity.forUser(principal.toString())
                     .withPrincipal(principal)
                     .build();
+            return userMapping.mapIdentity(principalIdentity);
         }
         catch (UserMappingException e) {
             throw new AuthenticationException(e.getMessage());
--- a/core/trino-main/src/main/java/io/trino/server/security/HeaderAuthenticator.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/HeaderAuthenticator.java
@@ -22,6 +22,7 @@ import java.security.Principal;
 import java.util.List;
 import java.util.Locale;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 
 import static com.google.common.base.Verify.verify;
@@ -38,9 +39,7 @@ public class HeaderAuthenticator
     public HeaderAuthenticator(HeaderAuthenticatorConfig authenticatorConfig, HeaderAuthenticatorManager authenticatorManager)
     {
         this.authenticatorManager = requireNonNull(authenticatorManager, "authenticatorManager is null");
-
         this.userMapping = createUserMapping(authenticatorConfig.getUserMappingPattern(), authenticatorConfig.getUserMappingFile());
-
         this.authenticatorManager.setRequired();
     }
 
@@ -52,14 +51,23 @@ public class HeaderAuthenticator
         Map<String, List<String>> lowerCasedHeaders = request.getHeaders().entrySet().stream()
                 .collect(Collectors.toMap(entry -> entry.getKey().toLowerCase(Locale.ENGLISH), Map.Entry::getValue));
 
-        for (io.trino.spi.security.HeaderAuthenticator authenticator : this.authenticatorManager.getAuthenticators()) {
+        for (io.trino.spi.security.HeaderAuthenticator authenticator : authenticatorManager.getAuthenticators()) {
             try {
-                Principal principal = authenticator.createAuthenticatedPrincipal(name -> lowerCasedHeaders.get(name.toLowerCase(Locale.ENGLISH)));
-                String authenticatedUser = this.userMapping.mapUser(principal.toString());
-
-                return Identity.forUser(authenticatedUser)
+                Optional<Identity> identity = authenticator.createAuthenticatedIdentity(name -> {
+                    String headerName = name.toLowerCase(Locale.ENGLISH);
+                    return lowerCasedHeaders.get(headerName);
+                });
+                if (identity.isPresent()) {
+                    return userMapping.mapIdentity(identity.get());
+                }
+                Principal principal = authenticator.createAuthenticatedPrincipal(name -> {
+                    String headerName = name.toLowerCase(Locale.ENGLISH);
+                    return lowerCasedHeaders.get(headerName);
+                });
+                Identity principalIdentity = Identity.forUser(principal.toString())
                         .withPrincipal(principal)
                         .build();
+                return userMapping.mapIdentity(principalIdentity);
             }
             catch (UserMappingException | AccessDeniedException e) {
                 if (exception == null) {
--- a/core/trino-main/src/main/java/io/trino/server/security/PasswordAuthenticator.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/PasswordAuthenticator.java
@@ -41,7 +41,6 @@ public class PasswordAuthenticator
     public PasswordAuthenticator(PasswordAuthenticatorManager authenticatorManager, PasswordAuthenticatorConfig config, ProtocolConfig protocolConfig)
     {
         this.userMapping = createUserMapping(config.getUserMappingPattern(), config.getUserMappingFile());
-
         this.authenticatorManager = requireNonNull(authenticatorManager, "authenticatorManager is null");
         authenticatorManager.setRequired();
         this.alternateHeaderName = protocolConfig.getAlternateHeaderName();
@@ -60,14 +59,16 @@ public class PasswordAuthenticator
         AuthenticationException exception = null;
         for (io.trino.spi.security.PasswordAuthenticator authenticator : authenticatorManager.getAuthenticators()) {
             try {
+                Optional<Identity> authenticatedIdentity = authenticator.createAuthenticatedIdentity(user, password);
+                if (authenticatedIdentity.isPresent()) {
+                    Identity mappedIdentity = userMapping.mapIdentity(authenticatedIdentity.get());
+                    return rewriteUserHeaderToMappedUser(mappedIdentity, basicAuthCredentials, request.getHeaders());
+                }
                 Principal principal = authenticator.createAuthenticatedPrincipal(user, password);
-                String authenticatedUser = userMapping.mapUser(principal.toString());
-
-                // rewrite the original "unmapped" user header to the mapped user (see method Javadoc for more details)
-                rewriteUserHeaderToMappedUser(basicAuthCredentials, request.getHeaders(), authenticatedUser);
-                return Identity.forUser(authenticatedUser)
+                Identity mappedIdentity = userMapping.mapIdentity(Identity.forUser(principal.toString())
                         .withPrincipal(principal)
-                        .build();
+                        .build());
+                return rewriteUserHeaderToMappedUser(mappedIdentity, basicAuthCredentials, request.getHeaders());
             }
             catch (UserMappingException | AccessDeniedException e) {
                 if (exception == null) {
@@ -90,7 +91,7 @@ public class PasswordAuthenticator
      * When the user in the basic authentication header matches the x-trino-user header, we assume that the client does
      * not want to force the runtime user name, and only wanted to communicate the authentication user.
      */
-    private void rewriteUserHeaderToMappedUser(BasicAuthCredentials basicAuthCredentials, MultivaluedMap<String, String> headers, String authenticatedUser)
+    private Identity rewriteUserHeaderToMappedUser(Identity mappedIdentity, BasicAuthCredentials basicAuthCredentials, MultivaluedMap<String, String> headers)
     {
         String userHeader;
         try {
@@ -98,11 +99,12 @@ public class PasswordAuthenticator
         }
         catch (ProtocolDetectionException _) {
             // this shouldn't fail here, but ignore and it will be handled elsewhere
-            return;
+            return mappedIdentity;
         }
         if (basicAuthCredentials.getUser().equals(headers.getFirst(userHeader))) {
-            headers.putSingle(userHeader, authenticatedUser);
+            headers.putSingle(userHeader, mappedIdentity.getUser());
         }
+        return mappedIdentity;
     }
 
     // Extract this out in a method so that the logic of preferring originalUser and fallback on user remains in one place
--- a/core/trino-main/src/main/java/io/trino/server/security/UserMapping.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/UserMapping.java
@@ -17,6 +17,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.annotations.VisibleForTesting;
 import com.google.common.collect.ImmutableList;
+import io.trino.spi.security.Identity;
 import io.trino.util.Case;
 
 import java.io.File;
@@ -67,6 +68,14 @@ public final class UserMapping
         }
 
         throw new UserMappingException("No user mapping patterns match the principal");
+    }
+
+    public Identity mapIdentity(Identity identity)
+            throws UserMappingException
+    {
+        return Identity.from(identity)
+                .withUser(mapUser(identity.getUser()))
+                .build();
     }
 
     public static final class UserMappingRules
--- a/core/trino-spi/src/main/java/io/trino/spi/security/CertificateAuthenticator.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/security/CertificateAuthenticator.java
@@ -16,6 +16,7 @@ package io.trino.spi.security;
 import java.security.Principal;
 import java.security.cert.X509Certificate;
 import java.util.List;
+import java.util.Optional;
 
 public interface CertificateAuthenticator
 {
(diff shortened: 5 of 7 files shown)


wendigo:
@dain do you want to chime in?

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
```diff
@@ -55,7 +56,11 @@ public Identity authenticate(ContainerRequestContext request)
         }
 
         try {
-            Principal principal = authenticatorManager.getAuthenticator().authenticate(certificates);
+            Optional<Identity> identity = authenticatorManager.getAuthenticator().authenticate(certificates);
+            if (identity.isPresent()) {
+                return identity.get();
```
findepi: `userMapping` is silently ignored in this code path

findepi:
> Introduce additional methods to the SPI authenticators to directly provide an Identity. This allows for more flexibility in by the implementing authenticators to define what the Identity should look like.

This describes the code change, but not the context. _The Why_ remains unclear.
What is this that you're trying to do, and is not possible in the current code? 
How important and non-niche use-case is this?

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
tbaeg: Good catch. Fixed!

tbaeg:
> > Introduce additional methods to the SPI authenticators to directly provide an Identity. This allows for more flexibility in by the implementing authenticators to define what the Identity should look like.
> 
> This describes the code change, but not the context. _The Why_ remains unclear. What is this that you're trying to do, and is not possible in the current code? How important and non-niche use-case is this?

I've updated the description to better outline the problem it's trying to solve along with additional context from another MR/comment.

core/trino-spi/src/main/java/io/trino/spi/security/CertificateAuthenticator.java
```diff
@@ -28,5 +29,20 @@ public interface CertificateAuthenticator
      * @return the authenticated entity
      * @throws AccessDeniedException if not allowed
      */
-    Principal authenticate(List<X509Certificate> certificates);
+    @Deprecated
+    Principal createAuthenticatedPrincipal(List<X509Certificate> certificates);
```
ksobolew: I'm mostly interested in the backwards compatibility story (as I'm sure I'll be adjusting a bunch of authenticators to that new API).

* Making this method deprecated indicates that it should not be used or implemented, but it looks like a normal fallback. Is the intent to remove it entirely in the future?
* Renaming this method forces the implementers to make adjustments anyway, can this be avoided?
* Naming the new method the same as the old one is confusing - the correct course of action when someone wants to keep the current behavior is to rename the method in the interface implementation, but the error message will say that the return type is wrong, which suggests something else

findepi: > * Making this method deprecated indicates that it should not be used or implemented, but it looks like a normal fallback. 

That's how deprecation is done. If the method was no longer called, we would remove it.

ksobolew: True, except that it's also renamed, making it technically a new method. So implementers of this interface need to adjust anyway. A typical deprecation process would deprecate the method, but implementations would work unchanged.

tbaeg: > Making this method deprecated indicates that it should not be used or implemented, but it looks like a normal fallback. Is the intent to remove it entirely in the future?

My preference would be to remove it entirely. I see limited/deprecated usage of the `Principal` (i.e. - `SystemAccessControl.checkCanSetUser`) that further supports removal.

> Renaming this method forces the implementers to make adjustments anyway, can this be avoided?

I generally wouldn't rename the deprecated method, but the main reason was for: 

1. Consistency between authenticators in it's current state.
2. The supplanting method name be `authenticate`. At it's core, it's really just a naming preference. (`authenticate` vs `createAuthenticatedIdentity`).

> Naming the new method the same as the old one is confusing - the correct course of action when someone wants to keep the current behavior is to rename the method in the interface implementation, but the error message will say that the return type is wrong, which suggests something else

I don't disagree it is a tad confusing and I'm not against naming the new method `createAuthenticatedIdentity`. That said, I don't think most people have a much trouble with this and I prefer to have some pain on some method names if we can get to the desired state faster.

ksobolew: Sure. Like I said, I *will* be updating a bunch of authenticators so this affects my work. It's not a big deal for me, though it may be for others. I just wanted to make sure that this is deliberate and well-thought-out. Thanks!

tbaeg: > Sure. Like I said, I _will_ be updating a bunch of authenticators so this affects my work. It's not a big deal for me, though it may be for others. I just wanted to make sure that this is deliberate and well-thought-out. Thanks!

I think your concern is valid and warranted. I will happily change the name to `createAuthenticatedIdentity`.

ksobolew: Thank you :)

tbaeg:
> > Some authentication methods can also have authorization information (i.e. - Identity.groups population using JWT claims
> 
> Populating groups as part of the authentication is wrong path. We had that for oauth2 and we realized it caused a lot of issues problems with views. For groups you should be using group providers.

I don't understand the specific problem that was encountered, but populating group information is just an example of what is possible; not a requirement. This just gives extension writers more flexibility in shaping the `Identity`, which I think is a step toward increased adoption without having to change authenticators in `trino-main`. Additionally, there is already support for groups via claims in the `OAUTH2` mode [here](https://github.com/trinodb/trino/blob/19cee94e2f3a4fcf093c636ec487559a9dae68f7/core/trino-main/src/main/java/io/trino/server/security/oauth2/OAuth2Authenticator.java#L87).

This is outside of the scope of this PR, but I'd love to see movement away from `Authenticator`s to something like `IdentityProvider`. This would streamline `Identity` provisioning and provide a simple interface to implement both authentication and authorization aspects. Happy to discuss more via the slack channel for those that are interested.

kokosing:
> I don't understand the specific problem that was encountered

https://github.com/trinodb/trino/pull/15669

> This is outside of the scope of this PR, but I'd love to see movement away from Authenticators to something like IdentityProvider. This would streamline Identity provisioning and provide a simple interface to implement both authentication and authorization aspects. 

Can you please elaborate?

> Happy to discuss more via the slack channel for those that are interested.

I like to post this on github as it is then easier to find in future.

tbaeg:
> > I don't understand the specific problem that was encountered
> 
> #15669

Based on the content in the mentioned issue, in the interim, impersonation issues could be mitigated with https://github.com/trinodb/trino/issues/16539. That said, from an `oauth` perspective, impersonation issues can be solved with RFC 8693.

https://datatracker.ietf.org/doc/html/rfc8693

https://auth0.com/docs/authenticate/custom-token-exchange

> > This is outside of the scope of this PR, but I'd love to see movement away from Authenticators to something like IdentityProvider. This would streamline Identity provisioning and provide a simple interface to implement both authentication and authorization aspects.
> 
> Can you please elaborate?

Will do. Need a bit of time to gather some thoughts.

tbaeg:
@kokosing Going to pause on the broader `IdentityProvider` notion, but could you weigh in on the current PR changes? Are we still against exposing this despite:

> Based on the content in the mentioned issue, in the interim, impersonation issues could be mitigated with https://github.com/trinodb/trino/issues/16539. That said, from an oauth perspective, impersonation issues can be solved with RFC 8693.
>
> https://datatracker.ietf.org/doc/html/rfc8693
>
> https://auth0.com/docs/authenticate/custom-token-exchange

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
```diff
@@ -55,6 +56,14 @@ public Identity authenticate(ContainerRequestContext request)
         }
 
         try {
+            Optional<Identity> identityOpt = authenticatorManager.getAuthenticator().createAuthenticatedIdentity(certificates);
+            if (identityOpt.isPresent()) {
+                Identity identity = identityOpt.get();
```
kokosing: inline `identity` and rename `identityOpt` to `identity` or `authenticatedIdentity`

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
```diff
@@ -23,5 +24,16 @@ public interface PasswordAuthenticator
      * @return the authenticated entity
      * @throws AccessDeniedException if not allowed
      */
+    @Deprecated
```
kokosing, line 27: What plan do you have for removal of that method?

Why not to remove it now?

Same question to other authenticators.

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
```diff
@@ -23,5 +24,16 @@ public interface PasswordAuthenticator
...
     Principal createAuthenticatedPrincipal(String user, String password);
+
+    /**
+     * If implemented and authenticated; provide an {@link Identity}.
+     *
+     * @return the authenticated {@link Identity}
+     */
+    default Optional<Identity> createAuthenticatedIdentity(String user, String password)
```
kokosing, line 35: What does it mean `Optional.empty()` here?


Same question to other authenticators.

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
```diff
@@ -55,6 +56,14 @@ public Identity authenticate(ContainerRequestContext request)
         }
 
         try {
+            Optional<Identity> identityOpt = authenticatorManager.getAuthenticator().createAuthenticatedIdentity(certificates);
```
kokosing: I believe we need to validate if the ideniity is correct.  For example don't we expect that principal is set?

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
tbaeg: I don't have a specific plan other than mark it as deprecated and allow for the maintainers to decide on full removal timeline.

Main reason not to remove it now is to allow those with plugins at least 1 release cycle to adjust.

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
tbaeg: In this context, it means there is is no authenticated `Identity` the `Authenticator` is able to provide, so fall back to using the `authenticate` which will provide a `Principal`.

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
tbaeg: I think the `Identity.Builder` does a good job guarding required fields of a `Identity` as is. We could do more validation in main if desired, but I wouldn't be entirely sure what to validate.

As for the `Principal`, given the optional nature of `Principal` in the `Identity` class, if `Principal` *is* required, it probably shouldn't be.

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
kokosing: I don't like this contract. That means that you introduced a new method to have a smoth transition persiod from one method to the other. However when you remove legacy method, you need to change the method too to remove the `Optional`.

I see two options. Either return exception here that method is not implemented `UnsupportedOperationException` or remove the old method.

I prefer the latter as I believe it will be painful anyway for consumers of that API, defering the pain is not making it easier, but it make it only longer.

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
kokosing: See my response here: https://github.com/trinodb/trino/pull/27387#discussion_r2653097267

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
tbaeg: The `Optional` wasn't just for transition. I prefer the `Optional` instead of relying on the implementation to throw the correct exception. This allows the callers (`trino-main`) to decide what exception to throw and provides a more consistent experience instead of relying on the implementation to "do the right thing". 

All that said, I'm not married to anything. Happy to change the interface immediately as I prefer to not defer the pain. So long as the primary goal of allowing more flexibility for the `Authenticator`s to shape the `Identity` is met.

kokosing: See the method above:

```
* @throws AccessDeniedException if not allowed
```

I would prefer to keep the same way. Notice that exception can actually pass some error message that could be useful for users.

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
```diff
@@ -23,5 +24,16 @@ public interface PasswordAuthenticator
...
+    {
+        return Optional.empty();
```
kokosing, line 37: I would change the default implementation to:


```
return Identity.ofUser(....).withPrincipal(createAuthenticatedPrincipal(...))...
```

What we should shoud pass here as user?

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
tbaeg: Personally, I don't think messages around access should be propagated at all. The implementation can log it, but I don't think there should be any real information being provided back to the user (from a security prospective).

core/trino-spi/src/main/java/io/trino/spi/security/PasswordAuthenticator.java
tbaeg: I prefer to keep it consistent between the `default` implementations and just return `Optional.empty()`. Otherwise, I rather just rip the band-aid off now and make a fully breaking change without a deprecation cycle.

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
```diff
@@ -55,6 +56,13 @@ public Identity authenticate(ContainerRequestContext request)
...
+            Optional<Identity> authenticatedIdentity = authenticatorManager.getAuthenticator().createAuthenticatedIdentity(certificates);
+            if (authenticatedIdentity.isPresent()) {
+                String authenticatedUser = userMapping.mapUser(authenticatedIdentity.get().getUser());
+                return Identity.from(authenticatedIdentity.get())
+                        .withUser(authenticatedUser)
+                        .build();
+            }
             Principal principal = authenticatorManager.getAuthenticator().authenticate(certificates);
             String authenticatedUser = userMapping.mapUser(principal.toString());
             return Identity.forUser(authenticatedUser)
```
kokosing: can we wrap this code as some method to make sure that now and in future we execute the same code to return the final idenity? I mean to wrap user maping. Same for other authenticators.

kokosing (approved):
% one comment

core/trino-main/src/main/java/io/trino/server/security/CertificateAuthenticator.java
```diff
@@ -55,11 +57,14 @@ public Identity authenticate(ContainerRequestContext request)
...
 
         try {
+            Optional<Identity> authenticatedIdentity = authenticatorManager.getAuthenticator().createAuthenticatedIdentity(certificates);
+            if (authenticatedIdentity.isPresent()) {
+                return tryMapUser(userMapping, authenticatedIdentity.get());
+            }
             Principal principal = authenticatorManager.getAuthenticator().authenticate(certificates);
-            String authenticatedUser = userMapping.mapUser(principal.toString());
-            return Identity.forUser(authenticatedUser)
+            return tryMapUser(userMapping, Identity.forUser(principal.toString())
```
kokosing: nit. Please extract variable for identity. It took me a while to understand that you are modyfining an argument of `tryMapUser` not its response.

core/trino-main/src/main/java/io/trino/server/security/HeaderAuthenticator.java
```diff
@@ -52,14 +52,16 @@ public Identity authenticate(ContainerRequestContext request)
         Map<String, List<String>> lowerCasedHeaders = request.getHeaders().entrySet().stream()
                 .collect(Collectors.toMap(entry -> entry.getKey().toLowerCase(Locale.ENGLISH), Map.Entry::getValue));
 
-        for (io.trino.spi.security.HeaderAuthenticator authenticator : this.authenticatorManager.getAuthenticators()) {
+        for (io.trino.spi.security.HeaderAuthenticator authenticator : authenticatorManager.getAuthenticators()) {
             try {
+                Optional<Identity> identity = authenticator.createAuthenticatedIdentity(name -> lowerCasedHeaders.get(name.toLowerCase(Locale.ENGLISH)));
```
kokosing: please extract variable for `lowerCasedHeaders.get(name.toLowerCase(Locale.ENGLISH)`

core/trino-main/src/main/java/io/trino/server/security/HeaderAuthenticator.java
```diff
@@ -52,14 +52,16 @@ public Identity authenticate(ContainerRequestContext request)
...
+                if (identity.isPresent()) {
+                    return tryMapUser(userMapping, identity.get());
+                }
                 Principal principal = authenticator.createAuthenticatedPrincipal(name -> lowerCasedHeaders.get(name.toLowerCase(Locale.ENGLISH)));
-                String authenticatedUser = this.userMapping.mapUser(principal.toString());
-
-                return Identity.forUser(authenticatedUser)
+                return tryMapUser(userMapping, Identity.forUser(principal.toString())
```
kokosing: same comment, extract identity.

Alternatively you can
```
return tryMapUser(
    userMapping, 
    Identity.forUser(principal.toString())
                        .withPrincipal(principal)
                        .build());
```

core/trino-main/src/main/java/io/trino/server/security/PasswordAuthenticator.java
```diff
@@ -60,14 +60,20 @@ public Identity authenticate(ContainerRequestContext request)
         AuthenticationException exception = null;
         for (io.trino.spi.security.PasswordAuthenticator authenticator : authenticatorManager.getAuthenticators()) {
             try {
+                Optional<Identity> authenticatedIdentity = authenticator.createAuthenticatedIdentity(user, password);
+                if (authenticatedIdentity.isPresent()) {
+                    Identity mappedIdentity = tryMapUser(userMapping, authenticatedIdentity.get());
```
kokosing: tryMapUser is nice here, but in this class I think we need this class specific method that would wrap rewriteUserHeaderToMappedUser too and the comment related to rewriteUserHeaderToMappedUser

core/trino-main/src/main/java/io/trino/server/security/UserMapping.java
```diff
@@ -69,6 +70,14 @@ public String mapUser(String principal)
         throw new UserMappingException("No user mapping patterns match the principal");
     }
 
+    public static Identity tryMapUser(UserMapping userMapping, Identity identity)
```
kokosing: why static? why `try`?

I would simply add a non-static method called `mapIdentity` like `mapUser`

tbaeg:
@kokosing Updated, thanks for the feedback!

core/trino-main/src/main/java/io/trino/server/security/UserMapping.java
tbaeg: I agree, adding an instance method feels a lot better. 

As for the `try`, I usually use that as a convention for attempting to do something that may or may not have an impact.

tbaeg:
@kokosing Gentle reminder.

kokosing:
Merged thanks!

PR #27388, opened 2025-11-20
Fix S3 createExclusive in case of AWS SDK retries

findepi:
AWS S3 supports S3 conditional writes, which Trino uses to implement
`S3OutputFile.createExclusive`. AWS SDK also has implicit request
retries in case of certain failures, including read timeouts.

It is possible, as reproduced by `TestS3Retries` modified in this
commit, that object is created successfully, but AWS SDK is not aware of
that success and retries the request. The retried request then fails
with precondition failure -- from the S3 server side perspective the
object already exists. In fact, in case of single PutObject requests,
there seems to be no way to distinguish requests on the server side.
ETag cannot be used for that, since it's usually content-based.
`createExclusive` cannot succeed for two different callers even if they
happen to be writing the same key.

What's surprising is that this problem is also present in case of
multi-part upload, at least in MinIO. It's unclear whether AWS SDK
retries combined with multi-part uploads to AWS S3 would also trigger
precondition failure.

The solution for this implemented in this commit is to tag objects
created by Trino with a random value. When precondition fails and AWS
SDK reports that the request was retried (`numAttempts > 1`), then tags
from the object are checked. `FileAlreadyExistsException` is raised only
when this request fails, or when the object does not have a tag
indicating it was created by the current thread.

- fixes https://github.com/trinodb/trino/issues/27402

## Release notes considerations

Delta Lake
- prevent log writer from failing with `FileAlreadyExistsException` when there are network failures during write
- writing to s3 now requires permissions for `PutObjectTagging` and `GetObjectTagging` operations (breaking change)

### DIFF
--- a/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
+++ b/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
@@ -13,6 +13,7 @@
  */
 package io.trino.filesystem.s3;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.trino.filesystem.encryption.EncryptionKey;
 import io.trino.memory.context.AggregatedMemoryContext;
 import io.trino.memory.context.LocalMemoryContext;
@@ -23,11 +24,15 @@ import software.amazon.awssdk.services.s3.model.AbortMultipartUploadRequest;
 import software.amazon.awssdk.services.s3.model.CompleteMultipartUploadRequest;
 import software.amazon.awssdk.services.s3.model.CompletedPart;
 import software.amazon.awssdk.services.s3.model.CreateMultipartUploadRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectTaggingRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectTaggingResponse;
 import software.amazon.awssdk.services.s3.model.ObjectCannedACL;
 import software.amazon.awssdk.services.s3.model.PutObjectRequest;
 import software.amazon.awssdk.services.s3.model.RequestPayer;
 import software.amazon.awssdk.services.s3.model.S3Exception;
 import software.amazon.awssdk.services.s3.model.StorageClass;
+import software.amazon.awssdk.services.s3.model.Tag;
+import software.amazon.awssdk.services.s3.model.Tagging;
 import software.amazon.awssdk.services.s3.model.UploadPartRequest;
 import software.amazon.awssdk.services.s3.model.UploadPartResponse;
 
@@ -43,7 +48,9 @@ import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
+import java.util.function.Predicate;
 
+import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Verify.verify;
 import static io.trino.filesystem.s3.S3FileSystemConfig.ObjectCannedAcl.getCannedAcl;
 import static io.trino.filesystem.s3.S3FileSystemConfig.S3SseType.NONE;
@@ -58,11 +65,14 @@ import static java.lang.System.arraycopy;
 import static java.net.HttpURLConnection.HTTP_PRECON_FAILED;
 import static java.util.Objects.checkFromIndexSize;
 import static java.util.Objects.requireNonNull;
+import static java.util.UUID.randomUUID;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
 
 final class S3OutputStream
         extends OutputStream
 {
+    private static final String TRINO_CREATION_TAG = "TrinoCreationId";
+
     private final List<CompletedPart> parts = new ArrayList<>();
     private final LocalMemoryContext memoryContext;
     private final Executor uploadExecutor;
@@ -261,7 +271,8 @@ final class S3OutputStream
         }
     }
 
-    private void waitForPreviousUploadFinish()
+    @VisibleForTesting
+    void waitForPreviousUploadFinish()
             throws IOException
     {
         if (inProgressUploadFuture == null) {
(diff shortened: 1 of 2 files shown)


lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -406,4 +466,20 @@ private void abortUploadSuppressed(Throwable throwable)
...
+    {
+        return Tagging.builder()
+                .tagSet(Tag.builder()
+                        .key(TRINO_CREATION_TAG)
+                        .value(trinoCompletionRequestId)
+                        .build())
+                .build();
+    }
+
+    private Predicate<Tag> isTrinoCreationTag()
```
losipiuk: nit: better name would be `hasOurTrinoCreationTag`

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -243,17 +271,31 @@ private void flushBuffer(boolean finished)
                 client.putObject(request, RequestBody.fromByteBuffer(bytes));
                 return;
             }
-            catch (S3Exception e) {
+            catch (SdkException putObjectException) {
+                // When `location` already exists, the operation will fail with `412 Precondition Failed`
```
losipiuk: can you extract exception handling for singlePart and multiPart to helper and use in both places? Are there any tricky differences?

findepi: it's extracted to the level reasonable. extracting catch clause logic IMO is not good for understanding control flow.
but i will take another look how to simplify this code.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
findepi: it's a predicate on a tag, not object, so "has..." is not appropriate.

findepi:
CI
- https://github.com/trinodb/trino/issues/27403 & `testVersionOnCompilerFailedError`

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
losipiuk: Yeah - but add Our/Ours (it could be other trino)

findepi: `isOurTrinoCreationTag` it is now

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -417,15 +434,78 @@ static void putObject(
...
+                if (exclusiveCreate && firstNonNull(putObjectException.numAttempts(), 0) > 1) {
+                    // Check if the object was actually created by a previous attempt of AWS SDK's implicit retries
+                    try {
+                        if (isCreatedByUs(client, context, location, trinoCreationId)) {
+                            createdByUs = UncertainBoolean.TRUE;
+                        }
+                    }
+                    catch (Exception getObjectTaggingException) {
+                        // The object may be gone. In such case we have no way to determine whether it was created by us.
+                        if (getObjectTaggingException instanceof NoSuchKeyException
```
losipiuk: It seems that introspection of `getObjectTaggingException` is not needed here. We can treat any exception which happened when checking if our marker tag exists on file `MAYBE`

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -406,4 +388,120 @@ private void abortUploadSuppressed(Throwable throwable)
...
+            // This is possible when
+            // - object already existed
+            // - object was created by us in a request, but then client side retry logic retried the request
+            boolean objectAlreadyExists = putObjectException instanceof S3Exception s3Exception &&
+                    s3Exception.statusCode() == HTTP_PRECON_FAILED;
+            if (objectAlreadyExists) {
+                UncertainBoolean createdByUs = UncertainBoolean.FALSE;
+                if (exclusiveCreate && firstNonNull(putObjectException.numAttempts(), 0) > 1) {
+                    // Check if the object was actually created by a previous attempt of AWS SDK's implicit retries
+                    // This is workaround for AWS SDK limitation: https://github.com/aws/aws-sdk-java-v2/issues/6580
```
findepi, line 447: https://github.com/aws/aws-sdk-java-v2/issues/6580

pettyjamesm:
The issue makes sense to me, and the approach of using tags to disambiguate should work- but it's worth noting that you do pay for S3 object tags so this does increase the cost profile. How frequently is `createExclusive` actually called here? Is it for each data file in Iceberg tables or only when updating a single manifest per table / partition?

You mention that eTags can't be used to distinguish between multiple calls to create the same object with the same content- but maybe the other question is: do we actually need to? If we're creating the same object that some other writer created, is it actually necessary to fail in that case? Because if we can tolerate that ambiguity, we _could_ use eTags for this purpose and avoid the cost of using tags.

findepi:
> The issue makes sense to me, and the approach of using tags to disambiguate should work- but it's worth noting that you do pay for S3 object tags so this does increase the cost profile.

@pettyjamesm appreciate chiming in here and sharing this perspective.


>  How frequently is `createExclusive` actually called here? Is it for each data file in Iceberg tables or only when updating a single manifest per table / partition?

only when `createExclusive` is explicitly called by application layer
AFAICT this is only in Delta transaction log json file atomic writer (single object per commit to a delta table).




> You mention that eTags can't be used to distinguish between multiple calls to create the same object with the same content- but maybe the other question is: do we actually need to? If we're creating the same object that some other writer created, is it actually necessary to fail in that case? Because if we can tolerate that ambiguity, we could use eTags for this purpose and avoid the cost of using tags.

We considered that, but I don't think it's viable path

- ETags are not guaranteed to be content based when encryption is involved. We don't know "our etag" before we complete the put request, which kills the purpose.
- This would require changing `createExclusive` semantics to allow concurrent writes to complete when they are identical. I wouldn't know how to call these new semantics, i.e. the new semantics would make code much harder to reason about. Also, I am not convinced it would be safe to do so in case of delta transaction log (someone with more delta knowledge would know).

However, i suppose you're asking this out of cost concern and that concern is better addressed by the fact that  usage of these object tags is very narrowly scoped.

PR #27392, opened 2025-11-20
Add support for PostgreSQL-style shorthand casts

### LINKED ISSUE #23795: Cast operator `::`

martint:
Standard SQL supports static method calls on types via the :: operator. While this syntax is generally incompatible with PostgreSQL shorthand cast syntax, there's a subset of that can be safely repurposed to support that functionality.

The SQL specification defines static method calls as:

```
   <static method invocation> ::=
      <path-resolved user-defined type name> <double colon> <method name>
      [ <SQL argument list> ]
```

where `<path-resolved user-defined type name> `translates to:

```
   <user-defined type name> ::=
      [ <schema name> <period> ] <qualified identifier>
```

To support casts, we need to extend the rule to support arbitrary expressions as the target of the invocation. To disambiguate a static method call from a cast, we distinguish between type-producing expressions and expressions that produce regular values. For the latter, if the method matches the name of a well-known type, we treat it as a cast. Otherwise, the expression assumed to be a static method call and fail the evaluation with a "not yet supported" error.

One limitation is that casts are only supported for simple types. Parametric types are not yet supported, but it wouldn't be too hard to add. Types whose name don't match the syntax of SQL function calls are not supported either and will be harder to support. That will require introducing type-producing expressions into the language and making types first-class expressions.

Fixes https://github.com/trinodb/trino/issues/23795

## Release notes

```markdown
## General
* Add support to simplified cast expressions with the `::` operator. ({issue}`23795`)
```

### DIFF
--- a/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
+++ b/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
@@ -586,58 +586,62 @@ valueExpression
     ;
 
 primaryExpression
-    : literal                                                                             #literals
-    | QUESTION_MARK                                                                       #parameter
-    | POSITION '(' valueExpression IN valueExpression ')'                                 #position
-    | '(' expression (',' expression)+ ')'                                                #rowConstructor
-    | ROW '(' expression (',' expression)* ')'                                            #rowConstructor
+    : literal                                                                               #literals
+    | QUESTION_MARK                                                                         #parameter
+    | POSITION '(' valueExpression IN valueExpression ')'                                   #position
+    | '(' expression (',' expression)+ ')'                                                  #rowConstructor
+    | ROW '(' expression (',' expression)* ')'                                              #rowConstructor
     | name=LISTAGG '(' setQuantifier? expression (',' string)?
         (ON OVERFLOW listAggOverflowBehavior)? ')'
         (WITHIN GROUP '(' orderBy ')')
-        filter? over?                                                                     #listagg
+        filter? over?                                                                       #listagg
     | processingMode? qualifiedName '(' (label=identifier '.')? ASTERISK ')'
-        filter? over?                                                                     #functionCall
+        filter? over?                                                                       #functionCall
     | processingMode? qualifiedName '(' (setQuantifier? expression (',' expression)*)?
-        orderBy? ')' filter? (nullTreatment? over)?                                       #functionCall
-    | identifier over                                                                     #measure
-    | identifier '->' expression                                                          #lambda
-    | '(' (identifier (',' identifier)*)? ')' '->' expression                             #lambda
-    | '(' query ')'                                                                       #subqueryExpression
+        orderBy? ')' filter? (nullTreatment? over)?                                         #functionCall
+    | identifier over                                                                       #measure
+    | identifier '->' expression                                                            #lambda
+    | '(' (identifier (',' identifier)*)? ')' '->' expression                               #lambda
+    | '(' query ')'                                                                         #subqueryExpression
     // This is an extension to ANSI SQL, which considers EXISTS to be a <boolean expression>
-    | EXISTS '(' query ')'                                                                #exists
-    | CASE operand=expression whenClause+ (ELSE elseExpression=expression)? END           #simpleCase
-    | CASE whenClause+ (ELSE elseExpression=expression)? END                              #searchedCase
-    | CAST '(' expression AS type ')'                                                     #cast
-    | TRY_CAST '(' expression AS type ')'                                                 #cast
-    | ARRAY '[' (expression (',' expression)*)? ']'                                       #arrayConstructor
-    | '[' (expression (',' expression)*)? ']'                                             #arrayConstructor
-    | value=primaryExpression '[' index=valueExpression ']'                               #subscript
-    | identifier                                                                          #columnReference
-    | base=primaryExpression '.' fieldName=identifier                                     #dereference
-    | name=CURRENT_DATE                                                                   #currentDate
-    | name=CURRENT_TIME ('(' precision=INTEGER_VALUE ')')?                                #currentTime
-    | name=CURRENT_TIMESTAMP ('(' precision=INTEGER_VALUE ')')?                           #currentTimestamp
-    | name=LOCALTIME ('(' precision=INTEGER_VALUE ')')?                                   #localTime
-    | name=LOCALTIMESTAMP ('(' precision=INTEGER_VALUE ')')?                              #localTimestamp
-    | name=CURRENT_USER                                                                   #currentUser
-    | name=CURRENT_CATALOG                                                                #currentCatalog
-    | name=CURRENT_SCHEMA                                                                 #currentSchema
-    | name=CURRENT_PATH                                                                   #currentPath
-    | TRIM '(' (trimsSpecification? trimChar=valueExpression? FROM)?
-        trimSource=valueExpression ')'                                                    #trim
-    | TRIM '(' trimSource=valueExpression ',' trimChar=valueExpression ')'                #trim
-    | SUBSTRING '(' valueExpression FROM valueExpression (FOR valueExpression)? ')'       #substring
-    | NORMALIZE '(' valueExpression (',' normalForm)? ')'                                 #normalize
-    | EXTRACT '(' identifier FROM valueExpression ')'                                     #extract
-    | '(' expression ')'                                                                  #parenthesizedExpression
-    | GROUPING '(' (qualifiedName (',' qualifiedName)*)? ')'                              #groupingOperation
-    | JSON_EXISTS '(' jsonPathInvocation (jsonExistsErrorBehavior ON ERROR)? ')'          #jsonExists
+    | EXISTS '(' query ')'                                                                  #exists
+    | CASE operand=expression whenClause+ (ELSE elseExpression=expression)? END             #simpleCase
+    | CASE whenClause+ (ELSE elseExpression=expression)? END                                #searchedCase
+    | CAST '(' expression AS type ')'                                                       #cast
+    | TRY_CAST '(' expression AS type ')'                                                   #cast
+    // the target is a primaryExpression to support PostgreSQL-style casts
+    // of the form <complex expression>::<type>, which are syntactically ambiguous with
+    // static method calls defined by the SQL spec (and we reserve it for future use)
+    | primaryExpression DOUBLE_COLON identifier ('(' (expression (',' expression)*)? ')')?  #staticMethodCall
+    | ARRAY '[' (expression (',' expression)*)? ']'                                         #arrayConstructor
+    | '[' (expression (',' expression)*)? ']'                                               #arrayConstructor
+    | value=primaryExpression '[' index=valueExpression ']'                                 #subscript
+    | identifier                                                                            #columnReference
+    | base=primaryExpression '.' fieldName=identifier                                       #dereference
+    | name=CURRENT_DATE                                                                     #currentDate
+    | name=CURRENT_TIME ('(' precision=INTEGER_VALUE ')')?                                  #currentTime
+    | name=CURRENT_TIMESTAMP ('(' precision=INTEGER_VALUE ')')?                             #currentTimestamp
+    | name=LOCALTIME ('(' precision=INTEGER_VALUE ')')?                                     #localTime
+    | name=LOCALTIMESTAMP ('(' precision=INTEGER_VALUE ')')?                                #localTimestamp
+    | name=CURRENT_USER                                                                     #currentUser
+    | name=CURRENT_CATALOG                                                                  #currentCatalog
+    | name=CURRENT_SCHEMA                                                                   #currentSchema
+    | name=CURRENT_PATH                                                                     #currentPath
+    | TRIM '(' (trimsSpecification? trimChar=valueExpression? FROM)?                        
+        trimSource=valueExpression ')'                                                      #trim
+    | TRIM '(' trimSource=valueExpression ',' trimChar=valueExpression ')'                  #trim
+    | SUBSTRING '(' valueExpression FROM valueExpression (FOR valueExpression)? ')'         #substring
+    | NORMALIZE '(' valueExpression (',' normalForm)? ')'                                   #normalize
+    | EXTRACT '(' identifier FROM valueExpression ')'                                       #extract
+    | '(' expression ')'                                                                    #parenthesizedExpression
+    | GROUPING '(' (qualifiedName (',' qualifiedName)*)? ')'                                #groupingOperation
+    | JSON_EXISTS '(' jsonPathInvocation (jsonExistsErrorBehavior ON ERROR)? ')'            #jsonExists
     | JSON_VALUE '('
         jsonPathInvocation
         (RETURNING type)?
         (emptyBehavior=jsonValueBehavior ON EMPTY)?
         (errorBehavior=jsonValueBehavior ON ERROR)?
-      ')'                                                                                 #jsonValue
+      ')'                                                                                  #jsonValue
     | JSON_QUERY '('
         jsonPathInvocation
         (RETURNING type (FORMAT jsonRepresentation)?)?
@@ -645,7 +649,7 @@ primaryExpression
         ((KEEP | OMIT) QUOTES (ON SCALAR TEXT_STRING)?)?
         (emptyBehavior=jsonQueryBehavior ON EMPTY)?
         (errorBehavior=jsonQueryBehavior ON ERROR)?
-      ')'                                                                                 #jsonQuery
+      ')'                                                                                   #jsonQuery
     | JSON_OBJECT '('
         (
           jsonObjectMember (',' jsonObjectMember)*
@@ -1123,6 +1127,7 @@ DISTINCT: 'DISTINCT';
 DISTRIBUTED: 'DISTRIBUTED';
 DO: 'DO';
 DOUBLE: 'DOUBLE';
+DOUBLE_COLON: '::';
 DROP: 'DROP';
 ELSE: 'ELSE';
 EMPTY: 'EMPTY';
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
@@ -145,6 +145,7 @@ import io.trino.sql.tree.SimpleCaseExpression;
 import io.trino.sql.tree.SkipTo;
 import io.trino.sql.tree.SortItem;
 import io.trino.sql.tree.SortItem.Ordering;
+import io.trino.sql.tree.StaticMethodCall;
 import io.trino.sql.tree.StringLiteral;
 import io.trino.sql.tree.SubqueryExpression;
 import io.trino.sql.tree.SubscriptExpression;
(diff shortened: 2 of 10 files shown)


core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -582,66 +582,70 @@ valueExpression
...
+    | name=CURRENT_SCHEMA                                                                   #currentSchema
+    | name=CURRENT_PATH                                                                     #currentPath
+    | TRIM '(' (trimsSpecification? trimChar=valueExpression? FROM)?                        
+        trimSource=valueExpression ')'                                                      #trim
+    | TRIM '(' trimSource=valueExpression ',' trimChar=valueExpression ')'                  #trim
+    | SUBSTRING '(' valueExpression FROM valueExpression (FOR valueExpression)? ')'         #substring
+    | NORMALIZE '(' valueExpression (',' normalForm)? ')'                                   #normalize
+    | EXTRACT '(' identifier FROM valueExpression ')'                                       #extract
+    | '(' expression ')'                                                                    #parenthesizedExpression
+    | GROUPING '(' (qualifiedName (',' qualifiedName)*)? ')'                                #groupingOperation
```
findepi, line 637: the diff here looks as if ~50 lines changed. this is probably a formatting change? if so, worth pulling to separate commit.

Viewing in github UI with "hide whitespace" helps, but it would be better if change was readable if possible regardless of where it's viewed.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -582,66 +582,70 @@ valueExpression
...
-    | GROUPING '(' (qualifiedName (',' qualifiedName)*)? ')'                              #groupingOperation
-    | JSON_EXISTS '(' jsonPathInvocation (jsonExistsErrorBehavior ON ERROR)? ')'          #jsonExists
+    | EXISTS '(' query ')'                                                                  #exists
+    | CASE operand=expression whenClause+ (ELSE elseExpression=expression)? END             #simpleCase
+    | CASE whenClause+ (ELSE elseExpression=expression)? END                                #searchedCase
+    | CAST '(' expression AS type ')'                                                       #cast
+    | TRY_CAST '(' expression AS type ')'                                                   #cast
+    // the target is a primaryExpression to support PostgreSQL-style casts
+    // of the form <complex expression>::<type>, which are syntactically ambiguous with
+    // static method calls
```
findepi: worth maybe elaborating what "static method calls" refers to. I understand this is something that exists in SQL specification, but not in Trino's implementation of SQL.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1704,6 +1705,44 @@ private void analyzeFrameRangeOffset(Expression offsetValue, FrameBound.Type bou
             frameBoundCalculations.put(NodeRef.of(offsetValue), function);
         }
 
+        @Override
+        protected Type visitStaticMethodCall(StaticMethodCall node, Context context)
+        {
+            // PostgreSQL-style are syntactically ambiguous with static method calls. So, static method call semantics take precendence.
```
findepi: > PostgreSQL-style are

+"casts"

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -582,66 +582,70 @@ valueExpression
...
+    | primaryExpression DOUBLE_COLON identifier ('(' (expression (',' expression)*)? ')')?  #staticMethodCall
```
findepi, line 615: use `type` on the rhs?

AFAIR engines allow  `a::timestamp with time zone` and likely `a::double precision` too

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1704,6 +1705,44 @@ private void analyzeFrameRangeOffset(Expression offsetValue, FrameBound.Type bou
...
+            // A static method call is characterized by the target being an expression whose type is "type". This not yet supported
+            // as a first-class concept, so we fake it by analyzing the expression normally. If the analysis succeeds, we treat it as
+            // the target of a cast. If the analysis fails, check whether the target is an identifier matching a known type name.
+            try {
+                process(node.getTarget(), context);
+            }
+            catch (TrinoException e) {
+                // assume it might be a type name, so check if it's an identifier matching a known type
+                if (node.getTarget() instanceof Identifier target) {
```
findepi: could there be separate grammar rule with higher precedence

```
identifier :: «type»
```
so that we don't need to do such instanceof checks in Java?

This could maybe allow us not to have `StaticMethodCall` "expression" which is actually only a transport for syntax, not a semantical thing.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1704,6 +1705,44 @@ private void analyzeFrameRangeOffset(Expression offsetValue, FrameBound.Type bou
             frameBoundCalculations.put(NodeRef.of(offsetValue), function);
         }
 
+        @Override
+        protected Type visitStaticMethodCall(StaticMethodCall node, Context context)
+        {
+            // PostgreSQL-style are syntactically ambiguous with static method calls. So, static method call semantics take precendence.
+            // A static method call is characterized by the target being an expression whose type is "type". This not yet supported
+            // as a first-class concept, so we fake it by analyzing the expression normally. If the analysis succeeds, we treat it as
+            // the target of a cast. If the analysis fails, check whether the target is an identifier matching a known type name.
```
findepi: What should be the result of 
```
SELECT bigint::real FROM a_table_that_has_column_named_bigint
```

i.e. treating lhs as type name takes precedence, or treating it as an expression does?

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1704,6 +1705,44 @@ private void analyzeFrameRangeOffset(Expression offsetValue, FrameBound.Type bou
...
+            }
+
+            // assume it's a PostgreSQL-style cast unless result type is not a known type
+            try {
+                Type type = plannerContext.getTypeManager().fromSqlType(node.getMethod().getValue());
+                return setExpressionType(node, type);
+            }
+            catch (Exception e) {
+                throw semanticException(NOT_SUPPORTED, node, "Static method calls are not supported");
+            }
```
findepi, line 1742: Move `return setExpressionType(node, type)` out of this try catch block

core/trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
```diff
@@ -467,6 +468,24 @@ protected String visitFunctionCall(FunctionCall node, Void context)
...
+
+            builder.append(process(node.getTarget(), context))
+                    .append("::")
+                    .append(process(node.getMethod(), context));
+
+            if (!node.getArguments().isEmpty()) {
+                builder.append('(')
+                        .append(joinExpressions(node.getArguments()))
+                        .append(')');
+            }
```
findepi, line 484: If this was actual static method call, then should we print braces also when there are no args?

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1704,6 +1705,44 @@ private void analyzeFrameRangeOffset(Expression offsetValue, FrameBound.Type bou
...
+
+            if (!node.getArguments().isEmpty()) {
+                throw semanticException(NOT_SUPPORTED, node, "Static method calls are not supported");
+            }
+
+            // assume it's a PostgreSQL-style cast unless result type is not a known type
+            try {
+                Type type = plannerContext.getTypeManager().fromSqlType(node.getMethod().getValue());
```
findepi, line 1737: This should take `node.getArguments()`

core/trino-main/src/test/java/io/trino/operator/scalar/TestStaticMethodCall.java
```diff
@@ -0,0 +1,77 @@
...
+    public void teardown()
+    {
+        assertions.close();
+        assertions = null;
+    }
+
+    @Test
+    void testPostgreSqlStyleCast()
+    {
+        assertThat(assertions.expression("1::double"))
```
findepi, line 52: add tests with cast to

- `decimal`, `decimal(3)`, `decimal(3, 2)`
- `double precision`
- `varchar`
- `timestamp`
- `timestamp with time zone` & friends 2
- `interval` & co

regardless of whether each of them is supported today -- there is a lot of special casing for types, so each constitutes a potential special case

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
martint: I had to move the labels to the right to make space for the new rule. It’s technically part of this change, unfortunately.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
martint: Can’t use type, and hence the limitation I mentioned in the description. The standard SQL syntax requires this to be a function invocation, so it has to look like it.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: That would make it syntactically ambiguous with the method invocation. Some types look like function calls. Also, it would result in potentially expensive backtracking on in case of deeply nested type definitions that only differ at the leaves.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: That should fail because it looks like a static method call (real method with no args on bigint type). It needs to behave like that for forward compatibility.

core/trino-parser/src/main/java/io/trino/sql/ExpressionFormatter.java
martint: It’s not required by the SQL spec, which allows omitting parentheses when there are no args

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: But it turns out it doesn't fail in the current code. I'll fix it.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: There's no possible value for type outside of the try-catch. Also, the only valid outcomes are returning from the try block or throwing an exception out of the method.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: Currently, it's only implemented for simple types. It's the limitation I mentioned in the description.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
findepi: > I had to move the labels to the right to make space for the new rule.

i know

> It’s technically part of this change, unfortunately.

it doesn't have to be part of one commit though.
"Move labels to the right to space for a new rule" is a perfect commit message and makes grammar changes obvious.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
findepi: to me, "syntactically ambiguous with static method calls" means that `foo::bar(1)` is ambiguous in a sense that it can be two different things

- PostgreSQL-style cast of expression `foo` to type `bar(1)`
- a static method `bar` invocation on type `foo` with parameters `1`

it is surprising, but it seems to me that `foo::bar` is also ambiguous because argument braces are optional

image

now,
`foo::timestamp` already departs from the spec
`foo::timestamp with time zone` also departs from the spec, but is consistent with `foo::timestamp`. And is not ambiguous.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
findepi: i meant this like below. We don't want to catch arbitrary exceptions from `setExpressionType`.

```
Type type;
try {
                type = plannerContext.getTypeManager().fromSqlType(node.getMethod().getValue());
            }
            catch (Exception e) {
                throw semanticException(NOT_SUPPORTED, node, "Static method calls are not supported");
            }
return setExpressionType(node, type);
```

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
findepi: Even under such limitation we should not ignore `node.getArguments()`.
If  `foo::varchar(100)` succeeds, it must be `varchar(100)`, not eg `varchar`

core/trino-main/src/test/java/io/trino/operator/scalar/TestStaticMethodCall.java
```diff
@@ -52,11 +53,18 @@ void testPostgreSqlStyleCast()
                 .hasType(DoubleType.DOUBLE)
                 .isEqualTo(1.0);
 
+        assertThat(assertions.expression("1::varchar"))
+                .hasType(VarcharType.VARCHAR)
```
findepi, line 57: Add a test with `varchar(10)`.
i think it currently would behave incorrectly.

core/trino-parser/src/test/java/io/trino/sql/parser/TestSqlParser.java
```diff
@@ -850,6 +850,16 @@ void testStaticMethodCall()
                                 new Identifier(location(1, 6), "b", false)),
                         new Identifier(location(1, 10), "double", false),
                         emptyList()));
+
+
+        assertThatThrownBy(() -> SQL_PARSER.createExpression("1::double precision"))
+                .hasMessageMatching(".*mismatched input.*");
+
+        assertThatThrownBy(() -> SQL_PARSER.createExpression("'abc'::timestamp with time zone"))
```
findepi, line 860: we should support this & decimals.
it's not nice to a user to allow `foo::type` for most but not all types that are commonly cast in sql
this syntax really helps write sql, but we punish users because they will have to sometimes go back to old syntax just because they need decimal instead of a double, or point in time instead of a zoneless timestamp

it's ok not to support this within this PR, but i would want to understand what it takes to support it in the future. also, can help with the implementation.

findepi (approved):

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: It's not ignored. It fails with a "static method calls not supported" error.

core/trino-parser/src/test/java/io/trino/sql/parser/TestSqlParser.java
martint: To do that, we'll need to unify these two productions:

* `identifier ('(' (expression (',' expression)*)? ')')?`
* `type`

Some implications:
* `type` will need to be a valid expression, since, for instance, you can have arbitrary nesting: `array(array(array(double precision)))`
* there's a syntactic ambiguity between `identifier ('(' (expression (',' expression)*)? ')')?` and `identifier ('(' typeParameter (',' typeParameter)* ')')?` if the arguments are all integer numbers. It can be parsed either as a static method invocation or as a generic type
* we'll need to figure out how to deal with the common that could lead to expensive backtracking. For example, in `array(array(array(double precision)))`, the parser wouldn't know which alternative to follow ahead of time. It would have to go down one of the alternatives until it finds it can't parse it, backtrack all the way to the top and try the other alternative
* We'll have to resolve other ambiguities or backtracking elsewhere. For example, `SELECT xxx::double precision` could be interpreted either as a cast to `double precision` or as `SELECT xxx::double AS precision`

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
martint: > "syntactically ambiguous with static method calls" means that foo::bar(1) is ambiguous in a sense that it can be two different things

Exactly. That's a syntactic ambiguity that's easy to solve during analysis. We parse all occurrences of `::` as static method calls and decide later if it was meant to be a call or a cast depending on what's on the left side of the expression. If it's a type-producing expression (currently, an identifier matching a type name), then we treat it as a static method call. Otherwise, we treat it as a cast.

> foo::timestamp with time zone also departs from the spec, but is consistent with foo::timestamp. And is not ambiguous.

Right, but to be able to parse that we'd need to allow either `identifier ('(' (expression (',' expression)*)? ')')` or `type` on the right side, which introduces all sorts of problems. See my comment below.

Also, as an aside, note that there's another ambiguous usage (I mention it in the comment below, too) with:

```
SELECT xxx::double precision
```

In Trino, that could be interpreted either as `SELECT xxx::double AS precision` or as a cast to `double precision`. PostgreSQL doesn't have that problem since `double` is not a valid type. Which goes to say, sometimes we can't just adopt a language feature from another database, since there are intricate interactions between all features that need to be taken into account.

core/trino-parser/src/test/java/io/trino/sql/parser/TestSqlParser.java
findepi: > `type` will need to be a valid expression,

this i don't understand.
types are not expressions in current trino (type is not a first class value) and i don't have intention to make them be.

> we'll need to figure out how to deal with the common that could lead to expensive backtracking. For example, in array(array(array(double precision))), the parser wouldn't know which alternative to follow ahead of time. It would have to go down one of the alternatives until it finds it can't parse it, backtrack all the way to the top and try the other alternative

yes, it could require backtracking. but would this be expensive backtracking?
it's not exponential, it's just two passes, right?
it could become exponential if in two passes we also can do backtracking, however `::` is not valid part of a type definition



> `SELECT xxx::double precision` could be interpreted either as a cast to `double precision` or as `SELECT xxx::double AS precision`

that's a very nice example. some systems solve ambiguity by having deterministic precedence, but i get the point.
i am not concerned about users really _needing_ to type `::double precision`. it's a understandable limitation to require them to type this as `::double`. they can still write what they need
i am more concern about casting to `timestamp with time zone` (that's my favorite type btw). how can we allow that?

findepi:
😞

martint:
It conflicts with https://github.com/trinodb/trino/pull/29385, so we'll need to think about a different approach (or syntax) once that goes in.

PR #27407, opened 2025-11-21
Parquet reader - avoid redundant dictionary reads

ShubhamChaurasia:
## Description
Avoid redundant dictionary reads - earlier the dictionary page is read at the time of rowgroup pruning (`PredicateUtils`) and then again read during the actual file reading(`ParquetReader`). This PR removes this redundant work and moves dictionary based rowgroup pruning logic to `ParquetReader`

Would like to thank @pettyjamesm for his initial suggestions on this.

#### Some numbers with patch

These queries were run on 1TB TPC-DS data(parquet, zstd compressed). 

```
select count(*) from store_sales where ss_store_sk=163;

Without Patch - 26.14 [2.88B rows, 2.08GiB] [110M rows/s, 81.4MiB/s]
With Patch - 21.09 [2.88B rows, 1.09GiB] [137M rows/s, 52.9MiB/s]


select count(*) from store_sales where ss_hdemo_sk < 5226;

Without Patch - 22.23 [2.88B rows, 2.4GiB] [130M rows/s, 111MiB/s]
With Patch - 17.17 [2.88B rows, 1.25GiB] [168M rows/s, 74.6MiB/s]


select count(*) from store_sales where ss_store_sk=163 and ss_hdemo_sk < 5226;

Without Patch - 28.37 [2.88B rows, 5.07GiB] [102M rows/s, 183MiB/s]
With Patch - 18.98 [2.88B rows, 2.94GiB] [152M rows/s, 158MiB/s]


select ss_cdemo_sk, count(*) from store_sales where ss_store_sk=163 and ss_hdemo_sk < 5226 group by ss_cdemo_sk;

Without Patch - 28.87 [2.88B rows, 7.36GiB] [99.8M rows/s, 261MiB/s]
With Patch - 24.23 [2.88B rows, 5.23GiB] [119M rows/s, 221MiB/s]
```

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
@@ -16,20 +16,10 @@ package io.trino.parquet.predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import io.airlift.slice.Slice;
-import io.airlift.slice.SliceInput;
-import io.airlift.slice.Slices;
 import io.trino.parquet.BloomFilterStore;
-import io.trino.parquet.DictionaryPage;
 import io.trino.parquet.ParquetCorruptionException;
 import io.trino.parquet.ParquetDataSource;
-import io.trino.parquet.ParquetDataSourceId;
-import io.trino.parquet.ParquetEncoding;
 import io.trino.parquet.ParquetReaderOptions;
-import io.trino.parquet.crypto.AesCipherUtils;
-import io.trino.parquet.crypto.ColumnDecryptionContext;
-import io.trino.parquet.crypto.FileDecryptionContext;
-import io.trino.parquet.crypto.ModuleType;
 import io.trino.parquet.metadata.BlockMetadata;
 import io.trino.parquet.metadata.ColumnChunkMetadata;
 import io.trino.parquet.metadata.ParquetMetadata;
@@ -39,20 +29,12 @@ import io.trino.spi.predicate.TupleDomain;
 import io.trino.spi.type.DecimalType;
 import io.trino.spi.type.Type;
 import org.apache.parquet.column.ColumnDescriptor;
-import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.statistics.Statistics;
-import org.apache.parquet.format.BlockCipher;
-import org.apache.parquet.format.DictionaryPageHeader;
-import org.apache.parquet.format.PageHeader;
-import org.apache.parquet.format.PageType;
-import org.apache.parquet.internal.column.columnindex.OffsetIndex;
 import org.apache.parquet.internal.filter2.columnindex.ColumnIndexStore;
-import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.parquet.schema.MessageType;
 import org.joda.time.DateTimeZone;
 
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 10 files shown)


lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
```diff
@@ -253,147 +214,9 @@ private static Map<ColumnDescriptor, Long> getColumnValueCounts(PrunedBlockMetad
...
-
-    private record PageHeaderWithData(PageHeader pageHeader, Slice compressedData)
-    {
-        private PageHeaderWithData(PageHeader pageHeader, Slice compressedData)
-        {
-            this.pageHeader = requireNonNull(pageHeader, "pageHeader is null");
-            this.compressedData = requireNonNull(compressedData, "compressedData is null");
-        }
+        private static final PredicateMatchResult FALSE_WITH_DICT_MATCHING_NOT_NEEDED = new PredicateMatchResult(false, null, null);
+        private static final PredicateMatchResult TRUE_WITH_DICT_MATCHING_NOT_NEEDED = new PredicateMatchResult(true, null, null);
```
Praveen2112: nit: Can we avoid null here ? Pass `TupleDomain#all` and an empty set ?

lib/trino-parquet/src/main/java/io/trino/parquet/reader/RowGroupInfo.java
```diff
@@ -14,8 +14,13 @@
...
 import io.trino.parquet.metadata.PrunedBlockMetadata;
+import io.trino.parquet.predicate.TupleDomainParquetPredicate;
+import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.internal.filter2.columnindex.ColumnIndexStore;
 
 import java.util.Optional;
+import java.util.Set;
 
-public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata, long fileRowOffset, Optional<ColumnIndexStore> columnIndexStore) {}
+public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata, long fileRowOffset, Optional<ColumnIndexStore> columnIndexStore,
```
pettyjamesm: Nit: break each field onto a separate line

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -467,38 +469,67 @@ private int nextBatch()
...
+            if (rowGroupInfo.indexPredicate().isPresent()) {
+                try {
+                    initializePageReader(field);
+                    boolean match = columnReaders.get(field.getId()).dictionaryPredicateMatch(rowGroupInfo);
+                    if (!match) {
+                        return false;
+                    }
+                }
+                catch (Exception e) {
+                    log.error(e, "Error while matching dictionary predicate for field " + field);
```
pettyjamesm: Do we really want to swallow an exception at this point? Seems like we should re-throw a variation of `RuntimeException` from `dictionaryPredicateMatch` (instead of throw `IOException`) and then not swallow that exception here since if we can't decode a dictionary page we should probably fail.

lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
ShubhamChaurasia: It's `TupleDomainParquetPredicate` so needs be converted to equivalent of `TupleDomain#all`. Made `indexPredicate` and `candidateColumnsForDictionaryMatching` `Optional` instead which removes wrapping them in Optional later when passing to `RowGroupInfo`

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
ShubhamChaurasia: yeah, we can let it throw `ParquetCorruptionException` which a few other methods also throw when they fail to read parquet. Made a change for that.

ShubhamChaurasia:
thanks for the comments @pettyjamesm @Praveen2112, addressed those, could you check?

PR #27408, opened 2025-11-21
Add active worker count and active coordinator count JMX metrics

gertjanal:
## Description
The `/metrics` endpoint lists active nodes, but not all nodes are workers and # of workers is not always equal to # nodes - # coordinators as coordinators can be workers as well.

This PR adds `CoordinatorNodeManager_ActiveCoordinatorCount` and `CoordinatorNodeManager_ActiveWorkerCount`  metrics. I can't find any tests for the current metrics in the `CoordinatorNodeManager`, so I'm not sure they should be added.

## Release notes

( X ) Release notes are required. Please propose a release note for me.

```markdown
## Metrics
* Added coordinator and worker counts to the metrics endpoint
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/node/CoordinatorNodeManager.java
+++ b/core/trino-main/src/main/java/io/trino/node/CoordinatorNodeManager.java
@@ -291,7 +291,7 @@ public final class CoordinatorNodeManager
     }
 
     @Managed
-    public long getActiveWorkerCount()
+    public int getActiveWorkerCount()
     {
         AllNodes allNodes = getAllNodes();
         return allNodes.activeNodes().size() - allNodes.activeCoordinators().size();


core/trino-main/src/main/java/io/trino/node/InternalNode.java
```diff
@@ -41,15 +41,22 @@ public class InternalNode
     private final URI internalUri;
     private final NodeVersion nodeVersion;
     private final boolean coordinator;
+    private final boolean worker;
     private final long longHashCode;
 
     public InternalNode(String nodeIdentifier, URI internalUri, NodeVersion nodeVersion, boolean coordinator)
```
gertjanal, line 46: This constructor  is used in quite some test files. Shall I remove this constructor and update the tests, or shall I keep it this way?

core/trino-spi/pom.xml
```diff
@@ -291,6 +291,12 @@
                                     <new>method void io.trino.spi.block.ShortArrayBlockEncoding::&lt;init&gt;(boolean)</new>
                                     <justification>ShortArrayBlockEncoding need to accept a parameter to enable SIMD support</justification>
                                 </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.addedToInterface</code>
+                                    <new>method boolean io.trino.spi.Node::isWorker()</new>
```
chenjian2664: It's unnecessary to add it, since the worker means not coordinator. we already has `isCoordinator`, why not just reuse it

gertjanal: Because a node can be a worker and coordinator at the same time using node-scheduler.include-coordinator as described here: https://trino.io/docs/current/installation/deployment.html#config-properties

chenjian2664: I think it is still a coordinator, that is allow schedule task on it but it is not a worker

gertjanal: With `node-scheduler.include-coordinator=true`, the UI adds it to the active worker count, but that number is not available via the metrics. The only metric available is the number of all nodes in the cluster.

So with `node-scheduler.include-coordinator=true`, the active node count is the number of available workers, with node-scheduler.include-coordinator=false` (the default), the active node count - the number of coordinators is the number of available workers. This PR introduces a new metric that is always consistent with the number of active workers as shown in the UI

chenjian2664: understood, I don't have a strong view on this yet, let's wait for more opinions, maybe slack?

core/trino-main/src/main/java/io/trino/node/InternalNode.java
```diff
@@ -41,15 +41,22 @@ public class InternalNode
...
+    {
+        this(nodeIdentifier, internalUri, nodeVersion, coordinator, !coordinator);
+    }
+
+    public InternalNode(String nodeIdentifier, URI internalUri, NodeVersion nodeVersion, boolean coordinator, boolean worker)
```
findepi: can a node be neither coordinator nor worker?
can a node be both coordinator and worker?

(`node-scheduler.include-coordinator` doesn't mean the coordinator needs to be called "a worker", even though it does work)

core/trino-main/src/main/java/io/trino/node/AllNodes.java
```diff
@@ -35,5 +36,6 @@ public record AllNodes(
         drainingNodes = ImmutableSet.copyOf(requireNonNull(drainingNodes, "drainingNodes is null"));
         shuttingDownNodes = ImmutableSet.copyOf(requireNonNull(shuttingDownNodes, "shuttingDownNodes is null"));
         activeCoordinators = ImmutableSet.copyOf(requireNonNull(activeCoordinators, "activeCoordinators is null"));
+        activeWorkers = ImmutableSet.copyOf(requireNonNull(activeWorkers, "activeWorkers is null"));
```
findepi: if all this is for is getting the count, why not store the count only, without keeping the set?

core/trino-main/src/main/java/io/trino/node/InternalNode.java
findepi: we definitely don't want to keep two constructors.
i don't know whether we need a new one (per above), but **IF** we do, the old one should be deprecated or perhaps simply removed. this is internal code.

core/trino-main/src/main/java/io/trino/node/AllNodes.java
gertjanal: Yes, I can do that

core/trino-main/src/main/java/io/trino/node/InternalNode.java
gertjanal: Sad part is that the whole "the node can be either coordinator and worker" is for a single node setup, in which one is most likely not very much interested in this metric at all. However, I don't really like returning the worker count as nodecount - 1

gertjanal: I have rewritten the code and removed the worker boolean. The workerCount is now activeNodes - coordinators, which is sufficient for almost all setups

core/trino-main/src/main/java/io/trino/node/CoordinatorNodeManager.java
```diff
@@ -284,6 +286,18 @@ public int getShuttingDownNodeCount()
...
+    @Managed
+    public int getActiveCoordinatorCount()
+    {
+        return getAllNodes().activeCoordinators().size();
+    }
+
+    @Managed
+    public long getActiveWorkerCount()
+    {
+        return getAllNodes().activeWorkerCount();
```
findepi: Why not
```java
AllNodes allNodes = getAllNodes();
return allNodes.activeNodes().size() - allNodes.activeCoordinators().size();
```

then no change in `AllNodes` is needed

gertjanal: Yes, that might be even better

chenjian2664:
ci: 
```
Error:  /home/runner/work/trino/trino/core/trino-main/src/main/java/io/trino/node/TestingInternalNodeManager.java:[61,25] constructor AllNodes in record io.trino.node.AllNodes cannot be applied to given types;
```

gertjanal:
@chenjian2664 Thanks, fixed the compilation issue

core/trino-main/src/main/java/io/trino/node/CoordinatorNodeManager.java
```diff
@@ -284,6 +284,19 @@ public int getShuttingDownNodeCount()
...
     }
 
+    @Managed
+    public int getActiveCoordinatorCount()
+    {
+        return getAllNodes().activeCoordinators().size();
+    }
+
+    @Managed
+    public long getActiveWorkerCount()
```
findepi:
```suggestion
    public int getActiveWorkerCount()
```

core/trino-main/src/main/java/io/trino/node/CoordinatorNodeManager.java
```diff
@@ -284,6 +284,19 @@ public int getShuttingDownNodeCount()
...
+    public int getActiveCoordinatorCount()
+    {
+        return getAllNodes().activeCoordinators().size();
+    }
+
+    @Managed
+    public int getActiveWorkerCount()
+    {
+        AllNodes allNodes = getAllNodes();
+        return allNodes.activeNodes().size() - allNodes.activeCoordinators().size();
```
chenjian2664, line 297: getActiveNodeCount() - getActiveCoordinatorCount() ?

findepi: Would that be racy?

PR #27409, opened 2025-11-21
Initialize PREFERRED_BIT_WIDTH before detect simd support

SongChujun:
## Description
Otherwise, PREFERRED_BIT_WIDTH will be uninitialized and equal to the default value of 0, causing detectSimd() to always return SimdSupport.NONE.

## Release notes

No release notes required because this fixes hardware detection for SIMD support that has not yet shipped in a release.


findepi:
thanks for fixing this

> Otherwise, PREFERRED_BIT_WIDTH will be uninitialized and equal to the default value of 0, causing detectSimd() to always return SimdSupport.NONE.

it's too late for this PR, but such info is useful not only as a PR description.
Commit message and maybe even code comment are also good places to share it.

PR #27417, opened 2025-11-23
Tests: Remove AWS region derivation in REST container and clean up catalog properties

smaheshwar-pltr:
## Description

1. The `IcebergRestCatalogBackendContainer` relies on the container deriving the AWS region from an environment variable, but this can be set instead with an Iceberg property, `client.region`
2. The environment variable setting readability may be improved with  a method that translates normal catalog properties to ones formatted how the container expects.

## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-testing-containers/src/main/java/io/trino/testing/containers/IcebergRestCatalogBackendContainer.java
+++ b/testing/trino-testing-containers/src/main/java/io/trino/testing/containers/IcebergRestCatalogBackendContainer.java
@@ -17,6 +17,8 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import org.testcontainers.containers.Network;
 
+import java.util.Locale;
+import java.util.Map;
 import java.util.Optional;
 
 import static io.trino.testing.containers.Minio.MINIO_REGION;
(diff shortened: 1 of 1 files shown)


testing/trino-testing-containers/src/main/java/io/trino/testing/containers/IcebergRestCatalogBackendContainer.java
```diff
@@ -36,16 +38,16 @@ public IcebergRestCatalogBackendContainer(
...
+                toCatalogEnvVars(ImmutableMap.of(
+                        "include-credentials", "true",
+                        "warehouse", warehouseLocation,
+                        "io-impl", "org.apache.iceberg.aws.s3.S3FileIO",
+                        "s3.access-key-id", minioAccessKey,
+                        "s3.secret-access-key", minioSecretKey,
+                        "s3.session-token", minioSessionToken,
+                        "s3.endpoint", "http://minio:4566",
+                        "s3.path-style-access", "true",
+                        "client.region", MINIO_REGION)),
```
smaheshwar-pltr, line 50: I think we can do this, using an explicit Iceberg property, instead of setting `AWS_REGION` as an env var for the AWS SDK to derive region from

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/IcebergRestCatalogBackendContainer.java
```diff
@@ -54,4 +56,18 @@ public String getRestCatalogEndpoint()
     {
         return getMappedHostAndPortForExposedPort(8181).toString();
     }
+
+    private static Map<String, String> toCatalogEnvVars(Map<String, String> properties)
```
smaheshwar-pltr, line 60: Reverses https://github.com/databricks/iceberg-rest-image/blob/62a5078032e0675d34b377d2ead2eea814c1da48/src/main/java/org/apache/iceberg/rest/RESTCatalogServer.java#L46-L62, so the constructor is more readable as it just uses standard Iceberg properties

PR #27426, opened 2025-11-24
Extend simd support detection logic for ARM

SongChujun:
This PR extends simd support for ARM based machines. 

Experiments show that on machines that supports ```sve```, the performance of vectorized code path is faster, compared to scalar path: https://jmh.morethan.io/?gists=a74f2b009fba51d2566a8283d8c941a0,0af9694b0e3cbdb8337bcd9580ff4c6d, experiments are performed on c7g.4xlarge, with with ```sve``` support.

We also conducted experiment on machines that only supports ```neon```, however, we observed regression, so we choose to gate the SIMD support on ```sve``` support: https://jmh.morethan.io/?gists=9ae472bec432b3b5a2887132756b160e,44e603976ebfa8b8a0c28727babe558c experiments are performed on c6g.4xlarge, with with ```neon``` support only.

This is a follow up of https://github.com/trinodb/trino/pull/26919

## Release notes

```markdown
## General
* Improve performance of data exchanges by using SIMD instructions on ARM CPUs that support SVE, such as graviton 3 and 4. This can be disabled by setting `experimental.blockserde-vectorized-null-suppression-strategy=NONE`({issue}`27426`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
+++ b/core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
@@ -13,12 +13,11 @@
  */
 package io.trino.simd;
 
+import com.google.common.base.StandardSystemProperty;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
 import io.trino.FeaturesConfig;
-import io.trino.util.MachineInfo;
 import jdk.incubator.vector.VectorShape;
-import oshi.hardware.CentralProcessor.ProcessorIdentifier;
 
 import java.util.EnumSet;
 import java.util.Set;
@@ -32,6 +31,14 @@ AVX512VBMI2 (for VPCOMPRESSB / VPCOMPRESSW instruction support over byte and sho
 whether Vector<T>#compress(VectorMask<T>) is supported natively in hardware or emulated by the JVM - because the emulated
 support is so much slower than the simple scalar code that exists, but since we don't have the ability to detect that
 directly from the JDK vector API we have to assume that native support exists whenever the CPU advertises it.
+
+On ARM, the situation is similar but the threshold for useful performance is different: NEON does not provide a native
+equivalent for the vector compress operation and JVM emulation tends to be significantly slower than our scalar fallback.
+SVE, on the other hand, offers predicate-driven operations that can lower Vector<T>#compress to efficient native
+instructions. Since the JDK vector API does not expose whether compress is lowered to SVE or emulated, we conservatively
+enable SIMD-only paths on ARM when the CPU advertises SVE (and the preferred vector width is large enough), and treat
+NEON-only systems as not supporting native compress for the purposes of this optimization. In AWS Graviton families,
+Graviton2 is NEON-only (no SVE), whereas Graviton3 provides SVE.
  */
 @Singleton
 public final class BlockEncodingSimdSupport
@@ -43,6 +50,7 @@ public final class BlockEncodingSimdSupport
             boolean expandAndCompressLong)
     {
         public static final SimdSupport NONE = new SimdSupport(false, false, false, false);
+        public static final SimdSupport ALL = new SimdSupport(true, true, true, true);
     }
 
     private static final int MINIMUM_SIMD_LENGTH = 256;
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -110,6 +125,37 @@ enum X86SimdInstructionSet {
...
+                }
+            }
+        }
+
+        // Ensure the JVM reports a sufficiently wide preferred vector size before enabling Vector API paths
+        if (PREFERRED_BIT_WIDTH < MINIMUM_SIMD_LENGTH) {
+            return SimdSupport.NONE;
+        }
+        else {
+            boolean hasSve = armFlags.contains(ArmSimdInstructionSet.sve);
```
pettyjamesm: Maybe just inline this into the `PREFERRED_BIT_WIDTH` branch, and return `SimdSupport.NONE`

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -72,11 +73,25 @@ private static SimdSupport detectSimd()
...
 
+        // Detect AArch64 / ARM64 platforms
+        if (arch != null) {
+            String normalizedArch = arch.toLowerCase(ENGLISH);
+            if (normalizedArch.contains("aarch64") || normalizedArch.contains("arm64")) {
+                return detectArmSimdSupport();
+            }
+        }
+
+        // Some platforms may report a non-standard vendor name for ARM CPUs
```
pettyjamesm: Any idea what those names would be? Seems like we probably want to prefer the simplest possible detection logic possible here.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: done

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: discussed offline, deleted the vendor based detection logic.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -110,6 +129,31 @@ enum X86SimdInstructionSet {
...
+                if (flags.contains(instructionSet.name())) {
+                    armFlags.add(instructionSet);
+                }
+            }
+        }
+
+        if (PREFERRED_BIT_WIDTH < MINIMUM_SIMD_LENGTH || !armFlags.contains(ArmSimdInstructionSet.sve)) {
+            return SimdSupport.NONE;
+        }
+        else {
```
raunaqmorarka: `else` is redundant

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -32,6 +31,13 @@ We need to specifically detect AVX512F (for VPCOMPRESSD / VPCOMPRESSQ instructio
 whether Vector<T>#compress(VectorMask<T>) is supported natively in hardware or emulated by the JVM - because the emulated
 support is so much slower than the simple scalar code that exists, but since we don't have the ability to detect that
 directly from the JDK vector API we have to assume that native support exists whenever the CPU advertises it.
+
+On ARM, the situation is similar but the threshold for useful performance is different: NEON does not provide a native
+equivalent for the vector compress operation and JVM emulation tends to be significantly slower than our scalar fallback.
+SVE, on the other hand, offers predicate-driven operations that can lower Vector<T>#compress to efficient native
+instructions. Since the JDK vector API does not expose whether compress is lowered to SVE or emulated, we conservatively
+enable SIMD-only paths on ARM when the CPU advertises SVE (and the preferred vector width is large enough), and treat
+NEON-only systems as not supporting native compress for the purposes of this optimization.
```
raunaqmorarka: It would be good to mention which graviton generations are NEON only and which have SVE.

SongChujun: added description for that.

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
SongChujun: removed ```eles```

pettyjamesm:
Confirmed from @raunaqmorarka that the TPCDS q09 regression is also present in master outside of this PR, so proceeding with merging. Thanks for extending the SIMD support to ARM, @SongChujun

PR #27433, opened 2025-11-25
Fix `EXPLAIN (TYPE IO)` failure when constraint contains type which cannot be cast to `varchar`

hqbhoho:
## Description
Avoid exception if constraint contain type which can't cast to varchar

## Release notes

```markdown
## General
* Fix `EXPLAIN (TYPE IO)` failure when query constraint contains type which cannot be cast to `varchar`. ({issue}`27433 `)
```


Praveen2112:
Can we add some test case to this changes

core/trino-main/src/main/java/io/trino/sql/planner/planprinter/IoPlanPrinter.java
```diff
@@ -783,7 +783,7 @@ private FormattedDomain parseDomain(Domain domain)
                                     .collect(toImmutableSet())),
                     discreteValues -> formattedRanges.addAll(
                             discreteValues.getValues().stream()
-                                    .map(value -> valuePrinter.castToVarcharOrFail(type, value))
```
Praveen2112, line 786: Now this `castToVarcharOrFail` is potentially unused right ? Can we remove that method or make it private ?

hqbhoho: Yes, I make it private

core/trino-main/src/main/java/io/trino/sql/planner/planprinter/IoPlanPrinter.java
```diff
@@ -783,7 +783,7 @@ private FormattedDomain parseDomain(Domain domain)
...
+                                    .map(value -> valuePrinter.castToVarchar(type, value))
```
findepi, line 786: `castToVarchar` is misnamed, exactly because may return something else when cast would fail
calling it "toPrintable"  or "toString" would be better. 
Let's rename as a separate PR (before or after this one)

kasiafi (approved):

core/trino-main/src/main/java/io/trino/sql/planner/planprinter/IoPlanPrinter.java
hqbhoho: okk, I will file a new PR to rename `castToVarchar ` to `toString` and remove `castToVarcharOrFail ` after this one merged.

chenjian2664:
@hqbhoho The pr got merged, FYI

PR #27439, opened 2025-11-25
Clean up dependencies on SPI test jar

martint:
Replace usages of TestingTypeManager and TestingBlockEncodingSerde, which are meant to be used only within the SPI module.

## Release notes

### DIFF
--- a/plugin/trino-delta-lake/pom.xml
+++ b/plugin/trino-delta-lake/pom.xml
@@ -425,13 +425,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
-            <artifactId>trino-spi</artifactId>
-            <type>test-jar</type>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
             <artifactId>trino-testing</artifactId>
             <scope>test</scope>
         </dependency>
--- a/core/trino-main/pom.xml
+++ b/core/trino-main/pom.xml
@@ -470,13 +470,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
-            <artifactId>trino-spi</artifactId>
-            <type>test-jar</type>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
             <artifactId>trino-testing-services</artifactId>
             <scope>test</scope>
         </dependency>
--- a/lib/trino-plugin-toolkit/pom.xml
+++ b/lib/trino-plugin-toolkit/pom.xml
@@ -175,13 +175,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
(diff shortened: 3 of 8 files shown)


core/trino-spi/src/test/java/io/trino/spi/block/TestingBlockEncodingSerde.java
```diff
@@ -29,7 +29,11 @@
 import static java.nio.charset.StandardCharsets.UTF_8;
 import static java.util.Objects.requireNonNull;
 
-// This class is exactly the same as BlockEncodingManager. They are in SPI and don't have access to InternalBlockEncodingSerde.
+/**
+ * This class is only meant to be used within the trino-spi module. For all other modules, use InternalBlockEncodingSerde.TESTING_BLOCK_ENCODING_SERDE
+ * <p>
+ * This class is exactly the same as BlockEncodingManager. They are in SPI and don't have access to InternalBlockEncodingSerde.
```
findepi: "They"?

core/trino-main/src/test/java/io/trino/sql/planner/TestSymbolKeySerialization.java
```diff
@@ -54,8 +53,7 @@ void testRoundTrip()
                 .contains("7|varchar|b")
                 .contains("7|varchar|%s".formatted("b".repeat(256)))
                 .contains("7|integer|1abcd")
-                .contains("7|integer|abcd")
-                .contains("2|id|a");
```
findepi, line 58: why this one is gone?

plugin/trino-delta-lake/pom.xml
```diff
@@ -423,13 +423,6 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-spi</artifactId>
-            <type>test-jar</type>
-            <scope>test</scope>
```
findepi, line 430: ❤️ 
the deps on `test-jar` are never cool

core/trino-spi/src/test/java/io/trino/spi/block/TestingBlockEncodingSerde.java
martint: That comes from the old comment that was there.

core/trino-main/src/test/java/io/trino/sql/planner/TestSymbolKeySerialization.java
martint: `id` is not a real type

core/trino-spi/src/test/java/io/trino/spi/block/TestingBlockEncodingSerde.java
martint: I'm rephrasing it

PR #27454, opened 2025-11-26
Use random table name even for fail scenarios in create smoke tests

vlad-lyutenko:
## Description

In trino engine, it could be situation that some checks could be executed on different levels (Metastore e.t.c)
So we can fall in situation, that with static table name (if for example it mimics some bucket storage), some location checks could potentially be executed before engine checks (and test could fail not because of some logic, but just because preexisting bucket). 
To omit such situation - let's use random naming even for fail scenarious.

## Release notes

```markdown
## Section
```


findinpath (approved):
Could you pls share in the description of the PR a stacktrace to reveal in more detail what kind of unexpected issues are appearing while running the tests concurrently?

Thank you for your contribution 🙏

PR #27455, opened 2025-11-26
Support prefetch & stale time for S3 web identity provider

jirislav:
## Description

When `s3.use-web-identity-token-credentials-provider` is configured, then the WebIdentityTokenFileCredentialsProvider is used in AWS SDK to communicate with S3. This provider supports [specifying the prefetchTime & staleTime](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credential-caching.html) (by default [it's 5 & 1 minute](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/sts/auth/StsCredentialsProvider.BaseBuilder.html#staleTime(java.time.Duration))).

Why is this important? Well, because when we presign an S3 URL (e.g. for Trino client to retrieve the data using the spooling protocol), the attached token will be valid only for as long, as is the remaining validity time of the temporary token that was used to sign the URL. For example, if the MaxSessionDuration in AWS IAM role is kept to it's default 1 hour, then what happens is that every hour new temporary credentials need to be fetched. If prefetch time is 5 minutes (and stale time 1 minute), then the worst-case scenario is that every hour the Trino clients might experience the `ExpiredToken` exception when downloading the spooled segments from a signed S3 url later than 1-5 minutes since their creation.

This issue is even worse in case some Trino workers finish their job early, upload their segments to S3, sign them, but then the Trino client still needs to wait for all the remaining workers to finish their part, which might take several more minutes. So if the query is executed just before the session is about to expire, then the clients running such heavy queries are almost guaranteed to experience the `ExpiredToken` exception.

## Release notes

Release notes are required. Please propose a release note for me.


chenjian2664:
@wendigo @electrum

findepi:
> then the worst-case scenario is that every hour the Trino clients might experience the `ExpiredToken` exception when downloading the spooled segments from a signed S3 url later than 1-5 minutes since their creation.

This sounds wrong. If i understand correctly, such an undesired situation happens under AWS's and Trino's default configuration.
We should improve out of the box behavior for such case. For example, trigger token refresh earlier than just 5 minutes before expiration. 

I understand a code change is needed for that.
I am not convinced yet, new configuration toggles are needed too.

jirislav:
> We should improve out of the box behavior for such case. For example, trigger token refresh earlier than just 5 minutes before expiration.

I agree with improving the out of the box behavior. However, it is not straightforward to define a universal configuration that would work for everyone.

For example, we might decide to use 30 minutes prefetch time with 25 minutes stale time. But then, queries running longer than that still might experience permission issues when attempting to download a segment from S3 that was presigned with now expired token.

Additionally, Trino itself has no information about the IAM role's session refresh interval. [The default (as well as the minimum) is 1h](https://docs.aws.amazon.com/singlesignon/latest/userguide/howtosessionduration.html), with maximum configurable interval being 12h. Now I honestly don't know what would happen in case the stale time would be bigger than the token refresh interval. Worst case scenario could be constant token refresh, resulting in rate limiting the presign requests, maybe even inability to interact with S3.

For that reason, it probably makes sense to keep it under 1h, but also not too close to the 1h, because then we would be essentially hitting the refresh every few minutes, which might be throttled.

> I understand a code change is needed for that.
I am not convinced yet, new configuration toggles are needed too. 

I would prefer to let people control the behavior, because they might be purposefully running very heavy queries, maybe even spanning several hours of wall time.

findepi:
> Additionally, Trino itself has no information about the IAM role's session refresh interval. [The default (as well as the minimum) is 1h](https://docs.aws.amazon.com/singlesignon/latest/userguide/howtosessionduration.html),

That's very comfortable if default and minimum validity time are the same.
This means that, if we make Trino work under default IAM's configuration (which is most strict), it should be expected to work under all other options (which are more relaxed).

> For example, we might decide to use 30 minutes prefetch time with 25 minutes stale time. But then, queries running longer than that still might experience permission issues when attempting to download a segment from S3 that was presigned with now expired token.

We are talking about client protocol spooling to S3, right?
It would expect it to work with queries longer than 1 hour, without need to change AWS IAM configuration.
Query time should, in fact, be of no relevance here. What matters is the time between spooled fragment is closed by the writer (worker node) and it's opened by the reader (the client).
Now, what matters is client ability to download the spooled chunks.

For "a very slow" client this would take "a very long time". If a client is unable to download query results in ⩽30 minutes[^1], then I would say `COORDINATOR_STORAGE_REDIRECT` should be used instead of `STORAGE`. 
Base on this PR description, this is not your case though.

> This issue is even worse in case some Trino workers finish their job early, upload their segments to S3, sign them, but then the Trino client still needs to wait for all the remaining workers to finish their part, which might take several more minutes. 

This sounds like ~a bug~ something we should fix? It's definitely not desirable for a worker to write spooled client data to S3, compute pre-signed URL and then sit on it for inordinate amount of time. These URLs should be flushed down the processing pipeline as soon as they are ready.



[^1]: 30 minutes was derived as half of IAM's default (and minimum) token validity time of 1 hour. It's assumed that refreshing a token when remaining validity is 30 minutes won't be abusive. It is just 2x more frequent than once per 1h refresh which is clearly insufficient. And remaining 30 minutes of freshness might be enough.

findepi:
cc @wendigo @electrum

findepi:
it sounds like this PR addresses a rather important problem
it's sad it gone stale.

> Why is this important? Well, because when we presign an S3 URL (e.g. for Trino client to retrieve the data using the spooling protocol), the attached token will be valid only for as long, as is the remaining validity time of the temporary token that was used to sign the URL. [...] then the worst-case scenario is that every hour the Trino clients might experience the `ExpiredToken` exception when downloading the spooled segments from a signed S3 url later than 1-5 minutes since their creation.

@jirislav would you mind filing a separate issue describing the problem in detail?
Think could help elevate importance of this change. The PR title is just a solution (normal for PR titles) and doesn't directly indicate the problem it's addressing.

Issue #21121, opened 2024-03-17
Flaky tests due to insufficient synchronization in FileHiveMetastore


findepi:
https://github.com/trinodb/trino/actions/runs/8310657389/job/22743572300?pr=21101

```
Error:  io.trino.plugin.hive.TestHiveConnectorTest.testShowTablesLike -- Time elapsed: 0.175 s <<< ERROR!
io.trino.testing.QueryFailedException: Error listing tables for catalog hive: com.google.common.util.concurrent.UncheckedExecutionException: com.google.common.util.concurrent.UncheckedExecutionException: io.trino.spi.TrinoException: Could not read table schema
	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:133)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:501)
	at io.trino.testing.DistributedQueryRunner.execute(DistributedQueryRunner.java:484)
	at io.trino.testing.AbstractTestQueryFramework.computeActual(AbstractTestQueryFramework.java:313)
	at io.trino.testing.AbstractTestQueryFramework.computeActual(AbstractTestQueryFramework.java:308)
	at io.trino.testing.AbstractTestQueries.testShowTablesLike(AbstractTestQueries.java:304)
	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
	at java.base/java.lang.reflect.Method.invoke(Method.java:580)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:728)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:131)
	at org.junit.jupiter.engine.extension.TimeoutExtension.intercept(TimeoutExtension.java:156)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestableMethod(TimeoutExtension.java:147)
	at org.junit.jupiter.engine.extension.TimeoutExtension.interceptTestMethod(TimeoutExtension.java:86)
	at org.junit.jupiter.engine.execution.InterceptingExecutableInvoker$ReflectiveInterceptorCall.lambda$ofVoidMethod$0(InterceptingExecutableInvoker.java:103)
[...]

PR #27459, opened 2025-11-26
Don't share file metastore catalog in tests

findepi:
Update `TestingHiveConnectorFactory` to configure file metastore in such a way that each catalog has its own metastore. The file metastore relies on Java synchronization for orchestrating writes.

- follows https://github.com/trinodb/trino/pull/27452
  - the previous PR fixed sharing between `hive` and `hive_bucketed` catalogs
  - this PR fixes sharing between `hive` and `hive_timestamp_nanos` catalogs

- again fixes https://github.com/trinodb/trino/issues/21121

### DIFF
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java
@@ -42,7 +42,6 @@ import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
 import java.util.function.Function;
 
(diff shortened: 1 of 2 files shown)


plugin/trino-hive/src/test/java/io/trino/plugin/hive/TestingHiveConnectorFactory.java
```diff
@@ -85,6 +76,19 @@ public Connector create(String catalogName, Map<String, String> config, Connecto
         if (metastore.isEmpty() && !config.containsKey("hive.metastore")) {
             configBuilder.put("hive.metastore", "file");
         }
+        Module module = binder -> {
```
losipiuk, line 79: how moving this from constructor here change anything?

findepi: now the metastore's "catalog dir" defaults to catalog name

findepi: applied same change in testing iceberg, delta and hudi connector factories. so i expect more tests to fail where there was intentional but hidden metastore sharing. will fix all of them and make each single-threaded

findepi:
CI
- https://github.com/trinodb/trino/issues/27469

PR #27464, opened 2025-11-26
Switch to pnpm for web-ui

wendigo:
Faster builds :)

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-web-ui/pom.xml
+++ b/core/trino-web-ui/pom.xml
@@ -84,9 +84,9 @@
                 </configuration>
                 <executions>
                     <execution>
-                        <id>install node and npm</id>
+                        <id>install node and pnpm</id>
                         <goals>
-                            <goal>install-node-and-npm</goal>
+                            <goal>install-node-and-pnpm</goal>
                         </goals>
                         <phase>generate-resources</phase>
                         <configuration>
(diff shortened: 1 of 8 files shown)


martint:
Can you add the motivation in the PR description?

mosabua:
I think this is actually a good move ..

wendigo:
@mosabua but doesn't work and even AI can't help :P

mosabua:
We use it in Trino Gateway .. moving is probably a bigger lift .. not to worry for now

wendigo:
@koszti do you have experience with pnpm?

koszti:
I don’t really have a strong preference. Faster builds and installs are nice but I usually stick with the most standard tools. I only install dependencies about once a week anyway so waiting a few extra seconds isn’t a big deal for me with a project of this size. :)

That’s just my personal opinion though and I’m not against pnpm at all. After all it’s a drop-in replacement.

koszti:
@wendigo ah sorry I think you wanted to know why this PR is not working? :) I don’t know :)

mosabua:
I have been using all these tools quite a bit recently with Chainguard Libraries for JavaScript .. if I find a chance to play with it I can look at the migration .. other things are more important though .. like publishing the query ui ;-)

koszti:
> other things are more important though .. like publishing the query ui 

unrelated to this PR, @mosabua since you mentioned the query-ui you may have seen my slack message and the PR to fix. Let me know how it goes or if I can help.

PR #27468, opened 2025-11-27
Clean up dependencies on test jars

martint:
## Release notes

### DIFF
--- a/core/trino-main/pom.xml
+++ b/core/trino-main/pom.xml
@@ -463,13 +463,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
-            <artifactId>trino-plugin-toolkit</artifactId>
-            <type>test-jar</type>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
             <artifactId>trino-testing-services</artifactId>
             <scope>test</scope>
         </dependency>
--- a/plugin/trino-base-jdbc/pom.xml
+++ b/plugin/trino-base-jdbc/pom.xml
@@ -188,13 +188,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
-            <artifactId>trino-cache</artifactId>
-            <type>test-jar</type>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
             <artifactId>trino-main</artifactId>
             <scope>test</scope>
         </dependency>
@@ -209,13 +202,6 @@
         <dependency>
             <groupId>io.trino</groupId>
             <artifactId>trino-parser</artifactId>
-            <scope>test</scope>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-plugin-toolkit</artifactId>
-            <type>test-jar</type>
             <scope>test</scope>
         </dependency>
 
--- a/plugin/trino-delta-lake/pom.xml
+++ b/plugin/trino-delta-lake/pom.xml
@@ -361,13 +361,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
-            <artifactId>trino-filesystem-cache-alluxio</artifactId>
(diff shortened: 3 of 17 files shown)


lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/eventlistener/TestingEventListenerContext.java
```diff
@@ -11,7 +11,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.trino.plugin.base.evenlistener;
```
findepi, line 14: 🤣

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/mapping/testing/RuleBasedIdentifierMappingUtils.java
```diff
@@ -0,0 +1,64 @@
...
+import java.nio.file.Path;
+import java.util.List;
+
+import static io.airlift.json.JsonCodec.jsonCodec;
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.nio.file.Files.createTempFile;
+import static java.nio.file.Files.write;
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+
+public final class RuleBasedIdentifierMappingUtils
```
findepi, line 31: This files appears as new in this commit

> Move test support classes to module source path
TestingEventListenerContext and TestingSystemAccessControlContext are used
for testing by dependent modules, so they should proper classes in the module.

the commit title says it is just _moves_, there should be no code in it.

Probably deletion of `lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/mapping/RuleBasedIdentifierMappingUtils.java` was meant to be part of this commit.

lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/mapping/RuleBasedIdentifierMappingUtils.java
```diff
@@ -1,61 +0,0 @@
...
-import java.nio.file.Path;
-import java.util.List;
-
-import static io.airlift.json.JsonCodec.jsonCodec;
-import static java.nio.charset.StandardCharsets.UTF_8;
-import static java.nio.file.Files.createTempFile;
-import static java.nio.file.Files.write;
-import static java.util.concurrent.TimeUnit.MILLISECONDS;
-
-public final class RuleBasedIdentifierMappingUtils
```
findepi, line 28: ah, and here we go. in `Move RuleBasedIdentifierMappingUtils to module source path` commit, this class shows up as deleted, not moved

(see earlier comment)

core/trino-main/pom.xml
```diff
@@ -461,13 +461,6 @@
             <scope>test</scope>
         </dependency>
 
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-plugin-toolkit</artifactId>
-            <type>test-jar</type>
```
findepi, line 467: ❤️

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/mapping/testing/RuleBasedIdentifierMappingUtils.java
martint: Good catch!

lib/trino-plugin-toolkit/src/test/java/io/trino/plugin/base/eventlistener/TestingEventListenerContext.java
wendigo: odd

chenjian2664: lol

Issue #27386, opened 2025-11-20
Flaky `TestGcsFileSystem`: `The bucket you tried to delete is not empty`


ebyhr:
https://github.com/trinodb/trino/actions/runs/19528166541/job/55905121654?pr=27382
```
Error:  io.trino.filesystem.gcs.TestGcsFileSystem -- Time elapsed: 132.0 s <<< ERROR!
com.google.cloud.storage.StorageException: The bucket you tried to delete is not empty.
	at com.google.cloud.storage.StorageException.translate(StorageException.java:230)
	at com.google.cloud.storage.spi.v1.HttpStorageRpc.translate(HttpStorageRpc.java:360)
	at com.google.cloud.storage.spi.v1.HttpStorageRpc.delete(HttpStorageRpc.java:744)
	at com.google.cloud.storage.StorageImpl.lambda$delete$19(StorageImpl.java:608)
	at com.google.cloud.storage.Retrying$DefaultRetrier.run(Retrying.java:165)
	at com.google.cloud.storage.Retrying$HttpRetrier.run(Retrying.java:206)
	at com.google.cloud.storage.StorageImpl.run(StorageImpl.java:1598)
	at com.google.cloud.storage.StorageImpl.delete(StorageImpl.java:608)
	at io.trino.filesystem.gcs.AbstractTestGcsFileSystem.tearDown(AbstractTestGcsFileSystem.java:73)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:511)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1450)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2019)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
	Suppressed: com.google.cloud.storage.RetryContext$RetryBudgetExhaustedComment: Unretryable error (attempts: 1, maxAttempts: 21, elapsed: PT0.156S, nextBackoff: PT0.013057078S, timeout: PT25S)
Caused by: com.google.api.client.googleapis.json.GoogleJsonResponseException: 409 Conflict
[...]

PR #27473, opened 2025-11-27
Delete all objects from testing GCS bucket

ebyhr:
## Description

Adding extra removal logic though we already have `cleanupFiles` with `@AfterEach`

* Hopefully fixes https://github.com/trinodb/trino/issues/27386

## Release notes

### DIFF
--- a/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
+++ b/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
@@ -13,8 +13,10 @@
  */
 package io.trino.filesystem.gcs;
 
+import com.google.cloud.storage.Blob;
 import com.google.cloud.storage.BlobId;
 import com.google.cloud.storage.BlobInfo;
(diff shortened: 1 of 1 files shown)


lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
```diff
@@ -71,7 +73,11 @@ protected void initialize(String gcpCredentialKey)
     void tearDown()
     {
         try {
-            storage.delete(rootLocation.host().get());
+            Bucket bucket = storage.get(new GcsLocation(rootLocation).bucket());
+            for (Blob blob : bucket.list().iterateAll()) {
+                storage.delete(blob.getBlobId());
```
findepi, line 78: That will break any other test that could be running concurrently (another test class, another build)

findepi: oh, the bucket is new & randomized name

findepi:
/test-with-secrets sha=33c74e73e8be6704d5e2934bcdbd65b0fac4d989

PR #27478, opened 2025-11-27
Enable SNI for single-label domains

lukasz-walkiewicz:
## Description

JDK TLS does not send SNI for single-label domains, which prevents certificates issued for such domains from being used correctly. This commit introduces a workaround to ensure SNI is sent for single-label domains as well.

## Release notes

```markdown
## Trino CLI/JDBC
* Fix TLS connection failures when using unqualified (single-label, without ".") hostnames. Default Java TLS does not send Server Name Indication (SNI) for such hostnames, which could cause certificate mismatches during TLS termination. 
```


client/trino-client/src/test/resources/certs/generate.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/sh
+
+set -eux
+
+openssl req -new -x509 -newkey rsa:4096 -sha256 -nodes -keyout localhost.key -days 3560 -out localhost.crt -config localhost.conf
```
findepi: The idea is that we need to re-run generate.sh every 10 years, correct?

client/trino-client/src/test/java/io/trino/client/TestSSL.java
```diff
@@ -0,0 +1,148 @@
...
+                Optional.empty(),
+                Optional.empty(),
+                false);
+
+        OkHttpClient client = clientBuilder.build();
+        try (Response response = client.newCall(new Request.Builder()
+                        .url("https://localhost:" + httpsPort)
+                        .build())
+                .execute()) {
+            assertThat(response.code()).isEqualTo(OK_200);
```
findepi: I see only successes here.
Should there be a test validating that not every hostname goes through?

client/trino-client/src/test/java/io/trino/client/TestSSL.java
```diff
@@ -0,0 +1,148 @@
...
+        setupSsl(
+                clientBuilder,
+                Optional.empty(),
+                Optional.empty(),
+                Optional.empty(),
+                false,
+                Optional.of(LOCALHOST_CERT),
+                Optional.empty(),
+                Optional.empty(),
+                false);
```
findepi: This looks like testing OkHttpClient, not the trino-client.
Would be possible to anchor this test on some APIs higher up?

findepi:
> (x) Release notes are required, with the following suggested text:
> 
> ```
> ## General
> * Fixed an issue where SNI (Server Name Indication) was not sent for single-label domains.
> ```

I don't understand this. Can you rephrase to the point where even I would understand?

client/trino-client/src/main/java/io/trino/client/OkHttpUtil.java
```diff
@@ -410,4 +416,93 @@ private static GSSContextProvider getDelegatedGSSContextProvider(Optional<GSSCre
                 .map(gssCred -> (GSSContextProvider) gssCred)
                 .orElse(new DelegatedUnconstrainedContextProvider());
     }
+
+    private static class SingleLabelDomainSNISSLSocketFactory
+            extends SSLSocketFactory
```
kokosing, line 423: introduce `ForwardingSSLSocketFactory`

client/trino-client/src/test/java/io/trino/client/TestSSL.java
```diff
@@ -0,0 +1,148 @@
+package io.trino.client;
```
kokosing, line 14: copyright

client/trino-client/src/test/resources/certs/generate.sh
lukasz-walkiewicz: I've used the same validity as: https://github.com/trinodb/trino/blob/master/core/trino-main/src/test/resources/cert/generate.sh
I can make it much longer of course if you want me to.

findepi: you can tune this up to the limit but 10y is ok for me

client/trino-client/src/main/java/io/trino/client/OkHttpUtil.java
lukasz-walkiewicz: What would be a benefit of doing that?

lukasz-walkiewicz:
> > (x) Release notes are required, with the following suggested text:
> > ```
> > ## General
> > * Fixed an issue where SNI (Server Name Indication) was not sent for single-label domains.
> > ```
> 
> I don't understand this. Can you rephrase to the point where even I would understand?

How about:
> Trino clients now sent Server Name Indication (SNI) for single-label domains (like localhost), improving TLS certificate compatibility for these hosts.

?

client/trino-client/src/test/java/io/trino/client/TestSSL.java
lukasz-walkiewicz: I'd say it's a test for `setupSsl ` method which is in `trino-client`

client/trino-client/src/test/resources/certs/generate.sh
lukasz-walkiewicz: Increased it to 100y

client/trino-client/src/main/java/io/trino/client/OkHttpUtil.java
kokosing: It will be easy to recognize what methods are simple pass through, and which methods are changed. This interface is quite simple, so I don't insist.

findepi:
> How about:
> 
> > Trino clients now sent Server Name Indication (SNI) for single-label domains (like localhost), improving TLS certificate compatibility for these hosts.
> 
> ?

```
Trino CLI
- Fix connection error when connecting with server using TLS and unqualified host name (without ".") 
  when internal communication also uses TLS.
```

?

lukasz-walkiewicz:
Thanks for the suggestion. This issue is not limited to cases where internal communication is over SSL however. It can also happen when the server has multiple SSL certificates for multiple domains and fails to match them accordingly due to Java not sending SNI for single-label domains. I've changed release notes, I hope it's better now.

PR #27480, opened 2025-11-27
 Rename castToVarchar to render in ValuePrinter

hqbhoho:
## Description
Remove unused method castToVarcharOrFail  and rename castToVarchar to render in ValuePrinter 
## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/planner/planprinter/ValuePrinter.java
+++ b/core/trino-main/src/main/java/io/trino/sql/planner/planprinter/ValuePrinter.java
@@ -38,7 +38,7 @@ public final class ValuePrinter
         this.session = requireNonNull(session, "session is null");
     }
 
-    public String castToVarchar(Type type, Object value)
+    public String render(Type type, Object value)
     {
         try {
             if (value == null) {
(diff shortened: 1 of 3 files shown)


findepi:
CI
- https://github.com/trinodb/trino/issues/27483
- https://github.com/trinodb/trino/issues/27484

core/trino-main/src/main/java/io/trino/sql/planner/planprinter/ValuePrinter.java
```diff
@@ -38,25 +38,19 @@ public ValuePrinter(Metadata metadata, FunctionManager functionManager, Session
         this.session = requireNonNull(session, "session is null");
     }
 
-    public String castToVarchar(Type type, Object value)
+    public String toString(Type type, Object value)
```
findepi: cc @martint

findepi (approved):

core/trino-main/src/main/java/io/trino/sql/planner/planprinter/ValuePrinter.java
martint: Let’s call it “render”. toString has a couple of issues:
* it’s an overload of Object’s toString(), which can cause confusion due to different semantics.
* It seems to imply that it can convert any value of a type to a string, which is not true. It only attempts to produce a rendering of the value so that it can be embedded in the plan.

hqbhoho: fixed

Issue #27082, opened 2025-10-24
Flaky `TestDirectTrinoClient.testDirectTrinoClientLongQuery`


ebyhr:
```
Error:  io.trino.client.direct.TestDirectTrinoClient.testDirectTrinoClientLongQuery -- Time elapsed: 1.892 s <<< ERROR!
io.trino.testing.QueryFailedException: Query 20251024_015052_00002_aqn7d was abandoned by the client, as it may have exited or stopped checking for query results. Query results have not been accessed since 2025-10-24T01:50:53.071504965Z: currentTime 2025-10-24T01:50:54.344700484Z
	at io.trino.testing.TestingDirectTrinoClient.toMaterializedRows(TestingDirectTrinoClient.java:79)
	at io.trino.testing.TestingDirectTrinoClient.lambda$execute$0(TestingDirectTrinoClient.java:67)
	at io.trino.testing.StandaloneQueryRunner.execute(StandaloneQueryRunner.java:112)
	at io.trino.client.direct.TestDirectTrinoClient.testDirectTrinoClientLongQuery(TestDirectTrinoClient.java:62)
Caused by: io.trino.spi.TrinoException: Query 20251024_015052_00002_aqn7d was abandoned by the client, as it may have exited or stopped checking for query results. Query results have not been accessed since 2025-10-24T01:50:53.071504965Z: currentTime 2025-10-24T01:50:54.344700484Z
	at io.trino.execution.QueryTracker.failAbandonedQueries(QueryTracker.java:276)
	at io.trino.execution.QueryTracker.lambda$start$0(QueryTracker.java:83)
	at com.google.common.util.concurrent.MoreExecutors$ScheduledListeningDecorator$NeverSuccessfulListenableFutureTask.run(MoreExecutors.java:639)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:369)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:310)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1474)
```

[...]

Issue #14526, opened 2022-10-07
Flaky test TestHttpRemoteTask.testRejectedExecution 


findepi:
https://github.com/trinodb/trino/actions/runs/3205780736/jobs/5240103859

```
Error:  io.trino.server.remotetask.TestHttpRemoteTask.testRejectedExecution  Time elapsed: 4.194 s  <<< FAILURE!
java.lang.AssertionError: expected [REMOTE_TASK_ERROR:65542] but found [GENERIC_INTERNAL_ERROR:65536]
	at org.testng.Assert.fail(Assert.java:94)
	at org.testng.Assert.failNotEquals(Assert.java:513)
	at org.testng.Assert.assertEqualsImpl(Assert.java:135)
	at org.testng.Assert.assertEquals(Assert.java:116)
	at org.testng.Assert.assertEquals(Assert.java:179)
	at io.trino.server.remotetask.TestHttpRemoteTask.runTest(TestHttpRemoteTask.java:398)
	at io.trino.server.remotetask.TestHttpRemoteTask.testRejectedExecution(TestHttpRemoteTask.java:167)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:104)
	at org.testng.internal.InvokeMethodRunnable.runOne(InvokeMethodRunnable.java:54)
	at org.testng.internal.InvokeMethodRunnable.run(InvokeMethodRunnable.java:44)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
	at java.base/java.lang.Thread.run(Thread.java:833)
```

piotrrzysko:
[...]

PR #27488, opened 2025-11-27
Fix testDirectTrinoClientLongQuery

piotrrzysko:
## Description

Attempt to fix https://github.com/trinodb/trino/issues/27082.

Record heartbeats more frequently to prevent query timeouts during large
result sets or when the query state changes more often than the heartbeat
interval.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
+++ b/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java
@@ -110,6 +110,8 @@ public class DirectTrinoClient
                             !(dispatchQuery.getState() == FINISHING && dispatchQuery.getFullQueryInfo().getStages().isEmpty());
                     state = queryManager.getQueryState(queryId)) {
                 for (Slice serializedPage = exchangeClient.pollPage(); serializedPage != null; serializedPage = exchangeClient.pollPage()) {
+                    // record heartbeat for each page to avoid query timeout during large result sets
+                    dispatchQuery.recordHeartbeat();
                     Page page = pageDeserializer.deserialize(serializedPage);
                     queryResultsListener.consumeOutputPage(page);
                 }
@@ -146,6 +148,8 @@ public class DirectTrinoClient
         while (!anyCompleteFuture.isDone()) {
             try {
                 anyCompleteFuture.get(heartBeatIntervalMillis, TimeUnit.MILLISECONDS);
+                // some time elapsed, so record a heartbeat
+                dispatchQuery.recordHeartbeat();
             }
             catch (TimeoutException e) {
                 // continue waiting until the query state changes or the exchange client is blocked.


findepi:
> As an experiment, I reverted the bugfix that this test is validating, and the test failed, so the intended behavior is still covered.

Thanks for doing that!

I understand this refers to https://github.com/trinodb/trino/pull/26371 where the test was added

that PR desc:

> When a query being executed by DirectTrinoClient takes longer than the query.client.timeout, the query is then canceled. This client is used in some cases when executing queries from within the coordinator

this PR desc

> Considering that processing the last page took 0.4s, it’s easy to imagine that sometimes the total interval between sending heartbeats exceeds 1s.


This test uses `blackhole.test_schema.slow_test_table` which produces pages slowly, but actually any other query/connector can produces pages slowly. for example aggregation will read all input before producing a page.
If pages coming slowly to `DirectTrinoClient` make query cancelled and abandoned, then it sounds like the bug is still there?

piotrrzysko:
Polling pages is non-blocking (if there are no pages, `pollPage` returns null immediately): https://github.com/trinodb/trino/blob/f1cdafa916b9bac59106d093ba5f881cd8b95c6f/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java#L112

The timings I attached mainly reflect deserialization:
https://github.com/trinodb/trino/blob/f1cdafa916b9bac59106d093ba5f881cd8b95c6f/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java#L113

My understanding is that if no new page is available, `DirectTrinoClient` immediately starts waiting on this future: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/client/direct/DirectTrinoClient.java#L12. If it doesn’t complete within `query.client.timeout / 2`, it sends a heartbeat and this repeats until the future completes.


The timeout is set to a very low value (1s), which is fine for tests, but in production it should be higher (the default is 5 minutes) to avoid getting close to the deserialization time.

testing/trino-tests/src/test/java/io/trino/client/direct/TestDirectTrinoClient.java
```diff
@@ -59,6 +59,6 @@ public void setup()
     @Timeout(value = 20, unit = TimeUnit.SECONDS)
     public void testDirectTrinoClientLongQuery()
     {
-        queryRunner.execute(TEST_SESSION, "SELECT * FROM blackhole.test_schema.slow_test_table");
+        queryRunner.execute(TEST_SESSION, "EXPLAIN ANALYZE SELECT * FROM blackhole.test_schema.slow_test_table");
```
findepi: would `SELECT count(*) FROM blackhole.test_schema.slow_test_table` have the same effect?

piotrrzysko:
@findepi I applied what we discussed offline. The test didn’t fail locally after running it in a loop for an hour - previously it would fail sporadically.

piotrrzysko:
CI failure doesn’t seem related. It looks like this: https://github.com/trinodb/trino/issues/14526

PR #27490, opened 2025-11-27
Add sdkmanrc with current Java version

wendigo:
## Release notes

```markdown
## Section
```


.sdkmanrc
```diff
@@ -0,0 +1 @@
+java=25.0.1-tem
```
findepi, line 1: How do we keep various places in sync?
should we have a list of places to update when bumping JDK?

wendigo: Yes. We can list them

wendigo:
@electrum I agree but that list is short. I don't use .java-version so I could raise a point that it's not useful for me. Same goes for .editorconfig. Why is one preferred over the other?

electrum:
The list is short because we worked hard to make it short and we are aggressive about keeping it short. If we say that "it's short, one more thing won't hurt" then it soon won't be short.

I see two options here:

1. If sdkman is better than jenv, then convince the other developers who use jenv to switch, then swap/replace the config file.
2. Send a PR to one or both of the projects to be compatible with the other one.

electrum:
`.java-version` is "preferred" because it was the first one added, by people who used it at the time. There wasn't any discussion about adding one or the other.

(I personally don't see the usefulness of having either one, as I use the latest Java as my default.)

`.editorconfig` is there for IntelliJ to configure things that for some reason can't set in code style (or that was true in the past), so you probably do use it.

PR #27491, opened 2025-11-27
Clean ConnectorMetadata deprecations

findepi:


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcMetadataConfig.java
```diff
@@ -71,11 +71,13 @@ public JdbcMetadataConfig setJoinPushdownEnabled(boolean joinPushdownEnabled)
...
 
+    @Deprecated
     public boolean isComplexJoinPushdownEnabled()
     {
         return complexJoinPushdownEnabled;
     }
 
+    @Deprecated
     @Config("join-pushdown.with-expressions")
     @ConfigDescription("Enable join pushdown with complex expressions")
```
findepi: cc @anusudarsan @findinpath @kokosing -- see `Deprecate join-pushdown.with-expressions config property` commit

findepi: moved out to https://github.com/trinodb/trino/pull/27498

PR #27500, opened 2025-11-28
Simplify cleanup logic for renaming tests in `BaseConnectorSmokeTest`

chenjian2664:
## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java
@@ -448,9 +448,8 @@ public abstract class BaseConnectorSmokeTest
             assertUpdate("ALTER TABLE " + oldTable + " RENAME TO " + newTable);
         }
         catch (Throwable e) {
-            try (AutoCloseable ignore = () -> assertUpdate("DROP TABLE " + oldTable)) {
-                throw e;
-            }
+            assertUpdate("DROP TABLE " + oldTable);
+            throw e;
         }
 
         assertThat(query("SHOW TABLES LIKE '" + oldTable + "'"))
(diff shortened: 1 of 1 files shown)


testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java
```diff
@@ -448,9 +448,8 @@ public void testRenameTable()
             assertUpdate("ALTER TABLE " + oldTable + " RENAME TO " + newTable);
         }
         catch (Throwable e) {
-            try (AutoCloseable ignore = () -> assertUpdate("DROP TABLE " + oldTable)) {
-                throw e;
-            }
+            assertUpdate("DROP TABLE " + oldTable);
+            throw e;
```
ebyhr, line 452: I think we should change more. The table will remain if the following assertion with `SHOW TABLES` or `SELECT` failed for some reason. 

Same for `testRenameTableAcrossSchemas`.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java
```diff
@@ -448,9 +448,8 @@ public void testRenameTable()
             assertUpdate("ALTER TABLE " + oldTable + " RENAME TO " + newTable);
         }
         catch (Throwable e) {
-            try (AutoCloseable ignore = () -> assertUpdate("DROP TABLE " + oldTable)) {
-                throw e;
```
findepi, line 452: The point here was to preserve `e` even if the DROP fails

chenjian2664: I don't think the "e" will be thrown if the DROP fails, did you want to put it in the finally? :
```
 try (AutoCloseable ignore = () -> assertUpdate("DROP TABLE " + oldTable)) {
}
finally {
    throw e;
} 
```

findepi: Can you test it locally?
```java
try (AutoCloseable ignore = () -> { throw ...  } ) {
                throw e;
}
```

chenjian2664: you are right. the "e" still be thrown even the DROP fails, good to learn.

Issue #27458, opened 2025-11-26
Building trino-web-ui sometimes fails on CI


findepi:
Building trino-web-ui sometimes fails on CI, making tests flaky 

https://github.com/trinodb/trino/actions/runs/19701684956/job/56439378400?pr=27453

```
[INFO] --- frontend:1.15.4:npm (package (webapp)) @ trino-web-ui ---
[INFO] Running 'npm run package:clean' in /home/runner/work/trino/trino/core/trino-web-ui/src/main/resources/webapp/src
[INFO] Recompiling the module because of changed source code.
[INFO] Recompiling the module because of changed source code.
[INFO] Compiling 14 source files with javac [debug deprecation parameters release 25] to target/classes
[INFO] Compiling 28 source files with javac [debug deprecation parameters release 25] to target/classes
[INFO] 
[INFO] > trino-webui@0.0.1 package:clean
[INFO] > npm clean-install && webpack --config webpack.config.js
[INFO] 

...

Error:  Failed to execute goal com.github.eirslett:frontend-maven-plugin:1.15.4:npm (package (webapp)) on project trino-web-ui: Failed to run task: 'npm run package:clean' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1]
```

findepi:
https://github.com/trinodb/trino/actions/runs/19743353503/job/56572433487?pr=27481

JohnFitzpatrick44:
```
[INFO] npm error code ECONNRESET
[INFO] npm error network aborted
[INFO] npm error network This is a problem related to network connectivity.
[INFO] npm error network In most cases you are behind a proxy or have bad network settings.
[INFO] npm error network
[INFO] npm error network If you are behind a proxy, please make sure that the
[INFO] npm error network 'proxy' config is set properly.  See: 'npm help config'
[INFO] npm error A complete log of this run can be found in: /home/runner/.npm/_logs/2025-11-26T11_12_24_897Z-debug-0.log
```

JohnFitzpatrick44:
Because these errors are rare, it seems that transient network errors are occasionally causing problems with clean npm installs. Npm already makes [3 attempts by default](https://docs.npmjs.com/cli/v11/using-npm/config#fetch-retries), but we could expand that number (and possibly the timeout min/max).

It's [unclear](https://github.com/npm/cli/issues/4159) whether npm can successfully retry after ECONNRESET, though.

[...]

Issue #26424, opened 2025-08-16
Flaky `TestIcebergParquetWithBloomFiltersMixedCase.testBloomFilterColumnWithDictionaryPage`


ebyhr:
https://github.com/trinodb/trino/actions/runs/17004812202/job/48212902847

```
Error:  io.trino.plugin.iceberg.TestIcebergParquetWithBloomFiltersMixedCase.testBloomFilterColumnWithDictionaryPage -- Time elapsed: 3.424 s <<< ERROR!
io.trino.testing.QueryFailedException: Failed to open input stream for file: s3://test-bucket-mixed-case/mixed_case_bloom_filter/metadata/snap-4355940061149223898-1-16be2c8e-cea2-4385-b80a-cb89ced57fd4.avro
	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:138)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:586)
	at io.trino.testing.DistributedQueryRunner.executeWithPlan(DistributedQueryRunner.java:575)
	at io.trino.testing.QueryAssertions.assertDistributedUpdate(QueryAssertions.java:106)
	at io.trino.testing.QueryAssertions.assertUpdate(QueryAssertions.java:60)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:426)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:421)
	at io.trino.plugin.iceberg.TestIcebergParquetWithBloomFiltersMixedCase.createParquetTableWithBloomFilter(TestIcebergParquetWithBloomFiltersMixedCase.java:84)
	at io.trino.testing.BaseTestParquetWithBloomFilters.testBloomFilterColumnWithDictionaryPage(BaseTestParquetWithBloomFilters.java:58)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:507)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1394)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1970)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
[...]

PR #27502, opened 2025-11-28
Support MATERIALIZED VIEW `WHEN STALE` in analyzer

piotrrzysko:
## Description

Follow-up to https://github.com/trinodb/trino/pull/27356

## Release notes

```markdown
## General
* Add the `WHEN STALE [INLINE | FAIL]` clause to `CREATE MATERIALIZED VIEW`. ({issue}`27502`)

## Iceberg connector
* Add support for the `WHEN STALE` clause in the Iceberg connector. ({issue}`27502`)

## Lakehouse connector
* Add support for the `WHEN STALE` clause in the Lakehouse connector. ({issue}`27502`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
+++ b/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
@@ -60,6 +60,7 @@ import static io.trino.spi.connector.ConnectorCapabilities.MATERIALIZED_VIEW_WHE
 import static io.trino.sql.SqlFormatterUtil.getFormattedSql;
 import static io.trino.sql.analyzer.ConstantEvaluator.evaluateConstant;
 import static io.trino.sql.analyzer.SemanticExceptions.semanticException;
+import static io.trino.sql.tree.CreateMaterializedView.WhenStaleBehavior.INLINE;
 import static io.trino.type.IntervalDayTimeType.INTERVAL_DAY_TIME;
 import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
@@ -159,11 +160,11 @@ public class CreateMaterializedViewTask
                 });
 
         Optional<WhenStaleBehavior> whenStale = statement.getWhenStaleBehavior()
-                .map(_ -> {
-                    if (!plannerContext.getMetadata().getConnectorCapabilities(session, catalogHandle).contains(MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR)) {
+                .map(whenStaleBehavior -> {
+                    if (!isWhenStaleBehaviorSupported(session, whenStaleBehavior, catalogHandle)) {
                         throw semanticException(NOT_SUPPORTED, statement, "Catalog '%s' does not support WHEN STALE", catalogName);
                     }
-                    throw semanticException(NOT_SUPPORTED, statement, "WHEN STALE is not supported yet");
+                    return toConnectorWhenStaleBehavior(whenStaleBehavior);
                 });
 
         MaterializedViewDefinition definition = new MaterializedViewDefinition(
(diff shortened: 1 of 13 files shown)


testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -1598,21 +1598,207 @@ public void testMaterializedViewGracePeriod()
...
+            assertUpdate("INSERT INTO " + baseTable + " (regionkey, name) VALUES (42, 'foo new region')", 1);
+            String updatedResults = initialResults + " UNION ALL VALUES (42, 'foo new region')";
+
+            // The materialization is stale now
+            assertThat(query(defaultSession, "TABLE " + viewName))
+                    .hasPlan(readFromStorageTable)
+                    .matches(initialResults);
+            assertThat(query(futureSession, "TABLE " + viewName))
+                    .hasPlan(readFromBaseTables)
+                    .matches(updatedResults);
```
piotrrzysko, line 1697: @findepi I'm moving your [comment](https://github.com/trinodb/trino/pull/27356#discussion_r2571186373) to this PR so it doesn’t get lost:

> The MV has GRACE PERIOD INTERVAL '1' HOUR.
The GRACE is defined earlier in the MV definition and should take precedence. I.e. we should keep querying the MV storage when it's known to be stale but grace period did not elapse.

piotrrzysko: I believe that in this case we should query the base tables, because we use a session (`futureSession`) that started after the grace period expired.

core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
```diff
@@ -193,4 +194,17 @@ Analysis executeInternal(
         plannerContext.getMetadata().createMaterializedView(session, name, definition, properties, statement.isReplace(), statement.isNotExists());
         return analysis;
     }
+
+    private boolean isWhenStaleBehaviorSupported(Session session, CreateMaterializedView.WhenStaleBehavior whenStaleBehavior, CatalogHandle catalogHandle)
+    {
+        return whenStaleBehavior == INLINE || plannerContext.getMetadata().getConnectorCapabilities(session, catalogHandle).contains(MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR);
```
findepi, line 201: add explanaory code comment

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -564,12 +566,21 @@ private Query showCreateMaterializedView(ShowCreate node)
                     false,
                     viewDefinition.get().getGracePeriod()
                             .map(DateTimeUtils::formatDayTimeInterval),
-                    Optional.empty(), // TODO support WHEN STALE
+                    viewDefinition.get().getWhenStaleBehavior()
+                            .map(Visitor::toSqlWhenStaleBehavior),
```
findepi: given INLINE is the default, we could `viewDefinition.get().getWhenStaleBehavior().orElse(INLINE)`

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -564,12 +567,20 @@ private Query showCreateMaterializedView(ShowCreate node)
                     false,
                     viewDefinition.get().getGracePeriod()
                             .map(DateTimeUtils::formatDayTimeInterval),
-                    Optional.empty(), // TODO support WHEN STALE
+                    Optional.of(toSqlWhenStaleBehavior(viewDefinition.get().getWhenStaleBehavior().orElse(INLINE))),
```
findepi, line 570: we should change the CreateMaterializedView field (in the non-connector class) to be non-Optional.

we could also change that in ConnectorMaterializedViewDefinition. in fact, backwards compatibility was broken, as the ConnectorMaterializedViewDefinition constructor already requires passing in `Optional<WhenStaleBehavior>`, so it could as well take `WhenStaleBehavior`

findepi: let's follow-up

piotrrzysko:
CI:
* https://github.com/trinodb/trino/issues/26424
* https://github.com/trinodb/trino/issues/27458

findepi:
> CI: [...]

thanks! make sure each of these issues is updated (comment). this helps understand most heavy offenders and have them fixed first

i retriggered the failed builds

chenjian2664:
For the release note about Lakehouse, is it for all lakehouse or just Iceberg, should we mention which connector supports it under the Lakehouse release entry

findepi:
> For the release note about Lakehouse, is it for all lakehouse or just Iceberg

I think it's simply for Iceberg and Lakehouse connectors.

> should we mention which connector supports it under the Lakehouse release entry


no

PR #27503, opened 2025-11-28
Add missing defaultValue for ColumnMetadata.Builder

hqbhoho:
## Description
Add missing defaultValue for ColumnMetadata.Builder

## Release notes
```
SPI

- Fix `ColumnMetadata.builderFrom` to retain column default value.
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/ColumnMetadata.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/ColumnMetadata.java
@@ -43,7 +43,8 @@ public class ColumnMetadata
         this(name, type, Optional.empty(), true, null, null, false, emptyMap());
     }
 
-    private ColumnMetadata(
+    // VisibleForTesting
+    ColumnMetadata(
             String name,
             Type type,
             Optional<String> defaultValue,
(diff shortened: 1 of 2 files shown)


core/trino-spi/src/main/java/io/trino/spi/connector/ColumnMetadata.java
```diff
@@ -190,6 +190,7 @@ private Builder(ColumnMetadata columnMetadata)
         {
             this.name = columnMetadata.getName();
             this.type = columnMetadata.getType();
+            this.defaultValue = columnMetadata.getDefaultValue();
```
findepi, line 194: just `this.defaultValue = columnMetadata.getDefaultValue();` is enough (pushed a fix to your branch)

can you please add a test that `Builder(ColumnMetada).build()` roundtrips?

core/trino-spi/src/test/java/io/trino/spi/connector/TestColumnMetadata.java
```diff
@@ -0,0 +1,46 @@
...
+        ColumnMetadata originColumnMetadata = ColumnMetadata.builder()
+                .setName("test_column")
+                .setType(INTEGER)
+                .setDefaultValue(Optional.of("1"))
+                .setNullable(false)
+                .setComment(Optional.ofNullable("test_comment"))
+                .setExtraInfo(Optional.of("test_extra_info"))
+                .setHidden(false)
+                .setProperties(ImmutableMap.of("test_key", "test_value"))
+                .build();
```
findepi: Make `ColumnMetadata` constructor package private, `@VisibleForTesting` and call here.
This would ensure the test stays up to date, should there any new fields be added.

hqbhoho: ok, `@VisibleForTesting` will need guava dependency. I add a doc rather than add dependency. what do you think?

PR #27506, opened 2025-11-28
Remove prefer_streaming_operators property

raunaqmorarka:
## Description
This is an old undocumented property that doesn't have a clear use case. It's effect can be achieved by setting task_concurrency to 1 as well

## Release notes

```markdown
## General
* Removed `prefer_streaming_operators` session property. `SET SESSION task_concurrency=1` can be used to achieve the same behaviour as `prefer_streaming_operators`. ({issue}`27506`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
+++ b/core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
@@ -68,7 +68,6 @@ public final class SystemSessionProperties
     public static final String MAX_HASH_PARTITION_COUNT = "max_hash_partition_count";
     public static final String MIN_HASH_PARTITION_COUNT = "min_hash_partition_count";
     public static final String MIN_HASH_PARTITION_COUNT_FOR_WRITE = "min_hash_partition_count_for_write";
-    public static final String PREFER_STREAMING_OPERATORS = "prefer_streaming_operators";
     public static final String TASK_MIN_WRITER_COUNT = "task_min_writer_count";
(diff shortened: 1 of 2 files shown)


core/trino-main/src/main/java/io/trino/SystemSessionProperties.java
```diff
@@ -300,11 +299,6 @@ public SystemSessionProperties(
                         queryManagerConfig.getMinHashPartitionCountForWrite(),
                         value -> validateIntegerValue(value, MIN_HASH_PARTITION_COUNT_FOR_WRITE, 1, false),
                         false),
-                booleanProperty(
```
wendigo, line 303: This is breaking change if that session property was used. I think that we should create a mechanism that basically adds `defunctProperty` that will generate a warning when set. Does it make sense? 

cc @findepi

raunaqmorarka: So far we have not given the "breaking changes"/defunct'ing treatment to session properties.
If we want to start doing that and prepare the framework to support that, then that's a separate discussion to be had in #core-dev

findepi (approved):

Praveen2112:
But we might need to update the release notes right ? I don't think this is a hidden session property.

martint:
This was from when we supported multiple table layouts. The goal was to be able to force the optimizer to choose a table layout that would allow running with streaming aggregation or window operators, which would result in approximately constant memory allocation. 

By picking a layout with data organized along the same dimensions as the aggregation and window function partitioning keys, the operators could compute a group at a time and flush the results once the group was complete.

PR #27509, opened 2025-11-28
Migrate ThreadLocal to ScopedValue usage

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-metastore/src/main/java/io/trino/metastore/cache/ReentrantBoundedExecutor.java
+++ b/lib/trino-metastore/src/main/java/io/trino/metastore/cache/ReentrantBoundedExecutor.java
@@ -17,6 +17,7 @@ import io.airlift.concurrent.BoundedExecutor;
 
 import java.util.concurrent.Executor;
 
+import static java.lang.ScopedValue.where;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -26,7 +27,7 @@ import static java.util.Objects.requireNonNull;
 public class ReentrantBoundedExecutor
         implements Executor
 {
-    private final ThreadLocal<Boolean> executorThreadMarkers = ThreadLocal.withInitial(() -> false);
+    private final ScopedValue<Void> executorThreadMarkers = ScopedValue.newInstance();
     private final Executor boundedExecutor;
     private final Executor coreExecutor;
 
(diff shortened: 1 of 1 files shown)


findepi:
`ThreadLocal`s (TL) distinguish between inheritable (`InheritableThreadLocal`) and non-inheritable (the default). Only the non-inheritable TLs behave reasonably, especially when thread pools are involved.  The inheritable TLs definitely have some safe application patterns, but I am yet to learn them.

`ScopedValue` are a much more convenient API than `ThreadLocal`. (`ScopedValue<T>` is like `ThreadLocal<Stack<T>>` with nice API).
However, they are not a safe replacement for TL. This is because `ScopedValue` are always inheritable (via `StructuredTaskScope`).

Here is a program demonstrating that `ReentrantBoundedExecutor` using `ScopedValue` no longer enforces its bounds.

```java
static void main()
{
    ReentrantBoundedExecutor executor = new ReentrantBoundedExecutor(
            Executors.newCachedThreadPool(),
            1);

    executor.execute(() -> {
        // This looks rogue, but that can be some other somewhere else using StructuredTaskScope to transport some other ScopedValue to some other place
        try (var scope = StructuredTaskScope.open()) {
            CyclicBarrier barrier = new CyclicBarrier(2);
            var first = scope.fork(() -> {
                executor.execute(() -> {
                    System.out.println("currentThread().getName() = " + currentThread().getName());
                    try {
                        barrier.await();
                    }
                    catch (Exception e) {
                    }
                    System.out.println("two threads are in");
                });
            });
            var second = scope.fork(() -> {
                executor.execute(() -> {
                    System.out.println("currentThread().getName() = " + currentThread().getName());
                    try {
                        barrier.await();
                    }
                    catch (Exception e) {
                    }
                    System.out.println("two threads are in");
                });
            });
            scope.join();
        }
        catch (Exception e) {
        }
    });
}
```

lib/trino-metastore/src/main/java/io/trino/metastore/cache/ReentrantBoundedExecutor.java
```diff
@@ -26,7 +27,7 @@
 public class ReentrantBoundedExecutor
         implements Executor
 {
-    private final ThreadLocal<Boolean> executorThreadMarkers = ThreadLocal.withInitial(() -> false);
+    private final ScopedValue<Void> executorThreadMarkers = ScopedValue.newInstance();
```
chenjian2664, line 30: Just got to know the usage of it 👍

findepi:
> Here is a program demonstrating that `ReentrantBoundedExecutor` using `ScopedValue` no longer enforces its bounds.

To be honest, `ReentrantBoundedExecutor` is pretty weak about enforcing its bounds -- that's the difference between `ReentrantBoundedExecutor` and regular `BoundedExecutor`. A task executing inside `ReentrantBoundedExecutor` has a free pass to insert arbitrary number of additional tasks into the executor. This was true before the changes and is true after. What this changed, is that there is another avenue for inserting arbitrary number of additional tasks -- via the `StructuredTaskScope` (preview API in 25).

PR #27513, opened 2025-12-01
Improve performance when parsing invalid queries

electrum:
Use `BailErrorStrategy` as recommended by the ANTLR team in https://github.com/antlr/antlr4/issues/192

## Release notes

### DIFF
--- a/core/trino-parser/src/main/java/io/trino/sql/jsonpath/PathParser.java
+++ b/core/trino-parser/src/main/java/io/trino/sql/jsonpath/PathParser.java
@@ -19,16 +19,19 @@ import io.trino.grammar.jsonpath.JsonPathParser;
 import io.trino.sql.jsonpath.tree.PathNode;
 import io.trino.sql.parser.ParsingException;
 import io.trino.sql.tree.NodeLocation;
+import org.antlr.v4.runtime.BailErrorStrategy;
 import org.antlr.v4.runtime.BaseErrorListener;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.CommonToken;
 import org.antlr.v4.runtime.CommonTokenStream;
+import org.antlr.v4.runtime.DefaultErrorStrategy;
 import org.antlr.v4.runtime.ParserRuleContext;
 import org.antlr.v4.runtime.RecognitionException;
 import org.antlr.v4.runtime.Recognizer;
 import org.antlr.v4.runtime.Token;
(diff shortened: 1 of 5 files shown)


core/trino-parser/src/main/java/io/trino/sql/parser/SqlParser.java
```diff
@@ -144,42 +146,27 @@ private Node invokeParser(String name, String sql, Optional<NodeLocation> locati
...
                 try {
                     // first, try parsing with potentially faster SLL mode
                     parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
+                    parser.setErrorHandler(new BailErrorStrategy());
                     tree = parseFunction.apply(parser);
                 }
-                catch (ParsingException ex) {
+                catch (ParseCancellationException e) {
                     // if we fail, parse with LL mode
-                    tokenStream.seek(0); // rewind input stream
```
martint: Why did this disappear?

electrum:
This is more about doing it in the correct way, that could theoretically improve performance. I don't actually care about the performance and didn't bother to benchmark it. This could alternatively be titled

> Use BailErrorStrategy for ANTLR two-stage parsing

core/trino-parser/src/main/java/io/trino/sql/parser/SqlParser.java
electrum: The `Parser.reset()` method already does this

core/trino-parser/src/main/java/io/trino/sql/jsonpath/PathParser.java
```diff
@@ -95,20 +98,20 @@ public PathNode parseJsonPath(String path)
             lexer.addErrorListener(errorListener);
 
             parser.removeErrorListeners();
-            parser.addErrorListener(errorListener);
 
             ParserRuleContext tree;
             try {
                 // first, try parsing with potentially faster SLL mode
                 parser.getInterpreter().setPredictionMode(PredictionMode.SLL);
+                parser.setErrorHandler(new BailErrorStrategy());
```
findepi, line 106: please add a code comment here & at `parser.setErrorHandler(new DefaultErrorStrategy());` line below

electrum: What comment? And in every file?

I prefer not to clutter the code with unnecessary comments as that generally decreases readability. The Javadoc for BailErrorStrategy already describes two-stage parsing.

PR #27516, opened 2025-12-01
Add BigQuery retries on transient UNAVAILABLE errors

vlad-lyutenko:
## Description
We observe flaky UNAVAILABLE errors, during read session creation, which should be retried
From Google documentation: Status.UNAVAILABLE is a transient network or backend condition and should be retried.

```
Caused by: io.trino.testing.QueryFailedException: Cannot create read sessionio.grpc.StatusRuntimeException: UNAVAILABLE: Connection to server broken (OnChannelError)
	at io.trino.testing.DistributedQueryRunner.executeWithQueryId(DistributedQueryRunner.java:634)
	at io.trino.testing.QueryAssertions.assertDistributedQuery(QueryAssertions.java:289)
	... 11 more
	Suppressed: java.lang.Exception: SQL: SELECT orderkey, totalprice FROM (SELECT orderkey, totalprice FROM orders ORDER BY 1, 2 LIMIT 10) ORDER BY 2, 1 LIMIT 5
```

## Release notes

### DIFF
--- a/plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
+++ b/plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
@@ -51,9 +51,12 @@ public final class BigQueryUtil
     private static boolean isRetryableInternalError(Throwable t)
     {
         if (t instanceof StatusRuntimeException statusRuntimeException) {
-            return statusRuntimeException.getStatus().getCode() == Status.Code.INTERNAL &&
+            return (statusRuntimeException.getStatus().getCode() == Status.Code.INTERNAL &&
                     INTERNAL_ERROR_MESSAGES.stream()
-                            .anyMatch(message -> statusRuntimeException.getMessage().contains(message));
+                            .anyMatch(message -> statusRuntimeException.getMessage().contains(message))) ||
+                    // from Google documentation: UNAVAILABLE - This is most likely a transient condition, which can be corrected by retrying with a backoff.
+                    // https://docs.cloud.google.com/bigquery/docs/reference/datatransfer/rest/v1/Code
+                    statusRuntimeException.getStatus().getCode() == Status.Code.UNAVAILABLE;
         }
         return false;
     }


plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
```diff
@@ -41,6 +41,11 @@ public final class BigQueryUtil
             "Received unexpected EOS on DATA frame from server",
             "INTERNAL: request failed: internal error");
 
+    // we observe flaky UNAVAILABLE errors, during read session creation, which should be retried
```
ebyhr: `isRetryable` is called from `ReadRowsHelper` which is not read session creation, right?

plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
```diff
@@ -41,6 +41,11 @@ public final class BigQueryUtil
...
+    // from Google documentation: Status.UNAVAILABLE is a transient network or backend condition and should be retried.
```
ebyhr: Could you add a link to the documentation?

plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
vlad-lyutenko: Yes, refactored in the way to retry only on read session creation, which we observed on CI/CD failures.

plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
```diff
@@ -58,6 +75,16 @@ private static boolean isRetryableInternalError(Throwable t)
         return false;
     }
 
+    private static boolean isUnavailableErrorWithRetryableMessage(Throwable t)
+    {
+        if (t instanceof StatusRuntimeException statusRuntimeException) {
+            return statusRuntimeException.getStatus().getCode() == Status.Code.UNAVAILABLE &&
+                    UNAVAILABLE_ERROR_MESSAGES.stream()
+                            .anyMatch(message -> statusRuntimeException.getMessage().contains(message));
```
findepi: - `statusRuntimeException.getMessage()` might be null
- why match on the exception message at all?

plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/ReadSessionCreator.java
```diff
@@ -123,7 +123,8 @@ public ReadSession create(ConnectorSession session, TableId remoteTable, List<Bi
                             .withMaxRetries(maxCreateReadSessionRetries)
                             .withBackoff(10, 500, MILLIS)
                             .onRetry(event -> log.debug("Request failed, retrying: %s", event.getLastException()))
-                            .handleIf(BigQueryUtil::isRetryable)
+                            .handleIf(BigQueryUtil::isRetryableInternalError)
+                            .handleIf(BigQueryUtil::isRetryableUnavailableError)
```
findepi: I liked single `BigQueryUtil::isRetryable` more. why differentiate between them?

vlad-lyutenko: We observe this problem only during read session creation, but this mechanism of retries is also used during read rows.
Do you think it's safe to also apply it there?

findepi: i'd reverse the question -- why would it be safe to retry in one place only?

vlad-lyutenko: My logic looks like this - create session kind of operation which could either succeed or not, so it's kind of safe to retry.
I prefer not to retry during reading rows, because we can skip/miss some rows during retry and get inconsistence data.
Of course it's only my guess, because we didn't encounter this situation.

findepi: that depends how that read rows request is defined.

vlad-lyutenko: Ok looks like it's safe to retry, because we keep offset:
 ```
catch (Exception e) {
                    // if relevant, retry the read, from the last read position
                    if (BigQueryUtil.isRetryable(e) && retries < helper.maxReadRowsRetries) {
                        log.debug("Request failed, retrying: %s", e);
                        serverResponses = helper.fetchResponses(nextOffset);
                        retries++;
                    }
                    else {
                        helper.client.close();
                        throw e;
                    }
                }
```
Will refactor to single BigQueryUtil::isRetryable

plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/BigQueryUtil.java
vlad-lyutenko: refactored, not relying on message

plugin/trino-bigquery/src/main/java/io/trino/plugin/bigquery/ReadSessionCreator.java
ebyhr: That is safe for now, but it will be unsafe if this method is used from different places such as write operations.

PR #27517, opened 2025-12-01
Improve handling of negative GRACE PERIOD

piotrrzysko:
## Description

Previously, providing a negative grace period resulted in a low-level internal exception:

```
Caused by: java.lang.IllegalArgumentException: gracePeriod cannot be negative: Optional[PT-1H]
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:217)
	at io.trino.metadata.MaterializedViewDefinition.<init>(MaterializedViewDefinition.java:49)
```

## Release notes

```markdown
## Section
```


core/trino-parser/src/main/java/io/trino/sql/parser/AstBuilder.java
```diff
@@ -610,7 +610,11 @@ public Node visitCreateMaterializedView(SqlBaseParser.CreateMaterializedViewCont
     {
         Optional<IntervalLiteral> gracePeriod = Optional.empty();
         if (context.GRACE() != null) {
-            gracePeriod = Optional.of((IntervalLiteral) visit(context.interval()));
+            IntervalLiteral interval = (IntervalLiteral) visit(context.interval());
+            if (interval.getSign() == IntervalLiteral.Sign.NEGATIVE) {
```
martint: This should be done in the semantic analyzer, not in the AST builder.

piotrrzysko: Thanks, fixed. I think I understand the reason. Just out of curiosity: if we had an _unsigned_ interval type and used it for GRACE PERIOD, would that be a syntax error then?

martint: Correct. But that would be handled in the grammar. Syntax checks in the AST builder should be limited to extreme cases that cannot handled easily via antlr4 rules.

ebyhr (approved):

PR #27519, opened 2025-12-01
Verify block types while generating JSON

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
+++ b/core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
@@ -119,6 +119,15 @@ public class OutputSpoolingOperatorFactory
             {
                 return queryDataEncoder.encoding();
             }
+
+            @Override
+            public String toString()
+            {
+                return toStringHelper(this)
+                        .add("queryDataEncoder", queryDataEncoder)
+                        .add("closed", closed)
+                        .toString();
+            }
         };
         return new OutputSpoolingOperator(operatorContext, trackingQueryDataEncoder, spoolingManager);
     }
--- a/core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
+++ b/core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
@@ -14,14 +14,25 @@
 package io.trino.server.protocol;
 
 import com.fasterxml.jackson.core.JsonGenerator;
+import com.google.common.collect.ImmutableList;
 import io.airlift.slice.Slice;
 import io.trino.Session;
 import io.trino.client.ClientCapabilities;
 import io.trino.spi.Page;
 import io.trino.spi.TrinoException;
+import io.trino.spi.block.ArrayBlock;
 import io.trino.spi.block.Block;
+import io.trino.spi.block.ByteArrayBlock;
+import io.trino.spi.block.DictionaryBlock;
+import io.trino.spi.block.IntArrayBlock;
+import io.trino.spi.block.LongArrayBlock;
+import io.trino.spi.block.MapBlock;
+import io.trino.spi.block.RowBlock;
+import io.trino.spi.block.RunLengthEncodedBlock;
+import io.trino.spi.block.ShortArrayBlock;
 import io.trino.spi.block.SqlMap;
 import io.trino.spi.block.SqlRow;
+import io.trino.spi.block.VariableWidthBlock;
 import io.trino.spi.type.ArrayType;
 import io.trino.spi.type.BigintType;
 import io.trino.spi.type.BooleanType;
@@ -48,8 +59,10 @@ import io.trino.type.SqlIntervalYearMonth;
 
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.util.Arrays;
 import java.util.List;
 import java.util.function.Consumer;
+import java.util.stream.IntStream;
 
 import static com.google.common.base.Verify.verify;
 import static io.trino.spi.StandardErrorCode.SERIALIZATION_ERROR;
@@ -64,6 +77,7 @@ import static io.trino.spi.type.TinyintType.TINYINT;
 import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static io.trino.spi.type.VarcharType.VARCHAR;
 import static java.util.Objects.requireNonNull;
+import static java.util.stream.Collectors.joining;
 
 public final class JsonEncodingUtils
 {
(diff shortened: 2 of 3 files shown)


core/trino-main/src/main/java/io/trino/operator/OutputSpoolingOperatorFactory.java
```diff
@@ -119,6 +119,15 @@ public String encoding()
             {
                 return queryDataEncoder.encoding();
             }
+
+            @Override
+            public String toString()
+            {
+                return toStringHelper(this)
+                        .add("delegate", queryDataEncoder)
```
findepi:
```suggestion
                        .add("queryDataEncoder", queryDataEncoder)
```

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -143,10 +163,55 @@ public static void writePagesToJsonGenerator(Consumer<TrinoException> throwableC
...
     }
 
+    private static Block unwrapBlock(Block block)
+    {
+        return switch (block) {
+            case DictionaryBlock dictionaryBlock -> dictionaryBlock.getDictionary();
+            case RunLengthEncodedBlock runLengthEncodedBlock -> runLengthEncodedBlock.getValue();
+            default -> block;
+        };
+    }
```
findepi: `block.getUnderlyingValueBlock()`

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -407,6 +616,33 @@ public void encode(JsonGenerator generator, Block block, int position)
...
+            return true;
+        }
+
+        @Override
+        public String toString()
+        {
+            ImmutableMap.Builder<String, String> fields = ImmutableMap.builderWithExpectedSize(fieldEncoders.length);
+            for (int i = 0; i < fieldEncoders.length; i++) {
+                fields.put(rowType.getFields().get(i).getName().orElse("col" + i), fieldEncoders[i].toString());
+            }
```
findepi: You can avoid materializing the hash map

```
IntStream.range(0, fieldEncoders.length)
        .mapToObj(i -> "%s: %s".formatted(rowType.getFields().get(i).getName().orElse("col" + i), fieldEncoders[i]))
        .collect(joining(", "))
```

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -143,10 +160,51 @@ public static void writePagesToJsonGenerator(Consumer<TrinoException> throwableC
         }
     }
 
+    private static String describeBlock(Block block)
```
raunaqmorarka, line 163: Can we improve the Block#toString implementations to print this instead ?
It would be beneficial for debugging in general

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -192,6 +274,18 @@ public void encode(JsonGenerator generator, Block block, int position)
             }
             generator.writeBoolean(BOOLEAN.getBoolean(block, position));
         }
+
+        @Override
+        public boolean isSupported(Block block)
+        {
+            return block instanceof ByteArrayBlock;
```
findepi, line 281:
```suggestion
            return block.getUnderlyingValueBlock() instanceof ByteArrayBlock;
```

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -143,10 +160,51 @@ public static void writePagesToJsonGenerator(Consumer<TrinoException> throwableC
         }
     }
 
+    private static String describeBlock(Block block)
+    {
+        return switch (block) {
+            case DictionaryBlock dictionaryBlock -> "DictionaryBlock{%s}".formatted(describeBlock(dictionaryBlock.getDictionary()));
+            case RunLengthEncodedBlock runLengthEncodedBlock -> "RleBlock{%s}".formatted(runLengthEncodedBlock.getValue());
```
findepi, line 167: Handle via `block.getUnderlyingValueBlock()`

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -143,10 +160,51 @@ public static void writePagesToJsonGenerator(Consumer<TrinoException> throwableC
...
+            case ArrayBlock arrayBlock -> "ArrayBlock{%s}".formatted(describeBlock(arrayBlock.getElementsBlock()));
+            case MapBlock mapBlock -> "MapBlock{keys=%s, values=%s}".formatted(describeBlock(mapBlock.getKeyBlock()), describeBlock(mapBlock.getValueBlock()));
+            case RowBlock rowBlock -> "RowBlock{fields=%s}".formatted(rowBlock.getFieldBlocks().stream().map(JsonEncodingUtils::describeBlock).collect(joining(", ")));
+            default -> block.toString();
```
findepi, line 171: This code should not be here for ever.
Let's file an issue describing problem we're trying to identify and add a TODO comment + link here to eventually remove this boiler-plate.
Also, it would probably be enough to handle primitive types only. Code would be much simpler, while debuggability would be near exact the same.

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
findepi: at least according to actual requirements in `encode`

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
losipiuk, line 159: This PR is quite a bit complex for a simple debugging log.
Did you consider maybe keeping track of deserialization state (change it as loop progresses) and add it to exception message here. This way you can easily get information like channel name or data types without all those code.

wendigo: I think that just a channel + type vs block mismatch is not enough

losipiuk: What else you get here?

findepi: we're getting more here indeed, but that's also easy to add with `block.toString` / `Array.toString(blocks)`

losipiuk: Yeah - I mean you can print whatever you want in the exception handler if you track that while loop iterates. You just need to keep pointer what block/channel you were processing when exception happened

wendigo: @losipiuk The whole point of this change is to verify blocks before they are serialized. Not to catch cast exception during serialization and then act on it.

wendigo:
Replaced by more generic approach: https://github.com/trinodb/trino/pull/27554

PR #27523, opened 2025-12-01
Use Stream#gather API instead of Guava code

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
+++ b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
@@ -57,8 +57,8 @@ import static com.google.cloud.storage.Storage.BlobListOption.pageSize;
 import static com.google.cloud.storage.Storage.SignUrlOption.withExtHeaders;
 import static com.google.cloud.storage.Storage.SignUrlOption.withV4Signature;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
-import static com.google.common.collect.Iterables.partition;
 import static io.airlift.concurrent.MoreFutures.getFutureValue;
 import static io.trino.filesystem.gcs.GcsUtils.encodedKey;
 import static io.trino.filesystem.gcs.GcsUtils.getBlob;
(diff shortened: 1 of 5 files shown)


lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
```diff
@@ -164,16 +165,19 @@ public void deleteFile(Location location)
...
             throws IOException
     {
-        List<ListenableFuture<?>> batchFutures = new ArrayList<>();
+        List<ListenableFuture<?>> batchFutures = locations.stream()
+                .gather(windowFixed(batchSize))
+                .map(locationBatch -> {
+                    StorageBatch batch = storage.batch();
+                    for (Location location : locationBatch) {
+                        GcsLocation gcsLocation = new GcsLocation(location);
+                        batch.delete(BlobId.of(gcsLocation.bucket(), gcsLocation.path()));
```
findepi, line 174: Why did it flow outside of the try catch block?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
```diff
@@ -186,16 +190,19 @@ public void deleteDirectory(Location location)
...
+                .map(blobBatch -> {
+                    StorageBatch batch = storage.batch();
+                    for (Blob blob : blobBatch) {
+                        batch.delete(blob.getBlobId());
+                    }
+                    return executorService.submit(batch::submit);
+                })
+                .collect(toImmutableList());
+
         try {
```
findepi, line 189: same here -- code moved outside of try

lib/trino-hdfs/src/main/java/io/trino/hdfs/s3/TrinoS3FileSystem.java
```diff
@@ -775,10 +775,9 @@ public void deleteFiles(Collection<Path> paths)
             throws IOException
     {
         try {
-            Iterable<List<Path>> partitions = Iterables.partition(paths, DELETE_BATCH_SIZE);
-            for (List<Path> currentBatch : partitions) {
-                deletePaths(currentBatch);
-            }
+            paths.stream()
+                    .gather(windowFixed(DELETE_BATCH_SIZE))
```
findepi, line 779: windowFixed  involves unnecessary data copying which `{Iterables,List}.partition` does not, so is inferior on performance aspect

it's also not superior on readability aspect. `{Iterables,List}.partition` very concisely describes what it does 

we should use `.gather(windowFixed` when we have a stream, but i don't see why we would want to use it on lists.

PR #27529, opened 2025-12-02
Add GRACE PERIOD to SHOW CREATE MATERIALIZED VIEW

piotrrzysko:
## Description

This PR adds support for displaying the `GRACE PERIOD` clause in the output of `SHOW CREATE MATERIALIZED VIEW`.

## Release notes

```markdown
## Section
* Add `GRACE PERIOD` to `SHOW CREATE MATERIALIZED VIEW` output. ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
+++ b/core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
@@ -562,7 +562,7 @@ public final class ShowQueriesRewrite
                     query,
                     false,
                     false,
-                    viewDefinition.flatMap(MaterializedViewDefinition::getGracePeriod)
+                    viewDefinition.get().getGracePeriod()
                             .map(DateTimeUtils::formatDayTimeInterval),
                     Optional.empty(), // TODO support WHEN STALE
                     propertyNodes,


core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -254,6 +263,69 @@ public static long parseDayTimeInterval(String value, IntervalField startField,
         throw invalidQualifier(startField, endField.orElse(startField));
     }
 
+    public static IntervalLiteral formatDayTimeInterval(Duration duration)
+    {
+        long millis = duration.toMillis();
+        Period period = new Period(Math.abs(millis)).normalizedStandard(PeriodType.dayTime());
```
findepi: add
```java
IntervalLiteral.Sign sign = millis < 0 ? NEGATIVE : POSITIVE;
```

so that abs is not scary

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -337,7 +409,8 @@ private static PeriodFormatter createPeriodFormatter(IntervalField startField, I
 
         List<PeriodParser> parsers = new ArrayList<>();
 
-        PeriodFormatterBuilder builder = new PeriodFormatterBuilder();
+        PeriodFormatterBuilder builder = new PeriodFormatterBuilder()
+                .printZeroIfSupported();
```
findepi, line 365: why so? code comment

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -372,6 +445,7 @@ private static PeriodFormatter createPeriodFormatter(IntervalField startField, I
                     break;
                 }
                 builder.appendLiteral(":");
+                builder.minimumPrintedDigits(2);
```
findepi, line 405: why so? code comment

core/trino-main/src/test/java/io/trino/util/TestDateTimeUtils.java
```diff
@@ -85,4 +97,81 @@ public void testParseIfIso8601DateFormat()
                 .isInstanceOf(DateTimeException.class)
                 .hasMessage("Invalid value for DayOfMonth (valid values 1 - 28/31): 41");
     }
+
+    @Test
+    void testDayTimeIntervalRoundTrip()
+    {
+        testDayTimeIntervalRoundTrip("0", SECOND, Optional.empty());
```
findepi: i'd rather have this take SQL interval string as input to test roundtrip.
That would be testing at the level at which things must be stable, and without adding too many layers IMO.

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -254,6 +263,69 @@ public static long parseDayTimeInterval(String value, IntervalField startField,
...
+            else {
+                endField = Optional.empty();
+                value = INTERVAL_MINUTE_FORMATTER.print(period);
+            }
+        }
+        else {
+            startField = SECOND;
+            endField = Optional.empty();
+            value = INTERVAL_SECOND_FORMATTER.print(period);
+        }
```
findepi: This looks terribly verbose, but i don't know how to improve it. ok to keep

piotrrzysko: One option is to reuse the function we use for casts: https://github.com/trinodb/trino/blob/518b244ba529bfa2d83f10759c467087f33ebac7/client/trino-client/src/main/java/io/trino/client/IntervalDayTime.java#L52-L73
but then the output is verbose: 
https://github.com/trinodb/trino/blob/518b244ba529bfa2d83f10759c467087f33ebac7/core/trino-main/src/test/java/io/trino/type/TestIntervalDayTime.java#L431-L434

martint: Keep in mind this is coming: https://github.com/trinodb/trino/pull/27539

findepi: > Keep in mind this is coming: #27539

Thanks for the link!

i skimmed the PR and it looks mostly backwards compatible. 
The code in this PR takes java time interval and formats it as a SQL literal, so it should work both before and after the same.

> One option is to reuse the function we use for casts

Sounds much simpler. How would that look like @piotrrzysko ?
Would we still output CREATE MV statement that can be pasted back into CLI and run?

piotrrzysko: A prototype: https://github.com/piotrrzysko/trino/commit/962fe5002bd3ba9e5029f4649b16a158cb9ac231

I added tests that demonstrate the formatted result can be parsed back into the original input (expressed in milliseconds, since the original format is not retained).

The only downside I see in this approach, as mentioned earlier, is that the formatted values are not succinct (e.g., `INTERVAL '1' HOUR` becomes `INTERVAL '0 01:00:00.000' DAY TO SECOND`).

findepi: We can't preserve original syntax (1 hour = 60 minutes and we don't distinguish)
I agree that `INTERVAL '1' HOUR` looks nicer  than `INTERVAL '0 01:00:00.000' DAY TO SECOND`.
But less code looks nicer to me.

What I am worried is that "reuse the function we use for casts" may be "opportunistic reuse" -- reuse of something that does what we need today, but has different design constrains and so can change in the future.
Unless/until we have confirmation "the function we use for casts" is guaranteed to produce a string literal that we can put directly into a SQL, let's not use it.

findepi:
please rebase

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
piotrrzysko: I agree that succinctness of the output is not critical in this case. Please take a look at the latest fixup commit. Since we cannot guarantee that the interval fields are preserved either way, it seems simplest to always use the widest range.

martint: One thing worth pointing out is that INTERVAL '1' HOUR and INTERVAL '...' DAY TO SECOND are actually different types in standard SQL. We don't distinguish between them *yet*, but it's something I'm trying to fix.

piotrrzysko: So this means that for GRACE PERIOD we will store the original interval fields and precision, right? If so, this is even better, because then we will explicitly know which formatter should be used.

martint: Eventually, yes.

findepi: > One thing worth pointing out is that INTERVAL '1' HOUR and INTERVAL '...' DAY TO SECOND are actually different types in standard SQL.

great


> So this means that for GRACE PERIOD we will store the original interval fields and precision, right?

i am against that.  in our simplified world,  1 day = 24 hours = 24 * 60 minutes = 24 * 60 * 60 seconds
furthermore, for MV functionality, it doesn't at all matter _how_ the grace period time duration was specified, so i don't see why we would want to go extra complexity to preserve that

anyhow, all this certainly does not matter from the perspective of _this PR_. it's undoubtedly a bug fix and way forward. let's merge it

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -254,6 +261,20 @@ public static long parseDayTimeInterval(String value, IntervalField startField,
         throw invalidQualifier(startField, endField.orElse(startField));
     }
 
+    public static IntervalLiteral formatDayTimeInterval(Duration duration)
+    {
+        long millis = duration.toMillis();
+        Period period = new Period(Math.abs(millis)).normalizedStandard(PeriodType.dayTime());
```
findepi: Do we still need `Period` class and `normalizedStandard` call?
Can the formatter operate on `long millis` value directly?

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -254,6 +261,20 @@ public static long parseDayTimeInterval(String value, IntervalField startField,
...
+
+        IntervalLiteral.Sign sign = millis < 0 ? NEGATIVE : POSITIVE;
```
findepi, line 267: please move `sign =` just before the `abs` line

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -561,7 +562,8 @@ private Query showCreateMaterializedView(ShowCreate node)
                     query,
                     false,
                     false,
-                    Optional.empty(), // TODO support GRACE PERIOD
+                    viewDefinition.flatMap(MaterializedViewDefinition::getGracePeriod)
```
findepi:
```suggestion
                    viewDefinition.get().getGracePeriod()
```

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
piotrrzysko: 1. The `PeriodFormatter.print()` method needs a `ReadablePeriod` object 
2. `normalizedStandard()` is needed to handle overflows between fields. Without it, for example, we would get `INTERVAL '0 25:01:01' DAY TO SECOND` instead of `INTERVAL '1 1:01:01' DAY TO SECOND`

chenjian2664:
I put RN under section "general", FYI

Issue #26147, opened 2025-07-08
ClassCastException when running sql with task retry policy


never-wang:
After upgrading Trino from version 436 to 476, one SQL in our daily offline tasks throws an error. 
Judging from the error stack, it seems to be related to the exchange manager, and we speculate that it has something to do with the task retry policy. 
We conducted a test: when we changed the retry-policy to QUERY, the SQL could run normally. 

The error stack is :
```
java.lang.ClassCastException: Cannot cast io.trino.spi.block.IntArrayBlock to io.trino.spi.block.VariableWidthBlock
	at java.base/java.lang.Class.cast(Class.java:3523)
	at io.trino.operator.InterpretedHashGenerator.nullSafeHash(InterpretedHashGenerator.java:90)
	at io.trino.operator.InterpretedHashGenerator.hashPosition(InterpretedHashGenerator.java:82)
	at io.trino.operator.exchange.LocalPartitionGenerator.getRawHash(LocalPartitionGenerator.java:53)
	at io.trino.operator.exchange.LocalPartitionGenerator.getPartition(LocalPartitionGenerator.java:47)
	at io.trino.operator.exchange.PartitioningExchanger.accept(PartitioningExchanger.java:62)
	at io.trino.operator.exchange.LocalExchangeSink.addPage(LocalExchangeSink.java:74)
	at io.trino.operator.exchange.LocalExchangeSinkOperator.addInput(LocalExchangeSinkOperator.java:137)
	at io.trino.operator.Driver.processInternal(Driver.java:408)
	at io.trino.operator.Driver.lambda$process$0(Driver.java:306)
	at io.trino.operator.Driver.tryWithLock(Driver.java:709)
	at io.trino.operator.Driver.process(Driver.java:298)
	at io.trino.operator.Driver.processForDuration(Driver.java:269)
	at io.trino.execution.SqlTaskExecution$DriverSplitRunner.processFor(SqlTaskExecution.java:889)
	at io.trino.execution.executor.dedicated.SplitProcessor.run(SplitProcessor.java:77)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.lambda$run$0(TaskEntry.java:201)
	at io.trino.$gen.Trino_476____20250708_001412_2.run(Unknown Source)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.run(TaskEntry.java:202)
[...]

PR #27538, opened 2025-12-02
Verify operator block types

losipiuk:
## Description

We observed occurrences where Trino queries failed at runtime with `ClassCastException` as pages produced by one of the `Operators did not match the contract driven by query plan, channels and their types.
Example is logged in this issue: https://github.com/trinodb/trino/issues/26147.

This PR adds diagnostics mechanism to verify for each `Page` produced by `Operator.getOutput` that blocks this page consists of match the types assigned to channels produced by this operator.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/Driver.java
+++ b/core/trino-main/src/main/java/io/trino/operator/Driver.java
@@ -14,6 +14,7 @@
 package io.trino.operator;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Joiner;
 import com.google.common.base.VerifyException;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
@@ -30,6 +31,9 @@ import io.trino.execution.SplitAssignment;
 import io.trino.metadata.Split;
 import io.trino.spi.Page;
 import io.trino.spi.TrinoException;
+import io.trino.spi.block.Block;
+import io.trino.spi.block.RowBlock;
+import io.trino.spi.block.ValueBlock;
 import io.trino.spi.type.Type;
 
 import java.io.Closeable;
@@ -53,6 +57,7 @@ import static com.google.common.util.concurrent.Futures.nonCancellationPropagati
 import static com.google.common.util.concurrent.Futures.withTimeout;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static io.airlift.concurrent.MoreFutures.getFutureValue;
+import static io.trino.SystemSessionProperties.isOperatorOutputValidationEnabled;
 import static io.trino.operator.Operator.NOT_BLOCKED;
 import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static java.lang.Boolean.TRUE;
@@ -72,6 +77,8 @@ public class Driver
     private static final Logger log = Logger.get(Driver.class);
 
     private static final Duration UNLIMITED_DURATION = new Duration(Long.MAX_VALUE, NANOSECONDS);
+
+    private final boolean operatorOutputValidationEnabled; // todo config
 
     private final DriverContext driverContext;
     private final List<TypedOperator> activeOperators;
@@ -160,6 +167,7 @@ public class Driver
         SettableFuture<Void> future = SettableFuture.create();
         future.set(null);
         driverBlockedFuture.set(future);
+        operatorOutputValidationEnabled = isOperatorOutputValidationEnabled(driverContext.getSession());
     }
 
     private static final class TypedOperator
@@ -448,7 +456,9 @@ public class Driver
 
         boolean movedPage = false;
         for (int i = 0; i < activeOperators.size() - 1 && !driverContext.isTerminatingOrDone(); i++) {
-            Operator current = activeOperators.get(i).operator();
+            TypedOperator currentTyped = activeOperators.get(i);
+            Operator current = currentTyped.operator();
+            Optional<List<Type>> currentOutputTypes = currentTyped.types();
             Operator next = activeOperators.get(i + 1).operator();
 
             // skip blocked operator
@@ -464,6 +474,10 @@ public class Driver
 
                 // if we got an output page, add it to the next operator
                 if (page != null && page.getPositionCount() != 0) {
+                    if (operatorOutputValidationEnabled && currentOutputTypes.isPresent()) {
+                        validateOperatorOutputPage(page, current, currentOutputTypes.get());
+                    }
+
                     next.addInput(page);
                     next.getOperatorContext().recordAddInput(operationTimer, page);
                     movedPage = true;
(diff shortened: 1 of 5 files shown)


TypedOperatorFactory.java
```diff
@@ -0,0 +1,22 @@
+package io.trino.sql.planner;
+
+import io.trino.operator.OperatorFactory;
+import io.trino.spi.type.Type;
+
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+
+public record TypedOperatorFactory(OperatorFactory operatorFactory, List<Type> types)
```
findepi: This copy of `TypedOperatorFactory` is outstide of source directories. Please remove.

core/trino-main/src/main/java/io/trino/operator/join/LookupJoinOperatorFactory.java
```diff
@@ -99,12 +99,20 @@ public LookupJoinOperatorFactory(
...
-            this.outerOperatorFactory = Optional.of(new LookupOuterOperatorFactory(
-                    operatorId,
-                    planNodeId,
-                    probeOutputTypes,
-                    buildOutputTypes,
-                    lookupSourceFactoryManager));
+            List<Type> types = ImmutableList.<Type>builder()
+                    .addAll(probeOutputTypes)
+                    .addAll(buildOutputTypes)
+                    .build();
```
findepi: We're duplicating logic from `LookupOuterOperatorFactory`.
(did you mean to use `LookupOuterOperatorFactory.getTypes()` also added in this commit?)

Maybe, instead, we could have an optional method on OperatorFactory 
```
Optional<List<Type>> returnedTypes();
```

then, also maybe, we wouldn't need the `TypedOperatorFactory`.

core/trino-main/src/main/java/io/trino/operator/join/unspilled/LookupJoinOperatorFactory.java
```diff
@@ -80,12 +80,20 @@ public LookupJoinOperatorFactory(
...
-            this.outerOperatorFactory = Optional.of(new LookupOuterOperatorFactory(
-                    operatorId,
-                    planNodeId,
-                    probeOutputTypes,
-                    buildOutputTypes,
-                    lookupSourceFactoryManager));
+            List<Type> types = ImmutableList.<Type>builder()
+                    .addAll(probeOutputTypes)
+                    .addAll(buildOutputTypes)
+                    .build();
```
findepi: as above

core/trino-main/src/main/java/io/trino/sql/planner/TypedOperatorFactory.java
```diff
@@ -0,0 +1,22 @@
+package io.trino.sql.planner;
+
+import io.trino.operator.OperatorFactory;
+import io.trino.spi.type.Type;
+
+import java.util.List;
+
+import static java.util.Objects.requireNonNull;
+
+public record TypedOperatorFactory(OperatorFactory operatorFactory, List<Type> types)
```
findepi: `record` is generally suitable for value classes. This one is not a value class to me. I see how `record` makes the impl shorter though. Should i change my perspective?

also, mind `instanceof.*OperatorFactory` & `LocalPlannerAware.class::isInstance` in the code.

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -118,15 +121,30 @@ public static Driver createDriver(DriverContext driverContext, Operator firstOpe
...
+        driver.initialize();
+        return driver;
     }
 
-    private Driver(DriverContext driverContext, List<Operator> operators)
+    private Driver(DriverContext driverContext, List<Operator> operators, Optional<List<List<Type>>> operatorOutputTypes)
     {
+        requireNonNull(operators, "operators is null");
+        requireNonNull(operatorOutputTypes, "operatorOutputTypes is null");
+        List<TypedOperator> typedOperators;
```
findepi: unused

core/trino-main/src/main/java/io/trino/sql/planner/TypedOperatorFactory.java
```diff
@@ -0,0 +1,22 @@
...
+{
+    public TypedOperatorFactory
+    {
+        requireNonNull(operatorFactory, "operatorFactory is null");
+        requireNonNull(types, "types is null");
```
findepi: defensive copy to guarantee immutability

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -145,12 +163,32 @@ private Driver(DriverContext driverContext, List<Operator> operators)
...
+
+        public TypedOperator(Operator operator, List<Type> types)
+        {
+            this(operator, Optional.of(types));
+        }
+
+        TypedOperator
+        {
+            requireNonNull(operator, "operator is null");
+            requireNonNull(types, "types is null");
```
findepi: defensive copy to guarantee immutability

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -99,17 +102,17 @@ private enum State
         ALIVE, NEED_DESTRUCTION, DESTROYING, DESTROYED
     }
 
-    public static Driver createDriver(DriverContext driverContext, List<Operator> operators)
+    public static Driver createDriver(DriverContext driverContext, List<Operator> operators, List<List<Type>> operatorTypes)
```
findepi, line 112: Wouldn't it be better to ask the caller of that method to marry operators and types, rather than have this as two lists?

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -118,15 +121,30 @@ public static Driver createDriver(DriverContext driverContext, Operator firstOpe
...
     {
+        requireNonNull(operators, "operators is null");
+        requireNonNull(operatorOutputTypes, "operatorOutputTypes is null");
+        List<TypedOperator> typedOperators;
+        if (operatorOutputTypes.isPresent()) {
+            checkArgument(operators.size() == operatorOutputTypes.get().size(), "operators and operatorOutputTypes size mismatch");
+            this.allOperators = Streams.zip(
+                    operators.stream(),
+                    operatorOutputTypes.get().stream(),
+                    (operator, types) -> new TypedOperator(operator, ImmutableList.copyOf(types))).collect(toImmutableList());
```
findepi: remove `ImmutableList.copyOf` from here. it's TypedOperator ctor responsibility

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -73,6 +75,8 @@ public class Driver
 
     private static final Duration UNLIMITED_DURATION = new Duration(Long.MAX_VALUE, NANOSECONDS);
 
+    private final boolean outputTypesCheckingEnabled = true; // todo config
```
findepi: \+ session property. Doing via session is usually easier.

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -524,6 +534,42 @@ private ListenableFuture<Void> processInternal(OperationTimer operationTimer)
         return NOT_BLOCKED;
     }
 
+    private void validateOperatorOuputPage(Page page, Operator operator, List<Type> types)
+    {
+        if (page.getChannelCount() != types.size()) {
+            throw new TrinoException(
+                    GENERIC_INTERNAL_ERROR,
+                    "Invalid number of channels; got %s expected %s; %s".formatted(page.getChannelCount(), types.size(), operatorDebugInfo(operator)));
```
findepi: Worth printing all expected types and all actual block types just as if there was a type mistmatch

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -524,6 +534,42 @@ private ListenableFuture<Void> processInternal(OperationTimer operationTimer)
...
+                mismatches.add("Bad block %s for channel %s of type %s".formatted(block, channel, type));
+            }
+        }
+        if (mismatches != null) {
+            throw new TrinoException(GENERIC_INTERNAL_ERROR,
+                    "Bad block types found for operator %s; %s".formatted(operatorDebugInfo(operator), mismatches));
+        }
+    }
+
+    private boolean isTypeValidForBlock(Block block, Type type)
```
findepi: `isBlockValidForType(type, block)` -- the type is what's expected and required, the block is actual.
also it's up to type to decide which bock type(s) is accepts, not the other way around

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -524,6 +534,42 @@ private ListenableFuture<Void> processInternal(OperationTimer operationTimer)
...
+    {
+        ValueBlock underlyingValueBlock = block.getUnderlyingValueBlock();
+        return type.getValueBlockType().isInstance(underlyingValueBlock);
```
findepi, line 631: neat

core/trino-main/src/main/java/io/trino/operator/Driver.java
```diff
@@ -524,6 +534,42 @@ private ListenableFuture<Void> processInternal(OperationTimer operationTimer)
...
+            if (!isTypeValidForBlock(block, type)) {
+                if (mismatches == null) {
+                    mismatches = new ArrayList<>();
+                }
+                mismatches.add("Bad block %s for channel %s of type %s".formatted(block, channel, type));
+            }
+        }
+        if (mismatches != null) {
+            throw new TrinoException(GENERIC_INTERNAL_ERROR,
+                    "Bad block types found for operator %s; %s".formatted(operatorDebugInfo(operator), mismatches));
```
findepi: Printing both matching and non-matching could help understanding what's going on.
```java
// mismatches is Map<Integer channel, String>
if (mismatches != null) {
    StringBuilder error = new StringBuilder("Bad block types found for operator " + operatorDebugInfo(operator));
    for (int channel = 0; channel < types.size(); channel++) {
        String line = mismatches.get(channel);
        if (line == null) {
            line = "Good block for channel %s of type %s".formatted(channel, types.get(channel));
        }
        error.append("\n\t\t").append(line);
    }
    throw new TrinoException(GENERIC_INTERNAL_ERROR, error.toString());
}
```

findepi:
please fill PR description

I think the change would be way simpler if we limited the check to connectors only
https://github.com/trinodb/trino/blob/9873ef3d260b3e3fde406878bb5f95eeb7707579/core/trino-main/src/main/java/io/trino/split/PageSourceManager.java#L62-L86

core/trino-main/src/main/java/io/trino/operator/join/LookupJoinOperatorFactory.java
losipiuk: Modifying `OperatorFactory` interface is problematic that we would either need to implement the new method in each of the classes implementing interface - for sure we do not want that - there are many of those.
Or create some wrapper implementation which adds the `Optional<List<Type>> returnedTypes();` - this would work but I think it is more error prone than using `TypedOperatorFactory` as you are not getting help from compiler and you can end up using non-wrapped OperatorFactory in place where you want it wrapped.

Also I am not sure if we sometimes do not inspect OperatorFactory class

core/trino-main/src/main/java/io/trino/sql/planner/TypedOperatorFactory.java
losipiuk: > Should i change my perspective?

I do not have strong opinion here. We are using often records like that, but I can change it to a class.

core/trino-main/src/main/java/io/trino/operator/Driver.java
losipiuk: I went this path to limit changes to external model. Currently TypedOperator is private class local to Driver. To marry those in external interface I need to make it public.
I do not have a strong opionion. Do you want me to do the change?

martint:
> I think the change would be way simpler if we limited the check to connectors only

+1 to this. This change is way too invasive. Verifying the outputs of every operator is overkill. We just need to ensure that connectors produce the right block types for the expected type -- that's where most of the mistakes can happen.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/avro/AvroPageSourceFactory.java
```diff
@@ -142,29 +141,6 @@ private static AvroPageSource createPageSource(
...
-            try {
-                return new AvroPageSource(inputFile, nullSchema.endRecord(), new HiveAvroTypeBlockHandler(createTimestampType(hiveTimestampPrecision)), start, length);
-            }
-            catch (IOException e) {
-                throw new TrinoException(HIVE_CANNOT_OPEN_SPLIT, e);
-            }
-            catch (AvroTypeException e) {
-                throw new TrinoException(HIVE_CANNOT_OPEN_SPLIT, "Avro type resolution error when initializing split from %s".formatted(inputFile.location()), e);
-            }
-        }
```
findepi, line 167: if it works... great!

cc @jklamer

findepi (commented):
Hive changes LGTM

losipiuk:
Alterantive: https://github.com/trinodb/trino/pull/27554

PR #27539, opened 2025-12-02
Arbitrary precision interval grammar

martint:
Add support grammar support for arbitrary precision interval qualifiers in preparation for implementing
arbitrary precision intervals, as described by the SQL specification.

It allows parsing intervals such as:

```
YEAR(4) 
YEAR(2) TO MONTH
DAY(5)
DAY(3) TO MINUTE
DAY(3) TO SECOND(6)
SECOND(5,2)
```

There's no functional difference from the existing implementation aside changes to parsing and semantic errors when the intervals are not compatible with what we currently support.

## Release notes

### DIFF
--- a/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
+++ b/core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
@@ -760,11 +760,7 @@ booleanValue
     ;
 
 interval
-    : INTERVAL sign=(PLUS | MINUS)? string from=intervalField (TO to=intervalField)?
-    ;
-
-intervalField
-    : YEAR | MONTH | DAY | HOUR | MINUTE | SECOND
+    : INTERVAL sign=(PLUS | MINUS)? string intervalQualifier
     ;
 
 normalForm
@@ -773,7 +769,7 @@ normalForm
 
 type
     : ROW '(' rowField (',' rowField)* ')'                                         #rowType
-    | INTERVAL from=intervalField (TO to=intervalField)?                           #intervalType
+    | INTERVAL intervalQualifier                                                   #intervalType
     | base=TIMESTAMP ('(' precision = typeParameter ')')? (WITHOUT TIME ZONE)?     #dateTimeType
     | base=TIMESTAMP ('(' precision = typeParameter ')')? WITH TIME ZONE           #dateTimeType
     | base=TIME ('(' precision = typeParameter ')')? (WITHOUT TIME ZONE)?          #dateTimeType
@@ -784,6 +780,18 @@ type
     | type ARRAY ('[' INTEGER_VALUE ']')?                                          #arrayType
     | identifier ('(' typeParameter (',' typeParameter)* ')')?                     #genericType
     ;
+
+intervalQualifier
+  : YEAR ('(' precision=INTEGER_VALUE ')')? TO MONTH                                              #compositeYearToMonthInterval
+  | field=(YEAR | MONTH) ('(' precision=INTEGER_VALUE ')')?                                       #simpleYearMonthInterval
+  | start=(DAY | HOUR | MINUTE) ('(' leadingPrecision=INTEGER_VALUE ')')?
+    TO (
+      end=HOUR |
+      end=MINUTE |
+      end=SECOND ('(' fractionalPrecision=INTEGER_VALUE ')')?)                                    #compositeDayTimeInterval
+  | field=(DAY | HOUR | MINUTE) ('(' precision=INTEGER_VALUE ')')?                                #simpleDayTimeInterval
+  | SECOND ('(' leadingPrecision=INTEGER_VALUE (',' fractionalPrecision=INTEGER_VALUE)? ')')?     #secondsDayTimeInterval
+  ;
 
 rowField
     : type
--- a/core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
+++ b/core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
@@ -18,6 +18,8 @@ import io.trino.client.IntervalDayTime;
 import io.trino.client.IntervalYearMonth;
 import io.trino.spi.TrinoException;
 import io.trino.spi.type.TimeZoneKey;
+import io.trino.sql.tree.CompositeIntervalQualifier;
+import io.trino.sql.tree.IntervalField;
 import io.trino.sql.tree.IntervalLiteral;
 import org.joda.time.DateTime;
 import org.joda.time.DurationFieldType;
@@ -48,9 +50,6 @@ import java.util.stream.Stream;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static io.trino.spi.StandardErrorCode.INVALID_LITERAL;
-import static io.trino.sql.tree.IntervalLiteral.IntervalField;
-import static io.trino.sql.tree.IntervalLiteral.IntervalField.DAY;
-import static io.trino.sql.tree.IntervalLiteral.IntervalField.SECOND;
 import static io.trino.sql.tree.IntervalLiteral.Sign.NEGATIVE;
 import static io.trino.sql.tree.IntervalLiteral.Sign.POSITIVE;
 import static io.trino.util.DateTimeZoneIndex.getChronology;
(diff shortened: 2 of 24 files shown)


core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -3787,10 +3787,6 @@ public void testLiteral()
                 .hasErrorCode(INVALID_LITERAL);
         assertFails("SELECT INTERVAL '12.1' DAY TO SECOND")
                 .hasErrorCode(INVALID_LITERAL);
-        assertFails("SELECT INTERVAL '12' YEAR TO DAY")
-                .hasErrorCode(INVALID_LITERAL);
-        assertFails("SELECT INTERVAL '12' SECOND TO MINUTE")
-                .hasErrorCode(INVALID_LITERAL);
```
findepi, line 3797: why remove?

core/trino-parser/src/main/java/io/trino/sql/tree/CompositeIntervalQualifier.java
```diff
@@ -0,0 +1,104 @@
...
+    private final IntervalField to;
+
+    public CompositeIntervalQualifier(NodeLocation location, OptionalInt precision, IntervalField from, IntervalField to)
+    {
+        super(location);
+        checkArgument(!from.equals(to), "'From' field must be wider that 'to' field");
+
+        this.precision = precision;
+        this.from = from;
+        this.to = to;
```
findepi: rnn `precision` and `to` (`from` is implicitly verified)

core/trino-spi/src/main/java/io/trino/spi/StandardErrorCode.java
```diff
@@ -160,6 +160,7 @@ public enum StandardErrorCode
     INVALID_BRANCH_PROPERTY(136, USER_ERROR),
     INVALID_DEFAULT_COLUMN_VALUE(137, USER_ERROR),
     INVALID_GRACE_PERIOD(138, USER_ERROR),
+    INVALID_INTERVAL_QUALIFIER(139, USER_ERROR),
```
findepi: To me this could also be a SYNTAX_ERROR perhaps?

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
martint: They fail during parsing now.

core/trino-spi/src/main/java/io/trino/spi/StandardErrorCode.java
martint: Fair. Removed.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
findepi: Okay, but why not have a test for them _somewhere_?
I see `SELECT INTERVAL '12' SECOND TO MINUTE` in removed lines, but not in lines added elsewhere

martint: There are tests in TestIntervalDayTime, already:

```
assertThatThrownBy(assertions.expression("INTERVAL '12-10' SECOND TO MINUTE")::evaluate)
                .hasMessage("line 1:39: mismatched input 'MINUTE'. Expecting: ')', ','");
```

findepi: it's a different test case, that's why i didn't find it. but fine

findepi:
Can you please update PR description with some examples showing what's new or changing?

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -785,6 +781,18 @@ type
...
 
+intervalQualifier
+  : YEAR ('(' precision=INTEGER_VALUE ')')? TO MONTH                                              #compositeYearToMonthInterval
+  | field=(YEAR | MONTH) ('(' precision=INTEGER_VALUE ')')?                                       #simpleYearMonthInterval
+  | start=(DAY | HOUR | MINUTE) ('(' leadingPrecision=INTEGER_VALUE ')')?
+    TO (
+      end=HOUR |
+      end=MINUTE |
+      end=SECOND ('(' fractionalPrecision=INTEGER_VALUE ')')?)                                    #compositeDayTimeInterval
+  | field=(MONTH | DAY | HOUR | MINUTE) ('(' precision=INTEGER_VALUE ')')?                        #simpleDayTimeInterval
```
kasiafi: Should MONTH be here?

core/trino-parser/src/main/java/io/trino/sql/tree/CompositeIntervalQualifier.java
```diff
@@ -0,0 +1,105 @@
...
+        extends IntervalQualifier
+{
+    private final OptionalInt precision;
+    private final IntervalField from;
+    private final IntervalField to;
+
+    public CompositeIntervalQualifier(NodeLocation location, OptionalInt precision, IntervalField from, IntervalField to)
+    {
+        super(location);
+        checkArgument(!from.equals(to), "'From' field must be wider that 'to' field");
```
kasiafi: Why handle this case here? It doesn't pass the `AstBuilder`, and is also handled in `ExpressionAnalyzer`.

core/trino-parser/src/test/java/io/trino/sql/parser/TestSqlParser.java
kasiafi, line 1: Since we can already parse intervals with precision, we should test them in `TestSqlParser`.
Specifically, we should have test cases to showcase the behavior of `SECOND` with precision in both cases:
`SECOND(3)` in composite qualifier --> `3` is the fractional precision,
`SECOND(3)` in simple qualifier --> `3` is the leading precision.

core/trino-parser/src/test/java/io/trino/sql/parser/TestTypeParser.java
kasiafi, line 1: Test qualifiers with precision.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
kasiafi, line 1: Assert failure for intervals with precision.

core/trino-parser/src/main/java/io/trino/sql/parser/AstBuilder.java
```diff
@@ -3555,15 +3556,89 @@ public Node visitTypeParameter(SqlBaseParser.TypeParameterContext context)
...
 
-        return new IntervalDayTimeDataType(
+    @Override
+    public Node visitSimpleYearMonthInterval(SqlBaseParser.SimpleYearMonthIntervalContext context)
+    {
+        return new SimpleIntervalQualifier(
+                getLocation(context),
+                context.precision != null ? OptionalInt.of(Integer.parseInt(context.precision.getText())) : OptionalInt.empty(),
+                switch (context.field.getType()) {
+                    case SqlBaseLexer.YEAR -> new IntervalField.Year();
```
kasiafi: Static import `YEAR` and other.

core/trino-main/src/main/java/io/trino/util/DateTimeUtils.java
```diff
@@ -271,7 +271,7 @@ public static IntervalLiteral formatDayTimeInterval(Duration duration)
         // determine the minimal field range and choose a specialized formatter.
         String value = INTERVAL_DAY_SECOND_FORMATTER.print(period);
 
-        return new IntervalLiteral(value, sign, DAY, Optional.of(SECOND));
+        return new IntervalLiteral(value, sign, new CompositeIntervalQualifier(new NodeLocation(1, 1), OptionalInt.empty(), new IntervalField.Day(), new IntervalField.Second(OptionalInt.empty())));
```
kasiafi: Wouldn't we rather have a constructor without `NodeLocation` than pass an arbitrary value?

core/trino-parser/src/test/java/io/trino/sql/TestExpressionFormatter.java
kasiafi, line 1: Test qualifiers with precision.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
martint: Good catch!

martint:
@kasiafi , updated.

kasiafi (approved):

PR #27540, opened 2025-12-02
Reduce log verbosity in tests

martint:
## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/testing/TestingTrinoServer.java
+++ b/core/trino-main/src/main/java/io/trino/server/testing/TestingTrinoServer.java
@@ -158,8 +158,12 @@ public class TestingTrinoServer
 {
     static {
         Logging logging = Logging.initialize();
+        logging.setLevel("io.trino.connector.CatalogStoreManager", Level.WARN);
+        logging.setLevel("io.trino.server.PluginManager", Level.WARN);
+        logging.setLevel("io.trino.memory.RemoteNodeMemory", Level.ERROR);
         logging.setLevel("io.trino.event.QueryMonitor", Level.ERROR);
         logging.setLevel("org.eclipse.jetty", Level.ERROR);
+        logging.setLevel("io.airlift.http.server", Level.WARN);
         logging.setLevel("io.airlift.concurrent.BoundedExecutor", Level.OFF);
 
         // Trino server behavior does not depend on locale settings.
--- a/core/trino-main/src/test/java/io/trino/execution/scheduler/faulttolerant/TestTaskDescriptorStorage.java
+++ b/core/trino-main/src/test/java/io/trino/execution/scheduler/faulttolerant/TestTaskDescriptorStorage.java
@@ -24,6 +24,8 @@ import io.airlift.bootstrap.Bootstrap;
 import io.airlift.jaxrs.JaxRsJsonMapper;
 import io.airlift.json.JsonCodec;
 import io.airlift.json.JsonModule;
+import io.airlift.log.Level;
+import io.airlift.log.Logging;
 import io.airlift.units.DataSize;
 import io.trino.connector.CatalogHandle;
 import io.trino.exchange.SpoolingExchangeInput;
@@ -53,6 +55,11 @@ import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 public class TestTaskDescriptorStorage
 {
+    static {
+        Logging logging = Logging.initialize();
+        logging.setLevel("org.hibernate.validator.internal.util.Version", Level.WARN);
+    }
+
     private static final QueryId QUERY_1 = new QueryId("query1");
     private static final QueryId QUERY_2 = new QueryId("query2");
 
@@ -363,7 +370,9 @@ public class TestTaskDescriptorStorage
                     jsonCodecBinder(binder).bindJsonCodec(Split.class);
                 });
 
-        Injector injector = app.initialize();
+        Injector injector = app
+                .doNotInitializeLogging()
+                .initialize();
         JsonCodec<TaskDescriptor> taskDescriptorJsonCodec = injector.getInstance(Key.get(new TypeLiteral<>() {}));
 
         TaskDescriptorStorage manager = new TaskDescriptorStorage(maxMemory, compressingHighWaterMark, compressingLowWaterMark, taskDescriptorJsonCodec);
--- a/core/trino-main/src/test/java/io/trino/server/protocol/spooling/TestPreferredQueryDataEncoderSelector.java
+++ b/core/trino-main/src/test/java/io/trino/server/protocol/spooling/TestPreferredQueryDataEncoderSelector.java
@@ -13,6 +13,8 @@
  */
 package io.trino.server.protocol.spooling;
 
+import io.airlift.log.Level;
+import io.airlift.log.Logging;
 import io.trino.server.protocol.spooling.QueryDataEncoder.EncoderSelector;
 import io.trino.server.protocol.spooling.encoding.JsonQueryDataEncoder;
 import org.junit.jupiter.api.Test;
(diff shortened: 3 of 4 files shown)


core/trino-main/src/main/java/io/trino/server/testing/TestingTrinoServer.java
```diff
@@ -158,8 +158,12 @@ public class TestingTrinoServer
 {
     static {
         Logging logging = Logging.initialize();
+        logging.setLevel("io.trino.connector.CatalogStoreManager", Level.ERROR);
+        logging.setLevel("io.trino.server.PluginManager", Level.ERROR);
+        logging.setLevel("io.trino.memory.RemoteNodeMemory", Level.ERROR);
```
findepi: In theory we could put ERROR as the default level, but that may make reproducing errors harder. Also, it opens a possibility for the code to work in tests, but not work under defaults settings (yes, i know no code semantics should depend on log levels being enabled).

So for every such special measure we should understand -- and document -- why we're doing that. Please add a line of comment for each of these perhaps exemplifying what's too verbose (at the moment of writing).

BTW sometimes we can fix the problem at source. For instance, airlift configuration logs all configured values, but afaict it's disabled in tests.

findepi: also, why not INFO -> WARN, when INFO is too verbose for tests?

core/trino-main/src/test/java/io/trino/execution/scheduler/faulttolerant/TestTaskDescriptorStorage.java
```diff
@@ -53,6 +55,12 @@
 
 public class TestTaskDescriptorStorage
 {
+    static {
+        Logging logging = Logging.initialize();
+        logging.setLevel("org.hibernate.validator.internal.util.Version", Level.ERROR);
+        logging.setLevel("Bootstrap", Level.ERROR);
```
findepi: What is Bootstrap?

core/trino-main/src/test/java/io/trino/server/protocol/spooling/TestPreferredQueryDataEncoderSelector.java
```diff
@@ -24,6 +26,11 @@
 
 class TestPreferredQueryDataEncoderSelector
 {
+    static {
+        Logging logging = Logging.initialize();
+        logging.setLevel("io.trino.server.protocol.spooling", Level.ERROR);
```
findepi: does this logger produce too much also in production?

cc @wendigo

plugin/trino-exchange-filesystem/src/test/java/io/trino/plugin/exchange/filesystem/s3/TestS3FileSystemExchangeManager.java
```diff
@@ -26,6 +28,11 @@
 public class TestS3FileSystemExchangeManager
         extends AbstractTestExchangeManager
 {
+    static {
+        Logging logging = Logging.initialize();
+        logging.setLevel("io.trino.bootstrap.exchange", Level.ERROR);
```
findepi:
```suggestion
        // Airlift configuration logs all the properties
        logging.setLevel("io.trino.bootstrap.exchange.filesystem", Level.WARN);
```

findepi: It would be much more robust to have `io.airlift.bootstrap.Bootstrap#quiet` initialize from env.

core/trino-main/src/test/java/io/trino/execution/scheduler/faulttolerant/TestTaskDescriptorStorage.java
wendigo: default logger used for unnamed `io.airlift.bootstrap.Bootstrap` instances

core/trino-main/src/test/java/io/trino/server/protocol/spooling/TestPreferredQueryDataEncoderSelector.java
wendigo: no. it just logs at startup (loading spooling manager...) similarily to loading exchange managers or plugins

plugin/trino-exchange-filesystem/src/test/java/io/trino/plugin/exchange/filesystem/s3/TestS3FileSystemExchangeManager.java
findepi: see
- https://github.com/airlift/airlift/pull/1667

wendigo (approved):

findepi:
Can this be simplified benefitting from
- https://github.com/airlift/airlift/pull/1667

?

findepi:
@martint ptal
- https://github.com/trinodb/trino/pull/27884

core/trino-main/src/main/java/io/trino/server/testing/TestingTrinoServer.java
martint: Addressed in b6cac03 — switched the new silencings to INFO→WARN where INFO chatter is the issue (CatalogStoreManager, PluginManager, io.airlift.http.server). ERROR is kept only on RemoteNodeMemory, where the existing WARN messages ("Memory info update request to %s has not returned") are transient memory-info-fetch timing artifacts under test load, not actionable.

core/trino-main/src/test/java/io/trino/execution/scheduler/faulttolerant/TestTaskDescriptorStorage.java
martint: Dropped in b6cac03. The unnamed-`Bootstrap` configuration noise is now suppressed globally by `-Dairlift.quiet=true` (airlift/airlift#1667 + trinodb/trino#27884 wiring it into `air.test.jvm.additional-arguments.default`). The remaining static block keeps only the Hibernate Validator banner silencing, and `.quiet()` on `app.initialize()` is gone for the same reason.

core/trino-main/src/test/java/io/trino/server/protocol/spooling/TestPreferredQueryDataEncoderSelector.java
martint: Narrowed and lowered in b6cac03. The silencing is now scoped to `io.trino.server.protocol.spooling.QueryDataEncoders` only (not the whole package), at WARN. In this test path the noisy line is actually `QueryDataEncoders` emitting `"Spooling protocol is enabled with encodings: ..."` once per construction (5 tests × `new QueryDataEncoders(..., enabled=true)` = 5 INFO lines per run). At WARN, the `PreferredQueryDataEncoderSelector` WARN diagnostics the tests exercise (`"Requested one of the spooled encodings: ..., but these are only available: ..."`) still print, so we're not masking anything the tests depend on.

plugin/trino-exchange-filesystem/src/test/java/io/trino/plugin/exchange/filesystem/s3/TestS3FileSystemExchangeManager.java
martint: Dropped entirely in b6cac03 — all five `io.trino.bootstrap.exchange*` silencings are gone now that `-Dairlift.quiet=true` lands in `air.test.jvm.additional-arguments.default` (airlift/airlift#1667 merged 2025-12-04 and propagated via trinodb/trino#27884). The `TestAlluxioFileSystemExchangeManager` / `TestLocalFileSystemExchangeManager` / `TestS3FileSystemExchangeManager{,SseKms,SseS3}` files are no longer touched by the PR at all.

core/trino-main/src/test/java/io/trino/server/protocol/spooling/TestPreferredQueryDataEncoderSelector.java
```diff
@@ -24,6 +26,11 @@
 
 class TestPreferredQueryDataEncoderSelector
 {
+    static {
+        Logging logging = Logging.initialize();
+        logging.setLevel("io.trino.server.protocol.spooling.QueryDataEncoders", Level.WARN);
```
findepi, line 31: it's our own class. is it logging too verbosely?

PR #27551, opened 2025-12-04
Skip underlying query analysis if MV is fresh

piotrrzysko:
## Description

Analyzing the underlying query can be costly in some scenarios, especially when it references many tables. An additional benefit of skipping this analysis is improved availability - MV data sources don’t need to be reachable when accessing the MV.

We are working on adding a `WHEN STALE` clause to `CREATE MATERIALIZED VIEW`. Relevant PRs:
* https://github.com/trinodb/trino/pull/27356
* https://github.com/trinodb/trino/pull/27502

The change proposed in this PR makes the definitions of the behaviors controlled by the new clause simpler and more symmetric; see this comment: https://github.com/trinodb/trino/pull/27356#discussion_r2537145553.

Additionally, it brings the current MV implementation closer to resolving https://github.com/trinodb/trino/discussions/23387#discussioncomment-10640385 and https://github.com/trinodb/trino/issues/23747.

- fixes https://github.com/trinodb/trino/issues/23747

## Release notes

```markdown
## Section
* Improve performance of queries referencing fresh materialized views. ({issue}`27551`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
@@ -2610,19 +2610,6 @@ class StatementAnalyzer
                 throw semanticException(VIEW_IS_RECURSIVE, table, "View is recursive");
             }
 
-            Query query = parseView(originalSql, name, table);
-
-            if (!query.getFunctions().isEmpty()) {
-                throw semanticException(NOT_SUPPORTED, table, "View contains inline function: %s", name);
-            }
-
-            analysis.registerTableForView(table, name, isMaterializedView);
-            RelationType descriptor = analyzeView(query, name, catalog, schema, owner, path, table);
-            analysis.unregisterTableForView();
-
-            checkViewStaleness(columns, descriptor.getVisibleFields(), name, table)
-                    .ifPresent(explanation -> { throw semanticException(VIEW_IS_STALE, table, "View '%s' is stale or in invalid state: %s", name, explanation); });
-
             // Derive the type of the view from the stored definition, not from the analysis of the underlying query.
             // This is needed in case the underlying table(s) changed and the query in the view now produces types that
             // are implicitly coercible to the declared view types.
@@ -2642,6 +2629,19 @@ class StatementAnalyzer
                 analysis.setMaterializedViewStorageTableFields(table, storageTableFields);
             }
             else {
+                Query query = parseView(originalSql, name, table);
+
+                if (!query.getFunctions().isEmpty()) {
+                    throw semanticException(NOT_SUPPORTED, table, "View contains inline function: %s", name);
+                }
+
+                analysis.registerTableForView(table, name, isMaterializedView);
+                RelationType descriptor = analyzeView(query, name, catalog, schema, owner, path, table);
+                analysis.unregisterTableForView();
+
+                checkViewStaleness(columns, descriptor.getVisibleFields(), name, table)
+                        .ifPresent(explanation -> { throw semanticException(VIEW_IS_STALE, table, "View '%s' is stale or in invalid state: %s", name, explanation); });
+
                 analysis.registerNamedQuery(table, query);
             }
 
--- a/core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
+++ b/core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
@@ -5798,9 +5798,7 @@ public class TestAnalyzer
     public void testAnalyzeFreshMaterializedView()
     {
         analyze("SELECT * FROM fresh_materialized_view");
-        assertFails("SELECT * FROM fresh_materialized_view_non_existent_table")
-                .hasErrorCode(INVALID_VIEW)
-                .hasMessage("line 1:15: Failed analyzing stored view 'tpch.s1.fresh_materialized_view_non_existent_table': line 1:18: Table 'tpch.s1.non_existent_table' does not exist");
+        analyze("SELECT * FROM fresh_materialized_view_non_existent_table");
         assertFails("REFRESH MATERIALIZED VIEW fresh_materialized_view_non_existent_table")
                 .hasErrorCode(TABLE_NOT_FOUND)
                 .hasMessage("line 1:18: Table 'tpch.s1.non_existent_table' does not exist");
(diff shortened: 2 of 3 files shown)


findepi:
> (x) Release notes are required, with the following suggested text:
> 
> ```
> ## Section
> * Skip analysis of underlying queries for fresh materialized views. ({issue}`issuenumber`)
> ```

"Improve performance ..."

core/trino-main/src/test/java/io/trino/connector/MockConnector.java
```diff
@@ -156,6 +156,7 @@ public class MockConnector
     private final BiFunction<ConnectorSession, SchemaTablePrefix, Map<SchemaTableName, ConnectorMaterializedViewDefinition>> getMaterializedViews;
     private final BiFunction<ConnectorSession, SchemaTableName, Boolean> delegateMaterializedViewRefreshToConnector;
     private final BiFunction<ConnectorSession, SchemaTableName, CompletableFuture<?>> refreshMaterializedView;
+    private final BiFunction<ConnectorSession, SchemaTableName, Optional<MaterializedViewFreshness>> getMaterializedViewFreshness;
```
findepi: why not `Optional<BiFunction<ConnectorSession, SchemaTableName, MaterializedViewFreshness>>`?
that would be simpler to reason about
also, in the future we should be able to remove the old `getMaterializedViewFreshness` handling code from mock connector.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -2642,6 +2629,19 @@ private Scope createScopeForView(
                 analysis.setMaterializedViewStorageTableFields(table, storageTableFields);
             }
             else {
+                Query query = parseView(originalSql, name, table);
```
findepi, line 2632: this is now under "view lacks storage table" condition
whereas it as supposed to be under "view is not fresh condition"
the calling code (2 up) suggests that `storageTable` is present _here_ only when view is considered fresh enough, but this intent is not clear in this function. requires at least a code comment

testing/trino-tests/src/test/java/io/trino/execution/TestEventListenerBasic.java
```diff
@@ -540,8 +542,6 @@ public void testReferencedTablesWithMaterializedViewsStale()
                 .hasViewText("SELECT nationkey AS test_column FROM tpch.tiny.nation");
     }
 
-    // Currently, the storage table for a materialized view is not included anywhere in the set of `tables` in the query event.
-    // See for more details: https://github.com/trinodb/trino/pull/18871#discussion_r1412247513
```
findepi, line 547: Why this _moved_ to `testReferencedTablesWithMaterializedViewsStale`?
Was this comment incorrect also before the changes? (yes => separate commit)

findepi:
We have a test `io.trino.plugin.iceberg.TestIcebergMetastoreAccessOperations#testSelectFromMaterializedView`.
We should add another test, with a fresh mv. Add it as a prep commit.
Then the `Skip underlying query analysis if MV is fresh` commit should update the counters in that test.

testing/trino-tests/src/test/java/io/trino/execution/TestEventListenerBasic.java
piotrrzysko: Instead of moving it, I should have copied it. Fixed.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
piotrrzysko: Good point. Renamed `storageTable` to `freshStorageTable`. I think with this change, a comment is not necessary.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMetastoreAccessOperations.java
```diff
@@ -210,6 +210,19 @@ public void testSelectFromMaterializedView()
                         .build());
     }
 
+    @Test
+    public void testSelectFromFreshMaterializedView()
```
piotrrzysko, line 214: @findepi 

> We have a test io.trino.plugin.iceberg.TestIcebergMetastoreAccessOperations#testSelectFromMaterializedView.
We should add another test, with a fresh mv. Add it as a prep commit.
Then the Skip underlying query analysis if MV is fresh commit should update the counters in that test.

~I added a test, but the number of accessed tables hasn’t changed because I didn’t implement the second part of your other comment: https://github.com/trinodb/trino/pull/27356#discussion_r2584368397. Implementing it might be challenging because `ConnectorMetadata::getMaterializedViewFreshness` is also used by `REFRESH MATERIALIZED VIEW` to determine whether a refresh can be skipped.~

~The question is: which `Freshness` should we return if we are at least in the grace period? IMO, from the `REFRESH MATERIALIZED VIEW` perspective, the grace period shouldn’t matter if the tables are stale, and to determine that, we need to access their metadata.~

~Performance note: the fact that we still access metadata of source tables applies only to Iceberg tables, and it is parallelized.~

Addressed in `Optimize Iceberg MV freshness checks based on grace period`.

core/trino-main/src/test/java/io/trino/connector/MockConnector.java
```diff
@@ -745,6 +748,10 @@ public Map<String, Object> getMaterializedViewProperties(ConnectorSession sessio
         @Override
         public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName viewName)
         {
+            MaterializedViewFreshness freshness = getMaterializedViewFreshness.apply(session, viewName);
+            if (freshness != null) {
+                return freshness;
```
findepi: We def don't like nullable return values and prefer `Optional`s instead.
However, I also don't want the old logic below to remain here for ever, so eventually `getMaterializedViewFreshness` should return `MaterializedViewFreshness` without optional wrapper.

To avoid scope creep, let's use Optional as getMaterializedViewFreshness return type, with a TODO comment to change this, and address in a follow-up (then default getMaterializedViewFreshness will just throw as not implemented).

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -5797,6 +5798,22 @@ public void testRowPatternCountFunction()
...
     {
         analyze("SELECT * FROM fresh_materialized_view");
+        analyze("SELECT * FROM fresh_materialized_view_non_existent_table");
+        assertFails("REFRESH MATERIALIZED VIEW fresh_materialized_view_non_existent_table")
+                .hasErrorCode(TABLE_NOT_FOUND)
+                .hasMessage("line 1:18: Table 'tpch.s1.non_existent_table' does not exist");
+    }
+
+    @Test
+    public void testAnalyzeStaleMaterializedView()
```
findepi, line 5808: Can you add this new test in a prep commit just like you did in `Test Iceberg metastore access count for fresh MVs`?

findepi:
> Additionally, it brings the current MV implementation closer to resolving [#23387 (reply in thread)](https://github.com/trinodb/trino/discussions/23387#discussioncomment-10640385) and #23747.

is this PR "fixes #23747"?

findepi:
> Shouldn't there be a change visible in `TestIcebergMetastoreAccessOperations`?

from offline feedback -- no, this cuts of AC checks. metastore calls are still there because getting freshness needs it

piotrrzysko:
> is this PR "fixes https://github.com/trinodb/trino/issues/23747"?

The issue states: `The only time we should be checking if the definer has permissions to run the view attached to the query is during refresh.` However, if an MV is stale, we still perform those checks when querying the MV (`INLINE` behavior). That’s why I didn’t say in the description that this fully resolves the issue.

On the other hand, this scenario might be interpreted as a refresh rather than a simple query -- a simplified version of `WHEN STALE REFRESH` that doesn't exist yet. If that interpretation is correct, then this PR does resolve the issue.

findepi:
> The issue states: `The only time we should be checking if the definer has permissions to run the view attached to the query is during refresh.` However, if an MV is stale, we still perform those checks when querying the MV (`INLINE` behavior). That’s why I didn’t say in the description that this fully resolves the issue.

I am sure the issue actually means "check underlying permissions during `REFRESH` AND also when inlining" a.k.a. "don't check underlying permissions when selecting from fresh view"
I therefore marked this PR as "fixes https://github.com/trinodb/trino/issues/23747"

PR #6643, opened 2021-01-19
Apache DataSketches plugin for Trino

ShashwatArghode:
Plugin to query Apache Datasketches (https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html). This includes merge and estimate function for theta sketch.

PR #27563, opened 2025-12-05
Apache DataSketches plugin for Trino

sopel39:
Plugin to query Apache DataSketches (https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html). This includes merge and estimate function for theta sketch.

## Release notes

```markdown
## General
* Add DataSketches functions. ({issue}`issuenumber`)
```

Supercedes https://github.com/trinodb/trino/pull/6643

### DIFF
--- a/docs/src/main/sphinx/functions.md
+++ b/docs/src/main/sphinx/functions.md
@@ -40,6 +40,7 @@ Color               <functions/color>
 Comparison          <functions/comparison>
 Conditional         <functions/conditional>
 Conversion          <functions/conversion>
+DataSketches        <functions/datasketches>
 Date and time       <functions/datetime>
 Decimal             <functions/decimal>
 Geospatial          <functions/geospatial>
--- a/docs/src/main/sphinx/functions/datasketches.md
+++ b/docs/src/main/sphinx/functions/datasketches.md
@@ -1,0 +1,100 @@
+# DataSketches functions
+
+[Apache DataSketches](https://datasketches.apache.org) is a high-performance
+library of stochastic streaming algorithms (sketch algorithms) that produce
+compact probabilistic summaries called sketches. A sketch is a small, stateful
+data structure that processes massive data as a stream and can provide
+approximate answers with mathematical guarantees much faster than traditional
+exact methods. DataSketches functions allow querying these serialized sketches
+from Trino. Support for the
+[Theta Sketch framework](https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html)
+is available through {func}`theta_sketch_union` and
+{func}`theta_sketch_cardinality`, typically used to replace expensive
+`COUNT(DISTINCT ...)` aggregations when sketches are precomputed and stored.
+
+## Configuration
+
+Because the DataSketches functions are provided by a connector, they are not
+available by default. To enable them, you must configure a
+[catalog properties file](catalog-properties) to register the functions with
+the specified catalog name.
+
+Create a catalog properties file `etc/catalog/datasketches.properties` that
+references the `datasketches` connector:
+
+```properties
+connector.name=datasketches
+```
+
+The DataSketches functions are available with the `theta` schema name. For the
+preceding example, the functions use the `datasketches.theta` catalog and
+schema prefix.
+
+To avoid needing to reference the functions with their fully qualified name,
+configure the `sql.path` [SQL environment
+property](/admin/properties-sql-environment) in the `config.properties` file to
+include the catalog and schema prefix:
+
+```properties
+sql.path=datasketches.theta
+```
+
+Configure multiple catalogs to use the same functions with different
+DataSketches configurations. In this case, the functions must be referenced
+using their fully qualified name, rather than relying on the SQL path.
+
+:::{note}
+Trino does not create new sketches. Build Theta sketches upstream (for example,
+in Spark, Hive, or Pig using the Apache DataSketches Theta APIs) and store the
+serialized sketch bytes as a ``VARBINARY`` column. The Trino functions operate
+on serialized Theta sketches only; other sketch families are not supported.
+:::
+
+## Functions
+
+:::{function} theta_sketch_union(sketch [, nominal_entries, seed]) -> varbinary
+Returns a serialized sketch as `varbinary`, which is a merged collection of
+sketches. The optional `nominal_entries` and `seed` parameters let you specify
+non-default sketch size and seed when merging sketches created with custom
+settings.
+:::
+
+:::{function} theta_sketch_cardinality(sketch) -> double
+Returns the estimated value of the sketch.
+:::
+
+::: {function} theta_sketch_cardinality(sketch, seed) -> double
+:noindex: true
+
+Returns the estimated value of the sketch using the supplied `seed`. Use this
+when the sketch was created with a non-default seed.
+:::
+
+## Examples
+
+The following query reads precomputed customer sketches from
+`tpch.sf100000.orders`, unions them per order date, and produces an approximate
+distinct customer count alongside exact spend. Using sketches avoids a heavy
+`COUNT(DISTINCT ...)` over billions of rows while retaining predictable error
+bounds.
+
+```sql
+SELECT
+  o_orderdate AS date,
+  theta_sketch_cardinality(theta_sketch_union(o_custkey_sketch)) AS unique_user_count,
+  SUM(o_totalprice) AS user_spent
+FROM tpch.sf100000.orders
+GROUP BY o_orderdate;
+```
+
+For comparison, the exact equivalent requires the raw keys and a costly distinct
+aggregation:
+
+```sql
+SELECT
+  o_orderdate AS date,
+  COUNT(DISTINCT o_custkey) AS unique_user_count,
+  SUM(o_totalprice) AS user_spent
+FROM tpch.sf100000.orders_raw_keys
+GROUP BY o_orderdate;
+```
--- a/plugin/trino-datasketches/pom.xml
+++ b/plugin/trino-datasketches/pom.xml
@@ -1,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.trino</groupId>
+        <artifactId>trino-root</artifactId>
+        <version>480-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>trino-datasketches</artifactId>
+    <packaging>trino-plugin</packaging>
+    <name>${project.artifactId}</name>
+    <description>Trino - DataSketches functions</description>
+
+    <properties>
+        <air.main.basedir>${project.parent.basedir}</air.main.basedir>
+    </properties>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-array</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-main</artifactId>
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.datasketches</groupId>
+            <artifactId>datasketches-java</artifactId>
+            <version>9.0.0</version>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>slice</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <!-- Trino SPI -->
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-spi</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>junit-extensions</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-tpch</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.assertj</groupId>
+            <artifactId>assertj-core</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-engine</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/state/SketchStateFactory.java
+++ b/plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/state/SketchStateFactory.java
@@ -1,0 +1,248 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.datasketches.state;
+
+import io.airlift.slice.Slice;
+import io.airlift.slice.Slices;
+import io.trino.array.ObjectBigArray;
+import io.trino.spi.function.AccumulatorStateFactory;
+import io.trino.spi.function.GroupedAccumulatorState;
+import org.apache.datasketches.theta.ThetaSetOperation;
+import org.apache.datasketches.theta.ThetaSketch;
+import org.apache.datasketches.theta.ThetaUnion;
+
+import java.lang.foreign.MemorySegment;
+
+import static com.google.common.base.Preconditions.checkState;
+import static io.airlift.slice.SizeOf.instanceSize;
+
+public class SketchStateFactory
+        implements AccumulatorStateFactory<SketchState>
+{
+    @Override
+    public SketchState createSingleState()
+    {
+        return new SingleSketchState();
+    }
+
+    @Override
+    public SketchState createGroupedState()
+    {
+        return new GroupedSketchState();
+    }
+
+    public static class GroupedSketchState
+            implements GroupedAccumulatorState, SketchState
+    {
+        private static final long INSTANCE_SIZE = instanceSize(GroupedSketchState.class);
+        private int nominalEntries;
+        private long seed;
+        private long groupId;
+        private final ObjectBigArray<ThetaUnion> unions = new ObjectBigArray<>();
+        private long totalUnionSize;
+
+        @Override
+        public void ensureCapacity(int size)
+        {
+            unions.ensureCapacity(size);
+        }
+
+        @Override
+        public long getEstimatedSize()
+        {
+            return INSTANCE_SIZE + totalUnionSize + unions.sizeOf();
+        }
+
+        @Override
+        public Slice getSketch()
+        {
+            ThetaUnion union = getUnion();
+            if (union == null) {
+                return null;
+            }
+            return Slices.wrappedBuffer(union.getResult().toByteArray());
+        }
+
+        @Override
+        public int getNominalEntries()
+        {
+            return nominalEntries;
+        }
+
+        @Override
+        public long getSeed()
+        {
+            return seed;
+        }
+
+        public ThetaUnion getUnion()
+        {
+            return unions.get(groupId);
+        }
+
+        @Override
+        public void setGroupId(int groupId)
+        {
+            this.groupId = groupId;
+        }
+
+        @Override
+        public void setNominalEntries(int value)
+        {
+            nominalEntries = value;
+        }
+
+        @Override
+        public void setSeed(long value)
+        {
+            seed = value;
+        }
+
+        @Override
+        public void addSketch(Slice value)
+        {
+            if (value == null) {
+                return;
+            }
+
+            checkState(nominalEntries > 0, "nominalEntries is not set");
+            checkState(seed != 0, "seed is not set");
+            addSketchToUnion(value, nominalEntries);
+        }
+
+        @Override
+        public void merge(SketchState otherState)
+        {
+            if (otherState.getSketch() == null) {
+                return;
+            }
+            addSketchToUnion(otherState.getSketch(), otherState.getNominalEntries());
+        }
+
+        private void addSketchToUnion(Slice value, int nominalEntries)
+        {
+            checkState(value != null, "sketch is null");
+            checkState(nominalEntries > 0, "nominalEntries is not set");
+            ThetaSketch sketch = ThetaSketch.wrap(MemorySegment.ofArray(value.getBytes()), seed);
+            ThetaUnion groupedUnion = getUnion();
+            int previousSize = groupedUnion == null ? 0 : groupedUnion.getCurrentBytes();
+            if (groupedUnion == null) {
+                groupedUnion = ThetaSetOperation.builder()
+                        .setSeed(seed)
+                        .setNominalEntries(nominalEntries)
+                        .buildUnion();
+                groupedUnion.union(sketch);
+                unions.set(groupId, groupedUnion);
+            }
+            else {
+                groupedUnion.union(sketch);
+            }
+            updateMemoryUsage(groupedUnion, previousSize);
+        }
+
+        private void updateMemoryUsage(ThetaUnion union, int previousSize)
+        {
+            int newSize = union.getCurrentBytes();
+            totalUnionSize += newSize - previousSize;
+        }
+    }
+
+    public static class SingleSketchState
+            implements SketchState
+    {
+        private static final long INSTANCE_SIZE = instanceSize(SingleSketchState.class);
+        private ThetaUnion union;
+        private int nominalEntries;
+        private long seed;
+
+        @Override
+        public Slice getSketch()
+        {
+            if (union == null) {
+                return null;
+            }
+            ThetaSketch result = union.getResult();
+            return Slices.wrappedBuffer(result.toByteArray());
+        }
+
+        @Override
+        public int getNominalEntries()
+        {
+            return nominalEntries;
+        }
+
+        @Override
+        public long getSeed()
+        {
+            return seed;
+        }
+
+        @Override
+        public void addSketch(Slice value)
+        {
+            if (value == null) {
+                return;
+            }
+
+            checkState(nominalEntries > 0, "nominalEntries is not set");
+            checkState(seed != 0, "seed is not set");
+            addSketchToUnion(value, nominalEntries);
+        }
+
+        @Override
+        public void setNominalEntries(int value)
+        {
+            nominalEntries = value;
+        }
+
+        @Override
+        public void setSeed(long value)
+        {
+            seed = value;
+        }
+
+        @Override
+        public long getEstimatedSize()
+        {
+            int sketchSize = union == null ? 0 : union.getCurrentBytes();
+            return INSTANCE_SIZE + sketchSize;
+        }
+
+        @Override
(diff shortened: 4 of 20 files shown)


plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/SketchFunctionsPlugin.java
```diff
@@ -0,0 +1,33 @@
...
+        implements Plugin
+{
+    @Override
+    public Set<Class<?>> getFunctions()
+    {
+        return ImmutableSet.<Class<?>>builder()
+                .add(Estimate.class)
+                .add(Union.class)
+                .add(UnionWithParams.class)
+                .build();
```
martint: This pollutes the global function namespace -- it's the old way for plugins to expose functions. The functions should be exposed via dynamic function resolution so that they can be namespaced properly.

martint (changes requested):

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/SketchFunctionsPlugin.java
sopel39: > The functions should be exposed via dynamic function resolution so that they can be namespaced properly.

Is there an example of that?

martint: Take a look at the AI functions plugin: https://github.com/trinodb/trino/tree/master/plugin/trino-ai-functions

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,43 @@
...
+from Trino. Support for the
+[Theta Sketch framework](https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html)
+is available through {func}`theta_sketch_union` and
+{func}`theta_sketch_estimate`, typically used in `COUNT DISTINCT` queries.
+DataSketches can be created with Hive or Pig using their respective sketch
+APIs.
+
+## Functions
+
+:::{function} theta_sketch_union(sketches) -> sketch
```
findepi: is "`sketch`" a type name?

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/UnionWithParams.java
```diff
@@ -0,0 +1,70 @@
...
+import io.trino.spi.function.SqlType;
+import io.trino.spi.type.StandardTypes;
+
+@AggregationFunction("theta_sketch_union")
+public final class UnionWithParams
+{
+    private UnionWithParams() {}
+
+    @InputFunction
+    public static void input(@AggregationState SketchState state, @SqlType(StandardTypes.VARBINARY) Slice inputValue, @SqlType(StandardTypes.INTEGER) Integer normEntries, @SqlType(StandardTypes.BIGINT) Long seed)
```
findepi: undocumented?

also - normEntries - avoid abbrevs

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/UnionWithParams.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+
+        if (state.getSketch() == null) {
+            state.setSeed(otherState.getSeed());
+            state.setNominalEntries(otherState.getNominalEntries());
+            state.setSketch(otherState.getSketch());
+            return;
+        }
+
+        state.merge(otherState);
```
findepi, line 55: I think this should explicitly invoke `SetOperation.builder().buildUnion()`
See https://github.com/trinodb/trino/blob/b390d6a969a1f953f60cc9a1f8ddec2131672765/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/aggregation/IcebergThetaSketchForStats.java#L69-L80

i am guessing now -- the current code works because `@CombineFunction` is not invoked on direct output from `@InputFunction`?

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,43 @@
...
+Returns the estimated value of the sketch using the supplied `seed`. Use
+this when the sketch was created with a non-default seed.
+:::
+
+## Examples
+
+```sql
+SELECT
+  o_orderdate AS date,
+  theta_sketch_estimate(theta_sketch_union(o_custkey_sketch)) AS unique_user_count,
```
findepi: I see only functions that take a sketch as a parameter.
How to i make my first sketch?

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/UnionWithParams.java
sopel39: This is already tested `io.trino.plugin.datasketches.theta.TestMergeEstimate#testSimpleMerge`

`SketchState.merge` creates a union via `ThetaSetOperation.builder().buildUnion()` when both sides have sketches. If the left state is empty it just copies the right.

docs/src/main/sphinx/functions/datasketches.md
sopel39: Sketches are produced either by other engines or ingestion pipelines (e.g. Spark or Hive) rather than Trino itself. Apache DataSketches are used quite commonly.

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/UnionWithParams.java
findepi: a unit test as in `TestMergeEstimate#testSimpleMerge` does not tell me what _guarantees_ the engine provides.
in particular, the test invokes `compact` which , afair, is pretty important for internal structure

i am guessing again -- the current code works because `@CombineFunction` is not invoked on direct output from `@InputFunction`?

plugin/trino-datasketches/src/test/java/io/trino/plugin/datasketches/TestDataSketches.java
```diff
@@ -0,0 +1,215 @@
...
+        DistributedQueryRunner queryRunner = DistributedQueryRunner.builder(session).build();
+        queryRunner.installPlugin(new SketchFunctionsPlugin());
+        queryRunner.createCatalog("datasketches", "datasketches");
+        queryRunner.installPlugin(new TpchPlugin());
+        queryRunner.createCatalog("tpch", "tpch");
+        return queryRunner;
+    }
+
+    @Test
+    public void testSimpleMerge()
```
findepi, line 75: This unit test should not be in `AbstractTestQueryFramework` class.
it doesn't use QueryRunner, so running this test should not require paying the cost of QR setup

plugin/trino-datasketches/src/test/java/io/trino/plugin/datasketches/theta/TestMergeEstimate.java
```diff
@@ -0,0 +1,215 @@
...
+    private static final long TEST_SEED = 95869L;
+    private static final int TEST_ENTRIES = 2048;
+    private static final int DEFAULT_ENTRIES = DEFAULT_NOMINAL_ENTRIES;
+
+    @BeforeAll
+    public void setup()
+    {
+        AccumulatorStateFactory<SketchState> factory = new SketchStateFactory();
+        state1 = factory.createSingleState();
+        state2 = factory.createSingleState();
```
findepi: is this shared mutable state in a concurrent test class?

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/UnionWithParams.java
```diff
@@ -0,0 +1,70 @@
...
+public final class UnionWithParams
+{
+    private UnionWithParams() {}
+
+    @InputFunction
+    public static void input(@AggregationState SketchState state, @SqlType(StandardTypes.VARBINARY) Slice inputValue, @SqlType(StandardTypes.INTEGER) Integer nominalEntries, @SqlType(StandardTypes.BIGINT) Long seed)
+    {
+        state.setNominalEntries(nominalEntries);
+        state.setSeed(seed);
+        state.setSketch(inputValue);
```
findepi: This looks the code _overrides_ state in the `@InputFunction` function.
Isn't the `@InputFunction` function supposed to accumulate?

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
+# DataSketches functions
+
+DataSketches is a high-performance library of stochastic streaming
+algorithms, commonly called sketches. Sketches are small, stateful programs
```
findepi: This document uses "sketch" to describe an algorithm (here), but also a data structure (when used as a type name).
This is confusing.

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+that process massive data as a stream and can provide approximate answers,
+with mathematical guarantees, much faster than traditional exact methods.
+The DataSketches functions allow querying the fast and memory-efficient
+[Apache DataSketches](https://datasketches.apache.org/docs/Community/Research.html)
+from Trino. Support for the
+[Theta Sketch framework](https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html)
+is available through {func}`theta_sketch_union` and
+{func}`theta_sketch_estimate`, typically used in `COUNT DISTINCT` queries.
```
findepi: If this is a typical usage, do you want to show it in an example?
Or did you mean "_instead_ of `count(DISTINCT ...)` queries"?

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+DataSketches can be created with Hive or Pig using their respective sketch
+APIs.
```
findepi: I'd rather see a concrete instructions user can follow to create these data sketches. IMO they should not need to go to a different system. It's really not a big lift especially that Trino already has an aggregation function for building them (`IcebergThetaSketchForStats`), it's just not user facing.

cc @mosabua for how to reference to other systems.

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+
+To use these functions, configure a catalog with the
+``datasketches`` connector (for example, name the catalog ``datasketches``)
+and either qualify function calls with ``datasketches.theta.`` or add
+``datasketches.theta`` to the session ``path``.
+
+Note: Trino does not create new sketches. Build the Theta sketches in your
+ingest pipeline or another engine (e.g., Spark or Hive) and store the
+serialized sketch bytes as a ``VARBINARY`` column. Trino functions operate on those serialized sketches.
```
findepi: DataSketches supports many sketch families and serialization format is to some extend configurable. This should say what's supported.

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Functions
+
+:::{function} theta_sketch_union(sketches [, nominal_entries, seed]) -> varbinary
```
findepi: Why is it aggregation function? should there be a corresponding scalar function for unioning two sketches?

Also sketches -> sketch? or sketches -> varbinary? cc @mosabua

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/state/SketchStateSerializer.java
```diff
@@ -0,0 +1,70 @@
...
+            Slice slice = state.getSketch();
+
+            SliceOutput sliceOutput = new DynamicSliceOutput(SIZE_OF_LONG + SIZE_OF_INT + SIZE_OF_INT + slice.length());
+            sliceOutput.appendInt(state.getNominalEntries());
+            sliceOutput.appendLong(state.getSeed());
+
+            sliceOutput.appendInt(slice.length());
+            sliceOutput.appendBytes(slice);
+
+            VARBINARY.writeSlice(out, sliceOutput.slice());
```
findepi: Compare with
https://github.com/trinodb/trino/blob/da1de5e4af3a59257fdeed6a2e14f3010e7c8e2f/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/aggregation/DataSketchStateSerializer.java#L51-L55

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
+# DataSketches functions
+
+DataSketches is a high-performance library of stochastic streaming
```
mosabua:
```suggestion
[Apache DataSketches](https://datasketches.apache.org) is a high-performance library of stochastic streaming
```

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+algorithms, commonly called sketches. Sketches are small, stateful programs
+that process massive data as a stream and can provide approximate answers,
+with mathematical guarantees, much faster than traditional exact methods.
+The DataSketches functions allow querying the fast and memory-efficient
```
mosabua:
```suggestion
DataSketches functions allow querying the fast and memory-efficient
```

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+[Apache DataSketches](https://datasketches.apache.org/docs/Community/Research.html)
```
mosabua:
```suggestion
Apache DataSketches
```

Link seems irrelevant for the specific mention and is better earlier

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+from Trino. Support for the
+[Theta Sketch framework](https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html)
+is available through {func}`theta_sketch_union` and
+{func}`theta_sketch_estimate`, typically used in `COUNT DISTINCT` queries.
+DataSketches can be created with Hive or Pig using their respective sketch
```
mosabua: How is that relevant? Maybe we need to say something like .. 

DataSketches must be created with Hive or Pig... , before you can invoke them from Trino.

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+APIs.
+
+To use these functions, configure a catalog with the
+``datasketches`` connector (for example, name the catalog ``datasketches``)
+and either qualify function calls with ``datasketches.theta.`` or add
+``datasketches.theta`` to the session ``path``.
+
+Note: Trino does not create new sketches. Build the Theta sketches in your
```
mosabua: Maybe combine above section about Hive/Pig with this. And if you really want this to be a note .. use the sphinx syntax for the right layout

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+The DataSketches functions allow querying the fast and memory-efficient
+[Apache DataSketches](https://datasketches.apache.org/docs/Community/Research.html)
+from Trino. Support for the
+[Theta Sketch framework](https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html)
+is available through {func}`theta_sketch_union` and
+{func}`theta_sketch_estimate`, typically used in `COUNT DISTINCT` queries.
+DataSketches can be created with Hive or Pig using their respective sketch
+APIs.
+
+To use these functions, configure a catalog with the
```
mosabua: Be more explicit what needs to be done .. use https://trino.io/docs/current/functions/ai.html as inspiration on how to reword and add example

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+``datasketches.theta`` to the session ``path``.
+
+Note: Trino does not create new sketches. Build the Theta sketches in your
+ingest pipeline or another engine (e.g., Spark or Hive) and store the
+serialized sketch bytes as a ``VARBINARY`` column. Trino functions operate on those serialized sketches.
+
+## Functions
+
+:::{function} theta_sketch_union(sketches [, nominal_entries, seed]) -> varbinary
+Returns a serialized sketch (VARBINARY) which is a merged collection of sketches. The optional
```
mosabua:
```suggestion
Returns a serialized sketch as `VARBINARY`, which is a merged collection of sketches. The optional
```

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,54 @@
...
+::: {function} theta_sketch_estimate(sketch, seed) -> double
+:noindex: true
+
+Returns the estimated value of the sketch using the supplied `seed`. Use
+this when the sketch was created with a non-default seed.
+:::
+
+## Examples
+
+```sql
```
mosabua, line 81: Add a sentence that describes what the queyr does and explain advantage of using the function. Maybe even have equivalent normal query and info about performance and accuracy ..

docs/src/main/sphinx/functions.md
```diff
@@ -40,6 +40,7 @@ Color               <functions/color>
 Comparison          <functions/comparison>
 Conditional         <functions/conditional>
 Conversion          <functions/conversion>
+Datasketches        <functions/datasketches>
```
mosabua:
```suggestion
DataSketches        <functions/datasketches>
```

mosabua (changes requested):
Mostly good to go. Left some minor comments. Main one is to address the installation/use better.

Fixes https://github.com/simpligility/contributions/issues/48

plugin/trino-datasketches/src/test/java/io/trino/plugin/datasketches/TestDataSketches.java
sopel39: We setup QR anyway for other tests, and IMO it's better to group Sketches tests in single class. Don't you agree?

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/state/SketchStateSerializer.java
sopel39: Simplified code to store `ThetaUnion` in accumulators. `CompactThetaSketch` is only used for outputs and serde now.

docs/src/main/sphinx/functions/datasketches.md
sopel39: Here we only support Theta sketches.

I think general theta sketches framework requires much boarder scope, possibly dedicated type (parametrized with seed) and we already have approximate functions that use HLL sketches in Trino. This PR means to provide upstream compatibility so that users can actually compute estimates (for reporting, etc) from already pre-computed Theta sketches.

docs/src/main/sphinx/functions/datasketches.md
sopel39: Removed this section

docs/src/main/sphinx/functions/datasketches.md
sopel39: It’s an aggregation because the primary use case is merging many sketch rows and we need the accumulator/state machinery for that. The current function is intentionally an aggregate to handle the general multi-row merge case.

plugin/trino-datasketches/src/test/java/io/trino/plugin/datasketches/TestDataSketches.java
findepi: > it's better to group Sketches tests in single class

i agree with that

yes, a test method not using QR should not be a in a class that creates one

docs/src/main/sphinx/functions/datasketches.md
findepi: > I'd rather see a concrete instructions user can follow to create these data sketches.

Is this addressed?

sopel39: What are you looking for specifically? I don't think we should provide explicit instructions on how to create sketches in other engines. It would be hard to maintain such documentation.

plugin/trino-datasketches/src/test/java/io/trino/plugin/datasketches/TestDataSketches.java
sopel39: > i agree with that

> yes, a test method not using QR should not be a in a class that creates one

These two are mutually exclusive. TestDataSketches is using QR and it's a single class for testing sketches

docs/src/main/sphinx/functions/datasketches.md
findepi: We have piece of documentation that a user may find interesting, but we don't help the user know how to start. That's not nice.

sopel39: I mean this feature is really for power users which require interoperability. They dont really require hand holding in this matter.

findepi: what are the user docs for?  How do we prevent the user from feeling dumb that they didn't infer from the docs how to use this feature, because user requires too much hand holding?

sopel39: I think the does are sufficient from trino side, wdyt @mosabua ?

mosabua: I think it is fine to just link to external docs for how to create the sketch. However if there is a Trino-specific tip or variant or consideration those should be added as well.

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,88 @@
+# DataSketches functions
+
+[Apache DataSketches](https://datasketches.apache.org) is a high-performance library of stochastic streaming
```
mosabua: Weird wrapping going on here.. can you just use 80char?

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,88 @@
...
+[Theta Sketch framework](https://datasketches.apache.org/docs/Theta/ThetaSketchFramework.html)
+is available through {func}`theta_sketch_union` and
+{func}`theta_sketch_estimate`, typically used to replace expensive
+`COUNT(DISTINCT ...)` aggregations when sketches are precomputed and stored.
+
+## Configuration
+
+Enable the functions by adding a catalog that uses the `datasketches` connector and exposes the `theta` schema.
+
+Create a catalog properties file (for example, `etc/catalog/datasketches.properties`) that references the `datasketches` connector:
```
mosabua:
```suggestion
Create a catalog properties file, for example `etc/catalog/datasketches.properties`, that references the `datasketches` connector:
```

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,88 @@
...
+
+```properties
+connector.name=datasketches
+```
+
+The functions are available with the `datasketches.theta` schema name. For the preceding example, the functions use the `datasketches.theta` catalog and schema prefix.
```
mosabua: The schema name is just theta though right .. and then you reference them with catalogname.theta .. see what wording we use in the AI functions maybe

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,88 @@
...
+
+To avoid needing to reference the functions with their fully qualified name, configure the `sql.path` [SQL environment property](/admin/properties-sql-environment)
+in the `config.properties` file to include the catalog and schema prefix:
+
+```properties
+sql.path=datasketches.theta
+```
+
+To use these functions, configure a catalog with the
```
mosabua: This duplicates the above.. clean up please

docs/src/main/sphinx/functions/datasketches.md
```diff
@@ -0,0 +1,88 @@
...
+```sql
+SELECT
+  o_orderdate AS date,
+  theta_sketch_estimate(theta_sketch_union(o_custkey_sketch)) AS unique_user_count,
+  SUM(o_totalprice) AS user_spent
+FROM tpch.sf100000.orders
+GROUP BY o_orderdate;
+```
+
+For comparison, the exact equivalent would need the raw keys and a costly distinct aggregation:
```
mosabua:
```suggestion
For comparison, the exact equivalent requires the raw keys and a costly distinct aggregation:
```

mosabua:
Adding `stale-ignore` label since I am pretty sure @sopel39 wants to continue driving this to merge. Maybe @findepi can help out.

plugin/trino-datasketches/pom.xml
```diff
@@ -0,0 +1,104 @@
...
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.datasketches</groupId>
+            <artifactId>datasketches-java</artifactId>
+            <version>9.0.0</version>
```
wendigo, line 46: Nice. I'm updating to the same version in iceberg connector here: https://github.com/trinodb/trino/pull/28046

plugin/trino-datasketches/src/test/java/io/trino/plugin/datasketches/TestDataSketches.java
```diff
@@ -0,0 +1,226 @@
...
+        union.union(sketch2);
+        ThetaSketch unionResult = union.getResult();
+
+        assertThat(estimate).isEqualTo(unionResult.getEstimate());
+    }
+
+    @Test
+    public void testMergeAndEstimate()
+    {
+        int[] idACatagory = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
```
martint: Typo: category

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/Estimate.java
```diff
@@ -0,0 +1,49 @@
...
+
+import java.lang.foreign.MemorySegment;
+
+import static org.apache.datasketches.common.Util.DEFAULT_UPDATE_SEED;
+
+public class Estimate
+{
+    private Estimate() {}
+
+    @ScalarFunction("theta_sketch_estimate")
```
martint: What does the estimate produce? Is it an approximate distinct count? Are there other types of sketches that can produce different types of estimates?

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/Estimate.java
```diff
@@ -0,0 +1,49 @@
...
+    @SqlType(StandardTypes.DOUBLE)
+    public static double estimate(@SqlType(StandardTypes.VARBINARY) Slice inputValue)
+    {
+        return estimate(inputValue, DEFAULT_UPDATE_SEED);
+    }
+
+    @ScalarFunction("theta_sketch_estimate")
+    @Description("Converts sketch bytearrays to double estimate")
+    @SqlType(StandardTypes.DOUBLE)
+    public static double estimate(@SqlType(StandardTypes.VARBINARY) Slice inputValue, @SqlType(StandardTypes.BIGINT) long seed)
```
martint, line 42: What's the purpose of `seed`?

plugin/trino-datasketches/pom.xml
sopel39: That one landed :)

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/Estimate.java
sopel39: `theta_sketch_estimate` returns estimated cardinality (approximate distinct count). Only Theta sketches are supported

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/Estimate.java
sopel39: It is the hash seed used to interpret/merge the sketch consistently. It must match the seed used when the sketch was built/unioned

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/theta/Estimate.java
martint: In that case, I'd name it "theta_sketch_cardinality"

sopel39: done

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/state/SketchStateFactory.java
```diff
@@ -0,0 +1,248 @@
...
+        public long getEstimatedSize()
+        {
+            return INSTANCE_SIZE + totalUnionSize + unions.sizeOf();
+        }
+
+        @Override
+        public Slice getSketch()
+        {
+            ThetaUnion union = getUnion();
+            if (union != null) {
```
wendigo: I'd reverse this condition:

```
if (union == null) {
   return null;
}
```

plugin/trino-datasketches/src/main/java/io/trino/plugin/datasketches/state/SketchStateSerializer.java
```diff
@@ -0,0 +1,70 @@
...
+    @Override
+    public void serialize(SketchState state, BlockBuilder out)
+    {
+        if (state.getSketch() == null) {
+            out.appendNull();
+        }
+        else {
+            Slice slice = state.getSketch();
+
+            SliceOutput sliceOutput = new DynamicSliceOutput(SIZE_OF_LONG + SIZE_OF_INT + SIZE_OF_INT + slice.length());
```
wendigo: You don't need DynamicSliceOutput.

`Slice serialized = Slices.allocate(SIZE_OF_LONG + SIZE_OF_INT + SIZE_OF_INT + slice.length())`

`SliceOutput output = serialized.getOutput()`

and then:

```
VARBINARY.writeSlice(out, serialized);
```

wendigo (approved):

sopel39:
Thanks for review

PR #27566, opened 2025-12-05
[Exasol] Switch off Exasol testsuite

pj-spoelders:
## Description

Turning off Exasol testsuite, for now. 
We were forced to update to 2025.1.8 this morning and it does not seem stable.
2025.1.3 is no longer available.
We will turn off the tests for now so we don't hinder everyone. 
We will turn them back on once we found the issue and have a new stable container to use.

## Release notes

(x ) This is not user-visible or is docs only, and no release notes are required.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -912,7 +912,7 @@ jobs:
             - suite-delta-lake-databricks154
             - suite-delta-lake-databricks164
             - suite-delta-lake-databricks173
-            - suite-exasol
+            # - suite-exasol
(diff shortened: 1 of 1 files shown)


wendigo (approved):

findepi:
> We were forced to update to 2025.1.8 this morning and it does not seem stable.
> 2025.1.3 is no longer available.

It's not cool when docker images disappear like this.
For example, it means that if i checkout a commit from a month ago, the tests may not run.
Can you please work with the image provider to ensure they don't delete images?

PR #27569, opened 2025-12-07
Relax validation for empty pages in PageValidations

losipiuk:
## Description

Allow empty pages without blocks. It is valid for connectors/table
functions to produce empty pages with empty block list interleaved with
pages with data. Empty pages are filtered out by Driver.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/split/PageValidations.java
+++ b/core/trino-main/src/main/java/io/trino/split/PageValidations.java
@@ -32,6 +32,14 @@ public final class PageValidations
 
     public static void validateOutputPageTypes(SourcePage page, List<Type> expectedTypes, Supplier<String> debugContextSupplier)
     {
+        if (page.getPositionCount() == 0 && page.getChannelCount() == 0) {
+            // Allow empty pages without blocks. It is valid for connectors/table functions to produce empty pages with empty block list
+            // interleaved with pages with data. Example of such behavior can be found in TableChangesFunctionProcessor (this is not exactly
+            // related to SourcePage, but illustrates the contract, which is the same for Pages and SourcePages).
+            // Empty pages are filtered out by Driver.
+            return;
+        }
+
         if (page.getChannelCount() != expectedTypes.size()) {
             throw new TrinoException(
                     GENERIC_INTERNAL_ERROR,


core/trino-main/src/main/java/io/trino/split/PageValidations.java
```diff
@@ -32,6 +32,11 @@ private PageValidations() {}
 
     public static void validateOutputPageTypes(SourcePage page, List<Type> expectedTypes, Supplier<String> debugContextSupplier)
     {
+        if (page.getPositionCount() == 0 && page.getChannelCount() == 0) {
```
raunaqmorarka, line 35: I would drop `&& page.getChannelCount() == 0`
It is valid to create an empty Page as `new Page(0, new Block[N])`, but that would throw a nullptr exception in the current code.
We should probably tighten up the Page constructor to disallow that.

findepi: > It is valid to create an empty Page as `new Page(0, new Block[N])`, but that would throw a nullptr exception in the current code.

how would that be a valid page?
The blocks are not nullable in the page

> We should probably tighten up the Page constructor to disallow that.

it doesn't have to be a runtime check (might be redundant in hot path)
the fact there is no such check should not be construed as implying that null blocks are OK

core/trino-main/src/main/java/io/trino/split/PageValidations.java
```diff
@@ -32,6 +32,11 @@ private PageValidations() {}
...
+            // allow empty pages without blocks
```
findepi: The comment says what we do, but doesn't say _why_. Please update & update PR description accordingly.

Also, it's not a universal truth that empty page can be any shape. For example
https://github.com/trinodb/trino/blob/26846cb021841998598ad22847c8e68ed85b73b3/core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java#L121-L128
This may not break because empty pages are filtered somewhere earlier.

losipiuk: > The comment says what we do, but doesn't say why. Please update & update PR description accordingly.

That is a good point. I do not know any concrete place in code which does produce such pages. It is based on hint from @raunaqmorarka that it is valid scenario, so I sent this PR defensively. 
@raunaqmorarka do you maybe have specific code pointers?

findepi: > I do not know any concrete place in code which does produce such pages.

example
https://github.com/trinodb/trino/blob/51087fb80428981e1adb99e2386bc00f2ffb7aea/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/functions/tablechanges/TableChangesFunctionProcessor.java#L137

losipiuk: Thank you

core/trino-main/src/main/java/io/trino/split/PageValidations.java
```diff
@@ -32,6 +32,14 @@ private PageValidations() {}
 
     public static void validateOutputPageTypes(SourcePage page, List<Type> expectedTypes, Supplier<String> debugContextSupplier)
     {
+        if (page.getPositionCount() == 0 && page.getChannelCount() == 0) {
+            // Allow empty pages without blocks. It is valid for connectors/table functions to produce empty pages with empty block list
```
findepi, line 36: per https://github.com/trinodb/trino/blob/993a08840651a3e7c7a6e16ffbbc6bc2c02346af/core/trino-main/src/main/java/io/trino/operator/Driver.java#L407
drop `page.getChannelCount() == 0` condition and update the comment

core/trino-main/src/main/java/io/trino/split/PageValidations.java
```diff
@@ -32,6 +32,14 @@ private PageValidations() {}
...
+            // interleaved with pages with data. Example of such behavior can be found in TableChangesFunctionProcessor (this is not exactly
```
findepi, line 37:
```suggestion
            // interleaved with pages with data. Example of such behavior can be found in Delta's TableChangesFunctionProcessor (this is not exactly
```

core/trino-main/src/main/java/io/trino/split/PageValidations.java
```diff
@@ -32,6 +32,14 @@ private PageValidations() {}
...
+            // related to SourcePage, but illustrates the contract, which is the same for Pages and SourcePages).
```
findepi, line 38: I realized that for a SourcePage, the check may be triggering a load of lazy data.
If we had applied the check outside of `ScanFilterAndProjectOperator`, this wouldn't be a problem (and we wouldn't see SourcePage), but now maybe it is?

core/trino-main/src/main/java/io/trino/split/PageValidations.java
losipiuk: I understood from the discussion that such pages are really broken and not created? Are we creating such anywhere. Or just drop just in case?

findepi: i don't know of any code creating empty pages with zero position count and non-zero, but incorrect, channel count

> Or just drop just in case?


this. since the Driver condition is what it and same is in Scan Filter operator
https://github.com/trinodb/trino/blob/5bdd75d1b5ce66ca721a0b2474acbcb9d1f8cddb/core/trino-main/src/main/java/io/trino/operator/project/PageProcessor.java#L115-L116

losipiuk:
Incorporated into https://github.com/trinodb/trino/pull/27582

Issue #4107, opened 2020-06-20
Improve statistics estimates for range predicates over int/long types


rzeyde-varada:
Following:
https://github.com/prestosql/presto/pull/3316#issuecomment-641937684_

PR #27570, opened 2025-12-07
Improve overlap percent estimation for low-density ranges in StatisticRange

anton-kutuzov:
## Description

Join order can be misestimated due to the uniform distribution assumption in `StatisticRange.overlapPercentWith()`.

When a column has a very wide numeric range but few distinct values `(e.g., distinctValues = 14, low = 1, high = 3.6e9)`, the current overlap estimation becomes extremely small `(e.g., 8.19e-10)`, underestimating join cardinalities.

**Example:**
```
SELECT *
FROM table1 t1
JOIN table2 t2
  ON t1.eid = t2.eid
WHERE CAST(event_date AS DATE) = DATE '2025-09-07'
  AND t1.platform_id IN (1, 2, 3, 4);
```

`table1` is large and `table2` is small.
The column `platform_id` in `table1` has `14` distinct values, with `low = 1` and `high = 3 662 098 119`.
In this case, the method `StatisticRange.overlapPercentWith()` estimates the overlap as
`(4 - 1) / (3,662,098,119 - 1) ≈ 8.19e-10`
which effectively means “all rows are filtered out”.
But in reality, the filter `IN (1,2,3,4)` should keep roughly `4` out of `14` values `(~29%)`.

**Solution:**
Introduce a density check `density = distinctValues / (high - low)` and combine uniform overlap with NDV-based estimate when density is low.

Fixes https://github.com/trinodb/trino/issues/4107

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
+++ b/core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
@@ -32,6 +32,7 @@ public class StatisticRange
 {
     private static final double INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR = 0.25;
     private static final double INFINITE_TO_INFINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR = 0.5;
+    private static final double DENSITY_HEURISTIC_THRESHOLD = 1e-3;
 
     // TODO unify field and method names with SymbolStatsEstimate
     /**
@@ -122,7 +123,17 @@ public class StatisticRange
         if (isInfinite(length()) && isFinite(lengthOfIntersect)) {
             return INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR;
         }
+
         if (lengthOfIntersect > 0) {
+            double thisDensity = this.distinctValues / length();
+
+            // Sparse columns can produce near-zero uniform overlap due to large ranges,
+            // leading to incorrect join ordering. Use NDV-based heuristic instead.
+            if (!isNaN(thisDensity)
+                    && lengthOfIntersect != length()
+                    && thisDensity < DENSITY_HEURISTIC_THRESHOLD) {
+                return minExcludeNaN(other.distinctValues / this.distinctValues, 1);
+            }
             return lengthOfIntersect / length();
         }
 
--- a/core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
+++ b/core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
@@ -60,6 +60,60 @@ public class TestStatisticRange
     }
 
     @Test
+    public void testLowDensityOverlapWithSparseLeftRange()
+    {
+        StatisticRange sparseRange = range(1, 4, NaN);
+        StatisticRange filterRange = range(1, 3662098119.0, 14);
+
+        assertOverlap(sparseRange, filterRange, 1);
+    }
+
+    @Test
+    public void testLowDensityOverlapWithSparseRightRange()
+    {
+        StatisticRange sparseRange = range(1, 3662098119.0, 14);
+        StatisticRange filterRange = range(1, 4, NaN);
+
+        assertOverlap(sparseRange, filterRange, 1);
+    }
+
+    @Test
+    public void testDensityThresholdBoundary()
+    {
+        StatisticRange boundaryRange = range(0, 10000, 10);
+        StatisticRange smallFilter = range(0, 100, 5);
+
+        assertOverlap(boundaryRange, smallFilter, 0.01);
+    }
+
+    @Test
+    public void testHighDensityOverlap()
+    {
+        StatisticRange denseRange = range(0, 100, 50);
+        StatisticRange filterRange = range(20, 30, 5);
+
+        assertOverlap(denseRange, filterRange, 0.1);
+    }
+
+    @Test
+    public void testVeryLowDensity()
+    {
+        StatisticRange verySparse = range(0, 1e9, 10);
+        StatisticRange filterRange = range(100, 200, 5);
+
+        assertOverlap(verySparse, filterRange, 0.5);
+    }
+
+    @Test
+    public void testDensityWithZeroDistinctValues()
+    {
+        StatisticRange zeroDistinct = range(0, 1000, 0);
+        StatisticRange filterRange = range(100, 200, 5);
+
+        assertOverlap(zeroDistinct, filterRange, 0);
+    }
+
+    @Test
     public void testIntersect()
     {
         StatisticRange zeroToTen = range(0, 10, 10);
--- a/core/trino-main/src/test/java/io/trino/cost/TestFilterStatsCalculator.java
+++ b/core/trino-main/src/test/java/io/trino/cost/TestFilterStatsCalculator.java
@@ -47,6 +47,7 @@ import java.math.BigDecimal;
 import java.util.function.Consumer;
 
 import static io.trino.SystemSessionProperties.FILTER_CONJUNCTION_INDEPENDENCE_FACTOR;
+import static io.trino.spi.type.BigintType.BIGINT;
 import static io.trino.spi.type.BooleanType.BOOLEAN;
 import static io.trino.spi.type.DecimalType.createDecimalType;
 import static io.trino.spi.type.DoubleType.DOUBLE;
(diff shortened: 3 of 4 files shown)


core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
```diff
@@ -122,7 +123,17 @@ public double overlapPercentWith(StatisticRange other)
         if (isInfinite(length()) && isFinite(lengthOfIntersect)) {
             return INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR;
         }
+
         if (lengthOfIntersect > 0) {
+            double thisDensity = this.distinctValues / length();
+            double otherDensity = other.distinctValues / other.length();
+            double minDensity = minExcludeNaN(thisDensity, otherDensity);
+
+            if (!isNaN(thisDensity) && !isNaN(otherDensity)
```
raunaqmorarka: `!isNaN(thisDensity) && !isNaN(otherDensity)` -> `!isNaN(minDensity)`

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
```diff
@@ -122,7 +123,17 @@ public double overlapPercentWith(StatisticRange other)
...
+                    && isFinite(length()) && isFinite(other.length())
+                    && minDensity < DENSITY_HEURISTIC_THRESHOLD) {
+                return minExcludeNaN(this.distinctValues, other.distinctValues) / this.distinctValues;
```
raunaqmorarka: can this be `return min(other.distinctValues / this.distinctValues, 1);` ?

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
```diff
@@ -122,7 +123,17 @@ public double overlapPercentWith(StatisticRange other)
         if (isInfinite(length()) && isFinite(lengthOfIntersect)) {
             return INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR;
         }
+
         if (lengthOfIntersect > 0) {
+            double thisDensity = this.distinctValues / length();
```
raunaqmorarka, line 128: Please add a code comment explaining this section

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
```diff
@@ -122,7 +123,17 @@ public double overlapPercentWith(StatisticRange other)
...
+            double otherDensity = other.distinctValues / other.length();
+            double minDensity = minExcludeNaN(thisDensity, otherDensity);
+
+            if (!isNaN(thisDensity) && !isNaN(otherDensity)
+                    && isFinite(length()) && isFinite(other.length())
```
raunaqmorarka: Why do we check that the lengths are finite ?
I think we want to skip `lengthOfIntersect == length()` case here

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
```diff
@@ -122,7 +123,17 @@ public double overlapPercentWith(StatisticRange other)
         if (isInfinite(length()) && isFinite(lengthOfIntersect)) {
             return INFINITE_TO_FINITE_RANGE_INTERSECT_OVERLAP_HEURISTIC_FACTOR;
         }
+
         if (lengthOfIntersect > 0) {
+            double thisDensity = this.distinctValues / length();
+            double otherDensity = other.distinctValues / other.length();
+            double minDensity = minExcludeNaN(thisDensity, otherDensity);
```
findepi: `ExcludeNaN` is redundant (we guard against this and other density being nan) but still is a cognitive overhead. I'd inline `minDensity` variable and remove `ExcludeNaN` from it's definition.

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
findepi: Add a comment explaining _why_ this particular logic is here.
In particular, a future person working on this code should understand what would break if they delete these lines. Not that people delete random lines -- however. sometimes new code lines cause new problems and removing or changing them is always a possibility. It must be understandable what circumstances we should be concerned about when doing so.

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
...
+    }
+
+    @Test
+    public void testDensityThresholdBoundary()
+    {
+        StatisticRange boundaryRange = range(0, 10000, 10);
+        StatisticRange smallFilter = range(0, 100, 5);
+
+        double overlap = boundaryRange.overlapPercentWith(smallFilter);
+        assertThat(overlap).isBetween(0.01, 0.5);
```
findepi: that's very wide

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
         assertOverlap(unboundedRange(0.0), unboundedRange(0), 0);
     }
 
+    @Test
+    public void testLowDensityOverlap()
+    {
+        StatisticRange sparseRange = range(1, 3662098119.0, 14);
+        StatisticRange filterRange = range(1, 4, 4);
+
+        double expectedOverlap = 4.0 / 14.0;
```
findepi: inline?

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
...
+        assertOverlap(sparseRange, filterRange, expectedOverlap);
+    }
+
+    @Test
+    public void testDensityThresholdBoundary()
+    {
+        StatisticRange boundaryRange = range(0, 10000, 10);
+        StatisticRange smallFilter = range(0, 100, 5);
+
+        double overlap = boundaryRange.overlapPercentWith(smallFilter);
```
findepi: I'd inline this variable and add line break before `.isBetween`

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
...
+    }
+
+    @Test
+    public void testVeryLowDensity()
+    {
+        StatisticRange verySparse = range(0, 1e9, 10);
+        StatisticRange filterRange = range(100, 200, 5);
+
+        double expected = 5.0 / 10.0;
+        double actual = verySparse.overlapPercentWith(filterRange);
```
findepi: I'd inline this variable and add line break before `.isCloseTo`

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
...
+        assertOverlap(denseRange, filterRange, 0.1);
+    }
+
+    @Test
+    public void testVeryLowDensity()
+    {
+        StatisticRange verySparse = range(0, 1e9, 10);
+        StatisticRange filterRange = range(100, 200, 5);
+
+        double expected = 5.0 / 10.0;
```
findepi: inline?

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
...
+        double actual = verySparse.overlapPercentWith(filterRange);
+        assertThat(actual).isCloseTo(expected, within(0.1));
```
findepi: 0.1 is a lot error margin given the expected is 0.5. 
could be 0.0001

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
```diff
@@ -59,6 +60,55 @@ public void testOverlapPercentWith()
         assertOverlap(unboundedRange(0.0), unboundedRange(0), 0);
     }
 
+    @Test
```
findepi, line 62: i love unit tests... except they actually don't really tell the story
sometimes it's just testing that a particular math formula produces a particular result... which feels like testing Java itself.

When it's not obvious _why_ we expect given result, then it's not clear to do what to do when a test fails on a change. Oftentimes, updating the test is there's needed (formula changed ⇒ result change ⇒ test expected value changed). 
It would be more interesting to see a test where expected value is harder to dispute. `TestTpcdsLocalStats` is an example of such test -- it simply tests that estimated row count matches  reality. The reality cannot be disputed and estimate being close to it is generally desired property of the estimate.

findepi:
Do i understand correctly that in OP context, for a `platform_id = 1 OR platform_id = 2 OR platform_id = 3 OR platform_id = 4` filter we would produce a good estimate, and for `platform_id IN (1, 2, 3, 4)` we produce a poor one?

raunaqmorarka:
> Do i understand correctly that in OP context, for a `platform_id = 1 OR platform_id = 2 OR platform_id = 3 OR platform_id = 4` filter we would produce a good estimate, and for `platform_id IN (1, 2, 3, 4)` we produce a poor one?

Those OR would get simplified to an IN. The problem here is from another optimization where we simplify filters like that to a BETWEEN 1 and 4 filter for better efficiency (SimplifyContinuousInValues), and that changes the estimate.

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
anton-kutuzov: I think that we cannot use:
`min(other.distinctValues / this.distinctValues, 1)`
because in cases like `IN (1, 2, 3, 4)`, the distinctValues for other is `NaN`. Then:
`min(other.distinctValues / this.distinctValues, 1) = min(NaN, 1) = NaN`
so the result would be `NaN`, which is not valid.

Instead, we should use:
`minExcludeNaN(this.distinctValues, other.distinctValues) / this.distinctValues`
or equivalently:
`minExcludeNaN(other.distinctValues / this.distinctValues, 1)`

---

Also, I looked more carefully at the idea of removing the `weight` I added before. If we do that, the estimate will go from about `0.29` to `1`, `because other.distinctValues` is `NaN`

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
anton-kutuzov: Added one more test to `TestFilterStatsCalculator`

core/trino-main/src/test/java/io/trino/cost/TestStatisticRange.java
anton-kutuzov: Thanks, simplified

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
anton-kutuzov: Done

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
anton-kutuzov: Thanks, simplified

core/trino-main/src/main/java/io/trino/cost/StatisticRange.java
anton-kutuzov: Thanks, fixed

anton-kutuzov:
> @anton-kutuzov please squash your commits

Sorry, I forget to do it. Done

PR #27574, opened 2025-12-08
Various cleanups and improvements in TypeParameter

martint:
## Description

See individual commit details.

## Release notes

```markdown
## SPI
* Remove support for `TypeSignatureParameter`. Use `TypeParameter`, instead. ({issue}`27574`)
* Remove support for `ParameterKind`. Use `TypeParameter.Type`, `TypeParameter.Numeric`, `TypeParameter.Variable`, instead. ({issue}`27574`)
* Remove support for `NamedType`, `NamedTypeSignature` and `NamedTypeParameter`. Use `TypeParameter.Type`, instead. ({issue}`27574`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
+++ b/core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
@@ -52,6 +52,7 @@ import io.trino.sql.tree.RowDataType;
 
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -149,20 +150,18 @@ public final class ProtocolUtil
 
     private static ClientTypeSignatureParameter toClientTypeSignatureParameter(String base, TypeParameter parameter, boolean supportsParametricDateTime)
     {
-        switch (parameter.getKind()) {
-            case TYPE:
+        return switch (parameter) {
+            case TypeParameter.Type(Optional<String> name, TypeSignature type) -> {
                 if (base.equalsIgnoreCase(ROW)) { // for backward compatibility with old clients, which expect NAMED_TYPE for row fields
-                    return ClientTypeSignatureParameter.ofNamedType(new NamedClientTypeSignature(
-                            parameter.name().map(RowFieldName::new),
-                            toClientTypeSignature(parameter.getTypeSignature(), supportsParametricDateTime)));
+                    yield ClientTypeSignatureParameter.ofNamedType(new NamedClientTypeSignature(
+                            name.map(RowFieldName::new),
+                            toClientTypeSignature(type, supportsParametricDateTime)));
                 }
-                return ClientTypeSignatureParameter.ofType(toClientTypeSignature(parameter.getTypeSignature(), supportsParametricDateTime));
-            case LONG:
-                return ClientTypeSignatureParameter.ofLong(parameter.getLongLiteral());
-            case VARIABLE:
-                // not expected here
-        }
-        throw new IllegalArgumentException("Unsupported kind: " + parameter.getKind());
+                yield ClientTypeSignatureParameter.ofType(toClientTypeSignature(type, supportsParametricDateTime));
+            }
+            case TypeParameter.Numeric number -> ClientTypeSignatureParameter.ofLong(number.value());
+            case TypeParameter.Variable _ -> throw new IllegalArgumentException("Unsupported parameter kind: " + parameter);
+        };
     }
 
     public static StatementStats toStatementStats(ResultQueryInfo queryInfo)
--- a/core/trino-main/src/main/java/io/trino/type/ArrayParametricType.java
+++ b/core/trino-main/src/main/java/io/trino/type/ArrayParametricType.java
@@ -14,12 +14,12 @@
 package io.trino.type;
 
 import io.trino.spi.type.ArrayType;
-import io.trino.spi.type.ParameterKind;
 import io.trino.spi.type.ParametricType;
 import io.trino.spi.type.StandardTypes;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.TypeManager;
 import io.trino.spi.type.TypeParameter;
+import io.trino.spi.type.TypeSignature;
 
 import java.util.List;
 
@@ -42,10 +42,10 @@ public final class ArrayParametricType
     public Type createType(TypeManager typeManager, List<TypeParameter> parameters)
     {
         checkArgument(parameters.size() == 1, "Array type expects exactly one type as a parameter, got %s", parameters);
-        checkArgument(
-                parameters.get(0).getKind() == ParameterKind.TYPE,
-                "Array expects type as a parameter, got %s",
-                parameters);
-        return new ArrayType(typeManager.getType(parameters.get(0).getTypeSignature()));
+
+        if (parameters.get(0) instanceof TypeParameter.Type(_, TypeSignature type)) {
+            return new ArrayType(typeManager.getType(type));
+        }
+        throw new IllegalArgumentException("Array expects type as a parameter, got " + parameters);
     }
 }
--- a/core/trino-main/src/main/java/io/trino/type/RowParametricType.java
+++ b/core/trino-main/src/main/java/io/trino/type/RowParametricType.java
@@ -13,7 +13,6 @@
  */
 package io.trino.type;
 
-import io.trino.spi.type.ParameterKind;
 import io.trino.spi.type.ParametricType;
 import io.trino.spi.type.RowType;
 import io.trino.spi.type.StandardTypes;
@@ -43,13 +42,11 @@ public final class RowParametricType
     public Type createType(TypeManager typeManager, List<TypeParameter> parameters)
     {
         checkArgument(!parameters.isEmpty(), "Row type must have at least one parameter");
-        checkArgument(
-                parameters.stream().allMatch(parameter -> parameter.getKind() == ParameterKind.TYPE),
-                "Expected only types as a parameters, got %s",
-                parameters);
-
         List<RowType.Field> fields = parameters.stream()
-                .map(parameter -> new RowType.Field(parameter.name(), typeManager.getType(parameter.getTypeSignature())))
+                .map(parameter -> switch (parameter) {
+                    case TypeParameter.Type type -> new RowType.Field(type.name(), typeManager.getType(type.type()));
+                    default -> throw new IllegalArgumentException("Expected only types as parameters, got " + parameter);
+                })
                 .collect(toList());
 
         return RowType.from(fields);
--- a/core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
@@ -15,161 +15,120 @@ package io.trino.spi.type;
 
 import com.google.errorprone.annotations.Immutable;
 
-import java.util.Objects;
 import java.util.Optional;
 
 import static java.lang.String.format;
-import static java.util.Objects.requireNonNull;
 
 @Immutable
-public final class TypeParameter
+public sealed interface TypeParameter
+        permits TypeParameter.Type, TypeParameter.Numeric, TypeParameter.Variable
 {
-    private final Optional<String> name;
-    private final ParameterKind kind;
-    private final Object value;
-
-    public static TypeParameter typeParameter(TypeSignature typeSignature)
+    static TypeParameter typeParameter(TypeSignature typeSignature)
     {
-        return new TypeParameter(Optional.empty(), ParameterKind.TYPE, typeSignature);
+        return new Type(Optional.empty(), typeSignature);
     }
 
-    public static TypeParameter typeParameter(Optional<String> name, TypeSignature typeSignature)
+    static TypeParameter typeParameter(Optional<String> name, TypeSignature typeSignature)
     {
-        return new TypeParameter(name, ParameterKind.TYPE, typeSignature);
+        return new Type(name, typeSignature);
     }
 
-    public static TypeParameter numericParameter(long longLiteral)
+    static TypeParameter numericParameter(long longLiteral)
     {
-        return new TypeParameter(Optional.empty(), ParameterKind.LONG, longLiteral);
+        return new Numeric(longLiteral);
     }
 
-    public static TypeParameter namedField(String name, TypeSignature type)
+    static TypeParameter namedField(String name, TypeSignature type)
     {
-        return new TypeParameter(Optional.of(name), ParameterKind.TYPE, type);
+        return new Type(Optional.of(name), type);
     }
 
-    public static TypeParameter anonymousField(TypeSignature type)
+    static TypeParameter anonymousField(TypeSignature type)
     {
-        return new TypeParameter(Optional.empty(), ParameterKind.TYPE, type);
+        return new Type(Optional.empty(), type);
     }
 
-    public static TypeParameter typeVariable(String variable)
+    static TypeParameter typeVariable(String variable)
     {
-        return new TypeParameter(Optional.empty(), ParameterKind.VARIABLE, variable);
+        return new Variable(variable);
     }
 
-    private TypeParameter(Optional<String> name, ParameterKind kind, Object value)
-    {
-        this.name = requireNonNull(name, "name is null");
-        this.kind = requireNonNull(kind, "kind is null");
-        this.value = requireNonNull(value, "value is null");
-    }
+    String jsonValue();
 
-    public Optional<String> name()
-    {
-        return name;
-    }
+    boolean isCalculated();
 
-    @Override
-    public String toString()
+    record Type(Optional<String> name, TypeSignature type)
+            implements TypeParameter
     {
-        return switch (kind) {
-            case VARIABLE, LONG -> value.toString();
-            case TYPE -> {
-                if (name.isPresent()) {
-                    yield format("\"%s\" %s", name.get().replace("\"", "\"\""), value.toString());
-                }
-                yield value.toString();
+        @Override
+        public String toString()
+        {
+            if (name.isPresent()) {
+                return format("\"%s\" %s", name.get().replace("\"", "\"\""), type.toString());
             }
-        };
-    }
-
-    public String jsonValue()
-    {
-        String prefix = "";
-
-        if (name.isPresent()) {
-            prefix = format("\"%s\" ", name.get().replace("\"", "\"\""));
+            return type.toString();
         }
 
-        if (kind == ParameterKind.VARIABLE) {
-            prefix += "@";
+        @Override
+        public String jsonValue()
+        {
+            String prefix = "";
+
+            if (name.isPresent()) {
+                prefix = format("\"%s\" ", name.get().replace("\"", "\"\""));
+            }
+
+            return prefix + type.jsonValue();
         }
 
-        String valueJson;
-        if (value instanceof TypeSignature typeSignature) {
-            valueJson = typeSignature.jsonValue();
+        @Override
+        public boolean isCalculated()
+        {
+            return type.isCalculated();
         }
-        else {
-            valueJson = value.toString();
+    }
+
+    record Numeric(long value)
+            implements TypeParameter
+    {
+        @Override
+        public String toString()
+        {
+            return Long.toString(value);
         }
-        return prefix + valueJson;
-    }
 
-    public ParameterKind getKind()
-    {
-        return kind;
-    }
-
-    public boolean isTypeSignature()
-    {
-        return kind == ParameterKind.TYPE;
-    }
-
-    public boolean isLongLiteral()
-    {
-        return kind == ParameterKind.LONG;
-    }
-
-    public boolean isVariable()
-    {
-        return kind == ParameterKind.VARIABLE;
-    }
-
-    private <A> A getValue(ParameterKind expectedParameterKind, Class<A> target)
-    {
-        if (kind != expectedParameterKind) {
-            throw new IllegalArgumentException(format("ParameterKind is [%s] but expected [%s]", kind, expectedParameterKind));
+        @Override
+        public String jsonValue()
+        {
+            return toString();
         }
-        return target.cast(value);
-    }
 
-    public TypeSignature getTypeSignature()
-    {
-        return getValue(ParameterKind.TYPE, TypeSignature.class);
-    }
-
-    public Long getLongLiteral()
-    {
-        return getValue(ParameterKind.LONG, Long.class);
-    }
-
-    public String getVariable()
-    {
-        return getValue(ParameterKind.VARIABLE, String.class);
-    }
-
-    public boolean isCalculated()
-    {
-        return switch (kind) {
-            case TYPE -> getTypeSignature().isCalculated();
-            case LONG -> false;
-            case VARIABLE -> true;
-        };
-    }
-
-    @Override
-    public boolean equals(Object o)
-    {
-        if (!(o instanceof TypeParameter that)) {
+        @Override
+        public boolean isCalculated()
+        {
             return false;
         }
-        return Objects.equals(name, that.name) && kind == that.kind && Objects.equals(value, that.value);
     }
 
-    @Override
-    public int hashCode()
+    record Variable(String name)
+            implements TypeParameter
     {
-        return Objects.hash(name, kind, value);
+        @Override
+        public String toString()
+        {
+            return name;
+        }
+
+        @Override
+        public String jsonValue()
+        {
+            return "@" + name;
+        }
+
+        @Override
+        public boolean isCalculated()
+        {
+            return true;
+        }
     }
 }
--- a/lib/trino-hive-formats/src/main/java/io/trino/hive/formats/avro/AvroPageDataReader.java
+++ b/lib/trino-hive-formats/src/main/java/io/trino/hive/formats/avro/AvroPageDataReader.java
@@ -46,13 +46,12 @@ public class AvroPageDataReader
         verifyNoCircularReferences(readerSchema);
         try {
             Type readerSchemaType = typeManager.typeFor(readerSchema);
-            if (readerSchemaType instanceof RowType rowType) {
-                pageBuilder = new PageBuilder(rowType.getFieldTypes());
-                initialize();
-            }
-            else {
+            if (!(readerSchemaType instanceof RowType rowType)) {
                 throw new AvroTypeException("Root Avro type must be a row");
             }
+
+            pageBuilder = new PageBuilder(rowType.getFieldTypes());
+            initialize();
         }
         catch (org.apache.avro.AvroTypeException e) {
             throw new AvroTypeException(e);
--- a/lib/trino-hive-formats/src/test/java/io/trino/hive/formats/avro/TestAvroPageDataReaderWithAvroNativeTypeManagement.java
+++ b/lib/trino-hive-formats/src/test/java/io/trino/hive/formats/avro/TestAvroPageDataReaderWithAvroNativeTypeManagement.java
@@ -47,7 +47,6 @@ import java.io.IOException;
 import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.util.Date;
-import java.util.List;
 import java.util.UUID;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
@@ -246,7 +245,7 @@ public class TestAvroPageDataReaderWithAvroNativeTypeManagement
             throws IOException, AvroTypeException
     {
         Location testLocation = createLocalTempLocation();
-        Type type = new NativeLogicalTypesAvroTypeBlockHandler().typeFor(ALL_SUPPORTED_TYPES_SCHEMA);
+        RowType type = (RowType) new NativeLogicalTypesAvroTypeBlockHandler().typeFor(ALL_SUPPORTED_TYPES_SCHEMA);
         try (AvroFileWriter fileWriter = new AvroFileWriter(
                 trinoLocalFilesystem.newOutputFile(testLocation).create(),
                 ALL_SUPPORTED_TYPES_SCHEMA,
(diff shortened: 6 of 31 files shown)


core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
```diff
@@ -92,84 +90,84 @@ public String jsonValue()
...
-    public boolean isLongLiteral()
+    @Override
+    public int hashCode()
     {
-        return kind == ParameterKind.LONG;
+        return Objects.hash(name, value);
     }
 
-    public boolean isVariable()
+    public sealed interface Value
```
martint: Still deciding whether TypeParameter should be a holder of `Optional<String>` (for the field name) and Value (for the parameter value), or whether to flatten the hierarchy and make TypeParameter be the interface. In the latter case, every implementation type would need to hold/be able to return the field name.

martint: Going with the flattened design, except that only the type-containing subtype can carry a name.

core/trino-main/src/main/java/io/trino/operator/scalar/Re2JCastToRegexpFunction.java
```diff
@@ -73,12 +75,18 @@ private Re2JCastToRegexpFunction(String sourceType, int dfaStatesLimit, int dfaR
     protected SpecializedSqlScalarFunction specialize(BoundSignature boundSignature)
     {
         Type inputType = boundSignature.getArgumentType(0);
-        Long typeLength = inputType.getTypeSignature().getParameters().get(0).getLongLiteral();
+
+        long length = switch (inputType) {
+            case CharType charType -> charType.getLength();
+            case VarcharType varcharType -> varcharType.getLength().orElse(VarcharType.MAX_LENGTH);
+            default -> throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Type " + inputType + " is not supported for cast to regexp");
```
findepi: `"...".formatted`

core/trino-main/src/main/java/io/trino/operator/scalar/Re2JCastToRegexpFunction.java
```diff
@@ -73,12 +75,18 @@ private Re2JCastToRegexpFunction(String sourceType, int dfaStatesLimit, int dfaR
     protected SpecializedSqlScalarFunction specialize(BoundSignature boundSignature)
     {
         Type inputType = boundSignature.getArgumentType(0);
-        Long typeLength = inputType.getTypeSignature().getParameters().get(0).getLongLiteral();
```
findepi, line 76: This didn't work for unbounded varchar, did it?

core/trino-main/src/main/java/io/trino/operator/scalar/Re2JCastToRegexpFunction.java
```diff
@@ -73,12 +75,18 @@ private Re2JCastToRegexpFunction(String sourceType, int dfaStatesLimit, int dfaR
...
+
+        long length = switch (inputType) {
```
findepi: `int` would be more appropriate.

core/trino-main/src/main/java/io/trino/operator/scalar/Re2JCastToRegexpFunction.java
```diff
@@ -73,12 +75,18 @@ private Re2JCastToRegexpFunction(String sourceType, int dfaStatesLimit, int dfaR
...
+            case VarcharType varcharType -> varcharType.getLength().orElse(VarcharType.MAX_LENGTH);
+            default -> throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Type " + inputType + " is not supported for cast to regexp");
+        };
+
         return new ChoicesSpecializedSqlScalarFunction(
                 boundSignature,
                 FAIL_ON_NULL,
                 ImmutableList.of(NEVER_NULL),
-                insertArguments(METHOD_HANDLE, 0, dfaStatesLimit, dfaRetries, padSpaces, typeLength));
+                insertArguments(METHOD_HANDLE, 0, dfaStatesLimit, dfaRetries, padSpaces, length));
```
findepi, line 89: I noticed the length matters only when padSpaces is true,  so we could replace `boolean padSpaces, long typeLength` with `OptionalInt padToLength`

no change requested here

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaHiveTypeTranslator.java
```diff
@@ -145,13 +144,13 @@ public static TypeInfo translate(Type type)
...
         }
-        if (type instanceof RowType) {
+        if (type instanceof RowType rowType) {
             ImmutableList.Builder<String> fieldNames = ImmutableList.builder();
-            for (TypeSignatureParameter parameter : type.getTypeSignature().getParameters()) {
-                if (!parameter.isNamedTypeSignature()) {
-                    throw new IllegalArgumentException(format("Expected all parameters to be named type, but got %s", parameter));
+            for (RowType.Field field : rowType.getFields()) {
+                if (field.getName().isEmpty()) {
+                    throw new TrinoException(NOT_SUPPORTED, format("Anonymous row type is not supported in Hive. Please give each field a name: %s", type));
```
findepi: Pre-existing, but this `throw` is redundant.
the same is raised from `field.getName().orElseThrow` in next line

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/HiveTypeTranslator.java
```diff
@@ -158,13 +158,13 @@ private static TypeInfo toTypeInfo(Type type)
...
         }
-        if (type instanceof RowType) {
+        if (type instanceof RowType rowType) {
             ImmutableList.Builder<String> fieldNames = ImmutableList.builder();
-            for (TypeSignatureParameter parameter : type.getTypeSignature().getParameters()) {
-                if (!parameter.isNamedTypeSignature()) {
-                    throw new IllegalArgumentException(format("Expected all parameters to be named type, but got %s", parameter));
+            for (RowType.Field field : rowType.getFields()) {
+                if (field.getName().isEmpty()) {
+                    throw new TrinoException(NOT_SUPPORTED, format("Anonymous row type is not supported in Hive. Please give each field a name: %s", type));
```
findepi: Pre-existing, but this `throw` is redundant.
the same is raised from `field.getName().orElseThrow` in next line

findepi (commented):
`Get type arguments directly from the type` LGTM

core/trino-main/src/main/java/io/trino/type/ArrayParametricType.java
```diff
@@ -39,13 +39,13 @@ public String getName()
     }
 
     @Override
-    public Type createType(TypeManager typeManager, List<TypeParameter> parameters)
```
findepi, line 42: The `TypeParameter` has the advantage that it's "resolved" version of the `TypeSignatureParameter`. From API perspective this might be better: the `ParametricType#createType` implementation gets something that's more processed and ready to use

core/trino-main/src/main/java/io/trino/type/RowParametricType.java
```diff
@@ -44,24 +40,19 @@ public String getName()
...
         checkArgument(!parameters.isEmpty(), "Row type must have at least one parameter");
         checkArgument(
                 parameters.stream().allMatch(parameter -> parameter.getKind() == ParameterKind.NAMED_TYPE),
                 "Expected only named types as a parameters, got %s",
                 parameters);
 
-        List<TypeSignatureParameter> typeSignatureParameters = parameters.stream()
-                .map(TypeParameter::getNamedType)
-                .map(parameter -> TypeSignatureParameter.namedTypeParameter(new NamedTypeSignature(parameter.getName(), parameter.getType().getTypeSignature())))
-                .collect(toList());
```
findepi, line 58: There and back again :)

core/trino-main/src/main/java/io/trino/type/RowParametricType.java
```diff
@@ -44,24 +40,19 @@ public String getName()
...
-
         List<RowType.Field> fields = parameters.stream()
-                .map(TypeParameter::getNamedType)
-                .map(parameter -> new RowType.Field(parameter.getName().map(RowFieldName::getName), parameter.getType()))
+                .map(TypeSignatureParameter::getNamedTypeSignature)
```
findepi: can it be just type signature parameter of kind TYPE (anonymous fields)

findepi (commented):
`Rename TypeSignatureParameter -> TypeParameter` LGTM assuming it's mechanical rename

core/trino-main/src/test/java/io/trino/operator/scalar/CustomFunctions.java
```diff
@@ -67,7 +67,7 @@ public static Slice varchar(ConnectorSession session, @SqlType(StandardTypes.VAR
 
     @ScalarFunction("connector_session_row")
     @SqlType(StandardTypes.VARCHAR)
-    public static Slice row(ConnectorSession session, @TypeParameter("row(greeting varchar)") RowType rowType, @SqlType("row(varchar)") SqlRow sqlRow)
+    public static Slice row(ConnectorSession session, @TypeParameter("row(greeting varchar)") RowType rowType, @SqlType("row(greeting varchar)") SqlRow sqlRow)
```
findepi, line 70: > The type parameter must match the argument it corresponds to.

is the validation for this?
do we need this duplication?

findepi, line 70: The commit title `Fix incorrect type declaration` could reveal that it's a very innocent change. E.g. add "in test"

core/trino-main/src/main/java/io/trino/type/ArrayParametricType.java
martint: Yes, I'm aware, but this is a very minor advantage. Besides, the method already takes a TypeManager, whose purpose is to resolve types.

core/trino-main/src/main/java/io/trino/type/RowParametricType.java
martint: Yes, but not in this commit. RowType only handles NamedType, which can have a name or be anonymous. Every other type (map, array, etc), takes either Type or Long. That's what I clean up in subsequent commits.

core/trino-main/src/test/java/io/trino/operator/scalar/CustomFunctions.java
martint: There doesn't seem to be any validation for this, but it starts failing with some of the changes I make in other commits.

The duplication is needed because it's what links the two arguments together. The SqlRow refers to the value, while the RowType refers to the reified type of the value passed as a SqlRow.

core/trino-main/src/main/java/io/trino/operator/scalar/Re2JCastToRegexpFunction.java
martint: It did. `varchar` gets a length of Integer.MAX_VALUE in boundSignature (which is technically wrong, but that's a different matter).

core/trino-main/src/main/java/io/trino/operator/scalar/Re2JCastToRegexpFunction.java
martint: Makes sense, but I'll leave that to a separate change.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaHiveTypeTranslator.java
martint: Removing

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -704,9 +704,9 @@ private static Object convertFromClientRepresentation(ClientTypeSignature column
                 for (int i = 0; i < fields.size(); i++) {
                     io.trino.client.RowField field = fields.get(i);
                     ClientTypeSignatureParameter clientTypeSignatureParameter = typeArguments.get(i);
-                    verify(clientTypeSignatureParameter.getKind() == ClientTypeSignatureParameter.ParameterKind.NAMED_TYPE, "Not a NAMED_TYPE: %s", clientTypeSignatureParameter);
-                    verify(field.getName().equals(clientTypeSignatureParameter.getNamedTypeSignature().getName()), "Name mismatch: %s, %s", field, clientTypeSignatureParameter);
-                    Object converted = convertFromClientRepresentation(clientTypeSignatureParameter.getNamedTypeSignature().getTypeSignature(), field.getValue());
+                    verify(clientTypeSignatureParameter.getKind() == ClientTypeSignatureParameter.ParameterKind.NAMED_TYPE, "Not a NAMED_TYPE or TYPE: %s", clientTypeSignatureParameter);
```
findepi: The condition and message are off

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
```diff
@@ -143,20 +143,20 @@ private static ClientTypeSignature toClientTypeSignature(TypeSignature signature
...
+                .map(parameter -> toClientTypeSignatureParameter(signature.getBase(), parameter, supportsParametricDateTime))
                 .collect(toImmutableList()));
     }
 
-    private static ClientTypeSignatureParameter toClientTypeSignatureParameter(TypeParameter parameter, boolean supportsParametricDateTime)
+    private static ClientTypeSignatureParameter toClientTypeSignatureParameter(String base, TypeParameter parameter, boolean supportsParametricDateTime)
     {
         switch (parameter.getKind()) {
             case TYPE:
+                if (base.equalsIgnoreCase(ROW)) { // for backward compatibility with old clients,which expect NAMED_TYPE for row fields
```
findepi: space after `,`

core/trino-main/src/main/java/io/trino/type/RowParametricType.java
```diff
@@ -44,13 +44,12 @@ public Type createType(TypeManager typeManager, List<TypeParameter> parameters)
     {
         checkArgument(!parameters.isEmpty(), "Row type must have at least one parameter");
         checkArgument(
-                parameters.stream().allMatch(parameter -> parameter.getKind() == ParameterKind.NAMED_TYPE),
+                parameters.stream().allMatch(parameter -> parameter.getKind() == ParameterKind.TYPE),
                 "Expected only named types as a parameters, got %s",
```
findepi: with the condition change,  this message should also change

core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
```diff
@@ -24,56 +24,76 @@
...
 public final class TypeParameter
 {
+    private final Optional<String> name;
     private final ParameterKind kind;
     private final Object value;
 
     public static TypeParameter typeParameter(TypeSignature typeSignature)
     {
-        return new TypeParameter(ParameterKind.TYPE, typeSignature);
+        return new TypeParameter(Optional.empty(), ParameterKind.TYPE, typeSignature);
```
findepi: in a way this means TYPE is now NAMED_TYPE :)

core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
```diff
@@ -24,56 +24,76 @@
...
     {
         String prefix = "";
+
+        if (name.isPresent()) {
+            prefix = format("\"%s\" ", name.get().replace("\"", "\"\""));
+        }
+
         if (kind == ParameterKind.VARIABLE) {
-            prefix = "@";
+            prefix += "@";
```
findepi: aren't `name.isPresent()` and `kind == ParameterKind.VARIABLE` exclusive conditions?
if so, better to have a check then lenient `+=`

plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveTestUtils.java
```diff
@@ -208,13 +207,15 @@ public static MapType mapType(Type keyType, Type valueType)
                 TypeParameter.typeParameter(valueType.getTypeSignature())));
     }
 
-    public static RowType rowType(List<NamedTypeSignature> elementTypeSignatures)
+    public static RowType rowType(List<TestHiveReaderProjectionsUtil.Field> elementTypeSignatures)
```
findepi: This import looks wrong.
HiveTestUtils definitely doesn't want to depend on a particular test class

plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveTestUtils.java
```diff
@@ -208,13 +207,15 @@ public static MapType mapType(Type keyType, Type valueType)
...
     {
         return (RowType) TESTING_TYPE_MANAGER.getParameterizedType(
                 StandardTypes.ROW,
                 elementTypeSignatures.stream()
-                        .map(TypeParameter::namedTypeParameter)
-                        .collect(toImmutableList()));
+                        .map(field -> TypeParameter.typeParameter(
+                                        Optional.of(field.name()),
+                                        field.type()))
+                .collect(toImmutableList()));
```
findepi: indent is off

core/trino-main/src/main/java/io/trino/sql/gen/columnar/InColumnarFilterGenerator.java
```diff
@@ -339,7 +342,7 @@ static boolean useSwitchCaseGeneration(Type type, List<RowExpression> values)
         // https://github.com/trinodb/trino/issues/17213
         // Until we support HASH_SWITCH strategy for code generation here, we treat structural type as an unsupported case
         // and fall back to existing expression evaluator for small lists
-        if (!type.getTypeParameters().isEmpty()) {
+        if (type instanceof ArrayType || type instanceof MapType || type instanceof RowType) {
```
findepi, line 345: - `ParametricType` is in SPI so one could imagine a plugin adding a structural type. The new check doesn't reject this.
  - in fact, `ClassifierType` is an example for which the new check returns different result

core/trino-spi/src/main/java/io/trino/spi/type/Type.java
```diff
@@ -89,7 +89,12 @@ default TypeOperatorDeclaration getTypeOperatorDeclaration(TypeOperators typeOpe
 
     /**
      * For parameterized types returns the list of parameters.
+     *
+     * @deprecated Use type-specific methods to access type parameters (e.g.,
+     *             {@link ArrayType#getElementType()}, {@link MapType#getKeyType()},
+     *             {@link MapType#getValueType()}, {@link RowType#getFields()})
```
findepi, line 95: in general i do agree. it's much more elegant

however this doesn't work when dealing with other-plugin-provided structural type

martint: There are very few places that need to operate on type arguments generically. Except for the type inferencer, which is based on the more generic "type signature" based representation, every other use case needs to be aware of the type they're dealing with.

core/trino-main/src/main/java/io/trino/sql/gen/columnar/InColumnarFilterGenerator.java
martint: Agreed. However, that's only a theoretical concern right now. There's also a difference from a structural type (i.e., a container such as map/array/row) and a specialized type for type-safety and static type validations.

core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
martint: Correct. It unifies all uses to TYPE with NAMED_TYPE

martint:
> i'd suggest separate PR.

Happy to do that. However, since all these commits affect the SPI, I wanted to get them all in the same timeframe to avoid thrash across multiple releases.

core/trino-spi/src/main/java/io/trino/spi/type/TypeParameter.java
martint: In theory, it would be possible to have a named field with an unbound type.

martint:
> many of these refactors looks good on their own, but it's hard to review all of them at once

They are meant to be reviewed individually. They are all logically independent from each other, although they build on/touch the same areas the previous commit touched.

core/trino-spi/src/main/java/io/trino/spi/type/Type.java
findepi: I'm ok with deprecation, but we may also undeprecated this at some point.
For example, if myplugin provides MyArrayType and your connector wants to support it, you will probably have to use this method.

i know the example is rather unlikely and deprecating this method has benefits, so let's do so now

core/trino-main/src/main/java/io/trino/sql/gen/columnar/InColumnarFilterGenerator.java
findepi: in fact, i don't know why we throw for some types, and return false for some other types.
it's looks like throwing isn't necessary in this method at all, because the final type validation/rejection needs to be done somewhere later ... and therefore it doesn't matter much what we do here. the new validation is weaker (more persmissive), and therefore likely incorrect, but it's whatever

findepi:
> > many of these refactors looks good on their own, but it's hard to review all of them at once
> 
> They are meant to be reviewed individually.

fair & my understanding as well. just the diff is longish

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/avro/AvroPageDataReader.java
```diff
@@ -46,9 +46,13 @@ public AvroPageDataReader(Schema readerSchema, AvroTypeBlockHandler typeManager)
...
-            verify(readerSchemaType instanceof RowType, "Root Avro type must be a row");
-            pageBuilder = new PageBuilder(readerSchemaType.getTypeParameters());
-            initialize();
+            if (readerSchemaType instanceof RowType rowType) {
+                pageBuilder = new PageBuilder(rowType.getFieldTypes());
+                initialize();
+            }
+            else {
+                throw new AvroTypeException("Root Avro type must be a row");
+            }
```
findepi: reverse the if condition so that throw comes close to it, and reduce nesting

lib/trino-hive-formats/src/test/java/io/trino/hive/formats/FormatTestUtils.java
```diff
@@ -281,16 +281,21 @@ private static SqlTimestamp decodeRecordReaderTimestamp(TimestampType timestampT
 
     private static List<Object> decodeRecordReaderList(Type type, List<?> list, Optional<DateTimeZone> hiveStorageTimeZone)
     {
-        Type elementType = type.getTypeParameters().get(0);
+        Type elementType = switch (type) {
+            case RowType rowType -> rowType.getFields().getFirst().getType();
+            case ArrayType arrayType -> arrayType.getElementType();
+            default -> throw new IllegalArgumentException("Unexpected type: " + type);
```
findepi, line 287: - it's weird to take just a first field of a row. worth code comment
- the code also worked for map type which is now unhandled. i understand map never lands in this (test) code?

lib/trino-hive-formats/src/test/java/io/trino/hive/formats/avro/TestAvroPageDataReaderWithAvroNativeTypeManagement.java
```diff
@@ -244,14 +246,16 @@ public void testWriting()
...
+        Type type = new NativeLogicalTypesAvroTypeBlockHandler().typeFor(ALL_SUPPORTED_TYPES_SCHEMA);
         try (AvroFileWriter fileWriter = new AvroFileWriter(
                 trinoLocalFilesystem.newOutputFile(testLocation).create(),
                 ALL_SUPPORTED_TYPES_SCHEMA,
                 new NativeLogicalTypesAvroTypeManager(),
                 AvroCompressionKind.NULL,
                 ImmutableMap.of(),
                 ALL_SUPPORTED_TYPES_SCHEMA.getFields().stream().map(Schema.Field::name).collect(toImmutableList()),
-                new NativeLogicalTypesAvroTypeBlockHandler().typeFor(ALL_SUPPORTED_TYPES_SCHEMA).getTypeParameters(), false)) {
+                type instanceof RowType rowType ? rowType.getFieldTypes() : List.of(),
```
findepi: the `?` is weird. the `type` is effectively constant, so it's always either row or not (i suppose it's always a row)

lib/trino-hive-formats/src/test/java/io/trino/hive/formats/avro/TestAvroPageDataWriterWithoutTypeManager.java
```diff
@@ -68,14 +69,16 @@ private void testAllTypesWriting(Schema writeSchema)
...
+        Type type = new BaseAvroTypeBlockHandler().typeFor(ALL_TYPES_RECORD_SCHEMA);
         try (AvroFileWriter fileWriter = new AvroFileWriter(
                 trinoLocalFilesystem.newOutputFile(tempTestLocation).create(),
                 writeSchema,
                 NoOpAvroTypeManager.INSTANCE,
                 AvroCompressionKind.NULL,
                 ImmutableMap.of(),
                 ALL_TYPES_RECORD_SCHEMA.getFields().stream().map(Schema.Field::name).collect(toImmutableList()),
-                new BaseAvroTypeBlockHandler().typeFor(ALL_TYPES_RECORD_SCHEMA).getTypeParameters(), false)) {
+                type instanceof RowType rowType ? rowType.getFieldTypes() : List.of(),
```
findepi: as above -- the `?` is weird. remove

lib/trino-hive-formats/src/test/java/io/trino/hive/formats/avro/TestAvroPageDataWriterWithoutTypeManager.java
```diff
@@ -133,14 +136,16 @@ public void testRLEAndDictionaryBlocks()
...
+        Type type = new BaseAvroTypeBlockHandler().typeFor(testBlocksSchema);
         try (AvroFileWriter avroFileWriter = new AvroFileWriter(
                 trinoLocalFilesystem.newOutputFile(testLocation).create(),
                 testBlocksSchema,
                 NoOpAvroTypeManager.INSTANCE,
                 AvroCompressionKind.NULL,
                 ImmutableMap.of(),
                 testBlocksSchema.getFields().stream().map(Schema.Field::name).collect(toImmutableList()),
-                new BaseAvroTypeBlockHandler().typeFor(testBlocksSchema).getTypeParameters(), false)) {
+                type instanceof RowType rowType ? rowType.getFieldTypes() : List.of(),
```
findepi: same

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -558,14 +556,17 @@ private ColumnChunk readArray(GroupField field)
...
             throws IOException
     {
-        List<Type> parameters = field.getType().getTypeParameters();
-        checkArgument(parameters.size() == 2, "Maps must have two type parameters, found %s", parameters.size());
-        Block[] blocks = new Block[parameters.size()];
+        if (!(field.getType() instanceof MapType mapType)) {
+            throw new ParquetCorruptionException(dataSource.getId(), "Expected map type, found: %s", field.getType());
+        }
+
+        checkArgument(field.getType() instanceof MapType, "Expected array type, found: %s", field.getType());
```
findepi: duplicates the `if` above?

also, the message is wrong

findepi:
> Happy to do that. However, since all these commits affect the SPI, I wanted to get them all in the same timeframe to avoid thrash across multiple releases.

I don't think it's a really big problem.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/avro/AvroPageDataReader.java
martint: I wrote it that way because I was optimizing for the declaration of `rowType` closer to its usage, and the exception condition as a secondary concern.

The alternative:

```java
if (!(readerSchemaType instanceof RowType rowType)) {
    throw new AvroTypeException("Root Avro type must be a row");
}

pageBuilder = new PageBuilder(rowType.getFieldTypes());
initialize();
```

makes it harder to figure out where `rowType` comes from. While it's valid Java, it's harder to intuitively grasp what's the scope of that variable.

So, it was intentional, but I can change it if you feel strongly about it.

lib/trino-hive-formats/src/test/java/io/trino/hive/formats/FormatTestUtils.java
martint: That was confusing to me, too. ~~But I think this method is designed to handle the case of decoding a list of uniform values and either pivot it into a row or map it to a Trino array.~~ I actually have no idea why this method would be called with a row type where only the first field type matters. It seems like a I'm not sure what the semantics would be for a Map (maybe key-value pairs?), either, but it doesn't matter since it doesn't happen in practice. We can always add that whenever the use case comes up.

BTW, getting the first field the row is not a new behavior -- that's implicitly how it it was done before. I just ported the behavior.

lib/trino-hive-formats/src/test/java/io/trino/hive/formats/avro/TestAvroPageDataReaderWithAvroNativeTypeManagement.java
martint: Ah, yes. I think I replicated that from another case where the type varies. Fixing.

lib/trino-hive-formats/src/main/java/io/trino/hive/formats/avro/AvroPageDataReader.java
findepi: > The alternative:
> 
> ```java
> if (!(readerSchemaType instanceof RowType rowType)) {
>     throw new AvroTypeException("Root Avro type must be a row");
> }
> ```

i agree it makes it harder to see where "rowType" comes from
but makes it easier to see under what condition do we throw
this is also closer to the `verify / check*` oneliners we use (when possible)

i see it's not optimal either way. it's sad we can't have both

martint: Ok, I'll change it.

martint:
@findepi , updated

Issue #27573, opened 2025-12-08
BigQuery `query` table function throws "Already Exists" when the result is reused


ebyhr:
It seems we introduced a bug in https://github.com/trinodb/trino/pull/22432. This query doesn't fail if I remove the following `, v1`
```sql
with v1 as (select * from table(system.query('select * from tpch.region'))) select * from v1, v1;
```

```Query 20251208_070209_00035_k5938, FAILED, 1 node
http://localhost:8080/ui/query.html?20251208_070209_00035_k5938
Splits: 1 total, 0 done (0.00%)
CPU Time: 0.0s total,     0 rows/s,     0B/s, 0% active
Per Node: 0.0 parallelism,     0 rows/s,     0B/s
Parallelism: 0.0
Peak Memory: 619B
5.40 [0 rows, 0B] [0 rows/s, 0B/s]

Query 20251208_070209_00035_k5938 failed: Already Exists: Table sep-bq-cicd:_504da3ba97a95b3ba8ecf5ea2dd9656b0c9b77fb._pbc_06422a8171b745ca8fcc3d070873bb59
com.google.cloud.bigquery.BigQueryException: Already Exists: Table sep-bq-cicd:_504da3ba97a95b3ba8ecf5ea2dd9656b0c9b77fb._pbc_06422a8171b745ca8fcc3d070873bb59
	at com.google.cloud.bigquery.BigQueryRetryHelper.runWithRetries(BigQueryRetryHelper.java:71)
	at com.google.cloud.bigquery.BigQueryImpl.getQueryResults(BigQueryImpl.java:2144)
	at com.google.cloud.bigquery.BigQueryImpl.getQueryResults(BigQueryImpl.java:2123)
	at com.google.cloud.bigquery.Job$1.call(Job.java:465)
	at com.google.cloud.bigquery.Job$1.call(Job.java:462)
	at com.google.api.gax.retrying.DirectRetryingExecutor.submit(DirectRetryingExecutor.java:102)
	at com.google.cloud.bigquery.BigQueryRetryHelper.run(BigQueryRetryHelper.java:108)
	at com.google.cloud.bigquery.BigQueryRetryHelper.runWithRetries(BigQueryRetryHelper.java:62)
	at com.google.cloud.bigquery.Job.waitForQueryResults(Job.java:461)
	at com.google.cloud.bigquery.Job.waitForInternal(Job.java:322)
	at com.google.cloud.bigquery.Job.waitFor(Job.java:233)
[...]

PR #27578, opened 2025-12-08
Fix `select` query when reusing BigQuery `query` table function result

krvikash:
## Description
Fixes: https://github.com/trinodb/trino/issues/27573

## Release notes

```markdown
## BigQuery
* Fix query failure when reusing `query` table function result. ({issue}`27573`)
```

### DIFF
--- a/plugin/trino-bigquery/src/test/java/io/trino/plugin/bigquery/BaseBigQueryConnectorTest.java
+++ b/plugin/trino-bigquery/src/test/java/io/trino/plugin/bigquery/BaseBigQueryConnectorTest.java
@@ -1122,6 +1122,13 @@ public abstract class BaseBigQueryConnectorTest
         }
     }
 
+    @Test // regression test for https://github.com/trinodb/trino/issues/27573
+    public void testNativeQueryWhenResultReused()
+    {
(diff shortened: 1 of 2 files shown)


ebyhr:
/test-with-secrets sha=9f07b7104484cabc9f88747501cf48f016891326

plugin/trino-bigquery/src/test/java/io/trino/plugin/bigquery/BaseBigQueryConnectorTest.java
```diff
@@ -1122,6 +1122,13 @@ public void testNativeQuerySelectFromTestTable()
         }
     }
 
+    @Test // regression test for https://github.com/trinodb/trino/issues/27573
+    public void testNativeQueryWhenResultReused()
+    {
+        assertThat(query("WITH t AS (SELECT * FROM TABLE(system.query('SELECT regionkey FROM tpch.region WHERE regionkey = 0'))) SELECT * FROM t, t"))
+                .matches("VALUES (BIGINT '0', BIGINT '0')");
```
ebyhr, line 1129: @krvikash I modified the test query so we can reuse the existing region table.

PR #27586, opened 2025-12-08
Vectorize block deserialization

pettyjamesm:
## Description
This PR includes two significant changes to BlockSerDe vectorization:

**1. Enable Vectorized Deserialization**
Extends the SIMD support in block encoding introduced by https://github.com/trinodb/trino/pull/26919 by adding SIMD logic for deserializing block values with nulls present.

This requires splitting the SIMD support detection logic for compress / expand operations from one another since ARM intrinsics for expand only land in JDK 26+ as part of https://github.com/openjdk/jdk/pull/26740

**2. Enable Vectorized Serialization and Deserialization on Graviton 4**
This PR _also_ enables vectorized serialization on graviton 4 CPU's that previously would not have it enabled due to only having 128-bit vector registers.

**Summary of Changes for Common CPU's**:
- amd64 with AVX512F only (e.g.: Intel Sky Lake): Vectorized deserialization of `int` and `long` enabled
- amd64 with AVX512F and AVX512VBMI2 (e.g.: Intel Sapphire Rapids, AMD Zen 4 and 5): vectorized deserialization enabled for `byte`, `short`, `int`, `long`
- aarch64 with NEON support only (e.g.: Graviton 1 and 2 or Apple Silicon): No change
- aarch64 with SVE 1 support only (e.g.: Graviton 3): No change
- aarch64 with SVE 2 support (e.g.: Graviton 4): Vectorized serialization of `byte`, `short`,  and `int` (but not `long`) enabled. Vectorized deserialization of `int` enabled.

#### Benchmarks
Switched to throughput (full benchmark iterations per second) instead of nano seconds per position to avoid rounding issues in the benchmark UI. Benchmark command used:
`
./mvnw exec:exec -pl :trino-main -Dair.check.skip-all=true -Dexec.classpathScope=test -Dexec.executable="java" -Dexec.args="-cp %classpath org.openjdk.jmh.Main -tu s -opi 1 -bm thrpt -f 1 -wi 10 -i 5 -rf json -rff ./deserialize_all_vectorized_throughput.json -jvmArgsPrepend '-Xms8g -Xmx8g -XX:+AlwaysPreTouch -XX:+EnableDynamicAgentLoading --add-modules jdk.incubator.vector -XX:ReservedCodeCacheSize=256M -Djdk.attach.allowAttachSelf' io.trino.execution.buffer.BenchmarkBlockSerde.deserialize"
`

**Intel Skylake (c5 instance)**: [BenchmarkBlockSerDe.deserialize*](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/45fdb7a7fe5a2e118b6128520c4deadd/raw/6d51000fe4cf918f7d61ba6a8fc88035ddadc088/deserialize_all_scalar_throughput.json,https://gist.githubusercontent.com/pettyjamesm/45fdb7a7fe5a2e118b6128520c4deadd/raw/4d87393d8df082b4ffcd23b6a77a5236aa9a19d9/deserialize_all_c5_vectorized_v2.json)
- Only supports AVX512F which enables vectorized deserialization for int and longs, but not shorts or bytes
- Minor to significant speedups for long and integer array deserialization with nulls after special casing null-heavy handling.

**Intel Sapphire Rapids (c7i instance)**: [BenchmarkBlockSerDe.deserialize*](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/45bdc11e950fcef89ad36c792c0323bb/raw/fc9538bb554c0d563e2c1049e2dc19133caa9d42/deserialize_all_scalar_c7i_throughput.json,https://gist.githubusercontent.com/pettyjamesm/45bdc11e950fcef89ad36c792c0323bb/raw/fc9538bb554c0d563e2c1049e2dc19133caa9d42/deserialize_all_vectorized_c7i_throughput.json)
- Supports AVX512F and AVX512VBMI2 which enabled vectorized deserialization for int, long, short, and byte
- Significant deserialization speedups for all vectorized types, up to 3.5x throughput increase

**Graviton 4 (c8g instance)**: [BenchmarkBlockSerDe](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/7e97959532f1c63ced45eb62a7d2dd06/raw/d7735eea30e941d9c482bfb1be0056e4631b1529/scalar_graviton4_throughput.json,https://gist.githubusercontent.com/pettyjamesm/7e97959532f1c63ced45eb62a7d2dd06/raw/d7735eea30e941d9c482bfb1be0056e4631b1529/vectorized_graviton4_throughput_v2.json)
- Enabled support for vectorized serialization for byte, short, and int but not long
- Enabled support for vectorized deserialization of int only
- Significant speedups for all vectorized operations, with a slight regression at 99% null rate deserialization for `int`

## Release notes

```markdown
## General
* Extend experimental performance improvements for remote data exchanges on newer CPU architectures ({issue}`27586`)
* Enable experimental performance improvements for remote data exchanges on Graviton 4 CPUs ({issue}`27586`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
+++ b/core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
@@ -17,6 +17,7 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.StandardSystemProperty;
 import com.google.inject.Inject;
 import com.google.inject.Singleton;
+import io.airlift.log.Logger;
 import io.trino.FeaturesConfig;
 import jdk.incubator.vector.VectorShape;
 
@@ -44,6 +45,8 @@ Graviton2 is NEON-only (no SVE), whereas Graviton3 provides SVE.
 @Singleton
 public final class BlockEncodingSimdSupport
 {
+    private static final Logger log = Logger.get(BlockEncodingSimdSupport.class);
+
     public record SimdSupport(
             boolean compressByte,
             boolean expandByte,
@@ -84,7 +87,15 @@ public final class BlockEncodingSimdSupport
         String archRaw = StandardSystemProperty.OS_ARCH.value();
         String arch = archRaw == null ? "" : archRaw.toLowerCase(ENGLISH);
         int preferredBitWidth = VectorShape.preferredShape().vectorBitSize();
-        return determineSimdSupport(arch, preferredBitWidth, readCpuFlags());
+        Set<String> cpuFlags = readCpuFlags();
+        SimdSupport detected = determineSimdSupport(arch, preferredBitWidth, cpuFlags);
+        if (log.isDebugEnabled()) {
+            log.info("Detected SIMD Support for architecture=%s, vectorBitWidth=%s, cpuFlags=%s: %s", arch, preferredBitWidth, cpuFlags, detected);
+        }
+        else {
+            log.info("Detected SIMD Support for architecture=%s, vectorBitWidth=%s: %s", arch, preferredBitWidth, detected);
+        }
+        return detected;
     }
 
     @VisibleForTesting


pettyjamesm:
The regression on TPCDS q09 seems like an ongoing issue since that keeps getting flagged on multiple PR's. @raunaqmorarka - did we ever figure out when that regression was introduced?

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -81,14 +85,53 @@ public IntArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput
...
+            if (vectorizeNullCompressAndExpand) {
+                return expandIntsWithNullsVectorized(sliceInput, positionCount, valueIsNull);
+            }
+            else {
+                return expandIntsWithNullsScalar(sliceInput, positionCount, valueIsNullPacked, valueIsNull);
+            }
+        }
+    }
+
+    private static IntArrayBlock expandIntsWithNullsVectorized(SliceInput sliceInput, int positionCount, boolean[] valueIsNull)
```
SongChujun: Probably we want to put the ```expandIntsWithNullsVectorized``` and ```compactIntsWithNullsVectorized``` in the same place place: Either in ```EncoderUtil``` or in the encoders for individual types.

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -82,14 +86,54 @@ public ByteArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInpu
...
+        if (valueIsNull.length != positionCount) {
+            throw new IllegalArgumentException("valueIsNull length must match positionCount");
+        }
+        int nonNullPositionCount = sliceInput.readInt();
+        int nonNullIndex = positionCount - nonNullPositionCount;
+        byte[] values = new byte[positionCount];
+        sliceInput.readBytes(values, nonNullIndex, nonNullPositionCount);
+
+        int position = 0;
+        for (; position < nonNullIndex && nonNullIndex + BYTE_SPECIES.length() < values.length; position += BYTE_SPECIES.length()) {
```
SongChujun, line 153: Why don't you use ```BYTE_SPECIES.loopBound(values.length)``` here?

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -82,14 +86,54 @@ public ByteArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInpu
...
+        }
+        for (; position < nonNullIndex; position++) {
+            values[position] = valueIsNull[position] ? 0 : values[nonNullIndex++];
+        }
+        return new ByteArrayBlock(0, positionCount, valueIsNull, values);
+    }
+
+    private static ByteArrayBlock expandBytesWithNullsScalar(SliceInput sliceInput, int positionCount, byte[] valueIsNullPacked, boolean[] valueIsNull)
+    {
+        if (valueIsNull.length != positionCount) {
```
SongChujun, line 169: We don't have such check in the existing impl, why do you want to add the check?

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -81,14 +85,70 @@ public IntArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput
...
+        if (valueIsNull.length != positionCount) {
+            throw new IllegalArgumentException("valueIsNull length must match positionCount");
+        }
+        int nonNullPositionCount = sliceInput.readInt();
+        int nonNullIndex = positionCount - nonNullPositionCount;
+        int[] values = new int[positionCount];
+        sliceInput.readInts(values, nonNullIndex, nonNullPositionCount);
+
+        int position = 0;
+        if ((nonNullPositionCount * (INT_SPECIES.length() * 4L)) <= positionCount) {
```
SongChujun: how is this heuristic decided? Or how do you define ```many``` nulls?

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -28,13 +31,14 @@
 public class ByteArrayBlockEncoding
```
SongChujun, line 30: Probably you cna add tests to ```TestEncoderUtil``` on the expand code path.

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
pettyjamesm: The check is a little different. For instance, the following is a potentially valid result: `SPECIES.loopBound(16) # => 16` is typical result- but we're loading a whole vector of non-null values starting from the `nonNullIndex` which:
1. doesn't start from `0` 
2. advances in arbitrary step sizes, not `SPECIES.length()` steps
3. must be able to load a full vector from the current index

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
pettyjamesm: Doing this to ensure that the JIT has "proof" within the scope of the method that the bounds checks align between `boolean[] valueIsNull` and `int[] values` (ensures that we don't end up with worse performance if inlining into the caller doesn't occur).

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
pettyjamesm: For the branching approach to be beneficial, the branch needs to be predictable- which means that we need to detect "all nulls" more than 50% of the time through the loop body. If we assume the data is mostly nulls and the nulls are randomly distributed, a threshold of `SPECIES.length() * 4` gives us a 75% "all nulls" rate. I initially tested this with `SPECIES.length() * 2` as the threshold and the results were worse than the branchless approach.

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
pettyjamesm: Tests added

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
pettyjamesm: Moved compact methods out of EncoderUtil into their corresponding block encoder class.

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -81,14 +87,70 @@ public IntArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput
...
         byte[] valueIsNullPacked = retrieveNullBits(sliceInput, positionCount);
-        int[] values = new int[positionCount];
-
         if (valueIsNullPacked == null) {
+            int[] values = new int[positionCount];
             sliceInput.readInts(values);
             return new IntArrayBlock(0, positionCount, null, values);
         }
-        boolean[] valueIsNull = decodeNullBits(valueIsNullPacked, positionCount);
+        else {
```
raunaqmorarka: else is redundant

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -81,14 +87,70 @@ public IntArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput
...
+            boolean[] valueIsNull = decodeNullBits(valueIsNullPacked, positionCount);
+            if (vectorizeNullExpand) {
+                return expandIntsWithNullsVectorized(sliceInput, positionCount, valueIsNull);
+            }
+            else {
```
raunaqmorarka: else is redundant

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -82,14 +88,54 @@ public ByteArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInpu
...
-        byte[] values = new byte[positionCount];
 
         if (valueIsNullPacked == null) {
-            sliceInput.readBytes(Slices.wrappedBuffer(values));
+            byte[] values = new byte[positionCount];
+            sliceInput.readBytes(values, 0, values.length);
             return new ByteArrayBlock(0, positionCount, null, values);
         }
-        boolean[] valueIsNull = decodeNullBits(valueIsNullPacked, positionCount);
+        else {
```
raunaqmorarka: else is redundant

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -82,14 +88,54 @@ public ByteArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInpu
...
+            boolean[] valueIsNull = decodeNullBits(valueIsNullPacked, positionCount);
+            if (vectorizeNullExpand) {
+                return expandBytesWithNullsVectorized(sliceInput, positionCount, valueIsNull);
+            }
+            else {
```
raunaqmorarka: else is redundant

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -82,14 +88,54 @@ public ByteArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInpu
...
+            }
+        }
+    }
 
+    private static ByteArrayBlock expandBytesWithNullsVectorized(SliceInput sliceInput, int positionCount, boolean[] valueIsNull)
+    {
+        if (valueIsNull.length != positionCount) {
+            throw new IllegalArgumentException("valueIsNull length must match positionCount");
+        }
+        int nonNullPositionCount = sliceInput.readInt();
```
raunaqmorarka: nonNullPositionCount -> nonNullPositionsCount

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
```diff
@@ -82,14 +88,54 @@ public ByteArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInpu
...
+        int nonNullIndex = positionCount - nonNullPositionCount;
+        byte[] values = new byte[positionCount];
+        sliceInput.readBytes(values, nonNullIndex, nonNullPositionCount);
+
+        int position = 0;
+        for (; position < nonNullIndex && nonNullIndex + BYTE_SPECIES.length() < values.length; position += BYTE_SPECIES.length()) {
```
raunaqmorarka, line 153: Could you add a code comment explaining the loop conditions here ? It is a bit difficult to follow

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
```diff
@@ -81,14 +87,70 @@ public IntArrayBlock readBlock(BlockEncodingSerde blockEncodingSerde, SliceInput
...
+        if (valueIsNull.length != positionCount) {
+            throw new IllegalArgumentException("valueIsNull length must match positionCount");
+        }
+        int nonNullPositionCount = sliceInput.readInt();
+        int nonNullIndex = positionCount - nonNullPositionCount;
+        int[] values = new int[positionCount];
+        sliceInput.readInts(values, nonNullIndex, nonNullPositionCount);
+
+        int position = 0;
+        if ((nonNullPositionCount * (INT_SPECIES.length() * 4L)) <= positionCount) {
```
raunaqmorarka: Do we get a significant gain from doing separate branchy vs branchless versions ?
So far we have stuck to using one or the other in all the hot code paths as it keeps a lid on the code complexity even if its not the most optimized implementation in every case. 
I think it would be nicer to retain just the branchless version and accept that it is slightly slower on the many NULLs path (which should be the less common case typically).

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -159,26 +164,30 @@ enum ArmSimdInstructionSet {
...
+        // SVE 1 is sufficient to have Vector#compress(VectorMask) intrinsics for all primitive types
+        boolean compressAll = true;
+        boolean compressLong = preferredVectorBitWidth > 128; // only vectorize long compression if we can handle more than 2 values per instruction
+
+        // As of JDK 25, SVE 2 intrinsics for Vector#expand(VectorMask) over int and long, but not byte or short. JDK 26 add intrinsics
+        // for byte and short on SVE 2, SVE 1, and NEON in https://bugs.openjdk.org/browse/JDK-8363989. We can reconsider enabling vectorized
+        // expansion for those types at some point in the future
+        boolean expandInt = armFlags.contains(ArmSimdInstructionSet.sve2);
+        boolean expandLong = armFlags.contains(ArmSimdInstructionSet.sve2) && preferredVectorBitWidth > 128; // ensure minimum register width for long
+        boolean expandByteAndShort = false; // no intrinsics in JDK 25
```
raunaqmorarka, line 192: Can we print a log line which will summarize what SIMD capabilities were detected by the code ?
This has gotten complex enough that I suspect we will need that info for debugging some day.

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
pettyjamesm: Renamed

core/trino-spi/src/main/java/io/trino/spi/block/ByteArrayBlockEncoding.java
pettyjamesm: Added a comment explaining the loop conditions

core/trino-spi/src/main/java/io/trino/spi/block/IntArrayBlockEncoding.java
pettyjamesm: Unfortunately yes, the regression compared to scalar is in the ~20% range compared to scalar code even on CPU's with the most complete intrinsic support for vectorization. The issue is most apparent with `long` (but also visible with `int`) and stems from the number of extra value loads that result in no non-null stores being incurred over and over again (hence why the penalty is highest for `long` and why this extra logic is omitted for `byte` and `short` where no regression is visible).

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
pettyjamesm: Added a log line

PR #27591, opened 2025-12-09
Remove QueryManager interface and rename SqlQueryManager

wendigo:
This interface had only a single implementation so it wasn't justified to keep around

## Release notes

```markdown
## Section
```


wendigo:
It turns out that QueryManager isn't used on the worker node, so generating proxy isn't needed

core/trino-main/src/main/java/io/trino/server/testing/TestingTrinoServer.java
```diff
@@ -402,7 +401,7 @@ private TestingTrinoServer(
         sessionPropertyManager = injector.getInstance(SessionPropertyManager.class);
         if (coordinator) {
             dispatchManager = injector.getInstance(DispatchManager.class);
-            queryManager = (SqlQueryManager) injector.getInstance(QueryManager.class);
```
wendigo, line 405: I'll drop this redundant cast

losipiuk (approved):

core/trino-main/src/main/java/io/trino/server/CoordinatorModule.java
```diff
@@ -202,9 +201,10 @@ protected void setup(Binder binder)
         jaxrsBinder(binder).bind(QueryResource.class);
         jaxrsBinder(binder).bind(ResourceGroupStateInfoResource.class);
         binder.bind(QueryIdGenerator.class).in(Scopes.SINGLETON);
-        binder.bind(SqlQueryManager.class).in(Scopes.SINGLETON);
-        newExporter(binder).export(SqlQueryManager.class).withGeneratedName();
-        binder.bind(QueryManager.class).to(SqlQueryManager.class);
+        binder.bind(QueryManager.class).in(Scopes.SINGLETON);
+        newExporter(binder).export(QueryManager.class).as(generator -> generator.generatedNameOf(QueryManager.class)
+                // For backward compatibility
+                .replaceFirst("QueryManager", "SqlQueryManager"));
```
martint, line 207: We don't need to preserve the old name. We make no guarantees about stability of JMX names.

wendigo: @martint we do that in various places already so if we don't do that, we should remove the backward compatibility in other metrics related places

losipiuk: I disagree. This is not a big problem for the codebase and it is nice for the users.

PR #27593, opened 2025-12-09
Fix duplicated plugin definition

wendigo:
This isn't compatible with Maven 4.0.0

## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-product-tests-launcher/pom.xml
+++ b/testing/trino-product-tests-launcher/pom.xml
@@ -194,46 +194,6 @@
                         <ignoredNonTestScopedDependency>org.assertj:assertj-core</ignoredNonTestScopedDependency>
                     </ignoredNonTestScopedDependencies>
                 </configuration>
-            </plugin>
-            <plugin>
-                <groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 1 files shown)


testing/trino-product-tests-launcher/pom.xml
```diff
@@ -280,6 +240,41 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-shade-plugin</artifactId>
```
findepi, line 245: is this just moved around?

findepi: i suppose the diff would be more readable, if you kept the shade plugin before the the dependency plugin?

wendigo: The change was to move dependency plugin configuration up, github renders this as the shade plugin move :)

findepi: It could be compensated with a description :)

wendigo: Does it matter? Change was structural only

PR #27602, opened 2025-12-10
Cleanups and improvements around MV analysis

piotrrzysko:
Extracted from
- https://github.com/trinodb/trino/pull/27551


core/trino-main/src/test/java/io/trino/connector/MockConnector.java
```diff
@@ -745,6 +749,12 @@ public Map<String, Object> getMaterializedViewProperties(ConnectorSession sessio
         @Override
         public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName viewName)
         {
+            if (getMaterializedViewFreshness.isPresent()) {
+                MaterializedViewFreshness freshness = getMaterializedViewFreshness.get().apply(session, viewName);
+                verifyNotNull(freshness, "freshness is null");
```
findepi, line 754: Not needed, but whatever. We can remove when addressing TODO below

PR #27129, opened 2025-10-28
Added queryText regex pattern to match in resource group selectors

gertjanal:
## Description
Some queries are quick and simple, like `SELECT 1` and a query to list all tables. These queries should not have to wait in the queue to be executed, but should return quickly for a responsive UI.

With this PR, a resource selector `queryText` regex pattern can be added to match these simple types of queries and the query can be passed on to a priority resource group.

## Additional context and related issues
This feature is also available in the Starburst version: https://docs.starburst.io/latest/admin/resource-groups.html#selector-rules

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( X ) Release notes are required, with the following suggested text:

```markdown
## Resource groups
* Add `queryText` as a regular expression resource group selector. ({issue}`27129`)
```

## Summary by Sourcery

Introduce `queryText` regex support for resource group selectors to route queries based on their SQL content, updating SPI, plugin, database mapping, dispatch logic, tests, and documentation.

New Features:
- Add optional `queryText` regex in resource group selectors to match SQL query strings

Enhancements:
- Extend SelectionCriteria to carry queryText and update StaticSelector, SelectorSpec, SelectorRecord, ResourceGroupsDao, and configuration managers to handle queryText patterns
- Update DispatchManager to pass the SQL text into resource group selection logic
- Add revapi exclusions for the changed SelectionCriteria constructor signature

Documentation:
- Document the new `queryText` selector property in resource groups documentation

Tests:
- Add unit tests for queryText matching and update existing selector tests to supply and verify the new queryText field

PR #27604, opened 2025-12-10
Minor fixes to query text pattern selector

ksobolew:
## Description

The issue is that `Pattern` does not implement `hashCode` so we can't use it directly for hashing. It also doesn't implement `equals`, but it's already properly handled in `SelectorSpec`'s `equals` method - we should do the analogous thing in `hashCode` as well. As a bonus, do that in `toString`, too.

Also, call `addNamedGroups` for `queryText` criterion in `StaticSelector`.

Follow-up to #27129.

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/StaticSelector.java
+++ b/plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/StaticSelector.java
@@ -88,7 +88,10 @@ public class StaticSelector
                 }))
                 .add(userGroupRegex.map(userGroupRegexValue ->
                             new BasicMatcher(criteria -> criteria.getUserGroups().stream().anyMatch(userGroup -> userGroupRegexValue.matcher(userGroup).matches()))))
-                .add(queryTextRegex.map(queryTextRegexValue -> new PatternMatcher(variableNames, queryTextRegexValue, SelectionCriteria::getQueryText)))
+                .add(queryTextRegex.map(queryTextRegexValue -> {
+                    addNamedGroups(queryTextRegexValue, variableNames);
+                    return new PatternMatcher(variableNames, queryTextRegexValue, SelectionCriteria::getQueryText);
+                }))
                 .add(queryType.map(queryTypeValue ->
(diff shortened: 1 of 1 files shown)


plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/SelectorSpec.java
```diff
@@ -153,7 +153,8 @@ public int hashCode()
                 authenticatedUserRegex.map(Pattern::flags),
                 sourceRegex.map(Pattern::pattern),
                 sourceRegex.map(Pattern::flags),
-                queryText,
+                queryText.map(Pattern::pattern),
+                queryText.map(Pattern::flags),
```
findepi, line 157: Or maybe SelectorSpec should not implement equals/hashcode at all?
what's the use of it anyway?

(potential follow-up)

ksobolew: > Or maybe SelectorSpec should not implement equals/hashcode at all?
> what's the use of it anyway?

Sadly, I don't know the answer to that question

gertjanal:
Thanks for your fixes, I noticed them today when using the classes. Can you rename queryText to queryTextRegex in the selectorspec? That seems consistent with the other patterns. (Note the textual rename on line 58)

plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/SelectorSpec.java
gertjanal: me neither, might be best to keep it this way

ksobolew:
> Can you rename queryText to queryTextRegex in the selectorspec? That seems consistent with the other patterns.

Yeah, I noticed that. I love consistency too, but I think it's low priority this time :)

plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/SelectorSpec.java
findepi: if the equals is meaningless, there is no point in keeping things like this.

PR #27606, opened 2025-12-10
Fix handling of dropped base tables for Iceberg MVs

piotrrzysko:
## Description

Follow-up to 
* https://github.com/trinodb/trino/pull/27551

## Release notes

```markdown
## Iceberg connector
* Ignore non-existent or corrupted base Iceberg tables when querying materialized views within the grace period ({issue}`27606`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -3928,12 +3928,9 @@ public class IcebergMetadata
                     firstTableChange = firstTableChange
                             .map(epochMilli -> Math.min(epochMilli, snapshot.timestampMillis()));
                 }
-                case UnknownTableChange() -> {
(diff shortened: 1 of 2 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3928,13 +3928,10 @@ case FirstChangeSnapshot(Snapshot snapshot) -> {
                     firstTableChange = firstTableChange
                             .map(epochMilli -> Math.min(epochMilli, snapshot.timestampMillis()));
                 }
-                case UnknownTableChange() -> {
+                case UnknownTableChange(), CorruptedTableChange() -> {
```
findepi: Note
```
 // Base table is gone or table is corrupted
            return new CorruptedTableChange();
```

they were using the same table change type  because for both it was irrecoverable
now it's recoverrable (for the sake of deleted tables). it might be still ok to use the same change type, but i'd suggest renaming to `GoneOrCorruptedTableChange`

PR #27608, opened 2025-12-10
Optimize Iceberg MV freshness checks based on grace period

piotrrzysko:
## Description

This PR introduces a mechanism to guide connectors on whether they can skip expensive freshness checks when a
materialized view's grace period allows it.

## Release notes

```markdown
## Section
* Optimize Iceberg materialized view freshness checks based on grace period ({issue}`27608`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
@@ -374,6 +374,7 @@ import static io.trino.spi.StandardErrorCode.UNSUPPORTED_SUBQUERY;
 import static io.trino.spi.StandardErrorCode.VIEW_IS_RECURSIVE;
 import static io.trino.spi.StandardErrorCode.VIEW_IS_STALE;
 import static io.trino.spi.connector.MaterializedViewFreshness.Freshness.FRESH;
+import static io.trino.spi.connector.MaterializedViewFreshness.Freshness.FRESH_WITHIN_GRACE_PERIOD;
 import static io.trino.spi.connector.StandardWarningCode.REDUNDANT_ORDER_BY;
 import static io.trino.spi.function.FunctionKind.AGGREGATE;
 import static io.trino.spi.function.FunctionKind.WINDOW;
@@ -741,7 +742,7 @@ class StatementAnalyzer
             TableHandle targetTableHandle = metadata.getTableHandle(session, targetTable)
                     .orElseThrow(() -> semanticException(TABLE_NOT_FOUND, refreshMaterializedView, "Table '%s' does not exist", targetTable));
 
-            analysis.setSkipMaterializedViewRefresh(metadata.getMaterializedViewFreshness(session, name).getFreshness() == FRESH);
+            analysis.setSkipMaterializedViewRefresh(metadata.getMaterializedViewFreshness(session, name, false).getFreshness() == FRESH);
 
             TableMetadata tableMetadata = metadata.getTableMetadata(session, targetTableHandle);
             List<String> insertColumns = tableMetadata.columns().stream()
@@ -2374,10 +2375,17 @@ class StatementAnalyzer
 
         private boolean isMaterializedViewSufficientlyFresh(Session session, QualifiedObjectName name, MaterializedViewDefinition materializedViewDefinition)
         {
-            MaterializedViewFreshness materializedViewFreshness = metadata.getMaterializedViewFreshness(session, name);
+            boolean gracePeriodZero = materializedViewDefinition.getGracePeriod()
+                    .map(Duration::isZero)
+                    .orElse(false);
+
+            // Do not allow connectors to optimize freshness checks when the grace period is zero,
+            // so that this special case is handled in the analyzer rather than duplicating the logic across all connectors.
+            boolean considerGracePeriod = !gracePeriodZero;
+            MaterializedViewFreshness materializedViewFreshness = metadata.getMaterializedViewFreshness(session, name, considerGracePeriod);
             MaterializedViewFreshness.Freshness freshness = materializedViewFreshness.getFreshness();
 
-            if (freshness == FRESH) {
+            if (freshness == FRESH || freshness == FRESH_WITHIN_GRACE_PERIOD) {
                 return true;
             }
             Optional<Instant> lastFreshTime = materializedViewFreshness.getLastFreshTime();
@@ -2389,13 +2397,13 @@ class StatementAnalyzer
                 // Unlimited grace period
                 return true;
             }
-            Duration gracePeriod = materializedViewDefinition.getGracePeriod().get();
-            if (gracePeriod.isZero()) {
+            if (gracePeriodZero) {
                 // Consider 0 as a special value meaning "do not accept any staleness". This makes 0 more reliable, and more likely what user wanted,
                 // regardless of lastFreshTime, query time or rounding.
                 return false;
             }
 
+            Duration gracePeriod = materializedViewDefinition.getGracePeriod().get();
             // Can be negative
             Duration staleness = Duration.between(lastFreshTime.get(), session.getStart());
             return staleness.compareTo(gracePeriod) <= 0;
--- a/core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
+++ b/core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
@@ -1405,6 +1405,15 @@ public class TracingConnectorMetadata
     }
 
     @Override
+    public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name, boolean considerGracePeriod)
+    {
+        Span span = startSpan("getMaterializedViewFreshness", name);
+        try (var _ = scopedSpan(span)) {
+            return delegate.getMaterializedViewFreshness(session, name, considerGracePeriod);
+        }
+    }
+
+    @Override
     public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name)
     {
         Span span = startSpan("getMaterializedViewFreshness", name);
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
@@ -1754,9 +1754,19 @@ public interface ConnectorMetadata
 
     /**
      * The method is used by the engine to determine if a materialized view is current with respect to the tables it depends on.
+     * <p>
+     * When {@code considerGracePeriod} is {@code true}, connectors may skip expensive freshness checks and return
+     * {@link MaterializedViewFreshness.Freshness#FRESH_WITHIN_GRACE_PERIOD} if the materialized view is within its
+     * configured grace period, evaluated relative to the session start time ({@link ConnectorSession#getStart()}).
      *
      * @throws MaterializedViewNotFoundException when materialized view is not found
      */
+    default MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name, boolean considerGracePeriod)
+    {
+        return getMaterializedViewFreshness(session, name);
+    }
+
+    @Deprecated(forRemoval = true)
     default MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name)
     {
         throw new TrinoException(GENERIC_INTERNAL_ERROR, "ConnectorMetadata getMaterializedView() is implemented without getMaterializedViewFreshness()");
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/MaterializedViewFreshness.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/MaterializedViewFreshness.java
@@ -76,6 +76,14 @@ public final class MaterializedViewFreshness
     public enum Freshness
     {
         FRESH,
+        /**
+         * The materialized view is fresh enough within its grace period. Returned when a connector
+         * is invoked with {@code considerGracePeriod = true} and determined it could skip expensive
+         * freshness checks (e.g., checking base table snapshots) because the MV is still within
+         * the grace period bounds. The actual freshness of base tables was not verified, but worst-case
+         * staleness is guaranteed to be within the acceptable grace period.
+         */
+        FRESH_WITHIN_GRACE_PERIOD,
         STALE,
         UNKNOWN,
         /**/
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -377,6 +377,7 @@ import static io.trino.spi.StandardErrorCode.TABLE_ALREADY_EXISTS;
 import static io.trino.spi.StandardErrorCode.TABLE_NOT_FOUND;
 import static io.trino.spi.StandardErrorCode.TYPE_MISMATCH;
 import static io.trino.spi.connector.MaterializedViewFreshness.Freshness.FRESH;
+import static io.trino.spi.connector.MaterializedViewFreshness.Freshness.FRESH_WITHIN_GRACE_PERIOD;
 import static io.trino.spi.connector.MaterializedViewFreshness.Freshness.STALE;
 import static io.trino.spi.connector.MaterializedViewFreshness.Freshness.UNKNOWN;
 import static io.trino.spi.connector.RetryMode.NO_RETRIES;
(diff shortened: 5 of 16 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3878,9 +3878,11 @@ public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession s
                 .map(snapshot -> Boolean.valueOf(snapshot.summary().getOrDefault(DEPENDS_ON_TABLE_FUNCTIONS, "false")))
                 .orElse(false);
 
-        Optional<Instant> refreshTime = currentSnapshot.map(snapshot -> snapshot.summary().get(TRINO_QUERY_START_TIME))
-                .map(Instant::parse)
-                .or(() -> currentSnapshot.map(snapshot -> Instant.ofEpochMilli(snapshot.timestampMillis())));
+        // refreshStartTime is captured at the beginning of refresh
```
findepi: isn't it obvious from the property name - TRINO_QUERY_START_TIME?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3878,9 +3878,11 @@ public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession s
...
+        Optional<Instant> refreshStartTime = currentSnapshot.map(snapshot -> snapshot.summary().get(TRINO_QUERY_START_TIME))
+                .map(Instant::parse);
+        // For MVs refreshed before TRINO_QUERY_START_TIME was introduced, fall back to snapshot timestamp (captured at end of refresh)
+        Optional<Instant> refreshTime = refreshStartTime.or(() -> currentSnapshot.map(snapshot -> Instant.ofEpochMilli(snapshot.timestampMillis())));
```
findepi: Adding a comment doesn't require adding a variable

```java
Optional<Instant> refreshTime = currentSnapshot.map(snapshot -> snapshot.summary().get(TRINO_QUERY_START_TIME))
.map(Instant::parse)
// Fallback to snapshot commit time (end of refresh) for MVs defined before TRINO_QUERY_START_TIME was introduced
.or(() -> currentSnapshot.map(snapshot -> Instant.ofEpochMilli(snapshot.timestampMillis())));
```

core/trino-spi/src/main/java/io/trino/spi/connector/MaterializedViewFreshnessCheckPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+     * grace period. A connector may skip expensive operations (e.g., metastore calls to check
+     * base table snapshots) if it can determine the MV is fresh enough based on the
+     * referenceTime and the grace period configured in the materialized view definition.
+     *
+     * @param referenceTime The point in time against which freshness is being evaluated
+     *                      (typically the query start time)
+     */
+    record ConsiderGracePeriod(Instant referenceTime)
+            implements MaterializedViewFreshnessCheckPolicy {}
+}
```
findepi: That's beautiful, but mouthful and makes following the logic harder.
I wanted this to be single boolean param like `boolean considerGracePeriod`...

It seems this SPI design was forced by the fact we don't want the connector to use `session.getStart()`, and we don't want the connector to use that value because... it's mocked in tests. 
So, SPI shape (and complexity) is a by product of existing tests. Not great. 
Let's find a way to avoid that.

For example, we could pass `SessionTimeProvider` to `io.trino.Session#toConnectorSession` so that `FullConnectorSession` can have its reported start time overriden.
Or, perhaps we could fix this to work (in tests only)
https://github.com/trinodb/trino/blob/3d347c5e8394b032fe2df419aafbbdce9e75c2ed/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java#L1520

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3895,6 +3898,27 @@ public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession s
...
+            // The refresh query can read base table snapshots created before or during its execution. In the most
+            // pessimistic scenario, a new base table snapshot is created immediately after the refresh started
+            // (at refreshStartTime + epsilon), but the refresh reads an older snapshot. This new snapshot would
+            // not be recorded in DEPENDS_ON_TABLES, making the MV technically stale. However, when ConsiderGracePeriod
+            // policy is used and refreshStartTime + gracePeriod > referenceTime, we can safely say that the MV
+            // is at least within the grace period because refreshStartTime is before the new snapshot creation time.
+            Optional<java.time.Duration> gracePeriod = materializedViewDefinition.get().getGracePeriod();
+            if (gracePeriod.isEmpty()) {
+                // infinite grace period
+                return new MaterializedViewFreshness(FRESH_WITHIN_GRACE_PERIOD, Optional.empty());
```
findepi: This case doesn't have to be guarded with `refreshStartTime.isPresent()` check. It's easy to address, by melding the `if`s into one:

```java
if (policy instanceof ConsiderGracePeriod(Instant referenceTime) && withinGracePeriod(gracePeriod, refreshTime, referenceTime)) {
    // bla bla bla
    return new MaterializedViewFreshness(FRESH_WITHIN_GRACE_PERIOD, Optional.empty());
}


boolean withinGracePeriod(gracePeriod, ...) {
  if gracePeriod.isEmpty() {
    // infinite grace period
    return true;
  }
 ...
}
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMetastoreAccessOperations.java
```diff
@@ -219,7 +219,7 @@ public void testSelectFromFreshMaterializedView()
 
         assertMetastoreInvocations("SELECT * FROM test_select_fresh_mview_view",
                 ImmutableMultiset.<MetastoreMethod>builder()
-                        .addCopies(GET_TABLE, 2)
+                        .addCopies(GET_TABLE, 1)
```
findepi, line 222: ❤️ 

maybe it would be wise to have an MV defined on couple source tables, so that impact is more visible in the test

(i should have thought about this in https://github.com/trinodb/trino/pull/27602/commits/37a7dca16fabe8e26092310e41d35f5a524536c9 ...)

core/trino-spi/src/main/java/io/trino/spi/connector/MaterializedViewFreshnessCheckPolicy.java
piotrrzysko: There were two reasons:
* More important: I wanted to avoid making the connector aware of what the reference time is. Even in the analyzer we have a TODO (which we probably should remove) that suggests we aren’t sure whether it should be the current time or the session start time.
* The second reason you almost figured out: I initially wanted to use the current time in the connector as the safest, most pessimistic option. But the already-mocked session start time stopped me…

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
piotrrzysko: It wasn’t obvious to me at first what query we were talking about. What do you think about renaming `TRINO_QUERY_START_TIME` to `LAST_REFRESH_START_TIME`?

findepi: Renaming the constant doesn't help and is actually more confusing. The constant name corresponds to it's value (the property name). It's what it is.

core/trino-spi/src/main/java/io/trino/spi/connector/MaterializedViewFreshnessCheckPolicy.java
piotrrzysko: I removed the TODO and made the session start time part of the contract with connectors (extended the javadoc for `getMaterializedViewFreshness`). Regarding mocking, I realized there were two mechanisms for it, so I consolidated them into one.

piotrrzysko:
Rebased to resolve a conflict.

core/trino-main/src/main/java/io/trino/testing/TransactionBuilder.java
```diff
@@ -137,9 +139,14 @@ public <T> T execute(Session session, Function<Session, T> callback)
         requireNonNull(session, "session is null");
         requireNonNull(callback, "callback is null");
 
-        session = Session.builder(session)
-                .setQueryId(QUERY_ID_GENERATOR.createNextQueryId())
-                .build();
+        Session.SessionBuilder sessionBuilder = Session.builder(session)
+                .setQueryId(QUERY_ID_GENERATOR.createNextQueryId());
+        if (session.getSystemProperties().containsKey(SESSION_START_TIME_PROPERTY)) {
+            Instant sessionStartTime = session.getSystemProperty(SESSION_START_TIME_PROPERTY, Instant.class);
```
findepi, line 145: Does this obsolete following?
https://github.com/trinodb/trino/blob/3452189dfa4d8f6b1f5bacddf3bcee84c7aba672/core/trino-main/src/main/java/io/trino/server/testing/TestingTrinoServer.java#L879-L882

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -93,7 +93,7 @@
 import static io.airlift.units.Duration.nanosSince;
 import static io.trino.SystemSessionProperties.IGNORE_STATS_CALCULATOR_FAILURES;
 import static io.trino.connector.informationschema.InformationSchemaTable.INFORMATION_SCHEMA;
-import static io.trino.server.testing.TestingSystemSessionProperties.TESTING_SESSION_TIME;
+import static io.trino.server.testing.TestingTrinoServer.SESSION_START_TIME_PROPERTY;
```
findepi, line 96: argh, TESTING_SESSION_TIME was a better name, because it indicates it's test-only thing :) 

no change requested

core/trino-main/src/main/java/io/trino/server/testing/TestingSessionTimeProvider.java
```diff
@@ -1,35 +0,0 @@
...
-package io.trino.server.testing;
-
-import io.trino.Session;
-import io.trino.sql.analyzer.SessionTimeProvider;
-
-import java.time.Instant;
-
-import static io.trino.server.testing.TestingSystemSessionProperties.TESTING_SESSION_TIME;
-
-public class TestingSessionTimeProvider
```
findepi, line 23: so nice!

core/trino-spi/src/main/java/io/trino/spi/connector/MaterializedViewFreshness.java
```diff
@@ -76,6 +76,14 @@ public String toString()
...
     {
         FRESH,
+        /**
+         * The materialized view is fresh enough within its grace period. Returned when a connector
+         * is invoked with {@code considerGracePeriod = true} and determined it could skip expensive
+         * freshness checks (e.g., checking base table snapshots) because the MV is still within
+         * the grace period bounds. The actual freshness of base tables was not verified, but worst-case
+         * staleness is guaranteed to be within the acceptable grace period.
+         */
+        FRESH_WITHIN_GRACE_PERIOD,
```
findepi, line 86: well written doc

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1754,10 +1754,14 @@ default Map<String, Object> getMaterializedViewProperties(ConnectorSession sessi
...
      * The method is used by the engine to determine if a materialized view is current with respect to the tables it depends on.
+     * <p>
+     * When {@code considerGracePeriod} is {@code true}, connectors may skip expensive freshness checks and return
+     * {@link MaterializedViewFreshness.Freshness#FRESH_WITHIN_GRACE_PERIOD} if the materialized view is within its
+     * configured grace period, evaluated relative to the session start time ({@link ConnectorSession#getStart()}).
      *
      * @throws MaterializedViewNotFoundException when materialized view is not found
      */
-    default MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name)
+    default MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name, boolean considerGracePeriod)
```
findepi, line 1764: That's a breaking change. A connector which implements `getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name)` should work for some time.

```java
   default MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name, boolean considerGracePeriod)
    {
        return getMaterializedViewFreshness(session, name);
    }

    @Deprecated(forRemoval = true)
    default MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name)
    {
        throw new TrinoException(GENERIC_INTERNAL_ERROR, "ConnectorMetadata getMaterializedView() is implemented without getMaterializedViewFreshness()");
    }
```

core/trino-main/src/main/java/io/trino/tracing/TracingConnectorMetadata.java
```diff
@@ -1405,11 +1405,11 @@ public Map<String, Object> getMaterializedViewProperties(ConnectorSession sessio
     }
 
     @Override
-    public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name)
+    public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession session, SchemaTableName name, boolean considerGracePeriod)
```
findepi, line 1408: When you restore the old `getMaterializedViewFreshness` signature, the Tracing, ClassLoaderSafe (and any other dumb wrappers if exist) should implement both

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -2374,10 +2375,17 @@ protected Scope visitTable(Table table, Optional<Scope> scope)
...
         private boolean isMaterializedViewSufficientlyFresh(Session session, QualifiedObjectName name, MaterializedViewDefinition materializedViewDefinition)
         {
-            MaterializedViewFreshness materializedViewFreshness = metadata.getMaterializedViewFreshness(session, name);
+            boolean gracePeriodZero = materializedViewDefinition.getGracePeriod()
+                    .map(Duration::isZero)
+                    .orElse(false);
+
+            // Do not allow connectors to optimize freshness checks when the grace period is zero,
+            // so that this special case is handled in the analyzer rather than duplicating the logic across all connectors.
+            boolean considerGracePeriod = !gracePeriodZero;
```
findepi, line 2384: I am not convinced this is necessary.

i know it follows my own logic
https://github.com/trinodb/trino/blob/3f8b4e056227f17be8f2a410877407c950250b37/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java#L2397-L2398

but it makes code flow more complicated.

anyway, no change requested

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3951,6 +3968,16 @@ case UnknownTableChange(), GoneOrCorruptedTableChange() -> {
...
     }
 
+    private boolean withinGracePeriod(Instant sessionStart, Optional<Instant> maybeRefreshStartTime, Optional<java.time.Duration> gracePeriod)
+    {
+        if (gracePeriod.isEmpty()) {
+            // infinite grace period
+            return true;
+        }
+        return maybeRefreshStartTime.map(refreshStartTime -> refreshStartTime.plus(gracePeriod.get()).isAfter(sessionStart))
+                .orElse(false);
```
findepi: i think the orElse is an awesome API to use when unwrapping optional gotten from someone else
I don't find the `.map(complex expression).orElse()` pattern readable

I prefer explicit flow (sadly this requires silencing intellij's check that's generally not unreasonable)
```java
    private boolean withinGracePeriod(Instant sessionStart, Optional<Instant> refreshStartTime, Optional<java.time.Duration> gracePeriod)
    {
        if (gracePeriod.isEmpty()) {
            // infinite grace period
            return true;
        }
        //noinspection OptionalIsPresent
        if (refreshStartTime.isEmpty()) {
            // refresh time unknown
            return false;
        }
        return refreshStartTime.get().plus(gracePeriod.get()).isAfter(sessionStart);
    }
```

or this can be written with `orElse` still:

```java
        Instant refreshTime = refreshStartTime
                // there are no MVs refreshed before that time
                .orElse(Instant.EPOCH);
        return refreshTime.plus(gracePeriod.get()).isAfter(sessionStart);
```

core/trino-main/src/main/java/io/trino/testing/TransactionBuilder.java
piotrrzysko: No. Unfortunately, the code from `TestingTrinoServer` is not used when asserting plans: https://github.com/trinodb/trino/blob/c162f1e41d93e39a32992f4bd0943c1be9aa7c75/core/trino-main/src/test/java/io/trino/sql/query/QueryAssertions.java#L575-L586
So I needed to duplicate the logic here.

findepi:
red

findepi:
> red

GH fault. retrying

PR #27613, opened 2025-12-11
Inline dependency versions in root pom.xml

ebyhr:
## Release notes


pom.xml
```diff
@@ -2415,7 +2411,7 @@
                 <plugin>
                     <groupId>org.apache.maven.plugins</groupId>
                     <artifactId>maven-failsafe-plugin</artifactId>
-                    <version>${dep.plugin.failsafe.version}</version>
+                    <version>${dep.plugin.surefire.version}</version>
```
findepi, line 2414: That feels wrong. failsafe and surefire are separate plugins IMO

PR #20437, opened 2024-01-22
Improve Delta lake caching of metadata

Pluies:
## Description

👋  This is a reworking of #17516 based off the current master branch. c/c the description of the previous issue:

> Currently, when a new commit is made to a Delta table, the cached metadata entry of the most up to date TableSnapshot is invalidated. This means that a metadata entry must be re-read from the checkpoint and the commits made after the checkpoint (please see https://github.com/trinodb/trino/commit/c0d09377b05000145852fff6e7e649c9f6ba76de). This is unnecessary work, as we always read the new commits, and could be reconciling the cached metadata entry with the possible metadata entries loaded from the new commits.
> 
> This PR modifies the TransactionLogTail to keep track of any metadata entries it may contain. In the TableSnapshot the cached metadata entry is reconciled with any metadata entry of the TransactionLogTail.
>
> This fixes the seconds part of https://github.com/trinodb/trino/issues/17406 .

Besides, this PR also:
- Extends the caching logic above to protocolEntries
- Fixes the type signature for getProtocolEntry to match getMetadataEntry (kept in a separate commit for ease of review)

## Additional context and related issues

Fixes #17406

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
# Section
* Improve caching of metadata and protocol entries from Delta Lake logs. ({issue}`17406`)
```

PR #27618, opened 2025-12-11
Delta Lake: Drop live files caching functionality in favor of checkpoint filtering

findinpath:
## Description

Checkpoint filtering has been the default way to retrieve active data files for more than a year now in the Delta Lake connector while live files caching has been rather a fallback in case of dealing with bugs in the checkpoint filtering functionality.

One side effect of using active data files caching is that it can build up and restrict the memory available on the coordinator.

This contribution retires the following settings:

- `metadata.live-files.cache-size`
- `metadata.live-files.cache-ttl`
- `checkpoint-filtering.enabled`

## Release notes

```markdown
## Delta Lake
* {{breaking}} Remove live files table metadata cache. The configuration properties `metadata.live-files.cache-size`, `metadata.live-files.cache-ttl` and `checkpoint-filtering.enabled` are now defunct and must be removed from server configurations. ({issue}`27618`)
```


chenjian2664:
Do we have any statistics or proven can share here, that user actually not set `checkpoint-filtering.enabled` to `false`

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeBasic.java
```diff
@@ -2052,9 +2052,6 @@ private void testPartitionValuesParsed(String resourceName)
         copyDirectoryContents(new File(Resources.getResource(resourceName).toURI()).toPath(), tableLocation);
         assertUpdate("CALL system.register_table(CURRENT_SCHEMA, '%s', '%s')".formatted(tableName, tableLocation.toUri()));
 
-        Session session = Session.builder(getQueryRunner().getDefaultSession())
-                .setCatalogSessionProperty("delta", "checkpoint_filtering_enabled", "false")
-                .build();
```
raunaqmorarka: remove extra newline

findinpath:
> Do we have any statistics or proven can share here, that user actually not set checkpoint-filtering.enabled to false ?

FWIW, I've verified the usage of the session property `checkpoint_filtering_enabled` over the past year internally on Starburst Galaxy and there hasn't been any usage of this session property.

@pettyjamesm / @zhaner08  do you happen to have made use of "live-files" cache functionality for **Delta Lake** at Athena?

@jkylling / @Pluies  maybe you can share as well insights whether this functionality does still make sense for you. I know you were relying on it a couple of years ago.

raunaqmorarka:
BTW if the filesystem cache is enabled, we will already get caching of all the accessed portions of delta metadata files on the coordinator disk.

Pluies:
Hey @raunaqmorarka @findinpath , circling back here!

We've decided to align with upstream on this, and remove our custom improvements from #20437 – effectively implementing the changes from this PR. 😄

So we're upgrading from:
- Trino 467 with our improvements
- `checkpoint-filtering.enabled=false`
- No filesystem-level caching of `/_delta_log`

To:
- Trino 479
- `checkpoint-filtering.enabled=true`
- File-level caching of `/_delta_log`

It's a bit early to tell, but our preliminary testing shows:
- Identical performance in `analysis` stage
- Lower performance in `planning` stage (25% slower)
- Better performance in `execution` stage
- _Overall performance gains_ due the execution stage gains overperforming the planning losses
- Much stabler coordinator: our coordinators are often running hot memory-wise, which causes high GC rates / GC pauses and high latency. We're seeing this being much improved with the upgrade.

While the planning performance decrease is disappointing, it's overall a net positive.

I'll report back after a few weeks of running this!

raunaqmorarka:
@Pluies can you check where the time is predominantly spent during planning ? (see `io.trino.execution.QueryStats#optimizerRulesSummaries`).
I suspect it's from reading table statistics, but would be good to know for sure.

chenjian2664:
If it possible, @Pluies can just put a timer [here](https://github.com/trinodb/trino/pull/27618/changes#diff-05a47c579f9acb4e90ab33c0b5cf47204d7ff01fb7ede859fe3de9e57b84e62cR386-R395) to see the difference for queries

PR #27622, opened 2025-12-11
Update jmxutils to 1.27

wendigo:
This fixes an issue with mbeans not being unexported
properly which breaks some of the tests that actually relied on a bug.

- fixes https://github.com/trinodb/trino/issues/27825

## Release notes

```markdown
## Section
```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -183,8 +183,6 @@
 
         <!-- keep dependency properties sorted -->
         <dep.airlift.version>390</dep.airlift.version>
-        <!-- TODO (https://github.com/trinodb/trino/issues/27825) temporary downgrade -->
-        <dep.jmxutils.version>1.26</dep.jmxutils.version>
         <dep.alluxio.version>2.9.6</dep.alluxio.version>
         <dep.antlr.version>4.13.2</dep.antlr.version>
         <dep.avro.version>1.12.1</dep.avro.version>
@@ -2480,12 +2478,6 @@
                     <artifactId>maven-enforcer-plugin</artifactId>
                     <configuration>
                         <rules>
-                            <requireUpperBoundDeps>
-                                <excludes combine.children="append">
-                                    <!-- TODO (https://github.com/trinodb/trino/issues/27825) update separately -->
-                                    <exclude>org.weakref:jmxutils</exclude>
-                                </excludes>
-                            </requireUpperBoundDeps>
                             <bannedDependencies>
                                 <excludes combine.children="append">
                                     <!-- We don't use log4j2, additionally versions < 2.15.0 are vulnerable to the RCE Log4Shell (CVE-2021-44228) -->
--- a/plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestJmxStats.java
+++ b/plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestJmxStats.java
@@ -16,6 +16,7 @@ package io.trino.plugin.jdbc;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.trino.spi.Plugin;
+import io.trino.spi.connector.Connector;
 import io.trino.spi.connector.ConnectorFactory;
 import io.trino.testing.TestingConnectorContext;
 import org.junit.jupiter.api.Test;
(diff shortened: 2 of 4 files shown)


plugin/trino-elasticsearch/src/test/java/io/trino/plugin/elasticsearch/ElasticsearchQueryRunner.java
```diff
@@ -91,14 +92,16 @@ public static Builder builder(ElasticsearchServer server)
...
         private final ElasticsearchServer server;
         private final Map<String, String> connectorProperties = new HashMap<>();
         private List<TpchTable<?>> initialTables = ImmutableList.of();
 
         private Builder(ElasticsearchServer server)
         {
             super(testSessionBuilder()
-                    .setCatalog("elasticsearch")
+                    // We need separate catalog names because these tests check for JMX beans that would clash otherwise
+                    .setCatalog("elasticsearch_" + instanceId.incrementAndGet())
```
findepi: use `randomNameSuffix`.

findepi: however, general purpose ElasticsearchQueryRunner should create a catalog with fixed name.

mbeans are exported by many connectors. what's special about ES?

pom.xml
```diff
@@ -197,6 +197,7 @@
         <dep.frontend-npm.version>11.2.0</dep.frontend-npm.version>
         <dep.httpcore5.version>5.3.6</dep.httpcore5.version>
         <dep.iceberg.version>1.10.0</dep.iceberg.version>
+        <dep.jmxutils.version>1.27</dep.jmxutils.version>
```
findepi: we should have automation that would help us prune these deps when updating airbase (when airbase has same or newer version)

plugin/trino-elasticsearch/src/test/java/io/trino/plugin/elasticsearch/ElasticsearchQueryRunner.java
wendigo: There are 3 query runner running at the same time, each registering the same mbeans with the same name, when the shortest test finishes, mbeans are correctly unexported which is causing other two query runner and their tests to fail on checking jmx stats

pom.xml
wendigo: good point, i will think how we can do that

plugin/trino-elasticsearch/src/test/java/io/trino/plugin/elasticsearch/ElasticsearchQueryRunner.java
findepi: I think the right solution would be to `@Isolate` the test that "ownership" of the JVM's mbean server

martint: That still a workaround. The real solution would be to have the engine provide the mbean server to use (or a mechanism to export mbeans) to the plugin, and use a fake implementation during testing. This is something we wanted to do years ago but never got to it.

wendigo: So ConnectorContext?

findepi: i suggest we use isolated test to allow this upgrade to go thru and file an issue covering addition of the testing mbean. I _suppose_ this upgrade is quite important from dynamic catalogs perspective, so i wouldn't want it to be blocked on more development effort.

martint: I’m ok with isolation for now. I was just commenting on where we should be aiming.

plugin/trino-elasticsearch/src/test/java/io/trino/plugin/elasticsearch/ElasticsearchQueryRunner.java
```diff
@@ -98,7 +99,8 @@ public static final class Builder
         private Builder(ElasticsearchServer server)
         {
             super(testSessionBuilder()
-                    .setCatalog("elasticsearch")
+                    // We need separate catalog names because these tests check for JMX beans that would clash otherwise
+                    .setCatalog("elasticsearch_" + randomNameSuffix())
```
findepi: Same change is also offered in 
- https://github.com/trinodb/trino/pull/27623
 
Let's clarify this in
- https://github.com/trinodb/trino/pull/27623#discussion_r2653175180

wendigo:
@findepi ptal here

PR #27434, opened 2025-11-25
Deprecate EXPLAIN (TYPE LOGICAL)

losipiuk:
## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
() Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
## General
* Deprecate `EXPLAIN (TYPE LOGICAL)`. `EXPLAIN (TYPE DISTRIBUTED)` should be used instead. ({issue}`27434`)
```

PR #27626, opened 2025-12-12
Deprecate EXPLAIN (TYPE DISTRIBUTED)

chenjian2664:
## Description

follow up of https://github.com/trinodb/trino/pull/27434
Going to deprecate the EXPLAIN (TYPE DISTRIBUTED) 

https://github.com/trinodb/trino/pull/27153#discussion_r2609180198

## Release notes

```markdown
## Section
* Deprecate EXPLAIN type LOGICAL and DISTRIBUTED. Use EXPLAIN without a type clause, instead
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/QueryExplainer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/QueryExplainer.java
@@ -60,8 +60,8 @@ import static java.util.Objects.requireNonNull;
 public class QueryExplainer
 {
     public static final String DEPRECATED_TYPE_LOGICAL_WARNING = """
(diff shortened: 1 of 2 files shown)


chenjian2664:
does this line needs to be updated https://github.com/trinodb/trino/blob/5e25a79456c8fa62435e14e9a7def21f5be4e326/core/trino-main/src/main/java/io/trino/sql/rewrite/ExplainRewrite.java#L113 ?

findepi (changes requested):
I don't think it's a good idea.

It's clear EXPLAIN has a type attribute
```
EXPLAIN (TYPE IO) ...
EXPLAIN (TYPE VALIDATE) ...
```

it's not user friendly when there is no TYPE option that corresponds to the default behavior

chenjian2664:
@findepi Thanks

PR #27632, opened 2025-12-12
Update React

mosabua:
## Description

Preventative measure to avoid alerts from the recent React security issues. See https://www.youtube.com/watch?v=NEWgj-KP4Sg 

Tested now - works fine. Update to 19.2 can be done later

## Release notes

### DIFF
--- a/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json
+++ b/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json
@@ -20,9 +20,9 @@
         "@xyflow/react": "^12.8.5",
         "axios": "^1.12.2",
         "lodash": "^4.17.21",
-        "react": "^19.1.1",
+        "react": "^19.1.4",
         "react-calendar-timeline": "^0.30.0-beta.3",
-        "react-dom": "^19.1.1",
+        "react-dom": "^19.1.4",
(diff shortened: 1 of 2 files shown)


koszti:
should we consider using regular `npm audit fix` instead of bumping vulnerable packages one by one? - somewhat related to https://github.com/trinodb/trino/pull/27869

mosabua:
> should we consider using regular `npm audit fix` instead of bumping vulnerable packages one by one? - somewhat related to #27869

I think we want to do that eventually ...  maybe after merging this PR already ... the only thing with these larger PRs that update a lot is that it is harder to test to some degree and also to figure out the right release notes entry (if any). And they might carry more related code changes as well

ebyhr:
Hmm, `package-lock.json` is different from a file I built locally.

mosabua:
> Hmm, `package-lock.json` is different from a file I built locally.

With new releases out, the lock file will change most likely. Ideally we update more often .. there is a separate PR for that somewhere. I purposely limited the changes in this PR.

Also thanks for merge @wendigo

ebyhr:
@mosabua I don't follow. The different `package-lock.json` means we have to revert the change before sending a PR. I updated the file in https://github.com/trinodb/trino/pull/27986.

mosabua:
Thanks for the follow up @ebyhr

PR #27633, opened 2025-12-12
Add support for `write.object-storage.partitioned-paths` Iceberg table property

ebyhr:
## Description

The PR adds support for `write.object-storage.partitioned-paths` table property. 
The property avoids including partition values in the location if it's disabled.

## Release notes

```markdown
## Iceberg
* Add support for `write.object-storage.partitioned-paths` Iceberg table property. ({issue}`27633`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
@@ -25,7 +25,10 @@ import java.util.Map;
 
 import static com.google.common.hash.Hashing.murmur3_32_fixed;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.apache.iceberg.TableProperties.WRITE_OBJECT_STORE_PARTITIONED_PATHS;
+import static org.apache.iceberg.TableProperties.WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT;
 import static org.apache.iceberg.util.LocationUtil.stripTrailingSlash;
+import static org.apache.iceberg.util.PropertyUtil.propertyAsBoolean;
 
 // based on org.apache.iceberg.LocationProviders.ObjectStoreLocationProvider
 public class ObjectStoreLocationProvider
@@ -40,12 +43,14 @@ public class ObjectStoreLocationProvider
     private static final int ENTROPY_DIR_DEPTH = 3;
     private final String storageLocation;
     private final String context;
+    private final boolean includePartitionPaths;
 
     public ObjectStoreLocationProvider(String tableLocation, Map<String, String> properties)
     {
         this.storageLocation = stripTrailingSlash(dataLocation(properties, tableLocation));
         // if the storage location is within the table prefix, don't add table and database name context
         this.context = storageLocation.startsWith(stripTrailingSlash(tableLocation)) ? null : pathContext(tableLocation);
+        this.includePartitionPaths = propertyAsBoolean(properties, WRITE_OBJECT_STORE_PARTITIONED_PATHS, WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT);
     }
 
     @SuppressWarnings("deprecation")
@@ -67,7 +72,10 @@ public class ObjectStoreLocationProvider
     @Override
     public String newDataLocation(PartitionSpec spec, StructLike partitionData, String filename)
     {
-        return newDataLocation("%s/%s".formatted(spec.partitionToPath(partitionData), filename));
+        if (includePartitionPaths) {
+            return newDataLocation("%s/%s".formatted(spec.partitionToPath(partitionData), filename));
+        }
+        return newDataLocation(filename);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


raunaqmorarka (approved):

findepi:
> [apache/iceberg#11112](https://github.com/apache/iceberg/pull/11112) changed the hash scheme from 32-bit base64 to 20-bit base2 in [ObjectStoreLocationProvider](https://github.com/apache/iceberg/blob/e90b06cc5f07b948a56021d129f128dd8b81123c/core/src/main/java/org/apache/iceberg/LocationProviders.java#L128). The first commit follows the upstream change.

the hash is derived from partitions, or is effectively a random number?


> The second commit adds support for `write.object-storage.partitioned-paths` table property.
> The property avoids including partition values in the location if it's disabled.

when the partitioning is not disabled, the hash part is written in the path before or after the partitions string?

findepi:
does this change require any docs update?

ebyhr:
> the hash is derived from partitions, or is effectively a random number?

It's derived from partitions + file names. Since the file name includes the query ID and a UUID, it's effectively random. 

> when the partitioning is not disabled, the hash part is written in the path before or after the partitions string?

The hash part comes first. e.g. `s3://table-location/data/1101/1101/0101/10100010/part=2/...`

> does this change require any docs update?

Good question. I didn't update the docs because I thought this was mostly an implementation detail.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
```diff
@@ -32,7 +32,12 @@ public class ObjectStoreLocationProvider
         implements LocationProvider
 {
     private static final HashFunction HASH_FUNC = murmur3_32_fixed();
-    private static final Base64.Encoder BASE64_ENCODER = Base64.getUrlEncoder().withoutPadding();
+    // Length of entropy generated in the file location
```
findinpath: Can we also link `LocationProviders` code snippet https://github.com/apache/iceberg/blob/849f218d8d9a6783ba04d2d4eafc07754ec3885c/core/src/main/java/org/apache/iceberg/LocationProviders.java#L131-L136 here ?
These values may seem rather mystical to a naive reader of the `trino-iceberg` code.

findinpath:
> does this change require any docs update?

>> this was mostly an implementation detail.

The functionality stays unchanged. The only benefit is _potentially_ improved object store layout for AWS S3.
I think that using a consistent object layout scheme to what `apache/iceberg` is using is worth calling in RN.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
```diff
@@ -40,12 +43,14 @@ public class ObjectStoreLocationProvider
...
     private final String storageLocation;
     private final String context;
+    private final boolean includePartitionPaths;
 
     public ObjectStoreLocationProvider(String tableLocation, Map<String, String> properties)
     {
         this.storageLocation = stripTrailingSlash(dataLocation(properties, tableLocation));
         // if the storage location is within the table prefix, don't add table and database name context
         this.context = storageLocation.startsWith(stripTrailingSlash(tableLocation)) ? null : pathContext(tableLocation);
+        this.includePartitionPaths = propertyAsBoolean(properties, WRITE_OBJECT_STORE_PARTITIONED_PATHS, WRITE_OBJECT_STORE_PARTITIONED_PATHS_DEFAULT);
```
findinpath, line 48: nit: skimming through the code, I see that the previous maintainers opted for `TableProperties.` approach. Consider a consistent referencing scheme within the class for the constants from `TableProperties` class.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
```diff
@@ -40,12 +43,14 @@ public class ObjectStoreLocationProvider
     private static final int ENTROPY_DIR_DEPTH = 3;
     private final String storageLocation;
     private final String context;
+    private final boolean includePartitionPaths;
```
findinpath, line 41: Given that we don't have a dedicated table property in the Trino Iceberg connector, I'm advocating for a compatibility product tests in spark to ensure end to end that Trino does honor the `write.object-storage.partitioned-paths` table property.

The unit test is fine, but it doesn't give us full certainty that the entire write operation (and eventual read operation afterwards) succeeds through the query engine.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
ebyhr: We already have a link here to the class. This is sufficient to me:

https://github.com/trinodb/trino/blob/400e08134b63fbc9d57e9f77b4cc5444546f4bdb/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java#L30

findepi:
> > the hash is derived from partitions, or is effectively a random number?
> 
> It's derived from partitions + file names. Since the file name includes the query ID and a UUID, it's effectively random.

if it's effectively just random, i don't get the point of trying to derive it from anything. It's confusing.

> > does this change require any docs update?
> 
> Good question. I didn't update the docs because I thought this was mostly an implementation detail.

in a sense it's not. when using this you may also need to talk to AWS support and have them tweak your S3 bucket configuration.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/ObjectStoreLocationProvider.java
```diff
@@ -99,7 +104,33 @@ private static String pathContext(String tableLocation)
 
     private static String computeHash(String fileName)
```
chenjian2664, line 113: Can we have a test for this method?  i,e in blow we testing the location provider, so we have "test" to "1011/1101/0010/00010011" ?

PR #27637, opened 2025-12-12
Clean up dynamic filter configs

raunaqmorarka:
## Description
Remove `enable-large-dynamic-filters` and `dynamic-filtering.small.*` configuration properties since large dynamic filters have been the default for a while.
Rename `dynamic-filtering.large.*` configs to drop `large` since the small/large distinction is no longer relevant.

## Release notes

```markdown
## General
* {{breaking}} `enable-large-dynamic-filters` configuration property and the corresponding system session property `enable_large_dynamic_filters` has been removed. Large dynamic filters are used by default. ({issue}`27637`)
* {{breaking}} `dynamic-filtering.small*` configuration properties are now defunct and must be removed from server configurations. ({issue}`27637`)
* {{breaking}} The previously deprecated `dynamic-filtering.large-broadcast*` configuration properties are now defunct and must be removed from server configurations. ({issue}`27637`).
* `dynamic-filtering.large*` configuration properties have been deprecated in favor of `dynamic-filtering.*`. ({issue}`27637`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/DynamicFilterConfig.java
+++ b/core/trino-main/src/main/java/io/trino/execution/DynamicFilterConfig.java
@@ -51,6 +51,10 @@ import static io.airlift.units.DataSize.Unit.MEGABYTE;
         "dynamic-filtering.small-partitioned.range-row-limit-per-driver",
         "dynamic-filtering.small-partitioned.max-size-per-operator",
         "enable-large-dynamic-filters",
+        "dynamic-filtering.large-broadcast.max-distinct-values-per-driver",
+        "dynamic-filtering.large-broadcast.max-size-per-driver",
+        "dynamic-filtering.large-broadcast.range-row-limit-per-driver",
+        "dynamic-filtering.large-broadcast.max-size-per-operator",
 })
 public class DynamicFilterConfig
 {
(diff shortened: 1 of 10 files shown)


sopel39 (approved):
that makes a lot of sense, thanks

core/trino-main/src/main/java/io/trino/execution/DynamicFilterConfig.java
```diff
@@ -130,244 +125,127 @@ public DynamicFilterConfig setDynamicRowFilterSelectivityThreshold(double dynami
...
-        return this;
-    }
-
-    @Min(0)
-    public int getSmallPartitionedMaxDistinctValuesPerDriver()
-    {
-        return smallPartitionedMaxDistinctValuesPerDriver;
-    }
-
-    @Config("dynamic-filtering.small-partitioned.max-distinct-values-per-driver")
```
findepi, line 194: Shouldn't there be a deprecation period during which config still works but using it triggers a warning?

cc @kokosing

raunaqmorarka: Not when we're outright removing some functionality.
The "small" configs became irrelevant when large DFs were enabled by default more than an year back.

PR #27639, opened 2025-12-12
Add seamless OpenTelemetry Java agent integration

### LINKED ISSUE #27638: Duplicate OpenTelemetry Instances When Using Java Agent
## Current Behavior

When deploying Trino with the [OpenTelemetry Java agent](https://opentelemetry.io/docs/zero-code/java/agent/) for auto-instrumentation (HTTP requests, JDBC calls, etc.) alongside Trino's built-in manual tracing (`tracing.enabled=true`), two separate OpenTelemetry SDK instances run concurrently:

1. **Java Agent SDK**: Auto-instruments libraries (HTTP, JDBC, gRPC, etc.)
2. **Trino TracingModule SDK**: Handles manual instrumentation (query planning, metadata operations, optimizer, etc.)

This creates several problems:

### 1. Duplicate Exporters and Configuration

Both SDK instances independently export spans to backends:
- Java agent exports to configured endpoint (via `OTEL_EXPORTER_OTLP_ENDPOINT` or `-Dotel.exporter.otlp.endpoint`)
- Trino exports to `tracing.exporter.endpoint`

Configuration must be duplicated:
```bash
 Java agent config (via env vars or system properties)
export OTEL_EXPORTER_OTLP_ENDPOINT=https://collector:4317
 OR
java -Dotel.exporter.otlp.endpoint=https://collector:4317 ...

 Plus config.properties
tracing.enabled=true
tracing.exporter.endpoint=https://collector:4317
tracing.exporter.protocol=grpc
```

### 2. Resource Attributes Not Shared

Resource attributes (service name, environment, cluster, namespace, etc.) configured in the Java agent don't apply to Trino's manual spans. This is particularly problematic for:

- **Custom Java agent extensions**: Many organizations have custom agent extensions that detect runtime environment and set resource attributes accordingly
- **Standard attributes**: Kubernetes namespace, pod name, node name, etc.
- **Organization-specific tags**: Team, cost center, deployment stage, etc.

**Example TraceQL query fails to find Trino spans:**
```traceql
{ resource.environment = "production" }  // Only finds auto-instrumented spans
```

Trino's manual spans (`query`, `Metadata.*`, `scheduler`, etc.) don't have these attributes and must be queried separately.

### 3. Trace Correlation Complexity

While spans from both instances share trace IDs (via context propagation), they have different resource attributes, making filtering and grouping more complex:

```traceql
 Hard to write queries that capture both types of spans
{
  (instrumentation.scope.name = "trino" && span.name = "query")
  ||
  (resource.service.name = "trino" && span.name = "HTTP POST")
}
```

### 4. Operational Overhead

- Two exporters means double the network traffic to collectors
- Two SDK instances means double the memory and CPU overhead for span processing
- Debugging requires checking two different configurations

## Broader Context: Comprehensive Observability with Java Agent

While Trino's built-in tracing provides valuable manual instrumentation for query planning, metadata operations, and optimizer behavior, the OpenTelemetry Java agent unlocks **comprehensive, automatic observability** across the entire stack:

- **Traces**: Not just Trino's manual spans, but also HTTP clients/servers, JDBC connections, gRPC calls, Kafka producers/consumers, Redis operations, and dozens more automatically instrumented libraries
- **Metrics**: JVM metrics, HTTP server metrics, connection pool metrics, etc.
- **Logs**: Automatic correlation of logs with trace IDs (via context propagation)

**All three signals share the same resource attributes** (service name, environment, cluster, etc.), enabling unified filtering and grouping. **Additionally, traces and logs are correlated** by trace ID and span ID, allowing you to jump from a log line to its corresponding trace.

The Java agent instruments at the library level, capturing:
- Network calls to metastores (Hive, Glue, etc.)
- Object storage operations (S3, GCS, Azure)
- Database queries to coordinators
- Internal HTTP communication between coordinator and workers
- JVM metrics and garbage collection

**Granular Control**: While the default configuration traces everything (which can be excessive), the Java agent provides hooks to [disable specific instrumentations](https://opentelemetry.io/docs/zero-code/java/agent/disable/), allowing users to find the right balance between visibility and overhead.

**Key Benefit**: By integrating with the Java agent's `GlobalOpenTelemetry`, Trino's intentional, business-level spans (query lifecycle, optimizer decisions) join the same telemetry pipeline as infrastructure-level spans (HTTP requests, DB calls). All telemetry signals (traces, metrics, logs) share resource attributes and exporter configuration, providing unified observability.

Without this integration, users must choose: either use Trino's manual tracing OR the Java agent - not both effectively. Fixing this issue would enable using both seamlessly.

## Use Case: Java Agent Extensions

A common pattern in enterprise deployments is to use Java agent extensions that:
1. Detect the runtime environment (Kubernetes, ECS, EC2, etc.)
2. Query metadata APIs to get cluster/namespace/region information
3. Set resource attributes automatically based on discovered metadata
4. Configure endpoints programmatically (e.g. some environments may have a local collector, others have to ship directly to a remote endpoint)

**Example:**
```java
// Custom Java agent extension
public class EnvironmentDetectorExtension implements AutoConfigurationCustomizer {
    @Override
    public void customize(AutoConfigurationCustomizerProvider config) {
        config.addResourceCustomizer((resource, configProperties) -> {
            return resource.merge(Resource.builder()
                .put("environment", detectEnvironment())  // prod/staging/dev
                .put("cluster", getClusterName())         // us-west-2-prod
                .put("namespace", getK8sNamespace())      // data-platform
                .put("team", "data-engineering")
                .build());
        });
        config.addLoggerProviderCustomizer(...);
        config.addTracerProviderCustomizer(...);
        // etc
    }
}
```

These resource attributes should apply to **all spans**, including Trino's manual instrumentation.

## Expected Behavior

When the OpenTelemetry Java agent is present:

1. ✅ Trino should use the globally registered `OpenTelemetry` instance from the agent
2. ✅ All spans (auto-instrumented and manual) should share the same resource attributes
3. ✅ Configuration should be done once via the Java agent
4. ✅ A single exporter should handle all spans

## Proposed Solution

Automatically detect if `GlobalOpenTelemetry` has been initialized (typically by the Java agent). If detected, use that instance; otherwise, fall back to creating a dedicated instance via `TracingModule`.

This should be:
- **Automatic**: No configuration required for the common case
- **Backwards compatible**: Existing deployments without Java agent work unchanged
- **Overridable**: System property to force specific behavior if needed

## Benefits

With this change:
- 🎯 Single OpenTelemetry instance (no duplicates)
- 🎯 Resource attributes from Java agent apply to all Trino spans
- 🎯 Configure once via Java agent (no duplicate config)
- 🎯 Single SSL/TLS configuration
- 🎯 Reduced overhead (one exporter, one SDK)
- 🎯 Simpler queries (all spans have same resources)

## Related

- OpenTelemetry Java Instrumentation: https://opentelemetry.io/docs/zero-code/java/agent/
- GlobalOpenTelemetry: https://www.javadoc.io/doc/io.opentelemetry/opentelemetry-api/latest/io/opentelemetry/api/GlobalOpenTelemetry.html

srstrickland:
Enable automatic integration with OpenTelemetry Java agent while maintaining full backwards compatibility with standalone tracing configuration.

Fixes #27638

Changes
-------
- Add ConditionalTracingModule that auto-detects GlobalOpenTelemetry
- Add GlobalOpenTelemetryModule to bind global instance when detected
- Update Server.java to use ConditionalTracingModule
- Add tests for auto-detection and override behavior
- Update OpenTelemetry documentation with Java agent integration guide

Behavior
--------
When Java agent is present:
- Auto-detects GlobalOpenTelemetry and uses it (logged as "Using GlobalOpenTelemetry")
- All spans share resource attributes from agent
- Configuration done once via agent (OTEL_* env vars)
- Single exporter, single SDK instance

When Java agent is NOT present:
- Falls back to TracingModule (logged as "Using dedicated TracingModule")
- Existing tracing.* configuration works unchanged
- 100% backwards compatible

Override
--------
System property or environment variable to force specific behavior:

  # Force use of global
  -Dtracing.use-global=true
  export TRACING_USE_GLOBAL=true

  # Force use of dedicated TracingModule
  -Dtracing.use-global=false

Testing
-------
Unit tests verify:
- Auto-detection when GlobalOpenTelemetry is initialized
- Fallback when not initialized
- System property override
- Backwards compatibility

Documentation
-------------
- docs/src/main/sphinx/admin/opentelemetry.md: Added Java agent integration section
- docs/src/main/sphinx/develop/opentelemetry-java-agent.md: Developer guide

## Release notes

```markdown
## OpenTelemetry
* Fix integration with opentelemetry javaagent (duplicate instances). ({issue}`27638`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/ConditionalTracingModule.java
+++ b/core/trino-main/src/main/java/io/trino/server/ConditionalTracingModule.java
@@ -19,6 +19,7 @@ import io.airlift.log.Logger;
 import io.airlift.tracing.TracingModule;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.trace.Tracer;
 
 import static java.util.Objects.requireNonNull;
 
@@ -78,18 +79,38 @@ public class ConditionalTracingModule
             return useGlobal;
         }
 
-        // Auto-detect: Check if GlobalOpenTelemetry has been initialized
+        // Auto-detect: Check if GlobalOpenTelemetry has been initialized with a real SDK
         // (typically by OpenTelemetry Java agent)
+        return isGlobalOpenTelemetryInitialized();
+    }
+
+    /**
+     * Checks if GlobalOpenTelemetry has been initialized with a real SDK
+     * (as opposed to the default noop implementation).
+     * <p>
+     * Note: We can't use equals() comparison because GlobalOpenTelemetry.get()
+     * returns an ObfuscatedOpenTelemetry wrapper that doesn't equal OpenTelemetry.noop().
+     * Instead, we check if the tracer it produces is a noop tracer.
+     */
+    private boolean isGlobalOpenTelemetryInitialized()
+    {
         try {
             OpenTelemetry global = GlobalOpenTelemetry.get();
-            boolean isNoop = global.equals(OpenTelemetry.noop());
+            Tracer tracer = global.getTracer("detection-probe");
+
+            // Check if the tracer is the noop implementation
+            // DefaultTracer is the noop tracer class name
+            String tracerClassName = tracer.getClass().getName();
+            boolean isNoop = tracerClassName.contains("DefaultTracer") ||
+                    tracerClassName.contains("NoopTracer") ||
+                    tracerClassName.equals("io.opentelemetry.api.trace.DefaultTracer");
 
             if (!isNoop) {
-                log.info("GlobalOpenTelemetry detected (likely from Java agent)");
+                log.info("GlobalOpenTelemetry has real SDK (tracer: %s)", tracerClassName);
                 return true;
             }
 
-            log.debug("GlobalOpenTelemetry is noop, will use dedicated instance");
+            log.debug("GlobalOpenTelemetry is noop (tracer: %s), will use dedicated instance", tracerClassName);
             return false;
         }
         catch (Exception e) {
--- a/core/trino-main/src/main/java/io/trino/server/GlobalOpenTelemetryModule.java
+++ b/core/trino-main/src/main/java/io/trino/server/GlobalOpenTelemetryModule.java
@@ -15,10 +15,14 @@ package io.trino.server;
 
 import com.google.inject.Binder;
 import com.google.inject.Module;
+import io.airlift.tracing.SpanSerialization.SpanDeserializer;
+import io.airlift.tracing.SpanSerialization.SpanSerializer;
 import io.opentelemetry.api.GlobalOpenTelemetry;
 import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.trace.Span;
 import io.opentelemetry.api.trace.Tracer;
 
+import static io.airlift.json.JsonBinder.jsonBinder;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -60,8 +64,10 @@ public class GlobalOpenTelemetryModule
         Tracer tracer = openTelemetry.getTracer(serviceName, serviceVersion);
         binder.bind(Tracer.class).toInstance(tracer);
 
-        // Note: We don't need to configure exporters, resource attributes, or span processors
-        // here because they are already configured in the global OpenTelemetry instance
-        // (typically by the Java agent or other initialization code)
+        // Register JSON serializers for Span (required for remote task communication)
+        // This is normally done by TracingModule, but we need to do it here since
+        // we're bypassing TracingModule when using GlobalOpenTelemetry
+        jsonBinder(binder).addSerializerBinding(Span.class).to(SpanSerializer.class);
+        jsonBinder(binder).addDeserializerBinding(Span.class).to(SpanDeserializer.class);
     }
 }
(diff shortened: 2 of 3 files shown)


core/trino-main/src/main/java/io/trino/server/ConditionalTracingModule.java
```diff
@@ -78,18 +79,38 @@ private boolean shouldUseGlobalOpenTelemetry()
...
+     */
+    private boolean isGlobalOpenTelemetryInitialized()
+    {
         try {
             OpenTelemetry global = GlobalOpenTelemetry.get();
-            boolean isNoop = global.equals(OpenTelemetry.noop());
+            Tracer tracer = global.getTracer("detection-probe");
+
+            // Check if the tracer is the noop implementation
+            // DefaultTracer is the noop tracer class name
```
kekwan, line 102: I found checking if the `TracerProvider` is noop also works here as it's a singleton and not wrapped in Obfsuscated wrapper like `GlobalOpenTelemetry` 

i.e. `global.getTracerProvider().equals(TracerProvider.noop())`

Once Trino upgrades `opentelemetry` to 1.57.0 and beyond, we can leverage `GlobalOpenTelemetry#isSet()` via
https://github.com/open-telemetry/opentelemetry-java/pull/7819 for a more robust check

PR #27645, opened 2025-12-14
Rename queryText to queryTextRegex variables for naming consistency

gertjanal:
## Description
In this PR: https://github.com/trinodb/trino/pull/27129 I missed one class when a rename was suggested.

Before:
```
   private final Optional<Pattern> originalUserRegex;
   private final Optional<Pattern> authenticatedUserRegex;
   private final Optional<Pattern> sourceRegex;
   private final Optional<Pattern> queryText;
```

After:
```
   private final Optional<Pattern> originalUserRegex;
   private final Optional<Pattern> authenticatedUserRegex;
   private final Optional<Pattern> sourceRegex;
   private final Optional<Pattern> queryTextRegex;
```
## Release notes


plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/SelectorSpec.java
```diff
@@ -44,7 +44,7 @@ public SelectorSpec(
             @JsonProperty("originalUser") Optional<Pattern> originalUserRegex,
             @JsonProperty("authenticatedUser") Optional<Pattern> authenticatedUserRegex,
             @JsonProperty("source") Optional<Pattern> sourceRegex,
-            @JsonProperty("queryText") Optional<Pattern> queryText,
+            @JsonProperty("queryText") Optional<Pattern> queryTextRegex,
```
wendigo, line 47: nit: queryTextRegex in JsonProperty

wendigo (approved):

plugin/trino-resource-group-managers/src/main/java/io/trino/plugin/resourcegroups/SelectorSpec.java
ebyhr, line 47: Should we update `resource-groups.md`? 

https://github.com/trinodb/trino/blob/d854aece4e06c2076b2e6ed4fc3df7ae477dac6b/docs/src/main/sphinx/admin/resource-groups.md?plain=1#L184

https://github.com/trinodb/trino/blob/d854aece4e06c2076b2e6ed4fc3df7ae477dac6b/docs/src/main/sphinx/admin/resource-groups.md?plain=1#L256

ksobolew: To be fair, we have `user`, `userGroup` etc. instead of `userRegex`, `userGroupRegex` etc

ksobolew: So I would say keeping `queryText` as the JSON property name is consistent with the rest

PR #27649, opened 2025-12-15
Remove unnecessary uses of TypeSignature

martint:
TypeSignature is an internal concept, akin to an IR for types. Currently, it's being over-used in places that shouldn't rely on it and should use Type, instead.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/BuiltinFunctionResolver.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/BuiltinFunctionResolver.java
@@ -107,7 +107,7 @@ class BuiltinFunctionResolver
         catch (UncheckedExecutionException e) {
             if (e.getCause() instanceof TrinoException cause) {
                 if (cause.getErrorCode().getCode() == FUNCTION_IMPLEMENTATION_MISSING.toErrorCode().getCode()) {
-                    throw new OperatorNotFoundException(operatorType, ImmutableList.of(fromType), toType.getTypeSignature(), cause);
+                    throw new OperatorNotFoundException(operatorType, ImmutableList.of(fromType), toType, cause);
                 }
                 throw cause;
             }
--- a/core/trino-main/src/main/java/io/trino/metadata/OperatorNotFoundException.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/OperatorNotFoundException.java
@@ -18,7 +18,6 @@ import com.google.common.collect.ImmutableList;
 import io.trino.spi.TrinoException;
 import io.trino.spi.function.OperatorType;
 import io.trino.spi.type.Type;
-import io.trino.spi.type.TypeSignature;
 
 import java.util.List;
 import java.util.Optional;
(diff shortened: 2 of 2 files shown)


core/trino-main/src/test/java/io/trino/execution/TestAddColumnTask.java
```diff
@@ -454,7 +454,7 @@ public void testAddFieldToNotExistingField()
 
         assertTrinoExceptionThrownBy(() -> getFutureValue(executeAddColumn(asQualifiedName(tableName), QualifiedName.of("col", "x", "c"), INTEGER, false, false)))
                 .hasErrorCode(COLUMN_NOT_FOUND)
-                .hasMessageContaining("Field 'x' does not exist within row(a row(b integer))");
+                .hasMessageContaining("Field 'x' does not exist within row(\"a\" row(\"b\" integer))");
```
findepi, line 457: > `Quote row field names`

In commit message, maybe include in which context the quoting gets applied
eg "Quote field names when rendering row type"

core/trino-main/src/main/java/io/trino/type/IpAddressType.java
```diff
@@ -94,6 +94,12 @@ public BlockBuilder createFixedSizeBlockBuilder(int positionCount)
         return new Int128ArrayBlockBuilder(null, positionCount);
     }
 
+    @Override
+    public String getDisplayName()
+    {
+        return "ipaddress";
```
findepi: Why not `StandardTypes.IPADDRESS`?

core/trino-main/src/main/java/io/trino/type/Json2016Type.java
```diff
@@ -41,6 +41,12 @@ public Json2016Type()
         super(new TypeSignature(StandardTypes.JSON_2016), JsonNode.class);
     }
 
+    @Override
+    public String getDisplayName()
+    {
+        return "json2016";
```
findepi: StandardTypes.JSON_2016

core/trino-spi/src/main/java/io/trino/spi/type/P4HyperLogLogType.java
```diff
@@ -49,6 +49,12 @@ public void writeSlice(BlockBuilder blockBuilder, Slice value, int offset, int l
         HYPER_LOG_LOG.writeSlice(blockBuilder, value, offset, length);
     }
 
+    @Override
+    public String getDisplayName()
+    {
+        return "P4HyperLogLog";
```
findepi: `StandardTypes.P4_HYPER_LOG_LOG`?

also, does it matter it's camel case?

findepi (commented):
`Specialize display names for each type`

findepi (commented):
> `Use Type instead of TypeSignature`

commit message to include "Mongo", it's important context

findepi (commented):
`Use Type.getDisplayName for user-visible renderings` lgtm

findepi (approved):
lgtm % earlier comments

core/trino-main/src/main/java/io/trino/type/IpAddressType.java
martint: I took a different approach. Introduced a NAME constant in the type and used it here and in the constructor.

PR #27610, opened 2025-12-10
Vectorize hashing for Exchange and Local Exchange Operators

pgandhi999:
## Description
Previously, there was work done in OSS Trino by @pettyjamesm to implement a vectorized approach for combined columnar hash calculation generation via codegen(PR: https://github.com/trinodb/trino/pull/19302) for FlatGroupByHash operator. This PR extends the work for Partitioned Exchange and Local Exchange Operators.

The results from running `BenchmarkPartitionedOutputOperator.verifyAddPage` for varying number of columns are summarized below:

**Environment**: MacOS Local Machine
**Partition Count**: 256
**Runtime JDK**: Java 25
**Page Count**: 5000

Number of Columns | Baseline(InterpretedHashGenerator)seconds | Prototype(Codegen based Vectorized HashGenerator)seconds | % Gain
-- | -- | -- | --
1 | 3.1 | 2.6 | 16.12903
5 | 5.4 | 3.8 | 29.62963
10 | 8 | 6.3 | 21.25
50 | 40 | 28 | 30
100 | 93 | 69 | 25.80645
500 | 603 | 343 | 43.11774
1000 | 1065 | 623 | 41.50235
  |   |   |  

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

PR #27657, opened 2025-12-15
Optimize hash code generation for exchanges

raunaqmorarka:
## Description
Optimize InterpretedHashGenerator using byte code generation and batched loops.
This also brings optimised handling of dictionaries to FlatGroupByHash

## Release notes

```markdown
## General
* Improve performance of queries with data exchanges or aggregations. ({issue}`27657`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/FlatHashStrategyCompiler.java
+++ b/core/trino-main/src/main/java/io/trino/operator/FlatHashStrategyCompiler.java
@@ -28,7 +28,6 @@ import io.airlift.bytecode.MethodDefinition;
 import io.airlift.bytecode.Parameter;
 import io.airlift.bytecode.Scope;
 import io.airlift.bytecode.Variable;
-import io.airlift.bytecode.control.ForLoop;
 import io.airlift.bytecode.control.IfStatement;
 import io.airlift.bytecode.expression.BytecodeExpression;
 import io.trino.annotation.UsedByGeneratedCode;
@@ -36,7 +35,6 @@ import io.trino.cache.CacheStatsMBean;
 import io.trino.operator.scalar.CombineHashFunction;
 import io.trino.spi.block.Block;
 import io.trino.spi.block.BlockBuilder;
-import io.trino.spi.block.RunLengthEncodedBlock;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.TypeOperators;
 import io.trino.sql.gen.CallSiteBinder;
@@ -45,11 +43,7 @@ import org.weakref.jmx.Nested;
 
 import java.lang.invoke.MethodHandle;
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
-import java.util.Objects;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Iterables.getOnlyElement;
@@ -61,14 +55,12 @@ import static io.airlift.bytecode.Access.a;
 import static io.airlift.bytecode.Parameter.arg;
 import static io.airlift.bytecode.ParameterizedType.type;
 import static io.airlift.bytecode.expression.BytecodeExpressions.add;
-import static io.airlift.bytecode.expression.BytecodeExpressions.and;
 import static io.airlift.bytecode.expression.BytecodeExpressions.constantBoolean;
 import static io.airlift.bytecode.expression.BytecodeExpressions.constantFalse;
 import static io.airlift.bytecode.expression.BytecodeExpressions.constantInt;
 import static io.airlift.bytecode.expression.BytecodeExpressions.constantLong;
 import static io.airlift.bytecode.expression.BytecodeExpressions.constantNull;
 import static io.airlift.bytecode.expression.BytecodeExpressions.constantTrue;
-import static io.airlift.bytecode.expression.BytecodeExpressions.equal;
 import static io.airlift.bytecode.expression.BytecodeExpressions.inlineIf;
 import static io.airlift.bytecode.expression.BytecodeExpressions.invokeDynamic;
 import static io.airlift.bytecode.expression.BytecodeExpressions.invokeStatic;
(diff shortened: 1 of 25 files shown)


raunaqmorarka:
Screenshot 2025-12-16 at 11 52 52 AM
Screenshot 2025-12-16 at 11 51 28 AM

core/trino-main/src/main/java/io/trino/operator/exchange/PartitioningExchanger.java
```diff
@@ -58,9 +59,12 @@ public void accept(Page page)
...
         Page partitionPage = partitionedPagePreparer.apply(page);
         // assign each row to a partition. The assignments lists are all expected to cleared by the previous iterations
+        if (partitions.length < partitionPage.getPositionCount()) {
+            partitions = new int[partitionPage.getPositionCount()];
+        }
+        partitionFunction.getPartitions(partitionPage, 0, partitionPage.getPositionCount(), partitions);
         for (int position = 0; position < partitionPage.getPositionCount(); position++) {
-            int partition = partitionFunction.getPartition(partitionPage, position);
-            partitionAssignments[partition].add(position);
+            partitionAssignments[partitions[position]].add(position);
```
pettyjamesm, line 67: There's probably an improvement to be gained by fusing this loop with `partitionAssignments.getPartitions`, but we can worry about that as a follow up.

core/trino-main/src/main/java/io/trino/operator/HashGenerator.java
```diff
@@ -21,9 +21,25 @@ public interface HashGenerator
...
+
     default int getPartition(int partitionCount, int position, Page page)
     {
         long rawHash = hashPosition(position, page);
+        return processRawHash(rawHash, partitionCount);
+    }
+
+    default void getPartitions(int partitionCount, int positionOffset, Page page, int length, int[] partitions)
+    {
+        long[] hashes = new long[length];
```
pettyjamesm, line 34: Reusing this array is likely to be beneficial for most operator use cases, since allocating a new instance on each invocation is non-trivial allocation pressure. Having this default implementation seems like a performance hazard

core/trino-main/src/main/java/io/trino/operator/InterpretedHashGenerator.java
```diff
@@ -95,13 +135,74 @@ private long nullSafeHash(int operatorIndex, Block block, int position)
...
+            }
+        }
+    }
+
+    private static boolean isDictionaryProcessingFaster(DictionaryBlock dictionaryBlock, int length)
+    {
+        // if the dictionary block length is greater than the number of elements in the dictionary,
+        // it will be faster to compute hash for the dictionary values only once and re-use it
+        // instead of recalculating it.
+        return length > dictionaryBlock.getDictionary().getPositionCount();
```
pettyjamesm: I'm betting the actual threshold here is more like `length >= (dictionaryBlock.getDictionary().getPositionCount() * 2)` since the overheads associated with this strategy are non-trivial.

core/trino-main/src/main/java/io/trino/operator/FlatHashStrategyCompiler.java
```diff
@@ -156,15 +150,21 @@ public static FlatHashStrategy compileFlatHashStrategy(List<Type> types, TypeOpe
                 type(Object.class),
                 type(FlatHashStrategy.class));
 
-        // the 'types' field is not used, but it makes debugging easier
-        // this is an instance field because a static field doesn't seem to show up in the IntelliJ debugger
-        FieldDefinition typesField = definition.declareField(a(PRIVATE, FINAL), "types", type(List.class, Type.class));
+        FieldDefinition hashGeneratorField = definition.declareField(a(PRIVATE, FINAL), "hashGenerator", type(InterpretedHashGenerator.class));
```
pettyjamesm: Seems like the responsibility of having a `HashGenerator` should live in `FlatHash` instead of `FlatHashStrategy`. I only added it to `FlatHashStrategy` in the first place because it needed code generation, but all of the logic is columnar and doesn't really overlap with the rest of the "flat" generated logic.

core/trino-main/src/main/java/io/trino/operator/NullSafeHashCompiler.java
```diff
@@ -175,7 +178,52 @@ private static void generateHashBlocksBatched(String methodName, ClassDefinition
...
+
+        BytecodeBlock computeHashLoop = new BytecodeBlock()
+                .append(mayHaveNull.set(valueBlock.invoke("mayHaveNull", boolean.class)))
+                .append(new ForLoop("for (int index = 0; index < length; index++)")
+                        .initialize(index.set(constantInt(0)))
+                        .condition(lessThan(index, length))
+                        .update(index.increment())
+                        .body(new BytecodeBlock()
+                                // position = dictionaryBlock.getUnderlyingValuePosition(offset + index)
+                                .append(position.set(dictionaryBlock.invoke("getUnderlyingValuePosition", int.class, add(offset, index))))
```
pettyjamesm: You might get a small performance lift by pulling `DictionaryBlock#getRawIds()` out of the loop body and accessing the array directly

pgandhi999:
I also got a chance to run the micro benchmark that I had performed for my PR(https://github.com/trinodb/trino/pull/27610) on your PR and the results align with what was stated on my PR. Thank you @raunaqmorarka.

core/trino-main/src/main/java/io/trino/operator/exchange/PartitioningExchanger.java
raunaqmorarka: That is better explored as a follow-up, I expect that might need a different method in `PartitionFunction`

core/trino-main/src/main/java/io/trino/operator/HashGenerator.java
raunaqmorarka: The allocation here is very short-lived and the JVM is pretty good at optimizing for that. Trying to reuse array adds some complexity as it has to be passed down from the calling operator, where it potentially needs to be tracked as a retained memory allocation. Since we didn't observe a problem with this in production for a while, I'm inclined to keep it simple for now and explore reuse as a follow-up.

core/trino-main/src/main/java/io/trino/operator/InterpretedHashGenerator.java
raunaqmorarka: I've updated this to `length > dictionaryBlock.getDictionary().getPositionCount() * 1.2`

core/trino-main/src/main/java/io/trino/operator/NullSafeHashCompiler.java
raunaqmorarka: done

findepi (approved):

core/trino-main/src/main/java/io/trino/operator/FlatHashStrategyCompiler.java
raunaqmorarka: I've removed `hashBlocksBatched` from FlatHashStrategy interface now and used HashGenerator directly in FlatGroupByHash instead of via delegation from generated code. I still want to inject NullSafeHashCompiler in FlatHashStrategyCompiler and create the HashGenerator from there to avoid having to pass NullSafeHashCompiler argument in too many places.

PR #27664, opened 2025-12-16
Cleanup around ServerInfo

losipiuk:
## Release notes

### DIFF
--- a/core/trino-main/src/test/java/io/trino/server/TestServerInfoSerialization.java
+++ b/core/trino-main/src/test/java/io/trino/server/TestServerInfoSerialization.java
@@ -1,0 +1,39 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


client/trino-client/src/main/java/io/trino/client/ServerInfo.java
```diff
@@ -133,3 +134,4 @@ public String toString()
                 .toString();
     }
 }
+
```
wendigo: Drop empty line

client/trino-client/src/main/java/io/trino/client/ServerInfo.java
```diff
@@ -101,22 +101,23 @@ public Optional<String> getNodeId()
     @Override
     public boolean equals(Object o)
     {
-        if (this == o) {
-            return true;
-        }
```
martint, line 106: Why? It's a good short-circuit

findepi: it's interesting that IntelliJ-generated equals doesn't include it (it looks exactly as the code in this PR).
I typically would use IntelliJ-generated unless there is a reason not to.
Wonder why they don't include this short-circuit check. Is it because it adds one more branch to the code?

losipiuk: Yeah - the I just regenerated equals with intellij - did not notice that it does not include this check. I agree it generally makes sense to have it - but given default intellij template does not include it we are for sure missing it in many places.

findepi: like here - https://github.com/trinodb/trino/pull/27757#discussion_r2652909450 :P

PR #27672, opened 2025-12-17
Add support for column aliases and comments in CREATE VIEW statements

hqbhoho:
## Description
Add support for the following syntax:
```
| CREATE (OR REPLACE)? VIEW qualifiedName
        columnComments?
        (COMMENT string)?
        (SECURITY (DEFINER | INVOKER))?
        (WITH properties)? AS rootQuery                                #createView
        
columnComments
    : '(' columnComment (',' columnComment)* ')'
    ;

columnComment
    : qualifiedName (COMMENT string)?
    ;        
```
So we can create view with column alias and comment as follow:
```sql
CREATE VIEW test (id , name) AS SELECT 123 , '456';
CREATE VIEW test (id , name) AS SELECT 123 as id1 , '456' as name1;
CREATE VIEW test (id COMMENT 'id', name COMMENT 'name') AS SELECT 123 , '456';
CREATE VIEW test (id COMMENT 'id', name ) AS SELECT 123 , '456';
// Maybe allow duplicate column name in query body?
CREATE VIEW test (id COMMENT 'id', name COMMENT 'name') AS SELECT 123 as id , '456' as id;
```
And `SHOW CREATE VIEW` will  expose these column alias and comment. 

## Release notes

```markdown
## General
* Add support for column aliases and comments in `CREATE VIEW` statements. ({issue}`issuenumber`)
```

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
@@ -2155,7 +2155,7 @@ public abstract class BaseConnectorTest
 
         // test SHOW CREATE VIEW
         String expectedSql = formatSqlText(format(
-                "CREATE VIEW %s.%s.%s SECURITY %s AS %s",
+                "CREATE VIEW %s.%s.%s (x, y) SECURITY %s AS %s",
                 getSession().getCatalog().get(),
                 getSession().getSchema().get(),
                 viewName,
@@ -2184,6 +2184,52 @@ public abstract class BaseConnectorTest
         assertUpdate("DROP VIEW IF EXISTS " + viewName);
         String ddl = format(
                 "CREATE VIEW %s.%s.%s SECURITY DEFINER AS\n" +
+                        "SELECT *\n" +
+                        "FROM\n" +
+                        "  (\n" +
+                        " VALUES \n" +
+                        "     ROW(1, 'one')\n" +
+                        "   , ROW(2, 't')\n" +
+                        ")  t (\"1_col1\", col2)",
+                getSession().getCatalog().get(),
+                getSession().getSchema().get(),
+                viewName);
+        String showCreateSql = format(
+                "CREATE VIEW %s.%s.%s (\n" +
+                        "   \"1_col1\",\n" +
+                        "   col2\n" +
+                        ") SECURITY DEFINER AS\n" +
+                        "SELECT *\n" +
+                        "FROM\n" +
+                        "  (\n" +
+                        " VALUES \n" +
+                        "     ROW(1, 'one')\n" +
+                        "   , ROW(2, 't')\n" +
+                        ")  t (\"1_col1\", col2)",
+                getSession().getCatalog().get(),
+                getSession().getSchema().get(),
+                viewName);
+        assertUpdate(ddl);
+
+        assertThat(computeScalar("SHOW CREATE VIEW " + viewName)).isEqualTo(showCreateSql);
+
+        assertUpdate("DROP VIEW " + viewName);
+    }
+
+    @Test
+    public void testShowCreateViewWithColumnComment()
+    {
+        skipTestUnless(hasBehavior(SUPPORTS_CREATE_VIEW));
+        checkState(getSession().getCatalog().isPresent(), "catalog is not set");
+        checkState(getSession().getSchema().isPresent(), "schema is not set");
+
+        String viewName = "test_show_create_view" + randomNameSuffix();
+        assertUpdate("DROP VIEW IF EXISTS " + viewName);
+        String ddl = format(
+                "CREATE VIEW %s.%s.%s (\n" +
+                        "   id COMMENT 'id',\n" +
+                        "   name COMMENT 'name'\n" +
+                        ") SECURITY DEFINER AS\n" +
                         "SELECT *\n" +
                         "FROM\n" +
                         "  (\n" +
--- a/core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
+++ b/core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
@@ -60,6 +60,7 @@ import io.trino.sql.tree.AllColumns;
 import io.trino.sql.tree.AstVisitor;
 import io.trino.sql.tree.BooleanLiteral;
 import io.trino.sql.tree.Cast;
+import io.trino.sql.tree.ColumnComment;
 import io.trino.sql.tree.ColumnDefinition;
 import io.trino.sql.tree.CreateMaterializedView;
 import io.trino.sql.tree.CreateMaterializedView.WhenStaleBehavior;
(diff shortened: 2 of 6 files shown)


ebyhr:
The logic around duplicate column names handling looks wrong. The below CRATE VIEW should fail:
```sql
CREATE VIEW duplicate_columns (regionkey, regionkey2) AS SELECT regionkey, regionkey FROM region;

TABLE duplicate_columns;
line 1:1: View 'memory.default.duplicate_columns' is stale or in invalid state: column [regionkey] of type bigint projected from query view at position 1 has a different name from column [regionkey2] of type bigint stored in view definition
```

hqbhoho:
@ebyhr Thanks for your feedback
The same SQL query that fails with a `Duplicate column name` error in Hive  but executes successfully in StarRocks.
Currently, there is indeed an issue: if the aliased column name does not match the name defined in the query, it will cause an execution error when query this view. 
What I'd like to confirm is, if we support aliased columns in views, can we then allow duplicate column names in a query?

hqbhoho:
@martint @kasiafi @Praveen2112 PTAL at your convenience. Thanks!

findepi:
> The logic around duplicate column names handling looks wrong. The below CRATE VIEW should fail:
> 
> ```sql
> CREATE VIEW duplicate_columns (regionkey, regionkey2) AS SELECT regionkey, regionkey FROM region;
> ```

It looks like this should work...




> ```sql
> ABLE duplicate_columns;
> line 1:1: View 'memory.default.duplicate_columns' is stale or in invalid state: column [regionkey] of type bigint projected from query view at position 1 has a different name from column [regionkey2] of type bigint stored in view definition
> ```


... of course, the created view should work too.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorViewDefinition.java
```diff
@@ -44,7 +45,8 @@ public ConnectorViewDefinition(
             @JsonProperty("comment") Optional<String> comment,
             @JsonProperty("owner") Optional<String> owner,
             @JsonProperty("runAsInvoker") boolean runAsInvoker,
-            @JsonProperty("path") List<CatalogSchemaName> path)
+            @JsonProperty("path") List<CatalogSchemaName> path,
+            @JsonProperty("withAlias") boolean withAlias)
```
findepi: What is the meaning of this field?

core/trino-spi/pom.xml
```diff
@@ -581,6 +581,13 @@
...
                                 </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.numberOfParametersChanged</code>
+                                    <old>method void io.trino.spi.connector.ConnectorViewDefinition::&lt;init&gt;(java.lang.String, java.util.Optional&lt;java.lang.String&gt;, java.util.Optional&lt;java.lang.String&gt;, java.util.List&lt;io.trino.spi.connector.ConnectorViewDefinition.ViewColumn&gt;, java.util.Optional&lt;java.lang.String&gt;, java.util.Optional&lt;java.lang.String&gt;, boolean, java.util.List&lt;io.trino.spi.connector.CatalogSchemaName&gt;)</old>
+                                    <new>method void io.trino.spi.connector.ConnectorViewDefinition::&lt;init&gt;(java.lang.String, java.util.Optional&lt;java.lang.String&gt;, java.util.Optional&lt;java.lang.String&gt;, java.util.List&lt;io.trino.spi.connector.ConnectorViewDefinition.ViewColumn&gt;, java.util.Optional&lt;java.lang.String&gt;, java.util.Optional&lt;java.lang.String&gt;, boolean, java.util.List&lt;io.trino.spi.connector.CatalogSchemaName&gt;, boolean)</new>
```
findepi: Backward incompatible SPI change should be avoided. Here it looks avoidable.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -4708,6 +4739,34 @@ public void testCommentViewColumn()
...
+    public void testCreateViewWithColumnAlias()
+    {
+        skipTestUnless(hasBehavior(SUPPORTS_CREATE_VIEW));
+
+        try (TestView view = new TestView(getQueryRunner()::execute, "test_create_view_with_column_alias", "(test_alias1, test_alias2, test_alias3)", "SELECT * FROM region")) {
+            assertQuery("SELECT test_alias1, test_alias2, test_alias3 FROM " + view.getName() + "", "SELECT * FROM region");
+        }
+
+        // test with query with duplicate column name
+        try (TestView view = new TestView(getQueryRunner()::execute, "test_create_view_with_column_alias", "(test_alias1, test_alias2, test_alias3)", "SELECT regionKey, regionKey, regionKey FROM region")) {
```
findepi: regionKey → regionkey

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -4708,6 +4739,34 @@ public void testCommentViewColumn()
         }
     }
 
+    @Test
+    public void testCreateViewWithColumnComment()
```
findepi, line 4948: nit: put this after `testCreateViewWithColumnAlias`
this is because adding column comments also involves defining aliases, so it is a more complicated test case, let's put it after a simpler one

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorViewDefinition.java
hqbhoho: I'm thinking about marking whether a view was created with an explicit column list or not.

Should SHOW CREATE VIEW only show the column aliases if the user actually typed them? Or should we enforce it so that SHOW CREATE VIEW always displays the column list (like (id1, name1)), even if the user didn't specify it originally?
```
-- User input
CREATE VIEW test AS SELECT 123 as id1 , '456' as name1 from test1;

-- Proposed 'Always normalize' output for SHOW CREATE VIEW:
CREATE VIEW test (id1, name1) AS SELECT 123 as id1 , '456' as name1 from test1;
```

core/trino-main/src/main/java/io/trino/execution/CreateViewTask.java
```diff
@@ -112,10 +117,24 @@ else if (metadata.getTableHandle(session, name).isPresent()) {
...
-                .getVisibleFields().stream()
-                .map(field -> new ViewColumn(field.getName().get(), field.getType().getTypeId(), Optional.empty()))
-                .collect(toImmutableList());
+        ImmutableList.Builder<ViewColumn> columns = ImmutableList.builder();
+        Optional<List<ColumnComment>> columnComments = statement.getColumnComments();
+        boolean withAlias = columnComments.isPresent();
+        if (withAlias) {
+            int aliasPosition = 0;
+            for (Field field : analysis.getOutputDescriptor(statement.getQuery()).getVisibleFields()) {
+                ColumnComment columnComment = columnComments.get().get(aliasPosition);
```
chenjian2664, line 126: Is the comment position always match output?
does the statement like: "CREATE VIEW v (a comment 'this is a', c comment 'this is c') AS SELECT a, b, c FROM t" allowed?

hqbhoho: It must ensure the view alias count matches the underlying query's column count.
It will check in `StatementAnalyzer#visitCreateView`.
```
validateColumnComments(node.getColumnComments().get(), queryScope.getRelationType().getVisibleFieldCount(), node);
```
In `TestAnalyzer`, some test as follow: 
```
 assertFails("CREATE VIEW test (id, name) AS SELECT 123")
                .hasErrorCode(MISMATCHED_COLUMN_ALIASES)
                .hasMessageContaining("Column alias list has 2 entries but relation has 1 columns");
```

chenjian2664: So, it either all or nothing for the comments?

hqbhoho: Yes, the current implementation requires consistent column counts for both if column aliases are used.

PR #16202, opened 2023-02-21
Add Exasol connector

kaklakariada:
## Description
This adds a new connector for the analytics database [Exasol](https://www.exasol.com/).

This closes #16083.

## Additional context and related issues
As discussed in https://github.com/trinodb/trino/issues/16011 the connector will be read-only for now.

## Release notes

(x) Release notes are required, with the following suggested text:

```markdown
# General
* Add read-only Exasol connector. ({issue}`16083`)
```

PR #27673, opened 2025-12-17
new Firebird connector

prrvchr:
Here is the new connector for **[Firebird 5.x](https://firebirdsql.org/file/documentation/html/en/refdocs/fblangref50/firebird-50-language-reference.html)**

Firebird has some peculiarities:
- It cannot rename a table.
- It cannot create a table from a SELECT statement.
- It does not support schemas.

Addressing these peculiarities required me to modify the code in two Java classes that this connector inherits from:
- [BaseJdbcConnectorTest.java](https://github.com/trinodb/trino/pull/27673/changes#diff-6d69520925faa8f876d78d002c7ad94202dfc4a81a9758a5bc3f2ba93b40e023)
- [TestingConnectorBehavior.java](https://github.com/trinodb/trino/pull/27673/changes#diff-a49d96a420928d09d76a487a1aadea18526d243ef132fdd297fc4a379e376ec1)

I'm not sure if it could have been done any other way, but in any case, it allows the connector to function correctly despite these peculiarities.

For now, this connector only works in read-only mode, as ebyhr advised me.

## Release notes


ebyhr:
We strongly recommend minimizing the scope of the initial PR. #16202 is a good example, supporting only a few type mappings and operations.

prrvchr:
@ebyhr Here's a version that doesn't do anything more than the Exasol connector. Thanks for your review.

PR #27674, opened 2025-12-17
Vectorize block null bit packing

pettyjamesm:
## Description
Continues to build on support for vectorized block serialization / deserialization introduced in https://github.com/trinodb/trino/pull/26919 and https://github.com/trinodb/trino/pull/27586 by adding support for encoding and decoding the null array itself using the vector API.

This is enabled on all x86_64 platforms with at least 256-bit SIMD register support and all ARM platforms with SVE or SVE2.

### Benchmarks

**Intel Skylake (c5 instance)** : [BenchmarkBlockSerDe](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/3fe5e3acb1c1674a5fe9f379d122035a/raw/a0cb42edd91e219cfd5c1ddb7e35dc870e8fbb7c/baseline_nulls_c5_throughput.json,https://gist.githubusercontent.com/pettyjamesm/3fe5e3acb1c1674a5fe9f379d122035a/raw/a0cb42edd91e219cfd5c1ddb7e35dc870e8fbb7c/vectorized_nulls_c5_throughput.json)
- ~10% throughput improvement for deserialization of `byte`, where the null bits represent a larger proportion of the total work involved. Smaller improvements for all other types.

**Intel Sapphire Rapids (c7i instance)**: [BenchmarkBlockSerDe](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/3fe5e3acb1c1674a5fe9f379d122035a/raw/a0cb42edd91e219cfd5c1ddb7e35dc870e8fbb7c/baseline_nulls_c7i_throughput.json,https://gist.githubusercontent.com/pettyjamesm/3fe5e3acb1c1674a5fe9f379d122035a/raw/9a9606e4e8a543c37f91c92e76439df716fa2f5e/vectorized_nulls_c7i_throughput.json)
- ~20% throughput improvement for deserialization of `byte`, where the null bits represent a larger proportion of the total work involved. Smaller improvements for all other types.

**Graviton 3 (c7g instance)**: [BenchmarkBlockSerDe](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/3b7474bf5030f2c932a730a22abfcf0a/raw/42de8d949267134c3bb23f82245e502f0cc0896f/baseline_nulls_c7g_throughput.json,https://gist.githubusercontent.com/pettyjamesm/3b7474bf5030f2c932a730a22abfcf0a/raw/42de8d949267134c3bb23f82245e502f0cc0896f/vectorized_nulls_c7g_throughput.json)
- ~40% throughput improvement for deserialization of `byte`, where the null bits represent a larger proportion of the total work involved. Smaller improvements for all other types.

**Graviton 4 (c8g instance)**: [BenchmarkBlockSerDe](https://jmh.morethan.io/?sources=https://gist.githubusercontent.com/pettyjamesm/3b7474bf5030f2c932a730a22abfcf0a/raw/42de8d949267134c3bb23f82245e502f0cc0896f/baseline_nulls_c8g_throughput.json,https://gist.githubusercontent.com/pettyjamesm/3b7474bf5030f2c932a730a22abfcf0a/raw/89c4602510accbeb2538f3ee46a3b800d992dff8/vectorized_nulls_c8g_throughput.json)
- ~40% throughput improvement for deserialization of `byte`, where the null bits represent a larger proportion of the total work involved. Smaller improvements for all other types.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
+++ b/core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
@@ -62,7 +62,6 @@ public final class BlockEncodingSimdSupport
         public static final SimdSupport ALL = new SimdSupport(true, true, true, true, true, true, true, true, true);
     }
 
-    private static final int MINIMUM_SIMD_LENGTH = 256;
     private static final SimdSupport AUTO_DETECTED_SUPPORT = detectSimd();
 
     private final SimdSupport simdSupport;
@@ -142,7 +141,7 @@ public final class BlockEncodingSimdSupport
 
         // Unclear which x86_64 platforms would lack 256 bit SIMD registers, so disable vectorization
         // because it's likely that intrinsics are missing or perform worse
-        if (preferredVectorBitWidth < MINIMUM_SIMD_LENGTH) {
+        if (preferredVectorBitWidth < 256) {
             return SimdSupport.NONE;
         }
 


core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -139,15 +140,20 @@ enum X86SimdInstructionSet {
             }
         }
 
+        // Unclear which x86_64 platforms would lack 256 bit SIMD registers, so disable vectorization
+        // because it's likely that intrinsics are missing or perform worse
         if (preferredVectorBitWidth < MINIMUM_SIMD_LENGTH) {
```
raunaqmorarka: Can we rename MINIMUM_SIMD_LENGTH to MINIMUM_SIMD_LENGTH_x86_64 and define another MINIMUM_SIMD_LENGTH_ARM for the `128` there ?
It's a bit confusing to see the static variable here and `128` in the ARM code path

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
```diff
@@ -48,6 +48,7 @@ public final class BlockEncodingSimdSupport
     private static final Logger log = Logger.get(BlockEncodingSimdSupport.class);
 
     public record SimdSupport(
+            boolean packIsNullBits,
```
raunaqmorarka: A separate `packIsNullBits` flag looks strange because this isn't a specific vector instruction. The code is in fact using the same operations as the parquet reader is.
I think just check that SimdSupport is not `NONE` might be sufficient for the code here

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
```diff
@@ -68,16 +72,53 @@ public static void encodeNullsAsBits(SliceOutput sliceOutput, @Nullable boolean[
...
+            return;
+        }
+        // inlined from Objects.checkFromIndexSize
+        if ((length | offset) < 0 || length > isNull.length - offset) {
+            throw new IndexOutOfBoundsException("Invalid offset: %s, length: %s for size: %s".formatted(offset, length, isNull.length));
+        }
+        byte[] packedIsNull = new byte[(length + 7) / 8];
+        int currentByte = 0;
+        int position = 0;
+        for (; position < ByteVector.SPECIES_64.loopBound(length); position += ByteVector.SPECIES_64.length(), currentByte++) {
```
raunaqmorarka: Can we write these for loops without initialiser as while loops ?
I think that would be a bit easier to read

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
pettyjamesm: I think that's a little indirect. Local experimentation shows that there are speedups on platforms that I can't otherwise figure out how to "target" at this point (e.g.: Apple Silicon) so this is something that should have the freedom to evolve with tweaks to the detection logic (and relying on the identity of `SimdSupport.NONE` seems like a generally risky idea)

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
pettyjamesm: Yeah, let me clean these constants up

pettyjamesm: Actually, let's just remove the constants altogether and use the inline literals. The comparison logic is specific to each, but only checked in one place for each. Specifically:
- In x86_64 we're just checking for 256 bit registers because anything else indicates "I have no idea what platform we're running on" if that condition isn't met
- In ARM logic, we're specifically checking the SIMD extension level version for instruction compatibility and only enabling the logic for `long` handling based on whether the register is wide enough (which is not a "minimum for all" but rather a "minimum for long, for ARM")

core/trino-spi/src/main/java/io/trino/spi/block/EncoderUtil.java
pettyjamesm: Rewritten as `while` loops

core/trino-main/src/main/java/io/trino/simd/BlockEncodingSimdSupport.java
raunaqmorarka: If we're going to have the flag, it should be named more generically. Unlike the other cases so far, we're not using any special instructions for null bits packing. Maybe call it `useVectorInstructions` or something similar ?

pettyjamesm: How about `vectorizeNullBitPacking` ?

PR #27677, opened 2025-12-17
Testing: Ensure in Redshift before retrying a CTAS statement that the table gets dropped

findinpath:
## Description

While `CREATE TABLE` is an atomic, `CREATE TABLE AS SELECT` statement happens in two phases for JDBC connectors from Trino:

- creating the table
- inserting content into the table

This change ensures that a test table created via CTAS through Trino, does get dropped to avoid receiving an exception saying that the table already exists when retrying the very same CTAS statement.

Here is a relevant stacktrace gathered from CI that showcases the issue
```
Error:  ....plugin.redshift.TestRedshiftConnectorSmokeTest.testRowLevelDelete -- Time elapsed: 13.15 s <<< ERROR!
io.trino.testing.QueryFailedException: line 1:1: Destination table 'redshift.test_schema.test_row_level_deletepbx2vn9g2n' already exists
	at io.trino.testing.DistributedQueryRunner.execute(DistributedQueryRunner.java:621)
	at io.trino.plugin.redshift.TrinoSqlExecutorWithRetries.lambda$execute$1(TrinoSqlExecutorWithRetries.java:52)
	at dev.failsafe.Functions.lambda$toCtxSupplier$9(Functions.java:228)
	at dev.failsafe.FailsafeExecutor.run(FailsafeExecutor.java:220)
	at io.trino.plugin.redshift.TrinoSqlExecutorWithRetries.execute(TrinoSqlExecutorWithRetries.java:52)
	at io.trino.testing.sql.TestTable.createAndInsert(TestTable.java:52)
	at io.trino.testing.sql.TestTable.<init>(TestTable.java:47)
	at io.trino.plugin.redshift.TestRedshiftConnectorSmokeTest.newTrinoTable(TestRedshiftConnectorSmokeTest.java:47)
	at io.trino.testing.AbstractTestQueryFramework.newTrinoTable(AbstractTestQueryFramework.java:700)
	at io.trino.testing.BaseConnectorSmokeTest.testRowLevelDelete(BaseConnectorSmokeTest.java:291)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:511)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.tryRemoveAndExec(ForkJoinPool.java:1490)
	at java.base/java.util.concurrent.ForkJoinPool.helpJoin(ForkJoinPool.java:2248)
	at java.base/java.util.concurrent.ForkJoinTask.awaitDone(ForkJoinTask.java:499)
	at java.base/java.util.concurrent.ForkJoinTask.join(ForkJoinTask.java:666)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:511)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1450)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2019)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
	Suppressed: java.lang.Exception: SQL: CREATE TABLE test_row_level_deletepbx2vn9g2n AS SELECT * FROM region
		... 24 more
Caused by: io.trino.spi.TrinoException: line 1:1: Destination table 'redshift.test_schema.test_row_level_deletepbx2vn9g2n' already exists
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitCreateTableAsSelect(StatementAnalyzer.java:960)
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitCreateTableAsSelect(StatementAnalyzer.java:533)
	at io.trino.sql.tree.CreateTableAsSelect.accept(CreateTableAsSelect.java:86)
	at io.trino.sql.tree.AstVisitor.process(AstVisitor.java:27)
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.process(StatementAnalyzer.java:552)
	at io.trino.sql.analyzer.StatementAnalyzer.analyze(StatementAnalyzer.java:512)
	at io.trino.sql.analyzer.StatementAnalyzer.analyze(StatementAnalyzer.java:501)
	at io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java:98)
	at io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java:87)
	at io.trino.execution.SqlQueryExecution.analyze(SqlQueryExecution.java:332)
	at io.trino.execution.SqlQueryExecution.<init>(SqlQueryExecution.java:241)
	at io.trino.execution.SqlQueryExecution$SqlQueryExecutionFactory.createQueryExecution(SqlQueryExecution.java:984)
	at io.trino.dispatcher.LocalDispatchQueryFactory.lambda$createDispatchQuery$0(LocalDispatchQueryFactory.java:160)
	at io.trino.$gen.Trino_testversion____20251208_175533_1.call(Unknown Source)
```

Relevant code from the method `testRowLevelDelete()` in `BaseConnectorSmokeTest.java`:

https://github.com/trinodb/trino/blob/d8c80579d60fee53174b3a4014d8d17cac26ac6e/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java#L286

## Release notes

```markdown
## Section
```


testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
```diff
@@ -49,8 +49,9 @@ public TestTable(SqlExecutor sqlExecutor, String namePrefix, @Language("SQL") St
 
     protected void createAndInsert(List<String> rowsToInsert)
     {
-        sqlExecutor.execute(format("CREATE TABLE %s %s", name, tableDefinition));
         try {
+            sqlExecutor.execute(format("CREATE TABLE %s %s", name, tableDefinition));
+
```
findinpath: For the reviewer:

Note that the code will ensure calling `close()` with does `DROP` the table (if it exists or throws an supressed exception otherwise).

```
 try (TestTable ignored = this) {
                throw e;
            }
```


I considered `DROP TABLE IF EXISTS` , but considered that given that the potential exception saying that the table is suppressed when the table does not yet exist should be practically enough.

testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
```diff
@@ -49,8 +49,9 @@ public TestTable(SqlExecutor sqlExecutor, String namePrefix, @Language("SQL") St
 
     protected void createAndInsert(List<String> rowsToInsert)
     {
-        sqlExecutor.execute(format("CREATE TABLE %s %s", name, tableDefinition));
```
findepi, line 52: This was correct way to write this.
if CREATE TABLE fails, there is nothing to drop

findepi:
> While `CREATE TABLE` is an atomic, `CREATE TABLE AS SELECT` statement happens in two phases for JDBC connectors from Trino:
> 
> * creating the table
> * inserting content into the table

If inserting data fails, the table should be rolled back to.
If it's not, it's a product bug, not a  test issue.

testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
findinpath: > if CREATE TABLE fails, there is nothing to drop

As mentioned in the description of this contribution `CREATE TABLE AS SELECT` is not atomic. 

I see that there is registered a rollback action when beginning the creation of the table:

https://github.com/trinodb/trino/blob/8a974f9f667ea6dbc84ae948decc97633ad81473/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java#L1203

However, this doesn't fully ensure that the table will actually be removed.

Going forward on this logic, it may as well happen that neither the statement executed immediately after (the logic that I'm adding) will succeed. :(

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TrinoSqlExecutorWithRetries.java
```diff
@@ -44,11 +40,6 @@ public TrinoSqlExecutorWithRetries(QueryRunner queryRunner, Session session)
     @Override
     public void execute(String sql)
     {
-        Failsafe.with(RetryPolicy.builder()
-                        .handleIf(e -> getCausalChain(e).stream().anyMatch(TestingRedshiftServer::isExceptionRecoverable))
-                        .withDelay(Duration.ofSeconds(10))
-                        .withMaxRetries(3)
-                        .build())
-                .run(() -> queryRunner.execute(session, sql));
+        REDSHIFT_RETRY_EXECUTOR.run(() -> queryRunner.execute(session, sql));
```
findepi: This is "type mismatch"
The failsafe definition in TestingRedshiftServer is redshift jdbc specific (it is declared with the intent that query is sent to redshift directly), while here it's sent to Trino

also, there is a fundamental problem with this -- this puts real production problem under the carpet

- a test can retry its infrastructure ops as many times it wants; test infrastructure is not the product we ship
- code under test is supposed to Just Work™. We don't want to execute Trino queries with retries. Rather, we want trino connectors to retry things they do, if retries are needed.


I understand both these problems are pre-existing to this PR. The `Extract Failsafe Redshift retry executor for reuse` commit makes them just less visible. I'd rather see this commit dropped and replaced with appropriate TODO comments that warn about unhygienic stuff this code does.

testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
findepi: > However, this doesn't fully ensure that the table will actually be removed.
> 
> Going forward on this logic, it may as well happen that neither the statement executed immediately after (the logic that I'm adding) will succeed. :(

indeed, it doesn't guarantee success, as the rollback may fail.
in fact, the current code in this PR doesn't "guarantee" much more -- it just issues a second DROP statement from the test code.

Maybe the rollback action should include retries? (again, a product change, not a test workaround)

findinpath: > Maybe the rollback action should include retries? 

I'm looking into adding retries for dropping the table - this should definitely be a safe operation to retry

findinpath: > Maybe the rollback action should include retries? (again, a product change, not a test workaround)

@findepi thank you for bringing this suggestion.

My concerns about doing retries on dropping destination tables is that they is not a safe operation.
Besides the logic of the operation being rolled-back, there may be as well other concurrent operations which maybe create a namesake table in the meantime and doing a DROP during a retry would leave the system in a (logically) corrupt state.

I'm assuming that use cases like the one above contribute to avoiding to add any retries (except for the `SQLTransientException` handling from `RetryingModule`) in the Trino JDBC connectors.

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TrinoSqlExecutorWithRetries.java
findinpath: > I understand both these problems are pre-existing to this PR. 

Incidentally, I'm the contributor who added the `TrinoSqlExecutorWithRetries.java`.


> this puts real production problem under the carpet

The retry logic is used specifically in setting up the test infrastructure, the "test" tables scaffolding on which the functionality of the connector is definitely tested **without retries**.
These changes are indeed a **compromise** which can lead indeed to hiding potential problems. However, these problems are definitely not linked to the accuracy of the connector.

FWIW, the existing test code of the connector makes already use of constructs like the following:

```
new TestTable(getRedshiftExecutor(), ...)
```
This scaffolding can't be used however in classes like `BaseConnectorTest` or `BaseConnectorSmokeTest` without needing to override the test methods where the SQL employed to create the test table is not compliant with the Redshift dialect.
So, one could say that we're basically making a pragmatic compromise on setting the test tables with retries through Trino without modifying the inherited test methods.


>  I'd rather see this commit dropped and replaced with appropriate TODO comments that warn about unhygienic stuff this code does.

This feels more of a cosmetic change at the moment and I will delay it for now.
Let's settle on whether the contribution's direction is worth investing further into or rather a different solution path needs to be explored.

testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
findepi: > My concerns about doing retries on dropping destination tables is that they is not a safe operation.
> Besides the logic of the operation being rolled-back, there may be as well other concurrent operations which maybe create a namesake table in the meantime and doing a DROP during a retry would leave the system in a (logically) corrupt state.

I am not sure what exact scenario are you concerned about. Can you give an example?

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TrinoSqlExecutorWithRetries.java
findepi: > The retry logic is used specifically in setting up the test infrastructure, the "test" tables scaffolding on which the functionality of the connector is definitely tested **without retries**.

When the setup is a CREATE TABLE that goes thru Trino, then it tests Trino as much as subsequent SELECT does.
If SELECT-based test is flaky, this means Trino is not reliable for its end users.
If test setup is flaky, it means the same -- Trino is not reliable for its end users.

testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
findinpath: t0 
Alice `CREATE TABLE t1 AS SELECT (VALUES (1), (2),(3)) AS t(data);` via Trino
Operation fails while inserting the values in Trino
Trino tries to rollback and DROP t1, but this statement fails (due to connect timeout). Retrying in 10 seconds

t5
Bob sees the table t1, drops it  and creates a new table
Bob `DROP TABLE IF EXISTS t1`
Bob `CREATE TABLE t1 AS SELECT (VALUES (1,'John'), (2, 'Mary'),(3, 'James')) AS t(id,data);`
Operation succeeds

t10
Retry attempt corresponding to the rollbacking operation of Alice is executed
Table t1 dropped successfully

t20
Bob wants to select from `t1`
Table does not exist anymore.

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TrinoSqlExecutorWithRetries.java
findinpath: > Trino is not reliable for its end users

Correct.
Trino, at the time of this writing, does not reliably perform WRITE operations on Redshift.

>  a product change, not a test workaround

I don't have visibility currently how to make the write operations involving ephemeral test tables more reliable and this is why I'm relying on retries to setup the test scaffolding :(

testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java
findepi: great example. If Bob follows this logic, then they are doing something dangerous anyway.
Same actions could interleave differently:

- Alice starts `CREATE TABLE t1 AS SELECT (VALUES (1), (2),(3)) AS t(data);` via Trino.
- Bob sees the table t1, drops it and creates a new table. Operation succeeds
- Alice operation eventually fails. Trino tries to rollback and  issues `DROP t1`.
- Bob wants to select from `t1`, but Table does not exist anymore.

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TrinoSqlExecutorWithRetries.java
findinpath: Continuing on https://github.com/trinodb/trino/pull/27702

findinpath:
Closing in favor of https://github.com/trinodb/trino/pull/27702

PR #27678, opened 2025-12-17
Fixes formatting issue for negated SQL subqueries

### LINKED ISSUE #19695: Trino Formatting Issue with Negation in SQL Queries
I am reaching out because I've encountered an issue with Trino's SQL query formatting when using negation operators. Specifically, it seems that Trino is adding extra parentheses in certain cases, resulting in failures when parsing the SQL query.

```
io.trino.spi.TrinoException: Query does not round-trip
	at io.trino.sql.SqlFormatterUtil.formattingFailure
	at io.trino.sql.SqlFormatterUtil.getFormattedSql
```

I've reviewed the syntax of the query, and is appears to be correct. However, during the formatting process, I notice that additional parentheses are introduced around negated expressions, leading to unexpected results. The same query successfully runs in Presto.

I wanted to check if any community members have experienced similar behavior and if there are known workarounds or solutions to address this formatting issue. I am currently using Trino version 417, and I've checked the documentation, but I couldn't find specific information related to this problem.

If you have any insights, tips, or workarounds that could help resolve or mitigate this issue, I would greatly appreciate your assistance. Additionally, if there are any plans to address this in future releases, please let me know.

JohnFitzpatrick44:
## Description

Fixes #19695 . Addresses issues with round-trips for negated subqueries. Specifically, the formatter adds parentheses around negative unary arithmetic expressions, even if the value being negated already has parentheses around it (i.e. for SQL subqueries). 

This change just checks whether the value being negated is already wrapped in parentheses.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-parser/src/test/java/io/trino/sql/TestSqlFormatter.java
+++ b/core/trino-parser/src/test/java/io/trino/sql/TestSqlFormatter.java
@@ -16,6 +16,7 @@ package io.trino.sql;
 import com.google.common.collect.ImmutableList;
 import io.trino.sql.tree.AddColumn;
 import io.trino.sql.tree.AllColumns;
+import io.trino.sql.tree.ArithmeticUnaryExpression;
 import io.trino.sql.tree.BooleanLiteral;
 import io.trino.sql.tree.ColumnDefinition;
 import io.trino.sql.tree.ColumnPosition;
@@ -52,6 +53,7 @@ import io.trino.sql.tree.ShowSchemas;
 import io.trino.sql.tree.ShowSession;
 import io.trino.sql.tree.ShowTables;
 import io.trino.sql.tree.StringLiteral;
+import io.trino.sql.tree.SubqueryExpression;
 import io.trino.sql.tree.Table;
 import io.trino.sql.tree.Update;
 import io.trino.sql.tree.UpdateAssignment;
@@ -66,6 +68,8 @@ import static io.trino.sql.QueryUtil.selectList;
 import static io.trino.sql.QueryUtil.simpleQuery;
 import static io.trino.sql.QueryUtil.table;
 import static io.trino.sql.SqlFormatter.formatSql;
+import static io.trino.sql.tree.ArithmeticUnaryExpression.Sign.MINUS;
+import static io.trino.sql.tree.ArithmeticUnaryExpression.Sign.PLUS;
 import static io.trino.sql.tree.CreateView.Security.DEFINER;
 import static io.trino.sql.tree.SaveMode.FAIL;
 import static io.trino.sql.tree.SaveMode.IGNORE;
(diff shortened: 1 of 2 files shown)


ebyhr:
Could you please add a test?

JohnFitzpatrick44:
@ebyhr I meant to ask where a good place to add the test would be. I know Trino typically leans hard on integration tests, but this seems much better suited for a unit test.

I see that there's a `TestExpressionFormatter` test, but I've tested this change locally by creating a new `TestSqlFormatter` test by hitting the `SqlFormatterUtil.getFormattedSql` method to confirm a query round-trips. Do you have a preference?

testing/trino-tests/src/test/java/io/trino/tests/TestSqlFormatter.java
```diff
@@ -0,0 +1,42 @@
...
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "SELECT +1",
+            "SELECT -1",
+            "SELECT + (SELECT 1)",
+            "SELECT - (SELECT 1)",
+    })
+    public void testArithmeticUnary(String sql)
+    {
+        SqlParser parser = new SqlParser();
```
marcinsbd: nit: how about extract to instance var or static const?

testing/trino-tests/src/test/java/io/trino/tests/TestSqlFormatter.java
```diff
@@ -0,0 +1,42 @@
...
+public class TestSqlFormatter
+{
+    @ParameterizedTest
+    @ValueSource(strings = {
+            "SELECT +1",
+            "SELECT -1",
+            "SELECT + (SELECT 1)",
+            "SELECT - (SELECT 1)",
+    })
+    public void testArithmeticUnary(String sql)
```
marcinsbd: nit:  package private

testing/trino-tests/src/test/java/io/trino/tests/TestSqlFormatter.java
```diff
@@ -0,0 +1,42 @@
...
+
+import io.trino.sql.SqlFormatter;
+import io.trino.sql.parser.SqlParser;
+import io.trino.sql.tree.Statement;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
+
+public class TestSqlFormatter
```
marcinsbd: nit:  package private final

testing/trino-tests/src/test/java/io/trino/tests/TestSqlFormatter.java
```diff
@@ -0,0 +1,49 @@
...
+        String formattedSql = SqlFormatter.formatSql(statement);
+        Statement roundTripStatement = parser.createStatement(formattedSql);
+
+        assertThat(roundTripStatement).isEqualTo(statement);
+    }
+
+    private static Stream<String> sqlQueries()
+    {
+        return Stream.of(
+                "SELECT +1",
```
marcinsbd: just curious if the following would be a valid case:
(SELECT 1) + SELECT
(SELECT 1) - SELECT

JohnFitzpatrick44: The SqlFormatter seems to correctly throw a syntax error for these cases

testing/trino-tests/src/test/java/io/trino/tests/TestSqlFormatter.java
```diff
@@ -0,0 +1,49 @@
...
+final class TestSqlFormatter
+{
+    private static final SqlParser parser = new SqlParser();
+
+    @ParameterizedTest
+    @MethodSource("sqlQueries")
+    void testArithmeticUnary(String sql)
+    {
+        Statement statement = parser.createStatement(sql);
+        String formattedSql = SqlFormatter.formatSql(statement);
```
wendigo: can we put that test in TestSqlFormatter?

core/trino-parser/src/test/java/io/trino/sql/TestSqlFormatter.java
```diff
@@ -832,4 +836,29 @@ public void testShowBranches()
                 new ShowBranches(new NodeLocation(1, 1), QualifiedName.of("a"))))
                 .isEqualTo("SHOW BRANCHES FROM TABLE a");
     }
+
+    @Test
+    public void testArithmeticUnary()
```
JohnFitzpatrick44, line 841: Verified that this fails with the current version of master

PR #27689, opened 2025-12-18
RowTransformer class to transform values in nested SqlRow structures

gertjanal:
## Description
This helper class is added to create Java functions that are able to do transformations on SqlRow field values.

## Example
_Given an array of users,  mask every city of the address with "xxxxxx"_

The supported way of doing this could be:
`transform(users, user -> cast(user.id, user.name, user.age, row(user.address.street, user.address.postalcode, NULL) as row(id integer, name varchar, age integer, address row(street varchar, postalcode varchar, city varchar))))`

In my case, the row had much more fields, so the cast was getting way too large.
With the `RowTransformer` class, a Java function can be created like this:

```java
@TypeParameter("E")
@SqlType("array(E)")
@ScalarFunction("mask_city")
public static Block maskCity(
        @TypeParameter("E") RowType rowType,
        @SqlType("array(E)") Block input)
{
    return build(rowType, (RowBlock) input, transformer -> {
        transformer.transform(setNativeValue("xxxxxx"), "address", "city");
    });
}
```

According to the tests, the class is working fine, but I'd like a Trino professional to look at it as I do not fully understand the block magic.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/RowReplaceFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/RowReplaceFunction.java
@@ -1,0 +1,109 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.operator.scalar;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.slice.Slice;
+import io.trino.metadata.SqlScalarFunction;
+import io.trino.spi.block.SqlRow;
+import io.trino.spi.function.BoundSignature;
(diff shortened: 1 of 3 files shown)


core/trino-main/src/main/java/io/trino/operator/scalar/RowReplaceFunction.java
```diff
@@ -0,0 +1,109 @@
...
+                .description("Replace the value of a field in a (nested) row")
+                .signature(Signature.builder()
+                        .rowTypeParameter("T")
+                        .typeVariable("V")
+                        .returnType(new TypeSignature("T"))
+                        .argumentType(new TypeSignature("T"))
+                        .argumentType(VARCHAR.getTypeSignature())
+                        .argumentType(new TypeSignature("V"))
+                        .build())
+                .build());
```
martint, line 64: This function is problematic for a few reasons:
* It breaks all type safety guarantees. In particular, there's no way for the analyzer to validate that V is compatible with the type of the corresponding field in the input row. Since the input type, the path, and the value to substitute are known statically, it should be possible to verify type safety statically (except that we don't currently have infrastructure for doing so)
* It defines yet another navigation syntax with unclear semantics. We already have functions that support field navigation (json_query, etc), so a different syntax is less than ideal. Additionally:
  * what happens when it encounters nested values that are not row types (arrays, maps, scalar values, etc)?
  * how does it deal with field names that contain non-standard characters and that need to be quoted?
  * how does it deal with row types with anonymous field names?

gertjanal: Thanks for the feedback

PR #27696, opened 2025-12-19
 Add allcolumnsearch() function for searching across all text columns

pchandra345:
## Summary
  Implements `allcolumnsearch()` function that searches for a term across all VARCHAR and CHAR columns in a query, similar to New Relic NRDB's allcolumnsearch() function.

  The function expands `allcolumnsearch('term')` into `(col1 LIKE '%term%' OR col2 LIKE '%term%' OR ...)` during query analysis, following Trino's pattern for row filter injection.

  ## Implementation Details
  - Table function implementation using AbstractConnectorTableFunction
  - Analysis phase: Validates search term, identifies VARCHAR/CHAR columns, validates at least one searchable column exists
  - Execution phase: Filters rows using Page.getPositions() for efficient page-level filtering
  - Regex support: Case-insensitive pattern matching with full regex support
  - Universal compatibility: Works across all Trino connectors (no connector-specific code)

  ## Features
  - Case-insensitive regex pattern matching
  - Efficient page-level row filtering
  - Handles NULL values correctly
  - Works with all query contexts: JOINs, subqueries, WHERE, ORDER BY, LIMIT
  - Comprehensive error handling (null/empty search term, invalid regex, no string columns)
  - Cross-connector support (TPCH, PostgreSQL, Iceberg, etc.)
  - Fully qualified name support (system.builtin.allcolumnsearch)

  ## Files Changed
  - core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java - Table function implementation
  - core/trino-main/src/main/java/io/trino/connector/system/SystemConnectorModule.java - Function registration
  - core/trino-main/src/main/java/io/trino/metadata/GlobalFunctionCatalog.java - Processor provider mapping
  - testing/trino-tests/src/test/java/io/trino/tests/TestAllColumnSearchFunction.java - Comprehensive test suite (19 test cases)

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
@@ -1,0 +1,214 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.operator.table;
+
+import com.google.common.collect.ImmutableList;
+import io.airlift.slice.Slice;
+import io.trino.spi.Page;
+import io.trino.spi.TrinoException;
+import io.trino.spi.block.Block;
+import io.trino.spi.connector.ConnectorAccessControl;
+import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.ConnectorTransactionHandle;
+import io.trino.spi.function.table.AbstractConnectorTableFunction;
+import io.trino.spi.function.table.Argument;
+import io.trino.spi.function.table.ConnectorTableFunctionHandle;
+import io.trino.spi.function.table.Descriptor;
+import io.trino.spi.function.table.ScalarArgument;
+import io.trino.spi.function.table.ScalarArgumentSpecification;
+import io.trino.spi.function.table.TableArgument;
+import io.trino.spi.function.table.TableArgumentSpecification;
+import io.trino.spi.function.table.TableFunctionAnalysis;
+import io.trino.spi.function.table.TableFunctionDataProcessor;
+import io.trino.spi.function.table.TableFunctionProcessorProvider;
+import io.trino.spi.type.CharType;
+import io.trino.spi.type.RowType;
+import io.trino.spi.type.Type;
+import io.trino.spi.type.VarcharType;
+
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.regex.Pattern;
+import java.util.regex.PatternSyntaxException;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableList.toImmutableList;
+import static com.google.common.collect.Iterables.getOnlyElement;
+import static io.trino.metadata.GlobalFunctionCatalog.BUILTIN_SCHEMA;
+import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.function.table.ReturnTypeSpecification.GenericTable.GENERIC_TABLE;
+import static io.trino.spi.function.table.TableFunctionProcessorState.Finished.FINISHED;
+import static io.trino.spi.function.table.TableFunctionProcessorState.Processed.usedInputAndProduced;
+import static io.trino.spi.type.BooleanType.BOOLEAN;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static java.util.Objects.requireNonNull;
+
+public class AllColumnSearchFunction
+        extends AbstractConnectorTableFunction
+{
+    public static final String NAME = "allcolumnsearch";
+
+    private static final String TABLE_ARGUMENT_NAME = "INPUT";
+    private static final String SEARCH_TERM_ARGUMENT_NAME = "SEARCH_TERM";
+    private static final String CASE_SENSITIVE_ARGUMENT_NAME = "CASE_SENSITIVE";
+
+    public AllColumnSearchFunction()
+    {
+        super(
+                BUILTIN_SCHEMA,
+                NAME,
+                ImmutableList.of(
+                        TableArgumentSpecification.builder()
+                                .name(TABLE_ARGUMENT_NAME)
+                                .rowSemantics()
+                                .build(),
+                        ScalarArgumentSpecification.builder()
+                                .name(SEARCH_TERM_ARGUMENT_NAME)
+                                .type(VARCHAR)
+                                .build(),
+                        ScalarArgumentSpecification.builder()
+                                .name(CASE_SENSITIVE_ARGUMENT_NAME)
+                                .type(BOOLEAN)
+                                .defaultValue(false)
+                                .build()),
+                GENERIC_TABLE);
+    }
+
+    @Override
+    public TableFunctionAnalysis analyze(
+            ConnectorSession session,
+            ConnectorTransactionHandle transaction,
+            Map<String, Argument> arguments,
+            ConnectorAccessControl accessControl)
+    {
+        // Get search term argument
+        ScalarArgument searchTermArgument = (ScalarArgument) arguments.get(SEARCH_TERM_ARGUMENT_NAME);
+        if (searchTermArgument.getValue() == null) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Search term cannot be null");
+        }
+
+        String searchTerm = ((Slice) searchTermArgument.getValue()).toStringUtf8();
+        if (searchTerm.isEmpty()) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Search term cannot be empty");
+        }
+
+        // Validate regex pattern
+        try {
+            Pattern.compile(searchTerm);
+        }
+        catch (PatternSyntaxException e) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Invalid regex pattern: " + e.getMessage());
+        }
+
+        // Get case_sensitive argument (defaults to false)
+        ScalarArgument caseSensitiveArgument = (ScalarArgument) arguments.get(CASE_SENSITIVE_ARGUMENT_NAME);
+        boolean caseSensitive = (Boolean) caseSensitiveArgument.getValue();
+
+        // Get input table schema
+        TableArgument tableArgument = (TableArgument) arguments.get(TABLE_ARGUMENT_NAME);
+        RowType inputRowType = tableArgument.getRowType();
+        List<RowType.Field> inputFields = inputRowType.getFields();
+
+        // Find all string columns (VARCHAR/CHAR types)
+        ImmutableList.Builder<Integer> stringColumnIndices = ImmutableList.builder();
+        for (int i = 0; i < inputFields.size(); i++) {
+            Type type = inputFields.get(i).getType();
+            if (type instanceof VarcharType || type instanceof CharType) {
+                stringColumnIndices.add(i);
+            }
+        }
+
+        List<Integer> searchableColumns = stringColumnIndices.build();
+        if (searchableColumns.isEmpty()) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "No searchable string columns found in input table");
+        }
+
+        // Request all columns from input
+        ImmutableList.Builder<Integer> requiredColumns = ImmutableList.builder();
+        for (int i = 0; i < inputFields.size(); i++) {
+            requiredColumns.add(i);
+        }
+
+        // Return type is same as input type
+        List<Descriptor.Field> returnedType = inputFields.stream()
+                .map(field -> new Descriptor.Field(field.getName(), Optional.of(field.getType())))
+                .collect(toImmutableList());
+
+        return TableFunctionAnalysis.builder()
+                .requiredColumns(TABLE_ARGUMENT_NAME, requiredColumns.build())
+                .returnedType(new Descriptor(returnedType))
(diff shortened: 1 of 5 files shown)


core/trino-main/src/main/java/io/trino/operator/scalar/AllColumnSearchFunction.java
```diff
@@ -0,0 +1,38 @@
...
+import io.airlift.slice.Slice;
+import io.trino.spi.TrinoException;
+import io.trino.spi.function.Description;
+import io.trino.spi.function.ScalarFunction;
+import io.trino.spi.function.SqlType;
+import io.trino.spi.type.StandardTypes;
+
+import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
+
+public final class AllColumnSearchFunction
```
onkar-r: It'd be good to provide a brief summary and an example usage of this function.

ebyhr:
@pchandra345 I recommend getting consensus on this syntax before doing further work. I'm not sure we really want this function.

pchandra345:
> @pchandra345 I recommend getting consensus on this syntax before doing further work. I'm not sure we really want this function.

Please take a look into this thread:- (https://trinodb.slack.com/archives/CP1MUNEUX/p1765463946141919)

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -5066,6 +5069,9 @@ private void analyzeSelectSingleColumn(
 
         private void analyzeWhere(Node node, Scope scope, Expression predicate)
         {
+            // Expand allcolumnsearch() BEFORE analysis so expanded AST gets properly analyzed
+            predicate = expandAllColumnSearch(predicate, scope);
```
Praveen2112: `StatementAnalyzer` is more of a generic framework for processing an AST, having a customization or hook at this layer makes it more brittle. Functionality like this can be implemented via TableFunctions but I guess it would be a bit more complex as we might need to somehow implement this `OR` of all `LIKE` expression as a part of TableFunctions. 

Semantics on this function would also be a bit tricky - how would we handle the hidden or meta columns ? A scalar function in Trino is not only applicable for filter i.e `WHERE` clause - but it can be used in a project clause as well - how would this behaviour translate during it ? Same for join filter clause or order by clause ?

Trino strictly adheres to ANSI SQL syntax - similar to `exclude_columns` which is not a part of ANSI SQL standard we have implemented it as a table function and the same restriction would be applied for `allcolumnsearch` function as well. 

cc: @martint

pchandra345: - Hidden/meta columns: My current implementation includes all visible columns. Should we explicitly exclude system columns?
  - Non-WHERE contexts: I added validations to prevent use in ORDER BY, JOIN ON  and select clause specifically because the semantics are unclear there.

pchandra345: I'm happy to refactor this as a TableFunction if that's the preferred approach. Could you point me to the `exclude_columns` implementation as a reference? I want to ensure I understand the pattern
  correctly before reworking the implementation.

pchandra345: Please take a look into this thread:- (https://trinodb.slack.com/archives/CP1MUNEUX/p1765463946141919)

findepi:
Did you try implementing this as a table function? Would it be possible?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -6185,4 +6206,181 @@ public List<TableArgumentAnalysis> getTableArgumentAnalyses()
...
+
+    /**
+     * Validator that detects allcolumnsearch() usage in invalid contexts.
+     * allcolumnsearch() is only allowed in WHERE clause predicates.
+     */
+
+    private static class AllColumnSearchValidator
+            extends DefaultExpressionTraversalVisitor<Void>
+    {
+        private final String context;
```
kbhatianr: What's the purpose of this `context`? It doesn't seem to be used anywhere. Do we require it?

pchandra345: the context field at line is used in the error message at line 6238 to provide users with information about which clause contained the invalid allcolumnsearch() usage.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -6185,4 +6206,181 @@ public List<TableArgumentAnalysis> getTableArgumentAnalyses()
...
+
+        public static void validate(Expression expression, String context)
+        {
+            new AllColumnSearchValidator(context).process(expression, null);
+        }
+
+        @Override
+        protected Void visitFunctionCall(FunctionCall node, Void context)
+        {
+            if (node.getName().getSuffix().equalsIgnoreCase("allcolumnsearch")) {
```
kbhatianr: Can you turn the `"allcolumnsearch"` into a constant? It's reused here and on line 6269. 

Two benefits: 
1. Consolidation of the uses
2. String allocation leads to slight performance benefit.

core/trino-main/src/main/java/io/trino/operator/scalar/AllColumnSearchFunction.java
pchandra345: Update the docs

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
pchandra345: Updated

docs/src/main/sphinx/functions/allcolumnsearch.md
```diff
@@ -0,0 +1,104 @@
...
+
+This function can only be used in WHERE clauses and is expanded at query
+analysis time into explicit LIKE predicates for each string column.
+
+**Requirements and behavior:**
+
+- `search_term` must be a constant string literal (not a column or expression)
+- Search is case-sensitive
+- Performs substring matching equivalent to `column LIKE '%search_term%'`
+- NULL values in columns are handled correctly (do not match)
```
kbhatianr: Instead of saying what "correctly" is, I would suggest rephrasing this to "`NULL` values in a column will evaluate to `false`"

pchandra345:
> Did you try implementing this as a table function? Would it be possible?

  Thanks for the suggestion! However, table functions wouldn't work well for this use case because they're connector-specific.

  Table functions are registered per-connector (via Connector.getTableFunctions()) and resolved with a CatalogHandle, meaning we'd need to implement allcolumnsearch() separately for every connector
  (MySQL, PostgreSQL, Iceberg, Hive, BigQuery, etc.).

  The current analyzer-based approach expands allcolumnsearch() to OR'd LIKE expressions during query analysis - before any connector is involved. This means it automatically works for all connectors
   without requiring any connector-specific code. By the time the query reaches the connector, it just sees a standard WHERE clause with LIKE predicates.

  This gives us universal functionality with a single implementation.

ebyhr:
> they're connector-specific

@pchandra345 Have you confirmed [ExcludeColumnsFunction](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/operator/table/ExcludeColumnsFunction.java) in the engine?

pchandra345:
> > they're connector-specific
> 
> @pchandra345 Have you confirmed [ExcludeColumnsFunction](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/operator/table/ExcludeColumnsFunction.java) in the engine?

 Thanks for bringing up ExcludeColumnsFunction! You're right that it's a built-in table function (registered with GlobalSystemConnector) that works across all connectors.

   ExcludeColumnsFunction works as a table function because it:
  - Performs column projection (schema manipulation), not row filtering
  - Naturally belongs in FROM clause (defining what table structure to query)
  - Passes through all rows unchanged (line 150: usedInputAndProduced)

  allcolumnsearch() performs row filtering based on content matching, which is semantically a WHERE clause operation.

  The analyzer-based expansion approach correctly implements WHERE clause semantics while maintaining universal cross-connector support and pushdown optimization potential.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
chenjian2664: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/operator/table/ExcludeColumnsFunction.java the `exclude_columns` function FYI

pchandra345: Thanks for bringing up ExcludeColumnsFunction! You're right that it's a built-in table function (registered with GlobalSystemConnector) that works across all connectors.

ExcludeColumnsFunction works as a table function because it:

Performs column projection (schema manipulation), not row filtering
Naturally belongs in FROM clause (defining what table structure to query)
Passes through all rows unchanged (line 150: usedInputAndProduced)
allcolumnsearch() performs row filtering based on content matching, which is semantically a WHERE clause operation.

The analyzer-based expansion approach correctly implements WHERE clause semantics while maintaining universal cross-connector support and pushdown optimization potential.

Praveen2112: Though the `analyzer-based expansion` works correctly or allows us to maintain the semantics - I'm afraid if it would be correct direction. `StatementAnalyzer` is a critical part or entry point for the planning and it should't be customized for a specific set of functions implementation.  This It could potentially increase the maintenance cost of `StatementAnalyzer` in future. Though `TableFunction` framework has some limitations, I guess implementing this feature as a `TableFunction` would be the preferred direction.

pchandra345: I've implemented allcolumnsearch() as a TableFunction, not using StatementAnalyzer expansion.

  The implementation uses AbstractConnectorTableFunction with execution-time filtering via Page.getPositions(). It's registered through SystemConnectorModule similar to other table functions like
  exclude_columns. No changes to StatementAnalyzer or query planning logic.

pchandra345:
@cla-bot check

core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
```diff
@@ -0,0 +1,212 @@
...
+        {
+            @Override
+            public TableFunctionDataProcessor getDataProcessor(ConnectorSession session, ConnectorTableFunctionHandle handle)
+            {
+                AllColumnSearchFunctionHandle functionHandle = (AllColumnSearchFunctionHandle) handle;
+                String searchTerm = functionHandle.searchTerm();
+                List<Integer> searchableColumns = functionHandle.searchableColumnIndices();
+
+                // Compile regex pattern once for reuse
+                Pattern searchPattern = Pattern.compile(searchTerm, Pattern.CASE_INSENSITIVE);
```
Praveen2112: Should we make this controllable via function arguments ?

core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
```diff
@@ -0,0 +1,212 @@
...
+
+                        // Check each searchable column
+                        for (int columnIndex : searchableColumns) {
+                            Block block = inputPage.getBlock(columnIndex);
+                            if (!block.isNull(position)) {
+                                Slice value = (Slice) VARCHAR.getObject(block, position);
+                                String columnValue = value.toStringUtf8();
+
+                                if (searchPattern.matcher(columnValue).find()) {
+                                    matches = true;
```
Praveen2112: We can inline `selectedPositions` right ?

core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
```diff
@@ -0,0 +1,212 @@
...
+                        return usedInputAndProduced(inputPage.getPositions(selectedPositions, 0, selectedCount));
+                    }
+                };
+            }
+        };
+    }
+
+    public record AllColumnSearchFunctionHandle(String searchTerm, List<Integer> searchableColumnIndices)
+            implements ConnectorTableFunctionHandle
+    {
```
Praveen2112, line 204: Can we add additional checks to ensure searchTerm and searchableColumnIndices are valid.

testing/trino-tests/src/test/java/io/trino/tests/TestAllColumnSearchFunction.java
```diff
@@ -0,0 +1,357 @@
...
+        // Test regex pattern matching
+        assertThat(query(
+                """
+                SELECT name
+                FROM TABLE(allcolumnsearch(
+                    input => TABLE(tpch.tiny.nation),
+                    search_term => '^UNITED.*'))
+                ORDER BY name
+                """))
+                .matches("SELECT name FROM tpch.tiny.nation WHERE REGEXP_LIKE(name, '^UNITED.*', 'i') OR REGEXP_LIKE(comment, '^UNITED.*', 'i') ORDER BY name");
```
Praveen2112: Can you verify the failures ?

Praveen2112 (commented):
Can we squash both the commits to one.

testing/trino-tests/src/test/java/io/trino/tests/TestAllColumnSearchFunction.java
pchandra345: Added a failure test as well

core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
pchandra345: Added some additional checks

core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
pchandra345: Done

core/trino-main/src/main/java/io/trino/operator/table/AllColumnSearchFunction.java
pchandra345: Added

Issue #27700, opened 2025-12-19
Permission issue when writing to Delta tables on S3


findepi:
This PR
-  https://github.com/trinodb/trino/pull/27388 

introduced a requirement for S3 `PutObjectTagging` permission and this is apparently a problem to some users.
It is also unnecessary friction. The S3 exclusive write indeed cannot be safely performed without tagging, but Delta connector knows more about content being written and can compensate that.

BTW Somehow this breaking change did not make it to final release notes, this is being fixed in
- https://github.com/trinodb/trino/pull/27699

Issue #27400, opened 2025-11-21
Writing to S3 sometimes fails with `FileAlreadyExistsException`


findepi:
When `s3.exclusive-create=true` (default), the S3 filesystem uses `If-None-Match: *` for all object uploads.
When upload succeeds on the server, but due to network conditions the AWS SDK S3 client doesn't know about that, it retries the request.
Such retried request then legitimately fails due to write precondition. 
As an effect, writing via `S3OutputFile.create` to a location that did not exist before may fail with `FileAlreadyExistsException`.

findepi:
Issue created for visibility. This is fixed in https://github.com/trinodb/trino/pull/27330

PR #27701, opened 2025-12-19
Recover Delta S3 transaction log atomic write after retries without tags

findepi:
This solves the problem described in detail in https://github.com/trinodb/trino/pull/27388. Previous solution
(that PR) was on S3 filesystem level and required additional permissions
to run. The new solution is on Delta Lake connector level and requires
no permissions. Delta Lake connector can guarantee the transaction log
content to be unique and can use this for determining provenance of the
object suspected of being a conflict with another writer.

- fixes https://github.com/trinodb/trino/issues/27700
- follows https://github.com/trinodb/trino/pull/27388
- follows https://github.com/trinodb/trino/pull/27699

### Release notes

Delta Lake
- Do not require `PutObjectTagging` AWS S3 permission when writing to Delta Lake tables on S3.

### DIFF
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/S3ConditionalWriteLogSynchronizer.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/S3ConditionalWriteLogSynchronizer.java
@@ -14,6 +14,7 @@
 package io.trino.plugin.deltalake.transactionlog.writer;
 
 import com.google.inject.Inject;
+import io.airlift.log.Logger;
 import io.airlift.slice.Slice;
 import io.airlift.slice.Slices;
 import io.trino.filesystem.FileMayHaveAlreadyExistedException;
@@ -35,6 +36,8 @@ import static java.util.Objects.requireNonNull;
 public class S3ConditionalWriteLogSynchronizer
         implements TransactionLogSynchronizer
 {
+    private static final Logger log = Logger.get(S3ConditionalWriteLogSynchronizer.class);
+
     private final DeltaLakeFileSystemFactory fileSystemFactory;
 
     @Inject
@@ -66,7 +69,16 @@ public class S3ConditionalWriteLogSynchronizer
                     Slice readBack = trinoInput.readFully(0, entryContents.length + 1);
                     // The contents are unique per query, so if they match, we know we created it
                     boolean createdByUs = readBack.equals(Slices.wrappedBuffer(entryContents));
-                    if (!createdByUs) {
+                    if (createdByUs) {
+                        // This should be a rare situation. Since we're masking a potential problem, log so that it surfaces if it becomes frequent.
+                        log.warn(
+                                """
+                                Potential conflict falsely detected while writing Transaction Log entry %s to S3. \
+                                This may happen in rare cases when network retries are involved. \
+                                If this happens frequently, consider inspecting network infrastructure between Trino and S3.""",
+                                newLogEntryPath);
+                    }
+                    else {
                         throw new TransactionConflictException("Conflict detected while writing Transaction Log entry %s to S3".formatted(newLogEntryPath), mayExistException);
                     }
                 }


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/S3ConditionalWriteLogSynchronizer.java
```diff
@@ -40,12 +46,36 @@ public class S3ConditionalWriteLogSynchronizer
     @Override
     public void write(ConnectorSession session, VendedCredentialsHandle credentialsHandle, String clusterId, Location newLogEntryPath, byte[] entryContents)
     {
+        // This is key to ensuring that two different queries/clusters can never write the same transaction log entry contents.
+        // Contents-based checks are used to recover when S3 write succeeds on the server, but is retried on the client and the fails.
+        String contentsAsString = new String(entryContents, StandardCharsets.UTF_8);
```
losipiuk: nit: static import `UTF_81

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -441,67 +429,14 @@ static void putObject(
             boolean objectAlreadyExists = putObjectException instanceof S3Exception s3Exception &&
                     s3Exception.statusCode() == HTTP_PRECON_FAILED;
             if (objectAlreadyExists) {
-                UncertainBoolean createdByUs = UncertainBoolean.FALSE;
                 if (exclusiveCreate && firstNonNull(putObjectException.numAttempts(), 0) > 1) {
```
chenjian2664, line 432: Question regarding the second condition:
Why do we use `firstNonNull(putObjectException.numAttempts(), 0) > 1` instead of `> 0`? Is it impossible for this situation to occur on the first retry attempt?

I also noticed that we handle `FileMayHaveAlreadyExistedException` later in the flow. given that, would it make sense to remove this condition entirely and allow more cases to fall through to that check?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/S3ConditionalWriteLogSynchronizer.java
```diff
@@ -40,12 +46,38 @@ public class S3ConditionalWriteLogSynchronizer
...
+            throw new TransactionConflictException("Conflict detected while writing Transaction Log entry %s to S3".formatted(newLogEntryPath), e);
+        }
+        catch (FileMayHaveAlreadyExistedException mayExistException) {
+            // Check if the object was actually created by a previous attempt of AWS SDK's implicit retries
+            // This is workaround for AWS SDK limitation: https://github.com/aws/aws-sdk-java-v2/issues/6580
+            try {
+                try (TrinoInput trinoInput = fileSystem.newInputFile(newLogEntryPath).newInput()) {
+                    Slice readBack = trinoInput.readFully(0, entryContents.length + 1);
+                    // The contents are unique per query, so if they match, we know we created it
+                    boolean createdByUs = readBack.equals(Slices.wrappedBuffer(entryContents));
```
chenjian2664, line 71: can we add a debug-level (or even warning-level) log here so we know when this condition occurs - when the `createdByUs` is true

docs/src/main/sphinx/release/release-479.md
```diff
@@ -53,7 +53,8 @@
 * {{breaking}} Remove support for unauthenticated access when GCS authentication type is set to `SERVICE_ACCOUNT`. ({issue}`26984`)
 * Rename `s3.exclusive-create` config to `delta.s3.transaction-log-conditional-writes.enabled`. ({issue}`27372`)
 * {{breaking}} Require `PutObjectTagging` AWS S3 permission when writing to tables in S3 buckets, 
-  unless `delta.s3.transaction-log-conditional-writes.enabled` is set to `false`. ({issue}`27388`)
+  unless `delta.s3.transaction-log-conditional-writes.enabled` is set to `false`.
+  This requirement is removed in a future release. ({issue}`27388`)
```
chenjian2664, line 57: will it reflected on https://trino.io/docs/current/release.html even the release is cut?

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
findepi: i thought numAttempts is "number of attempts", but I am hearing it might be read as "number of retries" i.e. "number of attempts - 1". Am i reading this correctly?

> I also noticed that we handle `FileMayHaveAlreadyExistedException` later in the flow. given that, would it make sense to remove this condition entirely and allow more cases to fall through to that check?

i still want to raise `FileAlreadyExistsException` in a happy path scenario where we just know that

i intended the code to strike this balance
- report `FileAlreadyExistsException` when we are sure object existed before we started (i.e. exists and not created by us)
- report `FileMayHaveAlreadyExistedException` when we're unsure

docs/src/main/sphinx/release/release-479.md
findepi: It's not the first time we're improving release notes after the release, so i hope the answer is yes. But i am not entirely sure. @martint @electrum will know better.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/S3ConditionalWriteLogSynchronizer.java
findepi: good idea, will add.

LeonidChistov:
@findepi 

I have a question regarding this Pull Request. 

`FileHiveMetastore` implementation has a call to `createExclusive` method, which may happen to be a call to S3 `createExclusive`.

```
try {
    output.createExclusive(json);
}
```

Do I understand correctly that after your revert of https://github.com/trinodb/trino/pull/27388 we now again have a potential bug in `FileHiveMetastore` with S3?

Also, if I understand correctly, code in https://github.com/trinodb/trino/pull/27388 could have been used to fix original issue https://github.com/trinodb/trino/issues/27400 and thus we won't have to remove logic related to `s3.create-exclusive` flag.

Your main motivation to do this revert was to avoid requirement of additional S3 permissions, correct?

Would appreciate any comments from your side.

findepi:
> Do I understand correctly that after your revert of #27388 we now again have a potential bug in `FileHiveMetastore` with S3?



Note that `FileHiveMetastore` cannot be safely used in multi-actor (multi-writer or read-writer) scenarios. It also does not provide no cross-version compatibility. It isn't really a recommended deployment model

You're, however, correct, that it may now fail also in single-cluster scenario. This is something I overlooked.
Do you want to file a new issue about this?

> Also, if I understand correctly, code in #27388 could have been used to fix original issue #27400 and thus we won't have to remove logic related to `s3.create-exclusive` flag.

Yes, but at the burden of requiring tag permission.
However, ``s3.create-exclusive`` was a wrong abstraction. Need for exclusive writes is quite rare (in terms of number of objects written) and Trino filesystem has dedicated API for callers to express need for an exclusive write. In retrospect it seems a mistake to add `s3.create-exclusive` flag as a configuration option on the filesystem level.


> Your main motivation to do this revert was to avoid requirement of additional S3 permissions, correct?

correct.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
chenjian2664: Thanks. got it, that means the first time of the putting action is also counted.

PR #27702, opened 2025-12-19
Add rollback action for CTAS destination table

findinpath:
## Description

While `CREATE TABLE` is an atomic operation, `CREATE TABLE AS SELECT` statement happens in two phases
for JDBC connectors from Trino:

- creating the table
- inserting content into the table

If the insertion of the data fails, the CTAS may leave residual empty tables behind.

The JDBC connectors, in dealing with `CREATE TABLE AS SELECT` scenarios,
remove the destination table in case of dealing with exceptions
while inserting the content.

**NOTE** that introducing the `rollbackConsumer`  in the `beginCreateTable` method may come with potential changes in the functionality of the connectors that extend from `trino-base-jdbc` module.

Here is a relevant stacktrace gathered from CI that showcases the issue of leaving behind empty tables after a CTAS 

```
Error:  ....plugin.redshift.TestRedshiftConnectorSmokeTest.testRowLevelDelete -- Time elapsed: 13.15 s <<< ERROR!
io.trino.testing.QueryFailedException: line 1:1: Destination table 'redshift.test_schema.test_row_level_deletepbx2vn9g2n' already exists
	at io.trino.testing.DistributedQueryRunner.execute(DistributedQueryRunner.java:621)
	at io.trino.plugin.redshift.TrinoSqlExecutorWithRetries.lambda$execute$1(TrinoSqlExecutorWithRetries.java:52)
	at dev.failsafe.Functions.lambda$toCtxSupplier$9(Functions.java:228)
	at dev.failsafe.FailsafeExecutor.run(FailsafeExecutor.java:220)
	at io.trino.plugin.redshift.TrinoSqlExecutorWithRetries.execute(TrinoSqlExecutorWithRetries.java:52)
	at io.trino.testing.sql.TestTable.createAndInsert(TestTable.java:52)
	at io.trino.testing.sql.TestTable.<init>(TestTable.java:47)
	at io.trino.plugin.redshift.TestRedshiftConnectorSmokeTest.newTrinoTable(TestRedshiftConnectorSmokeTest.java:47)
	at io.trino.testing.AbstractTestQueryFramework.newTrinoTable(AbstractTestQueryFramework.java:700)
	at io.trino.testing.BaseConnectorSmokeTest.testRowLevelDelete(BaseConnectorSmokeTest.java:291)
	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:511)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.tryRemoveAndExec(ForkJoinPool.java:1490)
	at java.base/java.util.concurrent.ForkJoinPool.helpJoin(ForkJoinPool.java:2248)
	at java.base/java.util.concurrent.ForkJoinTask.awaitDone(ForkJoinTask.java:499)
	at java.base/java.util.concurrent.ForkJoinTask.join(ForkJoinTask.java:666)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:511)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1450)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:2019)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:187)
	Suppressed: java.lang.Exception: SQL: CREATE TABLE test_row_level_deletepbx2vn9g2n AS SELECT * FROM region
		... 24 more
Caused by: io.trino.spi.TrinoException: line 1:1: Destination table 'redshift.test_schema.test_row_level_deletepbx2vn9g2n' already exists
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitCreateTableAsSelect(StatementAnalyzer.java:960)
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.visitCreateTableAsSelect(StatementAnalyzer.java:533)
	at io.trino.sql.tree.CreateTableAsSelect.accept(CreateTableAsSelect.java:86)
	at io.trino.sql.tree.AstVisitor.process(AstVisitor.java:27)
	at io.trino.sql.analyzer.StatementAnalyzer$Visitor.process(StatementAnalyzer.java:552)
	at io.trino.sql.analyzer.StatementAnalyzer.analyze(StatementAnalyzer.java:512)
	at io.trino.sql.analyzer.StatementAnalyzer.analyze(StatementAnalyzer.java:501)
	at io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java:98)
	at io.trino.sql.analyzer.Analyzer.analyze(Analyzer.java:87)
	at io.trino.execution.SqlQueryExecution.analyze(SqlQueryExecution.java:332)
	at io.trino.execution.SqlQueryExecution.<init>(SqlQueryExecution.java:241)
	at io.trino.execution.SqlQueryExecution$SqlQueryExecutionFactory.createQueryExecution(SqlQueryExecution.java:984)
	at io.trino.dispatcher.LocalDispatchQueryFactory.lambda$createDispatchQuery$0(LocalDispatchQueryFactory.java:160)
	at io.trino.$gen.Trino_testversion____20251208_175533_1.call(Unknown Source)
```

Relevant code from the method `testRowLevelDelete()` in `BaseConnectorSmokeTest.java`:

https://github.com/trinodb/trino/blob/d8c80579d60fee53174b3a4014d8d17cac26ac6e/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java#L286

## Relevant issues and PRs

Inspired from the discussions done during https://github.com/trinodb/trino/pull/27677

## Release notes

```markdown
## JDBC
* Drop the table in CTAS scenario in case of an exception during data insertion phase. ({issue}`issuenumber`)
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
@@ -829,12 +829,13 @@ public abstract class BaseJdbcClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
         try {
             if (shouldUseFaultTolerantExecution(session)) {
                 // Create the target table
-                createTable(session, tableMetadata);
+                JdbcOutputTableHandle destinationTableHandle = createTable(session, tableMetadata, tableMetadata.getTable().getTableName());
+                rollbackActionConsumer.accept(() -> rollbackCreateDestinationTable(session, destinationTableHandle.getRemoteTableName()));
                 // Create the temporary table
                 ColumnMetadata pageSinkIdColumn = getPageSinkIdColumn(
                         tableMetadata.getColumns().stream().map(ColumnMetadata::getName).toList());
@@ -847,6 +848,11 @@ public abstract class BaseJdbcClient
         catch (SQLException e) {
             throw new TrinoException(JDBC_ERROR, e);
         }
+    }
+
+    protected void rollbackCreateDestinationTable(ConnectorSession session, RemoteTableName remoteTableName)
+    {
+        dropTable(session, remoteTableName, false);
     }
 
     protected JdbcOutputTableHandle createTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, String targetTableName)
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
@@ -137,7 +137,7 @@ public class DefaultJdbcMetadata
     private final boolean precalculateStatisticsForPushdown;
     private final Set<JdbcQueryEventListener> jdbcQueryEventListeners;
 
-    private final List<Runnable> rollbackActions = new ArrayList<>();
+    protected final List<Runnable> rollbackActions = new ArrayList<>();
 
     public DefaultJdbcMetadata(
             JdbcClient jdbcClient,
@@ -1199,7 +1199,7 @@ public class DefaultJdbcMetadata
         if (replace) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
-        JdbcOutputTableHandle handle = jdbcClient.beginCreateTable(session, tableMetadata);
+        JdbcOutputTableHandle handle = jdbcClient.beginCreateTable(session, tableMetadata, rollbackActions::add);
         rollbackActions.add(() -> jdbcClient.rollbackCreateTable(session, handle));
         return handle;
     }
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
@@ -210,7 +210,7 @@ public interface JdbcClient
 
     void createTable(ConnectorSession session, ConnectorTableMetadata tableMetadata);
 
-    JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata);
+    JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer);
 
     void commitCreateTable(ConnectorSession session, JdbcOutputTableHandle handle, Set<Long> pageSinkIds);
 
--- a/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteClient.java
+++ b/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteClient.java
@@ -83,6 +83,7 @@ import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Strings.isNullOrEmpty;
@@ -381,7 +382,7 @@ public class IgniteClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
         if (tableMetadata.getComment().isPresent()) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support creating tables with table comment");
@@ -417,13 +418,14 @@ public class IgniteClient
 
         try (Connection connection = connectionFactory.openConnection(session)) {
             execute(session, connection, sql);
-
-            return new IgniteOutputTableHandle(
+            IgniteOutputTableHandle destinationTableHandle = new IgniteOutputTableHandle(
                     new RemoteTableName(Optional.empty(), Optional.of(schemaTableName.getSchemaName()), schemaTableName.getTableName()),
                     columnNames,
                     columnTypes.build(),
                     Optional.empty(),
                     primaryKeys.isEmpty() ? Optional.of(IGNITE_DUMMY_ID) : Optional.empty());
+            rollbackActionConsumer.accept(() -> rollbackCreateDestinationTable(session, destinationTableHandle.getRemoteTableName()));
+            return destinationTableHandle;
         }
         catch (SQLException e) {
             throw new TrinoException(JDBC_ERROR, e);
--- a/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
+++ b/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
@@ -220,7 +220,7 @@ public class IgniteMetadata
         if (saveMode == REPLACE) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
-        igniteClient.beginCreateTable(session, tableMetadata);
+        igniteClient.beginCreateTable(session, tableMetadata, _ -> {});
     }
 
     @Override
@@ -232,7 +232,7 @@ public class IgniteMetadata
         if (replace) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
-        return igniteClient.beginCreateTable(session, tableMetadata);
+        return igniteClient.beginCreateTable(session, tableMetadata, rollbackActions::add);
     }
 
     @Override
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/CachingJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/CachingJdbcClient.java
@@ -600,9 +600,9 @@ public class CachingJdbcClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
-        return delegate.beginCreateTable(session, tableMetadata);
+        return delegate.beginCreateTable(session, tableMetadata, rollbackActionConsumer);
     }
 
     @Override
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ForwardingJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ForwardingJdbcClient.java
@@ -265,9 +265,9 @@ public abstract class ForwardingJdbcClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
-        return delegate().beginCreateTable(session, tableMetadata);
+        return delegate().beginCreateTable(session, tableMetadata, rollbackActionConsumer);
     }
 
     @Override
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/RetryingJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/RetryingJdbcClient.java
@@ -368,10 +368,10 @@ public class RetryingJdbcClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
         // no retrying as it could be not idempotent operation
-        return delegate.beginCreateTable(session, tableMetadata);
+        return delegate.beginCreateTable(session, tableMetadata, rollbackActionConsumer);
     }
 
     @Override
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/jmx/StatisticsAwareJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/jmx/StatisticsAwareJdbcClient.java
@@ -353,9 +353,9 @@ public final class StatisticsAwareJdbcClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
-        return stats.getBeginCreateTable().wrap(() -> delegate().beginCreateTable(session, tableMetadata));
+        return stats.getBeginCreateTable().wrap(() -> delegate().beginCreateTable(session, tableMetadata, rollbackActionConsumer));
     }
 
     @Override
--- a/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
+++ b/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
@@ -74,6 +74,7 @@ import java.util.OptionalLong;
 import java.util.Set;
 import java.util.TimeZone;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Iterables.getOnlyElement;
@@ -476,7 +477,7 @@ public class DruidJdbcClient
     }
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
         throw new TrinoException(NOT_SUPPORTED, "This connector does not support creating tables with data");
     }
--- a/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
+++ b/plugin/trino-exasol/src/main/java/io/trino/plugin/exasol/ExasolClient.java
@@ -56,6 +56,7 @@ import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Set;
 import java.util.function.BiFunction;
+import java.util.function.Consumer;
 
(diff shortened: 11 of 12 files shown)


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -1449,7 +1454,22 @@ public void dropTable(ConnectorSession session, JdbcTableHandle handle)
...
+                        .build())
+                .run(context -> {
+                    if (context.getAttemptCount() > 0) {
+                        log.debug("Retrying to drop table %s: attempt %s", remoteTableName, context.getAttemptCount());
+                        if (getTableHandle(session, remoteTableName.getSchemaTableName()).isEmpty()) {
+                            // Table is already dropped
+                            return;
+                        }
+                    }
+                    execute(session, sql);
```
findepi: i agree  that it makes sense for ``jdbcClient.dropTable`` to be retried in general. 
it feels that this is the mission with which `RetryingJdbcClient` was added. 
i therefore think it may be wrong to add retries for this particular method in BaseJdbcClient if retries are generally handled by `RetryingJdbcClient`.

also, `jdbcClient.dropTable(non_existent_table)` should fail. now it will succeed
OTOH proactively checking table existence would be penalty for all usages.

since the motivation for this change is failed cleanup in CTAS and there we have knowledge of table existence (since we create it), did you consider apply retries only in this flow?
of course, having retries for dropTable makes sense, but we can follow-up on these.
in the worst case the query user can retry their query. with CTAS we have bigger problem as they can't retry if we fail to cleanup.

findinpath: thank you for the suggestion - will try to implement it.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
```diff
@@ -1200,6 +1200,7 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
         JdbcOutputTableHandle handle = jdbcClient.beginCreateTable(session, tableMetadata);
+        rollbackActions.add(() -> jdbcClient.rollbackCreateTableAsSelect(session, handle));
```
chenjian2664: There will be a failure if the table not exists, but it seems previous exists in `rollbackCreateTable` as well

chenjian2664: Why put it in this method, usually the target table is created in the finish stage, or I am missing sth

findinpath: > Why put it in this method, usually the target table is created

The destination table is created in `jdbcClient.beginCreateTable`

https://github.com/trinodb/trino/blob/137c601cb891e5ab08c377dfa0d4edad6fc3c872/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L835-L836

findinpath: > There will be a failure if the table not exists

The (empty) table is created during the previous statement so it will actually exist.

chenjian2664: You are right, but the case is when `shouldUseFaultTolerantExecution` is true, there is another case which will create temp table + rename. in that case when the `beginCreateTable` rollback the target table does not yet exist.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -1465,6 +1465,17 @@ public void rollbackCreateTable(ConnectorSession session, JdbcOutputTableHandle
...
 
+    @Override
+    public void rollbackCreateTableAsSelect(ConnectorSession session, JdbcOutputTableHandle handle)
+    {
+        dropTable(session,
+                new RemoteTableName(
+                        handle.getRemoteTableName().getCatalogName(),
+                        handle.getRemoteTableName().getSchemaName(),
+                        handle.getRemoteTableName().getTableName()),
+                false);
```
chenjian2664:
```suggestion
        dropTable(session,                        
                handle.getRemoteTableName(),
                false);
```

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
findinpath: Correct. I have overseen this aspect.
I've changed slightly the implementation to have more control when to register for rolling back the creation of the destination table.

Thank you for the heads'up.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -838,17 +838,27 @@ public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, Connecto
                 // Create the temporary table
                 ColumnMetadata pageSinkIdColumn = getPageSinkIdColumn(
                         tableMetadata.getColumns().stream().map(ColumnMetadata::getName).toList());
-                return createTable(session, tableMetadata, generateTemporaryTableName(session), Optional.of(pageSinkIdColumn));
+                JdbcOutputTableHandle handle = createTable(session, tableMetadata, generateTemporaryTableName(session), Optional.of(pageSinkIdColumn));
+                rollbackActionConsumer.accept(() -> rollbackCreateDestinationTable(session, handle));
```
chenjian2664: We should move this logic earlier (after line 837),  after the target table is successfully created, we should register the rollback action. otherwise, if the temporary table creation fails, we may miss the opportunity to drop the target table

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -838,17 +838,27 @@ public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, Connecto
...
+                rollbackActionConsumer.accept(() -> rollbackCreateTable(session, handle));
+                return handle;
             }
             else {
-                return createTable(session, tableMetadata, generateTemporaryTableName(session));
+                JdbcOutputTableHandle handle = createTable(session, tableMetadata, generateTemporaryTableName(session));
+                rollbackActionConsumer.accept(() -> rollbackCreateTable(session, handle));
+                return handle;
```
chenjian2664: I think we don't need to change the existing logic to remove the temporary table

findinpath: It may seem a bit strange while maintaining the code why there is rollbacking done internally in jdbc.beginCreateTable and also in DefaultJdbcMetadata.beginCreateTable.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
chenjian2664: How about: 
```suggestion
                JdbcOutputTableHandle targetTable = createTable(session, tableMetadata, tableMetadata.getTable().getTableName());
                rollbackActionConsumer.accept(() -> drop the target table);
```

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -818,23 +818,24 @@ protected PreparedQuery applyQueryTransformations(JdbcTableHandle tableHandle, P
...
 
     @Override
-    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
+    public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer)
     {
         try {
             if (shouldUseFaultTolerantExecution(session)) {
                 // Create the target table
-                createTable(session, tableMetadata);
+                JdbcOutputTableHandle destinationTableHandle = createTable(session, tableMetadata);
```
chenjian2664: How about using `createTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, String targetTableName)` ?

findinpath: We lose the Trino exception wrapping this way.

```
    public void createTable(ConnectorSession session, ConnectorTableMetadata tableMetadata)
    {
        try {
            createTable(session, tableMetadata, tableMetadata.getTable().getTableName());
        }
        catch (SQLException e) {
            throw new TrinoException(JDBC_ERROR, e);
        }
    }
```

I find the changing of the method signature for `createTable(session, tableMetadata)` call not dramatic.

findinpath: Nevermind - understood the hint.

plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
```diff
@@ -232,7 +232,7 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
         if (replace) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
-        return igniteClient.beginCreateTable(session, tableMetadata);
+        return igniteClient.beginCreateTable(session, tableMetadata, rollbackActions::add);
```
chenjian2664, line 235: It is not used in `Ignite`, we could put `_ -> {}` like above

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
```diff
@@ -137,7 +137,7 @@ public class DefaultJdbcMetadata
     private final boolean precalculateStatisticsForPushdown;
     private final Set<JdbcQueryEventListener> jdbcQueryEventListeners;
 
-    private final List<Runnable> rollbackActions = new ArrayList<>();
+    protected final List<Runnable> rollbackActions = new ArrayList<>();
```
chenjian2664, line 140: revert since it is not necessary exposed to `Ignite`

plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
chenjian2664: Please ignore my above commits, it seems we didn't handle rollback correctly in Ignite

plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteClient.java
chenjian2664: Could you add the rollback for target table here?

plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteClient.java
```diff
@@ -417,13 +418,15 @@ public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, Connecto
...
-            return new IgniteOutputTableHandle(
+            IgniteOutputTableHandle destinationTableHandle = new IgniteOutputTableHandle(
                     new RemoteTableName(Optional.empty(), Optional.of(schemaTableName.getSchemaName()), schemaTableName.getTableName()),
                     columnNames,
                     columnTypes.build(),
                     Optional.empty(),
                     primaryKeys.isEmpty() ? Optional.of(IGNITE_DUMMY_ID) : Optional.empty());
+            rollbackActionConsumer.accept(() -> rollbackCreateDestinationTable(session, destinationTableHandle));
+            return destinationTableHandle;
+
```
chenjian2664: remove this empty line

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -849,6 +850,11 @@ public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, Connecto
         }
     }
 
+    protected void rollbackCreateDestinationTable(ConnectorSession session, JdbcOutputTableHandle handle)
+    {
+        dropTable(session, handle.getRemoteTableName(), false);
```
ebyhr: Why do we want to take `JdbcOutputTableHandle` instead of `RemoteTableName` as the method parameter?

chenjian2664:
Adding the release note into `ClickHouse`, `DuckDB`, `Ignite`, `MariaDB`, `MySQL`, `Oracle`, `PostgreSQL`, `Redshift`, `SingleStore`, `Snowflake`, `SQL Server`, `Vertica`. let me know if you think there are connectors I should add the RN as well

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -849,6 +850,11 @@ public JdbcOutputTableHandle beginCreateTable(ConnectorSession session, Connecto
         }
     }
 
+    protected void rollbackCreateDestinationTable(ConnectorSession session, RemoteTableName remoteTableName)
```
findepi, line 853: `rollbackTableCreation` / `rollbackCreateTable`

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
```diff
@@ -1199,7 +1199,7 @@ public ConnectorOutputTableHandle beginCreateTable(ConnectorSession session, Con
         if (replace) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
-        JdbcOutputTableHandle handle = jdbcClient.beginCreateTable(session, tableMetadata);
+        JdbcOutputTableHandle handle = jdbcClient.beginCreateTable(session, tableMetadata, rollbackActions::add);
         rollbackActions.add(() -> jdbcClient.rollbackCreateTable(session, handle));
```
findepi, line 1203: we had rollback here. do we rollback twice now?

plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
```diff
@@ -220,7 +220,7 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
         if (saveMode == REPLACE) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
-        igniteClient.beginCreateTable(session, tableMetadata);
+        igniteClient.beginCreateTable(session, tableMetadata, _ -> {});
```
findepi, line 223: document why it's ok to ignore a rollback action
(it should be actually safer not to ignore -- this is CREATE TABLE flow. if nothing follows CREATE TABLE, nothing fails later and no rollback will take place)

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
```diff
@@ -210,7 +210,7 @@ default void setTableProperties(ConnectorSession session, JdbcTableHandle handle
 
     void createTable(ConnectorSession session, ConnectorTableMetadata tableMetadata);
 
-    JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata);
+    JdbcOutputTableHandle beginCreateTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, Consumer<Runnable> rollbackActionConsumer);
```
findepi, line 213: `rollbackActionCollector` would be a better name

more importantly, this is a rather odd abstraction and doesn't seem necessary.
so far it was JdbcMetadata who was respomnsible for instantiating and remembering rollback actions. let's not change that, if we don't have to

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
chenjian2664: the `rollbackCreateTable` is mainly for the clean ups of the temporary tables

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
chenjian2664: > rollbackActionCollector would be a better name

+1

> so far it was JdbcMetadata who was respomnsible for instantiating and remembering rollback actions


We are entering a situation where it is unclear when it is safe to roll back the target table during CTAS. For example, in `BaseJdbcClient#beginCreateTable`, when FTE is enabled, both the target table and a temporary table are created in the method, during rollback in `DefaultJdbcMetadata`, we cannot reliably determine which tables should be cleaned up

> let's not change that, if we don't have to

For now, this works as a solution, but I believe we could evolve the API to shift the responsibility back to `DefaultJdbcMetadata`. For example, we could change the method's return value so that `DefaultJdbcMetadata` knows exactly which tables need to be cleaned up

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
findepi: the responsibilities are now blurred, aren't they?
maybe we can clarify them with better naming or something?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcClient.java
findepi: > > rollbackActionCollector would be a better name
> 
> +1

Included in 
- https://github.com/trinodb/trino/pull/27794

however, the name may still be imperfect -- https://github.com/trinodb/trino/pull/27702#discussion_r2657456179




> in `BaseJdbcClient#beginCreateTable`, when FTE is enabled, both the target table and a temporary table are created in the method

I made the same mistake. It's not FTE-related despite the code making it look so.

> during rollback in DefaultJdbcMetadata, we cannot reliably determine which tables should be cleaned up

It's unclear why destination table is created there. I think it might be possible to defer that statement until later.


> > let's not change that, if we don't have to
> 
> For now, this works as a solution, but I believe we could evolve the API to shift the responsibility back to `DefaultJdbcMetadata`. For example, we could change the method's return value so that `DefaultJdbcMetadata` knows exactly which tables need to be cleaned up

That was also my intention. 
I stopped short from execution on it in 
- https://github.com/trinodb/trino/pull/27794

when i saw the beginMerge flow. We create high number of tables there, and someone needs to clean them up if we fail mid-way.

findinpath:
Following up the open discussions in https://github.com/trinodb/trino/pull/27848

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DefaultJdbcMetadata.java
chenjian2664: > the responsibilities are now blurred, aren't they?

now, yes. But I think the name is ok since we are inside the "createTable" method, that's is fair to have a rollback method like "rollbackCreateTable", it just our implementation only do cleanups of temporary table. Ideally if we could do the cleanup of target table inside this method as well

@findinpath https://github.com/trinodb/trino/pull/27848 the method is going to split such cleanups into two methods(we already do), but finally, I think that's more elegant that we should only has one rollback methods for "createTable", caller don't care or needs to know the cleanups table is "temporary" or not

findinpath: While rolling back the creation of the destination table is characteristic only for CTAS statements, rolling back the creation of a temporary table can occur on any DML statement.
I see a good rationale on having two separate methods for doing rollback for table creation.

PR #27703, opened 2025-12-19
Introduce ASOF and ASOF LEFT joins with single-inequality anchor

Pluies:
## Description

- Adds ASOF/ASOF LEFT joins for nearest‑neighbor matching.
- ON clause requires exactly one inequality comparing a right‑side expression to a left‑side expression; additional equi and same‑side predicates are allowed.
- Implements analyzer/planner support and predicate pushdown semantics.
- Updates documentation and adds focused analyzer, planner, and query tests.

Hey folks 👋 Seeing #27641 being pushed in order to implement #21759, I'm rushing our plans of upstreaming ASOF joins as implemented recently in Dune 🎁 

We have developed and built this feature on top of Trino 467, and this is a quick rebase on top of `master` to get the ball rolling. Expect potential bugs and test failures; I'll come back and clean this up asap.

I'm not expecting either this PR or #27641 to be adopted wholesale, but hopefully this can help in the conversation wrt which syntax and implementation of ASOF joins ends up in Trino!

cc @martint 

Related, our user-facing Dune docs: https://docs.dune.com/query-engine/Functions-and-operators/asof-join

## Release notes

```markdown
## Section
```


findepi:
That's a cool feature!

> Seeing #27641 being pushed in order to implement #21759, I'm rushing our plans of upstreaming ASOF joins as implemented recently in Dune 🎁

Would be nice if we could have one PR to collaborate on.
Given you created this PR knowing about #27641, does this mean this implementation is more complete than the other one?

Can you please chime in on #27641 too?

Pluies:
> Would be nice if we could have one PR to collaborate on. Given you created this PR knowing about #27641, does this mean this implementation is more complete than the other one?

I believe this PR is slightly more production-ready.

My understanding is:
- This PR supports ASOF LEFT as well as ASOF
- Broader support for the inequality condition: if I read #27641 correctly, the inequality condition _has to be_ the last column ([here](https://github.com/trinodb/trino/pull/27641/files#diff-446009e3ffe5a9383e716ac92d3cbf5474f96ac86f1c20d0f539927c56777ddeR1166)), whereas in this PR there is no such limitation
- This PR includes stronger syntax checks; see checks in [StatementAnalyzer.java](https://github.com/trinodb/trino/pull/27703/files#diff-f7e345d57c9a04d4f8c2409e09245adca19a7fe8a20b3255bbae7a97dc0f4a9d) and inequality checks in [LocalExecutionPlanner.java](https://github.com/trinodb/trino/pull/27703/files#diff-efde3b5b14ba20862096f59fd6550810866baa868ddc68f9393ec3fbf201ccc8)
- Less of a technical point and more... social proof?, but this PR [includes documentation](https://github.com/trinodb/trino/blob/613d12c30cfc4bd61b1071f44199017479df7bd0/docs/src/main/sphinx/sql/select.md) and has been in use in prod at Dune for a few months already!

But these could all be ported between PR if needed. I believe the main difference is in the way these PRs approach the issue in the first place: this PR supports ASOF joins as a native construct, whereas #27641 supports ASOF joins by means of rewriting them into a LEFT JOIN via [RewriteAsofJoinToLeftJoinWithTop1.java](https://github.com/trinodb/trino/pull/27641/files#diff-ec2b7ec93755574200161c807b162dadc0cacfb7fa0d4605d2d57d41ec36d846) ; an SQL syntax that Trino already understands.

cc @jirislav , let me know if I misunderstood changes 🙏 

> Can you please chime in on #27641 too?

Linking this comment there!

jirislav:
I agree with @Pluies about this PR being more production-ready. My PR ([#27641](https://github.com/trinodb/trino/pull/27641)) was intended to only establish the syntax with a simple translation to a LEFT JOIN, so we could ship the improvements of ASOF join implementation incrementally (for example by shipping a sort-merge operator using binary search).

> Broader support for the inequality condition: if I read https://github.com/trinodb/trino/pull/27641 correctly, the inequality condition has to be the last column ([here](https://github.com/trinodb/trino/pull/27641/files#diff-446009e3ffe5a9383e716ac92d3cbf5474f96ac86f1c20d0f539927c56777ddeR1166)), whereas in this PR there is no such limitation

It is only a requirement when the user chooses the `USING` expression (e.g. `ASOF JOIN b USING (a, b, ts)`. When the `USING` keyword is not used, then it is perfectly possible to use `a ASOF JOIN b ON a.ts <= b.ts AND a.key = b.key`).

Note that having the time-dimension as the last one after the `USING` keyword is the most common behavior in major platforms, see:
- [DuckDB docs](https://duckdb.org/docs/stable/guides/sql_features/asof_join#asof-joins-with-the-using-keyword) (they came up with the ASOF join idea)
- [ClickHouse docs](https://clickhouse.com/docs/sql-reference/statements/select/join#asof-join-usage)
- with only [Snowflake introducing MATCHING_CONDITIONS](https://docs.snowflake.com/en/sql-reference/constructs/asof-join#label-asof-join-using),
- curiously [KDB allowing the time dimension to not be the last](https://code.kx.com/q/ref/aj/#performance), but incurring a performance cost (but it's not SQL anyway, so less relevant).

jirislav:
@Pluies what is the reason to close this? Do we have some alternative PR somewhere I'm not aware of?

Pluies:
@jirislav sorry, my bad - I was cleaning up branches in my fork and deleted this one by mistake 🤦

kasiafi (commented):
Hi @Pluies, I have some high level comments/questions:

1. The syntax. I understand that the proposed syntax is something invented and adopted by other DBs. One thing I don't like about this syntax is that the "ASOF-inequality" does not stand out in any way from other join conditions, while it's semantics is very different. Are there alternative syntax approaches that consider this?

2. The "ASOF-inequality" also doesn't get any special treatment in the planner -- it becomes a regular part of the Join filter. And then it has to be extracted again after the query is optimized. This approach seems brittle. What if the inequality gets transformed or optimized-out altogether? What if some optimization adds another inequality to the Join filter? I think that the "ASOF-inequality" must be carried separately from the Join filter if we want to be sure that we still have it on the other end of the Planner.

3. Where is the logic responsible for returning one match for each row?

4. What happens if there are other join conditions than equalities/inequalities?

5. If there are ties on the build side, how are they solved?

6. If there are ties on the probe side matching ties on the build side, is the result consistent for each tied probe row?

jirislav:
The feature request https://github.com/trinodb/trino/issues/21759 was resolved by the `NEAREST` join support (https://github.com/trinodb/trino/pull/28937), so I'm closing this. Anyway, thanks for the effort @Pluies !

Issue #27704, opened 2025-12-19
Async query analysis may block resource group scheduling when queue full


jinyangli34:
Resource group wants to fail the query due to queue full. 
```
dispatcher-query-46667 BLOCKED #4122888 4123830
  at io.trino.plugin.hive.metastore.SemiTransactionalHiveMetastore.cleanupQuery(SemiTransactionalHiveMetastore.java:1430)
  at io.trino.plugin.hive.HiveMetadata.cleanupQuery(HiveMetadata.java:4101)
  at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorMetadata.cleanupQuery(ClassLoaderSafeConnectorMetadata.java:585)
  at io.trino.tracing.TracingConnectorMetadata.cleanupQuery(TracingConnectorMetadata.java:657)
  at io.trino.metadata.MetadataManager$QueryCatalogs.finish(MetadataManager.java:2978)
  at io.trino.metadata.MetadataManager.cleanupQuery(MetadataManager.java:1130)
  at io.trino.tracing.TracingMetadata.cleanupQuery(TracingMetadata.java:679)
  at io.trino.execution.QueryStateMachine.cleanupQuery(QueryStateMachine.java:1312)
  at io.trino.execution.QueryStateMachine.cleanupQueryQuietly(QueryStateMachine.java:1319)
  at io.trino.execution.QueryStateMachine.transitionToFailed(QueryStateMachine.java:1267)
  at io.trino.execution.QueryStateMachine.transitionToFailed(QueryStateMachine.java:1254)
  at io.trino.dispatcher.LocalDispatchQuery.fail(LocalDispatchQuery.java:295)
  at io.trino.execution.resourcegroups.InternalResourceGroup.run(InternalResourceGroup.java:880)
  at io.trino.execution.resourcegroups.InternalResourceGroupManager.submit(InternalResourceGroupManager.java:126)
```

SemiTransactionalHiveMetastore lock is hold by the async query analysis. 
```
Query-20251218_175855_00041_eu8vz-4123830 RUNNABLE #4123830
  at sun.nio.ch.Net.poll(Net.java:-2)
  at sun.nio.ch.NioSocketImpl.park(NioSocketImpl.java:190)
  at sun.nio.ch.NioSocketImpl.timedRead(NioSocketImpl.java:279)
  at sun.nio.ch.NioSocketImpl.implRead(NioSocketImpl.java:303)
[...]

PR #27705, opened 2025-12-19
Avoid failing query blocked by async query analysis

jinyangli34:
## Description
A fix proposal for https://github.com/trinodb/trino/issues/27704

## Release notes

```markdown
## Section
* Fix some things. ({issue}`27704`)
```

### DIFF
--- a/core/trino-main/src/test/java/io/trino/dispatcher/TestLocalDispatchQuery.java
+++ b/core/trino-main/src/test/java/io/trino/dispatcher/TestLocalDispatchQuery.java
@@ -71,6 +71,9 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.Executor;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
 
 import static com.google.common.util.concurrent.Futures.immediateVoidFuture;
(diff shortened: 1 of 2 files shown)


losipiuk:
Would it be possible to change Hive connector code to not hold on to lock instead?

core/trino-main/src/test/java/io/trino/dispatcher/TestLocalDispatchQuery.java
```diff
@@ -169,6 +172,150 @@ public void testSubmittedForDispatchedQuery()
...
     }
 
+    /**
+     * Tests that fail() does not block when the queryExecutionFuture is still running,
+     * and verifies that the original failure cause is preserved (not replaced by CancellationException).
+     * This reproduces the scenario where:
+     * 1. Query analysis is in progress (e.g., blocked on HMS call)
+     * 2. Resource group rejects the query and calls fail()
+     * 3. fail() should not block waiting for the analysis to complete
+     * 4. The original failure cause should be preserved in the query state
```
findepi, line 182: > original failure cause is preserved (not replaced by CancellationException)

Cancellation is _the_ failure cause when query is being forced to fail.

findepi:
Thanks for reporting the issue (https://github.com/trinodb/trino/issues/27704) and offering a fix!
I think the issue needs a bit more clarity. I added some questions there. Converting this PR to draft for now, to avoid parallel discussions in two places.

PR #27706, opened 2025-12-19
Add array_first(array(E), function(E, boolean)) function

gertjanal:
## Description
Find an element in an array by using a predicate
`array_first(users, user -> user.uid = 1337)`

## Release notes

```markdown
## Functions
* Added array_first(array(E), function(E, boolean)) to return an element matching the predicate. ({issue}`27706`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayFirstFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayFirstFunction.java
@@ -27,22 +27,47 @@ import java.lang.invoke.MethodHandle;
 import static io.trino.operator.scalar.ArrayElementAtFunction.elementAt;
 import static io.trino.spi.function.InvocationConvention.InvocationArgumentConvention.BLOCK_POSITION_NOT_NULL;
 import static io.trino.spi.function.InvocationConvention.InvocationReturnConvention.FAIL_ON_NULL;
+import static io.trino.spi.function.InvocationConvention.InvocationReturnConvention.NULLABLE_RETURN;
 import static io.trino.spi.function.OperatorType.READ_VALUE;
+import static java.lang.Boolean.TRUE;
 
-@ScalarFunction("array_first")
-@Description("Get the first element of an array")
 public final class ArrayFirstFunction
 {
+    private static final String NAME = "array_first";
+
     private ArrayFirstFunction() {}
 
-    @TypeParameter("E")
     @SqlNullable
     @SqlType("E")
+    @TypeParameter("E")
+    @ScalarFunction(NAME)
+    @Description("Get the first element of an array")
     public static Object arrayFirst(
             @OperatorDependency(operator = READ_VALUE, argumentTypes = "E", convention = @Convention(arguments = BLOCK_POSITION_NOT_NULL, result = FAIL_ON_NULL)) MethodHandle readValue,
             @SqlType("array(E)") Block array)
             throws Throwable
     {
         return elementAt(readValue, array, 1);
+    }
+
+    @SqlNullable
+    @SqlType("E")
+    @TypeParameter("E")
+    @ScalarFunction(NAME)
+    @Description("Get the first element of an array matching the given predicate")
+    public static Object arrayFirstMatch(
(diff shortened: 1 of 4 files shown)


martint:
I'd suggest an overload to `array_first` that takes a lambda expression. `first_match` is inconsistent with `all_match` and `any_match`, which are intended to answer the questions "do all elements match", "does any element match"?

Therefore: `array_first(array(E), E -> boolean) -> E`

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayFirstFunction.java
```diff
@@ -45,4 +46,24 @@ public static Object arrayFirst(
...
+    @SqlNullable
+    @SqlType("E")
+    public static Object arrayFirstMatch(
+            @OperatorDependency(operator = READ_VALUE, argumentTypes = "E", convention = @Convention(arguments = BLOCK_POSITION_NOT_NULL, result = FAIL_ON_NULL)) MethodHandle readValue,
+            @SqlType("array(E)") Block arrayBlock,
+            @SqlType("function(E, boolean)") ObjectToBooleanFunction function)
+            throws Throwable
+    {
+        for (int i = 0; i < arrayBlock.getPositionCount(); i++) {
+            if (!arrayBlock.isNull(i)) {
```
wendigo: You could do:
```
if (arrayBlock.isNull(i)) {
    continue; 
}
```

to reduce code nesting

gertjanal: Updated

chenjian2664:
ci @gertjanal :
```
1) [Guice/ErrorInCustomProvider]: IllegalArgumentException: Implementations with type parameters must all have matching signatures. <E>(array(e),function(e,boolean)):e does not match <E>(array(e)):e
  at ServerMainModule.systemFunctionBundle(ServerMainModule.java:515)
  while locating FunctionBundle annotated with @Element(setName=,uniqueId=4572, type=MULTIBINDER, keyType=)
  ```

**Comment by gertjanal** [comment]
@ chenjian2664 Fixed

File: core/trino-main/src/main/java/io/trino/operator/scalar/ArrayFirstFunction.java [hunk 2]
```diff
@@ -28,21 +28,48 @@
...
+    @ScalarFunction(NAME)
+    @Description("Get the first element of an array matching the given predicate")
+    public static Object arrayFirstMatch(
+            @OperatorDependency(operator = READ_VALUE, argumentTypes = "E", convention = @Convention(arguments = BLOCK_POSITION_NOT_NULL, result = FAIL_ON_NULL)) MethodHandle readValue,
+            @SqlType("array(E)") Block arrayBlock,
+            @SqlType("function(E, boolean)") ObjectToBooleanFunction function)
+            throws Throwable
+    {
+        for (int i = 0; i < arrayBlock.getPositionCount(); i++) {
+            if (arrayBlock.isNull(i)) {
```
chenjian2664: Just curious, do we allow to use `array_first(users, user -> user = null)` ?

gertjanal: Yes I did think of that and thought it wouldn't be a problem as in that case the function as written here returns null anyway (except without using the lambda). But now if I think a bit further, I could think of a usecase where the current implementation becomes problematic:
```
array_first(ARRAY['foo', 'bar', NULL, 'hello'], element -> contains(ARRAY['hello', 'world', NULL], element))
-- This returns 'hello', as the NULL element in the first array is skipped
```

While we would expect NULL to match earlier. I'll update the test and code

gertjanal: I added a test and updated the code

gertjanal: My previous test code idea didn't work as `contains(ARRAY['hello', NULL], NULL)` returns NULL

wendigo:
@gertjanal please fix checkstyle formatting issue

wendigo:
I've amended commit message to make it shorter and aligned with our commits guidelines

PR #27707, opened 2025-12-20
Fix Jdk 25 release notes entry

mosabua:
## Description

Its breaking change just like it was in prior releases such as 476 with JDK 24. PR link was wrong.

Found during recording https://youtube.com/live/7clvlAxGFOI

## Release notes


mosabua:
Thank you @ebyhr

chenjian2664:
thanks @mosabua for fixing it :)

PR #27708, opened 2025-12-20
Add Lakehouse connector in various places

mosabua:
## Description

See title

Going forward in release notes we should treat it like the other four and add relevant entries. also CC @chenjian2664 

Found during https://youtube.com/live/7clvlAxGFOI and also had some already in mind and filed in https://github.com/simpligility/contributions/issues/38

## Release notes

### DIFF
--- a/docs/src/main/sphinx/installation/plugins.md
+++ b/docs/src/main/sphinx/installation/plugins.md
@@ -248,6 +248,10 @@ with the listed coordinates.
   - [](/admin/event-listeners-kafka)
   - [io.trino:trino-kafka-event-listener](https://central.sonatype.com/search?q=io.trino%3Atrino-kafka-event-listener)
   - {download_gh}`kafka-event-listener`
+* - lakehouse
+  - [](/connector/lakehouse)
+  -
+  - {download_gh}`lakehouse`
 * - loki
   - [](/connector/loki)
   - [io.trino:trino-loki](https://central.sonatype.com/search?q=io.trino%3Atrino-loki)
--- a/docs/release-template.md
+++ b/docs/release-template.md
@@ -46,6 +46,8 @@
 
 ## Kafka connector
 
+## Lakehouse connector
+
 ## Loki connector
 
 ## MariaDB connector
--- a/docs/src/main/sphinx/ext/download.py
+++ b/docs/src/main/sphinx/ext/download.py
@@ -49,6 +49,7 @@ ARTIFACTS = {
     'jmx':  ('trino-jmx', 'zip'),
     'kafka':  ('trino-kafka', 'zip'),
     'kafka-event-listener':  ('trino-kafka-event-listener', 'zip'),
+    'lakehouse':  ('trino-lakehouse', 'zip'),
     'loki':  ('trino-loki', 'zip'),
     'mariadb':  ('trino-mariadb', 'zip'),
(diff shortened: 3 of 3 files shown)


docs/src/main/sphinx/installation/plugins.md
```diff
@@ -248,6 +248,10 @@ with the listed coordinates.
   - [](/admin/event-listeners-kafka)
   - [io.trino:trino-kafka-event-listener](https://central.sonatype.com/search?q=io.trino%3Atrino-kafka-event-listener)
   - {download_gh}`kafka-event-listener`
+* - lakehouse
+  - [](/connector/lakehouse)
+  - [io.trino:trino-lakehouse](https://central.sonatype.com/search?q=io.trino%3Atrino-lakehouse)
```
ebyhr: https://central.sonatype.com/search?q=io.trino%3Atrino-lakehouse returns an empty result unlike other plugins. It would be nice to drop the link in my opinion.

chenjian2664:
If we make a change in Iceberg, (it affects Lakehouse) should we also add a corresponding entry in Lakehouse?

If the feature is supported only by some connectors (of Hive, Hudi, Iceberg, and Delta Lake) but not by others, how should this be reflected in the release notes?

docs/src/main/sphinx/installation/plugins.md
electrum: @martint should we publish this to central?

mosabua: Done .. maybe we should remove all these links .. it seems like nothing is published at all anymore for the plugins.. although that kinda feels like a bug .. I thought we still publish everything aside from the big zip files .. so the pom and the simple JAR would still go there .. but currently it is not. Is that a bug or on purpose @martint @wendigo

mosabua:
> If we make a change in Iceberg, (it affects Lakehouse) should we also add a corresponding entry in Lakehouse?
> 
> If the feature is supported only by some connectors (of Hive, Hudi, Iceberg, and Delta Lake) but not by others, how should this be reflected in the release notes?

It should appear in each connector that is affected by a change .. part the job of writing a PR and merging as maintainer is to clarify what it affects and add that info the the PR description so whoever assembles the release notes can just take the info and does not have to spend hours reading code and understanding each PR.

chenjian2664:
@mosabua What I am trying to understand is the changes in partial of Lake house connecotors should be copied into LakeHouse section or it should mention in a different phrase? or we don't mention 
consider the case where you added a support on Iceberg which actually enhance the Lakehouse, what's suggestion of the such release note?

mosabua:
> @mosabua What I am trying to understand is the changes in partial of Lake house connecotors should be copied into LakeHouse section or it should mention in a different phrase? or we don't mention consider the case where you added a support on Iceberg which actually enhance the Lakehouse, what's suggestion of the such release note?

I think we should treat the Lakehouse connector just like the Hive, Iceberg, Delta Lake, and Hudi connectors, so

* if a file system support aspects applies to the connector - add it
* if a metastore support aspect applied - add it
* if a specific change on the lakehouse connector itself is done - add it
* and reverse .. so whatever does not apply .. dont add it. 

Makes sense?

chenjian2664:
Thanks, make sense for the add cases.
For the reverse, I still has the confusion, but we will see if there is a reality case let's discuss then

PR #27715, opened 2025-12-21
Update ignored properties in GCS config test

chenjian2664:
Removed 'gcs.json-key' and 'gcs.json-key-file-path' from the set of ignored properties in the `assertFullMapping` call.

## Description

The `gcs.json-key` and `gcs.json-key-file-path` properties were moved to `GcsServiceAccountAuthConfig` in #27106, so there is no need to keep them in `TestGcsFileSystemConfig` anymore.

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
+++ b/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
@@ -88,7 +88,7 @@ public class TestGcsFileSystemConfig
                 .setMinBackoffDelay(new Duration(20, MILLISECONDS))
                 .setMaxBackoffDelay(new Duration(20, MILLISECONDS))
                 .setApplicationId("application id");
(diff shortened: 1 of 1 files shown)


lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystemConfig.java
```diff
@@ -88,7 +88,7 @@ void testExplicitPropertyMappings()
                 .setMinBackoffDelay(new Duration(20, MILLISECONDS))
                 .setMaxBackoffDelay(new Duration(20, MILLISECONDS))
                 .setApplicationId("application id");
-        assertFullMapping(properties, expected, Set.of("gcs.json-key", "gcs.json-key-file-path", "gcs.use-access-token"));
```
findepi, line 91: would be great if `assertFullMapping` validated the set of ignored properties is accurate
this would make exclusions self-maintained

findepi: - https://github.com/airlift/airlift/pull/1701

PR #27731, opened 2025-12-22
Clean ups for Delta CdcEntry

chenjian2664:
## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/CdcEntry.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/CdcEntry.java
@@ -27,6 +27,7 @@ import static io.airlift.slice.SizeOf.instanceSize;
 import static io.airlift.slice.SizeOf.sizeOf;
 import static io.trino.plugin.deltalake.transactionlog.TransactionLogUtil.canonicalizePartitionValues;
 import static java.lang.String.format;
+import static java.util.Objects.requireNonNull;
 
 public class CdcEntry
 {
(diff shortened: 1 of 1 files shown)


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/CdcEntry.java
```diff
@@ -75,17 +76,10 @@ public long getSize()
         return size;
     }
 
-    @JsonProperty("dataChange")
-    public boolean isDataChange()
-    {
-        return dataChange;
-    }
```
ebyhr, line 82: Why do you remove this method? 

https://github.com/delta-io/delta/blob/master/PROTOCOL.md#add-cdc-file
> dataChange: Should always be set to false for cdc actions because they do not change the underlying data of the table

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/CdcEntry.java
```diff
@@ -36,19 +38,18 @@ public class CdcEntry
...
     @JsonCreator
     public CdcEntry(
             @JsonProperty("path") String path,
             @JsonProperty("partitionValues") Map<String, String> partitionValues,
             @JsonProperty("size") long size)
     {
-        this.path = path;
-        this.partitionValues = partitionValues;
+        this.path = requireNonNull(path,  "path is null");
+        // the partition value could be null, we check the `partitionValues` non-null
```
ebyhr:
```suggestion
        // Avoid ImmutableMap.copyOf because the value may have nulls
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/CdcEntry.java
chenjian2664: missing this one, was thought there is no usage, but actual write the "dataChange" entry in json when writing

PR #27741, opened 2025-12-23
Make task outputDataSize track network bytes

lukasz-stec:
Previously, `outputDataSize` for tasks and stages was equal to the memory retained by the pages sent to the output operator, and not the actual number of bytes sent over the network. The latter is more useful, as it can be used to discover internal network bottlenecks.

## Release notes

( X) Release notes are required, with the following suggested text:

```markdown
## Section
* Make the stage output in the explain analyze, display the actual size of data sent over the network
```


lukasz-stec:
It turns out `FTE` does not use `createBufferResultResponse`, so tracking there only won't work for this mode.

PR #27744, opened 2025-12-23
Configure Oracle UCP wait connection timeout

cccs-nik:
## Description
This change adds support for configuring a connection wait timeout for the Oracle connector’s connection pool.

When all connections in the pool are borrowed, connection requests can fail with UCP-45064 during bursts of queries where connections would become available shortly after. This can cause otherwise healthy workloads to fail unnecessarily.

The new catalog property, oracle.connection-pool.wait-timeout, allows connection requests to wait for a free connection for a configurable amount of time. By setting a higher wait timeout, queries can be serviced during brief bursts of high concurrency instead of occasionally failing.

This is particularly useful for bursty workloads with many short-running queries, where brief contention on the pool can lead to query failures despite sufficient database capacity and a high maximum connection pool size, because the pool cannot always scale fast enough to handle sudden bursts.

## Release notes

```markdown
## Oracle
* Add support for configuring a connection wait timeout via the `oracle.connection-pool.wait-timeout` catalog property. ({issue}`27744`)
```

### DIFF
--- a/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
+++ b/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
@@ -26,6 +26,7 @@ import java.math.RoundingMode;
 import java.util.Optional;
 
 import static java.util.concurrent.TimeUnit.MINUTES;
+import static java.util.concurrent.TimeUnit.SECONDS;
 
 @DefunctConfig("oracle.disable-automatic-fetch-size")
 public class OracleConfig
@@ -38,6 +39,7 @@ public class OracleConfig
     private int connectionPoolMinSize = 1;
     private int connectionPoolMaxSize = 30;
     private Duration inactiveConnectionTimeout = new Duration(20, MINUTES);
+    private Duration connectionPoolWaitDuration = new Duration(3, SECONDS);
     private Integer fetchSize;
 
     public boolean isSynonymsEnabled()
@@ -139,6 +141,20 @@ public class OracleConfig
     public OracleConfig setInactiveConnectionTimeout(Duration inactiveConnectionTimeout)
     {
         this.inactiveConnectionTimeout = inactiveConnectionTimeout;
+        return this;
+    }
+
+    @NotNull
+    public Duration getConnectionPoolWaitDuration()
+    {
+        return connectionPoolWaitDuration;
+    }
+
+    @Config("oracle.connection-pool.wait-duration")
+    @ConfigDescription("Maximum amount of time a request will wait to obtain an available connection from the pool if all connections are currently in use")
+    public OracleConfig setConnectionPoolWaitDuration(Duration connectionPoolWaitDuration)
+    {
+        this.connectionPoolWaitDuration = connectionPoolWaitDuration;
         return this;
     }
 
--- a/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OraclePoolConnectionFactory.java
+++ b/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OraclePoolConnectionFactory.java
@@ -43,6 +43,7 @@ public class OraclePoolConnectionFactory
             int connectionPoolMinSize,
             int connectionPoolMaxSize,
             Duration inactiveConnectionTimeout,
+            Duration connectionPoolWaitDuration,
             OpenTelemetry openTelemetry)
             throws SQLException
     {
@@ -59,6 +60,7 @@ public class OraclePoolConnectionFactory
         dataSource.setValidateConnectionOnBorrow(true);
         dataSource.setConnectionProperties(connectionProperties);
         dataSource.setInactiveConnectionTimeout(toIntExact(inactiveConnectionTimeout.roundTo(SECONDS)));
+        dataSource.setConnectionWaitDuration(connectionPoolWaitDuration.toJavaTime());
         credentialProvider.getConnectionUser(Optional.empty())
                 .ifPresent(user -> {
                     try {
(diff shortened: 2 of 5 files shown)


plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OraclePoolConnectionFactory.java
```diff
@@ -59,6 +60,7 @@ public OraclePoolConnectionFactory(
         dataSource.setValidateConnectionOnBorrow(true);
         dataSource.setConnectionProperties(connectionProperties);
         dataSource.setInactiveConnectionTimeout(toIntExact(inactiveConnectionTimeout.roundTo(SECONDS)));
+        dataSource.setConnectionWaitTimeout(toIntExact(connectionWaitTimeout.roundTo(SECONDS)));
```
ebyhr: `setConnectionWaitTimeout` is deprecated. Can we use `setConnectionWaitDuration` instead?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
```diff
@@ -38,6 +39,7 @@ public class OracleConfig
     private int connectionPoolMinSize = 1;
     private int connectionPoolMaxSize = 30;
     private Duration inactiveConnectionTimeout = new Duration(20, MINUTES);
+    private Duration connectionWaitTimeout = new Duration(5, SECONDS);
```
ebyhr: How did you choose 5s as the default value?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OraclePoolConnectionFactory.java
cccs-nik: Done!

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
cccs-nik: It's just an arbitrary value I picked. I have it configured to 30s for my deployments but figured that might be too high for a broad recommendation.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
```diff
@@ -142,6 +144,20 @@ public OracleConfig setInactiveConnectionTimeout(Duration inactiveConnectionTime
...
     }
 
+    @NotNull
+    public Duration getConnectionWaitDuration()
+    {
+        return connectionWaitDuration;
+    }
+
+    @Config("oracle.connection-pool.wait-duration")
+    @ConfigDescription("How long to wait for a used connection to be released by a client")
```
chenjian2664: I found "specifies the maximum amount of time an application request will wait to obtain an available connection from the pool if all connections are currently in use", could you refer this can rephrase it

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
chenjian2664: in https://docs.oracle.com/cd/B28359_01/java.111/e10788.pdf "Setting the Connection Wait Timeout" suggests that the default value is 3s :
```
The connection wait timeout value represents seconds. A value of 0 indicates that the
feature is disabled. The default value is set to 3 seconds. 
```

Can we just follow this as well

cccs-nik: Yes absolutely. I changed the default to 3s and updated the description of the configuration option to match the new wait-duration setting as requested earlier.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
```diff
@@ -38,6 +39,7 @@ public class OracleConfig
     private int connectionPoolMinSize = 1;
     private int connectionPoolMaxSize = 30;
     private Duration inactiveConnectionTimeout = new Duration(20, MINUTES);
+    private Duration connectionWaitDuration = new Duration(3, SECONDS);
```
ebyhr:
```suggestion
    private Duration connectionPoolWaitDuration = new Duration(3, SECONDS);
```

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleConfig.java
```diff
@@ -142,6 +144,20 @@ public OracleConfig setInactiveConnectionTimeout(Duration inactiveConnectionTime
...
     }
 
+    @NotNull
+    public Duration getConnectionWaitDuration()
+    {
+        return connectionWaitDuration;
+    }
+
+    @Config("oracle.connection-pool.wait-duration")
+    @ConfigDescription("Maximum amount of time an application request will wait to obtain an available connection from the pool if all connections are currently in use")
```
ebyhr:
```suggestion
    @ConfigDescription("Maximum amount of time a request will wait to obtain an available connection from the pool if all connections are currently in use")
```

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OraclePoolConnectionFactory.java
```diff
@@ -59,6 +60,7 @@ public OraclePoolConnectionFactory(
         dataSource.setValidateConnectionOnBorrow(true);
         dataSource.setConnectionProperties(connectionProperties);
         dataSource.setInactiveConnectionTimeout(toIntExact(inactiveConnectionTimeout.roundTo(SECONDS)));
+        dataSource.setConnectionWaitDuration(java.time.Duration.ofSeconds(connectionWaitDuration.roundTo(SECONDS)));
```
ebyhr: Airlift Duration provides `toJavaTime()` method.

PR #27745, opened 2025-12-23
Add the monaco editor to the Preview UI distribution

gertjanal:
## Description
The Preview UI uses the monaco editor to display queries. This uses https://cdn.jsdelivr.net/npm/monaco-editor@0.52.2/min/vs/loader.js to load the npm module. With this fix, the npm module is added and used to the dist via vite. This fix is useful for air gapped networks.

## Before
Screenshot 2025-12-23 at 22-29-48 Trino - Preview UI
(Fails to load sql view)

##  After
Screenshot 2025-12-23 at 22-19-20 Trino - Preview UI

## Release notes

### DIFF
--- a/core/trino-web-ui/src/main/resources/webapp-preview/src/App.tsx
+++ b/core/trino-web-ui/src/main/resources/webapp-preview/src/App.tsx
@@ -34,10 +34,13 @@ import { darkTheme, lightTheme } from './theme'
 import trinoLogo from './assets/trino.svg'
 import { QueryDetails } from './components/QueryDetails'
 import { WorkerStatus } from './components/WorkerStatus'
+import { loader } from '@monaco-editor/react'
+import * as monaco from 'monaco-editor'
 
 const App = () => {
     const config = useConfigStore()
     const prefersDarkMode = useMediaQuery('(prefers-color-scheme: dark)')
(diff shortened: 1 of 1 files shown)


gertjanal:
> It disables the CDN but I think monace also requires explicit web workers handling in vite. If I’m not mistaken without that the editor can break in prod.
> 
> can we add https://www.npmjs.com/package/vite-plugin-monaco-editor (or similar) to wire monaco workers correctly and keep everything bundled locally? Also would be cool to include only the requires language packs too to keep the bundle smaller.

The current fix works good enough for me, we only use the editor to render the SQL view, right?

koszti:
> we only use the editor to render the SQL view, right?

monaco is currently used for rendering queries, java exceptions, and query/task JSON plans. And we’re likely to reuse it in the query UI at some stage later.

Please forget what I asked above, I checked your PR and it works just fine. I think it's good to go.

And for reference, I checked total bundle size and it's ~5 MB gzip, which is fine. Monaco accounts for about 65% of that which I think is also acceptable given how feature-rich the editor is. If bundle size ever becomes a concern we can optimize monaco bundle size separately. It doesn't look trivial.

koszti (approved):

wendigo (approved):

PR #25550, opened 2025-04-10
Remove support for deletion vectors in Iceberg

ebyhr:
## Description

This PR reverts #24882. This isn't a breaking change because #24882 isn't yet released. 

The upcoming Iceberg 1.9.0 makes row-lineage and default-values required in their V3 spec.
Shipping deletion vector without supporting those features will be troublesome. 

I think we should wait until the V3 spec has stabilized. 

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #27753, opened 2025-12-24
[Iceberg v3] VARIANT type

### LINKED ISSUE #24538: Add support for variant type in Iceberg
Relates to:
* https://github.com/apache/iceberg/issues/10392
* https://github.com/apache/iceberg/pull/11831
* https://docs.google.com/document/d/1sq70XDiWJ2DemWyA5dVB80gKzwi0CWoM0LOWM7VJVd8/edit?tab=t.0#heading=h.rt0cvesdzsj7

dain:
This PR introduces support for the `VARIANT` type in Trino, based on the
Apache Iceberg Variant specification.

Key points:

- Adds a new `VARIANT` type to the SPI and engine
- Implements casts to and from `VARIANT`, including arrays, rows, maps,
  and JSON
- Supports dereferencing using the SQL subscript operator
- Adds client protocol support for `VARIANT`
- Adds Iceberg integration for `VARIANT` 
- Documents `VARIANT` functions, operators, casts, and Iceberg constraints

Iceberg notes:

- `VARIANT` is supported only for Iceberg tables using format version 3
- Variant is currently transported to clients as variant spec binary format.
- Variant is not integrated into clients, and clients will either error or render as binary.

This PR does not modify existing Delta Lake variant-to-JSON behavior.

- fixes https://github.com/trinodb/trino/issues/24538

## Release notes

```markdown
# Iceberg
* Add experimental support for the `VARIANT` type for Iceberg tables using format version 3.

# SPI
* Add `VARIANT` type defined by the Iceberg specification.
```

### DIFF
--- a/client/trino-client/src/main/java/io/trino/client/ClientStandardTypes.java
+++ b/client/trino-client/src/main/java/io/trino/client/ClientStandardTypes.java
@@ -47,6 +47,7 @@ public final class ClientStandardTypes
     public static final String MAP = "map";
     public static final String JSON = "json";
     public static final String JSON_2016 = "json2016";
+    public static final String VARIANT = "variant";
     public static final String IPADDRESS = "ipaddress";
     public static final String UUID = "uuid";
     public static final String GEOMETRY = "Geometry";
--- a/client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
+++ b/client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
@@ -13,12 +13,14 @@
  */
 package io.trino.client;
 
+import com.fasterxml.jackson.core.JsonGenerator;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.core.JsonToken;
 import com.fasterxml.jackson.databind.json.JsonMapper;
 import com.google.common.collect.ImmutableList;
 
 import java.io.IOException;
+import java.io.StringWriter;
 import java.util.Base64;
 import java.util.HashMap;
 import java.util.LinkedList;
@@ -64,6 +66,8 @@ import static io.trino.client.ClientStandardTypes.TINYINT;
 import static io.trino.client.ClientStandardTypes.UUID;
 import static io.trino.client.ClientStandardTypes.VARBINARY;
 import static io.trino.client.ClientStandardTypes.VARCHAR;
+import static io.trino.client.ClientStandardTypes.VARIANT;
+import static io.trino.client.JsonIterators.createJsonFactory;
 import static java.lang.String.format;
 import static java.util.Collections.unmodifiableList;
 import static java.util.Collections.unmodifiableMap;
@@ -81,6 +85,7 @@ public final class JsonDecodingUtils
     private static final RealDecoder REAL_DECODER = new RealDecoder();
     private static final BooleanDecoder BOOLEAN_DECODER = new BooleanDecoder();
     private static final StringDecoder STRING_DECODER = new StringDecoder();
+    private static final VariantDecoder VARIANT_DECODER = new VariantDecoder();
     private static final Base64Decoder BASE_64_DECODER = new Base64Decoder();
     private static final ObjectDecoder OBJECT_DECODER = new ObjectDecoder();
 
@@ -117,6 +122,8 @@ public final class JsonDecodingUtils
                 return REAL_DECODER;
             case BOOLEAN:
                 return BOOLEAN_DECODER;
+            case VARIANT:
+                return VARIANT_DECODER;
             case ARRAY:
                 return new ArrayDecoder(signature);
             case MAP:
@@ -284,6 +291,21 @@ public final class JsonDecodingUtils
                 throw illegalToken(parser);
             }
             return parser.getValueAsString();
+        }
+    }
+
+    private static class VariantDecoder
+            implements TypeDecoder
+    {
+        @Override
+        public Object decode(JsonParser parser)
+                throws IOException
+        {
+            StringWriter writer = new StringWriter();
+            try (JsonGenerator generator = createJsonFactory().createGenerator(writer)) {
+                generator.copyCurrentStructure(parser);
+            }
+            return writer.toString();
         }
     }
 
--- a/core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
@@ -17,8 +17,10 @@ import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.VerifyException;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
+import io.trino.connector.system.GlobalSystemConnector;
 import io.trino.spi.TrinoException;
 import io.trino.spi.function.BoundSignature;
+import io.trino.spi.function.CatalogSchemaFunctionName;
 import io.trino.spi.function.FunctionId;
 import io.trino.spi.function.LongVariableConstraint;
 import io.trino.spi.function.Signature;
@@ -30,7 +32,6 @@ import io.trino.spi.type.TypeParameter;
 import io.trino.spi.type.TypeSignature;
 import io.trino.sql.analyzer.TypeSignatureProvider;
 import io.trino.type.FunctionType;
-import io.trino.type.JsonType;
 import io.trino.type.TypeCoercion;
 import io.trino.type.UnknownType;
 
@@ -47,10 +48,13 @@ import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Verify.verify;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableSortedMap.toImmutableSortedMap;
+import static io.trino.metadata.GlobalFunctionCatalog.BUILTIN_SCHEMA;
+import static io.trino.metadata.OperatorNameUtil.mangleOperatorName;
 import static io.trino.metadata.SignatureBinder.RelationshipType.EXACT;
 import static io.trino.metadata.SignatureBinder.RelationshipType.EXPLICIT_COERCION_FROM;
 import static io.trino.metadata.SignatureBinder.RelationshipType.EXPLICIT_COERCION_TO;
 import static io.trino.metadata.SignatureBinder.RelationshipType.IMPLICIT_COERCION;
+import static io.trino.spi.function.OperatorType.CAST;
 import static io.trino.sql.analyzer.TypeSignatureProvider.fromTypes;
 import static io.trino.type.TypeCalculation.calculateLiteralValue;
 import static io.trino.type.TypeCoercion.isCovariantTypeBase;
@@ -598,7 +602,8 @@ public class SignatureBinder
 
     private boolean canCast(Type fromType, Type toType)
     {
-        if (toType instanceof UnknownType) {
+        // NULL can be cast to any type; avoid re-entering coercion cache.
+        if (fromType instanceof UnknownType || toType instanceof UnknownType) {
             return true;
         }
         if (fromType instanceof RowType) {
@@ -615,14 +620,14 @@ public class SignatureBinder
                 }
                 return true;
             }
-            if (toType instanceof JsonType) {
+            if (isRecursiveCastFromRow(toType)) {
                 return fromType.getTypeParameters().stream()
                         .allMatch(fromTypeParameter -> canCast(fromTypeParameter, toType));
             }
             return false;
         }
-        if (fromType instanceof JsonType) {
-            if (toType instanceof RowType) {
+        if (toType instanceof RowType) {
+            if (isRecursiveCastToRow(fromType)) {
                 return toType.getTypeParameters().stream()
                         .allMatch(toTypeParameter -> canCast(fromType, toTypeParameter));
             }
@@ -634,6 +639,61 @@ public class SignatureBinder
         catch (TrinoException e) {
             return false;
         }
+    }
+
+    /// Check if there is a recursive variadic CAST from ROW.
+    /// This needs special handling because the cast is applied to each field of ROW individually.
+    private boolean isRecursiveCastFromRow(Type toType)
+    {
+        return metadata.getFunctions(null, new CatalogSchemaFunctionName(GlobalSystemConnector.NAME, BUILTIN_SCHEMA, mangleOperatorName(CAST))).stream()
+                .map(cast -> cast.functionMetadata().getSignature())
+                .anyMatch(signature -> isRecursiveCastFromRow(toType, signature));
+    }
+
+    private static boolean isRecursiveCastFromRow(Type toType, Signature signature)
+    {
+        // the return type must match toType
+        if (!toType.getTypeSignature().equals(signature.getReturnType())) {
+            return false;
+        }
+
+        // there must be exactly one type variable constraint and no long variable constraints
+        if (signature.getTypeVariableConstraints().size() != 1 || !signature.getLongVariableConstraints().isEmpty()) {
+            return false;
+        }
+        TypeVariableConstraint typeVariableConstraint = signature.getTypeVariableConstraints().getFirst();
+
+        // The argument type must be a type variable with variadic bound of "row"
+        return signature.getArgumentTypes().size() == 1 &&
+                signature.getArgumentTypes().getFirst().getBase().equals(typeVariableConstraint.getName()) &&
+                typeVariableConstraint.isRowType();
+    }
+
+    /// Check if there is a recursive variadic CAST to ROW.
+    /// This needs special handling because the cast is applied to each field of ROW individually.
+    private boolean isRecursiveCastToRow(Type fromType)
+    {
+        return metadata.getFunctions(null, new CatalogSchemaFunctionName(GlobalSystemConnector.NAME, BUILTIN_SCHEMA, mangleOperatorName(CAST))).stream()
+                .map(cast -> cast.functionMetadata().getSignature())
+                .anyMatch(signature -> isRecursiveCastToRow(fromType, signature));
+    }
+
+    private static boolean isRecursiveCastToRow(Type fromType, Signature signature)
+    {
+        // the argument type must match fromType
+        if (signature.getArgumentTypes().size() != 1 || !fromType.getTypeSignature().equals(signature.getArgumentTypes().getFirst())) {
+            return false;
+        }
+
+        // there must be exactly one type variable constraint and no long variable constraints
+        if (signature.getTypeVariableConstraints().size() != 1 || !signature.getLongVariableConstraints().isEmpty()) {
+            return false;
+        }
+        TypeVariableConstraint typeVariableConstraint = signature.getTypeVariableConstraints().getFirst();
+
+        // The return type must be a type variable with variadic bound of "row"
+        return signature.getReturnType().getBase().equals(typeVariableConstraint.getName()) &&
+                typeVariableConstraint.isRowType();
     }
 
     private static List<TypeSignature> getLambdaArgumentTypeSignatures(TypeSignature lambdaTypeSignature)
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayToVariantCast.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayToVariantCast.java
@@ -1,0 +1,86 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.operator.scalar;
+
+import com.google.common.collect.ImmutableList;
+import io.trino.metadata.SqlScalarFunction;
+import io.trino.spi.block.Block;
+import io.trino.spi.function.BoundSignature;
+import io.trino.spi.function.FunctionMetadata;
+import io.trino.spi.function.Signature;
+import io.trino.spi.type.ArrayType;
+import io.trino.spi.type.TypeSignature;
+import io.trino.spi.variant.Variant;
+import io.trino.util.variant.VariantWriter;
+
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+
+import static io.trino.spi.StandardErrorCode.INVALID_CAST_ARGUMENT;
+import static io.trino.spi.function.InvocationConvention.InvocationArgumentConvention.NEVER_NULL;
+import static io.trino.spi.function.InvocationConvention.InvocationReturnConvention.FAIL_ON_NULL;
+import static io.trino.spi.function.OperatorType.CAST;
+import static io.trino.spi.type.TypeSignature.arrayType;
+import static io.trino.spi.type.VariantType.VARIANT;
+import static io.trino.util.Failures.checkCondition;
+import static io.trino.util.variant.VariantUtil.canCastToVariant;
+import static java.lang.invoke.MethodType.methodType;
+
+public class ArrayToVariantCast
+        extends SqlScalarFunction
+{
+    public static final ArrayToVariantCast ARRAY_TO_VARIANT = new ArrayToVariantCast();
+
+    private static final MethodHandle METHOD_HANDLE;
+
+    static {
+        try {
+            METHOD_HANDLE = MethodHandles.lookup().findStatic(ArrayToVariantCast.class, "toVariant", methodType(Variant.class, VariantWriter.class, Block.class));
+        }
+        catch (IllegalAccessException | NoSuchMethodException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+    }
+
+    private ArrayToVariantCast()
+    {
+        super(FunctionMetadata.operatorBuilder(CAST)
+                .signature(Signature.builder()
+                        .castableToTypeParameter("T", VARIANT.getTypeSignature())
+                        .returnType(VARIANT)
+                        .argumentType(arrayType(new TypeSignature("T")))
+                        .build())
+                .build());
+    }
+
+    @Override
+    protected SpecializedSqlScalarFunction specialize(BoundSignature boundSignature)
+    {
+        ArrayType arrayType = (ArrayType) boundSignature.getArgumentTypes().getFirst();
+        checkCondition(canCastToVariant(arrayType), INVALID_CAST_ARGUMENT, "Cannot cast %s to VARIANT", arrayType);
+
+        VariantWriter writer = VariantWriter.create(arrayType);
+        MethodHandle methodHandle = METHOD_HANDLE.bindTo(writer);
+        return new ChoicesSpecializedSqlScalarFunction(
+                boundSignature,
+                FAIL_ON_NULL,
+                ImmutableList.of(NEVER_NULL),
+                methodHandle);
+    }
+
+    private static Variant toVariant(VariantWriter writer, Block block)
+    {
+        return writer.write(block);
+    }
+}
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/annotations/OperatorValidator.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/annotations/OperatorValidator.java
@@ -53,15 +53,18 @@ public final class OperatorValidator
                 break;
             case SUBSCRIPT:
                 validateOperatorSignature(operatorType, returnType, argumentTypes, 2);
-                checkArgument(argumentTypes.get(0).getBase().equals(StandardTypes.ARRAY) || argumentTypes.get(0).getBase().equals(StandardTypes.MAP), "First argument must be an ARRAY or MAP");
-                if (argumentTypes.get(0).getBase().equals(StandardTypes.ARRAY)) {
-                    checkArgument(argumentTypes.get(1).getBase().equals(StandardTypes.BIGINT), "Second argument must be a BIGINT");
-                    TypeSignature elementType = ((TypeParameter.Type) argumentTypes.get(0).getParameters().get(0)).type();
-                    checkArgument(returnType.equals(elementType), "[] return type does not match ARRAY element type");
-                }
-                else {
-                    TypeSignature valueType = ((TypeParameter.Type) argumentTypes.get(0).getParameters().get(1)).type();
-                    checkArgument(returnType.equals(valueType), "[] return type does not match MAP value type");
+                checkArgument(argumentTypes.get(0).getBase().equals(StandardTypes.ARRAY) || argumentTypes.get(0).getBase().equals(StandardTypes.MAP) || argumentTypes.get(0).getBase().equals(StandardTypes.VARIANT), "First argument must be an ARRAY, MAP, or VARIANT");
+                switch (argumentTypes.get(0).getBase()) {
+                    case StandardTypes.ARRAY -> {
+                        checkArgument(argumentTypes.get(1).getBase().equals(StandardTypes.BIGINT), "Second argument must be a BIGINT");
+                        TypeSignature elementType = ((TypeParameter.Type) argumentTypes.get(0).getParameters().get(0)).type();
+                        checkArgument(returnType.equals(elementType), "[] return type does not match ARRAY element type");
+                    }
+                    case StandardTypes.MAP -> {
+                        TypeSignature valueType = ((TypeParameter.Type) argumentTypes.get(0).getParameters().get(1)).type();
+                        checkArgument(returnType.equals(valueType), "[] return type does not match MAP value type");
+                    }
+                    default -> {}
                 }
                 break;
             case HASH_CODE:
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/timestamp/VarcharToTimestampCast.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/timestamp/VarcharToTimestampCast.java
@@ -13,7 +13,6 @@
  */
 package io.trino.operator.scalar.timestamp;
 
-import com.google.common.annotations.VisibleForTesting;
 import io.airlift.slice.Slice;
 import io.trino.spi.TrinoException;
 import io.trino.spi.function.LiteralParameter;
@@ -69,7 +68,6 @@ public final class VarcharToTimestampCast
         }
     }
 
-    @VisibleForTesting
     public static long castToShortTimestamp(int precision, String value)
     {
         checkArgument(precision <= MAX_SHORT_PRECISION, "precision must be less than max short timestamp precision");
@@ -125,7 +123,6 @@ public final class VarcharToTimestampCast
         return epochSecond * MICROSECONDS_PER_SECOND + rescale(fractionValue, actualPrecision, 6);
     }
 
-    @VisibleForTesting
     public static LongTimestamp castToLongTimestamp(int precision, String value)
     {
         checkArgument(precision > MAX_SHORT_PRECISION && precision <= MAX_PRECISION, "precision out of range");
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/timestamptz/VarcharToTimestampWithTimeZoneCast.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/timestamptz/VarcharToTimestampWithTimeZoneCast.java
@@ -81,7 +81,7 @@ public final class VarcharToTimestampWithTimeZoneCast
         }
     }
 
-    private static long toShort(int precision, String value, Function<String, ZoneId> zoneId)
+    public static long toShort(int precision, String value, Function<String, ZoneId> zoneId)
     {
         checkArgument(precision <= MAX_SHORT_PRECISION, "precision must be less than max short timestamp precision");
 
@@ -134,7 +134,7 @@ public final class VarcharToTimestampWithTimeZoneCast
         return packDateTimeWithZone(epochSecond * MILLISECONDS_PER_SECOND + millisOfSecond, getTimeZoneKey(zone.getId()));
     }
 
-    private static LongTimestampWithTimeZone toLong(int precision, String value, Function<String, ZoneId> zoneId)
+    public static LongTimestampWithTimeZone toLong(int precision, String value, Function<String, ZoneId> zoneId)
     {
         checkArgument(precision > MAX_SHORT_PRECISION && precision <= MAX_PRECISION, "precision out of range");
 
--- a/core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
+++ b/core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
@@ -44,8 +44,11 @@ import io.trino.spi.type.TinyintType;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.VarbinaryType;
 import io.trino.spi.type.VarcharType;
+import io.trino.spi.type.VariantType;
+import io.trino.spi.variant.Variant;
 import io.trino.type.SqlIntervalDayTime;
 import io.trino.type.SqlIntervalYearMonth;
+import io.trino.util.variant.VariantUtil;
 
 import java.io.IOException;
 import java.math.BigDecimal;
@@ -64,6 +67,7 @@ import static io.trino.spi.type.SmallintType.SMALLINT;
 import static io.trino.spi.type.TinyintType.TINYINT;
 import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static io.trino.spi.type.VarcharType.VARCHAR;
+import static io.trino.spi.type.VariantType.VARIANT;
 import static java.util.Objects.requireNonNull;
 
 public final class JsonEncodingUtils
@@ -79,6 +83,7 @@ public final class JsonEncodingUtils
     private static final TinyintEncoder TINYINT_ENCODER = new TinyintEncoder();
     private static final VarcharEncoder VARCHAR_ENCODER = new VarcharEncoder();
     private static final VarbinaryEncoder VARBINARY_ENCODER = new VarbinaryEncoder();
+    private static final VariantEncoder VARIANT_ENCODER = new VariantEncoder();
 
     public static TypeEncoder[] createTypeEncoders(Session session, List<Type> types)
     {
@@ -106,6 +111,7 @@ public final class JsonEncodingUtils
             case VarcharType _ -> VARCHAR_ENCODER;
             case VarbinaryType _ -> VARBINARY_ENCODER;
             case CharType charType -> new CharEncoder(charType.getLength());
+            case VariantType _ -> VARIANT_ENCODER;
             // TODO: add specialized Short/Long decimal encoders
             case ArrayType arrayType -> new ArrayEncoder(arrayType, createTypeEncoder(arrayType.getElementType(), supportsParametricDateTime));
             case MapType mapType -> new MapEncoder(mapType, createTypeEncoder(mapType.getValueType(), supportsParametricDateTime));
@@ -309,6 +315,23 @@ public final class JsonEncodingUtils
             // Optimization: avoid copying Slice to byte array
             Slice slice = VARBINARY.getSlice(block, position);
             generator.writeBinary(slice.byteArray(), slice.byteArrayOffset(), slice.length());
+        }
+    }
+
+    private static final class VariantEncoder
+            implements TypeEncoder
+    {
+        @Override
+        public void encode(JsonGenerator generator, Block block, int position)
+                throws IOException
+        {
+            if (block.isNull(position)) {
+                generator.writeNull();
+                return;
+            }
+
+            Variant variant = VARIANT.getObject(block, position);
+            generator.writeRawValue(VariantUtil.asJson(variant).toStringUtf8());
         }
     }
 
--- a/core/trino-main/src/main/java/io/trino/type/VariantOperators.java
+++ b/core/trino-main/src/main/java/io/trino/type/VariantOperators.java
@@ -1,0 +1,400 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.type;
+
+import io.airlift.slice.Slice;
+import io.trino.spi.TrinoException;
+import io.trino.spi.function.LiteralParameter;
+import io.trino.spi.function.LiteralParameters;
+import io.trino.spi.function.ScalarOperator;
+import io.trino.spi.function.SqlNullable;
+import io.trino.spi.function.SqlType;
+import io.trino.spi.type.LongTimestamp;
+import io.trino.spi.type.LongTimestampWithTimeZone;
+import io.trino.spi.type.StandardTypes;
+import io.trino.spi.variant.Header;
+import io.trino.spi.variant.Variant;
+import io.trino.util.variant.VariantUtil;
+import io.trino.util.variant.VariantWriter;
+
+import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.function.OperatorType.CAST;
+import static io.trino.spi.function.OperatorType.SUBSCRIPT;
+import static io.trino.spi.type.DateTimeEncoding.unpackMillisUtc;
+import static java.lang.Float.intBitsToFloat;
+import static java.lang.Math.multiplyExact;
+import static java.lang.Math.toIntExact;
+import static java.util.Locale.ENGLISH;
+
+public final class VariantOperators
+{
+    private VariantOperators() {}
+
+    @ScalarOperator(SUBSCRIPT)
+    @SqlType(StandardTypes.VARIANT)
+    public static Variant dereference(@SqlType(StandardTypes.VARIANT) Variant value, @SqlType(StandardTypes.BIGINT) long index)
+    {
+        if (value.basicType() != Header.BasicType.ARRAY) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "VARIANT value is %s, not an array".formatted(variantTypeName(value)));
+        }
+        checkArrayIndex(index, value.getArrayLength());
+        return value.getArrayElement(toIntExact(index) - 1);
+    }
+
+    private static void checkArrayIndex(long index, int arrayLength)
+    {
+        if (index == 0) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "VARIANT array indices start at 1");
+        }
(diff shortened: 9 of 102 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -53,7 +53,7 @@
 public class IcebergConfig
 {
     public static final int FORMAT_VERSION_SUPPORT_MIN = 1;
-    public static final int FORMAT_VERSION_SUPPORT_MAX = 2;
+    public static final int FORMAT_VERSION_SUPPORT_MAX = 3;
```
raunaqmorarka, line 58: Is it safe to bump the supported iceberg version without support for some of the other V3 features ?
E.g. we don't support reading deletion vectors yet and if other engines write those, then we may have correctness issues or errors.
It looks we backed out changes for deletion vectors previously https://github.com/trinodb/trino/pull/25550 because V3 support requires handling of row lineage as well.

raunaqmorarka: You could consider adding variant type support to delta lake connector first, as we already support variant there. That way the rest of the changes in this PR won't need to wait for iceberg V3.

martint:
Thanks for putting this up. We talked about it a few weeks ago over Slack, so I wanted to capture my thoughts here.

SQL 2023 defines a rich SQL/JSON data model and set of operations and semantics that's almost a 1-to-1 mapping with Iceberg Variant. Note that, importantly, SQL/JSON does not mandate a textual encoding, but only conversion functions to and from text that produce RFC 8259 JSON.

image (4)
image (3)
image (2)
image (1)

Instead of adding a new data type, we should:
* Change Trino's JSON data type to use a more efficient encoding internally (we can adopt Variant's encoding if that's suitable enough)
* Add support for conversions to/from the additional data types (datetime types, etc)
* Adjust the existing JSON functions (json_query, json_value, json_table, etc) to operate on the new encoding
* Map Iceberg's and Delta Variant to Trino JSON. We already do that in Delta, but we'll need to adjust the implementation a bit to account for the change in format.

This gives us:
* A single data type that has all the semantics we need. That means less confusion for users on when to use one vs the other.
* Trino's JSON type, which predates SQL's standardization of the JSON type, up to compliance with the specification
* All the [functions and syntax](https://trino.io/docs/current/functions/json.html) that work for JSON will work out of the box, without having to define a parallel set of functions to operate on Variant.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
dain: I don't know anything about delta, but I expect it will take less than 10 mintues to integrate for someone that knows what they are doing.  Basically, instead of calling the "deserialize variant to json" method, you call "deserialize variant to variant" method.

dain:
@martint Thanks for writing this up — I agree that SQL:2023 defines a rich SQL/JSON data model, and at the abstract model level there is significant overlap with Iceberg Variant (null / scalar / array / object / sequence).

However, I think it is important to separate data model similarity from type semantics, interoperability guarantees, and physical representation.

This PR is intentionally focused on first-class Iceberg Variant support. That implies a number of constraints that are part of the Iceberg specification and cannot be relaxed or evolved independently:

* A closed and spec-defined scalar set
* A fixed binary encoding that must round-trip exactly
* Stable semantics across engines for read/write interoperability

Concretely, Iceberg Variant has a number of limitations that are part of the contract:

* Timestamps are limited to microsecond and nanosecond precision
* There is no stored time zone information
* There are no interval types
* There are no IP address types
* There are no geometry or geography types (including Iceberg’s own geo extensions)
* Equality and comparison semantics are explicitly defined by the Variant spec and differ from SQL equality in important ways

Because these constraints are fundamental to Variant, equating it with SQL/JSON is dangerous in both directions:

* SQL/JSON allows (and encourages) richer and potentially extensible scalar semantics that Variant cannot support without a spec revision
* Treating JSON as backed by Variant would implicitly restrict JSON semantics and create surprising behavior or broken round-tripping guarantees

For these reasons, representing Variant as a distinct Trino type is deliberate. It makes the interoperability boundary explicit and avoids accidental promises about future extensibility or semantic compatibility.

It is also important to note that JSON is already a trivial cast away. This PR supports casts in and out of Variant, including JSON ↔ VARIANT, and users who want JSON semantics can get them immediately via explicit casts. The hard problem here is not JSON manipulation — it is correctness and fidelity of Variant itself.

More broadly, if Trino were to implement deeper SQL/JSON compliance, Variant is not a good internal encoding for that purpose. Variant values are expensive and difficult to construct:

* Values must be sorted according to Variant rules
* Sizes are encoded as prefixes, which prevents true streaming construction
* Building a Variant often requires multiple passes over the data (this PR uses three passes)

These characteristics are acceptable and necessary for an interchange format, but they are not a good foundation for a general-purpose, extensible SQL/JSON implementation.

Refactoring Trino’s JSON type to adopt a new internal encoding and updating the full SQL/JSON function surface would be a valuable effort, but it is a large, orthogonal project with broad impact. This PR is intentionally scoped to deliver correct, interoperable Iceberg Variant support without blocking on that work.

In short, this PR is about delivering real Variant semantics with strong interoperability guarantees. SQL/JSON compliance and JSON internals can evolve independently without conflating the two or weakening either.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
dain: As for v3, I think a reasonable solution is to add a config property that let's you use v3 until DVs are implemented.  This PR is huge, and we should check it in instead of waiting.

I'm going to take a look at adding DV support because it seems trivial, but I don't know if I'll have time to finish it.

dain: Actually, the DV code looks simple.  I've already hacked together the read side, and should be able to quicly get the write side working.

ebyhr: Only supporting DV is insufficient. Please also add tests for default values and row lineage.

The connector throws exceptions for unsupported types (for example, timestamp nanos and geospatial types), which makes it safer. However, there are still cases where we could violate the spec if unsupported features are neither properly supported nor explicitly rejected.

ebyhr (dismissed):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
dain: I added a commit to the end that checks for unsupported v3 features and just fails the queries.  So if you have a table with default values, v3 encryption, or DV, you get an exception.  We should not need to do anything for row lineage since we do not expose the new columns (so you would get an exception if you try to use them), and the existing optimizer code checks if the version is exactly 2 (so it fails on v3).

I believe this is every feature that effects correctness, and they should all just fail.

martint:
> This PR is intentionally focused on first-class Iceberg Variant support. That implies a number of constraints that are part of the Iceberg specification and cannot be relaxed or evolved independently:

In that case, we should keep the type and functions confined to the Iceberg plugin.

> Concretely, Iceberg Variant has a number of limitations that are part of the contract:
>    Timestamps are limited to microsecond and nanosecond precision
>    There is no stored time zone information

That's fair. There would need to be some conversion/checking on write to ensure the contents have the expected precision that the connector needs. Alternatively, we could have an envelope with metadata about the contained data types so that the connector can decide whether to store as is or convert.

>    There are no interval types
>    There are no IP address types
>    There are no geometry or geography types (including Iceberg’s own geo extensions)

SQL/JSON doesn't have those either, so that's not a problem.

>    Equality and comparison semantics are explicitly defined by the Variant spec and differ from SQL equality in important ways

This could be a problem, yes. But in what ways? Can you elaborate?

> SQL/JSON allows (and encourages) richer and potentially extensible scalar semantics that Variant cannot support without a spec revision

Not really. It doesn't encourage extensions -- that's a choice we would make.

> It is also important to note that JSON is already a trivial cast away. This PR supports casts in and out of Variant, including JSON ↔ VARIANT, and users who want JSON semantics can get them immediately via explicit casts.

I could get behind that as long as we don't then go and try to replicate all the operations we have for JSON to work for variant. It's a slippery slope.

> These characteristics are acceptable and necessary for an interchange format, but they are not a good foundation for a general-purpose, extensible SQL/JSON implementation.

Not sure I understand this. Variant is not just an interchange format. If that were the purpose, we'd might as well use varbinary instead.

> Refactoring Trino’s JSON type to adopt a new internal encoding and updating the full SQL/JSON function surface would be a valuable effort, but it is a large, orthogonal project with broad impact.

There are only a handful of functions that expect JSON type at this point: 
* is_json_scalar, json_array_contains, json_array_length, which are pretty trivial
* json_array_get, which is currently broken, anyway
* json_extract, json_extract_scalar, json_format, json_parse, which could be implemented via an internal conversion to json first and then invoke the existing logic
* casts, which you've already implemented

> In short, this PR is about delivering real Variant semantics with strong interoperability guarantees. SQL/JSON compliance and JSON internals can evolve independently without conflating the two or weakening either.

Databases implement variant semantics in different ways, with no interoperable semantics. In particular databrick's, snowflake's and Iceberg variants differ from each other in important ways. Other databases use different types to achieve those semantics:

| Database | Variant / Equivalent |
|----------|-----------------|
| Snowflake   | VARIANT |
| SQL Server | sql_variant |
| Redshift      | SUPER |
| Databricks  | VARIANT |
| Oracle         | ANYDATA, JSON |
| PostgreSQL | JSON, JSONB |
| MySQL | JSON |
| SQLite | Dynamic typing |
| MongoDB | BSON documents |
| BigQuery | JSON, STRUCT |

So, if by interoperable we mean interoperable with other engines that read/write Iceberg, this should by an Iceberg connector feature, not an engine-level feature.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
findepi: with current state of the PR, is Trino allowed to v3 tables?
does it fulfill all the writer duties?

findepi:
> > It is also important to note that JSON is already a trivial cast away. This PR supports casts in and out of Variant, including JSON ↔ VARIANT, and users who want JSON semantics can get them immediately via explicit casts.
> 
> I could get behind that as long as we don't then go and try to replicate all the operations we have for JSON to work for variant

As much as i don't like duplicated logic especially if it's complex, I also want Trino to be less performant than it could be (and less performant than other systems).

For example, having to convert whole variant document to json type in order to perform a path lookup  would sound wasteful. 
FWIW Snowflake has JSON functions, but doesn't have a JSON type. All the JSON functions operate on their VARIANT type, which is quite similar to (but not same as) Parquet/Iceberg VARIANT.

My point is not really to decide on the direction here. Rather, I only want us not to make promises that we may regret in the future, and to keep the door open for future improvements around Trino's JSON and VARIANT story.

dain:
> > This PR is intentionally focused on first-class Iceberg Variant support. That implies a number of constraints that are part of the Iceberg specification and cannot be relaxed or evolved independently:
> 
> In that case, we should keep the type and functions confined to the Iceberg plugin.

The type is required by DeltaLake. I also think Hive supports this.  The problem is the stack type is a Java object and for these system to interop you need the type to be available to both and that means the SPI.  This is a type that is defined by a specification that is well established.  IMO we should just have this in the SPI, so any connector can use it.  I expect that this is something we might want to use in the snowflake connector.

I don't want to get into a debate about types, but the IMO the isolation of types into plugins does not work, and I think we should back this feature out of Trino.  Referencing a type by name and trying to do anything non trivial with the type is a total mess.  The complexity isn't worth the effort to keep types out. 

If you insist I'll do it because I want this feature in more than I want to debate this point, but I feel extremely strongly that this is the wrong direction (and I feel the same about the two other isolated types in trino).

> > Concretely, Iceberg Variant has a number of limitations that are part of the contract:
> > Timestamps are limited to microsecond and nanosecond precision
> > There is no stored time zone information
> 
> That's fair. There would need to be some conversion/checking on write to ensure the contents have the expected precision that the connector needs. Alternatively, we could have an envelope with metadata about the contained data types so that the connector can decide whether to store as is or convert.

My point is ther is no freedom.  I think we are much better off having connectors expose their native type and then they can trivially be cast to our internal JSON to use the JSON syntax and functions (this would be super trivial if we landed the postgress style casts). So Postgres can expose raw jsonb, redshift can expose it custom thing, and so on. You can manipulate them directly without loosing any fidelity.  If you want to move to a different type, you just cast.

IMO this is the best of both worlds.  You can work directly with your native type, and then you can move to something else if you want a different feature or if you want to interoperate with a different system.  JSON IMO isn't always the best.

> > There are no interval types
> > There are no IP address types
> > There are no geometry or geography types (including Iceberg’s own geo extensions)
> 
> SQL/JSON doesn't have those either, so that's not a problem.

This is bad in both directions.  What if Iceberg adds interval (there is a proposal), then what do you do?  Does SQL/JSON support UUID, because Iceberg Variant already has that.  There will always be an impendency mismatch here.

> > Equality and comparison semantics are explicitly defined by the Variant spec and differ from SQL equality in important ways
> 
> This could be a problem, yes. But in what ways? Can you elaborate?

The Iceberg spec defines equivilence classes, which tell you how you are supposed to extract and compare types.  One strange thing is that exact number, float, and decimal are separate types, and therefore not equal.

> > SQL/JSON allows (and encourages) richer and potentially extensible scalar semantics that Variant cannot support without a spec revision
> 
> Not really. It doesn't encourage extensions -- that's a choice we would make.

Again cuts both ways.  The main point is we do not control the variant spec.  Expanding it is bad, not fully support it is bad, not having choices in Trino would suck.

> > It is also important to note that JSON is already a trivial cast away. This PR supports casts in and out of Variant, including JSON ↔ VARIANT, and users who want JSON semantics can get them immediately via explicit casts.
> 
> I could get behind that as long as we don't then go and try to replicate all the operations we have for JSON to work for variant. It's a slippery slope.

I don't think we should be reimplementing the json operations.  We have JSON to do that.  I think the function set we already have (casts+dereference) is what we need. 

> > These characteristics are acceptable and necessary for an interchange format, but they are not a good foundation for a general-purpose, extensible SQL/JSON implementation.
> 
> Not sure I understand this. Variant is not just an interchange format. If that were the purpose, we'd might as well use varbinary instead.

Variant is a good for a format for Iceberg and Delta to store data.  It does not make a good format for in memory trino manipulation.  If we want a good format for dynamic structures like this, I can suggest a better design.

dain:
@ebyhr @raunaqmorarka @findepi, I split out the code to enable v3 and the commit to reject tables/commands for unsupported v3 features into PR #27786.  That PR is designed to hande and test all cases you have mentioned and any thing else I could think of.  The variant branch is not based on that PR.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
dain: I believe that the current state of this PR rejects anything not supported, and that we are fulfilling our duties for the write we do support (only insert today)

findepi:
> I think we are much better off having connectors expose their native type and then they can trivially be cast to our internal JSON to use the JSON syntax and functions

That would work but have undesired effect that direct field lookup is significantly faster on a large variant value compared with equivalent json function call.

I envision a future where Trino JSON type can be simply backed by Iceberg VARIANT value directly, without any data conversion and json functions can operate on those documents. I acknowledge that Trino JSON may want to be more type-rich than Iceberg VARIANT, so that runtime representation doesn't have to be the only one allowed.

Why? For intersection of two important assumptions

- Iceberg Variant is going to be popular source of unstructured data Trino will be processing in the future
  - it's not ubiquitous today, because it doesn't exist yet. But it's so cool that i bet it will be ubiquitous
- user queries often involve more than single-level field lookup
  - in real-world queries dealing with Snowflake Variant values, complex lookups are not uncommon. We definitely won't want to express them as chained single-step lookups for efficiency reasons, which will naturally lead to more demand for JSON path-like traversal of variant values.

My point is not really to decide on the direction here. Rather, I only want us not to make promises that we may regret in the future, and to keep the door open for future improvements around Trino's JSON and VARIANT story.

client/trino-client/src/main/java/io/trino/client/ClientStandardTypes.java
```diff
@@ -44,6 +44,7 @@ public final class ClientStandardTypes
     public static final String MAP = "map";
     public static final String JSON = "json";
     public static final String JSON_2016 = "json2016";
+    public static final String VARIANT = "variant";
```
findepi, line 50: i would consider calling this "iceberg variant" or perhaps "parquet variant".

dain: The type must have a globally unique name.  We at our core are a datalake engine, and variant now means iceberg variant, so we should go with that name.  Anything else is just going to be annoying for users for no real benefit to us.

dain:
I updated this PR with support for Iceberg Variant with ORC and AVRO formats, so all file formats are now supported

dain:
@ebyhr Can you remove your request changes from this?  The issue you are point out was resolved in the PRs we merge to enable Iceberg.

dain:
@coderabbit full review

core/trino-spi/src/main/java/io/trino/spi/variant/Header.java
```diff
@@ -0,0 +1,209 @@
...
+        TIMESTAMP_UTC_NANOS,
+        TIMESTAMP_NTZ_NANOS,
+        UUID;
+
+        private static PrimitiveType fromHeader(byte header)
+        {
+            // Bits 2-7 represent the primitive type
+            int primitiveTypeBits = (header & 0b1111_1100) >>> 2;
+            return values()[primitiveTypeBits];
+        }
```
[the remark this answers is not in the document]
dain: ArrayIndexOutOfBoundsException is fine here

core/trino-spi/src/main/java/io/trino/spi/variant/Metadata.java
```diff
@@ -0,0 +1,425 @@
...
+            int value = readOffset(metadata, position, offsetSize);
+            checkArgument(i != 0 || value == 0, "First dictionary offset must be 0");
+            checkArgument(i == 0 || value > previous, "dictionary offsets must be strictly increasing");
+            previous = value;
+            position += offsetSize;
+        }
+
+        int dictionaryStart = 1 + offsetSize + (dictionarySize + 1) * offsetSize;
+        int dictionaryLength = metadata.length() - dictionaryStart;
+        checkArgument(previous == dictionaryLength, "Last dictionary offset must equal dictionary length");
```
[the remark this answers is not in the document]
dain: Zero-length field names are not allowed by the iceberg spec. I added an explicit check/message for this.

core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
```diff
@@ -636,6 +641,59 @@ private boolean canCast(Type fromType, Type toType)
         }
     }
 
+    /// Check if there is a recursive variadic CAST from ROW.
+    /// This needs special handling because the cast is applied to each field of ROW individually.
+    private boolean isRecursiveCastFromRow(Type toType)
+    {
+        for (CatalogFunctionMetadata cast : metadata.getFunctions(null, new CatalogSchemaFunctionName(GlobalSystemConnector.NAME, BUILTIN_SCHEMA, mangleOperatorName(CAST)))) {
```
electrum: This can use
```java
metadata.getBuiltinFunctions(mangleOperatorName(CAST))

File: core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java [hunk 6]
```diff
@@ -636,6 +641,59 @@ private boolean canCast(Type fromType, Type toType)
...
+            }
+        }
+        return false;
+    }
+
+    /// Check if there is a recursive variadic CAST to ROW.
+    /// This needs special handling because the cast is applied to each field of ROW individually.
+    private boolean isRecursiveCastToRow(Type fromType)
+    {
+        for (CatalogFunctionMetadata cast : metadata.getFunctions(null, new CatalogSchemaFunctionName(GlobalSystemConnector.NAME, BUILTIN_SCHEMA, mangleOperatorName(CAST)))) {
```
electrum: This can use
```java
metadata.getBuiltinFunctions(mangleOperatorName(CAST))

File: core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java [hunk 7]
```diff
@@ -636,6 +641,59 @@ private boolean canCast(Type fromType, Type toType)
         }
     }
 
+    /// Check if there is a recursive variadic CAST from ROW.
+    /// This needs special handling because the cast is applied to each field of ROW individually.
+    private boolean isRecursiveCastFromRow(Type toType)
+    {
+        for (CatalogFunctionMetadata cast : metadata.getFunctions(null, new CatalogSchemaFunctionName(GlobalSystemConnector.NAME, BUILTIN_SCHEMA, mangleOperatorName(CAST)))) {
+            Signature signature = cast.functionMetadata().getSignature();
```
electrum: Consider extracting the logic to a static helper method `isRecursiveCastFromRow(Type toType, Signature signature)`, then you can simplify this to
```java
return metadata.getBuiltinFunctions(mangleOperatorName(CAST)).stream()
        .map(cast -> cast.functionMetadata().getSignature())
        .anyMatch(signature -> isRecursiveCastFromRow(toType, signature));

File: core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java [hunk 8]
```diff
@@ -636,6 +641,59 @@ private boolean canCast(Type fromType, Type toType)
...
+        }
+        return false;
+    }
+
+    /// Check if there is a recursive variadic CAST to ROW.
+    /// This needs special handling because the cast is applied to each field of ROW individually.
+    private boolean isRecursiveCastToRow(Type fromType)
+    {
+        for (CatalogFunctionMetadata cast : metadata.getFunctions(null, new CatalogSchemaFunctionName(GlobalSystemConnector.NAME, BUILTIN_SCHEMA, mangleOperatorName(CAST)))) {
+            Signature signature = cast.functionMetadata().getSignature();
```
electrum: Similar comment as above about static helper method

core/trino-spi/src/main/java/io/trino/spi/block/VariantBlockEncoding.java
```diff
@@ -0,0 +1,85 @@
...
+    public void writeBlock(BlockEncodingSerde blockEncodingSerde, SliceOutput sliceOutput, Block block)
+    {
+        VariantBlock variantBlock = (VariantBlock) block;
+
+        sliceOutput.appendInt(variantBlock.getPositionCount());
+
+        blockEncodingSerde.writeBlock(sliceOutput, variantBlock.getMetadata());
+        blockEncodingSerde.writeBlock(sliceOutput, variantBlock.getValues());
+
+        if (vectorizeNullBitPacking) {
```
electrum, line 59: Unrelated to this PR, but we should create unified versions of these methods that take a boolean for vectorized. That would cleanup all of the callers.

core/trino-main/src/test/java/io/trino/type/TestVariantType.java
```diff
@@ -0,0 +1,78 @@
...
+{
+    TestVariantType()
+    {
+        super(VARIANT, Variant.class, createTestBlock());
+    }
+
+    public static ValueBlock createTestBlock()
+    {
+        BlockBuilder blockBuilder = VARIANT.createBlockBuilder(null, 15);
+        VARIANT.writeObject(blockBuilder, Variant.ofDouble(11.11));
```
electrum: Why only doubles?

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -7,7 +7,8 @@
 Apache Iceberg is an open table format for huge analytic datasets. The Iceberg
 connector allows querying data stored in files written in Iceberg format, as
 defined in the [Iceberg Table Spec](https://iceberg.apache.org/spec/). The
-connector supports Apache Iceberg table spec versions 1 and 2.
+connector supports Apache Iceberg table spec versions 1 and 2. Support for
```
electrum: Nit: move sentence to its own line in markdown source

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -392,6 +397,12 @@ the following table:
   - `MAP(k,v)`
 :::
 
+:::{note}
+Iceberg `VARIANT` is supported only for tables using Iceberg format version `3`
+or higher. To create a table with `VARIANT` columns, set
+`format_version = 3` in the `WITH` clause. The default is `2`.
```
electrum, line 414: Add "format version" to sentence
```markdown
The default format version is `3`.
```

docs/src/main/sphinx/functions/variant.md
```diff
@@ -0,0 +1,159 @@
+# VARIANT functions and operators
+
+The `VARIANT` type represents a semi-structured value as defined by the
+[Apache Iceberg Variant specification](https://iceberg.apache.org/spec/#variant).
```
electrum: Looks like the correct link is now https://iceberg.apache.org/spec/#semi-structured-types

docs/src/main/sphinx/functions/variant.md
```diff
@@ -0,0 +1,159 @@
...
+* `NaN` is not equal to any value, including itself.
+
+## Subscript operator
+
+Elements of a `VARIANT` value can be accessed using the SQL subscript operator
+(`[]`). The result of a subscript operation is always a `VARIANT` value.
+
+### Objects
+
+When the underlying value is an object, the subscript operator uses a string
```
electrum: The language "the subscript operator uses" is a bit strange. Instead, tell the user what to do:

> When the underlying value is an object, use a `varchar` for the subscript operator:

docs/src/main/sphinx/functions/variant.md
```diff
@@ -0,0 +1,159 @@
...
+
+```sql
+variant_expression['key']
+```
+
+If the specified key does not exist in the object, the result is SQL `NULL`.
+
+### Arrays
+
+When the underlying value is an array, the subscript operator uses one-based
```
electrum: > When the underlying value is an array, use a `bigint` with one-based indexing:

docs/src/main/sphinx/functions/variant.md
```diff
@@ -0,0 +1,159 @@
...
+indexing, following the SQL standard for arrays:
+
+```sql
+variant_expression[index]
+```
+
+The following rules apply:
```
electrum: > The same rules as for SQL array indexes apply:

client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
```diff
@@ -298,6 +302,33 @@ public Object decode(JsonParser parser)
...
+            switch (parser.currentToken()) {
+                case START_ARRAY:
+                    break;
+                case VALUE_NULL:
+                    return null;
+                default:
+                    throw illegalToken(parser);
+            }
+
+            requireNonNull(parser.nextToken());
```
electrum: Throwing NPE here for bad input seems wrong

client/trino-client/src/main/java/io/trino/client/ClientStandardTypes.java
electrum, line 1: Per our discussion, we should add a new `ClientCapabilities` flag, maybe `RAW_VARIANT` or `BINARY_VARIANT`, and send the client JSON if they don't support it.

For JDBC, `getString()` should return the JSON value (this is what Snowflake does). For `getObject()`, we can return our own `TrinoVariant` in the JDBC package, similar to `TrinoIntervalDayTime`.

core/trino-main/src/main/java/io/trino/operator/scalar/annotations/OperatorValidator.java
```diff
@@ -53,13 +53,13 @@ public static void validateOperator(OperatorType operatorType, TypeSignature ret
                 break;
             case SUBSCRIPT:
                 validateOperatorSignature(operatorType, returnType, argumentTypes, 2);
-                checkArgument(argumentTypes.get(0).getBase().equals(StandardTypes.ARRAY) || argumentTypes.get(0).getBase().equals(StandardTypes.MAP), "First argument must be an ARRAY or MAP");
+                checkArgument(argumentTypes.get(0).getBase().equals(StandardTypes.ARRAY) || argumentTypes.get(0).getBase().equals(StandardTypes.MAP) || argumentTypes.get(0).getBase().equals(StandardTypes.VARIANT), "First argument must be an ARRAY, MAP, or VARIANT");
```
electrum, line 56: Make this a switch, then we don't need to duplicate the check

core/trino-main/src/main/java/io/trino/operator/scalar/timestamp/VarcharToTimestampCast.java
```diff
@@ -69,7 +69,6 @@ public static LongTimestamp castToLong(@LiteralParameter("p") long precision, @S
         }
     }
 
-    @VisibleForTesting
```
electrum, line 72: Where is this used in production code?

core/trino-main/src/main/java/io/trino/operator/scalar/timestamptz/VarcharToTimestampWithTimeZoneCast.java
```diff
@@ -81,7 +81,7 @@ public static LongTimestampWithTimeZone castToLong(@LiteralParameter("p") long p
         }
     }
 
-    private static long toShort(int precision, String value, Function<String, ZoneId> zoneId)
+    public static long toShort(int precision, String value, Function<String, ZoneId> zoneId)
```
electrum, line 84: Where are these used from outside?

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayToVariantCast.java
```diff
@@ -0,0 +1,88 @@
...
+    public static final ArrayToVariantCast ARRAY_TO_VARIANT = new ArrayToVariantCast();
+
+    private static final MethodHandle METHOD_HANDLE;
+
+    static {
+        try {
+            METHOD_HANDLE = MethodHandles.lookup().findStatic(ArrayToVariantCast.class, "toVariant", methodType(Variant.class, VariantWriter.class, Block.class));
+        }
+        catch (IllegalAccessException | NoSuchMethodException e) {
+            throw new TrinoException(GENERIC_INTERNAL_ERROR, e);
```
electrum: Same for all of these that use `MethodHandles.lookup()`:

```java
        catch (ReflectiveOperationException e) {
            throw new ExceptionInInitializerError(e);

File: core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java [hunk 22]
```diff
@@ -106,6 +110,7 @@ public static TypeEncoder createTypeEncoder(Type type, boolean supportsParametri
             case VarcharType _ -> VARCHAR_ENCODER;
             case VarbinaryType _ -> VARBINARY_ENCODER;
             case CharType charType -> new CharEncoder(charType.getLength());
+            case VariantType _ -> VARIANT_ENCODER;
```
electrum, line 114: See other comment. This should do JSON or raw, depending on client capability, like `supportsParametricDateTime`.

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -312,6 +317,28 @@ public void encode(JsonGenerator generator, Block block, int position)
...
+        public void encode(JsonGenerator generator, Block block, int position)
+                throws IOException
+        {
+            if (block.isNull(position)) {
+                generator.writeNull();
+                return;
+            }
+
+            Variant variant = VARIANT.getObject(block, position);
+            generator.writeStartArray();
```
electrum: Making this an object with `metadata` and `data` fields might be better and more future proof.

core/trino-main/src/main/java/io/trino/util/variant/VariantUtil.java
```diff
@@ -0,0 +1,1302 @@
...
+            .disable(CANONICALIZE_FIELD_NAMES)
+            // prevents characters outside BMP (e.g., emoji) from being escaped as surrogate pairs
+            .enable(COMBINE_UNICODE_SURROGATES_IN_UTF8)
+            .build());
+
+    private VariantUtil() {}
+
+    public static boolean canCastToVariant(Type type)
+    {
+        if (type instanceof UnknownType ||
```
electrum, line 133: Can use switch expression

core/trino-main/src/main/java/io/trino/util/variant/VariantUtil.java
```diff
@@ -0,0 +1,1302 @@
...
+        }
+        if (type instanceof RowType) {
+            return type.getTypeParameters().stream().allMatch(VariantUtil::canCastToVariant);
+        }
+        return false;
+    }
+
+    public static boolean canCastFromVariant(Type type)
+    {
+        if (type instanceof UnknownType ||
```
electrum, line 168: Can use switch expression

core/trino-main/src/main/java/io/trino/type/VariantOperators.java
```diff
@@ -0,0 +1,387 @@
...
+public final class VariantOperators
+{
+    private VariantOperators() {}
+
+    @ScalarOperator(SUBSCRIPT)
+    @SqlType(StandardTypes.VARIANT)
+    public static Variant dereference(@SqlType(StandardTypes.VARIANT) Variant value, @SqlType(StandardTypes.BIGINT) long index)
+    {
+        if (value.basicType() != Header.BasicType.ARRAY) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "VARIANT value is not an array");
```
electrum: Should we include the actual type here?

Same for object cast

core/trino-main/src/main/java/io/trino/util/variant/RowVariantWriter.java
```diff
@@ -0,0 +1,222 @@
...
+    private final List<Type> fieldTypes;
+    private final List<Optional<PrimitiveVariantEncoder>> fieldPrimitiveEncoders;
+    private final List<Optional<VariantWriter>> fieldWriters;
+    private final IntUnaryOperator outputIndexToRowIndex;
+
+    RowVariantWriter(RowType type)
+    {
+        List<Slice> fieldNames = type.getFields().stream()
+                .map(Field::getName)
+                .map(name -> name.orElseThrow(() -> new IllegalArgumentException("Anonymous fields are not supported in RowType for Variant encoding")))
```
electrum: Can users ever see this error? If so it should be `TrinoException`.

core/trino-main/src/main/java/io/trino/util/variant/RowVariantWriter.java
```diff
@@ -0,0 +1,222 @@
...
+            }
+
+            size = encodedObjectSize(maxFieldId, fieldPrimitiveEncoders.size(), totalElementLength);
+        }
+
+        @Override
+        public int size()
+        {
+            if (size < 0) {
+                throw new IllegalStateException("finalize() must be called before size()");
```
electrum: Can use `checkState()` for these

core/trino-main/src/main/java/io/trino/util/variant/VariantUtil.java
```diff
@@ -0,0 +1,1302 @@
...
+    }
+
+    // given a VARIANT parser, write to the BlockBuilder
+    public interface BlockBuilderAppender
+    {
+        void append(Variant variant, BlockBuilder blockBuilder);
+
+        static BlockBuilderAppender createBlockBuilderAppender(Type type)
+        {
+            if (type instanceof BooleanType) {
```
electrum, line 705: Can use switch expression

core/trino-main/src/test/java/io/trino/type/TestRowOperators.java
```diff
@@ -569,10 +569,6 @@ public void testRowCast()
                 .binding("a", "row(2, 3)"))
                 .isEqualTo(true);
 
-        assertThat(assertions.expression("cast(a AS row(aa bigint, bb double))[2]")
```
electrum, line 568: Why remove this test?

core/trino-main/src/test/java/io/trino/type/TestVariantFunctions.java
```diff
@@ -0,0 +1,67 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+import static org.junit.jupiter.api.parallel.ExecutionMode.CONCURRENT;
+
+@TestInstance(PER_CLASS)
+@Execution(CONCURRENT)
+class TestVariantFunctions
+{
+    // Some of the tests in this class are expected to fail when coercion between primitive Trino types changes behavior
```
electrum: What does this mean?

lib/trino-orc/src/main/java/io/trino/orc/reader/ColumnReaders.java
```diff
@@ -62,6 +65,14 @@ public static ColumnReader createColumnReader(
                     ICEBERG_BINARY_TYPE, column);
             return new UuidColumnReader(column);
         }
+        if (type instanceof VariantType) {
+            checkArgument(orcTypeKind == STRUCT, "Variant type can only be read from STRUCT column but got %s", column);
```
electrum: I think these should use `invalidStreamType()` so that they throw the correct exception type. The UUID one above might be wrong.

lib/trino-orc/src/main/java/io/trino/orc/writer/ColumnWriters.java
```diff
@@ -63,6 +67,16 @@ public static ColumnWriter createColumnWriter(
             checkArgument("TIME".equals(orcType.getAttributes().get(ICEBERG_LONG_TYPE)), "wrong attributes %s for type %s", orcType.getAttributes(), type);
             return new TimeColumnWriter(columnId, type, compression, bufferSize, () -> new TimeMicrosStatisticsBuilder(bloomFilterBuilder.get()));
         }
+        if (type instanceof VariantType) {
+            checkArgument(orcType.getOrcTypeKind() == STRUCT, "wrong ORC type %s for type %s", orcType, type);
+            checkArgument("true".equals(orcType.getAttributes().get(ICEBERG_VARIANT_TYPE_KIND)), "wrong attributes %s for type %s", orcType.getAttributes(), type);
+            checkArgument(orcType.getFieldCount() == 2, "Variant ORC struct must have 2 fields (metadata, value), but found %s", orcType.getFieldCount());
```
electrum, line 73: Also validate the structure field types?

lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
```diff
@@ -370,7 +371,7 @@ private static Optional<Field> constructField(Type type, ColumnIO columnIO, bool
 
     private static boolean isVariantType(Type type, ColumnIO columnIO)
     {
-        return type.getBaseName().equals(JSON) &&
+        return (type.getBaseName().equals(JSON) || type == VARIANT) &&
```
electrum: Could update check to `type == JSON || type == VARIANT`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/HiveSchemaUtil.java
```diff
@@ -48,8 +48,7 @@ private static String convertToTypeString(Type type)
             case FIXED, BINARY -> "binary";
             case DECIMAL -> "decimal(%s,%s)".formatted(((DecimalType) type).precision(), ((DecimalType) type).scale());
             case UNKNOWN, GEOMETRY, GEOGRAPHY -> throw new TrinoException(NOT_SUPPORTED, "Unsupported Iceberg type: " + type);
-            // TODO https://github.com/trinodb/trino/issues/24538 Support variant type
-            case VARIANT -> throw new TrinoException(NOT_SUPPORTED, "Unsupported Iceberg type: VARIANT");
+            case VARIANT -> "struct<metadata:binary,value:binary>";
```
electrum, line 49: Is this correct for all file formats?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5027,6 +5036,29 @@ public void testAllAvailableTypes()
         assertUpdate("DROP TABLE test_all_types");
     }
 
+    @Test
+    public void testNestedVariant()
+    {
+        // Tests variant nested inside array and row types
+        try (TestTable table = newTrinoTable(
+                "test_nested_variant",
+                "(" +
```
electrum, line 5240: See if you can use text block

core/trino-spi/src/main/java/io/trino/spi/variant/VariantUtils.java
```diff
@@ -0,0 +1,1106 @@
...
+                high = mid - 1;
+            }
+            else {
+                return mid;
+            }
+        }
+        return -1;
+    }
+
+    public static boolean equals(
```
electrum, line 190: The equality code is long and complex. Let's move it to a new class `VariantEquality`.

core/trino-spi/src/main/java/io/trino/spi/variant/VariantUtils.java
```diff
@@ -0,0 +1,1106 @@
...
+            int rightElementStart = rightValuesStart + readOffset(rightSlice, rightOffsetsStart + index * rightOffsetSize, rightOffsetSize);
+
+            if (!equals(leftMetadata, leftSlice, leftElementStart, rightMetadata, rightSlice, rightElementStart)) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    public static long hashCode(Metadata metadata, Slice slice, int offset)
```
electrum, line 401: How about moving this to a new class `VariantHashing`?

core/trino-spi/src/main/java/io/trino/spi/variant/VariantUtils.java
```diff
@@ -0,0 +1,1106 @@
...
+    }
+
+    static void verify(boolean test, String message)
+    {
+        if (!test) {
+            throw new VerifyException(message);
+        }
+    }
+
+    private static class VerifyException
```
electrum: Seems like most usages of `IllegalStateException` in this commit should use `VerifyException` instead

core/trino-spi/src/main/java/io/trino/spi/variant/VariantFieldRemapper.java
```diff
@@ -0,0 +1,360 @@
...
+                };
+
+                // write remapped elements
+                for (int index = 0; index < array.count(); index++) {
+                    int start = array.elementStart(index);
+                    int end = array.elementEnd(index);
+                    written += write(input, start, output, outputOffset + written, end - start);
+                }
+
+                verify(written == (remapMode == RemapMode.SAME_SIZE ? length : containerSizeCache.get(inputOffset)), "unexpected size mismatch in complete remap");
```
electrum: Extract a variable `expectedSize`

core/trino-spi/src/main/java/io/trino/spi/variant/VariantFieldRemapper.java
```diff
@@ -0,0 +1,360 @@
...
+                        outputOffset);
+
+                // write remapped elements
+                for (int index = 0; index < object.count(); index++) {
+                    int start = object.valueStart(index);
+                    int end = object.valueEnd(index);
+                    written += write(input, start, output, outputOffset + written, end - start);
+                }
+
+                verify(written == (remapMode == RemapMode.SAME_SIZE ? length : containerSizeCache.get(inputOffset)), "unexpected size mismatch in complete remap");
```
electrum: Extract a variable `expectedSize`

core/trino-spi/pom.xml
```diff
@@ -939,6 +951,16 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
+                <groupId>org.basepom.maven</groupId>
+                <artifactId>duplicate-finder-maven-plugin</artifactId>
+                <configuration>
+                    <ignoredResourcePatterns combine.children="append">
+                        <!-- ignore the Iceberg build metadata file -->
```
electrum, line 291: No need for this comment

core/trino-spi/pom.xml
```diff
@@ -939,6 +951,16 @@
                     </execution>
                 </executions>
             </plugin>
+            <plugin>
```
electrum, line 286: Nit: blank line between plugin declarations

electrum:
@coderabbitai check that this PR has sufficient test coverage, both for the variant code and for the Iceberg integration

electrum (approved):

client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
```diff
@@ -298,6 +302,33 @@ public Object decode(JsonParser parser)
...
+            byte[] metadata = Base64.getDecoder().decode(parser.getValueAsString());
+            requireNonNull(parser.nextToken());
+            byte[] value = Base64.getDecoder().decode(parser.getValueAsString());
+            if (parser.nextToken() != END_ARRAY) {
+                throw illegalToken(parser);
+            }
+            return List.of(metadata, value);
```
ebyhr: As a user, I expect the variant type to return a JSON-like result instead of a binary array. The metadata is an implementation detail, and I don’t see much benefit in exposing it.


```java
trino> SELECT CAST('{"abc": 123}' AS variant);
                       _col0
----------------------------------------------------
 [01 00 00, 31 7b 22 61 62 63 22 3a 20 31 32 33 7d]
```

core/trino-spi/src/main/java/io/trino/spi/type/VariantType.java
```diff
@@ -0,0 +1,296 @@
...
+import static java.lang.Math.abs;
+import static java.lang.Math.min;
+import static java.lang.Math.toIntExact;
+import static java.lang.invoke.MethodHandles.lookup;
+
+public class VariantType
+        extends AbstractType
+        implements VariableWidthType
+{
+    public static final String NAME = "VARIANT";
```
ebyhr: Note that most other types are written in lowercase.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5027,6 +5036,29 @@ public void testAllAvailableTypes()
...
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (NULL, NULL)", 1);
+
+            assertThat(query("SELECT * FROM " + table.getName()))
+                    .matches("VALUES (" +
+                            "ARRAY[CAST(1 AS VARIANT), CAST('hello' AS VARIANT), CAST(NULL AS VARIANT)], " +
+                            "CAST(ROW(42, 123) AS ROW(v variant, i integer))), " +
+                            "(NULL, NULL)");
+        }
+    }
+
```
ebyhr, line 5258: The test coverage in the Iceberg connector seems too limited. Please add more tests:  
* All column types as value fields across the three file formats
* Product test to ensure that Spark can read tables using our variant type

client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
findepi: I agree that we should return VARIANT as JSON text by default.
A client should be able to opt-in (client capabilities) into getting VARIANT binary data losslessly.
Trino CLI and JDBC can opt into this. JDBC can expose VARIANT via resultset.getobject.

core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
dain: Unfortunately, `getBuiltInFunctions` is not public

core/trino-main/src/main/java/io/trino/metadata/SignatureBinder.java
dain: Unfortunately, getBuiltInFunctions is not public

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
dain: This is the base test class that is used to test all iceberg file formats, so all formats should be covered.  

I also added a new test that has variants with every kind in it. I do not think this is necessary, as if you look at the code in Iceberg, it never actually inspects the variants. It's binary in and binary out. 

I added a spark product test covering all variant types.

docs/src/main/sphinx/connector/iceberg.md
dain: The default format version is still `2`

core/trino-main/src/main/java/io/trino/operator/scalar/timestamptz/VarcharToTimestampWithTimeZoneCast.java
dain: They are use in VariantUtils which is in a different package.

core/trino-main/src/main/java/io/trino/operator/scalar/timestamp/VarcharToTimestampCast.java
dain: Yes. This is used in VariantUtils.

core/trino-main/src/test/java/io/trino/type/TestRowOperators.java
dain: It is a duplicate assertion, but I moved this to a separate commit.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/HiveSchemaUtil.java
dain: Yes.  This is for the Hive Metastore Catalog in Iceberg.  This is the same value used by Apache Iceberg.

client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
dain: For now, we are converting the variant to JSON, so this code is no longer present.

client/trino-client/src/main/java/io/trino/client/JsonDecodingUtils.java
dain: Please see the comment above from David. This was always intended as temporary code, but for now we're just going to use JSON and add the pure VARIANT version back later, with opt-in controls.

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
dain: We use plain json for now.

core/trino-main/src/test/java/io/trino/type/TestVariantFunctions.java
dain: removed

dain:
I talked to @martint and he has OKed merging this.

soumilshah1995:
When are we planning to release 481?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/HiveSchemaUtil.java
ebyhr: Iceberg doesn't use this value. See https://github.com/apache/iceberg/issues/15220 for more details.

PR #27756, opened 2025-12-24
Correctly remove position delete files on optimize

### LINKED ISSUE #27732: POSITION_DELETES files in Iceberg are not optimized when using $partition filter
`POSITION_DELETES` files are not optimized after executing OPTIMIZE command with filter by `"$partition"` column.
Example:
```
CREATE TABLE test_tbl (id bigint, dt timestamp with time zone, payload varchar)
WITH (partitioning = ARRAY['month(dt)']);

INSERT INTO test_tbl VALUES(1, timestamp'2025-12-22 13:00', 'data1'), (2, timestamp'2025-11-10 12:00', 'ababa');
UPDATE test_tbl SET payload='data11' WHERE id = 1;

SELECT partition, content, count(*) files FROM "test_tbl$files" GROUP BY AUTO ORDER BY 1,2;
+--------------+-------+-----+
|partition     |content|files|
+--------------+-------+-----+
|{dt_month=670}|0      |1    |
|{dt_month=671}|0      |2    |
|{dt_month=671}|1      |1    |
+--------------+-------+-----+

ALTER TABLE test_tbl EXECUTE OPTIMIZE WHERE dt >= date'2025-12-01' and dt < date'2026-01-01';

SELECT partition, content, count(*) files FROM "test_tbl$files" GROUP BY AUTO ORDER BY 1,2;
+--------------+-------+-----+
|partition     |content|files|
+--------------+-------+-----+
|{dt_month=670}|0      |1    |
|{dt_month=671}|0      |1    |
+--------------+-------+-----+

UPDATE test_tbl SET payload='data1' WHERE id = 1;

SELECT partition, content, count(*) files FROM "test_tbl$files" GROUP BY AUTO ORDER BY 1,2;
+--------------+-------+-----+
|partition     |content|files|
+--------------+-------+-----+
|{dt_month=670}|0      |1    |
|{dt_month=671}|0      |2    |
|{dt_month=671}|1      |1    |
+--------------+-------+-----+

-- Optimizing with `"$partition"` filter
ALTER TABLE iceberg.temp.test_tbl EXECUTE OPTIMIZE WHERE "$partition"='dt_month=2025-12';

SELECT partition, content, count(*) files FROM iceberg.temp."test_tbl$files" GROUP BY AUTO ORDER BY 1,2;
+--------------+-------+-----+
|partition     |content|files|
+--------------+-------+-----+
|{dt_month=670}|0      |1    |
|{dt_month=671}|0      |1    |
|{dt_month=671}|1      |1    | <---- POSITION_DELETE file (content=1) exists 
+--------------+-------+-----+

```

JohnFitzpatrick44:
## Description

Fixes #27732 . Corrects an underlying bug that only affected predicates using the `$partition` column.

I updated an existing test to verify this behavior, as the test previously obscured this bug by using `EXECUTE REMOVE_ORPHAN_FILES`, which removes the dangling deletes that should've been removed by `OPTIMIZE`.

I believe the logic used to determine whether a position delete is fully applied is faulty, but **I could use a careful review here to verify my understanding.**

From the comment on line 348: `OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up position deletes`

This seems to contradict the logic on line 351: `case POSITION_DELETES -> partitionDomain.isAll() && pathDomain.isAll() && fileModifiedTimeDomain.isAll();`

As I understand this, `partitionDomain.isAll()` just checks whether there is any filter on the `$partition` column. As position delete files are only granular to the partition level, just checking the path and fileModifiedTime domains should be sufficient, as that would mean we're fully covering the partition. 

For predicates on the filter column (i.e. not using `$partition`), `partitionDomain.isAll()` is true, because we are not explicitly filtering on the `$partition` column. This is why the bug doesn't surface for these predicates.

## Release notes

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
@@ -348,7 +348,7 @@ public class IcebergSplitSource
                 // OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up position deletes
                 List<org.apache.iceberg.DeleteFile> fullyAppliedDeletes = wholeFileTask.deletes().stream()
                         .filter(deleteFile -> switch (deleteFile.content()) {
-                            case POSITION_DELETES -> partitionDomain.isAll() && pathDomain.isAll() && fileModifiedTimeDomain.isAll();
+                            case POSITION_DELETES -> pathDomain.isAll() && fileModifiedTimeDomain.isAll();
                             case EQUALITY_DELETES -> tableHandle.getEnforcedPredicate().isAll();
                             case DATA -> throw new IllegalStateException("Unexpected delete file: " + deleteFile);
                         })
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -5612,57 +5612,40 @@ public abstract class BaseIcebergConnectorTest
             throws IOException
     {
         String tableName = "test_optimize_" + randomNameSuffix();
-        Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
         assertUpdate("CREATE TABLE " + tableName + " WITH (partitioning = ARRAY['regionkey']) AS SELECT * FROM nation", 25);
 
-        List<String> allDataFilesInitially = getAllDataFilesFromTableDirectory(tableName);
-        assertThat(allDataFilesInitially).hasSize(5);
+        // There are 5 distinct region keys, so we expect 5 data files. A 'content' value of 0 corresponds to a data file, and a value of 1
+        // corresponds to a delete file
+        assertQuery(
+                "SELECT content, count(*) FROM \"" + tableName + "$files\" GROUP BY content",
+                "VALUES (0, 5)");
 
+        // Test that OPTIMIZE filtered on the partition column rewrites the corresponding data file and removes the delete file
         assertUpdate("DELETE FROM " + tableName + " WHERE nationkey = 7", 1);
+        assertQuery(
+                "SELECT content, count(*) FROM \"" + tableName + "$files\" GROUP BY content",
+                "VALUES (0, 5), (1, 1)");
+
+        Session session = withSingleWriterPerTask(getSession());
+        computeActual(session, "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE WHERE regionkey = 3");
+        assertQuery(
+                "SELECT content, count(*) FROM \"" + tableName + "$files\" GROUP BY content",
+                "VALUES (0, 5)");
+
+        // Test that OPTIMIZE using the $partition column rewrites the corresponding data file and removes the delete file
+        assertUpdate("DELETE FROM " + tableName + " WHERE nationkey = 8", 1);
+        assertQuery(
+                "SELECT content, count(*) FROM \"" + tableName + "$files\" GROUP BY content",
+                "VALUES (0, 5), (1, 1)");
+
+        computeActual(session, "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE WHERE \"$partition\" = 'regionkey=2'");
+        assertQuery(
+                "SELECT content, count(*) FROM \"" + tableName + "$files\" GROUP BY content",
+                "VALUES (0, 5)");
 
         assertQuery(
-                "SELECT summary['total-delete-files'] FROM \"" + tableName + "$snapshots\" WHERE snapshot_id = " + getCurrentSnapshotId(tableName),
-                "VALUES '1'");
-
-        List<String> allDataFilesAfterDelete = getAllDataFilesFromTableDirectory(tableName);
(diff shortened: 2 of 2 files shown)


guyco33:
Thanks @JohnFitzpatrick44 for this quick fix!
I can approve that it fixes the issue after doing some tests on some real datasets as well.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5612,57 +5612,43 @@ public void testOptimizeCleansUpDeleteFiles()
             throws IOException
     {
         String tableName = "test_optimize_" + randomNameSuffix();
-        Session sessionWithShortRetentionUnlocked = prepareCleanUpSession();
-        assertUpdate("CREATE TABLE " + tableName + " WITH (partitioning = ARRAY['regionkey']) AS SELECT * FROM nation", 25);
+        assertUpdate("CREATE TABLE " + tableName + " WITH (PARTITIONING = ARRAY['regionkey']) AS SELECT * FROM nation", 25);
```
ebyhr: No need to uppercase `partitioning`. Please revert this change.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -348,7 +348,7 @@ private synchronized Iterator<FileScanTaskWithDomain> prepareFileTasksIterator(L
                 // OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up position deletes
                 List<org.apache.iceberg.DeleteFile> fullyAppliedDeletes = wholeFileTask.deletes().stream()
                         .filter(deleteFile -> switch (deleteFile.content()) {
-                            case POSITION_DELETES -> partitionDomain.isAll() && pathDomain.isAll() && fileModifiedTimeDomain.isAll();
+                            case POSITION_DELETES -> pathDomain.isAll() && fileModifiedTimeDomain.isAll();
```
ebyhr, line 351: Please squash commits into one. We don't separate a commit for product code and test in this project.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5612,57 +5612,43 @@ public void testOptimizeCleansUpDeleteFiles()
...
-                "SELECT summary['total-delete-files'] FROM \"" + tableName + "$snapshots\" WHERE snapshot_id = " + getCurrentSnapshotId(tableName),
-                "VALUES '1'");
-
-        List<String> allDataFilesAfterDelete = getAllDataFilesFromTableDirectory(tableName);
-        assertThat(allDataFilesAfterDelete).hasSize(6);
-
-        // For optimize we need to set task_min_writer_count to 1, otherwise it will create more than one file.
-        computeActual(withSingleWriterPerTask(getSession()), "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE WHERE regionkey = 3");
-        computeActual(sessionWithShortRetentionUnlocked, "ALTER TABLE " + tableName + " EXECUTE EXPIRE_SNAPSHOTS (retention_threshold => '0s')");
-        computeActual(sessionWithShortRetentionUnlocked, "ALTER TABLE " + tableName + " EXECUTE REMOVE_ORPHAN_FILES (retention_threshold => '0s')");
```
ebyhr, line 5632: Could you explain why these commands were removed? Also, was there a reason for modifying the existing test instead of adding a new one?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5612,57 +5612,43 @@ public void testOptimizeCleansUpDeleteFiles()
...
+                "SELECT content, count(*) FROM \"" + tableName + "$files\" GROUP BY content",
+                "VALUES (0, 5), (1, 1)");
 
+        computeActual(withSingleWriterPerTask(
+                getSession()),
+                "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE WHERE regionkey = 3");
```
ebyhr: nit: We extract a variable for `withSingleWriterPerTask(getSession())` and unwrap the parameter here and L5642-5644.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
JohnFitzpatrick44: I felt that this test does not properly verify what it's supposed to, and that it should've caught the bug. It's named `testOptimizeCleansUpDeleteFiles`, but in practice, it does not verify that the `OPTIMIZE` process correctly removes dangling deletion files. `REMOVE_ORPHAN_FILES` does this instead, and obscures the correct behavior we're testing for. As `EXPIRE_SNAPSHOTS` and `REMOVE_ORPHAN_FILES` have their own tests, I opted to remove them.

Let me know if there's some behavior we should test for that I removed, and I can add it back or add a separate test.

chenjian2664:
Feel like it deserves a release note

PR #27757, opened 2025-12-24
Various cleanups in type inferencer

martint:
Various cleanups in the type inferencer an TypeSignature usages.

## Release notes


core/trino-main/src/main/java/io/trino/connector/system/RuleStatsSystemTable.java
```diff
@@ -64,7 +64,7 @@ public RuleStatsSystemTable(Optional<RuleStatsRecorder> ruleStatsRecorder, TypeM
                 .column("matches", BIGINT)
                 .column("failures", BIGINT)
                 .column("average_time", DOUBLE)
-                .column("time_distribution_percentiles", typeManager.getType(mapType(DOUBLE.getTypeSignature(), DOUBLE.getTypeSignature())))
+                .column("time_distribution_percentiles", new MapType(DOUBLE, DOUBLE, typeManager.getTypeOperators()))
```
findepi, line 67: The previous way of constructing the map abstracted away the map type dependencies (the key hash/equal/identical operators). This was a good thing.

The caller of `new MapType` needs to provide type operators. if they provide `new TypeOperators()`, things will still work, but likely not as well as they should.
Should the `MapType` constructor take `TypeManager` to nudge toward correct use of the API?

findepi: Also, the `Create Array, Map, and Row types directly` commit doesn't have any justification, so it's unclear what's the intention here. 

- Is `TypeManager.getType(TypeSignature)` supposed to be deprecated?
- add a line of explanation to the commit message

findepi (commented):
`Remove unused method \n\n It's only used in tests` LGTM

core/trino-spi/src/main/java/io/trino/spi/function/TypeVariableConstraint.java
```diff
@@ -214,10 +210,10 @@ public static TypeVariableConstraint fromJson(
             @JsonProperty("name") String name,
             @JsonProperty("comparableRequired") boolean comparableRequired,
             @JsonProperty("orderableRequired") boolean orderableRequired,
-            @JsonProperty("variadicBound") Optional<String> variadicBound,
+            @JsonProperty("variadicBound") boolean rowType,
```
findepi: update JsonProperty value

why didn't it break any tests?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPartitioningHandle.java
```diff
@@ -145,7 +145,7 @@ public long getCacheKeyHint()
         for (IcebergPartitionFunction function : partitionFunctions) {
             hasher.putInt(function.transform().ordinal());
             function.dataPath().forEach(hasher::putInt);
-            hasher.putString(function.type().getTypeSignature().toString(), UTF_8);
+            hasher.putString(function.type().getTypeId().toString(), UTF_8);
```
findepi: `Typeid.toString` lacks documentation guaranteeing that it represents the TypeId _value_ (since TypeId is opaque id, its toString could even be the default Object.toString)

`TypeId.hashCode` could be used here % the fact TypeId does not formally declare equality semantics.


why not `Type.getDisplayName` or `Type.hashCode`?

findepi (commented):
`Use TypeId instead of TypeSignature` (in IPH)

core/trino-main/src/main/java/io/trino/metadata/VariableBindings.java
```diff
@@ -0,0 +1,79 @@
...
+import io.trino.spi.type.Type;
+
+import java.util.Map;
+import java.util.Objects;
+
+import static com.google.common.base.Preconditions.checkState;
+import static java.lang.String.CASE_INSENSITIVE_ORDER;
+import static java.util.Objects.requireNonNull;
+
+public class VariableBindings
```
findepi: `final`, `@Immutable`

core/trino-main/src/main/java/io/trino/metadata/VariableBindings.java
```diff
@@ -0,0 +1,79 @@
...
+{
+    private final Map<String, Type> typeVariables;
+    private final Map<String, Long> longVariables;
+
+    public VariableBindings(Map<String, Type> typeVariables, Map<String, Long> longVariables)
+    {
+        this.typeVariables = ImmutableSortedMap.copyOf(typeVariables, CASE_INSENSITIVE_ORDER);
+        this.longVariables = ImmutableSortedMap.copyOf(longVariables, CASE_INSENSITIVE_ORDER);
```
findepi, line 37: Why does this want to be case-insensitive? To me it seems it shouldn't need to.
i know it's pre-existing from `BoundVariables`.
Add a comment.

core/trino-main/src/main/java/io/trino/metadata/VariableBindings.java
```diff
@@ -0,0 +1,79 @@
...
+    public boolean containsLongVariable(String variableName)
+    {
+        requireNonNull(variableName, "variableName is null");
+        return longVariables.containsKey(variableName);
+    }
+
+    @Override
+    public boolean equals(Object o)
+    {
+        if (!(o instanceof VariableBindings that)) {
```
findepi, line 71: we generally compare `this.class` vs `o.class`  rather than `instacenof` in `equals` impls.
This pattern is shorter and equivalent only when the class is final. Keep, but make class `final`.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinding.java
```diff
@@ -55,38 +48,9 @@ public int getArity()
...
-        T value = map.get(variableName);
-        checkState(value != null, "value for variable '%s' is null", variableName);
-        return value;
-    }
-
-    private static boolean containsValue(Map<String, ?> map, String variableName)
-    {
-        checkState(variableName != null, "variableName is null");
-        return map.containsKey(variableName);
+        return boundVariables;
```
findepi: Should the field and ctor param be also `variables`, without "bound" prefix?

core/trino-spi/src/main/java/io/trino/spi/function/Signature.java
```diff
@@ -71,6 +71,11 @@ public boolean isVariableArity()
         return variableArity;
     }
 
+    public boolean isGeneric()
```
findepi, line 77: Document. It should be clear that `array(x)` is generic, but `varchar(x)` is not.

core/trino-main/src/main/java/io/trino/connector/system/RuleStatsSystemTable.java
martint: I know. This is one of those unfortunate inconsistencies we have across the codebase. Except for MapType, every other type can be constructed directly, and in most cases (but not all) are. Even MapType is constructed directly in a large number of callsites, with TypeOperators passed explicitly.

This is an attempt at cleaning up usages and bringing some uniformity. This will evolve as I continue to refactor how types, constraints and type inference are handled.

> Is TypeManager.getType(TypeSignature) supposed to be deprecated?

Not yet, but it will take a different shape once I revamp TypeSignature.

core/trino-spi/src/main/java/io/trino/spi/function/TypeVariableConstraint.java
martint: Probably because this class never gets serialized in practice.

core/trino-main/src/main/java/io/trino/connector/system/RuleStatsSystemTable.java
findepi: > This is one of those unfortunate inconsistencies we have across the codebase. Except for MapType, every other type can be constructed directly

perhaps because almost no types have dependencies

what about this option:

> Should the MapType constructor take TypeManager to nudge toward correct use of the API?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPartitioningHandle.java
martint: Replaced with TypeId.getId()

> TypeId.hashCode could be used here % the fact TypeId does not formally declare equality semantic

Not sure what you mean. It implements equals/hashcode.

> why not Type.getDisplayName or Type.hashCode?

Type.hashCode may not be a good enough hash for that purpose. No guarantees.

Type.getDisplayName() is not canonical. It could potentially produce different outputs for equivalent types. For example, `decimal(3,0)` is the same as `decimal(3)`. Display name could in theory return either, depending on how it was constructed.

core/trino-main/src/main/java/io/trino/metadata/VariableBindings.java
martint: Yeah, pattern matching makes it a little more readable. It's what IntelliJ generates nowadays, FWIW

core/trino-main/src/main/java/io/trino/metadata/VariableBindings.java
martint: I do not know why we need that case-insensitive comparison, and haven't had a chance to verify it's actually required. I plan to fix this as part of the larger revamp of types, typesignature and type inference.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPartitioningHandle.java
findepi: > > TypeId.hashCode could be used here % the fact TypeId does not formally declare equality semantic
> 
> Not sure what you mean. It implements equals/hashcode.

i know it does.
it would be a nice practice to document equality semantics, as they are part of class contract.

anyway, no change requested here


> > why not Type.getDisplayName or Type.hashCode?
> 
> Type.hashCode may not be a good enough hash for that purpose. No guarantees.

i don't think it matters here. in fact it seems okay to skip the type in the hash here
the datapath is a more important field to hash 

> Type.getDisplayName() is not canonical. It could potentially produce different outputs for equivalent types. For example, `decimal(3,0)` is the same as `decimal(3)`. Display name could in theory return either, depending on how it was constructed.

it's not canonical, but it still should be a function of the type object's semantics
i.e. i expect same type (equal type objects) => same display name (equal) 
any other implementation i would detest as a trap

findepi: > Replaced with TypeId.getId()

👍

core/trino-main/src/main/java/io/trino/metadata/VariableBindings.java
findepi: makes sense to me. let's have a comment that this behavior is here, but not very intentionally (just say it in proper english ;) )

core/trino-main/src/main/java/io/trino/connector/system/RuleStatsSystemTable.java
martint: That's a false dependency. MapType doesn't need anything from TypeManager aside from TypeOperators, so I don't think that's necessarily a more correct use of the API.

findepi: Sure, if it needed something besides TypeManager, then it would already take the TypeManager.
However, reality is, `TypeManager.getTypeOperators()` is the only truly correct way to get type operators.
`new TypeOperators()` is a potential mistake someone could make, but not so much for `TypeManager`

PR #27755, opened 2025-12-24
Remove unnecessary DROP TABLE statement in TestIgniteTypeMapping

findinpath:
## Description

After the changes from https://github.com/trinodb/trino/pull/27702 there have been spotted signs of flakiness in the test

`io.trino.plugin.ignite.TestIgniteTypeMapping#testUnsupportedDateRange()`

```
io.trino.testing.QueryFailedException: Table doesn't exist: TEST_UNSUPPORTED_DATE_RANGE_CTAS
	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:138)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:621)
	at io.trino.testing.DistributedQueryRunner.executeWithPlan(DistributedQueryRunner.java:603)
	at io.trino.testing.QueryAssertions.assertDistributedUpdate(QueryAssertions.java:107)
	at io.trino.testing.QueryAssertions.assertUpdate(QueryAssertions.java:61)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:419)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:414)
	at io.trino.plugin.ignite.TestIgniteTypeMapping.testUnsupportedDateRange(TestIgniteTypeMapping.java:338)
	at io.trino.plugin.ignite.TestIgniteTypeMapping.testUnsupportedDateRange(TestIgniteTypeMapping.java:327)
	Suppressed: java.lang.Exception: SQL: DROP TABLE IF EXISTS test_unsupported_date_range_ctas
		at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:628)
		... 7 more
Caused by: io.trino.spi.TrinoException: Table doesn't exist: TEST_UNSUPPORTED_DATE_RANGE_CTAS
	at io.trino.plugin.jdbc.BaseJdbcClient.execute(BaseJdbcClient.java:1626)
	at io.trino.plugin.jdbc.BaseJdbcClient.dropTable(BaseJdbcClient.java:1459)
	at io.trino.plugin.jdbc.BaseJdbcClient.dropTable(BaseJdbcClient.java:1453)
	at io.trino.plugin.jdbc.ForwardingJdbcClient.dropTable(ForwardingJdbcClient.java:311)
[...]

PR #27760, opened 2025-12-26
Handle concurrent `DROP TABLE IF EXISTS` correctly

chenjian2664:
Adds logic to treat `TableNotFoundException` as a successful outcome when executing `DROP TABLE IF EXISTS`, addressing race conditions where
 the table may be dropped concurrently.

## Description

The current implementation first checks whether the table exists and then issues the `DROP TABLE` statement. However, under concurrent execution, the table may be dropped by another session between these two steps. In such cases, some JDBC backends raise a semantic "table does not exist" exception during the drop operation.
This PR updates the behavior to treat these exceptions as a successful outcome when `IF EXISTS` is specified, aligning the implementation with the intended semantics of `DROP TABLE IF EXISTS`.

A follow-up improvement would be to push `DROP TABLE IF EXISTS` to connectors that natively support this semantic.

* https://github.com/trinodb/trino/pull/27702
* https://github.com/trinodb/trino/pull/27755

## Release notes

```markdown
## General
* Fix failure when concurrent executing `DROP TABLE IF EXISTS` statement.
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
+++ b/core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
@@ -94,7 +94,7 @@ public class DropTableTask
                 throw semanticException(TABLE_NOT_FOUND, statement, "Table '%s' does not exist", tableName);
             }
             // This means we verified that the table existed, but by the time we executed DROP TABLE
-            // it may have already been dropped by another concurrent thread or task.
+            // it may have already been dropped by another concurrent task or due to concurrent processes.
             // Since DROP TABLE IF EXISTS is used, this case should be treated as successful.
         }
 
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
@@ -1465,21 +1465,15 @@ public abstract class BaseJdbcClient
             execute(session, sql);
         }
         catch (Throwable e) {
-            if (!(e instanceof TrinoException exception)) {
-                throw e;
-            }
-            if (isTableNotFoundException(exception)) {
+            if (e instanceof TrinoException exception && isTableNotFoundException(exception)) {
                 throw new TableNotFoundException(schemaTableName);
             }
-            throw exception;
+            throw e;
         }
     }
 
     protected boolean isTableNotFoundException(TrinoException exception)
     {
-        if (exception.getCause() instanceof SQLException sqlException) {
-            return "42S02".equals(sqlException.getSQLState());
-        }
         return false;
     }
 
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
@@ -9683,4 +9683,12 @@ public abstract class BaseHiveConnectorTest
                 .setCatalogSessionProperty(getSession().getCatalog().orElseThrow(), "orc_tiny_stripe_threshold", "0B")
                 .build();
     }
+
+    @Test
+    @Override
+    public void testDropTableIfExistsConcurrently()
+    {
+        assertThatThrownBy(super::testDropTableIfExistsConcurrently)
+                .hasMessageMatching("The following metastore delete operations failed: drop table .*");
+    }
 }
--- a/plugin/trino-mariadb/src/main/java/io/trino/plugin/mariadb/MariaDbClient.java
+++ b/plugin/trino-mariadb/src/main/java/io/trino/plugin/mariadb/MariaDbClient.java
@@ -752,6 +752,15 @@ public class MariaDbClient
         }
     }
 
+    @Override
+    protected boolean isTableNotFoundException(TrinoException exception)
+    {
+        if (exception.getCause() instanceof SQLException sqlException) {
+            return "42S02".equals(sqlException.getSQLState());
+        }
+        return false;
+    }
+
     private TableStatistics readTableStatistics(ConnectorSession session, JdbcTableHandle table)
             throws SQLException
     {
--- a/plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlClient.java
+++ b/plugin/trino-mysql/src/main/java/io/trino/plugin/mysql/MySqlClient.java
@@ -1321,6 +1321,15 @@ public class MySqlClient
         }
     }
 
+    @Override
+    protected boolean isTableNotFoundException(TrinoException exception)
+    {
+        if (exception.getCause() instanceof SQLException sqlException) {
+            return "42S02".equals(sqlException.getSQLState());
+        }
+        return false;
+    }
+
     private static Optional<ColumnHistogram> getColumnHistogram(Map<String, String> columnHistograms, String columnName)
     {
         return Optional.ofNullable(columnHistograms.get(columnName))


testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -5802,6 +5803,40 @@ public void testDropTableIfExists()
         assertThat(getQueryRunner().tableExists(getSession(), "test_drop_if_exists")).isFalse();
     }
 
+    @Test
+    public void testDropTableIfExistsConcurrently()
```
ebyhr, line 5991: We should test the change within engine in my opinion. Such concurrently tests with connectors often become flaky especially when the test runs with SaaS. 

I think you can use extended MockMetadata that throws TableNotFoundException for a specific table name in TestDropTableTask. You can refer to TestSetDefaultValueTask.

chenjian2664: What kinds of tests should we add when introducing support for each connector?

findepi: > We should test the change within engine in my opinion. Such concurrently tests with connectors often become flaky especially when the test runs with SaaS.

agreed. it's my fault that i added some of these. we could perhaps revisit and maybe remove those problematic ones.

chenjian2664: If we remove the connector tests, what is the appropriate place to validate this functionality for each connector once it is added? 
Seems now it is still the good one to work at the base connector test, though it is (higher possibility) flaky right?

core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
```diff
@@ -85,7 +86,17 @@ public ListenableFuture<Void> execute(
...
-        metadata.dropTable(session, redirectionAwareTableHandle.tableHandle().get(), tableName.asCatalogSchemaTableName());
+        try {
+            metadata.dropTable(session, redirectionAwareTableHandle.tableHandle().get(), tableName.asCatalogSchemaTableName());
+        }
+        catch (TableNotFoundException e) {
+            if (!statement.isExists()) {
+                throw semanticException(TABLE_NOT_FOUND, statement, "Table '%s' does not exist", tableName);
+            }
+            // This means we verified that the table existed, but by the time we executed DROP TABLE
+            // it may have already been dropped by another concurrent thread or task.
```
findinpath: Can happen also when the dropping is done from a concurrent process.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -1450,13 +1450,37 @@ public void dropNotNullConstraint(ConnectorSession session, JdbcTableHandle hand
...
+                throw new TableNotFoundException(schemaTableName);
+            }
+            throw exception;
+        }
+    }
+
+    protected boolean isTableNotFoundException(TrinoException exception)
+    {
+        if (exception.getCause() instanceof SQLException sqlException) {
+            return "42S02".equals(sqlException.getSQLState());
```
findinpath: magic number

findinpath: do we have any certainty that all the sql data sources throw this sql state for signaling a table not found exception?

chenjian2664: Yes, it is for the mysql, mariadb ... just part of the connectors throws it, different db has different sql state for the table not found exception

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
findepi: There is no better place than connector tests. It's just that, as practice showed, concurrency tests are timing and load sensitive and this leads to flakiness for _some_ of the connectors. 
Maybe we keep the tests but are more aggressive about disabling them for problematic connectors.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
findepi: > Yes, it is for the mysql, mariadb, Ignite ..

is 42S02 mandated by JDBC specification to mean "table/view not found"?
If not, this must not be in BaseJdbcClient.

in https://download.oracle.com/otn-pub/jcp/jdbc-4_3-mrel3-eval-spec/jdbc4.3-fr-spec.pdf i see this

> a SQLState. The String containing the SQLState can be retrieved by calling the
method SQLException.getSQLState.
The value of the SQLState string will depend on the underlying data source
setting the value. Both X/Open and SQL:2003 define SQLState values and the
conditions in which they should be set. Although the sets of values overlap,
the values defined by SQL:2003 are not a superset of X/Open.
The DatabaseMetaData method getSQLStateType allows an application to
determine if the SQLStates being returned by a data source are X/Open or
SQL:2003.

so it seems that, under some additional conditions,  this is at least _supposed_ to be portable way for checking not found.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
chenjian2664: Agree about the disabling we needs to look at that case by case.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
chenjian2664: Removed from base jdbc

core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
chenjian2664: added in comment

core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
```diff
@@ -85,7 +86,20 @@ public ListenableFuture<Void> execute(
...
+        catch (TableNotFoundException e) {
+            if (!statement.isExists()) {
+                throw semanticException(TABLE_NOT_FOUND, statement, "Table '%s' does not exist", tableName);
+            }
+            // This means we verified that the table existed, but by the time we executed DROP TABLE
+            // it may have already been dropped by another concurrent task or due to concurrent processes.
+            // Since DROP TABLE IF EXISTS is used, this case should be treated as successful.
+        }
+        catch (Throwable e) {
+            System.out.println("not ok drop table " + e.getMessage());
```
findepi: remove

core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
```diff
@@ -85,7 +86,20 @@ public ListenableFuture<Void> execute(
         QualifiedObjectName tableName = redirectionAwareTableHandle.redirectedTableName().orElse(originalTableName);
         accessControl.checkCanDropTable(session.toSecurityContext(), tableName);
 
-        metadata.dropTable(session, redirectionAwareTableHandle.tableHandle().get(), tableName.asCatalogSchemaTableName());
+        try {
+            metadata.dropTable(session, redirectionAwareTableHandle.tableHandle().get(), tableName.asCatalogSchemaTableName());
+        }
+        catch (TableNotFoundException e) {
+            if (!statement.isExists()) {
+                throw semanticException(TABLE_NOT_FOUND, statement, "Table '%s' does not exist", tableName);
```
findepi, line 94: why not `throw e`?

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -339,6 +340,7 @@ Optional<TableExecuteHandle> getTableHandleForExecute(
      * Drops the specified table
      *
      * @throws RuntimeException if the table cannot be dropped or table handle is no longer valid
+     * @throws TableNotFoundException if the table not exists
```
findepi, line 343: This should be added to ConnectorMetadata.dropTable as well.

However, we should also consider adding `boolean ifExists` to `ConnectorMetadata.dropTable` instead of this change. The implementation in JDBC connectors would be trivial.
Was this approach considered and rejected? Do we have reasoning noted down somewhere?


cc @martint

findepi: if we add `boolean ifExists` to `ConnectorMetadata.dropTable`, then perhaps there should also be a return value indicating "did we drop" to inform whether we should call `systemSecurityMetadata.tableDropped`.

core/trino-main/src/main/java/io/trino/execution/DropTableTask.java
chenjian2664: I wanna keep the exception consistent with above - https://github.com/trinodb/trino/blob/bcebf07a5bed0cd50cbe851ec1ed1359c6947c41/core/trino-main/src/main/java/io/trino/execution/DropTableTask.java#L80-L81

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
chenjian2664: Yes, it's considered, I am planning it as the follow ups - let connectors to handle "IF EXISTS" for those connectors support it - usually jdbc connectors

findepi: I am not sure I see need for both approaches being implemented. We either make it engine responsibility (and connector responsible for throwing appropriate exception), or we make it fully connector responsibility.

If we make it engine responsibility (this PR), then I'd expect `MetadataManager` do this (not the task).

Issue #27764, opened 2025-12-27
Allow to change the readonly status of the connection before it is established


prrvchr:
I am working on a connector for SQLite and it does not allow changing the readOnly (ie: `connection.setReadOnly(Boolean readOnly)`) status of the connection once it is established. This must be done before establishing the connection with connection's properties.

In `BaseJdbcClient.java` this connection is delivered by the [getConnection(Session session)](https://github.com/trinodb/trino/blob/922659a4c04a2547715cc4edc16460795a7b78c3/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L655) method which after establishing it switches to read-only mode. This isn't really a problem since a connector like SQLite can handle this by overriding this method.

The problem lies in the `ExecuteProcedure.java` class which in method [doExecute(ConnectorSession session, String query)](https://github.com/trinodb/trino/blob/922659a4c04a2547715cc4edc16460795a7b78c3/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/procedure/ExecuteProcedure.java#L76) get a connection and switches to write mode after obtaining it.

Couldn't we introduce a new method to obtain the connection with a parameter giving the desired ReadOnly status and use this method instead if we want a connection with the possibility of writing?

prrvchr:
An implementation allowing properties to be passed during connection in order to obtain it in read-only mode has been submitted in [PR#27768](https://github.com/trinodb/trino/pull/27768)

findepi:
This issue on its own is not actionable. It's only useful if we want to build a connector using a JDBC driver than needs that.
Thus, addressing those limitations may be part of some other work stream (such as https://github.com/trinodb/trino/pull/27768 or https://github.com/trinodb/trino/issues/10060), and won't happen otherwise. Keeping the issue open won't increase chances of this being addressed. Closing for clarity.

Issue #10060, opened 2021-11-24
Support a SQLite connector


erikerlandson:
There is some philosophical resistance to just supporting a generalized JDBC connector (#3105), so here is a tracking issue for supporting specific SQLite connector

lhofhansl:
Curious how you envision this would work.

Would each worker load its own, independent SQLite database? Or would only one worker load the database?
Or do you have a shared (posix compliant) media, and concurrent access is handled by locking, and the each worker opens the same database file?

erikerlandson:
@lhofhansl  I would probably be satisfied with whatever makes it easiest to implement.  If that means only a single worker can access the SQLite DB and that reduces parallelism, I would consider that a fair tradeoff.

Our particular use case would allow for read-only access to the SQLite DB, which usually makes concurrency easier but I don't know if a read-only connector is OK with trino.

dforsber:
This may be related: https://github.com/dacort/athena-sqlite

MichaelTiemannOSC:
> This may be related: https://github.com/dacort/athena-sqlite

Nice bit of archeology, but that repo has not been active for 4 years.  It requires Python 3.7 which many packages no longer support at all (and many more deprecate).

henri9813:
Hello,

Any news on 2026 ? ^^'

PR #27768, opened 2025-12-28
New Sqlite connector

prrvchr:
## Description

Add new SQLite connector

This new connector allows Trino to open an SQLite database file. This file must be accessible to the connector.

Since SQLite does not support changing the read-only status of the connection after establishment (see: issue #27764), the implementation of this new connector required modifying the connection factories to allow dynamically passing properties during connection establishment.
Perhaps this would have been done in a separate PR?

Currently, this connector is read-only and only supports a minimal set of SQL types.
Ultimately, it should allow very easy access to scattered microdata. I hope scientists and others will find it very useful.

Let's not forget that SQLite is the leading database in terms of usage.

## Release notes

```markdown
## Section
* Fix some things. ({issue}`27764`)
```


findepi:
Sqlite is a great embeddable database that usually runs off a local file.
Trino is a distributed system. Nodes usually run on separate machines.
It seems they do not match very well. Is the assumption this connector is supposed to work in single node setup only?
I tried to infer this from the PR description (well written!), but i didn't find that detail there.

chenjian2664:
Maybe this is a specific federation use case :)

prrvchr:
> Nodes usually run on separate machines.

The SQLite database file must be accessible from the connector (ie: worker) via an absolute path on the hard drive, preferably local. But (I haven't tested this) nothing should prevent the worker from being remote?
Concurrent write problems (ie: database lock) should occur if multiple write accesses are made to the connector, but this is inherent to SQLite.

chenjian2664:
Please add documentation for the new connector, covering supported behaviors and known limitations, you can use existing connector documentation (for example, ignite.md) as a reference

findepi:
> Concurrent write problems (ie: database lock) should occur if multiple write accesses are made to the connector, but this is inherent to SQLite.

i see no reason to make this connector read-write.
make it read-only and avoid problems like https://github.com/trinodb/trino/issues/27764 or with Sqlite file being actually copies on each of machines.

findepi (changes requested):
> The SQLite database file must be accessible from the connector (ie: worker) via an absolute path on the hard drive, preferably local

i am not sold on the use-case yet. Yes, I know it's technically feasible to do (under some constraints that you and I probably both understand). I just am not convinced it will be applicable outside of a very niche use-cases. 

An issue voicing addition of the sqlite connector does exist

- https://github.com/trinodb/trino/issues/10060

however the issue wasn't outstandingly active in last 4 years.

I think it's a great idea for an sqlite connector to exists somewhere. I am just not convinced it should be part of trino official repository and releases, which would imply maintainers spend their maintainership budget on it, without clear understanding why they do so.

cc @trinodb/maintainers

prrvchr:
> i see no reason to make this connector read-write.
> make it read-only and avoid problems like #27764 or with Sqlite file being actually copies on each of machines.

This connector is currently read-only, as you advised for the first connector.
All databases lock data during a write operation, but usually only the affected tables are locked. SQLite, however, locks all tables during a write operation, that's the difference. I don't think this will cause any more problems than with other databases. This can be verified when implementing the write functionality for the connector.

> I just am not convinced it will be applicable outside of a very niche use-cases.

Imagine we have many sensors distributed throughout the environment that connect to the data center at fixed times. When they connect, it would be possible to view all the SQLite databases embedded in the sensors as a single database. I don't believe this is currently possible, and yet Trino will make it incredibly simple.
In fact, if we look at all smart devices, there's a high probability that they will have an embedded SQLite database.

Perhaps one day this connector will be the reason for more than half of Trino's downloads? I myself was astonished at how my [SQLiteOOo](https://prrvchr.github.io/) extension is surpassing all others in terms of downloads. You can check it out.

> I am just not convinced it should be part of trino official repository and releases, which would imply maintainers spend their maintainership budget on it, without clear understanding why they do so.

Well if we do it together, then I commit to maintaining this connector. Otherwise, as I already explained to David Phillips, I'll still do it alone, and certainly not as well as we could together, but I'll do it.

prrvchr:
> Please add documentation for the new connector, covering supported behaviors and known limitations

@chenjian2664 Ok done.

> make it read-only and avoid problems like #27764

@findepi Even though the driver only works in read-only mode, it does not support the use of the `connection.setReadOnly(boolean readOnly)` method and generates an SQL error if used. This method is called by Trino code, as I mentioned in #27764.

Giving connectors the ability to perform necessary pre-connection actions to ensure they deliver the connection with the correct read-only status seems like a good idea to me. If the connector needs to open files to establish the connection, then knowing the file status (read only or read write) can be important when opening those files. In fact, I think that's pretty much what's happening with SQLite.

And wanting to implement only a read-only connector simply because SQLite doesn't support concurrent writes doesn't seem justified to me. SQLite offers tools to work around such limitations. The [Database Remote-Copy Tool for SQLite](https://www.sqlite.org/rsync.html) is a good example.
It is up to those using this connector to do what is necessary with SQLite to prevent this from happening.
This is a limitation of SQLite that all these users already have to deal with.

PR #27778, opened 2025-12-29
Fix trino-parquet  lib can't read a list-inner-list column

### LINKED ISSUE #27766: Fix trino-parquet lib - can't read a nested list-inner-list column  array(array(varchar))
Use flowing parquet schema to generate  a parquet file with some test data,   Call trino-parquet api  to read the parquet file.  

```
message schema {
  optional group nested_list (LIST) {
    repeated group array {
      optional group inner_list (LIST) {
        repeated group array {
          optional binary element (STRING);
        }
      }
    }
  }
}
```
System report the error: 
```
Caused by: java.lang.ClassCastException: class org.apache.parquet.io.GroupColumnIO cannot be cast to class org.apache.parquet.io.PrimitiveColumnIO (org.apache.parquet.io.GroupColumnIO and org.apache.parquet.io.PrimitiveColumnIO are in unnamed module of loader 'app')
	at io.trino.parquet.ParquetTypeUtils.constructField(ParquetTypeUtils.java:366)
	at io.trino.parquet.ParquetTypeUtils.constructField(ParquetTypeUtils.java:363)
	at io.trino.parquet.ParquetTypeUtils.constructField(ParquetTypeUtils.java:286)
```

menghot:
* Fix trino-parquet  lib can't read a list-inner-list column array(array(varchar)) fixes #27766

## Release notes

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
@@ -72,7 +72,7 @@ public final class ParquetTypeUtils
     /* For backward-compatibility, the type of elements in LIST-annotated structures should always be determined by the following rules:
      * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists
      */
-    public static ColumnIO getArrayElementColumn(ColumnIO repeatedColumnIO)
+    public static ColumnIO getArrayElementColumn(Type type, ColumnIO repeatedColumnIO)
     {
         // Rule: The field annotated with LIST must have a single REPEATED child.
         if (!repeatedColumnIO.getType().isRepetition(REPEATED)) {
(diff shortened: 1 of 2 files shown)


lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
```diff
@@ -70,41 +70,59 @@ public static GroupColumnIO getMapKeyValueColumn(GroupColumnIO groupColumnIO)
...
 
     /* For backward-compatibility, the type of elements in LIST-annotated structures should always be determined by the following rules:
-     * 1. If the repeated field is not a group, then its type is the element type and elements are required.
-     * 2. If the repeated field is a group with multiple fields, then its type is the element type and elements are required.
-     * 3. If the repeated field is a group with one field and is named either array or uses the LIST-annotated group's name with _tuple appended then the repeated type is the element type and elements are required.
-     * 4. Otherwise, the repeated field's type is the element type with the repeated field's repetition.
      * https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#lists
      */
-    public static ColumnIO getArrayElementColumn(ColumnIO columnIO)
+    public static ColumnIO getArrayElementColumn(Type type, ColumnIO repeatedColumnIO)
```
Praveen2112, line 75: Can we rename the `Type type` as `elementType` ?

lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
```diff
@@ -70,41 +70,59 @@ public static GroupColumnIO getMapKeyValueColumn(GroupColumnIO groupColumnIO)
...
     {
-        while (columnIO instanceof GroupColumnIO && !columnIO.getType().isRepetition(REPEATED)) {
```
Praveen2112, line 81: Are we missing recursive aspect here ? Previously it would work for nested type with more levels, how would it work in this case ?

lib/trino-parquet/src/main/java/io/trino/parquet/ParquetTypeUtils.java
```diff
@@ -70,41 +70,59 @@ public static GroupColumnIO getMapKeyValueColumn(GroupColumnIO groupColumnIO)
     }
 
     /* For backward-compatibility, the type of elements in LIST-annotated structures should always be determined by the following rules:
-     * 1. If the repeated field is not a group, then its type is the element type and elements are required.
```
chenjian2664, line 73: If the underlying logic has changed, could we update the documentation to reflect the new behavior instead of deleting it?

PR #27786, opened 2025-12-30
Enable reading Iceberg v3 failing on all unimplemented features

dain:
## Description

Add support for creating Iceberg format version 3 tables, upgrading v2 tables to v3, and inserting into v3 tables.

This change intentionally does not implement Iceberg v3 features beyond allowing v3 metadata and validating that inserts produce the required row-lineage metadata (as observed through the Iceberg library).

To avoid spec violations while v3 support is incomplete, the connector now explicitly rejects v3 features that are not yet supported. The goal is to safely unlock v3 table creation and incremental adoption, while making unsupported behavior fail fast and predictably.

Unsupported v3 features that now throw `NOT_SUPPORTED` include:

* Row-level mutations on v3 tables: `DELETE`, `UPDATE`, `MERGE`
* `OPTIMIZE` on v3 tables
* `add_files` / `add_files_from_table` procedures on v3 tables
* Deletion vectors (PUFFIN delete files)
* Column default values (`initial-default`, `write-default`)
* Iceberg table encryption (`encryption-keys` / snapshot `key-id`)

Tests:

* Add `TestIcebergV3` to cover:

  * create v3 tables and upgrade v2→v3
  * inserts into v3 tables produce required lineage metadata (`nextRowId`, `firstRowId`, `dataSequenceNumber`)
  * unsupported v3 features fail with clear exceptions

## Release notes

```markdown
## Iceberg
* Allow creating Iceberg format version 3 tables, upgrading v2 tables to v3, and inserting into v3 tables. Unsupported v3 features are explicitly rejected.
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
@@ -53,7 +53,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
 public class IcebergConfig
 {
     public static final int FORMAT_VERSION_SUPPORT_MIN = 1;
-    public static final int FORMAT_VERSION_SUPPORT_MAX = 2;
+    public static final int FORMAT_VERSION_SUPPORT_MAX = 3;
     public static final String EXTENDED_STATISTICS_CONFIG = "iceberg.extended-statistics.enabled";
     public static final String EXTENDED_STATISTICS_DESCRIPTION = "Enable collection (ANALYZE) and use of extended statistics.";
     public static final String COLLECT_EXTENDED_STATISTICS_ON_WRITE_DESCRIPTION = "Collect extended statistics during writes";
@@ -77,7 +77,7 @@ public class IcebergConfig
     private boolean registerTableProcedureEnabled;
     private boolean addFilesProcedureEnabled;
     private Optional<String> hiveCatalogName = Optional.empty();
-    private int formatVersion = FORMAT_VERSION_SUPPORT_MAX;
+    private int formatVersion = 2;
     private Duration expireSnapshotsMinRetention = new Duration(7, DAYS);
     private Duration removeOrphanFilesMinRetention = new Duration(7, DAYS);
     private DataSize targetMaxFileSize = DataSize.of(1, GIGABYTE);
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -177,6 +177,7 @@ import org.apache.iceberg.SortField;
 import org.apache.iceberg.SortOrder;
 import org.apache.iceberg.StatisticsFile;
 import org.apache.iceberg.Table;
+import org.apache.iceberg.TableMetadata;
 import org.apache.iceberg.TableProperties;
 import org.apache.iceberg.TableScan;
 import org.apache.iceberg.Transaction;
@@ -633,6 +634,40 @@ public class IcebergMetadata
         return tableHandleForCurrentSnapshot(session, tableName, table);
     }
 
+    // TODO: Remove when Iceberg v3 is fully supported
+    private static void validateTableForTrino(BaseTable table, Optional<Long> tableSnapshotId)
+    {
+        Snapshot snapshot = tableSnapshotId
+                .map(table::snapshot)
+                .orElse(table.currentSnapshot());
+        if (snapshot == null) {
+            // empty table, nothing to validate
+            return;
+        }
+
+        TableMetadata metadata = table.operations().current();
+        if (metadata.formatVersion() < 3) {
+            return;
+        }
+
+        Schema schema = metadata.schemasById().get(snapshot.schemaId());
+        if (schema == null) {
+            schema = metadata.schema();
+        }
+
+        // Reject schema default values (initial-default / write-default)
+        for (Types.NestedField field : schema.columns()) {
+            if (field.initialDefault() != null || field.writeDefault() != null) {
+                throw new TrinoException(NOT_SUPPORTED, "Iceberg v3 column default values are not supported");
+            }
+        }
+
+        // Reject Iceberg table encryption
+        if (!metadata.encryptionKeys().isEmpty() || snapshot.keyId() != null || metadata.properties().containsKey("encryption.key-id")) {
+            throw new TrinoException(NOT_SUPPORTED, "Iceberg table encryption is not supported");
+        }
+    }
+
     private IcebergTableHandle tableHandleForCurrentSnapshot(ConnectorSession session, SchemaTableName tableName, BaseTable table)
     {
         return tableHandleForSnapshot(
@@ -652,6 +687,7 @@ public class IcebergMetadata
             Schema tableSchema,
             Optional<PartitionSpec> partitionSpec)
     {
+        validateTableForTrino(table, tableSnapshotId);
         Map<String, String> tableProperties = table.properties();
         return new IcebergTableHandle(
                 tableName.getSchemaName(),
@@ -1427,9 +1463,10 @@ public class IcebergMetadata
     public ConnectorInsertTableHandle beginInsert(ConnectorSession session, ConnectorTableHandle tableHandle, List<ColumnHandle> columns, RetryMode retryMode)
     {
         IcebergTableHandle table = (IcebergTableHandle) tableHandle;
-        Table icebergTable = catalog.loadTable(session, table.getSchemaTableName());
+        BaseTable icebergTable = (BaseTable) catalog.loadTable(session, table.getSchemaTableName());
 
         validateNotModifyingOldSnapshot(table, icebergTable);
+        validateTableForTrino(icebergTable, Optional.ofNullable(icebergTable.currentSnapshot()).map(Snapshot::snapshotId));
 
         beginTransaction(icebergTable);
 
@@ -1689,12 +1726,27 @@ public class IcebergMetadata
 
         Table icebergTable = catalog.loadTable(session, tableHandle.getSchemaTableName());
 
+        verifyTableVersionForExecute(ADD_FILES, 2, icebergTable);
+
         return Optional.of(new IcebergTableExecuteHandle(
                 tableHandle.getSchemaTableName(),
                 ADD_FILES,
                 new IcebergAddFilesHandle(location, format, recursiveDirectory),
                 icebergTable.location(),
                 icebergTable.io().properties()));
+    }
+
+    private static void verifyTableVersionForExecute(IcebergTableProcedureId icebergTableProcedureId, int maxVersion, Table icebergTable)
+    {
+        int tableFormatVersion = formatVersion(icebergTable);
+        if (tableFormatVersion > maxVersion) {
+            throw new TrinoException(NOT_SUPPORTED, format(
+                    "%s is not supported for Iceberg table format version > %d. Table %s format version is %s.",
+                    icebergTableProcedureId.name(),
+                    maxVersion,
+                    icebergTable.name(),
+                    tableFormatVersion));
+        }
     }
 
     private Optional<ConnectorTableExecuteHandle> getTableHandleForAddFilesFromTable(ConnectorSession session, ConnectorAccessControl accessControl, IcebergTableHandle tableHandle, Map<String, Object> executeProperties)
@@ -1707,6 +1759,9 @@ public class IcebergMetadata
         Map<String, String> partitionFilter = (Map<String, String>) executeProperties.get("partition_filter");
         RecursiveDirectory recursiveDirectory = (RecursiveDirectory) executeProperties.getOrDefault("recursive_directory", "fail");
 
+        Table icebergTable = catalog.loadTable(session, tableHandle.getSchemaTableName());
+        verifyTableVersionForExecute(ADD_FILES_FROM_TABLE, OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION, icebergTable);
+
         HiveMetastore metastore = metastoreFactory.orElseThrow(() -> new TrinoException(NOT_SUPPORTED, "This catalog does not support add_files_from_table procedure"))
                 .createMetastore(Optional.of(session.getIdentity()));
         SchemaTableName sourceName = new SchemaTableName(schemaName, tableName);
@@ -1714,8 +1769,6 @@ public class IcebergMetadata
         accessControl.checkCanSelectFromColumns(null, sourceName, Stream.concat(sourceTable.getDataColumns().stream(), sourceTable.getPartitionColumns().stream())
                 .map(Column::getName)
                 .collect(toImmutableSet()));
-
-        Table icebergTable = catalog.loadTable(session, tableHandle.getSchemaTableName());
 
         checkProcedureArgument(
                 icebergTable.schema().columns().size() >= sourceTable.getDataColumns().size(),
@@ -1853,15 +1906,7 @@ public class IcebergMetadata
 
         validateNotModifyingOldSnapshot(table, icebergTable);
 
-        int tableFormatVersion = formatVersion(icebergTable);
-        if (tableFormatVersion > OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION) {
-            throw new TrinoException(NOT_SUPPORTED, format(
-                    "%s is not supported for Iceberg table format version > %d. Table %s format version is %s.",
-                    OPTIMIZE.name(),
-                    OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION,
-                    table.getSchemaTableName(),
-                    tableFormatVersion));
-        }
+        verifyTableVersionForExecute(OPTIMIZE, OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION, icebergTable);
 
         beginTransaction(icebergTable);
 
@@ -3119,8 +3164,12 @@ public class IcebergMetadata
 
     private static void verifyTableVersionForUpdate(IcebergTableHandle table)
     {
-        if (table.getFormatVersion() < 2) {
+        int formatVersion = table.getFormatVersion();
+        if (formatVersion < 2) {
             throw new TrinoException(NOT_SUPPORTED, "Iceberg table updates require at least format version 2");
+        }
+        if (formatVersion > 2) {
+            throw new TrinoException(NOT_SUPPORTED, "Iceberg table updates for format version %s are not supported yet".formatted(formatVersion));
         }
     }
 
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
@@ -35,6 +35,7 @@ import io.trino.plugin.base.metrics.LongCount;
 import io.trino.plugin.iceberg.delete.DeleteFile;
 import io.trino.plugin.iceberg.util.DataFileWithDeleteFiles;
 import io.trino.spi.SplitWeight;
+import io.trino.spi.TrinoException;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorSplit;
@@ -52,6 +53,7 @@ import jakarta.annotation.Nullable;
 import org.apache.iceberg.BaseFileScanTask;
 import org.apache.iceberg.CombinedScanTask;
 import org.apache.iceberg.ContentFile;
+import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.PartitionField;
 import org.apache.iceberg.PartitionSpec;
@@ -114,6 +116,7 @@ import static io.trino.plugin.iceberg.IcebergUtil.getPathDomain;
 import static io.trino.plugin.iceberg.IcebergUtil.primitiveFieldTypes;
 import static io.trino.plugin.iceberg.StructLikeWrapperWithFieldIdToIndex.createStructLikeWrapper;
 import static io.trino.plugin.iceberg.TypeConverter.toIcebergType;
+import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static io.trino.spi.type.DateTimeEncoding.packDateTimeWithZone;
 import static io.trino.spi.type.TimeZoneKey.UTC_KEY;
 import static java.lang.Math.clamp;
@@ -726,6 +729,7 @@ public class IcebergSplitSource
                 PartitionSpecParser.toJson(task.spec()),
                 PartitionData.toJson(task.file().partition()),
                 task.deletes().stream()
+                        .peek(file -> verifyNotDeletionVector(task, file))
                         .map(DeleteFile::fromIceberg)
                         .collect(toImmutableList()),
                 SplitWeight.fromProportion(clamp(getSplitWeight(task), minimumAssignedSplitWeight, 1.0)),
@@ -733,6 +737,13 @@ public class IcebergSplitSource
                 fileIoProperties,
                 cachingHostAddressProvider.getHosts(getSplitKey(task.file().location(), task.start(), task.length()), ImmutableList.of()),
                 task.file().dataSequenceNumber());
+    }
+
+    private static void verifyNotDeletionVector(FileScanTask task, org.apache.iceberg.DeleteFile deleteFile)
+    {
+        if (deleteFile.format() == FileFormat.PUFFIN) {
+            throw new TrinoException(NOT_SUPPORTED, "Iceberg deletion vector is not supported yet: data file %s, delete file %s".formatted(task.file().location(), deleteFile.location()));
+        }
     }
 
     private double getSplitWeight(FileScanTask task)
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
@@ -1,0 +1,630 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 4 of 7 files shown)


findepi:
For something as central as table format compatibility, we should have tests with some other system (Spark?) reading tables produced by Trino.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -629,9 +631,44 @@ public ConnectorTableHandle getTableHandle(
                     schemaFor(table, snapshotId),
                     Optional.empty());
         }
+        validateTableForTrino(table, Optional.ofNullable(table.currentSnapshot()).map(Snapshot::snapshotId));
         return tableHandleForCurrentSnapshot(session, tableName, table);
     }
 
+    // TODO: Remove when Iceberg v3 is fully supported
+    private static void validateTableForTrino(BaseTable table, Optional<Long> tableSnapshotId)
```
chenjian2664, line 638: If the table is upgraded from v2 to v3, all read operations on the table will be blocked, even for snapshots that do not include any v3 data

dain: There is a test showing that insert and read works.  Are you seeing something I'm not?

chenjian2664: I see that you are explicitly checking for v3 features (for example, default value) and rejecting queries against v3 tables. There is a possible scenario where another engine upgrades a table from v2 to v3, performs inserts and/or updates (including deletes), and then enables v3-specific features such as default value

In this case, what is the expected behavior when Trino queries such a table? how should this behave when time travel is used to query a snapshot from before upgrade?

dain: Yes. That is the point of this PR.  Today, there are no v3 tables allowed at all, so any query against them will fail.  With this PR you can create, insert and read v3 tables, and most other things fail.  If you have defaults the table can not be used, but remember it can't be used today.

This is an iterative PR. Allow what works to be used and explicitly fail on everything else.  Then additional PRs will implement the remaining features one at a time.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -622,6 +623,7 @@ public ConnectorTableHandle getTableHandle(
 
         if (endVersion.isPresent()) {
             long snapshotId = getSnapshotIdFromVersion(session, table, endVersion.get());
+            validateTableForTrino(table, Optional.of(snapshotId));
```
electrum: Move this call into `tableHandleForSnapshot` so that it also covers the materialized view storage table path from above.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1717,6 +1769,8 @@ private Optional<ConnectorTableExecuteHandle> getTableHandleForAddFilesFromTable
 
         Table icebergTable = catalog.loadTable(session, tableHandle.getSchemaTableName());
 
+        verifyTableVersionForExecute(ADD_FILES_FROM_TABLE, OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION, icebergTable);
```
electrum: It seems strange to have a separate version property for optimize, but I suppose we'll remove that when v3 is fully supported.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -726,6 +730,7 @@ private IcebergSplit toIcebergSplit(FileScanTaskWithDomain taskWithDomain)
                 PartitionSpecParser.toJson(task.spec()),
                 PartitionData.toJson(task.file().partition()),
                 task.deletes().stream()
+                        .peek(df -> verifyNotDeletionVector(task, df))
```
electrum: Nit: rename `df` to `file`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -735,6 +740,13 @@ private IcebergSplit toIcebergSplit(FileScanTaskWithDomain taskWithDomain)
                 task.file().dataSequenceNumber());
     }
 
+    private static void verifyNotDeletionVector(FileScanTask task, org.apache.iceberg.DeleteFile deleteFile)
+    {
+        if (deleteFile.format() == FileFormat.PUFFIN) {
+            throw new TrinoException(NOT_SUPPORTED, format("Iceberg deletion vector is not supported yet: data file %s, delete file %s", task.file().location(), deleteFile.location()));
```
electrum: Nit: use `.formatted()` instead of `format()`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+        String tableName = "test_upgrade_v2_to_v3_" + randomNameSuffix();
+
+        assertUpdate("CREATE TABLE " + tableName + " (id integer) WITH (format_version = 2)");
+        assertUpdate("INSERT INTO " + tableName + " VALUES 1", 1);
+
+        assertUpdate("ALTER TABLE " + tableName + " SET PROPERTIES format_version = 3");
+
+        // After upgrade, inserts should continue to work.
+        assertUpdate("INSERT INTO " + tableName + " VALUES 2", 1);
+        assertThat(query("SELECT count(*) FROM " + tableName))
```
electrum: It would be better to make this `SELECT *` and verify the values here

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+    void testDeleteV3TableFails()
+    {
+        String tableName = "test_delete_v3_fails_" + randomNameSuffix();
+
+        assertUpdate("CREATE TABLE " + tableName + " (id integer) WITH (format_version = 3)");
+        assertUpdate("INSERT INTO " + tableName + " VALUES 1, 2, 3", 3);
+
+        assertThat(query("DELETE FROM " + tableName + " WHERE id = 2"))
+                .failure()
+                .hasMessage("Iceberg table updates for format version 3 are not supported yet");
```
electrum, line 185: Add `.hasErrorCode(NOT_SUPPORTED)` for all of these

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+
+        String source = "hive.tpch.add_files_src_" + suffix;
+        String target = "add_files_target_" + suffix;
+
+        // Create a single ORC data file in Hive
+        assertUpdate("CREATE TABLE " + source + " WITH (format = 'ORC') AS SELECT 1 x", 1);
+
+        // Find that file’s directory (we’ll point add_files at it)
+        String sourceFilePath = (String) computeScalar("SELECT \"$path\" FROM " + source + " LIMIT 1");
+        String sourceDir = new java.io.File(sourceFilePath).getParent();
```
electrum: Why does this need to be qualified?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+
+    @Test
+    void testV3RejectsAddFilesProcedure()
+    {
+        String suffix = randomNameSuffix();
+
+        String source = "hive.tpch.add_files_src_" + suffix;
+        String target = "add_files_target_" + suffix;
+
+        // Create a single ORC data file in Hive
```
electrum: This isn't a Hive table

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+        // Find that file’s directory (we’ll point add_files at it)
+        String sourceFilePath = (String) computeScalar("SELECT \"$path\" FROM " + source + " LIMIT 1");
+        String sourceDir = new java.io.File(sourceFilePath).getParent();
+        String sourceDirUri = new java.io.File(sourceDir).toURI().toString(); // e.g. file:/.../
+
+        // Create empty Iceberg table, then upgrade to v3
+        assertUpdate("CREATE TABLE " + target + " (x integer) WITH (format = 'ORC', format_version = 2)");
+        assertUpdate("ALTER TABLE " + target + " SET PROPERTIES format_version = 3");
+
+        assertThat(query("ALTER TABLE " + target + " EXECUTE add_files(location => '" + sourceDirUri + "', format => 'ORC')"))
```
electrum: For the purpose of this test, any location should be fine (doesn't have to exist)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1717,6 +1769,8 @@ private Optional<ConnectorTableExecuteHandle> getTableHandleForAddFilesFromTable
 
         Table icebergTable = catalog.loadTable(session, tableHandle.getSchemaTableName());
```
electrum: If we load this first, then we don't need the source table in the test

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+
+        String registered = "registered_v3_write_defaults_" + randomNameSuffix();
+        assertUpdate("CALL system.register_table(CURRENT_SCHEMA, '%s', '%s')"
+                .formatted(registered, hadoopTableLocation));
+
+        assertQueryFails(
+                "SELECT * FROM " + registered,
+                ".*Iceberg v3 column default values are not supported.*");
+
+        assertUpdate("DROP TABLE " + temp);
```
electrum, line 376: Also drop `registered`? Do we allow that?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+                .rootCause()
+                .hasMessageContaining("Iceberg deletion vector is not supported yet");
+
+        assertUpdate("DROP TABLE " + registered);
+        deleteRecursively(hadoopTableLocation);
+
+        assertUpdate("DROP TABLE " + temp);
+    }
+
+    private static DataFile orcsDataFile(String dataFilePath, long size)
```
electrum: Should be named `orc` not `orcs`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+
+    private static void injectEncryptionKeysIntoMetadataJson(Path tableLocation, String keyId)
+            throws IOException
+    {
+        Path metadataFile = latestMetadataJson(tableLocation);
+
+        ObjectMapper mapper = new ObjectMapper();
+        ObjectNode root = (ObjectNode) mapper.readTree(metadataFile.toFile());
+
+        // Add "encryption-keys"
```
electrum: This is confusing because the `encryption-keys` part comes at the end, and there is a comment in the middle. If we create `key` first, that would improve the readability.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+                        ((ObjectNode) snapshotNode).put("key-id", keyId);
+                        break;
+                    }
+                }
+            }
+        }
+
+        Files.writeString(metadataFile, mapper.writeValueAsString(root));
+        // delete the crc file, since it is no longer valid
+        Path crc = metadataFile.resolveSibling("." + metadataFile.getFileName() + ".crc");
```
electrum, line 606: Is this needed? I assume HDFS would handle this itself

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+        Files.deleteIfExists(crc);
+    }
+
+    private static Path latestMetadataJson(Path tableLocation)
+            throws IOException
+    {
+        Path metadataDir = tableLocation.resolve("metadata");
+        try (var stream = Files.list(metadataDir)) {
+            return stream
+                    .filter(p -> p.getFileName().toString().endsWith(".metadata.json"))
```
electrum: Rename `p` to `path` or `file`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+                    .max(Comparator.comparing(p -> p.getFileName().toString()))
```
electrum: I think this can be `naturalOrder()` since `Path` is comparable

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,592 @@
...
+                    .orElseThrow(() -> new IllegalStateException("No metadata.json found in " + metadataDir));
+        }
+    }
+
+    private static void deleteRecursively(Path path)
```
electrum: Use `deleteRecursively(path, ALLOW_INSECURE)` from `MoreFiles`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
dain: It is needed.  We are modifying the file directly outside of HDFS.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -633,6 +634,40 @@ public ConnectorTableHandle getTableHandle(
...
+
+        // Reject schema default values (initial-default / write-default)
+        for (Types.NestedField field : schema.columns()) {
+            if (field.initialDefault() != null || field.writeDefault() != null) {
+                throw new TrinoException(NOT_SUPPORTED, "Iceberg v3 column default values are not supported");
+            }
+        }
+
+        // Reject Iceberg table encryption
+        if (!metadata.encryptionKeys().isEmpty() || snapshot.keyId() != null) {
```
findinpath: Should we have in `io.trino.plugin.iceberg.IcebergMetadata#getTableHandle` ?

```
        if (table.properties().containsKey(ENCRYPTION_TABLE_KEY)) {
            // TODO https://starburstdata.atlassian.net/browse/CONNECT-577 Support table encryption
            throw new TrinoException(NOT_SUPPORTED, "Table encryption is not supported for: " + tableName);
        }
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -633,6 +634,40 @@ public ConnectorTableHandle getTableHandle(
...
     }
 
+    // TODO: Remove when Iceberg v3 is fully supported
+    private static void validateTableForTrino(BaseTable table, Optional<Long> tableSnapshotId)
+    {
+        Snapshot snapshot = tableSnapshotId
+                .map(table::snapshot)
+                .orElse(table.currentSnapshot());
+        if (snapshot == null) {
+            // the snapshot does not exist, this is an error that will be handled elsewhere
```
findinpath: That's not an error

```
spark-sql (default)> create table t1 (data integer) using iceberg tblproperties ('format-version' = 3);
```

Spark does not create a snapshot when doing `CREATE TABLE`

findinpath:
Related failues


```
Error:    TestIcebergV3.testV3RejectsColumnDefaults:283->AbstractTestQueryFramework.assertUpdate:411->AbstractTestQueryFramework.assertUpdate:416 » QueryFailed Iceberg v3 column default values are not supported
Error:    TestIcebergV3.testV3RejectsColumnWriteDefaults:330->AbstractTestQueryFramework.assertUpdate:411->AbstractTestQueryFramework.assertUpdate:416 » QueryFailed Iceberg v3 column default values are not supported
Error:    TestIcebergV3.testV3RejectsDeletionVectorsPuffinDeleteFile:480 » FileSystem /tmp/TrinoTest13055050090781735228/iceberg_data/hadoop_v3_dv_qdovndcpa6: failed to delete one or more files; see suppressed exceptions for details
Error:    TestIcebergV3.testV3RejectsEncryptionKeys:416->AbstractTestQueryFramework.assertUpdate:411->AbstractTestQueryFramework.assertUpdate:416 » QueryFailed Iceberg table encryption is not supported
```

ebyhr (commented):
There is a bug about column defaults. The INSERT statement in the below test doesn't throw an exception. 

```java
    @Test
    void testWriteDefault()
    {
        String tableName = "tmp_v3_defaults_src_" + randomNameSuffix();
        assertUpdate("CREATE TABLE " + tableName + " (id INTEGER, data INTEGER) WITH (format_version = 3, format = 'ORC')");
        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 10)", 1);

        Table icebergTable = loadTable(tableName);
        icebergTable.updateSchema()
                .updateColumnDefault("data", Expressions.lit(42))
                .commit();

        assertQueryFails(
                "INSERT INTO " + tableName + " (id) VALUES (2)",
                ".*Iceberg v3 column default values are not supported.*");

    }
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,554 @@
...
+
+        assertQueryFails(
+                "SELECT * FROM " + registered,
+                ".*Iceberg table encryption is not supported.*");
+
+        assertUpdate("DROP TABLE " + registered);
+        assertUpdate("DROP TABLE " + temp);
+        deleteRecursively(hadoopTableLocation, ALLOW_INSECURE);
+    }
+
```
ebyhr, line 495: The handling of `encryption.key-id` table property looks missing. The INSERT statement in the below test doesn't throw an exception:
```java
    @Test
    void testUnsupportedTableEncryption()
    {
        String tableName = "test_encryption" + randomNameSuffix();
        assertUpdate("CREATE TABLE " + tableName + "(x int)");
        try {
            BaseTable icebergTable = loadTable(tableName);
            icebergTable.updateProperties().set(ENCRYPTION_TABLE_KEY, "test_key").commit();

            assertQueryFails("SELECT * FROM " + tableName, "...");
        }
        finally {
            metastore.dropTable("tpch", tableName, true);
        }
    }
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,554 @@
...
+        Schema schema = new Schema(Types.NestedField.optional(1, "id", Types.IntegerType.get()));
+
+        Table icebergTable = new HadoopTables(new Configuration(false)).create(
+                schema,
+                PartitionSpec.unpartitioned(),
+                SortOrder.unsorted(),
+                ImmutableMap.of(
+                        "format-version", "3",
+                        "write.format.default", "ORC"),
+                hadoopTableLocation.toString());
```
ebyhr, line 519: We don't need to use HadoopTables in most tests. We can use `loadTable` method for existing tables, or `TrinoCatalog` (there's `IcebergTestUtils#getTrinoCatalog`) for new tables.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -53,7 +53,7 @@
 public class IcebergConfig
 {
     public static final int FORMAT_VERSION_SUPPORT_MIN = 1;
-    public static final int FORMAT_VERSION_SUPPORT_MAX = 2;
+    public static final int FORMAT_VERSION_SUPPORT_MAX = 3;
```
ebyhr, line 56: Did you skip updating `iceberg.md` intentionally?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -633,6 +634,40 @@ public ConnectorTableHandle getTableHandle(
...
+            return;
+        }
+
+        TableMetadata metadata = table.operations().current();
+        if (metadata.formatVersion() < 3) {
+            return;
+        }
+
+        Schema schema = metadata.schemasById().get(snapshot.schemaId());
+        if (schema == null) {
```
chenjian2664, line 654: Under what circumstances did you encounter this? I would not expect it to be null

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,597 @@
...
+        assertUpdate("CREATE TABLE " + tableName + " (id integer) WITH (format_version = 3)");
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
+    @Test
+    void testUpgradeV2ToV3Allowed()
+    {
+        String tableName = "test_upgrade_v2_to_v3_" + randomNameSuffix();
+
+        assertUpdate("CREATE TABLE " + tableName + " (id integer) WITH (format_version = 2)");
```
chenjian2664: Could you use `newTrinoTable` with `TestTable` to handle test table creation and cleanup?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,597 @@
...
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a'), (2, 'b'), (3, 'c')", 3);
+
+        BaseTable table = loadTable(tableName);
+        table.refresh();
+
+        Snapshot snapshot = table.currentSnapshot();
+        assertThat(snapshot).isNotNull();
+
+        TableMetadata metadata = table.operations().current();
+        assertThat(metadata.formatVersion()).isGreaterThanOrEqualTo(3);
```
chenjian2664:
```suggestion
        assertThat(metadata.formatVersion()).isEqualTo(3);
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -0,0 +1,597 @@
...
+    void testCreateV3TableAllowed()
+    {
+        String tableName = "test_create_v3_table_allowed_" + randomNameSuffix();
+
+        assertUpdate("CREATE TABLE " + tableName + " (id integer) WITH (format_version = 3)");
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
+    @Test
+    void testUpgradeV2ToV3Allowed()
```
chenjian2664, line 125: Could you add v1 -> v3 is allowed, and v3 -> v2 or v1 is failed?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findinpath: see https://github.com/trinodb/trino/pull/27786#discussion_r2668767728

dain: The snapshot can be null for an empty table that has been created but has no data inserted yet. As findinpath noted above, Spark (and other engines) don't create a snapshot when doing CREATE TABLE. The null check prevents NPE when validating an empty table.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
dain: Updated

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
dain: testV3RejectsColumnDefaults and testV3RejectsColumnWriteDefaults require HadoopTables because they need to create tables with v3 column default features that cannot be created through Trino SQL. These tests verify Trino correctly rejects v3 features when reading tables created by other engines.

ebyhr: That isn't true. You can create such tables like this:
```java
        catalog.newCreateTableTransaction(
                        SESSION,
                        schemaTableName,
                        new Schema(
                                Types.NestedField.optional("id")
                                        .withId(1)
                                        .ofType(Types.IntegerType.get())
                                        .withInitialDefault(Expressions.lit(42))
                                        .build()),
                        PartitionSpec.unpartitioned(),
                        SortOrder.unsorted(),
                        Optional.ofNullable(catalog.defaultTableLocation(SESSION, schemaTableName)),
                        ImmutableMap.of("format-version", "3"))
                .commitTransaction();
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
chenjian2664: @findinpath @dain If the table was created by Spark and is empty, then an empty snapshot is expected. However, reaching this line indicates that a snapshot exists but has no schema, which should be an invalid state. I do not see a valid scenario where this can happen. I would suggest either removing this code path or explicitly failing fast by throwing an exception. thoughts?
cc @ebyhr

PR #27789, opened 2025-12-30
Compress test tables as a zip binary for Trino Hudi smoke tests

vamsikarnika:
## Description
Compress test tables as a zip binary for Trino smoke tests to reduce number of file changes every time we add a new test table.

## Release notes

```markdown
## Section
```


ebyhr:
I don't think we should merge this PR, since the ZIP file is hard to review and the git diff isn't human-readable.

Adding a new test table as resource isn't mandatory. Creating the table with Hudi library on the fly is also fine.

PR #27790, opened 2025-12-30
Change default column mapping mode to name in Delta Lake

ebyhr:
## Description

Change the default column mapping mode in Delta Lake to name so the connector can allow more column operations by default. As a side effect, this change updates the default reader version to 2 and the writer version to 5.

This PR also adds `delta.column-mapping-mode` config property so users can revert the default value to `none` if necessary.

All Databricks LTS versions support the name column mapping mode and those reader and writer versions.

## Release notes

```markdown
## Delta Lake
* TBD. ({issue}`issuenumber`)
```


findepi:
> All Databricks LTS versions support the name column mapping mode and those reader and writer versions.

What is the default column mapping mode used by databricks and latest delta oss?

docs/src/main/sphinx/connector/delta-lake.md
```diff
@@ -189,6 +189,9 @@ values. Typical usage does not require you to configure them.
     the [VACUUM](delta-lake-vacuum) procedure. The equivalent catalog session
     property is `vacuum_min_retention`.
   - `7 DAYS`
+* - `delta.column-mapping-mode`
+  - Column mapping mode. Possible values are: `ID`, `NAME`, and `NONE`.
+  - `NAME`
```
chenjian2664, line 194: Why `NAME` to be the default?

ebyhr:
> What is the default column mapping mode used by databricks and latest delta oss?

Both use none as the default column mapping mode.

docs/src/main/sphinx/connector/delta-lake.md
findepi: That's a good question. With Iceberg and Delta working out some compatibility frictions, i wonder what the default mode should be (or whether the ideal exists yet).

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -458,38 +459,6 @@ public void testDropNonEmptySchemaWithTable()
         assertUpdate("DROP SCHEMA " + schemaName);
     }
 
-    @Test
-    @Override
-    public void testDropColumn()
```
findepi, line 463: Nice!

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeConfig.java
```diff
@@ -534,6 +537,20 @@ public DeltaLakeConfig setQueryPartitionFilterRequired(boolean queryPartitionFil
         return this;
     }
 
+    public ColumnMappingMode getColumnMappingMode()
+    {
+        return columnMappingMode;
+    }
+
+    @Config("delta.column-mapping-mode")
```
findepi, line 545: `delta.default-column-mapping-mode` to indicate the value selected here is still overridable during CT

docs/src/main/sphinx/connector/delta-lake.md
```diff
@@ -768,7 +771,7 @@ The following table properties are available for use:
     * `NAME`
     * `NONE`
 
-    Defaults to `NONE`.
+    Defaults to `NAME`.
```
findepi, line 774: Any other documentation pieces to update? Like do we eg say drop/rename column don't work out of the box and now they do?

docs/src/main/sphinx/connector/delta-lake.md
```diff
@@ -189,6 +189,9 @@ values. Typical usage does not require you to configure them.
     the [VACUUM](delta-lake-vacuum) procedure. The equivalent catalog session
     property is `vacuum_min_retention`.
   - `7 DAYS`
+* - `delta.column-mapping-mode`
```
findepi, line 192: update name here too

findepi (approved):
I am OK with the change directionally and i believe it's an improvement.
I haven't review the code carefully, so you may want second pair of eyes.

I am concerned about

- this may be not the last time we change default mapping mode. maybe the ideal iceberg-compact mode doesn't exist yet?
- DBX still using 'none' as the default. They also want the best for their uses (drop/rename columns working OOB), so maybe we're overlooking something when doing this switch?

chenjian2664 (commented):
I think this is an improvement and we should move it forward, even though we're not fully certain whether `ID` or `NAME` is the better long-term choice, either option is clearly better than having none by default(easy to use/manage perspective). my only concern is whether this could introduce a breaking change here - especially for the `create or replace` scenarios, but still I think it worth to do

chenjian2664:
Can we add a commit body for recording that why are we not use none as default?

PR #27791, opened 2025-12-30
Cleanup DynamicPageFilter

pettyjamesm:
## Description
Small cleanups in DynamicPageFilter to:
- Stop filter evaluation once no more positions remain selected
- Record ineffective filters in a boolean[] instead of a hash set
- Avoid branching when recording filter selectivity
- Use `SelectAllEvaluator` when dynamic filters do not remove any elements

Also adjusts `AndFilterEvaluator` to use an array instead of `List<FilterEvaluator>` and short-circuit when no more positions are active in the selection.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/gen/columnar/AndFilterEvaluator.java
+++ b/core/trino-main/src/main/java/io/trino/sql/gen/columnar/AndFilterEvaluator.java
@@ -25,8 +25,8 @@ import java.util.Optional;
 import java.util.function.Supplier;
 
 import static com.google.common.base.Preconditions.checkArgument;
-import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.trino.sql.relational.SpecialForm.Form.AND;
+import static java.util.Objects.requireNonNull;
 
 public final class AndFilterEvaluator
         implements FilterEvaluator
@@ -34,10 +34,12 @@ public final class AndFilterEvaluator
     public static Optional<Supplier<FilterEvaluator>> createAndExpressionEvaluator(ColumnarFilterCompiler compiler, SpecialForm specialForm)
     {
         checkArgument(specialForm.form() == AND, "specialForm %s should be AND", specialForm);
-        checkArgument(specialForm.arguments().size() >= 2, "AND expression %s should have at least 2 arguments", specialForm);
 
-        ImmutableList.Builder<Supplier<FilterEvaluator>> builder = ImmutableList.builder();
-        for (RowExpression expression : specialForm.arguments()) {
+        List<RowExpression> arguments = specialForm.arguments();
+        checkArgument(arguments.size() >= 2, "AND expression %s should have at least 2 arguments", specialForm);
+
+        ImmutableList.Builder<Supplier<FilterEvaluator>> builder = ImmutableList.builderWithExpectedSize(arguments.size());
+        for (RowExpression expression : arguments) {
             Optional<Supplier<FilterEvaluator>> subExpressionEvaluator = FilterEvaluator.createColumnarFilterEvaluator(expression, compiler);
             if (subExpressionEvaluator.isEmpty()) {
                 return Optional.empty();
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/sql/gen/columnar/DynamicPageFilter.java
```diff
@@ -134,32 +136,39 @@ private Supplier<FilterEvaluator> createDynamicFilterEvaluator(ColumnarFilterCom
...
+            }
+            return new DynamicFilterEvaluator(filterEvaluators, selectivityThreshold);
+        };
     }
 
     static final class DynamicFilterEvaluator
             implements FilterEvaluator
     {
-        private final List<FilterEvaluator> subFilterEvaluators;
+        private final FilterEvaluator[] subFilterEvaluators;
```
raunaqmorarka, line 151: Can we continue to use `List` ?
I don't think there will be a non-trivial improvement to performance from using an array here and the existing code was a bit easier to read and had stronger immutability guarantees.

core/trino-main/src/main/java/io/trino/sql/gen/columnar/DynamicPageFilter.java
```diff
@@ -169,38 +178,35 @@ public SelectionResult evaluate(ConnectorSession session, SelectedPositions acti
...
+            long filterInputPositions = this.filterInputPositions[filterIndex];
             filterOutputPositions[filterIndex] += selectedPositions.size();
-            if (filterInputPositions[filterIndex] < MIN_SAMPLE_POSITIONS) {
-                return;
-            }
-            if (filterOutputPositions[filterIndex] > (selectivityThreshold * filterInputPositions[filterIndex])) {
-                ineffectiveFilterChannels.add(filterIndex);
-            }
+            long filterOutputPositions = this.filterOutputPositions[filterIndex];
+            filterChannelIneffective[filterIndex] = filterInputPositions >= MIN_SAMPLE_POSITIONS & filterOutputPositions > (selectivityThreshold * filterInputPositions);
```
raunaqmorarka: I'd prefer `&&` over `&` for clarity unless there is a proven benefit

core/trino-main/src/main/java/io/trino/sql/gen/columnar/AndFilterEvaluator.java
```diff
@@ -25,42 +25,51 @@
...
+        return Optional.of(() -> {
+            FilterEvaluator[] filterEvaluators = new FilterEvaluator[subExpressionEvaluators.size()];
+            for (int i = 0; i < filterEvaluators.length; i++) {
+                filterEvaluators[i] = requireNonNull(subExpressionEvaluators.get(i).get(), "subExpressionEvaluator is null");
+            }
+            return new AndFilterEvaluator(filterEvaluators);
+        });
     }
 
-    private final List<FilterEvaluator> subFilterEvaluators;
```
raunaqmorarka, line 51: Similar comment as previous commit

core/trino-main/src/main/java/io/trino/sql/gen/columnar/AndFilterEvaluator.java
```diff
@@ -25,42 +25,51 @@
...
+        checkArgument(subFilterEvaluators.length >= 2, "must have at least 2 subexpressions to AND");
         this.subFilterEvaluators = subFilterEvaluators;
     }
 
     @Override
     public SelectionResult evaluate(ConnectorSession session, SelectedPositions activePositions, SourcePage page)
     {
         long filterTimeNanos = 0;
-        for (FilterEvaluator evaluator : subFilterEvaluators) {
+        for (int i = 0; i < subFilterEvaluators.length && !activePositions.isEmpty(); i++) {
```
raunaqmorarka: It would be clearer to write
```
if (activePositions.isEmpty()) {
      break;
}
```
within for loop body

core/trino-main/src/main/java/io/trino/sql/gen/columnar/DynamicPageFilter.java
pettyjamesm: Immutability is still guaranteed since the constructor is private and only called by the supplier creating the factory. Agreed that the performance is negligible, but still worthwhile I think in this context since it's removing a layer of pointer indirection in a performance-critical path (filter evaluation).

core/trino-main/src/main/java/io/trino/sql/gen/columnar/DynamicPageFilter.java
pettyjamesm: Probably not a huge deal here, so sure- will update to use `&&`

core/trino-main/src/main/java/io/trino/sql/gen/columnar/AndFilterEvaluator.java
pettyjamesm: Sure, changed.

core/trino-main/src/main/java/io/trino/sql/gen/columnar/AndFilterEvaluator.java
pettyjamesm: See previous comment

core/trino-main/src/main/java/io/trino/sql/gen/columnar/DynamicPageFilter.java
raunaqmorarka: I meant that an array can still be mutated in-place while an immutable list will throw an exception if that were to happen.
I'm not sure I follow the point about pointer indirection, `com.google.common.collect.RegularImmutableList#get` is still doing only an array lookup internally and JIT can inline this method if that is a concern.

pettyjamesm: I agree that this is minor, but small things can add up. There's two things going on here at a minor level for latency sensitive loops:
- `List` requires pointer chasing from `this->list->array->element` during dereferencing, whereas the array requires `this->array->element` (the extra pointer indirection is through the list). The JIT may or may not be able to flatten that initial pointer chase down to the `array` during loop iteration but the first iteration will have that extra indirection and getting the inlined iteration of the elements array from the JIT is not a given.
- The `List` elements array is an `Object[]` which incurs a check cast on each access but not when storing elements into the array. Alternatively, the typed-array incurs the check cast when storing an element- but not on access which is assumed to be much more frequent than the initial construction.

As you're pointing out, mutation of the contents of the array is a maintainability risk- but when access or ownership of the array is scoped sufficiently narrowly (like a private field that doesn't leak into callers / callees) then the array form should be preferred when elements are accessed repeatedly in performance sensitive sections of code.

PR #27799, opened 2025-12-31
Fix errorprone warning in Hudi

ebyhr:
## Release notes

### DIFF
--- a/plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
+++ b/plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
@@ -334,24 +334,22 @@ public class TpchHudiTablesInitializer
 
         static TpchColumnTypeAdapter of(TpchColumnType columnType)
         {
-            if (columnType == TpchColumnTypes.INTEGER) {
+            if (columnType.equals(TpchColumnTypes.INTEGER)) {
                 return INTEGER;
             }
-            else if (columnType == TpchColumnTypes.IDENTIFIER) {
+            if (columnType.equals(TpchColumnTypes.IDENTIFIER)) {
                 return IDENTIFIER;
             }
-            else if (columnType == TpchColumnTypes.DATE) {
+            if (columnType.equals(TpchColumnTypes.DATE)) {
                 return DATE;
             }
-            else if (columnType == TpchColumnTypes.DOUBLE) {
+            if (columnType.equals(TpchColumnTypes.DOUBLE)) {
                 return DOUBLE;
             }
-            else {
-                if (columnType.getBase() != TpchColumnType.Base.VARCHAR || columnType.getPrecision().isEmpty()) {
-                    throw new IllegalArgumentException("Illegal column type: " + columnType);
-                }
-                return VARCHAR;
+            if (columnType.getBase() != TpchColumnType.Base.VARCHAR || columnType.getPrecision().isEmpty()) {
+                throw new IllegalArgumentException("Illegal column type: " + columnType);
             }
+            return VARCHAR;
         }
 
         static HiveType toHiveType(TpchColumnType columnType)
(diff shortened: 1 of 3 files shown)


plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
```diff
@@ -334,24 +334,22 @@ private enum TpchColumnTypeAdapter
 
         static TpchColumnTypeAdapter of(TpchColumnType columnType)
         {
-            if (columnType == TpchColumnTypes.INTEGER) {
```
findepi, line 337: In fact this is correct. `TpchColumnTypes` guarantees only single instance of INTEGER type.

findepi:
> Fix errorprone warning in Hudi 

errorprone spits our many warnings. 
those we care about should be converted into errors (unless they produce false positives)
those we do not care about should be silenced
therefore the warnings being output are those which we didn't triage yet

Math-ias (approved):
Found these error prone checks without these changes (from a recent PR so should approximate the _before_ well).
 * [Something in HudiTransactionManager is effectively private](https://github.com/trinodb/trino/actions/runs/20616940825/job/59211439832#step:5:5674)
   * This one I understand the least. The reference page is [a bit empty](https://errorprone.info/bugpattern/EffectivelyPrivate).
   * Though using the private modifier on methods not implementing an interface follows [the pattern set here](https://github.com/trinodb/trino/blob/fc170a76953941d77cddd879307d439ab9892fc7/lib/trino-orc/src/main/java/io/trino/orc/OrcWriteValidation.java#L571).
 * [Empty catch in HudiPageSourceProvider](https://github.com/trinodb/trino/actions/runs/20616940825/job/59211439832#step:5:5677)
    * I was curious if we should've done something here, for example, make a log. [But it's existed since the hudi connector was added](https://github.com/trinodb/trino/blob/97882fb32b7fedf384288f09fdf86bbc8a2daba5/plugin/trino-hudi/src/main/java/io/trino/plugin/hudi/HudiPageSourceProvider.java#L253-L254), we'll error out with a trino exception anyways, and it's from reading, not writing, to a file I think it's OK to keep ignoring.
 * [Range of reference equality in TpchHudiTablesInitializer](https://github.com/trinodb/trino/actions/runs/20616940825/job/59211439832#step:5:5877)
    * @findepi mentioned we don't have to make changes here. To triage the warning I think there's 3 options.
      * Using `@SuppressWarnings("ReferenceEquality")`
      * The [error prone reference equality page mentions the use of](https://errorprone.info/bugpattern/ReferenceEquality#how-about-a-reference-equality-comparison-before-a-more-expensive-content-equality-comparison) `Objects.equals` for supporting the narrow case we've lost support for where `columnType` could be null (whether or not that's a real possibility ...).
      * Using equals like you do ([which does have precedence](https://github.com/trinodb/trino/blob/fc170a76953941d77cddd879307d439ab9892fc7/plugin/trino-hive/src/test/java/io/trino/plugin/hive/TestHivePageSink.java#L442)).
        * You could also flip the variables I suppose to support the null case mentioned above `TpchColumnTypes.INTEGER.equals(columnType)`. 

I know in your commit message you also mention handling a redundant else (that appears to be in `TpchHudiTablesInitializer.java‎`).

That matches all the changes of the PR. 👍

findepi:
> [Range of reference equality in TpchHudiTablesInitializer](https://github.com/trinodb/trino/actions/runs/20616940825/job/59211439832#step:5:5877)
> * @findepi mentioned we don't have to make changes here. To triage the warning I think there's 3 options.
>   
>   * Using `@SuppressWarnings("ReferenceEquality")`

To be clear.
This "reference equality" check doesn't sound unreasonable. In this case, it doesn't matter, it's a false positive. 

We can make the "reference equality" check blessed (i.e. make it an error and care about it). This requires doing something with every false positive (a code change if small, a suppression if particular code wording is important).

If we don't make the check an error, then addressing it's false positives looks like a redundant change to me.

I generally do care and don't ignore warnings from my IDE (pretty much zero tolerance for warnings in the IDE for code i write). However, the way we configured error-prone, means there are many (untriaged) warnings, which makes me approach these warnings differently than those from IDE.  I am open to changes in the process, but i would want to see further beyond this PR.

plugin/trino-hudi/src/test/java/io/trino/plugin/hudi/testing/TpchHudiTablesInitializer.java
chenjian2664: curios does it make sense to use a switch here

PR #27800, opened 2025-12-31
Document `WHEN STALE` clause

piotrrzysko:
## Description

This PR shouldn't be merged before https://github.com/trinodb/trino/pull/27502.

## Release notes

```markdown
## Section
```

### DIFF
--- a/docs/src/main/sphinx/sql/create-materialized-view.md
+++ b/docs/src/main/sphinx/sql/create-materialized-view.md
@@ -6,6 +6,7 @@
 CREATE [ OR REPLACE ] MATERIALIZED VIEW
 [ IF NOT EXISTS ] view_name
 [ GRACE PERIOD interval ]
+[ WHEN STALE ( INLINE | FAIL ) ]
 [ COMMENT string ]
 [ WITH properties ]
 AS query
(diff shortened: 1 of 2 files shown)


findepi:
> This PR shouldn't be merged before #27502.

drafting

docs/src/main/sphinx/sql/create-materialized-view.md
```diff
@@ -49,6 +50,20 @@ is used for querying:
...
+
+* `INLINE` - When the materialized view is stale, it is expanded like a logical
+  view, and queries accessing the materialized view will use the underlying
+  query definition to retrieve up-to-date data. This is the default behavior
+  when `WHEN STALE` is not specified.
+* `FAIL` - When the materialized view is stale, queries accessing it fail with
+  an error. 
+
+Note that the `WHEN STALE` clause requires connector support. If the connector
+does not support this feature, specifying `WHEN STALE` results in an error.
```
findepi: Given INLINE is the default behavior, i wonder why we fail CREATE MV WHEN STALE INLINE for a connector that doesn't support WHEN STALE. 

There are these user expectations in play here

1. given no WHEN STALE declaration is equivalent to WHEN STALE INLINE, a CREATE MV without WHEN STALE should work same (and succeed same) as a query with WHEN STALE INLINE
2. given SHOW CREATE MV output, i can always feed it back into trino and it should succeed
   - but the SHOW CREATE MV output doesn't have to 'remember' whether WHEN STALE was specified or leveraged the default
   - in fact, it's nice when SHOW CREATE MV clears out any ambiguities, filling in the default values user did not specify explicitly
3. when CREATE MV with WHEN STALE INLINE succeeds, i can change it to FAIL and it will also succeed

Clearly, 1 and 3 are contradictory, cannot be satisfied at the same time.
Our current approach delivers 3, but i think i like 1 more and this also gives as 2 easily.

We could relay on our own backward compatibility and allow the CREATE MV WHEN STALE INLINE statement even for a connector that is not updated for  WHEN STALE yet.
The connector capability would be only checked when requested WHEN STALE behavior is != the default one.


What do you think?

piotrrzysko: Makes sense, thanks. I’ve updated the validation to skip the connector capability check when INLINE is used. I’ll update this PR after the PR with the actual implementation is approved/merged.

PR #27821, opened 2026-01-02
Make `whenStaleBehavior` non-optional in `MaterializedViewDefinition`

piotrrzysko:
## Description

Addresses:
* https://github.com/trinodb/trino/pull/27502#discussion_r2655414258.

I didn’t address the second part regarding `ConnectorMaterializedViewDefinition`, because that would require connectors to know what the default value is. See the mapping from `IcebergMaterializedViewDefinition` to `ConnectorMaterializedViewDefinition` in `AbstractTrinoCatalog#getMaterializedViewDefinition`.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
+++ b/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
@@ -51,6 +51,7 @@ import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.util.concurrent.Futures.immediateVoidFuture;
 import static io.trino.execution.ParameterExtractor.bindParameters;
+import static io.trino.metadata.MaterializedViewDefinition.DEFAULT_WHEN_STALE_BEHAVIOR;
 import static io.trino.metadata.MetadataUtil.createQualifiedObjectName;
 import static io.trino.metadata.MetadataUtil.getRequiredCatalogHandle;
 import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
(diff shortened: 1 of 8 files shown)


core/trino-main/src/main/java/io/trino/metadata/MaterializedViewDefinition.java
```diff
@@ -80,7 +83,7 @@ public ConnectorMaterializedViewDefinition toConnectorMaterializedViewDefinition
                         .map(column -> new ConnectorMaterializedViewDefinition.Column(column.name(), column.type(), column.comment()))
                         .collect(toImmutableList()),
                 getGracePeriod(),
-                whenStaleBehavior,
+                Optional.of(whenStaleBehavior),
```
findepi, line 86: We can followup with similar change in ConnectorMaterializedViewDefinition.
the compatibity was already broken when we changed ConnectorMaterializedViewDefinition constructor 🤷‍♂️

core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java
```diff
@@ -195,15 +192,18 @@ Analysis executeInternal(
...
+    private boolean isWhenStaleBehaviorSupported(Session session, WhenStaleBehavior whenStale, CatalogHandle catalogHandle)
     {
-        // INLINE is always supported as it's the default behavior
-        return whenStaleBehavior == INLINE || plannerContext.getMetadata().getConnectorCapabilities(session, catalogHandle).contains(MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR);
+        return whenStale == DEFAULT_WHEN_STALE_BEHAVIOR ||
+               plannerContext.getMetadata().getConnectorCapabilities(session, catalogHandle).contains(MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR);
     }
 
-    private static WhenStaleBehavior toConnectorWhenStaleBehavior(CreateMaterializedView.WhenStaleBehavior whenStale)
+    private static WhenStaleBehavior toConnectorWhenStaleBehavior(Optional<CreateMaterializedView.WhenStaleBehavior> whenStale)
```
findepi, line 201: should "connector" be here?
i think it's converting to engine's when stale, not SPI's (connector's)

findepi: pre-existing, to be follow-ed up

PR #27826, opened 2026-01-02
Do not overwrite discovery.uri if specified

losipiuk:
## Release notes

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
@@ -263,7 +263,7 @@ public final class DistributedQueryRunner
             Optional<List<SystemAccessControl>> systemAccessControls,
             List<EventListener> eventListeners)
     {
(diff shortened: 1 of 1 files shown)


testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
```diff
@@ -263,13 +263,15 @@ private TestingTrinoServer createServer(
...
     {
-        if (this.coordinator != null) {
-            String discoveryUri = this.coordinator.getCurrentNode().getInternalUri() +
-                    backupCoordinator.map(backup -> "," + backup.getCurrentNode().getInternalUri()).orElse("");
-            extraProperties = ImmutableMap.<String, String>builder()
-                    .putAll(extraProperties)
-                    .put("discovery.uri", discoveryUri)
-                    .buildOrThrow();
+        if (!extraProperties.containsKey("discovery.uri")) {
+            if (this.coordinator != null) {
```
findepi: Can this be one `if`?

PR #27779, opened 2025-12-29
Reduce Preview UI Monaco Editor languages

gertjanal:
## Description
~~Reduce the dist folder with `6919918−5303972=1615946 bytes` by removing unused features and languages.~~

~~See https://www.npmjs.com/package/@bithero/monaco-editor-vite-plugin~~

This PR is now a test case for the npm check #27832 

## Release notes

( X ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #27832, opened 2026-01-02
Add npm check to validate no unwanted licenses are used in node modules

gertjanal:
## Description
Added `npm run check:licenses` to check if the `package-lock.json` contains any unwanted license like AGPL.

Live demo in this PR https://github.com/trinodb/trino/pull/27779
```
[INFO] > webapp-preview@0.0.0 check:licenses
[INFO] > TEMP=$(mktemp); jq .packages[].license package-lock.json | grep -v 'null' | sort | uniq > $TEMP; DIFFERENCES=$(grep -vxf ../allowed-licenses.txt $TEMP); DIFF_COUNT=$(echo -n $DIFFERENCES | wc -c); rm $TEMP; if [ $DIFF_COUNT != '0' ]; then echo '\033[0;31mLicense found in package-lock.json that is not in allowed-licenses.txt'; echo $DIFFERENCES; exit 1; fi
[INFO] 
[INFO] License found in package-lock.json that is not in allowed-licenses.txt
[INFO] "AGPL-3.0-or-later"
[INFO] 
```

## Release notes


mosabua:
I think thats a good idea. We might need to expand to other licenses that are incompatible with Apache  but this is a good start

gertjanal:
Note to self:
`grep '"license": "' core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json | sort | uniq`

```json
{
      "license": "0BSD",
      "license": "AGPL-3.0-or-later", // In PR #27779
      "license": "Apache-2.0",
      "license": "BSD-2-Clause",
      "license": "BSD-3-Clause"
      "license": "BSD-3-Clause",
      "license": "CC-BY-4.0"
      "license": "ISC"
      "license": "ISC",
      "license": "MIT",
      "license": "MIT",
      "license": "MIT AND ISC",
      "license": "OFL-1.1",
      "license": "Python-2.0",
      "license": "Unlicense"
}
```

And for the original webapp:

```json
{
      "license": "Apache-2.0"
      "license": "Apache-2.0",
      "license": "BSD-2-Clause"
      "license": "BSD-2-Clause",
      "license": "BSD-3-Clause"
      "license": "BSD-3-Clause",
      "license": "CC-BY-4.0"
      "license": "ISC"
      "license": "ISC",
      "license": "MIT"
      "license": "MIT",
}
```

core/trino-web-ui/src/main/resources/webapp/src/package.json
```diff
@@ -35,8 +35,9 @@
     "package:clean": "npm clean-install && webpack --config webpack.config.js",
     "watch": "npm clean-install && webpack --config webpack.config.js --watch",
     "flow": "flow",
-    "check": "npm install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx",
+    "check": "npm install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx && npm run check:licenses",
     "check:clean": "npm clean-install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx",
+    "check:licenses": "if grep -q -i '\"license\": \"AGPL-' package-lock.json ; then echo '\\033[0;31mAGPL license found in package-lock.json'; exit 1; fi",
```
findepi: Rather than looking for disallowed license strings, we should rather look for any licenses that are not explicitly allowed.

also, `package-lock.json` is a JSON file. can we use a JSON-aware tool such as `jq` to process its contents?

.github/workflows/ci.yml
```diff
@@ -96,6 +96,10 @@ jobs:
         run: |
           export MAVEN_OPTS="${MAVEN_INSTALL_OPTS}"
           $MAVEN clean verify -B --strict-checksums -V -T 1C -DskipTests -P ci
+      - name: Maven NPM Checks
+        run: |
+          export MAVEN_OPTS="${MAVEN_INSTALL_OPTS}"
+          $MAVEN clean verify -pl io.trino:trino-web-ui -DskipTests -Dfrontend.package.goal=package -Dfrontend.check.goal=check
```
findepi: Shouldn't these checks be run by `mvn verify` out of the box? then we wouldn't need any ci.yml change.

core/trino-web-ui/src/main/resources/webapp/src/package.json
gertjanal: Yes I considered to look for unknown licenses, but as my previous "note to self" already shows, the values seem pretty "free-format", which makes it harder to add valid dependencies that have a license that is not in my allow list.

findepi: > which makes it harder to add valid dependencies that have a license that is not in my allow list.

if we allow "Apache-2.0" and a dep has "Apache 2.0" then we may choose to update the check adding "Apache 2.0" as also allowed. this is not a big deal.

gertjanal: I've created a `allow-licenses.txt` file

.github/workflows/ci.yml
gertjanal: Found it:

```xml
    <profiles>
        <profile>
            <id>ci</id>
            <activation>
                <activeByDefault>false</activeByDefault>
                <property>
                    <name>env.CI</name>
                </property>
            </activation>
            <properties>
                <frontend.package.goal>package:clean</frontend.package.goal>
                <frontend.check.goal>check:clean</frontend.check.goal>
            </properties>
        </profile>
    </profiles>
</project>
```

core/trino-web-ui/src/main/resources/webapp/src/package.json
```diff
@@ -35,8 +35,9 @@
     "package:clean": "npm clean-install && webpack --config webpack.config.js",
     "watch": "npm clean-install && webpack --config webpack.config.js --watch",
     "flow": "flow",
-    "check": "npm install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx",
-    "check:clean": "npm clean-install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx",
+    "check": "npm install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx && npm run check:licenses",
```
chenjian2664, line 38: Is the check triggered while building, or at some later stage, just wanna confirm if user push commits multi times, will the check run every time?

gertjanal: Yes, it will run every time

core/trino-web-ui/src/main/resources/allowed-licenses.txt
```diff
@@ -0,0 +1,11 @@
...
+"Apache-2.0"
+"BSD-2-Clause"
+"BSD-3-Clause"
+"CC-BY-4.0"
+"ISC"
+"MIT"
+"MIT AND ISC"
+"OFL-1.1"
+"Python-2.0"
+"Unlicense"
```
findepi, line 11: @martint PTAL

core/trino-web-ui/src/main/resources/webapp/src/package.json
```diff
@@ -35,8 +35,9 @@
...
+    "check:clean": "npm clean-install && flow && prettier --check **/*.js **/*.jsx *.js *.jsx && npm run check:licenses",
+    "check:licenses": "TEMP=$(mktemp); jq .packages[].license package-lock.json | grep -v 'null' | sort | uniq > $TEMP; DIFFERENCES=$(grep -vxf ../../allowed-licenses.txt $TEMP); DIFF_COUNT=$(echo -n $DIFFERENCES | wc -c); rm $TEMP; if [ $DIFF_COUNT != '0' ]; then echo '\\033[0;31mLicense found in package-lock.json that is not in allowed-licenses.txt'; echo $DIFFERENCES; exit 1; fi",
```
findepi: why `grep -v 'null'`? sounds like lack of license is a strong no?

nit: `sort | uniq` → `sort -u`

`grep -vxf ../../allowed-licenses.txt` - you likely want `-F` (fixed strings) too. The license names are not to be treated as patterns.

nit: `DIFF_COUNT` variable can likely be inlined. Did you try this?

```sh
if ! DIFFERENCES=$(grep ...); then
  echo "bad bad" >&2
  exit 1
fi
```

the error message should be echoed into stderr with `>&2`

gertjanal: Done!

gertjanal: The `grep -v 'null'` is indeed because one or multiple npm packages have no license

findepi: That sounds like a problem. If we don't know their license, what are the basis for including them?

core/trino-web-ui/src/main/resources/allowed-licenses.txt
```diff
@@ -0,0 +1,11 @@
+"0BSD"
+"Apache-2.0"
+"BSD-2-Clause"
+"BSD-3-Clause"
+"CC-BY-4.0"
+"ISC"
+"MIT"
+"MIT AND ISC"
+"OFL-1.1"
```
martint, line 9: This one is for fonts, not for code. Why do we need it?

core/trino-web-ui/src/main/resources/allowed-licenses.txt
```diff
@@ -0,0 +1,11 @@
+"0BSD"
+"Apache-2.0"
+"BSD-2-Clause"
+"BSD-3-Clause"
+"CC-BY-4.0"
```
martint, line 5: This one is compatible with the Apache License but requires attribution. In practice, it means we need to have entries for each dependency that uses this license in the NOTICE file (https://github.com/trinodb/trino/blob/master/core/trino-server-core/NOTICE)

core/trino-web-ui/src/main/resources/allowed-licenses.txt
martint: This one is ok.

core/trino-web-ui/src/main/resources/allowed-licenses.txt
gertjanal: It's just the reported license in https://github.com/trinodb/trino/blob/master/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json#L1170 that is added to the list of allowed licenses
```json
{
"node_modules/@fontsource/roboto": {
      "version": "5.2.8",
      "resolved": "https://registry.npmjs.org/@fontsource/roboto/-/roboto-5.2.8.tgz",
      "integrity": "sha512-oh9g4Cg3loVMz9MWeKWfDI+ooxxG1aRVetkiKIb2ESS2rrryGecQ/y4pAj4z5A5ebyw450dYRi/c4k/I3UBhHA==",
      "license": "OFL-1.1",
      "funding": {
        "url": "https://github.com/sponsors/ayuhito"
      }
    },
}
```

core/trino-web-ui/src/main/resources/allowed-licenses.txt
gertjanal: I'm not sure if you want me to do something, this file is just a list of all the unique licenses that are declared in both package-lock.json files (https://github.com/trinodb/trino/blob/master/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json and https://github.com/trinodb/trino/blob/master/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json)

core/trino-web-ui/src/main/resources/webapp/src/package.json
gertjanal: I see that the web ui is part of the package-lock as well (first element with key `""`) that is causing the `null` license as there is no `license` entry. I've update the `jq` selector to omit this entry, so the `grep -v null` is removed

core/trino-web-ui/src/main/resources/allowed-licenses.txt
gertjanal: Reference: https://github.com/trinodb/trino/blob/master/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json#L6708

core/trino-web-ui/src/main/resources/allowed-licenses.txt
martint: No change requested. However, we may want to add a comment next to it indicating that an entry needs to be added to the NOTICE file whenever a new dependency with this license is added.

findepi: filed issue to track this
- https://github.com/trinodb/trino/issues/28272

findepi:
Merged, thanks!

prrvchr:
I can no longer compile unless I modify the file: `core/trino-web-ui/src/main/resources/allowed-licenses.txt` with:

```
"0BSD"
"Apache-2.0"
"BSD-2-Clause"
"BSD-3-Clause"
"CC-BY-4.0"
"ISC"
"MIT"
"MIT AND ISC"
"OFL-1.1"
"Python-2.0"
"Unlicense"
"(MPL-2.0 OR Apache-2.0)"
```
Is this normal?

findepi:
thank you @prrvchr for heads up. this should help

- https://github.com/trinodb/trino/pull/28275

prrvchr:
@findepi Your welcome

PR #27836, opened 2026-01-03
[Iceberg v3]  Row lineage

dain:
## Description
This PR adds comprehensive support for Iceberg v3 row lineage, enabling Trino to read and preserve row identity metadata (`$row_id` and `$last_updated_sequence_number`).

Changes include:
- Reading lineage columns: Expose `$row_id` and `$last_updated_sequence_number` as queryable columns for v3 tables
- OPTIMIZE is now enabled for v3 tables and preserves lineage data
- UPDATE and MERGE operations preserve the original `$row_id` for modified rows, maintaining row identity across updates
- Enable cleanup procedures (expire_snapshots, remove_orphan_files) on v3 tables
- Run Iceberg connector tests against both v2 and v3 format versions

## Release notes

```markdown
## Iceberg
* Add support for Iceberg v3 row lineage, including reading `$row_id` and `$last_updated_sequence_number` columns, preserving row identity during UPDATE/MERGE operations, and OPTIMIZE support. ({issue}`issuenumber)

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -196,13 +196,15 @@ public abstract class BaseIcebergConnectorTest
     private static final Pattern WITH_CLAUSE_EXTRACTOR = Pattern.compile(".*(WITH\\s*\\([^)]*\\))\\s*$", Pattern.DOTALL);
 
     protected final IcebergFileFormat format;
+    protected final int formatVersion;
 
     protected TrinoFileSystem fileSystem;
     protected TimeUnit storageTimePrecision;
 
-    protected BaseIcebergConnectorTest(IcebergFileFormat format)
+    protected BaseIcebergConnectorTest(IcebergFileFormat format, int formatVersion)
     {
         this.format = requireNonNull(format, "format is null");
+        this.formatVersion = formatVersion;
     }
 
     @Override
@@ -218,6 +220,7 @@ public abstract class BaseIcebergConnectorTest
         return IcebergQueryRunner.builder()
                 .setIcebergProperties(ImmutableMap.<String, String>builder()
                         .put("iceberg.file-format", format.name())
+                        .put("iceberg.format-version", String.valueOf(formatVersion))
                         // Only allow some extra properties. Add "sorted_by" so that we can test that the property is disallowed by the connector explicitly.
                         .put("iceberg.allowed-extra-properties", "extra.property.one,extra.property.two,extra.property.three,sorted_by")
                         // Allows testing the sorting writer flushing to the file system with smaller tables
@@ -277,22 +280,26 @@ public abstract class BaseIcebergConnectorTest
             case SUPPORTS_CREATE_OR_REPLACE_TABLE,
                  SUPPORTS_REPORTING_WRITTEN_BYTES -> true;
             case SUPPORTS_ADD_COLUMN_NOT_NULL_CONSTRAINT,
-                 SUPPORTS_DEFAULT_COLUMN_VALUE,
                  SUPPORTS_LIMIT_PUSHDOWN,
                  SUPPORTS_REFRESH_VIEW,
                  SUPPORTS_RENAME_MATERIALIZED_VIEW_ACROSS_SCHEMAS,
                  SUPPORTS_TOPN_PUSHDOWN -> false;
+            case SUPPORTS_DEFAULT_COLUMN_VALUE -> formatVersion >= 3;
             default -> super.hasBehavior(connectorBehavior);
         };
     }
 
-    @Override
     @Test
+    @Override
     public void testCreateTableWithDefaultColumn()
     {
-        String tableName = "test_default_value_" + randomNameSuffix();
-        assertThatThrownBy(() -> assertUpdate("CREATE TABLE " + tableName + " (x int DEFAULT 1)"))
-                .hasMessageContaining("Default column values are not supported for Iceberg table format version < 3");
+        if (formatVersion < 3) {
+            String tableName = "test_default_value_" + randomNameSuffix();
+            assertQueryFails("CREATE TABLE " + tableName + " (x int DEFAULT 1)", "Default column values are not supported for Iceberg table format version < 3");
+            return;
+        }
+
+        super.testCreateTableWithDefaultColumn();
     }
 
     @Test
@@ -424,7 +431,7 @@ public abstract class BaseIcebergConnectorTest
                         ")\n" +
                         "WITH (\n" +
                         "   format = '" + format.name() + "',\n" +
-                        "   format_version = 2,\n" +
+                        "   format_version = " + formatVersion + ",\n" +
                         "   location = '\\E.*/tpch/orders-.*\\Q'\n" +
                         ")\\E");
     }
@@ -2064,11 +2071,12 @@ public abstract class BaseIcebergConnectorTest
                 """
                 WITH (
                    format = '%s',
-                   format_version = 2,
+                   format_version = %s,
                    location = '%s',
                    partitioning = ARRAY['adate']
                 )""",
                 format,
+                formatVersion,
                 tempDirPath));
 
         assertUpdate("CREATE TABLE test_create_table_like_copy0 (LIKE test_create_table_like_original, col2 INTEGER)");
@@ -2080,10 +2088,11 @@ public abstract class BaseIcebergConnectorTest
                 """
                 WITH (
                    format = '%s',
-                   format_version = 2,
+                   format_version = %s,
                    location = '%s'
                 )""",
                 format,
+                formatVersion,
                 getTableLocation("test_create_table_like_copy1")));
 
         assertUpdate("CREATE TABLE test_create_table_like_copy2 (LIKE test_create_table_like_original EXCLUDING PROPERTIES)");
@@ -2091,10 +2100,11 @@ public abstract class BaseIcebergConnectorTest
                 """
                 WITH (
                    format = '%s',
-                   format_version = 2,
+                   format_version = %s,
                    location = '%s'
                 )""",
                 format,
+                formatVersion,
                 getTableLocation("test_create_table_like_copy2")));
         assertUpdate("DROP TABLE test_create_table_like_copy2");
 
@@ -9296,6 +9306,41 @@ public abstract class BaseIcebergConnectorTest
                 },
                 _ -> {});
         assertUpdate("DROP MATERIALIZED VIEW " + materializedViewName);
+    }
+
+    @Test
+    void testRowLineageWithMaterializedViews()
+    {
+        if (formatVersion < 3) {
+            return;
+        }
+
+        try (TestTable table = newTrinoTable("test_materialized_views", "(id int, name varchar) WITH (format_version = 3)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, 'Alice'), (2, 'Bob')", 2);
+
+            String materializedViewName = "test_materialized_view_" + randomNameSuffix();
+            assertUpdate("CREATE MATERIALIZED VIEW " + materializedViewName + " AS SELECT id, name, \"$row_id\", \"$last_updated_sequence_number\" FROM " + table.getName());
+
+            assertUpdate("REFRESH MATERIALIZED VIEW " + materializedViewName, 2);
+
+            assertThat(query("SELECT id, name, \"$row_id\", \"$last_updated_sequence_number\" FROM " + materializedViewName + " ORDER BY id"))
+                    .matches("""
+                            VALUES (1, VARCHAR 'Alice', BIGINT '0', BIGINT '2'),
+                                   (2, VARCHAR 'Bob', BIGINT '1', BIGINT '2')
+                            """);
+
+            assertUpdate("UPDATE " + table.getName() + " SET name = 'Alice Updated' WHERE id = 1", 1);
+
+            assertUpdate("REFRESH MATERIALIZED VIEW " + materializedViewName, 2);
+
+            assertThat(query("SELECT id, name, \"$row_id\", \"$last_updated_sequence_number\" FROM " + materializedViewName + " ORDER BY id"))
+                    .matches("""
+                            VALUES (1, VARCHAR 'Alice Updated', BIGINT '0', BIGINT '3'),
+                                   (2, VARCHAR 'Bob', BIGINT '1', BIGINT '2')
+                            """);
+
+            assertUpdate("DROP MATERIALIZED VIEW " + materializedViewName);
+        }
     }
 
     private static TableFinishInfo getTableFinishInfo(QueryStats queryStats)
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergParquetConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergParquetConnectorTest.java
@@ -39,12 +39,12 @@ import static io.trino.testing.QueryAssertions.assertEqualsIgnoreOrder;
 import static java.time.ZoneOffset.UTC;
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class TestIcebergParquetConnectorTest
+public abstract class BaseIcebergParquetConnectorTest
         extends BaseIcebergConnectorTest
 {
-    public TestIcebergParquetConnectorTest()
+    public BaseIcebergParquetConnectorTest(int formatVersion)
     {
-        super(PARQUET);
+        super(PARQUET, formatVersion);
     }
 
     @Override
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
@@ -107,6 +107,7 @@ public class TestIcebergV3
         queryRunner.installPlugin(new TestingIcebergPlugin(dataDirectory));
         queryRunner.createCatalog(ICEBERG_CATALOG, "iceberg", ImmutableMap.of(
                 "iceberg.catalog.type", "TESTING_FILE_METASTORE",
+                "iceberg.format-version", "3",
                 "iceberg.register-table-procedure.enabled", "true",
                 "iceberg.add-files-procedure.enabled", "true",
                 "iceberg.hive-catalog-name", "hive",
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergAvroConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergAvroConnectorTest.java
@@ -18,12 +18,12 @@ import org.junit.jupiter.api.Test;
 import static io.trino.plugin.iceberg.IcebergFileFormat.AVRO;
 import static org.junit.jupiter.api.Assumptions.abort;
 
-public class TestIcebergAvroConnectorTest
+public abstract class BaseIcebergAvroConnectorTest
         extends BaseIcebergConnectorTest
 {
-    public TestIcebergAvroConnectorTest()
+    public BaseIcebergAvroConnectorTest(int formatVersion)
     {
-        super(AVRO);
+        super(AVRO, formatVersion);
     }
 
     @Override
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergMinioOrcConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergMinioOrcConnectorTest.java
@@ -43,14 +43,14 @@ import static org.junit.jupiter.api.parallel.ExecutionMode.SAME_THREAD;
  * Iceberg connector test ORC and with S3-compatible storage (but without real metastore).
  */
 @Execution(SAME_THREAD)
-public class TestIcebergMinioOrcConnectorTest
+public abstract class BaseIcebergMinioOrcConnectorTest
         extends BaseIcebergConnectorTest
 {
     private final String bucketName = "test-iceberg-orc-" + randomNameSuffix();
 
-    public TestIcebergMinioOrcConnectorTest()
+    public BaseIcebergMinioOrcConnectorTest(int formatVersion)
     {
-        super(ORC);
+        super(ORC, formatVersion);
     }
 
     @Override
@@ -65,6 +65,7 @@ public class TestIcebergMinioOrcConnectorTest
                 .setIcebergProperties(
                         ImmutableMap.<String, String>builder()
                                 .put("iceberg.file-format", format.name())
+                                .put("iceberg.format-version", String.valueOf(formatVersion))
                                 .put("fs.hadoop.enabled", "true")
                                 .put("fs.native-s3.enabled", "true")
                                 .put("s3.aws-access-key", MINIO_ROOT_USER)
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2AvroConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2AvroConnectorTest.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+class TestIcebergV2AvroConnectorTest
+        extends BaseIcebergAvroConnectorTest
+{
+    TestIcebergV2AvroConnectorTest()
+    {
+        super(2);
+    }
+}
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2MinioOrcConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2MinioOrcConnectorTest.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+class TestIcebergV2MinioOrcConnectorTest
+        extends BaseIcebergMinioOrcConnectorTest
+{
+    TestIcebergV2MinioOrcConnectorTest()
+    {
+        super(2);
+    }
+}
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2ParquetConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2ParquetConnectorTest.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+class TestIcebergV2ParquetConnectorTest
+        extends BaseIcebergParquetConnectorTest
+{
+    TestIcebergV2ParquetConnectorTest()
+    {
+        super(2);
+    }
+}
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3AvroConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3AvroConnectorTest.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+class TestIcebergV3AvroConnectorTest
+        extends BaseIcebergAvroConnectorTest
+{
+    TestIcebergV3AvroConnectorTest()
+    {
+        super(3);
+    }
+}
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3MinioOrcConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3MinioOrcConnectorTest.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+class TestIcebergV3MinioOrcConnectorTest
+        extends BaseIcebergMinioOrcConnectorTest
+{
+    TestIcebergV3MinioOrcConnectorTest()
+    {
+        super(3);
+    }
+}
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3ParquetConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3ParquetConnectorTest.java
@@ -1,0 +1,23 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg;
+
+class TestIcebergV3ParquetConnectorTest
+        extends BaseIcebergParquetConnectorTest
+{
+    TestIcebergV3ParquetConnectorTest()
+    {
+        super(3);
+    }
+}
--- a/testing/trino-faulttolerant-tests/src/test/java/io/trino/faulttolerant/iceberg/TestIcebergParquetFaultTolerantExecutionConnectorTest.java
+++ b/testing/trino-faulttolerant-tests/src/test/java/io/trino/faulttolerant/iceberg/TestIcebergParquetFaultTolerantExecutionConnectorTest.java
@@ -15,8 +15,8 @@ package io.trino.faulttolerant.iceberg;
 
 import io.trino.filesystem.Location;
 import io.trino.plugin.exchange.filesystem.containers.MinioStorage;
+import io.trino.plugin.iceberg.BaseIcebergParquetConnectorTest;
 import io.trino.plugin.iceberg.IcebergQueryRunner;
-import io.trino.plugin.iceberg.TestIcebergParquetConnectorTest;
 import org.junit.jupiter.api.AfterAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
@@ -33,9 +33,14 @@ import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 @Isolated
 @TestInstance(PER_CLASS)
 public class TestIcebergParquetFaultTolerantExecutionConnectorTest
-        extends TestIcebergParquetConnectorTest
+        extends BaseIcebergParquetConnectorTest
 {
     private MinioStorage minioStorage;
+
+    public TestIcebergParquetFaultTolerantExecutionConnectorTest()
+    {
+        super(2);
+    }
 
     @Override
     protected IcebergQueryRunner.Builder createQueryRunnerBuilder()

### CODE REVIEW

**Comment by chenjian2664** [comment]
Is the second commit "[add support for Iceberg v3 deletion vectors](https://github.com/trinodb/trino/pull/27836/commits/d6f077e0a251d666e157a3c2f874e8446146b584)" added intentionally?

**Comment by dain** [comment]
> Is the second commit "[add support for Iceberg v3 deletion vectors](https://github.com/trinodb/trino/pull/27836/commits/d6f077e0a251d666e157a3c2f874e8446146b584)" added intentionally?

Yep.  The first two commits are the base from another PR.  Row lineage needs deletions to work to fully test that rowid works for update commands.

File: plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java [hunk 1]
```diff
@@ -1036,6 +1040,8 @@ public Map<String, ColumnHandle> getColumnHandles(ConnectorSession session, Conn
         for (IcebergColumnHandle columnHandle : getTopLevelColumns(SchemaParser.fromJson(table.getTableSchemaJson()), typeManager)) {
             columnHandles.put(columnHandle.getName(), columnHandle);
         }
+        columnHandles.put(ROW_ID.getColumnName(), rowIdColumnHandle());
```
chenjian2664: I think we should first check the table format version here. For format versions 1 and 2, users should not be able to query these two columns. The concept of row lineage is only meaningful starting from format version 3

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -664,14 +665,128 @@ void testV3InsertProducesRowLineageMetadata()
...
+                        "  count_if(\"$row_id\" IS NULL) AS row_id_nulls, " +
+                        "  count_if(\"$last_updated_sequence_number\" IS NULL) AS lusn_nulls, " +
+                        "  count(DISTINCT \"$row_id\") AS distinct_row_id, " +
+                        "  count(DISTINCT \"$last_updated_sequence_number\") AS distinct_lusn_values_in_snapshot " +
+                        "FROM " + tableName + " FOR VERSION AS OF " + v2SnapshotId))
+                .matches("VALUES (BIGINT '3', BIGINT '3', BIGINT '0', BIGINT '0', BIGINT '1')");
+
+        // Upgrade the table metadata to v3. This does not rewrite old snapshots, so existing rows still have NULL $row_id.
+        assertUpdate("ALTER TABLE " + tableName + " SET PROPERTIES format_version = 3");
+
```
chenjian2664, line 977: nit: add a assert for the table version is equals to 3 now

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -647,6 +647,7 @@ void testV3InsertProducesRowLineageMetadata()
         assertUpdate("CREATE TABLE " + tableName + " (id INTEGER, v VARCHAR) WITH (format = 'PARQUET', format_version = 3)");
```
chenjian2664: ideally, we should also test AVRO and ORC format

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -664,14 +665,128 @@ void testV3InsertProducesRowLineageMetadata()
...
             assertThat(file.dataSequenceNumber()).as("data file dataSequenceNumber must be set").isNotNull();
         }
 
         assertThat(fileCount).isGreaterThan(0);
         assertThat(totalRecords).isEqualTo(3);
 
+        assertThat(query("SELECT \"$row_id\" FROM " + tableName))
+                .matches("VALUES (BIGINT '0'), (BIGINT '1'), (BIGINT '2')");
+
+        assertQuery("SELECT count(DISTINCT \"$last_updated_sequence_number\") FROM " + tableName, "VALUES 1");
```
chenjian2664: If `$last_updated_sequence_number` is entirely `NULL`, the test would still pass

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -664,14 +665,128 @@ void testV3InsertProducesRowLineageMetadata()
...
+        // Value updated.
+        assertThat(query("SELECT v FROM " + tableName + " WHERE id = 2"))
+                .matches("VALUES (VARCHAR 'bb')");
+
+        // Current expected behavior: UPDATE does not preserve $row_id (it behaves like delete+insert with a new row id).
+        assertThat(query(
+                "SELECT " +
+                        "  (SELECT \"$row_id\" FROM " + tableName + " WHERE id = 2) = " +
+                        "  (SELECT \"$row_id\" FROM " + tableName + " FOR VERSION AS OF " + beforeUpdateSnapshotId + " WHERE id = 2)"))
+                .matches("VALUES (BOOLEAN 'false')");
```
chenjian2664: Except for the update row, we should also verify the remained rows' `$row_id` still keep the same as before

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -988,15 +1025,24 @@ else if (column.isFileModifiedTimeColumn()) {
...
                 }
                 else if (column.isRowPositionColumn()) {
                     appendRowNumberColumn = true;
                     transforms.transform(new GetRowPositionFromSource());
                 }
                 else if (!parquetIdToFieldName.containsKey(column.getBaseColumn().getId())) {
-                    Object initialDefault = getInitialDefault(tableSchema, column.getBaseColumn().getId());
-                    transforms.constantValue(nativeValueToBlock(column.getType(), initialDefault));
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
+                        transforms.constantValue(nativeValueToBlock(column.getType(), dataSequenceNumber));
```
chenjian2664: verify the the format version support last updated sequence number

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -988,15 +1025,24 @@ else if (column.isFileModifiedTimeColumn()) {
...
+                    }
+                    else if (column.isRowIdColumn() && fileFirstRowId.isPresent()) {
+                        appendRowNumberColumn = true;
```
chenjian2664, line 1099: verify the the format version support row id

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1021,7 +1067,14 @@ else if (!parquetIdToFieldName.containsKey(column.getBaseColumn().getId())) {
 
                         parquetColumnFieldsBuilder.add(new Column(parquetFieldName, field.get()));
                     }
-                    if (column.isBaseColumn()) {
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
+                        transforms.transform(new DataSequenceNumberTransform(dataSequenceNumber, ordinal));
+                    }
+                    else if (column.isRowIdColumn() && fileFirstRowId.isPresent()) {
+                        appendRowNumberColumn = true;
+                        transforms.transform(new RowIdTransform(fileFirstRowId.get(), ordinal));
```
chenjian2664: ditto

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1144,7 +1197,9 @@ private static ReaderPageSourceWithRowPositions createAvroPageSource(
             Optional<NameMapping> nameMapping,
             String partition,
             List<IcebergColumnHandle> columns,
-            Map<Integer, Optional<String>> partitionKeys)
+            Map<Integer, Optional<String>> partitionKeys,
+            long dataSequenceNumber,
+            Optional<Long> fileFirstRowId)
```
chenjian2664: nit: OptionalLong

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -664,14 +665,128 @@ void testV3InsertProducesRowLineageMetadata()
...
+        assertThat(query("SELECT \"$row_id\" FROM " + tableName))
+                .matches("VALUES (BIGINT '0'), (BIGINT '1'), (BIGINT '2')");
+
+        assertQuery("SELECT count(DISTINCT \"$last_updated_sequence_number\") FROM " + tableName, "VALUES 1");
+
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
+    @Test
+    void testRowLineageUpgradeBehavior()
```
chenjian2664, line 920: Consider adding row linage tests for:
1. time travel, latter when supporting writes we should verify the behavior changes as well
2. dropping `$row_id` and `$last_updated_sequence_number` - should fails in all the version

chenjian2664 (commented):
Reviewed: "Add support for reading $row_id and $last_updated_sequence_number"

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
```diff
@@ -987,21 +988,33 @@ private RelationPlan createTableExecutePlan(Analysis analysis, TableExecute stat
...
         TableWriterNode.TableExecuteTarget tableExecuteTarget = new TableWriterNode.TableExecuteTarget(
                 executeHandle,
                 Optional.empty(),
                 tableName.asSchemaTableName(),
                 metadata.getInsertWriterScalingOptions(session, tableHandle));
 
-        Optional<TableLayout> layout = metadata.getLayoutForTableExecute(session, executeHandle);
-
-        List<Symbol> symbols = visibleFields(tableScanPlan);
+        Set<String> expectedColumnNames = metadata.getColumnNamesForTableExecute(session, executeHandle);
```
chenjian2664: What about returns  `Optional<List<ColumnHandle>>` ?

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -118,6 +118,8 @@ Optional<TableExecuteHandle> getTableHandleForExecute(
             String procedureName,
             Map<String, Object> executeProperties);
 
+    Set<String> getColumnNamesForTableExecute(Session session, TableExecuteHandle tableExecuteHandle);
```
chenjian2664: Involve @Praveen2112 @ebyhr 
Please help to review this [commit](https://github.com/trinodb/trino/pull/27836/changes/03b3b8f12a9c57c8b49cd12f1506f871e4d87c21)

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergParquetConnectorTest.java
```diff
@@ -39,12 +39,12 @@
 import static java.time.ZoneOffset.UTC;
 import static org.assertj.core.api.Assertions.assertThat;
 
-public class TestIcebergParquetConnectorTest
+public abstract class BaseIcebergParquetConnectorTest
         extends BaseIcebergConnectorTest
 {
-    public TestIcebergParquetConnectorTest()
+    public BaseIcebergParquetConnectorTest(int formatVersion)
```
raphaelsolarski, line 45: `formatVersion` is ignored here and `TestIcebergParquetFaultTolerantExecutionConnectorTest`, `TestIcebergV2ParquetConnectorTest` and `TestIcebergV3ParquetConnectorTest` pass various versions. Is it by design?

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
dain: This API is required and null is not an allowed response.  The connector metadata has a default implementation so this is backwards compatible.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
dain: I don't think this is necessary.  These columns are synthetic and handled by the engine directly. I don't think file format has any impact on this feature.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1055,6 +1059,11 @@ public Map<String, ColumnHandle> getColumnHandles(ConnectorSession session, Conn
         for (IcebergColumnHandle columnHandle : getTopLevelColumns(SchemaParser.fromJson(table.getTableSchemaJson()), typeManager)) {
             columnHandles.put(columnHandle.getName(), columnHandle);
         }
+        if (table.getFormatVersion() >= 3) {
+            // It is critical that ROW_ID comes before LAST_UPDATED_SEQUENCE_NUMBER, because the optimize writer expects this ordering
```
electrum: "optimize writer" -> "optimize command"

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -720,8 +739,17 @@ else if (column.isRowPositionColumn()) {
                     transforms.transform(new GetRowPositionFromSource());
                 }
                 else if (!fileColumnsByIcebergId.containsKey(column.getBaseColumnIdentity().getId())) {
-                    Object initialDefault = getInitialDefault(tableSchema, column.getBaseColumnIdentity().getId());
-                    transforms.constantValue(nativeValueToBlock(column.getType(), initialDefault));
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
```
electrum: These should probably be `else if` on the outer level, after the other `if (column.isXxx)`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -720,8 +739,17 @@ else if (column.isRowPositionColumn()) {
...
+                        transforms.constantValue(nativeValueToBlock(column.getType(), dataSequenceNumber));
+                    }
+                    else if (column.isRowIdColumn() && fileFirstRowId.isPresent()) {
```
electrum: Order these checks so that row ID comes before the sequence number, the same as the order of the columns in the same

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -739,7 +767,14 @@ else if (!fileColumnsByIcebergId.containsKey(column.getBaseColumnIdentity().getI
                                 projectionsByFieldId.get(baseColumn.getId())));
                     }
 
-                    if (column.isBaseColumn()) {
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
```
electrum: Order these checks so that row ID comes before the sequence number

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -997,8 +1034,17 @@ else if (column.isRowPositionColumn()) {
                     transforms.transform(new GetRowPositionFromSource());
                 }
                 else if (!parquetIdToFieldName.containsKey(column.getBaseColumn().getId())) {
-                    Object initialDefault = getInitialDefault(tableSchema, column.getBaseColumn().getId());
-                    transforms.constantValue(nativeValueToBlock(column.getType(), initialDefault));
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
```
electrum: Order these checks so that row ID comes before the sequence number

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1023,7 +1069,14 @@ else if (!parquetIdToFieldName.containsKey(column.getBaseColumn().getId())) {
 
                         parquetColumnFieldsBuilder.add(new Column(parquetFieldName, field.get()));
                     }
-                    if (column.isBaseColumn()) {
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
```
electrum: Order these checks so that row ID comes before the sequence number

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1203,8 +1258,17 @@ else if (column.isRowPositionColumn()) {
                     transforms.transform(new GetRowPositionFromSource());
                 }
                 else if (!fileColumnsByIcebergId.containsKey(column.getBaseColumn().getId())) {
-                    Object initialDefault = getInitialDefault(fileSchema, column.getBaseColumn().getId());
-                    transforms.constantValue(nativeValueToBlock(column.getType(), initialDefault));
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
```
electrum: Order these checks so that row ID comes before the sequence number

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1218,7 +1282,14 @@ else if (!fileColumnsByIcebergId.containsKey(column.getBaseColumn().getId())) {
                         columnTypes.add(baseColumn.getType());
                     }
 
-                    if (column.isBaseColumn()) {
+                    if (column.isLastUpdatedSequenceNumberColumn()) {
```
electrum: Order these checks so that row ID comes before the sequence number

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -625,14 +626,141 @@ void testV3InsertProducesRowLineageMetadata()
             fileCount++;
             totalRecords += file.recordCount();
 
-            // These are the lineage “inputs” Iceberg uses to materialize _row_id and last_updated_sequence_number
```
electrum, line 629: Why remove the comment?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
electrum: Why would it pass? The count would be zero in that case as count ignores nulls

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -139,6 +140,14 @@ default Optional<ConnectorTableExecuteHandle> getTableHandleForExecute(
         throw new TrinoException(NOT_SUPPORTED, "This connector does not support table procedures");
     }
 
+    default Set<String> getColumnNamesForTableExecute(ConnectorSession connectorSession, ConnectorTableHandle tableHandle, ConnectorTableExecuteHandle connectorTableExecuteHandle)
+    {
+        return getTableMetadata(connectorSession, tableHandle).getColumns().stream()
+                .filter(column -> !column.isHidden())
```
electrum: Can use method reference `not(ColumnMetadata::isHidden)`

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -139,6 +140,14 @@ default Optional<ConnectorTableExecuteHandle> getTableHandleForExecute(
...
+                .map(ColumnMetadata::getName)
+                .collect(Collectors.toSet());
```
electrum: Static import

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -139,6 +140,14 @@ default Optional<ConnectorTableExecuteHandle> getTableHandleForExecute(
         throw new TrinoException(NOT_SUPPORTED, "This connector does not support table procedures");
     }
 
+    default Set<String> getColumnNamesForTableExecute(ConnectorSession connectorSession, ConnectorTableHandle tableHandle, ConnectorTableExecuteHandle connectorTableExecuteHandle)
```
electrum: Add Javadoc explaining the purpose. I assume that the connector should return the columns that the specific execute invocation requires.

electrum: Can this return column handles instead?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1908,6 +1927,46 @@ private static Object requireProcedureArgument(Map<String, Object> properties, S
         return value;
     }
 
+    @Override
+    public Set<String> getColumnNamesForTableExecute(ConnectorSession session, ConnectorTableHandle tableHandle, ConnectorTableExecuteHandle tableExecuteHandle)
+    {
+        IcebergTableExecuteHandle executeHandle = (IcebergTableExecuteHandle) tableExecuteHandle;
+        IcebergTableHandle icebergTableHandle = (IcebergTableHandle) tableHandle;
+        switch (executeHandle.procedureId()) {
```
electrum: If this is a switch expression, we don't need the throw

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -232,19 +232,78 @@ void testMergeV3Table()
...
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a'), (2, 'b'), (3, 'c')", 3);
+        assertUpdate("INSERT INTO " + tableName + " VALUES (4, 'd'), (5, 'e')", 2);
+
+        BaseTable table = loadTable(tableName);
+        table.refresh();
+        long beforeOptimizeSnapshotId = table.currentSnapshot().snapshotId();
+
+        // Before: row_id materialized (no nulls), unique, and we have > 1 data file.
+        assertThat(query(
+                "SELECT " +
```
electrum: Can these be text blocks with `.formatted()`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -232,19 +232,78 @@ void testMergeV3Table()
...
+        // row_id still unique
+        // row lineage preserved ($row_id and $last_updated_sequence_number match before/after)
+        assertThat(query(
+                "SELECT " +
+                        "  count(*) AS rows, " +
+                        "  count(DISTINCT a.\"$row_id\") AS distinct_row_id, " +
+                        "  count(DISTINCT a.\"$last_updated_sequence_number\") AS distinct_last_updated_sequence_number, " +
+                        "  count_if(b.\"$path\" <> a.\"$path\") AS moved, " +
+                        "  count_if(b.\"$row_id\" = a.\"$row_id\" AND b.\"$last_updated_sequence_number\" = a.\"$last_updated_sequence_number\") AS lineage_ok " +
+                        "FROM " + tableName + " FOR VERSION AS OF " + beforeOptimizeSnapshotId + " b " +
```
electrum: The version trick is cool, but to be safe, we should have another test that selects the IDs and sequence numbers before running optimize, then validates against the values afterwards.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -277,22 +280,28 @@ protected boolean hasBehavior(TestingConnectorBehavior connectorBehavior)
...
 
-    @Override
     @Test
+    @Override
     public void testCreateTableWithDefaultColumn()
     {
-        String tableName = "test_default_value_" + randomNameSuffix();
-        assertThatThrownBy(() -> assertUpdate("CREATE TABLE " + tableName + " (x int DEFAULT 1)"))
-                .hasMessageContaining("Default column values are not supported for Iceberg table format version < 3");
+        skipTestUnless(hasBehavior(TestingConnectorBehavior.SUPPORTS_CREATE_TABLE) && hasBehavior(TestingConnectorBehavior.SUPPORTS_INSERT));
```
electrum: This `skipTestUnless` isn't needed as the feature set for Iceberg is fixed.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -5463,7 +5475,7 @@ public void testOptimize()
                     .hasSizeGreaterThan(workerCount);
 
             // For optimize we need to set task_min_writer_count to 1, otherwise it will create more than one file.
-            computeActual(withSingleWriterPerTask(getSession()), "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+                computeActual(withSingleWriterPerTask(getSession()), "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
```
electrum: Revert this formatting change

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMergeSink.java
```diff
@@ -227,6 +229,134 @@ private Optional<PartitionData> createPartitionData(PartitionSpec partitionSpec,
...
 
+    private record MergePage(Optional<Page> removals, Optional<Page> additions)
+    {
+        MergePage
+        {
+            requireNonNull(removals, "removals is null");
+            requireNonNull(additions, "additions is null");
+        }
+
+        static MergePage createMergePage(Page inputPage, int dataColumnCount, boolean includeRowId)
```
electrum: Could we reuse `MergePage` in the SPI and simply append the row ID block to the insertions page for v3?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -213,9 +213,28 @@ void testUpdateV3Table()
...
+            BaseTable baseTable = loadTable(table.getName());
+            baseTable.refresh();
+            long beforeUpdateSnapshotId = baseTable.currentSnapshot().snapshotId();
+
             assertUpdate("UPDATE " + table.getName() + " SET v = 'bb' WHERE id = 2", 1);
             assertThat(query("SELECT * FROM " + table.getName()))
                     .matches("VALUES (INTEGER '1', VARCHAR 'a'), (INTEGER '2', VARCHAR 'bb')");
+
+            // Verify row_id is preserved for the updated row
+            assertThat(query(
```
electrum, line 251: It would be better to select it before the update, rather than doing a version read. We could do the version read in addition to the check against the saved row ID values.

Same for the other tests.

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
chenjian2664: @Praveen2112 Please take a look

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
chenjian2664: ah, you are right, the count ignores nulls. 
I think it would be better to verify the value directly, as we don't know if the value may null

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
chenjian2664: where is the test case exercise the `IcebergPageSourceProvider` changes (for AVRO and ORC)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMergeSink.java
```diff
@@ -227,6 +229,134 @@ private Optional<PartitionData> createPartitionData(PartitionSpec partitionSpec,
...
+                        rowIdNulls[additionIndex] = true;
+                    }
+                    additionIndex++;
+                    if (additionIndex < additionCount) {
+                        nextAdditionPosition = additionPositions[additionIndex];
+                    }
+                }
+            }
+
+            Block rowIdBlock = new LongArrayBlock(additionCount, Optional.of(rowIdNulls), rowIdValues);
```
chenjian2664: inline

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMergeSink.java
```diff
@@ -227,6 +229,134 @@ private Optional<PartitionData> createPartitionData(PartitionSpec partitionSpec,
...
+                        // Pure INSERT or UPDATE with null source row ID
+                        rowIdNulls[additionIndex] = true;
+                    }
+                    additionIndex++;
+                    if (additionIndex < additionCount) {
+                        nextAdditionPosition = additionPositions[additionIndex];
+                    }
+                }
+            }
+
```
chenjian2664: add a check to verify that the `pendingSourceRowId` is null - that we don't have unhandled row ids

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMergeSink.java
```diff
@@ -227,6 +229,134 @@ private Optional<PartitionData> createPartitionData(PartitionSpec partitionSpec,
...
+            return new MergePage(removalsPage, additionsPage);
+        }
+
+        private static Block createRowIdBlock(Page inputPage, int dataColumnCount, int[] additionPositions, int additionCount)
+        {
+            // For V3, we need to extract source_row_id from the merge row ID for UPDATE_INSERT rows.
+            // UPDATE_DELETE is immediately followed by UPDATE_INSERT, so we track pending source row IDs.
```
chenjian2664: `UPDATE_DELETE is immediately followed by UPDATE_INSERT,` where is the logic that guarantee it?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -313,6 +317,7 @@ public ConnectorPageSource createPageSource(
             IcebergFileFormat fileFormat,
             Map<String, String> fileIoProperties,
             long dataSequenceNumber,
+            OptionalLong fileFirstRowId,
```
ebyhr, line 324: There is a logical conflict in TestIcebergPageSourceProvider. Please rebase on master.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -196,13 +196,15 @@ public abstract class BaseIcebergConnectorTest
     private static final Pattern WITH_CLAUSE_EXTRACTOR = Pattern.compile(".*(WITH\\s*\\([^)]*\\))\\s*$", Pattern.DOTALL);
 
     protected final IcebergFileFormat format;
+    protected final int formatVersion;
 
     protected TrinoFileSystem fileSystem;
     protected TimeUnit storageTimePrecision;
 
-    protected BaseIcebergConnectorTest(IcebergFileFormat format)
+    protected BaseIcebergConnectorTest(IcebergFileFormat format, int formatVersion)
```
ebyhr, line 204: Please test SHOW STATS with the new metadata columns. I believe it returns incorrect results.

ebyhr: Also, add MV test case: 
```java
    @Test
    void testRowLineageWithMaterializedViews()
    {
        try (TestTable table = newTrinoTable("test_materialized_views", "(id int, name varchar) WITH (format_version = 3)")) {
            assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, 'Alice'), (2, 'Bob')", 2);

            String materializedViewName = "test_materialized_view_" + randomNameSuffix();
            assertUpdate("CREATE MATERIALIZED VIEW " + materializedViewName + " AS SELECT id, name, \"$row_id\", \"$last_updated_sequence_number\" FROM " + table.getName());

            assertUpdate("REFRESH MATERIALIZED VIEW " + materializedViewName, 2);

            assertThat(query("SELECT id, name, \"$row_id\", \"$last_updated_sequence_number\" FROM " + materializedViewName))
                    .matches("""
                            VALUES (1, VARCHAR 'Alice', BIGINT '0', BIGINT '2'),
                                   (2, 'Bob', BIGINT '1', BIGINT '2')
                            """);

            assertUpdate("UPDATE " + table.getName() + " SET name = 'Alice Updated' WHERE id = 1", 1);

            assertUpdate("REFRESH MATERIALIZED VIEW " + materializedViewName, 2);

            assertThat(query("SELECT id, name, \"$row_id\", \"$last_updated_sequence_number\" FROM " + materializedViewName))
                    .matches("""
                            VALUES (1, VARCHAR 'Alice Updated', BIGINT '0', BIGINT '3'),
                                   (2, 'Bob', BIGINT '1', BIGINT '2')
                            """);

            assertUpdate("DROP MATERIALIZED VIEW " + materializedViewName);
        }
    }
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
dain: I change the metadata so rowId and LUSN to only exist in v3.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
dain: I change the metadata so rowId and LUSN to only exist in v3.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
dain: I change the metadata so rowId and LUSN to only exist in v3.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
dain: I change the metadata so rowId and LUSN to only exist in v3.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
dain: I did this everywhere

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
dain: Actually, @electrum pointed out that we could be using ColumHandles here instead of names.  I was reacting to the optional part not the column handle part.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergParquetConnectorTest.java
dain: fixed

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
dain: Actually, this is correct.  This handles the case where the row id or LUSN is not present in file and we must synthesize them, otherwise we must read them from the file.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMergeSink.java
dain: I rewrote this with more explict handling verification. Generally, none of these update systems actually verify this stuff, but I am happy to add it.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMergeSink.java
dain: I rewrote this

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
dain: I added that comment after I pushed.  I just finally got through applying all of the comments below.  This is now covered.

dain:
@ebyhr I believe this is ready to go now

ebyhr:
@dain Thanks for addressing the comments. I'll review this PR again tomorrow or shortly later.

dain:
@ebyhr any updates?

ebyhr:
@dain Sorry, I had to work on a different issue yesterday. It looks like there’s still a bug with the Avro format.

```sql
CREATE TABLE test (name varchar) WITH (format_version = 3, format = 'AVRO');
INSERT INTO test VALUES 'alice', 'bob';
INSERT INTO test VALUES 'carol', 'david';
UPDATE test SET name = 'BOB' WHERE name = 'bob';
SELECT name, "$row_id", "$last_updated_sequence_number" FROM test;
```

```
 name  | $row_id | $last_updated_sequence_number
-------+---------+-------------------------------
 carol |       2 |                             3
 david |       3 |                             3
 alice |       0 |                             2
 BOB   |       4 |                             4
```

BOB should return 1 as $row_id.

ebyhr:
The partition table has a bug regardless of the file format: 
```sql
CREATE TABLE test (name varchar, x bigint) WITH (format_version = 3, partitioning = ARRAY['x']);
INSERT INTO test VALUES ('alice', 1), ('bob', 2);
INSERT INTO test VALUES ('carol', 1), ('david', 2);
UPDATE test SET name = 'BOB' WHERE name = 'bob';
SELECT name, "$row_id" FROM test;
```

```
 name  | $row_id
-------+---------
 BOB   |       4
 carol |       2
 david |       3
 alice |       1
(4 rows)
```

BOB should return 0 as $row_id.

dain:
@ebyhr sorry about that, I missed the avro case, and fixing that help find another issue.  This should be all working now and tested.

dain:
@ebyhr any update?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
ebyhr: @dain The above `testRowLineageWithMaterializedViews` fails if we add `"iceberg.format-version", "3"` to `TestIcebergV3.createQueryRunner`. Could you check and fix the failure?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -621,18 +830,216 @@ void testV3InsertProducesRowLineageMetadata()
...
+                .matches("VALUES (BIGINT '0'), (BIGINT '1'), (BIGINT '2')");
+
+        assertQuery(
+                "SELECT id, \"$last_updated_sequence_number\" FROM " + tableName + " ORDER BY id",
+                "VALUES (1, %d), (2, %d), (3, %d)".formatted(
+                        expectedLastUpdatedSequenceNumber,
+                        expectedLastUpdatedSequenceNumber,
+                        expectedLastUpdatedSequenceNumber));
+
+        MaterializedResult showStats = computeActual("SHOW STATS FOR (SELECT \"$row_id\", \"$last_updated_sequence_number\" FROM " + tableName + ")");
```
ebyhr, line 870: Please add another assertion using WHERE clause with $row_id, $last_updated_sequence_number condition.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -621,18 +830,216 @@ void testV3InsertProducesRowLineageMetadata()
...
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
+    @Test
+    void testRowLineageUpgradeBehavior()
+    {
+        ALL_FILE_FORMATS.forEach(this::assertRowLineageUpgradeBehavior);
+    }
+
+    private void assertRowLineageUpgradeBehavior(String fileFormat)
```
ebyhr, line 954: There is a bug about upgraded behavior: 
```java
    @Test
    void testUpgradeTableToV3FromTrinoWithRowLineage()
    {
        String tableName = "test_upgrade_table_to_v3_from_trino_" + randomNameSuffix();
        assertUpdate("CREATE TABLE " + tableName + " WITH (format_version = 2) AS SELECT * FROM tpch.tiny.nation", 25);
        assertThat(formatVersion(loadTable(tableName))).isEqualTo(2);

        assertThat(query("SELECT \"$row_id\" FROM " + tableName)).failure()
                .hasMessage("line 1:8: Column '$row_id' cannot be resolved");
        assertThat(query("SELECT \"$last_updated_sequence_number\" FROM " + tableName)).failure()
                .hasMessage("line 1:8: Column '$last_updated_sequence_number' cannot be resolved");

        // v2 -> v3
        assertUpdate("ALTER TABLE " + tableName + " SET PROPERTIES format_version = 3");
        assertThat(formatVersion(loadTable(tableName))).isEqualTo(3);
        assertQuery("SELECT * FROM " + tableName, "SELECT * FROM nation");

        // data sequence number from file is 1, first_row_id is null
        assertThat(query("SELECT \"$row_id\",\"$last_updated_sequence_number\", nationkey FROM " + tableName + " WHERE nationkey = 10"))
                .matches("VALUES ( CAST(NULL AS bigint), BIGINT '1', BIGINT '10')");
        assertUpdate("UPDATE " + tableName + " SET nationkey = 110 WHERE nationkey = 10", 1);
        assertThat(query("SELECT \"$last_updated_sequence_number\", nationkey FROM " + tableName + " WHERE nationkey = 110"))
                .matches("VALUES (BIGINT '2', BIGINT '110')");

        // sometimes first_row_id from file is 0, as freshly created table, sometimes it is 24, as next index of row_id in nation table
        assertThat(query("SELECT \"$row_id\" FROM " + tableName + " WHERE nationkey = 110")).result().onlyColumnAsSet()
                .containsAnyOf(0L, 24L);
    }
```

```
io.trino.testing.QueryFailedException: source_row_id must be present for UPDATE_INSERT

	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:591)
	at io.trino.testing.DistributedQueryRunner.executeWithPlan(DistributedQueryRunner.java:580)
	at io.trino.testing.QueryAssertions.assertDistributedUpdate(QueryAssertions.java:111)
	at io.trino.testing.QueryAssertions.assertUpdate(QueryAssertions.java:65)
	at io.trino.testing.QueryAssertions.assertUpdate(QueryAssertions.java:59)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:426)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:421)
	at io.trino.plugin.iceberg.TestIcebergV3.testUpgradeTableToV3FromTrinoWithRowLineage(TestIcebergV3.java:111)
	Suppressed: java.lang.Exception: SQL: UPDATE test_upgrade_table_to_v3_from_trino_n6069m43ye SET nationkey = 110 WHERE nationkey = 10
		at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:598)
		... 7 more
Caused by: com.google.common.base.VerifyException: source_row_id must be present for UPDATE_INSERT
	at com.google.common.base.Verify.verify(Verify.java:124)
	at io.trino.plugin.iceberg.IcebergMergeSink.createRowIdBlock(IcebergMergeSink.java:277)
	at io.trino.plugin.iceberg.IcebergMergeSink.createInsertionsPageWithRowId(IcebergMergeSink.java:246)
	at io.trino.plugin.iceberg.IcebergMergeSink.lambda$storeMergedRows$0(IcebergMergeSink.java:112)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at io.trino.plugin.iceberg.IcebergMergeSink.storeMergedRows(IcebergMergeSink.java:110)
	at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorMergeSink.storeMergedRows(ClassLoaderSafeConnectorMergeSink.java:44)
	at io.trino.operator.MergeWriterOperator.addInput(MergeWriterOperator.java:149)
	at io.trino.operator.Driver.processInternal(Driver.java:408)
	at io.trino.operator.Driver.lambda$process$0(Driver.java:306)
	at io.trino.operator.Driver.tryWithLock(Driver.java:709)
	at io.trino.operator.Driver.process(Driver.java:298)
	at io.trino.operator.Driver.processForDuration(Driver.java:269)
	at io.trino.execution.SqlTaskExecution$DriverSplitRunner.processFor(SqlTaskExecution.java:852)
	at io.trino.execution.executor.dedicated.SplitProcessor.run(SplitProcessor.java:77)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.lambda$run$0(TaskEntry.java:205)
	at io.trino.$gen.Trino_testversion____20260310_044849_71.run(Unknown Source)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.run(TaskEntry.java:206)
	at io.trino.execution.executor.scheduler.FairScheduler.runTask(FairScheduler.java:177)
	at io.trino.execution.executor.scheduler.FairScheduler.lambda$submit$0(FairScheduler.java:164)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
```

dain: I found the issue. The problem was in IcebergMergeSink: for v3 updates we need to preserve source_row_id when it is available, but for rows coming from pre-lineage v2 files after a v2 -> v3 upgrade it can legitimately be absent. The fix is to read source_row_id directly from each UPDATE_INSERT row when present, and otherwise let Iceberg assign a fresh row id to the new row.

We did not catch this before because our coverage did not include the immediate-update-after-upgrade case. I added a focused regression test for that scenario, and I also verified the broader v3 merge/update coverage that was failing in CI now passes with the corrected logic.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
ebyhr: @dain The `testRowLineageWithMaterializedViews` still fails with the latest code.

chenjian2664: Simply verify that the column handle is included in the columns, should enough:
```
        if (table.getFormatVersion() >= 3) {
            // It is critical that ROW_ID comes before LAST_UPDATED_SEQUENCE_NUMBER, because the optimize command expects this ordering
            if (!containsColumnHandle(topLevelColumns, ROW_ID.getColumnName())) {
                columnHandles.put(ROW_ID.getColumnName(), rowIdColumnHandle());
            }
            if (!containsColumnHandle(topLevelColumns, LAST_UPDATED_SEQUENCE_NUMBER.getColumnName())) {
                columnHandles.put(LAST_UPDATED_SEQUENCE_NUMBER.getColumnName(), lastUpdatedSequenceNumberColumnHandle());
            }
        }
```
cc @dain

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
ebyhr: Reminder. We could add the following code or something similar to `IcebergMetadata#getTableStatistics` since it's hard to provide the correct table statistics when these columns are specified in WHERE clause:
```java
        if (originalHandle.getProjectedColumns().contains(rowIdColumnHandle()) || originalHandle.getProjectedColumns().contains(lastUpdatedSequenceNumberColumnColumnHandle())) {
            log.warn("Statistics for $row_id and $last_updated_sequence_number columns are not supported");
            return TableStatistics.empty();
        }
```

PR #27847, opened 2026-01-05
Document limitations of `table_changes` in Iceberg connector

chenjian2664:
Added a section describing limitations of the `table_changes` table function,
 including lack of support for delete files in `merge-on-read` tables and the
 net changes without cross snapshots.

## Release notes

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -2136,6 +2136,18 @@ ORDER BY _change_ordinal ASC;
 (6 rows)
 ```
 
+##### Limitations
+
+* Tables with delete files are not supported. The `table_changes` table function does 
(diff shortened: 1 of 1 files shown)


findinpath (approved):

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -2136,6 +2136,18 @@ ORDER BY _change_ordinal ASC;
 (6 rows)
 ```
 
+##### Limitations
+
+* Tables with delete files are not supported. For `merge-on-read` Iceberg tables,
```
ebyhr: > `merge-on-read`

There is no such a property in OS Trino. I think we should avoid using MoR keyword here. It's not explained  in the current documentation.

chenjian2664: removed keyword

PR #27848, opened 2026-01-05
Refactorings on the JdbcClient methods

findinpath:
## Description

This is a follow-up PR generated by the discussions that happened after merging https://github.com/trinodb/trino/pull/27702

cc @findepi @chenjian2664 

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
+++ b/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
@@ -218,6 +218,7 @@ public class IgniteMetadata
         if (saveMode == REPLACE) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
+        // No rollback action is needed when creating an empty table
         igniteClient.beginCreateTable(session, tableMetadata, _ -> {});
     }
 


findepi:
> This is a follow-up PR generated by the discussions that happened after merging #27702

Does this PR take https://github.com/trinodb/trino/pull/27794 into account, or is it meant to obsolete it?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/jmx/StatisticsAwareJdbcClient.java
```diff
@@ -400,9 +400,9 @@ public void dropTable(ConnectorSession session, JdbcTableHandle jdbcTableHandle)
     }
 
     @Override
-    public void rollbackCreateTable(ConnectorSession session, JdbcOutputTableHandle handle)
+    public void rollbackTemporaryTableCreation(ConnectorSession session, JdbcOutputTableHandle handle)
     {
-        stats.getRollbackCreateTable().wrap(() -> delegate().rollbackCreateTable(session, handle));
+        stats.getRollbackCreateTable().wrap(() -> delegate().rollbackTemporaryTableCreation(session, handle));
```
chenjian2664: The stats needs to renamed as well, thoughts?

findinpath:
> Does this PR take https://github.com/trinodb/trino/pull/27794 into account, or is it meant to obsolete it?

@findepi my intention is to build on top of the above mentioned PR. I'm hoping that https://github.com/trinodb/trino/pull/27794 will land before this PR and I'll adjust this one afterwards.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/jmx/StatisticsAwareJdbcClient.java
findinpath: Does this need RN ?

findinpath: It feels a bit strange to have `JdbcClientStats` for rolling back temporary tables, but not for the destination tables.

Added `rollbackDestinationTableCreation` to `JdbcStats`

findinpath:
Rebased on `master` to fix the code conflicts generated by the changes in https://github.com/trinodb/trino/pull/27794

findinpath:
@chenjian2664 , @findepi  could you look over this PR to close on the discussions that emerged from  https://github.com/trinodb/trino/pull/27702 ?

plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteMetadata.java
```diff
@@ -220,6 +220,7 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
         if (saveMode == REPLACE) {
             throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
         }
+        // No rollback action is needed because the creation of the table is atomic in Ignite
```
chenjian2664: It is not "atomic", see other BaseJdbcClient#createTable, we don't(need) put the rollback collector at all.
Btw, do you remember what's the problem if we pass the rollback?
how about "No rollback action is needed in this method"
another option is that we could discard this commit

findinpath: > Btw, do you remember what's the problem if we pass the rollback?

`createTable` corresponds to `CREATE TABLE` statements
`beginCreateTable` & `finishCreateTable` correspond to `CREATE TABLE AS SELECT` statements

The problem when taking into account the rollback action is that we end up trying to `DROP` a table that does not exist.

https://github.com/trinodb/trino/blob/e6c62390f155e70e27d485a3db9dbf0d361cc837/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteClient.java#L427

findinpath:
@chenjian2664  AC 

FYI
I've rebased the code to ensure that it is compatible with the latest `master`.

Issue #27435, opened 2025-11-25
Flaky test TestWorkerRestart.testRestartDuringQuery


findepi:
https://github.com/trinodb/trino/actions/runs/19671056140/job/56340013722?pr=27434

```
Error:  Tests run: 3, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 43.88 s <<< FAILURE! -- in io.trino.tests.TestWorkerRestart
Error:  io.trino.tests.TestWorkerRestart.testRestartDuringQuery()[1] -- Time elapsed: 23.00 s <<< FAILURE!
java.lang.AssertionError: 

Expecting message:
  "Error fetching http://127.0.0.1:37995/v1/task/20251125_133205_00001_yd995.0.0.0/results/0/0: Expected response code to be 200, but was 503:
Trino server is still initializing
"
to be found for regex:
  "^Expected response code from \S+ to be 200, but was 500|Expected response from \S+ is empty|Error fetching \S+: Expected response code to be 200, but was 500|Could not communicate with the remote task. The node may have crashed or be under too much load|Error fetching \S+: Content-Type header is not set"
but did not.

Throwable that failed the check:

io.trino.testing.QueryFailedException: Error fetching http://127.0.0.1:37995/v1/task/20251125_133205_00001_yd995.0.0.0/results/0/0: Expected response code to be 200, but was 503:
Trino server is still initializing

	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:138)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:587)
	at io.trino.testing.DistributedQueryRunner.execute(DistributedQueryRunner.java:570)
	at io.trino.testing.QueryRunner.execute(QueryRunner.java:82)
	at io.trino.tests.TestWorkerRestart.lambda$testRestartDuringQuery$0(TestWorkerRestart.java:96)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
	at java.base/java.lang.Thread.run(Thread.java:1474)
[...]

Issue #26194, opened 2025-07-14
Flaky test TestHiveConnectorTest -- fail to start


chenjian2664:
https://github.com/trinodb/trino/actions/runs/16250597112/job/45903138827?pr=26184#step:5:2545

ebyhr:
@chenjian2664 Could you add the stacktraces to the issue description?

ebyhr:
`TestVerticaConnectorSmokeTest` also failed to start the query runner. https://github.com/trinodb/trino/actions/runs/16334740729/job/46144513233

@wendigo Do you have any ideas about the cause? e.g. recent library bump

findepi:
still relevant?

chenjian2664:
Not sure, not see it for a while

findepi:
🤞

piotrrzysko:
Similar stacktrace, but a different test: https://github.com/trinodb/trino/actions/runs/21060681766/job/60566957361?pr=27849:

```
tests               | 2026-01-16 14:39:43 INFO: FAILURE     /    io.trino.tests.product.TestSqlCancel.cancelCreateTable (Groups: ) took 5.1 seconds
tests               | 2026-01-16 14:39:43 SEVERE: Failure cause:
tests               | java.lang.NullPointerException: Cannot invoke "org.eclipse.jetty.util.Pool.acquire()" because "this.pool" is null
tests               | 	at org.eclipse.jetty.client.AbstractConnectionPool.activate(AbstractConnectionPool.java:322)
tests               | 	at org.eclipse.jetty.client.AbstractConnectionPool.acquire(AbstractConnectionPool.java:233)
tests               | 	at org.eclipse.jetty.client.transport.HttpDestination.process(HttpDestination.java:365)
tests               | 	at org.eclipse.jetty.client.transport.HttpDestination.send(HttpDestination.java:351)
tests               | 	at org.eclipse.jetty.client.transport.HttpDestination.send(HttpDestination.java:345)
tests               | 	at org.eclipse.jetty.client.transport.HttpDestination.send(HttpDestination.java:322)
tests               | 	at org.eclipse.jetty.client.transport.HttpDestination.send(HttpDestination.java:300)
tests               | 	at org.eclipse.jetty.client.transport.HttpRequest.sendAsync(HttpRequest.java:799)
tests               | 	at org.eclipse.jetty.client.transport.HttpDestination.send(HttpDestination.java:295)
[...]

PR #27849, opened 2026-01-05
Reduce resource group lock contention

piotrrzysko:
## Description

Please see the individual commit messages for the reasoning behind these changes.

### Before
image

### After
image

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
+++ b/core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
@@ -33,13 +33,14 @@ import org.weakref.jmx.Nested;
 
 import java.time.Duration;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.Executor;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.BiConsumer;
 
 import static com.google.common.base.MoreObjects.toStringHelper;
@@ -84,6 +85,7 @@ public class InternalResourceGroup
     private final ResourceGroupId id;
     private final BiConsumer<InternalResourceGroup, Boolean> jmxExportListener;
     private final Executor executor;
+    private final AtomicBoolean resourceUsageStagingInProgress;
 
     // Configuration
     // =============
@@ -122,12 +124,10 @@ public class InternalResourceGroup
     @GuardedBy("root")
     private Queue<InternalResourceGroup> eligibleSubGroups = new FifoQueue<>();
     // Sub groups whose memory usage may be out of date. Most likely because they have a running query.
-    @GuardedBy("root")
-    private final Set<InternalResourceGroup> dirtySubGroups = new HashSet<>();
+    private final Set<InternalResourceGroup> dirtySubGroups = ConcurrentHashMap.newKeySet();
     @GuardedBy("root")
     private UpdateablePriorityQueue<ManagedQueryExecution> queuedQueries = new FifoQueue<>();
-    @GuardedBy("root")
-    private final Map<ManagedQueryExecution, ResourceUsage> runningQueries = new HashMap<>();
+    private final Map<ManagedQueryExecution, StagedResourceUsage> runningQueries = new ConcurrentHashMap<>();
     @GuardedBy("root")
     private int descendantRunningQueries;
     @GuardedBy("root")
@@ -154,10 +154,12 @@ public class InternalResourceGroup
         if (parent.isPresent()) {
             id = new ResourceGroupId(parent.get().id, name);
             root = parent.get().root;
+            resourceUsageStagingInProgress = root.resourceUsageStagingInProgress;
         }
         else {
             id = new ResourceGroupId(name);
             root = this;
+            resourceUsageStagingInProgress = new AtomicBoolean();
         }
     }
 
@@ -743,7 +745,7 @@ public class InternalResourceGroup
     {
         checkState(Thread.holdsLock(root), "Must hold lock to start a query");
         synchronized (root) {
-            runningQueries.put(query, ZERO);
+            runningQueries.put(query, StagedResourceUsage.zero());
             InternalResourceGroup group = this;
             group.getStartedQueries().update(1);
             while (group.parent.isPresent()) {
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
```diff
@@ -639,9 +638,7 @@ public void setJmxExport(boolean export)
     @Override
     public boolean isDisabled()
     {
-        synchronized (root) {
```
findepi, line 642: Why is this removed but synchronization in `setDisabled` is not?

findepi (commented):
`Reduce contention in InternalResourceGroupManager`

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
```diff
@@ -1079,4 +1128,41 @@ public int hashCode()
...
+        }
+
+        public ResourceUsage getStaged()
+        {
+            return stagedResourceUsage;
+        }
+
+        public void promoteStagedToCurrent()
+        {
+            currentResourceUsage = stagedResourceUsage;
```
findepi: With both being volatile, we're relying on external synchronization for this to mean anything.
Rather than having `StagedResourceUsage` with volatile state (that's kinda hard to reason about), we could have `record StagedResourceUsage` that's _replaced_ instead of being mutated. Would that work?

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
```diff
@@ -33,13 +33,14 @@
 
 import java.time.Duration;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
```
findepi, line 41: > Previously, concurrent threads attempting to perform updates were
blocked, causing updates to be executed serially. Now, when multiple
threads attempt to update resource usage, only the first performs the
update; the remaining threads skip it and use existing statistics.
> 
> This slightly changes the previous behavior, as threads starting new or
queued queries may observe outdated statistics. In practice, this is
acceptable since resource usage statistics are inherently stale, and
there has never been a guarantee of observing fully up-to-date data.




- small query finishes
- big query finishes at about same time
- "only the first performs the update; the remaining threads skip it and use existing statistics"

Sounds like the effect might be that the stats remain stale and wrong for some time. Will they be fixed after 100ms?


> Resource usage for a resource group tree is updated in two cases:
* periodically (every 100 ms)
* on query completion
> 
> Previously, concurrent threads attempting to perform updates were
blocked, causing updates to be executed serially.

Sounds like every 10ms on a relatively busy cluster (and more often on a very busy one).
How long can this take?

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
```diff
@@ -760,15 +762,65 @@ private void startInBackground(ManagedQueryExecution query)
...
+        }
+
+        for (InternalResourceGroup subGroup : dirtySubGroups) {
+            subGroup.stageResourceUsage();
+        }
+    }
+
+    private boolean acquireResourceUsageStaging()
+    {
+        if (resourceUsageStagingInProgress.get()) {
```
lukasz-stec, line 817: Does the early return have a visible impact on performance, given that it is also done in `compareAndSet`?

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
```diff
@@ -786,23 +838,24 @@ public void triggerProcessQueuedQueries()
 
     private void queryFinished(ManagedQueryExecution query)
     {
+        // CPU and data scan are measured cumulatively (i.e. total CPU & physical input data used until this moment by the query).
+        // Memory is measured instantaneously (how much memory the query is using at this moment). At query completion, memory usage drops to zero.
+        ResourceUsage finalUsage = new ResourceUsage(
```
lukasz-stec, line 847: Why did you move it here? To reduce the size of the critical section?

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
lukasz-stec: > small query finishes
big query finishes at about same time
"only the first performs the update; the remaining threads skip it and use existing statistics"
Sounds like the effect might be that the stats remain stale and wrong for some time. Will they be fixed after 100ms?

As I understand it, every query that finishes updates the resource usage. This happens in the `synchronized` section of the `queryFinished` method. The only thing that won't happen is the update of running queries with the delta usage between when the small and big query triggered stats collection, which should not matter. @piotrrzysko correct?

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
```diff
@@ -831,8 +884,9 @@ private void queryFinished(ManagedQueryExecution query)
             }
 
             updateEligibility();
-            root.triggerProcessQueuedQueries();
         }
+
+        root.triggerProcessQueuedQueries();
```
lukasz-stec, line 897: nit: Do I understand correctly that this is the core of the change?. Could you add a comment that even though the current query already updated its usage, other queries may not update their usage if there is a concurrent update running, but we don't care, as the update is small anyway?

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
piotrrzysko: Mainly to keep the scope of this PR as narrow as possible. `setDisabled` does not need synchronization now, but it is not on the hot path I am targeting - namely, where queries are executed and the resource group configuration is not modified.

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
piotrrzysko: Yes. In general, calls for stats (here, `query.getTotalCpuTime()` and `query.getBasicQueryInfo()`) are relatively heavy, especially when many tasks are running. In this case there is only one query and it is finished, so it likely doesn’t matter, but it shouldn’t hurt either.

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
piotrrzysko: Good point. Fixed.

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
piotrrzysko: In a highly contended scenario, where the flag is already set to true, microbenchmarks show a significant difference:

```
Benchmark                                          Mode  Cnt     Score     Error   Units
AtomicBooleanCasBenchmark.alreadyTrue_checkFirst  thrpt   10  7252,937 ± 684,847  ops/us
AtomicBooleanCasBenchmark.alreadyTrue_directCas   thrpt   10   246,783 ±   0,564  ops/us
```

Whether this matters here is hard to tell. I might try to run some macrobenchmarks.

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
piotrrzysko: Yes, @lukasz-stec's explanation is correct.

core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
lukasz-stec: 7252,937 ops/us gives 0.13ns per operation, which is so small that it raises suspicion. Feel free to ignore the comment, though. The early return is fine by me.

piotrrzysko:
CI:
* https://github.com/trinodb/trino/issues/27435
* https://github.com/trinodb/trino/issues/26194
* https://github.com/trinodb/trino/issues/27913

Rebasing because the last was fixed recently.

Issue #27730, opened 2025-12-22
Buffering capacity exceeded exception


vmtuan12:
Hi,

I am running into issues with `Buffering capacity 33554432 exceeded` exceptions. This is the stack trace in the admin UI:

```
io.trino.operator.PageTransportTimeoutException: Encountered too many errors talking to a worker node. The node may have crashed or be under too much load. This is probably a transient issue, so please retry your query in a few minutes. (http://10.4.8.69:8080/v1/task/20251222_083217_00010_hfgnw.1.0.0/results/0/0 - 108 failures, failure duration 60.47s, total failed request time 60.66s)
	at io.trino.operator.HttpPageBufferClient$1.onFailure(HttpPageBufferClient.java:505)
	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1575)
Caused by: java.lang.IllegalArgumentException: Buffering capacity 33554432 exceeded
	at org.eclipse.jetty.client.BufferingResponseListener.onContent(BufferingResponseListener.java:116)
	at org.eclipse.jetty.client.Response$ContentListener.onContent(Response.java:162)
	at org.eclipse.jetty.client.Response$AsyncContentListener.onContentSource(Response.java:213)
	at org.eclipse.jetty.client.Response$AsyncContentListener.lambda$onContentSource$0(Response.java:191)
	at org.eclipse.jetty.util.thread.Invocable$ReadyTask.run(Invocable.java:175)
...
```

Taking a look at class `HttpPageBufferClient`, `DirectExchangeClient` and relevant classes, I have tried increasing `exchange.max-buffer-size` to 1GB, but it still does not work. The query uses `array_agg()`, and seems that the array size is too large.

Is there any configuration that can help my query work? Thanks for helping me

raunaqmorarka:
This is a regression from jetty upgrade in https://github.com/trinodb/trino/pull/26453 in 477
https://github.com/trinodb/trino/pull/27373 should address it
cc: @wendigo @findepi

findepi:
> [#27373](https://github.com/trinodb/trino/pull/27373) should address it

[...]

PR #27851, opened 2026-01-05
Increase default exchange HTTP max content length to 64MB

raunaqmorarka:
## Description
Avoids "Remote page is too large" exceptions when dealing with large rows

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/ServerMainModule.java
+++ b/core/trino-main/src/main/java/io/trino/server/ServerMainModule.java
@@ -328,7 +328,9 @@ public class ServerMainModule
                 .withConfigDefaults(config -> {
                     config.setIdleTimeout(new Duration(30, SECONDS));
                     config.setRequestTimeout(new Duration(10, SECONDS));
-                    config.setMaxContentLength(DataSize.of(32, MEGABYTE));
+                    // Bumping the default limit to allow large rows to be transferred between nodes
+                    // 64MB is chosen based on real world experience of using this value in production clusters
+                    config.setMaxContentLength(DataSize.of(64, MEGABYTE));
                     config.setMaxRequestsQueuedPerDestination(65536);
                     if (internalCommunicationConfig.isHttp2Enabled()) {
                         config.setMaxConnectionsPerServer(64);


core/trino-main/src/main/java/io/trino/server/ServerMainModule.java
```diff
@@ -328,7 +328,7 @@ protected void setup(Binder binder)
                 .withConfigDefaults(config -> {
                     config.setIdleTimeout(new Duration(30, SECONDS));
                     config.setRequestTimeout(new Duration(10, SECONDS));
-                    config.setMaxContentLength(DataSize.of(32, MEGABYTE));
+                    config.setMaxContentLength(DataSize.of(64, MEGABYTE));
```
findepi, line 333: Maybe worth a code comment how the value was obtained.

findepi (approved):

findepi:
Do you want to include [Add a test which pushes a large row through a remote exchange](https://github.com/trinodb/trino/pull/27373/changes/dce8aaabab619a8716de0771b45b53b152dbf7da) from https://github.com/trinodb/trino/pull/27373 ?

raunaqmorarka:
> Do you want to include [Add a test which pushes a large row through a remote exchange](https://github.com/trinodb/trino/pull/27373/changes/dce8aaabab619a8716de0771b45b53b152dbf7da) from #27373 ?

Bumping the config is not enough to pass that test.
It still fails with `Buffering capacity 67108864 exceeded`.
It's a bug in 2 ways:
- Config limit of 64mb won't let a row of 48mb pass
- The exception message is misleading, we should've gotten PageTooLargeException instead

findepi:
> It's a bug in 2 ways:

If you're not going to fix these in this PR (which I am ok with), worth having issue(s)?


it might still be worth having a test though

- successful one with a large row value that works 
- and maybe a negative one that is supposed to work (so that we know when code works as expected, which might be intentional fix or incidental)

raunaqmorarka:
> > It's a bug in 2 ways:
> 
> If you're not going to fix these in this PR (which I am ok with), worth having issue(s)?
> 
> it might still be worth having a test though
> 
> * successful one with a large row value that works
> * and maybe a negative one that is supposed to work (so that we know when code works as expected, which might be intentional fix or incidental)

There was already a reported issue https://github.com/trinodb/trino/issues/27730 for the same thing, we should re-open that. The buffer in "Buffering capacity ... exceeded" is clearly oversized compared to the actual row size and that needs to be fixed.
https://github.com/trinodb/trino/pull/27373 was fixing that, I'm fine with any other way of fixing it as well.

findepi:
> There was already a reported issue #27730 for the same thing, we should re-open that. 

That issue description looks like general "it doesn't work for me" report. It was closed on the assumption it works for the user now.
I'd rather have an "internal" issue with description that starts with what we know rather that hijacking a user report for that purpose.

> The buffer in "Buffering capacity ... exceeded" is clearly oversized compared to the actual row size and that needs to be fixed.

if this needs to be fixed, we definitely should have an issue.

> https://github.com/trinodb/trino/pull/27373 was fixing that

💪  & 🤞

Issue #27646, opened 2025-12-14
Got error with clickhouse connector when upgrade clickhouse to ver25.12.1.503


ufosky:
I use trino 478, and got this error:
Error listing schemas for catalog clickhouse: java.io.IOException: Magic is not correct - expect [-126] but got [123].

```
connector.name=clickhouse
connection-url=jdbc:clickhouse://localhost:8123/
connection-user=
connection-password=
clickhouse.map-string-as-varchar=true
```

ebyhr:
Could you please add the full stacktrace to the issue description?

vietvudanh:
Same error. I use Trino on k8s, this is the only logs I got on Trino Coordinator container.
Chart 1.41.0, Trino 477, Clickhouse 25.11.2.24.

```
2026-01-05T10:08:21.458Z	INFO	dispatcher-query-16	io.trino.event.QueryMonitor	TIMELINE: Query 20260105_100821_00015_sg42m :: FAILED (JDBC_ERROR) :: elapsed 32ms :: planning 0ms :: waiting 0ms :: scheduling 32ms :: running 0ms :: finishing 32ms :: begin 2026-01-05T10:08:21.424Z :: end 2026-01-05T10:08:21.456Z
```

CLI output:
```
trino> use clickhouse_db.db;
Query 20260105_100821_00015_sg42m failed: java.io.IOException: Magic is not correct - expect [-126] but got [123]
```

vietvudanh:
OK looks like this issue pin point out the correct problems: https://github.com/ClickHouse/clickhouse-java/issues/2636
Clickhouse version > 25.10 cause the error, and Clickhouse JDBC 0.9.4 plugin solved the problem.

Currently Trino was using 0.7.1-patch1 (https://github.com/trinodb/trino/blob/f9b723f1bfe3b163d82f9ec5c1c5e57a13be4dda/pom.xml#L380). I am not familiar with Trino so anyone can help making the change to fix? @ebyhr ?

wendigo:
https://github.com/trinodb/trino/pull/28964

rakkhin:
Trino 481 & Clickhouse 26.3.12, still broken

```
io.trino.spi.TrinoException: Error listing schemas for catalog some_catalog_name: java.io.IOException: Magic is not correct - expect [-126] but got [123]
	at io.trino.metadata.MetadataListing.handleListingException(MetadataListing.java:375)
[...]

PR #26735, opened 2025-09-29
Upgrade ClickHouse client version

krvikash:
## Description

Upgrade ClickHouse client version to 0.9.2

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(X) Release notes are required, with the following suggested text:

```markdown
## ClickHouse
* Upgrade ClickHouse client version. ({issue}`issuenumber`)
```

PR #27853, opened 2026-01-06
Update ClickHouse JDBC dependency version to 0.9.5

vietvudanh:
## Description
 Fix https://github.com/trinodb/trino/issues/27646

Clickhouse version > 25.10 breaks changes. Need to update clickhouse-jdbc > 0.9.5 to fix it.


vietvudanh:
> FYI, there is a braking change in 0.9.x #26735

Ah thank you, I will take a look.

PR #27857, opened 2026-01-06
feat: generate Hudi test tables on the fly

vamsikarnika:
## Description

Generate hudi test tables on the fly for Trino hudi tests to reduce number of file changes every time we add a new test table.

## Release notes

```markdown


vamsikarnika:
@ebyhr Can you please review this pr?

PR #27862, opened 2026-01-06
Propagate access control exception

Chaho12:
## Description

**Background**

This issue occurs when using StorageBasedAuthorizationProvider in Hive Metastore.
In this case, an AccessControlException can be thrown during authorization checks, but the error is not propagated to the client immediately, causing the client to wait until the metastore Thrift timeout expires (default: 10s).

Relevant code:
- https://github.com/apache/hive/blob/master/ql/src/java/org/apache/hadoop/hive/ql/security/authorization/StorageBasedAuthorizationProvider.java#L429-L451

**Fix**
This change propagates AccessControlExceptions to the client immediately, avoiding unnecessary waits for the Thrift timeout.

Here is the exception message sent from metastore.
```
MetaException(message:java.security.AccessControlException: Permission denied: user=..., access=EXECUTE, inode="/user/...":...:services:drwx------
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.check(FSPermissionChecker.java:399)
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkTraverse(FSPermissionChecker.java:315)
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:242)
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkDefaultEnforcer(RangerHdfsAuthorizer.java:589)
	at org.apache.ranger.authorization.hadoop.RangerHdfsAuthorizer$RangerAccessControlEnforcer.checkPermission(RangerHdfsAuthorizer.java:377)
	at org.apache.hadoop.hdfs.server.namenode.FSPermissionChecker.checkPermission(FSPermissionChecker.java:193)
	at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPermission(FSDirectory.java:1877)
	at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPermission(FSDirectory.java:1861)
	at org.apache.hadoop.hdfs.server.namenode.FSDirectory.checkPathAccess(FSDirectory.java:1811)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.checkAccess(FSNamesystem.java:8048)
	at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.checkAccess(NameNodeRpcServer.java:2234)
	at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.checkAccess(ClientNamenodeProtocolServerSideTranslatorPB.java:1659)
	at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:523)
	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:991)
	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:872)
	at org.apache.hadoop.ipc.Server$RpcCall.run(Server.java:818)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1729)
	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2678)
)
```

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/RetryDriver.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/RetryDriver.java
@@ -13,6 +13,7 @@
  */
 package io.trino.plugin.hive.metastore.thrift;
 
+import com.google.common.base.Throwables;
 import com.google.common.collect.ImmutableList;
 import io.airlift.log.Logger;
 import io.airlift.units.Duration;
@@ -74,6 +75,24 @@ public class RetryDriver
         return new RetryDriver();
     }
 
+    /**
+     * Check if the exception indicates an access control failure.
+     * These exceptions should not be retried as they represent permanent
+     * authorization failures.
+     * For example: MetaException wrapping AccessControlException from HDFS
+     * permission denied errors.
+     */
+    private static boolean isAccessControlException(Exception exception)
+    {
+        // Check the exception message and cause chain for AccessControlException
+        // e.g. io.trino.hive.thrift.metastore.MetaException:
+        // org.apache.hadoop.security.AccessControlException: Permission denied: ...
+        return Throwables.getCausalChain(exception)
+                .stream()
+                .map(Throwable::toString)
+                .anyMatch(message -> message != null && message.contains("AccessControlException"));
+    }
+
     public final RetryDriver maxAttempts(int maxAttempts)
     {
         return new RetryDriver(maxAttempts, minSleepTime, maxSleepTime, scaleFactor, maxRetryTime, stopOnExceptions);
@@ -127,6 +146,11 @@ public class RetryDriver
                         addSuppressed(e, suppressedExceptions);
                         throw e;
                     }
+                }
+                // Do not retry on access control exceptions - these are permanent failures
+                if (isAccessControlException(e)) {
+                    addSuppressed(e, suppressedExceptions);
+                    throw e;
                 }
                 if (attempt >= maxAttempts || Duration.nanosSince(startTime).compareTo(maxRetryTime) >= 0) {
                     addSuppressed(e, suppressedExceptions);
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/metastore/thrift/TestRetryDriver.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/metastore/thrift/TestRetryDriver.java
@@ -1,0 +1,160 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.hive.metastore.thrift;
+
+import io.airlift.units.Duration;
+import io.trino.hive.thrift.metastore.MetaException;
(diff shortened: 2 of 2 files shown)


plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/RetryDriver.java
Chaho12, line 144: stopOnExceptions passed on here was

```
0 = {Class@37632} "class io.trino.spi.TrinoException"
1 = {Class@42658} "class io.trino.hive.thrift.metastore.NoSuchObjectException"
2 = {Class@42226} "class java.lang.NullPointerException"
3 = {Class@41494} "class java.lang.IllegalStateException"
4 = {Class@42222} "class java.lang.IllegalArgumentException"
```

Chaho12:
Ps. Hive recently sent a PR (Hive 4.3) to handles similar issue in "Propagate HiveAccessControlException to HiveCatalog" 
- https://github.com/apache/hive/pull/6171

It seems that Hive 4.3 will return in ForbiddenException type, so maybe we can add that when checking isAccessControlException

Chaho12:
Can you take a look at this? @raunaqmorarka

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/RetryDriver.java
```diff
@@ -74,6 +74,27 @@ public static RetryDriver retry()
...
+     * These exceptions should not be retried as they represent permanent
+     * authorization failures.
+     * For example: MetaException wrapping AccessControlException from HDFS
+     * permission denied errors.
+     */
+    private static boolean isAccessControlException(Exception exception)
+    {
+        // Check the exception message and cause chain for AccessControlException
+        // e.g. io.trino.hive.thrift.metastore.MetaException:
+        // org.apache.hadoop.security.AccessControlException: Permission denied: ...
```
wendigo, line 89: this could be simplified to:

```
Throwables.getCausalChain(exception)
                .stream()
                .map(Throwable::toString)
                .anyMatch(message -> message != null && message.contains("AccessControlException"));
```

Chaho12: Simplified :)

chenjian2664:
cc @Praveen2112

Chaho12:
It would be great if this PR is included in next release.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/RetryDriver.java
```diff
@@ -74,6 +75,24 @@ public static RetryDriver retry()
...
+        return Throwables.getCausalChain(exception)
+                .stream()
+                .map(Throwable::toString)
+                .anyMatch(message -> message != null && message.contains("AccessControlException"));
```
chenjian2664, line 93: Just an idea : maybe we could use the class to check if it is a `AccessControlException`, also since the class is marked deprecated, so could use the `SecurityException` instead?

Chaho12: You mean do this instead of contains?
- `err.getClass().getName().equals("org.apache.hadoop.fs.permission.AccessControlException")) {`

Seems like it is superseded by "org.apache.hadoop.security.AccessControlException" 
Nonetheless, depending on Hive version, any of the class name could occur, so we could check for both class

Chaho12:
Any furthere comments? I can rebase it later

PR #27869, opened 2026-01-07
Upgrade npm packages

gertjanal:
## Description
`npm audit fix`

Validated, everything seems to work

## Release notes

### DIFF
--- a/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json
+++ b/core/trino-web-ui/src/main/resources/webapp-preview/package-lock.json
@@ -523,9 +523,9 @@
       "license": "MIT"
     },
     "node_modules/@esbuild/aix-ppc64": {
-      "version": "0.25.10",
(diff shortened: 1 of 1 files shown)


koszti:
This is somewhat related to and partially addressed by https://github.com/trinodb/trino/pull/27632.

I’m generally in favor of running `npm audit fix` (and `npm update`) regularly but it raises the question of how we want to handle vulnerabilities going forward. From what I can see in the main trino java codebase we usually bump dependencies one by one in separate commits which doesn’t align with how npm commands operates. Do we want to adopt them or prefer more safe and semi-manual one-dependency-per-commit approach? cc @mosabua 

UPDATE: I had another look, react bump isn't covered by this PR so please ignore that.

wendigo (approved):

wendigo:
@koszti updating one-by-one would be nice. We can enable dependabot for npm packages

PR #27871, opened 2026-01-07
Retain avg size estimates through CAST expressions

raunaqmorarka:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/cost/ScalarStatsCalculator.java
+++ b/core/trino-main/src/main/java/io/trino/cost/ScalarStatsCalculator.java
@@ -177,6 +177,8 @@ public class ScalarStatsCalculator
                     .setLowValue(lowValue)
                     .setHighValue(highValue)
                     .setDistinctValuesCount(distinctValuesCount)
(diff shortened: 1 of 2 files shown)


core/trino-main/src/main/java/io/trino/cost/ScalarStatsCalculator.java
```diff
@@ -177,6 +177,8 @@ protected SymbolStatsEstimate visitCast(Cast node, Void context)
                     .setLowValue(lowValue)
                     .setHighValue(highValue)
                     .setDistinctValuesCount(distinctValuesCount)
+                    // Source average row size is a reasonable approximation for CAST result
+                    .setAverageRowSize(sourceStats.getAverageRowSize())
```
Praveen2112, line 181: For some primitive data types it could be specific right ?

raunaqmorarka: For fixed width types this is going to be NaN, `io.trino.cost.PlanNodeStatsEstimate#getOutputSizeForSymbol` figures that out automatically

PR #27880, opened 2026-01-08
Fix failure reading `FLOAT` type in `Query` table function in Oracle

chenjian2664:
## Description

Reading `FLOAT` in `Query` table function in Oracle will fail with:
```
io.trino.testing.QueryFailedException: com.google.common.util.concurrent.UncheckedExecutionException: 
java.lang.UnsupportedOperationException: 
Unsupported type: JdbcTypeHandle[jdbcType=2, jdbcTypeName=Optional[NUMBER], columnSize=Optional[126], decimalDigits=Optional[-127], arrayDimensions=Optional.empty, caseSensitivity=Optional[CASE_INSENSITIVE]] of column: Y
```
The return types in `Query` table function are derived from a prepared query https://github.com/trinodb/trino/blob/ce0492437e72d9e7483ef0028a941b333f82e488/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ptf/Query.java#L99C8-L106
Oracle will declare type to NUMBER for the `FLOAT` type.

## Release notes

```markdown
## Oracle
* Fix failure when reading `float` type in `query` table function. ({issue}`27880`)
```

### DIFF
--- a/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
+++ b/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
@@ -26,6 +26,7 @@ import io.trino.plugin.base.projection.ProjectFunctionRule;
 import io.trino.plugin.jdbc.BaseJdbcClient;
 import io.trino.plugin.jdbc.BaseJdbcConfig;
 import io.trino.plugin.jdbc.BooleanWriteFunction;
+import io.trino.plugin.jdbc.CaseSensitivity;
 import io.trino.plugin.jdbc.ColumnMapping;
 import io.trino.plugin.jdbc.ConnectionFactory;
 import io.trino.plugin.jdbc.DoubleWriteFunction;
@@ -81,6 +82,7 @@ import java.sql.Connection;
 import java.sql.DatabaseMetaData;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
+import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Types;
 import java.time.Instant;
@@ -108,6 +110,8 @@ import static com.google.common.base.Strings.emptyToNull;
 import static com.google.common.base.Verify.verify;
 import static io.airlift.slice.Slices.utf8Slice;
 import static io.airlift.slice.Slices.wrappedBuffer;
+import static io.trino.plugin.jdbc.CaseSensitivity.CASE_INSENSITIVE;
+import static io.trino.plugin.jdbc.CaseSensitivity.CASE_SENSITIVE;
 import static io.trino.plugin.jdbc.JdbcErrorCode.JDBC_ERROR;
 import static io.trino.plugin.jdbc.PredicatePushdownController.DISABLE_PUSHDOWN;
 import static io.trino.plugin.jdbc.PredicatePushdownController.FULL_PUSHDOWN;
@@ -162,6 +166,7 @@ import static java.lang.Math.min;
 import static java.lang.Math.toIntExact;
 import static java.lang.String.format;
 import static java.lang.String.join;
+import static java.sql.Types.FLOAT;
 import static java.util.Locale.ENGLISH;
 import static java.util.concurrent.TimeUnit.DAYS;
 
@@ -391,6 +396,75 @@ public class OracleClient
             throws SQLException
     {
         return Optional.ofNullable(emptyToNull(resultSet.getString("REMARKS")));
+    }
+
+    @Override
+    protected List<JdbcColumnHandle> getColumns(ConnectorSession session, Connection connection, ResultSetMetaData metadata)
+            throws SQLException
+    {
+        ImmutableList.Builder<JdbcColumnHandle> columns = ImmutableList.builder();
+        for (int column = 1; column <= metadata.getColumnCount(); column++) {
+            JdbcTypeHandle jdbcTypeHandle = getJdbcTypeHandle(session, metadata, column);
+
+            // Use getColumnLabel method because query pass-through table function may contain column aliases
+            String name = metadata.getColumnLabel(column);
+            Type type = toColumnMapping(session, connection, jdbcTypeHandle)
+                    .orElseThrow(() -> new UnsupportedOperationException(format("Unsupported type: %s of column: %s", jdbcTypeHandle, name)))
+                    .getType();
+            columns.add(new JdbcColumnHandle(name, jdbcTypeHandle, type));
+        }
+        return columns.build();
+    }
+
+    private static JdbcTypeHandle getJdbcTypeHandle(ConnectorSession session, ResultSetMetaData metadata, int column)
+            throws SQLException
+    {
+        int columnType = metadata.getColumnType(column);
+        int columnSize = metadata.getPrecision(column);
+        int scale = metadata.getScale(column);
+        CaseSensitivity caseSensitive = metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE;
+        if (columnType == OracleTypes.NUMBER) {
+            String columnClassName = metadata.getColumnClassName(column);
+            // Oracle report both NUMBER and FLOAT column as NUMBER type via `getColumnTypeName` (in table function),
+            // so we have to rely on the column class name here, Oracle NUMBER column are expected "java.math.BigDecimal",
+            // when the column class name is "java.lang.Double" it means it's actual a FLOAT type in Oracle side
+            // and the scale is expected to be -127
+            if ("java.lang.Double".equals(columnClassName)) {
+                int precision = columnSize + max(-scale, 0);
+                verify(scale == -127, "FLOAT scale is expected to be -127 but got %s", scale);
+                // we just handle the precision here is not able to handle by `toColumnMapping` method
+                if (!isAllowedNumber(session, precision, scale, columnSize)) {
+                    return new JdbcTypeHandle(FLOAT, Optional.of("FLOAT"), Optional.of(columnSize), Optional.of(scale), Optional.empty(), Optional.of(caseSensitive));
+                }
+            }
+        }
+
+        return new JdbcTypeHandle(
+                columnType,
+                Optional.ofNullable(metadata.getColumnTypeName(column)),
+                Optional.of(columnSize),
+                Optional.of(scale),
+                Optional.empty(), // TODO support arrays
+                Optional.of(caseSensitive));
+    }
+
+    private static boolean isAllowedNumber(ConnectorSession session, int precision, int scale, int columnSize)
+    {
+        Optional<Integer> numberDefaultScale = getNumberDefaultScale(session);
+        RoundingMode roundingMode = getNumberRoundingMode(session);
+        if (precision < scale) {
+            if (roundingMode == RoundingMode.UNNECESSARY) {
+                return false;
+            }
+        }
+        else if (numberDefaultScale.isPresent() && precision == PRECISION_OF_UNSPECIFIED_NUMBER) {
+            return true;
+        }
+        else if (precision > Decimals.MAX_PRECISION || columnSize <= 0) {
+            return false;
+        }
+
+        return true;
     }
 
     // Iterates over filtered schemas to fetch column details per schema, preventing unnecessary columns for Oracle's internal schemas
--- a/plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
+++ b/plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
@@ -121,6 +121,32 @@ public abstract class BaseOracleConnectorTest
     }
 
     @Test
+    void testReadingFloatWithQueryTableFunction()
+    {
+        testReadingFloatWithQueryTableFunction("FLOAT");
+        testReadingFloatWithQueryTableFunction("FLOAT(23)");
+        testReadingFloatWithQueryTableFunction("FLOAT(24)");
+        testReadingFloatWithQueryTableFunction("FLOAT(53)");
+        testReadingFloatWithQueryTableFunction("FLOAT(126)");
+    }
+
+    private void testReadingFloatWithQueryTableFunction(String floatType)
+    {
+        try (TestTable table = new TestTable(onRemoteDatabase(), "test_float_" + randomNameSuffix(), "(x int, y %s)".formatted(floatType))) {
+            String tableName = table.getName();
+            assertUpdate("INSERT INTO " + tableName + " VALUES (1, 0.123), (2, 456.789), (3, NULL)", 3);
+
+            // test both query with and without through query table function, make sure the type and values are the same
+            // for the oracle FLOAT type
+            String expectedValues = "VALUES CAST(0.123 AS DOUBLE), CAST(456.789 AS DOUBLE), CAST(NULL as DOUBLE)";
+            assertThat(query("SELECT y FROM " + tableName))
+                    .matches(expectedValues);
+            assertThat(query("SELECT y FROM TABLE(system.query('SELECT * FROM " + tableName + "'))"))
+                    .matches(expectedValues);
+        }
+    }
+
+    @Test
     @Override
     public void testShowColumns()
     {


plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,67 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+        int columnType = metadata.getColumnType(column);
+        int columnSize = metadata.getPrecision(column);
+        int scale = metadata.getScale(column);
+        CaseSensitivity caseSensitive = metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE;
+        if (columnType == Types.NUMERIC) {
+            String columnClassName = metadata.getColumnClassName(column);
+            int precision = columnSize + max(-scale, 0);
+            // Oracle NUMBER are expected "java.math.BigDecimal", when the class name is "java.lang.Double" it means
+            // it's actual a FLOAT type in Oracle side
+            if ("java.lang.Double".equals(columnClassName)) {
```
findepi, line 432: Did you try using `metadata.getColumnTypeName` to distinguish between them?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,67 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+    }
+
+    private static JdbcTypeHandle getJdbcTypeHandle(ResultSetMetaData metadata, int column)
+            throws SQLException
+    {
+        int columnType = metadata.getColumnType(column);
+        int columnSize = metadata.getPrecision(column);
+        int scale = metadata.getScale(column);
+        CaseSensitivity caseSensitive = metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE;
+        if (columnType == Types.NUMERIC) {
```
findepi: Do i understand correctly that in get table columns flow we get `OracleTypes.FLOAT` (6)
and in ResultSetMetaData we get `Types.NUMERIC` (2)?

(nit: you can use a constant from `OracleTypes` here too)

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,67 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+            // Oracle NUMBER are expected "java.math.BigDecimal", when the class name is "java.lang.Double" it means
+            // it's actual a FLOAT type in Oracle side
+            if ("java.lang.Double".equals(columnClassName)) {
+                // we just handle the precision here is not able to handle by `toColumnMapping` method
+                if (!isAllowedNumber(precision, columnSize)) {
+                    return new JdbcTypeHandle(FLOAT, Optional.of("FLOAT"), Optional.of(columnSize), Optional.of(scale), Optional.empty(), Optional.of(caseSensitive));
+                }
+            }
+            else {
+                verify("java.math.BigDecimal".equals(columnClassName), "Unexpected class name %s for Oracle NUMBER type", columnClassName);
```
findepi: I am not sure we want this check. Would perhaps feel safer without it.
We explicitly request conversion to BD with `resultSet.getBigDecimal`, so we don't need the driver to have this as default.

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,67 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+                Optional.of(caseSensitive));
+    }
+
+    private static boolean isAllowedNumber(int precision, int columnSize)
+    {
+        if (precision > Decimals.MAX_PRECISION) {
+            return false;
+        }
+        if (columnSize <= 0) {
+            return false;
```
findepi: Document

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,67 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+        return new JdbcTypeHandle(
+                columnType,
+                Optional.ofNullable(metadata.getColumnTypeName(column)),
+                Optional.of(columnSize),
+                Optional.of(scale),
+                Optional.empty(), // TODO support arrays
+                Optional.of(caseSensitive));
+    }
+
+    private static boolean isAllowedNumber(int precision, int columnSize)
```
findepi: Why this doesn't compare `columnSize` and `precision` values?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: The `metadata.getColumnTypeName` returns 2 -> `NUMBER`
So can't use it to identify the actual type

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: > Do i understand correctly that in get table columns flow we get OracleTypes.FLOAT (6)
and in ResultSetMetaData we get Types.NUMERIC (2)?

Yes it is.

updated to use `OracleTypes`

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: Ok, removed this check

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: Updated the logic to follow https://github.com/trinodb/trino/blob/2d361057a3d78926bcbb578057f1e96bc32e9d19/plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java#L507-L520 
The purpose of the method is to check if it is able to be handled as the `NUMBER` - as we could always correct interpret the type and with less "lossy"
If it not, we use it actual type - FLOAT

plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
```diff
@@ -120,6 +120,23 @@ protected TestTable createTableWithUnsupportedColumn()
                 "(one NUMBER(19), two NUMBER, three VARCHAR2(10 CHAR))");
     }
 
+    @Test
+    void testReadingFloatWithQueryTableFunction()
+    {
+        String tableName = "test_float_" + randomNameSuffix();
+        onRemoteDatabase().execute("CREATE TABLE " + tableName + " (x int, y FLOAT)");
```
krvikash: Can we use `TestTable` here?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,72 @@ public Optional<String> getTableComment(ResultSet resultSet)
         return Optional.ofNullable(emptyToNull(resultSet.getString("REMARKS")));
     }
 
+    @Override
+    protected List<JdbcColumnHandle> getColumns(ConnectorSession session, Connection connection, ResultSetMetaData metadata)
```
krvikash, line 402: Why are we not fixing this in `OracleClient#toColumnMapping`? Is it because in `OracleClient#toColumnMapping`, we don't have access to metadata to get `columnClassName` to make the decision?

krvikash: For this case, I see that actualPrecision is coming as 126 and decimalDigits as -127. I am not sure if this can be used to identify this failing case.

chenjian2664: The `OracleClient#toColumnMapping` is correct that we are mapping `FLOAT` to `DOUBLE`.
When resolve the type from query function, we use this [method](https://github.com/trinodb/trino/blob/b3c5bab7ec84f52a63c7d07a463f821470505c37/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L291) to get the output type, Oracle returns `NUMBER` type instead of `FLOAT` for `FLOAT` column in the query table function.

chenjian2664: @krvikash the decimalDigits seems stick to `-127`, but the precision will be different if you specify the p of `FLOAT(p)`, i,e to 10

krvikash: Can we do something like this in `case OracleTypes.NUMBER`:?

```
                else if (decimalDigits == -127) {
                    return Optional.of(ColumnMapping.doubleMapping(
                            DOUBLE,
                            ResultSet::getDouble,
                            oracleDoubleWriteFunction(),
                            FULL_PUSHDOWN));
                }
```

chenjian2664: Do you mean handle it in the `toColumnMapping`, any reason why we do in here?
I was considered but think it is the place might have bigger scope - it will influence all normal the column mapping, while in this pr we only touch the place the derived the type from a query

krvikash: > Do you mean handle it in the toColumnMapping

Yes,

> it will influence all normal the column mapping, while in this pr we only touch the place the derived the type from a query

Yes, I feel the same.

plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
```diff
@@ -120,6 +120,23 @@ protected TestTable createTableWithUnsupportedColumn()
                 "(one NUMBER(19), two NUMBER, three VARCHAR2(10 CHAR))");
     }
 
+    @Test
+    void testReadingFloatWithQueryTableFunction()
+    {
+        try (TestTable table = new TestTable(onRemoteDatabase(), "test_float_" + randomNameSuffix(), "(x int, y FLOAT)")) {
```
krvikash: Can we add a column with precision as well?

chenjian2664: updated

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,72 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+    private static JdbcTypeHandle getJdbcTypeHandle(ConnectorSession session, ResultSetMetaData metadata, int column)
+            throws SQLException
+    {
+        int columnType = metadata.getColumnType(column);
+        int columnSize = metadata.getPrecision(column);
+        int scale = metadata.getScale(column);
+        CaseSensitivity caseSensitive = metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE;
+        if (columnType == OracleTypes.NUMBER) {
+            String columnClassName = metadata.getColumnClassName(column);
+            int precision = columnSize + max(-scale, 0);
```
ebyhr: `precision` can be moved inside the `if ("java.lang.Double".equals(columnClassName)) {`.

plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
```diff
@@ -120,6 +120,33 @@ protected TestTable createTableWithUnsupportedColumn()
...
+        testReadingFloatWithQueryTableFunction(OptionalInt.empty());
+        testReadingFloatWithQueryTableFunction(OptionalInt.of(23));
+        testReadingFloatWithQueryTableFunction(OptionalInt.of(24));
+        testReadingFloatWithQueryTableFunction(OptionalInt.of(53));
+        testReadingFloatWithQueryTableFunction(OptionalInt.of(126));
+    }
+
+    private void testReadingFloatWithQueryTableFunction(OptionalInt precision)
+    {
+        String floatType = precision.isEmpty() ? "FLOAT" : "FLOAT(" + precision.getAsInt() + ")";
```
ebyhr: nit: We could change the parameter to String:
```java
    @Test
    void testReadingFloatWithQueryTableFunction()
    {
        testReadingFloatWithQueryTableFunction("FLOAT");
        testReadingFloatWithQueryTableFunction("FLOAT(23)");
        testReadingFloatWithQueryTableFunction("FLOAT(24)");
        testReadingFloatWithQueryTableFunction("FLOAT(53)");
        testReadingFloatWithQueryTableFunction("FLOAT(126)");
    }

    private void testReadingFloatWithQueryTableFunction(String floatType)
...
```

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
ebyhr: I think it's worth leaving a code comment.

plugin/trino-oracle/src/test/java/io/trino/plugin/oracle/BaseOracleConnectorTest.java
chenjian2664: thanks!

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: updated the comment to reflect the `getColumnTypeName` can't use to distinguish the NUMBER with FLOAT.
Let me know if you think there is a better phrase for it

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,73 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+            String name = metadata.getColumnLabel(column);
+            Type type = toColumnMapping(session, connection, jdbcTypeHandle)
+                    .orElseThrow(() -> new UnsupportedOperationException(format("Unsupported type: %s of column: %s", jdbcTypeHandle, name)))
+                    .getType();
+            columns.add(new JdbcColumnHandle(name, jdbcTypeHandle, type));
+        }
+        return columns.build();
+    }
+
+    private static JdbcTypeHandle getJdbcTypeHandle(ConnectorSession session, ResultSetMetaData metadata, int column)
```
Praveen2112, line 419: nit: columnIndex

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,73 @@ public Optional<String> getTableComment(ResultSet resultSet)
         return Optional.ofNullable(emptyToNull(resultSet.getString("REMARKS")));
     }
 
+    @Override
+    protected List<JdbcColumnHandle> getColumns(ConnectorSession session, Connection connection, ResultSetMetaData metadata)
+            throws SQLException
+    {
+        ImmutableList.Builder<JdbcColumnHandle> columns = ImmutableList.builder();
+        for (int column = 1; column <= metadata.getColumnCount(); column++) {
+            JdbcTypeHandle jdbcTypeHandle = getJdbcTypeHandle(session, metadata, column);
```
Praveen2112, line 407: Should we introduce `getJdbcTypeHandle` as an extension point so that connectors could implement their logic without mimicing the boilder plate code ?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
```diff
@@ -393,6 +398,73 @@ public Optional<String> getTableComment(ResultSet resultSet)
...
+        int columnSize = metadata.getPrecision(column);
+        int scale = metadata.getScale(column);
+        CaseSensitivity caseSensitive = metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE;
+        if (columnType == OracleTypes.NUMBER) {
+            String columnClassName = metadata.getColumnClassName(column);
+            // Oracle report both NUMBER and FLOAT column as NUMBER type via `getColumnTypeName` (in table function),
+            // so we have to rely on the column class name here, Oracle NUMBER column are expected "java.math.BigDecimal",
+            // when the column class name is "java.lang.Double" it means it's actual a FLOAT type in Oracle side
+            if ("java.lang.Double".equals(columnClassName)) {
+                int precision = columnSize + max(-scale, 0);
```
Praveen2112, line 433: When would scale be negative here

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: I want to make consistency with https://github.com/trinodb/trino/blob/195d53668c5d2484a5c9b735fab74e4fb4c5f821/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L295 so, let's keep the `column` as the name

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: Currently only Oracle has the requirements, so seems not so urgent to do such extension, wdyt?

plugin/trino-oracle/src/main/java/io/trino/plugin/oracle/OracleClient.java
chenjian2664: When the `FLOAT` type was selected from a `Query` table function, the type is `NUMBER` and the scale is `-127`

Praveen2112: Can we add it as a code comment ?

Issue #27850, opened 2026-01-05
Failure when reading `$file_modified_time` column on Iceberg table with equality deletes


mcarmonagonzalez:
An Iceberg table with equality deletes was generated by following [this blog post](https://aws.amazon.com/blogs/big-data/stream-real-time-data-into-apache-iceberg-tables-in-amazon-s3-using-amazon-data-firehose/) to manage an Iceberg table via Amazon Data Firehose.

Using Trino 476 via Amazon EMR 7.12, queries attempting to read the `$file_modified_time` column from the table fails with the following error:
```
trino:firehose_iceberg_db> select "$file_modified_time" from firehose_events_2;
...
Query 20260105_160953_00010_vva49 failed: Timestamp precision (3) not supported for Iceberg. Use "timestamp(6) with time zone" instead.
io.trino.spi.TrinoException: Timestamp precision (3) not supported for Iceberg. Use "timestamp(6) with time zone" instead.
	at io.trino.plugin.iceberg.TypeConverter.toIcebergTypeInternal(TypeConverter.java:186)
	at io.trino.plugin.iceberg.TypeConverter.toIcebergType(TypeConverter.java:126)
	at io.trino.plugin.iceberg.IcebergUtil.lambda$structTypeFromHandles$0(IcebergUtil.java:456)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:214)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1716)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:570)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:560)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:265)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:723)
	at io.trino.plugin.iceberg.IcebergUtil.structTypeFromHandles(IcebergUtil.java:457)
	at io.trino.plugin.iceberg.delete.EqualityDeleteFilter.createPredicate(EqualityDeleteFilter.java:58)
	at io.trino.plugin.iceberg.delete.DeleteManager.lambda$getDeletePredicate$1(DeleteManager.java:97)
[...]

PR #27882, opened 2026-01-08
Add custom handling for "$file_modified_time" column for equality deletes

findinpath:
## Description

The metadata column "$file_modified_time" has the type `timestamp(3) with time zone` which is not supported natively by Iceberg.
This change adds logic to work around this limitation that becomes apparent while projecting the "$file_modified_time" metadata column on a table which contains equality deletes.

Fixes https://github.com/trinodb/trino/issues/27850

## Release notes

```markdown
## Iceberg
* Fix failure when reading "$file_modified_time" metadata column on tables with equality deletes ({issue}`27850`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/LazyTrinoRow.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/LazyTrinoRow.java
@@ -13,6 +13,7 @@
  */
 package io.trino.plugin.iceberg.delete;
 
+import io.trino.spi.block.Block;
 import io.trino.spi.connector.SourcePage;
 import io.trino.spi.type.Type;
 import org.apache.iceberg.StructLike;
@@ -20,6 +21,7 @@ import org.apache.iceberg.StructLike;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkElementIndex;
 import static io.trino.plugin.iceberg.IcebergPageSink.getIcebergValue;
+import static io.trino.spi.type.TimestampWithTimeZoneType.TIMESTAMP_TZ_MILLIS;
 import static java.util.Objects.requireNonNull;
 
 /**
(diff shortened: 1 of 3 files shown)


ebyhr (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/LazyTrinoRow.java
```diff
@@ -72,4 +74,16 @@ private Object get(int i)
...
+    public static Object getIcebergValue(Block block, int position, Type type)
+    {
+        if (block.isNull(position)) {
+            return null;
+        }
+        if (type.equals(TIMESTAMP_TZ_MILLIS)) {
+            // $file_modified_time has the type TIMESTAMP_TZ_MILLIS in Trino, while Iceberg works with TIMESTAMP_TZ_MICROS
+            return TIMESTAMP_TZ_MILLIS.getLong(block, position);
+        }
+        return IcebergPageSink.getIcebergValue(block, position, type);
```
findinpath: should we maybe move `getIcebergValue()` method from `IcebergPageSink` ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/LazyTrinoRow.java
```diff
@@ -72,4 +74,16 @@ private Object get(int i)
         values[i] = value;
         return value;
     }
+
```
chenjian2664, line 77: What about the `TrinoRow`, why that class don't needs to change

findinpath: `TrinoRow` is being used in reading equality delete files (which cannot have timestamp(3) with time zone columns).
No need to add special handling in `TrinoRow`.

findinpath:
Rebased on `master`

findinpath:
> I think you are resolving the two problems, one is filter the metadata columns from the eq predicate.
another is support reading Trino timestamp millis but Iceberg micros type data?(not sure I am writing correctly)


Note that Iceberg supports only micros timestamp with time zone type.
The current fix is addressing only the need to work with metadata columns of millis timestamp with time zone type - specifically supporting `$file_modified_time` projections.

`LazyTrinoRow` works on top of a page that may contain as well `$file_modified_time` and therefore this contribution adds a workaround to deal with such cases.

Issue #27805, opened 2025-12-31
Trino-created Delta Lake tables left in broken/unreadable state


Pluies:
Trino-created Delta Lake tables can be left in an unreadable state if recreated with a different schema _after_ a checkpoint has been written.

To reproduce,
1. Create a table
2. Run actions on it until a checkpoint is written (default every 10 actions)
3. Run CREATE OR REPLACE with a different schema
4. Result: any reads to this table will now fail with the error message: `Failed to generate splits`

Reproduction (using the `S3DeltaLakeQueryRunnerMain` from [DeltaLakeQueryRunner](https://github.com/trinodb/trino/blob/master/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/DeltaLakeQueryRunner.java) and Trino CLI):

```sql
trino> CREATE TABLE delta.tpch.orders2 (orderkey bigint, orderdate timestamp);
CREATE TABLE
trino> INSERT INTO delta.tpch.orders2 (SELECT orderkey, orderdate from tpch.sf1000.orders limit 100);
INSERT: 100 rows

Query 20251231_103237_00026_3ra56, FINISHED, 3 nodes
Splits: 62 total, 62 done (100.00%)
0.68 [2.02M rows, 20.2KiB] [2.97M rows/s, 29.7KiB/s]

trino> INSERT INTO delta.tpch.orders2 (SELECT orderkey, orderdate from tpch.sf1000.orders limit 100);
INSERT: 100 rows

Query 20251231_103238_00027_3ra56, FINISHED, 3 nodes
Splits: 62 total, 62 done (100.00%)
0.40 [1.72M rows, 29.9KiB] [4.35M rows/s, 75.6KiB/s]

trino> INSERT INTO delta.tpch.orders2 (SELECT orderkey, orderdate from tpch.sf1000.orders limit 100);
INSERT: 100 rows

Query 20251231_103239_00028_3ra56, FINISHED, 3 nodes
Splits: 62 total, 62 done (100.00%)
0.39 [1.72M rows, 12.8KiB] [4.41M rows/s, 32.8KiB/s]

// ... Repeat the INSERT INTO 10 times so that Trino writes a checkpoint

trino> CREATE OR REPLACE TABLE delta.tpch.orders2 (orderkey bigint, orderdate date);
CREATE TABLE
trino> SELECT * FROM delta.tpch.orders2;
Query 20251231_103256_00038_3ra56 failed: Failed to generate splits for tpch.orders2

[...]

PR #27886, opened 2026-01-09
Fix failure reading checkpoint created by `CREATE OR REPLACE` with different schema

chenjian2664:
## Description

Close https://github.com/trinodb/trino/issues/27805

write a checkpoint for `CREATE OR REPLACE` when there is schema changed:
* when changed a column type
* when removed a column
* when renamed a column

## Release notes

```markdown
## Delta Lake
* Fix potential table corruption when executing `CREATE OR REPLACE` with table definition changes. ({issue}`27805`)
```


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1376,7 +1376,8 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
                 transactionLogWriter.flush();
 
                 if (replaceExistingTable) {
-                    writeCheckpointIfNeeded(session, schemaTableName, location, tableHandle.toCredentialsHandle(), tableHandle.getReadVersion(), checkpointInterval, commitVersion);
+                    verify(commitVersion > tableHandle.getReadVersion());
```
findinpath: not needed

see https://github.com/trinodb/trino/blob/2823d175de4b34ea29d9a717db16f79a25f7dbbd/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L1346 



This condition will always be true

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1376,7 +1376,8 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
...
+                    writeCheckpointMandatory(session, schemaTableName, location, tableHandle.toCredentialsHandle(), commitVersion);
```
findinpath: `writeCheckpointMandatory` -> `writeCheckpoint`

findinpath: IIUC we're supposed to write the checkpoint (and ignore the `checkpoint_interval` property) if the table schema has changed.

If the table schema does not change, we could avoid writing the checkpoint.
However, i think replacing of the table it is a good reason to checkpoint the table.

findinpath: Adjust `TestDeltaLakeFileOperations`

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: removed

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: Yes, ideally we should write checkpoint in the case if the schema changes, the `CREATE OR REPLACE` is the such case.

ebyhr:
> Fix failure reading checkpoint created by `CREATE OR REPLACE` with different schema.

This is a bug fix to the write logic, right? The current release note entry looks misleading if so.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -284,6 +284,42 @@ protected MaterializedResult getDescribeOrdersResult()
...
+
+    @Test
+    void testCreateReplaceReadingCheckpointWithDifferentSchemaCTAS()
+    {
+        try (TestTable table = newTrinoTable("test_create_replace_reading_checkpoint_", "(x int, y varchar) with (checkpoint_interval = 2)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, 'aa')", 1);
+            // generate a checkpoint
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (2, 'bb')", 1);
+
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + "  AS SELECT 3 AS x, 3 AS y", 1);
```
ebyhr:
```suggestion
            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " AS SELECT 3 AS x, 3 AS y", 1);
```

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -284,6 +284,42 @@ protected MaterializedResult getDescribeOrdersResult()
                 .build();
     }
 
+    @Test
+    void testCreateReplaceReadingCheckpointWithDifferentSchema()
```
ebyhr, line 288: This test passes even without this PR change. I think we should add SELECT statement to L295-296.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -284,6 +284,42 @@ protected MaterializedResult getDescribeOrdersResult()
...
+    {
+        try (TestTable table = newTrinoTable("test_create_replace_reading_checkpoint_", "(x int, y varchar) with (checkpoint_interval = 2)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, 'aa')", 1);
+            // generate a checkpoint
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (2, 'bb')", 1);
+
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " (x int, y int) with (checkpoint_interval = 2)");
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (3, 3)", 1);
+            assertThat(query("SELECT * FROM " + table.getName()))
```
ebyhr: (For consistency with the following assertQueryReturnsEmptyResult)
```suggestion
            assertThat(query("TABLE " + table.getName()))
```

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -284,6 +284,42 @@ protected MaterializedResult getDescribeOrdersResult()
...
+    @Test
+    void testCreateReplaceReadingCheckpointWithDifferentSchemaCTAS()
+    {
+        try (TestTable table = newTrinoTable("test_create_replace_reading_checkpoint_", "(x int, y varchar) with (checkpoint_interval = 2)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, 'aa')", 1);
+            // generate a checkpoint
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (2, 'bb')", 1);
+
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + "  AS SELECT 3 AS x, 3 AS y", 1);
+            assertThat(query("SELECT * FROM " + table.getName()))
```
ebyhr:
```suggestion
            assertThat(query("TABLE " + table.getName()))
```

ebyhr (approved):

findinpath:
> Does spark or other engines or delta lake spec mentions about writing checkpoint after we run `CREATE OR REPLACE` operation ?


I tried on Databricks (Spark OSS does not support `CREATE OR REPLACE`) and have seen a similar behavior:

```
CREATE TABLE test_cort.t1 (x int, y int)  using delta TBLPROPERTIES ('delta.checkpointInterval' = '2');
insert into test_cort.t1 values (1, 1);
insert into test_cort.t1 values (2, 2);
select * from test_cort.t1;

CREATE OR REPLACE TABLE test_cort.t1 (x int, y string)  using delta TBLPROPERTIES ('delta.checkpointInterval' = '2');
```

s3 listing

```
2026-01-09 08:02:13          0 
2026-01-09 08:02:14          0 .s3-optimization-0
2026-01-09 08:02:14          0 .s3-optimization-1
2026-01-09 08:02:14          0 .s3-optimization-2
2026-01-09 08:02:17       1979 00000000000000000000.crc
2026-01-09 08:02:14        986 00000000000000000000.json
2026-01-09 08:02:20       2430 00000000000000000001.crc
2026-01-09 08:02:19        977 00000000000000000001.json
2026-01-09 08:02:24      17420 00000000000000000002.checkpoint.parquet
2026-01-09 08:02:23       2880 00000000000000000002.crc
2026-01-09 08:02:21        977 00000000000000000002.json
2026-01-09 08:02:30      16285 00000000000000000003.checkpoint.parquet
2026-01-09 08:02:29       1964 00000000000000000003.crc
2026-01-09 08:02:27       1680 00000000000000000003.json
2026-01-09 08:02:30       5059 _last_checkpoint
```

FWIW on Databricks, if the schema of the table is not being changed, there is no new checkpoint being created.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1376,7 +1376,7 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
                 transactionLogWriter.flush();
 
                 if (replaceExistingTable) {
-                    writeCheckpointIfNeeded(session, schemaTableName, location, tableHandle.toCredentialsHandle(), tableHandle.getReadVersion(), checkpointInterval, commitVersion);
+                    writeCheckpoint(session, schemaTableName, location, tableHandle.toCredentialsHandle(), commitVersion);
```
findinpath: As pointed out in https://github.com/trinodb/trino/pull/27886#issuecomment-3727497351 if the schema of the table does not change, there is no new checkpoint created by default (if it doesn't correspond to the checkpoint interval property) in DBX. Consider whether it makes sense to follow the same pattern here as well.

Praveen2112: Agreed - We would also think about handling about schema evolution (we don't support it now) - but when we support it do we need to create a new checkpoint for each alter operation ?

chenjian2664: just tested dbx, only when changing column type it will add a checkpoint.

chenjian2664: added tests cover:
* add column
* remove column
* rename column
* change column type
Only add column doesn't create a new checkpoint.

plugin/trino-delta-lake/src/test/resources/databricks164/test_write_checkpoint_on_schema_change/README.md
```diff
@@ -0,0 +1,33 @@
...
+CREATE OR REPLACE TABLE test_checkpoint_on_schema_change (z int) 
+TBLPROPERTIES('delta.checkpointInterval' = 20);
+INSERT INTO table test_checkpoint_on_schema_change values 10;
+
+CREATE OR REPLACE TABLE test_checkpoint_on_schema_change (z string) 
+TBLPROPERTIES('delta.checkpointInterval' = 20);
+INSERT INTO test_checkpoint_on_schema_change values 'version12';
+```
+
+Note the 'delta.checkpointInterval' table property is set when creating or replacing the table
```
findinpath:
```suggestion
Note the 'delta.checkpointInterval' table property, set when creating or replacing the table,
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3193,6 +3206,37 @@ private void writeCheckpointIfNeeded(
...
+                .filter(column -> !isMetadataColumnHandle(column))
+                .collect(toImmutableMap(DeltaLakeColumnHandle::columnName, DeltaLakeColumnHandle::type));
+
+        for (DeltaLakeColumnHandle existingColumn : existingColumns) {
+            if (isMetadataColumnHandle(existingColumn)) {
+                continue;
+            }
+
+            if (!newColumnHandles.containsKey(existingColumn.columnName())) {
+                // remove column treated as schema change
```
findinpath: I'm assuming then that adding new columns is then "no schema change" ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3193,6 +3206,37 @@ private void writeCheckpointIfNeeded(
...
+            throws IOException
+    {
+        TransactionLogReader transactionLogReader = new FileSystemTransactionLogReader(tableLocation, credentialsHandle, fileSystemFactory);
+        TableSnapshot snapshot = transactionLogAccess.loadSnapshot(session, transactionLogReader, table, tableLocation, Optional.of(newVersion), credentialsHandle);
+        checkpointWriterManager.writeCheckpoint(session, snapshot, credentialsHandle);
+    }
+
+    private static boolean isSchemaChanged(List<DeltaLakeColumnHandle> existingColumns, List<DeltaLakeColumnHandle> newColumns)
+    {
+        Map<String, Type> newColumnHandles = newColumns.stream()
```
findinpath, line 3219: Why not compare `tableHandle.getMetadataEntry().schemaString` with `metadataEntry.schemaString` ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3193,6 +3206,37 @@ private void writeCheckpointIfNeeded(
...
+                .filter(column -> !isMetadataColumnHandle(column))
+                .collect(toImmutableMap(DeltaLakeColumnHandle::columnName, DeltaLakeColumnHandle::type));
+
+        for (DeltaLakeColumnHandle existingColumn : existingColumns) {
+            if (isMetadataColumnHandle(existingColumn)) {
+                continue;
+            }
+
+            if (!newColumnHandles.containsKey(existingColumn.columnName())) {
```
findinpath, line 3228: Have you taken into consideration as well `ColumnMappingMode.NAME` / `ColumnMappingMode.ID` ?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath: It is worth mentioning that the partition columns may change even if the schema remains the same

```
create or replace table delta.tiny.t9(x integer, y integer, part integer) with (checkpoint_interval = 2, column_mapping_mode='NAME');
```

```
create or replace table delta.tiny.t9(x integer, y integer, part integer) with (checkpoint_interval = 2, column_mapping_mode='NAME', partitioned_by = array['part']);
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath: FWIW, I haven't found any issues while testing locally with those mappings.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3193,6 +3206,37 @@ private void writeCheckpointIfNeeded(
...
+        TableSnapshot snapshot = transactionLogAccess.loadSnapshot(session, transactionLogReader, table, tableLocation, Optional.of(newVersion), credentialsHandle);
+        checkpointWriterManager.writeCheckpoint(session, snapshot, credentialsHandle);
+    }
+
+    private static boolean isSchemaChanged(List<DeltaLakeColumnHandle> existingColumns, List<DeltaLakeColumnHandle> newColumns)
+    {
+        Map<String, Type> newColumnHandles = newColumns.stream()
+                .filter(column -> !isMetadataColumnHandle(column))
+                .collect(toImmutableMap(DeltaLakeColumnHandle::columnName, DeltaLakeColumnHandle::type));
+
```
findinpath, line 3222: column order may change as well (x integer,  y integer ) -> (y integer, x integer)
although this is not something problematic in checkpoint reading.

findinpath (commented):
LGTM % `isSchemaChanged` method implementation.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: Yes, tested it is not required to create a checkpoint in dbx as well

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: added a test to shows that we can read correctly even the partition changed.
dbx did create a checkpoint in this case, but seems we don't need to

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: right, it is same as partition changes, in dbx they think it is necessary to create a checkpoint, but seems it is not necessary in Trino.

chenjian2664: add a case in connector test for it

ebyhr: > in dbx they think it is necessary to create a checkpoint, but seems it is not necessary in Trino.

Can Databricks read the table even if Trino didn't create the checkpoint?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath: The naming of the method may be a bit misleading then.
`isSchemaChanged`

Same note for `DeltaLakeOutputTableHandle#isSchemaChanged`

Consider renaming to `isNewCheckpointFileRequired`

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath: Yes, the old files will be mentioned in `remove` entries and we don't do in the checkpoint for such files any partition pruning - so it is probably safe to omit creating a checkpoint in this case.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath:
```suggestion
                // remove column requires creating a new checkpoint file
```

findinpath (approved):
LGTM % comments

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: Add a test https://github.com/trinodb/trino/pull/27886/changes/854bb235be1fcc5a8c9236ec1fb2b1a108a63ad0 to show case that with current logic Spark can read successfully

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath: Do you feel that `schemaChanged` is more appropriate? I am missing your response on this thread.

chenjian2664:
squashed fixups

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3193,6 +3206,37 @@ private void writeCheckpointIfNeeded(
...
+        for (DeltaLakeColumnHandle existingColumn : existingColumns) {
+            if (isMetadataColumnHandle(existingColumn)) {
+                continue;
+            }
+
+            if (!newColumnHandles.containsKey(existingColumn.columnName())) {
+                // remove column requires creating a new checkpoint file
```
findinpath: remove -> remove / rename

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: thanks, renamed to `isNewCheckpointFileRequired`

ebyhr:
@chenjian2664 There's another corruption issue Spark can't read. Could you confirm the details? 

```sql
CREATE TABLE test (x int, y varchar) WITH (checkpoint_interval = 2, location = 's3://test-bucket/test', deletion_vectors_enabled = true);
INSERT INTO test VALUES (1, 'aa');
INSERT INTO test VALUES (1, 'bb');

CREATE OR REPLACE TABLE test (x int, y varchar) WITH (checkpoint_interval = 2, location = 's3://test-bucket/test');

SELECT * FROM default.test;
-- Trino can read. Spark throws an exception
```

chenjian2664:
The reason is that the spark doesn't write `ProtocolEntry` in this case, and we write the entry without any `FeadFeatures` and `writeFeatures` but with min read/write version - Spark think it is invalid

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1535,9 +1541,11 @@ public DeltaLakeOutputTableHandle beginCreateTable(
         OptionalLong readVersion = OptionalLong.empty();
         ProtocolEntry protocolEntry;
 
+        boolean isSchemaChanged = false;
         if (replaceExistingTable) {
             protocolEntry = protocolEntryForTable(handle.getProtocolEntry().minReaderVersion(), handle.getProtocolEntry().minWriterVersion(), containsTimestampType, tableMetadata.getProperties());
             readVersion = OptionalLong.of(handle.getReadVersion());
+            isSchemaChanged = isNewCheckpointFileRequired(getColumns(handle.getMetadataEntry(), handle.getProtocolEntry()), columnHandles.build());
```
findinpath: pls consider renaming isSchemaChanged as well to reflect the method name change.

chenjian2664:
I find https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features, it means we do wrote the illegal protocol entry.

I am not able to find the behavior of create or replace in the protocol, but I find https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features-for-new-and-existing-tables, it says:
```
When upgrading an existing table to Reader Version 3 and/or Writer Version 7, the client should, on a best effort basis,
 determine which features supported by the original protocol version are used in any historical version of the table, and
 add only used features to reader and/or writer feature sets. The client must assume a feature has been used, unless it 
can prove that the feature is definitely not used in any historical version of the table that is reachable by time travel.
```

combined with https://github.com/delta-io/delta/blob/master/PROTOCOL.md#supported-features, my understanding is that no matter what's the version it is, if the procotol has `readFeatures` or `writeFeatures`, we should inherit it from original version, unless we pretty sure that all the historic data not use the feature at all.

which means we should keep the original read features and write features if we know they were used.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1541,11 +1541,11 @@ public DeltaLakeOutputTableHandle beginCreateTable(
         OptionalLong readVersion = OptionalLong.empty();
         ProtocolEntry protocolEntry;
 
-        boolean isSchemaChanged = false;
+        boolean isisNewCheckpointFileRequired = false;
```
Math-ias: With this name to not conflict with the method name? Can this be `boolean newCheckpointFileRequired` instead?

Math-ias:
Recreating a table with a different schema seems to work in spark without creating a new checkpoint. I'm curious why we need to write a new checkpoint?

```
docker run -it spark:4.0.0-scala2.13-java21-python3-ubuntu /opt/spark/bin/pyspark --packages io.delta:delta-spark_2.13:4.0.0 --conf "spark.jars.ivy=/tmp/.ivy2" --conf "spark.sql.extensions=io.delta.sql.DeltaSparkSessionExtension" --conf "spark.sql.catalog.spark_catalog=org.apache.spark.sql.delta.catalog.DeltaCatalog"
[...]

[creating table]
>>> spark.createDataFrame([("George", "Clooney")],["first","last"]).write.format("delta").mode("overwrite").save("/tmp/my_delta")

[overwriting table]
>>> spark.createDataFrame([("George", "Clooney the 2nd")],["first","last"]).write.format("delta").mode("overwrite").save("/tmp/my_delta")

[data is replaced]
>>> spark.read.format("delta").load("/tmp/my_delta").show()
+------+----------+
| first|      last|
+------+----------+
|George|Clooney II|
+------+----------+

[we can change schema with this overwriteSchema option]
>>> spark.createDataFrame([("George", "Clooney")],["different","fields"]).write.format("delta").mode("overwrite").option("overwriteSchema", "true").save("/tmp/my_delta")
```

This doesn't appear to create a new checkpoint.
```
tree /tmp/my_delta
/tmp/my_delta
├── _delta_log
│  ├── 00000000000000000000.crc
|  ├── 00000000000000000000.json
│  ├── 00000000000000000001.crc
│  ├── 00000000000000000001.json
│  ├── 00000000000000000002.crc
│  ├── 00000000000000000002.json
|  ├── 00000000000000000003.crc
│  ├── 00000000000000000003.json
│  └── _staged_commits
├── part-00000-11c38e5a-4634-473a-a9f8-8e76c6e2aa4e-c000.snappy.parquet
├── part-00000-2656cff1-dff4-4353-ad77-d9e630d30a90-c000.snappy.parquet
├── part-00000-2bbb6ac4-de9c-4492-8564-fb6542d7c949-c000.snappy.parquet
├── part-00000-d3e9c766-1612-409d-a691-a0bf9d131761-c000.snappy.parquet
├── part-00010-04065d3f-e027-48fb-9eac-46c8c2e900f1-c000.snappy.parquet
├── part-00010-5801955f-c966-44bf-87d7-216014718ea0-c000.snappy.parquet
├── part-00010-63e27f30-e312-41d4-9d85-08910c098596-c000.snappy.parquet
└── part-00010-64b91f48-0cb3-4d15-83fb-83dd25076223-c000.snappy.parquet
```

Funnily enough, I could re-upload these files created by spark, and it reads fine in this branch. 🤔 
```
CALL delta.system.register_table(schema_name => 'tpch', table_name => 'testtable', table_location => 's3://test-bucket
/my_delta');
SELECT * FROM delta.tpch.testtable;
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: No, it's my mistaken renamed, should be `isNewCheckpointFileRequired`

chenjian2664:
@Math-ias Thanks for stepping in, according to issue https://github.com/trinodb/trino/issues/27805 description: 
```
The checkpoint is necessary for this bug to be triggered
```

Math-ias:
> @Math-ias Thanks for stepping in, according to issue #27805 description:
> 
> ```
> The checkpoint is necessary for this bug to be triggered
> ```

Oh right! My little experiment was incomplete.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -98,6 +98,24 @@ public void testCreateOrReplaceTableOnTrinoWithSchemaChange()
...
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (ts VARCHAR) " +
+                               "with (location = 's3://" + bucketName + "/databricks-compatibility-test-" + tableName + "', checkpoint_interval = 10)");
+        try {
+            List<Row> expectedRows = performInsert(onTrino(), tableName, 12);
+
+            onTrino().executeQuery("CREATE OR REPLACE TABLE delta.default." + tableName + " AS SELECT CAST(ts AS TIMESTAMP(6)) as ts FROM " + tableName);
+            assertThat(onDelta().executeQuery("SELECT date_format(ts, \"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\") FROM default." + tableName)).containsOnly(expectedRows);
+        }
+        finally {
+            dropDeltaTableWithRetry(tableName);
```
ebyhr, line 115: `dropDeltaTableWithRetry` method exists for Glue environment. We can use the normal DROP TABLE statement here because this test runs only on DELTA_LAKE_OSS (= Spark & HMS).

PR #27888, opened 2026-01-09
Simplify tests using `newTrinoTable` in `TestIcebergV2`

chenjian2664:
Using `newTrinoTable` is also more robust for correctly creating and cleaning up test tables.

## Release notes

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
@@ -183,23 +183,21 @@ public class TestIcebergV2
     @Test
     public void testSettingFormatVersion()
     {
-        String tableName = "test_seting_format_version_" + randomNameSuffix();
-        assertUpdate("CREATE TABLE " + tableName + " WITH (format_version = 2) AS SELECT * FROM tpch.tiny.nation", 25);
-        assertThat(formatVersion(loadTable(tableName))).isEqualTo(2);
-        assertUpdate("DROP TABLE " + tableName);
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -182,23 +182,21 @@ protected QueryRunner createQueryRunner()
     @Test
     public void testSettingFormatVersion()
     {
-        String tableName = "test_seting_format_version_" + randomNameSuffix();
-        assertUpdate("CREATE TABLE " + tableName + " WITH (format_version = 2) AS SELECT * FROM tpch.tiny.nation", 25);
-        assertThat(formatVersion(loadTable(tableName))).isEqualTo(2);
-        assertUpdate("DROP TABLE " + tableName);
+        try (TestTable table = newTrinoTable("test_setting_format_version_", " WITH (format_version = 2) AS SELECT * FROM tpch.tiny.nation")) {
```
ebyhr: A leading space is redundant. Same for others.
```suggestion
        try (TestTable table = newTrinoTable("test_setting_format_version_", "WITH (format_version = 2) AS SELECT * FROM tpch.tiny.nation")) {
```

ebyhr (commented):
This change is unnecessary because TestIcebergV2 doesn't run in SaaS. That said, I'm okay with it.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
chenjian2664: thanks, done

ebyhr (approved):

PR #27893, opened 2026-01-11
[Iceberg v3]  Geometry and Geography types

dain:
## Description

This PR is based on  #27881, and only the last commit is relevent to Iceberg.

### Geometry

`geometry(crs)` → Trino `Geometry`

- CRS is converted to SRID (e.g., `EPSG:3857` → SRID 3857, `OGC:CRS84` → SRID 4326)
- Default CRS (null) is treated as OGC:CRS84 (SRID 4326)
- On read: SRID is injected into the WKB to produce EWKB
- On write: SRID is validated against column CRS, then stripped from EWKB to produce WKB (SRID 0 is accepted for any CRS)

### Geography

`geography(crs, algorithm)` → Trino `SphericalGeography`

- CRS must be WGS84 (OGC:CRS84, EPSG:4326, or null for default)
- Algorithm must be SPHERICAL (or null for default)
- Non-WGS84 CRS is not supported
- Non-spherical algorithms (e.g., VINCENTY) are not supported

## Release notes

```markdown
## Iceberg
* Add support for reading and writing `geometry` and `geography` column types. ({issue}`issuenumber`)
```

### DIFF
--- a/lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
+++ b/lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
@@ -24,9 +24,11 @@ import org.locationtech.jts.io.ParseException;
 import org.locationtech.jts.io.WKBReader;
 import org.locationtech.jts.io.WKBWriter;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Verify.verify;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
 import static java.lang.String.format;
+import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -36,6 +38,10 @@ import static java.util.Objects.requireNonNull;
 public final class JtsGeometrySerde
 {
     private static final GeometryFactory GEOMETRY_FACTORY = new GeometryFactory();
+    public static final int OGC_CRS84_SRID = 4326;
+
+    // EWKB flag for SRID presence (bit 29)
+    private static final int EWKB_SRID_FLAG = 0x20000000;
 
     // WKB type codes (2D)
     private static final int WKB_POINT = 1;
@@ -183,5 +189,109 @@ public final class JtsGeometrySerde
     {
         result.setSRID(validateAndGetSrid(left, right));
         return serialize(result);
+    }
+
+    /**
+     * Extract SRID from EWKB without full parsing.
+     * Returns 0 if no SRID is embedded.
+     */
+    public static int extractSrid(Slice ewkb)
+    {
+        if (ewkb.length() < 9) {
+            return 0;
+        }
+        boolean bigEndian = ewkb.getByte(0) == 0;
+        int type = ewkb.getInt(1);
+        if (bigEndian) {
+            type = Integer.reverseBytes(type);
+        }
+        if ((type & EWKB_SRID_FLAG) == 0) {
+            return 0;
+        }
+        int srid = ewkb.getInt(5);
+        if (bigEndian) {
+            srid = Integer.reverseBytes(srid);
+        }
+        return srid;
+    }
+
+    /**
+     * Strip SRID from EWKB to produce standard WKB.
+     * If the input is already standard WKB (no SRID flag), returns it unchanged.
+     */
+    public static Slice ewkbToWkb(Slice ewkb)
+    {
+        if (ewkb.length() < 9) {
+            return ewkb;
+        }
+        boolean bigEndian = ewkb.getByte(0) == 0;
+        int type = ewkb.getInt(1);
+        if (bigEndian) {
+            type = Integer.reverseBytes(type);
+        }
+        if ((type & EWKB_SRID_FLAG) == 0) {
+            return ewkb;
+        }
+
+        // Strip SRID flag and 4 SRID bytes
+        int newType = type & ~EWKB_SRID_FLAG;
+        Slice wkb = Slices.allocate(ewkb.length() - 4);
+        wkb.setByte(0, ewkb.getByte(0)); // endianness
+        wkb.setInt(1, bigEndian ? Integer.reverseBytes(newType) : newType);
+        wkb.setBytes(5, ewkb, 9, ewkb.length() - 9); // geometry data
+        return wkb;
+    }
+
+    /**
+     * Convert a CRS string to an SRID integer.
+     * Supports formats:
+     * - "EPSG:XXXX" → XXXX
+     * - "OGC:CRS84" or "CRS84" → 4326 (WGS84)
+     */
+    public static int crsToSrid(String crs)
+    {
+        if (crs == null || crs.isEmpty()) {
+            return 0;
+        }
+        String upperCrs = crs.toUpperCase(ENGLISH);
+        if (upperCrs.equals("OGC:CRS84") || upperCrs.equals("CRS84")) {
+            return OGC_CRS84_SRID; // WGS84
+        }
+        if (upperCrs.startsWith("EPSG:")) {
+            try {
+                int srid = Integer.parseInt(crs.substring(5));
+                if (srid <= 0) {
+                    throw new IllegalArgumentException("Invalid EPSG code: " + crs);
+                }
+                return srid;
+            }
+            catch (NumberFormatException e) {
+                throw new IllegalArgumentException("Invalid EPSG code: " + crs);
+            }
+        }
+        throw new IllegalArgumentException("Unsupported CRS format: " + crs);
+    }
+
+    /**
+     * Inject SRID into WKB to produce EWKB.
+     */
+    public static Slice wkbToEwkb(Slice wkb, int srid)
+    {
+        checkArgument(wkb.length() >= 5, "WKB too short");
+        boolean bigEndian = wkb.getByte(0) == 0;
+        int type = wkb.getInt(1);
+        if (bigEndian) {
+            type = Integer.reverseBytes(type);
+        }
+        checkArgument((type & EWKB_SRID_FLAG) == 0, "Input already has SRID flag set (expected WKB, got EWKB)");
+
+        // Add SRID flag
+        int newType = type | EWKB_SRID_FLAG;
+        Slice ewkb = Slices.allocate(wkb.length() + 4);
+        ewkb.setByte(0, wkb.getByte(0)); // endianness
+        ewkb.setInt(1, bigEndian ? Integer.reverseBytes(newType) : newType);
+        ewkb.setInt(5, bigEndian ? Integer.reverseBytes(srid) : srid);
+        ewkb.setBytes(9, wkb, 5, wkb.length() - 5); // geometry data
+        return ewkb;
     }
 }
--- a/lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
+++ b/lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
@@ -27,6 +27,7 @@ import org.locationtech.jts.io.WKBWriter;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Verify.verify;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static java.lang.String.format;
 import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
@@ -42,6 +43,9 @@ public final class JtsGeometrySerde
 
     // EWKB flag for SRID presence (bit 29)
     private static final int EWKB_SRID_FLAG = 0x20000000;
+    private static final int EWKB_M_FLAG = 0x40000000;
+    private static final int EWKB_Z_FLAG = 0x80000000;
+    private static final String UNSUPPORTED_COORDINATE_DIMENSIONS_MESSAGE = "Geospatial values with more than 2 coordinate dimensions are not supported";
 
     // WKB type codes (2D)
     private static final int WKB_POINT = 1;
@@ -226,10 +230,8 @@ public final class JtsGeometrySerde
             return ewkb;
         }
         boolean bigEndian = isBigEndian(ewkb.getByte(0));
-        int type = ewkb.getInt(1);
-        if (bigEndian) {
-            type = Integer.reverseBytes(type);
-        }
+        int type = readType(ewkb, bigEndian);
+        verifySupportedCoordinateDimensions(type);
         if ((type & EWKB_SRID_FLAG) == 0) {
             return ewkb;
         }
@@ -280,11 +282,9 @@ public final class JtsGeometrySerde
     {
         checkArgument(wkb.length() >= 5, "WKB too short");
         boolean bigEndian = isBigEndian(wkb.getByte(0));
-        int type = wkb.getInt(1);
-        if (bigEndian) {
-            type = Integer.reverseBytes(type);
-        }
+        int type = readType(wkb, bigEndian);
         checkArgument((type & EWKB_SRID_FLAG) == 0, "Input already has SRID flag set (expected WKB, got EWKB)");
+        verifySupportedCoordinateDimensions(type);
 
         // Add SRID flag
         int newType = type | EWKB_SRID_FLAG;
@@ -300,5 +300,26 @@ public final class JtsGeometrySerde
     {
         checkArgument(endianness == 0 || endianness == 1, "invalid WKB endianness: %s", endianness);
         return endianness == 0;
+    }
+
+    private static int readType(Slice wkb, boolean bigEndian)
+    {
+        int type = wkb.getInt(1);
+        if (bigEndian) {
+            type = Integer.reverseBytes(type);
+        }
+        return type;
+    }
+
+    private static void verifySupportedCoordinateDimensions(int type)
+    {
+        if ((type & (EWKB_Z_FLAG | EWKB_M_FLAG)) != 0 || (type & 0xFFFF) > 999) {
+            throw unsupportedCoordinateDimensions();
+        }
+    }
+
+    private static TrinoException unsupportedCoordinateDimensions()
+    {
+        return new TrinoException(NOT_SUPPORTED, UNSUPPORTED_COORDINATE_DIMENSIONS_MESSAGE);
     }
 }
--- a/lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
+++ b/lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
@@ -24,9 +24,11 @@ import org.locationtech.jts.io.ParseException;
 import org.locationtech.jts.io.WKBReader;
 import org.locationtech.jts.io.WKBWriter;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Verify.verify;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
 import static java.lang.String.format;
+import static java.util.Locale.ENGLISH;
 import static java.util.Objects.requireNonNull;
 
 /**
@@ -36,6 +38,10 @@ import static java.util.Objects.requireNonNull;
 public final class JtsGeometrySerde
 {
     private static final GeometryFactory GEOMETRY_FACTORY = new GeometryFactory();
+    public static final int OGC_CRS84_SRID = 4326;
+
+    // EWKB flag for SRID presence (bit 29)
+    private static final int EWKB_SRID_FLAG = 0x20000000;
 
     // WKB type codes (2D)
     private static final int WKB_POINT = 1;
@@ -97,9 +103,9 @@ public final class JtsGeometrySerde
         // WKB format: [1 byte endianness] [4 bytes type]
         // endianness: 0 = big endian (XDR), 1 = little endian (NDR)
         byte endianness = shape.getByte(0);
-        verify(endianness == 0 || endianness == 1, "invalid WKB endianness: %s", endianness);
+        boolean bigEndian = isBigEndian(endianness);
         int wkbType;
-        if (endianness == 0) {
+        if (bigEndian) {
             // Big endian - read bytes manually
             wkbType = ((shape.getByte(1) & 0xFF) << 24) |
                     ((shape.getByte(2) & 0xFF) << 16) |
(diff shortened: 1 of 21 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/aggregation/IcebergThetaSketchForStats.java
```diff
@@ -59,6 +59,7 @@ public static void input(@TypeParameter("T") Type type, @AggregationState DataSk
         Object trinoValue = readNativeValue(type, block, index);
         org.apache.iceberg.types.Type icebergType = toIcebergTypeForNewColumn(type, new AtomicInteger(1));
         Object icebergValue = convertTrinoValueToIceberg(type, trinoValue);
+
```
wendigo: Revert

lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
```diff
@@ -184,4 +189,104 @@ public static Slice serializeBinaryOp(Geometry result, Geometry left, Geometry r
...
+        }
+        boolean bigEndian = ewkb.getByte(0) == 0;
+        int type = ewkb.getInt(1);
+        if (bigEndian) {
+            type = Integer.reverseBytes(type);
+        }
+        if ((type & EWKB_SRID_FLAG) == 0) {
+            return 0;
+        }
+        int srid = ewkb.getInt(5);
```
wendigo, line 216: can we do a single getLong instead?

lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
```diff
@@ -184,4 +189,104 @@ public static Slice serializeBinaryOp(Geometry result, Geometry left, Geometry r
...
+     */
+    public static int crsToSrid(String crs)
+    {
+        if (crs == null || crs.isEmpty()) {
+            return 0;
+        }
+        if (crs.equalsIgnoreCase("OGC:CRS84") || crs.equalsIgnoreCase("CRS84")) {
+            return 4326; // WGS84
+        }
+        if (crs.toUpperCase(Locale.ROOT).startsWith("EPSG:")) {
```
wendigo: ENGLISH ?

wendigo: can we upper case after null check and compare with .equals() ?

lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
```diff
@@ -184,4 +189,104 @@ public static Slice serializeBinaryOp(Geometry result, Geometry left, Geometry r
...
+        }
+        throw new IllegalArgumentException("Unsupported CRS format: " + crs);
+    }
+
+    /**
+     * Inject SRID into WKB to produce EWKB.
+     */
+    public static Slice wkbToEwkb(Slice wkb, int srid)
+    {
+        if (wkb.length() < 5) {
```
wendigo: nit: `checkArgument`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -183,6 +190,20 @@ public IcebergPageSink(
                 .map(IcebergColumnHandle::getType)
                 .collect(toImmutableList());
 
+        // Build mapping of geometry column indexes to their target SRIDs
+        Map<Integer, Integer> geometrySrids = new HashMap<>();
```
wendigo: use ImmutableMap builder

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -327,6 +351,56 @@ private void writePage(Page page)
...
+        VariableWidthBlockBuilder builder = new VariableWidthBlockBuilder(null, positionCount, positionCount * 32);
+
+        for (int position = 0; position < positionCount; position++) {
+            if (block.isNull(position)) {
+                builder.appendNull();
+                continue;
+            }
+
+            Slice ewkb = GEOMETRY.getSlice(block, position);
+            int sourceSrid = JtsGeometrySerde.extractSrid(ewkb);
```
wendigo: nit: static import `extractSrid`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -327,6 +351,56 @@ private void writePage(Page page)
...
+
+            // Validate SRID: fail only if both source and target are non-zero and different
+            if (sourceSrid != 0 && targetSrid != 0 && sourceSrid != targetSrid) {
+                throw new TrinoException(INVALID_FUNCTION_ARGUMENT,
+                        "SRID mismatch: cannot write geometry with SRID %d into column %d with SRID %d".formatted(sourceSrid, columnIndex, targetSrid));
+            }
+
+            // Strip SRID from EWKB to produce standard WKB for storage
+            Slice wkb = JtsGeometrySerde.ewkbToWkb(ewkb);
```
wendigo: nit: static import `ewkbToWkb`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1659,6 +1686,73 @@ private static TrinoException handleException(ParquetDataSourceId dataSourceId,
...
+
+        // Find the field in the schema to get the CRS
+        Types.NestedField field = tableSchema.findField(column.getId());
+        if (field == null || !(field.type() instanceof Types.GeometryType geometryType)) {
+            return Optional.empty();
+        }
+
+        String crs = geometryType.crs();
+        // Iceberg uses null CRS to represent the default (OGC:CRS84 = SRID 4326)
+        int srid = (crs == null) ? 4326 : JtsGeometrySerde.crsToSrid(crs);
```
wendigo: can we have const for this 4326 magic number?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -1659,6 +1686,73 @@ private static TrinoException handleException(ParquetDataSourceId dataSourceId,
...
+        int positionCount = block.getPositionCount();
+        Slice[] results = new Slice[positionCount];
+        boolean[] nulls = new boolean[positionCount];
+
+        for (int i = 0; i < positionCount; i++) {
+            if (block.isNull(i)) {
+                nulls[i] = true;
+            }
+            else {
+                results[i] = JtsGeometrySerde.wkbToEwkb(variableWidthBlock.getSlice(i), srid);
```
wendigo: nit: static import wkbToEwkb

dain:
@coderabbitai review

lib/trino-geospatial-toolkit/src/main/java/io/trino/geospatial/serde/JtsGeometrySerde.java
dain: It turned out more complex.

dain:
@coderabbitai review

dain:
@coderabbitai review

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -107,6 +115,7 @@ protected QueryRunner createQueryRunner()
         queryRunner.installPlugin(new TpchPlugin());
         queryRunner.createCatalog("tpch", "tpch");
 
+        queryRunner.installPlugin(new GeoPlugin());
```
findinpath, line 121: Can you pls add documentation in `iceberg.md` on how to enable using geospatial types?

I was trying IcebergQueryRunner without this line and stumbled in 

```
create table iceberg.tpch.testgeo(id INTEGER, geom geometry) with (format_version = 3);
Query 20260327_053017_00017_2q373 failed: line 1:47: Unknown type 'geometry' for column 'geom'
io.trino.spi.TrinoException: line 1:47: Unknown type 'geometry' for column 'geom'
	at io.trino.execution.CreateTableTask.internalExecute(CreateTableTask.java:184)
	at io.trino.execution.CreateTableTask.execute(CreateTableTask.java:129)
```

raunaqmorarka:
Can we separate the commits for geometry and geography ?
Now that parquet spec has added these types https://github.com/apache/parquet-format/blob/master/Geospatial.md#logical-types https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry, can we use those logical annotations ?

dain:
> Can we separate the commits for geometry and geography ?

@raunaqmorarka  Why? Seems like a bunch of what for not much win.  This is not a big PR.

> Now that parquet spec has added these types https://github.com/apache/parquet-format/blob/master/Geospatial.md#logical-types https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry, can we use those logical annotations ?

I can take a look.

dain:
> > Now that parquet spec has added these types https://github.com/apache/parquet-format/blob/master/Geospatial.md#logical-types https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry, can we use those logical annotations ?
> 
> I can take a look.

I took a look, and this is not as simple as adding a hard-coded logical annotation. The geometry/geography parameters would need to be carried through the Parquet schema annotations, and as far as I can tell that would require Parquet/Iceberg library support updates on our side to propagate those parameters end-to-end. So I think this is better handled as follow-up work in the Parquet schema conversion stack rather than folded into this PR.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
dain: I thought about this for a while, and I don't actually think we should put this in the docs. The default packaging ships with the geo plugin installed, and if we tell people they have to have it installed, I think it will cause more confusion than it removes.

raunaqmorarka:
> > Can we separate the commits for geometry and geography ?
> 
> @raunaqmorarka Why? Seems like a bunch of what for not much win. This is not a big PR.

We usually make separate commits for independent pieces of functionality. It helps with review, nicer git history etc. Unless these functionalities cannot exist separately, I think it makes sense for the commits to be separate. Hopefully claude/codex can do this without too much fuss.

raunaqmorarka:
> > > Now that parquet spec has added these types https://github.com/apache/parquet-format/blob/master/Geospatial.md#logical-types https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry, can we use those logical annotations ?
> > 
> > 
> > I can take a look.
> 
> I took a look, and this is not as simple as adding a hard-coded logical annotation. The geometry/geography parameters would need to be carried through the Parquet schema annotations, and as far as I can tell that would require Parquet/Iceberg library support updates on our side to propagate those parameters end-to-end. So I think this is better handled as follow-up work in the Parquet schema conversion stack rather than folded into this PR.

I'm mainly worried about cross-engine compatibility. I see that Databricks/Spark doesn't support it yet, but SnowFlake does https://docs.snowflake.com/en/en/user-guide/tables-iceberg-data-types#label-tables-iceberg-v3-data-types
It would be nice if we can do this in a way that aligns with other iceberg/parquet reader/writers.

wendigo:
Rebased to solve conflicts

plugin/trino-iceberg/pom.xml
```diff
@@ -127,6 +127,16 @@
             <artifactId>trino-filesystem-s3</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-geospatial</artifactId>
```
wendigo: This seems wrong - trino-geospatial is a Trino plugin. Should we move classes that we import to a toolkit or to the SPI (I believe that these are the types that we import)?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -1628,6 +1637,254 @@ void testOrcTimestampNanoFiltering()
...
+    @Test
+    void testGeographyRoundTrip()
+    {
+        for (String format : List.of("PARQUET", "ORC", "AVRO")) {
+            try (TestTable table = newTrinoTable("test_geography_roundtrip_" + format.toLowerCase(Locale.ROOT) + "_",
+                    "(id INTEGER, geog sphericalgeography) WITH (format = '" + format + "', format_version = 3)")) {
+                assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, to_spherical_geography(ST_Point(-122.4194, 37.7749)))", 1);
+
+                assertThat(query("SELECT id FROM " + table.getName()))
+                        .matches("VALUES 1");
```
wendigo: This test is incomplete. Doesn't validate the geography

dain:
> > > > Now that parquet spec has added these types https://github.com/apache/parquet-format/blob/master/Geospatial.md#logical-types https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#geometry, can we use those logical annotations ?
> > > 
> > > 
> > > I can take a look.
> > 
> > 
> > I took a look, and this is not as simple as adding a hard-coded logical annotation. The geometry/geography parameters would need to be carried through the Parquet schema annotations, and as far as I can tell that would require Parquet/Iceberg library support updates on our side to propagate those parameters end-to-end. So I think this is better handled as follow-up work in the Parquet schema conversion stack rather than folded into this PR.
> 
> I'm mainly worried about cross-engine compatibility. I see that Databricks/Spark doesn't support it yet, but SnowFlake does https://docs.snowflake.com/en/en/user-guide/tables-iceberg-data-types#label-tables-iceberg-v3-data-types It would be nice if we can do this in a way that aligns with other iceberg/parquet reader/writers.

I added a commit that does this.

raunaqmorarka:
@coderabbitai review

raunaqmorarka:
@coderabbitai full review

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/OrcTypeConverter.java
```diff
@@ -88,7 +88,7 @@ private static List<OrcType> toOrcType(int nextFieldTypeIndex, Type type, Map<St
                 yield ImmutableList.of(new OrcType(timestampKind, ImmutableList.of(), ImmutableList.of(), Optional.empty(), Optional.empty(), Optional.empty(), attributes));
             }
             case STRING -> ImmutableList.of(new OrcType(OrcTypeKind.STRING, ImmutableList.of(), ImmutableList.of(), Optional.empty(), Optional.empty(), Optional.empty(), attributes));
-            case FIXED, BINARY -> ImmutableList.of(new OrcType(OrcTypeKind.BINARY, ImmutableList.of(), ImmutableList.of(), Optional.empty(), Optional.empty(), Optional.empty(), attributes));
+            case FIXED, BINARY, GEOMETRY, GEOGRAPHY -> ImmutableList.of(new OrcType(OrcTypeKind.BINARY, ImmutableList.of(), ImmutableList.of(), Optional.empty(), Optional.empty(), Optional.empty(), attributes));
```
raunaqmorarka: Should this be doing
```
.put(ICEBERG_BINARY_TYPE, "GEOMETRY/GEOGRAPHY")
```
?

lib/trino-parquet/src/main/java/io/trino/parquet/ParquetMetadataConverter.java
raunaqmorarka, line 551: Parquet spec is requiring these new tyes to have no min/max stats, I think we have to return SortOrder.UNKNOWN for Geometry/GeographyLogicalTypeAnnotation

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -327,6 +359,121 @@ private void writePage(Page page)
...
+    }
+
+    private static boolean containsGeometry(org.apache.iceberg.types.Type type)
+    {
+        return switch (type.typeId()) {
+            case GEOMETRY -> true;
+            case LIST -> containsGeometry(type.asListType().elementType());
+            case MAP -> containsGeometry(type.asMapType().keyType()) || containsGeometry(type.asMapType().valueType());
+            case STRUCT -> type.asStructType().fields().stream().anyMatch(field -> containsGeometry(field.type()));
+            default -> false;
```
raunaqmorarka, line 393: > [!NOTE]
> This is Claude-generated analysis — please verify the reproducer and reasoning before acting on it.

EWKB bytes can leak into GEOGRAPHY columns on write, producing files that aren't valid ISO WKB.

This switch only matches `case GEOMETRY`, and `appendTransformedGeometryValueForWrite` only handles `Types.GeometryType`, so `transformGeometryColumnsForWrite` skips GEOGRAPHY columns entirely and the block bytes flow to the writer unchanged.

That matters because SPHERICAL_GEOGRAPHY block bytes are produced by `JtsGeometrySerde.serialize` (`WKBWriter(2, true)`), which sets the EWKB SRID flag (`0x20000000`) when the JTS Geometry's SRID is non-zero. The read path here always seeds a non-zero SRID into geometry blocks (`appendTransformedGeometryValueForRead` → `wkbToEwkb(slice, getGeometrySrid(...))`, defaulting to `OGC_CRS84_SRID = 4326`), so it's easy to construct a SPHERICAL_GEOGRAPHY value carrying that SRID and have it written verbatim.

Reproducer using only standard Trino SQL:

    CREATE TABLE iceberg.s.geom_t (g geometry) WITH (format_version = 3);
    INSERT INTO iceberg.s.geom_t VALUES (ST_GeometryFromText('POINT(1 2)'));

    CREATE TABLE iceberg.s.geog_t (g geography) WITH (format_version = 3);
    INSERT INTO iceberg.s.geog_t
      SELECT to_spherical_geography(g) FROM iceberg.s.geom_t;

The Parquet column backing `geog_t` then contains bytes like `01 01 00 00 20 ...` — a Point with the EWKB SRID flag set and an SRID int prefix. A spec-conforming reader hits an unknown WKB type code on the very first record.

Suggested fix: extend both `containsGeometry(org.apache.iceberg.types.Type)` switches (here and in `IcebergPageSourceProvider`) to also match `GEOGRAPHY`, and add a `Types.GeographyType` branch to `appendTransformedGeometryValueForWrite` that strips the SRID via `ewkbToWkb`, symmetric to the GEOMETRY branch. The read-side counterpart should mirror the GEOMETRY path so a registered `geography(EPSG:NNN)` table round-trips its SRID into the block. A test along the lines of the reproducer (assert the on-disk bytes don't have the EWKB flag bit, or read with parquet-mr / pyiceberg and confirm WKB parses) would catch this and prevent regressions.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -327,6 +359,121 @@ private void writePage(Page page)
...
+
+        if (icebergType instanceof Types.GeometryType geometryType) {
+            Slice ewkb = trinoType.getSlice(block, position);
+            int sourceSrid = extractSrid(ewkb);
+            int targetSrid = getGeometrySrid(geometryType);
+
+            // Validate SRID: fail only if both source and target are non-zero and different
+            if (sourceSrid != 0 && targetSrid != 0 && sourceSrid != targetSrid) {
+                throw new TrinoException(INVALID_FUNCTION_ARGUMENT,
+                        "SRID mismatch: cannot write geometry with SRID %d into column %d with SRID %d".formatted(sourceSrid, columnIndex, targetSrid));
```
raunaqmorarka: > [!NOTE]
> This is Claude-generated analysis — verified by running a test against this PR branch (results below).

The validation at line 417 treats `sourceSrid == 0` as a wildcard, so a value with no embedded SRID can be inserted into a column declared with any non-default CRS without error. On read, `getGeometrySrid` injects the column's CRS-derived SRID, so the value the user gets back is silently re-tagged with a CRS the original data didn't have. Easy to do accidentally with `ST_Point(...)` / `ST_GeometryFromText(...)` (both produce SRID=0) against a column registered with e.g. EPSG:3857.

Verified against this PR's branch (`52e283b`) with this test added to `TestIcebergV3` — it passes, demonstrating the bug:

    @Test
    void testWriteSridZeroIntoCustomCrsColumn_demonstratesSilentReassignment()
    {
        String hadoopTableName = "hadoop_srid_zero_into_custom_" + randomNameSuffix();
        Path hadoopTableLocation = dataDirectory.resolve(hadoopTableName);

        Schema schema = new Schema(
                Types.NestedField.optional(1, "id", Types.IntegerType.get()),
                Types.NestedField.optional(2, "geom", Types.GeometryType.of("EPSG:3857")));

        new HadoopTables(new Configuration(false)).create(
                schema,
                PartitionSpec.unpartitioned(),
                SortOrder.unsorted(),
                ImmutableMap.of("format-version", "3", "write.format.default", "PARQUET"),
                hadoopTableLocation.toString());

        String registered = "registered_srid_zero_into_custom_" + randomNameSuffix();
        assertUpdate("CALL system.register_table(CURRENT_SCHEMA, '%s', '%s')"
                .formatted(registered, hadoopTableLocation));

        // ST_Point produces SRID=0; column is EPSG:3857
        assertUpdate("INSERT INTO " + registered + " VALUES (1, ST_Point(-122.4, 37.7))", 1);

        // The point comes back tagged with SRID=3857 even though we wrote a SRID=0 value.
        assertThat(query("SELECT ST_SRID(geom) FROM " + registered))
                .matches("VALUES 3857");

        assertUpdate("DROP TABLE " + registered);
    }

The user wrote WGS84-looking lon/lat values; downstream tooling now reads them as web-mercator metres. There's no error, no warning, and no way for a user to spot it in the SQL output.

Suggested fix: tighten the predicate to fail unless `sourceSrid == targetSrid`, treating SRID=0 as a real (mismatching) value rather than a wildcard. If users still want to opt into "trust me, my SRID-0 data is in the column's CRS", make that an explicit cast / function call (e.g. require the user to call `ST_SetSRID(g, 3857)` themselves), not the implicit default.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSink.java
```diff
@@ -327,6 +359,121 @@ private void writePage(Page page)
...
+    }
+
+    private static void appendTransformedGeometryValueForWrite(Type trinoType, org.apache.iceberg.types.Type icebergType, Block block, int position, BlockBuilder builder, int columnIndex)
+    {
+        if (block.isNull(position)) {
+            builder.appendNull();
+            return;
+        }
+
+        if (icebergType instanceof Types.GeometryType geometryType) {
```
raunaqmorarka: > [!NOTE]
> This is Claude-generated analysis — verified by running a test against this PR branch (results below).

The bytes flowing through this geometry-write branch have already lost Z/M dimensions before they reach the sink. `JtsGeometrySerde.serialize` is hard-coded to `new WKBWriter(2, true)` (line 80 of `lib/trino-geospatial-toolkit/.../JtsGeometrySerde.java`), so any Trino-side serialization path — `ST_GeometryFromText('POINT Z (...)')`, `ST_Point` plus a future Z constructor, casts, anything that round-trips through this type's `writeObject` — silently truncates to 2D. By the time `trinoType.getSlice(block, position)` runs above, Z is gone.

Verified against this PR's branch (`52e283b`) with this test added to `TestIcebergV3` — it fails, demonstrating the loss:

    @Test
    void testWriteThreeDimensionalGeometry_demonstratesZDrop()
    {
        try (TestTable table = newTrinoTable("test_z_drop_",
                "(geom geometry) WITH (format = 'PARQUET', format_version = 3)")) {
            assertUpdate("INSERT INTO " + table.getName() + " VALUES (ST_GeometryFromText('POINT Z (1 2 3)'))", 1);

            // Expected POINT Z (1 2 3); actual POINT (1 2)  — Z silently dropped at insert time.
            assertThat(query("SELECT ST_AsText(geom) FROM " + table.getName()))
                    .matches("VALUES VARCHAR 'POINT Z (1 2 3)'");
        }
    }

Test output:

    [Rows for query [SELECT ST_AsText(geom) FROM test_z_drop_...]]
    Expecting actual:
      (POINT (1 2))
    to contain exactly in any order:
      [(POINT Z (1 2 3))]

Read fidelity is fine — bytes from XYZ files written by other engines pass through `wkbToEwkb` unchanged (it only flips the type-code bit), so external XYZ data round-trips faithfully through `SELECT g`. The loss is strictly on writes Trino itself produces.

Suggested fix: derive the WKBWriter's `outputDimension` from the source `Geometry`'s coordinate dimension (`geometry.getCoordinate().getDimension()` or check Z/M presence) instead of hard-coding 2. Iceberg v3 advertises Z/M support, so silently demoting users' 3D data to 2D is a real surprise. At minimum, raising an error when the sink receives a WKB type code outside `1..7` would convert silent loss into a clear failure.

dain: Good point. I'm happy to support preserving Z, but this looks larger than an Iceberg-only fix. Trino’s geospatial serialization and semantics have historically been 2D-oriented, so changing this cleanly would likely need a broader Trino geospatial effort rather than a local change in the Iceberg connector. I think it’s better handled as follow-up work so we can make sure the behavior is consistent across Trino, not just in this connector path.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/util/OrcTypeConverter.java
dain: I took a look at this, and I am not seeing an upstream Iceberg ORC convention for geometry/geography-specific attributes here yet. Icebergs ORC schema utility currently only uses `iceberg.binary-type` for the existing binary-like cases such as UUID/fixed/binary, and does not appear to define geometry/geography-specific handling. Because of that, I would prefer not to invent a Trino-only ORC encoding hint in this PR without a stronger upstream/spec signal.  Or said another way, I don't think geo is supported for Iceberg ORC yet.

raunaqmorarka: I found this in the iceberg spec https://iceberg.apache.org/spec/#orc

Screenshot 2026-05-30 at 11 31 12 AM

dain: Awesome!  I added that

raunaqmorarka (approved):

PR #27900, opened 2026-01-12
Documentation the possible lossy of the Oracle FLOAT type

chenjian2664:
## Release notes

```markdown
## Section
```

### DIFF
--- a/docs/src/main/sphinx/connector/oracle.md
+++ b/docs/src/main/sphinx/connector/oracle.md
@@ -176,7 +176,9 @@ Trino data type mapping:
   - See [](oracle-number-mapping)
 * - `FLOAT[(p)]`
   - `DOUBLE`
-  -
+  - When `p` exceeds 53, numeric values may be subject to precision loss. 
+    The default precision of the [FLOAT data type](https://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj27281.html) 
+    in Oracle is 53.
 * - `BINARY_FLOAT`
   - `REAL`
   -


docs/src/main/sphinx/connector/oracle.md
```diff
@@ -176,7 +176,8 @@ Trino data type mapping:
   - See [](oracle-number-mapping)
 * - `FLOAT[(p)]`
   - `DOUBLE`
-  -
+  - When p exceeds 53, numeric values may be subject to precision loss. See Oracle
```
mosabua: Is that even valid in Trino to set P that high? Also would that not be a problem with other databases?

And in terms of links - https://developers.google.com/style/cross-references#descriptive-link-text

mosabua (dismissed):

docs/src/main/sphinx/connector/oracle.md
chenjian2664: > Is that even valid in Trino to set P that high? Also would that not be a problem with other databases?

The oracle can set high itself, while we are mapping to use DOUBLE to interpret it, it is a know behavior that possibliy cause lossy.

ebyhr: p → precision or ``` `p` ```

docs/src/main/sphinx/connector/oracle.md
```diff
@@ -176,7 +176,8 @@ Trino data type mapping:
   - See [](oracle-number-mapping)
 * - `FLOAT[(p)]`
   - `DOUBLE`
-  -
+  - When p exceeds 53, numeric values may be subject to precision loss. See Oracle 
+    documentation for details: [FLOAT data type](https://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj27281.html)
```
ebyhr: I'm not sure what you're trying to show with this link. Maybe:

```
The default precision of the [FLOAT data type](https://docs.oracle.com/javadb/10.6.2.1/ref/rrefsqlj27281.html) in Oracle is 53.
```

ebyhr (approved):

mosabua:
In the future please keep commit message and PR title aligned if that works

Also maybe

Document lossiness with high precision for the Oracle FLOAT type

ebyhr:
> In the future please keep commit message and PR title aligned if that works

Why? There's no need to do that. We can use a longer PR title to provide more context, while keeping commit titles limited to 50 characters.

mosabua:
Sure .. but if you fix grammar in the commit message it would be good to also fix it in the PR title .. often that means they can be consistent.

Of course that only applies to PRs with one commit. PR titles can be longer but often it will work fine to keep it in sync and then its easier .. you can always use the description and also the commit body to provide more info.

Its not a rule or anything .. just a recommendation to make it easier for you and for reviewers.

Issue #21725, opened 2024-04-26
Flaky `TestDeltaLakeLocalConcurrentWritesTest.testConcurrentInsertsSelectingFromTheSameVersionedTable`


ebyhr:
```
Error:  io.trino.plugin.deltalake.TestDeltaLakeLocalConcurrentWritesTest.testConcurrentInsertsSelectingFromTheSameVersionedTable -- Time elapsed: 3.660 s <<< ERROR!
io.trino.testing.QueryFailedException: Failed to write Delta Lake transaction log entry
	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:133)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:501)
	at io.trino.testing.DistributedQueryRunner.execute(DistributedQueryRunner.java:484)
	at io.trino.testing.QueryRunner.execute(QueryRunner.java:82)
	at io.trino.plugin.deltalake.TestDeltaLakeLocalConcurrentWritesTest.lambda$testConcurrentInsertsSelectingFromTheSameVersionedTable$9(TestDeltaLakeLocalConcurrentWritesTest.java:230)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1570)
	Suppressed: java.lang.Exception: SQL: INSERT INTO test_concurrent_inserts_select_from_same_versioned_table_gu3yeqy900 SELECT 2, 'c' AS part FROM test_concurrent_inserts_select_from_same_versioned_table_gu3yeqy900 FOR VERSION AS OF 0
		at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:508)
		... 7 more
Caused by: io.trino.spi.TrinoException: Failed to write Delta Lake transaction log entry
	at io.trino.plugin.deltalake.DeltaLakeMetadata.finishInsert(DeltaLakeMetadata.java:2028)
	at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorMetadata.finishInsert(ClassLoaderSafeConnectorMetadata.java:633)
[...]

Issue #27484, opened 2025-11-27
Incorrect transaction isolation when inserting into Delta table while selecting from it


findepi:
As found by CI https://github.com/trinodb/trino/actions/runs/19737237040/job/56552097779?pr=27480

```
Error:  Failures: 
Error:    TestDeltaLakeLocalConcurrentWritesTest.testConcurrentInsertsSelectingFromTheSameTable:136->testConcurrentInsertsSelectingFromTheSameTable:186->AbstractTestQueryFramework.assertQuery:356 For query 20251127_132020_00206_h9jyp: 
 SELECT * FROM test_concurrent_inserts_select_from_same_table_6htfdgywiq
not equal
Actual rows (up to 100 of 1 extra rows shown, 3 rows in total):
    [1, 10]
Expected rows (up to 100 of 1 missing rows shown, 3 rows in total):
    [2, 10]

```

cc @findinpath @ebyhr

findinpath:
I have explored several avenues on when this kind of unusual exceptions may happen:
- `LocalOutputFile`  not failing when creating a namesake transaction file
```
OutputStream stream = Files.newOutputStream(path, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE);
```
Unlikely. Even if so, the `SELECT` statement from the table would rather return 2 entries (given that one transaction log was swallowed) instead of 3


- `DeltaLakeMetadata` unobvious bug 

`io.trino.plugin.deltalake.DeltaLakeMetadata#checkForConcurrentTransactionConflicts`
When inserting while reading from the same table the operation should fail:
https://github.com/trinodb/trino/blob/1e3b669c4c7447ee7820397ba13e5940fa997e6c/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L2453-L2456

One option that I can think of is that the INSERT operation is thought as blind, because in such a situation we don't evaluate for conflicts with other INSERT operations.
https://github.com/trinodb/trino/blob/1e3b669c4c7447ee7820397ba13e5940fa997e6c/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L2409

In such a situation we would end up with the tuples and this would match the failure reported above :
```
(0,10)
(1,10)
(1,10)
```
instead of having
```
(0,10)
(1,10)
(2,10)
```



[...]

PR #27904, opened 2026-01-13
Make transaction log publish atomic in local file system

chenjian2664:
## Description

- Fixes https://github.com/trinodb/trino/issues/21725

The Delta Lake connector does not guarantee that queries are executed in a strict order. Queries may start concurrently, which makes count-based assertions in tests non-deterministic.

~~Implementation refer https://github.com/trinodb/trino/blob/master/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/AzureTransactionLogSynchronizer.java~~

The current Delta transaction commit flow for `INSERT` operations works as follows:
1. Read the current table version.
2. If the table version is greater than the current log version, this indicates a concurrent transaction is being committed. Before writing a new log entry, we perform conflict detection:

    a) - Read all actions newly added to the table since the current transaction started. 
    b) - If any of those actions conflict (overlap) with the current transaction, the commit is considered unsafe and fails.

3. If no conflicts are detected, write the transaction log entry with currentVersion + 1.

There is a subtle race condition in the retry path, since we have retry when failing the commit log.
On a retry, it is possible (with very low probability) that when the commit logic reaches the conflict detection step (2.a), another concurrent writer has created the log file but has not finished writing its contents. In this case:
	* The retry observes a partially written transaction log.
	* The newly added actions appear empty or incomplete when current commit reading from the log file.
As a result, the conflict detection passes trivially (no overlapping actions are observed). The retry then proceeds to successfully write its own log entry, step 3.

This leads to an incorrect outcome: the commit succeeds even though it should have detected a conflict with the concurrent transaction.

**Update**

the pr using the workaround by implementing the `createExclusive` for the `LocalOutputFile`

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
@@ -22,7 +22,6 @@ import java.io.File;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.nio.channels.FileChannel;
-import java.nio.channels.FileLock;
 import java.nio.file.FileAlreadyExistsException;
 import java.nio.file.Files;
 import java.nio.file.Path;
@@ -76,27 +75,27 @@ public class LocalOutputFile
             throw new FileAlreadyExistsException(location.toString());
         }
 
-        Closer closer = Closer.create();
         Path lockPath = path.resolveSibling(path.getFileName() + ".lock");
-        FileChannel channel = FileChannel.open(lockPath, CREATE_NEW, WRITE);
-        closer.register(channel);
-        closer.register(() -> Files.deleteIfExists(lockPath));
-        try (FileLock lock = channel.lock()) {
-            if (Files.exists(path)) {
-                throw new FileAlreadyExistsException(location.toString());
-            }
+        Closer closer = Closer.create();
+        try {
+            try (FileChannel _ = FileChannel.open(lockPath, CREATE_NEW, WRITE)) {
+                closer.register(() -> Files.deleteIfExists(lockPath));
+                if (Files.exists(path)) {
+                    throw new FileAlreadyExistsException(location.toString());
+                }
 
-            Path tmpFilePath = path.resolveSibling(path.getFileName() + "." + randomUUID() + ".tmp");
-            try (OutputStream out = Files.newOutputStream(tmpFilePath, CREATE_NEW, WRITE)) {
-                closer.register(() -> Files.deleteIfExists(tmpFilePath));
-                out.write(data);
-            }
+                Path tmpFilePath = path.resolveSibling(path.getFileName() + "." + randomUUID() + ".tmp");
+                try (OutputStream out = Files.newOutputStream(tmpFilePath, CREATE_NEW, WRITE)) {
+                    closer.register(() -> Files.deleteIfExists(tmpFilePath));
+                    out.write(data);
+                }
 
-            // Ensure that the file is only visible when fully written
-            Files.move(tmpFilePath, path, ATOMIC_MOVE);
+                // Ensure that the file is only visible when fully written
+                Files.move(tmpFilePath, path, ATOMIC_MOVE);
+            }
         }
         catch (IOException e) {
-            throw handleException(location, e);
+            throw closer.rethrow(handleException(location, e));
         }
         finally {
             closer.close();


vlad-lyutenko (approved):

findinpath:
@chenjian2664  note that this PR relates to https://github.com/trinodb/trino/issues/27484

> Queries may start concurrently, which makes count-based assertions in tests non-deterministic.

This is not true.
There seems to be indeed an issue when testing against the local filesystem which  has not been reproduced against object storage providers.

findinpath:
@chenjian2664  could you please add reference in the description to recorded CI failures against object storage providers?

findinpath (changes requested):
The flakiness is generated by testing against the local file system.

See https://github.com/trinodb/trino/issues/27484#issuecomment-3586559213

If the same issue would occur against an object storage, we would have a bug in the implementation.

chenjian2664:
@findinpath just considering the reading phrase, when the queries arrive at the connector, it is possible it reads the same version of the table right?

findinpath:
> @findinpath just considering the reading phrase, when the queries arrive at the connector, it is possible it reads the same version of the table right?

yes, sure.

```
INSERT INTO " + tableName + " SELECT COUNT(*), 10 AS part FROM " + tableName
```

This is an operation which reads from the same table it inserts into.
This operation is therefore not a "blind append".

In case of dealing with a non-blind append, we will be checking whether other concurrent operations that are also non-blind inserts have inserted data  and if so, fail the operation:

https://github.com/trinodb/trino/blob/7269eba48931c86a849d83c6f811680e572b847a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L2452-L2455


So, even if two operations are reading the same version of the table, only one of these operations will actually succeed in performing the commit successfully.

chenjian2664:
@findinpath Thanks for guiding me to the place that committing the insert.
I think the problem is here https://github.com/trinodb/trino/blob/b3c5bab7ec84f52a63c7d07a463f821470505c37/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L2354, concurrent reads are possibly here and still reads the same readVersion - if it is, we won't step into `checkForConcurrentTransactionConflicts` at all.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
```diff
@@ -41,14 +41,34 @@ public LocalTransactionLogSynchronizer(DeltaLakeFileSystemFactory fileSystemFact
     @Override
     public void write(ConnectorSession session, VendedCredentialsHandle credentialsHandle, String clusterId, Location newLogEntryPath, byte[] entryContents)
```
findinpath, line 41: For other reviewers:
The implementation strategy used here is the same as for `io.trino.plugin.deltalake.transactionlog.writer.AzureTransactionLogSynchronizer`

findinpath (approved):
Well done @chenjian2664  ❤️

findinpath:
Please add `Fixes https://github.com/trinodb/trino/issues/27484` to the description of the PR

ebyhr:
@findinpath Can we re-enable disabled tests in TestDeltaLakeLocalConcurrentWritesTest now?

findinpath:
@ebyhr yes, the tests disabled with the issue https://github.com/trinodb/trino/issues/21725 should be re-enabled

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
```diff
@@ -41,14 +41,34 @@ public LocalTransactionLogSynchronizer(DeltaLakeFileSystemFactory fileSystemFact
...
     {
+        Location tmpFilePath = newLogEntryPath.appendSuffix("." + UUID.randomUUID() + ".tmp");
```
findepi: nit: import `randomUUID` statically

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
```diff
@@ -41,14 +41,34 @@ public LocalTransactionLogSynchronizer(DeltaLakeFileSystemFactory fileSystemFact
...
+                conflict = true;
+                throw e;
+            }
         }
         catch (IOException e) {
+            try {
+                fileSystem.deleteFile(tmpFilePath);
+            }
+            catch (IOException | RuntimeException ex) {
+                if (!e.equals(ex)) {
```
findepi: rename:
e → createException
ex → deleteException

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
```diff
@@ -41,14 +41,34 @@ public LocalTransactionLogSynchronizer(DeltaLakeFileSystemFactory fileSystemFact
...
+                    e.addSuppressed(ex);
+                }
+            }
+            if (conflict) {
+                throw new TransactionConflictException("Conflict detected while writing Transaction Log entry " + newLogEntryPath, e);
```
findepi: `LocalTransactionLogSynchronizer` is apparently a test only facility today

- we don't need to detect conflict and report a nice error. the `conflict` variable can be removed IMO
- the class could be moved to src/test/ folder

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
```diff
@@ -41,14 +41,34 @@ public LocalTransactionLogSynchronizer(DeltaLakeFileSystemFactory fileSystemFact
...
-            outputStream.write(entryContents);
-        }
-        catch (FileAlreadyExistsException e) {
-            throw new TransactionConflictException("Conflict detected while writing Transaction Log entry " + newLogEntryPath, e);
+        try {
+            try (OutputStream outputStream = fileSystem.newOutputFile(tmpFilePath).create()) {
+                outputStream.write(entryContents);
+            }
+            try {
+                fileSystem.renameFile(tmpFilePath, newLogEntryPath);
```
findepi: Is rename guaranteed to be atomic?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
findepi: `LocalTransactionLogSynchronizer` is apparently a test only facility today, the class could be moved to src/test/ folder.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
chenjian2664: good question, I see the https://github.com/trinodb/trino/blob/9c1abc44e162ae8c63fcc07d072c56a16def1ed6/lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalFileSystem.java#L155-L156 not specify the `ATOMIC_MOVE`, and the comment is not clear to me :)
I am going to append the `ATOMIC_MOVE` option

chenjian2664: ~~After thinking a bit, It seems to me we don't need the rename operation to be atomic(copy+delete) in delta log publishing case as long as it can guarantee that the target file not expose partial content.
Of course, the atomic requirements will be more safe guarantee in this case.~~

Ignore above.
we don't know if the copy(move = copy+delete) attributes will be atomic or not if we don't specify the `ATOMIC_MOVE`.
I will append the `ATOMIC_MOVE` option

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/writer/LocalTransactionLogSynchronizer.java
chenjian2664: Make sense, will do (remove to test folder) after we agree on the implementation part

ebyhr:
> the tests disabled with the issue #21725 should be re-enabled

@findinpath Nice! 

@chenjian2664 Could you re-enable those tests?

chenjian2664:
ci failures https://github.com/trinodb/trino/actions/runs/21194534248/job/60967948329?pr=27904

The rename is atomic but not exclusive :(

chenjian2664:
@findinpath @ebyhr I am afraid we can't enable the tests of https://github.com/trinodb/trino/issues/21725 , the `updateExtendedStatistics` https://github.com/trinodb/trino/blob/9c1abc44e162ae8c63fcc07d072c56a16def1ed6/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/statistics/MetaDirStatisticsAccess.java#L95 does not currently provide atomic guarantees. As a result, two concurrent commits may both succeed and independently update the statistics

findinpath:
> I am afraid we can't enable the tests of https://github.com/trinodb/trino/issues/21725 , the `updateExtendedStatistics` does not currently provide atomic guarantees. 

@chenjian2664  yes, i remember now - let's handle this in a separate contribution.

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +56,57 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
         catch (IOException e) {
             throw handleException(location, e);
         }
     }
 
+    @Override
+    public void createExclusive(byte[] data)
+            throws IOException
+    {
+        Files.createDirectories(path.getParent());
```
findepi, line 71: It wasn't needed before, so i suppose it still isn't needed.

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +56,57 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+            // Use atomic move to ensure that the file is only visible when fully written
+            Files.move(tmpFilePath, path, ATOMIC_MOVE);
+        }
+        catch (IOException e) {
+            throw handleException(location, e);
+        }
+        finally {
+            if (lockCreated) {
+                try {
+                    Files.deleteIfExists(lockPath);
```
findepi: We may skip deletion if file was created but `channel.lock()` failed.

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +56,57 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+            if (Files.exists(path)) {
+                throw new FileAlreadyExistsException(location.toString());
+            }
+
+            Path tmpFilePath = path.resolveSibling(path.getFileName() + "." + randomUUID() + ".tmp");
+            try (OutputStream out = Files.newOutputStream(tmpFilePath, CREATE_NEW, WRITE)) {
+                out.write(data);
+            }
+
+            // Use atomic move to ensure that the file is only visible when fully written
```
findepi:
```suggestion
            // Ensure that the file is only visible when fully written
```

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +56,57 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+            Files.move(tmpFilePath, path, ATOMIC_MOVE);
```
findepi: When move fails, the `tmpFilePath` is left behind.

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +56,57 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+    {
+        Files.createDirectories(path.getParent());
+
+        // see if we can stop early without acquire locking
+        if (Files.exists(path)) {
+            throw new FileAlreadyExistsException(location.toString());
+        }
+
+        boolean lockCreated = false;
+        Path lockPath = path.resolveSibling(path.getFileName() + ".lock");
```
findepi, line 78: It looks like we have two mechanisms (file locks and atomic moves) to achieve same goal (exclusion).
So, what is lock needed for?

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
findinpath: see `io.trino.filesystem.local.LocalOutputFile#create`

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
chenjian2664: the file locks is for the exclusive - only one thread should able to create target file at a time.
The move is for the visibility(atomic), readers should either not see the target file or see the file exists and with the full content.

findinpath: pls document this line of thinking with a comment

findepi: @chenjian2664 lock file is "cooperative locking" - the parties involved need to agree on the "protocol" they are following (the lock file existence, its name, locked region).
rename seems to be sufficient for "cooperative exclusion"

i.e. if there is no lock file at all, what could possibly go wrong?

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +57,52 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+            throws IOException
+    {
+        Files.createDirectories(path.getParent());
+
+        // see if we can stop early without acquire locking
+        if (Files.exists(path)) {
+            throw new FileAlreadyExistsException(location.toString());
+        }
+
+        Closer closer = Closer.create();
```
findepi: try with resources?

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +57,52 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+        Path lockPath = path.resolveSibling(path.getFileName() + ".lock");
+        FileChannel channel = FileChannel.open(lockPath, CREATE_NEW, WRITE);
+        closer.register(channel);
+        closer.register(() -> Files.deleteIfExists(lockPath));
```
findepi: The lock handling is quite problematic

- what if lock file was left behind by previous attempt (process killed and didn't remove)? the CREATE_NEW will trigger lock file creation failure, application won't recover
- assuming this is fixed, imagine 3 threads contending for the write. A locks, B fails and removes lock file during its cleanup, C creates new lock file and also locks. Now A and C both think they have the lock.

if locks are gone (https://github.com/trinodb/trino/pull/27904/changes#r2730875323), great
if not, these problems need to be addressed somehow

chenjian2664: For the first question, I think it is the problem, but note there will be always problematic if we use the `CREATE_NEW` strategy to create the file, i,e in previous implementation we use it to create the file then writing into it, and when process killed during writing we are facing the same issue.

Second question, the `FileChannel channel = FileChannel.open(lockPath, CREATE_NEW, WRITE);` will guarantee that only one thread can register the line 83, then will cleanup the lock. there shouldn't be a thread fail without creating the lock but registered the cleanup logic

findepi: > For the first question, I think it is the problem, but note there will be always problematic if we use the `CREATE_NEW` strategy to create the file, i,e in previous implementation we use it to create the file then writing into it, and when process killed during writing we are facing the same issue.

Except that we're writing to a temp file with randomized name, so it's less likely to matter that we left a file behind.

> Second question, the `FileChannel channel = FileChannel.open(lockPath, CREATE_NEW, WRITE);` will guarantee that only one thread can register the line 83, then will cleanup the lock

If line numbers match what GH is showing me

image

then, the deletion in 83 is registered before lock is taken.

findepi: I am sure we can make file locks work reasonable well, but let's first consider this one:

> if locks are gone (https://github.com/trinodb/trino/pull/27904/changes#r2730875323), great


i.e. maybe we don't need any file locks at all

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
chenjian2664: @findepi The reason is `Files.move` is not guarantee to be exclusive it is only guarantee to be atomic.
If no lock there then probably multi-threads will think it success moved, but actually being overwrite by others-who commit last win.

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
chenjian2664: I am not sure understand you point correctly - if you are saying we could remove the `FileLock lock = channel.lock()` line, then I think it is make sense, I will do it, since there should only one thread can pass the line 81 at a time.

I would like to discuss about the point if you think we should remove FileChannel at all :)
refer https://github.com/trinodb/trino/pull/27904#discussion_r2734437908 - the move itself doesn't guarantee the exclusive

chenjian2664: > then, the deletion in 83 is registered before lock is taken.

Oh, I think here is a gap we understanding the "lock", actually the line 81 -> "FileChannel _ = FileChannel.open(lockPath, CREATE_NEW, WRITE)", it has the "lock"(exclusive) semantic, there should only one thread can pass here, the taken lock from file channel I put there is misleading so I removed now.

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
findepi: I believe that `Files.move` with `ATOMIC_MOVE` and without `REPLACE_EXISTING` is exclusive.
At least it's IMO documented as such.

Am i wrong?

chenjian2664: I tested with below code :
```

    static void main()
            throws Exception
    {
        Path tempDirectory = Files.createTempDirectory("trino-test-");
        ExecutorService executorService = Executors.newFixedThreadPool(10);
        CyclicBarrier barrier = new CyclicBarrier(10);
        Path targetFilePath = tempDirectory.resolve("target.test");
        for (int i = 0; i < 10; i++) {
            Path tmpFilePath = tempDirectory.resolve("file-" + i + ".test");
            executorService.execute(() -> {
                try {
                    writeFile(tmpFilePath);
                    barrier.await();
                    Files.move(tmpFilePath, targetFilePath, ATOMIC_MOVE);
                    System.out.println("Moved " + tmpFilePath.getFileName() + " to target.test");
                }
                catch (Exception e) {
                }
            });
        }

        Thread.sleep(5000);
        executorService.shutdown();
    }

    private static void writeFile(Path tmpFilePath)
            throws IOException
    {
        try (OutputStream out = Files.newOutputStream(tmpFilePath, CREATE_NEW, WRITE)) {
            out.write(("this is " + tmpFilePath.getFileName()).getBytes());
        }
    }
```
   
 multiple tmp files think there success moved. could you please help to confirm, or I am using wrongly with the move?

findepi: You're right. I have not read `Files.move` lengthy javadoc carefully enough. Thanks for standing me corrected.

We may get away from this by

- call some native rename / renameat / renameat2 function. Linux kernel awesomely supports this basic operation of renaming a file without overriding a target, should it exist. It's Java's fault it doesn't work
  - not great as it's going to be some unpretty JNA / FFM call and may or may not work on MacOS & Linux the same way
- workaround the limitation with `Files.createLink`. It seems to deliver exclusive target creation (on MacOS at least)
  - not great as this function seems quite platform-dependent. it doesn't seem to guarantee exclusiveness
- remind ourselves we're doing this for our tests only. While `LocalFileSystem` is a production facility which needs to be implemented well, the only call site we're really concerned about is `LocalTransactionLogSynchronizer` ([which was supposed to be moved to test dir](https://github.com/trinodb/trino/pull/27904#discussion_r2707872913)). For internal test purposes we can simply synchronize in memory on a lock. BTW this is how `FileHiveMetastore` performs synchronization around filesystem operations.

Let's do this last option.

chenjian2664: Sorry for the late reply.
I did consider an in-memory lock at first. My concern is about how we validate the log synchronizer design. We're currently relying entirely on the semantics of `createExclusive` and assuming the synchronizer logic is correct.

If we can't prove the local concurrency behavior is correct, how can we trust it in the cloud, even if the filesystem provides the same `createExclusive` guarantees, that’s why I wanted this to work across multiple clusters. It's still not a strong guarantee, but at least it would give us some confidence that matching `createExclusive` semantics won't lead to corrupted Delta logs or wrongly thinking it is success committed. 

But I looked the `FileHiveMetastore` you've mentioned seems I am overthinking ? 

And I don't found tests about the concurrent writing of cloud log synchronizer, is it because of the flaky, or I am missing the tests at somewhere, what's the reason behind?

lib/trino-filesystem/src/main/java/io/trino/filesystem/local/LocalOutputFile.java
```diff
@@ -50,14 +56,52 @@ public OutputStream create(AggregatedMemoryContext memoryContext)
...
+                // Ensure that the file is only visible when fully written
+                Files.move(tmpFilePath, path, ATOMIC_MOVE);
+            }
+        }
+        catch (IOException e) {
+            throw closer.rethrow(handleException(location, e));
+        }
+        finally {
+            closer.close();
+        }
```
findepi, line 102:
```suggestion
               Files.createDirectories(path.getParent());

        // Check target path before creating lock file
        if (Files.exists(path)) {
            throw new FileAlreadyExistsException(location.toString());
        }

        Path lockPath = path.resolveSibling(path.getFileName() + ".lock");
        Path tmpFilePath = path.resolveSibling(path.getFileName() + "." + randomUUID() + ".tmp");
        try (Closer closer = Closer.create()) {
            Files.write(lockPath, new byte[0], CREATE_NEW);
            closer.register(() -> Files.delete(lockPath));
            Files.write(tmpFilePath, data, CREATE_NEW);
            closer.register(() -> Files.deleteIfExists(tmpFilePath));

            // Ensure that the file is only visible when fully written
            // Files.move with ATOMIC_MOVE is not guaranteed to be exclusive, hence the lock file
            Files.move(tmpFilePath, path, ATOMIC_MOVE);
        }
```

chenjian2664:
superseded by https://github.com/trinodb/trino/pull/28092

PR #27908, opened 2026-01-13
Cleanup ScalarStatsCalculator row size estimates

pettyjamesm:
## Description
Uses the fixed size for casts and constants of fixed-width types to estimate the size per position.

Follows up from https://github.com/trinodb/trino/pull/27871

## Release notes


pettyjamesm:
Similar logic in [PlanNodeStatsEstimate](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/cost/PlanNodeStatsEstimate.java#L101-L104) does not include nulls in the average per-row size, although it's not clear why that would be the case (nulls still consume the fixed width size per position in-memory) cc: @sopel39

PR #27912, opened 2026-01-13
Add Debug Logs for Spilling and Unspilling Partitions in GenericPartitioningSpiller

Richard-Durkee:
Changes:
- Log partition number, operator name, and data sizes during spill/unspill
- Add operatorName parameter to PartitioningSpillerFactory.create()
- Update LookupJoinOperator to pass operator name for probe side spilling

## Description

This PR adds debug logging to `GenericPartitioningSpiller` to improve observability of spilling behavior, where currently no debug logs exist. This PR came out of an investigation of spilling and unspilling that occurs during the probe side of hash joins.

The new logs include operator name, partition number, and data sizes for each spill/unspill operation (using `sizeInBytes` and `retainedSizeInBytes`). See attached screenshot. 

Examples of Spilling and Unspilling in GenericPartitioningSpiller

This is my first contribution to Trino. I am open to any notes or revisions. 

To avoid passing the entire OperatorContext, I pass only the name of the operator. 

I followed the `FormatStringShouldUsePlaceholders` rule that was enabled in [this PR](https://github.com/trinodb/trino/pull/27643). 

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/spiller/GenericPartitioningSpiller.java
+++ b/core/trino-main/src/main/java/io/trino/spiller/GenericPartitioningSpiller.java
@@ -19,6 +19,7 @@ import com.google.common.io.Closer;
 import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.errorprone.annotations.ThreadSafe;
+import io.airlift.log.Logger;
 import io.airlift.units.DataSize;
 import io.trino.memory.context.AggregatedMemoryContext;
 import io.trino.operator.PartitionFunction;
@@ -45,18 +46,22 @@ import static com.google.common.base.Verify.verify;
 import static com.google.common.util.concurrent.Futures.immediateFuture;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static io.airlift.concurrent.MoreFutures.getFutureValue;
+import static io.airlift.units.DataSize.succinctBytes;
 import static java.util.Objects.requireNonNull;
 
 @ThreadSafe
 public class GenericPartitioningSpiller
         implements PartitioningSpiller
 {
+    private static final Logger log = Logger.get(GenericPartitioningSpiller.class);
+
     private final List<Type> types;
     private final PartitionFunction partitionFunction;
     private final Closer closer = Closer.create();
     private final SingleStreamSpillerFactory spillerFactory;
     private final SpillContext spillContext;
     private final AggregatedMemoryContext memoryContext;
+    private final String operatorName;
 
     private final List<PageBuilder> pageBuilders;
     private final List<Optional<SingleStreamSpiller>> spillers;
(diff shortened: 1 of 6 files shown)


sopel39:
please fix compilation issues

chenjian2664:
Please sign the cla as the bot suggested if you haven't done so

Richard-Durkee:
@chenjian2664 I emailed my signed CLA on January 9th, but it was from an email address no longer associated with my Github account. This morning, I resubmitted from a different address -- hoping that will work.

pettyjamesm:
> @chenjian2664 I emailed my signed CLA on January 9th, but it was from an email address no longer associated with my Github account. This morning, I resubmitted from a different address -- hoping that will work.

cc: @martint - CLA check

Richard-Durkee:
@sopel39 @chenjian2664 CLA is signed and all checks are passing. Thank you for your patience.

core/trino-main/src/main/java/io/trino/spiller/GenericPartitioningSpiller.java
```diff
@@ -186,6 +199,14 @@ private synchronized ListenableFuture<DataSize> flush(int partition)
         }
         Page page = pageBuilder.build();
         pageBuilder.reset();
+
+        log.debug(
+                "Spilling partition %d for operator %s, sizeInBytes %s, retainedSizeInBytes %s",
+                partition,
+                operatorName,
+                succinctBytes(page.getSizeInBytes()),
+                succinctBytes(page.getRetainedSizeInBytes()));
```
findepi: No need to guard succinctBytes with isDebugEnabled?

core/trino-main/src/main/java/io/trino/spiller/PartitioningSpillerFactory.java
```diff
@@ -26,11 +26,21 @@ PartitioningSpiller create(
             List<Type> types,
             PartitionFunction partitionFunction,
             SpillContext spillContext,
-            AggregatedMemoryContext memoryContext);
+            AggregatedMemoryContext memoryContext,
+            String operatorName);
+
+    default PartitioningSpiller create(
```
pettyjamesm: Looks like this is only called from `DefaultPageJoiner` and then test code, so maybe just inline passing "unknown" or "testOperator" from the test usages instead of adding a default method overload.

core/trino-main/src/main/java/io/trino/spiller/GenericPartitioningSpiller.java
pettyjamesm: This is going to be on a somewhat busy path for each flush, so the overhead of the varargs call (and `succinctBytes`) might be modest but measurable. I think it's justified to leave the `isDebugEnabled()` check here out of caution.

Richard-Durkee: To clarify, @findepi left a comment recommending the addition of isDebugEnabled(). I thought this was a good suggestion -- I wasn't actually aware of isDebugEnabled(). I addressed this shortly after, but never resolved the conversation. I think we are all aligned.

core/trino-main/src/main/java/io/trino/spiller/PartitioningSpillerFactory.java
Richard-Durkee: I think this is better, thank you for the suggestion. Implemented.

Issue #27913, opened 2026-01-14
Flaky `TestJmxStats.testJmxStatsExposure`


ebyhr:
```
Error:  io.trino.plugin.jdbc.TestJmxStats.testJmxStatsExposure -- Time elapsed: 0.926 s <<< FAILURE!
org.opentest4j.AssertionFailedError: 

Expecting value to be true but was false
	at io.trino.plugin.jdbc.TestJmxStats.testJmxStatsExposure(TestJmxStats.java:55)
```

https://github.com/trinodb/trino/actions/runs/20979150067/job/60300343371

antriksh-9:
hey @ebyhr, I would like to work on this issue.
Please assign it to me.
Also, can u give some more info. regarding the test case or related function?

ebyhr:
@antriksh-9 You don't need to get assigned. Please feel free to open a PR once it's ready.

PR #27930, opened 2026-01-15
Expose test MBeans in a separate domain

wendigo:
## Description

Fixes https://github.com/trinodb/trino/issues/27913

## Release notes

```markdown
## Section
```


plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestJmxStats.java
```diff
@@ -38,21 +40,24 @@ public class TestJmxStats
     public void testJmxStatsExposure()
             throws Exception
     {
-        Plugin plugin = new JdbcPlugin("base_jdbc", TestingH2JdbcModule::new);
+        String jmxBaseName = UUID.randomUUID().toString().replace("-", "");
```
findepi: why not use `getClass().getSimpleName() + "_" + randomNameSuffix()`?

nit: import randomUUID static

PR #27788, opened 2025-12-30
[Iceberg v3] Deletion vector

dain:
## Description

This PR adds support for Iceberg format v3 deletion vectors in the Trino Iceberg connector.

Read path:

* Recognizes v3 deletion vectors stored as Puffin `deletion-vector-v1` blobs referenced from delete file entries.
* Applies deletion vectors during reads using the row position column, integrating with the existing delete filtering pipeline.

Write path:

* Enables row-level operations (DELETE/UPDATE/MERGE) for v3 tables by writing deletion vectors rather than legacy position delete files.
* Supports v2 → v3 upgrade scenarios by merging existing v2 position delete files into deletion vectors once the table is upgraded.
* Maintains a single deletion vector per data file and prefers deletion vectors over position delete files when present.

Testing:

* Adds coverage for writing/reading deletion vectors in v3 tables (including multiple Puffin files mid-stream and convergence).
* Adds coverage for v2 tables with deletes upgraded to v3, ensuring existing deletes remain effective and new deletes use deletion vectors.
* Updates v3 “updates blocked” tests (DELETE/UPDATE/MERGE) to validate the operations now succeed.

## Release notes

(X) Release notes are required, with the following suggested text:

```markdown
## Iceberg
* Add support for Iceberg format v3 deletion vectors to enable DELETE/UPDATE/MERGE on v3 tables.
```

PR #27938, opened 2026-01-16
Minor updates for `DeleteVector`

chenjian2664:
## Description

some left or reviews from https://github.com/trinodb/trino/pull/27788

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/delete/TestDeletionVector.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/delete/TestDeletionVector.java
@@ -208,12 +208,20 @@ class TestDeletionVector
     {
         long key0Pos = 42;
         long key2Pos = (2L << 32) + 100;
+        long key2Pos2 = (2L << 32) + 1000;
+        long key2Pos3 = (2L << 32) + 10000;
         long key5Pos = (5L << 32) + 999;
+        long key5Pos2 = (5L << 32) + 1000;
+        long key5Pos3 = (5L << 32) + 1111;
 
         DeletionVector original = DeletionVector.builder()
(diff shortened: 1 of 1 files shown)


raunaqmorarka (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/delete/DeletionVector.java
```diff
@@ -238,21 +238,21 @@ public Builder add(long pos)
...
         {
-            for (int i = 0; i < other.deletedRows.length; i++) {
-                RoaringBitmap bitmap = other.deletedRows[i];
+            return addAll(other.deletedRows);
+        }
+
+        private Builder addAll(RoaringBitmap[] deletedRows)
+        {
+            // reverse order to minimize resizing the array
+            for (int key = deletedRows.length - 1; key >= 0; key--) {
```
dain: I don't think we should make this change.  It is more complicated and isn't going to help any real world usecases.  Each roaring bitmap holds 4B entires, we do not need to optimzie for files with > 4b entries, and instead we should be focusing on simplicity.

chenjian2664: For this one, make sense, I revert this one(reverse visiting logic), but keep the logic that extracting the method and reuse it.

However, adding an explicit position count is more readable to me. I mistakenly removed the comment earlier, also revert that.
I agree that in practice the real use case should not have 4B entries, but we cannot strictly guarantee that. we are still using an array and dynamically extending it based on length checks, rather than fixing the array size to exactly one bitmap, using the count is more obvious to understand the state of the vector, and it reduce the code - we don't need to calculate it when serializing

PR #27943, opened 2026-01-17
Preserve table reader and writer features on `CREATE OR REPLACE TABLE`

chenjian2664:
## Description

Fixes the issue https://github.com/trinodb/trino/pull/27886#issuecomment-3752516781 :
```
CREATE TABLE test (x int, y varchar) WITH (checkpoint_interval = 2, location = 's3://test-bucket/test', deletion_vectors_enabled = true);
INSERT INTO test VALUES (1, 'aa');
INSERT INTO test VALUES (1, 'bb');

CREATE OR REPLACE TABLE test (x int, y varchar) WITH (checkpoint_interval = 2, location = 's3://test-bucket/test');

SELECT * FROM default.test; -- Spark fails 
```

Follow the https://github.com/trinodb/trino/pull/27886#issuecomment-3753733798 :
```
I find https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features, it means we do wrote the illegal protocol entry.

I am not able to find the behavior of create or replace in the protocol, but I find https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features-for-new-and-existing-tables, it says:

When upgrading an existing table to Reader Version 3 and/or Writer Version 7, the client should, on a best effort basis,
 determine which features supported by the original protocol version are used in any historical version of the table, and
 add only used features to reader and/or writer feature sets. The client must assume a feature has been used, unless it 
can prove that the feature is definitely not used in any historical version of the table that is reachable by time travel.
combined with https://github.com/delta-io/delta/blob/master/PROTOCOL.md#supported-features, my understanding is that no matter what's the version it is, if the procotol has readFeatures or writeFeatures, we should inherit it from original version, unless we pretty sure that all the historic data not use the feature at all.

which means we should keep the original read features and write features if we know they were used.
```

## Release notes

```markdown
## Delta Lake
* Preserve table reader and writer features on `CREATE OR REPLACE TABLE`. ({issue}`27943`)
```

### DIFF
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
@@ -1352,12 +1352,12 @@ public class DeltaLakeMetadata
                             transactionLogWriter.appendRemoveFileEntry(new RemoveFileEntry(addFileEntry.getPath(), addFileEntry.getPartitionValues(), writeTimestamp, true, Optional.empty()));
                         }
                     }
-                    protocolEntry = protocolEntryForTable(tableHandle.getProtocolEntry().minReaderVersion(), tableHandle.getProtocolEntry().minWriterVersion(), containsTimestampType, tableMetadata.getProperties());
+                    protocolEntry = protocolEntryForTable(ProtocolEntry.builder(tableHandle.getProtocolEntry()), containsTimestampType, tableMetadata.getProperties());
                     statisticsAccess.deleteExtendedStatistics(session, schemaTableName, location, tableHandle.toCredentialsHandle());
                 }
                 else {
                     setRollback(() -> deleteRecursivelyIfExists(fileSystem, deltaLogDirectory));
-                    protocolEntry = protocolEntryForTable(DEFAULT_READER_VERSION, DEFAULT_WRITER_VERSION, containsTimestampType, tableMetadata.getProperties());
+                    protocolEntry = protocolEntryForTable(ProtocolEntry.builder(DEFAULT_READER_VERSION, DEFAULT_WRITER_VERSION), containsTimestampType, tableMetadata.getProperties());
                 }
 
                 MetadataEntry metadataEntry = MetadataEntry.builder()
@@ -1542,7 +1542,7 @@ public class DeltaLakeMetadata
 
         boolean isNewCheckpointFileRequired = false;
         if (replaceExistingTable) {
-            protocolEntry = protocolEntryForTable(handle.getProtocolEntry().minReaderVersion(), handle.getProtocolEntry().minWriterVersion(), containsTimestampType, tableMetadata.getProperties());
+            protocolEntry = protocolEntryForTable(ProtocolEntry.builder(handle.getProtocolEntry()), containsTimestampType, tableMetadata.getProperties());
             readVersion = OptionalLong.of(handle.getReadVersion());
             isNewCheckpointFileRequired = isNewCheckpointFileRequired(getColumns(handle.getMetadataEntry(), handle.getProtocolEntry()), columnHandles.build());
         }
@@ -1550,7 +1550,7 @@ public class DeltaLakeMetadata
             TrinoFileSystem fileSystem = fileSystemFactory.create(session, location);
             checkPathContainsNoFiles(fileSystem, finalLocation);
             setRollback(() -> deleteRecursivelyIfExists(fileSystem, finalLocation));
-            protocolEntry = protocolEntryForTable(DEFAULT_READER_VERSION, DEFAULT_WRITER_VERSION, containsTimestampType, tableMetadata.getProperties());
+            protocolEntry = protocolEntryForTable(ProtocolEntry.builder(DEFAULT_READER_VERSION, DEFAULT_WRITER_VERSION), containsTimestampType, tableMetadata.getProperties());
         }
 
         return new DeltaLakeOutputTableHandle(
@@ -3130,10 +3130,10 @@ public class DeltaLakeMetadata
         }
     }
 
-    private ProtocolEntry protocolEntryForTable(int readerVersion, int writerVersion, boolean containsTimestampType, Map<String, Object> properties)
+    private ProtocolEntry protocolEntryForTable(ProtocolEntry.Builder builder, boolean containsTimestampType, Map<String, Object> properties)
     {
         return protocolEntry(
-                ProtocolEntry.builder(readerVersion, writerVersion),
+                builder,
                 containsTimestampType,
                 getChangeDataFeedEnabled(properties),
                 getColumnMappingMode(properties),
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
@@ -61,7 +61,7 @@ public record ProtocolEntry(
         // Spark done the similar logic here:
         // https://github.com/delta-io/delta/blob/f5ebfbb10bda87b6b2b7c9032e86edaf40d446bc/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L155-L157
         if (minReaderVersion >= MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isEmpty()) {
-            log.warn("readerFeatures must exist when minReaderVersion is greater than or equal to " + MIN_VERSION_SUPPORTS_READER_FEATURES);
+            log.debug("readerFeatures must exist when minReaderVersion is greater than or equal to %s", MIN_VERSION_SUPPORTS_READER_FEATURES);
         }
         if (minWriterVersion < MIN_VERSION_SUPPORTS_WRITER_FEATURES && writerFeatures.isPresent()) {
             throw new IllegalArgumentException("writerFeatures must not exist when minWriterVersion is less than " + MIN_VERSION_SUPPORTS_WRITER_FEATURES);
@@ -72,13 +72,13 @@ public record ProtocolEntry(
         // Spark done the similar logic here:
         // https://github.com/delta-io/delta/blob/f5ebfbb10bda87b6b2b7c9032e86edaf40d446bc/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L158-L160
         if (minWriterVersion >= MIN_VERSION_SUPPORTS_WRITER_FEATURES && writerFeatures.isEmpty()) {
-            log.warn("writerFeatures must exist when minWriterVersion is greater than or equal to " + MIN_VERSION_SUPPORTS_WRITER_FEATURES);
+            log.debug("writerFeatures must exist when minWriterVersion is greater than or equal to %s", MIN_VERSION_SUPPORTS_WRITER_FEATURES);
         }
 
         // Spark done the similar logic here:
         // https://github.com/delta-io/delta/blob/f5ebfbb10bda87b6b2b7c9032e86edaf40d446bc/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L162-L166
         if (minReaderVersion >= MIN_VERSION_SUPPORTS_READER_FEATURES && minWriterVersion < MIN_VERSION_SUPPORTS_WRITER_FEATURES) {
-            log.warn("When reader is on table features, writer must be on table features too");
+            log.debug("When reader is on table features, writer must be on table features too");
         }
 
         readerFeatures = requireNonNull(readerFeatures, "readerFeatures is null").map(ImmutableSet::copyOf);
--- a/testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
+++ b/testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
@@ -140,6 +140,24 @@ public class TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility
         }
     }
 
+    @Test(groups = {DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
+    public void testCreateOrReplaceTableWithOnlyFeaturesChangedOnTrino()
+    {
+        String tableName = "test_create_or_replace_table_discard_features_" + randomNameSuffix();
+
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (x INT) " +
+                "WITH (location = 's3://" + bucketName + "/delta-lake-test-" + tableName + "', deletion_vectors_enabled = true)");
+        try {
+            onTrino().executeQuery("INSERT INTO delta.default." + tableName + " VALUES 1, 2, 3");
+
+            onTrino().executeQuery("CREATE OR REPLACE TABLE delta.default." + tableName + " (x INT)");
+            assertThat(onDelta().executeQuery("SELECT * FROM " + tableName)).hasNoRows();
+        }
+        finally {
+            onTrino().executeQuery("DROP TABLE delta.default." + tableName);
+        }
+    }
+
     private static List<Row> performInsert(QueryExecutor queryExecutor, String tableName, int numberOfRows)
     {
         ImmutableList.Builder<Row> expectedRowBuilder = ImmutableList.builder();
--- a/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeBasic.java
+++ b/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeBasic.java
@@ -494,6 +494,27 @@ public class TestDeltaLakeBasic
         }
     }
 
+    @Test
+    void testCreateOrReplacePreservesFeatures()
+            throws Exception
+    {
+        try (TestTable table = newTrinoTable("test_create_or_replace_preserves_features_", "(x int) WITH (deletion_vectors_enabled = true)")) {
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .contains("deletion_vectors_enabled = true");
+
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " (x int)");
+            // verify deletion_vector_enabled is preserved after CREATE OR REPLACE TABLE
+            Path tableLocation = Path.of(getTableLocation(table.getName()).replace("file://", ""));
+            ProtocolEntry protocolEntry = loadProtocolEntry(1, tableLocation);
+            assertThat(protocolEntry.readerFeaturesContains("deletionVectors")).isTrue();
+            assertThat(protocolEntry.writerFeaturesContains("deletionVectors")).isTrue();
+
+            // deletion_vectors_enabled is not enabled, since we created the table without it
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
+                    .doesNotContain("deletion_vectors_enabled = true");
+        }
+    }
+
     @Test // regression test for https://github.com/trinodb/trino/issues/24121
     void testPartitionValuesParsedCheckpoint()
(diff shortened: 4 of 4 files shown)


ebyhr:
The CI failure is related to this change.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
         }
     }
 
+    @Test(groups = {DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
+    public void testCreateOrReplaceTableOnlyDiscardFeaturesOnTrino()
+    {
+        String tableName = "test_create_or_replace_table_discard_features_" + randomNameSuffix();
+
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (ts VARCHAR) " +
```
findinpath: nit: we don't need 10 writes  

we can create the table with a smaller `checkpoint_interval`

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
         }
     }
 
+    @Test(groups = {DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
```
findinpath, line 143: This contribution is about ensuring that we're preserving in case of `CREATE OR REPLACE` the previously existing table features. Let's have such a test and showcase through `SHOW CREATE TABLE` that although a bit counterintuitive, the table preserves its features when doing a `CREATE OR REPLACE`.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
```diff
@@ -54,9 +54,15 @@ public record ProtocolEntry(
         if (minReaderVersion < MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isPresent()) {
             throw new IllegalArgumentException("readerFeatures must not exist when minReaderVersion is less than " + MIN_VERSION_SUPPORTS_READER_FEATURES);
         }
+        if (minReaderVersion >= MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isEmpty()) {
+            throw new IllegalArgumentException("readerFeatures must exist when minReaderVersion is greater than or equal to " + MIN_VERSION_SUPPORTS_READER_FEATURES);
+        }
```
ebyhr, line 65: Can you share the protocol statement disallowing this condition, or link to the relevant Spark code?

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
...
+    public void testCreateOrReplaceTableOnlyDiscardFeaturesOnTrino()
+    {
+        String tableName = "test_create_or_replace_table_discard_features_" + randomNameSuffix();
+
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (ts VARCHAR) " +
+                               "with (location = 's3://" + bucketName + "/databricks-compatibility-test-" + tableName + "', checkpoint_interval = 2, deletion_vectors_enabled = true)");
```
ebyhr:
```suggestion
        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + "(ts VARCHAR)" +
                "WITH (location = 's3://" + bucketName + "/databricks-compatibility-test-" + tableName + "', checkpoint_interval = 2, deletion_vectors_enabled = true)");
```

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
         }
     }
 
+    @Test(groups = {DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
+    public void testCreateOrReplaceTableOnlyDiscardFeaturesOnTrino()
```
ebyhr: I feel that the test name doesn't match what the test actually verifies. Which assertion ensures "OnlyDiscardFeaturesOnTrino"?

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
...
+    {
+        String tableName = "test_create_or_replace_table_discard_features_" + randomNameSuffix();
+
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (ts VARCHAR) " +
+                               "with (location = 's3://" + bucketName + "/databricks-compatibility-test-" + tableName + "', checkpoint_interval = 2, deletion_vectors_enabled = true)");
+        try {
+            List<Row> expectedRows = performInsert(onTrino(), tableName, 2);
+
+            onTrino().executeQuery("CREATE OR REPLACE TABLE delta.default." + tableName + " AS SELECT CAST(ts AS TIMESTAMP(6)) as ts FROM " + tableName);
+            assertThat(onDelta().executeQuery("SELECT date_format(ts, \"yyyy-MM-dd'T'HH:mm:ss.SSS'Z'\") FROM default." + tableName)).containsOnly(expectedRows);
```
ebyhr: Inserting string value and casting it to timestamp looks redundant. I recommend simplifying this test without `performInsert` call.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
chenjian2664: renamed to "testCreateOrReplaceTableWithOnlyFeaturesChangedOnTrino", I want to say: it tests the create or replace and with only discard a feature on Trino, do you have suggestions on the name?

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/transactionlog/TestProtocolEntry.java
```diff
@@ -49,7 +49,7 @@ public void testInvalidProtocolEntryFromJson()
                 .hasStackTraceContaining("readerFeatures must not exist when minReaderVersion is less than 3");
```
findinpath, line 50:
```
        @Language("JSON")
        String jsonWithoutFeatures = "{\"minReaderVersion\":3,\"minWriterVersion\":7}";
        assertThat(codec.fromJson(jsonWithoutFeatures)).isEqualTo(new ProtocolEntry(3, 7, Optional.empty(), Optional.empty()));

```

This test passes, although it should fail.

`supportsReaderFeatures()` & `supportsWriterFeatures()` rely on the `ProtocolEntry` fields, but they are not initialized yet. Test on the provided parameter instead.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
...
     }
 
+    @Test(groups = {DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
+    public void testCreateOrReplaceTableWithOnlyFeaturesChangedOnTrino()
+    {
+        String tableName = "test_create_or_replace_table_discard_features_" + randomNameSuffix();
+
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (x INT) " +
+                "WITH (location = 's3://" + bucketName + "/databricks-compatibility-test-" + tableName + "', deletion_vectors_enabled = true)");
+        try {
```
findinpath, line 150: Without the checkpoint we don't have anymore the failure reported in this PR's description

https://github.com/trinodb/trino/pull/27886#issuecomment-3752516781

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/transactionlog/TestProtocolEntry.java
chenjian2664: Thanks :)

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
chenjian2664: This one isn't relates to the checkpoint, it is that we write an illegal protocol entry caused the failure.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
```diff
@@ -54,9 +54,30 @@ public record ProtocolEntry(
...
         }
+        // when writer version is 7, writerFeatures must exist in protocol - https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features
+        //
+        // Spark done the similar logic here:
+        // https://github.com/delta-io/delta/blob/f5ebfbb10bda87b6b2b7c9032e86edaf40d446bc/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L158-L160
+        if (minWriterVersion >= MIN_VERSION_SUPPORTS_WRITER_FEATURES && writerFeatures.isEmpty()) {
+            throw new IllegalArgumentException("writerFeatures must exist when minWriterVersion is greater than or equal to " + MIN_VERSION_SUPPORTS_WRITER_FEATURES);
+        }
+
+        // Spark done the similar logic here:
```
Math-ias: I believe this rule [is codified into the protocol here](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features-for-new-and-existing-tables).

findepi:
> Preserve table features on CREATE OR REPLACE TABLE

Reader/writer Features, but not configuration, right?

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/transactionlog/TestProtocolEntry.java
```diff
@@ -49,7 +49,7 @@ public void testInvalidProtocolEntryFromJson()
...
 
         @Language("JSON")
-        String invalidMinWriterVersion = "{\"minReaderVersion\":3,\"minWriterVersion\":6,\"writerFeatures\":[\"timestampNTZ\"]}";
+        String invalidMinWriterVersion = "{\"minReaderVersion\":2,\"minWriterVersion\":6,\"writerFeatures\":[\"timestampNTZ\"]}";
```
findinpath: Let's add to `io.trino.plugin.deltalake.transactionlog.TestProtocolEntry#testInvalidProtocolEntryFromJson`

```
        @Language("JSON")
        String jsonWithoutFeatures = "{\"minReaderVersion\":3,\"minWriterVersion\":7}";
        assertThatThrownBy(() -> codec.fromJson(jsonWithoutFeatures))
                .hasMessageContaining("Invalid JSON string")
                .hasStackTraceContaining("readerFeatures must exist when minReaderVersion is greater than or equal to 3");
```

chenjian2664: Thanks, added

chenjian2664:
> > Preserve table features on CREATE OR REPLACE TABLE
> 
> Reader/writer Features, but not configuration, right?

Yes, updated the title and release note

chenjian2664:
@ebyhr @findepi @Praveen2112 Could you please take a look

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
```diff
@@ -54,9 +54,30 @@ public record ProtocolEntry(
         if (minReaderVersion < MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isPresent()) {
             throw new IllegalArgumentException("readerFeatures must not exist when minReaderVersion is less than " + MIN_VERSION_SUPPORTS_READER_FEATURES);
         }
+        // When reader version is 3, readerFeatures must exist in protocol - https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features
```
ebyhr, line 59: > Avoid writing illegal `ProtocolEntry`

Is this change backward-compatible? In particular, could it cause failures when reading tables that were created with older Trino versions?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1352,12 +1352,12 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
                             transactionLogWriter.appendRemoveFileEntry(new RemoveFileEntry(addFileEntry.getPath(), addFileEntry.getPartitionValues(), writeTimestamp, true, Optional.empty()));
                         }
                     }
-                    protocolEntry = protocolEntryForTable(tableHandle.getProtocolEntry().minReaderVersion(), tableHandle.getProtocolEntry().minWriterVersion(), containsTimestampType, tableMetadata.getProperties());
+                    protocolEntry = protocolEntryForTable(ProtocolEntry.builder(tableHandle.getProtocolEntry()), containsTimestampType, tableMetadata.getProperties());
```
ebyhr, line 1355:
```
Preserve table features on `CREATE OR REPLACE TABLE`
According to
https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features-for-new-and-existing-tables,
retain existing read and write features when executing
 `CREATE OR REPLACE TABLE`, as previous table data may
 still depend on them.
```

Remove redundant leading spaces from the bottom 2 sentences.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
chenjian2664: It's not backward-compatible, reading existing table that created by Trino old versions might fail

ebyhr: I don't think we want to backward-compatibility in this case. The existing tables are already there, and users can't fix it easily. Does it make sense to remove these checks, or replace with warn logger?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
```diff
@@ -47,16 +48,39 @@ public record ProtocolEntry(
...
     {
         if (minReaderVersion < MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isPresent()) {
             throw new IllegalArgumentException("readerFeatures must not exist when minReaderVersion is less than " + MIN_VERSION_SUPPORTS_READER_FEATURES);
         }
+        // When reader version is 3, readerFeatures must exist in protocol - https://github.com/delta-io/delta/blob/master/PROTOCOL.md#table-features
+        //
+        // Spark done the similar logic here:
+        // https://github.com/delta-io/delta/blob/f5ebfbb10bda87b6b2b7c9032e86edaf40d446bc/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L155-L157
+        if (minReaderVersion >= MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isEmpty()) {
+            log.warn("readerFeatures must exist when minReaderVersion is greater than or equal to " + MIN_VERSION_SUPPORTS_READER_FEATURES);
```
ebyhr: We avoid string concat in the logger. Same for other places. 

Also, consider using the debug log level if the message is not actionable. Otherwise, users will see it without understanding how to resolve the issue.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
```diff
@@ -140,6 +140,24 @@ public void testCreateOrReplaceTableWithSchemaChangeOnCheckpoint()
         }
     }
 
+    @Test(groups = {DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
+    public void testCreateOrReplaceTableWithOnlyFeaturesChangedOnTrino()
+    {
+        String tableName = "test_create_or_replace_table_discard_features_" + randomNameSuffix();
+
+        onTrino().executeQuery("CREATE TABLE delta.default." + tableName + " (x INT) " +
+                "WITH (location = 's3://" + bucketName + "/databricks-compatibility-test-" + tableName + "', deletion_vectors_enabled = true)");
```
ebyhr: nit: "databricks-compatibility" doesn't make sense since this test doesn't run on Databricks cluster.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
chenjian2664: Make sense, updatded to debug level.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeCreateOrReplaceTableAsSelectCompatibility.java
chenjian2664: renamed to "delta-lake-test"

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/ProtocolEntry.java
```diff
@@ -61,7 +61,7 @@ public record ProtocolEntry(
         // Spark done the similar logic here:
         // https://github.com/delta-io/delta/blob/f5ebfbb10bda87b6b2b7c9032e86edaf40d446bc/spark/src/main/scala/org/apache/spark/sql/delta/actions/actions.scala#L155-L157
         if (minReaderVersion >= MIN_VERSION_SUPPORTS_READER_FEATURES && readerFeatures.isEmpty()) {
-            log.warn("readerFeatures must exist when minReaderVersion is greater than or equal to " + MIN_VERSION_SUPPORTS_READER_FEATURES);
+            log.debug("readerFeatures must exist when minReaderVersion is greater than or equal to %s", MIN_VERSION_SUPPORTS_READER_FEATURES);
```
ebyhr, line 64: FYI, this change should have went to the first commit.

chenjian2664: you are right :)

PR #27944, opened 2026-01-17
trino-weaviate: Add connector for Weaviate 

bevzzz:
## Description

This PR introduces a Trino plugin for [Weaviate](https://github.com/weaviate/weaviate), an open-source vector database. For our first iteration we would like to provide a minimal connector which only supports simple SELECT statements.

The resulting connector uses Weaviate's cursor API for efficient data fetching (with configurable page sizes), allows querying both single- and multi-tenant collections, and supports all Weaviate's data types.

Unfortunately, our latest `client6:6.0.0` version has conflicting gRPC/protobuf-java dependency versions, so the CI build will likely fail. We are planning to address this during the next week (probably by releasing a dedicated `6.0.0-trino` artifact), after which I'll update the PR.

Trino is a fantastic project, and we hope that `trino-weaviate` will be a valuable contribution to its ecosystem. Working on this PR has been my first encounter with Trino, so I might've overlooked something. Looking forward to hearing your feedback.

We've been looking into retiring our GraphQL API and, with that, exploring the alternatives for interactive data exploration. At one point we discovered Trino, thought it an impressive project, and decided to take a shot at developing a plugin for Weaviate.

## Release notes

```markdown
## Weaviate
* Add Weaviate connector. ({issue}`N/A`)
```


.github/config/labeler-config.yml
Praveen2112, line 1: This indentation change is not required.

plugin/trino-weaviate/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.inject</groupId>
+            <artifactId>guice</artifactId>
+            <classifier>classes</classifier>
```
Praveen2112, line 28: Any specific reason for this classifier ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/Decoder.java
```diff
@@ -0,0 +1,178 @@
...
+
+    private static void decodePrimitive(BlockBuilder blockBuilder, Type type, Object raw)
+    {
+        if (raw == null) {
+            requireNonNull(blockBuilder, "blockBuilder is null");
+            blockBuilder.appendNull();
+            return;
+        }
+        switch (type) {
+            case VarcharType t -> t.writeString(blockBuilder, (String) raw);
```
Praveen2112: nit: Can we avoid abbrevation here ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateService.java
```diff
@@ -0,0 +1,246 @@
...
+    private final WeaviateClient w;
+    private final ConsistencyLevel consistencyLevel;
+    private final OptionalInt pageSize;
+
+    @Inject
+    public WeaviateService(WeaviateConfig config)
+    {
+        requireNonNull(config, "config is null");
+
+        Config.Custom cfg = new Config.Custom();
```
Praveen2112: nit: Can we avoid abbrevations ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateConfig.java
```diff
@@ -0,0 +1,268 @@
...
+    private Duration timeout;
+    private ConsistencyLevel consistencyLevel;
+
+    private String apiKey;
+    private String accessToken;
+    private String refreshToken;
+    private Duration accessTokenLifetime;
+    private String username;
+    private String password;
+    private String clientSecret;
```
Praveen2112, line 47: Can we start with a basic auth and extend them as a follow-up PRs ? In case of authentication we might need to isolate the config properties for each auth mechanism.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateMetadata.java
```diff
@@ -0,0 +1,151 @@
...
+    @Override
+    public ConnectorTableHandle getTableHandle(
+            ConnectorSession session,
+            SchemaTableName name,
+            Optional<ConnectorTableVersion> startVersion,
+            Optional<ConnectorTableVersion> endVersion)
+    {
+        requireNonNull(name, "name is null");
+
+        WeaviateTableHandle tableHandle = weaviateService.getTableHandle(name.getTableName());
```
Praveen2112, line 71: Is a table name is mapped to only one tenant ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateMetadata.java
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    @Override
+    public List<SchemaTableName> listTables(ConnectorSession session, Optional<String> schemaName)
+    {
+        requireNonNull(schemaName, "schemaName is null");
+
+        ImmutableList.Builder<SchemaTableName> tables = ImmutableList.builder();
+
+        String schema = schemaName.orElse(null);
```
Praveen2112: I guess this assumption would fail for certain types of queries - if `schemaName` is passed as an `Optional#empty` we should scan the tables for all schema. But setting the schema as `null` would fail when creating `SchemaTableName`

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestDecoder.java
```diff
@@ -0,0 +1,301 @@
...
+import static io.trino.spi.type.DoubleType.DOUBLE;
+import static io.trino.spi.type.IntegerType.INTEGER;
+import static io.trino.spi.type.TimestampType.TIMESTAMP_MILLIS;
+import static io.trino.spi.type.Timestamps.MICROSECONDS_PER_MILLISECOND;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+
+public class TestDecoder
+{
+    @ParameterizedTest
+    @MethodSource("testCases")
```
Praveen2112, line 54: Can we avoid parameterized test ? In the long term parameterized test makes the code a bit less readable. Instead can we try like we handle in `TypeMappingTest` ?

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestWeaviate.java
```diff
@@ -0,0 +1,266 @@
...
+    static final OffsetDateTime NOW = OffsetDateTime.now();
+    static final GeoCoordinates GEO = new GeoCoordinates(123f, 456f);
+
+    private WeaviateServer server;
+    private WeaviateClient client;
+
+    public TestWeaviate()
+            throws IOException
+    {
+        server = new WeaviateServer();
```
Praveen2112: Can we create `WeaviateServer` as a part of `createQueryRunner` this would require us to create only the query runner for future test cases

plugin/trino-weaviate/pom.xml
bevzzz: No, not really, I think I saw it used in some other connector and figured it was required that way. Let me try to remove it.

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestWeaviate.java
bevzzz: Do you mean instantiating `WeaviateServer` in the `createQueryRunner` method, or making `WeaviateServer` an attribute of the query runner itself? I'm guessing it's the ~latter~ former, but just to clarify with you.

plugin/trino-weaviate/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+    </parent>
+
+    <artifactId>trino-weaviate</artifactId>
+    <packaging>trino-plugin</packaging>
+    <name>${project.artifactId}</name>
+    <description>Trino - Weaviate connector</description>
+
+    <properties>
+        <dep.client6.version>6.0.1-trino</dep.client6.version>
+        <dep.testcontainers.version>2.0.2</dep.testcontainers.version>
```
Praveen2112: We already refer to version 2.0.3  in `pom.xml`, any specific reason for this version ?

plugin/trino-weaviate/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+    </properties>
+
+    <dependencies>
+
```
Praveen2112: nit: Can we remove the empty new line ?

plugin/trino-weaviate/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>testcontainers-weaviate</artifactId>
+            <version>${dep.testcontainers.version}</version>
+            <scope>test</scope>
+        </dependency>
+
```
Praveen2112: nit: Remove new line.

plugin/trino-weaviate/pom.xml
```diff
@@ -0,0 +1,186 @@
...
+    </dependencies>
+
```
Praveen2112: nit: Remove line.

.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ on:
       - master
   pull_request:
   repository_dispatch:
-    types: [test-with-secrets-command]
+    types: [ test-with-secrets-command ]
```
Praveen2112: nit: Can we revert this change

.github/workflows/ci.yml
```diff
@@ -637,7 +639,7 @@ jobs:
           $MAVEN test ${MAVEN_TEST} -pl ${{ matrix.modules }} ${{ format('-P {0}', matrix.profile) }}
       - name: Cloud Delta Lake Tests
         id: tests-delta
-      # Cloud tests are separate because they are time intensive, requiring cross-cloud network communication
+        # Cloud tests are separate because they are time intensive, requiring cross-cloud network communication
```
Praveen2112: Revert this indentation change

docs/src/main/sphinx/ext/download.py
```diff
@@ -118,7 +119,7 @@ def download_gh_link_role(role, rawtext, text, lineno, inliner, options={}, cont
         return [node], []
 
     def download_mc_link_role(role, rawtext, text, lineno, inliner, options={}, content=[]):
-        version = app.config.release 
+        version = app.config.release
```
Praveen2112: Can we revert the indentation changes

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateConfig.java
```diff
@@ -0,0 +1,268 @@
...
+        return scheme;
+    }
+
+    @Config("weaviate.scheme")
+    @ConfigDescription("Connection URL scheme (http/https)")
+    public WeaviateConfig setScheme(String scheme)
+    {
+        this.scheme = scheme;
+        return this;
+    }
```
Praveen2112, line 64: In case of `https` we also need to configure the TLS specific configuration right ? Like keystore and truststore - Can we add support for https as a follow-up ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateTableHandle.java
```diff
@@ -0,0 +1,105 @@
...
+
+    public WeaviateTableHandle(CollectionConfig collection)
+    {
+        this(
+                Optional.of(DEFAULT_SCHEMA),
+                requireNonNull(collection, "collection is null").collectionName(),
+                collection.description(),
+                OptionalLong.empty(),
+                makeColumns(collection.properties(), collection.vectors()));
+    }
```
Praveen2112: Can we move this to `WeaviateConnectorMetadata` - `WeaviateTableHandle` is more of a POJO and any additional way of creating it should be moved to some sort of util class or invoker side.

plugin/trino-weaviate/pom.xml
Praveen2112: Apologies. It looks like we need it.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateService.java
```diff
@@ -0,0 +1,246 @@
...
+        this.weaviateClient = new WeaviateClient(config.build());
+    }
+
+    List<String> listTenants()
+    {
+        ImmutableList.Builder<String> tenants = ImmutableList.<String>builder();
+        try {
+            List<CollectionConfig> collections = weaviateClient.collections.list();
+
+            for (var c : collections) {
```
Praveen2112: nit: Can we avoid abbrevations ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateService.java
```diff
@@ -0,0 +1,246 @@
...
+                }
+                weaviateClient.collections.use(c.collectionName())
+                        .tenants.list().stream()
+                        .map(Tenant::name).forEach(tenants::add);
+            }
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
+        return tenants.build();
```
Praveen2112: How about 
```
            weaviateClient.collections.list().stream()
                    .filter(collectionConfig -> !collectionConfig.multiTenancy().enabled())
                    .map(CollectionConfig::collectionName)
                    .flatMap(collectionName -> weaviateClient.collections.use(collectionName).tenants.list().stream())
                    .map(Tenant::name)
                    .collect(toImmutableList());
```

plugin/trino-weaviate/pom.xml
bevzzz: So it appears that all other connectors import `guice` with the `classes` classifier. Removing it prevents the project from building.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/CountQueryPageSource.java
```diff
@@ -0,0 +1,84 @@
...
+        int pageSize = toIntExact(Math.min(MAX_PAGE_SIZE, remaining));
+        remaining -= pageSize;
+
+        return SourcePage.create(pageSize);
+    }
+
+    @Override
+    public boolean isFinished()
+    {
+        return true;
```
Praveen2112: This would return only a singe source page right ? We might need to check if remaining reaches 0.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/FetchObjectsQueryPageSource.java
```diff
@@ -0,0 +1,110 @@
...
+    public FetchObjectsQueryPageSource(
+            WeaviateService weaviateService,
+            WeaviateTableHandle tableHandle,
+            List<WeaviateColumnHandle> columns)
+    {
+        requireNonNull(weaviateService, "weaviateService is null");
+        requireNonNull(tableHandle, "tableHandle is null");
+        requireNonNull(columns, "columns is null");
+
+        this.columns = columns;
```
Praveen2112: nit: ImmutableList#copyOf

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestWeaviate.java
bevzzz: I made `WeaviateQueryRunner.Builder` accept `WeaviateServer` as a DI. It will register it with the `DistributedQueryRunner` as a resource, so that it is disposed of when the query runner is closed.

Now both resources are created as part of the `createQueryRunner` step.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/FetchObjectsQueryPageSource.java
```diff
@@ -0,0 +1,110 @@
...
+        this.columnBuilders = columns.stream()
+                .map(WeaviateColumnHandle::trinoType)
+                .map(type -> type.createBlockBuilder(null, 1))
+                .toArray(BlockBuilder[]::new);
+
+        long start = System.nanoTime();
+        this.iterator = weaviateService.getTableIterator(tableHandle);
+        this.readTimeNanos += System.nanoTime() - start;
```
Praveen2112, line 57: Tracking this time would be incorrect right ? I assume this happens as a PageIterator - so if the table is too big we would make multiple request to the server right ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/Decoder.java
```diff
@@ -0,0 +1,178 @@
...
+        }
+    }
+
+    private static void decodePrimitive(BlockBuilder blockBuilder, Type type, Object raw)
+    {
+        if (raw == null) {
+            requireNonNull(blockBuilder, "blockBuilder is null");
+            blockBuilder.appendNull();
+            return;
+        }
```
Praveen2112: This is common for array, row and primitive type. So can we move it to `decode` before applying the switch case ?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateColumnHandle.java
```diff
@@ -0,0 +1,150 @@
...
+            case DataType.TEXT_ARRAY -> TEXT_ARRAY;
+            case DataType.BOOL_ARRAY -> BOOL_ARRAY;
+            case DataType.INT_ARRAY -> INT_ARRAY;
+            case DataType.NUMBER_ARRAY -> NUMBER_ARRAY;
+            case DataType.DATE_ARRAY -> DATE_ARRAY;
+            case DataType.OBJECT -> nestedObjectType(property.nestedProperties());
+            case DataType.OBJECT_ARRAY -> new ArrayType(nestedObjectType(property.nestedProperties()));
+            default -> throw new TrinoException(WEAVIATE_UNSUPPORTED_DATA_TYPE, dataType + " is not supported");
+        };
+    }
```
Praveen2112: Can we move all this logic on mapping WeaviateType to TrinoType as a util class like `WeaviateUtil` ?

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestDecoder.java
Praveen2112, line 1: Along with this - Can we have a TypeMapping test for this connector as well. ?

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestWeaviate.java
```diff
@@ -0,0 +1,261 @@
...
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+import java.io.IOException;
+import java.time.OffsetDateTime;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestWeaviate
```
Praveen2112: nit: `TestWeaviateConnectorTest`

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestWeaviate.java
```diff
@@ -0,0 +1,261 @@
...
+        extends BaseConnectorTest
+{
+    private static final String LARGE_COLLECTION = "TrinoTestLargs";
+    private static final String COLLECTION_NO_TENANT = "TrinoTestNoTenant";
+    private static final String COLLECTION_MULTI_TENANT = "TrinoCollectionMultiTenant";
+    private static final String JOHN_DOE = "john_doe";
+    private static final String JANE_DOE = "jane_doe";
+
+    static final OffsetDateTime NOW = OffsetDateTime.now();
```
Praveen2112: Can we have a specific point in time instead of ever increasing `NOW` ?

Praveen2112 (commented):
Thanks a lot for working on this.

plugin/trino-weaviate/pom.xml
Praveen2112: Adding this classifier allows us to use unshaded variant of guice.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/FetchObjectsQueryPageSource.java
bevzzz: This tracks the time elapsed in the constructor, as `getTableIterator` will include some network calls.

I now see that we don't track that time anywhere in the `getNextSourcePage`, so you're right it doesn't reflect additional page fetches. Will fix.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateService.java
bevzzz: Sure, that works 👌

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateMetadata.java
bevzzz: A table may have a single or multiple tenants -- it's part of that table's configuration. Because Weaviate doesn't have "schemas" in the conventional sense, we use `schema` to model tenants.

For a normal table these 2 queries are equivalent:

```sql
SELECT * FROM cities;
SELECT * FROM default.cities;
```
In this case there are _no tenants_ at all.

For a multi-tenant table the schema name is required, because exactly 1 tenant can be queries at a time.

```sql
SELECT * FROM john_doe.things; -- GOOD: query things under "john_doe" tenant

SELECT * FROM things; -- BAD: there's no way to query ALL tenants
SELECT * FROM default.things; -- BAD: there's no way to query ALL tenants
```

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateMetadata.java
bevzzz: > we should scan the tables for all schema

This is not possible, as multiple schemas implies multiple tenants; exactly 1 tenant can be queried at once (see comment above).

What I think this line should do instead is:

```java
String schema = schemaName.orElse(DEFAULT_SCHEMA);
```

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestDecoder.java
bevzzz: My understanding is that using this pattern:

```java
SqlDataTypeTest.create()
  .addRoundtrip("boolean", "true", BOOLEAN)
  .execute(getSession(), trinoCreateAsSelect("test_boolean"));
```

requires the connector to support `CREATE TABLE` query -- something WeaviateConnector does not do at the moment.

https://github.com/trinodb/trino/blob/b2f6e0ff8f8af16250cf85bd24071e8a10ef527a/testing/trino-testing/src/main/java/io/trino/testing/sql/TestTable.java#L52

I tried getting this to work instead, whereby, afaiu, the table is created directly by the test (and not via the query executor)

```java
SqlDataTypeTest.create()
  .addRoundtrip("boolean", "true", BOOLEAN)
  .execute(getSession(), weaviateCreateAsSelect("test_boolean"));
```

but I'm hitting another issue. After setting up the target table, the testing framework will send a `SELECT *` and expect `BOOLEAN true` to be returned.

https://github.com/trinodb/trino/blob/b2f6e0ff8f8af16250cf85bd24071e8a10ef527a/testing/trino-testing/src/main/java/io/trino/testing/datatype/SqlDataTypeTest.java#L101

Every object in Weaviate has a non-hidden `_uuid` field, so a query like `SELECT *` will return a tuple `(<uuid>, true)` instead of a single value. This breaks the next assertion, which expects `boolean` type but finds a `varchar`. The `.addRoundtrip()` method also does not provide a way to expect a tuple (understandably).

Have I overlooked something?

___

I wanted to have `TestDecoder` as a dedicated unit test for the Decoder util, and I've seen other connectors to that, e.g. Avro. Happy to move these tests to `TestWeaviateConnectorTest`, if that suits the repo better.

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/TestDecoder.java
bevzzz: > Instead can we try like we handle in TypeMappingTest ?

Please see my comment about `TypeMappingTest` above.

> Can we avoid parameterized test ?

I thought it was an appropriate place for a parameterized test because the surface of each test case is small -- type, input, expected output; besides, I saw it was used in other connectors (Parquet) and in the Trino core iself (trino-main and trino-parser). Happy to rewrite them as individual test cases, if that suits the repo better.

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateConfig.java
bevzzz: The client library does not currently expose a way to provide custom keystore/trustore and unfortunately I'm not sure when we might get a chance to add that. 

Being able to use `https` would allow our users to connect to remote clusters securely, so we'd like to provide this configuration option in the first version already.

Do you think that'd be possible?

plugin/trino-weaviate/src/main/java/io/trino/plugin/weaviate/WeaviateConfig.java
bevzzz: > Can we start with a basic auth [...] ?

Sure, I think we should be fine with just the `apiKey` authentication. This is the simplest way to have an authenticated connection rn.

> isolate the config properties for each auth mechanism.

Is this something you have in mind?

https://github.com/trinodb/trino/blob/b2f6e0ff8f8af16250cf85bd24071e8a10ef527a/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/ElasticsearchConfig.java#L49-L55

Only in this case it'd be Security.API_KEY, Security.PASSWORD, Security.TOKEN, etc.?

bevzzz:
Thanks for your review. I've resolved the conversations related to trivial changes.

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateServer.java
```diff
@@ -0,0 +1,82 @@
...
+    public WeaviateServer(String image)
+    {
+        this(Network.SHARED, image);
+    }
+
+    public WeaviateServer(Network network, String image)
+    {
+        container = new WeaviateContainer(image);
+        container.withNetwork(network);
+        container.withNetworkAliases("weaviate-server");
```
findepi: I guess the network is always `Network.SHARED`, so fixed alias may lead to conflicts.
What about not having these two lines?

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateServer.java
```diff
@@ -0,0 +1,82 @@
...
+        container.withStartupTimeout(Duration.ofMinutes(5));
+
+        container.start();
+    }
+
+    public WeaviateClient getClient()
+    {
+        if (!container.isRunning()) {
+            container.start();
```
findepi: It's started in ctor, so it's either running, or died.
the `getClient()` shouldn't try to restart it. remove these lines

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateServer.java
```diff
@@ -0,0 +1,82 @@
...
+    {
+        this(Network.SHARED, image);
+    }
+
+    public WeaviateServer(Network network, String image)
+    {
+        container = new WeaviateContainer(image);
+        container.withNetwork(network);
+        container.withNetworkAliases("weaviate-server");
+        container.withStartupTimeout(Duration.ofMinutes(5));
```
findepi: Is this different than default startup timeout of `org.testcontainers.weaviate.WeaviateContainer`?
Why? document

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateQueryRunner.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    static Builder builder(WeaviateServer weaviateServer)
+    {
+        HostAndPort httpAddress = weaviateServer.getHttpAddress();
+        HostAndPort grpcAddress = weaviateServer.getGrpcAddress();
+        return new Builder(weaviateServer)
+                .addConnectorProperty("scheme", "http")
+                .addConnectorProperty("http-host", "localhost")
+                .addConnectorProperty("grpc-host", "localhost")
```
findepi: Add `Test...Plugin` test class. See `TestIcebergPlugin` for example.

- the configs specific to weaviate connector should typically be prefixed with `weaviate.`
- Is `http-host` always eqaul to `grpc-host`? do we need both?

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateQueryRunner.java
```diff
@@ -0,0 +1,81 @@
...
+
+        private Builder(WeaviateServer weaviateServer)
+        {
+            super(testSessionBuilder().setCatalog(WEAVIATE).setSchema("default").build());
+            this.weaviateServer = weaviateServer;
+        }
+
+        private Builder addConnectorProperty(String key, String value)
+        {
+            this.connectorProperties.put(WEAVIATE + "." + key, value);
```
findepi: it is not a big deal for the caller to prefix properties with "weaviate.", and helps readability

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateQueryRunner.java
```diff
@@ -0,0 +1,81 @@
...
+import io.trino.testing.DistributedQueryRunner;
+
+import java.util.HashMap;
+import java.util.Map;
+
+import static io.trino.testing.TestingSession.testSessionBuilder;
+
+public final class WeaviateQueryRunner
+{
+    static final String WEAVIATE = "weaviate";
```
findepi: give this constant semantics -- eg `CATALOG_NAME` -- and use it only when the use matches that semantic

alternatively, inline.

bevzzz: > use it only when the use matches that semantic

That is a great advice, thank you!

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateQueryRunner.java
bevzzz: > Is http-host always eqaul to grpc-host? do we need both?

They are usually equal when connecting to a local instance, but in many cases (including instances running it the Weaviate Cloud) the two hosts differ. The WeaviateConnector does not differentiate between local/remote connections though, and always creates a "full" configu, so both options should be settable via `weaviate.properties`.

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateServer.java
bevzzz: We don't need to override the default value here, removed

plugin/trino-weaviate/src/test/java/io/trino/plugin/weaviate/WeaviateServer.java
bevzzz: We don't need to override the default value here, removed

PR #27955, opened 2026-01-19
Avoid reading delete manifests for iceberg table statistics

raunaqmorarka:
## Description
Reduces some unnecessary complexity in this code.
The main improvement is to avoid reading delete manifests for iceberg table statistics.
Using the planFiles API for getting file statistics results in
unnecessary reads of delete manifests. It also consumes significant
resources in `org.apache.iceberg.DeleteFileIndex` when there are a
significant number of deletes in the table.

## Release notes

```markdown
## Iceberg
* Reduce planning time of queries with scans on tables with delete files. ({issue}`27955`)
```


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergFileOperations.java
```diff
@@ -657,6 +658,15 @@ public void testShowStatsForTable()
...
                         .add(new FileOperation(MANIFEST, "InputFile.newStream"))
                         .build());
+
+        assertUpdate("DELETE FROM test_show_stats WHERE age = 10", 1);
+        assertFileSystemAccesses("SHOW STATS FOR test_show_stats",
+                ImmutableMultiset.<FileOperation>builder()
+                        .add(new FileOperation(METADATA_JSON, "InputFile.newStream"))
+                        .add(new FileOperation(SNAPSHOT, "InputFile.length"))
+                        .add(new FileOperation(SNAPSHOT, "InputFile.newStream"))
+                        .addCopies(new FileOperation(MANIFEST, "InputFile.newStream"), 2)
```
chenjian2664: the additional access is because we add one more delete manifest?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
```diff
@@ -153,38 +163,50 @@ public static TableStatistics makeTableStatistics(
...
 
         List<Types.NestedField> columns = icebergTable.schema().columns()
                 .stream()
                 .filter(column -> columnIds.contains(column.fieldId()))
                 .collect(toImmutableList());
         IcebergStatistics.Builder icebergStatisticsBuilder = new IcebergStatistics.Builder(columns, typeManager);
-        try (CloseableIterable<FileScanTask> fileScanTasks = tableScan.planFiles()) {
-            fileScanTasks.forEach(fileScanTask -> {
-                if (!partitionDomain.isAll() && !partitionDomain.includesNullableValue(utf8Slice(fileScanTask.spec().partitionToPath(fileScanTask.partition())))) {
+        try (CloseableIterable<DataFile> dataFiles = new ParallelIterable<>(dataFileIterables, icebergPlanningExecutor)) {
```
chenjian2664, line 182: Note this seems missing the logic to filter data file as [here](https://github.com/apache/iceberg/blob/8967729beac20f1fbcbe6b1f43ae010525a2c6f5/core/src/main/java/org/apache/iceberg/ManifestGroup.java#L333-L339)
though that filter seems not used at previous :) , just wanna discuss do think it worth that apply similar logic here - use the `filter` to build a evaluator for the data file

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -6282,109 +6282,6 @@ public void testCollectingStatisticsWithPathColumnPredicate()
         assertUpdate("DROP TABLE " + tableName);
     }
 
-    @Test
-    public void testCollectingStatisticsWithFileModifiedTimeColumnPredicate()
```
chenjian2664, line 6288: Can we keep the test but with a disabled annotation with a concise comment?

chenjian2664 (commented):
Mainly has some confusions on last 2 commits

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergFileOperations.java
raunaqmorarka: yes

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
raunaqmorarka: There's no benefit to keeping a disabled test, the code can be brought back from history if its ever needed again

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
raunaqmorarka: That filter is used only for predicates on `file_path`. We don't need to do that since we apply the path predicate separately in our own code.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
```diff
@@ -146,46 +146,50 @@ public static TableStatistics makeTableStatistics(
...
+            dataFiles.forEach(dataFile -> {
+                PartitionSpec spec = icebergTable.specs().get(dataFile.specId());
+                if (!partitionDomain.isAll() && !partitionDomain.includesNullableValue(utf8Slice(spec.partitionToPath(dataFile.partition())))) {
                     return;
                 }
-                if (!pathDomain.isAll() && !pathDomain.includesNullableValue(utf8Slice(fileScanTask.file().location()))) {
+                if (!pathDomain.isAll() && !pathDomain.includesNullableValue(utf8Slice(dataFile.location()))) {
                     return;
                 }
-                if (!fileModifiedTimeDomain.isAll()) {
```
chenjian2664, line 181: isn't removing this line result in incorrect statistics for queries that include a file modified time predicate?

raunaqmorarka: Yes, the statistics will be inaccurate. But to get accurate statistics, we are making HEAD call on every data file selected by the query in a serial fashion from the coordinator. This is not a reasonable trade-off just to make the estimates more accurate. CBO estimates don't need to be precise to be useful.

chenjian2664: If we decided to remove the column here then it makes me to think maybe it is "wrong" for us to expose such metadata column? 

I think it's ok to remove the check now, but since the time predicate is still useful, i,e in optimize, if it is expensive to get maybe we should consider another time concept, sth like, "committed timestamp", it could use file sequence number to find the corresponding snapshot committed timestamp.

Also the `cacheKey` logic seems needs to be updated https://github.com/trinodb/trino/blob/bc2945b4201d5217629a6228b9dcd4bb90b202a9/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java#L3815, the predicate should filter `$file_modified_time` out.

raunaqmorarka: It's fine to have the functionality around `file_modified_time`, we just shouldn't make statistics generation expensive. The estimates are supposed to be fast and best effort, we shouldn't do anything resource intensive to make it accurate.

chenjian2664: Could you leave a code comment here for it, we will know why we don't consider the filter out the file using `file_modified_time`

raunaqmorarka: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -71,8 +71,8 @@ public static class Builder
         private long fileCount;
         private long size;
         private final Map<Integer, ColumnStatistics> columnStatistics = new HashMap<>();
```
findinpath, line 73: From the commit message

> Statistics are allowed to be in-exact, so there is no need to add complexity by skipping statistics when they are not found in one file.

Statistics are allowed to be not exact, so there is no need to add complexity invalidating statistics when they are not found in one file.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -158,39 +151,17 @@ public IcebergStatistics build()
...
-                    existingCount.isPresent() && newCount.isPresent() ? Optional.of(existingCount.get() + newCount.get()) : Optional.empty());
-        }
-
-        private void updateNanCountStats(int id, Optional<Long> nanCount)
-        {
-            // If one file is missing nanCounts for a column, invalidate the estimate
-            nanCounts.merge(id, nanCount, (existingCount, newCount) ->
-                    (existingCount.isPresent() && newCount.isPresent()) ? Optional.of(existingCount.get() + newCount.get()) : Optional.empty());
-        }
-
```
findinpath, line 193: This change represents a functionality change.

If my reading is correct, the `SHOW STATS` for an Iceberg table will deliver data that corresponds to the metadata that exists for the data files of the table.
Can there be any negative consequences to it?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -80,8 +76,7 @@ public Builder(
                 TypeManager typeManager)
         {
             this.typeManager = requireNonNull(typeManager, "typeManager is null");
-            this.fieldIdToTrinoType = columns.stream()
-                    .collect(toImmutableMap(Types.NestedField::fieldId, column -> toTrinoType(column.type(), typeManager)));
```
findinpath, line 84: this could still stay here, no need to calculate for each file

```
io.trino.spi.type.Type trinoType = toTrinoType(column.type(), typeManager);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
```diff
@@ -169,6 +164,10 @@ public static TableStatistics makeTableStatistics(
                                 .collect(toImmutableList()))
                 .planWith(icebergPlanningExecutor);
 
+        List<Types.NestedField> columns = icebergTable.schema().columns()
```
findinpath, line 167: `icebergTable.schema()` -> `snapshotSchema`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3812,7 +3812,8 @@ public TableStatistics getTableStatistics(ConnectorSession session, ConnectorTab
                 originalHandle.getPartitionSpecJson(),
                 originalHandle.getFormatVersion(),
                 originalHandle.getUnenforcedPredicate(),
-                originalHandle.getEnforcedPredicate(),
+                // Skip $file_modified_time in cache key as the statistics do not depend on it
```
findinpath, line 3815: If the use case is 

```
SHOW STATS FOR (SELECT * FROM t1 WHERE \"$file_modified_time\" = ....)
```

and the filter is simply ignored, we should rather point through an exception that the filter is not supported.

I don't see the gain in removing this functionality.
If the user wants to filter on `file_modified_time`, this will incur as you pointed out serial retrieval of file modified time for each file. For all the other use-cases though, there is no performance decrease in having this functionality.

Advocating for keeping the functionality.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
```diff
@@ -14,11 +14,16 @@
 package io.trino.plugin.iceberg;
```
findinpath, line 15: Change the title of the PR to `Avoid reading delete manifests for iceberg table statistics`

That's the main added value of this contribution.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
```diff
@@ -153,38 +163,50 @@ public static TableStatistics makeTableStatistics(
...
+
+        NonEvictableLoadingCache<Integer, ManifestEvaluator> manifestPartitionFilterEvaluators = buildNonEvictableCache(
+                CacheBuilder.newBuilder().maximumSize(1000),
+                CacheLoader.from(specId -> {
+                    PartitionSpec partitionSpec = icebergTable.specs().get(specId);
+                    return ManifestEvaluator.forRowFilter(filter, partitionSpec, true);
+                }));
+        List<ManifestFile> filteredManifests = icebergTable.snapshot(snapshotId)
+                // We only need to look at data files to estimate statistics
+                .dataManifests(icebergTable.io())
```
findinpath, line 176: Main performance gain is here.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: There's no requirement for statistics to be fully accurate, so there's no reason to block any queries using that filter. I don't think there are good reasons in general to be using `$file_modified_time` in SELECT queries or SHOW STATS queries. I'm stopping short of removing that metadata column altogether because that would be a breaking change that I don't want to pursue for now. But still I'd rather not carry the code complexity and unexpectedly high runtime overhead for computing accurate stats for this case.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
raunaqmorarka: `toTrinoType` is a switch case, which is internally just a table look-up. Using a hashmap to look-up the value is still essentially doing the same thing with more verbosity.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
raunaqmorarka: Nan statistics are not used by CBO, so that won't matter.
Sizes are used only for string columns. We will now supply a nulls count and size value where we didn't previously. It's likely to improve the estimate because the most common scenario where this will happen is when a new column is added to a table.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
raunaqmorarka: This is unchanged from the existing code. We were using `snapshotSchema` only to look-up column name from column id for satisfying the planFiles API

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/TableStatisticsReader.java
```diff
@@ -205,6 +204,10 @@ public static TableStatistics makeTableStatistics(
                 snapshotId,
                 columnIds);
 
+        Map<Integer, org.apache.iceberg.types.Type> idToType = columns.stream()
+                .map(column -> Maps.immutableEntry(column.fieldId(), column.type()))
+                .collect(toUnmodifiableMap(Map.Entry::getKey, Map.Entry::getValue));
+
```
Praveen2112, line 216: nit: Is it more of re-arrangement of code ?

raunaqmorarka: This code block is moved down so that we use projected `columns` list rather than the full `icebergTable.schema().columns()`

PR #27958, opened 2026-01-20
Replace `return` with `continue` in SET DATA TYPE tests

ebyhr:
## Description

Otherwise, the test may exit unexpectedly without running all cases.

## Release notes


chenjian2664:
The ci looks related

wendigo (approved):

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -3440,7 +3440,7 @@ public void testSetColumnTypes()
                 if (setup.unsupportedType) {
                     assertThatThrownBy(setColumnType::run)
                             .satisfies(this::verifySetColumnTypeFailurePermissible);
-                    return;
```
findepi, line 3443: good find!

PR #27973, opened 2026-01-21
Avoid reading manifest list in applyFilter for single partition spec

raunaqmorarka:
## Description
We don't need to check the partition spec ids from the manifest list
when we know that there aren't multiple partition specs in the table.
Currently the manifest list read is cached for later re-use in splits generation
or in table statistics fetch. However, reading this during applyFilter is still 
worse because these reads will happen sequentially for all tables in a query 
during plan optimization. 
Splits generation on the other hand, is parallel for each scan and statistics
fetch will not be need for simple select queries with filter.

## Release notes

```markdown
## Iceberg
* Reduce planning time for simple queries with a scan and filter. ({issue}`issuenumber`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -3680,7 +3680,7 @@ public class IcebergMetadata
         }
     }
 
-    private static List<ManifestFile> loadAllManifestsFromSnapshot(Table icebergTable, Snapshot snapshot)
+    private static List<ManifestFile> loadAllManifestsFromSnapshot(Table icebergTable, @Nullable Snapshot snapshot)
     {
         if (snapshot == null) {
             return ImmutableList.of();


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3650,6 +3645,38 @@ else if (isMetadataColumnId(columnHandle.getId())) {
...
+
+        Snapshot snapshot = icebergTable.snapshot(snapshotId.get());
+        return loadDataManifestsFromSnapshot(icebergTable, snapshot).stream()
+                .map(ManifestFile::partitionSpecId)
+                .collect(toImmutableSet());
+    }
+
+    private static List<ManifestFile> loadDataManifestsFromSnapshot(Table icebergTable, Snapshot snapshot)
+    {
+        if (snapshot == null) {
```
findepi, line 3672: Worth annotating argument as nullable. Could be even better to handle snapshot-not-found by the caller, but i see it's aligned with pre-existing `loadAllManifestsFromSnapshot`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3650,6 +3645,25 @@ else if (isMetadataColumnId(columnHandle.getId())) {
...
+
+        Snapshot snapshot = icebergTable.snapshot(snapshotId.get());
+        return loadAllManifestsFromSnapshot(icebergTable, snapshot).stream()
+                .map(ManifestFile::partitionSpecId)
+                .collect(toImmutableSet());
+    }
+
     private static List<ManifestFile> loadAllManifestsFromSnapshot(Table icebergTable, Snapshot snapshot)
     {
         if (snapshot == null) {
```
findepi, line 3685: Worth annotating argument as nullable. But maybe even better to handle snapshot-not-found by the caller.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergFileOperations.java
```diff
@@ -662,8 +662,6 @@ public void testExplainSelect()
         assertFileSystemAccesses(withoutStatistics, "EXPLAIN SELECT * FROM test_explain WHERE age = 2",
                 ImmutableMultiset.<FileOperation>builder()
                         .add(new FileOperation(METADATA_JSON, "InputFile.newStream"))
-                        .add(new FileOperation(SNAPSHOT, "InputFile.length"))
-                        .add(new FileOperation(SNAPSHOT, "InputFile.newStream"))
```
findepi: nice!

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3659,11 +3659,24 @@ private static Set<Integer> getPartitionSpecIdsFromSnapshot(Table icebergTable,
         }
 
         Snapshot snapshot = icebergTable.snapshot(snapshotId.get());
-        return loadAllManifestsFromSnapshot(icebergTable, snapshot).stream()
+        return loadDataManifestsFromSnapshot(icebergTable, snapshot).stream()
```
findepi, line 3665: > Since we're primarily concerned about predicate pushdown on partitioning
of the data files, there's no need to consider delete manifests

Worth capturing this smart observation as a code comment?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3571,12 +3577,7 @@ public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(C
         else {
             Table icebergTable = catalog.loadTable(session, table.getSchemaTableName());
 
-            Set<Integer> partitionSpecIds = table.getSnapshotId().map(
-                            snapshot -> loadAllManifestsFromSnapshot(icebergTable, icebergTable.snapshot(snapshot)).stream()
-                                    .map(ManifestFile::partitionSpecId)
-                                    .collect(toImmutableSet()))
-                    // No snapshot, so no data. This case doesn't matter.
-                    .orElseGet(() -> ImmutableSet.copyOf(icebergTable.specs().keySet()));
+            Set<Integer> partitionSpecIds = getPartitionSpecIdsFromSnapshot(icebergTable, table.getSnapshotId());
```
chenjian2664, line 3580: Using a supplier would be even better :)

raunaqmorarka: We would still end up calling that supplier in most cases

PR #27976, opened 2026-01-21
Allow a JDBC connector to define the column type

### LINKED ISSUE #27890: Allow a JDBC connector to define the column type
Currently, it's difficult, if not impossible, for the connector to determine the JDBC types assigned to each column in a table.
This is because, at present, only the type returned by the database's JDBC driver is taken into account.

However, it's sufficient to add two methods to the [BaseJdbcClient](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java) class that can be overridden by the connector to make this possible.

Can I submit a pull request to enable this?

prrvchr:
## Description
 
This small change allows a connector to influence the creation of SQL types.

This is in response to [issue#27890](https://github.com/trinodb/trino/issues/27890). (fixes #27890)

## Release notes

```markdown
## Section
* Fix some things. ({issue}`27890`)
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
@@ -296,19 +296,25 @@ public abstract class BaseJdbcClient
         for (int column = 1; column <= metadata.getColumnCount(); column++) {
             // Use getColumnLabel method because query pass-through table function may contain column aliases
             String name = metadata.getColumnLabel(column);
-            JdbcTypeHandle jdbcTypeHandle = new JdbcTypeHandle(
-                    metadata.getColumnType(column),
-                    Optional.ofNullable(metadata.getColumnTypeName(column)),
-                    Optional.of(metadata.getPrecision(column)),
-                    Optional.of(metadata.getScale(column)),
-                    Optional.empty(), // TODO support arrays
-                    Optional.of(metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE));
+            JdbcTypeHandle jdbcTypeHandle = getColumnTypeHandle(metadata, column);
             Type type = toColumnMapping(session, connection, jdbcTypeHandle)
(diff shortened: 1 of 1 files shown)


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -310,6 +304,18 @@ protected List<JdbcColumnHandle> getColumns(ConnectorSession session, Connection
         return columns.build();
     }
 
+    protected JdbcTypeHandle getColumnTypeHandle(ResultSetMetaData metadata, int column)
```
findepi, line 308: if we pass `session` into this function, we should be able to remove `io.trino.plugin.oracle.OracleClient#getColumns` and have  `OracleClient` just implement this one.
cc @chenjian2664 


nit: Should the two `getColumnTypeHandle` overloads live close to each other?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -527,6 +520,19 @@ private Optional<RelationColumnsMetadata> finishCurrentTable()
         }
     }
 
+    protected JdbcTypeHandle getColumnTypeHandle(ResultSet resultSet, Optional<CaseSensitivity> caseSensitivity)
```
findepi, line 523: This should document what `ResultSet` is / how is obtained so that the implementor knows what to do.
... which will be a bit hard to document given current code flow.

prrvchr: Overloading a method is quite basic since you have the method you are overloading as an example?
And often I prefer that to explanations that are difficult to understand, like the ones you can find in the interfaces.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
prrvchr: Could we make a decision, as I really need this in the main branch because I'm having conflict issues between my active branches?
Thank you.

Issue #27891, opened 2026-01-09
Allow a JDBC Connector to define its default schema


prrvchr:
Currently, it's not impossible, for the connector to determine the default schema used.
This is because, at present, only the `java.sql.ResultSet` returned by the database's JDBC driver is taken into account.

However, it is sufficient to make rewritable the method [getRemoteTable(ResultSet resultSet)](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L1885) in the `BaseJdbcClient` class to make this possible.

Can I submit a pull request to enable this?

findepi:
There already are some connectors which have non-standard handling of catalogs and schemas: MySQL and MySQL derivatives (MariaDB, SingleStore), Druid, ClickHouse.
Maybe they can serve as an inspiration how to cope with some catalog/schema issues.

prrvchr:
The new implementation of the Firebird connector in [PR#30483](https://github.com/trinodb/trino/pull/30483) is a good example of what the `BaseJdbcClient` class is missing in order to support a database that does not handle schemas.

PR #27977, opened 2026-01-21
Allow a JDBC Connector to define its default schema

prrvchr:
## Description

SQLite is unique in that it does not support schemas but manages two dummy schemas in its SQL and DDL commands:
- main
- temp

This minor modification enables support for connectors like this.

This is in response to [issue#27891](https://github.com/trinodb/trino/issues/27891)

## Release notes

```markdown
## Section
* Fix some things. ({issue}`27891`)
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
@@ -1882,7 +1882,7 @@ public abstract class BaseJdbcClient
         return name;
     }
 
-    private static RemoteTableName getRemoteTable(ResultSet resultSet)
+    protected RemoteTableName getRemoteTable(ResultSet resultSet)
             throws SQLException
     {
         return new RemoteTableName(


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -1882,7 +1882,7 @@ protected String escapeObjectNameForMetadataQuery(String name, String escape)
         return name;
     }
 
-    private static RemoteTableName getRemoteTable(ResultSet resultSet)
+    protected RemoteTableName getRemoteTable(ResultSet resultSet)
```
findepi, line 1885: What would be SQLite's (or someone elses's) implementation of this method?

prrvchr: the code is:


```
    private static final String MAIN_SCHEMA = "main";

    @Override
    protected RemoteTableName getRemoteTable(ResultSet resultSet)
            throws SQLException
    {
        return new RemoteTableName(
                Optional.ofNullable(resultSet.getString("TABLE_CAT")),
                Optional.ofNullable(getTableSchemaName(resultSet)),
                resultSet.getString("TABLE_NAME"));
    }

    @Override // FIXME: Must be override if connector does not support SCHEMA
    protected String getTableSchemaName(ResultSet resultSet)
            throws SQLException
    {
        return Optional.ofNullable(resultSet.getString("TABLE_SCHEM")).orElse(MAIN_SCHEMA);
    }

```

In fact, SQLite manages two pseudo-schemas:
- main
- temp

These two schemas can be used in all SQL and/or DDL commands.

However, SQLite returns null as the schema name if the table is not a temporary table and furthermore, **Trino does not support a JDBC driver that does not support schema nor catalog**.
In any case, I've never managed to do it with a JDBC driver that doesn't support schema nor catalog (ie: returning a null schema and a null catalog). It would appear that **Trino requires that at least one of them not be null**.

findepi: > However, SQLite returns null as the schema name if the table is not a temporary table

that should still work

> and furthermore, **Trino does not support a JDBC driver that does not support schema nor catalog**.

`schemaName` is optional in `RemoteTableName` (and so is `catalogName`)
Can you please share a stacktrace of what doesn't work?

prrvchr: Stack trace:

```
java.util.NoSuchElementException: No value present
	at java.base/java.util.Optional.orElseThrow(Optional.java:377)
	at io.trino.plugin.jdbc.RemoteTableName.getSchemaTableName(RemoteTableName.java:64)
	at io.trino.plugin.sqlite.SqliteClient.getPrimaryKeys(SqliteClient.java:502)
	at io.trino.plugin.sqlite.SqliteClient.getTableProperties(SqliteClient.java:489)
	at io.trino.plugin.jdbc.ForwardingJdbcClient.getTableProperties(ForwardingJdbcClient.java:495)
	at io.trino.plugin.jdbc.jmx.StatisticsAwareJdbcClient.getTableProperties(StatisticsAwareJdbcClient.java:516)
	at io.trino.plugin.jdbc.RetryingJdbcClient.lambda$getTableProperties$0(RetryingJdbcClient.java:520)
	at dev.failsafe.Functions.lambda$toCtxSupplier$11(Functions.java:243)
```

Or

```
Caused by: java.lang.NullPointerException: Cannot invoke "String.toLowerCase(java.util.Locale)" because "remoteSchemaName" is null
	at io.trino.plugin.base.mapping.DefaultIdentifierMapping.fromRemoteSchemaName(DefaultIdentifierMapping.java:26)
	at io.trino.plugin.base.mapping.CachingIdentifierMapping.fromRemoteSchemaName(CachingIdentifierMapping.java:73)
	at io.trino.plugin.base.mapping.RuleBasedIdentifierMapping.fromRemoteSchemaName(RuleBasedIdentifierMapping.java:63)
	at io.trino.plugin.base.mapping.ForwardingIdentifierMapping.fromRemoteSchemaName(ForwardingIdentifierMapping.java:44)
	at io.trino.plugin.jdbc.BaseJdbcClient.getAllTableComments(BaseJdbcClient.java:217)
	at io.trino.plugin.jdbc.BaseJdbcClient.getTableNames(BaseJdbcClient.java:198)
	at io.trino.plugin.jdbc.ForwardingJdbcClient.getTableNames(ForwardingJdbcClient.java:88)
	at io.trino.plugin.jdbc.jmx.StatisticsAwareJdbcClient.lambda$getTableNames$0(StatisticsAwareJdbcClient.java:110)
	at io.trino.plugin.jdbc.jmx.JdbcApiStats.wrap(JdbcApiStats.java:34)
	at io.trino.plugin.jdbc.jmx.StatisticsAwareJdbcClient.getTableNames(StatisticsAwareJdbcClient.java:110)
	at io.trino.plugin.jdbc.RetryingJdbcClient.lambda$getTableNames$0(RetryingJdbcClient.java:85)
	at dev.failsafe.Functions.lambda$toCtxSupplier$11(Functions.java:243)
	at dev.failsafe.FailsafeExecutor.get(FailsafeExecutor.java:112)
	at io.trino.plugin.jdbc.RetryingModule.retry(RetryingModule.java:67)
	at io.trino.plugin.jdbc.RetryingJdbcClient.getTableNames(RetryingJdbcClient.java:85)
	at io.trino.plugin.jdbc.CachingJdbcClient.lambda$getTableNames$0(CachingJdbcClient.java:179)
	at io.trino.cache.EmptyCache.get(EmptyCache.java:95)
	... 67 more
```
I haven't done any further testing on this issue.
Maybe Trino doesn't support a null schema only? Because I have not seen any connector that does not provide information regarding the schema.

findepi: > ```
> 	at io.trino.plugin.jdbc.RemoteTableName.getSchemaTableName(RemoteTableName.java:64)
> 	at io.trino.plugin.sqlite.SqliteClient.getPrimaryKeys(SqliteClient.java:502)
> ```

It's indeed true that `RemoteTableName.getSchemaTableName` will fail when schema is not present.
I don't like this method, it's not a getter and it also moves from remote names to internal Trino names without any conversion.

However, is there anything preventing `SqliteClient.getPrimaryKeys` from _not_ calling this method?


For the second stacktrace, you may get away the lazy way by overriding `getTableRemoteSchemaName` only to return `"main"` always.

prrvchr: I agree that it is possible to get around the problem in other ways.

But I won't hide from you that I prefer my way in the sense that it allows the management of temporary tables just by specifying `temp` as schema during table creation and the regular tables are found under the `main` schema...

And last precision, if I remember correctly, concerning the tests it is a real hassle to try to adapt them so that it works without a schema... This is what made me abandon the Firebird connector. I'm waiting for the version 6 which manages schemas.

prrvchr:
Could we make a decision, as I really need this in the main branch because I'm having conflict issues between my active branches?
Thank you.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
chenjian2664: The method seems introduced by me :) - that's try to get the `SchemaTableName` instance from a `RemoteTableName` is just for simple purpose if that is allowed, yes, is it possible to avoid call the method directly if the remote and local schema isn't consistency

prrvchr: @chenjian2664 Is it a problem if it becomes non-static and protected (ie: overwritable)?

chenjian2664: @prrvchr I don't think it's a problem, but change like this - expose the method for overwriting for a not exist connector also feels redundant for me. I am +1 for  @findepi suggesting, that you could avoid calling this method in your own getPrimaryKeys method

prrvchr: This method is called 4 times in the BaseJdbcClient class
- [getTableHandle(ConnectorSession session, SchemaTableName schemaTableName) line 243](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L243)
- [getColumns(ConnectorSession session, SchemaTableName schemaTableName, RemoteTableName remoteTableName) line 324](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L324)
- [getAllTableColumns(ConnectorSession session, Optional<String> schema) line 392](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L392)
- [computeNext() line 444](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L444)

So either I have to overwrite all these methods, but what @findepi says is wrong if the ResultSet of the database metadata returns a null schema when **Trino cannot handle null schema with an API as it is**.

I tried using the [Firebird connector](https://github.com/trinodb/trino/pull/27673) which does not support schemas, I finally had to give up.

chenjian2664: We also considered the case where the remote table’s schema does not exist. when handling internal schema table names, we assumed the schema would exist - we have lots of place to build the SchemaTableName from a RemoteTableName previously (the getSchemaTableName is just for convience). In practice, all JDBC connectors we currently support do have schema support, so this assumption has held so far.
Regarding the interface change, it is neither complex nor particularly risky. But since no existing connector requires this change, it may not be an obvious or necessary improvement for the community, I'd like to hear more opinions on this change

findepi: > This method is called 4 times in the BaseJdbcClient class
> 
> * [getTableHandle(ConnectorSession session, SchemaTableName schemaTableName) line 243](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L243)
> * [getColumns(ConnectorSession session, SchemaTableName schemaTableName, RemoteTableName remoteTableName) line 324](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L324)
> * [getAllTableColumns(ConnectorSession session, Optional schema) line 392](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L392)
> * [computeNext() line 444](https://github.com/trinodb/trino/blob/master/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java#L444)
> 
> So either I have to overwrite all these methods,



wasn't that answered already in:

> It's indeed true that RemoteTableName.getSchemaTableName will fail when schema is not present.
I don't like this method, it's not a getter and it also moves from remote names to internal Trino names without any conversion.
> 
> However, is there anything preventing SqliteClient.getPrimaryKeys from not calling this method?

PR #27984, opened 2026-01-22
Fix internal error caused by dynamic filters on metadata columns

alexjo2144:
## Description

When a dynamic filter is calculated against one of the Iceberg metadata columns like $path or $file_modified_time the split source attempts to convert it to an Iceberg expression rather than apply it at the split source level. Instead we should remove it from the pushed down portion of the predicate and apply it as splits are coming in. 

## Release notes
This feels like a pretty rare bug, I'd call a release note for it optional. 

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
@@ -271,7 +271,7 @@ public class IcebergSplitSource
         if (fileScanIterable == null) {
             this.pushedDownDynamicFilterPredicate = dynamicFilter.getCurrentPredicate()
                     .transformKeys(IcebergColumnHandle.class::cast)
-                    .filter((columnHandle, domain) -> isConvertibleToIcebergExpression(domain));
+                    .filter((columnHandle, domain) -> isConvertibleToIcebergExpression(domain) && !isMetadataColumnId(columnHandle.getId()));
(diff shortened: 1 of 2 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -271,7 +271,7 @@ private synchronized ConnectorSplitBatch getNextBatchInternal(int maxSize)
         if (fileScanIterable == null) {
             this.pushedDownDynamicFilterPredicate = dynamicFilter.getCurrentPredicate()
                     .transformKeys(IcebergColumnHandle.class::cast)
-                    .filter((columnHandle, domain) -> isConvertibleToIcebergExpression(domain));
+                    .filter((columnHandle, domain) -> isConvertibleToIcebergExpression(domain) && !isMetadataColumnId(columnHandle.getId()));
```
raunaqmorarka, line 274: Would it work if we converted path domains to Iceberg expression on `file_path` ?

alexjo-dd: Doesn't look like it `ValidationException: Cannot find field 'file_path' in struct: struct<1: id: optional int, 2: value: optional string>`

chenjian2664:
Added "Fix query failure caused by dynamic filters contain metadata columns" to release note, FYI

PR #27985, opened 2026-01-22
Stop running CI with Databricks 12.2 LTS 

ebyhr:
## Description

It reaches EOL on March 1, 2026.
https://docs.databricks.com/aws/en/release-notes/runtime/#supported-databricks-runtime-lts-releases

## Release notes


testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeChangeDataFeedCompatibility.java
```diff
@@ -540,7 +539,7 @@ public void testMergeDeleteIntoTableWithCdfEnabled(String columnMappingMode)
         }
     }
 
-    @Test(groups = {DELTA_LAKE_DATABRICKS, DELTA_LAKE_DATABRICKS_133, DELTA_LAKE_DATABRICKS_143, DELTA_LAKE_DATABRICKS_154, DELTA_LAKE_DATABRICKS_164, DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
```
findepi, line 543: The PR removes 12.2, but the group 13.3 is being removed. Worth splitting commits?

ebyhr: I don't think so. DELTA_LAKE_DATABRICKS runs tests on 13.3 after this change.

findepi: it's confusing to me. i thought that DELTA_LAKE_DATABRICKS is the group for DBX tests that are run on all environments, but clearly it's not so
https://github.com/trinodb/trino/blob/b4b2d9a3d285efca71298acea8f181d775b964bd/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/suite/suites/SuiteDeltaLakeDatabricks154.java#L36

if DELTA_LAKE_DATABRICKS actually means "tests that we want to run on DBX 13.3 only", i would rather keep the DELTA_LAKE_DATABRICKS_133 name and remove DELTA_LAKE_DATABRICKS

btw maybe it's worth documenting the test groups `DELTA_LAKE_DATABRICKS_vX` -- whether it means "run on version X" or "run on version X and all newer versions".

ebyhr: DELTA_LAKE_DATABRICKS actually means oldest (13.3) and newest (17.3) LTS versions, not _a_ specific version.

findepi: worth code comment what the intention (pattern to follow) is?
or do we prefer this be inferred from usages?

ebyhr: Added the following comment to `TestGroups.DELTA_LAKE_DATABRICKS`:
```
// DELTA_LAKE_DATABRICKS runs tests against oldest and newest Databricks runtimes
```

PR #27992, opened 2026-01-23
Unmark updateCount as nullable in QueryResults

wendigo:
Since it was refactored to be an OptionalInt, it's required in the constructor to be non-null so it never actually is and annotation is no longer correct.

## Release notes

```markdown
## Section
```


client/trino-client/src/main/java/io/trino/client/QueryResults.java
```diff
@@ -157,7 +157,6 @@ public String getUpdateType()
         return updateType;
     }
 
-    @Nullable
```
findepi, line 160: i wish there was a static check for that

PR #27998, opened 2026-01-24
Add support for altering list and map nested types in Iceberg

JohnFitzpatrick44:
## Description

This PR adds support for ALTERing a list or map column in the Iceberg connector, along with a minor bugfix for related logic. 

I noticed that the `buildUpdateSchema` method in `IcebergMetadata` did not actually process each field once, as the deduplication logic was based on the `NestedField`s rather than the field names. This caused issues when implementing support for other nested types. The new logic should correctly handle each distinct field name once.

Additionally, some test cases I added for this support in `BaseConnectorTest` were not supported by every connector (i.e. MongoDB), so I updated those tests accordingly. Additionally, some regex was misformatted, which I corrected.

## Release notes

```markdown
## Iceberg
* Add support for changing nested types in `array` and `map` types. ({issue}`27998`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -251,7 +251,6 @@ import static com.google.common.base.Verify.verifyNotNull;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
-import static com.google.common.collect.Iterables.getLast;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static com.google.common.collect.Maps.transformValues;
 import static com.google.common.collect.Sets.difference;
@@ -2747,25 +2746,11 @@ public class IcebergMetadata
     @Override
     public void addField(ConnectorSession session, ConnectorTableHandle tableHandle, List<String> parentPath, String fieldName, io.trino.spi.type.Type type, boolean ignoreExisting)
     {
-        // Iceberg disallows ambiguous field names in a table. e.g. (a row(b int), "a.b" int)
-        String parentName = String.join(".", parentPath);
-
         Table icebergTable = catalog.loadTable(session, ((IcebergTableHandle) tableHandle).getSchemaTableName());
-        NestedField parent = icebergTable.schema().caseInsensitiveFindField(parentName);
-
+        NestedField parent = icebergTable.schema().caseInsensitiveFindField(String.join(".", parentPath));
         String caseSensitiveParentName = icebergTable.schema().findColumnName(parent.fieldId());
 
-        Types.StructType structType;
-        if (parent.type().isListType()) {
-            // list(struct...)
-            structType = parent.type().asListType().elementType().asStructType();
-        }
-        else {
-            // just struct
-            structType = parent.type().asStructType();
-        }
-
-        NestedField field = structType.caseInsensitiveField(fieldName);
+        NestedField field = icebergTable.schema().caseInsensitiveFindField(caseSensitiveParentName + "." + fieldName);
         if (field != null) {
             if (ignoreExisting) {
                 return;
@@ -2852,16 +2837,13 @@ public class IcebergMetadata
     public void renameField(ConnectorSession session, ConnectorTableHandle tableHandle, List<String> fieldPath, String target)
     {
         Table icebergTable = catalog.loadTable(session, ((IcebergTableHandle) tableHandle).getSchemaTableName());
-        String parentPath = String.join(".", fieldPath.subList(0, fieldPath.size() - 1));
-        NestedField parent = icebergTable.schema().caseInsensitiveFindField(parentPath);
+        NestedField field = icebergTable.schema().caseInsensitiveFindField(String.join(".", fieldPath));
+        verify(field != null, "Field %s not found", String.join(".", fieldPath));
 
-        String caseSensitiveParentName = icebergTable.schema().findColumnName(parent.fieldId());
-        NestedField source = parent.type().asStructType().caseInsensitiveField(getLast(fieldPath));
-
-        String sourcePath = caseSensitiveParentName + "." + source.name();
+        String caseSensitiveFieldName = icebergTable.schema().findColumnName(field.fieldId());
         try {
             icebergTable.updateSchema()
-                    .renameColumn(sourcePath, target)
+                    .renameColumn(caseSensitiveFieldName, target)
                     .commit();
         }
         catch (RuntimeException e) {


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9347,6 +9347,11 @@ protected Optional<SetColumnTypeSetup> filterSetColumnTypesDataProvider(SetColum
             return Optional.of(setup.withNewValueLiteral("TIMESTAMP '2020-02-12 14:03:00.123000 +00:00'"));
         }
         switch ("%s -> %s".formatted(setup.sourceColumnType(), setup.newColumnType())) {
+            // Iceberg allows updating column types if the update is safe. Safe updates are:
```
JohnFitzpatrick44, line 9350: Just reordered some comments to make more sense here. Let me know if this (or the regex fixes) deserve their own commit or not

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9376,7 +9377,8 @@ protected void verifySetColumnTypeFailurePermissible(Throwable e)
     {
         assertThat(e).hasMessageMatching(".*(Failed to set column type: Cannot change (column type:|type from .* to )" +
                 "|Time(stamp)? precision \\(3\\) not supported for Iceberg. Use \"time(stamp)?\\(6\\)\" instead" +
-                "|Type not supported for Iceberg: tinyint|smallint|char\\(20\\)).*");
```
JohnFitzpatrick44, line 9379: This regex could match messages we don't expect it to. Technically, any message that matches `.*smallint.*` would also match this, so I added parentheses to be more explicit.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2899,18 +2899,20 @@ private static void buildUpdateSchema(String name, Type sourceType, Type newType
         }
         if (sourceType instanceof StructType sourceRowType && newType instanceof StructType newRowType) {
             // Add, update or delete fields
-            List<NestedField> fields = Streams.concat(sourceRowType.fields().stream(), newRowType.fields().stream())
+            List<String> fieldNames = Streams.concat(sourceRowType.fields().stream(), newRowType.fields().stream())
+                    .filter(Objects::nonNull)
```
findinpath: When can we deal with a null field?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2899,18 +2899,20 @@ private static void buildUpdateSchema(String name, Type sourceType, Type newType
         }
         if (sourceType instanceof StructType sourceRowType && newType instanceof StructType newRowType) {
             // Add, update or delete fields
-            List<NestedField> fields = Streams.concat(sourceRowType.fields().stream(), newRowType.fields().stream())
+            List<String> fieldNames = Streams.concat(sourceRowType.fields().stream(), newRowType.fields().stream())
```
findinpath: Commit comment

> Fix schema update to run once per unique field


It is not immediately obvious what the "fix" is.
At a first glance this change looks more of a code refactoring.


Maybe add in the commit comment:

> Change deduplication logic from `NestedField` to the name of `NestedField`


I'm looking though at `Types.NestedField` implementation and see `hashCode()` and `equals()` methods implemented.
It would be worthwhile to add a test to showcase the effectiveness of the fix.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2897,28 +2898,45 @@ private static void buildUpdateSchema(String name, Type sourceType, Type newType
...
+            List<String> fieldNames = Streams.concat(sourceNestedType.fields().stream(), newNestedType.fields().stream())
                     .filter(Objects::nonNull)
                     .map(NestedField::name)
                     .distinct()
                     .collect(toImmutableList());
+
+            verify(!sourceNestedType.isListType() || (sourceFieldCount == 1 && newFieldCount == 1 && fieldNames.size() == 1),
+                    "Source list type %s and new list type %s must have exactly 1 element field with the same name".formatted(sourceNestedType, newNestedType));
+            verify(!sourceNestedType.isMapType() || (sourceFieldCount == 2 && newFieldCount == 2 && fieldNames.size() == 2),
+                    "Source map type %s and new map type %s must have exactly 2 key and value fields with the same names".formatted(sourceNestedType, newNestedType));
```
findinpath: Aren't these checks already covered by the initial `verify()` you've added?

```
(sourceNestedType.isMapType() && newNestedType.isMapType())
                    || (sourceNestedType.isListType() && newNestedType.isListType())
```

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseConnectorTest.java
```diff
@@ -290,6 +293,10 @@ protected Optional<SetColumnTypeSetup> filterSetFieldTypesDataProvider(SetColumn
             case "row(x integer, y integer) -> row(\"z\" integer, \"y\" integer, \"x\" integer)":
             case "row(x row(nested integer)) -> row(\"x\" row(\"nested\" bigint))":
             case "row(x row(a integer, b integer)) -> row(\"x\" row(\"b\" integer, \"a\" integer))":
+            case "array(array(integer)) -> array(array(bigint))":
```
findinpath: In the previous commit the following line was removed (aka it was supported)

```
case "array(array(integer)) -> array(array(bigint))":
```

Why is it being added in this commit?

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseConnectorTest.java
```diff
@@ -256,6 +257,8 @@ protected Optional<SetColumnTypeSetup> filterSetColumnTypesDataProvider(SetColum
             case "varchar -> char(20)":
             case "time(6) -> time(3)":
             case "timestamp(6) -> timestamp(3)":
+            // Iceberg cannot update map keys
```
findinpath, line 260: duplicated line - see 297

plugin/trino-mongodb/src/test/java/io/trino/plugin/mongodb/TestMongoConnectorTest.java
```diff
@@ -1929,6 +1929,10 @@ protected Optional<SetColumnTypeSetup> filterSetColumnTypesDataProvider(SetColum
             case "timestamp(6) -> timestamp(3)":
             case "timestamp(3) with time zone -> timestamp(6) with time zone":
             case "timestamp(6) with time zone -> timestamp(3) with time zone":
+            // MongoDB does not support updating map types
+            case "map(integer, varchar) -> map(bigint, varchar)":
```
findinpath, line 1932: @krvikash  FYI

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findinpath: Theoretically the `ListType` & `MapType` can be extended. Maybe it is worth either making them `final` in `apache/iceberg`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
JohnFitzpatrick44: I can remove this. There was a compiler warning for possible null object exceptions here, but fields should never be null.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
JohnFitzpatrick44: I wanted to explicitly confirm the expected fields here, but if we're fine assuming a certain structure to Map and List fields, these verifications can be removed. I wanted to be cautious around changing `StructType` to `NestedType` type checks on L2901 and throw a clearer error message if we see something unexpected.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseConnectorTest.java
JohnFitzpatrick44: The previous commit removes `case "array(integer) -> array(bigint)":` from `filterSetColumnTypesDataProvider`, which is now supported.

This change removes support from the _nested_ array update from `filterSetFieldTypesDataProvider`, as setting fields does not support any nested types. You can see `case "array(integer) -> array(bigint)":` on L286 for this method.

plugin/trino-lakehouse/src/test/java/io/trino/plugin/lakehouse/TestLakehouseConnectorTest.java
JohnFitzpatrick44: There're map cases in `filterSetFieldTypesDataProvider`, but that's a different method for a different test

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
JohnFitzpatrick44: I could tighten up the message here, but the `equals` implementation is not suitable to find distinct subfields. For one, we know that sourceType and newType are not equal, so there will be a subfield that is different in some way. Additionally, ordering fields will change their IDs, which will cause the equality check to fail.

To demonstrate both of these cases, if you run the following test cases of `testSetColumnTypes`:
```
row(x integer) -> row(x bigint)
row(x integer, y integer) -> row(y integer, x integer)
```

You'll see `fields` will contain two versions of `x` in both cases, due to the change in type for the first case, and due to the change in id for the second. In both cases, this leads to duplicate calls to `buildUpdateSchema` with the exact same arguments, as those calls are only dependent on the field name.

I didn't add a unit test as this behavior doesn't actually lead to a failure state, and the tests we'd need to add would just be unit tests verifying the number of times `buildUpdateSchema` is called.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2932,7 +2952,7 @@ else if (fieldExists(newRowType, field.name())) {
         throw new IllegalArgumentException("Cannot change type from %s to %s".formatted(sourceType, newType));
     }
 
-    private static boolean fieldExists(StructType structType, String fieldName)
+    private static boolean fieldExists(NestedType structType, String fieldName)
```
raphaelsolarski: rename: `structType`->`nestedType`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2897,26 +2898,45 @@ private static void buildUpdateSchema(String name, Type sourceType, Type newType
             schemaUpdate.updateColumn(name, newType.asPrimitiveType());
             return;
         }
-        if (sourceType instanceof StructType sourceRowType && newType instanceof StructType newRowType) {
-            // Add, update or delete fields
-            List<NestedField> fields = Streams.concat(sourceRowType.fields().stream(), newRowType.fields().stream())
+        if (sourceType instanceof NestedType sourceNestedType && newType instanceof NestedType newNestedType) {
```
raphaelsolarski: I wonder if we could:
1. Extract `MapType` and `ListType` to separate simpler cases (most of code in this `if` body doesn't relate to them + those verification maybe could be skipped due to type safety)
2. Get rid of `return` statements in the `buildUpdateSchema` method
3. Extract remaining code that relate to `StructType` to a separate method - let's say `buildStructTypeUpdateSchema`.

We possibly could end up with something similar to:
```java
private static void buildUpdateSchema(String name, Type sourceType, Type newType, UpdateSchema schemaUpdate)
{
    if (sourceType.equals(newType)) {
        // no update needed
    }
    else if (sourceType.isPrimitiveType() && newType.isPrimitiveType()) {
        schemaUpdate.updateColumn(name, newType.asPrimitiveType());
    }
    else if (sourceType instanceof ListType sourceListType && newType instanceof ListType newListType) {
        buildUpdateSchema(name + ".element", sourceListType.elementType(), newListType.elementType(), schemaUpdate);
    }
    else if (sourceType instanceof org.apache.iceberg.types.Types.MapType sourceMapType
            && newType instanceof org.apache.iceberg.types.Types.MapType newMapType) {
        buildUpdateSchema(name + ".key", sourceMapType.keyType(), newMapType.keyType(), schemaUpdate);
        buildUpdateSchema(name + ".value", sourceMapType.valueType(), newMapType.valueType(), schemaUpdate);
    }
    else if (sourceType instanceof StructType sourceNestedType && newType instanceof StructType newNestedType) {
        buildStructTypeUpdateSchema(name, schemaUpdate, sourceNestedType, newNestedType);
    }
    else {
        throw new IllegalArgumentException("Cannot change type from %s to %s".formatted(sourceType, newType));
    }
}

private static void buildStructTypeUpdateSchema(String name, UpdateSchema schemaUpdate, StructType sourceNestedType, StructType newNestedType)
{
    List<String> fieldNames = Streams.concat(sourceNestedType.fields().stream(), newNestedType.fields().stream())
            .filter(Objects::nonNull)
            .map(NestedField::name)
            .distinct()
            .collect(toImmutableList());

    // Add, update or delete fields
    for (String fieldName : fieldNames) {
        if (fieldExists(sourceNestedType, fieldName) && fieldExists(newNestedType, fieldName)) {
            buildUpdateSchema(name + "." + fieldName, sourceNestedType.fieldType(fieldName), newNestedType.fieldType(fieldName), schemaUpdate);
        }
        else if (fieldExists(newNestedType, fieldName)) {
            schemaUpdate.addColumn(name, fieldName, newNestedType.fieldType(fieldName));
        }
        else {
            schemaUpdate.deleteColumn(name + "." + fieldName);
        }
    }

    // Order fields based on the new column type
    String currentName = null;
    for (NestedField field : newNestedType.fields()) {
        String path = name + "." + field.name();
        if (currentName == null) {
            schemaUpdate.moveFirst(path);
        }
        else {
            schemaUpdate.moveAfter(path, currentName);
        }
        currentName = path;
    }
}
```

@JohnFitzpatrick44 wdyt?

JohnFitzpatrick44: This idea is similar to a first pass I took, but it seemed unnecessary to split out handling for maps and lists if the existing logic for structs generally applies to all nested types. I did this with the verification steps that @findinpath suggests [we might remove](https://github.com/trinodb/trino/pull/27998#discussion_r2726894627), and if we do remove those checks, your approach would be cleaner. I would probably replace `buildStructTypeUpdateSchema` with something like `sortFields` that we do as a separate step, though.

Just calling out, if we take this code at face value, we're assuming:
- Maps have exactly 2 fields named "key" and "value"
- Lists have exactly 1 field called "element"

If a type does not follow those rules, we might get some unintuitive errors, or even add a field instead of updating it. I haven't tested if that's possible though. I wanted this section of code to avoid those kinds of assumptions.

JohnFitzpatrick44: I'll implement a version of this with some checks still there as a compromise

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -198,9 +198,11 @@
 import org.apache.iceberg.expressions.Term;
 import org.apache.iceberg.io.CloseableIterable;
```
findinpath, line 199: `Support altering nested list and map types in iceberg` commit message

Let's change it to 

`Support altering nested list and map value types in iceberg`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9347,6 +9347,11 @@ protected Optional<SetColumnTypeSetup> filterSetColumnTypesDataProvider(SetColum
             return Optional.of(setup.withNewValueLiteral("TIMESTAMP '2020-02-12 14:03:00.123000 +00:00'"));
```
findinpath, line 9347: It would be worthwhile to add this as the 2nd commit and the commit "Support altering nested list and map types in iceberg" to be the last commit.
You are adding here the map type test cases that are relevant for the 3rd commit.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9356,16 +9361,12 @@ protected Optional<SetColumnTypeSetup> filterSetColumnTypesDataProvider(SetColum
             case "varchar -> char(20)":
             case "time(6) -> time(3)":
             case "timestamp(6) -> timestamp(3)":
-            case "array(integer) -> array(bigint)":
-                // Iceberg allows updating column types if the update is safe. Safe updates are:
-                // - int to bigint
-                // - float to double
-                // - decimal(P,S) to decimal(P2,S) when P2 > P (scale cannot change)
-                // https://iceberg.apache.org/docs/latest/spark-ddl/#alter-table--alter-column
+            // Iceberg cannot update map keys
```
findinpath, line 9364: Would it make sense to explore in this PR also this TODO 

https://github.com/trinodb/trino/blob/d3756e646b7f15e1749a1336b4a82b889addda8a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java#L2966

?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
JohnFitzpatrick44: I ordered it in this way because, with this proposed ordering, we'd need to add cases to mark maps as unsupported in the second commit, and immediately remove them in the third. But I can make that change if need be.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
JohnFitzpatrick44: I can take a look, shouldn't be too different than what we have already

plugin/trino-mongodb/src/test/java/io/trino/plugin/mongodb/TestMongoConnectorTest.java
ebyhr: We don't need this code comment. All cases in this method indicate that the connector does not support updating these types.

Also, "MongoDB does not support" is misleading. This is a limitation of the connector, not MongoDB database.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
JohnFitzpatrick44: Implementing this expands the impact radius of this PR a good deal, due to some parquet-specific issues I've been running into. For now, keeping this focused on column setting, and I can create a follow up PR to address field setting.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2897,44 +2899,61 @@ private static void buildUpdateSchema(String name, Type sourceType, Type newType
...
-                    buildUpdateSchema(name + "." + field.name(), sourceRowType.fieldType(field.name()), newRowType.fieldType(field.name()), schemaUpdate);
-                }
-                else if (fieldExists(newRowType, field.name())) {
-                    schemaUpdate.addColumn(name, field.name(), field.type());
-                }
-                else {
-                    schemaUpdate.deleteColumn(name + "." + field.name());
-                }
+
+        switch (sourceType) {
```
raphaelsolarski, line 2885: awesome. I'll give you `Scala Champion` badge for this pattern matching 😄

JohnFitzpatrick44:
@findinpath @raphaelsolarski Added a new commit to support nested fields in field alteration methods.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2747,25 +2746,11 @@ public void dropDefaultValue(ConnectorSession session, ConnectorTableHandle tabl
...
-            // list(struct...)
-            structType = parent.type().asListType().elementType().asStructType();
-        }
-        else {
-            // just struct
-            structType = parent.type().asStructType();
-        }
-
-        NestedField field = structType.caseInsensitiveField(fieldName);
+        NestedField field = icebergTable.schema().caseInsensitiveFindField(caseSensitiveParentName + "." + fieldName);
```
ebyhr, line 2753: > Support altering non-primitive types for fields in iceberg

How does this change in `addField` method relate to altering types? Same for `renameField` method.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9357,11 +9362,11 @@ protected Optional<SetColumnTypeSetup> filterSetColumnTypesDataProvider(SetColum
...
-                // - decimal(P,S) to decimal(P2,S) when P2 > P (scale cannot change)
-                // https://iceberg.apache.org/docs/latest/spark-ddl/#alter-table--alter-column
+            case "array(array(integer)) -> array(array(bigint))":
+            case "map(varchar, integer) -> map(varchar, bigint)":
+            case "map(integer, row(x integer)) -> map(integer, row(\"x\" bigint))":
+            // Iceberg cannot update map keys
+            case "map(integer, varchar) -> map(bigint, varchar)":
                 return Optional.of(setup.asUnsupported());
 
             case "varchar(100) -> varchar(50)":
```
ebyhr, line 9369: Readers may wonder why the above cases have comments above `case`, while this case has a comment below it.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
ebyhr: I recommend separating a commit.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
ebyhr: I recommend separating a commit.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
JohnFitzpatrick44: Overall, I was trying to make this code more flexible, as this (and the following methods) assume the parent to be either a list with a struct sub-type or a struct. If it were any different configuration, we'd get an error when trying to cast the parent type as a struct, which is unnecessarily restrictive and less clear than the other error we'd receive if it was invalid to add the field (e.g., when trying to add a field to a map).

- [These lines](https://github.com/trinodb/trino/pull/27998/changes/1e24f22c38952aa0646c862c80ff0b39a7c27bd2#diff-e1cb17efec6787989f9df9ee40c4f2809ff3fe946cd2ec721ff8932b131997b8L2758-L2766) were removed, as they assume the structure of the parent.
- Without an assumed struct type, we can't try getting the field from the parent using its name. 
- We can either cast the parent type as a nested type or find the field using `caseInsensitiveFindField`. I just opted for the latter

Happy to partially revert these changes and just add some explicit checks that the parent is a struct, as I think it's important to return a reasonable error message, and minimize the other changes here.

JohnFitzpatrick44: Moved to different commit

PR #27999, opened 2026-01-25
Support `DESCRIBE OUTPUT (subquery)` syntax

Max-Cheng:
## Description
Add support for `DESCRIBE OUTPUT` with inline queries, allowing users to describe query output columns without creating a prepared statement first.

Previously, users had to use a two-step process:
```sql
PREPARE my_query FROM SELECT * FROM nation;
DESCRIBE OUTPUT my_query;
```
Now they can use the more concise syntax:
```sql
DESCRIBE OUTPUT (SELECT * FROM nation);
```
This enhancement improves usability by:
- Reducing verbosity for ad-hoc query analysis
- Avoiding the need to manage prepared statement names
- Working around HTTP header length limits in some cloud environments where long prepared statement SQL can exceed limits

The implementation adds:
- New DescribeOutputWithQuery AST node to represent inline query descriptions
- Grammar changes in SqlBase.g4 to support the DESCRIBE OUTPUT (query) syntax
- DescribeOutputWithQueryRewrite to handle the rewriting logic
- Parser tests and documentation updates

The feature follows the same execution path as prepared statement descriptions internally, ensuring consistent behavior and column metadata output.

## Release notes
```markdown
## General
* Add support for `DESCRIBE OUTPUT` with inline queries, allowing direct description of query results without requiring a `PREPARE` statement. For example: `DESCRIBE OUTPUT (SELECT * FROM nation)`.
```

- fixes https://github.com/trinodb/trino/issues/28002

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/AbstractTestEngineOnlyQueries.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/AbstractTestEngineOnlyQueries.java
@@ -1442,50 +1442,41 @@ public abstract class AbstractTestEngineOnlyQueries
     @Test
     public void testDescribeOutput()
     {
-        Session session = Session.builder(getSession())
-                .addPreparedStatement("my_query", "SELECT * FROM nation")
-                .build();
-
-        MaterializedResult actual = computeActual(session, "DESCRIBE OUTPUT my_query");
+        Session session = getSession();
+        String sql = "SELECT * FROM nation";
         MaterializedResult expected = resultBuilder(session, VARCHAR, VARCHAR, VARCHAR, VARCHAR, VARCHAR, BIGINT, BOOLEAN)
                 .row("nationkey", session.getCatalog().get(), session.getSchema().get(), "nation", "bigint", 8, false)
                 .row("name", session.getCatalog().get(), session.getSchema().get(), "nation", "varchar(25)", 0, false)
                 .row("regionkey", session.getCatalog().get(), session.getSchema().get(), "nation", "bigint", 8, false)
                 .row("comment", session.getCatalog().get(), session.getSchema().get(), "nation", "varchar(152)", 0, false)
                 .build();
-        assertEqualsIgnoreOrder(actual, expected);
+        assertDescribeOutputWithBothSyntax(session, sql, expected);
     }
 
     @Test
     public void testDescribeOutputDateTimeTypes()
     {
-        Session session = Session.builder(getSession())
-                .addPreparedStatement("my_query", "SELECT localtimestamp a, current_timestamp b, localtime c")
-                .build();
-
-        MaterializedResult actual = computeActual(session, "DESCRIBE OUTPUT my_query");
+        Session session = getSession();
+        String sql = "SELECT localtimestamp a, current_timestamp b, localtime c";
         MaterializedResult expected = resultBuilder(session, VARCHAR, VARCHAR, VARCHAR, VARCHAR, VARCHAR, BIGINT, BOOLEAN)
                 .row("a", "", "", "", "timestamp(3)", 8, true)
                 .row("b", "", "", "", "timestamp(3) with time zone", 8, true)
                 .row("c", "", "", "", "time(3)", 8, true)
                 .build();
-        assertEqualsIgnoreOrder(actual, expected);
+        assertDescribeOutputWithBothSyntax(session, sql, expected);
     }
 
     @Test
     public void testDescribeOutputNamedAndUnnamed()
     {
-        Session session = Session.builder(getSession())
-                .addPreparedStatement("my_query", "SELECT 1, name, regionkey AS my_alias FROM nation")
-                .build();
-
-        MaterializedResult actual = computeActual(session, "DESCRIBE OUTPUT my_query");
+        Session session = getSession();
+        String sql = "SELECT 1, name, regionkey AS my_alias FROM nation";
         MaterializedResult expected = resultBuilder(session, VARCHAR, VARCHAR, VARCHAR, VARCHAR, VARCHAR, BIGINT, BOOLEAN)
                 .row("_col0", "", "", "", "integer", 4, false)
                 .row("name", session.getCatalog().get(), session.getSchema().get(), "nation", "varchar(25)", 0, false)
                 .row("my_alias", session.getCatalog().get(), session.getSchema().get(), "nation", "bigint", 8, true)
                 .build();
-        assertEqualsIgnoreOrder(actual, expected);
+        assertDescribeOutputWithBothSyntax(session, sql, expected);
     }
 
     @Test
@@ -1564,16 +1555,27 @@ public abstract class AbstractTestEngineOnlyQueries
     @Test
     public void testDescribeOutputOnAliasedColumnsAndExpressions()
     {
-        Session session = Session.builder(getSession())
-                .addPreparedStatement("my_query", "SELECT count(*) AS this_is_aliased, 1 + 2 FROM nation")
-                .build();
-
-        MaterializedResult actual = computeActual(session, "DESCRIBE OUTPUT my_query");
+        Session session = getSession();
+        String sql = "SELECT count(*) AS this_is_aliased, 1 + 2 FROM nation";
         MaterializedResult expected = resultBuilder(session, VARCHAR, VARCHAR, VARCHAR, VARCHAR, VARCHAR, BIGINT, BOOLEAN)
                 .row("this_is_aliased", "", "", "", "bigint", 8, true)
                 .row("_col1", "", "", "", "integer", 4, false)
                 .build();
-        assertEqualsIgnoreOrder(actual, expected);
+        assertDescribeOutputWithBothSyntax(session, sql, expected);
+    }
+
+    private void assertDescribeOutputWithBothSyntax(Session session, @Language("SQL") String sql, MaterializedResult expected)
+    {
+        // Test prepared statement syntax
+        Session sessionWithStatement = Session.builder(session)
+                .addPreparedStatement("my_query", sql)
+                .build();
+        MaterializedResult preparedResult = computeActual(sessionWithStatement, "DESCRIBE OUTPUT my_query");
+        assertEqualsIgnoreOrder(preparedResult, expected);
+
+        // Test inline query syntax
+        MaterializedResult inlineResult = computeActual(session, format("DESCRIBE OUTPUT (%s)", sql));
+        assertEqualsIgnoreOrder(inlineResult, expected);
     }
 
     @Test


Max-Cheng:
Link https://github.com/trinodb/trino/issues/28002

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -204,6 +204,7 @@ statement
     | EXECUTE IMMEDIATE string (USING expression (',' expression)*)?   #executeImmediate
     | DESCRIBE INPUT identifier                                        #describeInput
     | DESCRIBE OUTPUT identifier                                       #describeOutput
+    | DESCRIBE OUTPUT '(' query ')'                                    #describeOutputWithQuery
```
findepi: cc @martint

Praveen2112: Should we treat it as `DESCRIBE (query)` - like we handle for `SHOW STATS FOR ..` - which would work for table and query

Max-Cheng: @Praveen2112 Thanks for the suggestion! I did consider the more general `DESCRIBE (query)` syntax similar to `SHOW STATS FOR`, but I ended up going with `DESCRIBE OUTPUT (query)` for a few reasons.

The main concern is that `DESCRIBE table_name` and `DESCRIBE OUTPUT` serve different purposes - one describes table structure while the other describes query results. Mixing these semantics under a single `DESCRIBE` keyword could create ambiguity and confuse users about what they're actually describing.

Also, `DESCRIBE OUTPUT (query)` keeps the same semantic consistency with the existing `DESCRIBE OUTPUT prepared_statement`, so users already familiar with that syntax will find this natural.

That said, if the community feels strongly about having a unified `DESCRIBE` syntax, I'm definitely open to exploring that approach. I just think the current design is more intuitive and avoids potential parsing ambiguities. What do you think?

docs/src/main/sphinx/sql/describe-output.md
```diff
@@ -71,7 +71,19 @@ DESCRIBE OUTPUT my_create;
...
 ```
-
+We also support describe a query whose output columns are expressions:
+```sql
+DESCRIBE OUTPUT (SELECT * FROM nation);
+```
+```text
+ Column Name | Catalog | Schema | Table  |     Type     | Type Size | Aliased
+-------------+---------+--------+--------+--------------+-----------+---------
+ n_nationkey | tpch    | sf1    | nation | bigint       |         8 | false
```
chenjian2664, line 81: Let's update the demo to show the aliased column here

core/trino-main/src/main/java/io/trino/sql/rewrite/DescribeOutputWithQueryRewrite.java
```diff
@@ -0,0 +1,190 @@
...
+        public Visitor(
+                Session session,
+                AnalyzerFactory analyzerFactory,
+                List<Expression> parameters,
+                Map<NodeRef<Parameter>, Expression> parameterLookup,
+                WarningCollector warningCollector,
+                PlanOptimizersStatsCollector planOptimizersStatsCollector)
+        {
+            this.session = requireNonNull(session, "session is null");
+            this.analyzerFactory = analyzerFactory;
```
chenjian2664: rnn

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
martint: This should use `rootQuery`, not `query`. Otherwise, it will miss queries with inline functions.

core/trino-parser/src/main/java/io/trino/sql/tree/DescribeOutputWithQuery.java
```diff
@@ -0,0 +1,77 @@
...
+package io.trino.sql.tree;
+
+import com.google.common.collect.ImmutableList;
+
+import java.util.List;
+import java.util.Objects;
+
+import static com.google.common.base.MoreObjects.toStringHelper;
+
+public class DescribeOutputWithQuery
```
martint: This should not be modeled as a separate kind of statement. The existing DescribeOutput should be extended to allow either a query or an identifier as an argument.

martint (changes requested):
The proposed syntax is reasonable. It's not covered by the SQL specification, but it aligns with the semantics of what DESCRIBE OUTPUT does.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -204,6 +204,7 @@ statement
     | EXECUTE IMMEDIATE string (USING expression (',' expression)*)?   #executeImmediate
     | DESCRIBE INPUT identifier                                        #describeInput
     | DESCRIBE OUTPUT identifier                                       #describeOutput
+    | DESCRIBE OUTPUT '(' rootQuery ')'                                #describeOutputWithQuery
```
martint: Rename the rule to `describeQueryOutput`

core/trino-main/src/main/java/io/trino/sql/rewrite/DescribeOutputRewrite.java
```diff
@@ -115,7 +116,7 @@ public Visitor(
         {
             this.session = requireNonNull(session, "session is null");
             this.parser = requireNonNull(parser, "parser is null");
-            this.analyzerFactory = analyzerFactory;
+            this.analyzerFactory = requireNonNull(analyzerFactory, "analyzerFactory is null");
```
martint, line 121: Unrelated change. Move it to a separate commit.

core/trino-parser/src/main/java/io/trino/sql/tree/DescribeOutput.java
```diff
@@ -17,25 +17,41 @@
...
 
 import static com.google.common.base.MoreObjects.toStringHelper;
+import static java.util.Objects.requireNonNull;
 
 public class DescribeOutput
         extends Statement
 {
-    private final Identifier name;
+    private final Optional<Identifier> name;
+    private final Optional<Query> query;
```
martint: It's better than before, but still not ideal. It makes the combinations where both are might present or both missing ambiguous.

A better approach is to introduce a sealed interface (e.g., `Target`) with two implementations and have a single field of that type.

```java
sealed interface Target
    permits Target.IdentifierTarget, Target.QueryTarget
{
    record IdentifierTarget(Identifier identifier)
            implements Target
    {
    }

    record QueryTarget(Query query) 
            implements Target
    {
    }
}
```

This approach makes it possible to use Java's enhanced switch with pattern matching wherever necessary. E.g, `SqlFormatter.visitDescribeOutput()` could be written as:

```java
...
builder.append(switch (node.getTarget()) {
    case IdentifierTarget(Identifier identifier) -> formatName(identifier);
    case QueryTarget(Query query) -> '(' + formatSql(query) + ')';
});
```

core/trino-parser/src/main/java/io/trino/sql/tree/DescribeOutput.java
```diff
@@ -45,13 +61,16 @@ public <R, C> R accept(AstVisitor<R, C> visitor, C context)
     @Override
     public List<Node> getChildren()
     {
+        if (query.isPresent()) {
+            return ImmutableList.of(query.get());
+        }
         return ImmutableList.of();
```
martint: This is a pre-existing issue, but `getChildren()` should return `name` in this case.

Max-Cheng:
@martint @Praveen2112 PTAL

core/trino-main/src/main/java/io/trino/sql/rewrite/DescribeOutputRewrite.java
```diff
@@ -139,6 +139,17 @@ protected Node visitDescribeOutput(DescribeOutput node, Void context)
             return createDescribeOutputQuery(rows, limit);
         }
 
+        private Statement getStatement(DescribeOutput node)
+        {
+            return switch (node.getTarget()) {
+                case DescribeOutput.Target.PreparedStatement(Identifier name) -> {
+                    String sqlString = session.getPreparedStatement(name.getValue());
```
martint: Inline `sqlString`, which will allow simplifying the structure of this block.

core/trino-main/src/main/java/io/trino/sql/rewrite/DescribeOutputRewrite.java
```diff
@@ -139,6 +139,14 @@ protected Node visitDescribeOutput(DescribeOutput node, Void context)
             return createDescribeOutputQuery(rows, limit);
         }
 
+        private Statement getStatement(DescribeOutput node)
+        {
+            return switch (node.getTarget()) {
+                case DescribeOutput.Target.PreparedStatement(Identifier name) -> parser.createStatement(session.getPreparedStatement(name.getValue()));
```
Praveen2112: nit: Can we use import for PreparedStatement and InlineQuery ?

Praveen2112: import io.trino.sql.tree.DescribeOutput.Target.PreparedStatement;
import io.trino.sql.tree.DescribeOutput.Target.InlineQuery;

core/trino-parser/src/test/java/io/trino/sql/parser/TestSqlParser.java
```diff
@@ -5466,6 +5466,18 @@ public void testDescribeOutput()
                 .isEqualTo(new DescribeOutput(location(1, 1), new Identifier(location(1, 17), "myquery", false)));
     }
 
+    @Test
+    public void testDescribeOutputWithQuery()
```
Praveen2112, line 5524: Can we add additional test cases in AbstractTestEngineOnlyQueries ?

core/trino-main/src/main/java/io/trino/sql/rewrite/DescribeOutputRewrite.java
chenjian2664: do you mean static import

core/trino-parser/src/test/java/io/trino/sql/parser/TestSqlParser.java
Max-Cheng: Sure.

core/trino-main/src/main/java/io/trino/sql/rewrite/DescribeOutputRewrite.java
Max-Cheng: > do you mean static import

I think the import should be shorter :D

chenjian2664:
@martint If the syntax looks good to you, could you clear the "requested changes"?

chenjian2664:
@Praveen2112 The comments are addressed, PTAL

Issue #27598, opened 2025-12-10
Materialized View documentation for Iceberg


PPL-TBSK:
In my opinion Trino Iceberg Materialized View documentation misses following information:

How you can get the exact name of the underlying storage table.
How you can optimize (optimize, expire snapshots, remove orphan files) the data for materialized view, therefore that you should run optimize on top of the underlying storage table.

Would it be good idea to include something like following statement to Iceberg Materialized View documentation?:

To expire snapshots, optimize data, remove orphan data regarding materialized views, you should run maintenance jobs on top of underlying storage table, which you can find using following query:
```
SELECT *
FROM system.metadata.materialized_views
WHERE catalog_name = 'iceberg'
;
```

Is my statement correct? and if yes, could it go into docs?

findepi:
> How you can get the exact name of the underlying storage table.

we considered storage table existence, and the fact that's a single table, to be implementation detail
we do not want users to access that table directly

> How you can optimize (optimize, expire snapshots, remove orphan files) the data for materialized view, therefore that you should run optimize on top of the underlying storage table.

I agree these operations should be supported on a materialized view.

cc @raunaqmorarka

PPL-TBSK:
@raunaqmorarka

raunaqmorarka:
I think REFRESH should be extended to perform cleanup of expired snapshots automatically.
We could consider allowing OPTIMZE to be run on MVs for compaction of underlying storage table.

raunaqmorarka:
> we considered storage table existence, and the fact that's a single table, to be implementation detail
we do not want users to access that table directly

> I agree these operations should be supported on a materialized view.

I find those two views to be in conflict.
If the storage table is just an internal detail, then the requirement to run all the **_table_** maintenance operations on it is leaking that implementation detail. The operations of compaction, snapshot expiry, removal of orphan files make sense only in the context of an iceberg table. If you want to expose all those concepts (and more like optimize_manifests), then you might as well expose the storage table explicitly and let the user operate on that directly.

JohnFitzpatrick44:
> I think REFRESH should be extended to perform cleanup of expired snapshots automatically.

Created https://github.com/trinodb/trino/pull/28008 to address this

PPL-TBSK:
Closing based on @JohnFitzpatrick44  comment

PR #25343, opened 2025-03-18
Cleanup previous snapshot files during materialized view refresh

mayankvadariya:
## Description

`REFRESH MATERIALIZED VIEW` https://trino.io/docs/current/sql/refresh-materialized-view.html doesn't cleanup old snapshot files which causes materialized view size to grow without having any benefit of retaining old data.
Proposed solution uses `org.apache.iceberg.ExpireSnapshots` procedure in `io.trino.plugin.iceberg.IcebergMetadata#finishRefreshMaterializedView` to cleanup previous snapshot created just before the execution of refresh materialized view command.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
## Iceberg
* Clean up old snapshots of the storage table on refresh of materialized views. ({issue}`25343`)
```

PR #28008, opened 2026-01-26
Remove expired materialized view iceberg snapshots on refresh

JohnFitzpatrick44:
## Description

Removes expired snapshots of MVs on refresh, up to a cap. Previously, this feature was implemented without a cap, but was reverted due to performance implications. The cap should mitigate those, and can be configured.

The two config properties that were added are:
- `iceberg.materialized-views.refresh-max-snapshots-to-expire`, defining the cap
- `iceberg.materialized-views.refresh-snapshot-retention-period`, defining a retention period to avoid removing snapshots used in ongoing queries

A new test was added to verify behavior. As this behavior depends on config properties, I used a pre-made MV to build a backlog of snapshots that should be removed to verify that the cap applies properly.

- First implementation: https://github.com/trinodb/trino/pull/25343
- Revert: https://github.com/trinodb/trino/pull/26051
- Related issue: https://github.com/trinodb/trino/issues/27598

## Release notes

```markdown
## Iceberg
* Clean up unused files from materialized views when they are refreshed. ({issue}`28008`)
```


docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -184,6 +184,20 @@ implementation is used:
...
     the new records.
   - `true`
+* - `iceberg.materialized-views.refresh-max-snapshots-to-expire`
+  - Maximum number of materialized view snapshots to expire when performing a
+    refresh. A value of `0` will result in no snapshots being removed. Higher
+    values will impact refresh performance. Any snapshots within the duration of
+    the `iceberg.materialized-views.refresh-snapshot-retention-period`
+    configuration property will be retained. Equivalent session property is
+    `materialized_view_refresh_max_expired_snapshots`.
+  - `0`
```
raunaqmorarka: 200

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -184,6 +184,20 @@ implementation is used:
     creation of more data files, since it uses the append operation to insert
     the new records.
   - `true`
+* - `iceberg.materialized-views.refresh-max-snapshots-to-expire`
+  - Maximum number of materialized view snapshots to expire when performing a
+    refresh. A value of `0` will result in no snapshots being removed. Higher
+    values will impact refresh performance. Any snapshots within the duration of
+    the `iceberg.materialized-views.refresh-snapshot-retention-period`
+    configuration property will be retained. Equivalent session property is
+    `materialized_view_refresh_max_expired_snapshots`.
```
raunaqmorarka: Drop the last line

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -184,6 +184,20 @@ implementation is used:
     creation of more data files, since it uses the append operation to insert
     the new records.
   - `true`
+* - `iceberg.materialized-views.refresh-max-snapshots-to-expire`
+  - Maximum number of materialized view snapshots to expire when performing a
+    refresh. A value of `0` will result in no snapshots being removed. Higher
+    values will impact refresh performance. Any snapshots within the duration of
```
raunaqmorarka: `Higher values will impact refresh performance` -> 
```
Higher values may increase the time taken to execute materialized view refreshes when there are a large number of snapshots in the materialized view.
```

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -184,6 +184,20 @@ implementation is used:
...
+    the `iceberg.materialized-views.refresh-snapshot-retention-period`
+    configuration property will be retained. Equivalent session property is
+    `materialized_view_refresh_max_expired_snapshots`.
+  - `0`
+* - `iceberg.materialized-views.refresh-snapshot-retention-period`
+  - The duration in which materialized view snapshots will be retained. During a
```
raunaqmorarka: `The duration in which` -> `The duration for which`

raunaqmorarka: 2nd line:
```
Any snapshots older than this value will be removed during a materialized view refresh, up to a cap ...
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
```diff
@@ -0,0 +1,234 @@
...
+            files.add(fileIterator.next());
+        }
+
+        return files.build().collect(Collectors.groupingBy(file ->
+                switch (file.location().toString()) {
+                    case String path when path.contains("/data/") -> FileType.DATA;
+                    case String path when path.contains("/metadata/") && path.endsWith(".json") -> FileType.METADATA;
+                    case String path when path.contains("/metadata/") && path.contains("/snap-") && path.endsWith(".avro") -> FileType.SNAPSHOT;
+                    case String path when path.contains("/metadata/") && !path.contains("/snap-") && path.endsWith(".avro") -> FileType.MANIFEST;
+                    default -> FileType.UNKNOWN;
```
raunaqmorarka: We can throw an exception here instead of having `UNKNOWN` enum

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -4020,10 +4040,28 @@ public Optional<ConnectorOutputMetadata> finishRefreshMaterializedView(
...
+
+        if (materializedViewRefreshMaxSnapshotsToExpire == 0) {
+            return icebergCommitMetadata;
         }
-        return Optional.of(new IcebergCommitMetadata(summary));
+
+        int snapshots = size(icebergTable.snapshots());
+        int snapshotsToRetain = max(1, snapshots - materializedViewRefreshMaxSnapshotsToExpire);
+        try {
+            executeExpireSnapshots(
```
findinpath, line 4052: This operation may take a while because it involves talking with the object storage on deleting metadata & data files.
This operation can be considered a "housekeeping" task and not critical to the success of refreshing the MV.
Advocating here to perform the operation in a different thread.

The end user will have to wait only until the MV refresh operation completes, not also until the additional housekeeping is done.

raunaqmorarka: The whole max-snapshots-to-expire logic is to prevent the runtime from going too high. In general, we don't want queries to leave behind async work in the cluster. Whatever the query does should happen within the context of the query, otherwise it is difficult to track and monitor what is going in the cluster.
The only other alternative is to expose the table maintenance operations on MVs and add running those to user responsibilities.

plugin/trino-iceberg/src/test/resources/iceberg/materialized_view_expired_snapshots/README.md
```diff
@@ -0,0 +1,32 @@
+Created with Trino using Iceberg version 1.10.1-e.1
```
findinpath: Let's remove `-e.1`
This version does not exist on any public repo

plugin/trino-iceberg/src/test/resources/iceberg/materialized_view_expired_snapshots/README.md
```diff
@@ -0,0 +1,32 @@
...
+```sql
+CREATE TABLE source_table (a INT);
+INSERT INTO source_table SELECT 0;
```
findinpath: `INSERT INTO source_table VALUES 0;`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
```diff
@@ -0,0 +1,234 @@
...
+                        "s3.aws-access-key", MINIO_ACCESS_KEY,
+                        "s3.aws-secret-key", MINIO_SECRET_KEY,
+                        "s3.region", MINIO_REGION,
+                        "s3.endpoint", minio.getMinioAddress(),
+                        "iceberg.register-table-procedure.enabled", "true",
+                        "s3.path-style-access", "true"))
+                .build();
+    }
+
+    @Test
```
findinpath, line 103: add
```
    /**
     * @see iceberg.materialized_view_expired_snapshots
     */
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
```diff
@@ -0,0 +1,234 @@
...
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@TestInstance(PER_CLASS)
+public class TestIcebergMaterializedViewExpiredSnapshotCleanup
+        extends AbstractTestQueryFramework
+{
+    private static final String MV_METADATA_FILE_NAME = "00005-b395c800-9c5a-48d2-9dbe-ba7b630881aa.metadata.json";
+    private static final String BUCKET_NAME = "test-bucket-mv-refresh-expired-snapshots";
+    private static final String RESOURCE_DIRECTORY = "materialized_view_expired_snapshots";
+    private static final String FULLY_QUALIFIED_RESOURCE_DIRECTORY = format("s3://%s/%s", BUCKET_NAME, RESOURCE_DIRECTORY);
```
findinpath: FULLY_QUALIFIED_RESOURCE_**DIRECTORY** - that's a prefix, not a directory


S3_LOCATION_PREFIX

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
```diff
@@ -0,0 +1,234 @@
...
+
+        // As generated, the MV will have 6 metadata files, 9 manifest files, 10 manifest lists, and 5 data files
+        assertThat(firstRefreshFiles.get(FileType.METADATA).size()).isEqualTo(6);
+        assertThat(firstRefreshFiles.get(FileType.MANIFEST).size()).isEqualTo(9);
+        assertThat(firstRefreshFiles.get(FileType.SNAPSHOT).size()).isEqualTo(10);
+        assertThat(firstRefreshFiles.get(FileType.DATA).size()).isEqualTo(5);
+
+        // Test that snapshots are removed up to the cap
+        assertUpdate(format("DELETE FROM %s WHERE a = 4", sourceTableName), 1);
+        assertUpdate(format("INSERT INTO %s SELECT 5", sourceTableName), 1);
```
findinpath:
```suggestion
        assertUpdate(format("INSERT INTO %s VALUES 5", sourceTableName), 1);
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
```diff
@@ -0,0 +1,234 @@
...
+        Map<FileType, List<FileEntry>> secondRefreshFiles = getMaterializedViewCurrentFiles(fileSystem);
+
+        // We expect 2 new metadata files: one created before the snapshots are cleaned, and one after, and expired metadata is not cleaned
+        assertThat(secondRefreshFiles.get(FileType.METADATA).size()).isEqualTo(8);
+        // We expect 2 new manifest files, but 4 to be removed: as there are 5 removed snapshots, one of which was for an empty table
+        assertThat(secondRefreshFiles.get(FileType.MANIFEST).size()).isEqualTo(7);
+        // We expect 2 new manifest lists, but 5 to be removed
+        assertThat(secondRefreshFiles.get(FileType.SNAPSHOT).size()).isEqualTo(7);
+        // We expect 1 new data file, but 3 to be removed as they are orphaned
+        assertThat(secondRefreshFiles.get(FileType.DATA).size()).isEqualTo(4);
```
findinpath: This is pretty hard to follow and very likely to maintain.

findinpath: Consider using this utility method

```
    private BaseTable getStorageTable(String materializedViewName)
    {
        HiveMetastore metastore = getHiveMetastore(getQueryRunner());
        TrinoFileSystemFactory fileSystemFactory = getFileSystemFactory(getQueryRunner());
        TrinoCatalog catalog = getTrinoCatalog(metastore, fileSystemFactory, ICEBERG_CATALOG);
        return catalog.getMaterializedViewStorageTable(SESSION, new SchemaTableName(getSession().getSchema().orElseThrow(), materializedViewName))
                .orElseThrow(() -> new TrinoException(TABLE_NOT_FOUND, "Storage table metadata not found for materialized view " + materializedViewName));
    }
```

and check the `Lists.newArrayList(storageTable.snapshots())` after every refresh to ensure that the expected snapshots have been expired.

We're checking here whether the snapshots have been expired, not how the expiry of the snapshots actually works. I'm assuming that `expired_snapshots` has been correspondingly tested.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -593,6 +596,34 @@ public IcebergConfig setIncrementalRefreshEnabled(boolean incrementalRefreshEnab
...
+    public Duration getMaterializedViewRefreshSnapshotRetentionPeriod()
+    {
+        return materializedViewRefreshSnapshotRetentionPeriod;
+    }
+
+    @Config("iceberg.materialized-views.refresh-snapshot-retention-period")
+    @ConfigDescription("Retention threshold to use when expiring snapshots during materialized view refresh")
+    public IcebergConfig setMaterializedViewRefreshSnapshotRetentionPeriod(Duration materializedViewRefreshSnapshotRetentionPeriod)
+    {
+        this.materializedViewRefreshSnapshotRetentionPeriod = materializedViewRefreshSnapshotRetentionPeriod;
```
findinpath, line 623: Should this  config value be correlated with `iceberg.expire-snapshots.min-retention`?
I'm looking at `io.trino.plugin.iceberg.IcebergMetadata#executeExpireSnapshots`

https://github.com/trinodb/trino/blob/a5b0dbadb21b49e0270d660383738e96ecd7a620/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java#L2158-L2165

findinpath: Nevermind, I see `ZERO` used for `minRetention`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -593,6 +596,34 @@ public IcebergConfig setIncrementalRefreshEnabled(boolean incrementalRefreshEnab
...
+    public int getMaterializedViewRefreshMaxSnapshotsToExpire()
+    {
+        return materializedViewRefreshMaxSnapshotsToExpire;
+    }
+
+    @Config("iceberg.materialized-views.refresh-max-snapshots-to-expire")
+    @ConfigDescription("Maximum number of snapshots to remove during materialized view refresh")
+    public IcebergConfig setMaterializedViewRefreshMaxSnapshotsToExpire(int materializedViewRefreshMaxSnapshotsToExpire)
+    {
+        this.materializedViewRefreshMaxSnapshotsToExpire = materializedViewRefreshMaxSnapshotsToExpire;
```
findinpath, line 609: I'm running locally `IcebergMinioHiveMetastoreQueryRunnerMain`

```
"iceberg.materialized-views.refresh-snapshot-retention-period", "1m",
```

and after doing a few operations and refreshing the MV (a few minutes later) i see 1 snapshot while debugging, but all the metadata files are still there.
Arent' they supposed to be removed?

findinpath: Why do we call expire snapshots without cleaning expired metadata ? what does the expired metadata serve for?

```
            executeExpireSnapshots(
                    session,
                    ((IcebergTableHandle) tableHandle).getSchemaTableName(),
                    materializedViewRefreshSnapshotRetentionPeriod,
                    ZERO,
                    snapshotsToRetain,
                    false);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2149,27 +2165,31 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
         }
     }
 
-    private void executeExpireSnapshots(ConnectorSession session, IcebergTableExecuteHandle executeHandle)
+    private void executeExpireSnapshots(
+            ConnectorSession session,
+            SchemaTableName schemaTableName,
+            Duration retentionThreshold,
+            Duration minRetention,
```
raphaelsolarski: `retentionThreshold` -> `retention`
`minRetention` -> `lowerRetentionLimit`
wdyt?

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -184,6 +184,20 @@ implementation is used:
...
   - `true`
+* - `iceberg.materialized-views.refresh-max-snapshots-to-expire`
+  - Maximum number of materialized view snapshots to expire when performing a
+    refresh. A value of `0` will result in no snapshots being removed. Higher
+    values will impact refresh performance. Any snapshots within the duration of
+    the `iceberg.materialized-views.refresh-snapshot-retention-period`
+    configuration property will be retained. Equivalent session property is
+    `materialized_view_refresh_max_expired_snapshots`.
+  - `0`
+* - `iceberg.materialized-views.refresh-snapshot-retention-period`
```
raphaelsolarski, line 196: I would consider naming it `iceberg.materialized-views.refresh-snapshot-ttl` to point out that it is time after which snapshot can be considered stale/expired (not cleaning period/interval).  But is see that it somehow refers to already existing `expire_snapshots_min_retention` session property.

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -184,6 +184,20 @@ implementation is used:
...
+  - The duration in which materialized view snapshots will be retained. During a
+    materialized view refresh, any snapshots older than this value will be
```
raphaelsolarski: Would it be correct?
`any snapshots older than this value` -> `any snapshots older than this value (except the latest one)`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
JohnFitzpatrick44: In the implementation PR for clean_expired_metadata, I saw it was disabled by default, and there was some discussion around leaving it disabled for performance purposes.

https://github.com/trinodb/trino/pull/27362/changes/47a4ba0dd37a5688829d9431093033b61d19dc1d#r2540573565

I wanted to limit the impact of this change to refresh performance so I left it disabled. We can always enable it though.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
JohnFitzpatrick44: These are taken from previous names: `executeHandle.retentionThreshold()` and `getExpireSnapshotMinRetention`. I'm inclined to keep these as-is

docs/src/main/sphinx/connector/iceberg.md
JohnFitzpatrick44: The "current" snapshot would not be older than this value, as it's still valid. I think it's safe to assume the current version of the table will not be expired

docs/src/main/sphinx/connector/iceberg.md
JohnFitzpatrick44: Forgot to update these from a previous version, thanks!

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
JohnFitzpatrick44: Removing this method and only checking snapshot counts as proposed by @findinpath

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
raunaqmorarka: `cleanExpiredMetadata` seems to be relevant only when table can have changes to partition spec or schema, that won't be the case for MV storage tables

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMaterializedViewExpiredSnapshotCleanup.java
JohnFitzpatrick44: Assertions should be easier to read now, just checking snapshot counts and verifying oldest snapshots are removed.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: minRetention -> minimumAllowedRetention

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -4020,10 +4040,28 @@ public Optional<ConnectorOutputMetadata> finishRefreshMaterializedView(
...
+        int snapshots = size(icebergTable.snapshots());
+        int snapshotsToRetain = max(1, snapshots - materializedViewRefreshMaxSnapshotsToExpire);
+        try {
+            executeExpireSnapshots(
+                    session,
+                    ((IcebergTableHandle) tableHandle).getSchemaTableName(),
+                    materializedViewRefreshSnapshotRetentionPeriod,
+                    ZERO,
+                    snapshotsToRetain,
+                    true);
```
raunaqmorarka: This should stay `false`

JohnFitzpatrick44: I may have misunderstood your previous comment. MVs do generate metadata files every time they're refreshed, but their partition spec/schema shouldn't change, so is there a downside to removing those additional metadata files when we refresh? Seems unnecessary to have a history.

raunaqmorarka: The partition spec and schema are part of the metadata json, so I think this is just about trimming some entries from the json rather than removing files. I doubt storage table will even have any extra entries around schema to be trimmed given that it doesn't change schema. 
In any case, we don't want to risk increased latency for that while we don't do that by default for normal tables as well.
You can look at tweaking `org.apache.iceberg.TableProperties#METADATA_DELETE_AFTER_COMMIT_ENABLED` or `org.apache.iceberg.TableProperties#METADATA_PREVIOUS_VERSIONS_MAX_DEFAULT` for storage tables to reduce the number of jsons in another PR/issue.

raunaqmorarka (approved):

PR #28012, opened 2026-01-27
Run modernizer with project.build.targetJdk

ebyhr:
## Description

This enables running modernizer according to the Java version required by each module.

## Release notes

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -159,7 +159,7 @@
         <air.check.skip-jacoco>true</air.check.skip-jacoco>
         <air.java.version>25.0.1</air.java.version>
         <air.javadoc.lint>syntax,reference,html,-missing</air.javadoc.lint>
-        <air.modernizer.java-version>11</air.modernizer.java-version>
+        <air.modernizer.java-version>${project.build.targetJdk}</air.modernizer.java-version>
         <air.release.preparation-goals>clean verify -DskipTests</air.release.preparation-goals>
         <!--
           America/Bahia_Banderas has:


ebyhr:
CI hit #28013

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -132,7 +132,7 @@ public static JsonParser createJsonParser(JsonFactory factory, Slice json)
         // is still valid for small inputs.
         if (json.length() < STRING_READER_LENGTH_LIMIT) {
             // StringReader is more performant than InputStreamReader for small inputs
-            return factory.createParser(new StringReader(json.toStringUtf8()));
+            return factory.createParser(Reader.of(json.toStringUtf8()));
```
findepi, line 135: update comment above

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
```diff
@@ -252,7 +251,7 @@ private static StageStats toStageStats(StageId stageId, BasicStagesInfo stages,
     private static int countStageAndAddGlobalUniqueNodes(BasicStageInfo stageInfo, Set<String> globalUniqueNodes)
     {
         List<TaskInfo> tasks = stageInfo.getTasks();
-        Set<String> stageUniqueNodes = Sets.newHashSetWithExpectedSize(tasks.size());
+        Set<String> stageUniqueNodes = HashSet.newHashSet(tasks.size());
```
findepi: import statically

plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/query/PinotQueryBuilder.java
```diff
@@ -185,7 +185,7 @@ private static Object convertValue(Type type, Object value)
             return ((Slice) value).toStringUtf8();
         }
         if (type instanceof VarbinaryType) {
-            return Hex.encodeHexString(((Slice) value).getBytes());
+            return HexFormat.of().formatHex(((Slice) value).getBytes());
```
findepi, line 188: is it the same in terms of upper/lower case?

pom.xml
```diff
@@ -159,7 +159,7 @@
         <air.check.skip-jacoco>true</air.check.skip-jacoco>
         <air.java.version>25.0.1</air.java.version>
         <air.javadoc.lint>syntax,reference,html,-missing</air.javadoc.lint>
-        <air.modernizer.java-version>11</air.modernizer.java-version>
+        <air.modernizer.java-version>${project.build.targetJdk}</air.modernizer.java-version>
```
findepi, line 162: I thought we pinned modernizer version because we were opinionated and modernizer is opinionated and we didn't agree about which code is better.
I am ok setting this here for now, but we should agree that we may revisit this and pin again.

plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/query/PinotQueryBuilder.java
```diff
@@ -29,9 +29,9 @@
 import io.trino.spi.type.Type;
 import io.trino.spi.type.VarbinaryType;
 import io.trino.spi.type.VarcharType;
-import org.apache.commons.codec.binary.Hex;
```
findepi, line 32: Would be nice to remove this dependency.
The other usage seems to be in https://github.com/trinodb/trino/blob/9e8029ae962658d890ea5c42cb03d7ac46a40a61/plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/decoders/VarbinaryDecoder.java#L51

plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/query/PinotQueryBuilder.java
ebyhr: They are same as far as I confirmed the javadoc & code.

PR #28018, opened 2026-01-27
Add default compression codec for parquet in Iceberg

hqbhoho:
## Description
Following PR #25755, if a user doesn't explicitly set the `compression-codec`  for Parquet, the `write.parquet.compression-codec` property is removed. Consequently, Trino falls back to the global `iceberg.compression-codec` configuration (defaulting to `ZSTD`) to write data files.

However, if Spark is used to rewrite these data files, it defaults to `GZIP` compression. To avoid this inconsistency, could we modify the logic to only remove the `write.parquet.compression-codec` property if the file format is not `Parquet`?

Or maybe we should configure a default compression codec for each file format?

## Release notes

```markdown
## Section
* Add default compression codec `ZSTD` for parquet in Iceberg
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
@@ -915,10 +915,10 @@ public final class IcebergUtil
             transaction = catalog.newCreateTableTransaction(session, schemaTableName, schema, partitionSpec, sortOrder, Optional.ofNullable(tableLocation), createTableProperties(tableMetadata, allowedExtraProperties));
         }
 
-        // If user doesn't set compression-codec for parquet, we need to remove write.parquet.compression-codec property,
+        // If user doesn't use parquet as the file format, we need to remove write.parquet.compression-codec property,
         // Otherwise Iceberg will set write.parquet.compression-codec to zstd by default.
-        String parquetCompressionValue = transaction.table().properties().get(PARQUET_COMPRESSION);
-        if (parquetCompressionValue != null && parquetCompressionValue.isEmpty()) {
+        IcebergFileFormat fileFormat = getFileFormat(transaction.table());
+        if (fileFormat.toIceberg() != FileFormat.PARQUET) {
             transaction.updateProperties()
                     .remove(PARQUET_COMPRESSION)
                     .commit();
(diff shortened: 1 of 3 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
raunaqmorarka, line 925: For my understanding can you clarify if the implemented behaviour is:

- There isn't a distinction between user setting compression_codec through table property and the catalog configuration property. For a new/replaced table, the table property defaults to the catalog configuration property and we always use the table property to define the iceberg property, regardless of whether user set the table property themselves or whether it is derived from the catalog configuration property.
- If file format is not parquet, we remove the `PARQUET_COMPRESSION` iceberg property.

If so, that makes sense to me.

chenjian2664:
Please fix the ci failures, they look related https://github.com/trinodb/trino/actions/runs/21392360484/job/61583054131?pr=28018

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergSystemTables.java
```diff
@@ -849,6 +849,12 @@ public void testPropertiesTable()
         try (TestTable table = newTrinoTable("test_properties", "(x BIGINT,y DOUBLE) WITH (sorted_by = ARRAY['y'])")) {
             Table icebergTable = loadTable(table.getName());
             Map<String, String> actualProperties = getTableProperties(table.getName());
+            if (format == PARQUET) {
+                assertThat(actualProperties).hasSize(8);
+                assertThat(actualProperties).contains(entry("write.parquet.compression-codec", "zstd"));
```
Math-ias, line 854: I find this behavior uneven. We will write a parquet compression-codec table property but not `write.orc.compression-codec`?

Math-ias (changes requested):
I think there's a bit of a mess of defaults here ...

[The iceberg specification sets default values for `compression-codec` table properties](https://iceberg.apache.org/docs/latest/configuration/#write-properties) ...

<table>
<thead>
<tr><th>Property</th><th>Default</th></tr>
</thead>
<tbody>
<tr><td>write.parquet.compression-codec</td><td>zstd</td></tr>
<tr><td>write.avro.compression-codec</td><td>gzip</td></tr>
<tr><td>write.orc.compression-codec</td><td>zlib</td></tr>
</tbody>
</table>

I think everyone is ignoring this part of the spec though.
 * Surely we're ignoring this in trino (because the catalog compression codec property uses a global default of zstd, as you mentioned. That matches parquet but not orc nor avro).
 * I suspect spark is ignoring these defaults as well if they write GZIPd parquet by default as you say.

If we were to take up having default values, then we'd risk making the `iceberg.compression-codec` catalog property (as it is today) do nothing? We'd want to change it from having [a default value](https://github.com/trinodb/trino/blob/a75c9d23147beefe345570849c64c6574d4fa769/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java#L64) to being nullable. Then the precedence could be "iceberg default" < "catalog property" < "table property".

That's still different to what's being done in this PR.

Is the mixing of codecs across trino and spark bad for your use-case? If we set the iceberg compression-codec table properties in parquet I'd rather do it for avro and orc as well though I'm curious about why we want the same trino/spark behavior here.

Issue #27358, opened 2025-11-18
Iceberg OPTIMIZE_MANIFESTS new strategy significantly increased planning time for large partitioned tables.


a1exandr0:
After #26616 changes new strategy to optimize manifests is the following:
* Calculate optimal number of manifests
* Cap it by 100
* For every data file get hash of top level partition and modulo it by number of manifests to determine which manifest would file be attached to

exact code reference - [click](https://github.com/trinodb/trino/pull/26616/files#diff-e1cb17efec6787989f9df9ee40c4f2809ff3fe946cd2ec721ff8932b131997b8L2127-R2136)

This creates following issue, that we are currently facing:
* You have big partitioned table with partitioning being `[hour(timestamp), {{ some medium cardinality column/columns up to 50 unique vals}}]` - all files are at least 128MB to not create many small files
* Your analytical use-cases are usually scanning few hours and up to 1 day of data, filtering on `timestamp`
* Before the change:
  * You would have single manifest per 1 hour partition (you could also add virtual day partition on top level to only have 1 manifest a day)
  * Your queries would read only few manifests thanks to manifest min/max pruning
  * Over time the only thing that would grow would be manifest list and that overhead is manageable
* After the change:
  * Every query reads ALL manifests due to new logic of distributing data files across manifests
  * For use-case above - we get approx 2MB of manifest data every day -> **`in few years we would have GBs of manifests that will be scanned every query`**
  * **`This means that planning time would grow significantly as more data files are added and manifest pruning is essentially not used`**

Is there any plan to address this scalability issue?

One way I see that - is by parametrizing the `optimize_manifests` function to have two modes:
* one manifest per partition (use old approach)
* use new approach

This was also proposed here - #26324

Without this our team will have to either fork the repo or stick to version 476 that did use old optimize_manifests strategy.

findepi:
cc @ebyhr

a1exandr0:
cc @raunaqmorarka

PR #28022, opened 2026-01-27
Cluster manifests by partitions in optimize_manifests

raunaqmorarka:
## Description
Sort unique partition values of the first partitioning column and group 
them by assigning common clustering key to adjacent partitions. 
We avoid creating too many manifests by limiting the number of cluster keys 
based on current size of manifests and `commit.manifest.target-size-bytes`

Clustering manifests by partitions helps to improve performance of reads with filters on partitions.

Fixes https://github.com/trinodb/trino/issues/27358
We had to drop the original naive clustering implementation which simply clustered by partition values due to https://github.com/trinodb/trino/issues/26323
This implementation avoids that problem by making a preparatory pass over the manifests which range partitions by the first partitioning column to a limited target number of cluster keys, and then uses that as input to `org.apache.iceberg.RewriteManifests#clusterBy`

## Release notes

```markdown
## Iceberg
* Improve `optimize_manifests` to cluster manifests by partitions for improving performance of reads with partition filters. ({issue}`27358`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -203,6 +203,9 @@ import org.apache.iceberg.expressions.Expressions;
 import org.apache.iceberg.expressions.Literal;
 import org.apache.iceberg.expressions.Term;
 import org.apache.iceberg.io.CloseableIterable;
+import org.apache.iceberg.metrics.CommitMetricsResult;
+import org.apache.iceberg.metrics.CommitReport;
+import org.apache.iceberg.metrics.MetricsReporters;
 import org.apache.iceberg.types.Comparators;
 import org.apache.iceberg.types.JavaHash;
 import org.apache.iceberg.types.Type;
@@ -2077,8 +2080,7 @@ public class IcebergMetadata
         IcebergTableExecuteHandle executeHandle = (IcebergTableExecuteHandle) tableExecuteHandle;
         switch (executeHandle.procedureId()) {
             case OPTIMIZE_MANIFESTS:
-                executeOptimizeManifests(session, executeHandle);
-                return ImmutableMap.of();
+                return executeOptimizeManifests(session, executeHandle);
             case DROP_EXTENDED_STATS:
                 executeDropExtendedStats(session, executeHandle);
                 return ImmutableMap.of();
@@ -2108,7 +2110,7 @@ public class IcebergMetadata
         }
     }
 
-    private void executeOptimizeManifests(ConnectorSession session, IcebergTableExecuteHandle executeHandle)
+    private Map<String, Long> executeOptimizeManifests(ConnectorSession session, IcebergTableExecuteHandle executeHandle)
     {
         checkArgument(executeHandle.procedureHandle() instanceof IcebergOptimizeManifestsHandle, "Unexpected procedure handle %s", executeHandle.procedureHandle());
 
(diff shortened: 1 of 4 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2112,24 +2122,131 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
+                manifestFile ->
+                        CloseableIterable.transform(
+                                liveEntries(
+                                        ManifestFiles.read(manifestFile, icebergTable.io()).select(ImmutableList.of("partition"))),
+                                dataFile -> dataFile.copyWithoutStats()));
+        // Collect unique values of top-level partitioning column
+        Set<Object> uniqueValues = new ObjectOpenCustomHashSet<>(icebergHashStrategy);
+        try (CloseableIterable<ContentFile<DataFile>> dataFiles = new ParallelIterable<>(dataFileIterables, icebergScanExecutor)) {
+            dataFiles.forEach(dataFile -> {
+                StructLike partition = dataFile.partition();
```
grantatspothero: This returns the transformed partition value, but we insert it into the set using the icebergHashStrategy generated from the untransformed partition value. 

Worried about correctness when the partition source field type does not match the partition transform type, can you add a test for this path?

raunaqmorarka: Good catch, that was indeed an issue. There was a problem with the partition transform evolution scenario as well. I've made changes to handle those now.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2113,24 +2124,132 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
-        long targetManifestClusters = Math.min(((totalManifestsSize + manifestTargetSizeBytes - 1) / manifestTargetSizeBytes), 100);
+        int targetManifestClusters = toIntExact(Math.min(((totalManifestsSize + manifestTargetSizeBytes - 1) / manifestTargetSizeBytes), 200));
+
+        // Cluster manifests by partitioning fields for better manifest pruning when reading data files with filters
+        Map<Object, Integer> clusteredPartitionValues;
+        if (icebergTable.spec().isPartitioned() && targetManifestClusters > 1) {
+            clusteredPartitionValues = getClusteredPartitionValues(icebergTable, manifests, targetManifestClusters);
+        }
+        else {
+            clusteredPartitionValues = ImmutableMap.of();
```
losipiuk, line 2137: nit: just initilaize with `ImmutableMap.of();` and drop else

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2113,24 +2124,132 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
+        checkArgument(icebergTable.spec().isPartitioned(), "Table %s must be partitioned", icebergTable);
+
+        // Clustering is limited to the top-level partitioning column as that is likely
+        // to be the most effective for filters during reads
+        Type.PrimitiveType firstPartitionFieldType = icebergTable.spec().partitionType()
+                .fields().getFirst()
+                .type().asPrimitiveType();
+        Hash.Strategy<Object> icebergHashStrategy = new IcebergHashStrategy(firstPartitionFieldType);
+
+        Iterable<CloseableIterable<ContentFile<DataFile>>> dataFileIterables = Iterables.transform(
```
losipiuk, line 2194: nit: static import transform

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2113,24 +2124,132 @@ private void executeOptimizeManifests(ConnectorSession session, IcebergTableExec
...
+                dataManifests,
+                manifestFile ->
+                        CloseableIterable.transform(
+                                liveEntries(
+                                        ManifestFiles.read(manifestFile, icebergTable.io(), icebergTable.specs()).select(ImmutableList.of("partition"))),
+                                dataFile -> dataFile.copyWithoutStats()));
```
losipiuk: `ContentFile::copyWithoutStats` - but I am not entirely sure what it does :)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: It's used in a lambda expression later on, so has to be effectively final

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: We use both `Iterables.transform` and `CloseableIterable.transform` in this class, so its better to retain the full name for clarity

PR #28038, opened 2026-01-29
Unbind `S3Presigner` from `S3FileSystemLoader`

findinpath:
## Description

`S3Presigner` instance does not need to be created
at the creation time of the `S3FileSystemLoader`.

In the context of dealing with S3 security mappings, the signer, needs to be corresponding to the `S3Client` instance which corresponds to the security mapping.

https://trino.io/docs/current/object-storage/file-system-s3.html#security-mapping

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemLoader.java
+++ b/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemLoader.java
@@ -20,9 +20,7 @@ import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystemFactory;
 import io.trino.filesystem.s3.S3Context.S3SseContext;
 import jakarta.annotation.PreDestroy;
-import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
-import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
 import software.amazon.awssdk.auth.credentials.WebIdentityTokenFileCredentialsProvider;
 import software.amazon.awssdk.core.checksums.RequestChecksumCalculation;
 import software.amazon.awssdk.core.client.config.ClientOverrideConfiguration;
@@ -34,10 +32,7 @@ import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.s3.LegacyMd5Plugin;
 import software.amazon.awssdk.services.s3.S3Client;
 import software.amazon.awssdk.services.s3.S3ClientBuilder;
-import software.amazon.awssdk.services.s3.S3Configuration;
 import software.amazon.awssdk.services.s3.presigner.S3Presigner;
-import software.amazon.awssdk.services.sts.StsClient;
-import software.amazon.awssdk.services.sts.StsClientBuilder;
 import software.amazon.awssdk.services.sts.auth.StsAssumeRoleCredentialsProvider;
 
 import java.net.URI;
@@ -51,6 +46,9 @@ import java.util.function.Function;
 import static com.google.common.base.Preconditions.checkState;
 import static io.airlift.concurrent.Threads.daemonThreadsNamed;
 import static io.trino.filesystem.s3.S3FileSystemConfig.RetryMode.getRetryStrategy;
+import static io.trino.filesystem.s3.S3FileSystemUtils.createS3PreSigner;
+import static io.trino.filesystem.s3.S3FileSystemUtils.createStaticCredentialsProvider;
+import static io.trino.filesystem.s3.S3FileSystemUtils.createStsClient;
 import static java.lang.Math.toIntExact;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.Executors.newCachedThreadPool;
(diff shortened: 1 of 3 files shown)


wendigo (approved):

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.awssdk.services.s3.S3Configuration;
+import software.amazon.awssdk.services.s3.presigner.S3Presigner;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.StsClientBuilder;
+import software.amazon.awssdk.services.sts.auth.StsAssumeRoleCredentialsProvider;
+
+import java.net.URI;
+import java.util.Optional;
+
+public final class S3FileSystemUtils
```
wendigo: We can just call it `S3PresignerFactory` since it contains only S3Presigner creation logic

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+                    .build());
+        }
+        else {
+            staticCredentialsProvider.ifPresent(s3::credentialsProvider);
+        }
+
+        return s3.build();
+    }
+
+    public static StsClient createStsClient(S3FileSystemConfig config, Optional<AwsCredentialsProvider> credentialsProvider)
```
wendigo: private

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+        StsClientBuilder sts = StsClient.builder();
+        Optional.ofNullable(config.getStsEndpoint()).map(URI::create).ifPresent(sts::endpointOverride);
+        Optional.ofNullable(config.getStsRegion())
+                .or(() -> Optional.ofNullable(config.getRegion()))
+                .map(Region::of).ifPresent(sts::region);
+        credentialsProvider.ifPresent(sts::credentialsProvider);
+        return sts.build();
+    }
+
+    public static Optional<AwsCredentialsProvider> createStaticCredentialsProvider(S3FileSystemConfig config)
```
wendigo: private

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
```diff
@@ -0,0 +1,96 @@
...
+import software.amazon.awssdk.services.sts.auth.StsAssumeRoleCredentialsProvider;
+
+import java.net.URI;
+import java.util.Optional;
+
+public final class S3FileSystemUtils
+{
+    private S3FileSystemUtils() {}
+
+    public static S3Presigner createS3PreSigner(S3FileSystemConfig config, S3Client s3Client)
```
wendigo, line 35: just `create`

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
findinpath: It is being used in `S3FileSystemLoader` as well.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
findinpath: It is being used in `S3FileSystemLoader` as well.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemUtils.java
findinpath: All the methods in the class are used both in `S3FileSystemLoader` and also on `S3FileSystemFactory`.
The name `S3PresignerFactory` seems not quite fit for what the class does - utility methods.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystemLoader.java
```diff
@@ -110,6 +107,7 @@ public TrinoFileSystemFactory apply(Location location)
             Optional<S3SecurityMappingResult> mapping = mappingProvider.orElseThrow().getMapping(identity, location);
 
             S3Client client = clients.computeIfAbsent(mapping, _ -> clientFactory.create(mapping));
+            S3Presigner preSigner = preSigners.computeIfAbsent(mapping, _ -> createS3PreSigner(config, client));
```
raphaelsolarski, line 110: Do we need to ensure exactly one presigner/s3 client per mapping or can it be just best effort?

findinpath: There was a singleton before, so I've followed the same pattern - one client & presigner per security mapping.

PR #28042, opened 2026-01-29
Rename getAccessTokenClaims method

ssheikin:
## Description

Be more specific from what token this method can get claims.
Implementation-wise, NimbusOAuth2Client, for example, utilises JWTProcessor configured specifically for accessTokenProcessor, which cannot be reused for other tokens.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/security/oauth2/JweTokenSerializer.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/oauth2/JweTokenSerializer.java
@@ -112,7 +112,7 @@ public class JweTokenSerializer
     {
         requireNonNull(tokenPair, "tokenPair is null");
 
-        Map<String, Object> claims = client.getClaims(tokenPair.accessToken()).orElseThrow(() -> new IllegalArgumentException("Claims are missing"));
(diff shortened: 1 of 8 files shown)


findepi:
Can you please fill PR description and remove the parts that don't need to be filled?

core/trino-main/src/main/java/io/trino/server/security/oauth2/JweTokenSerializer.java
```diff
@@ -112,7 +112,7 @@ public String serialize(TokenPair tokenPair)
     {
         requireNonNull(tokenPair, "tokenPair is null");
 
-        Map<String, Object> claims = client.getClaims(tokenPair.accessToken()).orElseThrow(() -> new IllegalArgumentException("Claims are missing"));
+        Map<String, Object> claims = client.getAccessTokenClaims(tokenPair.accessToken()).orElseThrow(() -> new IllegalArgumentException("Claims are missing"));
```
chenjian2664: The error message could be updated as well

ssheikin:
Comments are addressed. Can we have this PR merged?

PR #28046, opened 2026-01-29
Update to datasketches-java 9.0.0

wendigo:
https://github.com/apache/datasketches-java/releases/tag/9.0.0

Important part:

> None of these changes have affected the binary compatibility of the serialized versions of these sketches. The sketches in this library can still interpret serialized versions of the same sketch from earlier release versions of this library and from other language versions of this library.

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/io/ByteBuffers.java
+++ b/lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/io/ByteBuffers.java
@@ -34,22 +34,4 @@ public final class ByteBuffers
         checkArgument(byteBuffer.remaining() == array.length, "buffer has %s remaining bytes while array length is %s", byteBuffer.remaining(), array.length);
         return array;
     }
-
-    /**
-     * Gets the bytes the provided {@link ByteBuffer} represents, without advancing buffer position.
-     * The returned byte array may be shared with the buffer.
-     */
-    public static byte[] getBytes(ByteBuffer byteBuffer)
-    {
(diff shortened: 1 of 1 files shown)


raunaqmorarka (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/aggregation/DataSketchStateSerializer.java
```diff
@@ -51,7 +51,7 @@ public static void serializeToVarbinary(DataSketchState state, BlockBuilder out)
             checkArgument(state.getUpdateSketch() == null || state.getCompactSketch() == null, "A state must not have both transient accumulator and combined form set");
             CompactThetaSketch compactSketch = Optional.ofNullable(state.getCompactSketch())
                     .orElseGet(() -> state.getUpdateSketch().compact());
-            Slice slice = Slices.wrappedBuffer(compactSketch.toByteArray());
+            Slice slice = Slices.wrappedBuffer(compactSketch.toByteArrayCompressed());
```
findinpath, line 54: Could you pls explain why adding the sketch compressed makes sense?
I'm thinking that it will get decompressed anyway.
Does it leave a high memory footprint?

findinpath (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/aggregation/DataSketchStateSerializer.java
wendigo: Compressed sketch is around 20% smaller when sketch gets big enough. This saves both memory, deserialization roundtrip and network latency.

PR #28062, opened 2026-02-01
Fix flaky tests when abort flow takes longer to cleanup

avibhstarburst:
Comparative analysis reveals a race condition between the asynchronous transactionManager.asyncAbort task and the post-rollback verification query. While the local file system processes these deletions near-instantaneously, high-latency environments like MinIO allow the test to outpace the background cleanup. This leads to a synchronization drift where the test asserts state while the cleanup task is still actively deleting files

## Release notes

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
@@ -2581,7 +2581,11 @@ public abstract class BaseConnectorTest
                 }))
                 .isInstanceOf(RollbackException.class);
 
-        assertQuery(format("SELECT count(*) FROM %s", table), "SELECT 0");
+        // retry due to async rollback cleanup which may outrun the next verification query on remote file systems,
+        // leading to FileNotFoundException on splits
+        assertEventually(
+                new Duration(1, SECONDS),
+                () -> assertQuery(format("SELECT count(*) FROM %s", table), "SELECT 0"));
     }
 
     private static class RollbackException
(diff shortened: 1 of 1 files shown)


testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -2581,7 +2581,8 @@ public void testRollback()
                 }))
                 .isInstanceOf(RollbackException.class);
 
-        assertQuery(format("SELECT count(*) FROM %s", table), "SELECT 0");
+        assertEventually(new Duration(1, SECONDS),
+                () -> assertQuery(format("SELECT count(*) FROM %s", table), "SELECT 0"));
```
findepi, line 2588: please add code comment for the why

nit: we put all arguments on one line, or each on separate line

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -5244,9 +5245,11 @@ public void testSelectAfterInsertInTransaction()
...
             }
             // SELECT again after transaction completes
-            assertQuery(
-                    "TABLE " + tableName,
-                    "SELECT nationkey, name, regionkey FROM nation WHERE nationkey IN " + (commit ? "(1, 2)" : "(1)"));
+            boolean isCommit = commit;
+            assertEventually(new Duration(1, SECONDS),
+                    () -> assertQuery(
+                            "TABLE " + tableName,
+                    "SELECT nationkey, name, regionkey FROM nation WHERE nationkey IN " + (isCommit ? "(1, 2)" : "(1)")));
```
findepi:
```suggestion
            assertEventually(
                    new Duration(1, SECONDS),
                    () -> assertQuery("TABLE " + tableName,
                    "SELECT nationkey, name, regionkey FROM nation WHERE nationkey IN " + (isCommit ? "(1, 2)" : "(1)")));
```

findepi: please add code comment for the why

PR #28064, opened 2026-02-01
Expand internal shared secret using HKDF-SHA256

wendigo:
## Description

Apply HKDF-SHA256 to the shared internal communication secret to
derive the actual key used for securing internal traffic.
That makes the effective internal key stronger and more uniformly sized/structured,
even if the configured secret was weak or short.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/test/java/io/trino/failuredetector/TestHeartbeatFailureDetector.java
+++ b/core/trino-main/src/test/java/io/trino/failuredetector/TestHeartbeatFailureDetector.java
@@ -33,8 +33,6 @@ import io.trino.server.InternalCommunicationConfig;
 import io.trino.server.StartupStatus;
 import io.trino.server.security.SecurityConfig;
 import io.trino.spi.NodeVersion;
-import jakarta.ws.rs.GET;
-import jakarta.ws.rs.Path;
 import org.junit.jupiter.api.Test;
 
 import java.net.SocketTimeoutException;
@@ -43,7 +41,6 @@ import java.net.URI;
 import static io.airlift.configuration.ConfigBinder.configBinder;
 import static io.airlift.discovery.client.DiscoveryBinder.discoveryBinder;
 import static io.airlift.discovery.client.ServiceTypes.serviceType;
-import static io.airlift.jaxrs.JaxrsBinder.jaxrsBinder;
 import static org.assertj.core.api.Assertions.assertThat;
 
 public class TestHeartbeatFailureDetector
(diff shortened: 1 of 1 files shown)


losipiuk (approved):
Does not seem crucial to me. But not much harm also.

findepi:
> ## Description
> ## Additional context and related issues
> ## Release notes

Please update the PR description.

core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
```diff
@@ -24,12 +23,16 @@
 import io.jsonwebtoken.JwtParser;
 import io.trino.server.security.InternalPrincipal;
 import io.trino.server.security.SecurityConfig;
+import io.trino.spi.NodeVersion;
 import io.trino.spi.security.Identity;
 import jakarta.ws.rs.container.ContainerRequestContext;
 import jakarta.ws.rs.core.Response;
 
+import javax.crypto.KDF;
```
findepi, line 31: > Expand internal shared secret using HKDF-SHA256
This produces stronger internal communication key,
even if configuration-provided one is weak.

Please add more color to the commit message.

And update the PR description.

wendigo: Done

core/trino-main/src/main/java/io/trino/server/InternalAuthenticationManager.java
```diff
@@ -189,4 +191,22 @@ public boolean isExpired()
...
+
+    private static SecretKey expandKey(String sharedSecret, NodeVersion nodeVersion)
+    {
+        try {
+            KDF hkdf = KDF.getInstance("HKDF-SHA256");
+
+            AlgorithmParameterSpec params =
+                    HKDFParameterSpec.ofExtract()
+                            .addIKM(sharedSecret.getBytes(UTF_8))
+                            .addSalt(nodeVersion.toString().getBytes(UTF_8))
```
findepi, line 203: Does this mean that, if two nodes have different version, they will not be able to authenticate each other? So manifestation of version mismatch changes?

wendigo: Yes. Correct.

losipiuk:
>That makes the effective internal key stronger

This is obviously not true. Overal entropy in the key is the same.

PR #28091, opened 2026-02-03
Populate sort_order_id in iceberg metadata during writes

raunaqmorarka:
## Description
Use `org.apache.iceberg.DataFiles.Builder#withSortOrder` to populate the table sort order 
in the manifests metadata when the sort order was enforced by the worker

## Release notes

```markdown
## Section
```


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1512,19 +1513,32 @@ private List<String> getChildNamespaces(ConnectorSession session, String parentN
...
+                supportedSortFields,
+                getSupportedSortOrderId(supportedSortFields, tableSortOrder),
                 getPartitionColumns(table, typeManager),
                 table.location(),
                 getFileFormat(table),
                 table.properties(),
                 table.io().properties());
     }
 
+    private static OptionalInt getSupportedSortOrderId(List<TrinoSortField> supportedSortFields, SortOrder tableSortOrder)
```
grantatspothero: nit: every table always has a sort order `SortOrder.unsorted()` with orderId 0, you could remove the optionality here and just return an int.

grantatspothero:
lgtm [non-binding]

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1513,19 +1513,32 @@ private List<String> getChildNamespaces(ConnectorSession session, String parentN
...
                 getFileFormat(table),
                 table.properties(),
                 table.io().properties());
     }
 
+    private static int getSupportedSortOrderId(List<TrinoSortField> supportedSortFields, SortOrder tableSortOrder)
+    {
+        // Trino supports sorting only by identity transforms, so the supported sort fields in Trino may be a subset of the iceberg table sort order.
+        // The writer can populate sort_order_id in the iceberg metadata only when the iceberg table sort order is fully supported by Trino.
+        if (tableSortOrder.isSorted() && supportedSortFields.size() == tableSortOrder.fields().size()) {
```
findepi: `getSupportedSortFields` and `getSupportedSortOrderId` could be merged into single method that returns both: the trino sort fields and sort ID to store.

That would make the code easier to understand and harder to use wrong.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1989,12 +2004,14 @@ private void finishOptimize(ConnectorSession session, IcebergTableExecuteHandle
...
         Set<DataFile> newFiles = new HashSet<>();
+        Map<Integer, SortOrder> sortOrders = icebergTable.sortOrders();
         for (CommitTaskData task : commitTasks) {
             DataFiles.Builder builder = DataFiles.builder(icebergTable.spec())
                     .withPath(task.path())
                     .withFileSizeInBytes(task.fileSizeInBytes())
                     .withFormat(optimizeHandle.fileFormat().toIceberg())
-                    .withMetrics(task.metrics().metrics());
+                    .withMetrics(task.metrics().metrics())
+                    .withSortOrder(sortOrders.get(task.sortOrderId()));
```
findepi, line 2023: This is where you'd want to use the following, right?
- https://github.com/apache/iceberg/pull/15217

findepi, line 2023: Can `sortOrders.get` can return null? Does it return null for `SortOrder.unsorted().orderId()`?
I know `withSortOrder` accepts null, but it might be more future proof to surface nullity to the reader (supposedly, "everything" in trino is meant to be non-null unless documented)

```
@Nullable SortOrder sortOrder = sortOrders.get(task.sortOrderId()); // document why nullable
...
.withSortOrder(sortOrder)
```


(same in other places)

findepi (approved):
> Description
Use org.apache.iceberg.DataFiles.Builder#withSortOrder to populate the table sort order
in the manifests metadata when the sort order was enforced by the worker

What's the consequence/impact/gain of doing so?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: yes

raunaqmorarka: It can be null when the table sort order is not supported by Trino or sorted writing is disabled in the worker, there is a test covering that scenario.
I don't want to add elaborate explanations of it in the code, since `withSortOrder` does the right thing today and this map look-up will go away entirely after https://github.com/apache/iceberg/pull/15217 is available to use.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: done

PR #28107, opened 2026-02-04
Change taskInstanceId to long & convert execution classes to records

wendigo:
## Description

According to the original change:

> When a client tries to talk to a task to a non-exisiting task, we create
> a new task. In some cases it recreates tasks that were previously torn
> down. Attach a task instance id to each task so that the client can
> identify that the task it was communicating with has changed.

For that purpose, we can use any source of entropy that will generate a random identifier
that allows to distinguish two separate task instances.

randomUUID uses SecureRandom, but we don't really need to use CSPRNG,
because PRNG is enough and much faster (nanoseconds vs hundreds of nanos)

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/buffer/OutputBufferStatus.java
+++ b/core/trino-main/src/main/java/io/trino/execution/buffer/OutputBufferStatus.java
@@ -13,48 +13,20 @@
  */
 package io.trino.execution.buffer;
 
-import com.fasterxml.jackson.annotation.JsonCreator;
-import com.fasterxml.jackson.annotation.JsonProperty;
-
 import java.util.OptionalLong;
 
 import static java.util.Objects.requireNonNull;
 
-public class OutputBufferStatus
+public record OutputBufferStatus(
+        OptionalLong outputBuffersVersion,
+        boolean overutilized,
+        boolean exchangeSinkInstanceHandleUpdateRequired)
 {
     private static final OutputBufferStatus INITIAL = new OutputBufferStatus(OptionalLong.empty(), false, false);
 
-    private final OptionalLong outputBuffersVersion;
-    private final boolean overutilized;
-    private final boolean exchangeSinkInstanceHandleUpdateRequired;
-
-    @JsonCreator
-    public OutputBufferStatus(
-            @JsonProperty("outputBuffersVersion") OptionalLong outputBuffersVersion,
-            @JsonProperty("overutilized") boolean overutilized,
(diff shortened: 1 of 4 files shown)


core/trino-main/src/main/java/io/trino/execution/SqlTask.java
```diff
@@ -85,9 +85,10 @@
 public class SqlTask
 {
     private static final Logger log = Logger.get(SqlTask.class);
+    private static final SplittableRandom random = new SplittableRandom();
```
raunaqmorarka: Capitalize `random` ?

raunaqmorarka: Please add a code comment explaining why SplittableRandom.

raunaqmorarka: Why not ThreadLocalRandom ?

wendigo: It is faster and avoid having a random generator per thread (this will be http worker threads and there are plenty of it)

wendigo:
Benchmarks show consistent 5% cpu improvement and 2% duration

martint:
> Benchmarks show consistent 5% cpu improvement and 2% duration

That doesn't sound real. The random number generator doesn't and all this machinery shouldn't be called that often compared to the cost of all the data processing.

core/trino-main/src/main/java/io/trino/execution/SqlTask.java
```diff
@@ -86,8 +86,17 @@ public class SqlTask
...
+    private static final SplittableRandom RANDOM = new SplittableRandom();
+
+    // Each thread gets its own SplittableRandom instance derived from the root.
+    // This avoids synchronization, eliminates false sharing, and ensures
+    // high-throughput, thread-safe random number generation on hot paths.
+    private static final ThreadLocal<SplittableRandom> RANDOM_THREAD_LOCAL = ThreadLocal.withInitial(RANDOM::split);
+
     private final TaskId taskId;
-    private final String taskInstanceId;
+    private final long taskInstanceId;
```
losipiuk, line 99: @wendigo Can you elaborate why you think just a 64 bit random is good enough, and we are not risking conflicts here. This is not that large space.

wendigo: this is a space within given taskId

PR #15686, opened 2023-01-12
Add metric to measure requests duration of HttpPageBufferClient

radek-starburst:
It adds the metric that measures distribution of requests duration sent by `HttpPageBufferClient`. It was found that it is very useful during investigation. Example of metric:

                 "requestDuration" : {
                    "@class" : "io.trino.plugin.base.metrics.TDigestHistogram",
[...]

PR #15369, opened 2022-12-12
Decide number of clients basing on average request size of client

radek-starburst:
## Description

Let's define the class of queries that are run on skewed data and as a result on some downstream stage there is one (or few nodes) that gather data from empty or nearly empty nodes at upstream stage - i.e. at some upstream stage most nodes are empty (serving no data) or nearly empty (serving little data). We observed that such queries are executed very poorly on Trino.
This PR addresses that issue.

The query listed below is a representant of that class:

```
SELECT
    count(*),
    cc.cc_name,
    cc.cc_class,
    cc.cc_manager,
    cc.cc_mkt_desc,
    cc.cc_market_manager,
    cc.cc_division_name
FROM catalog_sales cs
    RIGHT JOIN call_center cc ON cc.cc_call_center_sk =  cs.cs_call_center_sk
    RIGHT JOIN store s ON cc.cc_closed_date_sk = s.s_closed_date_sk
GROUP BY 2, 3, 4, 5, 6, 7
ORDER BY 1, 2, 3, 4, 5, 6, 7
LIMIT 100;
```

**Before the change it takes 11,5 minutes to finish that query on 32 nodes r6g.4xlarge cluster. After the change it takes 4 minutes to finish on same cluster.** 
No regresion in serial and throughput benchmarks. 

## Release notes

( *) This is not user-visible or docs only and no release notes are required.
( ) Release notes are required, please propose a release note for me.
( ) Release notes are required, with the following suggested text:

PR #28111, opened 2026-02-04
Remove request duration histograms from DirectExchangeClient

wendigo:
These histograms are captured globally for the exchange http client

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/DirectExchangeClient.java
+++ b/core/trino-main/src/main/java/io/trino/operator/DirectExchangeClient.java
@@ -21,7 +21,6 @@ import com.google.errorprone.annotations.concurrent.GuardedBy;
 import io.airlift.http.client.HttpClient;
 import io.airlift.log.Logger;
 import io.airlift.slice.Slice;
-import io.airlift.stats.TDigest;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
 import io.trino.FeaturesConfig.DataIntegrityVerification;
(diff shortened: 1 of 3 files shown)


raunaqmorarka:
> These histograms are captured globally for the exchange http client

Can you mention where these globally captured histograms can be found ?

wendigo:
@raunaqmorarka 
```
MBean Information Item	Value
MBean Name	io.airlift.http.client:type=HttpClient,name=ForExchange
```

raunaqmorarka:
> @raunaqmorarka
> 
> ```
> MBean Information Item	Value
> MBean Name	io.airlift.http.client:type=HttpClient,name=ForExchange
> ```

Those are JMH metrics, do we get them in operator summaries ? That what the original use case of https://github.com/trinodb/trino/pull/15686 was

wendigo:
@raunaqmorarka I've never seen these numbers to be used. Do you have an evidence that this is actually needed?

raunaqmorarka:
> @raunaqmorarka I've never seen these numbers to be used. Do you have an evidence that this is actually needed?

I think that metric came about during investigation of https://github.com/trinodb/trino/pull/15369
I don't know how much it has been used since. The JMX metrics you're pointing to (and most JMX metrics in general) are also almost never used, yet we're keeping them.
Do we have a gain in some benchmark by removing this metric ?

wendigo:
@raunaqmorarka it does show reduction in duration

PR #28116, opened 2026-02-04
Use XxHash3 for calculating page checksums

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/buffer/PagesSerdeUtil.java
+++ b/core/trino-main/src/main/java/io/trino/execution/buffer/PagesSerdeUtil.java
@@ -40,6 +40,8 @@ import static java.util.Objects.requireNonNull;
 
 public final class PagesSerdeUtil
 {
+    private static ThreadLocal<XxHash3Hasher> HASHERS = ThreadLocal.withInitial(XxHash3Native::newHasher);
+
     private PagesSerdeUtil() {}
 
     static final int SERIALIZED_PAGE_POSITION_COUNT_OFFSET = 0;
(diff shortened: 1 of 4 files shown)


findepi (dismissed):

findepi:
> Description

Please fill the PR description:

- what changes
- why

core/trino-main/src/main/java/io/trino/execution/buffer/PagesSerdeUtil.java
```diff
@@ -80,12 +81,24 @@ static Page readRawPage(int positionCount, SliceInput input, BlockEncodingSerde
 
     public static long calculateChecksum(List<Slice> pages)
     {
-        XxHash3Hasher hasher = HASHERS.get();
-        hasher.reset();
-        for (Slice page : pages) {
-            hasher.update(page.byteArray(), page.byteArrayOffset(), page.length());
+        int size = pages.stream().mapToInt(Slice::length).sum();
+        long checksum;
+        if (size > 16384) {
```
martint, line 86: Why?

wendigo: It seems that for small slices the cost of calling native outweights the benefits of a faster hashing function

wendigo:
cc @dain

core/trino-main/src/main/java/io/trino/execution/buffer/PagesSerdeUtil.java
findepi: could the magic live in aircompressor ?

findepi: also, if we cannot move this switch there, the code as it stands now absolutely requires code comment

dain: The overhead of xxhash3 with a 64 bit output is less than xxhash64 in all cases.

dain:
If you make the change I suggested and reopen this, I'll approve it

PR #28117, opened 2026-02-04
Remove unnecessary chunking of pages in lake connector page sinks

raunaqmorarka:
## Description
ORC and parquet writers already do chunking of pages in OrcWriter#write and ParquetWriter#write respectively. 
There is no need for the page sink layer to be chunking pages as the file writers are better suited to apply this logic based on their internal details.

## Release notes

```markdown
## Section
```


raunaqmorarka:
Note that there are small adjustments to `testTargetMaxFileSize` methods because the chunking impacts how often the target max file size check is applied and the tests are dealing with very small file sizes. For real workloads with much larger target file sizes, checking file max size after each Page without the 4096 positions limit should still be good enough.

PR #28130, opened 2026-02-05
Convert engine classes to records

wendigo:
## Release notes

```markdown
## Section
```


core/trino-main/src/main/java/io/trino/sql/PlannerContext.java
```diff
@@ -31,61 +31,28 @@
  * This is used to ease the addition of new services in the future without
  * having to modify large portions the planner and analyzer just to pass around
  * the service.
+ *
+ * @param metadata Be careful when adding services here. This context is used throughout the analyzer and planner, so it is easy to create circular dependencies, just create a junk drawer of services.
  */
-public class PlannerContext
+public record PlannerContext(
```
martint: Value-based equality and identity don't make sense for this, so I don't think it's appropriate to make it a record.

core/trino-main/src/main/java/io/trino/sql/planner/plan/PlanNodeId.java
```diff
@@ -15,24 +15,19 @@
...
 import com.fasterxml.jackson.annotation.JsonValue;
-import com.google.errorprone.annotations.Immutable;
 
 import static io.airlift.slice.SizeOf.estimatedSizeOf;
 import static io.airlift.slice.SizeOf.instanceSize;
 import static java.util.Objects.requireNonNull;
 
-@Immutable
-public class PlanNodeId
+public record PlanNodeId(String id)
```
martint: Converting this to a record makes the id visible to users. Not ideal, as it allows callers to inspect and operate on the underlying id, which defeats the purpose of the abstraction.

wendigo: Id was previously exposed as well. PlanNodeId is just a plain alias to a String so maybe we should just remove this type altogether?

core/trino-main/src/main/java/io/trino/sql/PlannerContext.java
wendigo: Dropped

core/trino-main/src/main/java/io/trino/sql/planner/plan/PlanNodeId.java
martint: It's not currently exposed. There's no getId() method in the current PlanNodeId.

> just a plain alias to a String so maybe we should just remove this type altogether?

No, it's a type-safe abstraction to represent a plan node id. Today it's internally represented as a String, but it doesn't need to be, and should not be confused with one.

wendigo: Beside that commit, is everything else ok? I can drop this one

martint: Yes, everything else is ok.

PR #28138, opened 2026-02-06
Document running tests from IntelliJ after vector API became required

hashhar:
## Description

Since trinodb/trino@bc028fde4424d33d1e99dd7d9d39af0f9962e8fc the vector API usage is no longer
optional. Maven handles the required JVM flags via surefire args but IntelliJ
does not consume these automatically.

## Release notes

```markdown
## Section
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,6 +96,13 @@ here is `--add-modules jdk.incubator.vector` but various `*QueryRunner` classes
 might require additional options (if necessary, check the `air.test.jvm.additional-arguments` 
 property in the `pom.xml` file of the module from which the runner comes).
 
+### Running tests from the IDE
+
+When running individual test classes directly from IntelliJ, you need to
+configure the JUnit run configuration template. Go to Run/Debug Configurations >
+Edit Configuration Templates > JUnit > VM options and set the value to
+`-ea --add-modules=jdk.incubator.vector`.
+
 ### Running the full server
 
 Trino comes with sample configuration that should work out-of-the-box for


hashhar:
Note since the file is gitignored and was added using `add --force` you can safely make local modifications to this file if you want. This PR just provides a good baseline for a "working out of the box" setup.

findepi:
> Note since the file is gitignored and was added using `add --force` you can safely make local modifications to this file if you want.

This is not how .gitignore works.
AFAIU, `.gitignore` rules affect only handling of files not tracked by `git` yet.

.idea/runConfigurations/_template__of_JUnit.xml
```diff
@@ -0,0 +1,9 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="true" type="JUnit" factoryName="JUnit">
+    <option name="TEST_OBJECT" value="class" />
+    <option name="VM_PARAMETERS" value="-ea --add-modules=jdk.incubator.vector" />
+    <method v="2">
+      <option name="Make" enabled="true" />
+    </method>
+  </configuration>
+</component>
```
wendigo: add trailing newline

chenjian2664:
I think the idea is good. Maybe we could add a demo or a short note in the documentation about the vector feature? e.g., similar to the testing server section in [README](https://github.com/trinodb/trino/blob/master/README.md#running-a-testing-server)
This can avoid put the file directly into git, but users would known there is another way to run tests easily

PR #28142, opened 2026-02-06
Verify file sorting from parquet data rather than footer

raunaqmorarka:
## Description
Updates checkParquetFileSorting to avoid relying on row group metadata to verify 
sorting and use a read of the actual data instead. 
This avoids test failures from "Test must produce at least two row groups"

## Release notes

```markdown
## Section
```


Praveen2112 (approved):
CI failures look legit.

PR #28162, opened 2026-02-08
Refactor coordinators management in DistributedQueryRunner

losipiuk:
Replace a concept of backupCoordinator in DistributedQueryRunner with more straightforward logic when one can just add one or more extra coordinators to cluster modeled as DQR. Do some refactors by the way which clean cruft and makes DistributedQueryRunner in line with changes which happened around NodeManager

## Release notes

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
@@ -140,7 +140,6 @@ public final class DistributedQueryRunner
             int workerCount,
             Map<String, String> extraProperties,
             Map<String, String> coordinatorProperties,
-            Optional<Map<String, String>> backupCoordinatorProperties,
             String environment,
             Module additionalModule,
             Optional<Path> baseDataDir,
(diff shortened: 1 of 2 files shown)


testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
```diff
@@ -122,7 +126,7 @@ public final class DistributedQueryRunner
 
     private final Closer closer = Closer.create();
 
-    private TestingTrinoClient trinoClient;
+    private LoadingCache<TestingTrinoServer, TestingTrinoClient> trinoClients;
```
martint: This introduces more complexity. What’s the purpose?

testing/trino-testing/pom.xml
```diff
@@ -125,6 +125,11 @@
             <artifactId>opentelemetry-sdk-trace</artifactId>
         </dependency>
 
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-cache</artifactId>
```
findepi: Needed?

findepi (approved):

findepi:
Please add PR description with the rationale

testing/trino-testing/src/main/java/io/trino/testing/DistributedQueryRunner.java
losipiuk: It is to be able to talk to the cluster over any of defined coordinators. Given current tests do not exploit this I will drop commit which adds this.

losipiuk:
> These seem like a bunch of nontrivial changes without a clear overarching goal. Can you add some explanation in the PR description?

Added

testing/trino-testing/pom.xml
losipiuk: dopped

PR #28164, opened 2026-02-09
Remove redundant HadoopTables from Iceberg tests

ebyhr:
## Release notes

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
@@ -19,12 +19,12 @@ import com.fasterxml.jackson.databind.node.ArrayNode;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 import com.google.common.collect.ImmutableMap;
 import io.trino.Session;
-import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystemFactory;
 import io.trino.metastore.HiveMetastore;
 import io.trino.plugin.hive.HivePlugin;
+import io.trino.plugin.iceberg.catalog.TrinoCatalog;
 import io.trino.plugin.tpch.TpchPlugin;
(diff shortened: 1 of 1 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV3.java
```diff
@@ -716,17 +677,14 @@ void testV3RejectsEncryptionKeysInMetadata()
...
-        long dataFileSize = getFileSize(dataFilePath);
+        Table tempTable = loadTable(temp);
 
         String hadoopTableName = "hadoop_v3_encryption_" + randomNameSuffix();
-        Path hadoopTableLocation = dataDirectory.resolve(hadoopTableName);
-
-        Schema schema = new Schema(Types.NestedField.optional(1, "id", Types.IntegerType.get()));
+        Path hadoopTableLocation = Path.of(tempTable.location()).resolveSibling(hadoopTableName);
 
+        // Use HadoopTables to prevent stale caches from direct metadata.json modification
```
findinpath, line 685: Which cache becomes stale?

findinpath: Consider using `"iceberg.metadata-cache.enabled", "false",` in the configuration of the catalog to avoid the issue you're facing and let's remove completely the usage of HadoopTables.

ebyhr: I intentionally avoided disabling the property since changing the default catalog config value just for a single test doesn't make sense to me.

findinpath: What about creating dynamically a different catalog to avoid caching issues then?

ebyhr: That's overkill to me. I'd merge as-is.

Issue #10814, opened 2022-01-26
Add support to cast type VARBINARY as JSON


cccs-joel:
# Summary:
As mentioned at [JSON doc](https://trino.io/docs/current/functions/json.html), some types are currently not supported with the function CAST AS JSON. This might be due to the fact that the JSON format natively doesn't support binary data. Our use case is to be able to cast a complex column (row, arrays...) as json with some values being unsupported types.

# Description:
```
SELECT CAST(
CAST(TIMESTAMP '2022-01-01' AS TIMESTAMP)
AS JSON)
```
does work even if the TIMESTAMP is not in the supported type list

```
SELECT CAST(
CAST(TIMESTAMP '2022-01-01' AS TIMESTAMP WITH TIME ZONE)
AS JSON)
```
does not work with error 'Cannot cast timestamp(3) with time zone to json'.

```
SELECT CAST(
X'65683F'
AS JSON)
```
does not work with error 'Cannot cast varbinary to json'.

As Spark SQL and potentially other engines do, Trino could convert these values. For example, the following Spark SQL example

```
SELECT TO_JSON(
struct(X'65683F')
) AS x
```
is fine and returns something like {"col1":"ZWg/"}.

Can the cast for timestamp with time zone as json be fixed?

Would Trino accept to add support to cast binaries to json?

kasiafi:
Hi @cccs-joel,
we have recently added a couple of functions to process JSON. You can use them to read JSON data from varbinary.
Please note that these functions won't return JSON type, but they return character string or binary string representation. JSON type is not ANSI-compliant, and with the new functions, we followed the spec carefully.

Let me show how it works.
The varbinary representation of your input is:
```
trino:tiny> SELECT CAST ('{"col1":"ZWg/"}' AS varbinary);
                    _col0
----------------------------------------------
 7b 22 63 6f 6c 31 22 3a 22 5a 57 67 2f 22 7d
(1 row)
```

I'll use the `json_query` function to read that varbinary, and return back the text:
```
trino:tiny> SELECT JSON_QUERY(X' 7b 22 63 6f 6c 31 22 3a 22 5a 57 67 2f 22 7d', 'strict $');
      _col0
-----------------
 {"col1":"ZWg/"}
(1 row)
```

The second argument `strict $` is the JSON path. Here I used it only to return the input object (`$`), but it is a very powerful tool to dig into your JSON data.

[...]

PR #28234, opened 2026-02-10
Add support to cast VARBINARY to JSON using base64

gertjanal:
## Description
Casting of `row(greeting varchar, varbinary planet)` to json was not possible.

This PR adds support by converting the varbinary to base64 string:
`{"greeting": "hello", "planet": "d29ybGQ="}`

Fixes https://github.com/trinodb/trino/issues/10814

## Release notes

```markdown
## General
* Add support for casting `varbinary` to `json` with base64. ({issue}`28234`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
@@ -19,6 +19,7 @@ import com.google.common.io.BaseEncoding;
 import com.google.common.primitives.Ints;
 import io.airlift.slice.Murmur3Hash128;
 import io.airlift.slice.Slice;
+import io.airlift.slice.SliceOutput;
 import io.airlift.slice.Slices;
 import io.airlift.slice.SpookyHashV2;
 import io.airlift.slice.XxHash64;
@@ -26,6 +27,7 @@ import io.trino.spi.TrinoException;
 import io.trino.spi.function.Description;
 import io.trino.spi.function.LiteralParameters;
 import io.trino.spi.function.ScalarFunction;
+import io.trino.spi.function.ScalarOperator;
 import io.trino.spi.function.SqlType;
 import io.trino.spi.type.StandardTypes;
 
@@ -35,7 +37,11 @@ import java.util.zip.CRC32;
 
 import static io.airlift.slice.Slices.EMPTY_SLICE;
 import static io.trino.operator.scalar.HmacFunctions.computeHash;
+import static io.trino.spi.StandardErrorCode.INVALID_CAST_ARGUMENT;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.function.OperatorType.CAST;
+import static io.trino.spi.type.StandardTypes.JSON;
+import static io.trino.spi.type.StandardTypes.VARBINARY;
 import static io.trino.util.Failures.checkCondition;
 
 public final class VarbinaryFunctions
@@ -500,5 +506,28 @@ public final class VarbinaryFunctions
             reverse.setByte(i, inputSlice.getByte((length - 1) - i));
         }
         return reverse;
+    }
+
+    @SqlType(JSON)
+    @ScalarOperator(CAST)
+    public static Slice castToJson(@SqlType(VARBINARY) Slice binary)
+    {
+        Slice encodedBinary = toBase64(binary);
+        Slice slice = Slices.allocate(2 + encodedBinary.length());
+        SliceOutput output = slice.getOutput();
+        output.writeByte('"');
+        output.writeBytes(encodedBinary);
+        output.writeByte('"');
+        return slice;
+    }
+
+    @SqlType(VARBINARY)
+    @ScalarOperator(CAST)
+    public static Slice castFromJson(@SqlType(JSON) Slice json)
+    {
+        if (json.length() < 2 || json.getByte(0) != '"' || json.getByte(json.length() - 1) != '"') {
+            throw new TrinoException(INVALID_CAST_ARGUMENT, "Expected quoted base64-encoded JSON string");
+        }
+        return fromBase64Varbinary(json.slice(1, json.length() - 2));
     }
 }
--- a/core/trino-main/src/main/java/io/trino/util/JsonUtil.java
+++ b/core/trino-main/src/main/java/io/trino/util/JsonUtil.java
@@ -51,6 +51,7 @@ import io.trino.spi.type.StandardTypes;
 import io.trino.spi.type.TimestampType;
 import io.trino.spi.type.TinyintType;
 import io.trino.spi.type.Type;
+import io.trino.spi.type.VarbinaryType;
 import io.trino.spi.type.VarcharType;
 import io.trino.type.BigintOperators;
 import io.trino.type.BooleanOperators;
@@ -81,6 +82,7 @@ import static com.fasterxml.jackson.core.JsonToken.FIELD_NAME;
 import static com.fasterxml.jackson.core.JsonToken.START_ARRAY;
 import static com.fasterxml.jackson.core.JsonToken.START_OBJECT;
 import static com.google.common.base.Verify.verify;
+import static io.trino.operator.scalar.VarbinaryFunctions.fromBase64Varchar;
 import static io.trino.plugin.base.util.JsonUtils.jsonFactoryBuilder;
 import static io.trino.spi.StandardErrorCode.INVALID_CAST_ARGUMENT;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
@@ -92,6 +94,7 @@ import static io.trino.spi.type.IntegerType.INTEGER;
 import static io.trino.spi.type.RealType.REAL;
 import static io.trino.spi.type.SmallintType.SMALLINT;
 import static io.trino.spi.type.TinyintType.TINYINT;
+import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static io.trino.spi.type.VarcharType.UNBOUNDED_LENGTH;
 import static io.trino.type.DateTimes.formatTimestamp;
 import static io.trino.type.JsonType.JSON;
@@ -181,6 +184,7 @@ public final class JsonUtil
                 type instanceof DoubleType ||
                 type instanceof DecimalType ||
                 type instanceof VarcharType ||
+                type instanceof VarbinaryType ||
                 type instanceof JsonType ||
                 type instanceof TimestampType ||
                 type instanceof DateType) {
@@ -211,6 +215,7 @@ public final class JsonUtil
                 type instanceof DoubleType ||
                 type instanceof DecimalType ||
                 type instanceof VarcharType ||
+                type instanceof VarbinaryType ||
                 type instanceof JsonType) {
             return true;
         }
@@ -318,6 +323,9 @@ public final class JsonUtil
             }
             if (type instanceof VarcharType) {
                 return new VarcharJsonGeneratorWriter(type);
+            }
+            if (type instanceof VarbinaryType) {
+                return new VarbinaryJsonGeneratorWriter();
             }
             if (type instanceof JsonType) {
                 return new JsonJsonGeneratorWriter();
@@ -508,6 +516,23 @@ public final class JsonUtil
             else {
                 Slice value = type.getSlice(block, position);
                 jsonGenerator.writeString(value.toStringUtf8());
+            }
+        }
+    }
+
+    private static class VarbinaryJsonGeneratorWriter
+            implements JsonGeneratorWriter
+    {
+        @Override
+        public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int position)
+                throws IOException
+        {
+            if (block.isNull(position)) {
+                jsonGenerator.writeNull();
+            }
+            else {
+                Slice value = VARBINARY.getSlice(block, position);
+                jsonGenerator.writeBinary(value.byteArray(), value.byteArrayOffset(), value.length());
             }
         }
     }
@@ -904,6 +929,9 @@ public final class JsonUtil
             if (type instanceof VarcharType) {
                 return new VarcharBlockBuilderAppender(type);
             }
+            if (type instanceof VarbinaryType) {
+                return new VarbinaryBlockBuilderAppender();
+            }
             if (type instanceof JsonType) {
                 return (parser, blockBuilder) -> {
                     String json = JSON_MAPPED_UNORDERED.writeValueAsString(parser.readValueAsTree());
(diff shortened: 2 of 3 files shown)


core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryToJsonCast.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.io.IOException;
+import java.util.Base64;
+
+import static io.trino.spi.StandardErrorCode.INVALID_CAST_ARGUMENT;
+import static io.trino.spi.function.OperatorType.CAST;
+import static io.trino.spi.type.StandardTypes.JSON;
+import static io.trino.spi.type.StandardTypes.VARBINARY;
+
+public final class VarbinaryToJsonCast
```
findepi: The tests live in `TestVarbinaryFunctions`. Should the implementation live in `VarbinaryFunctions`?

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryToJsonCast.java
```diff
@@ -0,0 +1,65 @@
...
+    {
+        byte[] data = Base64.getEncoder().encode(binary.getBytes());
+        try (SliceOutput output = new DynamicSliceOutput(data.length + 2)) {
+            output.writeByte('"');
+            output.writeBytes(data);
+            output.writeByte('"');
+            return output.slice();
+        }
+        catch (IOException e) {
+            throw new TrinoException(INVALID_CAST_ARGUMENT, e.getMessage());
```
findepi: This isn't reachable under normal circumstances, so shouldn't raise INVALID_CAST_ARGUMENT error code.
It can use "generic internal error"

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryToJsonCast.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @SqlType(VARBINARY)
+    @ScalarOperator(CAST)
+    public static Slice castToBinary(@SqlType(JSON) Slice json)
+    {
+        if (json.length() < 2 || json.getByte(0) != '"' || json.getByte(json.length() - 1) != '"') {
+            throw new TrinoException(INVALID_CAST_ARGUMENT, "Expected quoted base64 String");
+        }
+        byte[] data = json.getBytes(1, json.length() - 2);
+        return Slices.wrappedBuffer(Base64.getDecoder().decode(data));
```
findepi: if base64 decoding fails here, what exception do we get?
Should be TrinoException with INVALID_CAST_ARGUMENT

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -166,7 +169,8 @@ public static boolean canCastToJson(Type type)
                 type instanceof VarcharType ||
                 type instanceof JsonType ||
                 type instanceof TimestampType ||
-                type instanceof DateType) {
+                type instanceof DateType ||
+                type instanceof VarbinaryType) {
```
findepi: nit: put VarbinaryType right after VarcharType

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -330,6 +334,9 @@ static JsonGeneratorWriter createJsonGeneratorWriter(Type type)
                 }
                 return new RowJsonGeneratorWriter(rowType, fieldWriters);
             }
+            if (type instanceof VarbinaryType) {
```
findepi: nit: handle VarbinaryType right after VarcharType

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -495,6 +502,23 @@ public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int positio
...
+        @Override
+        public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int position)
+                throws IOException
+        {
+            if (block.isNull(position)) {
+                jsonGenerator.writeNull();
+            }
+            else {
+                Slice value = VARBINARY.getSlice(block, position);
+                jsonGenerator.writeString(Base64.getEncoder().encodeToString(value.getBytes()));
```
findepi: @wendigo is there a more performant API for that?

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryToJsonCast.java
gertjanal: Moved

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -495,6 +503,23 @@ public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int positio
...
```
wendigo:
```
jsonGenerator.writeBinary(Base64Variants.MIME_NO_LINEFEEDS, value.byteArray(), value.byteArrayOffset(), value.length());
```

wendigo: This entirely avoids copying Slice to new byte[] array

ebyhr:
Does this PR fix #10814? Please update the PR description if so.

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
wendigo: I've pushed this change.

wendigo:
I've applied some changes to the serde

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -495,6 +504,23 @@ public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int positio
...
+        @Override
+        public void writeJsonValue(JsonGenerator jsonGenerator, Block block, int position)
+                throws IOException
+        {
+            if (block.isNull(position)) {
+                jsonGenerator.writeNull();
+            }
+            else {
+                Slice value = VARBINARY.getSlice(block, position);
+                jsonGenerator.writeBinary(Base64Variants.MIME, value.byteArray(), value.byteArrayOffset(), value.length());
```
gertjanal: @wendigo The MIME variant has line breaks, are you sure you want those?

gertjanal:
I added extra tests for varbinary arrays (from and to json)

gertjanal:
Fixed the checkstyle issue

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
```diff
@@ -501,4 +507,27 @@ public static Slice reverse(@SqlType("varbinary") Slice inputSlice)
         }
         return reverse;
     }
+
+    @SqlType(JSON)
+    @ScalarOperator(CAST)
+    public static Slice castToJson(@SqlType(VARBINARY) Slice binary)
+    {
+        int base64EncodedLength = 4 * ((binary.length() + 2) / 3);
```
chenjian2664: what's meaning of dividing 3 here?

gertjanal: Base64 encoding works by:
* Taking 3 bytes (24 bits) at a time.
* Splitting those 24 bits into 4 groups of 6 bits.
* Mapping each 6-bit group to one Base64 character.

Since:
* 1 byte = 8 bits
* 3 bytes = 24 bits
* 24 bits ÷ 6 bits per character = 4 characters

So:
Every 3 bytes → 4 Base64 characters

wendigo:
@ebyhr @raunaqmorarka PTAL

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
ebyhr: I think we should leave a code comment.

wendigo: @ebyhr this is well known formula - easy to find, personally - overkill

raunaqmorarka: Why does this need to be pre-calculated ?
Why not use the length of the output of `toBase64(binary)` instead ?

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
```diff
@@ -501,4 +507,27 @@ public static Slice reverse(@SqlType("varbinary") Slice inputSlice)
...
+    }
+
+    @SqlType(VARBINARY)
+    @ScalarOperator(CAST)
+    public static Slice castFromJson(@SqlType(JSON) Slice json)
+    {
+        if (json.length() < 2 || json.getByte(0) != '"' || json.getByte(json.length() - 1) != '"') {
+            throw new TrinoException(INVALID_CAST_ARGUMENT, "Expected quoted base64-encoded JSON string");
+        }
+        return fromBase64Varbinary(json.slice(1, json.length() - 2));
```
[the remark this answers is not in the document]
raunaqmorarka: +1

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -1124,6 +1135,28 @@ public void append(JsonParser parser, BlockBuilder blockBuilder)
...
+        public void append(JsonParser parser, BlockBuilder blockBuilder)
+                throws IOException
+        {
+            Slice result = currentTokenAsVarchar(parser);
+            if (result == null) {
+                blockBuilder.appendNull();
+            }
+            else {
+                try {
+                    VARBINARY.writeSlice(blockBuilder, Slices.wrappedHeapBuffer(Base64.getDecoder().decode(result.toByteBuffer())));
```
raunaqmorarka: can we re-use `io.trino.operator.scalar.VarbinaryFunctions#fromBase64Varchar` ?

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -1107,6 +1135,28 @@ public void append(JsonParser parser, BlockBuilder blockBuilder)
...
+                }
+                catch (IllegalArgumentException e) {
+                    throw new JsonCastException(e.getMessage());
```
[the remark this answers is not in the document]
gertjanal: Thanks Copilot, but the `IllegalArgumentException` message only contains information about the invalid base64 bytes, thrown by `.decode()`. There is no cause in `JsonCastException`, so the current impl is fine

core/trino-main/src/test/java/io/trino/operator/scalar/TestVarbinaryFunctions.java
```diff
@@ -990,6 +997,46 @@ public void testReverse()
                 .isEqualTo(sqlVarbinary("racecar"));
     }
 
+    @Test
+    public void testVarbinaryToJson()
+    {
+        assertThat(assertions.expression("CAST(binary as JSON)")
+                .binding("binary", "from_base64('" + encodeBase64("hello") + "')"))
+                .hasType(JSON)
+                .isEqualTo(format("\"%s\"", encodeBase64("hello")));
```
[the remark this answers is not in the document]
gertjanal: Code is fine
```java
    private static String encodeBase64(byte[] value)
    {
        return Base64.getEncoder().encodeToString(value);
    }

    private static String encodeBase64(String value)
    {
        return encodeBase64(value.getBytes(UTF_8));
    }
```

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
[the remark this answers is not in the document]
gertjanal: Added

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
gertjanal: Good idea, done

core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
gertjanal: Done

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -1124,6 +1152,28 @@ public void append(JsonParser parser, BlockBuilder blockBuilder)
...
+            if (result == null) {
+                blockBuilder.appendNull();
+            }
+            else {
+                try {
+                    VARBINARY.writeSlice(blockBuilder, fromBase64Varchar(result));
+                }
+                catch (IllegalArgumentException e) {
+                    throw new JsonCastException(e.getMessage());
+                }
```
raunaqmorarka: catch is likely unnecessary now

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
```diff
@@ -1124,6 +1152,28 @@ public void append(JsonParser parser, BlockBuilder blockBuilder)
...
+            }
```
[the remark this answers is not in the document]
gertjanal: Applied

core/trino-main/src/main/java/io/trino/util/JsonUtil.java
gertjanal: yes, removed

core/trino-main/src/test/java/io/trino/operator/scalar/TestVarbinaryFunctions.java
```diff
@@ -13,8 +13,12 @@
  */
 package io.trino.operator.scalar;
 
+import com.google.common.collect.ImmutableList;
 import com.google.common.io.BaseEncoding;
+import io.trino.spi.type.ArrayType;
+import io.trino.spi.type.SqlVarbinary;
 import io.trino.sql.query.QueryAssertions;
```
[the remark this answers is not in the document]
gertjanal: Done

martint:
This change still needs "syntax review". Why did it get merged?

wendigo:
@martint sorry, I missed that

wendigo:
Is there anything wrong with the syntax here?

martint:
There are a number of things to consider:
* Why base64? Why not hex-encoded, or SQL-binary-literal encoded?
* How does it relate to SQL/JSON data model and JSON type defined in the latest version of the SQL spec. In fact, there are changes we need to make to the JSON data type to make it compatible with the SQL spec. This could be a step in the wrong direction
* How does it interoperate with the JSON processing functions?
* etc.

PR #28248, opened 2026-02-11
Add return after abort when server is still initializing

TheEdgeOfRage:
## Description

From what I can tell, this method should not continue execution after calling `abortWith`, as all other places return after it.
Queries will still fail in this case, but the error that's reported in these cases was very misleading, since the response that's sent from here was not valid json:

```json
{
    "type": "com.fasterxml.jackson.core.JsonParseException",
    "message": "Unrecognized token 'io': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')\n at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 4]",
}
```

## Release notes

```markdown
## Section
```


TheEdgeOfRage:
Yeah, I'm not gonna be giving up my personal address and phone for merging a PR to an open source project lol

PR #28256, opened 2026-02-11
Use TypeId to represent Type in OutputColumnMetadata

wendigo:
## Description

TypeId is described as:

> Represents an opaque identifier for a Type than can be used for serialization or storage in external systems.

It should be used over Type.toString() representation

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/planner/InputExtractor.java
+++ b/core/trino-main/src/main/java/io/trino/sql/planner/InputExtractor.java
@@ -33,7 +33,6 @@ import io.trino.sql.planner.plan.PlanNodeId;
 import io.trino.sql.planner.plan.PlanVisitor;
 import io.trino.sql.planner.plan.TableScanNode;
 
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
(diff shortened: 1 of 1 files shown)


core/trino-spi/pom.xml
```diff
@@ -887,6 +887,21 @@
...
+                                    <old>parameter void io.trino.spi.eventlistener.OutputColumnMetadata::&lt;init&gt;(java.lang.String, ===java.lang.String===, java.util.Set&lt;io.trino.spi.eventlistener.ColumnDetail&gt;)</old>
+                                    <new>parameter void io.trino.spi.eventlistener.OutputColumnMetadata::&lt;init&gt;(java.lang.String, ===io.trino.spi.type.TypeId===, java.util.Set&lt;io.trino.spi.eventlistener.ColumnDetail&gt;)</new>
+                                    <parameterIndex>1</parameterIndex>
+                                    <justification>TypeId is used to carry over Type information</justification>
+                                </item>
+                                <item>
+                                    <ignore>true</ignore>
+                                    <code>java.method.returnTypeChanged</code>
+                                    <old>method java.lang.String io.trino.spi.eventlistener.OutputColumnMetadata::getColumnType()</old>
+                                    <new>method io.trino.spi.type.TypeId io.trino.spi.eventlistener.OutputColumnMetadata::getColumnType()</new>
```
martint, line 902: TypeId is for internal use, not meant to be exposed to the event listeners. Today, the type id happens to match the SQL type representation, but that’s going to change soon.

martint (changes requested):

core/trino-spi/pom.xml
wendigo: Type.toString is even worse. What's the alternative?

martint: For user visible representations, Type.getDisplayName()

martint: I’m working on a bunch of changes related to Type, TypeId, TypeSignature, so try to keep any changes in those areas well contained.

Issue #9142, opened 2021-09-07
Support LZ4 compression with native Parquet writer


findepi:
per https://trino.io/blog/2020/01/01/2019-summary.html

> Support for LZ4 and Zstd compression in Parquet and ORC. LZ4 is currently the recommended algorithm for fast, lightweight compression, and Zstd otherwise.

let's support LZ4 with the native Parquet writer

currently it fails:

```
Caused by: java.lang.RuntimeException: Unsupported codec: LZ4
	at io.trino.parquet.writer.ParquetCompressor.getCompressor(ParquetCompressor.java:56)
	at io.trino.parquet.writer.PrimitiveColumnWriter.<init>(PrimitiveColumnWriter.java:101)
	at io.trino.parquet.writer.ParquetWriters$WriteBuilder.primitive(ParquetWriters.java:147)
	at io.trino.parquet.writer.ParquetWriters$WriteBuilder.primitive(ParquetWriters.java:78)
	at io.trino.parquet.writer.ParquetTypeVisitor.visit(ParquetTypeVisitor.java:42)
	at io.trino.parquet.writer.ParquetTypeVisitor.visitField(ParquetTypeVisitor.java:118)
	at io.trino.parquet.writer.ParquetTypeVisitor.visitFields(ParquetTypeVisitor.java:129)
	at io.trino.parquet.writer.ParquetTypeVisitor.visit(ParquetTypeVisitor.java:39)
	at io.trino.parquet.writer.ParquetWriters.getColumnWriters(ParquetWriters.java:74)
	at io.trino.parquet.writer.ParquetWriter.<init>(ParquetWriter.java:96)
	at io.trino.plugin.hive.parquet.ParquetFileWriter.<init>(ParquetFileWriter.java:64)
	at io.trino.plugin.hive.parquet.ParquetFileWriterFactory.createFileWriter(ParquetFileWriterFactory.java:119)
	at io.trino.plugin.hive.HiveWriterFactory.createWriter(HiveWriterFactory.java:462)
	at io.trino.plugin.hive.HivePageSink.getWriterIndexes(HivePageSink.java:377)
	at io.trino.plugin.hive.HivePageSink.writePage(HivePageSink.java:277)
	at io.trino.plugin.hive.HivePageSink.doAppend(HivePageSink.java:272)
	at io.trino.plugin.hive.HivePageSink.lambda$appendPage$2(HivePageSink.java:258)
[...]

PR #28259, opened 2026-02-11
Fix compression_codec property value issues

### LINKED ISSUE #28291: [Iceberg] Overvalidation of compression codec table property in read use-cases
Splitting [the discussion I started here](https://github.com/trinodb/trino/pull/28259#issue-3929019526) into 2 different issues. Here's the first.

A customer of ours found that when they upgraded from 476 to 477 that they could no longer read (spark-compressed) data from their table. I've reproduced their problem like so ...

<details>
<summary>My product test environment reproduction ... </summary>

#### Before
```
testing/bin/ptl env up --environment singlenode-spark-iceberg --server-package trino-server-476.tar.gz
```
Where I've downloaded the [server tarball here](https://trino.io/docs/476/installation/deployment.html#installing-trino).

Then connecting to the spark console ...
```
docker exec -it ptl-spark spark-sql
spark-sql (default)> CREATE TABLE my_table (a STRING) USING iceberg TBLPROPERTIES('write.format.default'='PARQUET', 'write.parquet.compression-codec'='lz4');
```

Reading the (empty table) with trino ...

```
java -jar trino-cli.jar
trino> SELECT * FROM iceberg.default.my_table;
 a
---
(0 rows)
```

#### After
Same steps but with 477 server package.

```
java -jar trino-cli.jar
trino> SELECT * FROM iceberg.default.my_table;
Query 20260213_161439_00000_bkqf2 failed: Compression codec LZ4 not supported for Parquet
```

#### After
```
testing/bin/ptl env up --environment singlenode-spark-iceberg --without-trino --server-package ~/Downloads/trino-server-479.tar.gz
```

</details>

I believe [this pull request to add support for setting compression codec table properties](https://github.com/trinodb/trino/pull/25755) applies [validation to loaded table metadata](https://github.com/trinodb/trino/pull/25755/changes#diff-2af3e19a6b656640a7d0bb73114ef224953a2efa04e569b1fe4da953b2cc6d15R324). And this call to load metadata applies to both read and write use-cases. And so we have inadvertently broke this specific _read-only_ use-case.

### LINKED ISSUE #28293: [Iceberg] Trino uses invalid compression_codec table property mappings.
Splitting [this PR discussion](https://github.com/trinodb/trino/pull/28259#issue-3929019526) into 2 different issues because we might want to fix these separately.

I found that trino uses improper mappings with the compression_codec table property ([improper because they are out of spec](https://iceberg.apache.org/docs/latest/configuration/#write-properties)). This has two effects, we fail to read some spark-set compression codec table properties, and we write compression codec table properties that spark can't read.

<details>
<summary>My product test reproduction ...</summary>

```
testing/bin/ptl env up --environment singlenode-spark-iceberg
```

In spark ...
```
docker exec -it ptl-spark spark-sql
spark-sql (default)> CREATE TABLE uncompressed_parquet (a string) USING ICEBERG TBLPROPERTIES('write.format.default'='parquet','write.parquet.compression-codec'='uncompressed');
```

In trino ...
```
java -jar trino-cli.jar
trino> SELECT * FROM iceberg.default.uncompressed_parquet;
Query 20260213_162840_00000_afgbn failed: Compression codec uncompressed is unsupported.
trino> CREATE TABLE iceberg.default.none_parquet (a VARCHAR) WITH (format='parquet', compression_codec='NONE');
trino> INSERT INTO iceberg.default.none_parquet VALUES ('Hello World!');
INSERT: 1 row
```

But "uncompressed" is the correct string. Instead trino writes "NONE" to table properties.

Back in spark ...
```
spark-sql (default)> INSERT INTO uncompressed_parquet VALUES 'Hello World!';
Time taken: 2.812 seconds
spark-sql (default)> INSERT INTO none_parquet VALUES 'Hello World!';
Exception in task 0.0 in stage 1.0 (TID 1)
java.lang.IllegalArgumentException: Unsupported compression codec: NONE
```

Both query engines mean the same thing as evidenced by what ends up in the parquet column metadata ...
```
spark-sql (default)> SELECT file_path FROM spark_catalog.default.uncompressed_parquet.files;
hdfs://hadoop-master:9000/user/hive/warehouse/uncompressed_parquet/data/00000-0-61086db6-4258-45c7-8397-ccab60c52340-0-00001.parquet
spark-sql (default)> SELECT file_path FROM spark_catalog.default.none_parquet.files;
hdfs://hadoop-master:9000/user/hive/warehouse/uncompressed_parquet/data/00000-0-61086db6-4258-45c7-8397-ccab60c52340-0-00001.parquet
hdfs://hadoop-master:9000/user/hive/warehouse/none_parquet-a5b7967d6b884423a49a01336f49e039/data/20260213_163245_00004_afgbn-eea98576-fa9a-4b5b-8421-66324ed5b804.parquet
```

```
docker exec -it ptl-hadoop-master hadoop fs -get hdfs://hadoop-master:9000/user/hive/warehouse/uncompressed_parquet/data/00000-0-61086db6-4258-45c7-8397-ccab60c52340-0-00001.parquet /tmp/spark.parquet
docker cp ptl-hadoop-master:/tmp/spark.parquet /tmp/spark.parquet
docker exec -it ptl-hadoop-master hadoop fs -get hdfs://hadoop-master:9000/user/hive/warehouse/none_parquet-a5b7967d6b884423a49a01336f49e039/data/20260213_163245_00004_afgbn-eea98576-fa9a-4b5b-8421-66324ed5b804.parquet /tmp/trino.parquet
docker cp ptl-hadoop-master:/tmp/trino.parquet /tmp/trino.parquet
```

```
D SELECT path_in_schema, compression FROM parquet_metadata('/tmp/trino.parquet');
┌────────────────┬──────────────┐
│ path_in_schema │ compression  │
│    varchar     │   varchar    │
├────────────────┼──────────────┤
│ a              │ UNCOMPRESSED │
└────────────────┴──────────────┘
```

</details>

One example, we translate properties `format='parquet',compression_codec='NONE'` into iceberg table properties key `write.format.default` value `parquet` and key `write.parquet.compression-codec` value `NONE` (this needs to be `uncompressed` according to the spec and the spark implementation).

Math-ias:
## Description
We found that after upgrading from 476 to 477 LZ4 compressed tables were no longer readable. We've found two issues with this https://github.com/trinodb/trino/issues/28293 and https://github.com/trinodb/trino/issues/28291.

[This pull request to add support for setting compression codec table properties](https://github.com/trinodb/trino/pull/25755) did two things.

It applied [validation to loaded table metadata](https://github.com/trinodb/trino/pull/25755/changes#diff-2af3e19a6b656640a7d0bb73114ef224953a2efa04e569b1fe4da953b2cc6d15R324). And this call to load metadata applies to both read and write use-cases. And so we have inadvertently broke this specific _read-only_ use-case.

Second [it removed this code to translate compression codec property values into the specific values the iceberg specification requires](https://github.com/trinodb/trino/pull/25755/changes#diff-2af3e19a6b656640a7d0bb73114ef224953a2efa04e569b1fe4da953b2cc6d15L946). These values mostly resemble HiveCompressionCodec but have some key differences (e.g we call no compression `NONE` and in parquet and avro these are called `uncompressed`).

[Iceberg has table properties to set the default(?) compression codec to use for the default(?) data file format](https://iceberg.apache.org/docs/latest/configuration/#write-properties). We map one property `compression_codec` to these 3 (parquet, avro, orc) flavors of properties.

- fixes #28291
- fixes #28293 

## Release notes

```markdown
## Iceberg
* Fix failures when reading from tables with `write.parquet.compression-codec` property set to `LZ4`. ({issue}`28291`)
* Fix value of compression-codec table property written by Trino to be compliant with iceberg spec. ({issue}`28293`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2793,9 +2793,39 @@ public class IcebergMetadata
 
         validateCompression(newFileFormat, compressionCodec);
 
-        compressionCodec.ifPresent(hiveCompressionCodec -> newCompressionProperties.put(getCompressionPropertyName(newFileFormat), hiveCompressionCodec.name()));
+        compressionCodec.ifPresent(hiveCompressionCodec ->
+                newCompressionProperties.put(
+                        getCompressionPropertyName(newFileFormat),
+                        toCompressionCodecTableProperty(newFileFormat, hiveCompressionCodec)));
 
         return newCompressionProperties.buildOrThrow();
+    }
+
+    static String toCompressionCodecTableProperty(IcebergFileFormat fileFormat, HiveCompressionCodec compressionCodec)
+    {
+        return switch (fileFormat) {
+            case PARQUET -> switch (compressionCodec) {
+                case NONE -> "uncompressed";
+                case SNAPPY -> "snappy";
+                case LZ4 -> "lz4";
+                case ZSTD -> "zstd";
+                case GZIP -> "gzip";
+            };
+            case ORC -> switch (compressionCodec) {
+                case NONE -> "none";
+                case SNAPPY -> "snappy";
+                case LZ4 -> "lz4";
+                case ZSTD -> "zstd";
+                case GZIP -> "zlib";
+            };
+            case AVRO -> switch (compressionCodec) {
+                case NONE -> "uncompressed";
+                case SNAPPY -> "snappy";
+                case LZ4 -> throw new TrinoException(NOT_SUPPORTED, "Compression codec LZ4 not supported for Avro");
+                case ZSTD -> "zstd";
+                case GZIP -> "gzip";
+            };
+        };
     }
 
     private static void updatePartitioning(Table icebergTable, Transaction transaction, List<String> partitionColumns)
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
@@ -115,6 +115,11 @@ import static io.airlift.slice.Slices.utf8Slice;
 import static io.trino.metastore.Table.TABLE_COMMENT;
 import static io.trino.parquet.writer.ParquetWriter.SUPPORTED_BLOOM_FILTER_TYPES;
 import static io.trino.plugin.base.io.ByteBuffers.getWrappedBytes;
+import static io.trino.plugin.hive.HiveCompressionCodec.GZIP;
+import static io.trino.plugin.hive.HiveCompressionCodec.LZ4;
+import static io.trino.plugin.hive.HiveCompressionCodec.NONE;
+import static io.trino.plugin.hive.HiveCompressionCodec.SNAPPY;
+import static io.trino.plugin.hive.HiveCompressionCodec.ZSTD;
 import static io.trino.plugin.iceberg.ColumnIdentity.createColumnIdentity;
 import static io.trino.plugin.iceberg.IcebergColumnHandle.fileModifiedTimeColumnHandle;
 import static io.trino.plugin.iceberg.IcebergColumnHandle.fileModifiedTimeColumnMetadata;
@@ -1023,16 +1028,43 @@ public final class IcebergUtil
     public static Optional<HiveCompressionCodec> getHiveCompressionCodec(IcebergFileFormat icebergFileFormat, Map<String, String> storageProperties)
     {
         String compressionProperty = getCompressionPropertyName(icebergFileFormat);
-
         return Optional.ofNullable(storageProperties.get(compressionProperty))
                 .filter(value -> !value.isEmpty())
                 .map(value -> {
+                    String upperCaseValue = value.toUpperCase(ENGLISH);
+                    // https://github.com/trinodb/trino/issues/28293
+                    // Not iceberg specification valid,
+                    // but Trino made these so we must read for backwards compatibility.
                     try {
-                        return HiveCompressionCodec.valueOf(value.toUpperCase(ENGLISH));
+                        return HiveCompressionCodec.valueOf(upperCaseValue);
                     }
-                    catch (IllegalArgumentException e) {
-                        throw new TrinoException(INVALID_TABLE_PROPERTY,
-                                format("Compression codec %s is unsupported.", value));
+                    catch (IllegalArgumentException _) {
+                        return switch (icebergFileFormat) {
+                            // Based on https://iceberg.apache.org/docs/1.10.1/docs/configuration/#write-properties.
+                            case PARQUET -> switch (upperCaseValue) {
+                                case "UNCOMPRESSED" -> NONE;
+                                case "SNAPPY" -> SNAPPY;
+                                case "ZSTD" -> ZSTD;
+                                case "LZ4" -> LZ4;
+                                case "GZIP" -> GZIP;
+                                default -> throw new TrinoException(INVALID_TABLE_PROPERTY, format("Unsupported compression codec: %s", upperCaseValue));
+                            };
+                            case AVRO -> switch (upperCaseValue) {
+                                case "UNCOMPRESSED" -> NONE;
+                                case "SNAPPY" -> SNAPPY;
+                                case "ZSTD" -> ZSTD;
+                                case "GZIP" -> GZIP;
+                                default -> throw new TrinoException(INVALID_TABLE_PROPERTY, format("Unsupported compression codec: %s", upperCaseValue));
+                            };
+                            case ORC -> switch (upperCaseValue) {
+                                case "ZLIB" -> GZIP;
+                                case "ZSTD" -> ZSTD;
+                                case "LZ4" -> LZ4;
+                                case "SNAPPY" -> SNAPPY;
+                                case "NONE" -> NONE;
+                                default -> throw new TrinoException(INVALID_TABLE_PROPERTY, format("Unsupported compression codec: %s", upperCaseValue));
+                            };
+                        };
                     }
                 });
     }
--- a/testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
+++ b/testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
@@ -1727,25 +1727,7 @@ public class TestIcebergSparkCompatibility
 
         assertThat(onSpark().executeQuery("SELECT * FROM " + sparkTableName))
                 .containsOnly(sparkRows);
-
-        String selectAll = "SELECT * FROM " + trinoTableName;
-        // https://github.com/trinodb/trino/issues/28293
-        if (storageFormat == StorageFormat.AVRO && compressionCodec == CompressionCodec.NONE) {
-            assertQueryFailure(() -> onTrino().executeQuery(selectAll))
-                    .hasMessageMatching("\\QQuery failed (#\\E\\S+\\Q): Compression codec uncompressed is unsupported.");
-            return;
-        }
-        if (storageFormat == StorageFormat.ORC && compressionCodec == CompressionCodec.GZIP) {
-            assertQueryFailure(() -> onTrino().executeQuery(selectAll))
-                    .hasMessageMatching("\\QQuery failed (#\\E\\S+\\Q): Compression codec zlib is unsupported.");
-            return;
-        }
-        if (storageFormat == StorageFormat.PARQUET && compressionCodec == CompressionCodec.NONE) {
-            assertQueryFailure(() -> onTrino().executeQuery(selectAll))
-                    .hasMessageMatching("\\QQuery failed (#\\E\\S+\\Q): Compression codec uncompressed is unsupported.");
-            return;
-        }
-        assertThat(onTrino().executeQuery(selectAll))
+        assertThat(onTrino().executeQuery("SELECT * FROM " + trinoTableName))
                 .containsOnly(sparkRows);
 
         List<Row> trinoRows = IntStream.range(0, 13)
@@ -1812,28 +1794,10 @@ public class TestIcebergSparkCompatibility
         List<Row> sparkRows = IntStream.range(0, 13)
                 .mapToObj(i -> row("spark" + i, i))
                 .collect(toImmutableList());
-        String insertQuery = "INSERT INTO " + sparkTableName + " VALUES " +
+        onSpark().executeQuery("INSERT INTO " + sparkTableName + " VALUES " +
                 sparkRows.stream()
                         .map(row -> format("('%s', %s)", row.getValues().get(0), row.getValues().get(1)))
-                        .collect(Collectors.joining(", "));
-
-        // https://github.com/trinodb/trino/issues/28293
-        if (storageFormat == StorageFormat.AVRO && compressionCodec == CompressionCodec.NONE) {
-            assertQueryFailure(() -> onSpark().executeQuery(insertQuery))
-                    .hasMessageContaining("Unsupported compression codec: NONE");
-           return;
-        }
-        if (storageFormat == StorageFormat.PARQUET && compressionCodec == CompressionCodec.NONE) {
-            assertQueryFailure(() -> onSpark().executeQuery(insertQuery))
-                    .hasMessageContaining("Unsupported compression codec: NONE");
-            return;
-        }
-        if (storageFormat == StorageFormat.ORC && compressionCodec == CompressionCodec.GZIP) {
-            assertQueryFailure(() -> onSpark().executeQuery(insertQuery))
-                    .hasMessageContaining("Unsupported compression codec: GZIP");
-            return;
-        }
-        onSpark().executeQuery(insertQuery);
+                        .collect(Collectors.joining(", ")));
 
         assertThat(onTrino().executeQuery("SELECT * FROM %s WHERE a LIKE 'spark%%'".formatted(trinoTableName)))
                 .containsOnly(sparkRows);
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -130,11 +130,11 @@ import static io.trino.SystemSessionProperties.TASK_MAX_WRITER_COUNT;
 import static io.trino.SystemSessionProperties.TASK_MIN_WRITER_COUNT;
 import static io.trino.SystemSessionProperties.TASK_SCALE_WRITERS_ENABLED;
 import static io.trino.SystemSessionProperties.USE_PREFERRED_WRITE_PARTITIONING;
-import static io.trino.hive.formats.compression.CompressionKind.ZSTD;
 import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_INVALID_METADATA;
 import static io.trino.plugin.iceberg.IcebergFileFormat.AVRO;
 import static io.trino.plugin.iceberg.IcebergFileFormat.ORC;
 import static io.trino.plugin.iceberg.IcebergFileFormat.PARQUET;
+import static io.trino.plugin.iceberg.IcebergMetadata.toCompressionCodecTableProperty;
 import static io.trino.plugin.iceberg.IcebergQueryRunner.ICEBERG_CATALOG;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.BUCKET_EXECUTION_ENABLED;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.COLLECT_EXTENDED_STATISTICS_ON_WRITE;
@@ -6943,7 +6943,7 @@ public abstract class BaseIcebergConnectorTest
 
                 assertThat(getTableProperties(tableName))
                         .containsEntry(DEFAULT_FILE_FORMAT, format.toString())
-                        .containsEntry(compressionProperty, compressionCodec.name());
+                        .containsEntry(compressionProperty, toCompressionCodecTableProperty(format, compressionCodec));
 
                 assertThat(query("SELECT * FROM " + tableName)).matches("SELECT * FROM nation");
                 assertThat(query(format("SELECT count(*) FROM \"%s$files\" WHERE file_path LIKE '%%.%s'", tableName, format.name().toLowerCase(ENGLISH))))
@@ -6988,13 +6988,13 @@ public abstract class BaseIcebergConnectorTest
                     .matches("SELECT count(*) FROM nation WHERE nationkey < 10");
             assertThat(getTableProperties(tableName))
                     .containsEntry(DEFAULT_FILE_FORMAT, format.toString())
-                    .containsEntry(compressionProperty, initialCompressionCodec.name());
+                    .containsEntry(compressionProperty, toCompressionCodecTableProperty(format, initialCompressionCodec));
 
             if (isCompressionCodecSupportedForFormat(format, compressionCodec)) {
                 assertUpdate(format("ALTER TABLE %s SET PROPERTIES compression_codec = '%s'", tableName, newCompressionCodec));
                 assertThat(getTableProperties(tableName))
                         .containsEntry(DEFAULT_FILE_FORMAT, format.toString())
-                        .containsEntry(compressionProperty, newCompressionCodec);
+                        .containsEntry(compressionProperty, toCompressionCodecTableProperty(format, compressionCodec));
                 assertUpdate(
                         "INSERT INTO " + tableName + " SELECT * FROM nation WHERE nationkey >= 10",
                         "SELECT count(*) FROM nation WHERE nationkey >= 10");
@@ -7006,7 +7006,7 @@ public abstract class BaseIcebergConnectorTest
             else {
                 assertQueryFails(
                         format("ALTER TABLE %s SET PROPERTIES compression_codec = '%s'", tableName, newCompressionCodec),
-                        "Compression codec LZ4 not supported for .*");
+                        "Compression codec %s not supported for .*".formatted(newCompressionCodec));
             }
             assertUpdate("DROP TABLE " + tableName);
         }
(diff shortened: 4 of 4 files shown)


Math-ias:
(Force pushed commit message changes)

Math-ias:
(force pushed to test positive cases of writing into trino tables from spark)

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1697,6 +1697,83 @@ public void testTrinoSparkConcurrentInsert()
...
+            {"orc", "write.orc.compression-codec", "zstd"},
+            {"orc", "write.orc.compression-codec", "lz4"},
+            {"orc", "write.orc.compression-codec", "lzo"},
+            {"orc", "write.orc.compression-codec", "zlib"},
+            {"orc", "write.orc.compression-codec", "snappy"},
+            {"orc", "write.orc.compression-codec", "none"}
+    };
+
     @Test(groups = {ICEBERG, PROFILE_SPECIFIC_TESTS}, dataProvider = "storageFormatsAndCompressionCodecs")
     public void testTrinoReadingSparkCompressedData(StorageFormat storageFormat, String compressionCodec)
```
raunaqmorarka: Is there any point in keeping testTrinoReadingSparkCompressedData if the SET statements used there are doing nothing ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -334,8 +334,6 @@ public static Map<String, Object> getIcebergTableProperties(BaseTable icebergTab
 
         Optional<HiveCompressionCodec> compressionCodec = getHiveCompressionCodec(fileFormat, icebergTable.properties());
 
-        validateCompression(fileFormat, compressionCodec);
```
raunaqmorarka, line 337: The validation should only prevent creation of tables that we don't support writing to, which i think is still happening elsewhere, so this is fine to drop.
If the compression codec is unsupported during read, the file format reader will throw an error about it anyway

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1794,6 +1871,24 @@ public void testSparkReadingTrinoCompressedData(StorageFormat storageFormat, Str
         assertThat(onSpark().executeQuery("SELECT * FROM " + sparkTableName))
                 .containsOnly(expected);
 
+        // Known bugs ...
```
raunaqmorarka: I think some of this is a regression from https://github.com/trinodb/trino/pull/25755
In particular we had
```
switch (fileFormat) {
            case PARQUET -> propertiesBuilder.put(PARQUET_COMPRESSION, toParquetCompressionCodecTableProperty(compressionCodec));
            case ORC -> propertiesBuilder.put(ORC_COMPRESSION, compressionCodec.getOrcCompressionKind().name().toLowerCase(Locale.ENGLISH));
            case AVRO -> propertiesBuilder.put(AVRO_COMPRESSION, toAvroCompressionCodecTableProperty(compressionCodec));
}

private static String toAvroCompressionCodecTableProperty(HiveCompressionCodec compressionCodec)
    {
        return switch (compressionCodec) {
            case NONE -> "uncompressed";
            case SNAPPY -> "snappy";
            case LZ4 -> throw new TrinoException(NOT_SUPPORTED, "Compression codec LZ4 not supported for Avro");
            case ZSTD -> "zstd";
            case GZIP -> "gzip";
        };
    }

    private static String toParquetCompressionCodecTableProperty(HiveCompressionCodec compressionCodec)
    {
        return switch (compressionCodec) {
            case NONE -> "uncompressed";
            case SNAPPY -> "snappy";
            case LZ4 -> "lz4";
            case ZSTD -> "zstd";
            case GZIP -> "gzip";
        };
    }
```
in `io.trino.plugin.iceberg.IcebergUtil#getIcebergTableProperties`
I think we should restore that part.
The current code is assuming that `HiveCompressionOption.valueOf(hiveCompressionCodec.name()))` is valid property name for iceberg, but its not

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: I suspect, but I will need to go into the debugger to confirm, that these `SET` statements change the compression codec of the data files, just not the _default_ compression codec of the iceberg table.

JohnFitzpatrick44:
I was curious why it seems that we don't support LZ4 for parquet, when it seems to be compatible in other engines. Looks like [support for LZ4 is being deprecated](https://github.com/apache/parquet-format/blob/master/Compression.md#lz4) in favor of LZ4_RAW.

Looks like LZ4_RAW support is on Trino's radar - https://github.com/trinodb/trino/issues/9142

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1697,6 +1697,83 @@ public void testTrinoSparkConcurrentInsert()
...
+    }
+
+    @DataProvider
+    public String[][] getIcebergTablePropertyList()
+    {
+        // From https://iceberg.apache.org/docs/latest/configuration/#write-properties on 2026-02-011
+        return ICEBERG_FORMATS_COMPRESSION_PROPERTIES_AND_VALUES;
+    }
+
+    private static final String[][] ICEBERG_FORMATS_COMPRESSION_PROPERTIES_AND_VALUES = new String[][] {
```
raphaelsolarski: nit: can we inline it?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1697,6 +1697,78 @@ public void testTrinoSparkConcurrentInsert()
...
+            {"avro", "write.avro.compression-codec", "gzip"},
+            {"avro", "write.avro.compression-codec", "zstd"},
+            {"avro", "write.avro.compression-codec", "snappy"},
+            {"avro", "write.avro.compression-codec", "uncompressed"},
+            {"parquet", "write.parquet.compression-codec", "zstd"},
+            // Classes not included in spark environment.
+            // {"parquet", "write.parquet.compression-codec", "brotli"},
```
raphaelsolarski: are we leaving it commented?

Math-ias: Wasn't sure how exactly to demonstrate completeness with iceberg specification. The more proper way is probably to update the spark test environment so it has the necessary brotli compression codec (so we can assert on its failure).

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: More a TIL for me but [the parquet format supports setting compression codecs per column metadata](https://github.com/apache/parquet-format/blob/38818fa0e7efd54b535001a4448030a40619c2a3/src/main/thrift/parquet.thrift#L892). BUT I checked and ...
```
spark-sql (default)> SET spark.sql.parquet.compression.codec = lz4;
spark-sql (default)> CREATE TABLE t1 (a string, b bigint) USING ICEBERG TBLPROPERTIES('write.format.default'='parquet');
spark-sql (default)> INSERT INTO t1 VALUES ('hello world!', 1);
spark-sql (default)> SELECT file_path FROM spark_catalog.default.t1.files;
hdfs://hadoop-master:9000/user/hive/warehouse/t1/data/00000-0-902ab7ac-9cb6-4886-ba4f-d41dbad1b764-0-00001.parquet
```

I copied that file to my local computer. And unfortunately that SET statement doesn't appear to do anything.
```
duckdb
SELECT path_in_schema, compression FROM parquet_metadata('~/Downloads/data.parquet');
┌────────────────┬─────────────┐
│ path_in_schema │ compression │
│    varchar     │   varchar   │
├────────────────┼─────────────┤
│ a              │ ZSTD        │
│ b              │ ZSTD        │
└────────────────┴─────────────┘
```
I'll adjust the PR.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: Re-applied but with a static bimap. Let me know if we want the changes to resemble these original functions.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: Now inlined and generated within the function.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: No longer using brotli or lzo examples.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1698,123 +1698,211 @@ public void testTrinoSparkConcurrentInsert()
     }
 
     @Test(groups = {ICEBERG, PROFILE_SPECIFIC_TESTS}, dataProvider = "storageFormatsAndCompressionCodecs")
-    public void testTrinoReadingSparkCompressedData(StorageFormat storageFormat, String compressionCodec)
+    public void testSparkTrinoCompressedDataExchange(StorageFormat storageFormat, CompressionCodec compressionCodec)
```
raunaqmorarka: I think the test method names should retain clarity about whether its Trino reading data compressed by Spark or vice versa. The new name is less clear about what is being tested.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
raunaqmorarka: I think separate methods are easier to read than the BiMap

Math-ias:
First force push about making compression codec lookup case insensitive. I think it should address many but not all of the test failures here. Will make subsequent force-pushes to address test failures and compression-codec lookup to use functions instead of the bimap.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1772,49 +1782,127 @@ public void testSparkReadingTrinoCompressedData(StorageFormat storageFormat, Str
...
         return Stream.of(StorageFormat.values())
-                .flatMap(storageFormat -> compressionCodecs.stream()
-                        .map(compressionCodec -> new Object[] {storageFormat, compressionCodec}))
+                .flatMap(storageFormat -> Stream.of(CompressionCodec.values())
+                        .flatMap(compressionCodec -> {
+                            if (storageFormat == StorageFormat.AVRO && compressionCodec == CompressionCodec.LZ4) {
+                                return Stream.empty();
+                            }
+                            else {
+                                Object[] array = new Object[] {storageFormat, compressionCodec};
```
marcinsbd: nit: inline array?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1698,72 +1698,82 @@ public void testTrinoSparkConcurrentInsert()
...
                                 .map(row -> format("('%s', %s)", row.getValues().get(0), row.getValues().get(1)))
                                 .collect(Collectors.joining(", ")));
+
         assertThat(onSpark().executeQuery("SELECT * FROM " + sparkTableName))
-                .containsOnly(rows);
-        assertThat(onTrino().executeQuery("SELECT * FROM " + trinoTableName))
-                .containsOnly(rows);
+                .containsOnly(sparkRows);
+
+        String firstSelect = "SELECT * FROM " + trinoTableName;
```
marcinsbd: nit: selectAll ?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1772,49 +1782,127 @@ public void testSparkReadingTrinoCompressedData(StorageFormat storageFormat, Str
...
-                "ZSTD",
-                "GZIP");
+        return switch(storageFormat) {
+            case AVRO -> "write.avro.compression-codec";
+            case PARQUET -> "write.parquet.compression-codec";
+            case ORC -> "write.orc.compression-codec";
+        };
+    }
+
+    public enum CompressionCodec {
```
marcinsbd, line 1830: nit: should we encapsulate this one?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1772,49 +1782,127 @@ public void testSparkReadingTrinoCompressedData(StorageFormat storageFormat, Str
...
+        GZIP,
+        ZSTD,
+        SNAPPY,
+        LZ4,
+        NONE,
+    }
+
+    String getIcebergCompressionCodecName(StorageFormat storageFormat, CompressionCodec compressionCodec)
```
marcinsbd: private ?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -1772,49 +1782,127 @@ public void testSparkReadingTrinoCompressedData(StorageFormat storageFormat, Str
...
+                            else {
+                                Object[] array = new Object[] {storageFormat, compressionCodec};
+                                return Stream.ofNullable(array);
+                            }
+                        }))
                 .toArray(Object[][]::new);
     }
 
-    private List<String> compressionCodecs()
+    String getIcebergCompressionCodecTableProperty(StorageFormat storageFormat)
```
marcinsbd: private?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -1023,16 +1030,43 @@ public static void verifyExtraProperties(Set<String> basePropertyKeys, Map<Strin
...
     {
         String compressionProperty = getCompressionPropertyName(icebergFileFormat);
-
         return Optional.ofNullable(storageProperties.get(compressionProperty))
                 .filter(value -> !value.isEmpty())
                 .map(value -> {
+                    String upperCaseValue = value.toUpperCase(ENGLISH);
+                    // https://github.com/trinodb/trino/issues/28293
+                    // Not iceberg specification valid,
+                    // but trino made these so we must read for backwards compatibility.
```
marcinsbd: nit: Trino

marcinsbd (approved):
LGTM,  few nits

Math-ias:
(force push to address failed maven checks, whitespace and imports changes)

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: Now named it `testSparkConfiguredCompressionCodecOnSparkAndTrino`. Just don't want to lose the intent that I'm writing to it from both spark and trino (important for surfacing these bugs).

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: In the next force push I'll simplify this and just filter at the bottom of the stream for the AVRO,LZ4 combo. I was experimenting with ways to use `Stream.of` here with an array object and not have java interpret me as wanting varargs.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: Sure I like that name more. Applied in next force-push.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: What kind of encapsulating class are you imagining? The storage format would encapsulate it's own compression codecs or something?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: Sure I've applied this to both `getIcebergCompressionCodecTableProperty` and `getIcebergCompressionCodecName` in the next force push.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
Math-ias: Applied in latest force push.

Math-ias:
(force pushed changes to address @marcinsbd 's review comments)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2792,11 +2791,45 @@ public static Map<String, String> calculateTableCompressionProperties(IcebergFil
...
+                        getCompressionPropertyName(newFileFormat),
+                        toCompressionCodecTableProperty(newFileFormat, hiveCompressionCodec)));
 
         return newCompressionProperties.buildOrThrow();
     }
 
+    static String toCompressionCodecTableProperty(IcebergFileFormat fileFormat, HiveCompressionCodec compressionCodec)
+    {
+        return switch (fileFormat) {
+            case PARQUET -> toAvroCompressionCodecTableProperty(compressionCodec);
```
SongChujun: Why ```toAvroCompressionCodecTableProperty``` is called on PARQUET format while  ```toParquetCompressionCodecTableProperty``` is called on AVRO format?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2791,11 +2791,45 @@ public static Map<String, String> calculateTableCompressionProperties(IcebergFil
...
+            case ORC -> compressionCodec.getOrcCompressionKind().name().toLowerCase(ENGLISH);
```
raunaqmorarka: Is relying on `io.trino.orc.metadata.CompressionKind` safe ?
I don't think there is a guarantee that a valid value in that enum is a valid value in iceberg table properties. It would be safer to do the mapping in similar way as avro/parquet, or look at how the iceberg library validates the string value and copy that.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
raunaqmorarka: @Math-ias  Is that a typo ?

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
marcinsbd: I meant make it private

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
Math-ias: Yes, this is a typo. I should swap these two or just do it inline.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
Math-ias: [We used to rely on toString'ing the orc compression kind](https://github.com/trinodb/trino/pull/25755/changes#diff-2af3e19a6b656640a7d0bb73114ef224953a2efa04e569b1fe4da953b2cc6d15L883) so I was restoring that. But I agree, we can do better with this change.

Looks like in the iceberg library [the "write.orc.compression-codec" is read here](https://github.com/apache/iceberg/blob/d06e6c22fecf00be81f2eb399f090fae320560a4/orc/src/main/java/org/apache/iceberg/orc/ORC.java#L301). And then that string value goes [into this method here](https://github.com/apache/iceberg/blob/d06e6c22fecf00be81f2eb399f090fae320560a4/orc/src/main/java/org/apache/iceberg/orc/ORC.java#L377) which uses `org.apache.orc.CompressionKind` (defined [here](https://github.com/apache/orc/blob/cd70240519fd7021ac5e1e6c3b1ac43ba26bf0e6/java/core/src/java/org/apache/orc/CompressionKind.java#L26)). That enum resembles what we have in `io.trino.orc.metadata.CompressionKind` (though we have some omissions like brotli).

Mapping strings to enum values and vice versa is much simpler, so I will do that instead.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: I'd agree. I don't want people importing this CompressionCodec elsewhere. I marked it public because IntelliJ warned me that this public test method ...

```
public void testSparkConfiguredCompressionCodecOnSparkAndTrino(StorageFormat storageFormat, CompressionCodec compressionCodec)
```

Was exposing my private enum. So I made it public like the `StorageFormat` enum. Going to resolve.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
Math-ias: https://github.com/trinodb/trino/pull/28259/changes/bfd1fcb5cf5c4b3da0af5dc2e56c7fa7ae3a4d42#diff-c6f2d27810f15140edca8709529de4d89bfb05cbb1d633c61be47d79f5744518R1731

Math-ias:
(force push to rebase against master)

PR #28260, opened 2026-02-12
Added trino-logger-event-listener

vjain143:
## Description

Add a QueryLogEventListener that emits structured per-query events to configured sinks (logs/HTTP/metrics).  
The listener captures query lifecycle events (submission, planning, execution, completion/failure) and forwards a compact, stable event payload suitable for auditing, observability, and downstream analytics.

Key changes:
- trino-main: add QueryLogEventListener implementation and lifecycle hook
- trino-spi: (if applicable) add SPI interfaces/types for event payload
- config: new configuration properties to enable/configure the listener and sinks
- tests: unit + integration coverage for happy path and failure cases

- Background: operators and observability tooling need a lightweight, stable per-query event stream for auditing and external monitoring without depending on system tables or the UI.
- Backward compatibility: opt-in by default (no behavioral change when listener is not enabled)
- Performance: low overhead when enabled; batching and async delivery used to avoid adding latency
- Risk: low — changes are isolated to event-listener wiring and an opt-in implementation
- Testing: unit tests added (files: [paths to unit tests]); integration test added (files: [paths to integration tests])

Manual test steps:
1. Enable the listener in etc/catalog or etc/config.properties:
   - querylog.listener.enabled=true
   - querylog.listener.sink=[log|http|noop]
2. Run a representative query that previously exercised planner/executor paths.
3. Verify an event is emitted containing: queryId, user, state, durationMs, error (if any), stats.

Reviewer guidance:
- Focus on correctness of the emitted payload (fields, nullability, stable names)
- Verify performance/async delivery (no blocking on query completion)
- Check configuration surface and docs
- Suggested reviewers: @trinodb/core-dev, maintainers of observability/instrumentation

## Release notes

Proposed release-note (short):
* Add QueryLogEventListener — optional per-query event emission for auditing and observability. (#{issuenumber})

Proposed release-note (detailed):
* Query logging — new QueryLogEventListener produces structured per-query events (submission, completion, failure) to configured sinks (log/HTTP). Opt-in via configuration; no migration required. (#{issuenumber})

### DIFF
--- a/plugin/trino-logger-event-listener/README.md
+++ b/plugin/trino-logger-event-listener/README.md
@@ -1,0 +1,375 @@
+# Trino Logger Event Listener
+
+This is a Trino plugin that logs query events to a file using Airlift Logger.
+
+## Configuration
+
+The Logger Event Listener can be configured through Trino properties. Add the following properties to your Trino configuration:
+
+### Basic Configuration
+
+```properties
+# Enable the query log event listener plugin
+event-listener.type=logger
+
+# Log query created events
+logger-event-listener.log-created=true
+
+# Log query completed events
+logger-event-listener.log-completed=true
+
+# Log query executed events
+logger-event-listener.log-executed=true
+
+# Path to the log file (optional, default: logger.log)
+logger-event-listener.log-file-path=/var/log/trino/logger.log
+```
+
+### Advanced Configuration: Field Control
+
+#### Exclude Sensitive Fields
+
+Exclude specific fields from being logged (their values will be replaced with null):
+
+```properties
+# Comma-separated list of fields to exclude
+# Example: payload,user,sourceCode,password
+logger-event-listener.excluded-fields=payload,user,sourceCode
+```
+
+#### Truncate Large Fields
+
+Control the maximum size of field values to prevent excessive log sizes:
+
+```properties
+# Maximum field size before truncation (default: 4KB)
+# Supported units: B, KB, MB, GB
+logger-event-listener.max-field-size=8KB
+```
+
+#### Selective Field Truncation (by Size)
+
+Truncate specific fields (like `query` and `stageInfo`) if they exceed a size limit:
+
+```properties
+# Comma-separated list of field names to truncate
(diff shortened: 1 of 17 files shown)


plugin/trino-querylog-event-listener/src/main/java/io/trino/plugin/eventlistener/querylog/QueryLogEventListenerConfig.java
```diff
@@ -0,0 +1,219 @@
...
+import java.util.Set;
+
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static io.airlift.units.DataSize.Unit.KILOBYTE;
+import static java.util.Objects.requireNonNull;
+
+public class QueryLogEventListenerConfig
+{
+    private final EnumSet<QueryLogEventType> loggedEvents = EnumSet.noneOf(QueryLogEventType.class);
+    private String logFilePath = "querylog.log";
```
martint: Where is this used?

plugin/trino-querylog-event-listener/src/main/java/io/trino/plugin/eventlistener/querylog/QueryLogEventListener.java
```diff
@@ -0,0 +1,126 @@
...
+ * Event listener that logs query events to a file using Airlift Logger.
+ * <p>
+ * The listener logs all configured event types (QUERY_CREATED, QUERY_COMPLETED, QUERY_EXECUTED)
+ * as JSON to the configured log file path.
+ * <p>
+ * Supports:
+ * - Excluding specific fields from log output
+ * - Truncating large field values to prevent excessive log sizes
+ */
+public class QueryLogEventListener
```
martint: LoggerEventListener would be a better name. "QueryLog" could be confused with this being a generic query logging listener. All implementations of EventListener are for logging queries.

vjain143: changing name to trino-logger-event-listener

PR #22632, opened 2024-07-09
Add support for reading from Unity catalog with Iceberg REST

ebyhr:
## Description

This PR supports only read operations because Iceberg REST interface in Unity catalog doesn't support creating tables yet. 

Note that Databricks Unity catalog provides HTTP interface, but it doesn’t return Iceberg UniForm tables. 

Example catalog config properties:
```properties
connector.name=iceberg
iceberg.catalog.type=rest
iceberg.rest-catalog.uri=https://dbc-12345678-9999.cloud.databricks.com/api/2.1/unity-catalog/iceberg
iceberg.security=read_only
iceberg.rest-catalog.security=OAUTH2
iceberg.rest-catalog.oauth2.token=***
iceberg.rest-catalog.parent-namespace=test_namespace
```

Fixes #22609

## Release notes

```markdown
# Iceberg
* Add support for reading from Unity catalog with REST catalog. ({issue}`22609`)
```

PR #18517, opened 2023-08-03
Accelerate system.metadata.table_comments with Iceberg Glue catalog

findepi:
With Glue, table listing already pulls a lot information about tables.
For Trino-managed Iceberg tables this is sufficient information to
answer `system.metadata.table_comments` queries, without having to fetch
Glue tables again, one-by-one. Trino-manged Iceberg tables keep table
comment up to date in Glue (along with additional information sufficient
to verify it's indeed up to date). The approach can be generalized to
Iceberg's own `GlueCatalog` later if the community is interested.

Builds on https://github.com/trinodb/trino/pull/18315

PR #18315, opened 2023-07-17
Cache Iceberg column types in Glue for faster access

findepi:
Make `information_schema.columns` queries faster by storing the
necessary directly in Glue so that loading Iceberg metadata from storage
is not need.

- Table comment is stored as `comment` Glue table parameter.

- Column Trino type is stored as Glue column type (for some cases) or as
  `trino_type` column table parameter. This is because, following
  Iceberg own's `org.apache.iceberg.aws.glue.IcebergToGlueConverter.toTypeString`
  the column type written to Glue is not accurate, so this piece of
  information may be lossy. In such cases the column parameter is used
  to store the Trino type.
  - For more compatibility we could store Iceberg type string, but
    Iceberg lacks API to reconstruct Type from string (except for
    primitive types). This is not surprising, as it needs IDs for all
    fields. Something that is not needed to answer metadata queries.

- Column NOT NULL constraint is stored as `trino_not_null=true` column table
      parameter (omitted for nullable columns).

Before the above cached information is used, the following conditions
are checked

- the `trino_table_metadata_info_valid_for` table property must be set
  to current metadata location. This ensures that the cached information
  is invalided whenever metadata location changes.

- at least one column must have `trino_type` property set. This ensures
  the cached information is not used when column parameters were lost in
  transit or otherwise erased.

`iceberg.glue.cache-table-metadata` serves as a kill-switch for the new
functionality (both write & read parts).

Alternative to https://github.com/trinodb/trino/pull/18299

PR #28266, opened 2026-02-12
Fix query-scope cache invalidation for Glue catalog

piotrrzysko:
## Description

This PR fixes a bug where `GlueIcebergTableOperations#commitTableUpdate` was not invalidating the query-scope table metadata cache after a successful commit. As a result, Iceberg observed stale metadata when determining committed files. This caused uncommitted manifest files from failed retry attempts to remain as orphans instead of being cleaned up.

Automatic cleanup is performed in [BaseTransaction#commitSimpleTransaction](https://github.com/apache/iceberg/blob/444e381dc786d6b61d8325e9526b816f881307cc/core/src/main/java/org/apache/iceberg/BaseTransaction.java#L409-L413.) However, it was never triggered because `committedFiles` was null due to stale information returned by [TableOperations::current](https://github.com/apache/iceberg/blob/444e381dc786d6b61d8325e9526b816f881307cc/core/src/main/java/org/apache/iceberg/BaseTransaction.java#L489.).

The fix refactors the code to wrap the Glue client in a caching client implementation, centralizing cache management and preventing similar bugs in the future.

## Release notes

```markdown
## Section
* Fix orphaned manifest files not being cleaned up after commit retries when using the Glue catalog. ({issue}`issuenumber`)
```


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergPolarisCatalogConnectorSmokeTest.java
```diff
@@ -252,4 +252,12 @@ public void testDeleteRowsConcurrently()
         //TODO: Fix https://github.com/trinodb/trino/issues/23941
         abort("Skipped for now due to #23941");
     }
+
+    @Test
+    @Override
+    public void testUncommittedManifestCleanupAfterConcurrentWrites()
+    {
+        //TODO: Fix https://github.com/trinodb/trino/issues/23941
+        abort("Skipped for now due to #23941");
```
findepi: the test is self-maintained when it follows the pattern

```
assertThatThrownBy(super::test_name)
  .hasMessage...(...);
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergS3TablesConnectorSmokeTest.java
```diff
@@ -147,6 +147,10 @@ public void testShowCreateTable()
                         "\\)");
     }
 
+    @Test
+    @Override // The TrinoFileSystem.listFiles is unsupported
+    public void testUncommittedManifestCleanupAfterConcurrentWrites() {}
```
findepi: Is it "TrinoFileSystem.listFiles is not supported for S3 tables buckets"?

Also,  the super impl could stop using `listFiles` but this override won't disappear. Using `assertThatThrownBy(super::test_name)` + exception assertion that captures listFiles-related limitation would be nice

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergUnityRestCatalogConnectorSmokeTest.java
```diff
@@ -261,6 +261,14 @@ public void testDeleteRowsConcurrently()
                 .hasMessageContaining("Access Denied");
     }
 
+    @Test
+    @Override
+    public void testUncommittedManifestCleanupAfterConcurrentWrites()
+    {
+        assertThatThrownBy(super::testUncommittedManifestCleanupAfterConcurrentWrites)
+                .hasMessageContaining("Access Denied");
```
findepi: Why? What's going on?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -184,6 +186,65 @@ public void testDeleteRowsConcurrently()
         }
     }
 
+    @RepeatedTest(3)
+    @Execution(ExecutionMode.SAME_THREAD)
```
findepi: Document both.
Is running once sufficient to reproduce the problem with reasonable probability?
Maybe, going forward, running once  will be enough on CI?

If so, making it ordinary `@Test` should be fine. If the test is probabilistic, this deserves a code comment

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMotoConnectorSmokeTest.java
```diff
@@ -135,6 +135,11 @@ protected boolean locationExists(String location)
     @Override
     public void testDeleteRowsConcurrently() {}
 
+    @Test
+    @Disabled("Moto is not concurrency safe")
```
findepi, line 139: 😮 

Does it make sense to continue to test with Moto? What value does it add?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/glue/TrinoGlueCatalog.java
```diff
@@ -449,7 +443,7 @@ private void getColumnsFromIcebergMetadata(
         for (SchemaTableName tableName : relationFilter.apply(glueTables.keySet())) {
             Table table = glueTables.get(tableName);
             // potentially racy with invalidation, but TrinoGlueCatalog is session-scoped
```
findepi, line 445: This comment was already here. Now we're adding `QueryScopedCachingTrinoGlueClient`.
Do we really need two levels of caching & invalidation?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
piotrrzysko: Changed to `@Test` - locally, I can reproduce the issue this test verifies on the first attempt.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMotoConnectorSmokeTest.java
piotrrzysko: > Does it make sense to continue to test with Moto?

That’s a good question, but I think it's outside the scope of this PR. I followed the existing pattern (see the test above).

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergUnityRestCatalogConnectorSmokeTest.java
piotrrzysko: I think the explanation is in this PR: https://github.com/trinodb/trino/pull/22632 and in the test setup: https://github.com/trinodb/trino/blob/b29013dd5b36bb446bab2a2e2021a75355c91cce/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergUnityRestCatalogConnectorSmokeTest.java#L71

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/glue/TrinoGlueCatalog.java
piotrrzysko: By `two levels of caching & invalidation` do you mean `TrinoGlueCatalog` (level 1) and `QueryScopedCachingTrinoGlueClient` (level 2)?

If so, I introduced `QueryScopedCachingTrinoGlueClient` to keep cache invalidation close to the state-modifying operations. Previously, they were decoupled, which in my opinion led to the bug I’m trying to fix.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergPolarisCatalogConnectorSmokeTest.java
piotrrzysko: I’m not sure yet about this. I think it fails nondeterministically, and I couldn’t find the error message with which it failed for me. Locally, I couldn’t reproduce failures for either this test or the one above.

findepi: you're right. this pattern is not applicable for tests which fail nondeterministically

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergUnityRestCatalogConnectorSmokeTest.java
findepi: it's not clear what access and to what is denied.
maybe we we asserted on more details of the exception message, this could be more self-documenting?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMotoConnectorSmokeTest.java
findepi: > That’s a good question, but I think it's outside the scope of this PR.

of course!

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/glue/TrinoGlueCatalog.java
findepi: I get there is a subtle bug manifesting in partial cleanup.

Is the bug fixiable without introducing `QueryScopedCachingTrinoGlueClient`? what would the fix look like, and what would be the cons of doing it?

piotrrzysko: I’m fixing a missing invalidation in `GlueIcebergTableOperations#commitTableUpdate`, where the raw `GlueClient` was used. I introduced `QueryScopedCachingTrinoGlueClient` to share invalidation logic between `TrinoGlueCatalog` and `GlueIcebergTableOperations`.

I didn’t intend to address this pre-existing code comment.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergUnityRestCatalogConnectorSmokeTest.java
piotrrzysko: I opened a PR to address this:

* https://github.com/trinodb/trino/pull/28281 

Once it is merged, I'll apply the same pattern here.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorSmokeTest.java
```diff
@@ -184,6 +188,72 @@ public void testDeleteRowsConcurrently()
...
+                    .describedAs("No orphaned manifests expected - uncommitted files should be cleaned up during commit")
+                    .isEmpty();
+        }
+        finally {
+            executor.shutdownNow();
+            assertThat(executor.awaitTermination(10, SECONDS)).isTrue();
+        }
+    }
+
+    private Set<String> listManifestFiles(String tableName)
```
raunaqmorarka, line 242: Can we use `computeActual("SELECT path FROM \"table$manifests\"").getOnlyColumnAsSet()` ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/glue/TrinoGlueCatalog.java
```diff
@@ -1551,4 +1507,102 @@ private record MaterializedViewData(
...
+
+        @Override
+        public void createDatabase(DatabaseInput database)
+        {
+            delegate.createDatabase(database);
+        }
+
+        public void populateCache(SchemaTableName tableName, Table table)
+        {
+            uncheckedCacheGet(tableCache, tableName, () -> table);
```
findepi, line 1605: comment why it's safe

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/glue/TestIcebergGlueCatalogAccessOperations.java
```diff
@@ -284,7 +283,7 @@ public void testRefreshMaterializedView()
 
             assertGlueMetastoreApiInvocations("REFRESH MATERIALIZED VIEW test_refresh_mview_view",
                     ImmutableMultiset.<GlueMetastoreMethod>builder()
-                            .addCopies(GET_TABLE, 6)
+                            .addCopies(GET_TABLE, 7)
```
findepi, line 286: 😞

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/glue/TestIcebergGlueCatalogAccessOperations.java
```diff
@@ -572,10 +568,7 @@ public void testInformationSchemaTableAndColumns()
                                 .add(GET_DATABASE)
                                 .add(GET_TABLE)
                                 .build(),
-                        ImmutableMultiset.<FileOperation>builder()
-                                .add(new FileOperation(METADATA_JSON, "InputFile.length"))
-                                .add(new FileOperation(METADATA_JSON, "InputFile.newInput"))
```
findepi, line 577: from commit message

> Changes in testInformationSchemaTableAndColumns were required because
now INSERTs in the test setup populate the filesystem cache so SELECTs
hit the cache.

this means the test is no longer testing what it appears it tries to
DESCRIBE still can result in `InputFile.newInput`, just that it's hidden from the test,
The test should be running with `iceberg.metadata-cache.enabled=false`, or flush FS cache in `assertInvocations`

raunaqmorarka: This test class is about testing Glue catalog accesses, and that's what it should focus on.
There are other tests like `TestIcebergFileOperations` which focus on iceberg metadata accesses on filesystem.

findepi: The filesystem access counts depend on catalog in use. In fact, they are here for metadata queries, which have have quite important catalog-specific component. Here we're asserting a metadata query. `TestIcebergFileOperations` is awesome for regular query access patterns and catalog-agnostic components. It doesn't run with Glue and doesn't, therefore, guard against potential metadata-related regressions in Glue catalog.

raunaqmorarka: I thought any catalog would need to look-up table entry in metastore and then read metadata.json from FS to show output of DESCRIBE. Not sure why the FS access would depend on catalog.

findepi: eg

- https://github.com/trinodb/trino/pull/18315
- https://github.com/trinodb/trino/pull/18517

PR #28267, opened 2026-02-12
Small optimizations

wendigo:
## Description

Spotted while going through the JFR

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/exchange/ExchangeId.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/exchange/ExchangeId.java
@@ -17,7 +17,6 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonValue;
 
 import java.util.Objects;
-import java.util.regex.Pattern;
 
 import static io.airlift.slice.SizeOf.estimatedSizeOf;
 import static io.airlift.slice.SizeOf.instanceSize;
@@ -27,8 +26,6 @@ import static java.util.UUID.randomUUID;
 public class ExchangeId
 {
     private static final long INSTANCE_SIZE = instanceSize(ExchangeId.class);
-
-    private static final Pattern ID_PATTERN = Pattern.compile("[a-zA-Z0-9_-]+");
 
     private final String id;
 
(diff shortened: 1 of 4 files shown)


core/trino-spi/src/main/java/io/trino/spi/exchange/ExchangeId.java
```diff
@@ -82,4 +79,15 @@ public long getRetainedSizeInBytes()
     {
         return INSTANCE_SIZE + estimatedSizeOf(id);
     }
+
+    private static boolean isValidId(String id)
```
losipiuk, line 83: `Validate exchangeId without regex ` drop.
This is much less readable

losipiuk (commented):
two last commits are fine

core/trino-spi/src/main/java/io/trino/spi/exchange/ExchangeId.java
wendigo: We do the same in the QueryId and other classes

losipiuk: We do lots of weird thing in various places. Why follow those patterns?

core/trino-spi/src/main/java/io/trino/spi/security/Identity.java
```diff
@@ -197,9 +197,9 @@ public static class Builder
         private String user;
         private Set<String> groups = new HashSet<>();
         private Optional<Principal> principal = Optional.empty();
-        private Set<String> enabledRoles = new HashSet<>();
-        private Map<String, SelectedRole> connectorRoles = new HashMap<>();
-        private Map<String, String> extraCredentials = new HashMap<>();
+        private Set<String> enabledRoles = new HashSet<>(0);
```
chenjian2664, line 200: Revert? the initial capacity would be 16 anyway

core/trino-spi/src/main/java/io/trino/spi/exchange/ExchangeId.java
chenjian2664: I found the original the commit mention performance [here](https://github.com/trinodb/trino/commit/e656ec4632d6f09a41c726e2e116a9e4325e9b85) 
Maybe considering reusing the method in QueryId?

PR #28270, opened 2026-02-12
Add function resolver cache

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/BuiltinFunctionResolver.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/BuiltinFunctionResolver.java
@@ -25,6 +25,7 @@ import io.trino.spi.function.OperatorType;
 import io.trino.spi.function.Signature;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.TypeManager;
+import io.trino.spi.type.TypeSignature;
 import io.trino.sql.analyzer.TypeSignatureProvider;
 
 import java.util.Collection;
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/metadata/BuiltinFunctionResolver.java
```diff
@@ -66,12 +68,24 @@ public BuiltinFunctionResolver(Metadata metadata, TypeManager typeManager, Globa
...
     }
 
     ResolvedFunction resolveBuiltinFunction(String name, List<TypeSignatureProvider> parameterTypes)
     {
-        CatalogFunctionBinding functionBinding = functionBinder.bindFunction(parameterTypes, getBuiltinFunctions(name), name);
-        return resolveBuiltin(functionBinding);
+        try {
+            return uncheckedCacheGet(functionCache, FunctionCacheKey.from(name, parameterTypes),
+                () -> {
+                    CatalogFunctionBinding functionBinding = functionBinder.bindFunction(parameterTypes, getBuiltinFunctions(name), name);
```
losipiuk, line 79: Seems fine to me. But are we sure this would be stable over time? E.g. if we have dynamic catalogs which provide functions, and catalog definition chagnes over time?

dain:
We did this a while back, and my memory is it had problems. The true issue here is why is function resolution so unbelievably slow? The answer is, it needs a real rewrite. The satisfaction algorithm is way too complicated, and when you are dealing with cast functions, it's just a total mess. Something much much simpler could be done here. This really just masks over a problem, and my memory is it causes additional problems.  I unfortunately can't remember what the actual problems were. I think part of it is there will now be like four layers of caching in the function system, which again is just a sign that there's something really fundamentally broken.

wendigo:
@dain does this change hurt?

dain:
My memory is that we used to have this and removed it for a reason. But I do not remember the reason, so maybe it hurts, maybe it doesn't. I have no idea. 

What is the actual problem you're trying to fix? The code is literally, I'm going to add a cache on this, but I assume you see a real problem and that this fixes the problem.

dain:
I mean, if you wanna just check in, then don't let me hold you back.

Issue #27416, opened 2025-11-23
Trino does not honour the vended credentials from rest iceberg catalog


SatvikNema:
I have a custom rest iceberg catalog (with AWS S3), which implements the rest iceberg catalog [spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml). We have been using it with spark with no issues, where it honours the vended credentials from the [LoadTableResponse](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L3396). 
But when we use the same with Trino, it does not use those credentials, which results in this error while reading the metadata file:
```
Caused by: io.trino.filesystem.TrinoFileSystemException: S3 HEAD request failed for file: s3://path/to/file/metadata/snap-5781680776850380500-1-4e41ec9d-88bb-40cc-9599-01e3956780a8.avro
	at io.trino.filesystem.s3.S3InputFile.headObject(S3InputFile.java:160)
	at io.trino.filesystem.s3.S3InputFile.length(S3InputFile.java:83)
	at io.trino.filesystem.tracing.Tracing.withTracing(Tracing.java:51)
	at io.trino.filesystem.tracing.TracingInputFile.length(TracingInputFile.java:81)
	at io.trino.filesystem.memory.MemoryFileSystemCache.load(MemoryFileSystemCache.java:179)
	at io.trino.filesystem.memory.MemoryFileSystemCache.lambda$getOrLoadFromCache$3(MemoryFileSystemCache.java:169)
	at com.google.common.cache.LocalCache$LocalManualCache$1.load(LocalCache.java:4903)
	at com.google.common.cache.LocalCache$LoadingValueReference.loadFuture(LocalCache.java:3574)
	at com.google.common.cache.LocalCache$Segment.loadSync(LocalCache.java:2316)
	at com.google.common.cache.LocalCache$Segment.lockedGetOrLoad(LocalCache.java:2190)
	at com.google.common.cache.LocalCache$Segment.get(LocalCache.java:2080)
	at com.google.common.cache.LocalCache.get(LocalCache.java:4017)
	at com.google.common.cache.LocalCache$LocalManualCache.get(LocalCache.java:4898)
	at io.trino.cache.EvictableCache.get(EvictableCache.java:119)
	at io.trino.filesystem.memory.MemoryFileSystemCache.getOrLoadFromCache(MemoryFileSystemCache.java:169)
	... 24 more
[...]

PR #28290, opened 2026-02-13
Fix Iceberg vended credentials not working with REST catalogs that return credentials via `storage-credentials` instead of `config`.

rcjverhoef:
## Description

Iceberg Credential Vending doesn't work with REST catalogs following latests specs, specifically catalogs that don't provide the `config` fallback. Per [latest spec](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L3485):

> Clients must first check whether the respective credentials exist in the storage-credentials field before checking the config for credentials.

The Iceberg REST spec defines two mechanisms for credential vending:
- **`config` map** (older): credentials are returned as key-value pairs in the `config` field of `LoadTableResponse`. Trino already supports this.
- **`storage-credentials` array** (newer, preferred): credentials are returned as typed `Credential` objects with a `prefix` and `config`. Trino did not support this.

When Trino provides a custom `ioBuilder` to `RESTSessionCatalog` (which it always does), `storage-credentials` are dropped in [`RESTSessionCatalog.newFileIO()`](https://github.com/apache/iceberg/blob/apache-iceberg-1.10.1/core/src/main/java/org/apache/iceberg/rest/RESTSessionCatalog.java#L981-L982) — only the `config` map is passed through. Without a custom `ioBuilder`, Iceberg handles this via the `SupportsStorageCredentials` interface, but that path is never taken in Trino.

## Release notes

(x ) Release notes are required, with the following suggested text:

## Iceberg connector
* Fix vended credentials not working with REST catalogs that return credentials via `storage-credentials` instead of `config`.

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/StorageCredentialsMergingRestClient.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/StorageCredentialsMergingRestClient.java
@@ -1,0 +1,170 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.iceberg.catalog.rest;
+
+import org.apache.iceberg.rest.RESTClient;
+import org.apache.iceberg.rest.RESTRequest;
+import org.apache.iceberg.rest.RESTResponse;
+import org.apache.iceberg.rest.auth.AuthSession;
+import org.apache.iceberg.rest.credentials.Credential;
+import org.apache.iceberg.rest.responses.ErrorResponse;
+import org.apache.iceberg.rest.responses.LoadTableResponse;
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+import static java.util.Objects.requireNonNull;
+
+/**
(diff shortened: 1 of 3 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/StorageCredentialsMergingRestClient.java
```diff
@@ -0,0 +1,170 @@
+/*
```
ebyhr, line 1: Can we add a test case to demonstrate #27416?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoIcebergRestCatalogFactory.java
```diff
@@ -135,10 +135,13 @@ public synchronized TrinoCatalog create(ConnectorIdentity identity)
             }
 
             RESTSessionCatalog icebergCatalogInstance = new RESTSessionCatalog(
-                    config -> HTTPClient.builder(config)
-                            .uri(config.get(CatalogProperties.URI))
-                            .withHeaders(RESTUtil.configHeaders(config))
-                            .build(),
+                    config -> {
+                        var client = HTTPClient.builder(config)
```
ebyhr: Please avoid var: https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md#avoid-var

rcjverhoef: removed

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/StorageCredentialsMergingRestClient.java
rcjverhoef: added a test-case.

as fyi: most existing tests use the tabular rest-catalog. That seems to have been moved to databricks docker registry post the acquisition, but still both versions are not kept up-to-date. Seems these should be moved to `iceberg-rest-fixture`, but that feels out-of-scope of this PR. I am ok doing the work for that in a separate PR, can make that pre-cursor for this if you'd prefer.

PR #28292, opened 2026-02-13
Allow to use `#`/`?` chars in Google Cloud Storage paths

raphaelsolarski:
## Description
This PR removes validation which blocked from using `#`/`?` characters in GCS file names e.g. when using hive connector.
Probably intend of the validation author was to block from using GCS Generation numbers in file names, but java google sdk uses a separate field to store the generation numbers so we should not encounter generation numbers with "google cloud cli notation" in names/paths. 

* GCS file names can contains `#` characters. It is discouraged because it is specially treated in google cloud cli. But it is not directly prohibited -> https://docs.cloud.google.com/storage/docs/objects#recommendations
* legacy hdfs based implementation seems to allow `#` chars
* java gcs sdk returns only current versions of blobs by default
* during http transport file name is encoded to url safe form for as described here -> https://docs.cloud.google.com/storage/docs/request-endpoints#encoding
Example of such namig in GCS console:
Screenshot 2026-02-13 at 12 23 55

## Stack trace (for prepared file)
```
Query 20260218_095645_00011_6qgmx failed: GCS generation numbers are not supported: gs://some-test/connection_check/20260218_095457_00008_6qgmx_71a31166-7fb7-41fb-8844-f36dc5b#9198
java.lang.IllegalArgumentException: GCS generation numbers are not supported: gs://some-test/connection_check/20260218_095457_00008_6qgmx_71a31166-7fb7-41fb-8844-f36dc5b#9198
	at com.google.common.base.Preconditions.checkArgument(Preconditions.java:217)
	at io.trino.filesystem.gcs.GcsLocation.<init>(GcsLocation.java:33)
	at io.trino.filesystem.gcs.GcsFileSystem.newInputFile(GcsFileSystem.java:125)
	at io.trino.filesystem.switching.SwitchingFileSystem.newInputFile(SwitchingFileSystem.java:72)
	at io.trino.filesystem.tracing.TracingFileSystem.newInputFile(TracingFileSystem.java:63)
	at io.trino.plugin.hive.parquet.ParquetPageSourceFactory.createPageSource(ParquetPageSourceFactory.java:187)
	at io.trino.plugin.hive.HivePageSourceProvider.createHivePageSource(HivePageSourceProvider.java:208)
	at io.trino.plugin.hive.HivePageSourceProvider.createPageSource(HivePageSourceProvider.java:143)
	at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorPageSourceProvider.createPageSource(ClassLoaderSafeConnectorPageSourceProvider.java:48)
	at io.trino.split.PageSourceManager$PageSourceProviderInstance.createPageSource(PageSourceManager.java:80)
	at io.trino.operator.TableScanOperator.getOutput(TableScanOperator.java:273)
	at io.trino.operator.OutputValidatingSourceOperator.getOutput(OutputValidatingSourceOperator.java:70)
	at io.trino.operator.Driver.processInternal(Driver.java:403)
	at io.trino.operator.Driver.lambda$process$0(Driver.java:306)
	at io.trino.operator.Driver.tryWithLock(Driver.java:709)
	at io.trino.operator.Driver.process(Driver.java:298)
	at io.trino.operator.Driver.processForDuration(Driver.java:269)
	at io.trino.execution.SqlTaskExecution$DriverSplitRunner.processFor(SqlTaskExecution.java:852)
	at io.trino.execution.executor.dedicated.SplitProcessor.run(SplitProcessor.java:77)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.lambda$run$0(TaskEntry.java:205)
	at io.trino.$gen.Trino_testversion____20260218_095333_36.run(Unknown Source)
	at io.trino.execution.executor.dedicated.TaskEntry$VersionEmbedderBridge.run(TaskEntry.java:206)
	at io.trino.execution.executor.scheduler.FairScheduler.runTask(FairScheduler.java:177)
	at io.trino.execution.executor.scheduler.FairScheduler.lambda$submit$0(FairScheduler.java:164)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
```

## Release notes

* Fix to situation when `#` char in GCS file name was causing the "GCS generation numbers are not supported:" error

### DIFF
--- a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
+++ b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
@@ -23,8 +23,6 @@ record GcsLocation(Location location)
     GcsLocation
     {
         // Note: Underscores are allowed in bucket names, see https://cloud.google.com/storage/docs/buckets#naming.
-        // Generation numbers are not supported, i.e. gs://bucket/path#generation.
-        // For more details on generation numbers see https://cloud.google.com/storage/docs/object-versioning.
 
         requireNonNull(location, "location");
         checkArgument(location.scheme().isPresent(), "No scheme for GCS location: %s", location);
@@ -32,8 +30,6 @@ record GcsLocation(Location location)
         checkArgument(location.host().isPresent(), "No bucket for GCS location: %s", location);
         checkArgument(location.userInfo().isEmpty(), "GCS location contains user info: %s", location);
         checkArgument(location.port().isEmpty(), "GCS location contains port: %s", location);
-        checkArgument(!location.path().contains("#"), "GCS generation numbers are not supported: %s", location);
-        checkArgument(!location.path().contains("?"), "Invalid character '?': %s", location);
     }
 
     public String scheme()
--- a/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
+++ b/lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
@@ -174,11 +174,11 @@ public abstract class AbstractTestGcsFileSystem
     void testRoundTripFileWithDiscouragedCharsName()
             throws Exception
     {
-        // According to https://docs.cloud.google.com/storage/docs/objects#recommendations some chars ([*]) are discouraged
+        // According to https://docs.cloud.google.com/storage/docs/objects#recommendations some chars ([*]#?) are discouraged
         // because they are specially treated in gcloud cli. But they are not directly prohibited.
         byte[] buffer = new byte[8];
         String stringToWrite = "test";
-        Location fileLocation = getRootLocation().appendPath("[*]");
+        Location fileLocation = getRootLocation().appendPath("[*]#?");
         TrinoOutputFile outputFile = getFileSystem().newOutputFile(fileLocation);
         outputFile.createOrOverwrite(stringToWrite.getBytes(UTF_8));
         TrinoInputFile inputFile = getFileSystem().newInputFile(fileLocation);
--- a/testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
+++ b/testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
@@ -133,14 +133,14 @@ public abstract class BaseTestTableFormats
 
     protected void testLocationContainsDiscouragedCharacter(String schemaLocation)
     {
-        // According to https://docs.cloud.google.com/storage/docs/objects#recommendations some chars ([*]) are discouraged
+        // According to https://docs.cloud.google.com/storage/docs/objects#recommendations some chars ([*]#?) are discouraged
         // because they are specially treated in gcloud cli. But they are not directly prohibited.
         // Chars used in schema location are not escaped like those in partition names
         // so it allows to test whether whole stack works e2e with discouraged characters.
         String schemaName = getCatalogName() + ".test";
         String tableName = "%s.test_location_contains_discouraged_character_%s".formatted(schemaName, randomNameSuffix());
         try {
-            String locationWithDiscouragedChars = schemaLocation + "/[*]";
+            String locationWithDiscouragedChars = schemaLocation + "/[*]#?";
             onTrino().executeQuery("CREATE SCHEMA " + schemaName + " WITH (location = '" + locationWithDiscouragedChars + "')");
             onTrino().executeQuery("CREATE TABLE " + tableName + " (id bigint, someValue varchar)");
 


raphaelsolarski:
I just realized that maybe we should also allow "?" character.

ebyhr:
/test-with-secrets sha=2e785256d22b13f80c67f16cb2922f032980f6d5

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystem.java
```diff
@@ -34,6 +38,23 @@ void setup()
         initialize(requireEnv("GCP_CREDENTIALS_KEY"));
     }
 
+    @Test
+    void testRoundTripFileWithHashInName()
```
ebyhr: Why do we add this test to TestGcsFileSystem, not AbstractTestGcsFileSystem or AbstractTestTrinoFileSystem?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
```diff
@@ -23,16 +23,13 @@ record GcsLocation(Location location)
...
-        // Generation numbers are not supported, i.e. gs://bucket/path#generation.
-        // For more details on generation numbers see https://cloud.google.com/storage/docs/object-versioning.
 
         requireNonNull(location, "location");
         checkArgument(location.scheme().isPresent(), "No scheme for GCS location: %s", location);
         checkArgument(location.scheme().get().equals("gs"), "Wrong scheme for GCS location: %s", location);
         checkArgument(location.host().isPresent(), "No bucket for GCS location: %s", location);
         checkArgument(location.userInfo().isEmpty(), "GCS location contains user info: %s", location);
         checkArgument(location.port().isEmpty(), "GCS location contains port: %s", location);
-        checkArgument(!location.path().contains("#"), "GCS generation numbers are not supported: %s", location);
```
ebyhr, line 35: Can we add a test to the Iceberg connector (ideally Hive and Delta as well) so we can verify the entire query behavior? Each connector may have additional handling to special characters.

findinpath: @raphaelsolarski we can't control the file names on Delta & Iceberg (they're UUID generated).
However we can do that by changing the table names or schema names to contain a special character `#` / `?`

For Hive we can repro the behavior by creating and uploading to an existing table a file containing special characters.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
```diff
@@ -23,16 +23,13 @@ record GcsLocation(Location location)
...
         // Note: Underscores are allowed in bucket names, see https://cloud.google.com/storage/docs/buckets#naming.
-        // Generation numbers are not supported, i.e. gs://bucket/path#generation.
-        // For more details on generation numbers see https://cloud.google.com/storage/docs/object-versioning.
 
         requireNonNull(location, "location");
         checkArgument(location.scheme().isPresent(), "No scheme for GCS location: %s", location);
         checkArgument(location.scheme().get().equals("gs"), "Wrong scheme for GCS location: %s", location);
         checkArgument(location.host().isPresent(), "No bucket for GCS location: %s", location);
         checkArgument(location.userInfo().isEmpty(), "GCS location contains user info: %s", location);
         checkArgument(location.port().isEmpty(), "GCS location contains port: %s", location);
```
findinpath, line 32: In the commit message.

Google Cloud Store -> Google Cloud Storage

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
ebyhr: We can still control the file name in Iceberg connector if we use `add_files` procedure. Or just using the character in the partition value should demonstrate the behavior in both Iceberg and Delta Lake connectors.

> Allow to use hash character in Google Cloud Storage **_file_** names

The current title might be misleading. This PR affects the entire path, not limited to the file name.

findepi (approved):
LGTM, but i agree with @ebyhr it would be good to also have end to end tests with connectors.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
raphaelsolarski: ok, I'm working on adding a test to validate the behavior e2e - in smoke or product tests.
I'll update the pr/commit title.

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
```diff
@@ -166,4 +169,21 @@ void testExistingFileWithTrailingSlash()
             storage.delete(blobId);
         }
     }
+
+    @Test
+    void testRoundTripFileWithHashInName()
```
marcinsbd: should we also have similar test for other char as '?' ?

ebyhr:
/test-with-secrets sha=e0b8da542a6d8e5fb4637f362990796bfd86cbdb

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
raphaelsolarski: good point. I will add all "discouraged" chars to the file name and rename the test.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
raphaelsolarski: Product test added for whole gcs suite (Iceberg, Hive, DeltaLake).

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/TestGcsFileSystem.java
raphaelsolarski: I renamed the test and I moved it to `AbstractTestGcsFileSystem`. Despite added product test I would like to left this one as it allows faster feedback than product tests + tests behavior when the chars are in the last "segment" (theoretically it should not make a difference)

raphaelsolarski:
@ebyhr @findepi @findinpath @marcinsbd Thanks for your feedback
According to @findinpath suggestion I also removed validation against `?` char in `GcsLocation` to prevent similar issues with that char.
I added product test + addressed comments.
I also added `GcsHiveQueryRunner` in a separate pre-commit to make debugging of the setup easier in the future (pls let me know if it should be a separate PR for that).

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
```diff
@@ -166,4 +169,21 @@ void testExistingFileWithTrailingSlash()
...
         }
     }
+
+    @Test
+    void testRoundTripFileWithDiscouragedCharsName()
+            throws Exception
+    {
+        byte[] buffer = new byte[8];
+        String stringToWrite = "test";
+        Location fileLocation = getRootLocation().appendPath("?#[*]");
```
findinpath: It would be worthwhile to add a comment

// According to https://docs.cloud.google.com/storage/docs/objects#recommendations the characters .... are discouraged in Google Cloud Storage file names.

findinpath (approved):

plugin/trino-hive/src/test/resources/io/trino/plugin/hive/gcs/hdp3.1-core-site.xml
```diff
@@ -0,0 +1,145 @@
+<?xml version="1.0"?>
+<!--
```
findinpath, line 2: pls name it `hdp3.1-core-site.xml.gcs-template` same as the other occurrences in the trino codebase.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
```diff
@@ -0,0 +1,175 @@
...
+        Path gcpCredentialsFile = Files.createTempFile("gcp-credentials", ".json", READ_ONLY_PERMISSIONS);
+        gcpCredentialsFile.toFile().deleteOnExit();
+        Files.writeString(gcpCredentialsFile, jsonKeyBytes);
+
+        String projectId = extractProjectId(jsonKeyBytes);
+
+        String coreSiteXmlContent = Resources.toString(
+                        getResource("io/trino/plugin/hive/gcs/core-site.xml.gcs-template"), UTF_8)
+                .replace("%GCP_CREDENTIALS_FILE_PATH%", "/etc/hadoop/conf/gcp-credentials.json")
+                .replace("%GCP_PROJECT_ID%", projectId); // needed for hive be able to create a bucket if it not exist
```
findinpath: does the comment make sense?

plugin/trino-hive/src/test/resources/io/trino/plugin/hive/gcs/core-site.xml.gcs-template
```diff
@@ -0,0 +1,145 @@
...
+      <description>
+        The JSON key file of the service account used for GCS
+        access when google.cloud.auth.service.account.enable is true.
+      </description>
+    </property>
+
+    <property>
+        <name>fs.gs.project.id</name>
+        <value>%GCP_PROJECT_ID%</value>
+    </property>
```
findinpath: Not sure whether this parameter is actually needed.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
findinpath: can the code run without `fs.gs.project.id` setting?

raphaelsolarski: I deleted project_id extraction/setting as we discussed.

raphaelsolarski: yes, that was only for scenario when bucket was not existing. I already removed the part.

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
```diff
@@ -131,6 +131,33 @@ protected void testPathContainsSpecialCharacter(String schemaLocation, String pa
...
+
+            onTrino().executeQuery("INSERT INTO " + getCatalogName() + ".test." + tableName + " VALUES " +
+                    "(1, 'someValue')");
+
+            assertThat(onTrino().executeQuery(format("SELECT * FROM %1$s.test.%2$s", getCatalogName(), tableName))).containsOnly(ImmutableList.of(
+                    row(1, "someValue")));
+        }
+        finally {
+            onTrino().executeQuery("DROP TABLE %1$s.test.%2$s".formatted(getCatalogName(), tableName));
+            onTrino().executeQuery("DROP SCHEMA %1$s.test".formatted(getCatalogName()));
```
ebyhr: These statements may throw an exception if CREATE SCHEMA/TABLE fail for some reason. Please add `IF EXISTS` option.

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
```diff
@@ -23,17 +23,13 @@ record GcsLocation(Location location)
     GcsLocation
     {
         // Note: Underscores are allowed in bucket names, see https://cloud.google.com/storage/docs/buckets#naming.
-        // Generation numbers are not supported, i.e. gs://bucket/path#generation.
-        // For more details on generation numbers see https://cloud.google.com/storage/docs/object-versioning.
```
ebyhr, line 27: `io.trino.filesystem.gcs.GcsFileSystem#listFiles` doesn't return non-current objects having generation numbers, right? 

Could you update a commit body to explain why allowing # is safe from object versioning perspective?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
```diff
@@ -0,0 +1,159 @@
...
+
+public final class GcsHiveQueryRunner
+{
+    static {
+        Logging.initialize();
+    }
+
+    private static final FileAttribute<?> READ_ONLY_PERMISSIONS = PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rw-r--r--"));
+
+    private GcsHiveQueryRunner() {}
```
ebyhr: This method can be static.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
```diff
@@ -0,0 +1,159 @@
...
+        }
+    }
+
+    static void main()
+            throws Exception
+    {
+        String gcpCredentialKey = requiredNonEmptySystemProperty("testing.gcp-credentials-key");
+        String gcpStorageBucket = requiredNonEmptySystemProperty("testing.gcp-storage-bucket");
+
+        HiveHadoop hiveHadoop = createGcsHiveContainer(gcpCredentialKey);
```
ebyhr, line 135: Could you add `@SuppressWarnings("resource")` here and the following `QueryRunner queryRunner`?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
```diff
@@ -0,0 +1,159 @@
...
+{
+    static {
+        Logging.initialize();
+    }
+
+    private static final FileAttribute<?> READ_ONLY_PERMISSIONS = PosixFilePermissions.asFileAttribute(PosixFilePermissions.fromString("rw-r--r--"));
+
+    private GcsHiveQueryRunner() {}
+
+    public static HiveHadoop createGcsHiveContainer(String gcpCredentialKey)
```
ebyhr: This method can be static. Same for `public static Builder builder(URI hiveMetastoreEndpoint)` and `public static class Builder`.

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
```diff
@@ -131,6 +131,33 @@ protected void testPathContainsSpecialCharacter(String schemaLocation, String pa
         }
     }
 
+    protected void testLocationContainsDiscouragedCharacter(String schemaLocation)
+    {
+        // According to https://docs.cloud.google.com/storage/docs/objects#recommendations some chars (?#[*]) are discouraged
+        // because they are specially treated in gcloud cli. But they are not directly prohibited.
+        // Chars used in schema location are not escaped like those in partition names
+        // so it allows to test whether whole stack works e2e with discouraged characters.
+        String tableName = "test_location_contains_discouraged_character" + randomNameSuffix();
```
ebyhr: We could extract a variable for a qualified table name: 
```java
String trinoTableName = "%s.test.%s".formatted(getCatalogName(), tableName);
```

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
```diff
@@ -131,6 +131,33 @@ protected void testPathContainsSpecialCharacter(String schemaLocation, String pa
...
+            onTrino().executeQuery(format("CREATE SCHEMA %1$s.test WITH (location = '%2$s')", getCatalogName(), locationWithDiscouragedChars));
+            onTrino().executeQuery(format(
+                    "CREATE TABLE %1$s.test.%2$s (id bigint, someValue varchar)",
+                    getCatalogName(),
+                    tableName));
+
+            onTrino().executeQuery("INSERT INTO " + getCatalogName() + ".test." + tableName + " VALUES " +
+                    "(1, 'someValue')");
+
+            assertThat(onTrino().executeQuery(format("SELECT * FROM %1$s.test.%2$s", getCatalogName(), tableName))).containsOnly(ImmutableList.of(
```
ebyhr: `ImmutableList.of()` is redundant.

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
```diff
@@ -131,6 +131,33 @@ protected void testPathContainsSpecialCharacter(String schemaLocation, String pa
         }
     }
 
+    protected void testLocationContainsDiscouragedCharacter(String schemaLocation)
```
ebyhr, line 134: Any reason we don't update `testPathContainsSpecialCharacter()`?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
```diff
@@ -0,0 +1,159 @@
...
+            throws Exception
+    {
+        String jsonKeyBytes = new String(Base64.getDecoder().decode(gcpCredentialKey), UTF_8);
+        Path gcpCredentialsFile = Files.createTempFile("gcp-credentials", ".json", READ_ONLY_PERMISSIONS);
+        gcpCredentialsFile.toFile().deleteOnExit();
+        Files.writeString(gcpCredentialsFile, jsonKeyBytes);
+
+        String coreSiteXmlContent = Resources.toString(
+                        getResource("io/trino/plugin/hive/gcs/hdp3.1-core-site.xml.gcs-template"), UTF_8)
+                .replace("%GCP_CREDENTIALS_FILE_PATH%", "/etc/hadoop/conf/gcp-credentials.json");
```
ebyhr: Why do we use template for a static path?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsLocation.java
raphaelsolarski: > io.trino.filesystem.gcs.GcsFileSystem#listFiles doesn't return non-current objects

that is right. It requires a separate `BlobListOption.versions(true)` option added to `Storage.list`. Even when it is added  generation numbers are not encoded into name (it afaik gcloud specific behavior).

I updated commit message.

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
raphaelsolarski: I started with `testPathContainsSpecialCharacter` but after that I extracted it to a separate test to not mix concerns as `testPathContainsSpecialCharacter` actually tests something different.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
raphaelsolarski: fixed

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
raphaelsolarski: fixed

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
raphaelsolarski: fixed

plugin/trino-hive/src/test/java/io/trino/plugin/hive/gcs/GcsHiveQueryRunner.java
raphaelsolarski: fixed

testing/trino-product-tests/src/main/java/io/trino/tests/product/BaseTestTableFormats.java
raphaelsolarski: fixed

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
```diff
@@ -166,4 +169,23 @@ void testExistingFileWithTrailingSlash()
...
+
+    @Test
+    void testRoundTripFileWithDiscouragedCharsName()
+            throws Exception
+    {
+        // According to https://docs.cloud.google.com/storage/docs/objects#recommendations some chars (?#[*]) are discouraged
+        // because they are specially treated in gcloud cli. But they are not directly prohibited.
+        byte[] buffer = new byte[8];
+        String stringToWrite = "test";
+        Location fileLocation = getRootLocation().appendPath("?#[*]");
```
marcinsbd: The character set put under the test is different than in the PR title

raphaelsolarski: yes, but we just deleted validation for `#` and `?` - rest was allowed. Do we need a separate commit? 😄

ebyhr:
/test-with-secrets sha=cb7376c6ad44fc215758e1f2e81c011211272007

lib/trino-filesystem-gcs/src/test/java/io/trino/filesystem/gcs/AbstractTestGcsFileSystem.java
marcinsbd: you can separate commits 
1. adding missing test for allowed special characters
2. remove validation for  # ? and add them to the characters set in test

raphaelsolarski: ok, I separated those commits.

ebyhr:
/test-with-secrets sha=c17fbc9b0fe837380da4ed4e91d878ecb7b9a9b3

PR #28295, opened 2026-02-13
Map google.protobuf.Struct to JSON in decoder

### LINKED ISSUE #28294: Kafka protobuf decoder rejects `google.protobuf.Struct` as recursive type
Protobuf schemas containing `google.protobuf.Struct` fields can't be queried through the Kafka connector. The schema parser rejects them with the error:

```
Protobuf schema containing fields with self-reference are not supported because they cannot be mapped to a Trino type
```

`google.protobuf.Struct`, `google.protobuf.Value`, and `google.protobuf.ListValue` are [well-known protobuf types][1] with a [defined JSON mapping][2] in the protobuf specification. They form a closed type family representing arbitrary JSON-like data:

- [`Struct`](https://protobuf.dev/reference/protobuf/google.protobuf/#struct) - JSON object (e.g., `map<string, Value>`)
- [`Value`](https://protobuf.dev/reference/protobuf/google.protobuf/#value) - any JSON value (`oneof` of `null`, `number`, `string`, `bool`, `Struct`,
  `ListValue`)
- [`ListValue`](https://protobuf.dev/reference/protobuf/google.protobuf/#list-value) - JSON array (`repeated Value`)

These types contain self references (`Struct` -> `Value` -> `Struct`), which triggers the [recursive type detection in `ProtobufSchemaParser`](https://github.com/trinodb/trino/blob/b8997c21a05bcc06d1f0dfd12e3fd528a925907e/plugin/trino-kafka/src/main/java/io/trino/plugin/kafka/encoder/protobuf/ProtobufSchemaParser.java#L154-L163). However, unlike arbitrary user-defined recursive messages (which may require an infinitely nested Trino row type), `Struct`, `Value`, and `ListValue` have a canonical JSON representation defined by the [protobuf JSON mapping specification][2] outlined above. They map naturally to Trino's JSON type, with serialization handled entirely by protobuf's `JsonFormat.printer()`, similar to the [existing treatment of the `Any` type](https://github.com/trinodb/trino/blob/b8997c21a05bcc06d1f0dfd12e3fd528a925907e/plugin/trino-kafka/src/main/java/io/trino/plugin/kafka/encoder/protobuf/ProtobufSchemaParser.java#L150-L152).

## Expected behavior

Fields of type `google.protobuf.Struct`, `Value`, or `ListValue` should be mapped to Trino's `JSON` type, following the same pattern used for `google.protobuf.Any` (mapped to JSON).

## Example schema that fails

```proto
syntax = "proto3";
import "google/protobuf/struct.proto";

message Event {
    string id = 1;
    google.protobuf.Struct metadata = 2;
}
```

Querying a Kafka topic using this schema produces the recursive type error instead of returning the `metadata` field as JSON.

[1]: https://protobuf.dev/reference/protobuf/google.protobuf/
[2]: https://protobuf.dev/programming-guides/json/

ccampo133:
## Description

Map `google.protobuf.Struct`, `google.protobuf.Value`, and `google.protobuf.ListValue` well-known types to Trino's `JSON` type in the Kafka protobuf schema parser and decoder. This follows the same pattern used for `google.protobuf.Any` (which is already mapped to JSON, per https://github.com/trinodb/trino/commit/5ebbd2e46ec318a4e35725f0cc5fe27d5c70ae00).

I don't believe any configuration flag is needed like the one used for `google.protobuf.Any`. Unlike `Any` (which
requires a `DescriptorProvider` to resolve the `type_url` field), `Struct` is entirely self-describing and always safe to decode as JSON.

See the description in https://github.com/trinodb/trino/issues/28294 for why this approach is appropriate.

The changes span two modules:

- Schema parsing in `ProtobufSchemaParser`: Intercepts `Struct`/`Value`/`ListValue` types *before* the recursive type detection and maps them to JSON.

- Runtime decoding in `ProtobufColumnDecoder` and `ProtobufValueProvider`: Converts `Struct` `DynamicMessage` instances to sorted JSON strings using `JsonFormat.printer()`. Handles `Struct` both as top-level fields and when nested inside arrays, maps, or row types.

The `sorted()` helper in `ProtobufColumnDecoder` was also generalized to handle non-object JSON values (e.g., `"foo"` from a `Value` field or `["foo","bar"]` from a `ListValue` field) by using `Object.class` instead of `Map.class` in the Jackson mapper. This is backward-compatible since existing callers (`Any`, `Oneof`) always produce JSON objects.

Closes https://github.com/trinodb/trino/issues/28294

## Release notes

```
* Add support for `google.protobuf.Struct`, `Value`, and `ListValue` types in the Kafka protobuf decoder, mapping them to Trino's `JSON` type.
```

### DIFF
--- a/lib/trino-record-decoder/src/main/java/io/trino/decoder/protobuf/ProtobufColumnDecoder.java
+++ b/lib/trino-record-decoder/src/main/java/io/trino/decoder/protobuf/ProtobufColumnDecoder.java
@@ -50,7 +50,6 @@ import io.trino.spi.type.VarcharType;
 import jakarta.annotation.Nullable;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
@@ -70,6 +69,10 @@ public class ProtobufColumnDecoder
     // Trino JSON types are expected to be sorted by key
     private static final ObjectMapper mapper = JsonMapper.builder().configure(ORDER_MAP_ENTRIES_BY_KEYS, true).build();
     private static final String ANY_TYPE_NAME = "google.protobuf.Any";
+    private static final Set<String> STRUCT_TYPES = Set.of(
+            "google.protobuf.Struct",
+            "google.protobuf.Value",
+            "google.protobuf.ListValue");
     private static final Slice EMPTY_JSON = Slices.utf8Slice("{}");
 
     private static final Set<Type> SUPPORTED_PRIMITIVE_TYPES = ImmutableSet.of(
@@ -174,6 +177,12 @@ public class ProtobufColumnDecoder
             return createAnyJson((Message) value, valueDescriptor.get());
         }
 
+        if (valueDescriptor.isPresent()
+                && STRUCT_TYPES.contains(valueDescriptor.get().getFullName())
+                && value instanceof Message structMessage) {
+            return structToJsonSlice(structMessage);
+        }
+
         return value;
     }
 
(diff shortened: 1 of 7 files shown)


wendigo:
(rebased to drop merge commit)

chenjian2664:
cc @findepi

Issue #26214, opened 2025-07-16
Repeated ANALYZE fails on Hive metastore 4.0.x


sbernauer:
We tested the following SQLs on Trino 451, 470 and 476, all versions behave the same.

On Hive metastore 3.1.3 everything works fine, on 4.0.0 and 4.0.1 the problem below occurs.

Minimal run for easy reproduction:

```sql
CREATE TABLE IF NOT EXISTS hive.demo.ny_taxi_data_raw5 (
    VendorID BIGINT,
    tpep_pickup_datetime TIMESTAMP,
    tpep_dropoff_datetime TIMESTAMP,
    passenger_count DOUBLE,
    trip_distance DOUBLE,
    payment_type BIGINT,
    Fare_amount DOUBLE,
    Tip_amount DOUBLE,
    Total_amount DOUBLE
) WITH (
    external_location = 's3a://demo/ny-taxi-data/raw/',
    format = 'parquet'
);
-- Works
ANALYZE hive.demo.ny_taxi_data_raw5;
-- Fails: Invalid column statistics data: ColumnStatisticsObj(colName:tpep_dropoff_datetime, colType:timestamp, statsData:<ColumnStatisticsData >)
ANALYZE hive.demo.ny_taxi_data_raw5;
```

This where we get the test data from. It's the well-known New York taxi dataset.

```bash
for month in \
2020-01 2020-02 2020-03 2020-04 2020-05 2020-06 2020-07 2020-08 2020-09 2020-10 \
2020-11 2020-12 2021-01 2021-02 2021-03 2021-04 2021-05 2021-06 2021-07 2021-08 \
2021-09 2021-10 2021-11 2021-12 2022-01 2022-02 2022-03 2022-04; do
curl -O "https://repo.stackable.tech/repository/misc/ny-taxi-data/yellow_tripdata_$month.parquet"
mc alias set minio "$MINIO_ENDPOINT" "$MINIO_ACCESS_KEY" "$MINIO_SECRET_KEY"
mc cp "yellow_tripdata_$month.parquet" minio/demo/ny-taxi-data/raw/
done
```

A second run with a bit more debug output

```sql
CREATE TABLE hive.demo.ny_taxi_data_raw6 [..] -- same statement as for ny_taxi_data_raw5
-- All NULL - expected
SHOW STATS FOR hive.demo.ny_taxi_data_raw6;
-- Works
ANALYZE hive.demo.ny_taxi_data_raw6;
[...]

Issue #5859, opened 2020-11-06
Add support for Hive min/max statistics for `TIMESTAMP` statistics in Hive


losipiuk:
Add proper support to min/max stats reading and writing in Hive for columns of type `TIMESTAMP(p)`

losipiuk:
Before working on this one we should do https://github.com/prestosql/presto/issues/5860

PR #26338, opened 2025-08-04
Added support for the timestamp type for statistics in trino-hive connector

anton-kutuzov:
## Description
Added support for the `timestamp` type in statistics calculation.
To maintain compatibility with `Hive` statistics, the computed `min` and `max` values are truncated to seconds during write by dividing by `MICROSECONDS_PER_SECOND`, using `floor` for `min` and `ceil` for `max` to avoid losing millisecond precision. When reading, values are converted back to Trino format by multiplying by `MICROSECONDS_PER_SECOND.`

Related [issue](https://github.com/trinodb/trino/issues/5860)

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

Issue #5860, opened 2020-11-06
Add support for lowValue and highValue statistics for TIMESTAMP type in engine


losipiuk:
Currenly lowValue, highValue statistics for `TIMESTAMP` type are not supported.
See:
https://github.com/prestosql/presto/blob/3c8cc38ce567cba6550b80c761e39fbe90a8df05/presto-main/src/main/java/io/prestosql/cost/StatsUtil.java#L40-L65

ericxiao251:
Hey team I would like to work on this issue, but I would like some guidance on where to start.

I started by looking at the tests for similar column stats ([date column stat tests](https://github.com/trinodb/trino/blob/master/plugin/trino-hive/src/test/java/io/trino/plugin/hive/metastore/thrift/TestThriftMetastoreUtil.java#L251-L289)) and I noticed that we do not have any tests for the timestamp column even though we do calculate some stats [here](https://github.com/trinodb/trino/blob/master/plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java#L956-L959) for the timestamp.

I also noticed we do not have related data classes for timestamp columns, we have `metastore.api.Date` and `metastore.api.DateColumnStatsData` but there is no `metastore.api.TimestampColumnStatsData` class. I think the first thing to address would to be add tests for the existing timestamp stats which require also adding a new `classmetastore.api.TimestampColumnStatsData`?

losipiuk:
The date tests you reference above are just for testing conversion between Hive's and Trino's statistics model. Hive uses `DateColumnStatsData` for that.
For TIMESTAMP columns though Hive uses  `LongColumnStatsData` (see [here](https://github.com/trinodb/trino/blob/c0276cea688da155f7f5bf0cb0b53169ece43e6e/plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java#L821-L822)). Unit tests for that class are [here](https://github.com/trinodb/trino/blob/96692e03c3c5abe88be5a0398745a43b722ccce6/plugin/trino-hive/src/test/java/io/trino/plugin/hive/metastore/thrift/TestThriftMetastoreUtil.java#L73).

[...]

PR #28296, opened 2026-02-14
Add support for timestamp column statistics in Hive 3 and 4

anton-kutuzov:
## Description
Added support for the `timestamp` type in statistics calculation.
To maintain compatibility with `Hive` statistics, the computed `min` and `max` values are truncated to seconds during write by dividing by `MICROSECONDS_PER_SECOND`, using `floor` for `min` and `ceil` for `max` to avoid losing millisecond precision. When reading, values are converted back to Trino format by multiplying by `MICROSECONDS_PER_SECOND.`
This solution supports both Hive Metastore version 3 and version 4.

- Related issues: https://github.com/trinodb/trino/issues/5860, https://github.com/trinodb/trino/pull/26338
- Fixes: https://github.com/trinodb/trino/issues/5859
- Fixes https://github.com/trinodb/trino/issues/26214 
- Fixes https://github.com/trinodb/trino/issues/28330

## Release notes

```markdown
## Section
```


plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java
```diff
@@ -435,8 +435,7 @@ static void main()
                             .setSchema("tpch")
                             .build())
                     .addCoordinatorProperty("http-server.http.port", "8080")
-                    .addHiveProperty("hive.metastore", "glue")
-                    .addHiveProperty("hive.metastore.glue.default-warehouse-dir", "local:///glue")
+                    .addHiveProperty("hive.metastore.uri", "thrift://hive-metastore:9083")
```
ebyhr: Revert unrelated change. This is Hive**Glue**QueryRunnerMain.

testing/trino-tests/src/test/java/io/trino/tests/tpch/TpchQueryRunner.java
```diff
@@ -83,6 +85,13 @@ static void main()
                         .buildOrThrow())
                 .withProtocolSpooling("json+zstd")
                 .build();
+
+        queryRunner.installPlugin(new MemoryPlugin());
+        queryRunner.createCatalog("memory", "memory");
+
+        queryRunner.installPlugin(new HivePlugin());
+        queryRunner.createCatalog("hive", "hive");
+
```
ebyhr: Revert. When a new plugin is required in the test class, please install it via `AbstractTestQueryFramework#createQueryRunner` method.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2326,6 +2326,46 @@ public void testUnsupportedDropSchemaCascadeWithNonHiveTable()
...
     }
 
+    @Test
+    public void testSupportTimestampStatisticsForHive3andHive4()
+    {
+        Session session = Session.builder(getQueryRunner().getDefaultSession())
+                .setCatalogSessionProperty("hive", "insert_existing_partitions_behavior", "APPEND")
+                .setCatalogSessionProperty("hive", "collect_column_statistics_on_write", "true")
+                .build();
+        String testTable = HIVE_TEST_SCHEMA + ".test_tabble_" + randomNameSuffix();
```
ebyhr: Fix typo > tabble

ebyhr (commented):
Just skimmed.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/HiveQueryRunner.java
anton-kutuzov: Thanks. Fixed

testing/trino-tests/src/test/java/io/trino/tests/tpch/TpchQueryRunner.java
anton-kutuzov: Thanks. Fixed

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Thanks. Fixed

findepi:
> Related issues: #5860, #26214, #26338

Is this PR "fixes https://github.com/trinodb/trino/issues/26214"?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
```diff
@@ -610,6 +631,14 @@ public static OptionalLong fromMetastoreNullsCount(long nullsCount)
         return OptionalLong.of(nullsCount);
     }
 
+    private static OptionalLong fromMetastoreTimestamp(Timestamp timestamp)
+    {
+        if (timestamp == null) {
+            return OptionalLong.empty();
+        }
+        return OptionalLong.of(LocalDateTime.ofEpochSecond(timestamp.getSecondsSinceEpoch(), 0, ZoneOffset.UTC).toInstant(ZoneOffset.UTC).toEpochMilli() * MICROSECONDS_PER_MILLISECOND);
```
findepi: Could this be done simpler?

```suggestion
        return OptionalLong.of(timestamp.getSecondsSinceEpoch() * MICROSECONDS_PER_SECOND);
```

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
```diff
@@ -820,6 +850,18 @@ private static ColumnStatisticsObj createLongStatistics(String columnName, HiveT
         return new ColumnStatisticsObj(columnName, columnType.toString(), longStats(data));
     }
 
+    private static ColumnStatisticsObj createTimestampStatistics(String columnName, HiveType columnType, HiveColumnStatistics statistics)
+    {
+        LongColumnStatsData data = new LongColumnStatsData();
+        statistics.getIntegerStatistics().ifPresent(timestampStatistics -> {
+            timestampStatistics.getMin().ifPresent(value -> data.setLowValue(floorDiv(value, MICROSECONDS_PER_SECOND)));
+            timestampStatistics.getMax().ifPresent(value -> data.setHighValue(ceilDiv(value, MICROSECONDS_PER_SECOND)));
```
findepi, line 884: ❤️ 

Although, stats are only estimates, so ceil div is not strictly necessary. 
In fact, it may cause some planning issues (a partition with single value will have min and max different).
My intuition is that, the code as currently written is a better way, so no changes requested.

cc @raunaqmorarka @sopel39

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
```diff
@@ -894,8 +936,7 @@ public static Set<HiveColumnStatisticType> getSupportedColumnStatistics(Type typ
             return ImmutableSet.of(MIN_VALUE, MAX_VALUE, NUMBER_OF_DISTINCT_VALUES, NUMBER_OF_NON_NULL_VALUES);
         }
         if (type instanceof TimestampType || type instanceof TimestampWithTimeZoneType) {
-            // TODO (https://github.com/trinodb/trino/issues/5859) Add support for timestamp MIN_VALUE, MAX_VALUE
```
findepi, line 898: Please add "fixes #5859" to the PR description.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/statistics/AbstractHiveStatisticsProvider.java
```diff
@@ -896,6 +900,14 @@ private static Optional<DoubleRange> createDateRange(DateStatistics statistics)
         return Optional.empty();
     }
 
+    private static Optional<DoubleRange> createTimestampRange(IntegerStatistics statistics)
+    {
+        if (statistics.getMin().isPresent() && statistics.getMax().isPresent()) {
+            return Optional.of(new DoubleRange(statistics.getMin().getAsLong(), statistics.getMax().getAsLong()));
```
findepi, line 906: Shouldn't this multiply by `MICROSECONDS_PER_SECOND`?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
```diff
@@ -288,6 +296,19 @@ private static Optional<LocalDate> getDateValue(Type type, Block block)
         return Optional.of(LocalDate.ofEpochDay(days));
     }
 
+    private static OptionalLong getTimestampValue(Type type, Block block)
+    {
+        verify(type instanceof TimestampType, "Unsupported type: %s", type);
+        if (block.isNull(0)) {
+            return OptionalLong.empty();
+        }
+        if (block instanceof Fixed12Block) {
```
findepi, line 311: Don't inspect the block type. Use the type API to get the value

```java
if (((TimestampType) type).isShort()) {
            return OptionalLong.of(type.getLong(block, 0));
        }
        else {
            return OptionalLong.of(((LongTimestamp) type.getObject(block, 0)).getEpochMicros());
        }
```

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
```diff
@@ -288,6 +296,19 @@ private static Optional<LocalDate> getDateValue(Type type, Block block)
         return Optional.of(LocalDate.ofEpochDay(days));
     }
 
+    private static OptionalLong getTimestampValue(Type type, Block block)
```
findepi: It doesn't read the timestamp value, only part of it. Call it `getTimestampEpochMicro`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
```diff
@@ -240,10 +246,12 @@ else if (type.equals(DOUBLE) || type.equals(REAL)) {
         else if (type.equals(DATE)) {
             result.setDateStatistics(new DateStatistics(getDateValue(type, min), getDateValue(type, max)));
         }
+        else if (type instanceof TimestampType) {
+            result.setIntegerStatistics(new IntegerStatistics(getTimestampValue(type, min), getTimestampValue(type, max)));
```
findepi: Don't you need to divide by MICROSECONDS_PER_SECOND?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
```diff
@@ -894,8 +936,7 @@ public static Set<HiveColumnStatisticType> getSupportedColumnStatistics(Type typ
...
-            return ImmutableSet.of(NUMBER_OF_DISTINCT_VALUES, NUMBER_OF_NON_NULL_VALUES);
+            return ImmutableSet.of(MIN_VALUE, MAX_VALUE, NUMBER_OF_DISTINCT_VALUES, NUMBER_OF_NON_NULL_VALUES);
```
findepi, line 965: This PR should change behavior for `type instanceof TimestampType` case, but not the `TimestampWithTimeZoneType` case. Split the `if`

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveTableStatistics.java
```diff
@@ -129,7 +129,7 @@ private static List<Row> getAllTypesTableStatistics()
                 row("c_double", null, 2.0, 0.0, null, "234.561", "235.567"),
                 row("c_decimal", null, 2.0, 0.0, null, "345.0", "346.0"),
                 row("c_decimal_w_params", null, 2.0, 0.0, null, "345.671", "345.678"),
-                row("c_timestamp", null, 2.0, 0.0, null, null, null),
+                row("c_timestamp", null, 2.0, 0.0, null, "2015-05-10 12:15:31.000", "2015-05-10 12:15:36.000"),
```
findepi, line 132: Was this table analyzed by hive?

(How do we verify hive timestamp stats being expressed in seconds in epoch?)

anton-kutuzov:
> > Related issues: #5860, #26214, #26338
> 
> Is this PR "fixes #26214"?

Yes, the [test](https://github.com/trinodb/trino/pull/28296/changes#diff-c6ac76ce5a102f34fe5cdca7794cff7404bf8ed13740ae7232c25ad6205a1245R2348) was added for this case

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
anton-kutuzov: Yes, I had that option, but I was getting an error `class io.trino.spi.block.Fixed12Block cannot be cast to class io.trino.spi.block.LongArrayBlock`, so I added a check on Fixed12Block

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: Done

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveTableStatistics.java
anton-kutuzov: If I understood the question correctly(does hive compute statistics for timestamp in seconds), then yes, it also compute it in seconds: https://github.com/apache/hive/blob/46f27831/ql/src/java/org/apache/hadoop/hive/ql/stats/ColumnStatisticsObjTranslator.java#L303-L310

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
anton-kutuzov: No, because we do it only before save to metastore
https://github.com/trinodb/trino/blob/master/plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftHiveMetastore.java#L570

plugin/trino-hive/src/main/java/io/trino/plugin/hive/statistics/AbstractHiveStatisticsProvider.java
anton-kutuzov: No, we multiply one time, when read from metastore.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: I used `floorDiv` and `ceilDiv` because I needed them to implement partition filtering based on statistics in the Hive catalog. With this implementation, I may include two extra partitions, which is not critical, but I won’t miss any.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: Done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: Thanks. Fixed

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2326,6 +2344,46 @@ public void testUnsupportedDropSchemaCascadeWithNonHiveTable()
...
+                "    passenger_count DOUBLE,\n" +
+                "    trip_distance DOUBLE,\n" +
+                "    payment_type BIGINT,\n" +
+                "    Fare_amount DOUBLE,\n" +
+                "    Tip_amount DOUBLE,\n" +
+                "    Total_amount DOUBLE\n" +
+                ")");
+        assertUpdate(session, "INSERT INTO " + testTable + " VALUES(1, TIMESTAMP '2025-09-19 10:00:00', TIMESTAMP '2025-09-19 10:10:00', 2, 3.5, 1, 10.0, 2.0, 12.0)", 1);
+        assertUpdate(session, "ANALYZE " + testTable, 1);
+        assertUpdate(session, "ANALYZE " + testTable, 1);
```
chenjian2664: Is it necessary to analyze twice?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2326,6 +2344,46 @@ public void testUnsupportedDropSchemaCascadeWithNonHiveTable()
         }
     }
 
+    @Test
+    public void testSupportTimestampStatisticsForHive3andHive4()
```
chenjian2664: "ForHive3andHive4" suffix looks redundant

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2326,6 +2344,46 @@ public void testUnsupportedDropSchemaCascadeWithNonHiveTable()
...
+                "    tpep_pickup_datetime TIMESTAMP,\n" +
+                "    tpep_dropoff_datetime TIMESTAMP,\n" +
+                "    passenger_count DOUBLE,\n" +
+                "    trip_distance DOUBLE,\n" +
+                "    payment_type BIGINT,\n" +
+                "    Fare_amount DOUBLE,\n" +
+                "    Tip_amount DOUBLE,\n" +
+                "    Total_amount DOUBLE\n" +
+                ")");
+        assertUpdate(session, "INSERT INTO " + testTable + " VALUES(1, TIMESTAMP '2025-09-19 10:00:00', TIMESTAMP '2025-09-19 10:10:00', 2, 3.5, 1, 10.0, 2.0, 12.0)", 1);
```
chenjian2664: nit: let two timestamp value different(though there is only 1 row)

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2326,6 +2344,46 @@ public void testUnsupportedDropSchemaCascadeWithNonHiveTable()
...
     }
 
+    @Test
+    public void testSupportTimestampStatisticsForHive3andHive4()
+    {
+        Session session = Session.builder(getQueryRunner().getDefaultSession())
+                .setCatalogSessionProperty("hive", "insert_existing_partitions_behavior", "APPEND")
+                .setCatalogSessionProperty("hive", "collect_column_statistics_on_write", "true")
+                .build();
+        String testTable = HIVE_TEST_SCHEMA + ".test_table_" + randomNameSuffix();
```
chenjian2664: test_table_ -> test_timestamp_statistics

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -319,6 +319,24 @@ public void testUpdateStatisticInsertOverwritePartitionedTable()
...
+                .build();
+        String tableName = "test_statistic" + randomNameSuffix();
+        String testTable = getFullyQualifiedTestTableName(tableName);
+        computeActual(session, "CREATE TABLE " + testTable + "(" +
+                                           "    history_batch_timestamp TIMESTAMP, " +
+                                           "    dt VARCHAR  ) " +
+                                    "WITH (partitioned_by=ARRAY['dt'])");
+
+        assertUpdate(session, "INSERT INTO " + testTable + "(history_batch_timestamp, dt) VALUES (CAST(CURRENT_TIMESTAMP AS TIMESTAMP), '2026-02-03')", 1);
+        assertUpdate(session, "INSERT INTO " + testTable + "(history_batch_timestamp, dt) VALUES (CAST(CURRENT_TIMESTAMP AS TIMESTAMP), '2026-02-03')", 1);
```
chenjian2664: Is the second insert operation for testing the overwrite behavior? if so, can we add a test i,e for value count about it?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
chenjian2664: Seems not?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -319,6 +319,24 @@ public void testUpdateStatisticInsertOverwritePartitionedTable()
...
+    @Test
+    public void testUpdateStatisticInsertOverwritePartitionedTableTimestamp()
+    {
+        Session session = Session.builder(getQueryRunner().getDefaultSession())
+                .setCatalogSessionProperty("hive", "insert_existing_partitions_behavior", "OVERWRITE")
+                .setCatalogSessionProperty("hive", "collect_column_statistics_on_write", "true")
+                .build();
+        String tableName = "test_statistic" + randomNameSuffix();
+        String testTable = getFullyQualifiedTestTableName(tableName);
+        computeActual(session, "CREATE TABLE " + testTable + "(" +
```
chenjian2664:
```suggestion
        assertUpdate(session, "CREATE TABLE " + testTable + "(" +
```

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Yes, we are checking this case https://github.com/trinodb/trino/issues/26214

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Thanks. Done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: Oh, you are right. I forgot to puth it. Done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
chenjian2664: Please add a comment for it

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
raunaqmorarka: > I needed them to implement partition filtering based on statistics in the Hive catalog

Since statistics are not guaranteed to be up to date in Hive, I don't think we'd accept changes in Trino to do partition filtering based on Hive table statistics.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
chenjian2664: reminder

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Yes, we testing the overwrite. Added comment. This is related with issue https://github.com/trinodb/trino/issues/28330

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Thanks. Done

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: Thanks. Done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: > Since statistics are not guaranteed to be up to date in Hive, I don't think we'd accept changes in Trino to do partition filtering based on Hive table statistics.

I know, I saw discussion in slack about statistics. But we in our production guarantee that statistics are always recalculated after changes and for this reason, we remove redundant partitions in io.trino.plugin.hive.HiveMetadata#applyFilter. I understand that you do not accept changes partition filtering and I will not contribute it.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
```diff
@@ -6672,7 +6672,7 @@ public void testCollectStatisticsOnCreateTableTimestampWithPrecision()
 
             assertQuery("SHOW STATS FOR " + tableName,
                     "SELECT * FROM VALUES " +
-                            "('c_timestamp', null, 9.0, 0.0, null, null, null), " +
+                            "('c_timestamp', null, 9.0, 0.0, null, '1988-04-08 02:03:04.111', '1988-04-08 02:03:04.119'), " +
```
raunaqmorarka, line 6675: All the tests are asserting on fractional second values, while the code is supposed to truncate it to seconds. I think this is because of discrepancy between how column statistics are handled in the thrift metastore implementation vs the file hive metastore implementation used in testing.
This is what co-pilot is complaining about as well

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveTableStatistics.java
raunaqmorarka: In product tests, we should verify that timestamp statistics produced by `COMPUTE STATISTICS FOR COLUMNS` in Hive is compatible with Trino and vice versa.

raunaqmorarka, line 132: Update the test timestamp values to have a non-zero fractional part, so that we can verify whether truncation to seconds is happening or not.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
raunaqmorarka: I've pushed a code comment explaining why the `block instanceof Fixed12Block` had to be used

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveTableStatistics.java
anton-kutuzov: Could you please explain your idea in more details? In method `getAllTypesTableStatistics()` we have expected values, and we expect that timestamp will be truncated to second. In test data [ALL_TYPES_TABLE](https://github.com/trinodb/trino/pull/28296/changes#diff-2bc8defb03cb680fc16eb09b44c4e82823fd09eac70c84e16fa6a3a13da60a2eR95) we have value `2015-05-10 12:15:31.123456`. So we can see that timestamp is truncated to second. Or do we need to test something else?

raunaqmorarka: Ah, then it already does what i wanted

anton-kutuzov: [Here](https://github.com/trinodb/trino/pull/28296/changes#diff-2bc8defb03cb680fc16eb09b44c4e82823fd09eac70c84e16fa6a3a13da60a2eR532) we verify that after `COMPUTE STATISTICS FOR COLUMNS` we have correct statistics

wendigo:
Hive failures are relevant

wendigo:
`ColumnStatisticsObj.getColType()` is marked as `@org.apache.thrift.annotation.Nullable`

wendigo:
@raunaqmorarka PTAL

wendigo:
@coderabbitai full review

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveTableStatistics.java
raunaqmorarka: Yes, it would be nice to cover the other direction as well. But I see that we haven't had that testing so far, so I won't require it here.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
findepi: do we have tests for this with thrift HMS too?

raunaqmorarka: Yes, the product tests at io.trino.tests.product.hive.TestHiveTableStatistics are testing with thrift and we can see that fractional seconds are truncated as expected there

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
```diff
@@ -517,12 +523,32 @@ public static Partition fromMetastoreApiPartition(io.trino.hive.thrift.metastore
...
     {
         if (columnStatistics.getStatsData().isSetLongStats()) {
             LongColumnStatsData longStatsData = columnStatistics.getStatsData().getLongStats();
             OptionalLong min = longStatsData.isSetLowValue() ? OptionalLong.of(longStatsData.getLowValue()) : OptionalLong.empty();
             OptionalLong max = longStatsData.isSetHighValue() ? OptionalLong.of(longStatsData.getHighValue()) : OptionalLong.empty();
+            // Hive 3 stores TIMESTAMP statistics as LongStats (seconds since epoch)
+            // Convert from seconds to microseconds for Trino
+            if (TIMESTAMP.equals(columnStatistics.getColType())) {
+                if (min.isPresent()) {
+                    min = OptionalLong.of(min.getAsLong() * MICROSECONDS_PER_SECOND);
```
findepi, line 546: Does this work when Hive session property timestamp_precision is set to NANOSECONDS?
Or should we be constructing min/max as a `LongTimestamp` object?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
```diff
@@ -587,6 +613,16 @@ public static HiveColumnStatistics fromMetastoreApiColumnStatistics(ColumnStatis
                     averageColumnLength,
                     nullsCount);
         }
+        // Hive 4+ stores TIMESTAMP statistics using dedicated TimestampColumnStatsData
+        // Values are stored as seconds since epoch and converted to microseconds
+        if (columnStatistics.getStatsData().isSetTimestampStats()) {
+            TimestampColumnStatsData timestampStatsData = columnStatistics.getStatsData().getTimestampStats();
+            OptionalLong min = timestampStatsData.isSetLowValue() ? fromMetastoreTimestamp(timestampStatsData.getLowValue()) : OptionalLong.empty();
+            OptionalLong max = timestampStatsData.isSetHighValue() ? fromMetastoreTimestamp(timestampStatsData.getHighValue()) : OptionalLong.empty();
```
findepi, line 621: same as https://github.com/trinodb/trino/pull/28296/changes#r2983280998

> Does this work when Hive session property timestamp_precision is set to NANOSECONDS?
Or should we be constructing min/max as a `LongTimestamp` object?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
findepi: The stats representation is micros for long timestamp too
https://github.com/trinodb/trino/blob/2576043b6b361256eae269776760a5604f5bdf8b/core/trino-spi/src/main/java/io/trino/spi/statistics/StatsUtil.java#L76
so maybe we're good.

Can we have a test for this?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
```diff
@@ -288,6 +296,24 @@ private static Optional<LocalDate> getDateValue(Type type, Block block)
...
+    {
+        verify(type instanceof TimestampType, "Unsupported type: %s", type);
+        if (block.isNull(0)) {
+            return OptionalLong.empty();
+        }
+        // TODO: The columnTypes map in HiveMetadata.finishChangingTable is built using the deprecated
+        //   getTypeSignature(HiveType) which always resolves TIMESTAMP to milliseconds,
+        //   ignoring the session's hive.timestamp-precision setting. The statistics
+        //   blocks however are computed by the engine using the session's precision.
+        //   The fix is to pass the session's timestamp precision when building columnTypes in HiveMetadata.
```
findepi: Can we have an issue for this maybe? (and link back here)

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
```diff
@@ -288,6 +296,24 @@ private static Optional<LocalDate> getDateValue(Type type, Block block)
...
+        if (block instanceof Fixed12Block) {
+            LongTimestamp timestamp = (LongTimestamp) TIMESTAMP_NANOS.getObject(block, 0);
+            return OptionalLong.of(timestamp.getEpochMicros());
```
findepi, line 314: This truncates (also for max). It's probably OK, stats are only estimates after all, but maybe worth a code comment

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
findepi: Then I think we're good here. it's OK for file metastore not to be exact behavioral equivalent of the thrift metastore.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -319,6 +319,56 @@ public void testUpdateStatisticInsertOverwritePartitionedTable()
...
+                .build();
+        String tableName = "test_statistic_overwrite_timestamp" + randomNameSuffix();
+        String testTable = getFullyQualifiedTestTableName(tableName);
+        assertUpdate(session, "CREATE TABLE " + testTable + "(" +
+                "    history_batch_timestamp TIMESTAMP, " +
+                "    dt VARCHAR  ) " +
+                "WITH (partitioned_by=ARRAY['dt'])");
+
+        // We check that overwrite works
+        // Related with https://github.com/trinodb/trino/issues/28330
```
findepi: Is this `// Regression test for https://github.com/trinodb/trino/issues/28330` , or only related to that issue?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
anton-kutuzov: Do we need to rewrite the file hive metastore implementation that used in tests?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2326,6 +2376,52 @@ public void testUnsupportedDropSchemaCascadeWithNonHiveTable()
...
+                "    payment_type BIGINT,\n" +
+                "    Fare_amount DOUBLE,\n" +
+                "    Tip_amount DOUBLE,\n" +
+                "    Total_amount DOUBLE\n" +
+                ")");
+        assertUpdate(session, "INSERT INTO " + testTable + " VALUES" +
+                "(1, TIMESTAMP '2025-09-19 11:00:00', TIMESTAMP '2025-09-19 10:10:00', 2, 3.5, 1, 10.0, 2.0, 12.0)," +
+                "(2, TIMESTAMP '2025-09-19 12:30:00', TIMESTAMP '2025-09-20 15:40:10', 4, 8.0, 2, 12.0, 5.0, 19.0)", 2);
+        // We should call analyze twice to check that a second call does not cause an exception
+        // Relate with https://github.com/trinodb/trino/issues/26214
```
findepi: `// Call ANALYZE twice as a regression test for https://github.com/trinodb/trino/issues/26214` ?

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
findepi: i don't think so. It will always behave at least slightly different from HMS anyway.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
anton-kutuzov: Created [issue](https://github.com/trinodb/trino/issues/28835) and added to comment

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/thrift/ThriftMetastoreUtil.java
anton-kutuzov: I tested the case with NANOSECONDS, and I also added a test just now.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: We checking bug 26214

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
anton-kutuzov: We checking bug 28330

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
findepi: changed comment myself

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
findepi: changed comment myself

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/Statistics.java
findepi: added a comment

anton-kutuzov: Thanks

Issue #26930, opened 2025-10-14
Elasticsearch connector does not support a path prefix


mjlshen:
There doesn't seem to be away to configure (specify a path prefix) for an elasticsearch connection to an endpoint like: "http://example.com:80/elasticsearch"

ebyhr:
cc: @martint

Praveen2112:
@mjlshen Is there any specific reason for specifying a path prefix ?

mjlshen:
That's where the elasticsearch instance I'm trying to connect to is hosted at is all. I'm able to provide `elasticsearch+http://example.com:80/elasticsearch` to other tools and query data (e.g. Superset), so I'm confident it would probably work in Trino as well, but I can't seem to pass the configuration values to enable that.

raphaelsolarski:
Hi, could I take this task on myself?

DHRUV6029:
@Praveen2112 @ebyhr can i work on this

daviddallakyan2005:
Taking this over from @mjlshen's #28300, which stale-bot closed after review. The property is still missing; no successor is open (#29534 was closed as a duplicate).

On the 405 question from that PR: the test set `elasticsearch.path-prefix=/test-prefix` against an Elasticsearch that is not behind a prefix, so the client requested `/test-prefix/orders/_mappings`. Elasticsearch 7+ treats `/{index}/{type}/_mappings` as a removed typed URL and returns 405 Method Not Allowed. That status was ES routing, not a reverse-proxy check. The new PR asserts the request path on a stub HTTP server instead.

#30733

PR #28300, opened 2026-02-14
Add support for path prefix in Elasticsearch connector

mjlshen:
## Description
This commit introduces support for configuring a path prefix when connecting to an Elasticsearch
instance. This feature is particularly useful when Elasticsearch is deployed behind a reverse proxy
that exposes it under a specific path, e.g. `http://example.com:80/elasticsearch`

#26930 

## Release notes

```markdown
## Elasticsearch Connector
* Add 'elasticsearch.path-prefix' configuration property to allow connecting to Elasticsearch instances through a reverse proxy.
```

### DIFF
--- a/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/ElasticsearchConfig.java
+++ b/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/ElasticsearchConfig.java
@@ -56,6 +56,7 @@ public class ElasticsearchConfig
 
     private List<String> hosts;
     private int port = 9200;
+    private String pathPrefix;
     private String defaultSchema = "default";
     private int scrollSize = 1_000;
     private Duration scrollTimeout = new Duration(1, MINUTES);
@@ -100,6 +101,19 @@ public class ElasticsearchConfig
     public ElasticsearchConfig setPort(int port)
     {
         this.port = port;
+        return this;
+    }
+
+    public Optional<String> getPathPrefix()
+    {
+        return Optional.ofNullable(pathPrefix);
+    }
+
+    @Config("elasticsearch.path-prefix")
+    @ConfigDescription("Path prefix for REST client")
+    public ElasticsearchConfig setPathPrefix(String pathPrefix)
+    {
+        this.pathPrefix = pathPrefix;
         return this;
     }
 
--- a/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
+++ b/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
@@ -206,6 +206,8 @@ public class ElasticsearchClient
                         .map(httpHost -> new HttpHost(httpHost, config.getPort(), config.isTlsEnabled() ? "https" : "http"))
                         .toArray(HttpHost[]::new));
 
+        config.getPathPrefix().ifPresent(builder::setPathPrefix);
+
         builder.setHttpClientConfigCallback(_ -> {
             RequestConfig requestConfig = RequestConfig.custom()
                     .setConnectTimeout(toIntExact(config.getConnectTimeout().toMillis()))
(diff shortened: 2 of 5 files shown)


plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
```diff
@@ -206,6 +206,8 @@ private static BackpressureRestHighLevelClient createClient(
                         .map(httpHost -> new HttpHost(httpHost, config.getPort(), config.isTlsEnabled() ? "https" : "http"))
                         .toArray(HttpHost[]::new));
 
+        config.getPathPrefix().ifPresent(builder::setPathPrefix);
```
ebyhr, line 209: Can we add a test? Is the test setup too complicated?

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/ElasticsearchConfig.java
```diff
@@ -77,6 +77,7 @@ public enum Security
     private boolean verifyHostnames = true;
 
     private Security security;
+    private String pathPrefix;
```
ebyhr: Let's move under `port` field. Same for the getter and setter.

mjlshen: Done!

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
mjlshen: I'm not super familiar with tests in this project, but I was able to come up with a relatively simple addition that verifies that this `pathPrefix` config property is being read/used correctly by the existing test (which fails as expected because the existing test server isn't setup to respond to a `pathPrefix`). What do you think?

mjlshen: It took me a bit to figure out how to implement that^, but I think I finally did 😅

chenjian2664:
@cla-bot check

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
```diff
@@ -148,11 +149,16 @@ public ElasticsearchClient(
...
         this.refreshInterval = config.getNodeRefreshInterval();
         this.tlsEnabled = config.isTlsEnabled();
+        this.pathPrefix = config.getPathPrefix();
     }
 
     @PostConstruct
     public void initialize()
     {
+        if (pathPrefix.isPresent()) {
+            return;
```
chenjian2664: This would prevent us from entering `refreshNodes`, so `nodes` would always be an empty set. Is that expected?

mjlshen: Thanks for catching this, no I missed that out of unfamiliarity with the codebase 😅. I removed this and added a  guard later to ensure that this works even if node discovery isn't exposed by the loadbalancer/proxy.

```java
                if (nodes.isEmpty()) {
                    shards.add(new Shard(chosen.getIndex(), chosen.getShard(), Optional.empty()));
                    continue;
                }
```

mjlshen:
cc @ebyhr or @chenjian2664 sorry for the ping, but do either of you have cycles for a re-review? Thanks!

cce-dsantos (approved):

mjlshen:
Maybe cc @martint ? I'm not sure which maintainer to reach out to, tagging you because @ebyhr added you on the original issue

plugin/trino-elasticsearch/src/test/java/io/trino/plugin/elasticsearch/BaseElasticsearchConnectorTest.java
```diff
@@ -139,6 +139,23 @@ public void testWithoutBackpressure()
...
+    public void testPathPrefix()
+            throws Exception
+    {
+        Map<String, String> properties = ImmutableMap.of("elasticsearch.path-prefix", "/test-prefix");
+
+        try (QueryRunner queryRunner = ElasticsearchQueryRunner.builder(server)
+                .addConnectorProperties(properties)
+                .build()) {
+            assertThatThrownBy(() -> queryRunner.execute("SELECT * FROM orders"))
+                    .hasMessageContaining("URI [/test-prefix/orders/_mappings], status line [HTTP/1.1 405 Method Not Allowed]");
```
chenjian2664, line 152: Please help me to understand the test, why it fails with 405?

PR #21609, opened 2024-04-18
Write checkpoint if necessary for CREATE OR REPLACE statements

belen-pruvost:
## Description
When running `CREATE [OR REPLACE] TABLE` statements for an existing table, the `finishCreate` method is currently not writing a checkpoint to the Delta Log. This PR makes it possible to write those checkpoints. It results in faster analysis times for queries that read from those tables.

## Additional context and related issues
We have noticed that tables recreated through this flow (`CREATE [OR REPLACE] TABLE`) have more than 2000 commits and no checkpoint, which leads to a very bad analysis time

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #28312, opened 2026-02-16
Write checkpoint if needed on CREATE OR REPLACE without schema change

### LINKED ISSUE #28310: CREATE OR REPLACE in Delta Lake not writing checkpoints as expected
This is a regression of #21609

Introduced in https://github.com/trinodb/trino/pull/27886

Specifically, here: https://github.com/trinodb/trino/pull/27886/changes#diff-79f367e47040dda32ac81ecec46b27cfd7cc822858bc8c31fe435a15a47446dcL1378-R1385

Pre-#27886, the behaviour was to always call `writeCheckpointIfNeeded`, which would write a checkpoint based on the checkpoint writing interval.

There was a bug in there: if the schema changed, we should write a checkpoint regardless of interval. The goal of #27886 was to fix this bug, but instead it changed the logic to write a checkpoint _only_ if the schema changed.

As a result, we're now not writing checkpoints on tables that get updated solely with CREATE AND REPLACE, which was originally fixed in #21609

This was not caught in CI because the unit tests introduced in #21609 unfortunately also did a schema change (!), which meant they were still passing after #27886.

Pluies:
## Description

Fixes regression in Delta Lake where we stopped writing checkpoints on CREATE OR REPLACE without a schema change. See #28310 for full explanation.

This PR has two commits:
- First commit improves the original test from #21609 to test without a schema change. These tests will fail, showing the issue written up in #28310.
- Second commit then fixes the underlying issue

I'm relying on tests included in #27886 to ensure the other use case of "Delta needs to write a checkpoint on schema change" works properly, e.g. [here](https://github.com/trinodb/trino/pull/27886/changes#diff-f8407debd31f7e168ce7c1259a228f6a5538c536d53f5e2c343df2f4e6fcf595R344).

I've taken the liberty of renaming `isNewCheckpointFileRequired` to specify exactly what it does, and including the schema-change check as part of `writeCheckpointIfNeeded`, which makes the logic easier to follow.

Fixes #28310

## Release notes

### DIFF
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
@@ -1382,9 +1382,7 @@ public class DeltaLakeMetadata
                 if (replaceExistingTable) {
                     List<DeltaLakeColumnHandle> existingColumns = getColumns(tableHandle.getMetadataEntry(), tableHandle.getProtocolEntry());
                     List<DeltaLakeColumnHandle> newColumns = getColumns(metadataEntry, protocolEntry);
-                    if (isNewCheckpointFileRequired(existingColumns, newColumns)) {
-                        writeCheckpoint(session, schemaTableName, location, tableHandle.toCredentialsHandle(), commitVersion);
-                    }
+                    writeCheckpointIfNeeded(session, schemaTableName, location, tableHandle.toCredentialsHandle(), tableHandle.getReadVersion(), checkpointInterval, commitVersion, Optional.of(existingColumns), Optional.of(newColumns));
                 }
             }
         }
@@ -1542,12 +1540,12 @@ public class DeltaLakeMetadata
 
         OptionalLong readVersion = OptionalLong.empty();
         ProtocolEntry protocolEntry;
+        Optional<List<DeltaLakeColumnHandle>> existingColumns = Optional.empty();
 
-        boolean isNewCheckpointFileRequired = false;
         if (replaceExistingTable) {
             protocolEntry = protocolEntryForTable(ProtocolEntry.builder(handle.getProtocolEntry()), containsTimestampType, tableMetadata.getProperties());
             readVersion = OptionalLong.of(handle.getReadVersion());
-            isNewCheckpointFileRequired = isNewCheckpointFileRequired(getColumns(handle.getMetadataEntry(), handle.getProtocolEntry()), columnHandles.build());
+            existingColumns = Optional.of(getColumns(handle.getMetadataEntry(), handle.getProtocolEntry()));
         }
         else {
             TrinoFileSystem fileSystem = fileSystemFactory.create(session, location);
@@ -1570,7 +1568,7 @@ public class DeltaLakeMetadata
                 columnMappingMode,
                 maxFieldId,
                 replace,
-                isNewCheckpointFileRequired,
+                existingColumns,
                 readVersion,
                 protocolEntry);
     }
@@ -1759,8 +1757,8 @@ public class DeltaLakeMetadata
             transactionLogWriter.flush();
             writeCommitted = true;
 
-            if (handle.replace() && handle.readVersion().isPresent() && handle.isSchemaChanged()) {
-                writeCheckpoint(session, schemaTableName, location, handle.toCredentialsHandle(), commitVersion);
+            if (handle.replace() && handle.readVersion().isPresent()) {
+                writeCheckpointIfNeeded(session, schemaTableName, handle.location(), handle.toCredentialsHandle(), handle.readVersion().getAsLong(), handle.checkpointInterval(), commitVersion, handle.existingColumns(), Optional.of(handle.inputColumns()));
             }
 
             if (isCollectExtendedStatisticsColumnStatisticsOnWrite(session) && !computedStatistics.isEmpty()) {
@@ -2323,7 +2321,7 @@ public class DeltaLakeMetadata
             long commitVersion = Failsafe.with(TRANSACTION_CONFLICT_RETRY_POLICY)
                     .get(context -> commitInsertOperation(session, handle, sourceTableHandles, isolationLevel, dataFileInfos, readVersion, context.getAttemptCount()));
             writeCommitted = true;
-            writeCheckpointIfNeeded(session, handle.tableName(), handle.location(), handle.credentialsHandle(), handle.readVersion(), handle.metadataEntry().getCheckpointInterval(), commitVersion);
+            writeCheckpointIfNeeded(session, handle.tableName(), handle.location(), handle.credentialsHandle(), handle.readVersion(), handle.metadataEntry().getCheckpointInterval(), commitVersion, Optional.empty(), Optional.empty());
             enqueueUpdateInfo(session, handle.tableName().getSchemaName(), handle.tableName().getTableName(), commitVersion, handle.metadataEntry().getSchemaString(), Optional.ofNullable(handle.metadataEntry().getDescription()));
 
             if (isCollectExtendedStatisticsColumnStatisticsOnWrite(session) && !computedStatistics.isEmpty() && !dataFileInfos.isEmpty()) {
@@ -2713,7 +2711,7 @@ public class DeltaLakeMetadata
                     handle.getMetadataEntry().getSchemaString(),
                     Optional.ofNullable(handle.getMetadataEntry().getDescription()));
 
-            writeCheckpointIfNeeded(session, handle.getSchemaTableName(), handle.getLocation(), handle.toCredentialsHandle(), handle.getReadVersion(), checkpointInterval, commitVersion);
+            writeCheckpointIfNeeded(session, handle.getSchemaTableName(), handle.getLocation(), handle.toCredentialsHandle(), handle.getReadVersion(), checkpointInterval, commitVersion, Optional.empty(), Optional.empty());
         }
         catch (RuntimeException e) {
             if (!writeCommitted) {
@@ -2995,7 +2993,9 @@ public class DeltaLakeMetadata
                     optimizeHandle.getCredentialsHandle(),
                     optimizeHandle.getCurrentVersion().orElseThrow(),
                     checkpointInterval,
-                    commitVersion);
+                    commitVersion,
+                    Optional.empty(),
+                    Optional.empty());
         }
         catch (Exception e) {
             if (!writeCommitted) {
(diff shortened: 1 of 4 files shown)


plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3181,23 +3181,29 @@ private void writeCheckpointIfNeeded(
             VendedCredentialsHandle credentialsHandle,
             long readVersion,
             Optional<Long> checkpointInterval,
-            long newVersion)
+            long newVersion,
+            boolean isSchemaChanged)
```
findinpath: It feels we're trying to unify here by adding `isSchemaChanged` the logic related to writing checkpoints.

However, note that besides doing `CREATE OR REPLACE` statements, the schema can change also through `ALTER TABLE` statements.


I just noticed that the we're not respecting the `checkpoint_interval` when working with such a scenario

```
trino> create schema delta.tiny with (location='s3://test-bucket/tiny');
CREATE SCHEMA
trino> create table delta.tiny.t1 (a integer) with (checkpoint_interval = 2) ;
CREATE TABLE
trino> alter table delta.tiny.t1 add column b integer;
ADD COLUMN
trino> alter table delta.tiny.t1 add column c integer;
ADD COLUMN
trino> alter table delta.tiny.t1 add column d integer;
ADD COLUMN
```
In the above example, there's no checkpoint created, although we've executed 4 statements (> checkpoint interval set to 2).

I doubt that this functionality is intended :( .
cc @ebyhr

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3214,7 +3220,7 @@ private void writeCheckpoint(ConnectorSession session, SchemaTableName table, St
         checkpointWriterManager.writeCheckpoint(session, snapshot, credentialsHandle);
     }
 
-    private static boolean isNewCheckpointFileRequired(List<DeltaLakeColumnHandle> existingColumns, List<DeltaLakeColumnHandle> newColumns)
```
findinpath, line 3219: Yes, we have missed here the check used to verify whether the checkpoint interval has been exceeded :( 

I see the reasoning why you opted for unifying into `writeCheckpointIfNeeded` method.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
Pluies: Good point @findinpath – another bug! Likely there's no call to `writeCheckpointIfNeeded` at all in ALTER TABLE 👀 

I'm all for simplifying/putting this all into a single call 👍

findinpath: Note that when calling `writeCheckpointIfNeeded` from alter table methods, we'll need to respect (probably) the `checkpoint_interval` rule and not the schema changed shortcircuit.

ebyhr:
Added a release blocker label since the regression comes from https://github.com/trinodb/trino/pull/27886 (v480).

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
Pluies: Makes sense. I'm thinking of adding checkpoints to ALTER TABLE in a separate PR, as it's technically a new feature, whereas this is a regression, to keep scope small on this PR and unblock the release of 480 – what do you think?

findinpath: Yes, let's proceed with the bugfix.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -1380,9 +1380,8 @@ public void createTable(ConnectorSession session, ConnectorTableMetadata tableMe
                 if (replaceExistingTable) {
                     List<DeltaLakeColumnHandle> existingColumns = getColumns(tableHandle.getMetadataEntry(), tableHandle.getProtocolEntry());
                     List<DeltaLakeColumnHandle> newColumns = getColumns(metadataEntry, protocolEntry);
```
findinpath, line 1384: Write checkpoint if needed on CREATE OR REPLACE without schema change -> Write checkpoint on CREATE OR REPLACE if checkpoint interval is exceeded

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
findinpath: Now a reader may think that if schema is changed there needs to happen always a new checkpoint write.
Can we use a more meaningful variable name here?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3181,23 +3181,29 @@ private void writeCheckpointIfNeeded(
...
+                }
 
-            // TODO: There is a race possibility here(https://github.com/trinodb/trino/issues/12004),
-            // which may result in us not writing checkpoints at exactly the planned frequency.
-            // The snapshot obtained above may already be on a version higher than `newVersion` because some other transaction could have just been committed.
-            // This does not pose correctness issue but may be confusing if someone looks into transaction log.
-            // To fix that we should allow for getting snapshot for given version.
-
-            writeCheckpoint(session, table, tableLocation, credentialsHandle, newVersion);
+                // TODO: There is a race possibility here(https://github.com/trinodb/trino/issues/12004),
```
findinpath: Let's do the writing in only one branch of the code.
You have now a call in 3192 and one in 3205 (and a large comment about concurrency which applies only to the call in 3205).

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
```diff
@@ -1463,27 +1463,27 @@ public void testCreateOrReplaceCheckpointing()
         assertThat(listCheckpointFiles(transactionLogDirectory)).isEmpty();
         assertQuery("SELECT * FROM " + tableName, "VALUES (1,'ala'),  (2,'kota'), (3, 'psa')");
 
-        // replace table
+        // replace table with same schema, expect a checkpoint
```
chenjian2664:
```suggestion
        // replace table with same schema, expect a checkpoint because the checkpoint interval threshold has been reached
```

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
```diff
@@ -1446,7 +1446,7 @@ public void testCheckpointWriteStatsAsStructWithPartiallyUnsupportedColumnStats(
     }
 
     @Test
-    public void testCreateOrReplaceCheckpointing()
+    public void testCreateOrReplaceSameSchemaWritesCheckpoint()
```
chenjian2664:
```suggestion
testCreateOrReplaceWithSameSchemaWritesCheckpoint
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: Indeed, we missed original writing the checkpoint logic in https://github.com/trinodb/trino/pull/27886

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
chenjian2664: The `isSchemaChanged` isn't accurate here, i,e the situation when adding a new column doesn't treated a checkpoint required situation, thus i think previous the method `isNewCheckpointFileRequired` would be better

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3214,7 +3220,7 @@ private void writeCheckpoint(ConnectorSession session, SchemaTableName table, St
         checkpointWriterManager.writeCheckpoint(session, snapshot, credentialsHandle);
     }
 
-    private static boolean isNewCheckpointFileRequired(List<DeltaLakeColumnHandle> existingColumns, List<DeltaLakeColumnHandle> newColumns)
+    private static boolean isSchemaChangedForCheckpoint(List<DeltaLakeColumnHandle> existingColumns, List<DeltaLakeColumnHandle> newColumns)
```
chenjian2664: I think the name is good and we should keep the original name

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3181,23 +3181,29 @@ private void writeCheckpointIfNeeded(
...
+            boolean isSchemaChanged)
     {
         try {
             // We are writing checkpoint synchronously. It should not be long lasting operation for tables where transaction log is not humongous.
             // Tables with really huge transaction logs would behave poorly in read flow already.
-            long lastCheckpointVersion = getLastCheckpointVersion(session, table, tableLocation, readVersion, credentialsHandle);
-            if (newVersion - lastCheckpointVersion < checkpointInterval.orElse(defaultCheckpointInterval)) {
-                return;
+            if (isSchemaChanged) {
+                // Always write a checkpoint on schema change
```
chenjian2664: This is not true, see the logic in `isNewCheckpointFileRequired`, we don't write a checkpoint in every schema change.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
findepi: Add this as a new test case.
There's definitely a value testing replace table with a schema change, so we should continue doing so.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
Pluies: You're correct that the name here isn't accurate 👍 However, I don't like `isNewCheckpointFileRequired` either because it's misleading in the other direction, implying that if it's false then a new checkpoint file is _not_ required... Which is not true! As it could still be required by the checkpoint interval.

I'm rewriting this in a new commit by:
- Passing (as `Optional`) the list of old & new columns into `writeCheckpointIfNeeded`, which stays the final arbiter of whether a new checkpoint is needed or not
- Renaming the helper method to a very explicit `columnChangeRequiresCheckpoint`

Will push this soon, running tests locally first 🤞

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
Pluies: I agree! Both parts need to be tested 👍 

This is already tested in https://github.com/trinodb/trino/pull/27886/changes#diff-f8407debd31f7e168ce7c1259a228f6a5538c536d53f5e2c343df2f4e6fcf595R344 ; but I guess there's no harm to test it there as well? Adding as `testCreateOrReplaceWithDifferentSchemaWritesCheckpoint`

Pluies:
@findepi @findinpath @chenjian2664 hey folks, sorry about the delay, I've incorporated your feedback in two new commits:
- https://github.com/trinodb/trino/pull/28312/commits/eb7ee03673a8a0a7a9e81c0d070ee17103fec2a9
- https://github.com/trinodb/trino/pull/28312/changes/bdb5aed3db545f2fadcc7f2a1a5dd080015d30d4

Hopefully this covers everything, ptal 🙏

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3184,21 +3184,26 @@ private void writeCheckpointIfNeeded(
...
+            Optional<List<DeltaLakeColumnHandle>> existingColumns,
+            Optional<List<DeltaLakeColumnHandle>> newColumns)
     {
         try {
             // We are writing checkpoint synchronously. It should not be long lasting operation for tables where transaction log is not humongous.
             // Tables with really huge transaction logs would behave poorly in read flow already.
-            if (isSchemaChanged) {
-                // Always write a checkpoint on schema change
-                writeCheckpoint(session, table, tableLocation, credentialsHandle, newVersion);
+            boolean checkpointNeeded = false;
```
chenjian2664: how about add a helper method similar to:
```

    private boolean isCheckpointNeeded(
            ConnectorSession session,
            SchemaTableName table,
            String tableLocation,
            VendedCredentialsHandle credentialsHandle,
            long readVersion,
            Optional<Long> checkpointInterval,
            long newVersion,
            Optional<List<DeltaLakeColumnHandle>> existingColumns,
            Optional<List<DeltaLakeColumnHandle>> newColumns)
    {
        // If columns have changed: we need to write a new checkpoint
        if (columnChangeRequiresCheckpoint(existingColumns, newColumns)) {
            return true;
        }

        // If we've reached the checkpoint writing interval: we need to write a checkpoint
        long lastCheckpointVersion = getLastCheckpointVersion(session, table, tableLocation, readVersion, credentialsHandle);
        return newVersion - lastCheckpointVersion >= checkpointInterval.orElse(defaultCheckpointInterval);
    }
```

Pluies: I like the idea here; on it!

Pluies: Extracted (including relevant comment) in https://github.com/trinodb/trino/pull/28312/changes/51d448704bf1e74a75617da064ef54ea7293a5a8 👍

chenjian2664:
@Pluies Please squash the commits

Issue #28318, opened 2026-02-16
Avoidable warning from Trino CLI used in a script


findepi:
```
client/trino-cli/target/trino-cli-480-SNAPSHOT-executable.jar <<EOF
SELECT DECIMAL '1';
EOF
Feb 16, 2026 7:53:55 PM org.jline.utils.Log logr
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
"1"
```

the part
```
Feb 16, 2026 7:53:55 PM org.jline.utils.Log logr
WARNING: Unable to create a system terminal, creating a dumb terminal (enable debug logging for more information)
```

doesn't need to be there.
It's typical for CLI tools to recognize whether output is a terminal and handle that without informing the user about it.

PR #28321, opened 2026-02-16
Create dumb terminal on non-interactive session

wendigo:
Fixes https://github.com/trinodb/trino/issues/28318

## Release notes

```markdown
## CLI
* Resolve warning generated when system terminal cannot be created ({issue}`issuenumber`)
```

### DIFF
--- a/client/trino-cli/src/main/java/io/trino/cli/TerminalUtils.java
+++ b/client/trino-cli/src/main/java/io/trino/cli/TerminalUtils.java
@@ -39,9 +39,21 @@ public class TerminalUtils
     private static Terminal createTerminal()
     {
         try {
-            return TerminalBuilder.builder()
-                    .name("Trino")
-                    .build();
+            try {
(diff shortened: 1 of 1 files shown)


client/trino-cli/src/main/java/io/trino/cli/TerminalUtils.java
```diff
@@ -39,9 +39,21 @@ public static Terminal getTerminal()
...
-                    .name("Trino")
-                    .build();
+            try {
+                return TerminalBuilder.builder()
+                        .system(true)
+                        .dumb(false)
+                        .name("Trino")
+                        .build();
+            }
+            catch (IllegalStateException e) {
```
findepi, line 49: is there equivalent of bash's `test -t 1` so that we could avoid having this exception driven?

wendigo: I couldn't find it

PR #28331, opened 2026-02-17
Add support for case sensitive identifiers

prrvchr:
## Description

This fix allows Trino to support identifiers in mixed cases.

As a result, now **Trino support two type of connector**: 
- connector witch does not support mixed case identifier (ie: legacy connector).
- connector witch support mixed case identifier.

For **legacy connectors**, everything remains unchanged except:
- Errors involving an identifier now return the identifier's canonicalized value, and no longer the value that was entered, as before.
- But perhaps there are differences that the tests have not yet revealed.

For **connectors supporting mixed case**, in addition to supporting mixed case, their operation differs from a legacy connector in the following ways:
- The metadata held by Trino is now exactly as delivered by the connectors, meaning it was not anymore converted to lowercase.
- All identifiers provided to Trino other than by an SQL query will be treated as delimited identifiers, that is to say without be canonicalized, unlike in legacy mode where they are converted to lowercase if the characters that compose them are within the set of characters describing an undelimited identifier.
- All `io.trino.sql.analyzer.Field` will be treated as delimited identifiers, meaning they will not be canonicalized, unlike in legacy mode where they are converted to lowercase.

Voila...

This is supposed to correct [issue#17](https://github.com/trinodb/trino/issues/17)
Supersedes: [PR#7837](https://github.com/trinodb/trino/pull/7837), [PR#2350](https://github.com/trinodb/trino/pull/2350) and [PR#26339](https://github.com/trinodb/trino/pull/26339)

## Release notes

```markdown
## Section
* Fix some things. ({issue}`17`)
```


core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -2755,6 +2754,9 @@ private List<Field> analyzeTableOutputFields(Table table, QualifiedObjectName ta
                         Optional.of(column.getName()),
                         false);
                 fields.add(field);
+                System.out.println("StatementAnalyser.analyzeTableOutputFields() column Name: " + column.getName());
+                System.out.println("StatementAnalyser.analyzeTableOutputFields() columnHandle.containsKey(): " + columnHandles.containsKey(column.getName()));
+                System.out.println("StatementAnalyser.analyzeTableOutputFields() columnHandle Keys: " + String.join(", ", columnHandles.keySet()));
                 ColumnHandle columnHandle = columnHandles.get(column.getName());
```
Praveen2112, line 2953: Can we remove the `System.out.println`

Praveen2112 (commented):
Thanks for working on this. I don't think removing the conversion to lower case alone would fix this issue. We need to introduce a contract between the engine and the connector - as on what would be its stand on case sensitive identifiers and convert them accordingly.

prrvchr:
> I don't think removing the conversion to lower case alone would fix this issue. We need to introduce a contract between the engine and the connector

If we want to be able to handle identifiers in mixed cases, then **Trino must become strict regarding character case matching on identifier**. And trying to offer more than that will only cause problems. No further conversions are performed by Trino.

The question is: Which connectors do not support mixed cases in their identifiers?

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
prrvchr, line 1: Does Druid `supportsMixedCaseIdentifiers`?

wendigo:
This is backward incompatible change hence for sure won't be accepted.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
prrvchr: Ok done.

prrvchr:
@wendigo How do you think this will be backward compatible? Trino doesn't differentiate between identifiers: `column`, `COLUMN`, or `CoLuMn`. If it's asked to differentiate between them, then it can't be backward compatible? Perhaps it's time to consider creating a new branch for this?

And if it's not going to be published for that reason, then you'll be stuck with issue 17 for a long time.

However, I'd like to know what the decision will be, because in that case, I'll fork and maintain only my connectors.
**Thank you for giving me a precise answer.**

core/trino-parser/src/main/java/io/trino/sql/tree/Identifier.java
```diff
@@ -78,11 +77,7 @@ public boolean isDelimited()
 
     public String getCanonicalValue()
     {
-        if (isDelimited()) {
-            return value;
-        }
-
-        return value.toUpperCase(ENGLISH);
```
Praveen2112, line 85: This is essential. As per SQL Spec - only delimited identifiers should be case sensitive i.e 
`colA` == `COLA` == `COlA` == `cola` (or upper case would be default mode)
`"colA"` == `"colA"` - i.e case sensitivity is considered only for delimited identifiers.

Praveen2112:
> Trino must become strict regarding character case matching on identifier.

Trino doesn't have to be strict regarding character case matching on identifier, it just needs to respect the underlying connector's decision.  Connector should be able to help on canonicalize the object name accordingly. Lets say for hive connector, it supports all object names is lower case so even it the user tries to run the query like `columnA`  or `ColumnA` or `"columnA"` - all of them should map to `columna`. Lets say for a JDBC based data source which supports case sensitive objects, connector should be able to provide how to map the name accordingly. Trino would pass an identifier and an information on whether it is delimited or not, it is the job of the connector to resolve it. 



> Which connectors do not support mixed cases in their identifiers?

A good example is hive, which doesn't support mixed case identifiers. 

> How do you think this will be backward compatible? 

Create views based on JDBC tables and storing them in Hive - is a pretty common use case. Enabling it directly to all the JDBC connector will make the views unusable.

core/trino-main/src/main/java/io/trino/execution/CreateTableTask.java
```diff
@@ -186,7 +185,7 @@ ListenableFuture<Void> internalExecute(CreateTable statement, Session session, L
                 if (type.equals(UNKNOWN)) {
                     throw semanticException(COLUMN_TYPE_UNKNOWN, element, "Unknown type '%s' for column '%s'", column.getType(), name);
                 }
-                if (columns.containsKey(name.getValue().toLowerCase(ENGLISH))) {
+                if (columns.containsKey(name.getValue())) {
```
Praveen2112: Name should be canonicalized based on the connectors expectation

prrvchr: If we want to be able to create tables with columns in mixed case, then this is mandatory.

Praveen2112: Yes - but the logic on whether
- `name.getValue().toLowerCase(ENGLISH)`
- `name.getValue().toUpperCase(ENGLISH)` 
- `name.getValue()`

Should be based on whether the Identifier is delimited and connector support case sensitive identifiers.

prrvchr: I agree.

prrvchr:
> Create views based on JDBC tables and storing them in Hive - is a pretty common use case. Enabling it directly to all the JDBC connector will make the views unusable.

@Praveen2112 If we want to maintain compatibility, as requested by @wendigo, then the connectors need to know whether an identifier is delimited or not. Currently, this information hasn't been relayed back to the connectors, it seems to me?

~~Okay, I just opened [issue#28356](https://github.com/trinodb/trino/issues/28356) and I'm going to submit a pull request just to fix this problem. Once that's done, I'll come back here.~~

**Edit:** I got lost, thanks to @mosabua for the advice which I followed in this latest version.

core/trino-main/src/main/java/io/trino/execution/CreateTableTask.java
prrvchr: The latest version supports this.

prrvchr:
This new version addresses @wendigo's request and is normally compatible with what is currently done in Trino.
However, two errors remain concerning the Iceberg connector:
- `TestIcebergRestCatalogNestedNamespaceConnectorSmokeTest.testView()`
- `TestIcebergRestCatalogNestedNamespaceConnectorSmokeTest.testCommentView()`

I need help fixing this.

**Edit**: This has been resolved and has nothing to do with Iceberg.

chenjian2664:
@prrvchr I think it is non trivial feature, could you please share your idea/thoughts briefly on pr description?

prrvchr:
> @prrvchr I think it is non trivial feature, could you please share your idea/thoughts briefly on pr description?

@chenjian2664 I'd like to explain in detail what was necessary to resolve this issue and the oversights in such an implementation.
**But I need guarantees regarding its publication**, as I can't waste time on a pull request that won't be published, and therefore I'll have to maintain it in my own fork.
Haven't you already done that with my SQLite connector?
I am ready for genuine collaboration, but not a one-way street; do what is necessary.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -263,11 +262,12 @@ public List<String> listSchemaNames(Session session, String catalogName)
         ImmutableSet.Builder<String> schemaNames = ImmutableSet.builder();
         if (catalog.isPresent()) {
             CatalogMetadata catalogMetadata = catalog.get();
+            boolean supportMixedCase = catalogMetadata.supportsMixedCase();
```
Praveen2112: I'm a bit afraid - if it would introduce some sort of a breaking change for connectors which were previously  working for JDBC based connectors like I mentioned their existing queries and views. I would envision them as encapsulating it as a config/session property which is disabled by default. The users could enable it share their feedback and then we could have it enabled by default.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -495,6 +498,8 @@ class StatementAnalyzer
         this.warningCollector = requireNonNull(warningCollector, "warningCollector is null");
         this.correlationSupport = requireNonNull(correlationSupport, "correlationSupport is null");
         this.functionResolver = plannerContext.getFunctionResolver(warningCollector);
+        this.nameCanonicalizer = getNameCanonicalizer();
+        this.supportsMixedCase = supportsMixedCase();
```
Praveen2112: Ideally the NameCanonicalizer should handle these stuffs right ? Why do we need supports mixed case ?

core/trino-main/src/main/java/io/trino/metadata/MetadataUtil.java
Praveen2112, line 1: But we also need to remove the lower case conversion in SPI side for SchemaTableName, CatalogSchemaName. Before merging this change, we should decouple the lower name conversion in SchemaTableName and other objects i.e conversion to lower case should happen in the engine side and connector side (initially) so as to decouple it from the SPI objects and then if we could introduce this change on the engine side and effectively on the connectors as well. 

cc: @martint WDYT ?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
prrvchr: > Why do we need supports mixed case ?

For handling `io.trino.sql.analyzer.Field`, who don't have delimiter. Two rules apply:
- If the connector supports mixed cases, it remains unchanged.
- If the connector does not support mixed cases, it is converted to lowercase.

see [issue#28440](https://github.com/trinodb/trino/issues/28440)

Praveen2112: But this could be embedded in the NameCanonicalizer right ?

core/trino-main/src/main/java/io/trino/metadata/MetadataUtil.java
prrvchr: > But we also need to remove the lower case conversion in SPI side for SchemaTableName, CatalogSchemaName

Of course, that's all that's left to raise. It will be done soon. For now, I've focused on integrating `NameCanonicalizer` and getting everything working.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
prrvchr: I'm not sure this is the best way to do it, but `supportsMixedCase` is necessary to handle `Field`. See my next answer.

core/trino-main/src/main/java/io/trino/metadata/MetadataUtil.java
Praveen2112: I would prefer it in the reverse direction though. First we could decouple it, then fix on the engine side and then  on the connector side. Bcoz now we don't have a way to test the behaviour on the changes in the engine side and then fixing it on the connector side (aka removing the lower case checks on the connector side) could break a few things.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
prrvchr: > But this could be embedded in the NameCanonicalizer right ?

I couldn't get it to work. Look at the three methods:
- `MetadataUtil.createQualifiedName()`
- `MetadataUtil.createQualifiedObjectName()`
- `MetadataUtil.createIdentifierParts()`

plugin/trino-pinot/src/main/java/io/trino/plugin/pinot/query/PinotSqlFormatter.java
```diff
@@ -283,7 +284,7 @@ public static String stripQuotes(String value)
 
     public static PinotColumnHandle getColumnHandle(String name, SchemaTableName schemaTableName, Map<String, ColumnHandle> columnHandles)
     {
-        PinotColumnHandle columnHandle = (PinotColumnHandle) columnHandles.get(name);
+        PinotColumnHandle columnHandle = (PinotColumnHandle) columnHandles.get(name.toLowerCase(ENGLISH));
```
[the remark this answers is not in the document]
prrvchr: This change come from [PR#26339](https://github.com/trinodb/trino/pull/26339). I don't known Pinot for giving the right answer.

core/trino-main/src/main/java/io/trino/metadata/MetadataUtil.java
```diff
@@ -162,48 +161,67 @@ private static String joinName(List<String> name)
...
+        ImmutableList.Builder<Identifier> partsBuilder = ImmutableList.builder();
+        Identifier identifier = parts.size() > 2 ? parts.get(2) : session.getCatalog().map(catalog -> new Identifier(catalog, false)).orElseThrow(() ->
                 semanticException(MISSING_CATALOG_NAME, node, "Catalog must be specified when session catalog is not set"));
-
-        return new QualifiedObjectName(catalogName, schemaName, objectName);
+        partsBuilder.add(identifier);
+        boolean supportsMixedCase = metadata.supportsMixedCase(session, Canonicalizer.LEGACY_CANONICALIZER.apply(identifier));
+        partsBuilder.add(parts.size() > 1 ? parts.get(1) : session.getSchema()
+                .map(schema -> new Identifier(schema, supportsMixedCase || !Identifier.isValidIdentifier(schema))).orElseThrow(() ->
+                semanticException(MISSING_SCHEMA_NAME, node, "Schema must be specified when session schema is not set")));
```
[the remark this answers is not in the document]
prrvchr: The logic is:
- If the connector supports mixed cases, then the identifiers are treated as delimited identifiers.
- Otherwise, we proceed as before.

For me, the copilot has lost its way...

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -598,33 +596,31 @@ private Query showCreateView(ShowCreate node)
...
                     node.getLocation().orElseThrow(),
-                    QualifiedName.of(ImmutableList.of(catalogName, schemaName, tableName)),
+                    createQualifiedName(session, node, node.getName(), metadata, canonicalizer),
                     query,
                     false,
                     viewDefinition.get().getComment(),
                     Optional.of(security),
                     propertyNodes))
                     .trim();
+            System.out.println("ShowQueriesRewrite.showCreateView() sql: " + sql);
```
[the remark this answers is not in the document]
prrvchr: Done

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -3642,6 +3688,7 @@ protected Scope visitUpdate(Update update, Optional<Scope> scope)
             // Update only have one update case number which default is 0
             updatedColumnHandles.forEach(columnHandle -> updateCaseColumnsBuilder.put(0, columnHandle));
             createMergeAnalysis(table, handle, tableSchema, tableScope, tableScope, ImmutableList.of(updatedColumnHandles), updateCaseColumnsBuilder.build());
+            System.out.println("StatementAnalyzer.visitUpdate() 4");
```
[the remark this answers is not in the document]
prrvchr: Done

lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/mapping/CachingIdentifierMapping.java
```diff
@@ -89,7 +89,7 @@ public String fromRemoteColumnName(String remoteColumnName)
     public String toRemoteSchemaName(RemoteIdentifiers remoteIdentifiers, ConnectorIdentity identity, String schemaName)
     {
         requireNonNull(schemaName, "schemaName is null");
-        verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(schemaName), "Expected schema name from internal metadata to be lowercase: %s", schemaName);
+        //verify(CharMatcher.forPredicate(Character::isUpperCase).matchesNoneOf(schemaName), "Expected schema name from internal metadata to be lowercase: %s", schemaName);
```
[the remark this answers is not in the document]
prrvchr: Done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/BaseTrinoCatalogTest.java
```diff
@@ -216,7 +216,7 @@ public void testCreateTable()
         TrinoCatalog catalog = createTrinoCatalog(false);
         String namespace = "test_create_table_" + randomNameSuffix();
         String table = "tableName";
-        SchemaTableName schemaTableName = new SchemaTableName(namespace, table);
+        SchemaTableName schemaTableName = new SchemaTableName(namespace, table.toLowerCase(ENGLISH));
```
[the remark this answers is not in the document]
prrvchr: This change come from [PR#26339](https://github.com/trinodb/trino/pull/26339). I don't known Iceberg for giving the right answer.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -3408,22 +3443,28 @@ protected Scope visitSetOperation(SetOperation node, Optional<Scope> scope)
             return createAndAssignScope(node, scope, outputDescriptorFields);
         }
 
-        private static Map<String, Integer> buildNameToIndex(SetOperation node, RelationType relationType)
+        private static Map<String, Integer> buildNameToIndex(SetOperation node, RelationType relationType, boolean supportsMixedCase)
         {
+            // FIXME: If we want to follow the SQL recommendations, then we need to resolve issue #28440.
+            // FIXME: To keep backward compatibility we compare field in lower case if mixed case is not supported.
```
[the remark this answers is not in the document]
prrvchr: For now, given how the Field class is used, the simplest approach is to keep this logic in place. For connectors supporting mixed case, the Fields will be case-sensitive; otherwise, they will be converted to lowercase.
FIXME is there to remind us of that.

core/trino-main/src/main/java/io/trino/metadata/MetadataUtil.java
```diff
@@ -78,22 +78,21 @@ public static String checkSchemaName(String schemaName)
...
     }
 
     public static String checkLowerCase(String value, String name)
     {
         if (value == null) {
             throw new NullPointerException(format("%s is null", name));
         }
-        checkArgument(value.equals(value.toLowerCase(ENGLISH)), "%s is not lowercase: %s", name, value);
         return value;
     }
```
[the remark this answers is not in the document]
prrvchr: Done

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -974,6 +975,18 @@ public WriterScalingOptions getInsertWriterScalingOptions(ConnectorSession sessi
...
+    public String canonicalize(ConnectorSession session, String identifier, boolean delimited)
+    {
+        return identifier.toLowerCase(ENGLISH);
+    }
+
+    @Override
+    public Boolean supportsMixedCase()
+    {
+        return false;
+    }
```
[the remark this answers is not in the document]
prrvchr: This change come from [PR#26339](https://github.com/trinodb/trino/pull/26339). I don't known Lakehouse for giving the right answer.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/TrinoRestCatalog.java
```diff
@@ -457,7 +457,7 @@ public Transaction newCreateOrReplaceTableTransaction(
     @Override
     public void registerTable(ConnectorSession session, SchemaTableName tableName, TableMetadata tableMetadata)
     {
-        TableIdentifier tableIdentifier = TableIdentifier.of(toRemoteNamespace(session, toNamespace(tableName.getSchemaName())), tableName.getTableName());
+        TableIdentifier tableIdentifier = TableIdentifier.of(toRemoteNamespace(session, toNamespace(tableName.getSchemaName())), tableName.getTableName().toLowerCase(ENGLISH));
```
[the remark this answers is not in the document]
prrvchr: This change come from [PR#26339](https://github.com/trinodb/trino/pull/26339). I don't known Iceberg for giving the right answer.

core/trino-main/src/main/java/io/trino/execution/AddColumnTask.java
```diff
@@ -143,8 +143,11 @@ public ListenableFuture<Void> execute(
             if (!element.isNullable() && !plannerContext.getMetadata().getConnectorCapabilities(session, catalogHandle).contains(NOT_NULL_COLUMN_CONSTRAINT)) {
                 throw semanticException(NOT_SUPPORTED, element, "Catalog '%s' does not support NOT NULL for column '%s'", catalogHandle, columnName);
             }
-            if (position instanceof ColumnPosition.After after && !columns.containsKey(after.column().getValue().toLowerCase(ENGLISH))) {
-                throw semanticException(COLUMN_NOT_FOUND, statement, "Column '%s' does not", after.column().getValue());
+            if (position instanceof ColumnPosition.After(Identifier after)) {
+                String afterColumn = canonicalizer.canonicalize(after);
+                if (!columns.containsKey(afterColumn)) {
+                    throw semanticException(COLUMN_NOT_FOUND, statement, "Column '%s' does not", after.getValue());
```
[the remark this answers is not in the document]
prrvchr: Done

core/trino-parser/src/main/java/io/trino/sql/tree/Identifier.java
```diff
@@ -81,7 +81,7 @@ public String getCanonicalValue()
         if (isDelimited()) {
             return value;
         }
-
+        // FIXME: We need access to NameCanonicalizer here to do what's appropriate.
         return value.toUpperCase(ENGLISH);
```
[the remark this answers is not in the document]
prrvchr: I haven't been able to find the correct reason for this conversion to uppercase. There are certainly people who know.
FIXME is only there so that those people can give their approval.

core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
```diff
@@ -601,9 +602,11 @@ public PlanNode plan(Update node)
         }
         Map<String, ColumnHandle> nameToHandle = nameToHandleBuilder.buildOrThrow();
 
+        // FIXME: The canonicalizer must be retrieved with the catalog of the table and not from the current session.
+        NameCanonicalizer canonicalizer = getNameCanonicalizer();
```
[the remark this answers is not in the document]
prrvchr: The question is: can a plan handle multiple catalogs at once? If so, then this needs to be corrected.

core/trino-parser/src/main/java/io/trino/sql/tree/Identifier.java
prrvchr: Done

plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
prrvchr: This has been resolved and has nothing to do with Druid

core/trino-main/src/main/java/io/trino/metadata/MetadataUtil.java
prrvchr: I just made the change and 4 errors appeared:
- In trino-base-jdbc:
  `Error:    TestJdbcCachingConnectorSmokeTest.createQueryRunner:29 » QueryFailed Expected table name from internal metadata to be lowercase: NATION`
  `Error:    TestJdbcFlushMetadataCacheProcedure.createQueryRunner:49 » QueryFailed Expected table name from internal metadata to be lowercase: NATION`
- In trino-sqlserver:
  `Error:    TestSqlServerConnectorTest.testRenameFromToTableWithSpecialCharacterName:223->AbstractTestQueryFramework.assertQuery:356 Execution of 'actual' query 20260227_193122_00321_9femh failed: SELECT x FROM "UPPERCASE"`
- In trino-pinot:
  `Error:    TestDynamicTable.testTextMatch:310 
expected: "SELECT "OriginCityName" FROM hybrid_REALTIME WHERE text_match("OriginCityName", 'new and york') LIMIT 70"
 but was: "SELECT "OriginCityName" FROM hybrid_REALTIME WHERE text_match("OriginCityName", 'new AND york') LIMIT 70"`

Since my changes were mixed in with my last pull, here is the list:
- `CachingIdentifierMapping` has be modified to don't check lowercase on identifier.
- The two last tests need to be modified to support metadata that is currently unchanged.
- The `TestingH2JdbcModule` has been switched to lower case database(*).
- `RenameTableTask.createTargetQualifiedObjectName()` has been canonicalized, it's an oversight.
- `TestJdbcConnectorTest.errorMessageForInsertIntoNotNullColumn()` has been modified so columnName is not converted to lower case anymore.
- `TestJdbcFlushMetadataCacheProcedure.testFlushMetadataCacheProcedureFlushIdentifierMapping()` has been modified to handle lower case.

(*) Implementing a `NameCanonicalizer` for the H2 client seems like the best solution. I await your confirmation.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
prrvchr: > But this could be embedded in the NameCanonicalizer right ?

Okay, it's finally been done. That's definitely better.

Note the new constructor for the class `QualifiedName`:
`public static QualifiedName of(List<Identifier> originalParts, Canonicalizer canonicalizer)`
This constructor must be used if you want to be able to propagate delimited identifiers correctly.
I'm not sure I applied this constructor everywhere it was necessary. In short, this has been applied everywhere the tests go.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1828,4 +1828,18 @@ default WriterScalingOptions getInsertWriterScalingOptions(ConnectorSession sess
...
+    /**
+     * Canonicalizes the provided SQL identifier according to connector-specific rules
+     * for the purpose of providing the name in metadata APIs
+     */
+    default String canonicalize(String value, boolean delimited)
+    {
+        return value.toLowerCase(ENGLISH);
+    }
+
+    default Boolean supportsMixedCase()
```
martint: This method should not be needed. All decisions on identifier canonicalization should be delegated to the connector via the `canonicalize` method.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
```diff
@@ -8805,7 +8805,7 @@ public void testSelectFromViewWithoutDefaultCatalogAndSchema()
                 .setCatalog(Optional.empty())
                 .setSchema(Optional.empty())
                 .build();
-        assertQueryFails(sessionNoCatalog, "SELECT count(*) FROM " + viewName, ".*Schema must be specified when session schema is not set.*");
```
martint, line 9083: Unrelated change?

martint:
@prrvchr . Thanks for working on this. It's been a long standing issue and it's great to have someone tackle it! As I'm going through the discussion above and reviewing the code, I want to make sure the approach is aligned with how we've envisioned this should be fixed.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
prrvchr: This change come from [PR#26339](https://github.com/trinodb/trino/pull/26339/changes#diff-7dad18a4dbe9d9a536d45655875f47ffd0d8e3b5d5c6d2f3447246763841fb44) which is the basis of this one.
I don't know Hive well enough to confirm if this is really suitable.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
prrvchr: If we want to be able to manage `io.trino.sql.analyzer.Field`, who don't have delimiter, in mixed case while maintaining compatibility with Trino's current operation, then it is necessary to have a boolean value determined by whether or not the connector supports mixed case.
Similarly, Trino metadata (i.e., `system.jdbc.schemas`, `system.jdbc.tables`, and `system.jdbc.columns`) must be converted to lowercase in legacy mode and left unchanged when mixed case support is enabled.
I cannot easily perform all these operations without having access to a method like `supportsMixedCase`.
At least I haven't found it?

prrvchr:
> Thanks for working on this. ... I want to make sure the approach is aligned with how we've envisioned this should be fixed.

@martint Your welcome and yes I'm ready to follow your instructions so I don't get lost.

After further testing, it seems difficult to get some connectors working in legacy mode and others in mixed case mode simultaneously.

If we want to handle mixed case, we need to migrate to native Trino metadata, rather than converting it to lowercase as before.

This necessary change calls into question the entire metadata management system, which currently only supports undelimited identifiers. It seems to me that the enum used to manage identifiers (ie: `io.trino.connector.informationschema.InformationSchemaTable`) should be specific to each connector?

Furthermore, it should be noted that the name of the schema containing the metadata ie: information_schema) is specific to the connector:
- mariadb speel it as `information_schema`
- hsqldb speel it as `INFORMATION_SCHEMA`
A non-delimited and canonicalized identifier can resolve this schema identifier issue if the underlying database has correctly implemented this rule, but the column name from the metadata tables seem to follow the JDBC API?

Thank you for your advice.

core/trino-main/src/main/java/io/trino/metadata/Canonicalizer.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return canonicalize(value, delimited);
+    }
+
+    default String canonicalizeColumn(Identifier identifier)
+    {
+        return canonicalizeColumn(identifier.getValue(), identifier.isDelimited());
+    }
+
+    default boolean requiresDelimiting(String value)
```
martint: This method is not used. Remove it

core/trino-main/src/main/java/io/trino/metadata/Canonicalizer.java
```diff
@@ -0,0 +1,46 @@
...
+    Canonicalizer LOWERCASE_CANONICALIZER = (value, delimited) -> delimited ? value : value.toLowerCase(ENGLISH);
+
+    String canonicalize(String value, boolean delimited);
+
+    default String canonicalize(Identifier identifier)
+    {
+        return canonicalize(identifier.getValue(), identifier.isDelimited());
+    }
+
+    default String canonicalizeColumn(String value, boolean delimited)
```
martint: This is only used by the method below. Remove it.

core/trino-main/src/main/java/io/trino/metadata/Canonicalizer.java
```diff
@@ -0,0 +1,46 @@
...
+    {
+        return canonicalize(value, delimited);
+    }
+
+    default String canonicalizeColumn(Identifier identifier)
```
martint: We don't need a separate method for columns. Canonicalization works the same way regardless of whether it's a table, column, role, user, etc.

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -936,4 +936,12 @@ default boolean isMaterializedView(Session session, QualifiedObjectName viewName
      * Returns list of functions authorization info
      */
     Set<FunctionAuthorization> getFunctionsAuthorizationInfo(Session session, QualifiedObjectPrefix prefix);
+
+    /**
+     * Canonicalizes the provided SQL identifier according to connector-specific rules
+     */
+    default Canonicalizer getCanonicalizer(Session session, String catalogName)
```
martint: This shouldn't have a default implementation. The only expected implementations of this interface are MetadataManager and the testing focused implementations. Update them directly.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -3084,4 +3081,30 @@ public Set<FunctionAuthorization> getFunctionsAuthorizationInfo(Session session,
         }
         return ImmutableSet.of();
     }
+
+    @Override
+    public Canonicalizer getCanonicalizer(Session session, String catalogName)
+    {
+        Optional<CatalogMetadata> catalog = getOptionalCatalogMetadata(session, catalogName);
+        System.out.println("MetadataManager.getCanonicalizer() 1");
```
martint: Remove

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1828,4 +1828,20 @@ default WriterScalingOptions getInsertWriterScalingOptions(ConnectorSession sess
...
         return WriterScalingOptions.DISABLED;
     }
+
+    /**
+     * Canonicalizes the provided SQL identifier according to connector-specific rules
+     * for the purpose of providing the name in metadata APIs
+     */
+    default String canonicalize(String value, boolean delimited)
+    {
+        System.out.println("ConnectorMetadata.canonicalize() 1");
```
martint: Remove

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1828,4 +1828,20 @@ default WriterScalingOptions getInsertWriterScalingOptions(ConnectorSession sess
...
+        return delimited ? value : value.toLowerCase(ENGLISH);
+    }
+
+    default String canonicalizeColumn(String value, boolean delimited)
```
martint: What's the reason to handle column names separately?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -526,6 +530,15 @@ private enum UpdateKind
...
     }
 
+    private Canonicalizer getCanonicalizer()
+    {
+        if (session.getCatalog().isPresent()) {
+            System.out.println("StatementAnalyzer.getCanonicalizer() catalog: " + session.getCatalog().get());
+            return metadata.getCanonicalizer(session, session.getCatalog().get());
+        }
+        return Canonicalizer.LEGACY_CANONICALIZER;
+    }
```
martint: This is problematic. The behavior of a query should not depend on whether a catalog has been set (i.e., via `USE `

core/trino-parser/src/main/java/io/trino/sql/tree/Canonicalizer.java
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.sql.tree;
+
+import java.util.function.Function;
+
+import static java.util.Locale.ENGLISH;
+
+public interface Canonicalizer
```
martint: This class is unused and unnecessary. Remove it.

core/trino-parser/src/main/java/io/trino/sql/tree/Identifier.java
```diff
@@ -40,27 +40,35 @@ public class Identifier
 
     private final String value;
     private final boolean delimited;
+    private final Optional<Canonicalizer> canonicalizer;
```
martint: Unused. Remove it and revert the changes in this class.

core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
```diff
@@ -601,9 +602,11 @@ public PlanNode plan(Update node)
         }
         Map<String, ColumnHandle> nameToHandle = nameToHandleBuilder.buildOrThrow();
 
+        // FIXME: The canonicalizer must be retrieved with the catalog of the table and not from the current session.
+        Canonicalizer canonicalizer = getCanonicalizer();
         io.trino.sql.tree.Expression[] orderedColumnValuesArray = new io.trino.sql.tree.Expression[updatedColumnHandles.size()];
         node.getAssignments().forEach(assignment -> {
-            String name = assignment.getName().getValue().toLowerCase(ENGLISH);
+            String name = canonicalizer.canonicalizeColumn(assignment.getName());
             ColumnHandle handle = nameToHandle.get(name);
```
martint: The query planner shouldn't be aware of canonicalization. Any necessary canonicalization should've been done by the analyzer.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorCapabilities.java
```diff
@@ -19,4 +19,5 @@ public enum ConnectorCapabilities
     NOT_NULL_COLUMN_CONSTRAINT,
     MATERIALIZED_VIEW_GRACE_PERIOD,
     MATERIALIZED_VIEW_WHEN_STALE_BEHAVIOR,
+    MIXED_CASE_IDENTIFIER_SUPPORTED,
```
martint: No need for connectors to describe this attribute. It's implicit in the implementation of `ConnectorMetadata.canonicalize()`

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -759,6 +760,8 @@ protected Type visitLocalTimestamp(LocalTimestamp node, Context context)
         @Override
         protected Type visitIdentifier(Identifier node, Context context)
         {
+            System.out.println("ExpressionAnalyzer.visitIdentifier() node value: " + node.getValue());
+            System.out.println("ExpressionAnalyzer.visitIdentifier() node isDelimited: " + node.isDelimited());
```
martint: Remove all System.out calls

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -3695,6 +3699,7 @@ public static ExpressionAnalysis analyzeExpression(
             WarningCollector warningCollector,
             CorrelationSupport correlationSupport)
     {
+        Canonicalizer canonicalizer = null;
```
martint: Unused. Remove it.

core/trino-main/src/main/java/io/trino/metadata/Canonicalizer.java
prrvchr: Thanks for your review.
Done

core/trino-main/src/main/java/io/trino/metadata/Canonicalizer.java
prrvchr: MariaDB and I suppose MySQL is case sensitive on schema and table but case insensitive on columns.
Perhaps there are other databases that work this way?

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
prrvchr: Done

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
prrvchr: For MariaDB and MySQL column are case insensitive but not table.
And for other DB even if it does the same thing, perhaps it's good to have a view of identifier processing only on the columns

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
prrvchr: There are situations where canonicalization is required but the catalog name isn't always accessible.
I'm not familiar enough with the API to resolve this.

core/trino-parser/src/main/java/io/trino/sql/tree/Canonicalizer.java
prrvchr: Done

core/trino-parser/src/main/java/io/trino/sql/tree/Identifier.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/sql/planner/QueryPlanner.java
prrvchr: Thank you for the advice. Done

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorCapabilities.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
prrvchr: Done

core/trino-main/src/main/java/io/trino/metadata/Canonicalizer.java
prrvchr: Done

client/trino-jdbc/src/main/java/io/trino/jdbc/TrinoDatabaseMetaData.java
```diff
@@ -937,14 +941,20 @@ public ResultSet getProcedureColumns(String catalog, String schemaPattern, Strin
...
-                "  PRECISION, LENGTH, SCALE, RADIX,\n" +
-                "  NULLABLE, REMARKS, COLUMN_DEF, SQL_DATA_TYPE, SQL_DATETIME_SUB,\n" +
-                "  CHAR_OCTET_LENGTH, ORDINAL_POSITION, IS_NULLABLE, SPECIFIC_NAME\n" +
+        String procedureCat = quoted("PROCEDURE_CAT");
+        String procedureSchem = quoted("PROCEDURE_SCHEM");
+        String procedureName = quoted("PROCEDURE_NAME");
+        String specificName = quoted("SPECIFIC_NAME");
+        String columnName = quoted("COLUMN_NAME");
+
+        String query = "SELECT " + procedureCat + ", " + procedureSchem + ", " + procedureName + ", " + columnName + ", " +
```
osscm: we shall use StringBuilder to contact?

martint:
@prrvchr Thanks for sticking with this. The discussion here, the companion issues you opened ([#28356](https://github.com/trinodb/trino/issues/28356), [#28786](https://github.com/trinodb/trino/issues/28786), [#29290](https://github.com/trinodb/trino/issues/29290)), and Praveen's points have surfaced most of the questions that weren't fully addressed in the original design. The [design page](https://github.com/trinodb/trino/wiki/Delimited-Identifiers) has been updated with the resolved version. Here's a summary of where it landed and what this PR needs to change to match.

## The shape we want

Single SPI method on `ConnectorMetadata`, with an `IdentifierKind` parameter:

```java
String canonicalize(String value, IdentifierKind kind);

enum IdentifierKind { SCHEMA, TABLE, COLUMN }
```

Called only on the text of non-delimited identifiers — delimited identifiers are already in canonical form and the engine uses them as-is. The branching on delimited-vs-non-delimited stays at the call site, which has the AST `Identifier` and knows.

The `IdentifierKind` parameter is the resolution to the per-kind point in [#28786](https://github.com/trinodb/trino/issues/28786): a single method covers MariaDB / MySQL's column-vs-table distinction without needing a parallel `canonicalizeColumn`.

Default is SQL-standard (uppercase non-delimited). That means:

- Oracle, SQL Server, Druid: drop the override; the default does the right thing.
- Hive / Lakehouse / Iceberg-over-Hive / PostgreSQL / Redshift: override to lowercase for all kinds.
- MariaDB / MySQL: identity for `SCHEMA`/`TABLE`, lowercase for `COLUMN`. Your existing override is right in spirit, just needs the new signature.

## What to drop

These pieces of the current PR aren't needed once the API above is in place:

- `canonicalizeColumn` — folded into `canonicalize(value, COLUMN)`.
- `ConnectorMetadata.supportsMixedCase()`, `Metadata.supportsMixedCase()`, and the `MIXED_CASE_IDENTIFIER_SUPPORTED` capability — the canonicalizer is sufficient.
- The parser-side `io.trino.sql.tree.Canonicalizer` interface — keep only `io.trino.metadata.Canonicalizer`.
- `LEGACY_CANONICALIZER` — only needed at the catalog-name pinning site.
- The `Optional<Canonicalizer>` field on `Identifier`, the constructor that takes one, and `Identifier.getUpperCaseValue()` — `Identifier` stays as `(value, delimited)`, and `getCanonicalValue()` keeps its current SQL-standard semantics.
- The canonicalizer access in `QueryPlanner` — canonicalization is complete by the time the planner runs. (Your FIXME on this site is right that the session catalog is the wrong source; the cleaner fix is to not have it in the planner at all.)
- `Canonicalizer.requiresDelimiting` as a method on the interface — it's a derived concept used by the SQL formatter, not part of the canonicalizer's contract.

## What stays

Your core architectural pieces are right:

- `ConnectorMetadata.canonicalize` as the SPI hook.
- `Metadata.getCanonicalizer(session, catalog)` reaching it.
- `MetadataUtil.createQualifiedObjectName` as the AST/metadata boundary where canonicalization happens.
- `QualifiedName` as a syntactic-only type (no normalization).
- The SPI value classes (`SchemaTableName`, `CatalogSchemaName`, `ColumnMetadata`, `ColumnSchema`) preserving their input verbatim. We try to maintain SPI compatibility as much as possible to avoid breaking third-party connectors, but here the break is unavoidable for the change to work; in-tree connectors are updated in the same change set.
- JDBC `IdentifierMapping` no longer pre-normalizing.

## Other places to adjust

- On `Metadata.getCanonicalizer`: this method shouldn't have a default implementation; only `MetadataManager` and the testing-focused implementations should supply it.
- On `StatementAnalyzer.getCanonicalizer`: the current fallback makes query behavior depend on whether `USE` was issued (i.e., on session catalog state). Canonicalization should be tied to the catalog of the referenced object, not to the session.
- On the CORRESPONDING fallback in `buildNameToIndex` (with the FIXME pointing at #28440): the right direction is binary string equality on stored names. The spec requires identifiers to be in case-normal form when stored, so equivalence reduces to plain string comparison. Cross-catalog mismatches (one arm from a SQL-compliant catalog, the other from a Hive-style catalog) are the user's problem to resolve via aliases.
- On `SHOW CREATE` formatting: the current output double-quotes every part; the direction to go is to quote only parts that wouldn't round-trip through canonicalization unquoted.
- On field matching against `RelationType`: the previous lowercasing was removed without being replaced. The right direction is to follow the originating catalog's canonicalization rules.
- Several "Done" replies to inline threads (printlns, dead code, the `getFieldName` helper, the misnamed `checkLowerCase`) don't appear to have landed in the latest force-push. Worth a sweep.
- PostgreSQL is missing a canonicalize override (one of the motivating examples in the original design page).

## Test changes

A few of the test diffs in the PR appear to accept behavior breaks; worth reconsidering whether each is the direction we want:

- `BaseConnectorTest.testRowUpdate` — the case-insensitive assertion was dropped.
- `testShowCreateInformationSchema` — expectations were changed to all-quoted output.
- `BaseHiveConnectorTest` — error message changed from "Schema must be specified" to "Catalog must be specified".
- `testAddColumnNonLowercase` — error formatting changed (echoing canonical form vs user input).

The broader testing infrastructure question in [#29290](https://github.com/trinodb/trino/issues/29290) — whether to parameterize shared test base classes by the catalog's canonicalizer, or rewrite shared tests to use delimited identifiers consistently — is worth a follow-up discussion but doesn't block this PR.

## Sequencing

Praveen's [Feb 27 suggestion](https://github.com/trinodb/trino/pull/28331#discussion_r2055207534) of doing the SPI-side decoupling (the `toLowerCase` removal in `SchemaTableName`, `CatalogSchemaName`, `ColumnMetadata`, `ColumnSchema`) as a separate, prior PR — with connectors adapting in lockstep — is worth considering. It would shrink this PR significantly and make the engine-side work testable in isolation.

prrvchr:
@martint Thank for the [design page](https://github.com/trinodb/trino/wiki/Delimited-Identifiers) this will be of great help.

After reading, I would like to make the following remarks:

- Regarding the management of MariaDB and MySQL connectors, it seems to me that the identifiers remain unchanged whether they are delimited or not (ie: identity canonicalizer). Delimiting identifiers simply allows the use of spaces or non-ASCII characters in identifier.

  However, they are only converted to lowercase beforehand for comparison purposes. This means that a MariaDB or MySQL table cannot have two columns that differ only in their case. And it would seem that this also applies to tables when the database is running under Windows.

  If this is confirmed, and that's what I've seen in my tests, then a canonicalizer alone isn't enough; a comparator is also necessary to be able to raise an exception, before the underlying driver, when inserting a new column into a table in case of a duplicate.

- > The engine reaches a per-catalog canonicalizer through Metadata.getCanonicalizer(session, catalogName). Metadata does not have a default implementation of this method; only MetadataManager and the testing-focused implementations of Metadata provide it.

  In the tests, some connectors (ie: H2) are used that do not have any implementation: they only use the base implementation, which necessitates a default implementation or modifying these connectors.

- > Field matching against RelationType follows the catalog's canonicalization rules. A column lookup in a query is resolved as follows: if the user-typed identifier is delimited, its body is used as-is; otherwise it is passed through the originating catalog's canonicalize(value, COLUMN). The resulting string is compared by binary equality against the stored column name in RelationType.

  This is the crux of the matter: RelationType lists the fields as retrieved from the underlying database.

  Therefore, they can only be compared using identifiers that have already been canonicalized, at least for non-delimited identifiers. And this is very difficult to implement in the class: [ExpressionAnalyzer.visitIdentifier()](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java#L763)

Otherwise, I completely agree with everything else.

martint:
> a canonicalizer alone isn't enough; a comparator is also necessary

The canonicalizer is the comparator. The contract is: **connectors return names already in canonical form**, and the engine canonicalizes only the user-typed non-delimited identifier at the call site. Comparison then reduces to binary string equality. There's no second comparison step and no separate comparator interface.

Concretely for MySQL/MariaDB columns: when the connector reads a column from JDBC metadata, it applies its own `canonicalize(name, COLUMN)` before populating `ColumnMetadata`. So if the underlying table has `ColumnA`, the connector hands back `columna`. `RelationType` is built from those already-canonical names — the engine doesn't re-canonicalize them. The user's `columna` (or `COLUMNA`, or `ColumnA`) goes through `canonicalize(value, COLUMN)` once at the call site, also producing `columna`, and the lookup is plain string equality.

The duplicate-column-differing-only-in-case scenario can't arise: the underlying database wouldn't permit it, so it can't show up in the metadata the connector receives. The connector doesn't need to defend against it.

> some connectors (ie: H2) ... only use the base implementation, which necessitates a default implementation or modifying these connectors

Two interfaces, two different rules:

- `ConnectorMetadata.canonicalize` (SPI) — has a default, returns `value.toUpperCase(ENGLISH)` (SQL standard). H2 and any other test connector inherit it without changes.
- `Metadata.getCanonicalizer` (engine-internal) — no default. Only `MetadataManager` and the testing-focused `Metadata` implementations supply it.

So nothing to do for H2.

> RelationType lists the fields as retrieved from the underlying database. Therefore, they can only be compared using identifiers that have already been canonicalized

Per the contract above, the names in `RelationType` *are* already canonicalized — by the connector, before they reach the analyzer. The analyzer only needs to canonicalize the user-typed identifier and compare. The work in `ExpressionAnalyzer.visitIdentifier` is making the catalog's canonicalizer reachable at that site, which is plumbing rather than a fundamental difficulty.

prrvchr:
@martint
> The canonicalizer is the comparator. The contract is: connectors return names already in canonical form
> Concretely for MySQL/MariaDB columns: when the connector reads a column from JDBC metadata, it applies its own canonicalize(name, COLUMN) before populating ColumnMetadata.

No, a database returns identifiers as they are stored. Canonization only occurs during SQL queries **originating from the user** (and this is so that the underlying connector can construct queries using only delimiter identifiers).

Initially, the databases only supported unique case identifiers and only with the ASCII characters [a-z0-9] or [A-Z0-9].
To add support for identifiers with, for example, spaces in their name, a delimiter has been introduced in SQL commands; but the rule for reading undelimited identifiers remains unchanged. All of this only applies to **SQL queries entered** by the user or any **program sending SQL queries** to the database.

When the database returns data, for example when querying metadata, the returned identifiers are exactly as they are stored in that database. No transformation (ie: canonicalization) should occur here. Anyway, how would that be possible?

- Should we consider the metadata output identifiers as delimited or undelimited?

- If I apply your rule to MariaDB or MySQL, then all the column names returned in the metadata will be lowercase, which isn't how it works... For Oracle, then, all columns will be in uppercase...

Clearly, the column names returned by the metadata must not be changed and must be delivered exactly as they are stored.

Therefore, there are two issues:

- How does a database read undelimited identifiers **in SQL queries**? And it is the **canonicalizer** who is in charge of that.

- How does a database **globally resolve** identifier names, regardless of their delimiter? The **comparator** handles this, and it occurs after canonicalization, if necessary.

It would be good if we could see the same thing, because the entire implementation depends on it.

This is easily verifiable with the tests present in the mariaDB connector; simply replace the `MariaDbMetadata.compareColumn(String value)` method with the identity function to see for yourself.

martint:
@prrvchr You're right. The case-preserving / case-insensitive regime (MySQL and MariaDB columns specifically) doesn't fit the design as I described it, and what you're calling out as needing a comparator is a real gap.

The original wiki page had a `delimited` boolean on `canonicalize` for exactly this case, but the motivation for it wasn't written down. In subsequent discussions we convinced ourselves it wasn't needed and dropped it — which is wrong for MySQL/MariaDB columns. Putting it back, this time with the rationale documented.

The corrected shape:

```java
String canonicalize(String value, IdentifierKind kind, boolean delimited);
```

Default implementation: `delimited ? value : value.toUpperCase(ENGLISH)` (SQL-standard).

How matching works: the engine canonicalizes **both sides** and compares by binary equality —

```java
canonicalize(userInput, kind, userDelimited)
    .equals(canonicalize(storedName, kind, true))
```

Stored names are passed with `delimited = true` by convention (they're concrete strings, not parsed input).

For SQL-standard and case-folding catalogs (Oracle, Hive, PostgreSQL, …), the stored side is identity — the database already enforced canonical storage, so passing `true` is a no-op. Behavior is the same as before.

For MariaDB/MySQL **column** kind, the override ignores the bit entirely:

```java
canonicalize(value, COLUMN, _) = value.toLowerCase(ENGLISH);
```

So the stored `ColumnA` canonicalizes to `columna`; user `ColumnA`, `columna`, `"ColumnA"`, `"columna"` all also canonicalize to `columna`; matching works case-insensitively regardless of delimiting; and `RelationType` keeps the verbatim stored name for display, so `SHOW COLUMNS` shows `ColumnA` as the user wrote it.

For SCHEMA and TABLE kinds in MariaDB/MySQL, the override is `delimited ? value : value` (identity) — those are case-sensitive on Linux, so the SQL-standard rule applies.

The `SHOW CREATE` quoting rule generalizes the same way: quote iff `!isValidIdentifier(value) || canonicalize(value, kind, false) != canonicalize(value, kind, true)`. For SQL-compliant catalogs this reduces to "quote iff casing differs from canonical"; for MySQL columns the two canonicalizations are always equal, so case alone never forces quoting and the original case round-trips faithfully.

I'll update the wiki page to spell this out — the API signature, the both-sides-canonicalize matching rule, and the override matrix with MySQL/MariaDB called out explicitly.

prrvchr:
@martint 
> You're right. The case-preserving / case-insensitive regime (MySQL and MariaDB columns specifically) doesn't fit the design as I described it

Thank you for your confirmation, it will help us move forward.

> `RelationType` keeps the verbatim stored name for display, so SHOW COLUMNS shows ColumnA as the user wrote it.

Yes, `RelationType` must hold the columns of a table as provided by the database metadata.

However, this raises the issue of needing to know which canonicalizer to use when processing identifiers in AST, since these are resolved by comparing the fields of `RelationType.allFields`. See method: [ExpressionAnalyzer.visitIdentifier()](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java#L763)

And if we want to be able to support two columns that differ only in case, then this resolution must be case-sensitive, thus requiring that undelimited identifiers received in method `visitIdentifier()` be canonicalized before this resolution.

martint:
@prrvchr Yes — that's the right framing. The plumbing the design relies on:

- `RelationType` carries the originating catalog identity per field. `Field` already knows where it came from in cases that mix catalogs (joins across catalogs, set operations, lateral references, subqueries). Passing the catalog through to `Field` is the work.
- At resolution time in `visitIdentifier`, for each candidate field, the analyzer looks up the canonicalizer via `Metadata.getCanonicalizer(session, field.getOriginCatalog())`, canonicalizes the user-typed identifier with `(value, COLUMN, identifier.isDelimited())`, canonicalizes the field's stored name with `(name, COLUMN, true)`, and compares by binary equality.
- For purely engine-internal scopes (e.g., `WITH` query columns, `JSON_TABLE` outputs, set-operation result columns) there's no originating catalog and the SQL-standard canonicalizer is used — same rule, just a fixed canonicalizer.

And on your second point — yes, exactly. Binary equality on canonicalized forms is the model. A catalog that genuinely distinguishes columns differing only in case (Oracle or PostgreSQL with `CREATE TABLE t ("Foo" INT, "foo" INT)`) will have stored names `Foo` and `foo` in `RelationType`, both canonicalizing to themselves under their catalog's `(name, COLUMN, true)` rule. A user reference `Foo` non-delimited folds to `FOO` (Oracle) or `foo` (PostgreSQL) and binary-equality-matches at most one of them; delimited references match exactly the one they spell. No case-insensitive "find any candidate that loosely matches" logic anywhere — that's the property the design protects.

Issue #27959, opened 2026-01-20
CREATE CATALOG should validate connection before returning success


hpopuri2:
Currently, `CREATE CATALOG` succeeds even when the provided connection configuration is invalid. The connection is only validated when the first query is executed against the catalog, leading to a poor user experience.

  ### Current Behavior

  ```sql
  CREATE CATALOG my_catalog USING snowflake
    WITH (
      "connection-url" = 'jdbc:snowflake://invalid-config...',
      "connection-user" = 'user'
    );
  -- Returns: CREATE CATALOG ✅ (Success)

  SHOW SCHEMAS IN my_catalog;
  -- Returns: Error listing schemas ❌ (Connection fails here)

  The catalog is created successfully, but users only discover configuration errors when running their first query.

  Expected Behavior

  CREATE CATALOG should:
  1. Parse and validate configuration
  2. Test the database connection (e.g., execute SELECT 1 or equivalent)
  3. Only return success if the connection test passes
  4. Return an immediate error if the connection fails

  CREATE CATALOG my_catalog USING snowflake
    WITH (
      "connection-url" = 'jdbc:snowflake://invalid-config...',
      "connection-user" = 'user'
    );
  -- Should return: Error - Unable to connect to database ❌

  Benefits

  - Immediate feedback: Users know instantly if their configuration is correct
  - Better UX: Prevents false success messages
  - Faster debugging: Configuration errors are caught at creation time, not later
  - Consistent behavior: Aligns with user expectations (catalog creation = validated connection)

  Scope

  This enhancement would apply to all JDBC-based connectors (PostgreSQL, MySQL, Snowflake, Oracle, etc.).

  Implementation Considerations

  - Optional validation via a flag like connection-validation=true/false (default: true)
  - Graceful handling of transient failures
  - Timeout configuration for validation attempts
  - Appropriate error messages for common failure scenarios

  Related

  This issue is particularly important for connectors that support multiple authentication methods where configuration errors are common.

  ---

findepi:
> `CREATE CATALOG` succeeds even when the provided connection configuration is invalid. The connection is only validated when the first query is executed against the catalog, leading to a poor user experience.

I don't disagree that it would be more user-friendly to have more validation from `CREATE CATALOG`.

Connectors generally should not validate external systems and their availability during system startup.
[...]

PR #28333, opened 2026-02-17
Drop catalog even if we are not able to get catalog metadata

vlad-lyutenko:
## Description

During getCatalogMetadataForWrite - we could actually load metadata from the connector (and interact with datasource), and this operation could fail, but we still want to drop the catalog, because catalogMetadata needed only for post-drop action - looks like this is just some hook, so we don't want to fail main operation. 

In the same way as Create Catalog works:

```
    @Override
    public void createCatalog(Session session, CatalogName catalog, ConnectorName connectorName, Map<String, String> properties, boolean notExists)
    {
        catalogManager.createCatalog(catalog, connectorName, properties, notExists);
        CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
        if (catalogMetadata.getSecurityManagement() == SYSTEM) {
            systemSecurityMetadata.catalogCreated(session, catalog);
        }
    }
```

We kind of create catalog in anyway, even if post create action will fail.
So I think the same approach should be applied for Drop catalog to be aligned with Create catalog flow.

## Release notes

```markdown
## General
* Prevent `DROP CATALOG` failure when catalog is not functioning properly.
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/Catalog.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/Catalog.java
@@ -25,9 +25,12 @@ import io.trino.spi.transaction.IsolationLevel;
 import io.trino.transaction.InternalConnector;
 import io.trino.transaction.TransactionId;
 
+import java.util.Optional;
+
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static io.trino.connector.CatalogHandle.createRootCatalogHandle;
+import static io.trino.metadata.CatalogMetadata.SecurityManagement;
 import static io.trino.metadata.CatalogStatus.FAILING;
 import static io.trino.spi.StandardErrorCode.CATALOG_UNAVAILABLE;
 import static java.lang.String.format;
@@ -96,6 +99,12 @@ public class Catalog
     public CatalogStatus getCatalogStatus()
     {
         return catalogStatus;
+    }
+
+    public Optional<SecurityManagement> getSecurityManagement()
+    {
+        return Optional.ofNullable(catalogConnector)
+                .map(ConnectorServices::getSecurityManagement);
     }
 
     public CatalogInfo toInfo()
--- a/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
@@ -848,14 +848,13 @@ public final class MetadataManager
     @Override
     public void dropCatalog(Session session, CatalogName catalog, boolean cascade)
     {
-        Optional<CatalogMetadata> catalogMetadata = Optional.empty();
         // there is a potential race condition here, TODO: https://github.com/trinodb/trino/issues/26927
-        Optional<Catalog> optionalCatalog = catalogManager.getCatalog(catalog);
-        if (optionalCatalog.isPresent() && optionalCatalog.get().getCatalogStatus() == OPERATIONAL) {
-            catalogMetadata = Optional.of(getCatalogMetadataForWrite(session, catalog.toString()));
-        }
+        boolean systemSecurityManagement = catalogManager.getCatalog(catalog)
+                .filter(foundCatalog -> foundCatalog.getCatalogStatus() == OPERATIONAL)
+                .flatMap(Catalog::getSecurityManagement)
+                .orElse(CONNECTOR) == SYSTEM;
         catalogManager.dropCatalog(catalog, cascade);
-        if (catalogMetadata.isPresent() && catalogMetadata.get().getSecurityManagement() == SYSTEM) {
+        if (systemSecurityManagement) {
             systemSecurityMetadata.catalogDropped(session, catalog);
         }
     }
--- a/testing/trino-tests/src/test/java/io/trino/tests/TestMetadataManager.java
+++ b/testing/trino-tests/src/test/java/io/trino/tests/TestMetadataManager.java
@@ -26,6 +26,7 @@ import io.trino.server.SessionContext;
 import io.trino.server.protocol.Slug;
 import io.trino.spi.Plugin;
 import io.trino.spi.QueryId;
+import io.trino.spi.TrinoException;
 import io.trino.spi.connector.ConnectorFactory;
 import io.trino.spi.connector.ConnectorViewDefinition;
 import io.trino.spi.connector.SchemaTableName;
@@ -55,6 +56,7 @@ import static io.airlift.concurrent.Threads.daemonThreadsNamed;
 import static io.trino.SessionTestUtils.TEST_SESSION;
 import static io.trino.execution.QueryState.FAILED;
 import static io.trino.execution.QueryState.RUNNING;
+import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static io.trino.spi.type.BigintType.BIGINT;
 import static java.util.concurrent.Executors.newCachedThreadPool;
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -73,6 +75,7 @@ import static org.junit.jupiter.api.parallel.ExecutionMode.SAME_THREAD;
 @Execution(SAME_THREAD) // metadataManager.getActiveQueryIds() is shared mutable state that affects the test outcome
 public class TestMetadataManager
 {
+    private final AtomicBoolean failMetadataCall = new AtomicBoolean();
     private QueryRunner queryRunner;
     private MetadataManager metadataManager;
 
@@ -89,6 +92,12 @@ public class TestMetadataManager
                 SchemaTableName viewTableName = new SchemaTableName("UPPER_CASE_SCHEMA", "test_view");
 
                 MockConnectorFactory connectorFactory = MockConnectorFactory.builder()
+                        .withMetadataWrapper(connectorMetadata -> {
+                            if (failMetadataCall.get()) {
+                                throw new TrinoException(GENERIC_INTERNAL_ERROR, new IllegalStateException("Failed to get connector metadata"));
+                            }
+                            return connectorMetadata;
+                        })
                         .withListSchemaNames(session -> ImmutableList.of("UPPER_CASE_SCHEMA"))
                         .withGetTableHandle((session, schemaTableName) -> {
                             if (schemaTableName.equals(viewTableName)) {
@@ -221,6 +230,21 @@ public class TestMetadataManager
         // TODO (https://github.com/trinodb/trino/issues/17) this should return 100 rows
         assertThat(queryRunner.execute("SELECT * FROM system.jdbc.columns WHERE table_schem = 'UPPER_CASE_TABLE' AND table_name = 'UPPER_CASE_TABLE'"))
                 .isEmpty();
+    }
+
+    @Test
+    public void testDropCatalogFailure()
+    {
+        try {
+            queryRunner.execute("CREATE CATALOG drop_mock_catalog USING mock");
+            assertThat(queryRunner.execute("SHOW CATALOGS").getOnlyColumnAsSet()).contains("drop_mock_catalog");
+            failMetadataCall.set(true);
+            queryRunner.execute("DROP CATALOG drop_mock_catalog");
+            assertThat(queryRunner.execute("SHOW CATALOGS").getOnlyColumnAsSet()).doesNotContain("drop_mock_catalog");
+        }
+        finally {
+            failMetadataCall.set(false);
+        }
     }
 
     // Probabilistic partial regression test for https://github.com/trinodb/trino/issues/28017


core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -823,12 +823,19 @@ public void createCatalog(Session session, CatalogName catalog, ConnectorName co
     public void dropCatalog(Session session, CatalogName catalog, boolean cascade)
     {
         Optional<CatalogMetadata> catalogMetadata = Optional.empty();
-        // there is a potential race condition here, TODO: https://github.com/trinodb/trino/issues/26927
-        Optional<Catalog> optionalCatalog = catalogManager.getCatalog(catalog);
-        if (optionalCatalog.isPresent() && optionalCatalog.get().getCatalogStatus() == OPERATIONAL) {
-            catalogMetadata = Optional.of(getCatalogMetadataForWrite(session, catalog.toString()));
+        try {
```
findinpath: > We kind of create catalog in anyway, even if post create action will fail.
> So I think the same approach should be applied for Drop catalog to be aligned with Create catalog flow.

It feels that we're putting under the rug a potential issue here .
Is there besides `CREATE / DROP CATALOG` any other method where this pattern needs to be employed?
`ALTER CATALOG` ?

vlad-lyutenko: I don't see Alter catalog operation, but I will check, where similar pattern could be applied

ksobolew: > It feels that we're putting under the rug a potential issue here .

I think the exception will be thrown anyway, and the error will be reported to the user. It's just that the catalog will be dropped along the way.

ksobolew: Actually, @findinpath is right to question this :) Perhaps the fix should be in `createCatalog` instead to prevent the catalog from being created if metadata can't be retrieved? Because now we have an inconsistent order of operations:

In `createCatalog`:
* create catalog
* get metadata
* call metadata

In `dropCatalog`:
* get metadata
* drop catalog
* call metadata

So the question is, which order is the correct one?

findinpath: I think it would be worthwhile to add an exception stacktrace, even if truncated, to better explain the problem that you're trying to solve.

marcinsbd: I think that the issue was that the catalog was created even when getting metadata where threw exception so in this way now we have an exisiting catalog which we can not delete because the metadata throws exception. 
imho
1. Preventing -> Make creating catalog and getting metadata atomic 
2. Treatment -> Drop catalog in finally

vlad-lyutenko: > In `dropCatalog`:
> 
> * get metadata
> * drop catalog
> * call metadata

@ksobolew I think we can not change this order, I mean we can not first drop catalog, and then get metadata.

@marcinsbd mention also 

> I think that the issue was that the catalog was created even when getting metadata where threw exception so in this way now we have an exisiting catalog which we can not delete because the metadata throws exception.
> imho
> 
> Preventing -> Make creating catalog and getting metadata atomic
> Treatment -> Drop catalog in finally

So maybe we can fix it in the way of making create catalog atomic, smth like this:

```
    @Override
    public void createCatalog(Session session, CatalogName catalog, ConnectorName connectorName, Map<String, String> properties, boolean notExists)
    {
        catalogManager.createCatalog(catalog, connectorName, properties, notExists);
        try {
            CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
            if (catalogMetadata.getSecurityManagement() == SYSTEM) {
                systemSecurityMetadata.catalogCreated(session, catalog);
            }
        }
        catch (TrinoException e) {
            // revert catalog creation if we fail to load metadata from the connector,
            // because without metadata catalog is not operational
            // TODO correct logic on exist/not exist cases
            catalogManager.dropCatalog(catalog, !notExists);
            throw e;
        }
    }
```
wdyt guys?

findinpath: Can we maybe retrieve `CatalogMetadata` from the `createCatalog/dropCatalog` methods in `CatalogManager` ?

The problematic call seems to be 

```
CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
```

marcinsbd: imho, we should not allow catalogs to be in an inconsistent state  - after we create catalog we need to check if it's operational
I assume that line guarantee that, isn't?

vlad-lyutenko: @findinpath
I checked code path as you suggested for static catalog initialisation with one of the QueryRunners:
`MetadaManager.createCatalog` is not called in code path.

the problematic call 
`CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());`

will be called only on schema creation.

So it looks like this code path is very specific only for Dynamic catalogs.

@marcinsbd, @ksobolew  

So I think we have 2 options as you suggested : 
- either fix on delete path (current code)
- or make create catalog atomic:
```
 @Override
    public void createCatalog(Session session, CatalogName catalog, ConnectorName connectorName, Map<String, String> properties, boolean notExists)
    {
        catalogManager.createCatalog(catalog, connectorName, properties, notExists);
        try {
            CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
            if (catalogMetadata.getSecurityManagement() == SYSTEM) {
                systemSecurityMetadata.catalogCreated(session, catalog);
            }
        }
        catch (TrinoException e) {
            // revert catalog creation if we fail to load metadata from the connector,
            // because without metadata catalog is not operational
            // TODO correct logic on exist/not exist cases
            catalogManager.dropCatalog(catalog, !notExists);
            throw e;
        }
    }
```

findepi:
please add tests using mock connector.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
vlad-lyutenko: As discussed offline with @findepi I refactored this code according to hints:

> CREATE CATALOG should succeed or fail cleanly without leaving a partial state.

and 

> DROP CATALOG should remove the catalog definition from Trino metadata without attempting to initialize a backend connection that requires user-specific credentials (which are not present during a DROP operation).

Also added some tests.

But @findinpath mentioned one scenario, which kind of break the flow:
- Admin user create dynamic catalog with Oauth2 passthrough
- Admin user doesn't have correct settings for Oauth2 passthrough in his IDP
- And he knows about this and accept it, because catalog is created for users, who will be using this catalog
- And these users has correct settings in IDPs, so it will work for them.
- So create operation will partially fail (catalog created, but connector metadata calls fails)

So if we will go with my changes in this PR, we will break this working flow, which @findinpath pointed me.

So as workaround I can go only with DROP changes (drop catalog, even if connector metadata call fails) , but didn't touch create flow.

wdyt @findepi ?

Thanks!

findepi: > So as workaround I can go only with DROP changes (drop catalog, even if connector metadata call fails) , but didn't touch create flow.

works for me.

> Admin user doesn't have correct settings for Oauth2 passthrough in his IDP

yes, this is possible

see _"what to do with catalogs that require client extra credentials to work properly (eg Delta in some configurations). During CREATE CATALOG we may lack these credentials and healthcheck would fail."_ in https://github.com/trinodb/trino/issues/27959#issuecomment-3913054649

> * So create operation will partially fail (catalog created, but connector metadata calls fails)

We can issue a warning in such case.

vlad-lyutenko: Big thx! for suggestions @findepi 
I refactored the code according to them

testing/trino-tests/src/test/java/io/trino/tests/TestMetadataManager.java
```diff
@@ -223,6 +257,35 @@ public void testColumnsQueryWithUpperCaseFilter()
...
+    @Test
+    public void testDropCatalogFailure()
+    {
+        try {
+            queryRunner.execute("CREATE CATALOG mock2 USING mock");
+            assertThat(queryRunner.execute("SHOW CATALOGS").getOnlyColumnAsSet()).contains("mock2");
+            failMetadataCall.set(true);
+            assertThatThrownBy(() -> queryRunner.execute("DROP CATALOG mock2"))
+                    .hasMessageContaining("Failed to get connector metadata");
+            assertThat(queryRunner.execute("SHOW CATALOGS").getOnlyColumnAsSet()).doesNotContain("mock2");
```
findepi: The DROP fails, but the catalog is dropped nonetheless?
Am i reading this correctly?

vlad-lyutenko: Yes, before this PR flow was like this:
- obtain connector metadata, needed for some post-drop action
- actual catalog drop

If first operation fails, actual drop is not executed.
With my change, we want to drop catalog even if obtaining connector metadata fails.

findepi: > With my change, we want to drop catalog even if obtaining connector metadata fails.

this makes sense to me

however, if dropping catalog is successful, i would expect DROP CATALOG statement not to end with a failure.

vlad-lyutenko: refactored

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -839,22 +839,37 @@ public List<RelationCommentMetadata> listRelationComments(Session session, Strin
...
+            // there is a potential race condition here, TODO: https://github.com/trinodb/trino/issues/26927
+            Optional<Catalog> optionalCatalog = catalogManager.getCatalog(catalog);
+            if (optionalCatalog.isPresent() && optionalCatalog.get().getCatalogStatus() == OPERATIONAL) {
+                catalogMetadata = Optional.of(getCatalogMetadataForWrite(session, catalog.toString()));
+            }
+        }
+        catch (TrinoException e) {
+            log.warn(e, "Exception occurred, during obtaining catalog metadata");
+        }
+        finally {
```
findinpath: why `finally` ?

Why not simply after the try/catch statement?

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -839,21 +839,34 @@ public List<RelationCommentMetadata> listRelationComments(Session session, Strin
     public void createCatalog(Session session, CatalogName catalog, ConnectorName connectorName, Map<String, String> properties, boolean notExists)
     {
         catalogManager.createCatalog(catalog, connectorName, properties, notExists);
```
findinpath, line 841: @vlad-lyutenko  let's add the stacktrace of the issue in the description of this issue to provide concrete context for other eventual participants.

```
io.trino.spi.TrinoException: Cannot obtain metadata
    at io.trino.plugin.iceberg.IcebergConnector.getMetadata(IcebergConnector.java:145)
    at io.trino.metadata.CatalogTransaction.getConnectorMetadata(CatalogTransaction.java:71)
    at io.trino.metadata.CatalogMetadata.getMetadata(CatalogMetadata.java:87)
    at io.trino.metadata.MetadataManager$QueryCatalogs.registerCatalog(MetadataManager.java:3003)
    at io.trino.metadata.MetadataManager.registerCatalogForQuery(MetadataManager.java:2976)
    at io.trino.metadata.MetadataManager.getCatalogMetadataForWrite(MetadataManager.java:2952)
    at io.trino.metadata.MetadataManager.createCatalog(MetadataManager.java:818)
    at io.trino.tracing.TracingMetadata.createCatalog(TracingMetadata.java:383)
    at io.trino.execution.CreateCatalogTask.execute(CreateCatalogTask.java:99)
    at io.trino.execution.CreateCatalogTask.execute(CreateCatalogTask.java:44)
    at io.trino.execution.DataDefinitionExecution.start(DataDefinitionExecution.java:161)
    at io.trino.execution.SqlQueryManager.createQuery(SqlQueryManager.java:306)
    at io.trino.dispatcher.LocalDispatchQuery.startExecution(LocalDispatchQuery.java:154)
    at io.trino.dispatcher.LocalDispatchQuery.lambda$waitForMinimumWorkers$1(LocalDispatchQuery.java:138)
    at io.airlift.concurrent.MoreFutures.lambda$addSuccessCallback$0(MoreFutures.java:570)
    at io.airlift.concurrent.MoreFutures$3.onSuccess(MoreFutures.java:545)
    at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1132)
    at io.trino.$gen.Trino_477_e_1____20260116_005757_2.run(Unknown Source)
Caused by: org.apache.iceberg.exceptions.NotAuthorizedException: Not authorized: {"error_code":401,"message":"Credential was not sent or was of an unsupported type for this API. [ReqId: c0bddac4-5cb9-4513-938f-324845f449b8]"}
    at org.apache.iceberg.rest.ErrorHandlers$DefaultErrorHandler.accept(ErrorHandlers.java:236)
    at org.apache.iceberg.rest.ErrorHandlers$DefaultErrorHandler.accept(ErrorHandlers.java:214)
    at org.apache.iceberg.rest.HTTPClient.throwFailure(HTTPClient.java:240)
    at org.apache.iceberg.rest.HTTPClient.execute(HTTPClient.java:336)
    at org.apache.iceberg.rest.HTTPClient.execute(HTTPClient.java:297)
    at org.apache.iceberg.rest.BaseHTTPClient.get(BaseHTTPClient.java:77)
    at org.apache.iceberg.rest.RESTSessionCatalog.fetchConfig(RESTSessionCatalog.java:1023)
    at org.apache.iceberg.rest.RESTSessionCatalog.initialize(RESTSessionCatalog.java:205)
    at io.trino.plugin.iceberg.catalog.rest.TrinoIcebergRestCatalogFactory.create(TrinoIcebergRestCatalogFactory.java:145)
    at io.trino.plugin.iceberg.IcebergMetadataFactory.create(IcebergMetadataFactory.java:114)
    at io.trino.plugin.iceberg.IcebergTransactionManager$MemoizedMetadata.get(IcebergTransactionManager.java:90)
    at io.trino.plugin.iceberg.IcebergTransactionManager.get(IcebergTransactionManager.java:56)
    at io.trino.plugin.iceberg.IcebergConnector.getMetadata(IcebergConnector.java:141)
    ... 20 more

```


Concrete context for eventual reproduction of the issue : creating dynamically a catalog with **vended credentials** for Iceberg REST Catalog integration is failing.


What is actually happening ?


```
CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
if (catalogMetadata.getSecurityManagement() == SYSTEM) {
    systemSecurityMetadata.catalogCreated(session, catalog);
}
```

The call `getCatalogMetadataForWrite` actually ends up in instantiating `IcebergMetadata` like it would happen in the case of starting a new query.

Something is not right here - we're creating an IcebergMetadata although there is no actual query happening.

Why are we going through all this effort?
Well, it seems we just want to know the "security management" attribute of the catalog.

https://github.com/trinodb/trino/blob/d2611800e8f4d290783411c9d0ae8f3a1e663a70/core/trino-main/src/main/java/io/trino/connector/ConnectorServices.java#L287-L290

My impression is that we don't need to create a new transaction to figure this out.

We rather need to expose the security management of the catalog connector

https://github.com/trinodb/trino/blob/d2611800e8f4d290783411c9d0ae8f3a1e663a70/core/trino-main/src/main/java/io/trino/metadata/Catalog.java#L135


@vlad-lyutenko  pls consider the following:
- return `Catalog` (instead of `void`) in `io.trino.connector.CoordinatorDynamicCatalogManager#createCatalog`
- add a method in `Catalog` to expose its "security management"
- refactor the code of `MetadataManager` to avoid calling `getCatalogMetadataForWrite` in the methods which relate to CRUD operations for catalogs.

findinpath: cc @ksobolew & @hashhar    because you were probably also involved in the process of shaping the functionality of dynamic catalogs

testing/trino-tests/src/test/java/io/trino/tests/TestMetadataManager.java
```diff
@@ -300,4 +362,76 @@ private static ConnectorViewDefinition getConnectorViewDefinition()
                 false,
                 ImmutableList.of());
     }
+
+    private abstract static class ForwardingConnectorMetadata
```
ksobolew: This could (should?) be a general-purpose utility class, with its own tests to ensure that all methods are overridden and properly delegate calls.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
ksobolew: I think the original issue was that we are able to create a misconfigured dynamic catalog, but then we are not able to drop it because of the misconfiguration. The stack trace you posted is a symptom of that.

I am only invested in the `systemSecurityMetadata.catalogCreated` and `systemSecurityMetadata.catalogDropped` calls, though, I was not involved in the implementation of dynamic catalog otherwise :) But yes, I suppose I'm responsible for calling `getCatalogMetadataForWrite` here.

vlad-lyutenko: @ksobolew what you think about current situation, should we just add some exception handling as it's done in this PR or go forward with more global changes as suggested above:

>  - return Catalog (instead of void) in io.trino.connector.CoordinatorDynamicCatalogManager#createCatalog
> add a method in Catalog to expose its "security management"
> - refactor the code of MetadataManager to avoid calling getCatalogMetadataForWrite in the methods which 
> - relate to CRUD operations for catalogs.

findepi: I don't think we should refocus the PR now on `createCatalog`.
The DROP needs a fix so that malfunctioning catalog can be dropped, regardless of what happened during catalog creation.

vlad-lyutenko: @findepi so removing all the code changes for `createCatalog` and just fix the DROP right?

ksobolew: Personally, I'm fine with how this PR looks right now (except maybe extracting the `Forwarding` class as a utility on its own).

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -839,21 +839,34 @@ public List<RelationCommentMetadata> listRelationComments(Session session, Strin
...
-        CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
-        if (catalogMetadata.getSecurityManagement() == SYSTEM) {
-            systemSecurityMetadata.catalogCreated(session, catalog);
+        try {
+            CatalogMetadata catalogMetadata = getCatalogMetadataForWrite(session, catalog.toString());
+            if (catalogMetadata.getSecurityManagement() == SYSTEM) {
```
findepi: We don't need `CatalogMetadata` to know what is the connector security management.
The value comes from here
https://github.com/trinodb/trino/blob/774f0a883394253f48a120e5298ce38ad1165f83/core/trino-main/src/main/java/io/trino/metadata/Catalog.java#L135

If we skip `getCatalogMetadataForWrite` during (CREATE and) DROP CATALOG, we should solve the problem with DROP being unable to proceed when catalog is malfunctioning.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
findepi: We should fix the DROP flow.
We may change CREATE's code along the way if it's needed for consistency or anything else.

> * refactor the code of MetadataManager to avoid calling getCatalogMetadataForWrite in the methods which
> * relate to CRUD operations for catalogs.

i came to the same conclusion
- https://github.com/trinodb/trino/pull/28333/changes#r3000023800

vlad-lyutenko: @findepi @ksobolew How about this - https://github.com/trinodb/trino/pull/28333/changes/48f68de87b2f1d504e6b739bc130375e7a21a978

Changes only in DROP, all other improvements can go as follow up for CREATE e.t.c
No need for `ForwardingConnectorMetadata` in tests

ksobolew: Fine by me too!

core/trino-main/src/main/java/io/trino/metadata/Catalog.java
```diff
@@ -98,6 +98,11 @@ public CatalogStatus getCatalogStatus()
         return catalogStatus;
     }
 
+    public ConnectorServices getCatalogConnector()
```
dain: Instead of exposing the whole ConnectorServices class, just add the method `getSecurityManagement()`

Also the catalogConnector will be null when you have a failed catalog, so this should return `Optional<SecurityManagement>`

vlad-lyutenko: refactored

PR #28334, opened 2026-02-17
Add alias-based dynamic filter assertions for plan matching

piotrrzysko:
### Summary
This PR introduces a new approach for asserting dynamic filters in plan matcher tests. The new framework uses logical aliases to connect dynamic filter producers (at Join/SemiJoin nodes) with their consumers (at `FilterNode`).

### Why?

The existing dynamic filter assertions lack explicit links between producers and consumers. Tests could only specify that a Join produces a dynamic filter and that _somewhere_ in the left subtree there's a matching consumer—but couldn't explicitly assert which `FilterNode` consumes a given dynamic filter. This made it impossible to verify correct DF placement in plans with multiple `FilterNodes` or multiple dynamic filters.

### Approach

Tests assign string aliases (e.g., `"DF1"`) to dynamic filters. During bottom-up plan matching, consumer matchers at `FilterNodes` collect candidate `DynamicFilterId`s that match the expected pattern and associate them with the alias. As matching progresses up the tree, candidates are narrowed: multiple `FilterNodes` with the same alias intersect their candidates, and the producer matcher at Join/SemiJoin resolves to exactly one ID by verifying the build symbol. 

#### Before
```java
  join(RIGHT, builder -> builder
      .equiCriteria("ORDERS_OK", "LINEITEM_OK")
      .dynamicFilter(BIGINT, "ORDERS_OK", "LINEITEM_OK")
      .left(anyTree(tableScan("orders", ...)))
      .right(exchange(tableScan("lineitem", ...))))
```

#### After
```java
  join(RIGHT, builder -> builder
      .equiCriteria("ORDERS_OK", "LINEITEM_OK")
      .addDynamicFilter("DF", "LINEITEM_OK")
      .left(
          filter(TRUE,
              df -> df.addConsumer(c -> c.alias("DF").expression(BIGINT, "ORDERS_OK")),
              tableScan("orders", ...)))
      .right(exchange(tableScan("lineitem", ...))))
```

## Release notes

```markdown
## Section
```


core/trino-main/src/test/java/io/trino/sql/planner/assertions/DynamicFilterConsumerMatcher.java
```diff
@@ -0,0 +1,242 @@
...
+                }
+                matchedConsumers.add(consumer);
+                break;
+            }
+        }
+
+        return matchedConsumers.size() == consumers.size();
+    }
+
+    private static <K, V> Map<K, Set<V>> groupByKey(List<V> values, Function<V, K> keyFunction)
```
raunaqmorarka, line 137: `java.util.stream.Collectors#groupingBy(java.util.function.Function<? super T,? extends K>, java.util.stream.Collector<? super T,A,D>)` ?

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
```diff
@@ -164,11 +166,11 @@ public void testDynamicFilterConsumedOnBuildSide()
                 removeUnsupportedDynamicFilters(root),
                 join(INNER, builder -> builder
                         .equiCriteria("ORDERS_OK", "LINEITEM_OK")
-                        .dynamicFilter(BIGINT, "ORDERS_OK", "LINEITEM_OK")
+                        .dynamicFilters(dynamicFilters -> dynamicFilters.addProducer("DF", "LINEITEM_OK"))
```
findepi: Before the change, in the call we were providing 2 symbol names. They were oddly named ("key" and "value"), but apparently they were coming from left and right side.

After the change, we provide only the one from right value. Why is that?

findepi: API choice
It appears that after  `.dynamicFilters(dynamicFilters -> dynamicFilters.` the only think that can be called is `addProducer(String alias, String buildAlias)`.
Do you envision many other options in the future?
Cause it seems we could cut of `dynamicFilters -> dynamicFilters.addProducer` from the callsite, and thus make it significantly simpler to read and write.

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
```diff
@@ -164,11 +166,11 @@ public void testDynamicFilterConsumedOnBuildSide()
...
                         .left(
                                 PlanMatchPattern.filter(
                                         TRUE,
-                                        createDynamicFilterExpression(metadata, new DynamicFilterId("DF"), BIGINT, new Reference(BIGINT, "ORDERS_OK")),
+                                        dynamicFilters -> dynamicFilters.addConsumer(consumer -> consumer.alias("DF").expression(BIGINT, "ORDERS_OK")),
```
findepi, line 173: API design

After `dynamicFilters -> dynamicFilters.` there are two options: a consumer, or no consumer.
It could be also done without a lambda over a builder, like this

```
filter(
   «static filter condition matcher»,
   «dynamic filter matcher (not builder)»,
   «source»)
```

where `«dynamic filter matcher (not builder)»` is one of 
- `noDynamicFilters()`
- `dynamicFilter("DF", BIGINT, "ORDERS_OK")`

findepi:
cc @sopel39

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
piotrrzysko: This change decouples producers from consumers.

Previously, `.dynamicFilter(BIGINT, "ORDERS_OK", "LINEITEM_OK")` meant: 
* match if I'm producing a dynamic filter based on `LINEITEM_OK` and, somewhere on the left side, there is a consumer that applies this dynamic filter to `ORDERS_OK`.

Now, `.dynamicFilters(dynamicFilters -> dynamicFilters.addProducer("DF", "LINEITEM_OK"))` means: 
* match if I'm producing a dynamic filter based on `LINEITEM_OK` and `DF` is matched by at least one consumer. I don't care how the consumer matches `DF`. I only require that it be matched before I attempt to match - i.e., the consumer must be my descendant.
In this case, the consumer matches `DF` if it applies this filter to `ORDERS_OK`.

piotrrzysko: To make sure I understand the suggestion correctly: for multiple DFs produced by one join, it would look like this:
```java
.dynamicFilters(
        "DF_ORDERKEY", "L_ORDERKEY",
        "DF_PARTKEY", "L_PARTKEY"))
```

Right?

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
piotrrzysko: > there are two options: a consumer, or no consumer.

there is a third option: multiple consumers (see for example: `io.trino.sql.planner.TestDynamicFilter#testCrossJoinBetweenDF`)

In this case, I think we need either a builder or a list of dynamic filter matchers.

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
findepi: for multiple DFs produced by one join, perhaps there could be multiple `dynamicFilters()` invocations

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
findepi: so maybe a builder is warranted here. fine

core/trino-main/src/test/java/io/trino/sql/planner/optimizations/TestRemoveUnsupportedDynamicFilters.java
piotrrzysko: I applied your suggestion. Based on other builders in the plan assertion framework, I chose `addDynamicFilter` instead of `dynamicFilter` or `dynamicFilters`.

core/trino-main/src/test/java/io/trino/sql/planner/assertions/DynamicFilterConsumerMatcher.java
```diff
@@ -0,0 +1,242 @@
...
+
+    private DynamicFilterConsumerMatcher(List<DynamicFilterConsumer> expectedConsumers)
+    {
+        this.expectedConsumers = groupByKey(expectedConsumers, consumer -> new DynamicFilterAlias(consumer.alias()));
+    }
+
+    @Override
+    public boolean shapeMatches(PlanNode node)
+    {
+        return node instanceof FilterNode;
```
Praveen2112: Apart from being a FilterNode - shouldn't we check it has predicate based on dynamic filters ?

core/trino-main/src/test/java/io/trino/sql/planner/assertions/DynamicFilterAlias.java
```diff
@@ -0,0 +1,30 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.sql.planner.assertions;
+
+import static java.util.Objects.requireNonNull;
+
+public record DynamicFilterAlias(String alias)
```
Praveen2112, line 18: nit: Any specific reason for this abstraction instead of using String ? Alias is more of a name right ?

core/trino-main/src/test/java/io/trino/sql/planner/assertions/MatchingDynamicFilters.java
```diff
@@ -0,0 +1,156 @@
...
+    }
+
+    public boolean containsUnresolvedAliases()
+    {
+        Set<DynamicFilterId> usedIds = new HashSet<>();
+
+        for (Map.Entry<DynamicFilterAlias, Set<DynamicFilterId>> entry : matching.entrySet()) {
+            if (entry.getValue().size() != 1) {
+                return true;
+            }
```
Praveen2112, line 95: If an alias can have only a single entry, shouldn't we fail it a bit early ?

core/trino-main/src/test/java/io/trino/sql/planner/assertions/MatchingDynamicFilters.java
```diff
@@ -0,0 +1,156 @@
...
+                .flatMap(Collection::stream)
+                .collect(toImmutableSet());
+    }
+
+    public Optional<Set<DynamicFilterId>> getCandidates(DynamicFilterAlias alias)
+    {
+        return Optional.ofNullable(matching.get(alias));
+    }
+
+    public boolean containsUnresolvedAliases()
```
Praveen2112, line 85: Is the method name a bit misleading, IIUC it looks for uniqueness of DynamicFilterId ? How does it correlate with unresolved alias, isn't more of a duplicate aliases ?

core/trino-main/src/test/java/io/trino/sql/planner/assertions/JoinMatcher.java
```diff
@@ -250,27 +250,9 @@ public Builder filter(Expression expectedFilter)
...
-        public Builder dynamicFilter(Type type, String key, String value)
-        {
-            this.dynamicFilter = Optional.of(ImmutableList.of(new PlanMatchPattern.DynamicFilterPattern(new Reference(type, key), EQUAL, value)));
-
-            return this;
-        }
-
-        @CanIgnoreReturnValue
-        public Builder dynamicFilter(List<PlanMatchPattern.DynamicFilterPattern> expectedDynamicFilter)
+        public Builder addDynamicFilter(String alias, String buildAlias)
```
Praveen2112, line 253: For consistency can we keep it as `dynamicFilter` ?

core/trino-main/src/test/java/io/trino/sql/planner/assertions/DynamicFilterConsumerMatcher.java
piotrrzysko: It shouldn't hurt. Added.

core/trino-main/src/test/java/io/trino/sql/planner/assertions/JoinMatcher.java
piotrrzysko: For consistency with other builders in the plan assertion framework, I chose `addDynamicFilter` rather than `dynamicFilter` or `dynamicFilters`. 

See, for example, `io.trino.sql.planner.assertions.WindowMatcher.Builder#addFunction(...)`. 

It seems that the convention is: when setting a value, we drop the verb; when adding a value, we prefix it with add.

core/trino-main/src/test/java/io/trino/sql/planner/assertions/MatchingDynamicFilters.java
piotrrzysko: The goal is to have a 1:1 mapping between aliases and DynamicFilterIds. If that's not achieved, we have unresolved aliases. 

Does that make sense or would you prefer a different name?

core/trino-main/src/test/java/io/trino/sql/planner/assertions/MatchingDynamicFilters.java
piotrrzysko: My reply above should make this clearer. We call this method at the end of plan matching to ensure a 1:1 mapping. We can have multiple entries here, which indicates that we detected a discrepancy in the plan.

core/trino-main/src/test/java/io/trino/sql/planner/assertions/DynamicFilterAlias.java
piotrrzysko: It’s more explicit. 

We store aliases in maps, e.g., `Map<DynamicFilterAlias, Set<DynamicFilterConsumer>>`. I prefer this over `Map<String, Set<DynamicFilterConsumer>>`. 

Also, we already have `io.trino.sql.planner.assertions.SymbolAlias`, which is similarly a thin wrapper over `String`.

core/trino-main/src/test/java/io/trino/sql/planner/assertions/DynamicFilterConsumerMatcher.java
```diff
@@ -0,0 +1,242 @@
...
+        FilterNode filterNode = (FilterNode) node;
+        Map<DynamicFilterId, Set<DynamicFilters.Descriptor>> actualDynamicFilters = groupByKey(
+                extractDynamicFilters(filterNode.getPredicate()).getDynamicConjuncts(),
+                DynamicFilters.Descriptor::getId);
+
+        if (actualDynamicFilters.size() != expectedConsumers.size()) {
+            return NO_MATCH;
+        }
+
+        SymbolMapper symbolMapper = new SymbolMapper(symbol -> Symbol.from(context.symbolAliases().get(symbol.name())));
```
Praveen2112: nit: How about `new SymbolMapper(symbol -> context.symbolAliases().getSymbol(symbol.name()))`

core/trino-main/src/test/java/io/trino/sql/planner/assertions/MatchingDynamicFilters.java
Praveen2112: But alias doesn't always mean a 1:1 mapping right i.e a DynamicFilterIds can have multiple aliases. Additionally unresolved alias means - an dynamic filter alias with an empty set. Or can we add a code comment for the same. How about `containsUniqueAliases`. 

Is there any specific reason behind maintaining this 1:1 mapping ?

piotrrzysko: > But alias doesn't always mean a 1:1 mapping right

An alias is just a pointer to an ID. We need aliases because we don’t know the IDs when writing assertions. Since IDs must be unique, in the end we should always have a 1:1 mapping.

Issue #28204, opened 2026-02-10
[Iceberg] Add Support for Iceberg Native Encryption (Client-side Table Encryption)


kmikmy:
### Summary

Trino currently does not support Apache Iceberg’s native table encryption (client-side encryption), which encrypts data files, manifest files, and manifest list files before they are written to the underlying object storage.
https://github.com/trinodb/trino/blob/544ee86448677317c4be080a85a7c0d05e7ae9f3/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java#L687-L690

This feature becomes especially important in environments where the underlying storage layer cannot be fully trusted.

### Background

[Apache Iceberg 1.10](https://iceberg.apache.org/releases/#1100-release) provides native encryption capabilities for tables by encrypting datafiles, manifest files and manifest list files before writing them to storage. Iceberg supports encryption using KMS systems such as AWS KMS, GCP KMS, and Azure Key Vault. In addition, Iceberg allows users to integrate any custom KMS by implementing the KeyStoreKmsClient interface and loading their own Java class, enabling the use of organization-specific or non‑standard key management systems.

kmikmy:
refs. 
 
- https://github.com/trinodb/trino/pull/28389

PR #24517, opened 2024-12-18
[WIP] Add Parquet decryption support for Hive tables

sopel39:
## Description
Adds support to read Hive tables with encrypted Parquet files. 
PS: This PR is work in progress and we are adding tests to it.

## Additional context and related issues
Parquet added support for encryption https://parquet.apache.org/docs/file-format/data-pages/encryption/. Spark also added support to read and write tables with parquet encrypted files. In this PR we are adding support to read Hive tables with encrypted Parquet files with Trino.

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
# Section
* Fix some things. ({issue}`issuenumber`)
```

PR #28354, opened 2026-02-18
Add support for Iceberg table encryption

kamijin-fanta:
Integrate Iceberg encryption handling into read and write paths and table operations across Iceberg catalogs.

## Description

- Add end-to-end support for Iceberg table encryption in the Trino Iceberg connector.
- Support encrypted table creation via table properties: `encryption_key_id` and `encryption_data_key_length`.
- Add catalog-level KMS implementation config: `iceberg.encryption.kms-impl`.
- Wire `EncryptionManager` into write paths for data files, position delete files, and deletion vectors.
- Persist and propagate file key metadata through commit tasks, splits, and table changes processing.
- Add decryption-aware read paths for encrypted data files and delete files by introducing encrypted Trino input/output wrappers.
- Update Iceberg table operations to use `EncryptingFileIO` for encrypted tables.

- `encryption_data_key_length` is validated to `16`, `24`, or `32` bytes and requires `encryption_key_id`.
- Changing `encryption_key_id` or `encryption_data_key_length` after they are set is rejected.
- KMS client behavior is explicit: catalog-level `iceberg.encryption.kms-impl` is reused, while table-level `encryption.kms-impl` is resolved per table.
- Added and updated tests include encrypted table DML flows, ORC writer validation with encryption, table changes on encrypted tables, config mapping, and encryption manager factory behavior.
- Iceberg References https://iceberg.apache.org/docs/nightly/encryption/
- Related issue: #28204

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
@@ -41,6 +41,8 @@ import io.trino.parquet.ParquetCorruptionException;
 import io.trino.parquet.ParquetDataSource;
 import io.trino.parquet.ParquetDataSourceId;
 import io.trino.parquet.ParquetReaderOptions;
+import io.trino.parquet.crypto.DecryptionKeyRetriever;
+import io.trino.parquet.crypto.FileDecryptionProperties;
 import io.trino.parquet.metadata.FileMetadata;
 import io.trino.parquet.metadata.ParquetMetadata;
 import io.trino.parquet.predicate.TupleDomainParquetPredicate;
@@ -52,6 +54,7 @@ import io.trino.plugin.hive.TransformConnectorPageSource;
 import io.trino.plugin.hive.orc.OrcPageSource;
 import io.trino.plugin.hive.parquet.ParquetPageSource;
 import io.trino.plugin.iceberg.IcebergParquetColumnIOConverter.FieldContext;
+import io.trino.plugin.iceberg.IcebergSplit.ParquetFileDecryptionData;
 import io.trino.plugin.iceberg.delete.DeleteFile;
 import io.trino.plugin.iceberg.delete.DeleteManager;
 import io.trino.plugin.iceberg.delete.DeletionVector;
@@ -111,6 +114,7 @@ import org.apache.iceberg.types.Types;
 import org.apache.iceberg.types.Types.NestedField;
 import org.apache.iceberg.util.StructLikeWrapper;
 import org.apache.parquet.column.ColumnDescriptor;
+import org.apache.parquet.hadoop.metadata.ColumnPath;
 import org.apache.parquet.io.MessageColumnIO;
 import org.apache.parquet.schema.GroupType;
 import org.apache.parquet.schema.MessageType;
@@ -301,6 +305,7 @@ public class IcebergPageSourceProvider
                 split.getDataSequenceNumber(),
                 tableHandle.getNameMappingJson().map(NameMappingParser::fromJson),
                 split.getEncryptionKeyMetadata(),
+                split.getParquetFileDecryptionData(),
                 encryptionManager);
     }
 
@@ -325,6 +330,7 @@ public class IcebergPageSourceProvider
             long dataSequenceNumber,
             Optional<NameMapping> nameMapping,
             Optional<byte[]> encryptionKeyMetadata,
+            Optional<ParquetFileDecryptionData> parquetFileDecryptionData,
             Optional<EncryptionManager> encryptionManager)
     {
         Map<Integer, Optional<String>> partitionKeys = getPartitionKeys(partitionData, partitionSpec);
@@ -342,7 +348,13 @@ public class IcebergPageSourceProvider
         String partition = partitionSpec.partitionToPath(partitionData);
         TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), fileIoProperties);
         TrinoInputFile inputFile = newInputFile(fileSystem, path, fileSize, encryptionKeyMetadata, isUseFileSizeFromMetadata(session));
-        TrinoInputFile decryptedInputFile = decryptInputFileIfNeeded(inputFile, encryptionKeyMetadata, encryptionManager);
+        boolean useLegacyParquetDecryption = fileFormat == IcebergFileFormat.PARQUET && encryptionKeyMetadata.isPresent();
+        Optional<ParquetFileDecryptionData> effectiveParquetFileDecryptionData = useLegacyParquetDecryption ? Optional.empty() : parquetFileDecryptionData;
+
+        TrinoInputFile decryptedInputFile = inputFile;
+        if (fileFormat != IcebergFileFormat.PARQUET || useLegacyParquetDecryption) {
+            decryptedInputFile = decryptInputFileIfNeeded(inputFile, encryptionKeyMetadata, encryptionManager);
+        }
         try {
             if (effectivePredicate.isAll() &&
                     start == 0 && length == decryptedInputFile.length() &&
@@ -366,7 +378,7 @@ public class IcebergPageSourceProvider
                 .forEach(requiredColumns::add);
 
         long decryptedFileSize = fileSize;
-        if (encryptionKeyMetadata.isPresent()) {
+        if ((fileFormat != IcebergFileFormat.PARQUET || useLegacyParquetDecryption) && encryptionKeyMetadata.isPresent()) {
             try {
                 decryptedFileSize = decryptedInputFile.length();
             }
@@ -388,6 +400,7 @@ public class IcebergPageSourceProvider
                 requiredColumns,
                 effectivePredicate,
                 nameMapping,
+                effectiveParquetFileDecryptionData,
                 partition,
                 partitionKeys);
 
@@ -533,12 +546,18 @@ public class IcebergPageSourceProvider
             Optional<EncryptionManager> encryptionManager)
     {
         TrinoInputFile trinoInputFile = newInputFile(fileSystem, delete.path(), delete.fileSizeInBytes(), delete.encryptionKeyMetadata(), true);
-        TrinoInputFile inputFile = decryptInputFileIfNeeded(
-                trinoInputFile,
-                delete.encryptionKeyMetadata(),
-                encryptionManager);
+        Optional<ParquetFileDecryptionData> parquetFileDecryptionData = delete.parquetFileDecryptionData();
+        boolean useLegacyParquetDecryption = delete.format() == org.apache.iceberg.FileFormat.PARQUET && delete.encryptionKeyMetadata().isPresent();
+        Optional<ParquetFileDecryptionData> effectiveParquetFileDecryptionData = useLegacyParquetDecryption ? Optional.empty() : parquetFileDecryptionData;
+        TrinoInputFile inputFile = trinoInputFile;
+        if (delete.format() != org.apache.iceberg.FileFormat.PARQUET || useLegacyParquetDecryption) {
+            inputFile = decryptInputFileIfNeeded(
+                    trinoInputFile,
+                    delete.encryptionKeyMetadata(),
+                    encryptionManager);
+        }
         long fileSize = delete.fileSizeInBytes();
-        if (delete.encryptionKeyMetadata().isPresent()) {
+        if ((delete.format() != org.apache.iceberg.FileFormat.PARQUET || useLegacyParquetDecryption) && delete.encryptionKeyMetadata().isPresent()) {
             try {
                 fileSize = inputFile.length();
             }
@@ -559,6 +578,7 @@ public class IcebergPageSourceProvider
                 columns,
                 tupleDomain,
                 Optional.empty(),
+                effectiveParquetFileDecryptionData,
                 "",
                 ImmutableMap.of())
                 .pageSource();
@@ -595,6 +615,35 @@ public class IcebergPageSourceProvider
         return new EncryptedTrinoInputFile(inputFile, decryptedInputFile);
     }
 
+    private static Optional<FileDecryptionProperties> parquetFileDecryptionProperties(Optional<ParquetFileDecryptionData> parquetFileDecryptionData)
+    {
+        return parquetFileDecryptionData.map(data -> FileDecryptionProperties.builder()
+                .withKeyRetriever(new StaticKeyRetriever(data.fileEncryptionKey()))
+                .withAadPrefix(data.fileAadPrefix())
+                .build());
+    }
+
+    private record StaticKeyRetriever(byte[] fileEncryptionKey)
+            implements DecryptionKeyRetriever
+    {
+        private StaticKeyRetriever
+        {
+            requireNonNull(fileEncryptionKey, "fileEncryptionKey is null");
+        }
+
+        @Override
+        public Optional<byte[]> getColumnKey(ColumnPath columnPath, Optional<byte[]> keyMetadata)
+        {
+            return Optional.of(fileEncryptionKey);
+        }
+
+        @Override
+        public Optional<byte[]> getFooterKey(Optional<byte[]> keyMetadata)
+        {
+            return Optional.of(fileEncryptionKey);
+        }
+    }
+
     private ReaderPageSourceWithRowPositions createDataPageSource(
             ConnectorSession session,
             TrinoInputFile inputFile,
@@ -608,6 +657,7 @@ public class IcebergPageSourceProvider
             List<IcebergColumnHandle> dataColumns,
             TupleDomain<IcebergColumnHandle> predicate,
             Optional<NameMapping> nameMapping,
+            Optional<ParquetFileDecryptionData> parquetFileDecryptionData,
             String partition,
             Map<Integer, Optional<String>> partitionKeys)
     {
@@ -657,6 +707,7 @@ public class IcebergPageSourceProvider
                     predicate,
                     fileFormatDataSourceStats,
                     nameMapping,
+                    parquetFileDecryptionData,
                     partition,
                     partitionKeys);
             case AVRO -> createAvroPageSource(
@@ -1006,6 +1057,7 @@ public class IcebergPageSourceProvider
             TupleDomain<IcebergColumnHandle> effectivePredicate,
             FileFormatDataSourceStats fileFormatDataSourceStats,
             Optional<NameMapping> nameMapping,
+            Optional<ParquetFileDecryptionData> parquetFileDecryptionData,
             String partition,
             Map<Integer, Optional<String>> partitionKeys)
     {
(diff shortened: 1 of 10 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergTableEncryption.java
```diff
@@ -0,0 +1,138 @@
...
+
+public class TestIcebergTableEncryption
+        extends AbstractTestQueryFramework
+{
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return IcebergQueryRunner.builder()
+                .addIcebergProperty("iceberg.encryption.kms-impl", TestingKmsClient.class.getName())
```
ebyhr: Can we use LocalStack instead? We should also test with real KMS such as [AWS Key Management Service](https://docs.aws.amazon.com/kms/latest/developerguide/overview.html).

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -695,4 +696,17 @@ public IcebergConfig setFileBasedConflictDetectionEnabled(boolean fileBasedConfl
...
         return this;
     }
+
+    public Optional<String> getEncryptionKmsImpl()
+    {
+        return encryptionKmsImpl;
+    }
+
+    @Config("iceberg.encryption.kms-impl")
+    @ConfigDescription("KMS implementation class for Iceberg table encryption")
```
ebyhr: We shouldn't expose Java class name to users. Please introduce a new enum and map to class name internally. 

Also, don't forget updating `iceberg.md`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/functions/tablechanges/TableChangesFunctionHandle.java
```diff
@@ -36,6 +39,7 @@ public record TableChangesFunctionHandle(
         requireNonNull(schemaTableName, "schemaTableName is null");
         requireNonNull(tableSchemaJson, "tableSchemaJson is null");
         columns = ImmutableList.copyOf(requireNonNull(columns, "columns is null"));
+        storageProperties = ImmutableMap.copyOf(requireNonNull(storageProperties, "storageProperties is null"));
```
ebyhr: We don't use `requireNonNull` with `ImmutableMap.copyOf`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
kamijin-fanta: According to the Iceberg documentation, both `encryption.kms-type` and `encryption.kms-impl` appear to be public configuration options. However, in Iceberg 1.10.1 and also in the latest `main` branch, `encryption.kms-type` doesn’t seem to be referenced/used in the code, and users effectively must set `encryption.kms-impl` all the time.

Also, I think we should keep some extensibility here because there are users who use KMS providers outside of major clouds like AWS/GCP (including myself). For consistency with other configuration surfaces (e.g., catalog configs) I’m currently leaning toward accepting the Java class name for now, and then later switching to an enum-based selection once Iceberg formalizes/supports `kms-type` (or an equivalent stable interface). Ideally we can evolve this to an internal enum → class mapping when Iceberg’s interface becomes clearer.

What do you think, @ebyhr? Another reasonable option is to wait to merge until Iceberg finalizes this interface, but I’m trying to balance that with keeping room for non-cloud KMS implementations.

Ref:
https://github.com/apache/iceberg/blob/a6c4e6aef06103db38be064534b7b0be09b3ebca/core/src/main/java/org/apache/iceberg/encryption/EncryptionUtil.java#L38-L80

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergTableEncryption.java
kamijin-fanta: I initially thought we couldn’t override the endpoint. However, after re-reading the code, I realized we can implement our own `AwsClientFactory` and point to it via the catalog property `client.factory`. That would let us call `endpointOverride(...)` in the client builder.

I’ll validate whether we can run the tests against LocalStack with this approach. Thanks!

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
ebyhr: > encryption.kms-type doesn’t seem to be referenced/used in the code

Did you confirm https://github.com/apache/iceberg/pull/15272?. We don't need to wait for 1.11.0 for using KMS type anyway. We can introduce `iceberg.encryption.kms-type` and internally set `encryption.kms-impl`. 

> I’m currently leaning toward accepting the Java class name for now, and then later switching to an enum-based selection once Iceberg formalizes/supports kms-type (or an equivalent stable interface).

The expected order is different. We should begin with the strictest option, such as an enum-based approach, and only allow arbitrary options if that proves too restrictive. Note that we generally avoid such generic options in this project.

kamijin-fanta: Thanks. I read your Iceberg PR.

One question to confirm: are you saying it’s acceptable to implement this logic on the Trino side without waiting for the next Iceberg release? For example, we could expose only the `iceberg.encryption.kms-type` property in Trino, and then internally set Iceberg’s `encryption.kms-impl` to the appropriate implementation class based on that value.

Also, my understanding is that KMS APIs don’t have an industry-standard interface like the S3 API for object storage. In particular, users who enable encryption often build their own KMS around on-prem HSMs or similar systems. Given that context, I assume Iceberg exposes `kms-impl` to allow users to plug in their own implementation. That said, I understand the point that Trino generally shouldn’t expect users to configure a Java class name directly, or, put differently, to distribute arbitrary jars/classes via configuration.

In that case, is there a good way to leave room for users outside cloud environments to customize KMS behavior without forking Trino? And is supporting that kind of extensibility something you’d consider in scope?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionManagerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+
+        if (catalogKmsClient != null) {
+            return catalogKmsClient;
+        }
+
+        synchronized (this) {
+            if (catalogKmsClient != null) {
+                return catalogKmsClient;
+            }
+            catalogKmsClient = EncryptionUtil.createKmsClient(properties);
```
ebyhr, line 88: I don't think calling `EncryptionUtil.createKmsClient` with table properties will work against a real KMS. As I understand it, it requires credentials provided via catalog properties.

This illustrates why we don't allow such generic implementations. It becomes very easy to ship a broken implementation.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -271,6 +279,41 @@ public static BaseTable getIcebergTableWithMetadata(
...
+    {
+        requireNonNull(table, "table is null");
+        if (table instanceof HasTableOperations hasTableOperations && hasTableOperations.operations() instanceof IcebergTableOperations icebergTableOperations) {
+            return icebergTableOperations.fileIoProperties();
+        }
+        FileIO fileIo = table.io();
+        try {
+            return fileIo.properties();
+        }
+        catch (UnsupportedOperationException e) {
```
ebyhr, line 292: We will be able to remove this catch once https://github.com/apache/iceberg/pull/15289 is released.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/encryption/TestIcebergEncryptionManagerFactory.java
```diff
@@ -0,0 +1,53 @@
...
+import org.apache.iceberg.TableProperties;
+import org.apache.iceberg.encryption.PlaintextEncryptionManager;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class TestIcebergEncryptionManagerFactory
```
ebyhr: Please follow https://trino.io/docs/current/develop/tests.html

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionManagerFactory.java
```diff
@@ -0,0 +1,90 @@
...
+public class IcebergEncryptionManagerFactory
+{
+    private final Optional<String> kmsImpl;
+
+    private volatile KeyManagementClient catalogKmsClient;
+
+    @Inject
+    public IcebergEncryptionManagerFactory(IcebergConfig config)
+    {
+        requireNonNull(config, "config is null");
```
ebyhr: We don't use requireNonNull for config classes. https://github.com/trinodb/trino/pull/13940

sopel39:
Thanks for the PR. I've been also working on PME in Iceberg, but haven't published the draft yet. Let me go though this PR, to see if it converges in right direction

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionManagerFactory.java
kamijin-fanta: I’ve removed `iceberg.encryption.kms-impl` for now and added `iceberg.encryption.kms-type` and `iceberg.encryption.kms-properties`.

For the KMS client configuration, table properties are not used. Instead, `iceberg.encryption.kms-properties` is used.

That said, my understanding is that Trino generally has very few places where we allow passing through an arbitrary set of options like this. So I’m considering listing the required properties (for example, the AWS region) and defining explicit, typed fields for them instead.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
```diff
@@ -48,8 +46,9 @@ public class IcebergSplit
     private final long fileSize;
     private final long fileRecordCount;
     private final IcebergFileFormat fileFormat;
+    private final Optional<byte[]> encryptionKeyMetadata;
```
sopel39, line 49: Split should contain fileKey AND aad.

`encryptionKeyMetadata` shouldn't be forwarded to workers are workers will not have access to Iceberg metadata (where keys are stored)


Example
```
    @VisibleForTesting
    record ParquetFileDecryptionData(byte[] fileEncryptionKey, byte[] fileAadPrefix)
    {
        ParquetFileDecryptionData
        {
            requireNonNull(fileEncryptionKey, "fileEncryptionKey is null");
            requireNonNull(fileAadPrefix, "fileAadPrefix is null");
        }
    }

    private static Optional<ParquetFileDecryptionData> parquetFileDecryptionData(
            EncryptedInputFile encryptedInputFile,
            EncryptionManager encryptionManager)
    {
        InputFile inputFile;
        try {
            inputFile = encryptionManager.decrypt(encryptedInputFile);
        }
        catch (RuntimeException e) {
            return Optional.empty();
        }

        if (!(inputFile instanceof NativeEncryptionInputFile nativeEncryptionInputFile)) {
            return Optional.empty();
        }

        NativeEncryptionKeyMetadata nativeKeyMetadata;
        try {
            nativeKeyMetadata = nativeEncryptionInputFile.keyMetadata();
        }
        catch (RuntimeException e) {
            return Optional.empty();
        }

        ByteBuffer encryptionKey = nativeKeyMetadata.encryptionKey();
        ByteBuffer aadPrefix = nativeKeyMetadata.aadPrefix();
        if (encryptionKey == null || aadPrefix == null) {
            return Optional.empty();
        }

        return Optional.of(new ParquetFileDecryptionData(
                ByteBuffers.toByteArray(encryptionKey),
                ByteBuffers.toByteArray(aadPrefix)));
    }
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -527,6 +564,37 @@ public ConnectorPageSource openDeleteFile(
...
+
+    private static TrinoInputFile newInputFile(TrinoFileSystem fileSystem, String path, long fileSize, Optional<byte[]> keyMetadata, boolean useMetadataLength)
+    {
+        if (keyMetadata.isPresent() || !useMetadataLength) {
+            return fileSystem.newInputFile(Location.of(path));
+        }
+        return fileSystem.newInputFile(Location.of(path), fileSize);
+    }
+
+    private static TrinoInputFile decryptInputFileIfNeeded(
```
sopel39, line 604: We already have framework for PME decryption (#24517) that works with Trino native Parquet reader that should be used instead.

Parquet has different encryption modes with footer/column keys, see https://parquet.apache.org/docs/file-format/data-pages/encryption/. I don't think `org.apache.iceberg.encryption.StandardEncryptionManager.StandardDecryptedInputFile` even works with reading encrypted parquet files.

IIUC `encryptionManager.get().decrypt` is for decrypting metadata files.

sopel39:
PTAL at https://github.com/trinodb/trino/pull/28389. Especially split handling, read-path test and `FixedKeyDecryptionKeyRetriever`. I'm fine either PR landing, just using proper PME code with testing

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergTableEncryption.java
```diff
@@ -0,0 +1,148 @@
...
+    {
+        try (TestTable table = new TestTable(
+                getQueryRunner()::execute,
+                "test_encrypted_table_",
+                "(id INT, data VARCHAR) WITH (" +
+                        "format_version = 3, " +
+                        "encryption_key_id = 'test-key', " +
+                        "encryption_data_key_length = 16)")) {
+            String tableName = table.getName();
+            assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a'), (2, 'b'), (3, 'c')", 3);
```
sopel39, line 67: encrypted table is created by Trino too, right? I think it's the reason why 
```
InputFile decryptedInputFile = encryptionManager.get().decrypt(EncryptedFiles.encryptedInput(encryptedInputFile, metadataWithLength));
```
works for Parquet files. However, this should really be tested with code like https://github.com/trinodb/trino/pull/28389/changes#diff-75c880cd52216050277554400af0c70c6a983d40426c347aae427123aa28542bR140 (e.g. use apache parquet writer like
```
        try (DataWriter<Record> writer = Parquet.writeData(encryptedOutputFile)
                .forTable(table)
                .withSpec(table.spec())
                .withPartition(null)
                .withKeyMetadata(encryptedOutputFile.keyMetadata())
                .createWriterFunc(GenericParquetWriter::create)
                .build()) {
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
kamijin-fanta: Thanks for the detailed feedback and for pointing to #24517. I aligned this PR with the PME approach from #24517, following the same pattern used with Trino’s native Parquet reader.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergConfig.java
```diff
@@ -695,4 +697,50 @@ public IcebergConfig setFileBasedConflictDetectionEnabled(boolean fileBasedConfl
...
+
+    @Config("iceberg.encryption.kms-properties")
+    @ConfigDescription("Catalog-level KMS client properties in key=value format")
+    public IcebergConfig setEncryptionKmsProperties(List<String> encryptionKmsProperties)
+    {
+        this.encryptionKmsProperties = Optional.ofNullable(encryptionKmsProperties)
+                .map(ImmutableList::copyOf)
+                .orElseGet(ImmutableList::of);
+        return this;
+    }
```
ebyhr, line 727: In this project, we avoid using arbitrary config properties as much as possible. With that model, it's easy to miss required properties and to permit invalid combinations of settings. It also makes it harder to correctly handle values that contain a `=`. Also, the current code has the risk of leaking credentials since `@ConfigSecuritySensitive` annotation is missing.

You can refer to [IcebergRestCatalogModule](https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogModule.java) for an example of how we handle multiple implementations.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergFileWriterFactory.java
```diff
@@ -214,10 +229,12 @@ private IcebergFileWriter createOrcWriter(
             Schema icebergSchema,
             ConnectorSession session,
             Map<String, String> storageProperties,
-            DataSize stringStatisticsLimit)
+            DataSize stringStatisticsLimit,
+            Optional<EncryptionManager> encryptionManager)
     {
         try {
-            OrcDataSink orcDataSink = OutputStreamOrcDataSink.create(fileSystem.newOutputFile(outputPath));
+            EncryptedOutput encryptedOutput = createOutputFile(fileSystem, outputPath, encryptionManager);
```
ggershinsky, line 236: Iceberg encryption is not supported yet in tables with ORC data files. In the future, the native ORC encryption will need to be leveraged.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergTableEncryption.java
```diff
@@ -0,0 +1,148 @@
...
+            assertUpdate("UPDATE " + tableName + " SET data = 'bb' WHERE id = 2", 1);
+            assertUpdate("DELETE FROM " + tableName + " WHERE id = 1", 1);
+            assertQuery("SELECT id, data FROM " + tableName + " ORDER BY id", "VALUES (2, 'bb'), (3, 'c')");
+
+            assertEncryptedFilesHaveKeyMetadata(tableName);
+        }
+    }
+
+    @Test
+    void testEncryptedOrcWriterValidation()
```
ggershinsky, line 77: Iceberg throws an exception when writing an encrypted ORC table,
https://github.com/apache/iceberg/blob/main/orc/src/main/java/org/apache/iceberg/orc/ORC.java#L116

sopel39:
@kamijin-fanta are you going to work on this PR?

ebyhr:
#28905 takes over this PR as we talked offline.

PR #28355, opened 2026-02-18
Fail analysis if window function uses partitioning on non orderable column

losipiuk:
Currently execution requires symbols used for window function partitioning to be orderable (not just comparable). We use PagesIndex and iterating over it in order to determine partition boundaries.

This commit fixes logic of determining invalid queries.

Regarding release notes: I do not think those are required - given the queries failed also previously and we just improve user experience making failure more understandable.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
@@ -1469,6 +1469,11 @@ public class ExpressionAnalyzer
                     if (!type.isComparable()) {
                         throw semanticException(TYPE_MISMATCH, expression, "%s is not comparable, and therefore cannot be used in window function PARTITION BY", type);
                     }
+                    if (!type.isOrderable()) {
+                        // TODO this is not a hard requirement, just the implication of how we do partitioning right now, using PagesIndex and iterating over ordered values
+                        // to find partition boundaries.
+                        throw semanticException(TYPE_MISMATCH, expression, "%s is not orderable, and therefore cannot be used in window function PARTITION BY", type);
+                    }
                 }
             }
 
--- a/core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
+++ b/core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
@@ -281,6 +281,14 @@ public class TestAnalyzer
     }
 
     @Test
+    public void testNonOrderableWindowPartition()
+    {
+        assertFails("SELECT row_number() OVER (PARTITION BY t.x) FROM (VALUES(MAP(ARRAY['key1', 'key2', 'key3' ], ARRAY[2373, 3463, 45837])) )AS t(x)")
+                .hasErrorCode(TYPE_MISMATCH)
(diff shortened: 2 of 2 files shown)


core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -280,6 +280,14 @@ public void testNonComparableWindowPartition()
                 .hasMessage("line 1:40: HyperLogLog is not comparable, and therefore cannot be used in window function PARTITION BY");
     }
 
+    @Test
+    public void testNonOrderableWindowPartition()
+    {
+        assertFails("SELECT row_number() OVER (PARTITION BY t.x) FROM (VALUES(MAP(ARRAY['key1', 'key2', 'key3' ], ARRAY[2373, 3463, 45837])) )AS t(x)")
+                .hasErrorCode(TYPE_MISMATCH)
+                .hasMessage("line 1:40: map(varchar(4), integer) is not orderable, and therefore cannot be used in window function PARTITION BY");
```
findepi, line 288: What was the failure before the fix?

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -1469,6 +1469,9 @@ private void analyzeWindow(ResolvedWindow window, Context context, Node original
                     if (!type.isComparable()) {
                         throw semanticException(TYPE_MISMATCH, expression, "%s is not comparable, and therefore cannot be used in window function PARTITION BY", type);
                     }
+                    if (!type.isOrderable()) {
+                        throw semanticException(TYPE_MISMATCH, expression, "%s is not orderable, and therefore cannot be used in window function PARTITION BY", type);
```
findepi, line 1475: Maybe worth a comment. In theory it could work, it's just a limitation of the current implementation.

.. That's also maybe why it wasn't checked here in the first place? Where would it fail, if we didn't fail here?

findepi (approved):

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
losipiuk: no failure in alaysis. We got `com.google.common.util.concurrent.UncheckedExecutionException: java.lang.UnsupportedOperationException: map(varchar, varchar) is not orderable code: 65536, name: GENERIC_INTERNAL_ERROR` at runtime

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
losipiuk: > In theory it could work

Yeah - even now we could implement ordering based on for example textual representation of the values. But this will be slow - so not sure if worth the effort. WDYT?

losipiuk: I added a comment

findepi: > even now we could implement ordering based on for example textual representation of the values. 

That may not be correct. For example two equal values may render different strings (thing HashMap, or timestamp tz)

I am not asking to add support for partitioning on non-orderable  types.

Issue #28356, opened 2026-02-18
Connectors need to know whether an identifier is delimited or not


prrvchr:
I am working on issue #17 and if we want to maintain a minimum of compatibility with the old way of working in Trino, then it seems necessary that the connectors know whether they should use a delimited identifier or not in their DDL commands.

And on the Trino side, it's necessary to know, for each connector, how it handles non-delimited identifiers (i.e: lowercase, uppercase, or mixed case).

Then, in this case, it will be possible to compare delimited identifiers with non-delimited identifiers and maintain forward compatibility.

Someone correct me if I'm wrong.

prrvchr:
I close and follow @mosabua's advice with the use of the `NameCanonicalizer` class which allows us to return to the connectors identifiers that can be delimited.

PR #28359, opened 2026-02-18
Add ConnectorIdentifier to Connector

prrvchr:
## Description

I am trying to add a functional comparator for identifiers that support delimiters.
The new class `ConnectorIdentifier` is intended to do so. It also allows the connector to know whether the DDL commands for which it is responsible should be executed with delimited identifiers or not.

- This should resolve https://github.com/trinodb/trino/issues/28356

## Release notes

```markdown
## Section
* Fix some things. ({issue}`28356`)
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorIdentifier.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorIdentifier.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.spi.connector;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import java.util.Objects;
+
+import static java.util.Locale.ENGLISH;
+import static java.util.Objects.requireNonNull;
+
+public class ConnectorIdentifier
+{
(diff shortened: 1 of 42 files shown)


prrvchr:
@wendigo Here's how it's possible to switch to mixed case with Trino while maintaining a minimum of backward compatibility. This requires more changes than expected and is only implemented for views.
Thanks for your feedback.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorIdentifier.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.util.Objects;
+
+import static java.util.Locale.ENGLISH;
+import static java.util.Objects.requireNonNull;
+
+public class ConnectorIdentifier
+{
+    private final String value;
+    private final boolean delimited;
```
martint, line 27: Names coming from connectors don't have a delimited / non-delimited characteristic. They are already canonicalized. The delimited characteristic only pertains to identifiers parsed from a SQL query, and matters only for deciding how to *match* identifiers to names.

prrvchr: > Names coming from connectors don't have a delimited / non-delimited characteristic.

Regarding the identifiers coming from the connectors, I agree. But here we're talking about identifiers coming from Trino tasks and destined for the connectors in order to execute the appropriate DDL commands.

If we want to be able to manage delimited and undelimited identifiers and compare them according to the underlying database rules. For example:

- in HsqlDB (uppercase) `"PUBLIC".Table1` is the same identifier as `public."TABLE1"`

- in SQLite (lowercase) `"main".Table1` is the same identifier as `MAIN."table1"`

If we need to test whether a schema exists before its creation, I don't see any other way without using the `ConnectorIdentifier` class, which allows these comparisons between identifiers.

I'd like you to explain how you plan to solve this problem, because so far I haven't found anything better?

prrvchr: @martint I think I understand what you were trying to tell me.

You're saying that Trino must normalize (ie: canonalize) all undelimited identifiers and that, therefore, **all identifiers received by connectors must be considered as delimited identifiers**.

Can you confirm this? Please excuse my lack of understanding.

prrvchr:
I got a little sidetracked, I'm closing this.

PR #28371, opened 2026-02-19
Clean up partition scoped equality deletes in OPTIMIZE with a partition filter

oskar-szwajkowski:
## Description

Link to iceberg spec scan planning: https://iceberg.apache.org/spec/#scan-planning
Under which it states:
```
An equality delete file must be applied to a data file when all of the following are true:
- The data file's data sequence number is strictly less than the delete's data sequence number
- The data file's partition (both spec id and partition values) is equal [4] to the delete file's partition or the delete file's partition spec is unpartitioned

In general, deletes are applied only to data files that are older and **in the same partition**, except for two special cases:

- Equality delete files stored with an unpartitioned spec are applied as global deletes. Otherwise, delete files do not apply to files in other partitions.
```

Until now, equality deletes required strictly full table optimize, and they wasn't included in split when partition predicate was used. But given the spec it could, unless its global (unpartitioned) equality delete file, which gets skipped as previously all equality delete files would.

## Release notes

```markdown
## Iceberg
* Improve `OPTIMIZE` to clean up partition scoped equality delete files when a partition filter is used.
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
@@ -171,6 +171,7 @@ public class IcebergSplitSource
     private Iterator<FileScanTaskWithDomain> fileTasksIterator = emptyIterator();
 
     private final boolean recordScannedFiles;
+    private final Map<Integer, PartitionSpec> specsById;
     private final int currentSpecId;
     @GuardedBy("this")
     private final ImmutableSet.Builder<DataFileWithDeleteFiles> scannedFiles = ImmutableSet.builder();
@@ -213,6 +214,7 @@ public class IcebergSplitSource
         this.partitionConstraintMatcher = new PartitionConstraintMatcher(constraint);
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.recordScannedFiles = recordScannedFiles;
+        this.specsById = icebergTable.specs();
         this.currentSpecId = icebergTable.spec().specId();
         this.minimumAssignedSplitWeight = minimumAssignedSplitWeight;
         this.projectedBaseColumns = tableHandle.getProjectedColumns().stream()
@@ -342,14 +344,21 @@ public class IcebergSplitSource
         for (FileScanTaskWithDomain fileScanTaskWithDomain : fileScanTasks) {
             FileScanTask wholeFileTask = fileScanTaskWithDomain.fileScanTask();
             if (recordScannedFiles) {
-                // Equality deletes can only be cleaned up if the whole table has been optimized.
+                // Equality deletes can be either global (written with an unpartitioned spec) or partition-scoped
+                // (written with a partitioned spec). Global equality deletes apply across all partitions and can only
+                // be cleaned up when the whole table is optimized.
                 // Equality and position deletes may apply to many files, however position deletes are always local to a partition
                 // https://github.com/apache/iceberg/blob/70c506ebad2dfc6d61b99c05efd59e884282bfa6/core/src/main/java/org/apache/iceberg/deletes/DeleteGranularity.java#L61
-                // OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up position deletes
+                // OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up deletes
                 List<org.apache.iceberg.DeleteFile> fullyAppliedDeletes = wholeFileTask.deletes().stream()
                         .filter(deleteFile -> switch (deleteFile.content()) {
-                            case POSITION_DELETES -> pathDomain.isAll() && fileModifiedTimeDomain.isAll();
-                            case EQUALITY_DELETES -> tableHandle.getEnforcedPredicate().isAll();
+                            case POSITION_DELETES -> isUnconstrainedPathAndTimeDomain();
+                            case EQUALITY_DELETES -> {
+                                if (specsById.get(deleteFile.specId()).isUnpartitioned()) {
+                                    yield tableHandle.getEnforcedPredicate().isAll();
+                                }
+                                yield isUnconstrainedPathAndTimeDomain();
+                            }
                             case DATA -> throw new IllegalStateException("Unexpected delete file: " + deleteFile);
                         })
                         .collect(toImmutableList());
@@ -370,6 +379,11 @@ public class IcebergSplitSource
             }
         }
         return scanTaskBuilder.build().iterator();
+    }
+
+    private boolean isUnconstrainedPathAndTimeDomain()
+    {
+        return pathDomain.isAll() && fileModifiedTimeDomain.isAll();
     }
 
     private synchronized List<FileScanTaskWithDomain> processFileScanTask(TupleDomain<IcebergColumnHandle> dynamicFilterPredicate)
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
@@ -45,6 +45,7 @@ import io.trino.testing.AbstractTestQueryFramework;
 import io.trino.testing.DistributedQueryRunner;
 import io.trino.testing.MaterializedResult;
 import io.trino.testing.MaterializedRow;
+import io.trino.testing.QueryFailedException;
 import io.trino.testing.QueryRunner;
 import io.trino.testing.sql.TestTable;
 import org.apache.iceberg.BaseTable;
@@ -128,6 +129,7 @@ import static org.apache.iceberg.TableProperties.SPLIT_SIZE;
 import static org.apache.iceberg.TableUtil.formatVersion;
 import static org.apache.iceberg.mapping.NameMappingParser.toJson;
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
 import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 import static org.junit.jupiter.api.parallel.ExecutionMode.SAME_THREAD;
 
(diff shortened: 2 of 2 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -342,14 +344,16 @@ private synchronized Iterator<FileScanTaskWithDomain> prepareFileTasksIterator(L
...
+                // (written with a partitioned spec). Global equality deletes apply across all partitions and can only
+                // be cleaned up when the whole table is optimized.
                 // Equality and position deletes may apply to many files, however position deletes are always local to a partition
                 // https://github.com/apache/iceberg/blob/70c506ebad2dfc6d61b99c05efd59e884282bfa6/core/src/main/java/org/apache/iceberg/deletes/DeleteGranularity.java#L61
                 // OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up position deletes
                 List<org.apache.iceberg.DeleteFile> fullyAppliedDeletes = wholeFileTask.deletes().stream()
                         .filter(deleteFile -> switch (deleteFile.content()) {
                             case POSITION_DELETES -> pathDomain.isAll() && fileModifiedTimeDomain.isAll();
-                            case EQUALITY_DELETES -> tableHandle.getEnforcedPredicate().isAll();
+                            case EQUALITY_DELETES -> isEqualityDeleteEligibleForOptimize(deleteFile);
```
findinpath: Let's choose a more generic name.
`isEqualityDeleteEligibleForOptimize` sounds like `prepareFileTasksIterator` is used only for `OPTIMIZE` scenarios.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -342,14 +344,16 @@ private synchronized Iterator<FileScanTaskWithDomain> prepareFileTasksIterator(L
         for (FileScanTaskWithDomain fileScanTaskWithDomain : fileScanTasks) {
             FileScanTask wholeFileTask = fileScanTaskWithDomain.fileScanTask();
             if (recordScannedFiles) {
-                // Equality deletes can only be cleaned up if the whole table has been optimized.
+                // Equality deletes can be either global (written with an unpartitioned spec) or partition-scoped
+                // (written with a partitioned spec). Global equality deletes apply across all partitions and can only
```
findinpath, line 348: // Global equality deletes apply across all partitions and can only
// be cleaned up when the whole table is optimized.

can be removed - the comment is already present in `isEqualityDeleteEligibleForOptimize`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -372,6 +376,21 @@ private synchronized Iterator<FileScanTaskWithDomain> prepareFileTasksIterator(L
...
+        if (deleteSpec.isUnpartitioned()) {
+            // Global delete applies across all partitions — only safe to clean up
+            // when the entire table is optimized
+            return tableHandle.getEnforcedPredicate().isAll();
+        }
+        // Partition-scoped delete only applies within its partition.
+        // OPTIMIZE uses enforced predicates that select whole partitions,
+        // so we can clean up if we're processing all files in the partition
+        // (no path or time filtering that would skip files).
+        return pathDomain.isAll() && fileModifiedTimeDomain.isAll();
```
findinpath, line 386: Maybe we can extract to a separate method this check to avoid code duplication 

```
pathDomain.isAll() && fileModifiedTimeDomain.isAll()
```

See `case POSITION_DELETES -> pathDomain.isAll() && fileModifiedTimeDomain.isAll();`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -742,6 +730,200 @@ public void testOptimizingWholeTableRemovesDeleteFiles()
...
+
+            assertQuery("SELECT count(*) FROM \"" + testTable.getName() + "$files\" WHERE content = " + EQUALITY_DELETES.id(), "VALUES 0");
+
+            assertThat(query("SELECT * FROM " + testTable.getName()))
+                    .matches("SELECT * FROM nation WHERE regionkey != 1");
+        }
+    }
+
+    @Test
+    public void testOptimizeWithInnerPartitionPredicateRemovesPartitionScopedEqualityDeletesWithMultiplePartitions()
```
findinpath: testOptimizeWithInnerPartitionPredicateRemovesPartitionScopedEqualityDeletesWithMultiplePartitions -> testOptimizeInnerPartitionsRemovesPartitionScopedEqualityDeletes

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -742,6 +730,200 @@ public void testOptimizingWholeTableRemovesDeleteFiles()
         }
     }
 
+    @Test
+    public void testOptimizeWithPartitionPredicateRemovesPartitionScopedEqualityDeletes()
```
findinpath: testOptimizeWithPartitionPredicateRemovesPartitionScopedEqualityDeletes -> testOptimizePartitionRemovesPartitionScopedEqualityDeletes

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -372,6 +376,21 @@ private synchronized Iterator<FileScanTaskWithDomain> prepareFileTasksIterator(L
...
 
+    private boolean isEqualityDeleteEligibleForOptimize(org.apache.iceberg.DeleteFile deleteFile)
+    {
+        PartitionSpec deleteSpec = specsById.get(deleteFile.specId());
+        if (deleteSpec.isUnpartitioned()) {
+            // Global delete applies across all partitions — only safe to clean up
+            // when the entire table is optimized
+            return tableHandle.getEnforcedPredicate().isAll();
+        }
+        // Partition-scoped delete only applies within its partition.
```
JohnFitzpatrick44: We can remove this comment - it basically duplicates the one on L350-352:
```
// Equality and position deletes may apply to many files, however position deletes are always local to a partition
// https://github.com/apache/iceberg/blob/70c506ebad2dfc6d61b99c05efd59e884282bfa6/core/src/main/java/org/apache/iceberg/deletes/DeleteGranularity.java#L61
// OPTIMIZE supports only enforced predicates which select whole partitions, so if there is no path or fileModifiedTime predicate, then we can clean up position deletes
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
JohnFitzpatrick44: Nit - Seems like a simple enough piece of logic to inline, and you could remove the comments as the one on line 347 covers it.
```
case EQUALITY_DELETES -> {
    if (specsById.get(deleteFile.specId()).isUnpartitioned()) {
        yield tableHandle.getEnforcedPredicate().isAll();
    }
    yield pathDomain.isAll() && fileModifiedTimeDomain.isAll();
}
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
oskar-szwajkowski: good point

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
oskar-szwajkowski: done, and also extracted common `pathDomain.isAll() && fileModifiedTimeDomain.isAll()` to method

oskar-szwajkowski: inlined that with extracing some common domain matching logic

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -742,6 +730,200 @@ public void testOptimizingWholeTableRemovesDeleteFiles()
...
+
+            assertQuery("SELECT count(*) FROM \"" + testTable.getName() + "$files\" WHERE content = " + EQUALITY_DELETES.id(), "VALUES 1");
+
+            assertThat(query("SELECT * FROM " + testTable.getName()))
+                    .matches("SELECT * FROM nation WHERE name != 'ARGENTINA'");
+        }
+    }
+
+    @Test
+    public void testOptimizeWithPartitionDoesNotRemoveGlobalPartitionColumnEqualityDeletes()
```
raunaqmorarka: testOptimizeWithGlobalEqualityDeleteOnPartitionedColumn

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -742,6 +730,200 @@ public void testOptimizingWholeTableRemovesDeleteFiles()
...
+
+            assertQuery("SELECT count(*) FROM \"" + testTable.getName() + "$files\" WHERE content = " + EQUALITY_DELETES.id(), "VALUES 1");
+
+            assertThat(query("SELECT * FROM " + testTable.getName()))
+                    .matches("SELECT CAST(country AS VARCHAR), CAST(age AS INT), CAST(name AS VARCHAR) FROM (VALUES('US', 30, 'Bob'), ('US', 20, 'Alice'), ('CA', 30, 'Carol'), ('CA', 15, 'Sarah'), ('UK', 30, 'Tom')) AS t(country, age, name)");
+        }
+    }
+
+    @Test
+    public void testOptimizeWithPartitionDoesNotRemoveGlobalRegularColumnEqualityDeletes()
```
raunaqmorarka: testOptimizeWithGlobalEqualityDelete

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
raunaqmorarka: Can this test just be included in the above test as an additional assertion ?

oskar-szwajkowski: actually it can, done

raunaqmorarka (approved):

Issue #28373, opened 2026-02-19
Resource groups: add per-query limits, driver-level admission control, and planning concurrency limit


kaveti:
*Problem*

We run Trino at scale with shared clusters serving multiple teams — analysts running ad-hoc queries, ETL pipelines, dashboards, etc. We use resource groups heavily to isolate these workloads from each other. Over time we've hit a few gaps in what resource groups can actually control today, and I wanted to describe them here before submitting a PR.

**1. No way to limit total drivers within a resource group**
This is the biggest one for us. Right now resource groups can limit concurrency (number of running queries) and memory, but there's nothing that limits the total number of drivers across all queries in a group.

In practice, this means a resource group with hardConcurrencyLimit: 5 could have 5 queries each with 200+ drivers, completely saturating task slots cluster-wide. Meanwhile other resource groups with lighter queries get starved because all the task slots are occupied.

We want to be able to say: "this resource group can have at most N total active drivers (running + queued + blocked) across all its queries". When the limit is reached, new queries queue instead of starting. This is a fundamentally different axis of control than query count or memory.

**2. Per-query limits should be configurable at the resource group level**
Trino has query.max-memory, query.max-cpu-time, and query.max-scan-physical-bytes as session/system properties. These work fine as global defaults, but when you want different per-query limits for different workloads, you need a session property manager to map users/sources to different values. That's a separate plugin with its own config, and in practice it's awkward to keep synchronized with resource group definitions.

For example, we want:

Ad-hoc group: no single query should use more than 10GB memory or scan more than 500GB
ETL group: queries can use up to 50GB memory and scan up to 5TB
Dashboard group: queries limited to 2GB memory and 30s CPU time
Today you'd configure these limits via a session property manager, referencing the same user/source patterns that your resource group selectors already match. It's duplicated config that has to stay in sync. It would be much more natural to just put these limits directly in the resource group definition:

```
{
  "name": "adhoc",
  "softMemoryLimit": "80%",
  "hardConcurrencyLimit": 20,
  "maxQueued": 100,
  "perQueryMemoryLimit": "10GB",
  "perQueryScanLimit": "500GB",
  "perQueryCpuLimit": "5m"
}
```
The enforcement semantics are the same — queries exceeding the limit get killed

[...]

PR #28374, opened 2026-02-19
Resource groups: add per-query limits, driver-level admission control, and planning concurrency limit

kaveti:
## Description

**Motivation**: In shared multi-tenant Trino clusters, resource groups today can limit query
count, memory, and CPU at the group level — but can't control driver-level resource consumption
or enforce per-query limits without a separate session property manager plugin. This forces
admins to maintain duplicated configuration across resource groups and session property managers
for workload isolation.

This PR Adds three categories of new controls to resource groups:

1. **Per-query resource limits** (`perQueryMemoryLimit`, `perQueryCpuLimit`, `perQueryScanLimit`) —
   Queries exceeding these limits within a resource group are killed. Enforced during
   **stageResourceUsage()** on the same periodic cadence as existing group-level resource tracking.
   Complements existing session-level properties (`query.max-memory`, `query.max-cpu-time`,
   `query.max-scan-physical-bytes`) by allowing workload-level policy directly in the resource
   group config without needing a separate session property manager.

2. **Hard total driver limit** (`hardTotalDriverLimit`) —
   Caps total active drivers (running + queued + blocked) across all queries in a group. New
   queries queue when the limit is reached. Driver counts aggregate hierarchically from subgroups.
   Addresses a real gap: a group with 5 concurrent queries can saturate cluster-wide task slots
   if each query has hundreds of drivers, starving other groups.

3. **Hard planning concurrency limit** (`hardPlanningConcurrencyLimit`) —
   Limits queries in PLANNING/STARTING state concurrently within a group. Acts as a smoothing
   mechanism for planning bursts on the coordinator.

All new limits default to MAX_VALUE (effectively disabled).

More Details in here : https://github.com/trinodb/trino/issues/28373

## Release notes

```markdown
## Section
* Fix some things. ({[issue](https://github.com/trinodb/trino/issues/28373)}`28373`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
+++ b/core/trino-main/src/main/java/io/trino/execution/resourcegroups/InternalResourceGroup.java
@@ -20,7 +20,9 @@ import com.google.errorprone.annotations.ThreadSafe;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import io.airlift.stats.CounterStat;
 import io.trino.execution.ManagedQueryExecution;
+import io.trino.execution.QueryState;
 import io.trino.execution.resourcegroups.WeightedFairQueue.Usage;
+import io.trino.server.BasicQueryStats;
 import io.trino.server.QueryStateInfo;
 import io.trino.server.ResourceGroupInfo;
 import io.trino.spi.TrinoException;
@@ -55,6 +57,9 @@ import static io.airlift.units.Duration.succinctDuration;
 import static io.trino.SystemSessionProperties.getQueryPriority;
 import static io.trino.execution.resourcegroups.ResourceUsage.ZERO;
 import static io.trino.server.QueryStateInfo.createQueryStateInfo;
+import static io.trino.spi.StandardErrorCode.EXCEEDED_CPU_LIMIT;
+import static io.trino.spi.StandardErrorCode.EXCEEDED_LOCAL_MEMORY_LIMIT;
+import static io.trino.spi.StandardErrorCode.EXCEEDED_SCAN_LIMIT;
 import static io.trino.spi.StandardErrorCode.INVALID_RESOURCE_GROUP;
 import static io.trino.spi.resourcegroups.ResourceGroupState.CAN_QUEUE;
 import static io.trino.spi.resourcegroups.ResourceGroupState.CAN_RUN;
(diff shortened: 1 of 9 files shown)


dain:
I understand your desire to make this change, but this is not really what the Trino project is focused on. The resource groups are really designed for normal people running normal clusters, and exposing stuff like driver limits to end-users is philosophically, something we try to avoid, specifically asking users questions that they are not qualified to answer.  If the driver limits should be adjusted, then the engine should be adjusting them for users. If you would like to change this in your environment, you should be able to write your own resource groups, plug in, and build the server.

kaveti:
@dain Thanks for the feedback — that's a fair point. I agree that exposing driver-level details like total active drivers or planning state counts isn't the right abstraction for resource groups. I'll drop hardTotalDriverLimit and hardPlanningConcurrencyLimit or may be enable them when advance settings for advance user.

Would you be open to a narrower version that only adds per-query memory, CPU, and scan limits? These are already user-facing concepts with existing session properties (query.max-memory, query.max-cpu-time, query.max-scan-physical-bytes). The motivation is letting admins set different per-query limits per resource group directly — e.g., "ad-hoc gets 10GB per query, ETL gets 50GB" — without needing a separate session property manager with duplicated selector patterns.

Happy to rework the PR or close it if this still doesn't align with the project's direction.

benrifkind:
Piggybacking on this, I also would like a way to limit memory usage per resource group. I current ran into an issue where a user in a resource group had a query that went over its parent group's softMemoryLimit. That prevented other users in the group from running and caused a back up on the cluster. A more granular way to restrict memory usage would, I think, prevent this issue. I believe preventing this issue with the current setup would require setting cluster wide query memory limit restrictions. That's more restrictive than what I'm hoping to accomplish.

RoeyoOgen:
@kaveti have you implemented the RG plugin yourself?

kaveti:
> @kaveti have you implemented the RG plugin yourself?

@RoeyoOgen no  we did not implemented it.  but i felt it would be better if the trino would support this as built in feature for admin user to configure there clusters with proper guardrails.

kaveti:
There's growing interest in supporting per-query limits on resource groups (benrifkind raised the noisy-neighbor case above, and I've seen the same ask elsewhere). Would appreciate another look when you have a moment — @findepi @findinpath  @dain @wendigo @Praveen2112 @electrum 
 
Quick summary of where it stands: the per-query memory/CPU/scan limits cap a single query within a group, which the existing group-level limits can't do — one runaway query can currently starve its peers even when the group is under its 
 aggregate limit. I'm happy to split out the driver-level and planning-concurrency limits into a separate PR so the per-query limits can move forward on their own.

RoeyoOgen:
@kaveti have you thought about adding a max_returned_rows config?

RoeyoOgen:
also ping me on slack when you want a review

PR #28378, opened 2026-02-19
WIP: Grow VariableWithBlockBuilder by average position size

wendigo:
## Release notes

```markdown
## Section
```


core/trino-spi/src/main/java/io/trino/spi/block/VariableWidthBlockBuilder.java
```diff
@@ -396,6 +396,17 @@ private void ensureCapacity(int capacity)
         int newSize = calculateNewArraySize(capacity, initialEntryCount);
         valueIsNull = Arrays.copyOf(valueIsNull, newSize);
         offsets = Arrays.copyOf(offsets, newSize + 1);
+
+        // proactively grow bytes array based on observed average entry size
+        int currentBytes = offsets[positionCount];
+        if (positionCount > 0 && currentBytes > 0) {
+            int averageBytesPerEntry = currentBytes / positionCount;
```
pettyjamesm, line 403: Consider using `double` instead of integer division. The rounding down behavior seems likely to result in underestimating the target size in most cases which will require another growing copy.

PR #28380, opened 2026-02-19
Mark more functions as infallible

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/type/DecimalCasts.java
+++ b/core/trino-main/src/main/java/io/trino/type/DecimalCasts.java
@@ -78,29 +78,31 @@ import static java.nio.charset.StandardCharsets.UTF_8;
 
 public final class DecimalCasts
 {
-    public static final SqlScalarFunction DECIMAL_TO_BOOLEAN_CAST = castFunctionFromDecimalTo(BOOLEAN.getTypeSignature(), "shortDecimalToBoolean", "longDecimalToBoolean");
-    public static final SqlScalarFunction BOOLEAN_TO_DECIMAL_CAST = castFunctionToDecimalFrom(BOOLEAN.getTypeSignature(), "booleanToShortDecimal", "booleanToLongDecimal");
-    public static final SqlScalarFunction DECIMAL_TO_BIGINT_CAST = castFunctionFromDecimalTo(BIGINT.getTypeSignature(), "shortDecimalToBigint", "longDecimalToBigint");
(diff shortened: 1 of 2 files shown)


core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -757,11 +757,12 @@ protected Node visitShowFunctions(ShowFunctions node, Void context)
                     .put("argument_types", "Argument Types")
                     .put("function_type", "Function Type")
                     .put("deterministic", "Deterministic")
+                    .put("infallible", "Infallible")
```
martint: This is an internal detail, not something we want to expose to users.

wendigo: Dropped

raunaqmorarka (approved):

core/trino-main/src/main/java/io/trino/operator/scalar/ArrayAllMatchFunction.java
```diff
@@ -31,7 +31,7 @@
 import static java.lang.Boolean.FALSE;
 
 @Description("Returns true if all elements of the array match the given predicate")
-@ScalarFunction("all_match")
+@ScalarFunction(value = "all_match", neverFails = true)
```
pettyjamesm, line 34: Isn't `neverFails = true` only true if the function passed never fails? Or is the idea that the analyzer will see `neverFails = false` on the inner function call itself and so we don't need to worry about representing that in this wrapping function call.

martint: `neverFails = true` for this function is incorrect. For functions that take lambda expressions, this characteristic depends on what the function actually does internally. E.g., the function may catch errors and convert them to nulls. The analyzer can't determine that.

martint: Good catch!

wendigo: @martint how is it consistent with: https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java#L70

which says "Lambda" may not fail

martint: Lamda expressions being able to fail is irrelevant. It depends on the semantics of the function that calls them. 

That's why in IrExpressions they are marked as "may not fail" -- the evaluation of the lambda argument cannot fail. It produces a function. What can fail is the invocation of such function, and that happens within the function that takes it as an argument.

PR #28383, opened 2026-02-20
Remove deprecated system.rollback_to_snapshot from Iceberg

ebyhr:
## Description

We added the alternative `rollback_to_snapshot` table procedure over a year ago.

## Release notes

```markdown
## Iceberg
* {{breaking}} Remove deprecated `system.rollback_to_snapshot`.
  Use `rollback_to_snapshot` table procedure instead. ({issue}`28383`)
```


chenjian2664 (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/procedure/RollbackToSnapshotProcedure.java
```diff
@@ -1,82 +0,0 @@
...
-import java.lang.invoke.MethodHandle;
-
-import static io.trino.plugin.base.util.Procedures.checkProcedureArgument;
-import static io.trino.spi.type.BigintType.BIGINT;
-import static io.trino.spi.type.VarcharType.VARCHAR;
-import static java.lang.invoke.MethodHandles.lookup;
-import static java.util.Objects.requireNonNull;
-
-@Deprecated
-public class RollbackToSnapshotProcedure
```
chenjian2664, line 35: Do we need to update the doc ?

ebyhr: Do we still have this procedure in our documentation? https://trino.io/docs/current/connector/iceberg.html doesn't contain that.

PR #28387, opened 2026-02-20
Rename ScalarFunction field and use enum values

wendigo:
This will allow to gradually mark all functions as either safe (never fails) or unsafe (possibly fails) while keeping a default being "undefined" which means that we can capture functions that are not yet described.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/function/ScalarFunction.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/function/ScalarFunction.java
@@ -18,6 +18,7 @@ import com.google.errorprone.annotations.Keep;
 import java.lang.annotation.Retention;
 import java.lang.annotation.Target;
 
+import static io.trino.spi.function.ScalarFunction.MayFail.UNDEFINED;
 import static java.lang.annotation.ElementType.METHOD;
 import static java.lang.annotation.ElementType.TYPE;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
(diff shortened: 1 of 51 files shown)


wendigo:
@findepi I've implemented what we've talked about offline

core/trino-spi/src/main/java/io/trino/spi/function/ScalarFunction.java
```diff
@@ -39,5 +40,12 @@
...
      * for any possible combination of input arguments.
      */
-    boolean neverFails() default false;
+    MayFail mayFail() default UNDEFINED;
+
+    enum MayFail
+    {
+        NEVER,
+        POSSIBLY,
+        UNDEFINED;
```
martint, line 49: Undefined doesn't mean anything. The optimizer can't make use of that.

The "neverFails" attribute is a conservative estimate. The optimizer only cares about knowing whether the function will never fail. If it can fail, or if it's unknown whether it can fail, it's all the same to it.

martint: So, defaulting to "neverFails = false" is a safe choice, always.

wendigo: @martint yes and no. We want to mark every function as either never failing or possibly failing. With true/false you can't model "not yet reviewed"

martint: We shouldn't pollute the API just to capture whether an existing function has been reviewed. For that, let's compile a list out of band and create an issue to track.

For new functions, the author needs to think about that characteristic when the function is introduced, just like whether the function is hidden or deterministic.

findepi: > For new functions, the author needs to think about that characteristic when the function is introduced, just like whether the function is hidden or deterministic

They author will unlikely to think about this -- unless this is one of us involved on this PR -- and reviewer will unlikely to notice lack of such attribute (possibly so even if it is one of us involved on this PR).

An API that's easier to get right is better than an API that's harder to get right.

cc @losipiuk

PR #26640, opened 2025-09-16
Update Iceberg to 1.11.0

ebyhr:
## Description

Update Iceberg to 1.11.0

* Fixes #23601
* Fixes #20511
* Fixes #29479

## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #28389, opened 2026-02-20
Add read support for Iceberg Parquet encryption in the Iceberg connector.

sopel39:
## Description

Add read support for Iceberg Parquet encryption in the Iceberg connector.

Iceberg table encryption protects the confidentiality and integrity of table data at rest
in untrusted storage. Data files, delete files, manifest files, and manifest list files are
encrypted and tamper-proofed before being sent to storage, using per-file encryption keys
and an envelope encryption scheme backed by a KMS (Key Management Service).

This PR wires the Iceberg encryption subsystem into Trino so that tables with the
`encryption.key-id` table property can be read. Writing to encrypted tables is not yet
supported and is explicitly blocked.

### What's included

- Catalog-level KMS configuration via the `iceberg.encryption.kms-type` property,
  which accepts `AWS` or `GCP`
- `EncryptionManagerFactory` that creates an Iceberg `EncryptionManager` from table
  properties, used to unwrap per-file DEKs via the configured KMS
- `EncryptionAwareFileIO` wrapper so that table operations (manifest/metadata reads)
  go through the Iceberg `EncryptingFileIO`
- Propagation of Parquet decryption data (file encryption key + AAD prefix) through
  splits to the Parquet reader
- Plaintext file detection: by default, reading an unencrypted file in an encrypted
  table fails; overridable via the `iceberg.encryption.plaintext-files-allowed-for-encrypted-tables`
  catalog property or the `plaintext_files_allowed_for_encrypted_tables` catalog session
  property
- Write blocking on all data-producing paths (INSERT, MERGE, OPTIMIZE, CREATE OR REPLACE,
  add_files, materialized view refresh)
- Support for encrypted delete files (position and equality deletes) and table changes

See https://iceberg.apache.org/docs/nightly/encryption/ for details on
Iceberg's encryption model.

## Release notes

Release notes are required, with the following suggested text:

## Iceberg
* Add read support for encrypted Parquet tables in Iceberg. ({issue}`28389`)

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -259,6 +259,18 @@ implementation is used:
   - Enable bucket-aware execution. This allows the engine to use physical
     bucketing information to optimize queries by reducing data exchanges.
   - `true`
+* - `iceberg.encryption.kms-type`
+  - Key Management Service type for
+    [Iceberg table encryption](https://iceberg.apache.org/docs/nightly/encryption/).
+    Possible values are `AWS` and `GCP`. Required to read encrypted tables.
+    Writing to encrypted tables is not supported.
+  -
+* - `iceberg.encryption.plaintext-files-allowed-for-encrypted-tables`
+  - Allow reading unencrypted files in tables with encryption enabled. When set
+    to `false`, an error is raised if a file with encryption key metadata is not
+    actually encrypted. The equivalent catalog session property is
+    `plaintext_files_allowed_for_encrypted_tables`.
+  - `false`
 :::
 
 (iceberg-fte-support)=
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -262,7 +262,7 @@ implementation is used:
 * - `iceberg.encryption.kms-type`
   - Key Management Service type for
     [Iceberg table encryption](https://iceberg.apache.org/docs/nightly/encryption/).
-    Possible values are `AWS` and `GCP`. Required to read encrypted tables.
+    Possible values are `AWS`, `AZURE`, and `GCP`. Required to read encrypted tables.
     Writing to encrypted tables is not supported.
   -
 * - `iceberg.encryption.plaintext-files-allowed-for-encrypted-tables`
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/crypto/FileDecryptionProperties.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/crypto/FileDecryptionProperties.java
@@ -22,12 +22,18 @@ public class FileDecryptionProperties
     private final DecryptionKeyRetriever keyRetriever;
     private final Optional<byte[]> aadPrefix;
     private final boolean checkFooterIntegrity;
+    private final boolean plaintextFilesAllowed;
 
-    private FileDecryptionProperties(DecryptionKeyRetriever keyRetriever, Optional<byte[]> aadPrefix, boolean checkFooterIntegrity)
+    private FileDecryptionProperties(
+            DecryptionKeyRetriever keyRetriever,
+            Optional<byte[]> aadPrefix,
+            boolean checkFooterIntegrity,
+            boolean plaintextFilesAllowed)
     {
         this.keyRetriever = requireNonNull(keyRetriever, "keyRetriever is null");
         this.aadPrefix = requireNonNull(aadPrefix, "aadPrefix is null");
         this.checkFooterIntegrity = checkFooterIntegrity;
+        this.plaintextFilesAllowed = plaintextFilesAllowed;
     }
 
     public static Builder builder()
@@ -50,11 +56,17 @@ public class FileDecryptionProperties
         return checkFooterIntegrity;
     }
 
+    public boolean isPlaintextFilesAllowed()
+    {
+        return plaintextFilesAllowed;
+    }
+
     public static class Builder
     {
         private DecryptionKeyRetriever keyRetriever;
         private Optional<byte[]> aadPrefix = Optional.empty();
         private boolean checkFooterIntegrity = true;
+        private boolean plaintextFilesAllowed;
 
         public Builder withKeyRetriever(DecryptionKeyRetriever keyRetriever)
         {
@@ -74,9 +86,15 @@ public class FileDecryptionProperties
             return this;
         }
 
+        public Builder withPlaintextFilesAllowed()
+        {
+            this.plaintextFilesAllowed = true;
+            return this;
+        }
+
         public FileDecryptionProperties build()
         {
-            return new FileDecryptionProperties(keyRetriever, aadPrefix, checkFooterIntegrity);
+            return new FileDecryptionProperties(keyRetriever, aadPrefix, checkFooterIntegrity, plaintextFilesAllowed);
         }
     }
 }
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
@@ -228,10 +228,10 @@ public final class MetadataReader
 
         return org.apache.parquet.column.statistics.Statistics
                 .getBuilderForReading(columnStatistics.type())
-                       .withMin(min)
-                       .withMax(max)
-                       .withNumNulls(!columnStatistics.isNumNullsSet() && statistics.isSetNull_count() ? statistics.getNull_count() : columnStatistics.getNumNulls())
-                       .build();
+                .withMin(min)
+                .withMax(max)
+                .withNumNulls(!columnStatistics.isNumNullsSet() && statistics.isSetNull_count() ? statistics.getNull_count() : columnStatistics.getNumNulls())
+                .build();
     }
 
     private static boolean isAscii(byte b)
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
@@ -27,6 +27,7 @@ import io.trino.parquet.crypto.AesCipherUtils;
 import io.trino.parquet.crypto.AesGcmEncryptor;
 import io.trino.parquet.crypto.FileDecryptionContext;
 import io.trino.parquet.crypto.FileDecryptionProperties;
+import io.trino.parquet.crypto.ParquetCryptoException;
 import io.trino.parquet.metadata.FileMetadata;
 import io.trino.parquet.metadata.ParquetMetadata;
 import org.apache.parquet.CorruptStatistics;
@@ -139,12 +140,19 @@ public final class MetadataReader
         }
 
         FileMetaData fileMetaData = readFileMetaData(metadataStream, footerDecryptor, aad);
-        if (!encryptedFooterMode && fileDecryptionProperties.isPresent() && fileMetaData.isSetEncryption_algorithm()) {
-            // footer is not encrypted, but some columns might be encrypted
-            decryptionContext = Optional.of(new FileDecryptionContext(dataSource.getId(), fileDecryptionProperties.get(), fileMetaData.getEncryption_algorithm(), Optional.ofNullable(fileMetaData.getFooter_signing_key_metadata())));
-            if (fileDecryptionProperties.get().isCheckFooterIntegrity()) {
-                // verify footer integrity
-                verifyFooterIntegrity(dataSource, metadataStream, decryptionContext.get(), metadataLength);
+        if (!encryptedFooterMode && fileDecryptionProperties.isPresent()) {
+            if (!fileMetaData.isSetEncryption_algorithm()) {
+                // Plaintext file — detect files that were not encrypted by mistake
+                if (!fileDecryptionProperties.get().isPlaintextFilesAllowed()) {
+                    throw new ParquetCryptoException("Applying decryptor on plaintext file: %s", dataSource.getId());
+                }
+            }
+            else {
+                // Footer is not encrypted, but some columns might be encrypted
+                decryptionContext = Optional.of(new FileDecryptionContext(dataSource.getId(), fileDecryptionProperties.get(), fileMetaData.getEncryption_algorithm(), Optional.ofNullable(fileMetaData.getFooter_signing_key_metadata())));
+                if (fileDecryptionProperties.get().isCheckFooterIntegrity()) {
+                    verifyFooterIntegrity(dataSource, metadataStream, decryptionContext.get(), metadataLength);
+                }
             }
         }
 
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -202,6 +202,7 @@ import org.apache.iceberg.UpdatePartitionSpec;
 import org.apache.iceberg.UpdateProperties;
 import org.apache.iceberg.UpdateSchema;
 import org.apache.iceberg.UpdateStatistics;
+import org.apache.iceberg.encryption.PlaintextEncryptionManager;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.CommitStateUnknownException;
@@ -768,11 +769,6 @@ public class IcebergMetadata
         TableMetadata metadata = table.operations().current();
         if (metadata.formatVersion() < 3) {
             return;
-        }
-
-        // Reject Iceberg table encryption
-        if (!metadata.encryptionKeys().isEmpty() || snapshot.keyId() != null || metadata.properties().containsKey("encryption.key-id")) {
-            throw new TrinoException(NOT_SUPPORTED, "Iceberg table encryption is not supported");
         }
     }
 
@@ -1477,6 +1473,7 @@ public class IcebergMetadata
                     throw new TrinoException(INVALID_TABLE_PROPERTY, format("The provided location '%s' does not match the existing table location '%s'", providedTableLocation.get(), icebergTable.location()));
                 }
                 validateNotModifyingOldSnapshot(table, icebergTable);
+                validateNotEncryptedForWrite(icebergTable);
                 tableLocation = icebergTable.location();
                 List<PartitionField> existingPartitionFields = getAllPartitionFields(icebergTable);
                 transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties, existingPartitionFields);
@@ -1606,6 +1603,7 @@ public class IcebergMetadata
 
         validateNotModifyingOldSnapshot(table, icebergTable);
         validateTableForTrino(icebergTable, getCurrentSnapshotId(icebergTable));
+        validateNotEncryptedForWrite(icebergTable);
 
         beginTransaction(icebergTable);
 
@@ -2093,6 +2091,7 @@ public class IcebergMetadata
         BaseTable icebergTable = catalog.loadTable(session, table.getSchemaTableName());
 
         validateNotModifyingOldSnapshot(table, icebergTable);
+        validateNotEncryptedForWrite(icebergTable);
 
         verifyTableVersionForExecute(OPTIMIZE, OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION, icebergTable);
 
@@ -2638,6 +2637,7 @@ public class IcebergMetadata
     {
         IcebergAddFilesHandle addFilesHandle = (IcebergAddFilesHandle) executeHandle.procedureHandle();
         Table table = catalog.loadTable(session, executeHandle.schemaTableName());
+        validateNotEncryptedForWrite(table);
         TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), table.io().properties());
         long addedDataFiles = addFiles(
                 session,
@@ -2655,6 +2655,7 @@ public class IcebergMetadata
     {
         IcebergAddFilesFromTableHandle addFilesHandle = (IcebergAddFilesFromTableHandle) executeHandle.procedureHandle();
         Table table = catalog.loadTable(session, executeHandle.schemaTableName());
+        validateNotEncryptedForWrite(table);
         TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), table.io().properties());
         long addedDataFiles = addFilesFromTable(
                 session,
@@ -3549,6 +3550,7 @@ public class IcebergMetadata
 
         Table icebergTable = catalog.loadTable(session, table.getSchemaTableName());
         validateNotModifyingOldSnapshot(table, icebergTable);
+        validateNotEncryptedForWrite(icebergTable);
 
         beginTransaction(icebergTable);
 
@@ -3576,6 +3578,13 @@ public class IcebergMetadata
     {
         if (table.getSnapshotId().isPresent() && (table.getSnapshotId().getAsLong() != icebergTable.currentSnapshot().snapshotId())) {
             throw new TrinoException(NOT_SUPPORTED, "Modifying old snapshot is not supported in Iceberg");
+        }
+    }
+
+    private static void validateNotEncryptedForWrite(Table table)
+    {
+        if (!(table.encryption() instanceof PlaintextEncryptionManager)) {
+            throw new TrinoException(NOT_SUPPORTED, "Writing to encrypted Iceberg tables is not supported");
         }
     }
 
@@ -4259,6 +4268,7 @@ public class IcebergMetadata
         checkState(fromSnapshotForRefresh.isEmpty(), "From Snapshot must be empty at the start of MV refresh operation.");
         IcebergTableHandle table = (IcebergTableHandle) tableHandle;
         Table icebergTable = catalog.loadTable(session, table.getSchemaTableName());
+        validateNotEncryptedForWrite(icebergTable);
         beginTransaction(icebergTable);
 
         Optional<String> dependencies = Optional.ofNullable(icebergTable.currentSnapshot())
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -185,6 +185,7 @@ import org.apache.iceberg.UpdatePartitionSpec;
 import org.apache.iceberg.UpdateProperties;
 import org.apache.iceberg.UpdateSchema;
 import org.apache.iceberg.UpdateStatistics;
+import org.apache.iceberg.encryption.PlaintextEncryptionManager;
 import org.apache.iceberg.exceptions.AlreadyExistsException;
 import org.apache.iceberg.exceptions.CommitFailedException;
 import org.apache.iceberg.exceptions.CommitStateUnknownException;
@@ -718,11 +719,6 @@ public class IcebergMetadata
         TableMetadata metadata = table.operations().current();
         if (metadata.formatVersion() < 3) {
             return;
-        }
-
-        // Reject Iceberg table encryption
-        if (!metadata.encryptionKeys().isEmpty() || snapshot.keyId() != null || metadata.properties().containsKey("encryption.key-id")) {
-            throw new TrinoException(NOT_SUPPORTED, "Iceberg table encryption is not supported");
         }
     }
 
@@ -1429,6 +1425,7 @@ public class IcebergMetadata
                     throw new TrinoException(INVALID_TABLE_PROPERTY, format("The provided location '%s' does not match the existing table location '%s'", providedTableLocation.get(), icebergTable.location()));
                 }
                 validateNotModifyingOldSnapshot(table, icebergTable);
+                validateNotEncryptedForWrite(icebergTable);
                 tableLocation = icebergTable.location();
                 List<PartitionField> existingPartitionFields = getAllPartitionFields(icebergTable);
                 transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties, existingPartitionFields);
@@ -1558,6 +1555,7 @@ public class IcebergMetadata
 
         validateNotModifyingOldSnapshot(table, icebergTable);
         validateTableForTrino(icebergTable, getCurrentSnapshotId(icebergTable));
+        validateNotEncryptedForWrite(icebergTable);
 
         beginTransaction(icebergTable);
 
@@ -2050,6 +2048,7 @@ public class IcebergMetadata
         BaseTable icebergTable = catalog.loadTable(session, table.getSchemaTableName());
 
         validateNotModifyingOldSnapshot(table, icebergTable);
+        validateNotEncryptedForWrite(icebergTable);
 
         verifyTableVersionForExecute(OPTIMIZE, OPTIMIZE_MAX_SUPPORTED_TABLE_VERSION, icebergTable);
 
@@ -2370,6 +2369,7 @@ public class IcebergMetadata
     {
         IcebergAddFilesHandle addFilesHandle = (IcebergAddFilesHandle) executeHandle.procedureHandle();
         Table table = catalog.loadTable(session, executeHandle.schemaTableName());
+        validateNotEncryptedForWrite(table);
         TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), table.io().properties());
         long addedDataFiles = addFiles(
                 session,
@@ -2387,6 +2387,7 @@ public class IcebergMetadata
     {
         IcebergAddFilesFromTableHandle addFilesHandle = (IcebergAddFilesFromTableHandle) executeHandle.procedureHandle();
         Table table = catalog.loadTable(session, executeHandle.schemaTableName());
+        validateNotEncryptedForWrite(table);
         TrinoFileSystem fileSystem = fileSystemFactory.create(session.getIdentity(), table.io().properties());
         long addedDataFiles = addFilesFromTable(
                 session,
@@ -3217,6 +3218,7 @@ public class IcebergMetadata
 
         Table icebergTable = catalog.loadTable(session, table.getSchemaTableName());
         validateNotModifyingOldSnapshot(table, icebergTable);
+        validateNotEncryptedForWrite(icebergTable);
 
         beginTransaction(icebergTable);
 
@@ -3244,6 +3246,13 @@ public class IcebergMetadata
     {
         if (table.getSnapshotId().isPresent() && (table.getSnapshotId().getAsLong() != icebergTable.currentSnapshot().snapshotId())) {
             throw new TrinoException(NOT_SUPPORTED, "Modifying old snapshot is not supported in Iceberg");
+        }
+    }
+
+    private static void validateNotEncryptedForWrite(Table table)
+    {
+        if (!(table.encryption() instanceof PlaintextEncryptionManager)) {
+            throw new TrinoException(NOT_SUPPORTED, "Writing to encrypted Iceberg tables is not supported");
         }
     }
 
@@ -3877,6 +3886,7 @@ public class IcebergMetadata
         checkState(fromSnapshotForRefresh.isEmpty(), "From Snapshot must be empty at the start of MV refresh operation.");
         IcebergTableHandle table = (IcebergTableHandle) tableHandle;
         Table icebergTable = catalog.loadTable(session, table.getSchemaTableName());
+        validateNotEncryptedForWrite(icebergTable);
         beginTransaction(icebergTable);
 
         Optional<String> dependencies = Optional.ofNullable(icebergTable.currentSnapshot())
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
@@ -23,6 +23,7 @@ import io.trino.plugin.hive.orc.OrcReaderConfig;
 import io.trino.plugin.hive.orc.OrcWriterConfig;
 import io.trino.plugin.hive.parquet.ParquetReaderConfig;
 import io.trino.plugin.hive.parquet.ParquetWriterConfig;
+import io.trino.plugin.iceberg.encryption.IcebergEncryptionConfig;
 import io.trino.spi.TrinoException;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.session.PropertyMetadata;
@@ -106,6 +107,7 @@ public final class IcebergSessionProperties
     private static final String QUERY_PARTITION_FILTER_REQUIRED_SCHEMAS = "query_partition_filter_required_schemas";
     private static final String INCREMENTAL_REFRESH_ENABLED = "incremental_refresh_enabled";
     public static final String BUCKET_EXECUTION_ENABLED = "bucket_execution_enabled";
+    private static final String PLAINTEXT_FILES_ALLOWED_FOR_ENCRYPTED_TABLES = "plaintext_files_allowed_for_encrypted_tables";
     private static final String MAX_PARTITIONS_PER_WRITER = "max_partitions_per_writer";
 
     private final List<PropertyMetadata<?>> sessionProperties;
@@ -113,6 +115,7 @@ public final class IcebergSessionProperties
     @Inject
     public IcebergSessionProperties(
             IcebergConfig icebergConfig,
+            IcebergEncryptionConfig encryptionConfig,
             OrcReaderConfig orcReaderConfig,
             OrcWriterConfig orcWriterConfig,
             ParquetReaderConfig parquetReaderConfig,
@@ -384,6 +387,11 @@ public final class IcebergSessionProperties
                         "Enable bucket-aware execution: use physical bucketing information to optimize queries",
                         icebergConfig.isBucketExecutionEnabled(),
                         false))
+                .add(booleanProperty(
+                        PLAINTEXT_FILES_ALLOWED_FOR_ENCRYPTED_TABLES,
+                        "Allow reading unencrypted files in tables with encryption enabled",
+                        encryptionConfig.isPlaintextFilesAllowedForEncryptedTables(),
+                        false))
                 .add(integerProperty(
                         MAX_PARTITIONS_PER_WRITER,
                         "Maximum number of partitions per writer",
@@ -638,5 +646,10 @@ public final class IcebergSessionProperties
     public static int maxPartitionsPerWriter(ConnectorSession session)
     {
         return session.getProperty(MAX_PARTITIONS_PER_WRITER, Integer.class);
+    }
+
+    public static boolean arePlaintextFilesAllowedForEncryptedTables(ConnectorSession session)
+    {
+        return session.getProperty(PLAINTEXT_FILES_ALLOWED_FOR_ENCRYPTED_TABLES, Boolean.class);
     }
 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
@@ -23,6 +23,7 @@ import io.trino.plugin.hive.orc.OrcReaderConfig;
 import io.trino.plugin.hive.orc.OrcWriterConfig;
 import io.trino.plugin.hive.parquet.ParquetReaderConfig;
 import io.trino.plugin.hive.parquet.ParquetWriterConfig;
+import io.trino.plugin.iceberg.encryption.IcebergEncryptionConfig;
 import io.trino.spi.TrinoException;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.session.PropertyMetadata;
@@ -107,6 +108,7 @@ public final class IcebergSessionProperties
     private static final String QUERY_PARTITION_FILTER_REQUIRED_SCHEMAS = "query_partition_filter_required_schemas";
     private static final String INCREMENTAL_REFRESH_ENABLED = "incremental_refresh_enabled";
     public static final String BUCKET_EXECUTION_ENABLED = "bucket_execution_enabled";
+    private static final String PLAINTEXT_FILES_ALLOWED_FOR_ENCRYPTED_TABLES = "plaintext_files_allowed_for_encrypted_tables";
     private static final String MAX_PARTITIONS_PER_WRITER = "max_partitions_per_writer";
 
     private final List<PropertyMetadata<?>> sessionProperties;
@@ -114,6 +116,7 @@ public final class IcebergSessionProperties
     @Inject
     public IcebergSessionProperties(
             IcebergConfig icebergConfig,
+            IcebergEncryptionConfig encryptionConfig,
             OrcReaderConfig orcReaderConfig,
             OrcWriterConfig orcWriterConfig,
             ParquetReaderConfig parquetReaderConfig,
@@ -390,6 +393,11 @@ public final class IcebergSessionProperties
                         "Enable bucket-aware execution: use physical bucketing information to optimize queries",
                         icebergConfig.isBucketExecutionEnabled(),
                         false))
+                .add(booleanProperty(
+                        PLAINTEXT_FILES_ALLOWED_FOR_ENCRYPTED_TABLES,
+                        "Allow reading unencrypted files in tables with encryption enabled",
+                        encryptionConfig.isPlaintextFilesAllowedForEncryptedTables(),
+                        false))
                 .add(integerProperty(
                         MAX_PARTITIONS_PER_WRITER,
                         "Maximum number of partitions per writer",
@@ -649,5 +657,10 @@ public final class IcebergSessionProperties
     public static int maxPartitionsPerWriter(ConnectorSession session)
     {
         return session.getProperty(MAX_PARTITIONS_PER_WRITER, Integer.class);
+    }
+
+    public static boolean arePlaintextFilesAllowedForEncryptedTables(ConnectorSession session)
+    {
+        return session.getProperty(PLAINTEXT_FILES_ALLOWED_FOR_ENCRYPTED_TABLES, Boolean.class);
     }
 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
@@ -23,7 +23,9 @@ import io.trino.spi.block.Block;
 import io.trino.spi.connector.ConnectorSplit;
 import io.trino.spi.predicate.TupleDomain;
 
+import java.util.Arrays;
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.OptionalLong;
 
@@ -49,7 +51,8 @@ public record IcebergSplit(
         @JsonProperty("fileStatisticsDomain") TupleDomain<IcebergColumnHandle> fileStatisticsDomain,
         @JsonProperty("affinityKey") Optional<String> affinityKey,
         @JsonProperty("dataSequenceNumber") long dataSequenceNumber,
-        @JsonProperty("fileFirstRowId") OptionalLong fileFirstRowId)
+        @JsonProperty("fileFirstRowId") OptionalLong fileFirstRowId,
+        @JsonProperty("parquetFileDecryptionData") Optional<ParquetFileDecryptionData> parquetFileDecryptionData)
         implements ConnectorSplit
 {
     private static final int INSTANCE_SIZE = instanceSize(IcebergSplit.class);
@@ -64,6 +67,7 @@ public record IcebergSplit(
         requireNonNull(fileStatisticsDomain, "fileStatisticsDomain is null");
         requireNonNull(affinityKey, "affinityKey is null");
         requireNonNull(fileFirstRowId, "fileFirstRowId is null");
+        requireNonNull(parquetFileDecryptionData, "parquetFileDecryptionData is null");
     }
 
     @Override
@@ -91,7 +95,8 @@ public record IcebergSplit(
                 + fileStatisticsDomain.getRetainedSizeInBytes(IcebergColumnHandle::getRetainedSizeInBytes)
                 + SIZE_OF_LONG // dataSequenceNumber
                 + sizeOf(affinityKey, SizeOf::estimatedSizeOf)
-                + (fileFirstRowId.isPresent() ? SIZE_OF_LONG : 0);
+                + (fileFirstRowId.isPresent() ? SIZE_OF_LONG : 0)
+                + sizeOf(parquetFileDecryptionData, ParquetFileDecryptionData::getRetainedSizeInBytes);
     }
 
     @Override
@@ -108,5 +113,41 @@ public record IcebergSplit(
                     .mapToLong(DeleteFile::recordCount).sum());
         }
         return helper.toString();
+    }
+
+    public record ParquetFileDecryptionData(byte[] fileEncryptionKey, byte[] fileAadPrefix)
+    {
+        private static final int INSTANCE_SIZE = instanceSize(ParquetFileDecryptionData.class);
+
+        public ParquetFileDecryptionData
+        {
+            requireNonNull(fileEncryptionKey, "fileEncryptionKey is null");
+            requireNonNull(fileAadPrefix, "fileAadPrefix is null");
+        }
+
+        @Override
+        public boolean equals(Object o)
+        {
+            return o instanceof ParquetFileDecryptionData other
+                    && Arrays.equals(fileEncryptionKey, other.fileEncryptionKey)
+                    && Arrays.equals(fileAadPrefix, other.fileAadPrefix);
+        }
+
+        @Override
+        public int hashCode()
+        {
+            return Objects.hash(Arrays.hashCode(fileEncryptionKey), Arrays.hashCode(fileAadPrefix));
+        }
+
+        @Override
+        public String toString()
+        {
+            return "[REDACTED]";
+        }
+
+        public long getRetainedSizeInBytes()
+        {
+            return INSTANCE_SIZE + SizeOf.sizeOf(fileEncryptionKey) + SizeOf.sizeOf(fileAadPrefix);
+        }
     }
 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
@@ -33,9 +33,11 @@ import io.trino.filesystem.cache.SplitAffinityProvider;
 import io.trino.plugin.base.metrics.DurationTiming;
 import io.trino.plugin.base.metrics.IntList;
 import io.trino.plugin.base.metrics.LongCount;
+import io.trino.plugin.iceberg.IcebergSplit.ParquetFileDecryptionData;
 import io.trino.plugin.iceberg.delete.DeleteFile;
 import io.trino.plugin.iceberg.util.DataFileWithDeleteFiles;
 import io.trino.spi.SplitWeight;
+import io.trino.spi.TrinoException;
 import io.trino.spi.block.Block;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ConnectorSession;
@@ -63,9 +65,15 @@ import org.apache.iceberg.Scan;
 import org.apache.iceberg.Schema;
 import org.apache.iceberg.StructLike;
 import org.apache.iceberg.Table;
+import org.apache.iceberg.encryption.EncryptedInputFile;
+import org.apache.iceberg.encryption.EncryptionManager;
+import org.apache.iceberg.encryption.NativeEncryptionInputFile;
+import org.apache.iceberg.encryption.NativeEncryptionKeyMetadata;
 import org.apache.iceberg.expressions.Expression;
 import org.apache.iceberg.io.CloseableIterable;
 import org.apache.iceberg.io.CloseableIterator;
+import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.io.InputFile;
 import org.apache.iceberg.metrics.InMemoryMetricsReporter;
 import org.apache.iceberg.metrics.ScanMetricsResult;
 import org.apache.iceberg.metrics.ScanReport;
@@ -118,6 +126,7 @@ import static io.trino.plugin.iceberg.IcebergUtil.primitiveFieldTypes;
 import static io.trino.plugin.iceberg.StructLikeWrapperWithFieldIdToIndex.createStructLikeWrapper;
 import static io.trino.plugin.iceberg.TypeConverter.toIcebergType;
 import static io.trino.plugin.iceberg.TypeConverter.toTrinoType;
+import static io.trino.spi.StandardErrorCode.NOT_SUPPORTED;
 import static io.trino.spi.type.DateTimeEncoding.packDateTimeWithZone;
 import static io.trino.spi.type.TimeZoneKey.UTC_KEY;
 import static io.trino.spi.type.TypeUtils.writeNativeValue;
@@ -127,7 +136,10 @@ import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 import static org.apache.iceberg.FileContent.EQUALITY_DELETES;
 import static org.apache.iceberg.FileContent.POSITION_DELETES;
+import static org.apache.iceberg.FileFormat.PARQUET;
+import static org.apache.iceberg.encryption.EncryptedFiles.encryptedInput;
 import static org.apache.iceberg.types.Conversions.fromByteBuffer;
+import static org.apache.iceberg.util.ByteBuffers.toByteArray;
 
 public class IcebergSplitSource
         implements ConnectorSplitSource
@@ -140,6 +152,8 @@ public class IcebergSplitSource
     private final ConnectorSession session;
     private final IcebergTableHandle tableHandle;
     private final Map<String, String> fileIoProperties;
+    private final FileIO fileIo;
+    private final EncryptionManager encryptionManager;
     private final Scan<?, FileScanTask, CombinedScanTask> tableScan;
     private final OptionalLong maxScannedFileSizeInBytes;
     private final Map<Integer, Type.PrimitiveType> fieldIdToType;
@@ -209,7 +223,9 @@ public class IcebergSplitSource
         this.fileSystemFactory = requireNonNull(fileSystemFactory, "fileSystemFactory is null");
         this.session = requireNonNull(session, "session is null");
         this.tableHandle = requireNonNull(tableHandle, "tableHandle is null");
-        this.fileIoProperties = requireNonNull(icebergTable.io().properties(), "fileIoProperties is null");
+        this.fileIo = requireNonNull(icebergTable.io(), "fileIo is null");
+        this.fileIoProperties = requireNonNull(fileIo.properties(), "fileIoProperties is null");
+        this.encryptionManager = requireNonNull(icebergTable.encryption(), "encryptionManager is null");
         this.tableScan = requireNonNull(tableScan, "tableScan is null");
         this.maxScannedFileSizeInBytes = maxScannedFileSize.isPresent() ? OptionalLong.of(maxScannedFileSize.orElseThrow().toBytes()) : OptionalLong.empty();
         this.fieldIdToType = primitiveFieldTypes(tableScan.schema());
@@ -590,6 +606,16 @@ public class IcebergSplitSource
                 .filter(column -> predicatedColumnIds.contains(column.fieldId()))
                 .map(column -> getColumnHandle(column, typeManager))
                 .collect(toImmutableList());
+        List<DeleteFile> deleteFiles = wholeFileTask.deletes().stream()
+                .peek(file -> verifyDeletionVectorReferencesDataFile(wholeFileTask, file))
+                .map(file -> DeleteFile.fromIceberg(file, parquetFileDecryptionData(
+                        file.format(),
+                        file.location(),
+                        file.fileSizeInBytes(),
+                        file.keyMetadata(),
+                        fileIo,
+                        encryptionManager)))
+                .collect(toImmutableList());
         return new FileScanTaskWithDomain(
                 wholeFileTask,
                 createFileStatisticsDomain(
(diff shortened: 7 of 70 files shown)


sopel39:
cc @osscm

findinpath:
@sopel39  pls add a bit of color to the description of the PR as well as to the commit comment.

The PR comes with approximately 2K LOC

Add  functional notes to the description about what Iceberg parquet encryption is good for
https://iceberg.apache.org/docs/nightly/encryption/

(Release notes would be welcome as well)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1609,6 +1605,7 @@ public ConnectorInsertTableHandle beginInsert(ConnectorSession session, Connecto
 
         validateNotModifyingOldSnapshot(table, icebergTable);
         validateTableForTrino(icebergTable, getCurrentSnapshotId(icebergTable));
+        validateNotEncryptedForWrite(icebergTable);
```
VVKot, line 1558: I think beginOptimize and beginRefreshMaterializedView should also have this because they write data and hence could write unencrypted data?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/EncryptionAwareFileIO.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.iceberg.io.FileIO;
+import org.apache.iceberg.io.InputFile;
+import org.apache.iceberg.io.OutputFile;
+
+import java.util.Map;
+
+/**
+ * Wraps {@link EncryptingFileIO} to handle encrypted manifests and data files,
+ * while delegating {@link #properties()} to the underlying {@link FileIO}
+ * since {@link EncryptingFileIO} does not support it.
```
VVKot: Maybe a TODO saying this could be deleted after Iceberg 1.11 bump, i.e. once we have https://github.com/apache/iceberg/commit/473d46a5c76ce7fc776199034650749572b3b1d5?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
```diff
@@ -251,4 +265,20 @@ public String toString()
         }
         return helper.toString();
     }
+
+    public record ParquetFileDecryptionData(byte[] fileEncryptionKey, byte[] fileAadPrefix)
```
VVKot, line 118: nit: you migth want to override equals/hashCode here because byte array is a reference type, so generate equals/hashCode are not useful.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
VVKot: @sopel39 looking through this more, you probably want to handle a few more:

1. beginCreateTable - at least the `replace` path
2. executeAddFiles
3. executeAddFilesFromTable
4. executeDelete
5. truncateTable

Thoughts?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -592,19 +606,69 @@ private FileScanTaskWithDomain createFileScanTaskWithDomain(FileScanTask wholeFi
...
+
+        if (!(inputFile instanceof NativeEncryptionInputFile nativeEncryptionInputFile)) {
+            return Optional.empty();
+        }
+
+        NativeEncryptionKeyMetadata nativeKeyMetadata = nativeEncryptionInputFile.keyMetadata();
+        ByteBuffer encryptionKey = requireNonNull(nativeKeyMetadata.encryptionKey(), "native encryption key is null");
+        ByteBuffer aadPrefix = requireNonNull(nativeKeyMetadata.aadPrefix(), "native AAD prefix is null");
+        return Optional.of(new ParquetFileDecryptionData(toByteArray(encryptionKey), toByteArray(aadPrefix)));
+    }
```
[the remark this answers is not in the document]
VVKot: Thinking about this more - should we override toString as well? Given that this is raw, unwrapped DEK. You get something like `[B@6d06d69c` by default anyway, but better safe than sorry in case someone decides to log this at a later point?  

```
        @Override
        public String toString()
        {
            return "[REDACTED]";
        }
```

Copilot's concern is valid, but that's what Spark does as well, and IMO it is correct as we're optimizing for number of KMS calls performed.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/util/EncryptedFileTestUtils.java
```diff
@@ -0,0 +1,248 @@
...
+                .withPartition(null)
+                .withKeyMetadata(encryptedOutputFile.keyMetadata())
+                .createWriterFunc(GenericParquetWriter::create)
+                .build()) {
+            for (Record record : records) {
+                writer.write(record);
+            }
+            writer.close();
+            return writer.toDataFile();
+        }
```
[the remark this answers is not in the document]
sopel39: toDataFile()/toDeleteFile() requires the writer to be explicitly closed first to finalize the file. The try-with-resources only closes at block exit, which is too late. Restoring the writer.close() calls.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
sopel39: makes sense

sopel39:
@ebyhr @VVKot addressed comments, PTAL

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
VVKot: I think executeDelete and truncateTable are still not covered? Appreciate those are less of a concern given that there isn't a way to insert new data there, but I believe the intent of the PR is to only enable reads and block all writes.

Covering all of those in tests would be good as well, to prevent this from accidentally breaking during some refactoring

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+        return this;
+    }
+
+    public boolean isAllowPlaintextFiles()
+    {
+        return allowPlaintextFiles;
+    }
+
+    @Config("iceberg.encryption.allow-plaintext-files")
+    @ConfigDescription("Allow reading unencrypted files in tables with encryption enabled")
```
VVKot, line 47: Not sure about this flag - maybe worth rewording it if you want to merge it?

There are two cases:

1. old plain-text files that didn't have encryption metadata
2. new plain-text files manifest for which has encryption metadata (== there was a bug on the write path)

One could think that if this is set to `true`, Trino would throw in both cases, but (1) will succeed and only (2) will throw.

Arguably in case of (2) it is already "too late" because you wrote data without encryption when you should have, but at least this gives you some indication about it.

VVKot (approved):
This looks good to me! I have also tested with Iceberg `main` branch against our internal REST catalog with encryption, and it works there too.

I had one more nit, and for some reason can't seem to un-resolve the comment: https://github.com/trinodb/trino/pull/28389#discussion_r3094876500

raunaqmorarka:
@coderabbitai review

lib/trino-parquet/src/main/java/io/trino/parquet/crypto/FileDecryptionProperties.java
```diff
@@ -22,12 +22,14 @@ public class FileDecryptionProperties
     private final DecryptionKeyRetriever keyRetriever;
     private final Optional<byte[]> aadPrefix;
     private final boolean checkFooterIntegrity;
+    private final boolean plaintextFilesAllowed;
```
findinpath, line 25: Please change the PR title to

Add read support for Iceberg Parquet encryption in the Iceberg connector.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetEncryption.java
```diff
@@ -0,0 +1,373 @@
...
+    public void setUp()
+    {
+        catalog = getTrinoCatalog(getHiveMetastore(getQueryRunner()), getFileSystemFactory(getQueryRunner()), "iceberg");
+        getQueryRunner().execute("CREATE SCHEMA IF NOT EXISTS tpch");
+    }
+
+    @Test
+    public void testReadEncryptedParquetTable()
+            throws Exception
+    {
```
findinpath, line 98: Please consider writing product tests to test the compatibility with Spark.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetEncryption.java
```diff
@@ -0,0 +1,373 @@
...
+                        SortOrder.unsorted(),
+                        Optional.of(catalog.defaultTableLocation(SESSION, schemaTableName)),
+                        ImmutableMap.of(TableProperties.DEFAULT_FILE_FORMAT, "PARQUET"))
+                .commitTransaction();
+    }
+
+    private void setTableEncryptionKey(Table table, SchemaTableName schemaTableName)
+    {
+        table.updateProperties()
+                .set(ENCRYPTION_TABLE_KEY, TEST_KEY_ID)
```
findinpath, line 403: https://iceberg.apache.org/docs/nightly/encryption/

> Currently, encryption is supported in the Hive and REST catalogs for tables with Parquet and Avro data formats.

Is there coverage in the tests to verify that decrypting avro does not currently work?
ORC decrypting should fail right?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/AbstractIcebergTableOperations.java
```diff
@@ -284,6 +301,7 @@ protected void refreshFromMetadataLocation(String newLocation, Function<String,
         currentMetadataLocation = newLocation;
         version = OptionalInt.of(parseVersion(Location.of(newLocation).fileName()));
         shouldRefresh = false;
+        updateEncryptionManager(newMetadata);
```
findinpath, line 320: https://iceberg.apache.org/docs/nightly/encryption/#catalog-security-requirements

>Catalogs must ensure the encryption.key-id property is not modified or removed during table lifetime.


Is this in line with the recommendations from the Iceberg docs?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/TrinoHiveCatalog.java
```diff
@@ -890,7 +890,7 @@ public Metrics getMetrics()
     }
 
     @Override
-    protected void invalidateTableCache(SchemaTableName schemaTableName)
+    public void invalidateTableCache(SchemaTableName schemaTableName)
```
findinpath, line 894: I see it used in `TestIcebergParquetEncryption` 

This could be eventually obtained with unregister/register table . No need to tamper this method for invalidating table's cache.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/jdbc/IcebergJdbcTableOperationsProvider.java
```diff
@@ -32,16 +33,19 @@ public class IcebergJdbcTableOperationsProvider
     private final TrinoFileSystemFactory fileSystemFactory;
     private final ForwardingFileIoFactory fileIoFactory;
     private final IcebergJdbcClient jdbcClient;
+    private final EncryptionManagerFactory encryptionManagerFactory;
 
     @Inject
     public IcebergJdbcTableOperationsProvider(
```
findinpath, line 39: https://iceberg.apache.org/docs/nightly/encryption/

> Currently, encryption is supported in the Hive and REST catalogs for tables with Parquet and Avro data formats.

Trivia question: Do you happen to have any idea why is it not supported for all the catalogs?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/nessie/IcebergNessieTableOperationsProvider.java
```diff
@@ -33,13 +34,15 @@ public class IcebergNessieTableOperationsProvider
     private final TrinoFileSystemFactory fileSystemFactory;
     private final ForwardingFileIoFactory fileIoFactory;
     private final NessieIcebergClient nessieClient;
```
findinpath, line 36: Could you pls document briefly this new feature in `iceberg.md` ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+        KeyManagementClient kmsClient = keyManagementClient.orElseThrow(() -> new TrinoException(
+                ICEBERG_CATALOG_ERROR,
+                "Can't create encryption manager, because key management client is not configured. Set iceberg.encryption.kms-type or iceberg.encryption.kms-impl catalog property."));
+        return EncryptionUtil.createEncryptionManager(tableProperties, kmsClient);
+    }
+
+    private static Optional<KeyManagementClient> createKeyManagementClient(IcebergEncryptionConfig encryptionConfig)
+    {
+        requireNonNull(encryptionConfig, "encryptionConfig is null");
```
findinpath: nit: maybe move it up in the constructor 

```
createKeyManagementClient(requireNonNull(encryptionConfig, "encryptionConfig is null"));
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+        ImmutableMap.Builder<String, String> properties = ImmutableMap.builder();
+        encryptionConfig.getKmsType().ifPresent(value -> properties.put(CatalogProperties.ENCRYPTION_KMS_TYPE, value));
+        encryptionConfig.getKmsImpl().ifPresent(value -> properties.put(CatalogProperties.ENCRYPTION_KMS_IMPL, value));
```
findinpath: Does not feel in line with Spark

https://github.com/apache/iceberg/blob/dde712ec9ed6c9d28183ee4615d50f97b246af5d/core/src/main/java/org/apache/iceberg/encryption/EncryptionUtil.java#L48-L55

Only one of them should be present in Spark.

findinpath: We're delegating to iceberg and the failure will happen, but we can provide a meaningful trino error code instead of an internal error.

findinpath: actually we can control this in `io.trino.plugin.iceberg.encryption.IcebergEncryptionConfig` with an assertion

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+ */
+package io.trino.plugin.iceberg.encryption;
+
+import io.airlift.configuration.Config;
+import io.airlift.configuration.ConfigDescription;
+import jakarta.validation.constraints.NotNull;
+
+import java.util.Optional;
+
+public class IcebergEncryptionConfig
```
findinpath, line 22: Let's document these properties in `iceberg.md`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    private boolean allowPlaintextFiles;
+
+    @NotNull
+    public Optional<String> getKmsType()
+    {
+        return kmsType;
+    }
+
+    @Config("iceberg.encryption.kms-type")
+    public IcebergEncryptionConfig setKmsType(String kmsType)
```
findinpath: Let's use an enum here - similar to what we're doing with `IcebergFileFormat`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
```diff
@@ -22,17 +23,24 @@
...
+import static java.util.Objects.requireNonNull;
 
 public record FilesTableSplit(
         TrinoManifestFile manifestFile,
         String schemaJson,
         String metadataTableJson,
         Map<Integer, String> partitionSpecsByIdJson,
-        Optional<Type> partitionColumnType)
+        Optional<Type> partitionColumnType,
+        Map<String, String> tableProperties)
```
findinpath: In the `tableProperties` we may have security sensitive data.

We recently removed this field in da91faf8

findinpath: for the purpose of encryption we can use a dedicated object to transport relevant data for `EncryptionManagerFactory`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
```diff
@@ -650,4 +658,9 @@ public static int maxPartitionsPerWriter(ConnectorSession session)
     {
         return session.getProperty(MAX_PARTITIONS_PER_WRITER, Integer.class);
     }
+
+    public static boolean isEncryptionAllowPlaintextFiles(ConnectorSession session)
```
findinpath: isEncryptionAllowPlaintextFiles -> not quite english

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSessionProperties.java
```diff
@@ -650,4 +658,9 @@ public static int maxPartitionsPerWriter(ConnectorSession session)
...
+    {
+        return session.getProperty(ENCRYPTION_ALLOW_PLAINTEXT_FILES, Boolean.class);
```
findinpath: ENCRYPTION_ALLOW_PLAINTEXT_FILES -> ALLOW_PLAINTEXT_FILES_ENCRYPTION

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+public class IcebergEncryptionConfig
+{
+    private Optional<String> kmsType = Optional.empty();
+    private Optional<String> kmsImpl = Optional.empty();
+    private boolean allowPlaintextFiles;
+
+    @NotNull
+    public Optional<String> getKmsType()
+    {
+        return kmsType;
```
findinpath, line 30: There's support for AWS, Azure and GCP in iceberg's `EncyrptionUtil`.
It would make sense to have a test against an actual KMS (in the product tests) besides using the `TestingFileMetastoreKeyManagementClient` implementation

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.Optional;
+
+import static io.trino.plugin.iceberg.IcebergErrorCode.ICEBERG_CATALOG_ERROR;
+import static java.util.Objects.requireNonNull;
+import static org.apache.iceberg.TableProperties.ENCRYPTION_TABLE_KEY;
+
+public class DefaultEncryptionManagerFactory
+        implements EncryptionManagerFactory
+{
+    private final Optional<KeyManagementClient> keyManagementClient;
```
ebyhr, line 37: Why does this field need to be optional?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,76 @@
...
+    {
+        String tableKeyId = tableProperties.get(ENCRYPTION_TABLE_KEY);
+        if (tableKeyId == null) {
+            return PlaintextEncryptionManager.instance();
+        }
+
+        KeyManagementClient kmsClient = keyManagementClient.orElseThrow(() -> new TrinoException(
+                ICEBERG_CATALOG_ERROR,
+                "Can't create encryption manager, because key management client is not configured. Set iceberg.encryption.kms-type or iceberg.encryption.kms-impl catalog property."));
+        return EncryptionUtil.createEncryptionManager(tableProperties, kmsClient);
```
ebyhr: `createEncryptionManager(Map<String, String> tableProperties, KeyManagementClient kmsClient)` is deprecated.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,76 @@
...
+
+        ImmutableMap.Builder<String, String> properties = ImmutableMap.builder();
+        encryptionConfig.getKmsType().ifPresent(value -> properties.put(CatalogProperties.ENCRYPTION_KMS_TYPE, value));
+        encryptionConfig.getKmsImpl().ifPresent(value -> properties.put(CatalogProperties.ENCRYPTION_KMS_IMPL, value));
+
+        Map<String, String> catalogProperties = properties.buildOrThrow();
+        if (catalogProperties.isEmpty()) {
+            return Optional.empty();
+        }
+        return Optional.of(EncryptionUtil.createKmsClient(catalogProperties));
```
ebyhr: `catalogProperties` only contains ENCRYPTION_KMS_TYPE / ENCRYPTION_KMS_IMPL. 

Does the subsequent `KeyManagementClient#initialize(catalogProperties)` in Iceberg library really work with real KMS? I suppose we need to pass more properties, e.g. credentials.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    @NotNull
+    public Optional<String> getKmsImpl()
+    {
+        return kmsImpl;
+    }
+
+    @Config("iceberg.encryption.kms-impl")
+    public IcebergEncryptionConfig setKmsImpl(String kmsImpl)
```
ebyhr: Allowing kms-impl (= java class name) is controversial. The property can be used in unexpected way such as untested custom jars. I don't think we should provide this option from the beginning.

sopel39: > Allowing kms-impl (= java class name) is controversial. The property can be used in unexpected way such as untested custom jars. I don't think we should provide this option from the beginning.

what would be the alternative?

> The property can be used in unexpected way such as untested custom jars.

Ordinary user cannot set these, only admin

ebyhr: > what would be the alternative?

Why did you remove kms-type?

sopel39: Iceberg 1.10.1 doesn't support it. It comes with 1.11

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
sopel39: It's optional because KMS may not be configured (no kms-impl set). Tables without encryption still work

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
sopel39: replaced with encryptionKeyId

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
sopel39: removed kms-type entirely, not present in iceberg 1.10

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetEncryption.java
sopel39: Iceberg doesn't support ORC encryption. Added AVRO tests

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/AbstractIcebergTableOperations.java
sopel39: Added protection for chaning encryption.key-id via extra properties

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/TrinoHiveCatalog.java
sopel39: This is test infrastructure only. The invalidateTableCache visibility change keeps tests simple and readable. Using unregister/register would add boilerplate and obscure the test's intent.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
sopel39: Kms type is removed for now

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/jdbc/IcebergJdbcTableOperationsProvider.java
sopel39: I believe this might be related to securely storing actual keys, which might not be trivial

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
sopel39: Let's do this as a follow-up maybe? I'm not sure we have KMS infra really and TBH I'm not sure how to set it up. `TestingFileMetastoreKeyManagementClient` is already used in PT, which mimics real setup

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
sopel39: Well, the other non-deprecated `createEncryptionManager` method  is not yet public. Will be public in 1.11

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
sopel39: Real KMS clients (e.g., AWS) fall back to the default credential chain (environment variables, instance profile) when no credentials are in the properties map same as Trino's S3 filesystem. So initialize() with just `kms-impl` works in environments with ambient credentials.

lib/trino-parquet/src/main/java/io/trino/parquet/crypto/FileDecryptionProperties.java
```diff
@@ -22,12 +22,14 @@ public class FileDecryptionProperties
...
 
-    private FileDecryptionProperties(DecryptionKeyRetriever keyRetriever, Optional<byte[]> aadPrefix, boolean checkFooterIntegrity)
+    private FileDecryptionProperties(DecryptionKeyRetriever keyRetriever, Optional<byte[]> aadPrefix, boolean checkFooterIntegrity, boolean plaintextFilesAllowed)
```
findinpath: nit: split to multiple lines

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -97,6 +98,7 @@ public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<
         validateParquet(MAGIC.equals(magic) || EMAGIC.equals(magic), dataSource.getId(), "Expected magic number: %s or %s got: %s", MAGIC.toStringUtf8(), EMAGIC.toStringUtf8(), magic.toStringUtf8());
 
         boolean encryptedFooterMode = EMAGIC.equals(magic);
+
```
findinpath: probably not intentionally added - remove

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/AbstractIcebergTableOperations.java
```diff
@@ -115,6 +125,7 @@ public void initializeFromMetadata(TableMetadata tableMetadata)
         currentMetadataLocation = tableMetadata.metadataFileLocation();
         shouldRefresh = false;
         version = OptionalInt.of(parseVersion(Location.of(currentMetadataLocation).fileName()));
+        updateEncryptionManager(tableMetadata);
```
findinpath, line 133: we should be "updating" the encryption manager only if the metadata properties are changing. let's track the previous metadata properties to reduce the need to redundantly re-create the encryption manager.

testing/trino-product-tests-launcher/src/main/resources/docker/trino-product-tests/conf/environment/singlenode-spark-iceberg/iceberg.properties
```diff
@@ -2,4 +2,5 @@ connector.name=iceberg
 hive.metastore.uri=thrift://hadoop-master:9083
 iceberg.register-table-procedure.enabled=true
 iceberg.allowed-extra-properties=*
+iceberg.encryption.kms-impl=io.trino.plugin.iceberg.TestingFileMetastoreKeyManagementClient
```
findinpath: i don't find this test relevant.
let's either use localstack or actual aws infra for testing this feature.

findinpath: here is a short integration test based on localstack (generated code)

```
import org.junit.jupiter.api.AfterAll;
import org.junit.jupiter.api.BeforeAll;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.DisplayName;
import org.junit.jupiter.api.Test;
import org.testcontainers.containers.localstack.LocalStackContainer;
import org.testcontainers.utility.DockerImageName;
import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
import software.amazon.awssdk.auth.credentials.StaticCredentialsProvider;
import software.amazon.awssdk.regions.Region;
import software.amazon.awssdk.services.kms.KmsClient;
import software.amazon.awssdk.services.kms.model.CreateKeyResponse;
import software.amazon.awssdk.services.s3.S3Client;
import software.amazon.awssdk.services.s3.model.HeadObjectResponse;
import software.amazon.awssdk.services.s3.model.PutObjectRequest;
import software.amazon.awssdk.services.s3.model.ServerSideEncryption;

import static org.junit.jupiter.api.Assertions.assertEquals;

class S3KmsIntegrationTest {

    private static final DockerImageName localstackImage = DockerImageName.parse("localstack/localstack:3.0");

    // Spin up LocalStack with S3 and KMS services
    static LocalStackContainer localstack = new LocalStackContainer(localstackImage)
            .withServices(LocalStackContainer.Service.S3, LocalStackContainer.Service.KMS);

    private S3Client s3;
    private KmsClient kms;

    @BeforeAll
    static void startContainer() {
        localstack.start();
    }

    @AfterAll
    static void stopContainer() {
        localstack.stop();
    }

    @BeforeEach
    void setupClients() {
        // Configure clients to point to the LocalStack proxy port
        s3 = S3Client.builder()
                .endpointOverride(localstack.getEndpoint())
                .credentialsProvider(StaticCredentialsProvider.create(
                        AwsBasicCredentials.create(localstack.getAccessKey(), localstack.getSecretKey())))
                .region(Region.of(localstack.getRegion()))
                .build();

        kms = KmsClient.builder()
                .endpointOverride(localstack.getEndpoint())
                .credentialsProvider(StaticCredentialsProvider.create(
                        AwsBasicCredentials.create(localstack.getAccessKey(), localstack.getSecretKey())))
                .region(Region.of(localstack.getRegion()))
                .build();
    }

    @Test
    @DisplayName("Should upload file to S3 using KMS encryption in LocalStack")
    void testS3KmsUpload() {
        String bucketName = "secure-test-bucket";
        String key = "test-file.txt";

        // 1. Create a KMS Key
        CreateKeyResponse keyResponse = kms.createKey(r -> r.description("Test Key"));
        String kmsKeyId = keyResponse.keyMetadata().keyId();

        // 2. Create an S3 Bucket
        s3.createBucket(b -> b.bucket(bucketName));

        // 3. Put Object with SSE-KMS
        s3.putObject(PutObjectRequest.builder()
                        .bucket(bucketName)
                        .key(key)
                        .serverSideEncryption(ServerSideEncryption.AWS_KMS)
                        .ssekmsKeyId(kmsKeyId)
                        .build(),
                software.amazon.awssdk.core.sync.RequestBody.fromString("Encrypted content"));

        // 4. Verify the metadata
        HeadObjectResponse head = s3.headObject(h -> h.bucket(bucketName).key(key));

        assertEquals(ServerSideEncryption.AWS_KMS.toString(), head.serverSideEncryptionAsString());
        // LocalStack returns the full ARN; we check if it contains our KeyId
        assert(head.ssekmsKeyId().contains(kmsKeyId));
    }
}

```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetEncryption.java
```diff
@@ -0,0 +1,407 @@
...
+    private static final TestingFileMetastoreKeyManagementClient KMS_CLIENT = new TestingFileMetastoreKeyManagementClient();
+
+    private TrinoCatalog catalog;
+
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        return IcebergQueryRunner.builder()
+                .addIcebergProperty("iceberg.encryption.kms-impl", TestingFileMetastoreKeyManagementClient.class.getName())
```
findinpath: consider using localstack with s3 and kms

testing/trino-product-tests-launcher/src/main/resources/docker/trino-product-tests/conf/environment/singlenode-spark-iceberg/iceberg.properties
sopel39: > i don't find this test relevant.
let's either use localstack or actual aws infra for testing this feature.

The key idea of this test is not KMS, but rather spark <-> Trino compatibility when writing, reading files

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
ebyhr: I don't think we should expose Java class name. `iceberg.encryption.kms-type` is sufficient because we can internally map the value to `encryption.kms-impl`.

sopel39: kms-type is not supported in Iceberg 1.10.1. It's added in 1.11, which still supports `kms-impl`. Hardcoding Iceberg's internal KMS class names in Trino as an enum mapping would be fragile. If we bump to 1.11, we can expose kms-type natively.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetEncryption.java
sopel39: @findinpath I think it's overkill.
1. We already have good UT-level tests for parquet encryption. It doesn't make sense to make them heavier for the sake of KMS testing. It also doesn't add value to what is fundamentally tested.
2. From Trino perspective, KMS is transparent, because Trino just calls `EncryptionUtil.createEncryptionManager`. It's internal Iceberg logic to instantiate KMS.
3. Valuable test for Spark <-> Trino integration is already added as PT. Therefore I think we have pretty good coverage.

sopel39:
@ebyhr @findinpath added KMS loading test. In 1.10 azure kms is not yet available

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -133,12 +134,20 @@ public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<
...
+                if (!fileDecryptionProperties.get().isPlaintextFilesAllowed()) {
+                    throw new ParquetCryptoException("Applying decryptor on plaintext file: %s", dataSource.getId());
+                }
+            }
+            else {
+                // Footer is not encrypted, but some columns might be encrypted
+                decryptionContext = Optional.of(new FileDecryptionContext(dataSource.getId(), fileDecryptionProperties.get(), fileMetaData.getEncryption_algorithm(), Optional.ofNullable(fileMetaData.getFooter_signing_key_metadata())));
+                if (fileDecryptionProperties.get().isCheckFooterIntegrity()) {
+                    // verify footer integrity
+                    verifyFooterIntegrity(dataSource, metadataStream, decryptionContext.get(), metadataLength);
```
ebyhr: This comment can be removed. It doesn't give additional info.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -222,10 +231,10 @@ private static org.apache.parquet.column.statistics.Statistics<?> tryReadOldUtf8
...
         return org.apache.parquet.column.statistics.Statistics
                 .getBuilderForReading(columnStatistics.type())
-                       .withMin(min)
-                       .withMax(max)
-                       .withNumNulls(!columnStatistics.isNumNullsSet() && statistics.isSetNull_count() ? statistics.getNull_count() : columnStatistics.getNumNulls())
-                       .build();
+                .withMin(min)
+                .withMax(max)
+                .withNumNulls(!columnStatistics.isNumNullsSet() && statistics.isSetNull_count() ? statistics.getNull_count() : columnStatistics.getNumNulls())
+                .build();
```
ebyhr, line 242: This change can be extacted into a commit. It is unrelated to "Add support for table encryption in Iceberg".

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
```diff
@@ -69,6 +75,7 @@ public IcebergSplit(
             @JsonProperty("splitWeight") SplitWeight splitWeight,
             @JsonProperty("fileStatisticsDomain") TupleDomain<IcebergColumnHandle> fileStatisticsDomain,
             @JsonProperty("dataSequenceNumber") long dataSequenceNumber,
+            @JsonProperty("parquetFileDecryptionData") Optional<ParquetFileDecryptionData> parquetFileDecryptionData,
```
ebyhr: Can we avoid adding this object to the split? We recently faced a security incident, and that is why we introduced `ConnectorTableCredentials` interface.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
ebyhr: I still disagree with the `kms-impl` config property. Trino users shouldn't need to worry about the Iceberg library version.

You mentioned that Iceberg 1.10.x doesn't support Azure KMS. That suggests we should only provide an enum with AWS and GCP.

If we expose the `kms-impl` config property, we won't be able to remove it later since we won't know how it's being used. I don't think we should open that door from the beginning.

findinpath: > `kms-type` is not supported in Iceberg 1.10.1. It's added in 1.11

You can temporarily add handling in the trino code to rely internally on the iceberg's impl mechanism until 1.11
This doesn't seem a constraint to me.

findinpath:
> @ebyhr @findinpath added KMS loading test. In 1.10 azure kms is not yet available

@sopel39 Karol, my request was about making the PR easier to grasp (and eventually to maintain)  by testing it end to end against  real-world KMS implementations. With that, I was not implying the `io.trino.plugin.iceberg.encryption.TestKmsClientInstantiation` unit test.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
sopel39: `parquetFileDecryptionData` is per file and not on a table level.

@ebyhr how is split POJO a security thread? I'm not sure what incident you mean, but I'm betting it's more around bad/unsecured infra rather than POJO java structures. Non-admins users should never have direct access to worker nodes.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/IcebergEncryptionConfig.java
sopel39: I've added `kms-type` and `kms-impl`. Mutually exclusive.

findinpath: i see other examples of dealing with class names in the `apache/iceberg` library code

`AwsProperties.CLIENT_FACTORY` https://github.com/apache/iceberg/blob/57409faedb184eb3b410a9cbbab52a4e5c0b6f0a/aws/src/main/java/org/apache/iceberg/aws/AwsProperties.java#L112-L117
AwsClientProperties.CLIENT_CREDENTIALS_PROVIDER https://github.com/apache/iceberg/blob/57409faedb184eb3b410a9cbbab52a4e5c0b6f0a/aws/src/main/java/org/apache/iceberg/aws/AwsClientProperties.java#L45-L60

however, whether intentionally or not, they haven't been added to `trinodb/trino` (yet).

@sopel39  what about sticking with `kms-type`  in `trinodb/trino` which covers the most wide array of use cases and leave `kms-impl` for Trino shops that actually need it ?
I think that this change would potentially move the contention point around `kms-impl`.

sopel39: Iceberg exposes `kms-impl` natively for other engines to use. Spark encrypted tables can (and in practice are) created with customized `kms-impl` value. Hence, if we want to preserve Spark <-> Trino compatibility (like in PT), we should expose `kms-impl` config property

sopel39:
@ebyhr @findinpath removed `kms-impl`, added localstack based PT

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -259,6 +259,18 @@ implementation is used:
   - Enable bucket-aware execution. This allows the engine to use physical
     bucketing information to optimize queries by reducing data exchanges.
   - `true`
+* - `iceberg.encryption.kms-type`
+  - Key Management Service type for
+    [Iceberg table encryption](https://iceberg.apache.org/docs/nightly/encryption/).
+    Possible values are `AWS` and `GCP`. Required to read encrypted tables.
+    Writing to encrypted tables is not supported.
```
findinpath, line 266: adjust the PR description to match the exposed configuration properties.

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -259,6 +259,18 @@ implementation is used:
   - Enable bucket-aware execution. This allows the engine to use physical
     bucketing information to optimize queries by reducing data exchanges.
   - `true`
+* - `iceberg.encryption.kms-type`
```
findinpath, line 262: change the commit title to 

> Add read support for encrypted tables in Iceberg

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/KmsType.java
```diff
@@ -11,16 +11,23 @@
...
-import java.util.Optional;
-
-sealed interface VendedCredentials
-        permits S3VendedCredentials, GcsVendedCredentials, AzureVendedCredentials
+public enum KmsType
 {
-    Map<String, String> toExtraCredentials();
+    AWS("org.apache.iceberg.aws.AwsKeyManagementClient"),
+    GCP("org.apache.iceberg.gcp.GcpKeyManagementClient"),
+    /**/;
```
findinpath, line 21: what speaks against adding support for 

```
AZURE("org.apache.iceberg.azure.keymanagement.AzureKeyManagementClient")
```
?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,72 @@
...
+
+public class DefaultEncryptionManagerFactory
+        implements EncryptionManagerFactory
+{
+    private final Optional<KeyManagementClient> keyManagementClient;
+
+    @Inject
+    public DefaultEncryptionManagerFactory(IcebergEncryptionConfig encryptionConfig)
+    {
+        this.keyManagementClient = createKeyManagementClient(requireNonNull(encryptionConfig, "encryptionConfig is null"));
```
findinpath: pls pass to `createKeyManagementClient` only the class name - no need for that method to know about the config class.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -0,0 +1,72 @@
...
+        return EncryptionUtil.createEncryptionManager(tableProperties, kmsClient);
+    }
+
+    private static Optional<KeyManagementClient> createKeyManagementClient(IcebergEncryptionConfig encryptionConfig)
+    {
+        if (encryptionConfig.getKmsType().isEmpty()) {
+            return Optional.empty();
+        }
+
+        Map<String, String> catalogProperties = ImmutableMap.of(
```
findinpath: not sure whether this is correct.

The argument for `EncryptionUtil` is `catalogProperties`

For a REST catalog it would look probably like this https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogPropertiesProvider.java

I see also in io.trino.plugin.iceberg.catalog.jdbc.TrinoJdbcCatalogFactory catalogProperties being built.
Maybe we should have a `CatalogPropertiesProvider` bound for each catalog type (hive/rest/glue/...)

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -2438,6 +2442,44 @@ public void testAlterTableExecuteProceduresOnEmptyTable()
...
+        }
+
+        onSpark().executeQuery("DROP TABLE IF EXISTS " + sparkTableName);
+
+        onSpark().executeQuery(format(
+                "CREATE TABLE %s (id INT, name STRING) USING ICEBERG " +
+                        "TBLPROPERTIES ('write.format.default'='PARQUET', 'encryption.key-id'='%s')",
+                sparkTableName, keyId));
+        onSpark().executeQuery(format("INSERT INTO %s VALUES (1, 'alice'), (2, 'bob'), (3, 'charlie')", sparkTableName));
+
```
findinpath, line 2489: let's verify as well that spark's read content is the same as trino's

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -2438,6 +2442,44 @@ public void testAlterTableExecuteProceduresOnEmptyTable()
...
+        assertThat(onTrino().executeQuery("SELECT * FROM " + trinoTableName + " ORDER BY id"))
+                .containsOnly(row(1, "alice"), row(2, "bob"), row(3, "charlie"));
+
+        assertQueryFailure(() -> onTrino().executeQuery("INSERT INTO " + trinoTableName + " VALUES (4, 'dave')"))
+                .hasMessageContaining("Writing to encrypted Iceberg tables is not supported");
+
```
findinpath, line 2501: can we check as well $files metadata table content ?

testing/trino-product-tests-launcher/src/main/resources/docker/trino-product-tests/conf/environment/singlenode-spark-iceberg/spark-defaults.conf
```diff
@@ -10,6 +10,7 @@ spark.sql.catalog.iceberg_test.type=hive
 spark.sql.catalog.iceberg_test.uri=thrift://hadoop-master:9083
 ; disabling caching allows us to run spark queries interchangeably with trino's
 spark.sql.catalog.iceberg_test.cache-enabled=false
+spark.sql.catalog.iceberg_test.encryption.kms-type=aws
```
findinpath, line 13: pretty impressive that it works out of the box with localstack without any extra settings.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/KmsType.java
sopel39: there is no AZURE KMS in 1.10...

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -2438,6 +2442,44 @@ public void testAlterTableExecuteProceduresOnEmptyTable()
...
+            keyId = kmsClient.createKey(builder -> builder.description("trino-iceberg-pt-" + randomNameSuffix())).keyMetadata().keyId();
+        }
+
+        onSpark().executeQuery("DROP TABLE IF EXISTS " + sparkTableName);
+
+        onSpark().executeQuery(format(
+                "CREATE TABLE %s (id INT, name STRING) USING ICEBERG " +
+                        "TBLPROPERTIES ('write.format.default'='PARQUET', 'encryption.key-id'='%s')",
+                sparkTableName, keyId));
+        onSpark().executeQuery(format("INSERT INTO %s VALUES (1, 'alice'), (2, 'bob'), (3, 'charlie')", sparkTableName));
```
findinpath, line 2488:
```
[root@hadoop-master /]# hdfs dfs -copyToLocal /user/hive/warehouse/t1/data/00002-2-7efcb6e4-7dd0-45c8-9882-c371bfb1992a-0-00001.parquet .
[root@hadoop-master /]# ls
00002-2-7efcb6e4-7dd0-45c8-9882-c371bfb1992a-0-00001.parquet  anaconda-post.log  bin  dev  docker  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
[root@hadoop-master /]# exit
exit
(base) ➜  ~ docker cp ptl-hadoop-master:/00002-2-7efcb6e4-7dd0-45c8-9882-c371bfb1992a-0-00001.parquet  .
Successfully copied 654B (transferred 2.56kB) to /Users/marius/.
(base) ➜  ~ parquet schema 00002-2-7efcb6e4-7dd0-45c8-9882-c371bfb1992a-0-00001.parquet 
{
  "type" : "record",
  "name" : "table",
  "fields" : [ {
    "name" : "id",
    "type" : "int"
  }, {
    "name" : "name",
    "type" : "string"
  } ]
}
(base) ➜  ~ parquet cat 00002-2-7efcb6e4-7dd0-45c8-9882-c371bfb1992a-0-00001.parquet 
{"id": 3, "name": "charlie"}
```

content doesn't seem encrypted to me.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
sopel39: I don't think we should apply this in this PR. A few reasons:

1. The KMS client doesn't actually need Trino's catalog properties. EncryptionUtil.createKmsClient ends up calling KeyManagementClient.initialize(catalogProperties). For AwsKeyManagementClient that's AwsClientFactories.from(map) + new AwsProperties(map) — both expect
  Iceberg-AWS keys (client.region, client.credentials-provider, s3.access-key-id, …). Trino doesn't populate those: our S3/GCS access goes through TrinoFileSystem, not through Iceberg's AwsClientFactories. So forwarding REST/JDBC catalog properties wouldn't make AWS KMS
  magically pick up Trino's S3 credentials — it would either be a no-op (the keys aren't there) or it would leak unrelated catalog config (REST OAuth tokens, JDBC connection strings) into KMS init.
2. Default credentials work for the standard deployment. With an empty map, AwsClientFactories.from(...) falls back to the AWS SDK default credentials chain (env/instance/IRSA/profile). That's exactly how a Trino worker on EC2/EKS would talk to AWS KMS today — no
  per-catalog plumbing required.
3. CatalogPropertiesProvider bound per catalog type is indirection without a concrete need. We have two supported KMS types (AWS, GCP), both with sensible defaults. If/when we need to thread custom KMS config (custom region for KMS, role-arn, etc.) we should add a
  dedicated, properly-scoped namespace like iceberg.encryption.kms.* and forward only those to initialize. That's a smaller, explicit surface than reusing the whole catalog property bag.
4. YAGNI for read-only support. This PR only enables reads of encrypted tables. Writing isn't supported yet. Let's keep the encryption wiring minimal until there's a real use case that the default credential chain doesn't cover; we can extend then without breaking anyone.

Happy to file a follow-up issue to track adding iceberg.encryption.kms.* pass-through properties if/when someone needs them.  We could also introduce dedicated per KMS Config properties loaded depending on chosen KMS. Let's do that as a potential follow-up

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/KmsType.java
findinpath: follow-up https://github.com/trinodb/trino/pull/26640#issuecomment-4351749396

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
sopel39: Iceberg 1.10.1 does not support reading encrypted Avro manifest list files (apache/iceberg#7770 is not in the 1.10.x branch), and 1.10.1's Hive Catalog also lacks the wiring to engage StandardEncryptionManager on writes (apache/iceberg#13066, apache/iceberg#15272 not in 1.10.x). End-to-end encrypted Iceberg tables test with Hive catalog requires Iceberg 1.11.0+ on both sides (Spark and Trino)

While I've managed to make Spark write encrypted tables with Hive catalog, these cannot be read by Trino using 1.10 (encrypted Avro read fails). I wonder if it makes sense to support PME without Hive catalog or we should just wait for 1.11

findinpath: Stack your PR on a branch that feeds from  https://github.com/trinodb/trino/pull/26640

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/encryption/DefaultEncryptionManagerFactory.java
```diff
@@ -59,14 +61,8 @@ public EncryptionManager create(Map<String, String> tableProperties)
...
+    private static KeyManagementClient createKeyManagementClient(String kmsClientClassName)
     {
-        if (encryptionConfig.getKmsType().isEmpty()) {
-            return Optional.empty();
-        }
-
-        Map<String, String> catalogProperties = ImmutableMap.of(
-                ENCRYPTION_KMS_IMPL, encryptionConfig.getKmsType().orElseThrow().getKmsClientClassName());
-        return Optional.of(EncryptionUtil.createKmsClient(catalogProperties));
+        return EncryptionUtil.createKmsClient(ImmutableMap.of(ENCRYPTION_KMS_IMPL, kmsClientClassName));
```
findinpath, line 68: I think we can switch to using `CatalogProperties.ENCRYPTION_KMS_TYPE` with `apache/iceberg` `1.11.0`

testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/EnvSinglenodeSparkIceberg.java
```diff
@@ -51,8 +51,10 @@ public class EnvSinglenodeSparkIceberg
     private static final String AWS_REGION = "us-east-1";
     private static final String AWS_ACCESS_KEY_ID = "test";
     private static final String AWS_SECRET_ACCESS_KEY = "test";
-    // Must be kept in sync with dep.iceberg.version in the root pom.xml
-    private static final String ICEBERG_VERSION = "1.10.1";
+    // Spark engages Iceberg encryption only with the 1.11.0+ catalog wiring
+    // (iceberg/apache#13066, #15272). Replace with 1.11.0 once released.
+    private static final String ICEBERG_SNAPSHOT_VERSION = "1.11.0-SNAPSHOT";
+    private static final String ICEBERG_SNAPSHOT_REPO = "https://repository.apache.org/content/repositories/snapshots/";
```
findinpath: is this still needed?

findinpath:
https://github.com/trinodb/trino/actions/runs/25163707867/job/73765598014?pr=28389
PT seems to fail

```
presto-master       | 2026-04-30T18:02:42.199+0545	INFO	dispatcher-query-11	io.trino.event.QueryMonitor	TIMELINE: Query 20260430_121742_02128_7nb5j :: FAILED (GENERIC_INTERNAL_ERROR) :: elapsed 35ms :: planning 6ms :: waiting 20ms :: scheduling 29ms :: running 0ms :: finishing 29ms :: begin 2026-04-30T18:02:42.161+05:45 :: end 2026-04-30T18:02:42.196+05:45
tests               | 2026-04-30 18:02:42 INFO: not retrying; @Flaky annotation not present
tests               | 2026-04-30 18:02:42 INFO: FAILURE     /    io.trino.tests.product.iceberg.TestIcebergSparkCompatibility.testTrinoReadingSparkEncryptedData (Groups: profile_specific_tests, iceberg) took 3.4 seconds
tests               | 2026-04-30 18:02:42 SEVERE: Failure cause:
tests               | io.trino.tempto.query.QueryExecutionException: java.sql.SQLException: Query failed (#20260430_121742_02128_7nb5j): io.trino.spi.TrinoException: Error processing metadata for table default.test_trino_reading_spark_encrypted_bv4wzrs330
tests               | 	at io.trino.tempto.query.JdbcQueryExecutor.execute(JdbcQueryExecutor.java:119)
tests               | 	at io.trino.tempto.query.JdbcQueryExecutor.executeQuery(JdbcQueryExecutor.java:84)
tests               | 	at io.trino.tests.product.utils.QueryExecutors$1.lambda$executeQuery$0(QueryExecutors.java:56)
tests               | 	at dev.failsafe.Functions.lambda$toCtxSupplier$11(Functions.java:243)
tests               | 	at dev.failsafe.Functions.lambda$get$0(Functions.java:46)
tests               | 	at dev.failsafe.internal.RetryPolicyExecutor.lambda$apply$0(RetryPolicyExecutor.java:74)
tests               | 	at dev.failsafe.SyncExecutionImpl.executeSync(SyncExecutionImpl.java:187)
tests               | 	at dev.failsafe.FailsafeExecutor.call(FailsafeExecutor.java:376)
tests               | 	at dev.failsafe.FailsafeExecutor.get(FailsafeExecutor.java:112)
tests               | 	at io.trino.tests.product.utils.QueryExecutors$1.executeQuery(QueryExecutors.java:56)
tests               | 	at io.trino.tests.product.iceberg.TestIcebergSparkCompatibility.testTrinoReadingSparkEncryptedData(TestIcebergSparkCompatibility.java:2465)
tests               | 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
tests               | 	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
tests               | 	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:104)
tests               | 	at org.testng.internal.Invoker.invokeMethod(Invoker.java:645)
tests               | 	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:851)
tests               | 	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1177)
tests               | 	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
tests               | 	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
tests               | 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
tests               | 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
tests               | 	at java.base/java.lang.Thread.run(Thread.java:1474)
tests               | Caused by: java.sql.SQLException: Query failed (#20260430_121742_02128_7nb5j): io.trino.spi.TrinoException: Error processing metadata for table default.test_trino_reading_spark_encrypted_bv4wzrs330
tests               | 	at io.trino.jdbc.ResultUtils.resultsException(ResultUtils.java:33)
tests               | 	at io.trino.jdbc.AsyncResultIterator.lambda$new$1(AsyncResultIterator.java:94)
tests               | 	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545)
tests               | 	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328)
tests               | 	... 3 more
tests               | 	Suppressed: java.lang.Exception: Query: SELECT * FROM iceberg.default.test_trino_reading_spark_encrypted_bv4wzrs330 ORDER BY id
tests               | 		at io.trino.tempto.query.JdbcQueryExecutor.executeQueryNoParams(JdbcQueryExecutor.java:136)
tests               | 		at io.trino.tempto.query.JdbcQueryExecutor.execute(JdbcQueryExecutor.java:112)
tests               | 		at io.trino.tempto.query.JdbcQueryExecutor.executeQuery(JdbcQueryExecutor.java:84)
tests               | 		at io.trino.tests.product.utils.QueryExecutors$1.lambda$executeQuery$0(QueryExecutors.java:56)
tests               | 		at dev.failsafe.Functions.lambda$toCtxSupplier$11(Functions.java:243)
tests               | 		at dev.failsafe.Functions.lambda$get$0(Functions.java:46)
tests               | 		at dev.failsafe.internal.RetryPolicyExecutor.lambda$apply$0(RetryPolicyExecutor.java:74)
tests               | 		at dev.failsafe.SyncExecutionImpl.executeSync(SyncExecutionImpl.java:187)
tests               | 		at dev.failsafe.FailsafeExecutor.call(FailsafeExecutor.java:376)
tests               | 		at dev.failsafe.FailsafeExecutor.get(FailsafeExecutor.java:112)
tests               | 		at io.trino.tests.product.utils.QueryExecutors$1.executeQuery(QueryExecutors.java:56)
tests               | 		at io.trino.tests.product.iceberg.TestIcebergSparkCompatibility.testTrinoReadingSparkEncryptedData(TestIcebergSparkCompatibility.java:2465)
tests               | 		at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
tests               | 		at java.base/java.lang.reflect.Method.invoke(Method.java:565)
tests               | 		at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:104)
tests               | 		at org.testng.internal.Invoker.invokeMethod(Invoker.java:645)
tests               | 		at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:851)
tests               | 		at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1177)
tests               | 		at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
tests               | 		at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
tests               | 		... 3 more
tests               | Caused by: java.util.concurrent.CompletionException: io.trino.spi.TrinoException: Error processing metadata for table default.test_trino_reading_spark_encrypted_bv4wzrs330
tests               | 	at java.base/java.util.concurrent.CompletableFuture.wrapInCompletionException(CompletableFuture.java:323)
tests               | 	at java.base/java.util.concurrent.CompletableFuture.encodeThrowable(CompletableFuture.java:359)
tests               | 	at java.base/java.util.concurrent.CompletableFuture.completeThrowable(CompletableFuture.java:364)
tests               | 	at java.base/java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:1015)
tests               | 	at java.base/java.util.concurrent.CompletableFuture$UniExceptionally.tryFire(CompletableFuture.java:995)
tests               | 	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:531)
tests               | 	at java.base/java.util.concurrent.CompletableFuture.completeExceptionally(CompletableFuture.java:2221)
tests               | 	at io.airlift.concurrent.MoreFutures$2.onFailure(MoreFutures.java:495)
tests               | 	at com.google.common.util.concurrent.Futures$CallbackListener.run(Futures.java:1111)
tests               | 	at com.google.common.util.concurrent.DirectExecutor.execute(DirectExecutor.java:30)
tests               | 	at com.google.common.util.concurrent.AbstractFuture.executeListener(AbstractFuture.java:1021)
tests               | 	at com.google.common.util.concurrent.AbstractFuture.complete(AbstractFuture.java:784)
tests               | 	at com.google.common.util.concurrent.AbstractFuture.setException(AbstractFuture.java:533)
tests               | 	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.afterRanInterruptiblyFailure(TrustedListenableFutureTask.java:138)
tests               | 	at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:89)
tests               | 	at com.google.common.util.concurrent.TrustedListenableFutureTask.run(TrustedListenableFutureTask.java:80)
tests               | 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
tests               | 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
tests               | 	at java.base/java.lang.Thread.run(Thread.java:1474)
tests               | Caused by: io.trino.spi.TrinoException: Error processing metadata for table default.test_trino_reading_spark_encrypted_bv4wzrs330
tests               | 	at io.trino.plugin.iceberg.IcebergExceptions.translateMetadataException(IcebergExceptions.java:54)
tests               | 	at io.trino.plugin.iceberg.IcebergSplitSource.lambda$getNextBatch$2(IcebergSplitSource.java:286)
tests               | 	at java.base/java.util.concurrent.CompletableFuture.uniExceptionally(CompletableFuture.java:1011)
tests               | 	... 15 more
tests               | Caused by: org.apache.iceberg.exceptions.RuntimeIOException: Failed to open file: hdfs://hadoop-master:9000/user/hive/warehouse/test_trino_reading_spark_encrypted_bv4wzrs330/metadata/snap-6482809109431514840-1-ce641b4c-f9e2-449b-b337-261f89f09b20.avro
tests               | 	at org.apache.iceberg.avro.AvroIterable.newFileReader(AvroIterable.java:104)
tests               | 	at org.apache.iceberg.avro.AvroIterable.iterator(AvroIterable.java:77)
tests               | 	at org.apache.iceberg.avro.AvroIterable.iterator(AvroIterable.java:37)
tests               | 	at org.apache.iceberg.relocated.com.google.common.collect.Iterables.addAll(Iterables.java:332)
tests               | 	at org.apache.iceberg.relocated.com.google.common.collect.Lists.newLinkedList(Lists.java:261)
tests               | 	at org.apache.iceberg.ManifestLists.read(ManifestLists.java:42)
tests               | 	at org.apache.iceberg.BaseSnapshot.cacheManifests(BaseSnapshot.java:185)
tests               | 	at org.apache.iceberg.BaseSnapshot.dataManifests(BaseSnapshot.java:211)
tests               | 	at org.apache.iceberg.DataTableScan.doPlanFiles(DataTableScan.java:68)
tests               | 	at org.apache.iceberg.SnapshotScan.planFiles(SnapshotScan.java:139)
tests               | 	at io.trino.plugin.iceberg.IcebergSplitSource.getNextBatchInternal(IcebergSplitSource.java:320)
tests               | 	at io.trino.plugin.iceberg.IcebergSplitSource.lambda$getNextBatch$1(IcebergSplitSource.java:281)
tests               | 	at com.google.common.util.concurrent.TrustedListenableFutureTask$TrustedFutureInterruptibleTask.runInterruptibly(TrustedListenableFutureTask.java:128)
tests               | 	at com.google.common.util.concurrent.InterruptibleTask.run(InterruptibleTask.java:74)
tests               | 	... 4 more
tests               | Caused by: org.apache.avro.InvalidAvroMagicException: Not an Avro data file
tests               | 	at org.apache.avro.file.DataFileReader.openReader(DataFileReader.java:79)
tests               | 	at org.apache.iceberg.avro.AvroIterable.newFileReader(AvroIterable.java:102)
tests               | 	... 17 more
```

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
findinpath: @sopel39 pls add a test to verify that the content of the written parquet files through spark is actually encrypted.

Gemini suggestion

```
try (ParquetFileReader reader = ParquetFileReader.open(conf, hadoopPath)) {
    boolean isColumnEncrypted = reader.getFileMetaData()
            .getBlocks().get(0).getColumns().get(0)
            .getCryptoMetaData() != null;
    assertThat(isColumnEncrypted).isTrue();
}
```

it would have been slightly easier if we would rely on AWS stack instead of HDFS for testing

```
// Fetch file metadata from your local S3 mock container
HeadObjectResponse headResponse = s3Client.headObject(HeadObjectRequest.builder()
        .bucket(bucket)
        .key(key)
        .build());

// Assert that encryption was applied on the storage layer
assertThat(headResponse.serverSideEncryptionAsString()).isEqualTo("aws:kms");
assertThat(headResponse.ssekmsKeyId()).contains("your-kms-key-id");
```

findinpath:
branch has conflicts - pls resolve them

ebyhr:
- Superseded by #29882

PR #28390, opened 2026-02-20
Determine whether try/try_cast and lambda expressions might fail

wendigo:
## Description

This is a follow up to the changes made in the https://github.com/trinodb/trino/commit/bfd6a4ccf7bcbaf8ec6b61e4501915c5ea67713e. Specifically it now analyzes the lambda expression to check whether it might fail. If this is the case, `all_match` and `any_match` might fail too, despite being marked as `neverFails` which means that the function in itself never fails if all its inputs never fail.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
+++ b/core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
@@ -81,6 +81,8 @@ public final class IrExpressions
             // These expressions need to verify their operands
             case Array e -> e.elements().stream().anyMatch(element -> mayFail(plannerContext, element));
             case Between e -> mayFail(plannerContext, e.value()) || mayFail(plannerContext, e.min()) || mayFail(plannerContext, e.max());
+            // Function call is considered never failing if the function itself is marked as never failing and all input arguments
+            // are never failing too.
             case Call e -> mayFail(e.function()) || e.arguments().stream().anyMatch(argument -> mayFail(plannerContext, argument));
             case Case e -> e.whenClauses().stream().anyMatch(clause -> mayFail(plannerContext, clause.getOperand()) || mayFail(plannerContext, clause.getResult())) ||
                     mayFail(plannerContext, e.defaultValue());


core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -80,6 +80,7 @@ public static boolean mayFail(PlannerContext plannerContext, Expression expressi
             case Comparison e -> mayFail(plannerContext, e.left()) || mayFail(plannerContext, e.right());
             case In e -> mayFail(plannerContext, e.value()) || e.valueList().stream().anyMatch(argument -> mayFail(plannerContext, argument));
             case IsNull e -> mayFail(plannerContext, e.value());
+            case Lambda e -> mayFail(plannerContext, e.body());
```
martint, line 94: This is incorrect. A lambda being able to fail doesn't imply the function call that invokes it fails. Lambdas cannot be called directly.

wendigo: Shouldn't lambda invocations failures be handled consistently across different functions? Why function X with lambda Y can fail but other Z with same lambda Y will not because it's not rethrowing an exception.

martint: Example:

A hypothetical `try_or_else(() -> 1 / x > 0, false)` will never fail even though the invocation of the `() -> 1 / x > 0` may fail for some x.

wendigo: I'm asking in the context of these two array functions. If the lambda can fail, should these functions catch and ignore exceptions coming from lambda?

martint: Ah, no, array processing functions such as `all_match`, `any_match`, `array_transform`, etc. should fail if the invocation of the lambda expression fails during evaluation. Note, however, that `any_match` doesn't necessarily evaluate all values in the array, as it will bail out as soon as a match is found.

wendigo: but we can determine whether lambda expression might fail using the same IR processing as we do for other things. Lambda is just an expression that involves operators, logical operators, function calls etc.

So if we have a `function x(a, b: lambda)` that is marked as `neverFails`

function call `x(c, d -> d > 0)` won't fail since `d > 0` won't fail

but function call `x(c, d -> 1 / d > 0)` might fail since `1 / d > 0` might fail.

Basically `neverFails` on a scalar function means `if my input arguments never fail, I won't fail too` which is exactly what we verify in the `IrExpressions`:

```
case Call e -> mayFail(e.function()) || e.arguments().stream().anyMatch(argument -> mayFail(plannerContext, argument));
```

wendigo: I've added second commit that adds explicit meaning to try and try_cast functions (these functions never fail regardless of whether their input arguments might fail)

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -31,6 +35,9 @@
 
 public final class IrExpressions
 {
+    private static final CatalogSchemaFunctionName TRY_FUNCTION_NAME = builtinFunctionName(TryFunction.NAME);
```
martint, line 38: This class should not hard-code knowledge about any function. The initial implementation had that as a placeholder until that attribute was added to the function definitions.

wendigo: These are two types of functions: 

Ones that never fail when arguments might fail

Ones that never fail when arguments never fail too

We can't express that with boolean neverFails today

We should have 3 options: NEVER_FAILS_ON_FAILING_ARGS (this is try, try_cast), NEVER_FAILS_ON_NON_FAILING_ARGS (like, strpos etc), MIGHT_FAIL (rest)

martint: > We should have 3 options: NEVER_FAILS_ON_FAILING_ARGS (this is try, try_cast), NEVER_FAILS_ON_NON_FAILING_ARGS (like, strpos etc), MIGHT_FAIL (rest)

The question `IrExpression.mayFail()` is trying to answer is: "can this expression, when evaluated, fail?". To do so, it looks at every subtree and asks: "While evaluating the expression, can the evaluation of this subtree fail?". 

Hence the check for `Call`: `mayFail(e.function()) || e.arguments().stream().anyMatch(argument -> mayFail(plannerContext, argument))` -- if the evaluation any of the arguments before calling the function fails, then the whole call will fails. If the function may inherently fail, then the whole call may fail.

This is capturing the following scenarios:
* `my_function(1/0)` fails because the argument will fail *before* the function is invoked
* `my_function(x)` may fail because of internal logic in `my_function` that depends on the actual value of `x`

Some clarifications are in order:
* Arguments are *always* evaluated before the function is invoked. 
* The evaluation of an argument fails or succeeds before the function is called. 
* Arguments are just "values" of different types (boolean, varchar, function, etc)

Consequently, talking about failing or non-failing arguments doesn't make sense. The failure happens when calling a function, which, in turn, might call a function value that gets passed in. 

A classifications such as `NEVER_FAILS_ON_FAILING_ARGS`, `NEVER_FAILS_ON_NON_FAILING_ARGS`, `MIGHT_FAIL` mixes overlapping concerns and introduces more confusion to the semantics. By definition, all arguments are non-failing (they are just values). If, as a thought experiment, by non-failing we meant "a value of type `function` that when evaluated may fail", then `NEVER_FAILS_ON_FAILING_ARGS` would be an invalid classifier for a first-order function (i.e., a function that doesn't take other functions as arguments). It would also be insufficient. What if the function takes two functions as arguments, but will never fail regardless of the function passed as the second argument failing when invoked?

PR #28410, opened 2026-02-21
Fix function declarations that might fail due to lambda argument

wendigo:
Verify whether this is the case and fail such functions

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayAllMatchFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/ArrayAllMatchFunction.java
@@ -31,7 +31,7 @@ import static io.trino.spi.function.OperatorType.READ_VALUE;
 import static java.lang.Boolean.FALSE;
 
 @Description("Returns true if all elements of the array match the given predicate")
-@ScalarFunction(value = "all_match", neverFails = true)
+@ScalarFunction("all_match")
 public final class ArrayAllMatchFunction
 {
     private ArrayAllMatchFunction() {}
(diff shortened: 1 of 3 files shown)


core/trino-spi/src/main/java/io/trino/spi/function/FunctionMetadata.java
```diff
@@ -70,7 +70,9 @@ private FunctionMetadata(
         if (functionNullability.getArgumentNullable().size() != signature.getArgumentTypes().size()) {
             throw new IllegalArgumentException("signature and functionNullability must have same argument count");
         }
-
+        if (neverFails && acceptsLambdaArgument(signature)) {
```
martint: Not always. Functions such as “try” takes a lambda but never fails.

wendigo: If this is the case it would have throw an exception. It seems that FunctionMetadata isn't created for try and try_cast

martint: For those specific functions, not yet. For historical reasons, they go through another path. However, it’s a valid use case for a function to swallow errors when invoking a lambda provided via an argument.

wendigo: So there is no way to validate function

martint: That’s right. You’d need to crack open the function and do flow analysis - but if it were that easy, we’d derive this automatically instead of asking the author to tag the function.

This is no different from the “deterministic” flag. There’s no easy way to derive or to verify that the tag is correct.

PR #28413, opened 2026-02-23
Create airlift group for dependabot

ebyhr:
## Description

This means:
- When io.airlift:airbase has an update → Dependabot checks if io.airlift:bom also has an update
- If both have updates → Creates one PR updating both
- If only one has an update → Creates a PR for just that one

## Release notes

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,11 @@ updates:
     schedule:
       interval: "daily"
     open-pull-requests-limit: 10
+    groups:
+      airlift:
+        patterns:
+          - "io.airlift:airbase"
+          - "io.airlift:bom"
     cooldown:
       # Apply 7 day cooldown to avoid updating dependencies right away. This reduces the opportunity window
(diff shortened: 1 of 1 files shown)


wendigo (approved):

.github/dependabot.yml
```diff
@@ -13,16 +13,18 @@ updates:
...
     cooldown:
       # Apply 7 day cooldown to avoid updating dependencies right away. This reduces the opportunity window
       # when supply chain is compromised. This doesn't apply to the dependencies that we own and release.
       default-days: 7
       exclude:
         - io.airlift:*
         - io.trino:*
-    ignore:
-      # Airbase and Airlift are best updated together. Update Airbase only when updating Airlift.
-      - dependency-name: "io.airlift:airbase"
```
findepi, line 25: I think it still is true.

This PR solves the problem when airbase and airlift are both ready to update -- and that's great solution.
But what if airlift is not ready to update _yet_?

PR #28450, opened 2026-02-24
Make BaseElasticsearchConnectorTest extensible with regards to jmxBaseName

ksobolew:
## Description

Initializing a private field in `createQueryRunner` is fraught with peril in case someone ever needs to override it. An overridden method would be called, field initialization will be skipped and tests will fail because the field is null. Initializing it in the constructor avoids that, but we also need to make this field `protected` so that the override can use it to properly initialize the coordinator properties.

- none

## Release notes

```markdown
## Section
```


ksobolew:
Ping

PR #28458, opened 2026-02-25
Fix parametric types data type in system.jdbc.types

### LINKED ISSUE #28451: Wrong values for DATA_TYPE in system.jdbc.types table
The following query:

```sql
SELECT TYPE_NAME, DATA_TYPE from system.jdbc.types where type_name = 'char';
```
returns
 > char	2000

It should return 
>  char    1

The same issue applies to other types, like varchar. 

That causes JDBC method `DatabaseMetaData.getTypeInfo()` to return wrong values for `DATA_TYPE` and causes for instance the SQL `CHAR` type to be mapped to `JDBC_OBJECT` instead of `CHAR`.

wendigo:
## Description

- Fixes https://github.com/trinodb/trino/issues/28451

## Release notes

```markdown
## Core
* Fix parametric data types in system.jdbc.types table ({issue}`issuenumber`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
+++ b/core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
@@ -39,10 +39,18 @@ import io.trino.spi.predicate.Domain;
 import io.trino.spi.predicate.NullableValue;
 import io.trino.spi.predicate.TupleDomain;
 import io.trino.spi.type.ArrayType;
+import io.trino.spi.type.BigintType;
+import io.trino.spi.type.BooleanType;
 import io.trino.spi.type.CharType;
+import io.trino.spi.type.DateType;
 import io.trino.spi.type.DecimalType;
+import io.trino.spi.type.DoubleType;
+import io.trino.spi.type.IntegerType;
+import io.trino.spi.type.NumberType;
 import io.trino.spi.type.ParametricType;
 import io.trino.spi.type.QuantileDigestParametricType;
+import io.trino.spi.type.RealType;
+import io.trino.spi.type.SmallintType;
 import io.trino.spi.type.TimeParametricType;
 import io.trino.spi.type.TimeType;
 import io.trino.spi.type.TimeWithTimeZoneParametricType;
@@ -51,7 +59,9 @@ import io.trino.spi.type.TimestampParametricType;
 import io.trino.spi.type.TimestampType;
 import io.trino.spi.type.TimestampWithTimeZoneParametricType;
 import io.trino.spi.type.TimestampWithTimeZoneType;
+import io.trino.spi.type.TinyintType;
 import io.trino.spi.type.Type;
+import io.trino.spi.type.VarbinaryType;
 import io.trino.spi.type.VarcharType;
 import io.trino.type.ArrayParametricType;
 import io.trino.type.CharParametricType;
@@ -59,6 +69,7 @@ import io.trino.type.DecimalParametricType;
 import io.trino.type.FunctionParametricType;
 import io.trino.type.MapParametricType;
 import io.trino.type.RowParametricType;
+import io.trino.type.UnknownType;
 import io.trino.type.VarcharParametricType;
 
 import java.sql.DatabaseMetaData;
@@ -86,7 +97,6 @@ import static io.trino.metadata.MetadataListing.listTableColumns;
 import static io.trino.metadata.MetadataListing.listTables;
 import static io.trino.metadata.MetadataUtil.TableMetadataBuilder.tableMetadataBuilder;
 import static io.trino.spi.type.BigintType.BIGINT;
-import static io.trino.spi.type.BooleanType.BOOLEAN;
 import static io.trino.spi.type.DateType.DATE;
 import static io.trino.spi.type.DoubleType.DOUBLE;
 import static io.trino.spi.type.IntegerType.INTEGER;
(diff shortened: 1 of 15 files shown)


core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
```diff
@@ -414,6 +424,22 @@ static int jdbcDataType(Type type)
         return Types.JAVA_OBJECT;
     }
 
+    static int jdbcDataType(ParametricType type)
+    {
+        return switch (type) {
+            case VarcharParametricType _ -> Types.VARCHAR;
+            case CharParametricType _ -> Types.CHAR;
+            case RowParametricType _ -> Types.STRUCT;
```
findepi: JAVA_OBJECT is used here: 
https://github.com/trinodb/trino/blob/b9654727ff502aea4c3065c17c63125cd238a4c2/client/trino-jdbc/src/main/java/io/trino/jdbc/ColumnInfo.java#L271-L272

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
```diff
@@ -414,6 +424,22 @@ static int jdbcDataType(Type type)
...
+            case ArrayParametricType _ -> Types.ARRAY;
+            case TimestampWithTimeZoneParametricType _ -> Types.TIMESTAMP_WITH_TIMEZONE;
+            case TimeWithTimeZoneParametricType _ -> Types.TIME_WITH_TIMEZONE;
+            case DecimalParametricType _ -> Types.DECIMAL;
+            case TimeParametricType _ -> Types.TIME;
+            case TimestampParametricType _ -> Types.TIMESTAMP;
```
findepi: can you please put date/times together in order: time, timetz, timestamp, timestamptz?

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
```diff
@@ -414,6 +424,22 @@ static int jdbcDataType(Type type)
...
+            default -> Types.JAVA_OBJECT;
```
findepi: Can we go without `default` so that this code gets updated whenever things change?

core/trino-main/src/main/java/io/trino/connector/system/jdbc/TypesJdbcTable.java
```diff
@@ -119,14 +118,14 @@ private static void addTypeRow(Builder builder, ParametricType type)
         String typeName = type.getName();
         builder.addRow(
                 typeName,
-                typeName.equalsIgnoreCase("array") ? Types.ARRAY : Types.JAVA_OBJECT,
```
findepi, line 122: 😆

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
findepi: Can we maybe have that logic implemented once?

core/trino-spi/src/main/java/io/trino/spi/type/ParametricType.java
```diff
@@ -20,4 +20,12 @@ public interface ParametricType
     String getName();
 
     Type createType(TypeManager typeManager, List<TypeParameter> parameters);
+
+    /**
+     * Returns true only if all created types are comparable.
+     */
+    default boolean areAllTypesComparable()
```
findepi: comparable in `io.trino.spi.type.Type#isComparable` sense and not eg `java.lang.Comparable`.
doesn't hurt to disambiguate.

findepi (commented):
"Fix parametric types data type in system.jdbc.types"

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
wendigo: How you'd achieve that?

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
wendigo: sure

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
wendigo: I can throw. Type system is open here (you can provide parametric type via plugin)

core/trino-spi/src/main/java/io/trino/spi/type/ParametricType.java
wendigo: Clarified

wendigo:
@findepi ptal

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
```diff
@@ -360,58 +383,45 @@ private static void addColumnRows(Builder builder, String catalog, SchemaTableNa
...
+            case CharParametricType _ -> Types.CHAR;
+            case RowParametricType _, MapParametricType _ -> Types.JAVA_OBJECT;
+            case ArrayParametricType _ -> Types.ARRAY;
+            case DecimalParametricType _ -> Types.DECIMAL;
+            case TimeParametricType _ -> Types.TIME;
+            case TimeWithTimeZoneParametricType _ -> Types.TIME_WITH_TIMEZONE;
+            case TimestampParametricType _ -> Types.TIMESTAMP;
+            case TimestampWithTimeZoneParametricType _ -> Types.TIMESTAMP_WITH_TIMEZONE;
+            case QuantileDigestParametricType _, FunctionParametricType _ -> Types.JAVA_OBJECT;
+            default -> throw new IllegalArgumentException("Unmapped parametric type " + type);
```
electrum: This means plugins can’t work.

wendigo: It never worked

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
wendigo: Do we want to couple jdbc with spi or spi with jdbc? Either doesn't make sense to me

wendigo:
@findepi PTAL

core/trino-main/src/main/java/io/trino/connector/system/jdbc/ColumnJdbcTable.java
findepi: neither code considered is SPI.
it's trino-main and trino-jdbc. Both seem to depend on trino-client, right?

wendigo: Is this a hard requirement for this PR to move forward? Duplication applies to master as well and this is a bug fix

PR #25751, opened 2025-05-08
Add CREATE/DROP/FAST FORWARD/SHOW BRANCHES syntax

ebyhr:
## Description

Extracted from #25152. This PR adds the following syntax. Creating a new branch with `FROM other-branch` will be handled as follow-up. 

```g4
    | CREATE (OR REPLACE)? BRANCH (IF NOT EXISTS)? identifier
        (WITH properties)? IN TABLE qualifiedName                      #createBranch
    | DROP BRANCH (IF EXISTS)? identifier
        IN TABLE qualifiedName                                         #dropBranch
    | ALTER BRANCH source=identifier (FROM | IN) TABLE qualifiedName
        FAST FORWARD TO target=identifier                              #fastForwardBranch
    | SHOW BRANCHES IN TABLE qualifiedName                             #showBranchs
```

https://docs.google.com/document/d/1jEF4IkWu-2Gzk5ii2Nb0exuEnAUeo98UbiM3i0xtgWQ/edit?usp=sharing

## Release notes

```markdown
## General
* Add support for creating, dropping, fast-forwarding and listing branches. ({issue}`25751`)
```

PR #24556, opened 2024-12-21
Add support for DML with properties in engine + Add support for branching in Iceberg

ebyhr:
## Description

The 1st commit allows properties after the target table name in DML:
```g4
    | INSERT INTO qualifiedName ('@' branch=string)?
       columnAliases? rootQuery                                        #insertInto
    | DELETE FROM qualifiedName ('@' branch=string)?
         (WHERE booleanExpression)?                                    #delete
    | UPDATE qualifiedName ('@' branch=string)?
        SET updateAssignment (',' updateAssignment)*
        (WHERE where=booleanExpression)?                               #update
    | MERGE INTO qualifiedName (AS? identifier)? ('@' branch=string)?
        USING relation ON expression mergeCase+                        #merge
```

Avoided session property approach because users may forget the current branch and modify unexpected branch. 

The 2nd commit introduces branching support and adds the following procedures in Iceberg connector:
* ALTER TABLE table EXECUTE create_branch(name)
  * creating a new branch
* ALTER TABLE table EXECUTE drop_branch(name)
  * removing a branch by name
* ALTER TABLE table EXECUTE fast_forward(from, to) 
  * performs a fast-forward of `from` up to the `to` snapshot if `from` is an ancestor of `to`.

TestIcebergBranching uses that new syntax and procedures.

Relates to #12844

References
* https://iceberg.apache.org/docs/latest/branching/
* https://docs.dremio.com/current/reference/sql/commands/nessie/

## Release notes

```markdown
# SPI
* TBD. ({issue}`issuenumber`)

# Iceberg
* Add support for `create_branch`, `drop_branch` and `fast_forward` table procedures. ({issue}`issuenumber`)
```

PR #28464, opened 2026-02-25
Add branch management to Iceberg connector

### LINKED ISSUE #12844: Support branching of Iceberg tables
```
trino:default> insert into "test1@3692406374410249224" values (10);
INSERT: 1 row

Query 20220614_205840_00020_m6z2x, FINISHED, 1 node
Splits: 19 total, 19 done (100.00%)
1.16 [0 rows, 0B] [0 rows/s, 0B/s]

trino:default> insert into test1 FOR VERSION AS OF 3692406374410249224 values (10);
Query 20220614_205904_00021_m6z2x failed: line 1:19: mismatched input 'FOR'. Expecting: '(', <query>
insert into test1 FOR VERSION AS OF 3692406374410249224 values (10)
```

By using the now deprecated syntax, we are able to do insertions into a versioned Iceberg table.
However, when using the syntax `AS OF` this functionality is not anymore supported.

This functionality will likely require the grammar `SqlBase.g4` to be extended.

SebastienN15:
Fixes https://github.com/trinodb/trino/issues/12844
## Description

Implement the `ConnectorMetadata` SPI branch methods introduced in #25751, #26300, and #26136 to expose Iceberg's native branching capabilities through the new SQL syntax.

**DDL operations:**
- `CREATE BRANCH [IF NOT EXISTS] b IN TABLE t [FROM source]`
- `CREATE OR REPLACE BRANCH b IN TABLE t`
- `DROP BRANCH [IF EXISTS] b IN TABLE t`
- `ALTER BRANCH b IN TABLE t FAST FORWARD TO target`
- `SHOW BRANCHES IN TABLE t`

**DML with branches:**
- `INSERT INTO t@branch`
- `DELETE FROM t@branch`
- `UPDATE t@branch`
- `MERGE INTO t@branch`

The implementation adds an `Optional<String> branch` field to `IcebergTableHandle` and `IcebergWritableTableHandle`, which is populated when the engine resolves a branch reference via `getTableHandle()`. Write operations call Iceberg's `.toBranch()` API to target the correct branch, and `validateNotModifyingOldSnapshot` is adjusted to allow writes to branch heads that differ from main.

Branch properties (`WITH` clause) are not implemented in this PR.

- SPI syntax PRs: #25751, #26300, #26136
- Prior Iceberg branch PR (closed, used procedures): #24556
- Iceberg natively supports branching via `manageSnapshots().createBranch()`, `removeBranch()`, and `fastForwardBranch()`

## Release notes

```markdown
## Iceberg connector
* Add support for branch management operations: `CREATE BRANCH`, `DROP BRANCH`, `ALTER BRANCH ... FAST FORWARD`, `SHOW BRANCHES`, and DML targeting branches with `@branch` syntax. ({issue}`issuenumber`)
```


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2344,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+        icebergTable.manageSnapshots().createBranch(branch, snapshotId).commit();
+    }
+
+    @Override
+    public void dropBranch(ConnectorSession session, ConnectorTableHandle tableHandle, String branch)
+    {
+        IcebergTableHandle table = checkValidTableHandle(tableHandle);
+
+        if (branch.equals(SnapshotRef.MAIN_BRANCH)) {
+            throw new TrinoException(NOT_SUPPORTED, "Cannot drop the main branch");
```
ebyhr: NOT_SUPPORTED should be used when we plan to support it in the future. We should use a different error code in this case.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9500,6 +9500,279 @@ protected Session withoutSmallFileThreshold(Session session)
                 .build();
     }
 
+    @Test
```
ebyhr, line 9969: Please move these tests to BaseConnectorTest.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9500,6 +9500,279 @@ protected Session withoutSmallFileThreshold(Session session)
...
+    public void testDeleteFromBranch()
+    {
+        String tableName = "test_delete_branch_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " WITH (format_version = 2) AS SELECT * FROM (VALUES (1, 'a'), (2, 'b'), (3, 'c')) AS t(id, name)", 3);
+
+        assertUpdate("CREATE BRANCH test_branch IN TABLE " + tableName);
+
+        // Delete from branch
+        assertUpdate("DELETE FROM " + tableName + "@test_branch WHERE id = 2", 1);
```
ebyhr: This is row level delete. Please add a test for metadata delete such as deleting all rows.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
SebastienN15: Changed to `INVALID_ARGUMENTS`

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,280 @@ protected String createSchemaSql(String schemaName)
         return "CREATE SCHEMA " + schemaName;
     }
 
+    @Test
+    public void testCreateAndShowBranches()
+    {
+        if (!hasBehavior(SUPPORTS_BRANCH)) {
+            return;
```
ebyhr, line 7987: Please check the failure message instead of existing silently. You can refer to other tests, e.g. `testAddColumn()`

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,280 @@ protected String createSchemaSql(String schemaName)
...
+        }
+        String tableName = "test_create_show_branches_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " (id INTEGER, name VARCHAR)");
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a')", 1);
+
+        assertThat(query("SHOW BRANCHES IN TABLE " + tableName))
+                .skippingTypesCheck()
+                .matches("VALUES VARCHAR 'main'");
+
```
ebyhr, line 7999: The default branch name depends on the connector. The name should be extendable.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,280 @@ protected String createSchemaSql(String schemaName)
...
+            return;
+        }
+        String tableName = "test_create_branch_from_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " (id INTEGER, name VARCHAR)");
+        assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'a')", 1);
+
+        assertUpdate("CREATE BRANCH branch_a IN TABLE " + tableName);
+        assertUpdate("CREATE BRANCH branch_b IN TABLE " + tableName + " FROM branch_a");
+
+        assertQuery(
```
ebyhr: Please use `assertThat(query(...))` instead.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,280 @@ protected String createSchemaSql(String schemaName)
...
+    }
+
+    @Test
+    public void testMergeIntoBranch()
+    {
+        if (!hasBehavior(SUPPORTS_BRANCH)) {
+            return;
+        }
+        String tableName = "test_merge_branch_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " WITH (format_version = 2) AS SELECT * FROM (VALUES (1, 'a'), (2, 'b')) AS t(id, name)", 2);
```
ebyhr: > format_version

Please avoid using a connector specific table property in this test class.

alexjo2144:
I haven't been following the changes recently but really excited to see this finally in. 
Is there a syntax for doing things like compacting a branch? 

Probably something like:
```java
ALTER TABLE foo@my_branch EXECUTE optimize;
```

SebastienN15:
@alexjo2144 it's not the purpose of this PR. I think you would need to implement [a new procedure](https://trino.io/docs/current/connector/iceberg.html#procedures) to achieve this, probably similar to already existing [Iceberg spark sql procedures](https://iceberg.apache.org/docs/latest/spark-procedures/)

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
SebastienN15: I wrapped it in `getDefaultBranchName` so that future connectors can extend it

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
SebastienN15: The default SPI listBranches() returns an empty list rather than throwing, so SHOW BRANCHES succeeds on all connectors. We can’t use assertQueryFails with it. createBranch() does throw NOT_SUPPORTED by default, but it requires a valid table handle, meaning we’d first need to create a table, but SUPPORTS_CREATE_TABLE may be false for some connectors.
I think the code is correct here, but if you prefer I can change the SPI default for listBranches() to throw instead of returning empty.

SebastienN15:
Thank you @vlad-lyutenko @ebyhr @findinpath for taking a look at the PR. I have two questions : 
 - Do you know how much time it usually takes for CLA reviews? I send the email 3 weeks ago but got no updates
 - There's still a failing test, but it seems unrelated to my change. Have you ever observed some flakiness on this one? Any hint on how to fix it ?

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,285 @@ protected String createSchemaSql(String schemaName)
...
+    }
+
+    @Test
+    public void testCreateBranchFromBranch()
+    {
+        if (!hasBehavior(SUPPORTS_BRANCH)) {
+            return;
+        }
+        String tableName = "test_create_branch_from_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " (id INTEGER, name VARCHAR)");
```
findinpath, line 8026: ideally the `master` , `branch_a` and `branch_b` should have different contents.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1643,7 +1671,14 @@ else if (!computedStatistics.isEmpty()) {
...
         transaction = null;
 
-        Map<String, String> summary = icebergTable.currentSnapshot().summary();
+        Snapshot resultSnapshot = table.branch()
+                .map(branch -> {
+                    SnapshotRef ref = icebergTable.refs().get(branch);
+                    verify(ref != null, "Branch ref not found after commit: %s", branch);
+                    return icebergTable.snapshot(ref.snapshotId());
+                })
+                .orElseGet(icebergTable::currentSnapshot);
```
findinpath: You may want to extract this method to a utility method - i see it is duplicated

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+                .filter(entry -> entry.getValue().isBranch())
+                .map(Map.Entry::getKey)
+                .collect(toImmutableList());
+    }
+
+    @Override
+    public boolean branchExists(ConnectorSession session, SchemaTableName tableName, String branch)
+    {
+        BaseTable table = catalog.loadTable(session, tableName);
+        SnapshotRef ref = table.refs().get(branch);
```
findinpath:
```suggestion
        SnapshotRef ref = catalog.loadTable(session, tableName).refs().get(branch);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+    {
+        IcebergTableHandle table = checkValidTableHandle(tableHandle);
+        BaseTable icebergTable = catalog.loadTable(session, table.getSchemaTableName());
+
+        if (saveMode == SaveMode.REPLACE) {
+            SnapshotRef existingRef = icebergTable.refs().get(branch);
+            if (existingRef != null && existingRef.isBranch()) {
+                icebergTable.manageSnapshots().removeBranch(branch).commit();
+            }
+        }
```
findinpath, line 2528: is there ability in the SQL syntax to replace branches?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+        else if (icebergTable.currentSnapshot() != null) {
+            snapshotId = icebergTable.currentSnapshot().snapshotId();
+        }
+        else {
+            // Table has no snapshots yet
+            icebergTable.manageSnapshots().createBranch(branch).commit();
+            return;
+        }
+
+        icebergTable.manageSnapshots().createBranch(branch, snapshotId).commit();
```
findinpath: refactor the code to have two branches only

with a source branch specified
```
manageSnapshots.createBranch(branch, ref.snapshotId());
```

without
```
manageSnapshots.createBranch(branch);
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+    @Override
+    public void dropBranch(ConnectorSession session, ConnectorTableHandle tableHandle, String branch)
+    {
+        IcebergTableHandle table = checkValidTableHandle(tableHandle);
+
+        if (branch.equals(SnapshotRef.MAIN_BRANCH)) {
+            throw new TrinoException(INVALID_ARGUMENTS, "Cannot drop the main branch");
+        }
+
+        BaseTable icebergTable = catalog.loadTable(session, table.getSchemaTableName());
```
findinpath, line 2559: add coverage for the situation of removing a branch two times.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+        icebergTable.manageSnapshots().removeBranch(branch).commit();
+    }
```
findinpath, line 2561: also add coverage for namesake but different case operations

`test-branch` vs `TEST-BRANCH`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+    @Override
+    public void fastForwardBranch(
+            ConnectorSession session,
+            ConnectorTableHandle tableHandle,
+            String sourceBranch,
+            String targetBranch)
+    {
+        IcebergTableHandle table = checkValidTableHandle(tableHandle);
+        BaseTable icebergTable = catalog.loadTable(session, table.getSchemaTableName());
+        icebergTable.manageSnapshots().fastForwardBranch(sourceBranch, targetBranch).commit();
```
findinpath: check whether  the forwarding is done to a branch that is not an ancestor of source branch

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2317,6 +2352,88 @@ private void executeRollbackToSnapshot(ConnectorSession session, IcebergTableExe
...
+
+    @Override
+    public void fastForwardBranch(
+            ConnectorSession session,
+            ConnectorTableHandle tableHandle,
+            String sourceBranch,
+            String targetBranch)
+    {
+        IcebergTableHandle table = checkValidTableHandle(tableHandle);
+        BaseTable icebergTable = catalog.loadTable(session, table.getSchemaTableName());
```
findinpath, line 2571: while code is correct, aim for more user-friendly exceptions in case that the operation fails due to illegal arguments.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3420,7 +3537,12 @@ private static void verifyTableVersionForUpdate(IcebergTableHandle table)
 
     private static void validateNotModifyingOldSnapshot(IcebergTableHandle table, Table icebergTable)
     {
-        if (table.getSnapshotId().isPresent() && (table.getSnapshotId().getAsLong() != icebergTable.currentSnapshot().snapshotId())) {
+        // When writing to a branch, the snapshot in the handle is the branch head, not main's head
```
findinpath: not sure whether the comment is needed.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3665,9 +3788,17 @@ public OptionalLong executeDelete(ConnectorSession session, ConnectorTableHandle
...
                 .scanManifestsWith(icebergScanExecutor);
+        handle.getBranch().ifPresent(deleteFiles::toBranch);
         commitUpdate(deleteFiles, session, "delete");
 
-        Map<String, String> summary = icebergTable.currentSnapshot().summary();
+        Snapshot snapshot = handle.getBranch()
+                .map(branch -> {
+                    SnapshotRef ref = icebergTable.refs().get(branch);
+                    verify(ref != null, "Branch ref not found after commit: %s", branch);
+                    return icebergTable.snapshot(ref.snapshotId());
```
findinpath: duplicated code

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3665,9 +3788,17 @@ public OptionalLong executeDelete(ConnectorSession session, ConnectorTableHandle
...
                 .deleteFromRowFilter(toIcebergExpression(handle.getEnforcedPredicate()))
                 .scanManifestsWith(icebergScanExecutor);
+        handle.getBranch().ifPresent(deleteFiles::toBranch);
         commitUpdate(deleteFiles, session, "delete");
 
-        Map<String, String> summary = icebergTable.currentSnapshot().summary();
+        Snapshot snapshot = handle.getBranch()
+                .map(branch -> {
+                    SnapshotRef ref = icebergTable.refs().get(branch);
+                    verify(ref != null, "Branch ref not found after commit: %s", branch);
```
findinpath: if (ref == null) {
            throw new TrinoException(BRANCH_NOT_FOUND, "Branch '%s' does not exist".formatted(branch));
        }

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,285 @@ protected String createSchemaSql(String schemaName)
...
+        assertUpdate("CREATE BRANCH test_branch IN TABLE " + tableName);
+        assertUpdate("DELETE FROM " + tableName + "@test_branch", 2);
+
+        assertThat(query("SELECT * FROM " + tableName))
+                .matches("VALUES (1, CAST('a' AS VARCHAR)), (2, CAST('b' AS VARCHAR))");
+        assertThat(query("SELECT * FROM " + tableName + " FOR VERSION AS OF 'test_branch'"))
+                .returnsEmptyResult();
+
+        assertUpdate("DROP TABLE " + tableName);
+    }
```
findinpath: It is unclear to me how this test is different from `testDeleteFromBranch`

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,285 @@ protected String createSchemaSql(String schemaName)
...
+
+    @Test
```
findinpath, line 8263: add more failure use cases 
- DML operation into non-existing branch
- DML operation into a tab

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -7972,6 +7973,285 @@ protected String createSchemaSql(String schemaName)
...
+        assertUpdate("DELETE FROM " + tableName + "@test_branch WHERE id = 2", 1);
+
+        assertThat(query("SELECT * FROM " + tableName))
+                .matches("VALUES (1, CAST('a' AS VARCHAR)), (2, CAST('b' AS VARCHAR)), (3, CAST('c' AS VARCHAR))");
+        assertThat(query("SELECT * FROM " + tableName + " FOR VERSION AS OF 'test_branch'"))
+                .matches("VALUES (1, CAST('a' AS VARCHAR)), (3, CAST('c' AS VARCHAR))");
+
+        assertUpdate("DROP TABLE " + tableName);
+    }
+
```
findinpath, line 8262: I suggest adding the tests in a dedicated class - `TestIcebergBranching` in the `trino-iceberg` connector.

We want to have coverage for both v2 & v3 iceberg tables and BCT does not seem the right place to do these tests.
Some test operations may be related for v3 tables only (e.g. testing with deletion vectors on branches).

findinpath (commented):
Production code looks generally good - let's focus on adding tests to ensure easy maintenance and avoiding regressions.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
SebastienN15: Yes — Trino’s SQL supports CREATE OR REPLACE BRANCH. The SPI passes SaveMode.REPLACE for that case, which is exactly what lines 2372-2377 handle. We already have a test testCreateOrReplaceBranch in BaseConnectorTest.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
SebastienN15: In a previous review, @ebyhr suggested I should add branch tests in BaseConnectorTest - so I'm not sure what's best here. Happy to do the change you require if you both agree.
Or we could keep the generic DDL/DML branch tests in BaseConnectorTest (since branching is an SPI-level feature, not Iceberg-specific), and add Iceberg-specific tests (like testMetadataDeleteFromBranch, testBranchVisibleInRefsTable, and future v3/deletion vector tests) in the Iceberg test classes. wdyt?

ebyhr: Adding TestIcebergBranching is fine. However, we should also update BaseConnectorTest.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
SebastienN15: Addressed via `getLatestSnapshot()` utility which uses `verify()` for null safety — slightly different pattern but same intent.

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
SebastienN15: Moved `testMetadataDeleteFromBranch` to `BaseIcebergConnectorTest` since it tests an Iceberg-specific code path (metadata delete via `DeleteFiles`) distinct from row-level delete (`RowDelta`).

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
SebastienN15: Added test for DML into non-existing branch. Tag test deferred to Iceberg-specific test class since tags require Iceberg-specific setup.

findinpath:
please add "Fixes https://github.com/trinodb/trino/issues/12844" to the description of the PR

PavelkoSemen:
@findinpath Hi! Could you please let me know what's still missing for this MR to be merged? Branches in Iceberg are already supported by almost all engines, except for Trino.

SebastienN15:
@PavelkoSemen I'm still waiting for my CLA to be reviewed / approved. It's been a while I've send it but got no news since then

findinpath:
@cla-bot check

testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
SebastienN15: Thanks for the suggestion — it’s in place.

`BaseConnectorTest` now carries the branch contract tests behind `TestingConnectorBehavior.SUPPORTS_BRANCH`, with `getDefaultBranchName()` so catalogs that don’t use `main` can override. Iceberg enables the behavior in `BaseIcebergConnectorTest` / smoke tests (and catalogs that don’t support branches opt out).

I also assert the `SHOW BRANCHES` result column name is `Branch` (as produced by `ShowQueriesRewrite`) in `BaseConnectorTest` and `BaseIcebergConnectorSmokeTest` for clearer failures if the protocol shape changes.

SebastienN15:
@ebyhr @findinpath would you mind taking a look 🙏 All the tests are passing now

kmurra:
@SebastienN15 Hello, I'm wondering if you're still watching this and waiting for a CLA to be approved?

mosabua:
@cla-bot check

SebastienN15:
@kmurra I couldn't get a review on this PR so I switched to another query engine.
Feel free to take it over if you need it

stefan-toubia:
Thanks for the update @SebastienN15. Coincidentally, I started looking into reworking this PR a few weeks ago, and made some more progress on it last week. That branch addresses most of the remaining open issues here, as well as a few more I found along the way. Should be ready for PR soon.

PR #28468, opened 2026-02-25
Track memory usage from equality deletes in IcebergPageSourceProvider

raunaqmorarka:
## Release notes

```markdown
## Iceberg
* Avoid worker crashes when reading from tables with a larger number of equality deletes. ({issue}`28468`)
```

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergPageSourceProvider.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergPageSourceProvider.java
@@ -1,0 +1,196 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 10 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergPageSourceProvider.java
```diff
@@ -0,0 +1,196 @@
...
+            // Memory should still be 0 before reading any pages (lazy loading)
+            assertThat(provider.getMemoryUsage()).isEqualTo(0);
+
+            // Read pages to trigger lazy loading of equality deletes
+            while (!pageSource.isFinished()) {
+                pageSource.getNextSourcePage();
+            }
+
+            // After reading, the equality delete filter should be loaded and tracked in memory
+            assertThat(provider.getMemoryUsage()).isGreaterThan(0);
```
lukasz-stec: Can we have a more real check around the actual memory usage?

lukasz-stec: Maybe an automatic check using JOL?

wendigo: What's the advantage of JOL? Current implementation of the Slice sizeOf is similar to RamUsageEstimator from lucene

raunaqmorarka: I updated it to `> 100`, its the size of a Map that is not exposed externally, so there isn't a way to get the actual memory usage.

PR #28471, opened 2026-02-26
Remove OSHI dependency

electrum:
## Summary

Remove OSHI from `trino-main` and replace `MachineInfo` with direct platform reads.

- Linux physical processor count: parse `/proc/cpuinfo` (`physical id` + `core id`) and fall back to logical processor count if topology keys are unavailable.
- Linux CPU flags: parse `/proc/cpuinfo` `flags`/`Features`, normalize tokens, then keep only flags present on every CPU section.
- macOS (dev-only): on `aarch64`, assume `neon`; otherwise return no SIMD flags.

## Why this is safe

Our OSHI usage here was only:
1. physical processor count
2. CPU feature flags

For Linux feature flags, OSHI already reads `/proc/cpuinfo` (`CpuInfo.queryFeatureFlags`). Trino then applied its own normalization/intersection logic. This change keeps Trino-owned behavior but removes the extra dependency layer.

For physical processor count, we intentionally use a simpler `/proc/cpuinfo` parser. When machine topology metadata is missing, behavior remains safe by falling back to `Runtime.availableProcessors()`.

For macOS, this only affects local development. Production is Linux-only, and Apple Silicon has baseline NEON support, so using `aarch64 -> neon` is sufficient to exercise SIMD paths in dev.

## Dependency/runtime impact

- Removes direct OSHI dependency from `trino-main`.
- Removes OSHI-related `udev` requirement from the Docker image.
- Removes no-longer-needed JPMS JNA entries (`jna-jpms`, `jna-platform-jpms`) while keeping `jna`/`jna-platform` where still required in the reactor.

## Validation

- Added Linux parser-focused tests for physical core parsing.
- Added Linux CPU flag tests for cross-CPU intersection and single-CPU parsing.
- Existing fallback behavior remains: if machine metadata cannot be read, code uses safe defaults.

### DIFF
--- a/core/trino-main/src/main/java/io/trino/util/MachineInfo.java
+++ b/core/trino-main/src/main/java/io/trino/util/MachineInfo.java
@@ -13,118 +13,210 @@
  */
 package io.trino.util;
 
+import com.google.common.base.CharMatcher;
+import com.google.common.base.Splitter;
 import com.google.common.base.StandardSystemProperty;
+import com.google.common.collect.HashMultiset;
 import com.google.common.collect.ImmutableSet;
-import oshi.SystemInfoFFM;
-import oshi.hardware.CentralProcessor;
-import oshi.hardware.CentralProcessor.ProcessorIdentifier;
+import com.google.common.collect.Iterables;
+import com.google.common.collect.Multiset;
 
-import java.util.Arrays;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Optional;
 import java.util.Set;
+import java.util.function.Supplier;
 
+import static com.google.common.base.Suppliers.memoize;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static java.lang.Math.min;
 import static java.util.Locale.ENGLISH;
+import static java.util.function.Predicate.not;
 
 public final class MachineInfo
 {
+    private static final Splitter FLAG_SPLITTER = Splitter.on(CharMatcher.whitespace())
+            .omitEmptyStrings();
+    private static final Splitter KEY_VALUE_SPLITTER = Splitter.on(':')
+            .limit(2)
+            .trimResults();
+
     // cache physical processor count, so that it's not queried multiple times during tests
-    private static volatile int physicalProcessorCount = -1;
-    private static final SystemInfoFFM SYSTEM_INFO = new SystemInfoFFM();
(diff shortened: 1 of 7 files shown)


electrum:
Given that no shipping Mac hardware (seemingly even the M5) supports SVE, we could simply return `neon` as the only CPU flag for Mac on ARM, as that has been supported since the M1. Thoughts?

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -13,118 +13,212 @@
...
-                    .getHardware()
-                    .getProcessor()
-                    .getPhysicalProcessorCount();
+        if ("Linux".equals(osName) && "amd64".equals(osArch)) {
+            totalPhysicalProcessorCount = readLinuxPhysicalProcessorCount()
+                    .orElse(availableProcessorCount);
         }
         else {
-            // ARM platforms do not support hyper threading, therefore each logical processor is separate core
+            // Fallback to logical processor count when physical core topology is not available.
```
dain, line 75: This comment seems off.  This is the non Linix on intel/amd path.

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -13,118 +13,212 @@
...
+    }
 
-            if (tokens.isEmpty()) {
+    static Set<String> parseLinuxCpuFlags(List<String> cpuInfoLines)
+    {
+        Set<String> intersection = null;
+        Set<String> sectionFlags = new HashSet<>();
+
+        // add a synthetic section separator so the final section is flushed
+        for (String line : withTrailingBlankLine(cpuInfoLines)) {
```
dain, line 150: It *might* be easier if you split the lines into sections and then process section chunks:
```java
for (List<String section> section : splitSections(cpuInfoLines))
```

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -13,118 +13,210 @@
...
+            if (pair.size() != 2) {
+                continue;
             }
-            else {
-                intersection.retainAll(tokens);
-                if (intersection.isEmpty()) {
-                    break; // nothing in common
-                }
+
+            String key = pair.get(0).toLowerCase(ENGLISH).trim();
```
wendigo: .get(0) -> getFirst

core/trino-main/src/main/java/io/trino/util/MachineInfo.java
```diff
@@ -13,118 +13,210 @@
...
+            if (!key.equals("flags") && !key.equals("features")) {
+                continue;
             }
+            sectionFlags.addAll(parseCpuFlags(pair.get(1)));
```
wendigo: getLast()

PR #28477, opened 2026-02-26
Use ExactConversionsSupport for narrowing checks

dain:
## Description
In Java 23, ExactConversionsSupport was added and this allows checking if a cast is exact without a catch/rethrow structure. This simplifies a bunch of numeric conversion logic.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/MathFunctions.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/MathFunctions.java
@@ -16,8 +16,6 @@ package io.trino.operator.scalar;
 import com.google.common.math.DoubleMath;
 import com.google.common.math.LongMath;
 import com.google.common.primitives.Doubles;
-import com.google.common.primitives.Shorts;
-import com.google.common.primitives.SignedBytes;
 import io.airlift.slice.Slice;
 import io.trino.metadata.SqlScalarFunction;
 import io.trino.spi.TrinoException;
@@ -69,6 +67,9 @@ import static java.lang.Float.floatToRawIntBits;
 import static java.lang.Float.intBitsToFloat;
 import static java.lang.Math.toIntExact;
 import static java.lang.String.format;
+import static java.lang.runtime.ExactConversionsSupport.isLongToByteExact;
+import static java.lang.runtime.ExactConversionsSupport.isLongToIntExact;
+import static java.lang.runtime.ExactConversionsSupport.isLongToShortExact;
 
 public final class MathFunctions
 {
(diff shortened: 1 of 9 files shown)


wendigo:
Nice one!

electrum (approved):

core/trino-main/src/main/java/io/trino/operator/scalar/MathFunctions.java
```diff
@@ -794,12 +791,10 @@ public static long roundSmallint(@SqlType(StandardTypes.SMALLINT) long num, @Sql
...
         long rounded = roundLong(num, decimals);
-        try {
-            return toIntExact(rounded);
-        }
-        catch (IllegalArgumentException e) {
-            throw new TrinoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for integer: " + rounded, e);
+        if (!isLongToIntExact(rounded)) {
+            throw new TrinoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for integer: " + rounded);
         }
+        return (int) rounded;
```
ksobolew, line 797: I believe this syntax is still in preview?

```java
        return switch (rounded) {
            case int exactInt -> exactInt;
            default -> throw new TrinoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for integer: " + rounded);
        };
```

electrum: The syntax is in preview, but the class is available unconditionally.

core/trino-main/src/main/java/io/trino/operator/scalar/MathFunctions.java
```diff
@@ -766,12 +767,10 @@ public static long round(@SqlType(StandardTypes.BIGINT) long num)
...
         long rounded = roundLong(num, decimals);
-        try {
-            return SignedBytes.checkedCast(rounded);
-        }
-        catch (IllegalArgumentException e) {
-            throw new TrinoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for tinyint: " + rounded, e);
+        if (!isLongToByteExact(rounded)) {
+            throw new TrinoException(NUMERIC_VALUE_OUT_OF_RANGE, "Out of range for tinyint: " + rounded);
         }
+        return (byte) rounded;
```
findepi, line 773: That feels like more error-prone version than the old pattern.
`isLongToXXXXExact` and `(XXX) value` need to match.

I appreciate this saves an exception, but this is only in case an exception is thrown anyway.
So what's the value?

PR #28484, opened 2026-02-27
Enable UnnecessarilyFullyQualified error prone check

electrum:


core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -4026,7 +4027,7 @@ private static void checkDefaultColumnValue(Session session, PlannerContext plan
         }
 
         plannerContext.getExpressionEvaluator().evaluate(
-                new io.trino.sql.ir.Cast(new io.trino.sql.ir.Constant(literalType, value), type),
```
martint, line 4029: The choice of which one to qualify was intentional. Revert.

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/PushProjectionIntoTableScan.java
```diff
@@ -103,9 +104,9 @@ public Result apply(ProjectNode project, Captures captures, Context context)
                         extractPartialTranslations(
                                 expression.getValue(),
                                 session
-                        ).entrySet().stream())
+                ).entrySet().stream())
```
martint: Unrelated

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
electrum: Yeah, that shouldn't have changed

PR #28486, opened 2026-02-28
Add TTL-based eviction and graceful degradation to TrinoFileSystemCache

kaveti:
TrinoFileSystemCache uses a ConcurrentHashMap with a hard limit (default 1000) and no eviction. In multi-tenant environments with HDFS impersonation, each unique user creates a permanent cache entry. Once the cache fills, all new FileSystem creation fails with "FileSystem max cache size has been reached", requiring worker restarts to recover.

This change adds:

- TTL-based background eviction: a daemon thread evicts shared entries idle longer than the configured expiry (default 10 minutes). Evicted entries are removed from the cache but NOT closed — FileSystemFinalizerService (PhantomReference) closes them when all references are GC'd, preventing breakage of in-flight queries.

- Graceful degradation: when the cache is full, return an uncached FileSystem instead of throwing. The query succeeds but doesn't benefit from caching. A JMX counter tracks degradation events.

- New config property hive.fs.cache.expire-after-access (default 10m, minimum 1m) to control the TTL.

- JMX counters for evictions and cacheFullDegradations.

Only shared entries (unique == 0) are evicted; unique entries from getUnique() are managed by callers via close(). The eviction executor is intentionally not shut down in closeAll() since it is a daemon thread and the singleton INSTANCE must survive test cleanup cycles.

## Description

# Fix TrinoFileSystemCache exhaustion in impersonation environments

## What

`TrinoFileSystemCache` has a hard limit (default 1000 entries) with no eviction. With HDFS impersonation enabled, each distinct user creates a permanent cache entry. Once full, every new query from a previously-unseen user fails:

```
ICEBERG_CANNOT_OPEN_SPLIT: FileSystem max cache size has been reached: 1000
```

The only recovery is restarting workers. This has caused production incidents in multi-tenant clusters running over days/weeks.

## Why this happened

The original design chose correctness over availability: no eviction means no risk of closing a `FileSystem` mid-query (`IOException: Filesystem closed`). The 1000 limit was a safety valve against unbounded memory growth. This works fine without impersonation (small key space: scheme + authority + single service user), but breaks down when `hive.hdfs.impersonation.enabled=true` creates hundreds of unique proxy user entries over time.

## What this changes

**1. TTL-based background eviction**

A daemon thread runs every 60 seconds, evicting shared entries idle longer than the configured TTL (default 10 minutes). Key safety properties:

- **Evicted entries are NOT closed.** They're removed from the cache map, but the `FileSystem` object stays alive. `FileSystemFinalizerService` (PhantomReference, already exists) closes it when all references are GC'd. This means in-flight queries holding a reference continue working.
- **Only shared entries (`unique == 0`) are evicted.** Entries from `getUnique()` have explicit lifecycle management via `close()`.
- **Double-check under `cache.compute` lock** prevents evicting entries that were accessed between the outer `forEach` check and the inner compute.

**2. Graceful degradation when cache is full**

Instead of throwing `RuntimeException`, we return an uncached `FileSystem`. The query succeeds but doesn't benefit from caching. A JMX counter (`cacheFullDegradations`) tracks this so operators can detect it.

**3. New config property**

`hive.fs.cache.expire-after-access` (default `10m`, minimum `1m`) controls the idle TTL.

**4. JMX observability**

New counters: `evictions` and `cacheFullDegradations`, both exposed via `@Managed @Nested`.

## Concurrency analysis

| Concern | Status |
|---------|--------|
| `cacheSize` consistency | `AtomicLong` with `Math.min` clamp inside `cache.compute`. Eviction and insertion on different keys use different bucket locks but share the atomic counter. No drift. |
| Race: eviction vs `getInternal` for same key | `ConcurrentHashMap.compute` serializes per-key. Double-check in eviction sees updated `lastAccessNanos` from `touchLastAccess()`. |
| Race: eviction of holder before `createFileSystemOnce()` | Theoretically possible with sub-ms TTL. Caller still gets a working FileSystem (just orphaned from cache). `FileSystemFinalizerService` handles cleanup. Impossible in practice with 10m default. |
| `System.nanoTime()` captured once per eviction scan | Conservative: stale `now` can only delay eviction (entry lives one extra cycle), never cause premature eviction. If `touchLastAccess()` ran between outer/inner check, `now - lastAccessNanos` goes negative, always `< expiryNanos`. |
| Uncached FS resource lifecycle | `createFileSystem()` registers every FS with `FileSystemFinalizerService`. Uncached FSes are cleaned up by GC + PhantomReference, same as evicted ones. |
| Singleton executor and `closeAll()` | Executor is intentionally NOT shut down in `closeAll()`. It's a daemon thread (won't block JVM exit), and the singleton must survive test `FileSystem.closeAll()` cycles. Iterating an empty cache is a no-op. |

## What this does NOT change

- No new dependencies (no Caffeine). Manual TTL with `ConcurrentHashMap` + `ScheduledExecutorService`, following the pattern used by `FileSystemFinalizerService`.
- No changes to `FileSystem.close()` semantics or the `remove()` path.
- No changes to how Kerberos credential rotation works.

## Testing

- `testFileSystemCacheGracefulDegradation` — fill cache to max, verify next request returns uncached FS (no exception), counter increments
- `testEviction` — create entry, wait for expiry, call `evictExpiredEntries()`, assert cache is empty
- `testEvictionDoesNotEvictRecentlyAccessedEntries` — access entry to reset TTL, verify it survives eviction
- `testEvictionDoesNotEvictUniqueEntries` — `getUnique()` entries are never evicted by TTL
- `testEvictionThenRecache` — evict, then verify new FS is created and cached for the same key
- `testGracefulDegradationThenEvictionRecovery` — fill cache, degrade, evict, verify cache accepts new entries again
- `testGracefulDegradationCountIsCorrect` / `testEvictionCountIsCorrect` — JMX counter verification
- `TestHdfsConfig` — default and explicit mapping for new property
- Full `lib/trino-hdfs` test suite passes

## Config

```properties
# Time after which an idle FileSystem is evicted from the cache (default: 10m, min: 1m)
hive.fs.cache.expire-after-access=10m
```

## Behavior under heavy impersonation burst

Consider 2000 unique users hitting the cluster within a 10-minute window, with `fs.cache.max-size=1000`.

**Phase 1 — Cache filling (users 1–1000):** Each new user creates a cached `FileSystem`. Normal behavior, no issues.

**Phase 2 — Cache full, degradation (users 1001–2000):** Cache is at capacity. New users trigger graceful degradation — they get an uncached `FileSystem` per request. Each uncached FS means a new HDFS connection (TCP socket, Kerberos handshake if applicable). More expensive than a cache hit, but the query succeeds. `cacheFullDegradations` JMX counter increments for each occurrence.

**Phase 3 — Eviction kicks in (within 60 seconds):** The background thread runs, evicts entries idle for >10 minutes. Early users who are done go away. Cache slots free up. New users start getting cached again.

### Potential concerns under heavy burst

1. **Connection storm during degradation:** If 1000 uncached FSes are created in quick succession, that's 1000 new HDFS connections (TCP + optional Kerberos). Each one has a socket, potentially a DFS client thread. This is temporary (until eviction frees space), but under extreme burst it could hit file descriptor limits (`ulimit -n`), HDFS NameNode connection limits, or Kerberos KDC load. **Mitigation:** The 60-second eviction interval limits the degradation window. In practice, the existing cache entries expire quickly if idle. `FileSystemFinalizerService` cleans up uncached FSes as soon as they're GC'd.

2. **No connection pooling for uncached FSes:** Each uncached FS is independent. If the same user makes 10 queries in 5 seconds during degradation, they get 10 separate uncached FSes (since the first one is returned to the caller and not cached, subsequent calls don't find it in cache). Wasteful but safe.

3. **Memory pressure from many FileSystem objects:** Each Hadoop `FileSystem` holds internal buffers, connection pools, etc. With 2000+ simultaneously live FS objects (1000 cached + many uncached), memory usage increases. GC will reclaim uncached ones eventually, but there's a transient spike.

### Bottom line

The fix **never fails queries** — that's the key improvement over the current behavior where the 1001st user's query simply crashes. The trade-off is degraded performance (uncached connections) during a burst that exceeds cache capacity, rather than failures. The eviction mechanism self-heals within 60 seconds as idle entries expire.

For clusters with very high user counts, operators should:
- Increase `hive.fs.cache.max-size` to match their peak concurrent user count
- Decrease `hive.fs.cache.expire-after-access` if users are truly transient (e.g., `5m`)
- Monitor `cacheFullDegradations` via JMX to detect when the cache is undersized

## Related issues

This addresses the well-known `FileSystem max cache size has been reached` error that affects any Trino deployment using HDFS impersonation with a diverse user base. The issue is fundamentally a missing eviction policy — the cache grows monotonically until it hits the hard limit.

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-hdfs/src/test/java/io/trino/hdfs/TestFileSystemCache.java
+++ b/lib/trino-hdfs/src/test/java/io/trino/hdfs/TestFileSystemCache.java
@@ -192,7 +192,7 @@ public class TestFileSystemCache
         cache.setCacheExpiry(Duration.ofMillis(50));
 
         Configuration conf = new Configuration(false);
-        java.net.URI uri = new URI("file:///tmp/evict_recache/");
+        URI uri = new URI("file:///tmp/evict_recache/");
         FileSystem fs1 = cache.get(uri, conf);
         assertThat(cache.getCacheSize()).isEqualTo(1);
 


wendigo:
I don't think we need this. trino-hdfs is in a maintenance mode and there is no point in adding features there. MemoryFileSystemCache already supports what you describe here

Issue #27553, opened 2025-12-04
Humongous allocations and low heap saturation when writing to S3


findepi:
`s3.streaming.part-size` defaults to 32MB and this is also the G1 region size we recommend to use (`-XX:G1HeapRegionSize=32M`)
A `byte[]` array of size `32 * 1024 * 1024`  has object size of `≈ 32 * 1024 * 1024 + object header + int for array length`, so slightly above region size. A single such allocation therefore occupies two regions, using almost exactly 50% of their capacity.
Furthermore, humongous allocations are likely to be considered old gen and garbage collected later then they could.

For an application writing to S3 with multiple threads, wasting 50% heap could be considered wasteful.

#### proposed solution

the buffer sized with `s3.streaming.part-size` is allocated in `S3OutputStream`
this buffer doesn't have to be contiguous. it can be composed of few smaller buffers and thus avoiding humongous allocation

relatedly, we should also verify that aws s3 client doesn't do internal data buffering additional to the buffering we do before invoking the s3 client

cc @pettyjamesm @raunaqmorarka @losipiuk

wendigo:
@findepi should we just lower the stream size or allocate couple of linked buffers instead?

findepi:
my proposed solution is in the desc -- linked buffers + verify that aws s3 client doesn't do internal data buffering additional to the buffering we do before invoking the s3 client

findepi:
cc @lukasz-stec

Issue #22455, opened 2024-06-20
TestDeltaLakeLocalConcurrentWritesTest.testConcurrentInsertsReconciliationForMixedInserts flaky


losipiuk:
```Error:  io.trino.plugin.deltalake.TestDeltaLakeLocalConcurrentWritesTest.testConcurrentInsertsReconciliationForMixedInserts -- Time elapsed: 0.279 s <<< ERROR!
io.trino.testing.QueryFailedException: Failed to write Delta Lake transaction log entry
	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:134)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:562)
	at io.trino.testing.DistributedQueryRunner.execute(DistributedQueryRunner.java:545)
	at io.trino.testing.QueryRunner.execute(QueryRunner.java:85)
	at io.trino.plugin.deltalake.TestDeltaLakeLocalConcurrentWritesTest.lambda$testConcurrentInsertsReconciliationForMixedInserts$17(TestDeltaLakeLocalConcurrentWritesTest.java:352)
	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
	Suppressed: java.lang.Exception: SQL: INSERT INTO test_concurrent_mixed_inserts_table_i5062zrybr SELECT COUNT(*) AS a, 20 AS part FROM test_concurrent_mixed_inserts_table_i5062zrybr WHERE part = 20
		at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:569)
		... 7 more
Caused by: io.trino.spi.TrinoException: Failed to write Delta Lake transaction log entry
	at io.trino.plugin.deltalake.DeltaLakeMetadata.finishInsert(DeltaLakeMetadata.java:2073)
	at io.trino.plugin.base.classloader.ClassLoaderSafeConnectorMetadata.finishInsert(ClassLoaderSafeConnectorMetadata.java:630)
	at io.trino.tracing.TracingConnectorMetadata.finishInsert(TracingConnectorMetadata.java:709)
[...]

PR #28488, opened 2026-03-01
Add linked buffers for S3OutputStream

tbaeg:
## Description
- Resolves https://github.com/trinodb/trino/issues/27553

Quick tests from M1 Max Macbook Pro with: 

1. Local minio.
2. Iceberg connector + JDBC catalog. 
3. Queries are from a cold start and run in succession.

## With linked buffer
```
trino> INSERT INTO iceberg.tpch_test.orders_test
    -> SELECT * FROM tpch.sf1000.orders LIMIT 100000000;
INSERT: 100000000 rows

Query 20260327_132501_00002_99kw5, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
39.39 [100M rows, 22.5GiB] [2.55M rows/s, 586MiB/s]

Query 20260327_132542_00003_99kw5, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
36.54 [102M rows, 22.5GiB] [2.79M rows/s, 631MiB/s]

Query 20260327_132620_00004_99kw5, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
36.46 [103M rows, 22.6GiB] [2.82M rows/s, 635MiB/s]

Query 20260327_132700_00005_99kw5, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
35.69 [102M rows, 22.6GiB] [2.87M rows/s, 647MiB/s]
```

## Without linked buffer (release 480)
```
trino> INSERT INTO iceberg.tpch_test.orders_test
    -> SELECT * FROM tpch.sf1000.orders LIMIT 100000000;
INSERT: 100000000 rows

Query 20260327_133103_00000_gin83, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
44.08 [101M rows, 22.5GiB] [2.29M rows/s, 523MiB/s]

Query 20260327_133209_00001_gin83, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
40.41 [102M rows, 22.5GiB] [2.53M rows/s, 571MiB/s]

Query 20260327_133309_00002_gin83, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
38.23 [101M rows, 22.5GiB] [2.65M rows/s, 603MiB/s]

Query 20260327_133350_00003_gin83, FINISHED, 1 node
Splits: 86 total, 86 done (100.00%)
37.61 [101M rows, 22.5GiB] [2.68M rows/s, 613MiB/s]
```
## Release notes

```markdown
## Delta Lake connector
* Reduce memory fragmentation and improve memory tracking for S3 file write
  operations. ({issue}`28488`)  

## Hive connector
* Reduce memory fragmentation and improve memory tracking for S3 file write
  operations. ({issue}`28488`)  

## Hudi connector
* Reduce memory fragmentation and improve memory tracking for S3 file write
  operations. ({issue}`28488`)  

## Iceberg connector
* Reduce memory fragmentation and improve memory tracking for S3 file write
  operations. ({issue}`28488`)  

## Lakehouse connector
* Reduce memory fragmentation and improve memory tracking for S3 file write
  operations. ({issue}`28488`)  
```

### DIFF
--- a/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
+++ b/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
@@ -14,10 +14,13 @@
 package io.trino.filesystem.s3;
 
 import com.google.common.annotations.VisibleForTesting;
+import com.google.common.collect.ImmutableList;
+import io.airlift.units.DataSize;
 import io.trino.filesystem.FileMayHaveAlreadyExistedException;
 import io.trino.filesystem.encryption.EncryptionKey;
 import io.trino.memory.context.AggregatedMemoryContext;
 import io.trino.memory.context.LocalMemoryContext;
+import jakarta.annotation.Nullable;
 import software.amazon.awssdk.core.exception.SdkException;
 import software.amazon.awssdk.core.sync.RequestBody;
 import software.amazon.awssdk.services.s3.S3Client;
@@ -33,39 +36,46 @@ import software.amazon.awssdk.services.s3.model.StorageClass;
 import software.amazon.awssdk.services.s3.model.UploadPartRequest;
 import software.amazon.awssdk.services.s3.model.UploadPartResponse;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.io.InputStream;
 import java.io.InterruptedIOException;
 import java.io.OutputStream;
-import java.nio.ByteBuffer;
+import java.io.SequenceInputStream;
 import java.nio.file.FileAlreadyExistsException;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Future;
 
+import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Verify.verify;
+import static io.airlift.units.DataSize.Unit.MEGABYTE;
 import static io.trino.filesystem.s3.S3FileSystemConfig.ObjectCannedAcl.getCannedAcl;
 import static io.trino.filesystem.s3.S3FileSystemConfig.S3SseType.NONE;
 import static io.trino.filesystem.s3.S3FileSystemConfig.StorageClassType.toStorageClass;
 import static io.trino.filesystem.s3.S3SseCUtils.encoded;
 import static io.trino.filesystem.s3.S3SseCUtils.md5Checksum;
 import static io.trino.filesystem.s3.S3SseRequestConfigurator.setEncryptionSettings;
-import static java.lang.Math.clamp;
-import static java.lang.Math.max;
 import static java.lang.Math.min;
+import static java.lang.Math.toIntExact;
 import static java.lang.System.arraycopy;
 import static java.net.HttpURLConnection.HTTP_PRECON_FAILED;
+import static java.util.Collections.enumeration;
 import static java.util.Objects.checkFromIndexSize;
 import static java.util.Objects.requireNonNull;
 import static java.util.Objects.requireNonNullElse;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
+import static software.amazon.awssdk.core.internal.util.Mimetype.MIMETYPE_OCTET_STREAM;
 
 final class S3OutputStream
         extends OutputStream
 {
+    private static final int INITIAL_BUFFER_SIZE = toIntExact(DataSize.of(1, MEGABYTE).toBytes());
+    private static final int MAXIMUM_BUFFER_SIZE = toIntExact(DataSize.of(8, MEGABYTE).toBytes());
+
     private final List<CompletedPart> parts = new ArrayList<>();
     private final LocalMemoryContext memoryContext;
     private final Executor uploadExecutor;
@@ -77,11 +87,11 @@ final class S3OutputStream
     private final StorageClass storageClass;
     private final ObjectCannedACL cannedAcl;
     private final Optional<EncryptionKey> key;
+    @Nullable // is null on last upload request
+    private LinkedBuffer buffer;
+    private int inFlightBytes;
 
     private int currentPartNumber;
-    private byte[] buffer = new byte[0];
-    private int bufferSize;
-    private int initialBufferSize = 64;
 
     private boolean closed;
     private boolean failed;
@@ -105,19 +115,18 @@ final class S3OutputStream
         this.storageClass = toStorageClass(context.storageClass());
         this.cannedAcl = getCannedAcl(context.cannedAcl());
         this.key = requireNonNull(key, "key is null");
-
+        this.buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
+        updateMemory();
         verify(key.isEmpty() || context.s3SseContext().sseType() == NONE, "Encryption key cannot be used with SSE configuration");
     }
 
-    @SuppressWarnings("NumericCastThatLosesPrecision")
     @Override
     public void write(int b)
             throws IOException
     {
         ensureOpen();
-        ensureCapacity(1);
-        buffer[bufferSize] = (byte) b;
-        bufferSize++;
+        buffer.write(b);
+        updateMemory();
         flushBuffer(false);
     }
 
@@ -125,19 +134,23 @@ final class S3OutputStream
     public void write(byte[] bytes, int offset, int length)
             throws IOException
     {
+        requireNonNull(bytes);
+        checkFromIndexSize(offset, length, bytes.length);
         ensureOpen();
 
+        // make sure we don't exceed the part size
         while (length > 0) {
-            ensureCapacity(length);
-
-            int copied = min(buffer.length - bufferSize, length);
-            arraycopy(bytes, offset, buffer, bufferSize, copied);
-            bufferSize += copied;
-
+            int capacity = partSize - buffer.size();
+            if (capacity >= length) {
+                buffer.write(bytes, offset, length);
+                updateMemory();
+                break;
+            }
+            buffer.write(bytes, offset, capacity);
+            updateMemory();
             flushBuffer(false);
-
-            offset += copied;
-            length -= copied;
+            offset += capacity;
+            length -= capacity;
         }
     }
 
@@ -170,12 +183,14 @@ final class S3OutputStream
 
         try {
             flushBuffer(true);
-            memoryContext.close();
             waitForPreviousUploadFinish();
         }
         catch (IOException | RuntimeException e) {
             abortUploadSuppressed(e);
             throw e;
+        }
+        finally {
+            memoryContext.close();
         }
 
         try {
@@ -189,6 +204,10 @@ final class S3OutputStream
             }
             throw new IOException(e);
         }
+        finally {
+            buffer = null;
+            memoryContext.close();
+        }
     }
 
     private void ensureOpen()
@@ -199,23 +218,10 @@ final class S3OutputStream
         }
     }
 
-    private void ensureCapacity(int extra)
-    {
-        int capacity = min(partSize, bufferSize + extra);
-        if (buffer.length < capacity) {
-            int target = max(buffer.length, initialBufferSize);
-            if (target < capacity) {
-                target += target / 2; // increase 50%
-                target = clamp(target, capacity, partSize);
-            }
-            buffer = Arrays.copyOf(buffer, target);
-            memoryContext.setBytes(buffer.length);
-        }
-    }
-
     private void flushBuffer(boolean finished)
             throws IOException
     {
+        DataStreamProvider dataStreamProvider = buffer;
         // skip multipart upload if there would only be one part
         if (finished && !multipartUploadStarted) {
             try {
@@ -225,9 +231,9 @@ final class S3OutputStream
                         location,
                         key,
                         false,
-                        buffer,
-                        0,
-                        bufferSize);
+                        dataStreamProvider);
+                buffer = null;
+                updateMemory();
                 return;
             }
             catch (Throwable e) {
@@ -237,20 +243,7 @@ final class S3OutputStream
         }
 
         // the multipart upload API only allows the last part to be smaller than 5MB
-        if ((bufferSize == partSize) || (finished && (bufferSize > 0))) {
-            byte[] data = buffer;
-            int length = bufferSize;
-
-            if (finished) {
-                this.buffer = null;
-            }
-            else {
-                this.buffer = new byte[0];
-                this.initialBufferSize = partSize;
-                bufferSize = 0;
-            }
-            memoryContext.setBytes(0);
-
+        if ((buffer.size() >= partSize) || (finished && (buffer.size() > 0))) {
             try {
                 waitForPreviousUploadFinish();
             }
@@ -259,8 +252,12 @@ final class S3OutputStream
                 abortUploadSuppressed(e);
                 throw e;
             }
+
             multipartUploadStarted = true;
-            inProgressUploadFuture = supplyAsync(() -> uploadPage(data, length), uploadExecutor);
+            inProgressUploadFuture = supplyAsync(() -> uploadPage(dataStreamProvider), uploadExecutor);
+            inFlightBytes = buffer.allocated();
+            buffer = finished ? null : buffer.startNextPart();
+            updateMemory();
         }
     }
 
@@ -282,9 +279,12 @@ final class S3OutputStream
         catch (ExecutionException e) {
             throw new IOException("Streaming upload failed", e);
         }
+        finally {
+            inFlightBytes = 0;
+        }
     }
 
-    private CompletedPart uploadPage(byte[] data, int length)
+    private CompletedPart uploadPage(DataStreamProvider dataStreamProvider)
     {
         if (uploadId.isEmpty()) {
             CreateMultipartUploadRequest request = CreateMultipartUploadRequest.builder()
@@ -312,7 +312,7 @@ final class S3OutputStream
                 .requestPayer(requestPayer)
                 .bucket(location.bucket())
                 .key(location.key())
-                .contentLength((long) length)
+                .contentLength((long) dataStreamProvider.size())
                 .uploadId(uploadId.get())
                 .partNumber(currentPartNumber)
                 .applyMutation(builder ->
@@ -324,9 +324,7 @@ final class S3OutputStream
                                 () -> setEncryptionSettings(builder, context.s3SseContext())))
                 .build();
 
-        ByteBuffer bytes = ByteBuffer.wrap(data, 0, length);
-
-        UploadPartResponse response = client.uploadPart(request, RequestBody.fromByteBuffer(bytes));
+        UploadPartResponse response = client.uploadPart(request, RequestBody.fromContentProvider(dataStreamProvider::takeInputStream, dataStreamProvider.size(), MIMETYPE_OCTET_STREAM));
 
         CompletedPart part = CompletedPart.builder()
                 .partNumber(currentPartNumber)
@@ -382,19 +380,20 @@ final class S3OutputStream
         }
     }
 
+    private void updateMemory()
+    {
+        memoryContext.setBytes((buffer == null ? 0 : buffer.allocated()) + inFlightBytes);
+    }
+
     static void putObject(
             S3Client client,
             S3Context context,
             S3Location location,
             Optional<EncryptionKey> key,
             boolean exclusiveCreate,
-            byte[] data,
-            int dataOffset,
-            int dataLength)
+            DataStreamProvider dataStreamProvider)
             throws IOException
     {
-        checkFromIndexSize(dataOffset, dataLength, data.length);
-
         PutObjectRequest request = PutObjectRequest.builder()
                 .overrideConfiguration(context::applyCredentialProviderOverride)
                 .acl(getCannedAcl(context.cannedAcl()))
@@ -402,7 +401,7 @@ final class S3OutputStream
                 .bucket(location.bucket())
                 .key(location.key())
                 .storageClass(toStorageClass(context.storageClass()))
-                .contentLength((long) dataLength)
+                .contentLength((long) dataStreamProvider.size())
                 .applyMutation(builder -> {
                     if (exclusiveCreate) {
                         builder.ifNoneMatch("*");
@@ -416,10 +415,8 @@ final class S3OutputStream
                 })
                 .build();
 
-        ByteBuffer bytes = ByteBuffer.wrap(data, dataOffset, dataLength);
-
         try {
-            client.putObject(request, RequestBody.fromByteBuffer(bytes));
+            client.putObject(request, RequestBody.fromContentProvider(dataStreamProvider::takeInputStream, dataStreamProvider.size(), MIMETYPE_OCTET_STREAM));
         }
         catch (SdkException putObjectException) {
             // When `location` already exists, the operation will fail with `412 Precondition Failed`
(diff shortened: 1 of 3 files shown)


lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -439,4 +415,89 @@ static void putObject(
...
+        {
+            writeInternal(bytes, offset, length);
+        }
+
+        public int size()
+        {
+            return size;
+        }
+
+        public byte[] asBytes()
```
wendigo: This will still result in a humongous allocation

tbaeg: Realized it after that fact, ha. Updated to be an Inputstream.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -439,4 +416,94 @@ static void putObject(
             throw new IOException("Put failed for bucket [%s] key [%s]: %s".formatted(location.bucket(), location.key(), putObjectException), putObjectException);
         }
     }
+
+    private static class LinkedBuffer
```
chenjian2664: can we add a test for it

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -439,4 +416,94 @@ static void putObject(
...
+            return size;
+        }
+
+        public InputStream asInputStream()
+        {
+            Enumeration<ByteArrayInputStream> source = enumeration(ImmutableList.copyOf(parts));
+            if (offset > 0) {
+                source = enumeration(ImmutableList.<ByteArrayInputStream>builder()
+                        .addAll(parts)
+                        .add(new ByteArrayInputStream(Arrays.copyOfRange(buffer, 0, offset)))
```
wendigo: You can use guava's LimitedInputStream to avoid copying

tbaeg: Used the other `ByteArrayInputStream` constructor instead.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: Done.

wendigo:
CI hit: https://github.com/trinodb/trino/issues/22455

wendigo:
I've reorganized code a little and fixed a small bug that was causing invalid writes. Now tests are passing :)

tbaeg:
> I've reorganized code a little and fixed a small bug that was causing invalid writes. Now tests are passing :)

Thank you for the updates! They all
make sense and and looks cleaner.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -324,9 +313,7 @@ private CompletedPart uploadPage(byte[] data, int length)
                                 () -> setEncryptionSettings(builder, context.s3SseContext())))
                 .build();
 
-        ByteBuffer bytes = ByteBuffer.wrap(data, 0, length);
-
-        UploadPartResponse response = client.uploadPart(request, RequestBody.fromByteBuffer(bytes));
+        UploadPartResponse response = client.uploadPart(request, RequestBody.fromInputStream(inputStream, length));
```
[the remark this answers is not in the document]
tbaeg: Valid concern. We'd need an `InputStream` implementation that would support `mark`/`reset` but rather not have yet another. Pushed something that wraps the `LinkedBuffer` so we can use the `RequestBody.fromContentProvider` and avoid a custom implementation. It's crude and looking for help on how folks would like it organized.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -416,10 +400,8 @@ static void putObject(
                 })
                 .build();
 
-        ByteBuffer bytes = ByteBuffer.wrap(data, dataOffset, dataLength);
-
         try {
-            client.putObject(request, RequestBody.fromByteBuffer(bytes));
+            client.putObject(request, RequestBody.fromInputStream(inputStream, dataLength));
```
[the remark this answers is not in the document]
tbaeg: See https://github.com/trinodb/trino/pull/28488#discussion_r2879956566

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -439,4 +421,95 @@ static void putObject(
...
+            this.parts = new ArrayList<>();
+            this.currentBufferSize = initialBufferSize;
+            this.currentBuffer = new byte[initialBufferSize];
+            this.currentOffset = 0;
+            this.totalSize = 0;
+        }
+
+        public void write(int b)
+        {
+            writeInternal(new byte[] {(byte) b}, 0, 1);
```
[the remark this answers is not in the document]
tbaeg: Having a consistent code path feels better, but open to this change.

tbaeg:
@wendigo Any thoughts on these updates?

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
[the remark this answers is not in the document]
wendigo: new byte[] {} allocation is an overkill

tbaeg:
@findepi We'll be working on this PR.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -33,39 +35,45 @@
...
+import static java.util.Collections.enumeration;
 import static java.util.Objects.requireNonNull;
 import static java.util.Objects.requireNonNullElse;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
+import static software.amazon.awssdk.core.internal.util.Mimetype.MIMETYPE_OCTET_STREAM;
 
 final class S3OutputStream
         extends OutputStream
 {
+    private static final int INITIAL_BUFFER_SIZE = toIntExact(DataSize.of(8, KILOBYTE).toBytes());
```
pettyjamesm: These constants seem a little small to me, typical uploads parts are going to be in the ~16MB or 32MB range right? Seems like the hazard we're avoiding is creating "humongous" objects which for G1GC (using the 32MB heap region size advised by Trino) would start just under 16MB. Maybe 4MB is a more reasonable maximum chunk size? Also, maybe 64KB is a better starting size?

I don't feel particularly strongly about the exact values chosen, other than that these values "feel" small.

pettyjamesm:
In the V1 SDK, there were issues around the specific `InputStream` sub-class that you passed in (it would special case `FileInputStream` and `ByteArrayInputStream`) that would affect whether:
- The input stream was materialized into memory to compute the MD5 signature of the upload
- The upload attempt was able to be automatically retried in the case of failure

The new SDK seems to be solving these issues by not requiring signature calculations during uploads and by using the `ContentStreamProvider` API to support arbitrary restarts- so I think that's not a problem, but I can't say for sure what gotchas may be hiding in the V2 SDK with this change. As long as the performance numbers look reasonable then we can assume no redundant copies are happening anyway.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: @wendigo can chime in a bit more for these specific numbers, but at the surface, I agree that they "feel" small.

64KB/4MB sizes for initial and maximum feel reasonable to me. I'd even argue a larger 8MB maximum doesn't feel "bad" either.  Let me know what you prefer.

tbaeg:
@pettyjamesm Let me know if there is anything else you'd like changed.

pettyjamesm:
@tbaeg - I posted this question on slack, but will copy here in case you missed it:
> Did we verify the results experimentally somehow? (e.g.: benchmark results, verifying the number of humongous regions in GC logs after running large S3 uploads, etc)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -33,39 +35,46 @@
...
 import static java.util.Objects.requireNonNull;
 import static java.util.Objects.requireNonNullElse;
 import static java.util.concurrent.CompletableFuture.supplyAsync;
+import static software.amazon.awssdk.core.internal.util.Mimetype.MIMETYPE_OCTET_STREAM;
 
 final class S3OutputStream
         extends OutputStream
 {
+    private static final int INITIAL_BUFFER_SIZE = toIntExact(DataSize.of(64, KILOBYTE).toBytes());
+    private static final int MAXIMUM_BUFFER_SIZE = toIntExact(DataSize.of(4, MEGABYTE).toBytes());
```
raunaqmorarka: The buffer sizes look unnecessarily conservative to me.
File sizes usually range from a few mbs to a gb. The default for s3.streaming.part-size is at 32mb and minimum allowed is 5mb.
I would put INITIAL_BUFFER_SIZE at 1MB and MAXIMUM_BUFFER_SIZE at 8MB

tbaeg: Increased to 1MB and 8MB.

raunaqmorarka:
@coderabbitai review

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -439,4 +417,128 @@ static void putObject(
...
+        public InputStream takeInputStream()
+        {
+            if (currentOffset == 0) {
+                return new SequenceInputStream(enumeration(parts));
+            }
+            return new SequenceInputStream(enumeration(ImmutableList.<InputStream>builder()
+                    .addAll(parts)
+                    .add(new ByteArrayInputStream(currentBuffer, 0, currentOffset))
+                    .build()));
+        }
```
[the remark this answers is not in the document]
tbaeg: Adjusted.

lib/trino-filesystem-s3/src/test/java/io/trino/filesystem/s3/TestLinkedBuffer.java
```diff
@@ -0,0 +1,168 @@
...
+
+        buffer.reset();
+
+        assertThat(buffer.size()).isZero();
+        assertThat(buffer.takeInputStream().readAllBytes()).isEmpty();
+
+        buffer.write(data, 0, data.length);
+        assertThat(buffer.size()).isEqualTo(5);
+        assertThat(buffer.takeInputStream().readAllBytes()).isEqualTo(data);
+    }
```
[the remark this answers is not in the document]
tbaeg: Test added.

raunaqmorarka:
@coderabbitai review

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -105,19 +112,17 @@ public S3OutputStream(AggregatedMemoryContext memoryContext, Executor uploadExec
...
     public void write(int b)
             throws IOException
     {
         ensureOpen();
-        ensureCapacity(1);
-        buffer[bufferSize] = (byte) b;
-        bufferSize++;
+        buffer.write(b);
+        memoryContext.setBytes(buffer.size());
         flushBuffer(false);
```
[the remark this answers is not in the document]
tbaeg: Added new `allocated` variable for tracking memory usage.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -237,17 +229,12 @@ private void flushBuffer(boolean finished)
...
-                this.buffer = null;
+                buffer = null;
             }
             else {
-                this.buffer = new byte[0];
-                this.initialBufferSize = partSize;
-                bufferSize = 0;
+                buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
             }
             memoryContext.setBytes(0);
```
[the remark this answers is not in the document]
tbaeg: Added new `allocated` variable for tracking memory usage.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -105,19 +112,17 @@ public S3OutputStream(AggregatedMemoryContext memoryContext, Executor uploadExec
         this.storageClass = toStorageClass(context.storageClass());
         this.cannedAcl = getCannedAcl(context.cannedAcl());
         this.key = requireNonNull(key, "key is null");
-
+        this.buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
         verify(key.isEmpty() || context.s3SseContext().sseType() == NONE, "Encryption key cannot be used with SSE configuration");
```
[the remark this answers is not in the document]
tbaeg: Added new `allocated` variable for tracking memory usage.

raunaqmorarka:
@coderabbitai review

findepi:
Some benchmark results posted in https://trinodb.slack.com/archives/C07ABNN828M/p1773765720522249.

@tbaeg since we retain only 90 of Slack history, can you maybe post a summary here?

tbaeg:
> Some benchmark results posted in https://trinodb.slack.com/archives/C07ABNN828M/p1773765720522249.
> 
> @tbaeg since we retain only 90 of Slack history, can you maybe post a summary here?

Yup. I actually already edited the PR description a few days ago.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -439,4 +418,147 @@ static void putObject(
...
+        @Override
+        public int size()
+        {
+            return totalSize;
+        }
+
+        @Override
+        public InputStream takeInputStream()
+        {
+            List<InputStream> inputs = parts.stream().map(ByteArrayInputStream::new).collect(toImmutableList());
```
pettyjamesm: Nit, feel free to ignore- you can avoid creating two `List<InputStream>` in the common case by checking `currentOffset == 0` ahead of time. Something like:

```
public InputStream takeInputStream()
{
  ImmutableList.Builder<InputStream> streams = ImmutableList.<InputStream>builderWithExpectedSize(parts.size() + (currentOffset == 0) ? 0 : 1);
  for (byte[] part : parts) {
    streams.add(new ByteArrayInputStream(part));
  }
  if (currentOffset > 0) {
    streams.add(new ByteArrayInputStream(currentBuffer, 0, currentOffset));
  }
  return new SequenceInputStream(enumeration(streams.build()));
}
```

tbaeg: I appreciate nits!

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -237,17 +228,12 @@ private void flushBuffer(boolean finished)
...
-                this.buffer = null;
+                buffer = null;
             }
             else {
-                this.buffer = new byte[0];
-                this.initialBufferSize = partSize;
-                bufferSize = 0;
+                buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
             }
             memoryContext.setBytes(0);
```
pettyjamesm: AI flagged this as under-reporting memory when we hit the `else` block here (and suggested calling `memoryContext.setBytes(buffer == null ? 0 : buffer.allocated());`

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -237,17 +228,12 @@ private void flushBuffer(boolean finished)
...
+        if ((buffer.size() >= partSize) || (finished && (buffer.size() > 0))) {
             if (finished) {
-                this.buffer = null;
+                buffer = null;
             }
             else {
-                this.buffer = new byte[0];
-                this.initialBufferSize = partSize;
-                bufferSize = 0;
+                buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
```
pettyjamesm: Do we want to reset the buffer to the `INITIAL_BUFFER_SIZE` after each part? Probably not critical for performance, but it would seem like we would carry over the buffer growth factor

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -225,9 +217,8 @@ private void flushBuffer(boolean finished)
                         location,
                         key,
                         false,
-                        buffer,
-                        0,
-                        bufferSize);
+                        dataStreamProvider);
+                buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
```
pettyjamesm: When `finished` below, we set `buffer = null;` instead of to a new buffer

tbaeg: Will be consistent and set to `null`.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -105,19 +111,17 @@ public S3OutputStream(AggregatedMemoryContext memoryContext, Executor uploadExec
         this.storageClass = toStorageClass(context.storageClass());
         this.cannedAcl = getCannedAcl(context.cannedAcl());
         this.key = requireNonNull(key, "key is null");
-
+        this.buffer = new LinkedBuffer(INITIAL_BUFFER_SIZE, MAXIMUM_BUFFER_SIZE);
```
pettyjamesm, line 118: AI review is flagging this as needing a `memoryContext.setBytes(buffer.allocated());` after creation to avoid under-reporting memory usage until the first write.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: I'm not sure it matters much since it only takes 3 iterations to hit the maximum buffer size (1 MB -> 8 MB). That said, I am happy to carry it over. Does a method like `copy()` or `duplicate()` make sense? It's probably moot, but I rather not reference the internal buffer size.

pettyjamesm: Sure, although both `copy()` and `duplicate()` would seem to suggest that the contents should be preserved which isn't the right mental model here. Maybe... `startNewBufferForNextPart()`

tbaeg: The name feels a bit verbose and specific (I could see the `LinkedBuffer` being used else where). 

What about `newLike()`?

pettyjamesm: I hear you on being overly specific, but you want to be able to infer the semantics from the name and here the semantics are “new (empty) instance with a continuation of the initial buffer size”. Since LinkedBuffer is nested within the S3OutputStream class it’s probably ok to refer to “startNextPart”

tbaeg: I will concede for now... but if `LinkedBuffer` is reused/moved.. I'll reignite this conversation, haha.

Updated method name.

pettyjamesm:
@coderabbitai review

pettyjamesm:
@tbaeg - expand the above AI review comments, they look relevant. One for handling negative indexes (sure, why not) but the other about memory tracking for the background upload looks right (and also hasn't been handled in the past, but we might as well fix that while we're here)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -105,39 +114,42 @@ public S3OutputStream(AggregatedMemoryContext memoryContext, Executor uploadExec
...
+        memoryContext.setBytes(buffer.allocated() + inFlightBytes);
         flushBuffer(false);
     }
 
     @Override
     public void write(byte[] bytes, int offset, int length)
             throws IOException
     {
+        checkNotNull(bytes);
+        checkPositionIndex(offset, length);
```
pettyjamesm: What you want is `checkFromIndexSize(offset, length, bytes.length);` (to verify the whole write is in-bounds for the `byte[] bytes` argument)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -237,30 +235,36 @@ private void flushBuffer(boolean finished)
...
+                // since the bytes written will already be accounted for
+                // by calls to write, we only need to track the additional
+                // allocation of the new buffer
+                bufferAllocationBytes = buffer.allocated();
+                memoryContext.addBytes(bufferAllocationBytes);
             }
-            memoryContext.setBytes(0);
 
             try {
                 waitForPreviousUploadFinish();
```
pettyjamesm, line 248: `waitForPreviousUploadFinish()` can / should set `inFlightBytes = 0;`

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -237,30 +235,36 @@ private void flushBuffer(boolean finished)
...
+                // after the previous upload is finished set memory to bytes that
```
pettyjamesm: I would move this to after the `try / catch` which should only be concerned with the waiting for the previous upload.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -225,9 +224,8 @@ private void flushBuffer(boolean finished)
                         location,
                         key,
                         false,
-                        buffer,
-                        0,
-                        bufferSize);
+                        dataStreamProvider);
+                buffer = null;
```
pettyjamesm, line 235: Missing `memoryContext` update on this path. The memory tracking code is now non-trivial, so maybe worth making it a separate method

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -237,30 +235,36 @@ private void flushBuffer(boolean finished)
...
             else {
-                this.buffer = new byte[0];
-                this.initialBufferSize = partSize;
-                bufferSize = 0;
+                buffer = buffer.startNextPart();
+                // since the bytes written will already be accounted for
+                // by calls to write, we only need to track the additional
+                // allocation of the new buffer
+                bufferAllocationBytes = buffer.allocated();
+                memoryContext.addBytes(bufferAllocationBytes);
```
pettyjamesm: 1. Instead of calling `memoryContext.addBytes` I would coalesce the memory handling logic into a single `memoryContext.setBytes` call (or a call to a new `updateMemory` method).
2. This whole block seems like it actually should come _after_ the call to `waitForPreviousUploadFinish` since that way we can ensure that memory tracked only comes from 2 places: (new buffer, new inFlightBytes) instead of 3 (old inFlightBytes, new inFlightBytes, new buffer)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: Unless I am missing something, calls to `write` will already track all the memory that is being used since this is not a multipart upload. Any update to the `memoryContext` would be redundant.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: > 1. Instead of calling `memoryContext.addBytes` I would coalesce the memory handling logic into a single `memoryContext.setBytes` call (or a call to a new `updateMemory` method).

I can try and move everything to an `updateMemory` call. I was trying to avoid more memory management refactoring, but I think we're past that point, heh.

>     2. This whole block seems like it actually should come _after_ the call to `waitForPreviousUploadFinish` since that way we can ensure that memory tracked only comes from 2 places: (new buffer, new inFlightBytes) instead of 3 (old inFlightBytes, new inFlightBytes, new buffer)

I think if we did that, it'd be an under representation of memory utilization, especially since the upload could enter a retry loop.

pettyjamesm: > I think if we did that, it'd be an under representation of memory utilization, especially since the upload could enter a retry loop.

I'm not sure I follow. If we first wait for the previous upload to finish, we know that the previous in flight memory consumption (if any) is cleared before we enter the next part which promotes the current buffer's memory usage into `inFlightBytes` (which is known to be `0` before that promotion because the previous upload finished) and creates a new buffer (which then contributes to the current allocation tracking)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
pettyjamesm: The memory consumed by `buffer` would have been previously tracked, but at this moment when we release the buffer our memory consumption should go to `0` after the putObject call completes.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: I misinterpreted what was being asked. Updated!

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: I can add, but felt like an additional call setting it to `0` was moot since immediately after we close the `memoryContext`. Thoughts?

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -259,8 +243,19 @@ private void flushBuffer(boolean finished)
                 abortUploadSuppressed(e);
                 throw e;
             }
+
+            inFlightBytes = buffer.allocated();
+            memoryContext.setBytes(inFlightBytes);
```
pettyjamesm: No need to call `memoryContext.setBytes` here, just call `updateMemory()` below unconditionally in something like this:
```
multipartUploadStarted = true;
inProgressUploadFuture = supplyAsync(() -> uploadPage(dataStreamProvider), uploadExecutor);
inFlightBytes = buffer.allocated();
buffer = finished ? null : buffer.startNextPart();
updateMemory();
```

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -382,27 +378,28 @@ private void abortUploadSuppressed(Throwable throwable)
         }
     }
 
+    private void updateMemory()
+    {
+        memoryContext.setBytes(buffer.allocated() + inFlightBytes);
```
pettyjamesm: handle `buffer == null` scenario here

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -77,11 +86,10 @@ final class S3OutputStream
     private final StorageClass storageClass;
     private final ObjectCannedACL cannedAcl;
     private final Optional<EncryptionKey> key;
+    private LinkedBuffer buffer;
```
pettyjamesm, line 91: Mark `@Nullable` and add a comment explaining that the buffer is null when the upload is finished

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
pettyjamesm: I would make this a call to `updateMemory()` (even though we expect a subsequent call to `close()`) just so that all of the memory update logic is in one place and that the accounting is updated any time that `inFlightBytes` or `buffer` changes.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
tbaeg: I specifically didn't add a `null` check to avoid continuous checks since this is called in tight loops. This is also why I explicitly call `memoryContext.setBytes` [here](https://github.com/trinodb/trino/pull/28488#discussion_r3018456255). I believe all code paths ensure there is no `NullPointerException`s.

tbaeg: I am happy to still consolidate the logic and add `null` check. Let me know if this is the preferred path.

pettyjamesm: Yeah, once you have a method for book keeping the memory it keeps things simpler to use that everywhere instead of having one-off special memory update calls or path-dependent scenarios on when it’s safe to call and when it’s not (maintainability).

The memory tracking call itself is the expensive operation in this context since it’s so lock-hungry, so no need to worry about the null check here in this context.

tbaeg: Definitely agree about better maintainability; ironically that's what I had first before changing it to this. That said, will update per your feedback!

pettyjamesm:
@coderabbitai review

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
```diff
@@ -170,8 +183,8 @@ public void close()
 
         try {
             flushBuffer(true);
-            memoryContext.close();
             waitForPreviousUploadFinish();
+            memoryContext.close();
```
pettyjamesm: Move `memoryContext.close()` to a `finally` block

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputStream.java
pettyjamesm, line 181: Add a `finally` block like:
```
catch (SdkException e) {
     throw new IOException(e);
}
finally {
    buffer = null;
    memoryContext.close();
}
```

pettyjamesm:
Merged, thanks for the contribution @tbaeg!

PR #28489, opened 2026-03-01
Use static import for JsonCodec factory methods

wendigo:
## Description

Just a cleanup

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/FailureFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/FailureFunction.java
@@ -23,9 +23,11 @@ import io.trino.spi.function.ScalarFunction;
 import io.trino.spi.function.SqlType;
 import io.trino.spi.type.StandardTypes;
 
+import static io.airlift.json.JsonCodec.jsonCodec;
+
 public final class FailureFunction
 {
-    private static final JsonCodec<FailureInfo> JSON_CODEC = JsonCodec.jsonCodec(FailureInfo.class);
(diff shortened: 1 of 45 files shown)


core/trino-main/src/main/java/io/trino/connector/CatalogPruneTask.java
```diff
@@ -82,7 +81,8 @@ public CatalogPruneTask(
             InternalNode currentNode,
             InternalNodeManager internalNodeManager,
             @ForCatalogPrune HttpClient httpClient,
-            CatalogPruneTaskConfig catalogPruneTaskConfig)
+            CatalogPruneTaskConfig catalogPruneTaskConfig,
+            JsonCodec<List<CatalogHandle>> jsonCodec)
```
martint: Why? There’s no advantage in injecting the codec.
There’s no customization, configuration or reuse needed.

dain: Ya. A long time ago we wanted these injected so we could do customization, and never did.  I actually think we should consider going the other direction and removing the injection of most (all?) json codecs.

electrum: It is needed for things like Trino types that need the type registry injected.

losipiuk (approved):

core/trino-main/src/main/java/io/trino/connector/CatalogPruneTask.java
wendigo: I'd prefer all of the json codecs to be injected. This avoids having multiple ObjectMapper instances that have internal caches etc

PR #28492, opened 2026-03-01
Add nested group resolution to LDAP group provider with optional AD matching-rule support

### LINKED ISSUE #30285: Support nested group resolution in LDAP group provider
### Problem

The LDAP group provider can resolve direct group memberships, but it does not support resolving nested group memberships in search-based LDAP group resolution.

For deployments where users are members of groups that are themselves members of parent groups, authorization checks can miss the parent groups. For example, if a user belongs to `group-a` and `group-a` belongs to `group-b`, Trino only sees `group-a`.

### Desired behavior

Search-based LDAP group resolution should optionally support nested group membership resolution, so Trino can include parent groups in the resolved group set.

For Active Directory deployments, it should also be possible to delegate nested membership resolution to LDAP using the `LDAP_MATCHING_RULE_IN_CHAIN` matching rule.

### Scope

This is intended for search-based LDAP group resolution. Attribute-based resolution using a user membership attribute such as `memberOf` should continue returning the groups provided by that attribute and should not perform recursive lookups itself.

### Related PR

Proposed implementation: #28492

OmerRaifler:
- Fixes #30285

### Motivation
- Enable resolving nested LDAP group memberships for the search-based LDAP group provider.
- Provide an option to delegate nested membership resolution to Active Directory using the `LDAP_MATCHING_RULE_IN_CHAIN` matching rule.
- Keep nested group behavior explicit and configurable without changing attribute-based (`memberOf`) group resolution.

### Description
- Added `ldap.group-search-mode` for search-based LDAP group resolution, with three supported modes:
  - `DIRECT` returns only direct group memberships and is the default.
  - `RECURSIVE` resolves nested memberships through client-side recursive LDAP searches.
  - `MATCHING_RULE_IN_CHAIN` delegates nested membership resolution to Active Directory.
- Extracted the existing direct group lookup into preparatory `LdapGroup`, `LdapGroupResolver`, and `LdapGroupSearch` components.
- Added direct, recursive, and Active Directory matching-rule resolver implementations selected at startup.
- Updated `LdapFilteringGroupProvider` to resolve the user DN and delegate group lookup to the configured resolver.
- Clarified that nested group resolution applies only to search-based group resolution. Attribute-based (`memberOf`) resolution returns the groups provided by the configured user membership attribute and does not recursively resolve parent groups.
- Updated configuration and integration tests to cover direct resolution, recursive resolution including cycles, mode mapping, and the default mode.

### Release notes
- ( ) This is not user-visible or is docs only, and no release notes are required.
- ( ) Release notes are required. Please propose a release note for me.
- (x) Release notes are required, with the following suggested text:

```markdown
# LDAP group provider
* Add support for nested group resolution in search-based LDAP group provider mode, with optional Active Directory `LDAP_MATCHING_RULE_IN_CHAIN` support.
```

### DIFF
--- a/docs/src/main/sphinx/security/group-mapping.md
+++ b/docs/src/main/sphinx/security/group-mapping.md
@@ -158,9 +158,9 @@ include the user DN. This requires the following properties:
 * - `ldap.group-search-member-attribute`
   - Attribute from group documents used for filtering by member. For example,
     `cn`.
-* - `ldap.group-search-enable-nested-groups`
+* - `ldap.group-search-nested-enabled`
   - Enable nested group resolution for search-based mode. Defaults to `false`.
-* - `ldap.group-search-use-matching-rule-in-chain`
+* - `ldap.group-search-nested-use-matching-rule-in-chain`
   - Use the Active Directory `LDAP_MATCHING_RULE_IN_CHAIN` matching rule
     (`1.2.840.113556.1.4.1941`) for nested group resolution. Defaults to
     `false`.
@@ -168,12 +168,12 @@ include the user DN. This requires the following properties:
 
 By default, search-based group resolution returns only direct groups.
 
-If `ldap.group-search-enable-nested-groups=true`, Trino recursively resolves
+If `ldap.group-search-nested-enabled=true`, Trino recursively resolves
 nested group memberships (for example user → group A → group B) by repeatedly
 searching groups that contain the previously resolved group DN as a member.
 
 For Active Directory deployments, you can additionally enable
-`ldap.group-search-use-matching-rule-in-chain=true` to delegate nested group
+`ldap.group-search-nested-use-matching-rule-in-chain=true` to delegate nested group
 resolution to LDAP itself (this is only relevant when nested group resolution
 is enabled).
 
--- a/docs/src/main/sphinx/security/group-mapping.md
+++ b/docs/src/main/sphinx/security/group-mapping.md
@@ -158,7 +158,24 @@ include the user DN. This requires the following properties:
 * - `ldap.group-search-member-attribute`
   - Attribute from group documents used for filtering by member. For example,
     `cn`.
+* - `ldap.group-search-enable-nested-groups`
+  - Enable nested group resolution for search-based mode. Defaults to `false`.
+* - `ldap.group-search-use-matching-rule-in-chain`
+  - Use the Active Directory `LDAP_MATCHING_RULE_IN_CHAIN` matching rule
+    (`1.2.840.113556.1.4.1941`) for nested group resolution. Defaults to
+    `false`.
 :::
+
+By default, search-based group resolution returns only direct groups.
+
+If `ldap.group-search-enable-nested-groups=true`, Trino recursively resolves
+nested group memberships (for example user → group A → group B) by repeatedly
+searching groups that contain the previously resolved group DN as a member.
+
+For Active Directory deployments, you can additionally enable
+`ldap.group-search-use-matching-rule-in-chain=true` to delegate nested group
+resolution to LDAP itself (this is only relevant when nested group resolution
+is enabled).
 
 In case of attribute-based group resolution, Trino reads the group list
 directly from a user attribute. This requires the following property:
--- a/docs/src/main/sphinx/security/group-mapping.md
+++ b/docs/src/main/sphinx/security/group-mapping.md
@@ -158,7 +158,20 @@ include the user DN. This requires the following properties:
 * - `ldap.group-search-member-attribute`
   - Attribute from group documents used for filtering by member. For example,
     `cn`.
+* - `ldap.group-search-mode`
+  - Group resolution mode. Supported values are `DIRECT`, `RECURSIVE`, and
+    `MATCHING_RULE_IN_CHAIN`. Defaults to `DIRECT`.
 :::
+
+The `DIRECT` mode returns only direct groups.
+
+The `RECURSIVE` mode recursively resolves
+nested group memberships (for example user → group A → group B) by repeatedly
+searching groups that contain the previously resolved group DN as a member.
+
+For Active Directory deployments, use `MATCHING_RULE_IN_CHAIN` to delegate
+nested group resolution to LDAP with the `LDAP_MATCHING_RULE_IN_CHAIN` matching
+rule (`1.2.840.113556.1.4.1941`).
 
 In case of attribute-based group resolution, Trino reads the group list
 directly from a user attribute. This requires the following property:
@@ -172,6 +185,11 @@ directly from a user attribute. This requires the following property:
 * - `ldap.user-member-of-attribute`
   - Group membership attribute in user documents. For example, `memberOf`.
 :::
+
+Nested group resolution is only supported with search-based group resolution.
+With attribute-based group resolution, Trino returns the groups provided by the
+configured user membership attribute and does not recursively resolve group
+membership.
 
 ### Example configurations
 
--- a/plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
+++ b/plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
@@ -67,8 +67,8 @@ public class LdapFilteringGroupProvider
         this.groupsNameAttribute = config.getLdapGroupsNameAttribute();
 
         String groupsSearchMemberAttribute = filteringConfig.getLdapGroupsSearchMemberAttribute();
-        this.enableNestedGroups = filteringConfig.isLdapGroupSearchEnableNestedGroups();
-        this.useMatchingRuleInChain = filteringConfig.isLdapGroupSearchUseMatchingRuleInChain();
+        this.enableNestedGroups = filteringConfig.isLdapGroupSearchNestedEnabled();
+        this.useMatchingRuleInChain = filteringConfig.isLdapGroupSearchNestedUseMatchingRuleInChain();
         String groupSearchMemberPredicate = useMatchingRuleInChain
                 ? String.format("%s:%s:={0}", groupsSearchMemberAttribute, LDAP_MATCHING_RULE_IN_CHAIN)
                 : String.format("%s={0}", groupsSearchMemberAttribute);
@@ -217,16 +217,21 @@ public class LdapFilteringGroupProvider
             throws NamingException
     {
         ImmutableSet.Builder<LdapGroup> groups = ImmutableSet.builder();
+        boolean missingConfiguredNameAttribute = false;
         while (search.hasMore()) {
             SearchResult groupResult = search.next();
             Attribute groupName = groupResult.getAttributes().get(groupsNameAttribute);
             if (groupName == null) {
-                log.warn("The group object [%s] does not have group name attribute [%s]. Falling back on object full name.", groupResult, groupsNameAttribute);
+                missingConfiguredNameAttribute = true;
+                log.debug("The group object [%s] does not have group name attribute [%s]. Falling back on object full name.", groupResult, groupsNameAttribute);
                 groups.add(new LdapGroup(groupResult.getNameInNamespace(), groupResult.getNameInNamespace()));
             }
             else {
                 groups.add(new LdapGroup(groupResult.getNameInNamespace(), groupName.get().toString()));
             }
+        }
+        if (missingConfiguredNameAttribute) {
+            log.warn("Some LDAP group objects do not have configured group name attribute [%s]. Falling back on object full name.", groupsNameAttribute);
         }
         return groups.build();
     }
--- a/plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
+++ b/plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
@@ -20,19 +20,26 @@ import io.trino.plugin.base.ldap.LdapClient;
 import io.trino.plugin.base.ldap.LdapQuery;
 import io.trino.spi.security.GroupProvider;
 
+import javax.naming.NamingEnumeration;
 import javax.naming.NamingException;
 import javax.naming.directory.Attribute;
 import javax.naming.directory.SearchResult;
 
+import java.util.ArrayDeque;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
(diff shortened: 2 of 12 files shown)


plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -96,38 +111,125 @@ public Set<String> getGroups(String user)
...
+
+    private Set<LdapGroup> extractGroups(NamingEnumeration<SearchResult> search)
+            throws NamingException
+    {
+        ImmutableSet.Builder<LdapGroup> groups = ImmutableSet.builder();
+        while (search.hasMore()) {
+            SearchResult groupResult = search.next();
+            Attribute groupName = groupResult.getAttributes().get(groupsNameAttribute);
+            if (groupName == null) {
+                log.warn("The group object [%s] does not have group name attribute [%s]. Falling back on object full name.", groupResult, groupsNameAttribute);
```
gertjanal: Consider debug log level to avoid this warning being thrown for every user?

docs/src/main/sphinx/security/group-mapping.md
```diff
@@ -158,8 +158,25 @@ include the user DN. This requires the following properties:
 * - `ldap.group-search-member-attribute`
   - Attribute from group documents used for filtering by member. For example,
     `cn`.
+* - `ldap.group-search-enable-nested-groups`
```
gertjanal: Can you sort the configuration keys alphabetically?

gertjanal: Or consider renaming them so all nested properties stay grouped (😉), so 'ldap.group-search-nested-enable' and 'ldap.group-search-nested-use-matching-rule-in-chain'

gertjanal:
I'm not able to test it IRL, but the code looks good.
I don't have the permission to approve it though

docs/src/main/sphinx/security/group-mapping.md
OmerRaifler: I renamed the properties 🙌

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: I agree that warning on every malformed group entry can become noisy.  
I changed the implementation to log the per-entry details at debug, and emit a single warn if any group is missing the configured ldap.group-name-attribute. That preserves operator visibility for misconfiguration or unexpected LDAP schema while avoiding repeated warnings for every user lookup.

OmerRaifler:
> I'm not able to test it IRL, but the code looks good. I don't have the permission to approve it though

Thanks for the review and positive feedback! I will bump this on Slack.

OmerRaifler:
I’ve deployed this change to a test Trino cluster and verified that it works as expected. Nested LDAP group resolution is working correctly in our environment.

This PR has also already received a review, so I’d like to revive it for another look now that it’s gone stale.

Praveen2112:
@OmerRaifler Can you rebase it on top of the master.

OmerRaifler:
@Praveen2112  I've rebased the branch onto master and resolved the merge conflicts. Since there were a few conflicts to work through, could you take a quick look to make sure I resolved them correctly? If anything should be handled differently, I'm happy to update it.

Praveen2112:
I think there is a code style issue as well, which needs to be fixed

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -98,40 +113,132 @@ public Set<String> getGroups(String user)
...
+                groups.add(new LdapGroup(groupResult.getNameInNamespace(), groupName.get().toString()));
+            }
+        }
+        if (missingConfiguredNameAttribute) {
+            log.warn("Some LDAP group objects do not have configured group name attribute [%s]. Falling back on object full name.", groupsNameAttribute);
+        }
+        return groups.build();
+    }
+
+    private record LdapGroup(String distinguishedName, String name) {}
```
Praveen2112: Can we move this LdapGroup extraction as a preparatory commit, so as to ease the review ?

docs/src/main/sphinx/security/group-mapping.md
```diff
@@ -158,8 +158,25 @@ include the user DN. This requires the following properties:
 * - `ldap.group-search-member-attribute`
   - Attribute from group documents used for filtering by member. For example,
     `cn`.
+* - `ldap.group-search-nested-enabled`
+  - Enable nested group resolution for search-based mode. Defaults to `false`.
+* - `ldap.group-search-nested-use-matching-rule-in-chain`
```
Praveen2112, line 163: Should we enable this only if ldap.group-search-nested-enabled set to true ?

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -20,20 +20,27 @@
...
+import java.util.ArrayDeque;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Queue;
 import java.util.Set;
 
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static java.util.Objects.requireNonNull;
 
 public class LdapFilteringGroupProvider
```
Praveen2112, line 31: So in case of a LdapSingleQueryGroupProvider we won't be supporting it out ? i.e in case of `memberOfAttribute` would it map them recursively ?

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -98,40 +113,132 @@ public Set<String> getGroups(String user)
...
         }
 
-        return userDistinguishedName.map(ldapUser -> {
+        return userDistinguishedName
+                .map(ldapUser -> {
+                    if (!enableNestedGroups) {
+                        return resolveDirectGroups(ldapUser);
+                    }
+                    return useMatchingRuleInChain ? resolveGroupsInSingleQuery(ldapUser) : resolveGroupsRecursively(ldapUser);
+                })
```
Praveen2112: Is it possible to extract them as a separate injectable class, which can be injected during the runtime.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: Sure, done.

docs/src/main/sphinx/security/group-mapping.md
OmerRaifler: Yes, good point. I updated this so `ldap.group-search-nested-use-matching-rule-in-chain` is only effective when `ldap.group-search-nested-enabled=true`.

When nested resolution is disabled, group search now uses the regular member predicate even if the matching-rule property is set. I also clarified this in the docs and added an integration test for that case.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: I avoided adding recursive resolution to `LdapSingleQueryGroupProvider` because it would no longer be a single-query provider. Recursing over `memberOf` would require querying each returned group and reading its parent membership attribute, which depends on group entries exposing `memberOf` and has different LDAP compatibility/performance characteristics. The search-based provider already has the group-search primitives needed for recursive resolution, so this PR scopes nested resolution there.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: Yes, I extracted this into an injectable resolver strategy. The filtering provider now delegates group lookup to `LdapGroupResolver`, and the module binds the direct, recursive, or matching-rule implementation based on the LDAP group filter configuration.

OmerRaifler:
Hi @Praveen2112 
I've addressed all of your review comments and pushed the requested changes. Whenever you have a chance, I'd appreciate another look. Thanks again for the thorough review!

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -134,4 +126,29 @@ public Set<String> getGroups(String user)
...
+                missingConfiguredNameAttribute = true;
+                log.debug("The group object [%s] does not have group name attribute [%s]. Falling back on object full name.", groupResult, groupsNameAttribute);
+                groups.add(new LdapGroup(groupResult.getNameInNamespace(), groupResult.getNameInNamespace()));
+            }
+            else {
+                groups.add(new LdapGroup(groupResult.getNameInNamespace(), groupName.get().toString()));
+            }
+        }
+        if (missingConfiguredNameAttribute) {
+            log.warn("Some LDAP group objects do not have configured group name attribute [%s]. Falling back on object full name.", groupsNameAttribute);
```
Praveen2112: nit: This is not a part of extraction right ?

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -134,4 +126,29 @@ public Set<String> getGroups(String user)
...
+        }
+        return groups.build();
+    }
+
+    private record LdapGroup(String distinguishedName, String name) {}
```
Praveen2112: nit: Can we add a rnn check for distinguishedName and name ?

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
Praveen2112, line 1: nit: I would also extract LdapGroupResolver as a prep commit.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapGroupProviderModule.java
```diff
@@ -30,11 +30,25 @@ protected void setup(Binder binder)
...
         }
         else {
             configBinder(binder).bindConfig(LdapSingleQueryGroupProviderConfig.class);
             binder.bind(GroupProvider.class).to(LdapSingleQueryGroupProvider.class).in(Scopes.SINGLETON);
         }
     }
+
+    private static void bindLdapGroupResolver(Binder binder, LdapFilteringGroupProviderConfig config)
+    {
+        if (!config.isLdapGroupSearchNestedEnabled()) {
```
Praveen2112: WDYT on exposing the mode as a ENUM ?

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/BaseLdapGroupResolver.java
```diff
@@ -0,0 +1,107 @@
...
+                        .withSearchFilter(combinedGroupSearchFilter)
+                        .withFilterArguments(memberDistinguishedName)
+                        .build(),
+                search -> {
+                    if (!search.hasMore()) {
+                        log.debug("No groups found using %s [pattern=%s, arguments={%s}]", searchDescription(), combinedGroupSearchFilter, memberDistinguishedName);
+                    }
+                    return extractGroups(search);
+                });
+    }
```
Praveen2112: Instead of inheriting this method, can we use composition specific to searching groups ? LDAP query can be provided by LdapGroupResolver or manually nesting through each groups. WDYT ?

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: Correct, this is not part of the extraction. I’ll keep the preparatory commit mechanical and move this behavioral change to the feature commit.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: Yes. I’ll add a compact constructor to `LdapGroup` with `requireNonNull` checks for both `distinguishedName` and `name`.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
OmerRaifler: Agreed. I’ll move the `LdapGroupResolver` extraction into the preparatory commit as well, so the feature commit is focused on adding the resolver implementations and selecting the configured mode.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapGroupProviderModule.java
OmerRaifler: I like that better. An enum such as `DIRECT`, `RECURSIVE`, and `MATCHING_RULE_IN_CHAIN` makes the modes explicit and avoids the invalid/ignored combination that is possible with two booleans. I’ll switch the configuration to a single enum-backed mode.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/BaseLdapGroupResolver.java
OmerRaifler: Agreed, composition is clearer here. I’ll replace `BaseLdapGroupResolver` inheritance with a shared LDAP group-search component responsible for query execution and result extraction. The resolvers can then compose it and provide either the direct/matching-rule query or the recursive traversal logic.

PR #28494, opened 2026-03-01
Remove location from SHOW CREATE TABLE for translated table

### LINKED ISSUE #28483: SHOW CREATE TABLE return managed table path as external_location for hive connector
**Problem Statement:**
SHOW CREATE TABLE cmd for hive connector is returning managed table path as external_location causing confusion that the table is managed or external and violates the Principle of Least Surprise.

 **Steps to reproduce:**
1. Create a schema: 
`CREATE SCHEMA hive.some_hive`

2. Create a managed table: ""
`CREATE TABLE hive.some_hive.new_table(id int, name varchar);`

3. show create table  hive.some_hive.new_table;

**Current Output:**
`trino> show create table  hive.some_hive.new_table;`
`                                      Create Table                                `
`--------------------------------------------------------------------------------------------`
`CREATE TABLE hive.some_hive.new_table (       `                                             
  `  id integer,             `                                                                
`    name varchar                     `                                                       
` )                  `                                                                        
` WITH (     `                                                                                
    `external_location = 'hdfs://localhost:9000/user/hive/warehouse/some_hive.db/new_table', `
    `format = 'ORC'                `                                                          
 `)         `                                                                                 
`(1 row)`

`Query 20260227_172350_00005_3ugvw, FINISHED, 1 node`
`Splits: 1 total, 1 done (100.00%)`
`0.37 [0 rows, 0B] [0 rows/s, 0B/s]`

`trino> `

**Expected output:**
The "external_location" keyword should be either "location" or it should not be present for managed table at all.

**Trino Version:**
Compiled from master

Kundannitp:
Fixes #28483

Currently, the SHOW CREATE TABLE command for the Hive connector returns the managed table's path as external_location. This causes confusion as to whether the table is actually managed or external, violating the Principle of Least Surprise.

This change restricts the external_location property from being printed for managed tables that were silently translated to external tables by Hive.

Additional context and related issues
Previously, the if condition only checked if the table type was EXTERNAL_TABLE before appending the external_location property to the DDL.

However, in Hive 3.x+, Hive intercepts non-transactional managed table creations and silently converts them to external tables to bypass ACID requirements. When doing so, Hive marks the table parameter "TRANSLATED_TO_EXTERNAL" as "true".

To fix this, we now explicitly check this parameter. If the table is external but was translated by Hive, we treat it as a managed table for the purpose of SHOW CREATE TABLE and omit the external_location property.

Release notes

## Hive Connector
* Fix ``SHOW CREATE TABLE`` exposing ``external_location`` for managed tables. ({issue}`28483`)

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
@@ -368,6 +368,7 @@ public class HiveMetadata
     private static final String BUCKETING_VERSION = "bucketing_version";
     public static final String STORAGE_TABLE = "storage_table";
     public static final String TRANSACTIONAL = "transactional";
+    public static final String TRANSLATED_TO_EXTERNAL = "TRANSLATED_TO_EXTERNAL";
     public static final String PRESTO_VIEW_EXPANDED_TEXT_MARKER = "/* Presto View */";
 
     public static final String ORC_BLOOM_FILTER_COLUMNS_KEY = "orc.bloom.filter.columns";
@@ -697,7 +698,11 @@ public class HiveMetadata
         // External location property
         ImmutableMap.Builder<String, Object> properties = ImmutableMap.builder();
         if (table.getTableType().equals(EXTERNAL_TABLE.name())) {
-            properties.put(EXTERNAL_LOCATION_PROPERTY, table.getStorage().getLocation());
+            boolean isHive3ConvertedManagedTable = "true".equalsIgnoreCase(
+                    table.getParameters().get(TRANSLATED_TO_EXTERNAL));
+            if (!isHive3ConvertedManagedTable) {
+                properties.put(EXTERNAL_LOCATION_PROPERTY, table.getStorage().getLocation());
+            }
         }
 
         // Storage format property
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
@@ -698,9 +698,8 @@ public class HiveMetadata
         // External location property
         ImmutableMap.Builder<String, Object> properties = ImmutableMap.builder();
         if (table.getTableType().equals(EXTERNAL_TABLE.name())) {
-            boolean isHive3ConvertedManagedTable = "true".equalsIgnoreCase(
-                    table.getParameters().get(TRANSLATED_TO_EXTERNAL));
-            if (!isHive3ConvertedManagedTable) {
+            boolean isHiveConvertedManagedTable = "true".equalsIgnoreCase(table.getParameters().get(TRANSLATED_TO_EXTERNAL));
+            if (!isHiveConvertedManagedTable) {
                 properties.put(EXTERNAL_LOCATION_PROPERTY, table.getStorage().getLocation());
             }
         }
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
@@ -2622,4 +2622,43 @@ abstract class BaseTestHiveOnDataLake
                 testTable,
                 "Overwriting existing partition in transactional tables doesn't support DIRECT_TO_TARGET_EXISTING_DIRECTORY write mode");
     }
+
+    @Test
+    public void testShowCreateTableForManagedTable()
+    {
+        String tableName = "test_hive_managed_" + randomNameSuffix();
+        String qualifiedTableName = getFullyQualifiedTestTableName(tableName);
+
+        assertUpdate(format("CREATE TABLE %s (id bigint, name varchar)", qualifiedTableName));
+        try {
+            assertThat((String) computeScalar(format("SHOW CREATE TABLE %s", qualifiedTableName)))
+                    .doesNotContain("external_location")
+                    .doesNotContain("location")
+                    .contains("format = 'ORC'");
+        }
+        finally {
+            assertUpdate(format("DROP TABLE %s", qualifiedTableName));
+        }
+    }
+
+    @Test
+    public void testShowCreateTableForExternalTable()
+    {
+        String tableName = "test_explicit_external_" + randomNameSuffix();
+        String qualifiedTableName = getFullyQualifiedTestTableName(tableName);
+        String tableLocation = format("s3a://%s/%s/%s", bucketName, HIVE_TEST_SCHEMA, tableName);
+
+        assertUpdate(format(
+                "CREATE TABLE %s (id bigint, name varchar) WITH (external_location = '%s')",
+                qualifiedTableName,
+                tableLocation));
+        try {
+            assertThat((String) computeScalar(format("SHOW CREATE TABLE %s", qualifiedTableName)))
+                    .contains("external_location = '" + tableLocation + "'")
+                    .contains("format = 'ORC'");
+        }
+        finally {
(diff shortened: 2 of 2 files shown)


ebyhr:
Could you please add a test?

Kundannitp:
@ebyhr Thanks for the review.
Actually tests are already present for this scenario but the error can only be observed if tests run on a actual hive and trino server. Please refer this https://github.com/trinodb/trino/issues/28483#issuecomment-3977648858 for why its not captured in current tests.
So not sure if we can add a test which will run on a real setup.
If you can point out any scenario where such a test added I can use it as a reference.

ebyhr:
@Kundannitp We already have test classes (e.g. BaseTestHiveOnDataLake) with dockerized Hive.

Kundannitp:
okay thanks 
Let me check

Kundannitp:
@ebyhr 
Thanks for pointing out the correct test file.
Added tests, pls check

Kundannitp:
@ebyhr I got my CLA, and checks are green. Can we merge this now?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
```diff
@@ -697,7 +698,11 @@ else if (isTrinoView || isTrinoMaterializedView) {
         // External location property
         ImmutableMap.Builder<String, Object> properties = ImmutableMap.builder();
         if (table.getTableType().equals(EXTERNAL_TABLE.name())) {
-            properties.put(EXTERNAL_LOCATION_PROPERTY, table.getStorage().getLocation());
+            boolean isHive3ConvertedManagedTable = "true".equalsIgnoreCase(
```
ebyhr: Why do we need to include 3 as the variable name? Hive 4 and later versions don't have this concept?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
```diff
@@ -697,7 +698,11 @@ else if (isTrinoView || isTrinoMaterializedView) {
...
+                    table.getParameters().get(TRANSLATED_TO_EXTERNAL));
```
ebyhr: Unwrap.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2622,4 +2622,38 @@ public void testInsertOverwritePartitionedAndBucketedAcidTable()
...
+        String qualifiedTableName = getFullyQualifiedTestTableName(tableName);
+        String tableLocation = format("s3a://%s/%s/%s", bucketName, HIVE_TEST_SCHEMA, tableName);
+
+        assertUpdate("CREATE TABLE " + qualifiedTableName + "(id bigint, name varchar) WITH (external_location = '" + tableLocation + "')");
+        try {
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + qualifiedTableName))
+                    .contains("external_location = '" + tableLocation + "'");
+        }
+        finally {
+            assertUpdate("DROP TABLE " + qualifiedTableName);
```
ebyhr, line 2657: These tests only use Trino executor. Could you add tests to verify Hive runtime?

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
chenjian2664: Yea, the name is misleading, the concept only exists in 3+(4 here)

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
```diff
@@ -2622,4 +2622,38 @@ public void testInsertOverwritePartitionedAndBucketedAcidTable()
...
+    public void testShowCreateTableForManagedTable()
+    {
+        String tableName = "test_hive_managed_" + randomNameSuffix();
+        String qualifiedTableName = getFullyQualifiedTestTableName(tableName);
+
+        assertUpdate("CREATE TABLE " + qualifiedTableName + "(id bigint, name varchar)");
+        try {
+            assertThat((String) computeScalar("SHOW CREATE TABLE " + qualifiedTableName))
+                    .doesNotContain("external_location")
+                    .doesNotContain("location");
```
chenjian2664: remove `location`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
Kundannitp: Yes makes sense I'll change this. Thanks

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
Kundannitp: Thought to be more specific with assertions so that no trace of location should be found.
But yea not required as such will remove it. Thanks!!

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
Kundannitp: Sure thanks

Kundannitp: Tried addressing your comment by adding two tests where table creation is happened via hive runtime and show create is done by trino cli.
Pls let me know if you are expecting the same thing or something else I can improve.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
Kundannitp: Tried addressing this in recent fixup commit.
Please check

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
Kundannitp: Tried addressing it in the recent fixup commit.
Please check.

chenjian2664:
I have a question regarding the `TRANSLATED_TO_EXTERNAL` property. Is it exclusively set and managed by the HMS?
Can this value be altered by any user or system operations after creation? I'm trying to assess whether it is robust enough to rely on @Kundannitp

Kundannitp:
@chenjian2664 
Yes, this property is automatically set by the Hive Metastore when translating non-ACID managed tables, making it a reliable indicator for our logic.

To address the mutability aspect:

From Trino: This property is highly restricted. Trino blocks users from modifying it via ALTER TABLE ... SET PROPERTIES, so it is effectively immutable for pure Trino users.

From Hive: It is technically mutable if a user connects directly to Hive and runs ALTER TABLE ... SET TBLPROPERTIES.

If a user explicitly connects to Hive and overwrites this to false, the external_location will indeed become visible again in SHOW CREATE TABLE. However, I don't see this as a security or behavioral concern. A user executing that command already has modify access to the table and is intentionally overriding Hive's default managed state, meaning they assume responsibility for manually altering that metadata.

Let me know if this aligns with your thoughts!

Kundannitp:
@chenjian2664 
Can you pls check my [explanation](https://github.com/trinodb/trino/pull/28494#issuecomment-4168543754) for your [question](https://github.com/trinodb/trino/pull/28494#issuecomment-4159803004).

Kundannitp:
@chenjian2664 and @ebyhr this PR is pending from long time can we please review this and close.

PR #28498, opened 2026-03-02
Use apache/iceberg-rest-fixture for Iceberg REST Catalog tests

findinpath:
## Description
Switch from using outdated `tabulario/iceberg-rest` Docker image to `apache/iceberg-rest-fixture` Docker image for testing Iceberg REST Catalog functionality.

`EnvSinglenodeSparkIcebergRest` usage of `tabulario/iceberg-rest` is intentionally omitted from this contribution because the actual image of `apache/iceberg-rest-fixture` does not include hadoop apache library dependency and can't therefore communicate with `hdfs` file system.

https://github.com/trinodb/trino/blob/e64b425bb194386f002abe36a2b2f56220ed2200/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/environment/EnvSinglenodeSparkIcebergRest.java#L51

## Release notes

```markdown
## Section
```


testing/trino-testing-containers/src/main/java/io/trino/testing/containers/IcebergRestCatalogBackendContainer.java
```diff
@@ -32,7 +32,7 @@ public IcebergRestCatalogBackendContainer(
             String minioSessionToken)
     {
         super(
-                "tabulario/iceberg-rest:1.6.0",
```
ebyhr, line 35: I intentionally hadn't updated this image because the older image was helpful to catch non-backward compatible changes in Iceberg repository. That said, I'm okay with this change since the image was too old.

Issue #28499, opened 2026-03-02
JtsGeometrySerde throws exception for some geometries


umartin:
### Steps to reproduce

`select st_points(
	st_geometryfromtext(
		'POLYGON ((489960.9893476785 6104581.829743207, 490020.7763635455 6104522.042727341, 490020.77636354364 6104522.042727342, 489960.9893476785 6104581.829743207))'
	)
);`

`SQL Error [65536]: Query failed (#20260302_111148_16652_gx8pn): shell is null but hole found`

### Cause

JtsGeometrySerde.isClockwise can return incorrect results for certain geometries due to floating-point rounding errors. This leads to misclassification of polygon rings and invalid polygon construction.

JtsGeometrySerde relies on ring orientation (clockwise vs. counterclockwise coordinate ordering) to distinguish between:

- Outer rings (shells)
- Inner rings (holes)

The method JtsGeometrySerde.isClockwise determines the orientation of a polygon ring. However, due to floating-point rounding errors in the orientation calculation, the method may return an incorrect result for some geometries.

When this occurs:

- An outer ring may be incorrectly identified as a hole.
- The polygon can be interpreted as containing only a hole without a corresponding outer shell.
- This results in an invalid or malformed polygon during deserialization.

PR #28500, opened 2026-03-02
Fix JtsGeometrySerde floating point rounding errors

umartin:
## Description

Fix error in JtsGeometrySerde caused by floating point rounding errors in isClockwise.

Replace isClockwise with Orientation.isCCW in JTS.

#28499

## Release notes

(x) Release notes are required, with the following suggested text: Fix serde failure for some geometries


wendigo:
I've amended commit message

PR #28506, opened 2026-03-03
Add random name suffix to SQL Server test tables

LBPeraza:
## Description

Multiple test functions create tables with names from testTableNameTestData, which can cause flaky failures when running them concurrently. Adding these random suffixes should make sure the table names never collide.

Example stack trace:

```
io.trino.testing.QueryFailedException: line 1:1: Table 'sqlserver.dbo.lowercase' already exists
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:631)
	at io.trino.testing.DistributedQueryRunner.executeWithPlan(DistributedQueryRunner.java:613)
	at io.trino.testing.QueryAssertions.assertDistributedUpdate(QueryAssertions.java:112)
	at io.trino.testing.QueryAssertions.assertUpdate(QueryAssertions.java:66)
	at io.trino.testing.QueryAssertions.assertUpdate(QueryAssertions.java:60)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:419)
	at io.trino.testing.AbstractTestQueryFramework.assertUpdate(AbstractTestQueryFramework.java:414)
	at io.trino.plugin.sqlserver.TestSqlServerConnectorTest.testCreateAndDropTableWithSpecialCharacterName(TestSqlServerConnectorTest.java:184)
```

## Release notes

### DIFF
--- a/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
+++ b/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
@@ -178,6 +178,7 @@ public class TestSqlServerConnectorTest
     public void testCreateAndDropTableWithSpecialCharacterName()
     {
         for (String tableName : testTableNameTestData()) {
+            tableName = addRandomNameSuffix(tableName);
             String tableNameInSql = "\"" + tableName.replace("\"", "\"\"") + "\"";
             // Until https://github.com/trinodb/trino/issues/17 the table name is effectively lowercase
             tableName = tableName.toLowerCase(ENGLISH);
@@ -215,6 +216,7 @@ public class TestSqlServerConnectorTest
     public void testRenameFromToTableWithSpecialCharacterName()
     {
         for (String tableName : testTableNameTestData()) {
+            tableName = addRandomNameSuffix(tableName);
             String tableNameInSql = "\"" + tableName.replace("\"", "\"\"") + "\"";
             String sourceTableName = "test_rename_source_" + randomNameSuffix();
             assertUpdate("CREATE TABLE " + sourceTableName + " AS SELECT 123 x", 1);
@@ -275,5 +277,17 @@ public class TestSqlServerConnectorTest
                 .add("open[bracket")
                 .add("close]bracket")
                 .build();
+    }
+
+    /**
+     * Returns a new name with a random suffix which maintains the shape of the name.
+     *
+     * <p>In particular, the suffix is added before any trailing spaces.
+     */
+    private static String addRandomNameSuffix(String name)
+    {
+        String trimmed = name.stripTrailing();
+        String trailingWhitespace = name.substring(trimmed.length());
+        return format("%s%s%s", trimmed, randomNameSuffix(), trailingWhitespace);
     }
 }


plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
```diff
@@ -276,4 +278,14 @@ private List<String> testTableNameTestData()
...
     }
+
+    /**
+     * Returns a new name with a random suffix which maintains the shape of the name.
+     *
+     * <p>In particular, the suffix is added before any trailing spaces.
+     */
+    private static String addNameSuffix(String name)
+    {
+        return name.replaceFirst("(\\s*)$", randomNameSuffix() + "$1");
```
chenjian2664: It's hard to read the test about the table name, does the fix of `testTableNameTestData` works ?

ebyhr:
@chenjian2664 It runs concurrently since AbstractTestQueryFramework has `@TestInstance(PER_CLASS)` & `@Execution(CONCURRENT)`.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
ebyhr: Another approach is to use AbstractTestQueryFramework#executeExclusively method. It still has a shared problem - a failure may affect other tests since the relevant test don't have finally block to drop tables. 

Or just changing the prefix in each test likes #28508 might be sufficient.

chenjian2664:
@ebyhr Thanks for the explanation

ebyhr:
There are 3 usages of testTableNameTestData, and each test has a different prefix: 
* testCreateAndDropTableWithSpecialCharacterName: (no prefix)
* testRenameColumnNameAdditionalTests: tcn_
* testRenameFromToTableWithSpecialCharacterName: test_rename_source_

I'm not sure where the conflict is occurring. TestSqlServerConnectorTest vs BaseConnectorTest?

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerConnectorTest.java
marcinsbd: I believe those randomized suffixes are good enough.
True, we can move drop to finally, and   iirc synapse does not support "IF EXIST" in drop statement

marcinsbd: @LBPeraza  is adding randomized suffix connected with running concurrently profiles|suites: weekly and nightly?

LBPeraza:
> we seems not run the test concurrently? It's better to share the logs of the failure test

I've added a failure stack trace to the PR description.

> There 3 usages of testTableNameTestData, and each test has a different prefix:
> 
> * testCreateAndDropTableWithSpecialCharacterName: (no prefix)
> * testRenameColumnNameAdditionalTests: tcn_
> * testRenameFromToTableWithSpecialCharacterName: test_rename_source_
> 
> I'm not sure where the conflict is occurring. TestSqlServerConnectorTest vs BaseConnectorTest?

`testRenameFromToTableWithSpecialCharacterName` initially creates the table with a `test_rename_source_` prefix, but then renames it to the name without a prefix, which conflicts with the table created in `testCreateAndDropTableWithSpecialCharacterName`.

ebyhr (approved):

PR #28507, opened 2026-03-03
Reuse IcebergPageSourceProvider across splits for a scan in Lakehouse

raunaqmorarka:
## Description
This is necessary to reuse equality deletes in an iceberg scan across splits in a worker

## Release notes

```markdown
## Lakehouse
* Improved performance and memory usage when [Equality Delete](https://iceberg.apache.org/spec/#equality-delete-files) files are used ({issue}`28507`)
```

### DIFF
--- a/plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
+++ b/plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
@@ -23,10 +23,17 @@ import io.trino.plugin.hudi.HudiTableHandle;
 import io.trino.plugin.iceberg.IcebergPageSourceProviderFactory;
 import io.trino.plugin.iceberg.IcebergTableHandle;
 import io.trino.plugin.iceberg.system.files.FilesTableSplit;
+import io.trino.spi.connector.ColumnHandle;
+import io.trino.spi.connector.ConnectorPageSource;
 import io.trino.spi.connector.ConnectorPageSourceProvider;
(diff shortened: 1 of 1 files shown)


chenjian2664 (approved):

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehousePageSourceProviderFactory.java
```diff
@@ -54,8 +61,36 @@ public LakehousePageSourceProviderFactory(
...
+        // createPageSourceProvider is called for each scan within a query
+        // we hold on to ConnectorPageSourceProvider instance to allow IcebergPageSourceProvider to reuse equality deletes between splits of the same scan
+        return new ConnectorPageSourceProvider()
+        {
+            private volatile ConnectorPageSourceProvider delegate;
+
+            @Override
+            public ConnectorPageSource createPageSource(ConnectorTransactionHandle transaction, ConnectorSession session, ConnectorSplit split, ConnectorTableHandle table, List<ColumnHandle> columns, DynamicFilter dynamicFilter)
+            {
+                if (delegate == null) {
```
oskar-szwajkowski, line 73: instead of explicit double checked locking you could use `Suppliers.memoize()` which under the hood does the same I believe, this way `volatile` can also be omitted as its encapsulated within memoizer

raunaqmorarka: The arguments to `forHandle` need to come from the first call to createPageSource, we end up either needing an AtomicReference to Supplier or maintaining additional fields for the arguments. So it's not any less verbose to use `Suppliers.memoize()`

PR #28508, opened 2026-03-03
Fix table name in BaseConnectorTest

ebyhr:
## Description

Fix table name in BaseConnectorTest

## Release notes

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
@@ -6208,7 +6208,7 @@ public abstract class BaseConnectorTest
     protected void testColumnName(String columnName, boolean delimited)
     {
         String nameInSql = toColumnNameInSql(columnName, delimited);
-        String tableName = "tcn_" + nameInSql.toLowerCase(ENGLISH).replaceAll("[^a-z0-9]", "") + randomNameSuffix();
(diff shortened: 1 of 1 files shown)


testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorTest.java
```diff
@@ -6208,7 +6208,7 @@ public void testColumnName()
     protected void testColumnName(String columnName, boolean delimited)
     {
         String nameInSql = toColumnNameInSql(columnName, delimited);
-        String tableName = "tcn_" + nameInSql.toLowerCase(ENGLISH).replaceAll("[^a-z0-9]", "") + randomNameSuffix();
+        String tableName = "test_column_" + nameInSql.toLowerCase(ENGLISH).replaceAll("[^a-z0-9]", "") + randomNameSuffix();
```
chenjian2664, line 6211: Is this abbreviation due to the Oracle connector, does the connector ok with it?

ebyhr: We've already upgraded the image in #26573, and used longer name in https://github.com/trinodb/trino/pull/26956

chenjian2664: ah ignore, oracle container version already updated.

PR #28512, opened 2026-03-03
Add and use linked byte buffers for S3OutputStream

### LINKED ISSUE #27553: Humongous allocations and low heap saturation when writing to S3

wendigo:
This avoid both data copying when buffer grows and single humongous allocations while uploading data to s3.

## Description

- Fixes https://github.com/trinodb/trino/issues/27553

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputFile.java
+++ b/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3OutputFile.java
@@ -19,6 +19,7 @@ import io.trino.filesystem.encryption.EncryptionKey;
 import io.trino.memory.context.AggregatedMemoryContext;
 import software.amazon.awssdk.services.s3.S3Client;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Optional;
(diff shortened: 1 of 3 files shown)


findepi:
is this "fixes https://github.com/trinodb/trino/issues/27553" ?

wendigo:
@findepi correct

findepi:
cc @tbaeg

tbaeg:
Let me know if I should close the other PR.

findepi:
@wendigo @tbaeg can you please work together and decide whether we go with this PR, or the previous one?
- https://github.com/trinodb/trino/pull/28488

PR #24030, opened 2024-11-04
Handle edge cases when converting double from JSON to Trino types

Pluies:
## Description

Currently, we are unable to optimise tables when they contain infinity/NaN values in fields of type "DOUBLE" because the conversion fails when we try to cast a string containing "Infinity", "-Infinity" or "NaN" into a double.

Instead, we try to check if the JSON value contains a string, and convert it appropriately.

## Additional context and related issues

Fixes #24029 
Fixes https://github.com/trinodb/trino/issues/28532

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
## Delta Lake
* Fix silent failure to write checkpoint files for tables containing double values Infinity or NaN. ({issue}`24029`)
```

PR #28539, opened 2026-03-04
Fix ClassCastException when Delta Lake DECIMAL stats are JSON Numbers

### LINKED ISSUE #28532: ClassCastException when writing Delta Lake checkpoint for Decimal columns (JSON Numeric vs String)
# Description
When Trino attempts to write a Delta Lake checkpoint for a table containing DECIMAL columns, it may fail with a java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.String.

This occurs when the preceding transaction logs (.json files) were authored by Spark, which sometimes serializes DECIMAL statistics as JSON Numbers (e.g., 36.17) rather than JSON Strings (e.g., "36.17"). Trino's CheckpointWriter strictly expects these values to be Strings during the parsing phase.

# Steps to Reproduce
Create a Delta table using Spark SQL with a DECIMAL column.
```Java
CREATE TABLE delta_part_for_insert (
    order_id BIGINT,
    customer_id BIGINT,
    order_amount DECIMAL(10, 2),
    order_ts TIMESTAMP
) USING delta
PARTITIONED BY (order_ts);
```

Insert data into the table using Spark. Spark may write the stats field in the JSON commit file as follows:

```JSON
"stats": "{\"numRecords\":1,\"minValues\":{\"amount\":36.17},\"maxValues\":{\"amount\":36.17},...}"
Note: 36.17 is a JSON Number, not a String.
```

Perform an operation in Trino that triggers a checkpoint (e.g., multiple INSERT statements until the checkpoint interval is reached).

Trino fails during the finishInsert phase while writing the checkpoint parquet file.

# Error Stacktrace
```Plaintext
java.lang.ClassCastException: class java.lang.Double cannot be cast to class java.lang.String (java.lang.Double and java.lang.String are in module java.base of loader 'bootstrap')
    at io.trino.plugin.deltalake.transactionlog.DeltaLakeParquetStatisticsUtils.jsonValueToTrinoValue(DeltaLakeParquetStatisticsUtils.java:134)
    at io.trino.plugin.deltalake.transactionlog.checkpoint.CheckpointWriter.lambda$preprocessMinMaxValues$21(CheckpointWriter.java:503)
    ...
    at io.trino.plugin.deltalake.transactionlog.checkpoint.CheckpointWriter.write(CheckpointWriter.java:169)
    at io.trino.plugin.deltalake.transactionlog.checkpoint.CheckpointWriterManager.writeCheckpoint(CheckpointWriterManager.java:170)
```

# Root Cause Analysis
In DeltaLakeParquetStatisticsUtils.jsonValueToTrinoValue, the logic for DecimalType assumes the jsonValue retrieved from the Jackson-parsed map is always a String:

```Java
if (type instanceof DecimalType decimalType) {
    BigDecimal decimal = new BigDecimal((String) jsonValue); // <--- Fails here if jsonValue is Double/Integer
    ...
}
```
While the Delta Lake protocol suggests statistics should be strings, Spark's implementation frequently optimizes small or simple Decimals as numeric types in the JSON log. Trino should be more resilient by handling both String and Number types.

# Proposed Fix
Modify jsonValueToTrinoValue to use String.valueOf(jsonValue) or perform an explicit type check before casting:

```Java
if (type instanceof DecimalType decimalType) {
    String stringValue = jsonValue instanceof String ? (String) jsonValue : String.valueOf(jsonValue);
    BigDecimal decimal = new BigDecimal(stringValue);
    ...
}
```

# Affected Versions
Tested on Trino 475 (and current master).

rameshreddy-adutla:
## Summary

Fixes #28532

Spark sometimes serializes DECIMAL column statistics as JSON Numbers (e.g., `36.17`) rather than Strings (`"36.17"`) in Delta Lake transaction log files. When Trino writes a checkpoint for such tables, `jsonValueToTrinoValue` casts the value to `String`, causing a `ClassCastException`.

## Changes

- **`DeltaLakeParquetStatisticsUtils.java`**: Use pattern matching to handle both `String` and `Number` JSON values for `DecimalType` statistics, with `String.valueOf()` as fallback for non-String types.
- **`TestDeltaLakeParquetStatisticsUtils.java`**: Added test verifying that numeric JSON values (Double, Integer) produce the same result as their String equivalents.

## Root Cause

```java
// Before — fails when Spark writes 36.17 as a JSON Number
BigDecimal decimal = new BigDecimal((String) jsonValue);

// After — handles both String and Number types
BigDecimal decimal = new BigDecimal(jsonValue instanceof String stringValue ? stringValue : String.valueOf(jsonValue));
```

### DIFF
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/DeltaLakeParquetStatisticsUtils.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/DeltaLakeParquetStatisticsUtils.java
@@ -132,7 +132,8 @@ public final class DeltaLakeParquetStatisticsUtils
             return (double) jsonValue;
         }
         if (type instanceof DecimalType decimalType) {
-            BigDecimal decimal = new BigDecimal((String) jsonValue);
+            // Spark may serialize DECIMAL statistics as JSON Numbers instead of Strings
+            BigDecimal decimal = new BigDecimal(jsonValue instanceof String stringValue ? stringValue : String.valueOf(jsonValue));
 
             if (decimalType.isShort()) {
                 return Decimals.encodeShortScaledValue(decimal, decimalType.getScale());
--- a/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/transactionlog/TestDeltaLakeParquetStatisticsUtils.java
+++ b/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/transactionlog/TestDeltaLakeParquetStatisticsUtils.java
@@ -16,6 +16,7 @@ package io.trino.plugin.deltalake.transactionlog;
 import com.google.common.collect.ImmutableMap;
 import io.airlift.slice.Slice;
 import io.airlift.slice.Slices;
+import io.trino.spi.type.DecimalType;
 import io.trino.spi.type.DoubleType;
 import io.trino.spi.type.IntegerType;
 import org.apache.parquet.column.statistics.Statistics;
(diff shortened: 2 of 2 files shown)


rameshreddy-adutla:
CLA signed and emailed to cla@trino.io — awaiting processing. 🙏

ebyhr:
Sorry, I was missiing #24030 has the same fix. Let me handle the PR instead.

rameshreddy-adutla:
No worries @ebyhr, thanks for letting me know! I see #24030 covers the same fix. Happy to help with any other issues in the future. 🙂

PR #28540, opened 2026-03-05
Switch to streaming JSON/page decoding

wendigo:
Switches some of the `ResponseHandler`-s to a streaming implementation that avoids materializing response to a single `byte[]` array. This avoid unnecessary data copying and possibly humongous allocation as well, when response is large.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/remotetask/TaskInfoFetcher.java
+++ b/core/trino-main/src/main/java/io/trino/server/remotetask/TaskInfoFetcher.java
@@ -17,9 +17,9 @@ import com.google.common.util.concurrent.Futures;
 import com.google.common.util.concurrent.ListenableFuture;
 import com.google.errorprone.annotations.concurrent.GuardedBy;
 import io.airlift.concurrent.SetThreadName;
-import io.airlift.http.client.FullJsonResponseHandler;
 import io.airlift.http.client.HttpClient;
 import io.airlift.http.client.HttpUriBuilder;
+import io.airlift.http.client.JsonResponse;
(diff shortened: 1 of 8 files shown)


wendigo:
@findepi this is not a draft. Why marking it as one?

findepi:
It has no description, so it's apparently a draft

wendigo:
@findepi please link to a doc stating that

findepi:
https://github.com/trinodb/trino/blob/master/.github/pull_request_template.md

wendigo:
@findepi there is no "mandatory" there

core/trino-main/src/main/java/io/trino/server/remotetask/TaskInfoFetcher.java
```diff
@@ -250,7 +250,7 @@ private synchronized void sendNextRequest()
                 .build();
 
         errorTracker.startRequest();
-        future = httpClient.executeAsync(request, createFullJsonResponseHandler(taskInfoCodec));
+        future = httpClient.executeAsync(request, streamingJsonResponseHandler(taskInfoCodec));
```
findepi, line 253: What is the exception message when parsing of JSON fails (e.g. required field is missing)?
What was it before the changes?

Are the responses here big enough to benefit from streaming?

wendigo: Exceptions are the same - parsing is the same except for consuming an input stream rather than a byte array

PR #28381, opened 2026-02-20
Read metadata and protocol information from Delta checksum files

adam-richardson-openai:
## Description
Read metadata and protocol information from Delta checksum files, when configured and where available

Compliant writers of Delta tables may optionally write "checksum" files alongside each commit. These checksum files contain a variety of (optional) useful information, including the Delta table metadata and protocol information. See https://github.com/delta-io/delta/blob/488c916931ca9d210f4cadd2d5520e0274d26b04/PROTOCOL.md#version-checksum-file for the full checksum file spec

Trino needs to load the table metadata and protocol information at planning time. Today, this is done by identifying and loading the latest table checkpoint, as well as replaying all intervening commits up to the latest. This can be extremely slow and expensive, as checkpoints can be enormous and there may be many intervening commits

Instead, we can simply determine the latest commit in the table, load the corresponding checksum file (if it exists), and parse the metadata and protocol information (if available in the checksum file). This takes only a single listing operation and a single load of a small JSON file, as opposed to potentially loading many files in the Delta log based approach (some of which may be extremely large depending on the size and configuration of the table)

The listing starts from `_last_checkpoint` (when present) using the `listFilesStartingFrom` filesystem primitive, so the scan only covers commits since the last checkpoint rather than the entire `_delta_log`.

If there is no checksum file for the latest eligible commit in the table, or if the checksum file doesn't capture both the metadata and the protocol information for the table, we fall back to the existing approach of scanning the Delta log. (Checksum files are considered optional under the Delta spec, as are all fields therein)

This new behavior is gated behind a session property, `load_metadata_from_checksum_file`, which in turn defaults to the value of the `delta.load-metadata-from-checksum-file` configuration. The config value itself defaults to true, since we expect this change to be a straightforward performance optimization in the overwhelming majority of cases

Repeated queries against the same table version reuse the parsed descriptor via a cross-query cache on `TransactionLogAccess` keyed by `(schema.table, location, version)`, so the `.crc` parse is skipped on subsequent calls. The cache is bounded to 1000 entries since each parsed descriptor is small. This matches the hot-cache behaviour of the transaction-log path, which already benefits from the `TableSnapshot` cache.

This optimization is particularly effective for tables using the v1 checkpoint spec, since v1 checkpoints files may be very large and heavy

We drove internal performance testing, using queries like
```
SELECT 1 FROM <table> LIMIT 1
```
where `<table>` is a large table using the v1 checkpoint spec. We observed that time spent in analysis fell from 10s on average to well under 500ms

[...]

Issue #28386, opened 2026-02-20
Delta Lake: Add support writing checksum files for Delta tables


findinpath:
Compliant writers of Delta tables may optionally write "checksum" files alongside each commit. These checksum files contain a variety of (optional) useful information, including the Delta table metadata and protocol information. See https://github.com/delta-io/delta/blob/488c916931ca9d210f4cadd2d5520e0274d26b04/PROTOCOL.md#version-checksum-file for the full checksum file spec.

This information can be use to accelerate reading Delta Lake metadata information

Inspired from https://github.com/trinodb/trino/pull/28381

PR #28549, opened 2026-03-05
Support listFilesStartingFrom in TrinoFileSystem

adam-richardson-openai:
## Description

Modern blob storage systems support performant APIs for enumerating objects starting from a user-provided cursor. This can be used to skip irrelevant objects when listing from large directories, which in turn can make the listing operation much faster and cheaper.

This PR introduces a `listFilesStartingFrom` method on `TrinoFileSystem`, with an emulated default implementation for file systems that don't support the operation natively. It includes native implementations for `AzureFileSystem`, `S3FileSystem`, and `GcsFileSystem`.

`startingFrom` is restricted to ASCII characters. Under that restriction, UTF-8 byte, Java char, and Unicode code point orderings all agree on which files to include, so the backend-native cursor (byte-ordered) and the client-side emulated filter (char-ordered) produce identical results regardless of what filenames exist in the directory.

This API was initially designed to accelerate operations on the `_delta_log` directory in Delta Lake tables. For large tables, `_delta_log` can grow to encompass tens of thousands of objects, so efficiently skipping irrelevant objects is critical to overall good performance. However, it may also be useful for optimizations on other file-based table formats as well.

This improvement came out of a previous PR, https://github.com/trinodb/trino/pull/28381, which accelerates reading Delta Lake metadata by identifying the latest available "checksum file" in the `_delta_log`. Since `_delta_log` may be enormous, it's only possible to do this performantly using `listFilesStartingFrom` or a similar API.

## Release notes

```markdown
```

### DIFF
--- a/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
+++ b/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
@@ -42,6 +42,7 @@ import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
+import io.trino.filesystem.EmulatedListFilesStartingFromIterator;
 import io.trino.filesystem.FileIterator;
 import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystem;
@@ -66,6 +67,7 @@ import java.util.concurrent.ExecutorService;
 import static com.azure.storage.common.implementation.Constants.HeaderConstants.ETAG_WILDCARD;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static io.trino.filesystem.TrinoFileSystem.checkStartingFrom;
 import static io.trino.filesystem.azure.AzureUtils.blobCustomerProvidedKey;
 import static io.trino.filesystem.azure.AzureUtils.encodedKey;
 import static io.trino.filesystem.azure.AzureUtils.handleAzureException;
@@ -307,21 +309,61 @@ public class AzureFileSystem
         try {
             // blob API returns directories as blobs, so it cannot be used when Gen2 is enabled
             return isHierarchicalNamespaceEnabled(azureLocation)
-                    ? listGen2Files(azureLocation)
-                    : listBlobFiles(azureLocation);
+                    ? listGen2Files(azureLocation, Optional.empty())
+                    : listBlobFiles(azureLocation, Optional.empty());
         }
         catch (RuntimeException e) {
             throw handleAzureException(e, "listing files", azureLocation);
         }
     }
 
-    private FileIterator listGen2Files(AzureLocation location)
+    /**
+     * {@inheritDoc}
+     *
+     * <p>On ADLS Gen2 (hierarchical namespace) the native recursive listing returns results in a
+     * depth-first order that does not match the lexicographic ordering required by this API, so
+     * the server-side {@code startFrom} cursor cannot be used directly. This implementation
+     * instead sends a conservative lower bound (truncating {@code startingFrom} at the first
+     * character {@code <= '/'}) and applies a client-side filter over the results. This preserves
+     * correctness while typically still allowing the server to skip a large portion of the
+     * directory. On flat-namespace (blob) storage the native cursor is used directly.
+     */
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        checkStartingFrom(startingFrom);
+        if (startingFrom.isEmpty()) {
+            return listFiles(location);
+        }
+
+        AzureLocation azureLocation = new AzureLocation(location);
+
+        try {
+            return isHierarchicalNamespaceEnabled(azureLocation)
+                    ? listGen2Files(azureLocation, Optional.of(startingFrom))
+                    : listBlobFiles(azureLocation, Optional.of(startingFrom));
+        }
+        catch (RuntimeException e) {
+            throw handleAzureException(e, "listing files", azureLocation);
+        }
+    }
+
+    private FileIterator listGen2Files(AzureLocation location, Optional<String> startingFrom)
             throws IOException
     {
         DataLakeFileSystemClient fileSystemClient = createFileSystemClient(location, Optional.empty());
+        ListPathsOptions options = new ListPathsOptions().setRecursive(true);
+
+        // In ADLS Gen2 recursive listing, startFrom does not implement the same lexicographic ordering as required by
+        // listFilesStartingFrom. Use a conservative server-side lower bound (truncating startingFrom), then filter client-side
+        startingFrom.map(AzureFileSystem::startingFromLowerBound)
+                .filter(not(String::isEmpty))
+                .ifPresent(options::setStartFrom);
+
         PagedIterable<PathItem> pathItems;
         if (location.path().isEmpty()) {
-            pathItems = fileSystemClient.listPaths(new ListPathsOptions().setRecursive(true), null);
+            pathItems = fileSystemClient.listPaths(options, null);
         }
         else {
             DataLakeDirectoryClient directoryClient = createDirectoryClient(fileSystemClient, location.path());
@@ -331,18 +373,27 @@ public class AzureFileSystem
             if (!directoryClient.getProperties().isDirectory()) {
                 throw new TrinoFileSystemException("Location is not a directory: " + location);
             }
-            pathItems = directoryClient.listPaths(true, false, null, null);
+
+            pathItems = directoryClient.listPaths(options, null, null);
         }
-        return new AzureDataLakeFileIterator(
+        FileIterator iterator = new AzureDataLakeFileIterator(
                 location,
                 pathItems.stream()
                         .filter(not(PathItem::isDirectory))
                         .iterator());
+        if (startingFrom.isPresent()) {
+            return new EmulatedListFilesStartingFromIterator(iterator, location.location(), startingFrom.orElseThrow());
+        }
+        return iterator;
     }
 
-    private FileIterator listBlobFiles(AzureLocation location)
+    private FileIterator listBlobFiles(AzureLocation location, Optional<String> startingFrom)
     {
-        PagedIterable<BlobItem> blobItems = createBlobContainerClient(location, Optional.empty()).listBlobs(new ListBlobsOptions().setPrefix(location.directoryPath()), null);
+        String directoryPath = location.directoryPath();
+        ListBlobsOptions options = new ListBlobsOptions().setPrefix(directoryPath);
+        startingFrom.ifPresent(startFrom -> options.setStartFrom(directoryPath.concat(startFrom)));
+
+        PagedIterable<BlobItem> blobItems = createBlobContainerClient(location, Optional.empty()).listBlobs(options, null);
         return new AzureBlobFileIterator(location, blobItems.iterator());
     }
 
@@ -660,6 +711,19 @@ public class AzureFileSystem
             throw new IllegalArgumentException();
         }
         return fileSystemClient;
+    }
+
+    private static String startingFromLowerBound(String startingFrom)
+    {
+        // Truncate startingFrom to immediately before the first character that is either equal to or lexicographically
+        // smaller than '/'. This is a workaround to disambiguate the startFrom cursor given that Azure Data Lake Storage
+        // Gen2 orders directories before files in listing results
+        for (int index = 0; index < startingFrom.length(); index++) {
+            if (startingFrom.charAt(index) <= '/') {
+                return startingFrom.substring(0, index);
+            }
+        }
+        return startingFrom;
     }
 
     private static DataLakeDirectoryClient createDirectoryClient(DataLakeFileSystemClient fileSystemClient, String directoryName)
--- a/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
+++ b/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
@@ -54,12 +54,14 @@ import java.util.Set;
 import static com.google.cloud.storage.Storage.BlobListOption.currentDirectory;
 import static com.google.cloud.storage.Storage.BlobListOption.matchGlob;
 import static com.google.cloud.storage.Storage.BlobListOption.pageSize;
+import static com.google.cloud.storage.Storage.BlobListOption.startOffset;
 import static com.google.cloud.storage.Storage.SignUrlOption.withExtHeaders;
 import static com.google.cloud.storage.Storage.SignUrlOption.withV4Signature;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.Iterables.partition;
 import static io.airlift.concurrent.MoreFutures.getFutureValue;
+import static io.trino.filesystem.TrinoFileSystem.checkStartingFrom;
 import static io.trino.filesystem.gcs.GcsUtils.encodedKey;
 import static io.trino.filesystem.gcs.GcsUtils.getBlob;
 import static io.trino.filesystem.gcs.GcsUtils.handleGcsException;
@@ -217,6 +219,24 @@ public class GcsFileSystem
         GcsLocation gcsLocation = new GcsLocation(normalizeToDirectory(location));
         try {
             return new GcsFileIterator(gcsLocation, getPage(gcsLocation));
+        }
+        catch (RuntimeException e) {
+            throw handleGcsException(e, "listing files", gcsLocation);
+        }
+    }
+
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        checkStartingFrom(startingFrom);
+        if (startingFrom.isEmpty()) {
+            return listFiles(location);
+        }
+
+        GcsLocation gcsLocation = new GcsLocation(normalizeToDirectory(location));
+        try {
+            return new GcsFileIterator(gcsLocation, getPage(gcsLocation, startOffset(gcsLocation.path() + startingFrom)));
         }
         catch (RuntimeException e) {
             throw handleGcsException(e, "listing files", gcsLocation);
--- a/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
+++ b/lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
@@ -16,6 +16,7 @@ package io.trino.filesystem.s3;
 import com.google.common.collect.HashMultimap;
 import com.google.common.collect.SetMultimap;
 import io.airlift.units.Duration;
+import io.trino.filesystem.EmulatedListFilesStartingFromIterator;
 import io.trino.filesystem.FileIterator;
 import io.trino.filesystem.Location;
 import io.trino.filesystem.TrinoFileSystem;
@@ -59,6 +60,7 @@ import static com.google.common.base.Verify.verify;
 import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.Iterables.partition;
 import static com.google.common.collect.Multimaps.toMultimap;
+import static io.trino.filesystem.TrinoFileSystem.checkStartingFrom;
 import static io.trino.filesystem.s3.S3FileSystemConfig.S3SseType.NONE;
 import static io.trino.filesystem.s3.S3SseCUtils.encoded;
 import static io.trino.filesystem.s3.S3SseCUtils.md5Checksum;
@@ -159,7 +161,7 @@ final class S3FileSystem
     public void deleteDirectory(Location location)
             throws IOException
     {
-        FileIterator iterator = listObjects(location, true);
+        FileIterator iterator = listObjects(location, true, "");
         while (iterator.hasNext()) {
             List<Location> files = new ArrayList<>();
             while ((files.size() < 1000) && iterator.hasNext()) {
@@ -237,38 +239,15 @@ final class S3FileSystem
     public FileIterator listFiles(Location location)
             throws IOException
     {
-        return listObjects(location, false);
+        return listObjects(location, false, "");
     }
 
-    private FileIterator listObjects(Location location, boolean includeDirectoryObjects)
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
             throws IOException
     {
-        S3Location s3Location = new S3Location(location);
-
-        String key = s3Location.key();
-        if (!key.isEmpty() && !key.endsWith("/")) {
-            key += "/";
-        }
-
-        ListObjectsV2Request request = ListObjectsV2Request.builder()
-                .overrideConfiguration(context::applyCredentialProviderOverride)
-                // Restore status will not be added to the response without requested
-                .optionalObjectAttributes(OptionalObjectAttributes.RESTORE_STATUS)
-                .requestPayer(requestPayer)
-                .bucket(s3Location.bucket())
-                .prefix(key)
-                .build();
-
-        try {
-            Stream<S3Object> s3ObjectStream = client.listObjectsV2Paginator(request).contents().stream();
-            if (!includeDirectoryObjects) {
-                s3ObjectStream = s3ObjectStream.filter(object -> !object.key().endsWith("/"));
-            }
-            return new S3FileIterator(s3Location, s3ObjectStream.iterator());
-        }
-        catch (SdkException e) {
-            throw new TrinoFileSystemException("Failed to list location: " + location, e);
-        }
+        checkStartingFrom(startingFrom);
+        return listObjects(location, false, startingFrom);
     }
 
     @Override
@@ -303,16 +282,7 @@ final class S3FileSystem
         S3Location s3Location = new S3Location(location);
         Location baseLocation = s3Location.baseLocation();
 
-        String key = s3Location.key();
-        if (!key.isEmpty() && !key.endsWith("/")) {
-            key += "/";
-        }
-
-        ListObjectsV2Request request = ListObjectsV2Request.builder()
-                .overrideConfiguration(context::applyCredentialProviderOverride)
-                .requestPayer(requestPayer)
-                .bucket(s3Location.bucket())
-                .prefix(key)
+        ListObjectsV2Request request = listObjectsRequest(s3Location, directoryKey(s3Location.key()))
                 .delimiter("/")
                 .build();
 
@@ -386,6 +356,64 @@ final class S3FileSystem
         catch (URISyntaxException e) {
             throw new TrinoFileSystemException("Failed to convert pre-signed URI to URI", e);
         }
+    }
+
+    private FileIterator listObjects(Location location, boolean includeDirectoryObjects, String startingFrom)
+            throws IOException
+    {
+        S3Location s3Location = new S3Location(location);
+        String keyPrefix = directoryKey(s3Location.key());
+        ListObjectsV2Request.Builder request = listObjectsRequest(s3Location, keyPrefix);
+
+        try {
+            if (!startingFrom.isEmpty()) {
+                // S3 startAfter is exclusive. Start slightly earlier and filter client-side to
+                // preserve the inclusive listFilesStartingFrom semantics.
+                String startingAfter = keyPrefix + truncateLastCodePoint(startingFrom);
+                if (!startingAfter.isEmpty()) {
+                    request.startAfter(startingAfter);
+                }
+            }
+
+            Stream<S3Object> s3ObjectStream = client.listObjectsV2Paginator(request.build()).contents().stream();
+            if (!includeDirectoryObjects) {
+                s3ObjectStream = s3ObjectStream.filter(object -> !object.key().endsWith("/"));
+            }
+
+            S3FileIterator iterator = new S3FileIterator(s3Location, s3ObjectStream.iterator());
+            if (!startingFrom.isEmpty()) {
+                return new EmulatedListFilesStartingFromIterator(iterator, s3Location.location(), startingFrom);
+            }
+            return iterator;
+        }
+        catch (SdkException e) {
+            throw new TrinoFileSystemException("Failed to list location: " + location, e);
+        }
+    }
+
+    private ListObjectsV2Request.Builder listObjectsRequest(S3Location location, String keyPrefix)
+    {
+        return ListObjectsV2Request.builder()
+                .overrideConfiguration(context::applyCredentialProviderOverride)
+                // Restore status will only be added to the response if requested
+                .optionalObjectAttributes(OptionalObjectAttributes.RESTORE_STATUS)
+                .requestPayer(requestPayer)
+                .bucket(location.bucket())
+                .prefix(keyPrefix);
+    }
+
+    private static String directoryKey(String key)
+    {
+        if (!key.isEmpty() && !key.endsWith("/")) {
+            return key + "/";
+        }
+        return key;
+    }
+
+    private static String truncateLastCodePoint(String value)
+    {
+        verify(!value.isEmpty(), "value is empty");
+        return value.substring(0, value.offsetByCodePoints(value.length(), -1));
     }
 
     private static Map<String, List<String>> filterHeaders(Map<String, List<String>> headers)
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.filesystem;
+
+import java.io.IOException;
+import java.util.NoSuchElementException;
+
+import static com.google.common.base.Preconditions.checkState;
+import static java.util.Objects.requireNonNull;
+
+public final class EmulatedListFilesStartingFromIterator
+        implements FileIterator
+{
+    private final FileIterator delegate;
+    private final String locationPath;
+    private final String startingFrom;
+    private FileEntry nextEntry;
+
+    public EmulatedListFilesStartingFromIterator(FileIterator delegate, Location location, String startingFrom)
+    {
+        this.delegate = requireNonNull(delegate, "delegate is null");
+
+        String locationPath = location.path();
+        this.locationPath = (locationPath.isEmpty() || locationPath.endsWith("/")) ? locationPath : locationPath + "/";
+
+        this.startingFrom = requireNonNull(startingFrom, "startingFrom is null");
+    }
+
+    @Override
+    public boolean hasNext()
+            throws IOException
+    {
+        loadNextEntry();
+        return nextEntry != null;
+    }
+
+    @Override
+    public FileEntry next()
+            throws IOException
+    {
+        loadNextEntry();
+        if (nextEntry == null) {
+            throw new NoSuchElementException();
+        }
+
+        FileEntry result = nextEntry;
+        nextEntry = null;
+        return result;
+    }
+
+    private void loadNextEntry()
+            throws IOException
+    {
+        if (nextEntry != null) {
+            return;
+        }
+
+        while (delegate.hasNext()) {
+            FileEntry entry = delegate.next();
+            String entryPath = entry.location().path();
+            checkState(entryPath.startsWith(locationPath), "Expected listed file to start with directory path '%s': %s", locationPath, entry.location());
+
+            String entryTail = entryPath.substring(locationPath.length());
+            if (entryTail.compareTo(startingFrom) >= 0) {
+                nextEntry = entry;
+                return;
+            }
+        }
+
+        nextEntry = null;
+    }
+}
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
@@ -24,6 +24,9 @@ import java.util.Collection;
 import java.util.Optional;
 import java.util.Set;
 
+import static com.google.common.base.Preconditions.checkArgument;
+import static java.util.Objects.requireNonNull;
+
 /**
  * TrinoFileSystem is the main abstraction for Trino to interact with data in cloud-like storage
  * systems. This replaces uses HDFS APIs in Trino.  This is not a full replacement of the HDFS, and
@@ -218,7 +221,7 @@ public interface TrinoFileSystem
      * exact name of the prefix, it is not included in the results.
      * <p>
      * The returned FileEntry locations will start with the specified location exactly
-     * and are lexicographically sorted (except for local HDFS which has the system-dependant
+     * and are lexicographically sorted (except for local HDFS which has the system-dependent
      * ordering).
      *
      * @param location the directory to list
@@ -226,6 +229,56 @@ public interface TrinoFileSystem
      */
     FileIterator listFiles(Location location)
             throws IOException;
+
+    /**
+     * Lists files within the specified directory recursively, including only files where the
+     * remainder of the path after the location is lexicographically greater than or equal to
+     * {@code startingFrom}. The location can be empty, listing all files in the file system where
+     * the path is greater than or equal to {@code startingFrom}. The location is always interpreted
+     * as a directory, whether or not it ends with a slash; {@code startingFrom} is compared against
+     * the portion of the path that follows the directory's (implicit or explicit) trailing slash.
+     * If the location does not exist, an empty iterator is returned. {@code startingFrom} must
+     * contain only ASCII characters (code points 0-127) and may be empty, in which case the
+     * behavior is identical to {@link #listFiles(Location)}.
+     * <p>
+     * For hierarchical file systems, if the path is not a directory, an exception is
+     * raised. For blob file systems, all blobs that start with the location are listed, subject
+     * to the {@code startingFrom} constraint.
+     * <p>
+     * The returned FileEntry locations start with the specified location exactly and are
+     * lexicographically sorted (except for local HDFS which has the system-dependent ordering).
+     * When non-ASCII filenames are present, the exact sort order may differ between
+     * implementations: object stores commonly sort by UTF-8 byte order, while in-memory and
+     * local implementations sort by Java {@link String#compareTo char-by-char order}. Because
+     * every non-ASCII code point is strictly greater than every ASCII character under all of
+     * these orderings, an ASCII {@code startingFrom} never excludes a non-ASCII filename.
+     * Callers that rely on a specific sort order should restrict usage to directories whose
+     * filenames are entirely ASCII.
+     *
+     * @param location the directory to list
+     * @param startingFrom the inclusive lower bound for the path remainder after {@code location};
+     *         must contain only ASCII characters
+     * @throws IllegalArgumentException if location is not valid for this file system, or if
+     *         {@code startingFrom} contains non-ASCII characters
+     */
+    default FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        checkStartingFrom(startingFrom);
+        if (startingFrom.isEmpty()) {
+            return listFiles(location);
+        }
+
+        return new EmulatedListFilesStartingFromIterator(listFiles(location), location, startingFrom);
+    }
+
+    static void checkStartingFrom(String startingFrom)
+    {
+        requireNonNull(startingFrom, "startingFrom is null");
+        for (int i = 0; i < startingFrom.length(); i++) {
+            checkArgument(startingFrom.charAt(i) <= 0x7F, "startingFrom must contain only ASCII characters");
+        }
+    }
 
     /**
      * Checks if a directory exists at the specified location. For all file system types,
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/tracing/TracingFileSystem.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/tracing/TracingFileSystem.java
@@ -120,6 +120,17 @@ final class TracingFileSystem
     }
 
     @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        Span span = tracer.spanBuilder("FileSystem.listFilesStartingFrom")
+                .setAttribute(FileSystemAttributes.FILE_LOCATION, location.toString())
+                .setAttribute(FileSystemAttributes.FILE_LIST_STARTING_FROM, startingFrom)
+                .startSpan();
+        return withTracing(span, () -> delegate.listFilesStartingFrom(location, startingFrom));
+    }
+
+    @Override
     public Optional<Boolean> directoryExists(Location location)
             throws IOException
     {
--- a/lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
+++ b/lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
@@ -1037,6 +1037,13 @@ public abstract class AbstractTestTrinoFileSystem
     }
 
     @Test
+    public void testListFilesStartingFrom()
+            throws IOException
+    {
+        testListFilesStartingFrom(isHierarchical());
+    }
+
+    @Test
     public void testPreSignedUris()
             throws IOException
     {
@@ -1148,6 +1155,177 @@ public abstract class AbstractTestTrinoFileSystem
                 // this lists a path in a directory with an empty name
                 assertThat(listPath("/")).isEmpty();
             }
+        }
+    }
+
+    protected void testListFilesStartingFrom(boolean hierarchicalNamingConstraints)
+            throws IOException
+    {
+        try (Closer closer = Closer.create()) {
+            createTestDirectoryStructure(closer, hierarchicalNamingConstraints);
+
+            ImmutableList.Builder<Location> rootAllBuilder = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0-file0"),
+                            createLocation("level0-file1"),
+                            createLocation("level0-file2"),
+                            createLocation("level0/level1-file0"),
+                            createLocation("level0/level1-file1"),
+                            createLocation("level0/level1-file2"),
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                rootAllBuilder.add(
+                        createLocation("level0"),
+                        createLocation("level0/level1"),
+                        createLocation("level0/level1/level2"));
+            }
+            ImmutableList<Location> rootAll = rootAllBuilder.build();
+
+            ImmutableList.Builder<Location> rootStartingFromLevel1File1Builder = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0/level1-file1"),
+                            createLocation("level0/level1-file2"),
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                rootStartingFromLevel1File1Builder.add(createLocation("level0/level1/level2"));
+            }
+            ImmutableList<Location> rootStartingFromLevel1File1 = rootStartingFromLevel1File1Builder.build();
+
+            ImmutableList.Builder<Location> level0AllBuilder = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0/level1-file0"),
+                            createLocation("level0/level1-file1"),
+                            createLocation("level0/level1-file2"),
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                level0AllBuilder.add(
+                        createLocation("level0/level1"),
+                        createLocation("level0/level1/level2"));
+            }
+            ImmutableList<Location> level0All = level0AllBuilder.build();
+
+            ImmutableList.Builder<Location> level0StartingFromLevel1File0zBuilder = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0/level1-file1"),
+                            createLocation("level0/level1-file2"),
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                level0StartingFromLevel1File0zBuilder.add(createLocation("level0/level1/level2"));
+            }
+            ImmutableList<Location> level0StartingFromLevel1File0z = level0StartingFromLevel1File0zBuilder.build();
+
+            ImmutableList<Location> level0StartingFromLevel1Level2File1 = ImmutableList.of(
+                    createLocation("level0/level1/level2-file1"),
+                    createLocation("level0/level1/level2-file2"));
+
+            ImmutableList.Builder<Location> level1AllBuilder = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                level1AllBuilder.add(createLocation("level0/level1/level2"));
+            }
+            ImmutableList<Location> level1All = level1AllBuilder.build();
+
+            ImmutableList.Builder<Location> level1StartingFromLevel2Builder = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                level1StartingFromLevel2Builder.add(createLocation("level0/level1/level2"));
+            }
+            ImmutableList<Location> level1StartingFromLevel2 = level1StartingFromLevel2Builder.build();
+
+            ImmutableList<Location> level1StartingFromLevel2File1 = ImmutableList.of(
+                    createLocation("level0/level1/level2-file1"),
+                    createLocation("level0/level1/level2-file2"));
+
+            assertThat(listPathStartingFrom("", ""))
+                    .containsExactlyInAnyOrderElementsOf(rootAll);
+            assertThat(listPathStartingFrom("", "level0/level1-file1"))
+                    .containsExactlyInAnyOrderElementsOf(rootStartingFromLevel1File1);
+            assertThat(listPathStartingFrom("", "zzz")).isEmpty();
+
+            assertThat(listPathStartingFrom("level0", ""))
+                    .containsExactlyInAnyOrderElementsOf(level0All);
+            assertThat(listPathStartingFrom("level0/", ""))
+                    .containsExactlyInAnyOrderElementsOf(level0All);
+            assertThat(listPathStartingFrom("level0", "level1"))
+                    .containsExactlyInAnyOrderElementsOf(level0All);
+            assertThat(listPathStartingFrom("level0", "level1-file0z"))
+                    .containsExactlyInAnyOrderElementsOf(level0StartingFromLevel1File0z);
+            assertThat(listPathStartingFrom("level0", "level1/level2-file1"))
+                    .containsExactlyInAnyOrderElementsOf(level0StartingFromLevel1Level2File1);
+            assertThat(listPathStartingFrom("level0/", "level1-file0z"))
+                    .containsExactlyInAnyOrderElementsOf(level0StartingFromLevel1File0z);
+            assertThat(listPathStartingFrom("level0", "zzz")).isEmpty();
+
+            assertThat(listPathStartingFrom("level0/level1/", ""))
+                    .containsExactlyInAnyOrderElementsOf(level1All);
+            assertThat(listPathStartingFrom("level0/level1", ""))
+                    .containsExactlyInAnyOrderElementsOf(level1All);
+            assertThat(listPathStartingFrom("level0/level1", "level2"))
+                    .containsExactlyInAnyOrderElementsOf(level1StartingFromLevel2);
+            assertThat(listPathStartingFrom("level0/level1", "level2-file1"))
+                    .containsExactlyInAnyOrderElementsOf(level1StartingFromLevel2File1);
+            assertThat(listPathStartingFrom("level0/level1", "zzz")).isEmpty();
+
+            assertThat(listPathStartingFrom("level0/level1/level2/", "")).isEmpty();
+            assertThat(listPathStartingFrom("level0/level1/level2", "")).isEmpty();
+            assertThat(listPathStartingFrom("level0/level1/level2", "zzz")).isEmpty();
+
+            assertThat(listPathStartingFrom("level0/level1/level2/level3", "")).isEmpty();
+            assertThat(listPathStartingFrom("level0/level1/level2/level3/", "")).isEmpty();
+
+            assertThat(listPathStartingFrom("unknown/", "")).isEmpty();
+
+            for (String directory : ImmutableList.of(
+                    "",
+                    "level0",
+                    "level0/level1",
+                    "level0/level1/level2",
+                    "level0/level1/level2/level3")) {
+                assertThat(listPathStartingFrom(directory, ""))
+                        .containsExactlyInAnyOrderElementsOf(listPath(directory));
+
+                if (!directory.isEmpty()) {
+                    assertThat(listPathStartingFrom(directory + "/", ""))
+                            .containsExactlyInAnyOrderElementsOf(listPath(directory + "/"));
+                }
+            }
+
+            if (isHierarchical()) {
+                assertThatThrownBy(() -> listPathStartingFrom("level0-file0", ""))
+                        .isInstanceOf(IOException.class)
+                        .hasMessageContaining(createLocation("level0-file0").toString());
+            }
+            else {
+                assertThat(listPathStartingFrom("level0-file0", "")).isEmpty();
+            }
+
+            if (!hierarchicalNamingConstraints && !normalizesListFilesResult()) {
+                assertThat(listPathStartingFrom("/", "")).isEmpty();
+            }
+
+            assertThatThrownBy(() -> listPathStartingFrom("", "é"))
+                    .isInstanceOf(IllegalArgumentException.class)
+                    .hasMessageContaining("ASCII");
+            assertThatThrownBy(() -> listPathStartingFrom("", "level0/é"))
+                    .isInstanceOf(IllegalArgumentException.class)
+                    .hasMessageContaining("ASCII");
+            assertThatThrownBy(() -> listPathStartingFrom("", "💡"))
+                    .isInstanceOf(IllegalArgumentException.class)
+                    .hasMessageContaining("ASCII");
         }
     }
 
@@ -1404,6 +1582,20 @@ public abstract class AbstractTestTrinoFileSystem
     {
         List<Location> locations = new ArrayList<>();
         FileIterator fileIterator = getFileSystem().listFiles(createListingLocation(path));
+        while (fileIterator.hasNext()) {
+            FileEntry fileEntry = fileIterator.next();
+            Location location = fileEntry.location();
+            assertThat(fileEntry.length()).isEqualTo(TEST_BLOB_CONTENT_PREFIX.length() + location.toString().length());
+            locations.add(location);
+        }
+        return locations;
+    }
+
+    private List<Location> listPathStartingFrom(String path, String startingFrom)
+            throws IOException
+    {
+        List<Location> locations = new ArrayList<>();
+        FileIterator fileIterator = getFileSystem().listFilesStartingFrom(createListingLocation(path), startingFrom);
         while (fileIterator.hasNext()) {
             FileEntry fileEntry = fileIterator.next();
             Location location = fileEntry.location();
--- a/lib/trino-filesystem/src/test/java/io/trino/filesystem/TestEmulatedListFilesStartingFromIterator.java
+++ b/lib/trino-filesystem/src/test/java/io/trino/filesystem/TestEmulatedListFilesStartingFromIterator.java
@@ -1,0 +1,261 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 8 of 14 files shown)


lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
```diff
@@ -307,21 +307,49 @@ public FileIterator listFiles(Location location)
...
             throw handleAzureException(e, "listing files", azureLocation);
         }
     }
 
-    private FileIterator listGen2Files(AzureLocation location)
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        requireNonNull(location, "location is null");
```
electrum: We generally don't use explicit null checks for method arguments, as the code using them will throw NPE with a useful stack trace

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +229,37 @@ void renameFile(Location source, Location target)
...
+     * ordering).
+     *
+     * @param location the directory to list
+     * @param startingFrom the inclusive lower bound for the path remainder after {@code location}
+     * @throws IllegalArgumentException if location is not valid for this file system
+     */
+    default FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        requireNonNull(location, "location is null");
```
electrum: Remove these explicit null checks

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -1413,6 +1438,31 @@ private List<Location> listPath(String path)
...
+        while (fileIterator.hasNext()) {
+            FileEntry fileEntry = fileIterator.next();
+            Location location = fileEntry.location();
+            assertThat(fileEntry.length()).isEqualTo(TEST_BLOB_CONTENT_PREFIX.length() + location.toString().length());
+            locations.add(location);
+        }
+        return locations;
+    }
+
+    private List<Location> listPathsStartingFromExpectation(Set<Location> locations, String path, String startingFrom)
```
electrum: Tests should be as simple and clear as possible, so that the expected behavior is obvious when reading the test. This type of utility method is complicated and would itself need tests to ensure correctness.

Instead, please hard-code the list of expected results in the test, so that it's easy to see how the method is expected to behave, similar to `testListDirectories`.

(I'd prefer if `testListFiles` was written more similar to `testListDirectories`)

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,91 @@
...
+        }
+
+        while (delegate.hasNext()) {
+            FileEntry entry = delegate.next();
+            String entryPath = entry.location().toString();
+            checkState(entryPath.startsWith(location), "Expected listed file to start with directory path '%s': %s", location, entryPath);
+
+            String entryTail = entryPath.substring(location.length());
+            if (Slices.utf8Slice(entryTail).compareTo(Slices.utf8Slice(startingFrom)) >= 0) {
+                nextEntry = entry;
```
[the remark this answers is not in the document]
adam-richardson-openai: Precomputing `Slices.utf8Slice(startingFrom)` seems like a valid improvement. I don't think that building a slice for each entry is a major performance issue, though

lib/trino-filesystem/src/test/java/io/trino/filesystem/TestEmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,174 @@
...
+import java.io.IOException;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Optional;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class TestEmulatedListFilesStartingFromIterator
```
electrum, line 27: Related to my comment in `AbstractTestTrinoFileSystem`, this test is great. It's easy to see the expected behavior and there are no concerns about the test code being correct.

lib/trino-filesystem/src/main/java/io/trino/filesystem/tracing/TracingFileSystem.java
```diff
@@ -119,6 +119,16 @@ public FileIterator listFiles(Location location)
         return withTracing(span, () -> delegate.listFiles(location));
     }
 
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        Span span = tracer.spanBuilder("FileSystem.listFilesStartingFrom")
+                .setAttribute(FileSystemAttributes.FILE_LOCATION, location.toString())
```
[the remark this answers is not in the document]
adam-richardson-openai: There are many existing examples of tracer spans that exclude method parameters in this class, so this feedback isn't obviously valid. My instinct is that this metadata is not useful enough to be worth introducing an additional public-facing attribute

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,91 @@
...
+            return;
+        }
+
+        while (delegate.hasNext()) {
+            FileEntry entry = delegate.next();
+            String entryPath = entry.location().toString();
+            checkState(entryPath.startsWith(location), "Expected listed file to start with directory path '%s': %s", location, entryPath);
+
+            String entryTail = entryPath.substring(location.length());
+            if (Slices.utf8Slice(entryTail).compareTo(Slices.utf8Slice(startingFrom)) >= 0) {
```
electrum: Nit: static import `utf8Slice`

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,91 @@
...
+
+import static com.google.common.base.Preconditions.checkState;
+import static java.util.Objects.requireNonNull;
+
+public final class EmulatedListFilesStartingFromIterator
+        implements FileIterator
+{
+    private final FileIterator delegate;
+    private final String location;
+    private final String startingFrom;
```
electrum, line 27: This could be `Slice` so that we only convert it once

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
```diff
@@ -307,21 +307,49 @@ public FileIterator listFiles(Location location)
...
+        requireNonNull(startingFrom, "startingFrom is null");
+
+        if (startingFrom.isEmpty()) {
+            return listFiles(location);
+        }
+
+        AzureLocation azureLocation = new AzureLocation(location);
+
+        try {
+            if (!isHierarchicalNamespaceEnabled(azureLocation)) {
```
electrum: Write this using a ternary so it looks the same as in `listFiles`

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystem.java
```diff
@@ -340,9 +369,15 @@ private FileIterator listGen2Files(AzureLocation location)
                         .iterator());
     }
 
-    private FileIterator listBlobFiles(AzureLocation location)
+    private FileIterator listBlobFiles(AzureLocation location, Optional<String> startingFrom)
     {
-        PagedIterable<BlobItem> blobItems = createBlobContainerClient(location, Optional.empty()).listBlobs(new ListBlobsOptions().setPrefix(location.directoryPath()), null);
+        String directoryPath = location.directoryPath();
+        ListBlobsOptions options = new ListBlobsOptions().setPrefix(directoryPath);
+        if (startingFrom.isPresent()) {
```
electrum: Can this use `startingFrom.ifPresent()`?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
```diff
@@ -223,6 +224,27 @@ public FileIterator listFiles(Location location)
         }
     }
 
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        requireNonNull(location, "location is null");
```
electrum: Same comment

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
```diff
@@ -223,6 +224,27 @@ public FileIterator listFiles(Location location)
...
+        requireNonNull(startingFrom, "startingFrom is null");
+
+        if (startingFrom.isEmpty()) {
+            return listFiles(location);
+        }
+
+        GcsLocation gcsLocation = new GcsLocation(normalizeToDirectory(location));
+
+        try {
+            return new GcsFileIterator(gcsLocation, getPage(gcsLocation, startOffset(gcsLocation.path() + startingFrom)));
```
electrum, line 239: Is this correct if the path doesn't end with slash?

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -240,37 +241,84 @@ public FileIterator listFiles(Location location)
         return listObjects(location, false);
     }
 
-    private FileIterator listObjects(Location location, boolean includeDirectoryObjects)
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
             throws IOException
     {
-        S3Location s3Location = new S3Location(location);
+        requireNonNull(startingFrom, "startingFrom is null");
```
electrum: Same comment

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -240,37 +241,84 @@ public FileIterator listFiles(Location location)
...
-                // Restore status will not be added to the response without requested
-                .optionalObjectAttributes(OptionalObjectAttributes.RESTORE_STATUS)
-                .requestPayer(requestPayer)
-                .bucket(s3Location.bucket())
-                .prefix(key)
-                .build();
+        return listObjects(location, false, Optional.of(startingFrom));
+    }
+
+    private FileIterator listObjects(Location location, boolean includeDirectoryObjects)
```
electrum: This method is only called in a couple of places, so we could inline it (remove it and have the callers pass `Optional.empty()` for the `startingFrom` argument)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -240,37 +241,84 @@ public FileIterator listFiles(Location location)
...
+    {
+        if (!key.isEmpty() && !key.endsWith("/")) {
+            return key + "/";
+        }
+        return key;
+    }
+
+    private static String truncateLastCodePoint(String value)
+    {
+        requireNonNull(value, "value is null");
```
electrum: Same comment about null check

electrum:
This API change looks good

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
adam-richardson-openai: I believe this is correct. As a function of `normalizeToDirectory` a few lines up, `gcsLocation.path()` either ends with a slash or is the empty string. Either way, concatenating `gcsLocation.path() + startingFrom` gives a valid cursor

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
adam-richardson-openai: Yes! Codex had the same comment, and I was fixing it concurrently with your review

adam-richardson-openai:
@electrum Thank you for the review! I've taken action on all your feedback

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,85 @@
...
+    private FileEntry nextEntry;
+
+    public EmulatedListFilesStartingFromIterator(FileIterator delegate, Location location, String startingFrom)
+    {
+        this.delegate = delegate;
+
+        String locationPath = location.path();
+        this.locationPath = (locationPath.isEmpty() || locationPath.endsWith("/")) ? locationPath : locationPath + "/";
+
+        this.startingFrom = utf8Slice(startingFrom);
```
wendigo: There is no need for Slice here

wendigo (approved):
One minor change

wendigo:
@electrum do you want to take a look?

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
adam-richardson-openai: Okay! We debated this internally quite a bit when I shipped this to our fork (I also wasn't sure if we needed it). I left it in out of caution, but I trust your judgment that it's unnecessary

ebyhr:
/test-with-secrets sha=42c5611279398a3d6a6e156a0d61b5c779d59a5d

adam-richardson-openai:
Looks like there are test failures on four `TestAzureFileSystemOAuthGen2Hierarchical*` tests. I will investigate and fix (at which point I'll need assistance to re-run the secrets-related tests)

adam-richardson-openai:
I have found the likely bug and committed a prospective fix. I believe the issue is that Azure has a confusing and arguably ill-defined behavior for the ordering semantics of `ListPathsOptions#listPaths`, and therefore for the behavior of `startFrom`

Specifically, `listPaths` does _not_ return objects and directories in lexicographical order, but rather in "depth-first" order (nested directories before objects). As far as I can tell, this is only documented in a random migration guide: https://learn.microsoft.com/en-ca/azure/storage/blobs/upgrade-to-data-lake-storage-gen2#list-operations. It's not documented AFAICT in any of the API or SDK docs

I took a deep look at the docs for the underlying REST API: https://learn.microsoft.com/en-us/rest/api/storageservices/datalakestoragegen2/path/list?view=rest-storageservices-datalakestoragegen2-2019-12-12. These do not mention any specific ordering for the results. The docs for the `beginFrom` parameter state
>A relative path within the specified directory where the listing will start from. For example, a recursive listing under directory folder1/folder2 with beginFrom as folder3/readmefile.txt will start listing from folder1/folder2/folder3/readmefile.txt

However, these docs don't clarify the ordering of the results, or the sequence against which `beginFrom` is evaluated

Even after all of this research, it's genuinely unclear to me what `startFrom/beginFrom` actually means in the general case. However, I believe that it's likely safe in practice to pass a `startFrom` that does not contain a slash, simply because there's only one plausible, sane interpretation (return objects and directories `>= startFrom`), whereas the semantics for `startFrom` w/ slashes are more ambiguous and confusing

On that basis, I tweaked the code to truncate the provided `startingAfter` before the first slash, if any, and additionally to filter the result set on the client side to achieve the desired semantics of `listFilesStartingAfter`

I will need a maintainer's assistance to re-run the secrets tests with the prospective fix. cc @wendigo @ebyhr

adam-richardson-openai:
I believe the CI failure on `ci / test (plugin/trino-delta-lake)` is likely unrelated to my changes

wendigo:
```
Error:  Failures: 
Error:    TestAzureFileSystemGen2Hierarchical>AbstractTestTrinoFileSystem.testListFilesStartingFrom:1043->AbstractTestTrinoFileSystem.testListFilesStartingFrom:1252 
Expecting actual:
  [abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2]
to contain exactly in any order:
  [abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]
but could not find the following elements:
  [abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-601dd857-520a-41c3-8457-12ef15b4e62f@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]

Error:    TestAzureFileSystemGen2HierarchicalWithEncryption>AbstractTestTrinoFileSystem.testListFilesStartingFrom:1043->AbstractTestTrinoFileSystem.testListFilesStartingFrom:1252 
Expecting actual:
  [abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2]
to contain exactly in any order:
  [abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]
but could not find the following elements:
  [abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-083bc35e-e12d-4785-8ca1-acba99e0837a@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]

Error:    TestAzureFileSystemOAuthGen2Hierarchical>AbstractTestTrinoFileSystem.testListFilesStartingFrom:1043->AbstractTestTrinoFileSystem.testListFilesStartingFrom:1252 
Expecting actual:
  [abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2]
to contain exactly in any order:
  [abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]
but could not find the following elements:
  [abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-a4529ebb-f47d-43e8-aa0d-471161fce027@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]

Error:    TestAzureFileSystemOAuthGen2HierarchicalWithEncryption>AbstractTestTrinoFileSystem.testListFilesStartingFrom:1043->AbstractTestTrinoFileSystem.testListFilesStartingFrom:1252 
Expecting actual:
  [abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2]
to contain exactly in any order:
  [abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1-file1,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1-file2,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]
but could not find the following elements:
  [abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file0,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file1,
    abfs://test-hierarchical-1ab5d3cc-c1f4-4454-b1a5-80c304d1619e@devcicdhierarchical.dfs.core.windows.net/level0/level1/level2-file2]

[INFO] 
Error:  Tests run: 216, Failures: 4, Errors: 0, Skipped: 14
```

adam-richardson-openai:
What a pain!

I'm thinking I'll set up an Azure storage account on my side so that I can just run the relevant tests locally and iterate on a prospective fix faster w/o needing help from maintainers

If we can't find a confident fix for ADLS gen2, maybe we'll have no choice but to simply use the inefficient emulated implementation for that platform

Edit: I'm able to repro locally, so we can hopefully get clarity on the fix, or the lack thereof, much faster from here.

adam-richardson-openai:
After some additional iteration locally, I've found a fix for the Azure issues. My [previous summary](https://github.com/trinodb/trino/pull/28549#issuecomment-4015750048) of the issue was accurate, but my attempted fix was not. Rather than simply truncating `startFrom` before the first slash, it's necessary to truncate `startFrom` before the first character that is lexicographically smaller than `/`. We also need to preserve the additional client-side filtering, since it's a conservative bound. This is a bit ugly, honestly, but I'm confident it does work, and it should accelerate many practical queries (since `/` is small)

If this feels too hacky, we can rip it out and just use the emulated implementation instead

`TestAzureFileSystemGen2Hierarchical#testListFilesStartingFrom` and `TestAzureFileSystemGen2HierarchicalWithEncryption#testListFilesStartingFrom` now pass locally on my machine. I have not run the oauth tests (they require addtl. setup) but I'm confident they'll pass given that they previously exhibited identical failures

@wendigo Please take another look when you have the time

wendigo:
@adam-richardson-openai azure is passing now

adam-richardson-openai:
I am reasonably sure that the failure on `ci / test (plugin/trino-iceberg) (pull_request)` is an infrastructure flake unrelated to my changes

adam-richardson-openai:
@wendigo @electrum I think this PR is in a shippable state. Do you think it's good to go, or does it need another round of review? Thanks!!

findepi:
I wasn't aware of this PR and reviewed https://github.com/trinodb/trino/pull/28581 instead.

assuming it's the same code, @adam-richardson-openai can you please see my review comments?
- https://github.com/trinodb/trino/pull/28581#pullrequestreview-3920513358

adam-richardson-openai:
Yes, same code!! I think @wendigo cut that PR to compose and validate a small direct fix to my code

wendigo:
@findepi test with secrets still doesn't work for me :) I've closed other PR that I've used to validate azure code

adam-richardson-openai:
>This is great for paging outside of FS layer. For paging to work, we might want to support limit here too.
>What is the intended use-case for this method?

Please see https://github.com/trinodb/trino/pull/28381/changes#diff-deb3ec23c1e161eb8846cf12ce1c45586dc1c2dc839804c7d420e2cdf98f3069R334-R343 and https://github.com/trinodb/trino/pull/28381/. The acute motivation here is to accelerate scanning of Delta Lake table logs by starting from the `_last_checkpoint` rather than scanning the whole directory. For a large, representative table in our data warehouse, this reduces the scan time from about 45s to about 250ms

More broadly, the intent of the method is to facilitate skipping irrelevant data while listing a directory

This new method isn't really intended to support pagination, which is why it doesn't have a limit clause today. I actually think that extending this method to support pagination or a limit clause is pretty hard, because there's no well-defined order that is universal across all file systems (and even among blob storage platforms, there is variance -- see [my prev. comment](https://github.com/trinodb/trino/pull/28549#issuecomment-4015750048)). I think it could make sense to introduce a separate API for pagination too, but this is hard enough and different enough that I think it merits a separate method and a separate PR

Please see https://github.com/trinodb/trino/pull/28381#discussion_r2848851393 for the specific thread where we tentatively aligned on introducing `listFilesStartingFrom`

findepi:
> @findepi test with secrets still doesn't work for me :) I've closed other PR that I've used to validate azure code

😞  for the first, ❤️  for the second

findepi:
> The acute motivation here is to accelerate scanning of Delta Lake table logs by starting from the `_last_checkpoint` rather than scanning the whole directory.

I thought that the connector doesn't list at all, and probes for those commit files after last checkpoint.
Typically there shouldn't be more than ~10.

I understand that there is a PR change which introduces file listing (and then the change makes sense). 
Is it that the file listing is indeed beneficial over probing technique?

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -1151,6 +1158,142 @@ protected void testListFiles(boolean hierarchicalNamingConstraints)
...
+            }
+
+            assertThat(listPathStartingFrom("", ""))
+                    .containsExactlyInAnyOrderElementsOf(rootAll.build());
+            assertThat(listPathStartingFrom("", "level0/level1-file1"))
+                    .containsExactlyInAnyOrderElementsOf(rootStartingFromLevel1File1.build());
+            assertThat(listPathStartingFrom("", "zzz")).isEmpty();
+
+            assertThat(listPathStartingFrom("level0", ""))
+                    .containsExactlyInAnyOrderElementsOf(level0All.build());
```
raunaqmorarka: We should build the list once rather than building it in every assertion.

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -1151,6 +1158,142 @@ protected void testListFiles(boolean hierarchicalNamingConstraints)
...
+            ImmutableList.Builder<Location> level1StartingFromLevel2 = ImmutableList.<Location>builder()
+                    .add(
+                            createLocation("level0/level1/level2-file0"),
+                            createLocation("level0/level1/level2-file1"),
+                            createLocation("level0/level1/level2-file2"));
+            if (!hierarchicalNamingConstraints) {
+                level1StartingFromLevel2.add(createLocation("level0/level1/level2"));
+            }
+
+            assertThat(listPathStartingFrom("", ""))
```
raunaqmorarka, line 1253: I think it would be also useful to assert for each level that the output of listPathStartingFrom with empty `startingFrom` String is matching the output of listFiles

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,82 @@
...
+        implements FileIterator
+{
+    private final FileIterator delegate;
+    private final String locationPath;
+    private final String startingFrom;
+    private FileEntry nextEntry;
+
+    public EmulatedListFilesStartingFromIterator(FileIterator delegate, Location location, String startingFrom)
+    {
+        this.delegate = delegate;
```
raunaqmorarka: requireNonNull on delegate and startingFrom, as they are not immediately used here

lib/trino-filesystem/src/main/java/io/trino/filesystem/tracing/TracingFileSystem.java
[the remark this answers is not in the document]
raunaqmorarka: I think it would be good idea to add this, so that we know what was called when we're trying to understand FS latencies from traces.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -237,40 +238,78 @@ public void renameFile(Location source, Location target)
...
-                // Restore status will not be added to the response without requested
-                .optionalObjectAttributes(OptionalObjectAttributes.RESTORE_STATUS)
-                .requestPayer(requestPayer)
-                .bucket(s3Location.bucket())
-                .prefix(key)
-                .build();
+        return listObjects(location, false, Optional.of(startingFrom));
+    }
+
+    private FileIterator listObjects(Location location, boolean includeDirectoryObjects, Optional<String> startingFrom)
```
raunaqmorarka: Please move the private methods below the public ones, and static methods below the non-static ones

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -237,40 +238,78 @@ public void renameFile(Location source, Location target)
...
+            return key + "/";
+        }
+        return key;
+    }
+
+    private static String truncateLastCodePoint(String value)
+    {
+        verify(!value.isEmpty(), "value is empty");
+        return value.substring(0, value.offsetByCodePoints(value.length(), -1));
+    }
```
raunaqmorarka: How about something like
```
private static String getImmediatePredecessor(String value) 
{
    verify(!value.isEmpty(), "value is empty");
    
    int lastCodePointOffset = value.offsetByCodePoints(value.length(), -1);
    int lastCodePoint = value.codePointAt(lastCodePointOffset);
    String prefix = value.substring(0, lastCodePointOffset);

    if (lastCodePoint > 0) {
        // Decrement the last character: "Apple" -> "Appld"
        // This is much "closer" than "Appl"
        return prefix + new String(Character.toChars(lastCodePoint - 1));
    }
    // If the last character is U+0000, the prefix is indeed the closest
    return prefix;
}
```
This should find a "closer" predecessor than simple prefix

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -237,40 +238,78 @@ public void renameFile(Location source, Location target)
...
+
+            Stream<S3Object> s3ObjectStream = client.listObjectsV2Paginator(request.build()).contents().stream();
             if (!includeDirectoryObjects) {
                 s3ObjectStream = s3ObjectStream.filter(object -> !object.key().endsWith("/"));
             }
-            return new S3FileIterator(s3Location, s3ObjectStream.iterator());
+
+            S3FileIterator iterator = new S3FileIterator(s3Location, s3ObjectStream.iterator());
+            if (startingFrom.isPresent()) {
+                return new EmulatedListFilesStartingFromIterator(iterator, s3Location.location(), startingFrom.orElseThrow());
```
raunaqmorarka: Would we be able to avoid use of `EmulatedListFilesStartingFromIterator` if we used `getImmediatePredecessor` as mentioned below ?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
```diff
@@ -223,6 +224,24 @@ public FileIterator listFiles(Location location)
...
+    @Override
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
+        if (startingFrom.isEmpty()) {
+            return listFiles(location);
+        }
+
+        GcsLocation gcsLocation = new GcsLocation(normalizeToDirectory(location));
+
```
raunaqmorarka: skip new line

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -237,40 +238,78 @@ public void renameFile(Location source, Location target)
...
+
+    private static String directoryKey(String key)
+    {
+        if (!key.isEmpty() && !key.endsWith("/")) {
+            return key + "/";
+        }
+        return key;
+    }
+
+    private static String truncateLastCodePoint(String value)
```
raunaqmorarka: Do we have an existing test where there are file names in the filesystem between the prefix value and actual startingFrom value ?

adam-richardson-openai:
@findepi

>I thought that the connector doesn't list at all, and probes for those commit files after last checkpoint.
Typically there shouldn't be more than ~10.
>I understand that there is a PR change which introduces file listing (and then the change makes sense).
Is it that the file listing is indeed beneficial over probing technique?

There is a bit of nuance here. Let me explain it as clearly as possible (but do see https://github.com/trinodb/trino/pull/28381 for the full pictures and for the code change in context)

* At analysis time, the Delta Lake connector needs to load the Delta table metadata (including schema and partition columns) and the version protocol information
* Prior to https://github.com/trinodb/trino/pull/28381, this requires downloading, parsing, and processing the latest checkpoint and all intervening commits, since the the metadata and protocol are just stored as entries in the commit log
* For very large tables, this may involve processing 10s of gigabytes of Parquet, which is extremely slow and resource-intensive. For one of our large internal tables, analysis takes 3-4 minutes
* We can instead parse the Delta table metadata and protocol information from the version checksum file, an additional JSON file in the _delta_log ([ref.](https://github.com/delta-io/delta/blob/488c916931ca9d210f4cadd2d5520e0274d26b04/PROTOCOL.md#version-checksum-file)). This is a single point lookup on a tiny JSON blob, so it's very fast
* However, to find the latest commit and the associated checksum file in the first place, we still need to list the files in the Delta log, which is quite slow for very large _delta_log directories

For our large internal test table, we see:
* 3-4min analysis time for the old Delta log based approach
* 45s analysis time for https://github.com/trinodb/trino/pull/28381 with `listFiles` rather than `listFilesStartingFrom`
* 250ms analysis time for https://github.com/trinodb/trino/pull/28381 using `listFilesStartingFrom` (starting from the latest checkpoint)

This motivates the introduction of the `listFilesStartingFrom`

@wendigo and @findinpath have broader context on both changes here

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +227,35 @@ void renameFile(Location source, Location target)
...
+     * and are lexicographically sorted (except for local HDFS which has the system-dependent
+     * ordering).
+     *
+     * @param location the directory to list
+     * @param startingFrom the inclusive lower bound for the path remainder after {@code location}
+     * @throws IllegalArgumentException if location is not valid for this file system
+     */
+    default FileIterator listFilesStartingFrom(Location location, String startingFrom)
+            throws IOException
+    {
```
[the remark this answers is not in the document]
raunaqmorarka: +1

wendigo:
@findepi @electrum agreed as well that this API is useful to have. Actually I have another use case for it - removing expires segments in the spooling protocol.

adam-richardson-openai:
@findepi Addressing your other comment from the previous PR

>It's important to clarify the / behavior (must, should, may [not])
>```
>// In both cases: can `start` start with a slash?
>// In both cases: can `start` not start with a slash?
>listFilesStartingFrom(Location.of("file:///tmp"), start);
>listFilesStartingFrom(Location.of("file:///tmp/"), start);
>```

The contract of `listFilesStartingFrom` states:
>Lists files within the specified directory recursively, including only files where the remainder of the path after the location is lexicographically greater than or equal to startingFrom

and all implementations comply with this property. There are no additional restrictions on `startingFrom`, so it's permitted and valid to pass a value of `startingFrom` that begins with a slash (or that contains a slash)

However, passing a value of `startingFrom` that begins with a slash would be an unusual thing to do. Typically `startingFrom` would be used to skip irrelevant files or nested directories within `location`. So, `startingFrom` would most often be a prefix of some nested file or directory name within `location`, and it would be odd to (e.g.) have a file or directory name starting with a slash

Given that leading slashes are technically valid, but weird, do you have a suggested change in the javadoc comment? It feels a bit hard to document this behavior clearly without unintentionally encouraging users to do the wrong thing

We could also:
* Ban leading slashes
* Strip leading slashes

since there isn't really a great use cases for them. But overall I'm biased towards keeping the implementation simple

(A third option would be to change the semantics of `listFilesStartingFrom`. I feel the current semantics are best because they're the closest fit to the native semantics of modern blob stores)

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
adam-richardson-openai: As a non-expert in Unicode encodings, the complexity here makes me a bit nervous particularly given the differences in encoding between UTF-16 and UTF-8. This looks like it should work, but I'm not 100% confident. Additionally, I think the difference in performance here should be small for the practical use cases of this method. On that basis, I have a preference to stick with the simpler implementation in the current version of the PR

If we're certain that this approach is correct and we'd really prefer to include it in this PR, I'm okay to pull it in. Please let me know

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
adam-richardson-openai: Probably, but see https://github.com/trinodb/trino/pull/28549#discussion_r2914407173

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -1151,6 +1158,142 @@ protected void testListFiles(boolean hierarchicalNamingConstraints)
...
+                    .containsExactlyInAnyOrderElementsOf(rootAll.build());
```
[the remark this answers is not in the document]
adam-richardson-openai: * `testListFiles` also uses `containsExactlyInAnyOrderElementsOf`, so there is precedent for the pattern
* Distinguishing ordered vs. non-ordered file systems would require additional methods or fragile branching based on classes, which doesn't seem like a worthwhile improvement

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: I realized that giving a guarantee is hard here, so lets keep this

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
adam-richardson-openai: Yes, this is exercised by the existing `AbstractTestTrinoFileSystem`. E.g. for `listPathStartingFrom("level0", "level1/level2-file1")`, we truncate to `level1/level2-file` for the purposes of the S3 call, then apply client-side filtering via the emulated iterator to discard the unwanted `level0/level1/level2-file0` entry

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: I did some more digging and came up with
```
private static String getPredecessor(String value) 
{
    verify(!value.isEmpty(), "value is empty");

    int lastOffset = value.offsetByCodePoints(value.length(), -1);
    int lastCodePoint = value.codePointAt(lastOffset);
    String prefix = value.substring(0, lastOffset);

    // If the last character is U+0000, the prefix is the closest predecessor
    if (lastCodePoint == 0) {
        return prefix;
    }

    // Decrement the last character: "Apple" -> "Appld"
    // This is much "closer" than "Appl"
    int decrementedCodePoint = lastCodePoint - 1;

    // If we decremented U+E000, we hit the top of the surrogate gap.
    // Jump over the entire gap down to U+D7FF
    if (decrementedCodePoint == Character.MAX_SURROGATE) {
        decrementedCodePoint = Character.MIN_SURROGATE - 1; 
    }

    return prefix + new String(Character.toChars(decrementedCodePoint));
}
```
This is not the exactly previous value, but closer than just the prefix.
@findepi you implemented something similar at `io.trino.operator.scalar.CharacterStringCasts#varcharToCharSaturatedFloorCast`, does the above look reasonable to you ?

lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsFileSystem.java
```diff
@@ -360,6 +363,28 @@ private Optional<UriLocation> preSignedUri(Location location, Duration ttl, Opti
         }
     }
 
+    private Map<String, List<String>> toMultiMap(Map<String, String> extHeaders)
```
raunaqmorarka: static

adam-richardson-openai: This is an unintentional change unrelated to the PR. Backing it out now. One moment

adam-richardson-openai: Removed. Thanks!

adam-richardson-openai:
Pushed to address all uncontroversial/actionable feedback. Remaining open threads:
* Discussion on more precise "predecessor" implementation for S3 (vs. truncating the last code point): https://github.com/trinodb/trino/pull/28549#discussion_r2914100264 @raunaqmorarka
* Discussion on potential javadoc changes on leading slashes: https://github.com/trinodb/trino/pull/28549#issuecomment-4034330758 @findepi

cc @wendigo

raunaqmorarka:
> Prior to https://github.com/trinodb/trino/pull/28381, this requires downloading, parsing, and processing the latest checkpoint and all intervening commits, since the the metadata and protocol are just stored as entries in the commit log
For very large tables, this may involve processing 10s of gigabytes of Parquet, which is extremely slow and resource-intensive. For one of our large internal tables, analysis takes 3-4 minutes

@adam-richardson-openai could you help me to understand this part better ?
I think the existing code first tries to find metadata/protocol entries in the json files. If we find metadata and protocol entries here, then we should terminate early.
If we read the parquet checkpoint files, then we should be reading specifically metadata and protocol columns from the parquet files, and these should be mostly `NULL` which should take minimal space to store and decode quickly. 
So I'm not following why the existing implementation would involve multi GB reads from parquet only for metadata and protocol entries.

adam-richardson-openai:
>I think the existing code first tries to find metadata/protocol entries in the json files

This is true

>If we find metadata and protocol entries here, then we should terminate early.

This is a rare case. The only situation where we'll find the metadata and protocol entries in the recent commits is if the metadata and protocol both changed very recently. The default checkpoint interval is 36 commits, whereas it'd be totally normal to have hundreds or thousands of commits between metadata and protocol changes. So, typically the metadata and protocol entries can be found only in the checkpoint Parquet, which is expensive to load

Even if we happen to find metadata and protocol entries in the recent commit JSON, these also can be quite large, and are downloaded and processed sequentially, so this is also slower and more expensive than necessary

>If we read the parquet checkpoint files, then we should be reading specifically metadata and protocol columns from the parquet files, and these should be mostly NULL which should take minimal space to store and decode quickly

Empirically, loading and parsing the transaction logs and checkpoints is extremely slow for our largest tables, resulting in poor query performance and user experience. I don't have an exact accounting of where all the time is spent or exactly how much IO we're doing, etc. but it's very clear from rudimentary benchmarking that overall performance is poor. Likewise, simple benchmarks show that the checksum file based approach is dramatically more performant than the Delta log based approach

As an additional data point, we have examples from other open source projects (e.g. Delta Kernel from Databricks) and also anecdotes from closed-source data processing systems that these techniques for accelerating reading of Delta tables are commonly used to great effect

It may ultimately be possible to improve the performance of the existing transaction log based code (in fact, I have additional patches in that space, e.g. https://github.com/trinodb/trino/pull/28470). However, I think that making full use of the available metadata files is our best strategy to achieve optimal performance for Delta Lake tables on Trino

raunaqmorarka:
I understand that the checksum file based approach is better and we should go ahead with that. I'm trying to understand why the current approach would result in multi GB parquet scans as you mentioned before. To me that indicates something is not working as intended, and that's worth figuring out separately as well since checksum files are not guaranteed to be there.

adam-richardson-openai:
>To me that indicates something is not working as intended, and that's worth figuring out separately as well since checksum files are not guaranteed to be there.

Yeah, that's a fair question! I don't personally have an exact mechanical understanding why the performance is so poor. One specific hypothesis: processing files from the _delta_log is intrinsically serial in Trino's current implementation (both within files and across files), so despite all of the various optimizations to accelerate reading individual files, it's hard to achieve low latency overall on very large _delta_log directories

>I'm trying to understand why the current approach would result in multi GB parquet scans as you mentioned before

My wording was a bit imprecise in my previous message. These statements are definitely true:
* Our largest Delta tables have many gigabytes of checkpoint Parquet
* In the absence of the checksum parsing fix, Trino needs to parse the metadata and protocol information out of the checkpoint Parquet in most cases
* Analysis is empirically extremely slow for such tables

On the other hand, I do not know for certain that we're actually downloading and parsing every Parquet file in full. I agree that optimizations like projection and filter pushdown on the checkpoint Parquet should reduce the actual data scanned a lot. Nonetheless, perf is poor overall in our experiments

>checksum files are not guaranteed to be there

This is true, and I'm not disputing your point here more broadly. But just to document the status quo:
* Spark w/ recent versions of the Delta connector does write checksum files ([ref.](https://delta.io/blog/delta-lake-3-3/))
* All recent Databricks DBR versions write checksum files
* Trino does not yet write checksum files, but there's interest in implementing this soon ([ref.](https://github.com/trinodb/trino/issues/28386)). (You may see a patch from either Marius, myself, or one of my teammates here soon!)

My general thought here is that a larger and larger fraction of tables will have full support for Delta Lake version checksums over time

electrum:
For analyzing why planning is slow, you can use [OpenTelemetry](https://trino.io/docs/current/admin/opentelemetry.html) tracing. It will tell you exactly which operations are happening, what files are being read along with the read size, etc. You shouldn't need to guess or speculate.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: @adam-richardson-openai you can go ahead with the above version

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +227,39 @@ void renameFile(Location source, Location target)
     FileIterator listFiles(Location location)
             throws IOException;
 
+    /**
+     * Lists files within the specified directory recursively, including only files where the
+     * remainder of the path after the location is lexicographically greater than or equal to
+     * {@code startingFrom}. The location can be empty, listing all files in the file system where the
```
findepi: "lexicographically greater" is likely generally understood term until codepoints outside of BMP are concerned.

do we mean "lexicographically greater" when comparing by codepoints or comparing by Java chars?

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +227,39 @@ void renameFile(Location source, Location target)
...
+     * path is greater than or equal to {@code startingFrom}. The location is always interpreted as a
+     * directory, whether or not it ends with a slash. If the location does not exist, an empty
+     * iterator is returned. {@code startingFrom} may be empty, in which case the behavior is identical to
+     * {@link #listFiles(Location)}.
+     * <p>
+     * For hierarchical file systems, if the path is not a directory, an exception is
+     * raised.
+     * For hierarchical file systems, if the path does not reference an existing
+     * directory, an empty iterator is returned. For blob file systems, all blobs
```
findepi: This looks like contradicting.
"For hierarchical file systems, if the path is not a directory" then "exception is raised" and "empty iterator is returned". Am i reading this wrong?

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +227,39 @@ void renameFile(Location source, Location target)
...
+     * that start with the location are listed, subject to the startingFrom constraint.
+     * <p>
+     * The returned FileEntry locations will start with the specified location exactly
+     * and are lexicographically sorted (except for local HDFS which has the system-dependent
```
findepi: grammar: "will start ... are"  -- should this use present simple in both places?

findepi: so the ordering is not guaranteed after all?
If it is not guaranteed, it's better to remove it from the javadoc.

Also, the method name kinda suggest results are ordered, so maybe it could be renamed e.g. `listFilesAfter` or `listFilesLexicographicallyGreaterOrEqual` (this is too long, I know)

cc @electrum

findepi:
> The contract of `listFilesStartingFrom` states:
> 
> > Lists files within the specified directory recursively, including only files where the remainder of the path after the location is lexicographically greater than or equal to startingFrom

This seems to imply that `listFilesStartingFrom(Location.of("file:///tmp"), "/bar");` is a valid call and returns `file:///tmp/bar`, `file:///tmp/foo`, but not `file:///tmp/abc`

This also seems to imply that `listFilesStartingFrom(Location.of("file:///tmp"), "bar");` should return empty iterator.
(`"/bar" < "bar"`;  `"/foo" < "bar"`; `"/abc" < "bar"`).



> However, passing a value of `startingFrom` that begins with a slash would be an unusual thing to do.

This says my understanding of what the implementation does does not match my understanding of the words "including only files where the remainder of the path after the location is lexicographically greater than or equal to startingFrom".

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
```diff
@@ -0,0 +1,83 @@
...
+            return;
+        }
+
+        while (delegate.hasNext()) {
+            FileEntry entry = delegate.next();
+            String entryPath = entry.location().path();
+            checkState(entryPath.startsWith(locationPath), "Expected listed file to start with directory path '%s': %s", locationPath, entry.location());
+
+            String entryTail = entryPath.substring(locationPath.length());
+            if (entryTail.compareTo(startingFrom) >= 0) {
```
raunaqmorarka, line 75: fyi @findepi

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
```diff
@@ -1151,6 +1158,167 @@ protected void testListFiles(boolean hierarchicalNamingConstraints)
...
+            assertThat(listPathStartingFrom("level0", "level1/level2-file1"))
+                    .containsExactlyInAnyOrderElementsOf(level0StartingFromLevel1Level2File1);
+            assertThat(listPathStartingFrom("level0/", "level1-file0z"))
+                    .containsExactlyInAnyOrderElementsOf(level0StartingFromLevel1File0z);
+            assertThat(listPathStartingFrom("level0", "zzz")).isEmpty();
+
+            assertThat(listPathStartingFrom("level0/level1/", ""))
+                    .containsExactlyInAnyOrderElementsOf(level1All);
+            assertThat(listPathStartingFrom("level0/level1", ""))
+                    .containsExactlyInAnyOrderElementsOf(level1All);
```
findepi, line 1276: Cool, so the trailing slash in the `location` doesn't matter. 
This is good, but probably needs to be reflected in javadoc.

see also https://github.com/trinodb/trino/pull/28549#issuecomment-4057675806

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
findepi: depicting the difference :


```
jshell> String A = "\uD800\uDC00";  // U+10000 (LINEAR B SYLLABLE B008 A) - surrogate pair
   ...>   String B = "\uE000";         // U+E000 (private use character) - single char
A ==> "𐀀"
B ==> ""

jshell> System.out.println(A.compareTo(B));
-2048

jshell> System.out.println(((Integer)A.codePointAt(0)).compareTo(B.codePointAt(0)));
1
```

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
findepi: This does lexicographical comparison by Java chars. 
If we delegate to an API such as S3, it may do something else

let's get the answer on desired semantics, see -- https://github.com/trinodb/trino/pull/28549#discussion_r2933427536

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
adam-richardson-openai: The Javadoc does state:
```
The location is always interpreted as a directory, whether or not it ends with a slash
```
That remark was intended to address this behavior (that `location` with or without a trailing slash has the same result). But I see your point that it could be clearer or more specific. I can amend the Javadoc to clarify that we always treat `location` as if it does have a trailing slash

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
adam-richardson-openai: The "lexicographically sorted" language (and also the "hierarchical file systems" points) were adapted from the Javadoc for `listFiles`. My rationale for including that language is these properties should hold equally well for `listFilesStartingFrom` as they do for `listFiles`, since the two methods have shared implementations or structurally-similar implementations

However, I generally agree that these remarks are confusing, and in some cases they may be ill-defined or even wrong. For instance, as documented in previous comments, ADLS Gen2 listings aren't lexicographically ordered, so the documented behavior of both `listFilesStartingFrom` (_and_ `listFiles`) is probably just incorrect

I think we can definitely remove the "lexicographically sorted" remark. We can likely simplify or clarify the "hierarchical file systems" points too. I will take a pass at this

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
adam-richardson-openai: Thanks for probing at this! I agree that the current semantics of `listFilesStartingFrom` are ambiguous, and I'd love to discuss both the ideal semantics and implementation strategies

adam-richardson-openai: Given that Trino is a Java project, I think that the most intuitive definition of "lexicographically greater than or equal to" is: `A.compareTo(B) >= 0`

One issue is that we're working with several different potential backend file system implementations with different semantics and varying levels of documentation. It's not trivial to map the desired semantics onto the backend while maintaining both correctness and high performance

In practice, I think the actual practical applications of this API will mostly use simple cursors -- cursors that solely use BMP characters, or in fact cursors that are simple ASCII strings. On that basis, I think there are a few simplifications we can make here:
1. Only use the "accelerated" implementations if `startingFrom` is a simple ASCII string. Use the emulated version otherwise
2. Explicitly ban/reject non-ASCII `startingFrom` in `listFilesStartingFrom`
3. Declare that behavior is implementation-defined for non-ASCII `startingFrom`

(ordered by my preference.) In all cases, I think it should be easy to use the accelerated implementations for ASCII strings only, while being confident in both the performance and the correctness

What do you think of these proposed simplifications?

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
adam-richardson-openai: We can continue the extended discussion in https://github.com/trinodb/trino/pull/28549#discussion_r2933427536 -- but as a very brief summary, I suggest that we can treat Java string ordering as the true semantics. In this case, the challenge is how to map this back to the backend file system semantics while maintaining correctness and performance

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
adam-richardson-openai: I think we should close on this thread: https://github.com/trinodb/trino/pull/28549#discussion_r2934040512

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
findepi: > The location is always interpreted as a directory, whether or not it ends with a slash

this is obvious and, AFAICT, consistent with `listFiles` 👍 

however, the words `where the remainder of the path after the location`
or `@param startingFrom the inclusive lower bound for the path remainder after {@code location}`
seem refer to the location textual form as it is, so either ending in slash or not

findepi: > > The location is always interpreted as a directory, whether or not it ends with a slash
> 
> this is obvious and, AFAICT, consistent with `listFiles` 👍

Actually something is odd here:
- https://github.com/trinodb/trino/issues/28665

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
findepi: > ADLS Gen2 listings aren't lexicographically ordered, so the documented behavior of both `listFilesStartingFrom` (_and_ `listFiles`) is probably just incorrect

I did not know that! Given pre-existing `listFiles` is affected, would you mind fixing it (perhaps just fixing the javadoc), as a separate PR?

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
findepi: Perhaps stating the obvious. In this context it's worth noting s3 allows paths like `s3://bucket/double//slash`.
I suppose the `startingFrom` "." should include the `/slash` child path (as "." < "/") 
and  `startingFrom` "a" should not include it.

I.e. that only one / gets removed before making the comparison.

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
findepi: > Given that Trino is a Java project, I think that the most intuitive definition of "lexicographically greater than or equal to" is: `A.compareTo(B) >= 0`

There is no single obviously intuitive direction here.

In SQL, comparison is by codepoints.
In Iceberg, Parquet -- places where spec tries to abstract away from Java -- comparison is by codepoints.

To me, FS layer is this system boundary, when comparison by codepoints is intuitive (unless the system is also Java). Note that  comparison by codepoints is the same as comparing unsigned bytes of the UTF-8 representation.


> One issue is that we're working with several different potential backend file system implementations with different semantics and varying levels of documentation. It's not trivial to map the desired semantics onto the backend while maintaining both correctness and high performance

That's fair point and very important. For this API to be efficient, it needs to delegate to the remote system. To delegate correctly, it needs to understand the semantics it's supposed to deliver and whether the other system fulfills these semantics.

In case of multiple remote systems (as here), we want to provide as little guarantees to the caller, to leave maximum flexibility for the implementor.



> * Only use the "accelerated" implementations if `startingFrom` is a simple ASCII string. Use the emulated version otherwise

That makes sense to me. We perhaps can safely extend this to non-ASCII, but not the whole BMP.
The "compare by Java chars" vs "compare by codepoints" ambiguity starts at `java.lang.Character#MIN_SURROGATE` (U+D800).


We still should define the method semantics though.
IMO the method semantics should **not** be defined as "compare by Java chars". I.e. it should be illegal to provide `startingFrom` that includes U+D800 or higher OR the method should be defined "compare by codepoints". This is because I believe "compare by codepoints" is more likely to be delegate-able to a remote system.

The next question is, where the values for `startingFrom`  will be coming from.
If they will be calculated based on existing paths, it may be impossible to guarantee U+D800+ isn't there.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: In the interest of making progress, let's stick with the easy implementation for now.

raunaqmorarka:
@coderabbitai review

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
raunaqmorarka: We started this feature for listing contents of `_delta_log`, since the file naming there follows a predictable pattern we can assume that only the ASCII characterset is interesting for us.
So +1 to restricting the optimized implementation only to ascii `startingFrom` inputs.
The usage looks like
```
String startingFrom = format("%020d.", startVersion.get() + 1);  // startVersion is a long
files = fileSystem.listFilesStartingFrom(transactionLogLocation, startingFrom);
```
So this is guaranteed ASCII.
I think we can even assume that the entire output of listing `_delta_log` is ASCII. If it makes things simpler, we could just add a rider to the new API that it is safe to use only when the caller can guarantee that everything is ASCII.

wendigo:
@raunaqmorarka is this ready to be merged?

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
raunaqmorarka: Restricted `startingFrom` to ASCII. Under that restriction, UTF-8 byte, Java char, and codepoint orderings all agree on inclusion decisions, so the choice is no longer observable. Guard + tests added.

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
raunaqmorarka: Fixed. Removed the redundant sentence — the "location does not exist" case is already covered by the sentence at the top of the Javadoc.

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
raunaqmorarka: Fixed.

raunaqmorarka: Kept the name. The ordering caveat here mirrors `listFiles`; correcting it properly is out of scope for this PR. Added a note that result order may differ between implementations when non-ASCII filenames are present.

lib/trino-filesystem/src/main/java/io/trino/filesystem/EmulatedListFilesStartingFromIterator.java
raunaqmorarka: Closing: non-ASCII input is rejected at the API boundary, so the emulated iterator's Java-char comparison and the native S3/GCS UTF-8 cursor always agree on inclusion.

lib/trino-filesystem/src/test/java/io/trino/filesystem/AbstractTestTrinoFileSystem.java
raunaqmorarka: Javadoc now says `startingFrom` is compared against the portion of the path after the directory's (implicit or explicit) trailing slash. The double-slash case falls out: remainder of `double//slash` under location `double` is `/slash`, so `startingFrom="."` includes it and `startingFrom="a"` doesn't. Test added.

raunaqmorarka:
@findepi Addressing your leading-slash concern in https://github.com/trinodb/trino/pull/28549#issuecomment-4057675806: the Javadoc now specifies that `startingFrom` is compared against the portion of the path after the directory's (implicit or explicit) trailing slash. So `startingFrom="/bar"` on `file:///tmp` compares `/bar` against remainders like `foo`, `abc`, `bar` — all of which are greater than `/bar` since `/` is less than any printable filename start — effectively a match-all. Unit test `testListFilesStartingFromLeadingSlashInStartingFromMatchesAllFiles` pins this down.

On the CodeRabbit surrogate-pair flag on Azure: moot with the ASCII restriction — non-ASCII input, including supplementary characters, is rejected before reaching the native cursor.

raunaqmorarka:
@wendigo Pushed a round addressing the remaining feedback and tightening the contract to ASCII-only `startingFrom`. PTAL.

raunaqmorarka:
@coderabbitai review

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -237,38 +239,19 @@ public void renameFile(Location source, Location target)
     public FileIterator listFiles(Location location)
             throws IOException
     {
-        return listObjects(location, false);
+        return listObjects(location, false, Optional.empty());
```
electrum: See comment below
```java
return listObjects(location, false, "");

File: lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java [hunk 34]
```diff
@@ -237,38 +239,19 @@ public void renameFile(Location source, Location target)
...
+    public FileIterator listFilesStartingFrom(Location location, String startingFrom)
             throws IOException
     {
-        S3Location s3Location = new S3Location(location);
-
-        String key = s3Location.key();
-        if (!key.isEmpty() && !key.endsWith("/")) {
-            key += "/";
+        checkStartingFrom(startingFrom);
+        if (startingFrom.isEmpty()) {
```
electrum: Let's change `listObjects` to take a `String` for `startingFrom`, then we can simplify this to
```java
return listObjects(location, false, startingFrom);
```

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
```diff
@@ -388,6 +362,64 @@ public Optional<UriLocation> encryptedPreSignedUri(Location location, Duration t
...
 
+    private FileIterator listObjects(Location location, boolean includeDirectoryObjects, Optional<String> startingFrom)
+            throws IOException
+    {
+        S3Location s3Location = new S3Location(location);
+        String keyPrefix = directoryKey(s3Location.key());
+        ListObjectsV2Request.Builder request = listObjectsRequest(s3Location, keyPrefix);
+
+        try {
+            if (startingFrom.isPresent()) {
```
electrum: Change `startingFrom` to `String` and make this
```java
if (!startingFrom.isEmpty()) {

File: lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java [hunk 36]
```diff
@@ -388,6 +362,64 @@ public Optional<UriLocation> encryptedPreSignedUri(Location location, Duration t
...
+        catch (SdkException e) {
+            throw new TrinoFileSystemException("Failed to list location: " + location, e);
+        }
+    }
+
+    private ListObjectsV2Request.Builder listObjectsRequest(S3Location location, String keyPrefix)
+    {
+        return ListObjectsV2Request.builder()
+                .overrideConfiguration(context::applyCredentialProviderOverride)
+                // Restore status will not be added to the response without requested
```
electrum: Grammar nit
```java
// Restore status will only be added to the response if requested

File: lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java [hunk 37]
```diff
@@ -227,6 +230,56 @@ void renameFile(Location source, Location target)
...
+     * implementations: object stores commonly sort by UTF-8 byte order, while in-memory and
+     * local implementations sort by Java {@link String#compareTo char-by-char order}. Because
+     * every non-ASCII code point is strictly greater than every ASCII character under all of
+     * these orderings, an ASCII {@code startingFrom} never excludes a non-ASCII filename.
+     * Callers that rely on a specific sort order should restrict usage to directories whose
+     * filenames are entirely ASCII.
+     *
+     * @param location the directory to list
+     * @param startingFrom the inclusive lower bound for the path remainder after {@code location};
+     *                     must contain only ASCII characters
```
electrum: Nit: Javadoc continuation lines should use normal 8 space continuation indent, not alignment

electrum (approved):

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: Done — `listObjects` now takes a plain `String`, and `listFilesStartingFrom` simplifies to `return listObjects(location, false, startingFrom);`.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: Fixed.

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
raunaqmorarka: Fixed.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: Fixed as part of the refactor above.

lib/trino-filesystem-s3/src/main/java/io/trino/filesystem/s3/S3FileSystem.java
raunaqmorarka: Fixed as part of the refactor above.

raunaqmorarka:
@coderabbitai full review

PR #28554, opened 2026-03-05
Added retryable tests pattern

p3tr0sz:
## Description
Added a `RetryableFailurePattern` SPI for automatically retrying tests that fail due to known transient infrastructure issues.

### Motivation
Currently, tests that fail due to transient infrastructure problems (e.g., Databricks JDBC communication failures, Hadoop HDFS replication issues) can only be retried if they are individually annotated with `@Flaky`. This has two problems:
  1. Every test interacting with flaky infrastructure must carry its own `@Flaky` annotation with duplicated regex patterns - this is tedious and error-prone.
  2. New tests are not flaky at the start and are easily forgotten. It may lead to spurious CI failures that waste developer time and erode trust in the test suite. Non flaky tests can become ones.
 
### Approach
Instead of requiring per-test annotations, this PR introduces a `RetryableFailurePattern` interface discovered via Java `ServiceLoader`. Each implementation provides a regex pattern that is matched against the failure stacktrace. When any registered pattern matches, `FlakyTestRetryAnalyzer` retries the test (up to the `ALLOWED_RETRIES_COUNT` times) - regardless of whether `@Flaky` is present. This means known infrastructure failure patterns are defined once in the appropriate module and automatically apply to all TestNG tests on the classpath.

### Documentation
Updated docs/src/main/sphinx/develop/tests.md with usage instructions

## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-product-tests/pom.xml
+++ b/testing/trino-product-tests/pom.xml
@@ -366,6 +366,23 @@
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <dependencies>
+                    <!-- allow both JUnit and TestNG -->
+                    <dependency>
+                        <groupId>org.apache.maven.surefire</groupId>
+                        <artifactId>surefire-junit-platform</artifactId>
+                        <version>${dep.plugin.surefire.version}</version>
+                    </dependency>
+                    <dependency>
+                        <groupId>org.apache.maven.surefire</groupId>
+                        <artifactId>surefire-testng</artifactId>
+                        <version>${dep.plugin.surefire.version}</version>
+                    </dependency>
+                </dependencies>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-shade-plugin</artifactId>
                 <executions>
                     <execution>
--- a/testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
+++ b/testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
@@ -1,0 +1,80 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.tests.product;
+
+import io.airlift.log.Logger;
+import io.trino.testng.services.FlakyTestRetryAnalyzer;
+import org.testng.IRetryAnalyzer;
+import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.regex.Pattern;
+
+import static com.google.common.base.Throwables.getStackTraceAsString;
+import static io.trino.testng.services.FlakyTestRetryAnalyzer.ALLOWED_RETRIES_COUNT;
+import static io.trino.tests.product.utils.HadoopTestUtils.RETRYABLE_FAILURES_MATCH;
+
+/**
+ * Retries convention-based product tests that fail due to known transient
+ * infrastructure failures that cannot be fixed in the test itself.
+ */
+public class ProductTestRetryAnalyzer
+        implements IRetryAnalyzer
+{
+    private static final Logger log = Logger.get(ProductTestRetryAnalyzer.class);
+
+    private static final String DISABLED_SYSTEM_PROPERTY = "io.trino.tests.product.ProductTestRetryAnalyzer.disabled";
+
+    private static final Pattern INFRASTRUCTURE_FAILURE_PATTERN = Pattern.compile(RETRYABLE_FAILURES_MATCH);
+
+    private final Map<String, Long> retryCounter = new ConcurrentHashMap<>();
+
+    @Override
+    public boolean retry(ITestResult result)
+    {
+        if (result.isSuccess()) {
+            return false;
+        }
+
+        if (Boolean.getBoolean(DISABLED_SYSTEM_PROPERTY)) {
+            log.info("ProductTestRetryAnalyzer disabled via system property '%s'", DISABLED_SYSTEM_PROPERTY);
+            return false;
+        }
+
+        if (result.getThrowable() == null) {
+            return false;
+        }
+
+        String stackTrace = getStackTraceAsString(result.getThrowable());
+        if (!INFRASTRUCTURE_FAILURE_PATTERN.matcher(stackTrace).find()) {
+            return false;
+        }
+
+        ITestNGMethod method = result.getMethod();
+        String testId = FlakyTestRetryAnalyzer.testId(method, result.getParameters());
+        long retryCount = retryCounter.merge(testId, 1L, Long::sum);
+        if (retryCount > ALLOWED_RETRIES_COUNT) {
+            return false;
+        }
+        log.warn(
+                result.getThrowable(),
+                "Test %s::%s attempt %s failed due to infrastructure failure, retrying...,",
+                result.getTestClass().getName(),
+                method.getMethodName(),
+                retryCount);
+        return true;
+    }
+}
--- a/testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryListener.java
+++ b/testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryListener.java
@@ -1,0 +1,46 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.tests.product;
+
+import io.airlift.log.Logger;
+import io.trino.tempto.internal.convention.ConventionBasedTestProxyGenerator;
+import org.testng.IClassListener;
+import org.testng.ITestClass;
+import org.testng.ITestNGMethod;
+
+public class ProductTestRetryListener
+        implements IClassListener
+{
+    private static final Logger log = Logger.get(ProductTestRetryListener.class);
+
+    @Override
+    public void onBeforeClass(ITestClass testClass)
+    {
+        Class<?> realClass = testClass.getRealClass();
+        if (isConventionBasedTest(realClass)) {
+            for (ITestNGMethod method : testClass.getTestMethods()) {
+                log.debug("Instrumenting method %s with %s.", method.getMethodName(), ProductTestRetryAnalyzer.class.getSimpleName());
+                method.setRetryAnalyzer(new ProductTestRetryAnalyzer());
+            }
+        }
+    }
+
+    @Override
+    public void onAfterClass(ITestClass testClass) {}
+
+    private static boolean isConventionBasedTest(Class<?> realClass)
+    {
+        return ConventionBasedTestProxyGenerator.ConventionBasedTestProxy.class.isAssignableFrom(realClass);
+    }
+}
--- a/testing/trino-product-tests/src/main/resources/META-INF/services/org.testng.ITestNGListener
+++ b/testing/trino-product-tests/src/main/resources/META-INF/services/org.testng.ITestNGListener
@@ -1,0 +1,1 @@
+io.trino.tests.product.ProductTestRetryListener
--- a/testing/trino-product-tests/src/test/java/io/trino/tests/product/TestProductTestRetryAnalyzer.java
+++ b/testing/trino-product-tests/src/test/java/io/trino/tests/product/TestProductTestRetryAnalyzer.java
@@ -1,0 +1,147 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.tests.product;
+
+import org.testng.ITestNGMethod;
+import org.testng.ITestResult;
+import org.testng.annotations.Test;
+import org.testng.internal.NoOpTestClass;
+import org.testng.internal.TestResult;
+
+import java.lang.reflect.Proxy;
+
+import static io.trino.testng.services.FlakyTestRetryAnalyzer.ALLOWED_RETRIES_COUNT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@Test(singleThreaded = true)
+public class TestProductTestRetryAnalyzer
+{
+    @Test
+    public void testRetryOnHadoopReplicationFailure()
+    {
+        ProductTestRetryAnalyzer analyzer = new ProductTestRetryAnalyzer();
+        ITestResult result = failedResult("testMethod", new RuntimeException(
+                "could only be replicated to 0 nodes instead of minReplication"));
+
+        for (int i = 0; i < ALLOWED_RETRIES_COUNT; i++) {
+            assertThat(analyzer.retry(result))
+                    .describedAs("retry attempt %d", i + 1)
+                    .isTrue();
+        }
+        assertThat(analyzer.retry(result))
+                .describedAs("should not retry beyond allowed count")
+                .isFalse();
+    }
+
+    @Test
+    public void testRetryOnHadoopWriteFailure()
+    {
+        ProductTestRetryAnalyzer analyzer = new ProductTestRetryAnalyzer();
+        ITestResult result = failedResult("testMethod", new RuntimeException(
+                "could only be written to 0 of the 1 minReplication"));
+
+        assertThat(analyzer.retry(result)).isTrue();
+    }
+
+    @Test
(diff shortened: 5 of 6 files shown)


martint (changes requested):
I'm not sure I agree with this approach.
* It makes it harder to track which tests are considered flaky. Currently, can be directly inferred by looking at the test and its annotations.
* It masks issues that may creep up over time. Existing tests that become flaky (e.g., because of broken retry logic within connectors) are automatically retried without anyone noticing.
* there's no one-size-fits-all retry policy. For slow, heavy tests, we may want fewer retries. For faster tests, we may want more, spaced out retries to give the underlying infrastructure a chance to recover (assuming that's the cause of the failure).

The main goal `@Flaky` was to 1) identify tests that are flakyt and 2) work around temporary flakiness until the underlying issue is fixed.

There's even this comment in the `@Flaky` javadocs:

> Deprecated
> Use of this is strongly discouraged. Instead prefer to find the cause of the issue and either fix the test or rewrite it to > not be flaky.

Finally, TestNG is only used in product tests and a limited number of other places that have yet not migrated away from it.

p3tr0sz:
> I'm not sure I agree with this approach.
> 
> * It makes it harder to track which tests are considered flaky. Currently, can be directly inferred by looking at the test and its annotations.
> * It masks issues that may creep up over time. Existing tests that become flaky (e.g., because of broken retry logic within connectors) are automatically retried without anyone noticing.
> * there's no one-size-fits-all retry policy. For slow, heavy tests, we may want fewer retries. For faster tests, we may want more, spaced out retries to give the underlying infrastructure a chance to recover (assuming that's the cause of the failure).
> 
> The main goal `@Flaky` was to 1) identify tests that are flakyt and 2) work around temporary flakiness until the underlying issue is fixed.
> 
> There's even this comment in the `@Flaky` javadocs:
> 
> > Deprecated
> > Use of this is strongly discouraged. Instead prefer to find the cause of the issue and either fix the test or rewrite it to > not be flaky.
> 
> Finally, TestNG is only used in product tests and a limited number of other places that have yet not migrated away from it.

@martint Thanks for the detailed feedback.  Let me address the points:

* Tracking flaky tests: Looking at the data, out of 200+ `@Flaky` annotations in the codebase, only 1 seems like it is for an actual flaky test (TestHiveTransactionalTable#testFilesForAbortedTransactionsIgnored). All other use either `DATABRICKS_COMMUNICATION_FAILURE_MATCH` or `RETRYABLE_FAILURES_MATCH`. These aren't flaky tests in the term of producing inconsistent results - they're stable tests hitting transient infrastructure issues.
* The current `@Flaky` mechanism also uses a fixed retry count with no per-test customization
* I saw that using `@Flaky` is discouraged and I totally agree with fixing root cause instead. But the root cause here (Databricks API instability and Hadoop communication failures) is external infrastructure we don't control and cannot fix. These aren't bugs in our code or poorly written tests.

I attached Retryable mechanism to the existing `FlakyTestRetryAnalyzer` and probably that was a mistake. My intention wasn’t to provide a workaround that hides test flakiness. On the contrary - I wanted to separate transient infrastructure issues from code that doesn’t always return the correct result.
Updated the PR to reflect that more.

If the plan is to migrate these product tests to JUnit soon, the investment may not be justified. Is there a timeline for that migration? If it's not imminent, this change would eliminate ~240 boilerplate annotations and prevent every new Databricks/Hadoop product test from needing to add the same `@Flaky` line (like recent https://github.com/trinodb/trino/pull/28531/changes).

findepi:
> * and Hadoop communication failures) is external infrastructure we don't control and cannot fix. These aren't bugs in our code or poorly written tests.

I agree with the conclusion. I just would word it differently. Let me focus on HDFS write failures.
I am pretty sure it's possible to configure HDFS so that it runs with redundancy and I would assume/hope it eliminates the failures. I just think that

-  it's unlikely someone will actually spend much time on it, especially given that HDFS is not a cutting edge tech stack people are excited about these days. The problem is known for years....
- it's possible that configuring HDFS for high availability requires more than one host, and tests are run on a single machine. Challenging single machine assumption is would be super expensive (for CI, for running tests locally). I don't believe community would find/fund engineering time to take care of that
- if we do nothing, we need to live with flaky tests. Flaky tests are evil. I don't want to do that.

testing/trino-product-tests/src/main/resources/META-INF/services/io.trino.testng.services.InfrastructureRetryableFailurePattern
```diff
@@ -0,0 +1,2 @@
+io.trino.tests.product.utils.HadoopRetryableFailurePattern
+io.trino.tests.product.deltalake.util.DatabricksRetryableFailurePattern
```
findepi: This is more complicated than needed.
IMO what we need is: a single class like `FlakyTestRetryAnalyzer` (let's call it `ProductTestRetryAnalyzer`) that simply wraps all tests (or maybe just all convention-based tests cc @losipiuk) unconditionally.
Applying for convention-based only would be equivalent to what we agreed upon and are already doing

```
trino master$ git grep '@Flaky(issue = RETRYABLE_FAILURES_ISSUES, match = RETRYABLE_FAILURES_MATCH)' | wc -l
159
```

We can also apply to Java tests based on their group or package, which would allow us to remove these 159 lines. (I am less concerned about those 159 lines we already have, I am more concerned about those X lines we didn't add yet.)

I don't think we need dynamic loading of patterns. The ProductTestRetryAnalyzer can hard-code any pattern(s) it needs to recognize.

I don't think we need to solve databricks failures with it. We can just keep these as-is:
```
trino master$ git grep '@Flaky(issue = DATABRICKS_COMMUNICATION_FAILURE_ISSUE' | wc -l
66
```

p3tr0sz: I simplified the codebase and enabled for convention based tests only.

docs/src/main/sphinx/develop/tests.md
```diff
@@ -178,6 +178,20 @@ removed.
 New tests with the `@Flaky` annotation can not be introduced, since new tests
 must use JUnit. Rewrite the test to be stable or avoid the test altogether.
 
+### Product test infrastructure failure retries
```
findepi: > Updated README


Squash with previous commit?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryListener.java
```diff
@@ -0,0 +1,47 @@
...
+            }
+        }
+    }
+
+    @Override
+    public void onAfterClass(ITestClass testClass) {}
+
+    private static boolean isConventionBasedTest(Class<?> realClass)
+    {
+        return realClass.getSuperclass() != null && CONVENTION_BASED_TEST_PROXY.equals(realClass.getSuperclass().getName());
```
findepi: This could also be

```
Class.forName("io.trino.tempto.internal.convention.ConventionBasedTestProxyGenerator$ConventionBasedTestProxy").isAssignableForm(realClass)
```

docs/src/main/sphinx/develop/tests.md
```diff
@@ -178,6 +178,20 @@ removed.
...
+
+Some product test failures are caused by transient infrastructure issues rather
+than actual bugs, such as Databricks JDBC communication failures or Hadoop HDFS
+replication errors. These are distinct from flaky tests — the tests themselves
+are correct, but the external infrastructure they depend on has known transient
+failure modes that cannot be fixed in the test.
+
+These failures are handled by ``ProductTestRetryAnalyzer``, which is applied to
+all product tests without ``@Flaky``. It hardcodes patterns for known
```
findepi: It's outdated, right?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
```diff
@@ -0,0 +1,123 @@
...
+
+    private static final String ENABLED_SYSTEM_PROPERTY = "io.trino.tests.product.ProductTestRetryAnalyzer.enabled";
+
+    static final int ALLOWED_RETRIES_COUNT = 2;
+
+    // Hadoop HDFS replication and MapReduce failures
+    // https://github.com/trinodb/trino/issues/4936
+    // https://github.com/trinodb/trino/issues/5427
+    @Language("RegExp")
+    private static final String HADOOP_RETRYABLE_FAILURE_PATTERN =
```
findepi: This constant already exists somewhere else, right?
Can we reuse instead of copying?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
```diff
@@ -0,0 +1,123 @@
...
+ * infrastructure failures that cannot be fixed in the test itself.
+ */
+public class ProductTestRetryAnalyzer
+        implements IRetryAnalyzer
+{
+    private static final Logger log = Logger.get(ProductTestRetryAnalyzer.class);
+
+    private static final String ENABLED_SYSTEM_PROPERTY = "io.trino.tests.product.ProductTestRetryAnalyzer.enabled";
+
+    static final int ALLOWED_RETRIES_COUNT = 2;
```
findepi: Flaky allows 2 or 3?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
```diff
@@ -0,0 +1,123 @@
...
+        boolean matchesPattern = INFRASTRUCTURE_FAILURE_PATTERNS.stream()
+                .anyMatch(pattern -> pattern.matcher(stackTrace).find());
+        if (!matchesPattern) {
+            return false;
+        }
+
+        long retryCount;
+        ITestNGMethod method = result.getMethod();
+        synchronized (this) {
+            String name = getName(method, result.getParameters());
```
findepi: outside of synchronized?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
```diff
@@ -0,0 +1,123 @@
...
+            retryCount = retryCounter.getOrDefault(name, 0L);
+            retryCount++;
```
findepi: Map read, update, write back can be in one call.

Map.merge or something?

testing/trino-product-tests/pom.xml
```diff
@@ -364,6 +364,28 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <systemPropertyVariables>
+                        <io.trino.tests.product.ProductTestRetryAnalyzer.enabled>true</io.trino.tests.product.ProductTestRetryAnalyzer.enabled>
+                    </systemPropertyVariables>
+                </configuration>
+                <dependencies>
+                    <!-- allow both JUnit and TestNG -->
```
findepi, line 371: Do we have any JUnit in this module?

testing/trino-product-tests/pom.xml
```diff
@@ -364,6 +364,28 @@
         </resources>
 
         <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <systemPropertyVariables>
+                        <io.trino.tests.product.ProductTestRetryAnalyzer.enabled>true</io.trino.tests.product.ProductTestRetryAnalyzer.enabled>
+                    </systemPropertyVariables>
```
findepi: This might be confusing.
This property is used only when tests-of-product-tests are executed.
ProductTestRetryAnalyzer needs to be enabled when product tests' executable jar is run, so let's just have it enabled by default.
There can still be a system property to _disable_ it.

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
p3tr0sz: Flaky allows 2.
https://github.com/trinodb/trino/blob/master/testing/trino-testing-services/src/main/java/io/trino/testng/services/FlakyTestRetryAnalyzer.java#L44

testing/trino-product-tests/pom.xml
p3tr0sz: Yup, TestProductTestGroups is Junit. https://github.com/trinodb/trino/blob/8f933c805be21b6968034fad1af218ac35ec6676/testing/trino-product-tests/src/test/java/io/trino/tests/product/TestProductTestGroups.java#L98-L100

For TestProductTestRetryAnalyzer I followed TestFlakyTestRetryAnalyzer which is TestNG. I thought of having TestProductTestRetryAnalyzer as JUnit, but I would require to programatically invoke TestNG. I decided it'd add the complexity, so went with TestNG.

martint:
> I saw that using @Flaky is discouraged and I totally agree with fixing root cause instead. But the root cause here (Databricks API instability and Hadoop communication failures) is external infrastructure we don't control and cannot fix. These aren't bugs in our code or poorly written tests.

If these are issues that surface via interactions from Trino connectors, then I'd argue that the connectors are either not configured correctly, or they are deficient. If these issues occur in a test environment, they will occur in non-test environments and lead to query failures. The right thing to do would be to update the connectors to do internal retries.

Now, if we're talking about, say, Hive or Spark within product tests having issues (e.g., when we use them to test compatibility, etc) and when there is no way to control or configure retries for those tools, then retrying at higher layers is warranted. However, are we able to distinguish between failures in third party software vs failures within in Trino? If not, retries would potentially mask legitimate problems.

testing/trino-product-tests/pom.xml
p3tr0sz: Done, take a look if you can.

p3tr0sz:
> Now, if we're talking about, say, Hive or Spark within product tests having issues (e.g., when we use them to test compatibility, etc) and when there is no way to control or configure retries for those tools, then retrying at higher layers is warranted. However, are we able to distinguish between failures in third party software vs failures within in Trino? If not, retries would potentially mask legitimate problems.

Failures of external tools are distinguishable from Trino failures via message patterns. The setting I introduced will retry on very specific Hadoop/HDFS and Hive MapReduce error messages. This is not new retry logic - the same patterns are already used by HDFS related tests. As @findepi suggested I enabled it for convention based tests only.

To your question about distinguishing failures in third-party software vs Trino: I considered restricting retries to failures that don't pass through Trino's query path (by checking for `TrinoException/io.trino.jdbc` in the stack trace). However, convention-based tests execute all SQL through Trino, so every exception - including transient HDFS errors - contains Trino frames. This filtering would prevent all retries, defeating the purpose.

That said, honestly:
1. If a Trino connector bug somehow degraded HDFS (e.g., resource exhaustion, connection leaks), these retries could mask that.
2. Nothing structurally prevents someone from adding a pattern that matches a Trino-specific error and masks a real bug. It relies on the author and reviewer discipline.

findepi:
> If these are issues that surface via interactions from Trino connectors, then I'd argue that the connectors are either not configured correctly, or they are deficient. If these issues occur in a test environment, they will occur in non-test environments and lead to query failures. The right thing to do would be to update the connectors to do internal retries.

I totally agree with this principle. That's what I am advocating for all the time (within this project and also in internal projects).

The tests tell us that Trino filesystem (and so eg hive connector) doesn't work reliably when using single-node HDFS installation.
This is not a production setup, and that setup is possible (but not certain) reason why it works unreliably. As far as test results are concerned, we should suspect production installations are affected as well.

However, 
- this failure mode (eg "could only be written to 0 of the 1 minReplication") is plaguing our product tests for years
- users do not report such failures. All the known issues seem to be tied to our own tests output
- AFAIK no-one has motivation to work on these tests or test infrastructure to eliminate the source of these errors (could perhaps require additional test resources and definitely considerable engineering time)
- we keep adding `@Flaky` to selected tests (159 occurrences so far) -- just retrying is established pattern we accepted for this exact problem (not in general!). This PR adds this retry pattern to tests that cannot get the annotation, because they are not defined in Java


I see following options we can take

1. do nothing. Accept CI flakes and just keep ignoring these failures with our eyeballs
2. delete all the "convention based tests" 
3. rewrite all the "convention based tests" to be Java-defined and apply retries via `@Flaky` on them.
4. apply retries on "convention based tests" (this PR)
5. have some other retry mechanism (e.g. retries on CI job level or something like that) 

if that matters, I'm personally against the option 1, 5 and 2.

docs/src/main/sphinx/develop/tests.md
```diff
@@ -178,6 +178,23 @@ removed.
 New tests with the `@Flaky` annotation can not be introduced, since new tests
 must use JUnit. Rewrite the test to be stable or avoid the test altogether.
 
+### Product test infrastructure failure retries
+
+Some product test failures are caused by transient infrastructure issues rather
+than actual bugs, such as Hadoop HDFS replication errors. These are distinct
```
findepi: As Marting correctly pointed out, these may be actual production bugs.


What about deleting this readme entry altogether?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
findepi: Can we read the value from `io.trino.testng.services.FlakyTestRetryAnalyzer#ALLOWED_RETRIES_COUNT`?

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+
+    @Override
+    public boolean retry(ITestResult result)
+    {
+        if (result.isSuccess()) {
+            return false;
+        }
+
+        Optional<String> disabledPropertyValue = Optional.ofNullable(System.getProperty(DISABLED_SYSTEM_PROPERTY));
+        if (disabledPropertyValue.map(Boolean::parseBoolean).orElse(false)) {
```
findepi:
```suggestion
        if (Boolean.getBoolean(DISABLED_SYSTEM_PROPERTY)) {
```

testing/trino-product-tests/src/main/java/io/trino/tests/product/ProductTestRetryAnalyzer.java
```diff
@@ -0,0 +1,97 @@
...
+{
+    private static final Logger log = Logger.get(ProductTestRetryAnalyzer.class);
+
+    private static final String DISABLED_SYSTEM_PROPERTY = "io.trino.tests.product.ProductTestRetryAnalyzer.disabled";
+
+    static final int ALLOWED_RETRIES_COUNT = 2;
+
+    private static final Pattern INFRASTRUCTURE_FAILURE_PATTERN = Pattern.compile(RETRYABLE_FAILURES_MATCH);
+
+    private final Map<String, Long> retryCounter = new HashMap<>();
```
findepi: concurrent map?

findepi (approved):
Added fixups in a commit. Please pull & squash.

Change commit message to sth like
```
Retry HDFS failures in convention-based product tests

In product tests Trino experiences intermittent HDFS failures.
This may be testing infrastructure deficiency or a product problem.
For Java-defined tests this is currently dealt with using the `@Flaky`
annotation. This adds same retries to convention-based tests too.
```

PR #28557, opened 2026-03-05
Fix invalid subtree reading

wendigo:
## Description

Found invalid deserializers while testing https://github.com/airlift/airlift/pull/1811

Correct way of decoding (sub)trees is by using DeserializationContext which is aware
whether decoded tree is a part of a bigger document. If FAIL_ON_TRAILING_TOKENS
is enabled, decoding without context will fail if the decoded tree is a substree in a bigger
object as the parser will check for next token that exists for the outer object.

## Release notes

```markdown
## Section
```


martint (approved):
Add a note to the commit or PR description explaining why they are invalid, as it's not entirely obvious that the former way is incorrect.

PR #28562, opened 2026-03-06
add `privateKey` as a JVM property for Snowflake JDBC Connector

### LINKED ISSUE #26514: Snowflake Connector: Add support for `privateKey` authentication property

parthshyara:
## Description
* Adds support for the privateKey authentication property in the Snowflake connector.
* The Snowflake JDBC driver already supports this property, but Trino previously had no way to pass it through.
* With this change, users can configure key-pair authentication directly in the catalog instead of relying on file-based approaches.

* Fixes #26514
* Provides a more secure and convenient alternative to using JDBC URL parameters with file paths and passwords.
* Particularly useful for dynamic catalog creation, where file system dependencies are harder to manage.

## Release notes

(&#x2713;) Release notes are required, with the following suggested text:

```markdown
## Snowflake connector
* Add support for the `privateKey` authentication property to enable key-pair authentication. (#26514)
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -514,7 +514,8 @@ jobs:
             - { modules: plugin/trino-resource-group-managers }
             - { modules: plugin/trino-singlestore }
             - { modules: plugin/trino-snowflake }
-            - { modules: plugin/trino-snowflake, profile: cloud-tests }
+            - { modules: plugin/trino-snowflake, profile: cloud-tests, auth: private-key }
+            - { modules: plugin/trino-snowflake, profile: cloud-tests, auth: password }
             - { modules: plugin/trino-sqlserver }
             - { modules: plugin/trino-vertica }
             - { modules: testing/trino-faulttolerant-tests, profile: default }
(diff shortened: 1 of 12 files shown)


parthshyara:
Re-opening from #26519

PR #28569, opened 2026-03-06
Do not expose JMX beans over http

wendigo:
JmxHttpModule exposes all mbeans along with the embededded UI that allows to browse MBeans.

This is obsolete way of monitoring Trino that was already
replaced by much modern approach like openmetrics and opentracing.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/security/ServerSecurityModule.java
+++ b/core/trino-main/src/main/java/io/trino/server/security/ServerSecurityModule.java
@@ -22,7 +22,6 @@ import com.google.inject.multibindings.MapBinder;
 import io.airlift.configuration.AbstractConfigurationAwareModule;
 import io.airlift.http.server.HttpServer.ClientCertificate;
 import io.airlift.http.server.HttpServerConfig;
-import io.airlift.jmx.MBeanResource;
 import io.airlift.openmetrics.MetricsResource;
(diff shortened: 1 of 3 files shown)


core/trino-main/src/main/java/io/trino/server/security/ServerSecurityModule.java
```diff
@@ -56,7 +55,6 @@ protected void setup(Binder binder)
         jaxrsBinder(binder).bind(ResourceSecurityDynamicFeature.class);
 
         resourceSecurityBinder(binder)
-                .managementReadResource(MBeanResource.class)
```
findepi, line 59: Should `MBeanResource` be deprecated in Airlift?
I know it's not automatic just because Trino no longer uses that, but the PR description seems applicable to any Java app

> JmxHttpModule exposes all mbeans along with the embededded UI that allows to browse MBeans.
> 
> This is obsolete way of monitoring Trino that was already
replaced by much modern approach like openmetrics and opentracing.

martint: > Should MBeanResource be deprecated in Airlift?

I don't think so. That's a simple way of getting data about mbeans when you don't want to or can't run VisualVM.

PR #28555, opened 2026-03-05
Add retries to Bigquery delete test

anusudarsan:
## Release notes

(x) This is not user-visible or is docs only, and no release notes are required.

PR #28570, opened 2026-03-06
Add retries to BigQuery smoke delete test

ebyhr:
## Description

* Follow-up of https://github.com/trinodb/trino/pull/28555

## Release notes

### DIFF
--- a/plugin/trino-bigquery/src/test/java/io/trino/plugin/bigquery/TestBigQueryWithDifferentProjectIdConnectorSmokeTest.java
+++ b/plugin/trino-bigquery/src/test/java/io/trino/plugin/bigquery/TestBigQueryWithDifferentProjectIdConnectorSmokeTest.java
@@ -13,6 +13,8 @@
  */
 package io.trino.plugin.bigquery;
 
+import dev.failsafe.Failsafe;
+import dev.failsafe.RetryPolicy;
 import io.trino.testing.BaseConnectorSmokeTest;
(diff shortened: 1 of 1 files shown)


plugin/trino-bigquery/src/test/java/io/trino/plugin/bigquery/TestBigQueryWithDifferentProjectIdConnectorSmokeTest.java
```diff
@@ -127,4 +129,24 @@ public void testNativeQuerySelectFromTestTable()
...
+                .run(super::testDeleteAllDataFromTable);
+    }
+
+    @Test
+    @Override
+    public void testRowLevelDelete()
+    {
+        // there is a brief delay before newly written data becomes visible in the BigQuery connector
+        // https://github.com/trinodb/trino/issues/20894
+        Failsafe.with(RetryPolicy.builder().withMaxAttempts(3).build())
```
wendigo: You could extract policy to a const to reduce repetitions

ebyhr: Makes sense, extracted a constant: 
```java
private static final RetryPolicy<?> DELETE_TEST_RETRY_POLICY = RetryPolicy.builder().withMaxAttempts(3).build();
```

PR #28573, opened 2026-03-06
Do not wrap with Slice for comparisons and hashing

wendigo:
## Description

Saves on the allocation while comparing

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/type/AbstractVariableWidthType.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/type/AbstractVariableWidthType.java
@@ -363,7 +363,13 @@ public abstract class AbstractVariableWidthType
                 leftBytes = leftVariableSizeSlice;
                 leftOffset = leftVariableSizeOffset;
             }
-            return rightRawSlice.equals(rightRawSliceOffset, rightLength, wrappedBuffer(leftBytes, leftOffset, leftLength), 0, leftLength);
+            return Arrays.equals(
+                    rightRawSlice.byteArray(),
+                    rightRawSlice.byteArrayOffset() + rightRawSliceOffset,
+                    rightRawSlice.byteArrayOffset() + rightRawSliceOffset + rightLength,
+                    leftBytes,
+                    leftOffset,
+                    leftOffset + leftLength);
         }
 
         @ScalarOperator(XX_HASH_64)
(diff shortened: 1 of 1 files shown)


core/trino-spi/src/main/java/io/trino/spi/type/AbstractVariableWidthType.java
```diff
@@ -363,7 +363,7 @@ private static boolean equalOperator(
                 leftBytes = leftVariableSizeSlice;
                 leftOffset = leftVariableSizeOffset;
             }
-            return rightRawSlice.equals(rightRawSliceOffset, rightLength, wrappedBuffer(leftBytes, leftOffset, leftLength), 0, leftLength);
+            return Arrays.equals(rightRawSlice.byteArray(), rightRawSlice.byteArrayOffset() + rightRawSliceOffset, rightLength, leftBytes, leftOffset, leftLength);
```
pettyjamesm: Arguments should be `fromIndex, toIndex` but this is `fromIndex, length`

wendigo: Right!

core/trino-spi/src/main/java/io/trino/spi/type/AbstractVariableWidthType.java
```diff
@@ -363,7 +363,13 @@ private static boolean equalOperator(
...
                 leftOffset = leftVariableSizeOffset;
             }
-            return rightRawSlice.equals(rightRawSliceOffset, rightLength, wrappedBuffer(leftBytes, leftOffset, leftLength), 0, leftLength);
+            return Arrays.equals(
+                    rightRawSlice.byteArray(),
+                    rightRawSlice.byteArrayOffset() + rightRawSliceOffset,
+                    rightRawSlice.byteArrayOffset() + rightRawSliceOffset + rightLength,
+                    leftBytes,
+                    leftOffset,
+                    leftOffset + leftLength);
```
raunaqmorarka, line 372: Can we instead provide an API in Slice which takes raw byte array as input ?
```
boolean equals(int offset, byte[] that, int otherOffset, int length)

```

wendigo: @raunaqmorarka I can add the API but I don't know whether it will be accepted

wendigo: https://github.com/airlift/slice/pull/188

pettyjamesm:
Will be interested to see if this yields a perf difference in benchmarks, but I kind of doubt it. The slice allocation and equals method may have been inlined before yielding essentially similar code. It shouldn't hurt though, so no objections to the change as is even without a measurable lift.

wendigo:
@pettyjamesm I doubt that but worth checking anyway :)

PR #28579, opened 2026-03-06
Update Slice to 2.6

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/type/AbstractVariableWidthType.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/type/AbstractVariableWidthType.java
@@ -363,13 +363,7 @@ public abstract class AbstractVariableWidthType
                 leftBytes = leftVariableSizeSlice;
                 leftOffset = leftVariableSizeOffset;
             }
-            return Arrays.equals(
-                    rightRawSlice.byteArray(),
-                    rightRawSlice.byteArrayOffset() + rightRawSliceOffset,
-                    rightRawSlice.byteArrayOffset() + rightRawSliceOffset + rightLength,
-                    leftBytes,
-                    leftOffset,
-                    leftOffset + leftLength);
+            return rightRawSlice.equals(rightRawSliceOffset, rightLength, leftBytes, leftOffset, leftLength);
         }
 
         @ScalarOperator(XX_HASH_64)


findepi:
> ## Description
> ## Additional context and related issues
> ## Release notes
> (x) This is not user-visible or is docs only, and no release notes are required. ( ) Release notes are required. Please propose a release note for me. ( ) Release notes are required, with the following suggested text:
> 
> ```
> ## Section
> * Fix some things. ({issue}`issuenumber`)
> ```

If there is no information to be provided in the PR description, would it be useful to delete the clutter that conveys no information at all?

core/trino-main/src/main/java/io/trino/operator/scalar/StringFunctions.java
```diff
@@ -270,15 +270,15 @@ private static long stringPositionFromEnd(Slice string, Slice substring, long in
...
+            index = string.lastIndexOf(substring, index - 1);
             if (index < 0) {
                 return 0;
             }
             foundInstances++;
         }
         while (foundInstances < instance);
 
-        return index + 1;
+        return countCodePoints(string, 0, index + 1);
```
findepi, line 281: That looks like behavior change.
Is this a bugfix?

wendigo: No. It's not a bug fix. Slice finds an index in bytes, not codepoints

wendigo:
@findepi can we remove the PR template as it's not helpful in majority of the PRs?

core/trino-main/src/main/java/io/trino/operator/scalar/StringFunctions.java
martint: It’s confusing because the code change is small but the mechanism is very different. Before, it was converting the slice to a Java string and iterating over code points. Now it’s iterating over bytes and counting the code points at the end.

The commit title is misleading. It should say “avoid converting slice to Java string”, not “use new lastIndexOf” to describe the real intent of the change.

wendigo: @martint right, let me fix commit title

wendigo: Done

dain (approved):

findepi:
> can we remove the PR template as it's not helpful in majority of the PRs?

I am not ignoring, but perhaps this could be brought up in a wider forum.

PR #28580, opened 2026-03-07
Document iceberg.file-based-conflict-detection config property

soooojinlee:
## Description
 Add the `iceberg.file-based-conflict-detection` catalog configuration property to the Iceberg connector documentation. This property exists in the source code(`IcebergConfig.java`) but is missing from the docs. It controls whether partition information from actual written files is used as the source for the conflict detection system, and defaults to `true`. 

## Release notes

```markdown
## Section
```

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -258,6 +258,11 @@ implementation is used:
   - Enable bucket-aware execution. This allows the engine to use physical
     bucketing information to optimize queries by reducing data exchanges.
   - `true`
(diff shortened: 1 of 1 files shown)


ebyhr:
@soooojinlee We intentionally left this property undocumented because we planned to remove it eventually. As far as I know, there have been no bug reports for over a year, so I think this is a good time to start considering its removal.

@pajaks @raunaqmorarka @findinpath

soooojinlee:
Thanks for the context, @ebyhr. I wasn't aware this was intentionally left undocumented. Makes sense to consider removing it instead. Closing this PR.

PR #28581, opened 2026-03-07
Support listFilesStartingFrom in TrinoFileSystem

wendigo:
Modern blob storage systems support performant APIs for enumerating objects starting from a user-provided cursor. This can be used to skip irrelevant objects when listing from large directories, which in turn can make the listing operation much faster and cheaper.

This PR introduces a listFilesStartingFrom method on TrinoFileSystem, with an emulated default implementation for file systems that don't support the operation natively. It includes native, performant implementations for AzureFileSystem, S3FileSystem, and GcsFileSystem.

This API was initially designed to accelerate operations on the _delta_log directory in Delta Lake tables. For large tables, _delta_log can grow to encompass tens of thousands of objects, so efficiently skipping irrelevant objects is critical to overall good performance. However, it may also be useful for optimizations on other file-based table formats as well.

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
@@ -218,7 +218,7 @@ public interface TrinoFileSystem
      * exact name of the prefix, it is not included in the results.
      * <p>
      * The returned FileEntry locations will start with the specified location exactly
-     * and are lexicographically sorted (except for local HDFS which has the system-dependant
+     * and are lexicographically sorted (except for local HDFS which has the system-dependent
      * ordering).
      *
      * @param location the directory to list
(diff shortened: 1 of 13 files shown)


lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +227,35 @@ void renameFile(Location source, Location target)
...
+     * <p>
+     * The returned FileEntry locations will start with the specified location exactly
+     * and are lexicographically sorted (except for local HDFS which has the system-dependent
+     * ordering).
+     *
+     * @param location the directory to list
+     * @param startingFrom the inclusive lower bound for the path remainder after {@code location}
+     * @throws IllegalArgumentException if location is not valid for this file system
+     */
+    default FileIterator listFilesStartingFrom(Location location, String startingFrom)
```
findepi, line 253: This is great for paging _outside_ of FS layer. For paging to work, we might want to support limit here too.

What is the intended use-case for this method?

lib/trino-filesystem/src/main/java/io/trino/filesystem/TrinoFileSystem.java
```diff
@@ -227,6 +227,35 @@ void renameFile(Location source, Location target)
     FileIterator listFiles(Location location)
             throws IOException;
 
+    /**
+     * Lists files within the specified directory recursively, including only files where the
+     * remainder of the path after the location is lexicographically greater than or equal to
+     * {@code startingFrom}. The location can be empty, listing all files in the file system where the
```
findepi, line 233: It's important to clarify the `/` behavior (must, should, may [not])

```
// In both cases: can `start` start with a slash?
// In both cases: can `start` not start with a slash?
listFilesStartingFrom(Location.of("file:///tmp"), start);
listFilesStartingFrom(Location.of("file:///tmp/"), start);
```

findepi:
Closed likely because it's continued here:

- https://github.com/trinodb/trino/pull/28549

PR #28582, opened 2026-03-07
Harden github action workflows

wendigo:
- Use SHA commit for actions
- Specify permissions needed to run a workflow

## Release notes

```markdown
## Section
```

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -1,0 +1,38 @@
+name: OpenSSF Scorecard
+
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '0 6 * * 1'
+  push:
+    branches: [master]
+
+permissions: read-all
+
+jobs:
+  analysis:
+    name: Scorecard analysis
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+      id-token: write
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
+      - name: Run analysis
+        uses: ossf/scorecard-action@4eaacf0543bb3f2c246792bd56e8cdeffafb205a # v2.4.3
+        with:
+          results_file: results.sarif
+          results_format: sarif
+          publish_results: true
+      - name: Upload artifact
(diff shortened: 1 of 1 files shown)


findepi:
> * Use SHA commit for actions

❤️ 

is there a static analysis we can use to make sure new jobs are safe?

wendigo:
@findepi this is the workflow that I've added (https://scorecard.dev/)

.github/actions/process-test-results/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
   using: composite
   steps:
   - name: Upload test results
-    uses: actions/upload-artifact@v4
+    uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4
```
findepi, line 20: `actions/*` are github's own and are as trusted as github itself

ASF has strict policies about workflow safety and they allow referencing `actions/` actions via tag names

given referencing by name has benefits (convenience) and for those particular ones are not considered supply chain risk, we should keep here using by tag

findepi: BTW we have debated about this already, it's not a new topic

wendigo: I do not recall. Anyway OpenSSF scores tag usage low regardless of whether this is github provided action or not

.github/workflows/scorecard.yml
```diff
@@ -0,0 +1,38 @@
+name: OpenSSF Scorecard
+
+on:
+  branch_protection_rule:
+  schedule:
+    - cron: '0 6 * * 1'
+  push:
+    branches: [master]
```
findepi, line 8: Worth documenting what this workflow does?
Does this workflow scan the code, repo or branch configuration, or?


worth documenting why we have those 3 triggers?
Might become obvious once we know what this does.

.github/actions/process-test-results/action.yml
findepi: it might be my memory fails me (perhaps internal convo, not a PR?).  See https://infra.apache.org/github-actions-policy.html

findepi: we have a test for this already

https://github.com/trinodb/trino/blob/2ef2bccf72834136ae06336656c92b673cbc8c8f/testing/trino-tests/src/test/java/io/trino/tests/ci/TestWorkflows.java#L146-L165

PR #28591, opened 2026-03-09
Use EmptyPageSource.empty() singleton instead of creating new instances

gertjanal:
## Description
I noticed the `new EmptyPageSource()` was invoked multiple times, I think a lazy singleton is more efficient.

## Release notes

(X ) This is not user-visible or is docs only, and no release notes are required.

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/EmptyPageSource.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/EmptyPageSource.java
@@ -16,6 +16,12 @@ package io.trino.spi.connector;
 public class EmptyPageSource
         implements ConnectorPageSource
 {
+    public static final EmptyPageSource EMPTY = new EmptyPageSource();
+
+    private EmptyPageSource()
+    {
+    }
+
     @Override
     public long getCompletedBytes()
     {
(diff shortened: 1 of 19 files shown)


core/trino-spi/src/main/java/io/trino/spi/connector/EmptyPageSource.java
```diff
@@ -16,6 +16,20 @@
...
         implements ConnectorPageSource
 {
+    private static EmptyPageSource empty;
+
+    private EmptyPageSource()
+    {
+    }
+
+    public static EmptyPageSource empty()
+    {
```
wendigo: I'd just create a static final EMPTY_INSTANCE instead

martint:
Have you seen evidence of this being a problem? Empty page sources are not generated very often, and pages in general represent coarser grained chunks of processing. Per-page overheads are usually not a factor (vs per-row overheads)

core/trino-spi/src/main/java/io/trino/spi/connector/EmptyPageSource.java
gertjanal: Done, reduced to `EMPTY`

gertjanal:
> Have you seen evidence of this being a problem? Empty page sources are not generated very often, and pages in general represent coarser grained chunks of processing. Per-page overheads are usually not a factor (vs per-row overheads)

No not hard evidence, but it seems a waste of heap to me.

PR #28593, opened 2026-03-09
Cleanup in TranslationMap

wendigo:
## Release notes

```markdown
## Section
```


core/trino-main/src/main/java/io/trino/sql/planner/TranslationMap.java
```diff
@@ -866,40 +866,29 @@ private io.trino.sql.ir.Expression translate(AtTimeZone node)
...
                     .addArgument(valueType, value)
                     .addArgument(timeZoneType, timeZone)
                     .build();
-        }
-        else {
-            throw new IllegalArgumentException("Unexpected type: " + valueType);
-        }
-
-        return call;
+            case null, default -> throw new IllegalArgumentException("Unexpected type: " + valueType);
```
martint: valueType should never be null

PR #28597, opened 2026-03-10
Do not allocate lambda on every scheduleLeafSplit

wendigo:
Since this lambda captures this, JVM is unable to cache it. By hosting it out of the loop where it is called, lambda allocations are removed.

Before:

CleanShot 2026-03-10 at 11 48 47@2x

After:

CleanShot 2026-03-10 at 11 49 14@2x

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/executor/dedicated/ThreadPerDriverTaskExecutor.java
+++ b/core/trino-main/src/main/java/io/trino/execution/executor/dedicated/ThreadPerDriverTaskExecutor.java
@@ -79,6 +79,9 @@ public class ThreadPerDriverTaskExecutor
     @GuardedBy("this")
     private int runningLeafDrivers;
 
+    // Do not inline this field to avoid creating lambdas that cannot be cached by JVM.
+    private final Runnable leafSplitDoneCallback = this::leafSplitDone;
+
     @Inject
     public ThreadPerDriverTaskExecutor(TaskManagerConfig config, Tracer tracer, VersionEmbedder versionEmbedder)
     {
@@ -177,7 +180,7 @@ public class ThreadPerDriverTaskExecutor
 
     private boolean scheduleLeafSplit(TaskEntry task)
     {
-        boolean scheduled = task.dequeueAndRunLeafSplit(this::leafSplitDone);
+        boolean scheduled = task.dequeueAndRunLeafSplit(leafSplitDoneCallback);
         if (scheduled) {
             runningLeafDrivers++;
         }


wendigo:
Second commit 

before:

CleanShot 2026-03-10 at 12 06 39@2x

after:

CleanShot 2026-03-10 at 12 06 11@2x

core/trino-main/src/main/java/io/trino/execution/executor/dedicated/ThreadPerDriverTaskExecutor.java
```diff
@@ -79,6 +79,8 @@ public class ThreadPerDriverTaskExecutor
     @GuardedBy("this")
     private int runningLeafDrivers;
 
+    private final Runnable leafSplitDoneCallback = this::leafSplitDone;
```
martint, line 83: This needs a comment. Otherwise, someone will inadvertently inline this usage in the future.

core/trino-main/src/main/java/io/trino/execution/executor/dedicated/TaskEntry.java
```diff
@@ -137,10 +137,8 @@ private synchronized void leafSplitDone(QueuedSplit split)
     public synchronized ListenableFuture<Void> runSplit(SplitRunner split)
     {
         int splitId = nextSplitId();
-        ListenableFuture<Void> done = scheduler.submit(
-                group,
-                splitId,
-                new VersionEmbedderBridge(versionEmbedder, new SplitProcessor(taskId, splitId, split, tracer)));
+        VersionEmbeddedProcessor processor = new VersionEmbeddedProcessor(versionEmbedder, new SplitProcessor(taskId, splitId, split, tracer));
+        ListenableFuture<Void> done = scheduler.submit(group, splitId, processor);
```
martint: Why did this change? Revert.

core/trino-main/src/main/java/io/trino/execution/executor/dedicated/TaskEntry.java
```diff
@@ -196,14 +194,31 @@ public synchronized int targetConcurrency()
...
+            implements Schedulable, Runnable
     {
+        private final Schedulable delegate;
+        private final Runnable versionEmbedded;
+        private SchedulerContext context;
+
+        VersionEmbeddedProcessor(VersionEmbedder versionEmbedder, Schedulable delegate)
+        {
+            this.delegate = requireNonNull(delegate, "delegate is null");
+            this.versionEmbedded = requireNonNull(versionEmbedder, "versionEmbedder is null")
```
martint: `requireNonNull` not needed. The call to `embedVersion` will fail with NPE.

core/trino-main/src/main/java/io/trino/execution/executor/dedicated/TaskEntry.java
```diff
@@ -196,14 +194,31 @@ public synchronized int targetConcurrency()
...
-            versionEmbedder.embedVersion(adapter).run();
+            this.context = context;
+            versionEmbedded.run();
+        }
+
+        @Override
+        public void run()
+        {
+            delegate.run(context);
         }
```
martint, line 207: This is confusing (and error prone). From looking at this code, it's not clear which of these methods will get called and when. 

The previous design was much more obvious. 

Does it even matter? This should be executed once per quanta, so the overhead of the allocation should be negligible. If it is a real problem, then we should think about whether VersionEmbedder needs a new design.

wendigo:
Dropped second commit. Added comment.

core/trino-main/src/main/java/io/trino/execution/executor/dedicated/ThreadPerDriverTaskExecutor.java
wendigo: Done

PR #28599, opened 2026-03-10
Add configurable view resolution to OpenLineage event listener

kaveti:
# Add configurable view resolution to OpenLineage event listener

## Description

The OpenLineage Trino plugin currently reports views as input datasets when a query references them. For example, `CREATE TABLE t AS SELECT * FROM my_view` emits `my_view` as the input — not the underlying base table that actually holds the data. This breaks the lineage graph for downstream consumers  because views don't represent physical storage.

This PR adds a new configuration property `openlineage-event-listener.resolve-views` that, when enabled, reports the underlying base tables instead of views as input datasets.

## Problem

When Trino analyzes a query that references a view, `QueryMetadata.getTables()` returns entries for both the view and its underlying base tables. Each entry carries metadata:
- Views have `isDirectlyReferenced() = true` and `getViewText()` present
- Base tables have `isDirectlyReferenced() = false` and `getViewText()` empty

The existing `buildInputs()` method filters on `TableInfo::isDirectlyReferenced`, which means only views are reported and base tables are discarded. This is correct from an access-control perspective but loses physical lineage information.

## Solution

A new boolean config property `openlineage-event-listener.resolve-views` (default `false`) controls which tables are emitted as inputs:

- `false` (default): existing behavior — filter on `isDirectlyReferenced()`, report views
- `true`: filter on `getViewText().isEmpty()`, report only physical base tables

The `viewText` check is the right discriminator because Trino's analyzer sets `viewText` on every view and materialized view entry, regardless of nesting depth. A table that is not a view will always have `viewText = Optional.empty()`.

This handles all cases correctly:
- Single view → base table
- Nested views (v2 → v1 → base_table) — intermediate views filtered out
- View over a JOIN (join_view → t1, t2) — both base tables reported
- Mixed queries (direct table JOIN view) — direct table kept, view replaced by its base table
- Direct table queries — no change in behavior

## Release notes

```markdown
## Section
```

### DIFF
--- a/docs/src/main/sphinx/admin/event-listeners-openlineage.md
+++ b/docs/src/main/sphinx/admin/event-listeners-openlineage.md
@@ -164,6 +164,13 @@ event-listener.config-files=etc/openlineage-event-listener.properties,...
       variables: `$QUERY_ID`, `$USER`, `$SOURCE`, `$CLIENT_IP`.
       For example: `As $USER from $CLIENT_IP via $SOURCE`.
     - `$QUERY_ID`.
+*
+    - openlineage-event-listener.resolve-views
+    - When set to `true`, input datasets report the underlying base tables
+      instead of the view name. This resolves views at any nesting depth,
+      including views over joins with multiple base tables. When `false`,
+      views are reported as input datasets directly.
+    - `false`
 
 :::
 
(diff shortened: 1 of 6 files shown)


kaveti:
closing this PR. as this is already addressed in other PR: https://github.com/trinodb/trino/pull/28625

kaveti:
closing this PR. as this is already addressed in other PR: https://github.com/trinodb/trino/pull/28625

PR #28623, opened 2026-03-11
Cleanup PlanFragment

wendigo:
jsonRepresentation was serialized but never read/used.

dead code, dead code everywhere

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
+++ b/core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
@@ -101,6 +101,7 @@ import java.util.stream.Stream;
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
+import static com.google.common.base.Throwables.getCausalChain;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.util.concurrent.Futures.immediateVoidFuture;
 import static io.airlift.http.client.FullJsonResponseHandler.createFullJsonResponseHandler;
@@ -733,7 +734,11 @@ public final class HttpRemoteTask
             sendUpdateInternal();
         }
         catch (Throwable e) {
-            fatalUnacknowledgedFailure(new TrinoException(GENERIC_INTERNAL_ERROR, "unexpected error calling sendUpdate()", e));
+            Optional<Throwable> trinoException = getCausalChain(e)
+                    .stream()
+                    .filter(TrinoException.class::isInstance)
+                    .findFirst();
+            fatalUnacknowledgedFailure(trinoException.orElseGet(() -> new TrinoException(GENERIC_INTERNAL_ERROR, "unexpected error calling sendUpdate()", e)));
         }
     }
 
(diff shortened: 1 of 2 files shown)


core/trino-main/src/main/java/io/trino/sql/planner/PlanFragment.java
```diff
@@ -107,8 +105,7 @@ public PlanFragment(
             @JsonProperty("outputSkewedBucketCount") OptionalInt outputSkewedBucketCount,
             @JsonProperty("statsAndCosts") StatsAndCosts statsAndCosts,
             @JsonProperty("activeCatalogs") List<CatalogProperties> activeCatalogs,
-            @JsonProperty("languageFunctions") Map<FunctionId, LanguageFunctionData> languageFunctions,
-            @JsonProperty("jsonRepresentation") Optional<String> jsonRepresentation)
```
losipiuk, line 111: Isn't it used by WEB Ui or sth outside of java? I guess not - but how can we be sure.

core/trino-spi/src/main/java/io/trino/spi/type/DateTimeEncoding.java
```diff
@@ -27,7 +30,7 @@ private DateTimeEncoding() {}
     private static long pack(long millisUtc, short timeZoneKey)
     {
         if (millisUtc << MILLIS_SHIFT >> MILLIS_SHIFT != millisUtc) {
-            throw new IllegalArgumentException("Millis overflow: " + millisUtc);
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Millis overflow: " + millisUtc);
```
losipiuk, line 33: fix this as a preparatory commit

core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
```diff
@@ -733,7 +734,11 @@ private void sendUpdate()
             sendUpdateInternal();
         }
         catch (Throwable e) {
-            fatalUnacknowledgedFailure(new TrinoException(GENERIC_INTERNAL_ERROR, "unexpected error calling sendUpdate()", e));
+            Optional<Throwable> trinoException = getCausalChain(e)
```
losipiuk, line 737: should this be a part of change which removed jsonRepresentation then?

losipiuk: I am not sure this is good logic also. Why would you unwrap random layers. Those should not be here. We should expect TrinoException on the top level

wendigo: I don't know honestly

core/trino-main/src/main/java/io/trino/sql/planner/PlanFragment.java
wendigo: It's not

core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
wendigo: Other types already throw TrinoException 🤷‍♂️

core/trino-main/src/main/java/io/trino/sql/planner/PlanFragment.java
martint: The original commit said:

> Add JSON plan representation to enrich live plan view
> 
> - Add remoteSources to NodeRepresentation
> - Add a JSON renderer that produces normalized plan representation
> - Attach rendered JSON to PlanFragment
> - Use rendered JSON to enrich nodes in the live plan with more more descriptive names and identifiers

When did we lose that functionality?

PR #28625, opened 2026-03-12
Add transitive inputs option to openlineage event listener

kekwan:
## Description
Currently, OpenLineage event listener builds input datasets from only directly referenced datasets, which means only the view or table named in the query is emitted. 

For example, given `SELECT * FROM view2` where `view2 → view1 → base_table`, the listener would emit only `view2` as an input dataset. This is not a problem if lineage for `view2` creation was collected from the start AND lineage system can interpret and link the dataset. But if not, current implementation causes problems as we lose `view1` and `base_table` from lineage graph entirely. 

### Solution
- Add a new boolean config property `openlineage-event-listener.include-transitive-inputs` (default false)
- When enabled, all `TableInfo` entries are included as input datasets. Trino already performs the full view/table hierarchy flattening so no additional resolution logic is needed. We just need to include these input datasets. 

## Release notes

```markdown
## Section
* Add transitive input dataset option to OpenLineage event listener
```

### DIFF
--- a/docs/src/main/sphinx/admin/event-listeners-openlineage.md
+++ b/docs/src/main/sphinx/admin/event-listeners-openlineage.md
@@ -164,6 +164,14 @@ event-listener.config-files=etc/openlineage-event-listener.properties,...
       variables: `$QUERY_ID`, `$USER`, `$SOURCE`, `$CLIENT_IP`.
       For example: `As $USER from $CLIENT_IP via $SOURCE`.
     - `$QUERY_ID`.
+*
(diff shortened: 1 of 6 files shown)


kaveti (approved):
thank kekwan for this PR. this is really helpful to align trino with other engines like spark for view handling.

Praveen2112:
@kekwan But with a table info like `view2 → view1 → base_table` - how would we identify which is a base_table and what are the views ? In case of complex queries, as the number of tables increases won't it cause a confusion as on what are the tables involved in the queries ?

PR #28627, opened 2026-03-12
Initial product-test migration to JUnit 5 and Testcontainers.

dain:
## Summary

This PR is the first review slice of the product-tests migration from Tempto/launcher to JUnit 5 and Testcontainers.

It introduces the framework foundation and fully ports the MySQL/MariaDB lane to the new model.

## Scope (in this PR)

- Add JUnit 5 flaky retry extension
- Add JUnit product-test framework infrastructure
- Add MySQL environment + suite
- Port MySQL coverage to JUnit:
  - `TestMySqlSqlTests`
  - `TestCreateTableAsSelect`
  - `TestJdbcDynamicFilteringJmx`
- Add MariaDB environment + suite
- Port `TestMariaDb` to JUnit
- Remove legacy launcher `SuiteMysql` once parity coverage is in place

## Parity and behavior intent

The goal is product-test parity and lane-by-lane safety.  
Changes are intentionally structured in small, reviewable commits (environment/suite/test migration/drain).

## Documentation and reviewer notes

- Reviewer guide: https://github.com/trinodb/trino/blob/user/dain/jpt/all/PRODUCT_TESTS_REVIEWER_GUIDE.md
- Developer guide: testing/trino-product-tests/DEVELOPER_GUIDE.md
- Running guide: testing/trino-product-tests/RUN_PRODUCT_TESTS.md

Please prioritize correctness/parity review for this PR slice.  
I’d prefer to defer broad refactors until after the migration lands to keep rebasing/review risk low.

## Audit records

For reviewers who want deeper parity traceability, the full lane/suite/environment audit records are available in the `user/dain/jpt/all` branch:

- Audit index: https://github.com/trinodb/trino/tree/user/dain/jpt/all/PRODUCT_TESTS_AUDIT
- Audit README: https://github.com/trinodb/trino/blob/user/dain/jpt/all/PRODUCT_TESTS_AUDIT/README.md
- Audit progress tracker: https://github.com/trinodb/trino/blob/user/dain/jpt/all/PRODUCT_TESTS_AUDIT/PROGRESS.md

## Release notes


testing/trino-product-tests/src/test/java/io/trino/tests/product/mariadb/TestMariaDb.java
```diff
@@ -0,0 +1,59 @@
...
+import java.sql.ResultSet;
+import java.sql.Statement;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests CREATE TABLE AS SELECT functionality with MariaDB connector.
+ */
+@ProductTest
+@RequiresEnvironment(MariaDbEnvironment.class)
```
findepi, line 34: When working on new product tests on an environment, the attach option (`testing/bin/ptl test run --attach`) is very useful to have reasonable feedback loop. 
What is the equivalent option in this new system?

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
```diff
@@ -0,0 +1,405 @@
...
+        implements AutoCloseable
+{
+    // Tracks whether this environment is being managed by an EnvironmentManager
+    private volatile boolean managed;
+    private volatile boolean closeAllowed;
+    private Connection trinoSessionConnection;
+
+    /**
+     * Starts the environment. This method should be idempotent - calling it
+     * multiple times should have no effect after the first successful call.
```
findepi, line 84: Why idempotency is important? 
Implementations would be simpler if we drop this requirement. `EnvironmentManager` could handle this if needed.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
```diff
@@ -0,0 +1,405 @@
...
+     *
+     * @param user the user to connect as
+     * @return a new JDBC connection
+     * @throws SQLException if connection cannot be established
+     */
+    public Connection createTrinoConnection(String user)
+            throws SQLException
+    {
+        // Default implementation - subclasses can override for more sophisticated handling
+        return createTrinoConnection();
```
findepi: This is not a good default implementation given `user` parameter gets ignored. Remove.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
```diff
@@ -0,0 +1,405 @@
...
+    }
+
+    private static int executeTrinoUpdateStatement(Statement stmt, String sql)
+            throws SQLException
+    {
+        if (stmt.execute(sql)) {
+            try (ResultSet ignored = stmt.getResultSet()) {
+                // intentionally ignored
+            }
+            return -1;
```
findepi: Looks like AI slope. The `sql` is supposed to be an update and it's supposed not to return any result set.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
```diff
@@ -0,0 +1,405 @@
...
+    public abstract boolean isRunning();
+
+    /**
+     * Hook invoked by {@link ProductTestExtension} before each test method.
+     * <p>
+     * Default implementation is a no-op. Environments may override to perform
+     * per-test setup or cleanup that is environment-owned.
+     */
+    protected void beforeEachTest()
+            throws Exception
```
findepi, line 343: We should not assume tests will always execute single-threaded. This makes CI time longer than necessary.

Do we need this hook?

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
```diff
@@ -0,0 +1,405 @@
...
+     * directly. The manager handles lifecycle automatically. Calling close() on a
+     * managed environment will log a warning with the call stack.
+     */
+    @Override
+    public final void close()
+    {
+        if (managed && !closeAllowed) {
+            // Environment is managed by EnvironmentManager - ignore spurious close calls.
+            // This can happen when JUnit tries to close AutoCloseables in stores.
+            return;
```
findepi, line 372: I don't think PTE needs to be aware of EM.

findepi:
I like the idea that product tests can be directly invoked from the IDE. This is nice quality of life improvement.

However, the reason why existing product tests don't do that is not incidental. 
AFAIR, we never managed to have tests runnable from IDE to perform Kerberos authentication or access HDFS datanodes correctly.
This might be a blocker, i.e. those tests may not be migrate-able to the new approach. 
@dain , is the intention here that we will have two frameworks side by side going forward?



cc @wendigo

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
dain: I change this method to be completely abstract and implemented it in all environments.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
dain: I updated the code to throw an exception where the commit it.  Basically, if `execute` returns true, we maksure the result is fully closed and then throw an exception

testing/trino-product-tests/src/test/java/io/trino/tests/product/mariadb/TestMariaDb.java
dain: There is not a direct equivalent to `ptl ... --attach` in this first tranche of work.

Today the closest options are running the test directly from IntelliJ or Surefire, and in some
later environments I also added small `main` methods that keep an environment running for ad hoc
inspection. That second option is useful for bringing the environment up and poking at it, but
it is not the same as the old `--attach` behavior of running only the tests process against an
already-running environment.

I do think adding an equivalent is possible and worthwhile (I played with this in a branch), but I would prefer not to take that on during the port because it is orthogonal to the framework port itself and would create a 
separate workflow discussion in the middle of the migration review.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
dain: This is intentionally single-threaded at the environment and test-execution level, and I do not
currently plan to make it more parallel than that.

It also gives a much more predictable memory profile than allowing multiple tests in the same
environment to execute concurrently and compete for the same cluster resources. The
`beforeEachTest()` and `afterEachTest()` hooks are there because some environments genuinely need
per-test cleanup or reset, and without this, every single test class and possibly every method needs to re-implement the clean up, which is just truly annoying. This is a much cleaner and easier to maintain approach. 

On runtime, I agree that it is something to keep watching, but the current numbers do not show a
dramatic regression. A recent `master` CI run that exercised the full product-test workload
(`22944582040`) finished in about 57m 02s, while the current `initial` branch run (`22992404817`)
finished in about 56m 07s with failures. So for now I think the simpler and safer execution model
is the right tradeoff.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
dain: I do not think this is problematic coupling in practice.

`ProductTestEnvironment`, `EnvironmentManager`, and the extension are not intended to be a generic
reusable framework with independently swappable layers. They are one small lifecycle system
designed to work together, and the `managed` / `allowClose` handshake is there deliberately to
prevent accidental close paths from tearing down a shared environment at the wrong time.

If we later want to refactor that boundary, I am open to it, but I would prefer not to do that in
this PR unless we find a real lifecycle bug that the current design cannot handle.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
dain: This is intentionally defensive.

`EnvironmentManager` should call `start()` correctly, but I also want environment implementations to behave safely if that contract is violated accidentally. The goal is to avoid the kind of rare lifecycle issue that only shows up in CI and then takes a long time to track down.

So this is less about expecting `start()` to be called twice in normal operation, and more about making both sides robust enough that mistakes fail less mysteriously.

dain:
> I like the idea that product tests can be directly invoked from the IDE. This is nice quality of life improvement.
> 
> However, the reason why existing product tests don't do that is not incidental. AFAIR, we never managed to have tests runnable from IDE to perform Kerberos authentication or access HDFS datanodes correctly. This might be a blocker, i.e. those tests may not be migrate-able to the new approach. @dain , is the intention here that we will have two frameworks side by side going forward?

This concern is real, and I agree it is exactly the kind of thing that would make the port not 
worth landing if it forced us into two long-term frameworks.

The answer is that yes, these cases have been ported in the full branch, and the intent is not
to keep the old framework around permanently. The branch for that work is `user/dain/jpt/all`
if you want to inspect the harder environments directly.

For the historically tricky hostname-sensitive paths, the general model is to keep the services 
inside the Testcontainers network, expose the necessary entrypoints to the host-side test JVM,
and then preserve client validation semantics explicitly where the protocol requires it. In 
practice that means host-side Kerberos config for Kerberos cases, explicitly hooking host-side
client DNS resolution where internal service names still need to resolve correctly, and
truststore/certificate validation for TLS-sensitive clients instead of falling back to insecure
connections.

I do not want to overclaim that everything is perfect, but these environments are not being 
hand-waved away. They have been ported and exercised in the full branch specifically because 
this concern is a real one.

testing/trino-product-tests/src/test/java/io/trino/tests/product/mariadb/TestMariaDb.java
findepi: > I do think adding an equivalent is possible and worthwhile (I played with this in a branch), but I would prefer not to take that on during the port because it is orthogonal to the framework port itself

it's not unimportant and it's not orthogonal, how would it be?

dain: I do not think I said it was unimportant, and that was not what I meant.

What I meant by "orthogonal" is that I do not think a true `--attach` equivalent is required to preserve the semantics or coverage of the product-test port itself.

I agree a real equivalent is possible. I have also experimented with it. The reason I do not want to take it on here is that doing it cleanly adds another environment-management mode with its own complexity, and I would prefer not to expand the scope of a very large port in that direction while it is still under review.

So I am not dismissing the idea. I am just trying to keep this PR focused on landing the ported tests and the framework they require. If there is appetite for a true `--attach` equivalent after that, I think it is a reasonable follow-up.

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/environment/ProductTestEnvironment.java
findepi: > `EnvironmentManager` should call `start()` correctly, but I also want environment implementations to behave safely if that contract is violated accidentally. The goal is to avoid the kind of rare lifecycle issue that only shows up in CI and then takes a long time to track down.

Sounds like AI. It's not how we write code in Trino, is it?

dain: I literally wrote the code to do this. Did this on purpose. Exactly what I want.

PR #28633, opened 2026-03-12
Fix casting exception when writing value to IntegerType or RealType

gertjanal:
## Description
```
writeNativeValue(INTEGER, 2);
java.lang.ClassCastException: class java.lang.Integer cannot be cast to class java.lang.Long

writeNativeValue(REAL, 2f);
java.lang.ClassCastException: class java.lang.Float cannot be cast to class java.lang.Long
```

## Release notes

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/type/TypeUtils.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/type/TypeUtils.java
@@ -45,6 +45,12 @@ public final class TypeUtils
         if (block.isNull(position)) {
             return null;
         }
+        if (type instanceof IntegerType integerType) {
+            return integerType.getInt(block, position);
+        }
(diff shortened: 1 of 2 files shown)


raunaqmorarka:
This is working as intended, you're supposed to use
```
writeNativeValue(INTEGER, 2L);

writeNativeValue(REAL, (long) floatToRawIntBits(2f));
```

gertjanal:
But now in another code project I have something like
```java
if (value instanceof Float f && type instanceof RealType realType) {
    realType.writeFLoat(builder, f);
}
else if (value instanceof Integer && type instanceof IntegerType integerType) {
    integerType.writeInt(builder, i);
}
else {
    writeNativeValue(type, builder, value);
}
```
And I don't think I'm the only one with this if-else instead of writing just writeNativeValue
So to me it felt like this would be a good addition

PR #28634, opened 2026-03-12
Close Response InputStream

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/ui/WorkerResource.java
+++ b/core/trino-main/src/main/java/io/trino/server/ui/WorkerResource.java
@@ -44,6 +44,7 @@ import jakarta.ws.rs.core.HttpHeaders;
 import jakarta.ws.rs.core.Response;
 
 import java.io.IOException;
+import java.io.InputStream;
 import java.util.HashSet;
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/server/ui/WorkerResource.java
```diff
@@ -221,11 +222,11 @@ public byte[] handleException(Request request, Exception exception)
         @Override
         public byte[] handle(Request request, io.airlift.http.client.Response response)
         {
-            try {
+            try (InputStream stream = response.getInputStream()) {
```
findepi, line 225: 1. `io.airlift.http.client.Response.getInputStream` has no javadoc clarifying whether the stream is supposed to be closed (who owns it).
2. In fact, the implementation returns a stream from a field, so it's clearly owned by the `Response` object, or someone who created the response. 
   If we didn't call the `Response.getInputStream` method, we wouldn't close the stream, but the stream still needs to be closed.
3. The `Response` itself isn't closeable, so it surely doesn't own the input stream, which is closeable


So, the stream should be closed by the airlift http client.

Issue #859, opened 2019-05-31
Many test failure for TestQueuesDb


electrum:
We frequently see variants of this.

```
[ERROR] tearDown(io.prestosql.execution.resourcegroups.db.TestQueuesDb)  Time elapsed: 4.838 s  <<< FAILURE!
java.lang.RuntimeException: java.lang.reflect.InvocationTargetException
	at io.prestosql.server.testing.TestingPrestoServer.close(TestingPrestoServer.java:323)
	at com.google.common.io.Closer.close(Closer.java:214)
	at io.prestosql.tests.DistributedQueryRunner.close(DistributedQueryRunner.java:432)
	at io.prestosql.execution.resourcegroups.db.TestQueuesDb.tearDown(TestQueuesDb.java:89)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:104)
	at org.testng.internal.Invoker.invokeConfigurationMethod(Invoker.java:515)
	at org.testng.internal.Invoker.invokeConfigurations(Invoker.java:217)
	at org.testng.internal.Invoker.invokeMethod(Invoker.java:712)
	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:851)
	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1177)
	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.lang.reflect.InvocationTargetException
	at sun.reflect.GeneratedMethodAccessor1956.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
[...]

PR #28643, opened 2026-03-13
Fix TestQueues flakiness: cancel queries after assertion

p3tr0sz:
## Description
`TestQueues.testSelectorResourceEstimateBasedSelection` or `TestQueues.testQueryTypeBasedSelection` (depending on the run) intermittently times out after 240 seconds waiting for a query to reach RUNNING state.

`assertResourceGroup` creates queries but never cancels them. When running the full `TestQueues` class, earlier test methods spin up a full `DistributedQueryRunner`. By the time `testSelectorResourceEstimateBasedSelection` runs, its 5 sequential `assertResourceGroup` calls accumulate 5 concurrent `SELECT COUNT(*) FROM lineitem` queries on `sf100000` against a single-node cluster. Under JVM pressure, this leads to thread pool saturation (RejectedExecutionException), preventing later queries from progressing through their state machine.

## Release notes

```markdown
## Section
```

### DIFF
--- a/testing/trino-tests/src/test/java/io/trino/execution/QueryRunnerUtil.java
+++ b/testing/trino-tests/src/test/java/io/trino/execution/QueryRunnerUtil.java
@@ -27,6 +27,7 @@ import java.util.Set;
 
 import static io.airlift.concurrent.MoreFutures.getFutureValue;
 import static io.trino.execution.QueryState.RUNNING;
+import static java.lang.String.format;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
 public final class QueryRunnerUtil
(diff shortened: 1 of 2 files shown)


findepi:
Does this fix any of https://github.com/trinodb/trino/issues?q=is%3Aissue%20state%3Aopen%20TestQueues ?

p3tr0sz:
> Does this fix any of https://github.com/trinodb/trino/issues?q=is%3Aissue%20state%3Aopen%20TestQueues ?

https://github.com/trinodb/trino/issues/859 is pretty old issue, but it looks like it is related. The comment from the author of the issue is:
> We see similar failures for TestQueues, in the query runner close() method called from the try-with-resources black. This stack trace seems to indicate that a query is still running when the query runner is closed.

Codebase evolved since then and original issue is about `TestQueuesDb`, but I'm addressing the `TestQueues` root cause described in the ticket as `assertResourceGroup` method wasn't canceling queries after assertions, leaving them running during teardown.

The second issue #26567 I think is is unrelated.

PR #28648, opened 2026-03-14
Fix CREATE OR REPLACE TABLE partition name conflict

### LINKED ISSUE #25339: Issues in Iceberg CREATE OR REPLACE TABLE operation
I've met 2 strange issues in the behavior of CREATE OR REPLACE TABLE operation on Iceberg catalog.
I've tried and confirmed them on Trino 471 release. 

## Issue 1. Bucketing
I'm creating table, replacing it with another bucketing number and trying to read $files metadata table.

```
create table bucketing_issue (a int, b varchar) WITH (partitioning = ARRAY['bucket(a, 10)']);

insert into bucketing_issue values (1, 'val_b');

create or REPLACE table bucketing_issue WITH (partitioning = ARRAY['bucket(a, 11)']) as select 2 a, 'val_b' b;

select * from "bucketing_issue$files";
```
Getting error
```
SQL Error [65536]: Query failed (#20250318_075342_27814_rkqhc): Invalid schema: multiple fields for name partition.a_bucket: 1000 and 1001
```

## Issue 2. Changing column type
I'm creating table with ```int``` column and trying to replace it with ```array(int)``` column
```
drop table if exists types_change_issue;
create table types_change_issue (a int);
create or replace table types_change_issue (a array(int));
```
Getting error
```
SQL Error [84148236]: Query failed (#20250318_080113_27852_rkqhc): Failed to commit the transaction during create table: The following columns have types incompatible with the existing columns in their respective positions :
a
Cannot determine whether the commit was successful or not, the underlying data files may or may not be needed. Manual intervention via the Remove Orphan Files Action can remove these files when a connection to the Catalog can be re-established if the commit was actually unsuccessful.
Please check to see whether or not your commit was successful before retrying this commit. Retrying an already successful operation will result in duplicate records or unintentional modifications.
At this time no files will be deleted including possibly unused manifest lists.
```

Documentation says nothing about such a behavior. Can such limitations be fixed or documented at least?

laserninja:
## Description

When replacing a table with a different bucket count (e.g., `bucket(a, 10)` -> `bucket(a, 11)`), `parsePartitionFields()` was called without the existing partition fields. Both the old and new specs ended up with the same name `a_bucket`, causing a duplicate field error when reading `$files`.

Fix by passing existing partition fields from the original table to `parsePartitionFields()` so the width-suffix deduplication logic in `PartitionFields` applies correctly (e.g., the new spec gets named `a_bucket_11` instead of `a_bucket`).

Fixes #25339 (1st part)

## Release notes

```markdown
## Iceberg connector
* Fix partition name conflict in `CREATE OR REPLACE TABLE` when changing bucket count. ({issue}`25339`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -1457,6 +1457,8 @@ public class IcebergMetadata
                 }
                 validateNotModifyingOldSnapshot(table, icebergTable);
                 tableLocation = icebergTable.location();
+                List<PartitionField> existingPartitionFields = getAllPartitionFields(icebergTable);
+                transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties, existingPartitionFields);
             }
         }
 
@@ -1464,7 +1466,10 @@ public class IcebergMetadata
             tableLocation = getTableLocation(tableMetadata.getProperties())
                     .orElseGet(() -> catalog.defaultTableLocation(session, tableMetadata.getTable()));
         }
-        transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties);
+
+        if (transaction == null) {
+            transaction = newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties, ImmutableList.of());
+        }
         Location location = Location.of(transaction.table().location());
         try {
             // S3 Tables internally assigns a unique location for each table
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
@@ -908,11 +908,18 @@ public final class IcebergUtil
                 .toList();
     }
 
-    public static Transaction newCreateTableTransaction(TrinoCatalog catalog, ConnectorTableMetadata tableMetadata, ConnectorSession session, boolean replace, String tableLocation, Predicate<String> allowedExtraProperties)
+    public static Transaction newCreateTableTransaction(
+            TrinoCatalog catalog,
+            ConnectorTableMetadata tableMetadata,
+            ConnectorSession session,
+            boolean replace,
+            String tableLocation,
+            Predicate<String> allowedExtraProperties,
+            List<PartitionField> existingPartitionFields)
     {
         SchemaTableName schemaTableName = tableMetadata.getTable();
         Schema schema = schemaFromMetadata(tableMetadata.getColumns());
-        PartitionSpec partitionSpec = parsePartitionFields(schema, getPartitioning(tableMetadata.getProperties()));
+        PartitionSpec partitionSpec = parsePartitionFields(schema, getPartitioning(tableMetadata.getProperties()), existingPartitionFields);
         SortOrder sortOrder = parseSortFields(schema, getSortOrder(tableMetadata.getProperties()));
 
         Transaction transaction;
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -7722,6 +7722,29 @@ public abstract class BaseIcebergConnectorTest
     }
 
     @Test
+    public void testCreateOrReplaceTableChangePartitionBucketCount()
+    {
+        try (TestTable table = newTrinoTable("test_create_or_replace_", "(a int, b varchar) WITH (partitioning = ARRAY['bucket(a, 10)'])")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (1, 'v1')", 1);
+            long v1SnapshotId = getCurrentSnapshotId(table.getName());
+
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " WITH (partitioning = ARRAY['bucket(a, 11)']) AS SELECT 2 a, 'v2' b", 1);
+            assertThat(query("SELECT a, b FROM " + table.getName()))
+                    .matches("VALUES (2, CAST('v2' AS VARCHAR))");
+
+            // Reading $files must not throw "Invalid schema: multiple fields for name partition.*_bucket"
+            assertThat(query("SELECT count(*) FROM \"" + table.getName() + "$files\""))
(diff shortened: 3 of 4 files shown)


chenjian2664:
Please drop the 2nd commit and remove the changes from release notes of 479, we will add the rn in 480 if the pr finally land

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -902,10 +900,15 @@ public static List<ViewColumn> viewColumnsFromSchema(TypeManager typeManager, Sc
     }
 
     public static Transaction newCreateTableTransaction(TrinoCatalog catalog, ConnectorTableMetadata tableMetadata, ConnectorSession session, boolean replace, String tableLocation, Predicate<String> allowedExtraProperties)
+    {
+        return newCreateTableTransaction(catalog, tableMetadata, session, replace, tableLocation, allowedExtraProperties, ImmutableList.of());
+    }
+
+    public static Transaction newCreateTableTransaction(TrinoCatalog catalog, ConnectorTableMetadata tableMetadata, ConnectorSession session, boolean replace, String tableLocation, Predicate<String> allowedExtraProperties, List<PartitionField> existingPartitionFields)
```
chenjian2664: Consider placing each parameter on a separate line

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -7696,6 +7627,26 @@ public void testCreateOrReplaceTableChangeColumnNamesAndTypes()
...
+            long v1SnapshotId = getCurrentSnapshotId(table.getName());
+
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " WITH (partitioning = ARRAY['bucket(a, 11)']) AS SELECT 2 a, 'v2' b", 1);
+            assertThat(query("SELECT a, b FROM " + table.getName()))
+                    .matches("VALUES (2, CAST('v2' AS VARCHAR))");
+
+            // Reading $files must not throw "Invalid schema: multiple fields for name partition.*_bucket"
+            assertThat(query("SELECT count(*) FROM \"" + table.getName() + "$files\""))
+                    .matches("VALUES BIGINT '1'");
+
```
chenjian2664, line 7738: Add a check for table partitioning:
```
            assertThat((String) computeScalar("SHOW CREATE TABLE " + table.getName()))
                    .contains("partitioning = ARRAY['bucket(a, 11)']");
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
chenjian2664, line 1446: we don't need to touch the exists logic, instead adding a check before `transaction` below
```
if (transaction == null) {
}
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -7696,6 +7627,26 @@ public void testCreateOrReplaceTableChangeColumnNamesAndTypes()
         }
     }
 
+    @Test
+    public void testCreateOrReplaceTableChangeBucketCount()
```
chenjian2664: rename to `testCreateOrReplaceTableChangePartitionBucketCount`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
laserninja: done

ebyhr:
/test-with-secrets sha=f04446a9c46ef4482f55779ad61670e27ef203c4

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -908,11 +899,29 @@ public static List<ViewColumn> viewColumnsFromSchema(TypeManager typeManager, Sc
                 .toList();
     }
 
-    public static Transaction newCreateTableTransaction(TrinoCatalog catalog, ConnectorTableMetadata tableMetadata, ConnectorSession session, boolean replace, String tableLocation, Predicate<String> allowedExtraProperties)
+    public static Transaction newCreateTableTransaction(
```
ebyhr: There're only 2 usages of this method. Could you migrate them instead of having another method?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -7721,6 +7627,29 @@ public void testCreateOrReplaceTableChangeColumnNamesAndTypes()
         }
     }
 
+    @Test
+    public void testCreateOrReplaceTableChangePartitionBucketCount()
+    {
+        try (TestTable table = newTrinoTable("test_create_or_replace_", " (a int, b varchar) WITH (partitioning = ARRAY['bucket(a, 10)'])")) {
```
ebyhr:
```suggestion
        try (TestTable table = newTrinoTable("test_create_or_replace_", "(a int, b varchar) WITH (partitioning = ARRAY['bucket(a, 10)'])")) {
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
laserninja: done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
laserninja: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
laserninja: done

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
laserninja: Didn't notice there were only two. Done.

ebyhr:
/test-with-secrets sha=d83b82293e8671f15f0fe612909a53b72fea3d0e

findepi:
> Fixes #25339 (1st part)

should we reopen?

PR #28649, opened 2026-03-14
Fix CREATE OR REPLACE TABLE type change on HMS

### LINKED ISSUE #25339: Issues in Iceberg CREATE OR REPLACE TABLE operation

laserninja:
When replacing a table using a Hive Metastore-backed catalog and the new column type is incompatible with the existing Hive type (e.g., INT -> ARRAY(INT)), HMS replaceTable() rejects the change with a type-mismatch error. This is correct for a regular ALTER TABLE but not for CREATE OR REPLACE TABLE, which should allow any schema change.

Fix by adding a createOrReplace flag to IcebergTableOperations. When this flag is set, commitToExistingTable() uses a drop-and-recreate strategy instead of replaceTable(), bypassing Hive column type-compatibility checks while preserving all Iceberg data files.

Fixes #25339 part 2

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
@@ -35,6 +35,7 @@ import java.util.function.BiFunction;
 
 import static com.google.common.base.Preconditions.checkState;
 import static io.trino.metastore.PrincipalPrivileges.NO_PRIVILEGES;
+import static io.trino.metastore.Table.TABLE_COMMENT;
 import static io.trino.plugin.hive.metastore.thrift.ThriftMetastoreUtil.fromMetastoreApiTable;
 import static io.trino.plugin.iceberg.IcebergTableName.tableNameFrom;
 import static io.trino.plugin.iceberg.IcebergUtil.fixBrokenMetadataLocation;
@@ -74,9 +75,23 @@ public class HiveMetastoreTableOperations
     protected void commitToExistingTable(TableMetadata base, TableMetadata metadata)
     {
         Table currentTable = getTable();
-        commitTableUpdate(currentTable, metadata, (table, newMetadataLocation) -> Table.builder(table)
-                .apply(builder -> updateMetastoreTable(builder, metadata, newMetadataLocation, Optional.of(currentMetadataLocation)))
-                .build());
+        // Deliberately do NOT call setDataColumns() here. For Iceberg tables, HMS column
+        // definitions are not used for reads — the real schema lives in the Iceberg metadata
+        // file. Preserving the existing Hive column types avoids HMS type-compatibility
+        // validation, which rejects any column type changes (e.g., INT -> ARRAY(INT)) that
+        // are valid for CREATE OR REPLACE TABLE or for subsequent writes after such a replace.
+        commitTableUpdate(currentTable, metadata, (table, newMetadataLocation) -> {
+            Table.Builder builder = Table.builder(table)
+                    .withStorage(storage -> storage.setLocation(metadata.location()))
+                    .setParameter(METADATA_LOCATION_PROP, newMetadataLocation)
+                    .setParameter(PREVIOUS_METADATA_LOCATION_PROP, Optional.of(currentMetadataLocation))
+                    .setParameter(TABLE_COMMENT, Optional.ofNullable(metadata.properties().get(TABLE_COMMENT)));
+            if (metadata.currentSnapshot() != null) {
+                builder.setParameter(CURRENT_SNAPSHOT_ID, String.valueOf(metadata.currentSnapshot().snapshotId()))
+                        .setParameter(CURRENT_SNAPSHOT_TIMESTAMP, String.valueOf(metadata.currentSnapshot().timestampMillis()));
+            }
+            return builder.build();
+        });
     }
 
     @Override
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/hms/TestIcebergHiveCatalogWithoutLock.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/hms/TestIcebergHiveCatalogWithoutLock.java
@@ -67,6 +67,39 @@ final class TestIcebergHiveCatalogWithoutLock
     }
 
     @Test
+    void testCreateOrReplaceTableWithIncompatibleColumnType()
+    {
+        // Regression test for https://github.com/trinodb/trino/issues/25339
+        // HMS replaceTable() validates Hive column type compatibility and rejects incompatible
+        // changes like INT -> ARRAY(INT). CREATE OR REPLACE TABLE should allow any type change.
+        try (TestTable table = newTrinoTable("test_create_or_replace_type_", "(a int)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES 1", 1);
+
+            // CTAS form: Changing column type from INT to ARRAY(INT) must succeed
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " AS SELECT ARRAY[2] a", 1);
+            assertThat(query("SELECT a FROM " + table.getName()))
+                    .matches("VALUES ARRAY[2]");
+        }
+    }
+
+    @Test
+    void testCreateOrReplaceTableWithIncompatibleColumnTypeDdlForm()
+    {
+        // Regression test for https://github.com/trinodb/trino/issues/25339 (DDL form)
+        // Verify that CREATE OR REPLACE TABLE with explicit column definitions also works
+        // when the new column type is incompatible with the existing Hive column type.
+        try (TestTable table = newTrinoTable("test_create_or_replace_ddl_", "(a int)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES 1", 1);
+
+            // DDL form: CREATE OR REPLACE TABLE with explicit incompatible column type must succeed
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " (a array(int))");
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES (ARRAY[2])", 1);
+            assertThat(query("SELECT a FROM " + table.getName()))
+                    .matches("VALUES ARRAY[2]");
+        }
+    }
+
+    @Test
     void testCommitWithoutLock()
     {
         try (TestTable table = newTrinoTable("test_lock", "(x int)", List.of("1", "2", "3"))) {


ebyhr:
Could you check `TestLakehouseConnectorTest` failures?

laserninja:
@ebyhr can you check please

docs/src/main/sphinx/release/release-479.md
```diff
@@ -109,6 +109,8 @@
 * Add `APPLICATION_DEFAULT` authentication type for GCS. ({issue}`26984`)
 * {{breaking}} Remove support for unauthenticated access when GCS authentication type is set to `SERVICE_ACCOUNT`. ({issue}`26984`)
 * Reduce memory pressure when the table contains highly nested fields. ({issue}`25077`)
+* Fix `CREATE OR REPLACE TABLE` failure when changing a column to an incompatible
+  type with a Hive Metastore-backed catalog. ({issue}`25339`)
```
Praveen2112: Do we need to update on older version release notes ?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/hms/TestIcebergHiveCatalogWithoutLock.java
```diff
@@ -66,6 +66,22 @@ protected QueryRunner createQueryRunner()
...
+    {
+        // Regression test for https://github.com/trinodb/trino/issues/25339
+        // HMS replaceTable() validates Hive column type compatibility and rejects incompatible
+        // changes like INT -> ARRAY(INT). CREATE OR REPLACE TABLE should allow any type change.
+        try (TestTable table = newTrinoTable("test_create_or_replace_type_", "(a int)")) {
+            assertUpdate("INSERT INTO " + table.getName() + " VALUES 1", 1);
+
+            // Changing column type from INT to ARRAY(INT) must succeed
+            assertUpdate("CREATE OR REPLACE TABLE " + table.getName() + " AS SELECT ARRAY[2] a", 1);
+            assertThat(query("SELECT a FROM " + table.getName()))
```
Praveen2112, line 80: But performing an INSERT query the CREATE OR REPLACE TABLE query still fails.

docs/src/main/sphinx/release/release-479.md
laserninja: Done

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/hms/TestIcebergHiveCatalogWithoutLock.java
laserninja: Added a dedicated test testCreateOrReplaceTableWithIncompatibleColumnTypeDdlForm() that covers exactly this case: creates a table with (a int), inserts a row, does CREATE OR REPLACE TABLE ... (a array(int)) (DDL form), then verifies an INSERT and SELECT work. Both CTAS and DDL forms go through the same commitTableReplacement() path, so the fix applies to both.

docs/src/main/sphinx/release/release-481.md
```diff
@@ -1,146 +1,6 @@
...
-* Reduce memory usage when writing files to S3. ({issue}`28488`)
-* Fix failure when creating tables in the Hive metastore version 3.1. ({issue}`28798`)
-* Fix failure when parsing bucket numbers on tables backed by files without a bucket name. ({issue}`28632`)
-
-## Hudi connector
-
-* Add support for configuring the Azure connection-pool idle time and HTTP request timeout via the `azure.connection-pool-max-idle-time` and `azure.http-request-timeout` configuration properties. ({issue}`29284`)
+# Release 481 (Not released yet)
 
 ## Iceberg connector
```
findinpath, line 81: Remove this part - it is done separately in a different PR
Just update the  Release Notes section in the description of the PR

docs/src/main/sphinx/release.md
findinpath, line 9: remove

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
```diff
@@ -74,9 +75,23 @@ public HiveMetastoreTableOperations(
     protected void commitToExistingTable(TableMetadata base, TableMetadata metadata)
     {
         Table currentTable = getTable();
```
findinpath, line 77: https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/AbstractMetastoreTableOperations.java#L139


Is there a benefit of filling in the schema of the Iceberg table in HMS if it is not actually used?
In case of MV storage tables we just fill in a dummy column.
Why can't we do the same for all the tables?

cc @ebyhr @chenjian2664

laserninja: The MV dummy-column approach is used for VIRTUAL_VIEW-type entries in HMS, where the columns are entirely irrelevant. For regular EXTERNAL_TABLE Iceberg tables, syncing real column definitions to HMS was likely done to support external tooling that reads HMS metadata (e.g., Hive's DESCRIBE, schema-sync tools, governance tooling).

This PR scopes the fix conservatively: initial table creation (commitNewTable) still writes real columns; commitToExistingTable no longer updates them. This avoids breaking existing tooling that might read those initial HMS columns, while fixing the type-compatibility rejection on subsequent commits.

If the broader approach (always write a dummy column for Iceberg tables) is preferred, I'm happy to do that instead, but wanted to confirm the scope with reviewers before touching commitNewTable as well. @ebyhr @chenjian2664, do you have a preference?

laserninja:
@findinpath can you review this please

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
```diff
@@ -74,9 +75,23 @@ public HiveMetastoreTableOperations(
...
-        commitTableUpdate(currentTable, metadata, (table, newMetadataLocation) -> Table.builder(table)
-                .apply(builder -> updateMetastoreTable(builder, metadata, newMetadataLocation, Optional.of(currentMetadataLocation)))
-                .build());
+        // Deliberately do NOT call setDataColumns() here. For Iceberg tables, HMS column
+        // definitions are not used for reads — the real schema lives in the Iceberg metadata
+        // file. Preserving the existing Hive column types avoids HMS type-compatibility
+        // validation, which rejects any column type changes (e.g., INT -> ARRAY(INT)) that
+        // are valid for CREATE OR REPLACE TABLE or for subsequent writes after such a replace.
+        commitTableUpdate(currentTable, metadata, (table, newMetadataLocation) -> {
```
chenjian2664, line 83: Is the only difference from the original `updateMetadatastoreTable` that it skips setting `dataColumns`? If so, I'd suggest extending the existing method (e.g., with a `skipDataColumns` flag) instead of creating a separate implementation

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
```diff
@@ -74,9 +75,23 @@ public HiveMetastoreTableOperations(
     protected void commitToExistingTable(TableMetadata base, TableMetadata metadata)
     {
         Table currentTable = getTable();
-        commitTableUpdate(currentTable, metadata, (table, newMetadataLocation) -> Table.builder(table)
-                .apply(builder -> updateMetastoreTable(builder, metadata, newMetadataLocation, Optional.of(currentMetadataLocation)))
-                .build());
+        // Deliberately do NOT call setDataColumns() here. For Iceberg tables, HMS column
```
chenjian2664, line 78: I don't see the `createOrReplace` flag you've mentioned in the pr description, am I missing sth?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/hms/TestIcebergHiveCatalogWithoutLock.java
```diff
@@ -66,6 +66,39 @@ protected QueryRunner createQueryRunner()
                 .build();
     }
 
+    @Test
+    void testCreateOrReplaceTableWithIncompatibleColumnType()
```
chenjian2664, line 70: Could you add tests covering both `ALTER TABLE` and `CREATE OR REPLACE TABLE` :
* A table with multiple complex columns, where ALTER TABLE drops one of the complex columns.
* A table with multiple complex columns, where ALTER TABLE adds a regular column.
* and what if the `ALTER TABLE` rename the complex column?
* The same two scenarios using CREATE OR REPLACE TABLE.

chenjian2664 (commented):
This PR seems to skip setting dataColumns when committing the table. Are (HMS) `dataColumns` no longer used by any code path?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
chenjian2664: Does that mean the table schema will become inconsistent with the HMS schema? If so, is that harmless?

laserninja: Yes, after a type-incompatible replace, the HMS dataColumns can lag the Iceberg schema. It's harmless for Trino because Iceberg reads/writes never consult HMS columns. The trade-off is that purely HMS-based external tooling may see a stale schema; that already happens for any Iceberg schema evolution and is preferable to failing the operation outright.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/hms/HiveMetastoreTableOperations.java
laserninja: Ahh yes, the description is stale. The flag-based approach was replaced with this simpler one (never update HMS data columns on commit). I'll update the PR description accordingly.

laserninja:
> Are (HMS) dataColumns no longer used by any code path?

They're still written on initial table creation (commitNewTable → updateMetastoreTable with skipDataColumns=false), so external tooling that reads the initial HMS schema still works. They are not used by Trino's Iceberg read/write paths — the authoritative schema is the Iceberg metadata file. We only skip updating them on subsequent commits to avoid HMS's type-compatibility validation rejecting valid Iceberg schema changes.

PR #28650, opened 2026-03-14
Add getExecutingQueries to QueryTracker to filter active non-finishing queries

jinyangli34:
## Description
Move getState() to TrackedQuery interface and introduce getExecutingQueries() which filters out done/finishing queries and those without an execution start time. Update QueryManager enforce* methods to use the new filtered view instead of iterating all queries with manual guards.

seeing query monitor checking queries already finished.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/QueryTracker.java
+++ b/core/trino-main/src/main/java/io/trino/execution/QueryTracker.java
@@ -25,6 +25,7 @@ import org.weakref.jmx.Managed;
 
 import java.time.Instant;
 import java.util.Collection;
+import java.util.List;
 import java.util.NoSuchElementException;
 import java.util.Optional;
 import java.util.Queue;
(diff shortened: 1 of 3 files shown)


core/trino-main/src/main/java/io/trino/execution/QueryTracker.java
```diff
@@ -142,6 +145,15 @@ public Collection<T> getAllQueries()
         return unmodifiableCollection(queries.values());
     }
 
+    public List<T> getExecutingQueries()
+    {
+        return queries.values().stream()
+                .filter(not(TrackedQuery::isDone))
+                .filter(q -> q.getState() != QueryState.FINISHING)
```
wendigo: Nit: rename q to query

wendigo (approved):

chenjian2664 (approved):

core/trino-main/src/main/java/io/trino/execution/QueryTracker.java
jinyangli34: Updated. Thanks!

PR #28653, opened 2026-03-15
Fix SHOW COLUMNS and information_schema.columns returning only first column per table

### LINKED ISSUE #28630: Trino SHOW COLUMNS only returns the first row
## Problem description

Since upgrading from Trino v477 to Trino v479, the behavior of `SHOW COLUMNS FROM table` has changed:

- In v477, the command show all columns of all tables. Querying the `information_schema.columns` table shows all columns from all tables.
- In v479, the command only shows the first column of the table. Querying the `information_schema.columns` table shows only the first column of each table.

## Expected behaviour

Aligned on v477, the command `SHOW COLUMNS FROM table` show all columns of the `table`. Querying the `information_schema.columns` table shows all columns from the `table`.

sachinsharma3191:
## Problem

- Fixes https://github.com/trinodb/trino/issues/28630

Since upgrading from Trino v477 to v479, `SHOW COLUMNS FROM table` and `information_schema.columns` only return the first column of each table instead of all columns.

## Root Cause

The bug was introduced in the bulk column listing feature (commit 1ac1ee15796 - "Bulk fetch all columns from all tables in JDBC connectors"). 

In `BaseJdbcClient.IterateTableColumns`, when transitioning from one table to the next in the result set, the code called `finishCurrentTable()` and immediately returned. The current row (which contains the first column of the new table) was never processed. On the next `computeNext()` call, `resultSet.next()` would advance past that unprocessed row, causing the first column of every table (except the first) to be skipped.

## Solution

When a table switch is detected, we must ensure the current row (first column of the new table) is processed before returning. By explicitly resetting `currentTable = null` after `finishCurrentTable()`, we ensure the code falls through to the initialization block that sets up the new table and processes the current row - adding the first column to `currentTableColumns` - before the loop exits and returns the previous table's metadata.

## Testing

The fix is covered by `testInformationSchemaColumnsWithMoreTables` in `BaseJdbcConnectorTest`, which tests both `bulkListColumns=false` and `bulkListColumns=true` modes. With the bug, the test would fail when `bulkListColumns=true` because the second table (region) would be missing its first column (regionkey).

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
@@ -444,6 +444,9 @@ public abstract class BaseJdbcClient
                     RemoteTableName nextTable = getRemoteTable(resultSet);
                     if (currentTable != null && !currentTable.equals(nextTable)) {
                         computedNext = finishCurrentTable().orElse(null);
+                        // Reset for new table - we must process the current row (first column of new table)
+                        // before returning, otherwise it would be skipped on the next computeNext() call
+                        currentTable = null;
                     }
 
                     try {
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
@@ -444,9 +444,6 @@ public abstract class BaseJdbcClient
                     RemoteTableName nextTable = getRemoteTable(resultSet);
                     if (currentTable != null && !currentTable.equals(nextTable)) {
                         computedNext = finishCurrentTable().orElse(null);
-                        // Reset for new table - we must process the current row (first column of new table)
-                        // before returning, otherwise it would be skipped on the next computeNext() call
-                        currentTable = null;
                     }
 
                     try {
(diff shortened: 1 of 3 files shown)


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
```diff
@@ -444,6 +444,9 @@ protected RelationColumnsMetadata computeNext()
                     RemoteTableName nextTable = getRemoteTable(resultSet);
                     if (currentTable != null && !currentTable.equals(nextTable)) {
                         computedNext = finishCurrentTable().orElse(null);
+                        // Reset for new table - we must process the current row (first column of new table)
+                        // before returning, otherwise it would be skipped on the next computeNext() call
+                        currentTable = null;
```
findepi: Isn't this responsibility of `finishCurrentTable()` call?

findepi (commented):
Can you please add a regression test?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/BaseJdbcClient.java
sachinsharma3191: removed it

findepi:
I see work here progresses. Would you be able now to provide the information in the associated issue - https://github.com/trinodb/trino/issues/28630#issuecomment-4050279941 ?

findepi:
- https://github.com/trinodb/trino/issues/28630#issuecomment-4765043951

PR #28655, opened 2026-03-15
Add SHOW COLUMNS support for nested ROW fields

### LINKED ISSUE #19131: Better SHOW COLUMNS support for nested data
Currently when running `SHOW COLUMNS FROM my_schema.table_with_nested_fields` on a table containing a complex column (_ARRAY, MAP, ROW_) the result returned is this: 

| Column  | Type.   | Extra | Comment                        |
| ------- | ------- | ----- | ------------------------------ |
| _id     | varchar |       | A unique identifier                                 |
| complex | row(id varchar, name varchar, value bigint) | | This is a complex object with nested fields |

Some data sources such as [Iceberg tables](https://iceberg.apache.org/spec/#schemas) contain field metadata for nested fields as well.

It would be very useful to extend `SHOW COLUMNS` to support listing the columns of nested fields, so that the full schema definition of tables with nested fields could be queried. 

Example to how it could look syntactically:

`SHOW COLUMNS FROM my_schema.table_with_nested_fields.complex`

or

`SHOW COLUMNS FROM my_schema."table_with_nested_fields.complex"`

 to return in this case:
 
| Column  | Type    | Extra | Comment                        |
| ------- | ------- | ----- | -------------------------------|
| id      | varchar |       | Object Id                   |
| name    | varchar |       | Object Name                 |
| value   | bigint  |       | Object Value                |

sachinsharma3191:
- Fixes https://github.com/trinodb/trino/issues/19131

Supports:
- SHOW COLUMNS FROM table.column (2 parts, uses session catalog/schema)
- SHOW COLUMNS FROM catalog.schema.table.column (4 parts)

Shows the nested fields of a ROW type column when the column name is specified after the table name.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
+++ b/core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
@@ -51,6 +51,7 @@ import io.trino.spi.predicate.Domain;
 import io.trino.spi.security.PrincipalType;
 import io.trino.spi.security.TrinoPrincipal;
 import io.trino.spi.session.PropertyMetadata;
+import io.trino.spi.type.RowType;
 import io.trino.spi.type.Type;
 import io.trino.sql.SqlEnvironmentConfig;
 import io.trino.sql.analyzer.AnalyzerFactory;
@@ -127,6 +128,7 @@ import static io.trino.metadata.MetadataUtil.getRequiredCatalogHandle;
 import static io.trino.metadata.MetadataUtil.processRoleCommandCatalog;
 import static io.trino.metadata.PropertyUtil.toSqlProperties;
 import static io.trino.spi.StandardErrorCode.CATALOG_NOT_FOUND;
+import static io.trino.spi.StandardErrorCode.COLUMN_NOT_FOUND;
 import static io.trino.spi.StandardErrorCode.INVALID_COLUMN_PROPERTY;
 import static io.trino.spi.StandardErrorCode.INVALID_DEFAULT_COLUMN_VALUE;
 import static io.trino.spi.StandardErrorCode.INVALID_MATERIALIZED_VIEW_PROPERTY;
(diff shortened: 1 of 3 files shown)


core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -456,7 +456,6 @@ protected Node visitShowColumns(ShowColumns showColumns, Void context)
             Optional<String> nestedFieldName = Optional.empty();
             QualifiedObjectName tableName;
 
-            // Parse table.column or catalog.schema.table.column for nested field display
```
findepi: Why is it improvement?

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -549,9 +549,6 @@ else if (parts.size() == 3 && session.getCatalog().isPresent()) {
                     ordering(ascending("ordinal_position")));
         }
 
-        /**
-         * Handles SHOW COLUMNS for nested ROW field: SHOW COLUMNS FROM table.complex shows fields of ROW column "complex".
```
findepi: This was added in the previous commit and now being removed. Why so?

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
```diff
@@ -620,22 +620,17 @@ private boolean relationExists(QualifiedObjectName name)
...
                 return value.equals(pattern);
             }
-            // Basic fallback: treat % as .* and _ as . for regex-like matching
             String regex = pattern
                     .replace("\\", "\\\\")
                     .replace(".", "\\.")
                     .replace("%", ".*")
                     .replace("_", ".");
-            if (escape.isPresent()) {
-                // Simplified: if escape is present, we'd need proper LIKE semantics
```
findepi: This probably should be a TODO comment with a link to an issue. Or a throw.
Removing the comment (while also keeping the unused escape parameter), isn't IMO improving the codebase.

findepi:
If i understand correctly, `syntax-needs-review` pertains the first commit (`Add SHOW COLUMNS support for nested ROW fields`).

@sachinsharma3191 the code cleanup commits are disputable (added some inline comments). Would you mind dropping them?

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
sachinsharma3191: @findepi The CI/CD pipeline was failing due to 'unused code' warnings; I've removed the redundant logic to clear the build.

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
sachinsharma3191: @findepi It isn’t meant as a product/feature “improvement” in the release-notes sense — that was poor wording on my side. I was only trying to describe **how** we parse `table` vs `table.column` (and the 4-part form). The line was redundant with the code (`nestedFieldName`, `parts`, etc.), so in the **single consolidated commit** it’s removed. No behavioral change from dropping that comment.

core/trino-main/src/main/java/io/trino/sql/rewrite/ShowQueriesRewrite.java
sachinsharma3191: @findepi The Javadoc on `visitShowColumnsNested` was added in one **follow-up** commit and removed in another while I was iterating on feedback and CI. That back-and-forth wasn’t intentional product design — it was churn.

The branch is now **one commit** on `master` without those intermediate cleanup commits. I didn’t re-add the long Javadoc block because the control flow is already visible in `visitShowColumns` (resolve name → optional nested column → `visitShowColumnsNested`). If you want a **short** stable Javadoc here anyway, I can add a single sentence in a follow-up.

PR #28656, opened 2026-03-15
Fix Cassandra INT partition key full scan when IN clause uses consecutive  values

### LINKED ISSUE #28624: Cassandra table has  INT partition keys, specifying consecutive values in IN or OR clauses triggers a full scan
When using the Cassandra connector with partition keys that are INT or BIGINT, specifying multiple consecutive values using the IN clause or OR operator causes them to be expanded into BETWEEN clauses.

However, Cassandra cannot perform partition key filtering on queries using BETWEEN with INT keys.

Consequently, when attempting to retrieve multiple partitions from an INT partition key, the connector retrieves all partitions even in cases where filtering should be possible.

For example, suppose the partition_key column contains values 1, 2, 3, 4, 5.

Queries like `WHERE partition_column = 1` or `WHERE partition_column IN (1, 3, 5)` work fine.
However, queries like `WHERE partition_column IN (1,2,3,4)` are converted to `BETWEEN 1 AND 4`, resulting in a full scan.

sachinsharma3191:
When partition keys are INT or BIGINT, specifying consecutive values in IN (e.g. IN (1,2,3,4)) gets optimized to BETWEEN by SimplifyContinuousInValues. Cassandra cannot use BETWEEN for partition key filtering, causing a full scan.

Expand integer range predicates to discrete values for partition pruning, using IN clause which Cassandra supports. Limit expansion to 1000 values for safety.

Fixes #28624

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-cassandra/src/main/java/io/trino/plugin/cassandra/CassandraPartitionManager.java
+++ b/plugin/trino-cassandra/src/main/java/io/trino/plugin/cassandra/CassandraPartitionManager.java
@@ -58,6 +58,8 @@ public class CassandraPartitionManager
      */
     private static final int MAX_PARTITION_KEY_RANGE_EXPANSION = 1000;
 
+    private static final Set<CassandraType.Kind> INTEGER_PARTITION_KEY_TYPES = ImmutableSet.of(INT, BIGINT, SMALLINT, TINYINT);
+
     private final CassandraSession cassandraSession;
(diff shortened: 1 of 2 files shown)


ebyhr:
Could you please add a test?

sachinsharma3191:
> Could you please add a test?

@ebyhr I added the test please review it

PR #28657, opened 2026-03-15
Fix OPA access denied to list only inaccessible columns

### LINKED ISSUE #28602: OPA Access Control Select From Columns Error message
Hi,  
Right now, Trino shows a list of all table columns when running a `SELECT *` query, instead of only listing the inaccessible columns. Is there a setting that controls this behavior? I’m using OPA for access control.

sachinsharma3191:
## Description

When a user lacks permission for some (but not all) of the columns referenced in a query, the OPA access control's "access denied" message now lists only the inaccessible columns instead of every column of the table. `filterColumns` is used to determine which columns the user can access, and only the denied subset is reported.

- Fixes #28602

## Release notes

```markdown
## OPA
* Only list the inaccessible columns in the access denied error message. ({issue}`28602`)
```

### DIFF
--- a/plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
+++ b/plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
@@ -772,10 +772,6 @@ public sealed class OpaAccessControl
                 OpaQueryInputResource.builder().table(new TrinoTable(table).withProperties(properties)).build());
     }
 
-    /**
-     * When access is denied, reports only the inaccessible columns (not all columns)
-     * by using filterColumns to determine which columns the user can access.
-     */
     private void checkTableAndColumnsOperation(SystemSecurityContext context, String actionName, CatalogSchemaTableName table, Set<String> columns, BiConsumer<String, Set<String>> deny)
     {
         opaHighLevelClient.queryAndEnforce(
--- a/plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
+++ b/plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
@@ -13,6 +13,7 @@
  */
 package io.trino.plugin.opa;
 
+import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSetMultimap;
 import com.google.common.collect.Multimaps;
@@ -58,6 +59,7 @@ import java.util.function.Consumer;
 
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static com.google.common.collect.ImmutableMap.toImmutableMap;
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
 import static com.google.common.collect.Iterables.getOnlyElement;
 import static io.trino.plugin.opa.OpaHighLevelClient.buildQueryInputForSimpleResource;
 import static io.trino.spi.security.AccessDeniedException.denyCreateCatalog;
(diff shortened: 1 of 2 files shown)


plugin/trino-opa/src/main/java/io/trino/plugin/opa/OpaAccessControl.java
```diff
@@ -775,7 +777,15 @@ private void checkTableAndColumnsOperation(SystemSecurityContext context, String
...
                 actionName,
-                () -> deny.accept(table.toString(), columns),
+                () -> {
+                    Set<String> allowedColumns = filterColumns(context, table.getCatalogName(),
+                            ImmutableMap.of(table.getSchemaTableName(), columns))
+                            .getOrDefault(table.getSchemaTableName(), ImmutableSet.of());
+                    Set<String> inaccessibleColumns = columns.stream()
+                            .filter(column -> !allowedColumns.contains(column))
+                            .collect(toImmutableSet());
+                    deny.accept(table.toString(), inaccessibleColumns.isEmpty() ? columns : inaccessibleColumns);
```
kjh0623, line 789: I applied this locally on master and instrumented the OPA requests on the deny path, because `checkTableAndColumnsOperation` is shared by three operations and I wanted to see what the extra `filterColumns` call does for each. With a 2-column table and a deny-all policy:

| operation | OPA requests | resulting message |
|---|---|---|
| `SelectFromColumns` | 1 + 2 | `Cannot select from columns [col_a, col_b] ...` |
| `UpdateTableColumns` | 1 + 2 | `Cannot update columns [col_a, col_b] ...` |
| `CreateViewWithSelectFromColumns` | 1 + 2 | `View owner 'source-user' cannot create view that selects from ...` |

Two things fall out of that:

- For `CreateViewWithSelectFromColumns` the deny lambda is `(tableAsString, _) -> denyCreateViewWithSelect(...)`, so the columns argument is discarded — the computed `inaccessibleColumns` never reaches the message and the extra OPA round trip is pure overhead.
- For `UpdateTableColumns`, `filterColumns` answers SELECT visibility rather than update permission, so the set it produces isn't really "the columns you may not update".

Worth noting for cost as well: in non-batch mode `filterColumns` goes through `parallelFilterFromOpa`, i.e. one request **per column** — I saw `columns:["col_a"]` and `columns:["col_b"]` go out as separate requests. A `SELECT *` denial on a 100-column table would issue 100 extra requests. With `opa.policy.batched-uri` configured it collapses to one. Given @kokosing's note on #28602 about avoiding unnecessary `filterColumns` calls (cost and audit-log noise), it may be worth scoping the new lookup to `SelectFromColumns` only.

_Disclosure, same as on #28602: I used Claude Opus 4.8 to set up the local build and write the instrumentation. The numbers above are from a run on my machine, not model output._

kokosing:
@takezoe Can you please help with review here?

takezoe:
As @kjh0623 mentioned, error message is generated based on the output of `filterColumns` in this PR. Those columns may not be actually denied columns (Even `SelectFromColumns` isn't exactly equivalent to `FilterColumns`). Also, it triggers additional OPA requests.

Showing denied columns in error message wouldn't be straightforward in OPA because it doesn't have a way to get denied columns other than sending a request to check permission per column, or including denined columns in OPA response.

Providing an option to check permission per column for more user-friendly error message might be better in OPA. Users can enable it if they can accept the additional overhead.

kokosing:
I am not sure if we want to provide option for such thing. I don't feel it is justified. I think 2 more calls are justified when done only in case of access denied. Is that a case @takezoe @kjh0623 ?

takezoe:
2 more calls is just in the case of selecting 2 columns. The number of additional calls will be the number of columns to be selected. It might be acceptable because it already happens in some of OPA access control methods, like `filterColumns()`.

Anyway, even in that case, we should check `SelectFromColumns` permission per column rather than using `filterColumns()` when access denied.

Issue #28415, opened 2026-02-23
Test Druid handling of decimal type with negative scale


findepi:
The connector claims support for decimal with negative scale

https://github.com/trinodb/trino/blob/dfd75cf4f92c5bbe4929375e701a31e5d8c548c4/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java#L268

it's not tested and may not work

- see https://github.com/trinodb/trino/issues/28388
- see https://github.com/trinodb/trino/pull/28414

kkauy:
Hi, I’d like to work on this issue.

I’m planning to review the related PostgreSQL issue (#28388) and fix (#28414) first, then use that as a reference to validate and add test coverage for Druid handling of decimal types with negative scale.

Please let me know if there are any constraints, suggested test locations.

kkauy:
I investigated this with an integration test.

Negative-scale decimals are ingested as declared, but when read back through the Druid connector, `DruidJdbcClient.toColumnMapping(...)` sees them as `VARCHAR` (`jdbcType=12`) rather than `DECIMAL/NUMERIC`.

So the negative-scale decimal mapping branch is not reached in this path, and the test fails with `expected decimal(6,0) but was varchar`.

findepi:
obsoleted by 
- https://github.com/trinodb/trino/pull/28754

PR #28659, opened 2026-03-15
Test Druid handling of decimal type with negative scale

kkauy:
## Summary

Add a reproducer test for negative-scale decimal handling in the Druid connector.

The test documents the current behavior: values ingested as `decimal(p, -s)` are read back through the connector as `VARCHAR`, so the decimal mapping branch is not reached.

## Related issue

Related to #28415

### DIFF
--- a/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
+++ b/plugin/trino-druid/src/main/java/io/trino/plugin/druid/DruidJdbcClient.java
@@ -46,7 +46,6 @@ import io.trino.spi.predicate.Range;
 import io.trino.spi.security.ConnectorIdentity;
 import io.trino.spi.type.CharType;
 import io.trino.spi.type.DecimalType;
-import io.trino.spi.type.Decimals;
 import io.trino.spi.type.TimestampType;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.VarcharType;
@@ -88,7 +87,6 @@ import static io.trino.plugin.jdbc.StandardColumnMappings.booleanWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.charWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.dateColumnMappingUsingSqlDate;
 import static io.trino.plugin.jdbc.StandardColumnMappings.dateWriteFunctionUsingSqlDate;
-import static io.trino.plugin.jdbc.StandardColumnMappings.decimalColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultCharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.defaultVarcharColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.doubleColumnMapping;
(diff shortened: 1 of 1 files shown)


kkauy:
Related to #28415

Adds an integration test covering negative-scale decimal handling in the Druid connector.
The test documents the current behavior: values ingested as `decimal(p, -s)` are read back through the connector as `VARCHAR`, so the decimal mapping branch is not reached.

kkauy:
Hi, I submitted the signed CLA to cla@trino.io about two weeks ago. Could someone please check whether it was received and matched to my GitHub account (`kkauy`)? Thanks.

kkauy:
This seems different from #28414.

In this Druid path, the integration test shows the columns come back as `jdbcType=12` (`VARCHAR`), so the DECIMAL/NUMERIC mapping branch is not reached.

This PR adds a reproducer test for #28415. I can adjust the test approach if a mergeable documentation-style version is preferred.

Praveen2112:
But does Druid supports a dedicated decimal type OOB ? In the documentation it was mentioned that decimal is mapped to double.

kkauy:
Good point. Based on the Druid SQL type documentation, `DECIMAL` maps to `DOUBLE` rather than fixed-point decimal semantics.

I will investigate  why this tested path comes back as `jdbcType=12` (`VARCHAR`) instead of `DOUBLE`.  Thanks in advance.

findepi:
> In this Druid path, the integration test shows the columns come back as `jdbcType=12` (`VARCHAR`), so the DECIMAL/NUMERIC mapping branch is not reached.

thanks for checking this!
it might be it's just copy-pasted code. can we remove it then?

kkauy:
That seems reasonable to me.

I checked and this pattern exists in a number of other JDBC connectors as well, so it does look copy-pasted. Based on the current Druid investigation, this branch appears unreachable here.

plugin/trino-druid/src/test/java/io/trino/plugin/druid/TestDruidTypeMapping.java
```diff
@@ -184,6 +185,17 @@ public void testVarchar()
                 .execute(getQueryRunner(), druidCreateAndInsert("test_unbounded_varchar"));
     }
 
+    @Test
+    public void testDecimalNegativeScale()
```
ebyhr: Could you add roundtrip with positive scale, and rename this test to `testDecimal`? 

Also, please move this method above testVarchar.

ebyhr, line 189: It seems Druid maps unsupported types to varchar. The decimal type doesn't exist in (Druid runtime type) https://druid.apache.org/docs/latest/querying/sql-data-types/. 

I tested with random type name (e.g. `xxx`) and the test maps to varchar type instead of failure.

kkauy: Thanks for the clarification.

Since Druid does not expose decimal as a runtime type and unsupported types are mapped to VARCHAR, this test does not validate meaningful behavior.

I'll go ahead and close this PR.

findepi:
We should remove the dead code:
- https://github.com/trinodb/trino/pull/28754

Issue #22533, opened 2024-06-27
Materialized view referencing current date/time not refreshed when underlying tables did not change


findepi:
reproducible with `io.trino.plugin.iceberg.IcebergQueryRunner.DefaultIcebergQueryRunnerMain#main`

```
CREATE MATERIALIZED VIEW mv GRACE PERIOD INTERVAL '3' SECOND AS SELECT CURRENT_TIMESTAMP t FROM nation WHERE nationkey=1;

REFRESH MATERIALIZED VIEW mv;

TABLE mv;
                   t
---------------------------------------
 2024-06-27 21:25:54.269 Europe/Warsaw
```

now, when user waits a few seconds and queries the MV again, they can expect a timestamp not older than current time - 3 seconds, but that's not the case

```
TABLE mv;
                   t
---------------------------------------
 2024-06-27 21:25:54.269 Europe/Warsaw
```

findepi:
cc @alexjo2144 @marton-bod @sopel39

PR #28682, opened 2026-03-17
Detect materialized views with non-deterministic functions as stale

hashhar:
## Description

Materialized views using non-deterministic scalar functions like current_timestamp, current_date, or random() were not detected as stale because the refresh only tracks table function dependencies.

Add a hasNonDeterministicFunctions flag that propagates through the refresh. When set the MV freshness is reported as UNKNOWN, which causes re-execution after the grace period expires to match the existing behavior for table functions.

It covers both current-time AST nodes (CurrentDate, CurrentTime, CurrentTimestamp, LocalTime, LocalTimestamp) and non-deterministic functions via analysis.getResolvedFunctions().

- Fixes https://github.com/trinodb/trino/issues/22533
- Fixes https://github.com/trinodb/trino/issues/28696

## Release notes

```markdown
## Section
```

Submitted from repo to ensure all tests are run.

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergMaterializedViewTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergMaterializedViewTest.java
@@ -42,6 +42,7 @@ import io.trino.spi.function.table.TableFunctionProcessorProvider;
 import io.trino.spi.function.table.TableFunctionProcessorState;
 import io.trino.spi.function.table.TableFunctionSplitProcessor;
 import io.trino.spi.security.ConnectorIdentity;
+import io.trino.spi.security.Identity;
 import io.trino.sql.tree.ExplainType;
 import io.trino.testing.AbstractTestQueryFramework;
 import io.trino.testing.MaterializedRow;
@@ -1001,6 +1002,47 @@ public abstract class BaseIcebergMaterializedViewTest
         assertUpdate("DROP MATERIALIZED VIEW " + mvName2);
         assertUpdate("DROP MATERIALIZED VIEW " + mvName3);
         assertUpdate("DROP MATERIALIZED VIEW " + mvName4);
+        assertUpdate("DROP TABLE " + sourceTableName);
+    }
+
+    @Test
+    public void testMaterializedViewWithSessionScopedExpressions()
+    {
+        String sourceTableName = "source_table_" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + sourceTableName + " (value INTEGER)");
+        assertUpdate("INSERT INTO " + sourceTableName + " VALUES 1", 1);
+
+        // Session-scoped expressions (current_user, current_catalog, current_schema, current_path)
+        // are constants for materialized views (fixed at creation time), so the MV should be FRESH after refresh
+        String mvName = "mv_session_scoped_" + randomNameSuffix();
+        assertUpdate("CREATE MATERIALIZED VIEW " + mvName + " AS SELECT *, current_user AS created_by FROM " + sourceTableName);
+
+        assertFreshness(mvName, "STALE");
+        assertUpdate("REFRESH MATERIALIZED VIEW " + mvName, 1);
+        assertFreshness(mvName, "FRESH");
+        assertQuery("SELECT created_by FROM " + mvName, "VALUES 'user'");
+
+        // A different user querying the MV still sees the original creator's name
+        // because the MV is FRESH and returns cached data
+        Session otherUserSession = Session.builder(getSession())
+                .setIdentity(Identity.ofUser("other_user"))
+                .build();
+        assertQuery(otherUserSession, "SELECT created_by FROM " + mvName, "VALUES 'user'");
+
+        assertUpdate("INSERT INTO " + sourceTableName + " VALUES 2", 1);
+        assertFreshness(mvName, "STALE");
+        // Stale MV still serves cached data from storage
+        assertQuery(otherUserSession, "SELECT created_by FROM " + mvName, "VALUES 'user'");
+        // Refresh by a different user picks up the new row
+        assertUpdate(otherUserSession, "REFRESH MATERIALIZED VIEW " + mvName, 1);
+        assertFreshness(mvName, "FRESH");
+        // TODO https://github.com/trinodb/trino/issues/28738 session-scoped expressions should resolve using
+        // the MV owner's identity during refresh (like the stale inline path does via analyzeView), but currently
+        // the refresh path resolves them from the refreshing user's session.
+        // When fixed, this should be: VALUES ('user'), ('user')
+        assertQuery("SELECT created_by FROM " + mvName, "VALUES ('user'), ('other_user')");
+
+        assertUpdate("DROP MATERIALIZED VIEW " + mvName);
         assertUpdate("DROP TABLE " + sourceTableName);
     }
 


core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
```diff
@@ -687,11 +693,18 @@ private RelationPlan createRefreshMaterializedViewPlan(Analysis analysis)
         List<String> tableFunctions = analysis.getPolymorphicTableFunctions().stream()
                 .map(polymorphicTableFunction -> polymorphicTableFunction.getNode().getName().toString())
                 .collect(toImmutableList());
+        boolean hasNonDeterministicFunctions = analysis.getResolvedFunctions().stream().anyMatch(function -> !function.deterministic())
+                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), CurrentDate.class).isEmpty()
+                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), CurrentTime.class).isEmpty()
+                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), CurrentTimestamp.class).isEmpty()
+                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), LocalTime.class).isEmpty()
+                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), LocalTimestamp.class).isEmpty();
         RefreshMaterializedViewReference writerTarget = new RefreshMaterializedViewReference(
```
hashhar: I'd appreciate a better way to identify (and keep updated) these.

piotrrzysko: Please take a look at `io.trino.sql.analyzer.DeterminismEvaluator`.

findepi:
> Materialized views using non-deterministic scalar functions like current_timestamp, current_date, or random() were not detected as stale because the refresh only tracks table function dependencies.

This is not necessarily a bug.

There is some usefulness to having e.g. `current_timestamp AS refresh_time` column in a MV or using `random()` for sampling the source.

If we consider every MV with such functions to be stale, there is no point in producing materialization. It should never be used[^1]. So we should rather forbid creation of such MV.

cc @martint 


[^1]: except during grace period, but that seems rather odd to me.

hashhar:
that might make sense, I think at a minimum the 2nd commit needs to be there to avoid incorrectly computing an incremental refresh.

findepi:
> that might make sense, I think at a minimum the 2nd commit needs to be there to avoid incorrectly computing an incremental refresh.

That matches my intuition too, but I'd rather clarify desired semantics first, before making the final call.

hashhar:
Sure, I'll discuss with @martint

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -4308,6 +4311,9 @@ public MaterializedViewFreshness getMaterializedViewFreshness(ConnectorSession s
         boolean dependsOnTableFunctions = currentSnapshot
                 .map(snapshot -> Boolean.valueOf(snapshot.summary().getOrDefault(DEPENDS_ON_TABLE_FUNCTIONS, "false")))
                 .orElse(false);
+        boolean dependsOnNonDeterministicFunctions = currentSnapshot
+                .map(snapshot -> Boolean.valueOf(snapshot.summary().getOrDefault(DEPENDS_ON_NON_DETERMINISTIC_FUNCTIONS, "false")))
```
findepi, line 4465: This is not strictly safe from backwards compatibility perspective: the existing MV don't have this property, so while they may depend on random functions, we won't know that.
I think it's a reasonable default though, otherwise new release would invalidate _all_ existing MVs.
If we keep as is, this deserves a code comment.

Can we do better? When we _don't know_ if MV depends on random functions, could the engine help?

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
```diff
@@ -696,11 +691,7 @@ private RelationPlan createRefreshMaterializedViewPlan(Analysis analysis)
                 .map(polymorphicTableFunction -> polymorphicTableFunction.getNode().getName().toString())
                 .collect(toImmutableList());
         boolean hasNonDeterministicFunctions = analysis.getResolvedFunctions().stream().anyMatch(function -> !function.deterministic())
-                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), CurrentDate.class).isEmpty()
-                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), CurrentTime.class).isEmpty()
-                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), CurrentTimestamp.class).isEmpty()
-                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), LocalTime.class).isEmpty()
-                || !ExpressionTreeUtils.extractExpressions(ImmutableList.of(query), LocalTimestamp.class).isEmpty();
+                || containsCurrentTimeFunctions(query);
```
findepi, line 697: current_user, current_catalog, current_schema -- does any of this matter too?

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
```diff
@@ -610,7 +610,9 @@ private RelationPlan getInsertPlan(
         TableStatisticsMetadata statisticsMetadata = metadata.getStatisticsCollectionMetadataForWrite(session, tableHandle.catalogHandle(), tableMetadata.metadata(), false);
 
         if (materializedViewRefreshWriterTarget.isPresent()) {
-            RefreshType refreshType = IncrementalRefreshVisitor.canIncrementallyRefresh(plan.getRoot());
+            RefreshType refreshType = materializedViewRefreshWriterTarget.get().hasNonDeterministicFunctions()
+                    ? RefreshType.FULL
```
findepi, line 610: unless the time unit didn't elapse, eg. current_date and it's still same day -- future optimization
let's add todo + link to issue

findepi:
> Fixes #22533, #28696

1. this will close only the first of two
2. use `-` to make it a list. GH renders link previews within list items.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
hashhar: The engine could but it would penalise all MVs. We do have access to original sql and we could use the analyzer on it (needed to resolve functions, just parsing is insufficient) to detect.
The MVs will "heal" on whenever the next REFRESH runs and it's a one-time cost instead of per-query even for fresh MVs.

I'd opt to not be smarter here.

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
hashhar: great catch. I looked through all AST nodes that don't represent functions and we need to add CurrentUser, CurrentCatalog, CurrentSchema, CurrentPath I think.

I'll do it in a separate first commit in this PR since the change is actually more generally about DeterminismEvaluator than just MVs.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findepi: We can have best of both worlds, if engine and connector cooperate.
If the MV is known whether it dependsOnNonDeterministicFunctions, then it knows, fine.
When it doesn't know, the engine could offer help.

I am fine leaving as is, but let's have a code comment on the why

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
martint: Those shouldn't matter. They are tied to the session (SQL path, etc), which cannot change from run to run.

BTW, we should use a term other than "non-deterministic" when talking about current_timestamp and friends. Currently, in Trino, that term is used to describe per-row evaluation within a query. The current_xxx functions are constant for a given query, so they are only non-deterministic from run to run, not from row to row.

It's important to not mix them up, as the within-query non-determinism is used to decide whether terms in an expression are safe to reorder, for instance.

hashhar: > They are tied to the session (SQL path, etc), which cannot change from run to run.

I don't see how they cannot vary from run to run? Or do you only mean deterministic within a single session? Within a single session the current_timestamp is also non-varying IIRC (constant over a transaction).

martint: A materialized view's current_path, current_user, etc. are supposed to be fixed when the materialized view is created. They should not be affected by the path/user/catalog/schema of the user querying the MV.

findepi: Can they be affected by who refreshes the view?

piotrrzysko:
> > Fixes #22533, #28696
> 
> 1. this will close only the first of two

I think this PR addresses both. We should create a third issue to track optimization for incremental refreshes -- detecting whether materialized data is valid despite non-deterministic expressions.

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
hashhar: Ah you are right. We store those in MV definition and analyzer gets a synthetic session with those filled in.

I think the logical split will need to be a follow up because I'd need guidance on what you are expecting there (and isn't directly related to the correctness fix).

hashhar:
> > > Fixes #22533, #28696
> > 
> > 
> > 
> > 1. this will close only the first of two
> 
> I think this PR addresses both. We should create a third issue to track optimization for incremental refreshes -- detecting whether materialized data is valid despite non-deterministic expressions.

see https://github.com/trinodb/trino/issues/28731

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
martint: They should not be. Anyone with permissions to call REFRESH on the MV should be able to do it, and the result should be the same (it should execute as if the original creator, with the same environment as when it got created, ran it again)

hashhar:
@martint Looks like a pre-existing bug too about session-scoped things.

I see that current_user (and other session-scoped expressions) are stored as literal SQL text in the MV definition (https://github.com/trinodb/trino/blob/6b4690e1c2fac9f05804a19a10f4255585322f85/core/trino-main/src/main/java/io/trino/execution/CreateMaterializedViewTask.java#L118), not
replaced with constants at creation time. This creates an inconsistency in what they resolve to depending on the code path:

- FRESH query path [reads directly from the storage table](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java#L2645) so `current_user` returns
whatever value was materialized during the last refresh without any re-evaluation.

- STALE inline path goes through [`createScopeForMaterializedView`](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java#L2570), passes `view.getRunAsIdentity()` (the MV owner)
through to `analyzeView` which [creates a
`viewSession` with the owner's identity](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java#L5188). So
`current_user` resolves to the **MV owner**.

- REFRESH path uses [`visitRefreshMaterializedView`](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java#L713), re-parses the original SQL but [calls `process(query, scope)` directly](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java#L741) and **no `viewSession` with the owner's identity** is created, so `current_user` resolves to the **refreshing user**.

**The refresh path is the wrong one. It probably should also create a view session with the owner's identity (like the inline path does).**

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
findepi: > BTW, we should use a term other than "non-deterministic" when talking about current_timestamp and friends.

non-idempotent

findepi:
> I think this PR addresses both.

but still doesn't close both, per https://github.com/trinodb/trino/pull/28682#issuecomment-4076647674

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
hashhar: actually doesn't seem to work for REFRESH path, see https://github.com/trinodb/trino/pull/28682#issuecomment-4080571571

hashhar:
I extracted the REFRESH and session-scoped issues into https://github.com/trinodb/trino/issues/28738, WILL NOT ADDRESS them as part of this PR. This PR is focused on the correctness issue + staleness logic.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
hashhar: added a code comment

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
hashhar: added a comment + linked to https://github.com/trinodb/trino/issues/28731

core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
hashhar: marking as resolved since it'll be discussed in separate issue

martint:
The planner changes look good. I haven't reviewed the tests.

hashhar:
Rebased to resolved conflicts. @findepi / @martint can you PTAL at the tests so that we can merge this and move on to the DEFINER semantics bug.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergMaterializedViewTest.java
```diff
@@ -947,6 +951,146 @@ public void testMaterializedViewCreatedFromTableFunctionWithGracePeriod()
...
+        // Stale MV still serves cached data from storage
+        assertQuery(otherUserSession, "SELECT created_by FROM " + mvName, "VALUES 'user'");
+        // Refresh by a different user picks up the new row
+        assertUpdate(otherUserSession, "REFRESH MATERIALIZED VIEW " + mvName, 1);
+        assertFreshness(mvName, "FRESH");
+        // TODO https://github.com/trinodb/trino/issues/28738 session-scoped expressions should resolve using
+        // the MV owner's identity during refresh (like the stale inline path does via analyzeView), but currently
+        // the refresh path resolves them from the refreshing user's session.
+        // When fixed, this should be: VALUES ('user'), ('user')
+        assertQuery("SELECT created_by FROM " + mvName, "VALUES ('user'), ('other_user')");
```
findepi, line 1043: - ( https://github.com/trinodb/trino/issues/28738 )

hashhar:
hit https://github.com/trinodb/trino/issues/23537 before retries

PR #28691, opened 2026-03-17
Reduce code duplication in Iceberg Glue catalog

piotrrzysko:
## Description

Keep `GlueClient` and `GlueMetastoreStats` together, allowing stats recording code to be reused instead of duplicated.

## Release notes

```markdown
## Section
```


piotrrzysko:
We could go even further and use `StatsRecordingGlueClient` in other places that rely on `GlueClient` and `GlueMetastoreStats`, such as `GlueHiveMetastore` and `DeltaLakeGlueMetastoreTableOperations`. However, the latter doesn’t record stats in one place: https://github.com/trinodb/trino/blame/master/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/metastore/glue/DeltaLakeGlueMetastoreTableOperations.java#L61. 
Is it intentional @ebyhr?

ebyhr:
I believe that was an oversight, not intentional.

piotrrzysko:
> We could go even further and use StatsRecordingGlueClient in other places that rely on GlueClient and GlueMetastoreStats, such as GlueHiveMetastore and DeltaLakeGlueMetastoreTableOperations

It turns out that `GlueHiveMetastore` uses significantly more Glue API calls than the Iceberg Glue Catalog. I’m not addressing this right now, but I’m open to feedback.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/glue/StatsRecordingGlueClient.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        stats.getDeleteDatabase().call(() ->
+                glueClient.deleteDatabase(request -> request.name(databaseName)));
+    }
+
+    public void createDatabase(DatabaseInput database)
+    {
+        stats.getCreateDatabase().call(() ->
+                glueClient.createDatabase(request -> request.databaseInput(database)));
+    }
```
ebyhr: Could we move these methods so they appear before the table-related methods?

ebyhr:
/test-with-secrets sha=c006d89d0c1a07da3556df0acd98757393580556

piotrrzysko:
The test that failed seems flaky -- it passes locally. I created an issue:
* https://github.com/trinodb/trino/issues/28693

PR #28697, opened 2026-03-17
Improve estimates for constants

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/cost/ScalarStatsCalculator.java
+++ b/core/trino-main/src/main/java/io/trino/cost/ScalarStatsCalculator.java
@@ -162,7 +162,7 @@ public class ScalarStatsCalculator
             double lowValue = sourceStats.getLowValue();
             double highValue = sourceStats.getHighValue();
 
-            if (isIntegralType(((Expression) node).type())) {
+            if (isIntegralType(node.type())) {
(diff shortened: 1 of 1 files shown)


core/trino-main/src/main/java/io/trino/cost/ScalarStatsCalculator.java
```diff
@@ -135,11 +139,14 @@ else if (node.function().name().equals(builtinFunctionName(ADD)) ||
...
-            if (value instanceof Constant) {
-                return SymbolStatsEstimate.builder()
+            if (value instanceof Constant constant) {
+                SymbolStatsEstimate.Builder builder = SymbolStatsEstimate.builder()
                         .setNullsFraction(0)
-                        .setDistinctValuesCount(1)
-                        .build();
+                        .setDistinctValuesCount(1);
+
+                setConstantSizeEstimate(constant.value(), builder);
```
findepi, line 147: This makes sense from SHOW STATS perspective, but may be misleading from the CBO perspective.

Average row size is used with estimated row count to calculate the data size estimate: `average row * row count = total data`.
However, constants are stored and processed as RLE. At runtime, they occupy same amount of memory irrespective of number of rows.

raunaqmorarka: +1
I wouldn't worry about accurate estimate on constant size, unless there is a use case for it

PR #28698, opened 2026-03-17
Skip encryption for direct exchanges

p3tr0sz:
<!-- Thank you for submitting a pull request! Find more information 
at https://trino.io/development/process.html, 
at https://github.com/trinodb/trino/blob/master/.github/DEVELOPMENT.md 
and contact us on #core-dev in Slack. -->
<!-- Provide an overview for maintainers and reviewers. -->
## Description
This PR makes exchange encryption dynamic - encryption is applied only when data flows through external storage (spooling exchanges), and skipped for direct exchanges even when an external exchange manager is configured.

Previously, the `fault-tolerant-execution-exchange-encryption-enabled` config was a static toggle tied to fault-tolerant execution. This was overly broad: direct (in-memory) exchanges don't need encryption, but were encrypted anyway when the flag was on.

<!-- Provide details that help an engineer who is unfamiliar with this part of the code. -->
### Problem                                                                                                                                                                                                                                                                                                             
                                                          
When exchange encryption is enabled, Trino encrypts all exchanges uniformly - both direct (in-memory, node-to-node) and spooling (written to external storage). Direct exchanges travel over TLS-secured internal communication channels, making the additional encryption redundant CPU overhead.

### Design
Each `OutputBuffer` (writer) and `ExchangeDataSource` (reader) is classified via new interface methods `usesExternalStorage()` - describing whether the exchange writes to external storage. Concrete implementations return a fixed value (true for spooling, false for direct). Lazy wrappers (`LazyOutputBuffer`, `LazyExchangeDataSource`) throw `IllegalStateException` if queried before initialization, enforcing that the method is only called once the exchange type is known - this prevents callers from caching a stale default that could silently flip later.                                                                                                      
                                                                                             
  A central `ExchangeEncryptionKey` helper resolves the effective key:                                                                                                                                                                                                                                             
  ```java                                                                                                                                                                                                                                                                                                          
  if (externalExchange) {
      return session.getExchangeEncryptionKey();                                                                                                                                                                                                                                                                   
  }                                                                                                                                                                                                                                                                                                                
  return Optional.empty(); 
```                                                                                                                                                                                                                                                                                        
The serde layer is unchanged - it still checks whether a key is present. The new logic sits above it.                                                                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                   
**Writer vs reader asymmetry**                                                                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                                                                                                   
Writer-side: `LocalExecutionPlanner` resolves the encryption key at planning time (the `OutputBuffer` is initialized before planning begins). `TaskOutputOperator` defers serializer creation to the first `addInput()` call, since operator construction may happen before the output buffer delegate is set.                                                                                                                                                   
                                                            
Reader-side (ExchangeOperator, Query) defers deserializer creation to the first getOutput()/pollPage() call because LazyExchangeDataSource may not yet be resolved at construction. By the time pollPage() returns data, the source is guaranteed to be resolved.                                                
                                                            
 **Writer-reader consistency**                                                                                                                                                                                                                                                                                        
Consistency is guaranteed by the coordinator's scheduling decision: a SpoolingExchangeOutputBuffer exists because the coordinator scheduled a spooling exchange, which means the reader gets a SpoolingExchangeDataSource. They are two ends of the same pipe and cannot disagree about the exchange type.       

**Alternatives considered**

instanceof checks, callback/resolver patterns, visitor pattern, enum exchange type, and PlanFragment field were all evaluated. boolean methods `usesExternalStorage()` were chosen for simplicity, symmetry between reader and writer, no coupling to encryption concerns, and no coordinator/schema changes needed.

<!-- Mark the appropriate option with an (x). Propose a release note if you can.
More info at https://trino.io/development/process#release-note -->
## Release notes

```markdown
## Section
```

<!--
Rename `fault-tolerant-execution-exchange-encryption-enabled` to `external-exchange-encryption-enabled`. The old name continues to work. Direct node-to-node exchanges now skip encryption automatically, reducing CPU overhead. Spooling exchanges to external storage remain encrypted.
-->

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/ExchangeOperator.java
+++ b/core/trino-main/src/main/java/io/trino/operator/ExchangeOperator.java
@@ -19,6 +19,7 @@ import com.google.errorprone.annotations.ThreadSafe;
 import io.airlift.slice.Slice;
 import io.trino.connector.CatalogHandle;
 import io.trino.exchange.ExchangeDataSource;
+import io.trino.exchange.ExchangeEncryptionKey;
 import io.trino.exchange.ExchangeManagerRegistry;
 import io.trino.exchange.LazyExchangeDataSource;
 import io.trino.execution.TaskId;
@@ -38,6 +39,7 @@ import it.unimi.dsi.fastutil.ints.IntOpenHashSet;
 import it.unimi.dsi.fastutil.ints.IntSet;
 
 import java.util.List;
+import java.util.Optional;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
@@ -119,7 +121,7 @@ public class ExchangeOperator
                     operatorContext,
                     sourceId,
                     exchangeDataSource,
-                    serdeFactory.createDeserializer(driverContext.getSession().getExchangeEncryptionKey().map(Ciphers::deserializeAesEncryptionKey)),
+                    serdeFactory,
                     noMoreSplitsTracker,
                     operatorInstanceId);
             noMoreSplitsTracker.operatorAdded(operatorInstanceId);
@@ -149,30 +151,27 @@ public class ExchangeOperator
     private final OperatorContext operatorContext;
     private final PlanNodeId sourceId;
     private final ExchangeDataSource exchangeDataSource;
-    private final PageDeserializer deserializer;
+    private final PagesSerdeFactory serdeFactory;
     private final NoMoreSplitsTracker noMoreSplitsTracker;
     private final int operatorInstanceId;
 
+    private PageDeserializer deserializer;
     private ListenableFuture<Void> isBlocked = NOT_BLOCKED;
 
     public ExchangeOperator(
             OperatorContext operatorContext,
             PlanNodeId sourceId,
             ExchangeDataSource exchangeDataSource,
-            PageDeserializer deserializer,
+            PagesSerdeFactory serdeFactory,
             NoMoreSplitsTracker noMoreSplitsTracker,
             int operatorInstanceId)
     {
         this.operatorContext = requireNonNull(operatorContext, "operatorContext is null");
         this.sourceId = requireNonNull(sourceId, "sourceId is null");
         this.exchangeDataSource = requireNonNull(exchangeDataSource, "exchangeDataSource is null");
-        this.deserializer = requireNonNull(deserializer, "serializer is null");
+        this.serdeFactory = requireNonNull(serdeFactory, "serdeFactory is null");
         this.noMoreSplitsTracker = requireNonNull(noMoreSplitsTracker, "noMoreSplitsTracker is null");
         this.operatorInstanceId = operatorInstanceId;
-
-        LocalMemoryContext memoryContext = operatorContext.localUserMemoryContext();
-        // memory footprint of deserializer does not change over time
-        memoryContext.setBytes(deserializer.getRetainedSizeInBytes());
 
         operatorContext.setInfoSupplier(exchangeDataSource::getInfo);
     }
@@ -251,6 +250,16 @@ public class ExchangeOperator
         Slice page = exchangeDataSource.pollPage();
         if (page == null) {
             return null;
+        }
+
+        // Lazy initialization is required because the encryption key depends on whether
+        // the exchange uses external storage (spooling) or direct exchange. This is determined
+        // by LazyExchangeDataSource, which resolves the concrete exchange type only after
+        // the first split is delivered — which happens after the operator is created.
+        if (deserializer == null) {
+            Optional<Slice> effectiveKey = ExchangeEncryptionKey.keyFor(operatorContext.getSession(), exchangeDataSource);
+            deserializer = serdeFactory.createDeserializer(effectiveKey.map(Ciphers::deserializeAesEncryptionKey));
+            operatorContext.localUserMemoryContext().setBytes(deserializer.getRetainedSizeInBytes());
         }
 
         Page deserializedPage = deserializer.deserialize(page);
--- a/core/trino-main/src/main/java/io/trino/server/protocol/Query.java
+++ b/core/trino-main/src/main/java/io/trino/server/protocol/Query.java
@@ -33,6 +33,7 @@ import io.trino.client.QueryData;
 import io.trino.client.QueryError;
 import io.trino.client.QueryResults;
 import io.trino.exchange.ExchangeDataSource;
+import io.trino.exchange.ExchangeEncryptionKey;
 import io.trino.exchange.ExchangeManagerRegistry;
 import io.trino.exchange.LazyExchangeDataSource;
 import io.trino.execution.BasicStageInfo;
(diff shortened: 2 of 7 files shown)


core/trino-main/src/main/java/io/trino/FeaturesConfig.java
```diff
@@ -522,6 +524,19 @@ public FeaturesConfig setFaultTolerantExecutionExchangeEncryptionEnabled(boolean
         return this;
     }
 
+    public boolean isSpoolingOnlyExchangeEncryptionEnabled()
```
findepi: Worth documenting how is this different from `isFaultTolerantExecutionExchangeEncryptionEnabled`?

I am ordinary user here, and am confused.

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
```diff
@@ -522,6 +524,19 @@ public FeaturesConfig setFaultTolerantExecutionExchangeEncryptionEnabled(boolean
...
+    {
+        return spoolingOnlyExchangeEncryptionEnabled;
+    }
+
+    @Config("spooling-only-exchange-encryption-enabled")
```
findepi: The property name might be confusing

it's named after "spooling exchanges", but it doesn't control spooling exchanges. It's opposite. 
Maybe `direct-exchange-encryption-enabled`?

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
p3tr0sz: Very good question.
How I did it:
- `fault-tolerant-execution-exchange-encryption-enabled` controls whether encryption is turned on for FTE at all. It is evaluated at query dispatch time in QueryStateMachine: when FTE (retry policy TASK) is active and this flag is true, an encryption key is generated and attached to the session. When false, no key is generated and no exchanges are encrypted.
- `direct-exchange-encryption-enabled` controls which exchanges use the key once it exists.

They are tied together as `direct-exchange-encryption-enabled` takes an effect when encryption is already enabled. I added this to the `@ConfigDescription` and the release notes proposal.

I considered to not introduce new Feature Config and just assume when it is direct then not encrypt. What stopped me was the current behavour: every exchange is encrypted when there is an encryption key. My change weakened the security so I decided to go with additional property and keep the current behaviour as default one.
Happy to hear your thoughts on that.

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
p3tr0sz: Done

core/trino-main/src/test/java/io/trino/exchange/TestLazyExchangeDataSource.java
```diff
@@ -69,4 +65,21 @@ public void testIsBlockedCancellationIsolationInInitializationPhase()
                     .isNotCancelled();
         }
     }
+
+    private static LazyExchangeDataSource createUnresolvedSource()
```
losipiuk, line 72: nit: method extractin can be a separate commit

core/trino-main/src/main/java/io/trino/FeaturesConfig.java
findepi: @losipiuk ^

p3tr0sz: Decided to go with one property.

core/trino-main/src/main/java/io/trino/exchange/LazyExchangeDataSource.java
```diff
@@ -72,6 +72,16 @@ public LazyExchangeDataSource(
         this.exchangeManagerRegistry = requireNonNull(exchangeManagerRegistry, "exchangeManagerRegistry is null");
     }
 
+    @Override
+    public boolean isExternalExchange()
+    {
+        ExchangeDataSource dataSource = delegate.get();
+        if (dataSource == null) {
+            return true;
```
losipiuk: this does not look good. 
The `isExternalExchange` should always return same value when called on the same exchange. Otherwise the contract is very unobvious and hard to follow (what if caller caches the returned result for example).

losipiuk: It may be that throwing here would do the job if we do not know yet what the delegate is.

core/trino-main/src/main/java/io/trino/exchange/LazyExchangeDataSource.java
```diff
@@ -72,6 +72,16 @@ public LazyExchangeDataSource(
         this.exchangeManagerRegistry = requireNonNull(exchangeManagerRegistry, "exchangeManagerRegistry is null");
     }
 
+    @Override
+    public boolean isExternalExchange()
```
losipiuk: Why did you pick `isExternalExchange` here instead keeping `usesExternalStorage` as designator consistently?

core/trino-main/src/main/java/io/trino/execution/buffer/LazyOutputBuffer.java
```diff
@@ -88,6 +88,16 @@ public LazyOutputBuffer(
         this.exchangeManagerRegistry = requireNonNull(exchangeManagerRegistry, "exchangeManagerRegistry is null");
     }
 
+    @Override
+    public boolean usesExternalStorage()
+    {
+        OutputBuffer outputBuffer = getDelegateOutputBuffer();
+        if (outputBuffer == null) {
+            return true;
```
losipiuk: same here. The `usesExternalStorage` should be idempotent.
I think you need to throw if outputBuffer is still null here, but you need to change how it is used in `TaskOutputOperator` too.

losipiuk: You probably want to use `getDelegateOutputBufferOrFail` and then in `TaskOutputOperator` you want to keep memoizing `serializerProvider` instead of `serializer`

core/trino-main/src/main/java/io/trino/exchange/LazyExchangeDataSource.java
p3tr0sz: No particular reason. Actually I started with `isExternalExchange` and later came up with `usesExternalStorage` as more friendly name/less technical details.

losipiuk:
@coderabbitai full review

core/trino-main/src/main/java/io/trino/operator/ExchangeOperator.java
```diff
@@ -253,6 +252,12 @@ public Page getOutput()
             return null;
         }
 
+        if (deserializer == null) {
```
losipiuk, line 259: Add a comment why we cannot do that eagerly in the constructor please.

core/trino-main/src/main/java/io/trino/server/protocol/Query.java
```diff
@@ -591,13 +596,19 @@ private synchronized QueryResultRows removePagesFromExchange(ResultQueryInfo que
                     break;
                 }
 
+                if (deserializer == null) {
```
losipiuk, line 603: worth a comment why here and not in constructor

PR #28736, opened 2026-03-18
Merge tests-bq-ci into tests-bq for BigQuery CI

ebyhr:
## Release notes


chenjian2664:
@copilot

.github/workflows/ci.yml
chenjian2664, line 671: Previously we use `BIGQUERY_CASE_INSENSITIVE_CREDENTIALS_KEY` to run the test, we use  `BIGQUERY_CREDENTIALS_KEY` to control now, does the twos occurs in pair?

ebyhr: We can assume so. There is no need to check every single one.

PR #28737, opened 2026-03-18
Fix MERGE statement formatting in SqlFormatter

takezoe:
## Description
Currently, `SqlFormatter` formats MERGE statement wrongly if the source table is specified without alias.

Original query:
```sql
MERGE INTO t
  USING changes
  ON true
WHEN MATCHED
  THEN DELETE
```

Formatted:
```sql
MERGE INTO t
  USING TABLE changes

  ON true
WHEN MATCHED
  THEN DELETE
```

## Release notes

### DIFF
--- a/core/trino-parser/src/test/java/io/trino/sql/TestSqlFormatter.java
+++ b/core/trino-parser/src/test/java/io/trino/sql/TestSqlFormatter.java
@@ -677,6 +677,41 @@ public class TestSqlFormatter
     }
 
     @Test
+    void testMerge()
+    {
+        assertThat(formatSql(new Merge(
+                new NodeLocation(1, 1),
+                new Table(new NodeLocation(1, 1), QualifiedName.of("t")),
+                table(QualifiedName.of("changes")),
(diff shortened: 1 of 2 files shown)


core/trino-parser/src/test/java/io/trino/sql/TestSqlFormatter.java
```diff
@@ -366,6 +366,41 @@ public void testCreateTableAsSelect()
                         "  t\n");
     }
 
+    @Test
+    void testMerge()
```
ebyhr: nit: The method is currently located between testCreateTableAsSelect and testCreateView. Can we move it above testInsertWithBranch?

takezoe: Sure. Actually, I was wondering where this case should be located :-)

takezoe: Moved.

PR #28764, opened 2026-03-19
Fix flaky Iceberg testBloomFilterColumnWithDictionaryPage

wendigo:
By making sure that test tables are independent from each other. This requires a copied manifest to be patched for unique table location.

Fixes https://github.com/trinodb/trino/issues/26424

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetWithBloomFiltersMixedCase.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetWithBloomFiltersMixedCase.java
@@ -34,6 +34,7 @@ import static io.trino.testing.containers.Minio.MINIO_REGION;
 import static io.trino.testing.containers.Minio.MINIO_ROOT_PASSWORD;
 import static io.trino.testing.containers.Minio.MINIO_ROOT_USER;
 import static java.lang.String.format;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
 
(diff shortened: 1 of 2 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetWithBloomFiltersMixedCase.java
```diff
@@ -107,4 +110,17 @@ public void destroy()
...
+
+    public Minio.ResourcePreProcessor rewriteManifests(String newLocation)
+    {
+        return (fileName, data) -> {
+            if (!fileName.endsWith(".metadata.json")) {
+                return data;
+            }
+
+            return new String(data, UTF_8)
+                    .replaceAll("s3://test-bucket-mixed-case/mixed_case_bloom_filter", newLocation)
```
findepi, line 122: are we updating avro file contents here?

findepi:
Would making this test execution serial (`@Execution(SAME_THREAD)`) also fix the problem?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergParquetWithBloomFiltersMixedCase.java
wendigo: No. Only manifests

findepi: https://iceberg.apache.org/spec/#manifests _A manifest is an immutable Avro file that ..._

Are you sure we're not editing avro files after all?
Are they guaranteed to be utf-8 text? Can they e.g. be compressed?

findepi:
> #26424

- https://github.com/trinodb/trino/pull/28772

PR #28767, opened 2026-03-19
Document truncate(x, n) two-argument variant

### LINKED ISSUE #28765: Document `truncate(number, position)` function
https://trino.io/docs/current/functions/math.html#truncate documents only `truncate(number)`
`truncate(number, position)` is currently not documented.
It's supported for `decimal` and `number` types. 

see `round` docs https://trino.io/docs/current/functions/math.html#round for example

cc @mosabua

alexchenai:
## Summary

- Add documentation for `truncate(x, n)` which truncates `x` to `n` decimal places
- The two-argument variant exists for decimal types but was missing from the docs
- Follows the same pattern as the existing `round(x, d)` documentation

Fixes #28765

## Test plan

- [ ] Verify the docs render correctly with the new function signature
- [ ] Confirm the `:noindex: true` directive prevents duplicate index entries

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/src/main/sphinx/functions/math.md
+++ b/docs/src/main/sphinx/functions/math.md
@@ -114,6 +114,12 @@ Returns the square root of `x`.
 Returns `x` rounded to integer by dropping digits after decimal point.
 :::
 
+:::{function} truncate(x, n) -> [same as input]
+:noindex: true
+
+Returns `x` truncated to `n` decimal places.
+:::
+
 :::{function} width_bucket(x, bound1, bound2, n) -> bigint
 Returns the bin number of `x` in an equi-width histogram with the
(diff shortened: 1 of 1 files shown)


docs/src/main/sphinx/functions/math.md
```diff
@@ -114,6 +114,12 @@ Returns the square root of `x`.
 Returns `x` rounded to integer by dropping digits after decimal point.
 :::
 
+:::{function} truncate(x, n) -> [same as input]
+:noindex: true
+
+Returns `x` truncated to `n` decimal places.
```
findepi, line 120: The docs should mention the `n` can be positive or negative and how it works

PR #28755, opened 2026-03-18
Add support for passing table credentials to Iceberg tables

wendigo:
The connector table credentials are used for cases involving dynamic credentials for accessing a table (e.g.: vended credentials in data lake connectors).

ConnectorTableCredentials are passed through TaskUpdateRequest to ConnectorPage/ConnectorPageSink providers.

## Description

Supersedes #28746
Fixes #28537

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

PR #28775, opened 2026-03-19
Use table credentials for reading from Iceberg `table_changes` & `$files`

findinpath:
## Description

Follow-ups from https://github.com/trinodb/trino/pull/28755 for using a similar mechanism to pass the table credentials:

- `table_changes` Iceberg polymorphic table function.

- `$files` table
trino-spi changes: `ConnectorSystemTableHandle` extends `ConnectorTableHandle`

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/connector/system/SystemTablesMetadata.java
+++ b/core/trino-main/src/main/java/io/trino/connector/system/SystemTablesMetadata.java
@@ -21,15 +21,18 @@ import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ConnectorMetadata;
 import io.trino.spi.connector.ConnectorSession;
+import io.trino.spi.connector.ConnectorTableCredentials;
 import io.trino.spi.connector.ConnectorTableHandle;
 import io.trino.spi.connector.ConnectorTableMetadata;
 import io.trino.spi.connector.ConnectorTableVersion;
+import io.trino.spi.connector.ConnectorWritableTableHandle;
 import io.trino.spi.connector.Constraint;
 import io.trino.spi.connector.ConstraintApplicationResult;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.SchemaTablePrefix;
 import io.trino.spi.connector.SystemColumnHandle;
 import io.trino.spi.connector.SystemTable;
+import io.trino.spi.function.table.ConnectorTableFunctionHandle;
 import io.trino.spi.predicate.TupleDomain;
 
 import java.util.List;
@@ -174,6 +177,25 @@ public class SystemTablesMetadata
         }
         table = new SystemTableHandle(table.schemaName(), table.tableName(), newDomain);
         return Optional.of(new ConstraintApplicationResult<>(table, constraint.getSummary(), constraint.getExpression(), false));
+    }
+
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorTableHandle tableHandle)
+    {
+        SystemTable systemTable = checkAndGetTable(session, tableHandle);
+        return systemTable.getTableCredentials(session);
+    }
+
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorWritableTableHandle tableHandle)
+    {
+        throw new TrinoException(NOT_SUPPORTED, "This connector does not support table credentials");
+    }
+
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorTableFunctionHandle tableFunctionHandle)
+    {
+        throw new TrinoException(NOT_SUPPORTED, "This connector does not support table credentials");
     }
 
     private Constraint effectiveConstraint(TupleDomain<ColumnHandle> oldDomain, Constraint newConstraint, TupleDomain<ColumnHandle> effectiveDomain)
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
@@ -100,4 +100,15 @@ public interface SystemTable
     {
         return Optional.empty();
     }
+
+    /**
+     * Returns {@link ConnectorTableCredentials} if the system table requires credentials to access the data.
+     * <p>
+     * Implementing this method requires implementing {@link #splitSource(ConnectorSession, TupleDomain)} too.
+     * The credentials returned by this method will be passed to {@link ConnectorPageSourceProvider }
+     */
+    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session)
+    {
+        return Optional.empty();
+    }
 }
--- a/lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/classloader/ClassLoaderSafeSystemTable.java
+++ b/lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/classloader/ClassLoaderSafeSystemTable.java
@@ -21,6 +21,7 @@ import io.trino.spi.connector.ConnectorPageSource;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorSplit;
 import io.trino.spi.connector.ConnectorSplitSource;
+import io.trino.spi.connector.ConnectorTableCredentials;
 import io.trino.spi.connector.ConnectorTableMetadata;
 import io.trino.spi.connector.ConnectorTransactionHandle;
 import io.trino.spi.connector.RecordCursor;
@@ -121,6 +122,14 @@ public class ClassLoaderSafeSystemTable
     {
         try (ThreadContextClassLoader _ = new ThreadContextClassLoader(classLoader)) {
             return delegate.splitSource(connectorSession, constraint);
+        }
+    }
+
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session)
+    {
+        try (ThreadContextClassLoader _ = new ThreadContextClassLoader(classLoader)) {
+            return delegate.getTableCredentials(session);
         }
     }
 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -936,8 +936,6 @@ public class IcebergMetadata
             return Optional.empty();
         }
 
-        tableCredentialsCache.put(tableName, IcebergTableCredentials.forFileIO(table.io()));
-
         TableType tableType = IcebergTableName.tableTypeFrom(tableName.getTableName());
         return switch (tableType) {
             case DATA, MATERIALIZED_VIEW_STORAGE -> throw new VerifyException("Unexpected table type: " + tableType); // Handled above.
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
@@ -261,7 +261,7 @@ public class IcebergPageSourceProvider
         if (connectorSplit instanceof FilesTableSplit filesTableSplit) {
             return new FilesTablePageSource(
                     typeManager,
-                    fileSystemFactory.create(session.getIdentity(), filesTableSplit.fileIoProperties()),
+                    fileSystemFactory.create(session.getIdentity(), getFileIoProperties(connectorTableCredentials)),
                     fileIoFactory,
                     columns.stream().map(SystemColumnHandle.class::cast).map(SystemColumnHandle::columnName).collect(toImmutableList()),
                     filesTableSplit);
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
@@ -14,11 +14,13 @@
 package io.trino.plugin.iceberg.system;
 
 import com.google.common.collect.ImmutableList;
+import io.trino.plugin.iceberg.IcebergTableCredentials;
 import io.trino.plugin.iceberg.system.files.FilesTableSplitSource;
 import io.trino.spi.connector.ColumnHandle;
 import io.trino.spi.connector.ColumnMetadata;
 import io.trino.spi.connector.ConnectorSession;
 import io.trino.spi.connector.ConnectorSplitSource;
+import io.trino.spi.connector.ConnectorTableCredentials;
 import io.trino.spi.connector.ConnectorTableMetadata;
 import io.trino.spi.connector.SchemaTableName;
 import io.trino.spi.connector.SystemTable;
@@ -34,7 +36,6 @@ import org.apache.iceberg.PartitionField;
 import org.apache.iceberg.PartitionSpecParser;
 import org.apache.iceberg.SchemaParser;
 import org.apache.iceberg.Table;
-import org.apache.iceberg.io.FileIO;
 
 import java.util.List;
 import java.util.Map;
@@ -134,18 +135,21 @@ public final class FilesTable
     @Override
     public Optional<ConnectorSplitSource> splitSource(ConnectorSession connectorSession, TupleDomain<ColumnHandle> constraint)
     {
-        try (FileIO fileIO = icebergTable.io()) {
-            return Optional.of(new FilesTableSplitSource(
-                    icebergTable,
-                    snapshotId,
-                    SchemaParser.toJson(icebergTable.schema()),
-                    SchemaParser.toJson(MetadataTableUtils.createMetadataTableInstance(icebergTable, MetadataTableType.FILES).schema()),
-                    icebergTable.specs().entrySet().stream().collect(toImmutableMap(
-                            Map.Entry::getKey,
-                            partitionSpec -> PartitionSpecParser.toJson(partitionSpec.getValue()))),
-                    partitionColumnType,
-                    fileIO.properties()));
-        }
+        return Optional.of(new FilesTableSplitSource(
+                icebergTable,
+                snapshotId,
+                SchemaParser.toJson(icebergTable.schema()),
+                SchemaParser.toJson(MetadataTableUtils.createMetadataTableInstance(icebergTable, MetadataTableType.FILES).schema()),
+                icebergTable.specs().entrySet().stream().collect(toImmutableMap(
+                        Map.Entry::getKey,
+                        partitionSpec -> PartitionSpecParser.toJson(partitionSpec.getValue()))),
+                partitionColumnType));
+    }
+
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session)
+    {
+        return Optional.of(IcebergTableCredentials.forFileIO(icebergTable.io()));
     }
 
     public static Type getColumnType(String columnName, TypeManager typeManager)
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
@@ -28,8 +28,7 @@ public record FilesTableSplit(
         String schemaJson,
         String metadataTableJson,
         Map<Integer, String> partitionSpecsByIdJson,
-        Optional<Type> partitionColumnType,
-        Map<String, String> fileIoProperties)
+        Optional<Type> partitionColumnType)
         implements ConnectorSplit
 {
     private static final int INSTANCE_SIZE = instanceSize(FilesTableSplit.class);
@@ -42,7 +41,6 @@ public record FilesTableSplit(
                 + manifestFile.getRetainedSizeInBytes()
                 + estimatedSizeOf(schemaJson)
                 + estimatedSizeOf(metadataTableJson)
-                + estimatedSizeOf(partitionSpecsByIdJson, SizeOf::sizeOf, SizeOf::estimatedSizeOf)
-                + estimatedSizeOf(fileIoProperties, SizeOf::estimatedSizeOf, SizeOf::estimatedSizeOf);
+                + estimatedSizeOf(partitionSpecsByIdJson, SizeOf::sizeOf, SizeOf::estimatedSizeOf);
     }
 }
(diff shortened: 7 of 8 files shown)


core/trino-spi/src/main/java/io/trino/spi/function/table/TableFunctionProcessorProvider.java
```diff
@@ -27,6 +30,19 @@ default TableFunctionDataProcessor getDataProcessor(ConnectorSession session, Co
...
+    /**
+     * This method returns a {@code TableFunctionSplitProcessor}. All the necessary information collected during analysis is available
+     * in the form of {@link ConnectorTableFunctionHandle}. It is called once per each split processed by the table function.
+     */
+    default TableFunctionSplitProcessor getSplitProcessor(
+            ConnectorSession session,
+            ConnectorTableFunctionHandle handle,
+            Optional<ConnectorTableCredentials> tableCredentials,
+            ConnectorSplit split)
+    {
```
wendigo, line 43: if you delegate, validate that tableCredentials.isEmpty() and throw if it's not

core/trino-main/src/main/java/io/trino/connector/system/SystemTablesMetadata.java
```diff
@@ -176,6 +179,18 @@ public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(C
         return Optional.of(new ConstraintApplicationResult<>(table, constraint.getSummary(), constraint.getExpression(), false));
     }
 
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorWritableTableHandle tableHandle)
+    {
+        throw new TrinoException(NOT_SUPPORTED, "This connector does not support writable table credentials");
```
wendigo: just `"This connector does not support table credentials`

core/trino-main/src/main/java/io/trino/connector/system/SystemTablesMetadata.java
```diff
@@ -176,6 +179,18 @@ public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(C
...
+    }
+
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorReadableTableHandle tableHandle)
+    {
+        throw new TrinoException(NOT_SUPPORTED, "This connector does not support readable table credentials");
```
wendigo: just `This connector does not support table credentials`

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -963,9 +964,30 @@ public void setMaterializedViewProperties(ConnectorSession session, SchemaTableN
     @Override
     public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorReadableTableHandle tableHandle)
     {
+        if (tableHandle instanceof ConnectorSystemTableHandle systemTableHandle) {
+            return forHandle(session, systemTableHandle).getTableCredentials(session, systemTableHandle);
+        }
         return forHandle(tableHandle).getTableCredentials(session, tableHandle);
     }
 
+    private ConnectorMetadata forHandle(ConnectorSession session, ConnectorSystemTableHandle tableHandle)
```
chenjian2664: Move the method to blow close to "private forHandle(...)"

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -963,9 +964,30 @@ public void setMaterializedViewProperties(ConnectorSession session, SchemaTableN
...
+    {
+        SchemaTableName tableName = tableHandle.schemaTableName();
+        if (hiveMetadata.getSystemTable(session, tableName).isPresent()) {
```
chenjian2664: I think it may potentially cause conflicts

findinpath: see `io.trino.plugin.lakehouse.LakehouseMetadata#getSystemTable `

Are you referring to `$properties` metadata table being present in both iceberg & hive?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorReadableTableHandle.java
```diff
@@ -11,13 +11,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.trino.plugin.iceberg;
+package io.trino.spi.connector;
 
-class TestIcebergV3AvroConnectorTest
-        extends BaseIcebergAvroConnectorTest
+public interface ConnectorReadableTableHandle
```
findepi: Document what it is and how connectors should use this interface.

findepi:
@coderabbitai full review

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorReadableTableHandle.java
```diff
@@ -11,13 +11,12 @@
...
-package io.trino.plugin.iceberg;
+package io.trino.spi.connector;
 
-class TestIcebergV3AvroConnectorTest
-        extends BaseIcebergAvroConnectorTest
+/**
+ * Represents a handle to a readable relation returned from the connector
+ * to the engine at the analysis time.
+ */
+public interface ConnectorReadableTableHandle
```
findepi: nit:

we can make this sealed
```
public sealed interface ConnectorReadableTableHandle
        permits ConnectorTableHandle, ConnectorTableFunctionHandle
```

(and then mark ConnectorTableHandle, ConnectorTableFunctionHandle as non-sealed)


(however, I am not convinced about this new entity yet)

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorReadableTableHandle.java
```diff
@@ -11,13 +11,12 @@
...
 {
-    TestIcebergV3AvroConnectorTest()
-    {
-        super(3);
-    }
 }
```
findepi: doc: "readable relation" - remove "readable". relations are generally readable 

class name - maybe `ConnectorRelationHandle` to avoid this "Readable" word?

(however, I am not convinced about this new entity yet)

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1853,7 +1829,7 @@ default WriterScalingOptions getInsertWriterScalingOptions(ConnectorSession sess
         return WriterScalingOptions.DISABLED;
     }
 
-    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorTableHandle tableHandle)
+    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorReadableTableHandle tableHandle)
```
findepi: The implementation will have to do `instanceof` checks do understand which type of handle this is.
If we're indeed OK with that, then we can merge this with `getTableCredentials(ConnectorSession session, ConnectorWritableTableHandle tableHandle)` into single function.

However, I am not convinced this is really desired. 
Did you consider simply adding `getTableCredentials` overload for table functions?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSystemTableHandle.java
```diff
@@ -0,0 +1,24 @@
...
+
+/**
+ * Represents a handle to a metadata relation returned from the connector
+ * to the engine at the analysis time.
+ */
+public interface ConnectorSystemTableHandle
+        extends ConnectorTableHandle
+{
+    SchemaTableName schemaTableName();
+}
```
findepi: What's different about `ConnectorSystemTableHandle` that requires new SPI marker interface?

Why this marker interface, unlike all others, has `schemaTableName` method?
Given that `ConnectorSystemTableHandle` is still a `ConnectorTableHandle`, the table name is obtainable via `ConnectorMetadata.getTableMetadata`, so the method looks redundant.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
findinpath: > Did you consider simply adding getTableCredentials overload for table functions?

Adjusting 1st commit with this hint.

losipiuk: This is actually already a common case that we connectors need to do instanceof checks. Like this in delta for example:
```
    @Override
    public SchemaTableName getTableName(ConnectorSession session, ConnectorTableHandle table)
    {
        if (table instanceof CorruptedDeltaLakeTableHandle corruptedTableHandle) {
            return corruptedTableHandle.schemaTableName();
        }
        return ((DeltaLakeTableHandle) table).getSchemaTableName();
    }

```
Is that a big deal? 
Also it is technically possible to have a single class which implements more than one `ConnectorReadableTableHandle` subinterface, then instanceof check would not be needed - but it is probably not a typical class chierarchy.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSystemTableHandle.java
chenjian2664: The name is mainly for exposing the schema table name in connectors, so here the Iceberg can use the name to get the table credentials

findepi: I think we should add the following in the `io.trino.spi.connector.SystemTable` interface

```java
default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session)
    {
        return Optional.empty();
    }
```

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
findepi: It's not a big deal, but it's inconsistent with how other `ConnectorMetadata` API methods were designed for far. That's a big deal.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSystemTableHandle.java
findinpath: I followed up on this question and I ended up introducing another method in `ConnectorMetadata`

```
    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, SchemaTableName tableName)
    {
        return Optional.empty();
    }
```
and modified `MetadataManager`

```
    public Optional<ConnectorTableCredentials> getTableCredentials(Session session, CatalogHandle catalogHandle, ConnectorTableHandle tableHandle)
    {
        if (tableHandle instanceof SystemTableHandle systemTableHandle) {
            verify(catalogHandle.getType().isInternal());
            return getMetadata(session, catalogHandle.getRootCatalogHandle())
                    .getTableCredentials(session.toConnectorSession(catalogHandle.getRootCatalogHandle()), systemTableHandle.schemaTableName());
        }

        return getMetadata(session, catalogHandle).getTableCredentials(session.toConnectorSession(catalogHandle), tableHandle);
    }
```

No more `ConnectorSystemTableHandle` now.

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -1005,6 +1019,31 @@ private ConnectorMetadata forHandle(ConnectorTableHandle handle)
...
+    {
+        if (tableFunctionHandle instanceof TableChangesFunctionHandle _) {
+            return icebergMetadata;
+        }
+        throw new IllegalArgumentException("Unsupported ConnectorTableFunctionHandle type: " + tableFunctionHandle.getClass().getName());
+    }
+
+    private ConnectorMetadata forHandle(ConnectorSession session, SchemaTableName tableName)
+    {
+        if (hiveMetadata.getSystemTable(session, tableName).isPresent()) {
```
findinpath: still needs improvements.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1858,6 +1858,16 @@ default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession
         return Optional.empty();
     }
 
+    default Optional<ConnectorTableCredentials> getSystemTableCredentials(ConnectorSession session, SchemaTableName tableName)
```
chenjian2664: If we don't add a super class, then we have to add a dedicated api in metadata for different type of handle. I am ok with both :)

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSystemTableHandle.java
findepi: I don't think `getTableCredentials` should receive  table by name (``SchemaTableName``).
At the very least, it's not consistent with any other APIs.

my mental model to think about this:

- name gets resolved into a handle once. handle is like a pointer to the data
- at this point, the name can be even dropped from the catalog, the connector should not resolve name again
- all future access goes via the table handle ("the pointer")

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1858,6 +1858,16 @@ default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession
...
+    {
+        return Optional.empty();
+    }
```
findepi: remove (see https://github.com/trinodb/trino/pull/28775#discussion_r2973548245)

core/trino-main/src/main/java/io/trino/connector/system/SystemTablesMetadata.java
```diff
@@ -176,6 +179,30 @@ public Optional<ConstraintApplicationResult<ConnectorTableHandle>> applyFilter(C
         return Optional.of(new ConstraintApplicationResult<>(table, constraint.getSummary(), constraint.getExpression(), false));
     }
 
+    @Override
+    public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorWritableTableHandle tableHandle)
```
findepi: nit: keep same order as in the interface. `getTableCredentials(ConnectorSession session, ConnectorTableHandle tableHandle)` is the first overload in `ConnectorMetadata`

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1858,6 +1858,16 @@ default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession
...
+
+    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorTableFunctionHandle tableFunctionHandle)
```
findepi, line 1875: nit: move it under `getTableCredentials(ConnectorSession session, ConnectorWritableTableHandle tableHandle)`, so that overloads `ConnectorTableHandle` and `ConnectorWritableTableHandle` are next to each other.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSystemTableHandle.java
findinpath: > I don't think getTableCredentials should receive table by name (SchemaTableName).
> At the very least, it's not consistent with any other APIs.

@findepi pls do take a look at 
`io.trino.spi.connector.ConnectorMetadata#getSystemTable`

https://github.com/trinodb/trino/blob/b5a4f5aa1da3b1bb5f23aebb424ab9a4db211dae/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java#L198-L207

findepi: if this is for system tables only, why not convey this in the method name?

also, did you consider https://github.com/trinodb/trino/pull/28775#discussion_r2966089270? was there any blocker with that?

findinpath: > did you consider https://github.com/trinodb/trino/pull/28775#discussion_r2966089270? was there any blocker with that?

Apologies, I missed that comment.

I've refactored the code to add in the `SystemTable` implementations the ability to get the table credentials (and implemented the method in `FilesTable`).

core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
```diff
@@ -100,4 +100,12 @@ default Optional<ConnectorSplitSource> splitSource(ConnectorSession connectorSes
     {
         return Optional.empty();
     }
+
+    /**
+     * Returns {@link ConnectorTableCredentials} if the system table requires credentials to access the data or {@link Optional#empty} if there are no credentials.
+     */
+    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session)
```
findepi, line 110: If something is returned from here, where the system table gets these credentials back?

My understanding is that this works only when `splitSource` is also enabled. Is this correct?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1862,4 +1862,9 @@ default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession
     {
         return Optional.empty();
     }
+
+    default Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession session, ConnectorTableFunctionHandle tableFunctionHandle)
```
losipiuk, line 1875: nit: add javadoc.
Given other overloads do not have javadoc already, this can be done as a followup.

core/trino-spi/src/main/java/io/trino/spi/function/table/TableFunctionProcessorProvider.java
```diff
@@ -27,6 +33,22 @@ default TableFunctionDataProcessor getDataProcessor(ConnectorSession session, Co
...
```
findepi, line 43: This should replace `getSplitProcessor(ConnectorSession session, ConnectorTableFunctionHandle handle, ConnectorSplit split)`. Mark the old overload as deprecated.

core/trino-spi/src/main/java/io/trino/spi/function/table/TableFunctionProcessorProvider.java
```diff
@@ -27,6 +33,22 @@ default TableFunctionDataProcessor getDataProcessor(ConnectorSession session, Co
         throw new UnsupportedOperationException("this table function does not process input data");
     }
 
+    /**
```
losipiuk, line 33: nit: you can refer to credentials and whats its purpose in javadoc

core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
losipiuk: This should be conveyed by the javadoc

losipiuk: You may also note that for other methods `RecordCursor cursor(...)` `ConnectorPageSource pageSource(...)` table credentials are handled internally.

It is quite hard to figure out what is the flow just by looking at this interface.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
findinpath: Created a preparatory commit

core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
findinpath: Added javadoc.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java
```diff
@@ -1853,11 +1853,17 @@ default WriterScalingOptions getInsertWriterScalingOptions(ConnectorSession sess
         return WriterScalingOptions.DISABLED;
     }
 
+    /**
+     * Returns {@link ConnectorTableCredentials} for specified {@link ConnectorTableHandle} or {@link Optional#empty} if there are no credentials.
```
findepi: This doesn't say anything useful that the method signature didn't already say. 

What would be more interesting is how those credentials can be used.
https://github.com/trinodb/trino/pull/28830 (now merged) helps in this regard, because it makes the credential-taking APIs easier to find. 

Either remove, or replace with something like

```
Returns credentials that can be used in PageSourceProvider.
```

core/trino-spi/src/main/java/io/trino/spi/function/table/TableFunctionProcessorProvider.java
```diff
@@ -27,10 +33,30 @@ default TableFunctionDataProcessor getDataProcessor(ConnectorSession session, Co
...
+     */
+    default TableFunctionSplitProcessor getSplitProcessor(
+            ConnectorSession session,
+            ConnectorTableFunctionHandle handle,
+            Optional<ConnectorTableCredentials> tableCredentials,
+            ConnectorSplit split)
+    {
+        if (tableCredentials.isPresent()) {
+            throw new TrinoException(NOT_SUPPORTED, "This table function does not support table credentials");
+        }
```
findepi: remove.

core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
```diff
@@ -100,4 +100,18 @@ default Optional<ConnectorSplitSource> splitSource(ConnectorSession connectorSes
     {
         return Optional.empty();
     }
+
+    /**
+     * Returns {@link ConnectorTableCredentials} if the system table requires credentials to access the data
+     * or {@link Optional#empty} if there are no credentials.
```
findepi: remove `or {@link Optional#empty} if there are no credentials.`. don't state the obvious

core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
```diff
@@ -100,4 +100,18 @@ default Optional<ConnectorSplitSource> splitSource(ConnectorSession connectorSes
...
+     * <p>
+     * The credentials returned by this method during analysis will be used by the operator when
+     * processing splits for the system table generated by the {@link #splitSource(ConnectorSession, TupleDomain)} method.
```
findepi: Analysis is an earlier stage of query processing, before the plan is created, optimized and scheduled for execution. Fortunately, it doesn't (and should not) matter from SPI perspective.

Also, from SPI perspective, when implementing a table reads or system table, there is no "operator".

Change to:
```
Implementing this method requires implementing {@link #splitSource(ConnectorSession, TupleDomain)} too.
The credentials returned by this method will be passed to { #PageSourceProvider }
```

core/trino-spi/src/main/java/io/trino/spi/connector/SystemTable.java
```diff
@@ -100,4 +100,18 @@ default Optional<ConnectorSplitSource> splitSource(ConnectorSession connectorSes
...
+     *
+     * @see #splitSource(ConnectorSession, TupleDomain)
```
findepi: Remove.

findepi:
SPI LGTM. I didn't review the rest. 
@chenjian2664 @wendigo are you confident with the rest, or need another 👀 ?

chenjian2664:
I am cool with it. cc @wendigo @losipiuk

chenjian2664:
Thank you! @findinpath

Issue #13990, opened 2022-09-05
Delta-Lake only support Timestamp(3) 


hudsondba:
Today the delta connector only supports timestamp column with precision of 3.

trino> create table IF NOT EXISTS delta.events_silver.dummy_table(
    -> event_source varchar,
    -> event_time timestamp(6)
    -> );
Query 20220905_113842_00007_vxs5p failed: Unsupported type: timestamp(6)

This a huge blocker considering that spark supports a timestamp column with a higher precision:

>>> spark.sql("select * from delta.events_silver.dummy_table").show(10,False)
+--------------+-------------------------
|event_source.      |event_time             
+--------------+-----------------------------------+
|device               |2022-09-05 10:23:37.848712

PR #28776, opened 2026-03-19
fix(delta-lake): preserve microsecond precision when reading timestamp with timezone

patrickmoch:
## Description

Fixes precision loss when reading timestamp with time zone columns from Delta Lake tables backed by Parquet files written with microsecond precision (e.g. by Spark).

The Delta Lake [protocol spec](https://github.com/delta-io/delta/blob/master/PROTOCOL.md#primitive-types) defines timestamp as microsecond precision stored as int64 with isAdjustedToUTC=true in Parquet. Trino was mapping this to TIMESTAMP_TZ_MILLIS (precision 3), causing the sub-millisecond digits to be silently dropped on read. For example, a value written by Spark as 2024-01-12 09:30:47.405423 UTC would be read by Trino as 2024-01-12 09:30:47.405 UTC.
The fix maps the Delta "timestamp" type to TIMESTAMP_TZ_MICROS (precision 6) and updates the validation guard in validatePrimitiveType accordingly.
Backward compatibility with Parquet files written by older Trino versions (which used TIMESTAMP(MILLIS) physical encoding) is preserved by handling the MILLIS case in ColumnReaderFactory for LongTimestampWithTimeZone rather than throwing. These files will read correctly with zeros in the sub-millisecond digits, which accurately reflects the precision they were written with.
Note: timestamp_ntz (without timezone) was already correctly mapped to TIMESTAMP_MICROS (precision 6) — this change brings timestamp (with timezone) into alignment with it.

## Release notes

(X ) Release notes are required, with the following suggested text:

```markdown
## Delta Lake
* Fix precision loss when reading `timestamp with time zone` columns from
  Delta Lake tables. Values are now read at microsecond precision
  (`timestamp(6) with time zone`) instead of being silently truncated to
  millisecond precision (`timestamp(3) with time zone`). ({issue} '20359')
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/Catalog.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/Catalog.java
@@ -25,9 +25,12 @@ import io.trino.spi.transaction.IsolationLevel;
 import io.trino.transaction.InternalConnector;
 import io.trino.transaction.TransactionId;
 
+import java.util.Optional;
+
 import static com.google.common.base.MoreObjects.toStringHelper;
 import static com.google.common.base.Preconditions.checkArgument;
 import static io.trino.connector.CatalogHandle.createRootCatalogHandle;
+import static io.trino.metadata.CatalogMetadata.SecurityManagement;
 import static io.trino.metadata.CatalogStatus.FAILING;
 import static io.trino.spi.StandardErrorCode.CATALOG_UNAVAILABLE;
 import static java.lang.String.format;
@@ -96,6 +99,12 @@ public class Catalog
     public CatalogStatus getCatalogStatus()
     {
         return catalogStatus;
+    }
+
+    public Optional<SecurityManagement> getSecurityManagement()
+    {
+        return Optional.ofNullable(catalogConnector)
+                .map(ConnectorServices::getSecurityManagement);
     }
 
     public CatalogInfo toInfo()
--- a/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
@@ -855,14 +855,13 @@ public final class MetadataManager
     @Override
     public void dropCatalog(Session session, CatalogName catalog, boolean cascade)
     {
-        Optional<CatalogMetadata> catalogMetadata = Optional.empty();
         // there is a potential race condition here, TODO: https://github.com/trinodb/trino/issues/26927
-        Optional<Catalog> optionalCatalog = catalogManager.getCatalog(catalog);
-        if (optionalCatalog.isPresent() && optionalCatalog.get().getCatalogStatus() == OPERATIONAL) {
-            catalogMetadata = Optional.of(getCatalogMetadataForWrite(session, catalog.toString()));
-        }
+        boolean systemSecurityManagement = catalogManager.getCatalog(catalog)
+                .filter(foundCatalog -> foundCatalog.getCatalogStatus() == OPERATIONAL)
+                .flatMap(Catalog::getSecurityManagement)
+                .orElse(CONNECTOR) == SYSTEM;
         catalogManager.dropCatalog(catalog, cascade);
-        if (catalogMetadata.isPresent() && catalogMetadata.get().getSecurityManagement() == SYSTEM) {
+        if (systemSecurityManagement) {
             systemSecurityMetadata.catalogDropped(session, catalog);
         }
     }
--- a/testing/trino-tests/src/test/java/io/trino/tests/TestMetadataManager.java
+++ b/testing/trino-tests/src/test/java/io/trino/tests/TestMetadataManager.java
@@ -26,6 +26,7 @@ import io.trino.server.SessionContext;
 import io.trino.server.protocol.Slug;
 import io.trino.spi.Plugin;
 import io.trino.spi.QueryId;
+import io.trino.spi.TrinoException;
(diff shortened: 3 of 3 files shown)


ebyhr (commented):
Could you update integration tests and [product tests](https://github.com/trinodb/trino/tree/master/testing/trino-product-tests)?

patrickmoch:
I will check that and try to provide the tests

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/transactionlog/DeltaLakeSchemaSupport.java
```diff
@@ -429,7 +430,7 @@ private static void validatePrimitiveType(Type type)
     {
         if (serializeSupportedPrimitiveType(type).isEmpty() ||
                 (type instanceof TimestampType timestampType && timestampType.getPrecision() != 6) ||
-                (type instanceof TimestampWithTimeZoneType timestampWithTimeZoneType && timestampWithTimeZoneType.getPrecision() != 3)) {
+                (type instanceof TimestampWithTimeZoneType timestampWithTimeZoneType && timestampWithTimeZoneType.getPrecision() != 6)) {  // was != 3
```
findinpath: remove ` // was != 3`

It does not have added value in the eventual NEW version of the code.

patrickmoch: youre right, deleted it

patrickmoch:
Adds enhancement #13990

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeSystemTables.java
```diff
@@ -470,7 +470,7 @@ public void testPartitionsTableColumnTypes()
                 ROW(ROW(TIMESTAMP '2001-05-06 12:34:56.123000', TIMESTAMP '2001-05-06 12:34:56.568000', BIGINT '0')),
                 ROW(ROW(TIMESTAMP '2001-05-06 12:34:56.123000', TIMESTAMP '2001-05-06 12:34:56.124000', BIGINT '0'))
                 """);
-        testPartitionsTableColumnTypes("TIMESTAMP(3) WITH TIME ZONE", "VALUES (TIMESTAMP '2001-05-06 12:34:56.123 UTC', 'a'), (TIMESTAMP '2001-05-06 12:34:56.234 -08:30', 'a'), (TIMESTAMP '2001-05-06 12:34:56.567 GMT-08:30', 'b'), (TIMESTAMP '2001-05-06 12:34:56.789 America/New_York', 'b')", 4,
+        testPartitionsTableColumnTypes("TIMESTAMP(6) WITH TIME ZONE", "VALUES (TIMESTAMP '2001-05-06 12:34:56.123 UTC', 'a'), (TIMESTAMP '2001-05-06 12:34:56.234 -08:30', 'a'), (TIMESTAMP '2001-05-06 12:34:56.567 GMT-08:30', 'b'), (TIMESTAMP '2001-05-06 12:34:56.789 America/New_York', 'b')", 4,
```
marcinsbd: can we modify microseconds part be different then 000 in the  test cases?

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
```diff
@@ -1390,7 +1390,7 @@ public void testCheckpointWriteStatsAsStruct()
         testCheckpointWriteStatsAsStruct("varchar", "'test'", "'ŻŻŻŻŻŻŻŻŻŻ'", "0.0", "null", "null");
         testCheckpointWriteStatsAsStruct("varbinary", "X'65683F'", "X'ffffffffffffffffffff'", "0.0", "null", "null");
         testCheckpointWriteStatsAsStruct("date", "date '2021-02-03'", "date '9999-12-31'", "0.0", "'2021-02-03'", "'9999-12-31'");
-        testCheckpointWriteStatsAsStruct("timestamp(3) with time zone", "timestamp '2001-08-22 03:04:05.321 -08:00'", "timestamp '9999-12-31 23:59:59.999 +12:00'", "0.0", "'2001-08-22 11:04:05.321 UTC'", "'9999-12-31 11:59:59.999 UTC'");
+        testCheckpointWriteStatsAsStruct("timestamp(6) with time zone", "timestamp '2001-08-22 03:04:05.321 -08:00'", "timestamp '9999-12-31 23:59:59.999 +12:00'", "0.0", "'2001-08-22 11:04:05.321 UTC'", "'9999-12-31 11:59:59.999 UTC'");
```
marcinsbd: can we add microseconds part to timestamps?

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeSystemTables.java
patrickmoch: yes sure, maybe mixing precision 3 and 6

patrickmoch:
Resolving wrong precisions in testcases today and submitting changes

findepi:
@patrickmoch thank you for your PR!
We don't use merge commits in this project. Can you please rebase on current master and make sure the CI build is green?

PR #28783, opened 2026-03-20
Allow to store externalAuthenticationToken in SYSTEM Cache

ssheikin:
(in dotfiles),
so it can be reused between processes, without requiring authentication. This may be helpful in case different CLI are run from the same machine.

## Release notes

```markdown
## Section
* Allow to store externalAuthenticationToken in SYSTEM Cache. ({issue}`issuenumber`)
```

### DIFF
--- a/client/trino-client/src/test/java/io/trino/client/auth/external/TestExternalAuthenticator.java
+++ b/client/trino-client/src/test/java/io/trino/client/auth/external/TestExternalAuthenticator.java
@@ -202,7 +202,7 @@ public class TestExternalAuthenticator
     }
 
     @Test
-    @Timeout(2)
+    @Timeout(10)
     public void testAuthenticationFromMultipleThreadsWithSystemCachedToken(@TempDir Path tempDir)
     {
         KnownToken knownToken = new SystemCachedKnownToken(tempDir);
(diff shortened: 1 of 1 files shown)


wendigo:
What if I want to authenticate as two different users in the two different CLIs?

ssheikin:
The same question as 
`What if I want to authenticate to two different servers in the two different CLIs?`
Don't use this type of caching, then.

client/trino-client/src/main/java/io/trino/client/auth/external/KnownToken.java
```diff
@@ -31,4 +31,9 @@ static KnownToken memoryCached()
     {
```
kokosing, line 31: please add doc changes

client/trino-client/src/main/java/io/trino/client/auth/external/SystemCachedKnownToken.java
```diff
@@ -0,0 +1,137 @@
...
+    public Optional<Token> getToken()
+    {
+        // Wait while lock file exists, mimicking readLock blocking while writeLock is held
+        Failsafe.with(RetryPolicy.builder()
+                        .handleResultIf(Boolean.TRUE::equals)
+                        .withMaxAttempts(-1)
+                        .withDelay(100, 500, MILLIS)
+                        .withMaxDuration(Duration.ofMinutes(10))
+                        .build())
+                .get(() -> Files.exists(lockFile));
```
kokosing, line 78: What is the error thrown when this file exists for longer than 10 minutes?

ssheikin: added test for that.

ssheikin:
@wendigo @lukasz-walkiewicz would you like to take another glance here?

augurpl:
not an expert here so just, ack.

ssheikin:
Is it a good time to move forward with this PR?

PR #28791, opened 2026-03-21
Prefer the most specific function overload during coercion

dain:
## Description

When overload resolution falls back to coercion, Trino should prefer the narrowest applicable
function declaration. The previous selection logic could still depend on candidate order because it
compared a bound signature of one candidate against another candidate's declared signature.

This change replaces that selection step with a deterministic specificity comparison.

The new logic:

- compares bound signatures first, preserving existing coercion behavior
- breaks ties using declared-signature specificity
- prefers narrower declarations such as exact over generic, fixed arity over varargs, concrete over
  calculated, and repeated equality constraints over independent ones

Type-variable constraints are still handled during applicability filtering, but they are not
considered during ranking.

## Release notes

```markdown
## General
* Improve function overload resolution during coercion to prefer the most specific applicable function. ({issue}`28763`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
@@ -17,6 +17,7 @@ import com.google.common.base.Joiner;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Ordering;
 import io.trino.connector.CatalogHandle;
+import io.trino.metadata.FunctionSpecificityComparator.Specificity;
 import io.trino.spi.TrinoException;
 import io.trino.spi.function.BoundSignature;
 import io.trino.spi.function.CatalogSchemaFunctionName;
@@ -27,7 +28,6 @@ import io.trino.spi.type.Type;
 import io.trino.spi.type.TypeManager;
 import io.trino.sql.analyzer.TypeSignatureProvider;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
@@ -57,11 +57,13 @@ class FunctionBinder
 {
     private final Metadata metadata;
     private final TypeManager typeManager;
+    private final FunctionSpecificityComparator functionSpecificityComparator;
 
     public FunctionBinder(Metadata metadata, TypeManager typeManager)
     {
         this.metadata = requireNonNull(metadata, "metadata is null");
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
+        this.functionSpecificityComparator = new FunctionSpecificityComparator(metadata, typeManager);
     }
 
     CatalogFunctionBinding bindFunction(List<TypeSignatureProvider> parameterTypes, Collection<CatalogFunctionMetadata> candidates, String displayName)
@@ -233,25 +235,55 @@ class FunctionBinder
 
     private List<ApplicableFunction> selectMostSpecificFunctions(List<ApplicableFunction> candidates)
     {
-        // Provided `isMoreSpecificThan` is a partial order relation, this finds all the minimum values among candidates.
-        // TODO Warning: `isMoreSpecificThan` compares bound signature of the left with declared signature of the right (asymmetric) and it is *not* proper partial order relation.
-        //  the result depends on candidates order, and order in which `isMoreSpecificThan` is applied.
-
-        List<ApplicableFunction> representatives = new ArrayList<>();
-
-        for (ApplicableFunction current : candidates) {
-            if (representatives.removeIf(representative -> isMoreSpecificThan(current, representative))) {
-                representatives.add(current);
-                continue;
+        SpecificityComparisonCache comparisonCache = new SpecificityComparisonCache(candidates);
+        ImmutableList.Builder<ApplicableFunction> mostSpecificFunctions = ImmutableList.builder();
+        for (int currentIndex = 0; currentIndex < candidates.size(); currentIndex++) {
+            boolean lessSpecificThanAnother = false;
+            for (int otherIndex = 0; otherIndex < candidates.size(); otherIndex++) {
+                if (currentIndex == otherIndex) {
+                    continue;
+                }
+                if (comparisonCache.isMoreSpecific(otherIndex, currentIndex)) {
+                    lessSpecificThanAnother = true;
+                    break;
+                }
             }
-            if (representatives.stream().anyMatch(representative -> isMoreSpecificThan(representative, current))) {
-                // Current is less specific than one of the retained representatives.
-                continue;
+            if (!lessSpecificThanAnother) {
+                mostSpecificFunctions.add(candidates.get(currentIndex));
             }
-            representatives.add(current);
+        }
+        return mostSpecificFunctions.build();
+    }
+
+    private final class SpecificityComparisonCache
+    {
+        private final List<ApplicableFunction> candidates;
+        private final Specificity[][] cache;
+
+        private SpecificityComparisonCache(List<ApplicableFunction> candidates)
+        {
+            this.candidates = requireNonNull(candidates, "candidates is null");
+            this.cache = new Specificity[candidates.size()][candidates.size()];
         }
 
-        return representatives;
+        private boolean isMoreSpecific(int leftIndex, int rightIndex)
+        {
+            Specificity cachedSpecificity = cache[leftIndex][rightIndex];
+            if (cachedSpecificity != null) {
+                return cachedSpecificity == Specificity.MORE_SPECIFIC;
+            }
+
+            ApplicableFunction left = candidates.get(leftIndex);
+            ApplicableFunction right = candidates.get(rightIndex);
+            Specificity specificity = functionSpecificityComparator.compareSpecificity(
+                    left.boundSignature(),
+                    left.declaredSignature(),
+                    right.boundSignature(),
+                    right.declaredSignature());
+            cache[leftIndex][rightIndex] = specificity;
+            cache[rightIndex][leftIndex] = specificity.reverse();
+            return specificity == Specificity.MORE_SPECIFIC;
+        }
     }
 
     private static boolean someParameterIsUnknown(List<Type> parameters)
@@ -322,16 +354,6 @@ class FunctionBinder
             resultBuilder.add(typeManager.getType(typeSignatureProvider.getTypeSignature()));
         }
         return Optional.of(resultBuilder.build());
-    }
-
-    /**
-     * One method is more specific than another if invocation handled by the first method could be passed on to the other one
-     */
-    private boolean isMoreSpecificThan(ApplicableFunction left, ApplicableFunction right)
-    {
-        List<TypeSignatureProvider> resolvedTypes = fromTypeSignatures(left.boundSignature().getArgumentTypes());
-        return new SignatureBinder(metadata, typeManager, right.declaredSignature(), true)
-                .canBind(resolvedTypes);
     }
 
     private CatalogFunctionBinding toFunctionBinding(CatalogFunctionMetadata functionMetadata, Signature signature)
--- a/core/trino-main/src/main/java/io/trino/metadata/FunctionSpecificityComparator.java
+++ b/core/trino-main/src/main/java/io/trino/metadata/FunctionSpecificityComparator.java
@@ -1,0 +1,474 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.metadata;
+
+import com.google.common.collect.ImmutableList;
+import io.trino.spi.function.Signature;
+import io.trino.spi.function.TypeVariableConstraint;
+import io.trino.spi.type.Type;
+import io.trino.spi.type.TypeManager;
+import io.trino.spi.type.TypeParameter;
+import io.trino.spi.type.TypeSignature;
+import io.trino.sql.analyzer.TypeSignatureProvider;
+import io.trino.type.TypeCoercion;
+import it.unimi.dsi.fastutil.ints.IntArrayList;
+
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.TreeMap;
+
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Verify.verify;
+import static java.lang.String.CASE_INSENSITIVE_ORDER;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Compares already-applicable functions to determine which one is more specific for overload
+ * selection in {@link FunctionBinder}.
+ * <p>
+ * This helper is intentionally specialized to FunctionBinder's current needs. It is not a general
+ * partial order over arbitrary signatures or bindings.
+ * <p>
+ * The comparison is performed in two phases:
+ * <ol>
+ * <li>Compare already-bound applicable functions using the existing forwardability rule: a bound
+ * call handled by one function is more specific if it can also bind to the other function's
+ * declared signature.</li>
+ * <li>If that relation is mutual, compare the declared signatures structurally to choose the
+ * narrower declaration deterministically.</li>
+ * </ol>
+ */
+final class FunctionSpecificityComparator
+{
+    enum Specificity
+    {
+        MORE_SPECIFIC,
+        LESS_SPECIFIC,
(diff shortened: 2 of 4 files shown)


core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
```diff
@@ -0,0 +1,415 @@
...
+    public Comparison compare(Signature left, Signature right, int arity)
+    {
+        requireNonNull(left, "left is null");
+        requireNonNull(right, "right is null");
+        checkArgument(arity >= 0, "arity is negative");
+
+        Optional<List<TypeSignature>> leftArguments = expandArguments(left, arity);
+        Optional<List<TypeSignature>> rightArguments = expandArguments(right, arity);
+        if (leftArguments.isEmpty() || rightArguments.isEmpty()) {
+            return Comparison.INCOMPARABLE;
```
findepi: It seems the Optional.empty here indicates a bug (signature does not match call arity). In such case, i'd expect the code to fail fast.

No need for `expandArguments` to return `Optional`.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
```diff
@@ -231,27 +232,53 @@ private List<ApplicableFunction> selectMostSpecificFunctions(List<ApplicableFunc
...
         }
 
-        return representatives;
+        return signatureSpecificityComparator.compare(left.declaredSignature(), right.declaredSignature(), arity);
+    }
+
+    /**
+     * One method is more specific than another if invocation handled by the first method could be passed on to the other one.
+     */
+    private boolean isMoreSpecificThan(ApplicableFunction left, ApplicableFunction right)
```
findepi: Can you perhaps avoid moving the method so the diff is smaller?

It looks like `isMoreSpecificThan` did not change at all.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
```diff
@@ -231,27 +232,53 @@ private List<ApplicableFunction> selectMostSpecificFunctions(List<ApplicableFunc
...
+        }
+        if (rightMoreSpecific && !leftMoreSpecific) {
+            return SignatureSpecificityComparator.Comparison.LESS_SPECIFIC;
+        }
+        if (!leftMoreSpecific && !rightMoreSpecific) {
+            return SignatureSpecificityComparator.Comparison.INCOMPARABLE;
         }
 
-        return representatives;
+        return signatureSpecificityComparator.compare(left.declaredSignature(), right.declaredSignature(), arity);
```
findepi: Can we have one comparator that does the job instead of two stacked one over another?
This could lead to code that's easier to reason about.
WDYT?

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
```diff
@@ -231,27 +232,53 @@ private List<ApplicableFunction> selectMostSpecificFunctions(List<ApplicableFunc
...
-        List<ApplicableFunction> representatives = new ArrayList<>();
-
+        ImmutableList.Builder<ApplicableFunction> mostSpecificFunctions = ImmutableList.builder();
         for (ApplicableFunction current : candidates) {
-            if (representatives.removeIf(representative -> isMoreSpecificThan(current, representative))) {
-                representatives.add(current);
-                continue;
+            boolean lessSpecificThanAnother = false;
+            for (ApplicableFunction other : candidates) {
+                if (current == other) {
```
findepi: This is now quadratic. Previously, if we were to consider a function overloads for smallint, integer, bigint, decimal, double, number, we did about 5-ish comparisons, because we generally kept 1 surviving representative

core/trino-main/src/test/java/io/trino/metadata/TestSignatureSpecificityComparator.java
```diff
@@ -0,0 +1,204 @@
...
+import static io.trino.spi.function.TypeVariableConstraint.typeVariable;
+import static io.trino.sql.analyzer.TypeSignatureTranslator.parseTypeSignature;
+import static io.trino.sql.planner.TestingPlannerContext.PLANNER_CONTEXT;
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class TestSignatureSpecificityComparator
+{
+    private static final Set<String> LITERAL_PARAMETERS = ImmutableSet.of("p", "s", "p1", "s1", "p2", "s2", "x", "y");
+
+    private final SignatureSpecificityComparator comparator = new SignatureSpecificityComparator(PLANNER_CONTEXT.getTypeManager());
```
findepi: This is good to have unit test for the function binder selection.
However, testing this comparator is not enough.
See https://github.com/trinodb/trino/pull/28791/changes#r2969304558 for the why

findepi:
From the two overloads
```
approx_distinct : <T:comparable>(t,double):bigint
approx_distinct : (boolean,double):bigint
```

which one gets selected now?

core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
```diff
@@ -0,0 +1,415 @@
...
+    SignatureSpecificityComparator(TypeManager typeManager)
+    {
+        this.typeManager = requireNonNull(typeManager, "typeManager is null");
+        this.typeCoercion = new TypeCoercion(typeManager::getType);
+    }
+
+    /**
+     * Compares two already-applicable declarations for a specific call arity.
+     */
+    public Comparison compare(Signature left, Signature right, int arity)
```
findepi: Do you have an example where comparison result of specificity of declared signatures depends on call site arity?

core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
dain: Agreed. I changed `expandArguments` to fail fast with `verify(...)` instead of returning `Optional.empty()`. An arity mismatch here should indicate a bug, not an incomparable pair.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
dain: I did that, but then I refactored the code more based on your further comments, so this is actually moot now.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
dain: I agree with the direction here. I refactored this so `FunctionBinder` now uses a single `FunctionSpecificityComparator` API.

Internally it still has two phases, because the primary comparison is over already-bound applicable functions, while the tie-break is over declared signatures. BTW, I tried replacing the former with a purely structural declared-signature comparison and that broke existing coercion behavior for concrete (non-generic) overloads.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
dain: I think this is a fair concern. I went back and compared the old and new code in units of one-way `SignatureBinder.canBind(...)` calls, since that is the expensive operation here.

For the old algorithm, the range is `n - 1` to `n(n - 1)`. The best case happens when the representative set stays very small and candidates arrive in a favorable order, so each new candidate is resolved quickly. The worst case happens when many candidates remain relevant and there is little short-circuiting.

I updated the new code to memoize pairwise comparisons within one selection pass. With that change, the new algorithm ranges from `2(n - 1)` to `n(n - 1)`. So compared to the old code, it is about 2x more expensive in the favorable case, while matching the old worst case. As with the old code, the worst case is most likely when there are many applicable candidates with weak or no dominance relationships, for example overlapping generic overloads.

I also agree with the general point that the old representative-based structure often behaved closer to its favorable case when overloads were close to a chain. The remaining extra cost here is the tradeoff for making selection order-independent, since the old algorithm was relying on a partial-order property that was not actually valid.

core/trino-main/src/test/java/io/trino/metadata/TestSignatureSpecificityComparator.java
dain: Agreed. The helper-level tests are there to pin the comparator behavior, but the actual regression coverage is in `TestGlobalFunctionCatalog`, where the binder is exercised end-to-end for the order-dependent and builtin cases. I also expanded `TestFunctionSpecificityComparator` to cover the full `compareSpecificity(...)` API directly, so the helper tests now cover both the public binder-facing comparison and the internal declared-signature tie-break logic.

core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
dain: The only reason declared-signature comparison depends on call-site arity is varargs expansion. For non-varargs declarations it doesn't materially matter.
A varargs declaration has to be expanded to the actual call arity before its argument positions can be compared against another declaration, so the internal declared-signature helper still needs arity for that case.

dain:
> From the two overloads
> 
> ```
> approx_distinct : <T:comparable>(t,double):bigint
> approx_distinct : (boolean,double):bigint
> ```
> 
> which one gets selected now?

The specialized `(boolean, double)` overload is selected. I added a binder-level regression for that in `TestGlobalFunctionCatalog` (`approx_distinct(boolean, real)` resolves to the boolean-specific implementation).

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
findepi: > BTW, I tried replacing the former with a purely structural declared-signature comparison and that broke existing coercion behavior for concrete (non-generic) overloads.

That's interesting. Do you maybe have an example?

core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
findepi: The varargs are at the end. At least in Java, the function overload prioritization does not depend on call-site arity. Which ones are applicable (obviously) does.

Do you have **an example** where comparison result of specificity of declared signatures depends on call site arity?

dain:
@coderabbitai full review

dain:
@findepi Do you want to give this a final review?

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
```diff
@@ -233,25 +235,55 @@ private List<ApplicableFunction> selectMostSpecificFunctions(List<ApplicableFunc
...
+        {
+            this.candidates = requireNonNull(candidates, "candidates is null");
+            this.cache = new Specificity[candidates.size()][candidates.size()];
+        }
+
+        private boolean isMoreSpecific(int leftIndex, int rightIndex)
+        {
+            Specificity cachedSpecificity = cache[leftIndex][rightIndex];
+            if (cachedSpecificity != null) {
+                return cachedSpecificity == Specificity.MORE_SPECIFIC;
```
findepi, line 273: Shouldn't this implement Comparator or Comparator-like interface and return 3-valued result (<, =, >)? 

Then we could replace `selectMostSpecificFunctions` with just general purpose sorting.

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
```diff
@@ -233,25 +235,55 @@ private List<ApplicableFunction> selectMostSpecificFunctions(List<ApplicableFunc
...
+            }
+
+            ApplicableFunction left = candidates.get(leftIndex);
+            ApplicableFunction right = candidates.get(rightIndex);
+            Specificity specificity = functionSpecificityComparator.compareSpecificity(
+                    left.boundSignature(),
+                    left.declaredSignature(),
+                    right.boundSignature(),
+                    right.declaredSignature());
```
findepi, line 282: Comparing bound signatures needs to happen at query analysis time, because they vary.
Comparing declared signatures could be done ahead of time. They don't change. That's why https://github.com/trinodb/trino/pull/28791#discussion_r2969375839 is important.
Is there a potential to move some computation out of query analysis path, or cache it?

core/trino-main/src/main/java/io/trino/metadata/FunctionSpecificityComparator.java
```diff
@@ -0,0 +1,474 @@
...
+import static com.google.common.base.Verify.verify;
+import static java.lang.String.CASE_INSENSITIVE_ORDER;
+import static java.util.Objects.requireNonNull;
+
+/**
+ * Compares already-applicable functions to determine which one is more specific for overload
+ * selection in {@link FunctionBinder}.
+ * <p>
+ * This helper is intentionally specialized to FunctionBinder's current needs. It is not a general
+ * partial order over arbitrary signatures or bindings.
```
findepi, line 46: Remove. The paragraph above said this.

core/trino-main/src/main/java/io/trino/metadata/FunctionSpecificityComparator.java
```diff
@@ -0,0 +1,474 @@
...
+ * <p>
+ * The comparison is performed in two phases:
+ * <ol>
+ * <li>Compare already-bound applicable functions using the existing forwardability rule: a bound
```
findepi, line 50: "existing" refers to code that was pre-existing before this PR started. Once the code is merged, it's confusing.

This needs rewording.

core/trino-main/src/main/java/io/trino/metadata/FunctionSpecificityComparator.java
```diff
@@ -0,0 +1,474 @@
...
+        this.typeCoercion = new TypeCoercion(typeManager::getType);
+    }
+
+    /**
+     * Compares two already-applicable functions for overload selection in {@link FunctionBinder}.
+     * <p>
+     * The caller must provide both the bound signature selected for the actual call site and the
+     * original declared signature for each function. Usage outside this exact FunctionBinder flow
+     * has not been considered.
+     */
```
findepi, line 93: Duplicates class-level javadoc and what java method signature says. Remove.

core/trino-main/src/main/java/io/trino/metadata/FunctionSpecificityComparator.java
```diff
@@ -0,0 +1,474 @@
...
+
+    /**
+     * Compares two type signatures structurally after applicability has already been established.
+     * Concrete types are always narrower than type variables at this stage.
+     */
+    private Specificity compareTypeSignatures(
+            TypeSignature left,
+            TypeSignature right,
+            Map<String, TypeVariableConstraint> leftConstraints,
+            Map<String, TypeVariableConstraint> rightConstraints)
```
findepi, line 176: @martint  can you PTAL?

findepi:
> @findepi Do you want to give this a final review?

Thank you @dain for the mention!
I do not understand the code yet. I hoped @martint  will review this. If he doesn't have time, i can try.

i had some questions

- https://github.com/trinodb/trino/pull/28791#discussion_r2973616734
  - why this matters: could we compare declared signatures only?
- https://github.com/trinodb/trino/pull/28791#discussion_r2973628837
  - why this matters: the code for comparing declared signatures is long and looks expensive. if we eliminate call-dependent `arity`, we can cache this, or even keep functions in registry sorted once

I may have further questions when i try to actually understand the declared signatures comparator. Sorry for being slow on this. The code isn't simple, kudos!

core/trino-main/src/main/java/io/trino/metadata/FunctionBinder.java
dain: I don't have a single builtin with exactly that decimal/double shape; that was the minimal synthetic repro I used to isolate the problem. The real family that exposed it in the codebase was arithmetic operators like +, where concrete overloads from BigintOperators, DoubleOperators, NumberOperators, and DecimalOperators stopped preserving the existing coercion ordering when I tried to use a purely structural signature comparison as the primary rule.

core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
dain: We do not currently have an example of that shape in Trino today. arity is still needed internally to expand varargs before comparing declarations, but I do not have a real current overload set where the tie-break outcome changes with arity.

That said, a simple example that would need a tie break on arity is f(), f(X), and f(X...): arity 0 selects f(), arity 1 selects f(X), and arity 2+ selects f(X...). So the declared-signature tie-break does depend on call-site arity in varargs cases, because arity determines how the declarations are expanded and compared.  This is useful if you want to specialize a few concrete arities while still keeping a varargs fallback, and it only matters after we already know we have multiple applicable signatures and need to choose among them.

wendigo:
@martint ptal

core/trino-main/src/main/java/io/trino/metadata/SignatureSpecificityComparator.java
findepi: > a simple example that would need a tie break on arity is f(), f(X), and f(X...): arity 0 selects f(), arity 1 selects f(X), and arity 2+ selects f(X...).

This is actually a bad example.
The compared overloads are only those matching a call site.
if a call site provides one or more argument to f, the f() cannot be selected and won't be among those being compared.

(The assumption that the comparator compares only applicable overloads for some call site is already there in the code doc)

martint:
I’m working on a revamp of the type inference engine, so this may be unnecessary.

Issue #10512, opened 2022-01-10
Metadata caching may return incorrect data


kokosing:
Guava cache that we are using for metadata caching in many places (like hive metastore caching or JDBC connector metadata caching) is vulnerable to: https://github.com/google/guava/issues/1881

The issue is causing that cache entry invalidation does not cancel (invalidate) the existing load execution and that could lead to stale data being loaded. This may lead to incorrect results.

kokosing:
The issue can be reproduced with: https://github.com/trinodb/trino/pull/10513

I tried to play with caffeine (as Internet suggested) to see if it helps, however it was still failing. And javadoc says behavior in such case is undefined: https://github.com/ben-manes/caffeine/blob/5e834f8fdcf3819a242da7a44d7e62c7fde0ab2a/caffeine/src/main/java/com/github/benmanes/caffeine/cache/Cache.java#L174

ben-manes:
Unfortunately this is a limitation due to decorating `ConcurrentHashMap`, because an iterator skips over in-flight computations. It's `clear()` method will block if it traverses over a pending load, but the externalized iterators do not. As the cache has to maintain additional metadata and functionality it has to know which entry was removed (e.g. to notify a removal listener).

A possible workaround is to use `AsyncCache` which stores a mapping to a `CompletableFuture<V>`. This way the map operation is immediate as the expensive computation is performed outside of its locks. This would load on a background thread, though you can have a caller-runs policy by using the classic putIfAbsent-style [memoization](https://github.com/ben-manes/caffeine/wiki/Faq#recursive-computations) (ala JCIP).

Another approach that is implementation agnostic is to use a generation id as part of the cache key. By simply incrementing the counter, the old entries cannot be queried for and will eventually be evicted. This is somewhat common for remote caches like memcached / redis. Ideally you might proactively discard existing entries so that only an in-flight load is retained. If those stragglers are a problem then the caller could validate that the entry is still part of the current generation and discard + retry if not. This strategy is likely the best as very explicit and easily understood in code, since you'll want to document that assertion to avoid future regressions.

kokosing:
The idea with generation id is really nice. However, it won't fly with guava cache as we also call invalidate per cache key. So we would need to do switch to Caffeinate or fix guava cache.

kokosing:
Also methods like `io.trino.plugin.jdbc.CachingJdbcClient#invalidateCache` have no chance to invalidate entries that are during load (or reload).

[...]

PR #28792, opened 2026-03-21
Refresh vended credentials during long-running Iceberg REST catalog queries

### LINKED ISSUE #25827: Issue with Vended Credential Renewal with Iceberg REST Catalog
Hi Team,

I'm integrating Trino with an external Iceberg REST Catalog and have enabled vended credentials using the setting:
`iceberg.rest-catalog.vended-credentials-enabled=true`
Basic queries on Iceberg tables are working as expected. However, when executing long-running queries that exceed the STS token's expiry duration (30 minutes), the query fails.

Upon reviewing the Trino logs, I observed that the client does not invoke the [/credentials endpoint](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L1184) to refresh the vended credentials. It only calls the [oauth/tokens endpoint](https://github.com/apache/iceberg/blob/main/open-api/rest-catalog-open-api.yaml#L171).

I've attached a relevant log snippet highlighting the failure for your reference.

> TIMELINE: Query 20250519_171345_00011_uzse9 :: FAILED (ICEBERG_CURSOR_ERROR) :: elapsed 1803662ms :: planning 22ms :: waiting 3266ms :: scheduling 3363ms :: running 1800315ms :: finishing 0ms :: begin 2025-05-20T07:07:37.337Z :: end 2025-05-20T07:37:40.999Z

Could you please help confirm if Trino is expected to refresh credentials during long-running queries, or if there’s additional configuration required?

[logs2.txt](https://github.com/user-attachments/files/20342851/logs2.txt)

Regards,
Ashwin

kaveti:
# Refresh vended credentials during long-running Iceberg REST catalog queries

## Problem

When a query against an Iceberg REST catalog exceeds the STS token lifetime (typically 30–60 minutes), workers fail with `ICEBERG_CURSOR_ERROR` because the vended credentials (S3 access key, secret key, session token) have expired:

```
Query failed: com.amazonaws.services.s3.model.AmazonS3Exception:
  The provided token has expired. (Service: Amazon S3; Status Code: 400)
```

This affects any long-running workload: large table scans over hundreds of partitions, `INSERT INTO ... SELECT` with expensive transformations, multi-table JOINs, and ETL jobs running 30–60+ minutes.

The root cause is that `SqlStage.extractTableCredentials()` fetches credentials **once at stage creation** and freezes them into an `ImmutableMap` shared across all tasks in that stage. Later tasks reuse the same stale snapshot — the coordinator never re-fetches for already-dispatched stages.

## Solution

Three-layer fix: coordinator cache → transport refresh → worker-side update.

### 1. Coordinator cache (`IcebergMetadata`)

`tableCredentialsCache` (`ConcurrentHashMap`) stores `IcebergTableCredentials` per table. Each entry carries its own `Optional<Instant> expiresAt` derived from the vended token's expiry property (S3, GCS, or Azure). On access, `shouldRefresh()` checks if the entry is within 5 minutes of expiry and reloads from the catalog if needed. Updates use `ConcurrentMap.replace(K, old, new)` for atomic swap.

`TableNotFoundException` during `CREATE TABLE AS SELECT` returns empty credentials instead of failing — the table doesn't exist yet.

### 2. Coordinator → Worker transport (`SqlStage` / `HttpRemoteTask`)

Changed `SqlStage.extractTableCredentials()` from returning a resolved `ImmutableMap<PlanNodeId, ConnectorTableCredentials>` to `extractTableCredentialSuppliers()` returning `Map<PlanNodeId, Supplier<ConnectorTableCredentials>>`. Each supplier calls back into `IcebergMetadata.getOrLoadTableCredentials()`.

`HttpRemoteTask` stores these suppliers and calls `resolveTableCredentials()` on **every** `sendUpdate()`, so each `TaskUpdateRequest` carries the freshest credentials available.

### 3. Worker-side update (`TaskContext` / Operators)

`TaskContext.tableCredentials` changed from `final ImmutableMap` to `volatile` with an `updateTableCredentials()` method. `SqlTask.updateTask()` calls it on every incoming `TaskUpdateRequest` — credentials are updated before splits or dynamic filters are processed.

All five credential-consuming operators re-read from `TaskContext` at the point of use instead of relying on captured fields:

- `TableScanOperator.getOutput()` — per-split, when creating page source
- `ScanFilterAndProjectOperatorFactory.create()` — per-driver creation
- `TableWriterOperator.createPageSink()` — when creating page sink
- `MergeWriterOperator.createOperator()` — when creating merge sink
- `LeafTableFunctionOperator.resetProcessor()` — per-split

Pattern: `freshCredentials.or(() -> tableCredentials)` — reads from `TaskContext` first, falls back to captured value.

### Multi-cloud credential detection

`IcebergTableCredentials.forProperties()` detects vended credentials from all three providers:

- **S3**: `s3.access-key-id`, expiry from `s3.session-token-expires-at-ms`
- **GCS**: `gcs.oauth2.token`, expiry from `gcs.oauth2.token-expires-at`
- **Azure ADLS**: `adls.sas-token.*` prefix, expiry from `adls.sas-token-expires-at-ms.*` (uses earliest across accounts)

When no vended credentials are detected, `expiresAt` is `Optional.empty()` — refresh is disabled entirely.

When expiry is present, `shouldRefresh()` triggers 5 minutes before the actual expiry (matching Iceberg's `VendedCredentialsProvider` pattern). When vended credentials are present but no expiry timestamp is found, a conservative 55-minute TTL is assumed (1 hour default STS lifetime minus 5-minute buffer).

### Multi-cloud credential detection

`IcebergTableCredentials.forProperties()` detects vended credentials from all three providers:

- **S3**: `s3.access-key-id`, expiry from `s3.session-token-expires-at-ms`
- **GCS**: `gcs.oauth2.token`, expiry from `gcs.oauth2.token-expires-at`
- **Azure ADLS**: `adls.sas-token.*` prefix, expiry from `adls.sas-token-expires-at-ms.*` (uses earliest across accounts)

When no vended credentials are detected, `expiresAt` is `Optional.empty()` — refresh is disabled entirely.

When expiry is present, `shouldRefresh()` triggers 5 minutes before the actual expiry (matching Iceberg's `VendedCredentialsProvider` pattern). When vended credentials are present but no expiry timestamp is found, a conservative 55-minute TTL is assumed (1 hour default STS lifetime minus 5-minute buffer).

## Known limitations

- An already-open page source reading a single large file cannot refresh mid-read. In practice, individual splits complete in minutes, not hours — the refresh granularity is per-split, which covers the vast majority of real-world cases.
- Long-running single writes exceeding the token TTL are similarly constrained. This would require credential rotation support in the underlying file system client.

## Release notes

```
## Iceberg connector
* Add credential refresh for long-running queries against Iceberg REST catalogs that vend temporary cloud credentials. Credentials are now automatically refreshed on both coordinator and worker sides before they expire, preventing query failures due to expired STS tokens.
* Add configurable TTL properties: `iceberg.credential-cache-ttl`, `iceberg.coordinator-credential-prefetch`
```

  
Fixes #25827

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
+++ b/core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
@@ -146,7 +146,7 @@ public final class HttpRemoteTask
     private final String nodeId;
     private final AtomicBoolean speculative;
     private final PlanFragment planFragment;
-    private final Map<PlanNodeId, ConnectorTableCredentials> tableCredentials;
+    private final Map<PlanNodeId, Supplier<ConnectorTableCredentials>> tableCredentialSuppliers;
 
     private final AtomicLong nextSplitId = new AtomicLong();
 
@@ -220,7 +220,7 @@ public final class HttpRemoteTask
             boolean speculative,
             URI location,
             PlanFragment planFragment,
-            Map<PlanNodeId, ConnectorTableCredentials> tableCredentials,
+            Map<PlanNodeId, Supplier<ConnectorTableCredentials>> tableCredentialSuppliers,
             Multimap<PlanNodeId, Split> initialSplits,
             OutputBuffers outputBuffers,
             HttpClient httpClient,
@@ -250,7 +250,7 @@ public final class HttpRemoteTask
         requireNonNull(node, "node is null");
         requireNonNull(location, "location is null");
         requireNonNull(planFragment, "planFragment is null");
-        requireNonNull(tableCredentials, "tableCredentials is null");
+        requireNonNull(tableCredentialSuppliers, "tableCredentialSuppliers is null");
         requireNonNull(outputBuffers, "outputBuffers is null");
         requireNonNull(httpClient, "httpClient is null");
         requireNonNull(executor, "executor is null");
@@ -269,7 +269,7 @@ public final class HttpRemoteTask
             this.nodeId = node.getNodeIdentifier();
             this.speculative = new AtomicBoolean(speculative);
             this.planFragment = planFragment;
-            this.tableCredentials = ImmutableMap.copyOf(tableCredentials);
+            this.tableCredentialSuppliers = ImmutableMap.copyOf(tableCredentialSuppliers);
             this.outputBuffers.set(outputBuffers);
             this.httpClient = httpClient;
             this.executor = executor;
@@ -733,6 +733,16 @@ public final class HttpRemoteTask
         return false;
     }
 
+    private Map<PlanNodeId, ConnectorTableCredentials> resolveTableCredentials()
+    {
+        ImmutableMap.Builder<PlanNodeId, ConnectorTableCredentials> builder = ImmutableMap.builder();
+        for (Map.Entry<PlanNodeId, Supplier<ConnectorTableCredentials>> entry : tableCredentialSuppliers.entrySet()) {
+            Supplier<ConnectorTableCredentials> credentialSupplier = entry.getValue();
+            builder.put(entry.getKey(), credentialSupplier.get());
+        }
+        return builder.buildOrThrow();
+    }
+
     private void sendUpdate()
     {
         try {
@@ -773,7 +783,7 @@ public final class HttpRemoteTask
                 session.getIdentity().getExtraCredentials(),
                 stageSpan,
                 fragment,
-                tableCredentials,
+                resolveTableCredentials(),
                 splitAssignments,
                 outputBuffers.get(),
                 dynamicFilterDomains.getDynamicFilterDomains(),
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -13,12 +13,12 @@
  */
 package io.trino.plugin.iceberg;
 
+import com.google.common.annotations.VisibleForTesting;
 import com.google.common.base.Joiner;
 import com.google.common.base.Splitter;
 import com.google.common.base.Splitter.MapSplitter;
 import com.google.common.base.Suppliers;
 import com.google.common.base.VerifyException;
-import com.google.common.cache.CacheBuilder;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
@@ -26,7 +26,6 @@ import com.google.common.collect.Iterables;
 import com.google.common.collect.Lists;
 import com.google.common.collect.Sets;
 import com.google.common.collect.Streams;
-import com.google.common.util.concurrent.UncheckedExecutionException;
 import io.airlift.concurrent.MoreFutures;
 import io.airlift.json.JsonCodec;
 import io.airlift.log.Logger;
@@ -34,7 +33,6 @@ import io.airlift.slice.Slice;
 import io.airlift.slice.Slices;
 import io.airlift.units.DataSize;
 import io.airlift.units.Duration;
-import io.trino.cache.NonEvictableCache;
 import io.trino.filesystem.FileEntry;
 import io.trino.filesystem.FileIterator;
 import io.trino.filesystem.Location;
@@ -269,7 +267,6 @@ import java.util.stream.Stream;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
-import static com.google.common.base.Throwables.throwIfUnchecked;
 import static com.google.common.base.Verify.verify;
 import static com.google.common.base.Verify.verifyNotNull;
 import static com.google.common.collect.ImmutableList.toImmutableList;
@@ -280,8 +277,6 @@ import static com.google.common.collect.Iterables.size;
 import static com.google.common.collect.Maps.transformValues;
 import static com.google.common.collect.Sets.difference;
 import static io.airlift.units.Duration.ZERO;
-import static io.trino.cache.CacheUtils.uncheckedCacheGet;
-import static io.trino.cache.SafeCaches.buildNonEvictableCache;
 import static io.trino.filesystem.Locations.isS3Tables;
 import static io.trino.plugin.base.filter.UtcConstraintExtractor.extractTupleDomain;
 import static io.trino.plugin.base.projection.ApplyProjectionUtil.extractSupportedProjectedColumns;
@@ -524,7 +519,6 @@ public class IcebergMetadata
     private static final String DEPENDS_ON_NON_DETERMINISTIC_FUNCTIONS = "dependsOnNonDeterministicFunctions";
     // Value should be ISO-8601 formatted time instant
     private static final String TRINO_QUERY_START_TIME = "trino-query-start-time";
-
     private final TypeManager typeManager;
     private final JsonCodec<CommitTaskData> commitTaskCodec;
     private final TrinoCatalog catalog;
@@ -541,7 +535,9 @@ public class IcebergMetadata
     private final int materializedViewRefreshMaxSnapshotsToExpire;
     private final Duration materializedViewRefreshSnapshotRetentionPeriod;
     private final Map<IcebergTableHandle, AtomicReference<TableStatistics>> tableStatisticsCache = new ConcurrentHashMap<>();
-    private final NonEvictableCache<SchemaTableName, IcebergTableCredentials> tableCredentialsCache;
+    // Unbounded ConcurrentHashMap is acceptable here: IcebergMetadata is per-query,
+    // so entries only accumulate for tables accessed within a single query lifetime.
+    private final ConcurrentHashMap<SchemaTableName, IcebergTableCredentials> tableCredentialsCache = new ConcurrentHashMap<>();
     private final DeletionVectorWriter deletionVectorWriter;
 
     private Transaction transaction;
@@ -581,7 +577,6 @@ public class IcebergMetadata
         this.deletionVectorWriter = requireNonNull(deletionVectorWriter, "deletionVectorWriter is null");
         this.materializedViewRefreshMaxSnapshotsToExpire = materializedViewRefreshMaxSnapshotsToExpire;
         this.materializedViewRefreshSnapshotRetentionPeriod = materializedViewRefreshSnapshotRetentionPeriod;
-        this.tableCredentialsCache = buildNonEvictableCache(CacheBuilder.newBuilder());
     }
 
     @Override
@@ -605,18 +600,49 @@ public class IcebergMetadata
     private Optional<ConnectorTableCredentials> getOrLoadTableCredentials(ConnectorSession session, SchemaTableName schemaTableName)
     {
         try {
-            return Optional.of(uncheckedCacheGet(
-                tableCredentialsCache,
-                schemaTableName,
-                () -> {
-                    BaseTable baseTable = catalog.loadTable(session, schemaTableName);
-                    return new IcebergTableCredentials(baseTable.io().properties());
-                }));
+            // Check if existing cached entry is still fresh (per-entry expiry).
+            // Each IcebergTableCredentials carries its own expiresAt, so entries with
+            // different token lifetimes are handled correctly without a cache-wide static TTL.
+            IcebergTableCredentials cached = tableCredentialsCache.get(schemaTableName);
+            if (cached != null && !cached.shouldRefresh()) {
+                return Optional.of(cached);
+            }
+            // When refreshing an expired entry, invalidate the catalog-level table cache
+            // so loadTable() fetches a fresh BaseTable with updated vended credentials.
+            // Skip invalidation on initial load — the catalog cache already has the table.
+            if (cached != null) {
+                catalog.invalidateTableCache(schemaTableName);
+            }
+            IcebergTableCredentials fresh = loadTableCredentials(session, schemaTableName);
+            if (cached != null) {
+                tableCredentialsCache.replace(schemaTableName, cached, fresh);
+            }
+            else {
+                tableCredentialsCache.putIfAbsent(schemaTableName, fresh);
+            }
+            return Optional.of(fresh);
         }
-        catch (UncheckedExecutionException e) {
-            throwIfUnchecked(e.getCause());
-            throw e;
+        catch (TableNotFoundException e) {
+            // Table may not exist yet (e.g. during CREATE TABLE AS SELECT);
+            // no credentials to vend in that case.
+            return Optional.empty();
         }
+    }
+
+    private IcebergTableCredentials loadTableCredentials(ConnectorSession session, SchemaTableName schemaTableName)
+    {
+        BaseTable baseTable = catalog.loadTable(session, schemaTableName);
+        return IcebergTableCredentials.forFileIO(baseTable.io());
+    }
+
+    @VisibleForTesting
+    public void invalidateTableCredentialsCache()
+    {
+        // Replace each entry with an immediately-expired copy so the next
+        // getOrLoadTableCredentials() call sees cached != null with shouldRefresh() == true,
+        // triggering catalog cache invalidation and a fresh credential fetch.
+        tableCredentialsCache.replaceAll((_, value) ->
+                new IcebergTableCredentials(value.fileIoProperties(), Optional.of(Instant.EPOCH)));
     }
 
     private static SchemaTableName getSchemaTableName(ConnectorTableHandle tableHandle)
@@ -627,14 +653,6 @@ public class IcebergMetadata
         throw new IllegalArgumentException("Unsupported ConnectorTableHandle type: " + tableHandle.getClass().getName());
     }
 
-    private static SchemaTableName getSchemaTableName(ConnectorTableFunctionHandle tableFunctionHandle)
-    {
-        if (tableFunctionHandle instanceof TableChangesFunctionHandle handle) {
-            return handle.schemaTableName();
-        }
-        throw new IllegalArgumentException("Unsupported ConnectorTableFunctionHandle type: " + tableFunctionHandle.getClass().getName());
-    }
-
     private static SchemaTableName getSchemaTableName(ConnectorWritableTableHandle tableHandle)
     {
         return switch (tableHandle) {
@@ -642,6 +660,14 @@ public class IcebergMetadata
             case IcebergWritableTableHandle handle -> handle.name();
             default -> throw new IllegalArgumentException("Unsupported ConnectorWritableTableHandle type: " + tableHandle.getClass().getName());
         };
+    }
+
+    private static SchemaTableName getSchemaTableName(ConnectorTableFunctionHandle tableFunctionHandle)
+    {
+        if (tableFunctionHandle instanceof TableChangesFunctionHandle handle) {
+            return handle.schemaTableName();
+        }
+        throw new IllegalArgumentException("Unsupported ConnectorTableFunctionHandle type: " + tableFunctionHandle.getClass().getName());
     }
 
     @Override
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
@@ -15,20 +15,93 @@ package io.trino.plugin.iceberg;
 
 import com.google.common.collect.ImmutableMap;
 import io.trino.spi.connector.ConnectorTableCredentials;
+import org.apache.iceberg.aws.s3.S3FileIOProperties;
+import org.apache.iceberg.azure.AzureProperties;
+import org.apache.iceberg.gcp.GCPProperties;
 import org.apache.iceberg.io.FileIO;
 
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
+import java.util.Optional;
+import java.util.OptionalLong;
 
-public record IcebergTableCredentials(Map<String, String> fileIoProperties)
+import static java.util.Objects.requireNonNull;
+
+public record IcebergTableCredentials(Map<String, String> fileIoProperties, Optional<Instant> expiresAt)
         implements ConnectorTableCredentials
 {
+    // S3FileIOProperties.SESSION_TOKEN_EXPIRES_AT_MS is package-private in Iceberg,
+    // so we duplicate the string here. See apache/iceberg#11389.
+    private static final String S3_SESSION_TOKEN_EXPIRES_AT_MS = "s3.session-token-expires-at-ms";
+    // Default credential TTL when vended credentials are present but no expiry timestamp
+    // property is found. Matches the default AWS STS session token lifetime (1 hour).
+    private static final long DEFAULT_CREDENTIAL_TTL_MINUTES = 60;
+    // Prefetch buffer before credential expiry, matching Iceberg's VendedCredentialsProvider pattern.
+    private static final long PREFETCH_BUFFER_MINUTES = 5;
+
     public IcebergTableCredentials
     {
         fileIoProperties = ImmutableMap.copyOf(fileIoProperties);
+        requireNonNull(expiresAt, "expiresAt is null");
+    }
+
+    public boolean shouldRefresh()
+    {
+        return expiresAt.map(expiry -> Instant.now().isAfter(expiry)).orElse(false);
     }
 
     public static IcebergTableCredentials forFileIO(FileIO io)
     {
-        return new IcebergTableCredentials(io.properties());
+        return forProperties(io.properties());
+    }
+
+    static IcebergTableCredentials forProperties(Map<String, String> properties)
+    {
+        // Only compute an expiry when actual vended credentials are present.
+        // Empty expiresAt means credentials never expire, so catalogs that do not vend
+        // credentials are unaffected even if getTableCredentials() is called.
+        if (!hasVendedCredentials(properties)) {
+            return new IcebergTableCredentials(properties, Optional.empty());
+        }
+        // Extract the actual credential expiry from the properties, following the pattern in
+        // Iceberg's VendedCredentialsProvider (apache/iceberg#11389, #11577, #11282).
+        // Prefetch PREFETCH_BUFFER_MINUTES before actual expiry so shouldRefresh() triggers
+        // proactive refresh on lazy access (CachingKerberosAuthentication pattern).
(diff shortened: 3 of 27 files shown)


findinpath:
Pls rebase on master

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
findinpath, line 257: I think we should change this parameter from `Optional<ConnectorTableCredentials>` to `Supplier<Optional<ConnectorTableCredentials>>`


See https://github.com/apache/iceberg/pull/11577/ for reference

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
```diff
@@ -19,16 +19,25 @@
 
 import java.util.Map;
 
-public record IcebergTableCredentials(Map<String, String> fileIoProperties)
+public record IcebergTableCredentials(Map<String, String> fileIoProperties, long credentialsFetchedAtMs)
```
findinpath: You could flip and store an Instant when the credentials expire,  instead of the time when they were retrieved.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findinpath, line 600: Could you pls verify whether it would be enough to act in this method for retrieving "refreshed"  table credentials without probably any additional changes.

The `tableCredentialsCache` would need to be become evictable.
Expiration policy should be 1-2 minutes before the token expires.

I think this approach should likely solve the table credentials need.

kaveti: I traced the credential flow through Trino's execution model — the coordinator-side evictable cache alone wouldn't cover all cases.

The issue: extractTableCredentials() in [SqlStage](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/execution/SqlStage.java#L138) is called once at stage creation and the result is frozen into an ImmutableMap:

```

// SqlStage.java:160
this.tableCredentials = ImmutableMap.copyOf(tableCredentials);
Every task scheduled for that stage receives this same snapshot ([line 311](https://github.com/trinodb/trino/blob/master/core/trino-main/src/main/java/io/trino/execution/SqlStage.java#L311)) — getTableCredentials() is never called again for subsequent tasks within the same stage.

```

For a long-running query:

1. Stage is created → credentials fetched once and frozen
2. Hundreds of tasks are scheduled over the next 30+ minutes, all sharing the same credential snapshot from 
3. The coordinator cache may evict and be ready to serve fresh credentials, but nobody asks for them again

Cases this wouldn't solve:

Large scans — a single stage with many splits that takes longer than the credential TTL; later tasks get stale credentials
Long-running inserts/merges — a writer stage running for an extended period with the original frozen credentials

kaveti:
@coderabbitai  full review

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -579,7 +579,16 @@ public IcebergMetadata(
...
         this.materializedViewRefreshMaxSnapshotsToExpire = materializedViewRefreshMaxSnapshotsToExpire;
         this.materializedViewRefreshSnapshotRetentionPeriod = materializedViewRefreshSnapshotRetentionPeriod;
-        this.tableCredentialsCache = buildNonEvictableCache(CacheBuilder.newBuilder());
+        this.credentialCacheTtl = requireNonNull(credentialCacheTtl, "credentialCacheTtl is null");
+        // Credentials are cached with a TTL so that the coordinator will re-fetch fresh credentials
+        // when getTableCredentials() is called for new tasks in long-running queries.
+        // The cache evicts coordinatorCredentialPrefetch before the token expires,
+        // ensuring the next getTableCredentials() call re-fetches while the old token is still valid.
+        // EvictableCache guarantees that invalidate() is immediately visible to subsequent get() calls.
+        this.tableCredentialsCache = EvictableCacheBuilder.newBuilder()
```
findinpath: I think that coming up with "statically" configured values for configuring the credentials cache is not the solution here.

The options I see we have at hand:

We could make use Caffeine

> In Caffeine, you can define an Expiry object:

```
Cache<String, String> cache = Caffeine.newBuilder()
    .expireAfter(new Expiry<String, String>() {
        public long expireAfterCreate(String key, String value, long currentTime) {
            // Logic to return custom duration in nanoseconds
            return TimeUnit.MINUTES.toNanos(customDurationForKey(key));
        }
        public long expireAfterUpdate(...) { ... }
        public long expireAfterRead(...) { ... }
    })
    .build();
```
However, Trino project is still not using (at the time of this writing) Caffeine
https://github.com/trinodb/trino/issues/10512#issuecomment-1016221168
cc @findepi 

Otherwise, you'd have to cobble together a "synchronized" mechanism of dealing with `tableCredentialsCache` as a ConcurrentHashMap and invalidate/remove entries that are expired/close to be expired and recompute / refresh them.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
```diff
@@ -28,7 +28,8 @@ public record FilesTableSplit(
         String schemaJson,
         String metadataTableJson,
         Map<Integer, String> partitionSpecsByIdJson,
-        Optional<Type> partitionColumnType)
+        Optional<Type> partitionColumnType,
+        Map<String, String> fileIoProperties)
```
findinpath: Is it intentional to undo the changes done to reuse the table credentials mechanism from the planning?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findinpath: @kaveti getting back on my suggestion after receiving a hint from @findepi 

We can keep the `tableCredentialsCache` as a non-evictable cache.
However, instead of retrieving a POJO for the table credentials, we can build functionality similar to what has been done in `io.trino.plugin.base.authentication.CachingKerberosAuthentication` to refreshIfSoonWillBeExpired

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
kaveti: yeah. i will revert this codechange. my bad.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
kaveti: will follow same pattern as defined in CachingKerberosAuthentication

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
kaveti: supplier can't be serialized across the coordinator → worker.

findinpath: it doesn't need to be serialized - you're just keeping a supplier in the `tableCredentialsCache` but actually offerring further `TableCredentials` from `getOrLoadTableCredentials()` method

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/functions/tablechanges/TableChangesFunctionProcessor.java
```diff
@@ -186,6 +186,15 @@ public TableFunctionProcessorState process()
...
     }
 
+    private static Map<String, String> getEffectiveFileIoProperties(Optional<ConnectorTableCredentials> tableCredentials, Map<String, String> splitFileIoProperties)
+    {
+        Map<String, String> credentialProperties = getFileIoProperties(tableCredentials);
+        if (!credentialProperties.isEmpty()) {
+            return credentialProperties;
+        }
+        return splitFileIoProperties;
+    }
```
findinpath: is this method still needed?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
findinpath: > Stage is created → credentials fetched once and frozen

Then we'll need to retrieve the table credentials right before sending `TaskUpdateRequest` to the workers.

https://github.com/trinodb/trino/blob/9cb6ee529ad845c33ad095be81042d34b69547f9/core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java#L771-L781

Consider changing `SqlStage.java` from relying on a table credentials to supplier of table credentials

Now: `Map<PlanNodeId, ConnectorTableCredentials>`
After the changes: `Map<PlanNodeId, Supplier<ConnectorTableCredentials>>`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTableSplit.java
```diff
@@ -33,6 +34,15 @@ public record FilesTableSplit(
 {
     private static final int INSTANCE_SIZE = instanceSize(FilesTableSplit.class);
 
+    public FilesTableSplit
```
findinpath: intentional ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSinkProvider.java
```diff
@@ -85,20 +99,26 @@ public IcebergPageSinkProvider(
...
+                .expireAfterWrite(WORKER_CREDENTIAL_CACHE_TTL.toMillis(), MILLISECONDS)
+                .shareNothingWhenDisabled()
+                .build();
     }
 
     @Override
     public ConnectorPageSink createPageSink(ConnectorTransactionHandle transactionHandle, ConnectorSession session, ConnectorOutputTableHandle outputTableHandle, Optional<ConnectorTableCredentials> tableCredentials, ConnectorPageSinkId pageSinkId)
     {
         IcebergWritableTableHandle tableHandle = (IcebergWritableTableHandle) outputTableHandle;
-        return createPageSink(session, tableHandle, getFileIoProperties(tableCredentials));
```
findinpath, line 94: The credentials should be "fresh" now because they were just retrieved when building the TaskUpdateRequest?

Why do we need to do "maybeRefreshCredentials" ?

core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
```diff
@@ -733,6 +733,23 @@ boolean adjustSplitBatchSize(List<SplitAssignment> splitAssignments, long reques
...
+    {
+        ImmutableMap.Builder<PlanNodeId, ConnectorTableCredentials> builder = ImmutableMap.builder();
+        for (Map.Entry<PlanNodeId, Supplier<ConnectorTableCredentials>> entry : tableCredentialSuppliers.entrySet()) {
+            try {
+                builder.put(entry.getKey(), entry.getValue().get());
+            }
+            catch (RuntimeException e) {
+                // Supplier may fail if the query transaction has ended (race between
+                // sendUpdate and task termination).
+                // cancelled and the worker will not use the credentials.
```
findinpath: what is this last statement about ? it seems not to be a continuation of the previous statement.

core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
```diff
@@ -733,6 +733,23 @@ boolean adjustSplitBatchSize(List<SplitAssignment> splitAssignments, long reques
         return false;
     }
 
+    private Map<PlanNodeId, ConnectorTableCredentials> resolveTableCredentials()
+    {
+        ImmutableMap.Builder<PlanNodeId, ConnectorTableCredentials> builder = ImmutableMap.builder();
+        for (Map.Entry<PlanNodeId, Supplier<ConnectorTableCredentials>> entry : tableCredentialSuppliers.entrySet()) {
+            try {
+                builder.put(entry.getKey(), entry.getValue().get());
```
findinpath: `entry.getValue().get()` hides the fact that we're calling a supplier here.
i recommend using a variable for `entry.getValue()`

core/trino-main/src/main/java/io/trino/server/remotetask/HttpRemoteTask.java
```diff
@@ -733,6 +733,23 @@ boolean adjustSplitBatchSize(List<SplitAssignment> splitAssignments, long reques
...
+            }
+            catch (RuntimeException e) {
```
findinpath: if there's a failure here, i doubt that it makes sense to continue the query.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
```diff
@@ -146,12 +144,6 @@ public Optional<ConnectorSplitSource> splitSource(ConnectorSession connectorSess
                 partitionColumnType));
     }
 
-    @Override
```
findinpath, line 149: why is this operation not needed anymore?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSinkProvider.java
```diff
@@ -85,20 +99,26 @@ public IcebergPageSinkProvider(
         this.sortingFileWriterLocalStagingPath = icebergConfig.getSortedWritingLocalStagingPath();
         this.typeManager = requireNonNull(typeManager, "typeManager is null");
         this.pageSorter = requireNonNull(pageSorter, "pageSorter is null");
+        this.catalogFactory = requireNonNull(catalogFactory, "catalogFactory is null");
+        this.refreshedCredentialCache = EvictableCacheBuilder.newBuilder()
```
findinpath: i don't think we should have any caching here. the provided table credentials are supposed to be fresh.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
```diff
@@ -15,20 +15,85 @@
...
 
-public record IcebergTableCredentials(Map<String, String> fileIoProperties)
+import static java.util.Objects.requireNonNull;
+
+public record IcebergTableCredentials(Map<String, String> fileIoProperties, Instant expiresAt)
         implements ConnectorTableCredentials
 {
+    // S3FileIOProperties.SESSION_TOKEN_EXPIRES_AT_MS is package-private in Iceberg,
+    // so we duplicate the string here. See apache/iceberg#11389.
+    private static final String S3_SESSION_TOKEN_EXPIRES_AT_MS = "s3.session-token-expires-at-ms";
```
findinpath, line 36: Use constants from iceberg org.apache.iceberg.aws.s3.S3FileIOProperties#SESSION_TOKEN_EXPIRES_AT_MS

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
```diff
@@ -15,20 +15,85 @@
...
 
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
+import java.util.OptionalLong;
 
-public record IcebergTableCredentials(Map<String, String> fileIoProperties)
+import static java.util.Objects.requireNonNull;
+
+public record IcebergTableCredentials(Map<String, String> fileIoProperties, Instant expiresAt)
```
findinpath: we do support now in `trinodb/trino` credential vending for all cloud storage providers.

we should handle all of them

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSinkProvider.java
kaveti: yeah this would not be needed anymore.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/functions/tablechanges/TableChangesFunctionProcessor.java
kaveti: nope

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
kaveti: it is   static final java.lang.String SESSION_TOKEN_EXPIRES_AT_MS;

package-private in Iceberg 1.10.1.  i think we cannot import it

kaveti: i think we need to open pr upstream for making it visible. but i'm not sure the intension behind keeping it package private.

kaveti: this is made public in this PR: https://github.com/apache/iceberg/pull/15249/changes . post this there is no official stable version published.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -596,29 +584,34 @@ public Optional<ConnectorTableCredentials> getTableCredentials(ConnectorSession
...
+            // Check if existing cached entry is still fresh (per-entry expiry).
+            // Each IcebergTableCredentials carries its own expiresAt, so entries with
+            // different token lifetimes are handled correctly without a cache-wide static TTL.
+            IcebergTableCredentials cached = tableCredentialsCache.get(schemaTableName);
+            if (cached != null && !cached.shouldRefresh()) {
+                return Optional.of(cached);
+            }
+            // Entry is missing or approaching expiry — reload from catalog
+            IcebergTableCredentials fresh = loadTableCredentials(session, schemaTableName);
+            tableCredentialsCache.put(schemaTableName, fresh);
```
findinpath: One could use as well `java.util.concurrent.ConcurrentMap#replace(K, V, V)`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -1630,7 +1615,9 @@ private List<String> getChildNamespaces(ConnectorSession session, String parentN
 
     private IcebergWritableTableHandle newWritableTableHandle(SchemaTableName name, Table table)
     {
-        tableCredentialsCache.put(name, IcebergTableCredentials.forFileIO(table.io()));
+        // Remove so the next getOrLoadTableCredentials() call re-loads fresh credentials
```
findinpath: The removal rationale is not very clear to me.
Why do the credentials need to be reloaded?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
```diff
@@ -15,20 +15,87 @@
...
+        String s3Expiry = properties.get(S3_SESSION_TOKEN_EXPIRES_AT_MS);
+        if (s3Expiry != null) {
+            return OptionalLong.of(Long.parseLong(s3Expiry));
+        }
+        // GCS: gcs.oauth2.token-expires-at (apache/iceberg#11282)
+        String gcsExpiry = properties.get(GCPProperties.GCS_OAUTH2_TOKEN_EXPIRES_AT);
+        if (gcsExpiry != null) {
+            return OptionalLong.of(Long.parseLong(gcsExpiry));
+        }
+        return OptionalLong.empty();
```
findinpath, line 98: pls handle `AzureProperties.ADLS_SAS_TOKEN_EXPIRES_AT_MS_PREFIX` as well.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableCredentials.java
```diff
@@ -15,20 +15,87 @@
...
 
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.Map;
+import java.util.OptionalLong;
 
-public record IcebergTableCredentials(Map<String, String> fileIoProperties)
+import static java.util.Objects.requireNonNull;
+
+public record IcebergTableCredentials(Map<String, String> fileIoProperties, Instant expiresAt)
```
findinpath: expiresAt can be `Optional<Instant>`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergMetadataCredentialCacheWithRestCatalog.java
```diff
@@ -0,0 +1,308 @@
...
+        assertThat(tableHandle).isNotNull();
+
+        Optional<ConnectorTableCredentials> first = metadata.getTableCredentials(SESSION, tableHandle);
+        assertThat(first).isPresent();
+        IcebergTableCredentials firstCredentials = (IcebergTableCredentials) first.get();
+        String firstAccessKey = firstCredentials.fileIoProperties().get("s3.access-key-id");
+        int loadsAfterFirst = vendingAdapter.getLoadTableCount();
+
+        // Inject an expired entry into the credentials cache to simulate token expiry
+        injectExpiredCredentialsCacheEntry(metadata, firstCredentials);
```
findinpath: you can have a package-protected `@VisibleForTesting`  getter for the `tableCredentialsCache` instead of doing reflection because the mechanism is rather brittle.

or even better

`invalidateTableCredentialsCache()`

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergTableCredentials.java
```diff
@@ -0,0 +1,164 @@
...
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.Map;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class TestIcebergTableCredentials
+{
+    @Test
```
findinpath, line 31: add handling for Azure

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
kaveti: You're right — reverted it. The shouldRefresh() check in getOrLoadTableCredentials() already handles stale credentials based on per-entry expiry, so eagerly invalidating here was unnecessary.

findinpath:
Claude

> The coordinator-side refresh is necessary but not sufficient. For truly long-running tasks (processing large           
  partitions, slow I/O, etc.), credentials could expire mid-operation on the worker.
                                                                                                                         
>  To fully solve this, the worker would need either:                                                                   
>  1. Accept credential updates from subsequent TaskUpdateRequest messages (add updateTableCredentials() to TaskContext)
>  2. Or implement its own refresh mechanism (passing credential suppliers to workers instead of resolved credentials

It seems that my suggestion of handling the refreshing of the credentials on the coordinator side was rather naive.
Apologies @kaveti , it seems that the functionality needs to be reworked to allow the refresh of the credentials on the worker side.

kaveti:
> Claude
> 
> > The coordinator-side refresh is necessary but not sufficient. For truly long-running tasks (processing large
> > partitions, slow I/O, etc.), credentials could expire mid-operation on the worker.
> 
> > To fully solve this, the worker would need either:
> > 
> > 1. Accept credential updates from subsequent TaskUpdateRequest messages (add updateTableCredentials() to TaskContext)
> > 2. Or implement its own refresh mechanism (passing credential suppliers to workers instead of resolved credentials
> 
> It seems that my suggestion of handling the refreshing of the credentials on the coordinator side was rather naive. Apologies @kaveti , it seems that the functionality needs to be reworked to allow the refresh of the credentials on the worker side.

FYI, Known limitations, even with these change are.
- Already-open page source reading a huge file can't refresh mid-read
- Long-running single writes exceeding TTL

findinpath:
> The coordinator-side refresh is necessary but not sufficient. For truly long-running tasks (processing large partitions, slow I/O, etc.), credentials could expire mid-operation on the worker.

> It seems that my suggestion of handling the refreshing of the credentials on the coordinator side was rather naive.

@kaveti i've been looking through the implementation you have currently and reached the following conclusion:

The suggestion of working with Suppliers for the table credentials is "good enough" for the time being.
We'll have to deal indeed with the situations when a task takes longer than the validity of the credentials.
For this we would need more involved changes on the trino-filesystem-xxx classes that relate to `IcebergRestCatalogFileSystemFactory` in order to reuse the following `apache/iceberg` classes in order to refresh transparently the table credentials:
- aws: `org.apache.iceberg.aws.s3.VendedCredentialsProvider`
- azure: `org.apache.iceberg.azure.adlsv2.VendedAdlsCredentialProvider`
- gcs: `org.apache.iceberg.gcp.gcs.OAuth2RefreshCredentialsHandler`

I would suggest though that this operation to be done EVENTUALLY in a follow-up.

Therefore, I advise wrapping up the work here, make sure that the commit  message (as well as the description of the PR) is well documented and let's continue with the PR having the current line of thinking.

findinpath:
@kaveti  sorry for going back and forth on this PR.
I want to explore a bit the avenue  - i'll be getting back with feedback later this week.

Pausing for now the review of this contribution.

> For this we would need more involved changes on the trino-filesystem-xxx classes that relate to IcebergRestCatalogFileSystemFactory in order to reuse the following apache/iceberg classes in order to refresh transparently the table credentials:
>
> aws: org.apache.iceberg.aws.s3.VendedCredentialsProvider
> azure: org.apache.iceberg.azure.adlsv2.VendedAdlsCredentialProvider
> gcs: org.apache.iceberg.gcp.gcs.OAuth2RefreshCredentialsHandler

kaveti:
> Pausing for now the review of this contribution.
> 
> >

yeah sure @findinpath

PR #28794, opened 2026-03-21
Hive: Preserve custom partition locations during INSERT OVERWRITE

### LINKED ISSUE #27271: Can't INSERT INTO existing partitions of Hive external table when partition location is customized
As title, when INSERT INTO an existing partition of an external Hive table with customized partition location, the partition won't be updated.
Can reproduce the issue in version 435 and 476.

### Reproducing Flow

1. Create an external table.
```
 CREATE TABLE tmp2 (v integer, k varchar)
 WITH (
    external_location = 's3://<table_location>',
    partitioned_by = ARRAY['k']
 );
```

2. Register a partition k=k1 to a customized location k=k1_plus
```
CALL system.register_partition('<schema>', 'tmp2', array['k'], array['k1'], 's3://<table_location>/k=k1_plus');
```

3. Change insert behavior to OVERWRITE
```
SET SESSION hive.insert_existing_partitions_behavior='OVERWRITE';
```

4. INSERT INTO the table
```
INSERT INTO tmp2 values(1, 'k1'),(2, 'k2');
```

5. Check the contents with paths
```
select *, "$path" from tmp2;
```
Result will look like
```
 v | k  |                                                             $path
---+----+-------------------------------------------------------------------------------------------------------------------------------
 2 | k2 | s3://<table_location>/k=k2/<some_file_name>
(1 row)
```
Note that the first row (1, 'k1') disappeared. 
Further more, we can find objects created by Trino under the path `s3://<table_location>/k=k1/`, though it should be created under `s3://<table_location>/k=k1_plus/`

### Investigation
Per my understanding, this issue is because the existing partition information is completely ignored when determining the target write path in [this line](https://github.com/trinodb/trino/blob/1b0b0d6e37655e3c0bf023b96fa6ae229e39b45d/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveWriterFactory.java#L406), while the potential partition change is not committed to metastore.

Kundannitp:
Fixes #27271

Hive: Fix custom partition location on overwrite

Description
When using insert_existing_partitions_behavior='OVERWRITE', Trino incorrectly ignored custom partition locations registered via system.register_partition. Instead of replacing data in the custom URI, it reverted to the default partition path.

This caused cross-directory rename failures on object stores like S3 (IOException: S3 does not support renames), and silent data movement to the wrong directory on HDFS, breaking the custom partition registration entirely.

This PR ensures the overwrite logic correctly resolves and respects existing custom partition URIs, maintaining the integrity of external data pipelines.

Additional context and related issues
Previously, when evaluating the write path for an OVERWRITE operation, the code passed an empty Optional for the partition into the getPartitionWriteInfo() function. This caused the function to treat the partition as brand new, returning a path that used the default directory nomenclature and abandoning the actual custom directory.

In this change, we pass the existing partition as a parameter instead. This allows getPartitionWriteInfo() to correctly determine and return the custom directory. Because there is already existing logic to safely clean up the target directory during an OVERWRITE, passing the existing partition correctly resolves the location without accidentally triggering APPEND behavior.

Release notes

Code snippet
## Bug Fixes
* Hive
  * Fix an issue where `INSERT OVERWRITE` ignores custom partition locations and incorrectly moves data to the default partition path. ({issue}`#27271`)

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveWriterFactory.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveWriterFactory.java
@@ -403,7 +403,7 @@ public class HiveWriterFactory
                     compressionCodec = selectCompressionCodec(session, partitionStorageFormat);
                     schema.putAll(getHiveSchema(table));
 
-                    writeInfo = locationService.getPartitionWriteInfo(locationHandle, Optional.empty(), partitionName.get());
+                    writeInfo = locationService.getPartitionWriteInfo(locationHandle, partition, partitionName.get());
                     break;
                 case ERROR:
                     throw new TrinoException(HIVE_PARTITION_READ_ONLY, "Cannot insert into an existing partition of Hive table: " + partitionName.get());
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseTestHiveOnDataLake.java
@@ -108,6 +108,7 @@ abstract class BaseTestHiveOnDataLake
                 .setHiveProperties(
                         ImmutableMap.<String, String>builder()
                                 .put("hive.insert-existing-partitions-behavior", "OVERWRITE")
+                                .put("hive.allow-register-partition-procedure", "true")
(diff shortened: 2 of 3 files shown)


Kundannitp:
@martint and @chenjian2664 can you please review this PR

PR #28795, opened 2026-03-22
Scan only required fields when reading Iceberg manifests

### LINKED ISSUE #28787: Improve schema projection performance for Iceberg tables
### Trino version

446.1.198

### Please describe the bug

Let's say we have an iceberg table like this,
```
CREATE TABLE
  db_1.tbl_1 (
    col_1 string,
    col_2 string,
    col_3 string
  )
```
If we have a query in trino,
`SELECT col_1 FROM db_1.tbl_1 WHERE col_2 = 'my data'`

The current schema projection for this query will include all of columns (`col_1, col_2, col_3`) regardless of the exact projection and filtering in the query.

The expected result of schema projection based on the query above should be `col_1, col_2`

### How to reproduce:
Write a simple query in TestIcebergSplitSource.java
`@Test
    public void testMyQuery()
    {
        assertUpdate("CREATE TABLE employees (id BIGINT, name VARCHAR, contacts ARRAY(ROW(email VARCHAR, phone VARCHAR, city VARCHAR)), dt VARCHAR) WITH (partitioning = ARRAY['dt'])");
        assertThat(computeActual("SELECT contacts[1].email FROM employees where dt = '123'").getRowCount()).isEqualTo(2);
    }`
Check how table is scanned in `getNextBatch` in IcebergSplitSource.java

Image

ebyhr:
## Description

- Fixes #28787

## Release notes

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
@@ -35,18 +35,24 @@ import org.apache.iceberg.CombinedScanTask;
 import org.apache.iceberg.DataOperations;
 import org.apache.iceberg.FileScanTask;
 import org.apache.iceberg.Scan;
+import org.apache.iceberg.Schema;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.Table;
 import org.apache.iceberg.metrics.InMemoryMetricsReporter;
 import org.apache.iceberg.metrics.MetricsReporter;
+import org.apache.iceberg.types.Types;
 import org.apache.iceberg.util.SnapshotUtil;
 
+import java.util.List;
+import java.util.Objects;
 import java.util.concurrent.ExecutorService;
 
+import static com.google.common.collect.ImmutableList.toImmutableList;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.getDynamicFilteringWaitTimeout;
 import static io.trino.plugin.iceberg.IcebergSessionProperties.getMinimumAssignedSplitWeight;
 import static io.trino.spi.connector.FixedSplitSource.emptySplitSource;
 import static java.util.Objects.requireNonNull;
+import static org.apache.iceberg.util.SnapshotUtil.schemaFor;
 
 public class IcebergSplitManager
         implements ConnectorSplitManager
@@ -145,8 +151,17 @@ public class IcebergSplitManager
             // (deletes or overwrites), so we cannot perform incremental refresh. Falling back to full refresh.
             icebergMetadata.disableIncrementalRefresh();
         }
+
+        Schema schema = schemaFor(icebergTable, table.getSnapshotId().getAsLong());
+        List<String> names = table.getProjectedColumns().stream()
+                .map(column -> schema.findField(column.getId()))
+                .filter(Objects::nonNull) // Newly added column may not be found in current snapshot schema until new files are added
+                .map(Types.NestedField::name)
+                .collect(toImmutableList());
+
         return icebergTable.newScan()
                 .useSnapshot(table.getSnapshotId().getAsLong())
+                .project(schema.select(names)) // Using Scan.project method because Scan.select throws an exception for nested variant
                 .planWith(executor)
                 .metricsReporter(metricsReporter);
     }
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -9613,7 +9613,7 @@ public abstract class BaseIcebergConnectorTest
     {
         MaterializedResultWithPlan result = getDistributedQueryRunner().executeWithPlan(getSession(), "SELECT name FROM region");
         assertThat(getOperatorStats(result.queryId()).getConnectorMetrics().getMetrics())
-                .contains(entry("projectedFieldIds", new IntList(ImmutableList.of(1, 2, 3))));
+                .contains(entry("projectedFieldIds", new IntList(ImmutableList.of(2))));
     }
 
     @Test


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergSplitManager.java
```diff
@@ -0,0 +1,121 @@
...
+        ClassLoaderSafeConnectorSplitSource splitSource = (ClassLoaderSafeConnectorSplitSource) splitManager.getSplits(
+                transactionHandle,
+                IcebergTestUtils.SESSION,
+                tableHandle(new SchemaTableName("tpch", "region"), "regionkey"),
+                DynamicFilter.EMPTY,
+                Constraint.alwaysTrue());
+        getFutureValue(splitSource.getNextBatch(1000));
+
+        assertThat(splitSource.getMetrics().getMetrics())
+                .contains(entry("projectedFields", new LongCount(3)));
```
raunaqmorarka: We should be able to test for this metric through normal query execution since the split source metrics are made part of the scan operator metrics.
Or you can test in TestIcebergSplitSource

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -537,6 +537,7 @@ public Metrics getMetrics()
         ScanMetricsResult scanMetrics = scanReport.scanMetrics();
         return new Metrics(ImmutableMap.<String, Metric<?>>builder()
                 .put("scanPlanningDuration", new DurationTiming(Duration.succinctDuration(scanMetrics.totalPlanningDuration().totalDuration().toMillis(), MILLISECONDS)))
+                .put("projectedFields", new LongCount(scanReport.projectedFieldNames().size()))
```
raunaqmorarka: We should print either projectedFieldIds, or projectedFieldNames, just size omits useful information

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
```diff
@@ -145,8 +151,17 @@ private Scan<?, FileScanTask, CombinedScanTask> getScan(IcebergMetadata icebergM
...
+        Schema schema = schemaFor(icebergTable, table.getSnapshotId().getAsLong());
+        List<String> names = table.getProjectedColumns().stream()
+                .map(column -> schema.findField(column.getId()))
+                .filter(Objects::nonNull) // Newly added column may not be found in current snapshot schema until new files are added
+                .map(Types.NestedField::name)
+                .collect(toImmutableList());
+
         return icebergTable.newScan()
                 .useSnapshot(table.getSnapshotId().getAsLong())
+                .project(schema.select(names))
```
raunaqmorarka: Can we use `.select(names)` ?

raunaqmorarka: What's the material impact of this ? 
I assume we just get to have a bit lesser metadata in memory from scanning manifests and the rest stays the same ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
```diff
@@ -145,8 +151,17 @@ private Scan<?, FileScanTask, CombinedScanTask> getScan(IcebergMetadata icebergM
             // (deletes or overwrites), so we cannot perform incremental refresh. Falling back to full refresh.
             icebergMetadata.disableIncrementalRefresh();
         }
+
+        Schema schema = schemaFor(icebergTable, table.getSnapshotId().getAsLong());
+        List<String> names = table.getProjectedColumns().stream()
```
chenjian2664, line 156: What if `table.getProjectedColumns()` is empty, are we allowing to project empty schema?

findepi:
> Scan only required fields in Iceberg

is it for data files or manifest?

(please update PR title and description)

ebyhr:
(Just rebased on master without any changes)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
ebyhr: Yes, the method may return an empty list for aggregation queries such as `SELECT count(*) FROM region`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
ebyhr: Yeah, that should be the impact of this change.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
raunaqmorarka: ping

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitManager.java
ebyhr: It seems `.select(names)` throws an exception for nested variant. I'll revert to `.project(schema.select(names))`.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
ebyhr: @raunaqmorarka Added a new Metric implementation. Could you take another look?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -537,6 +538,8 @@ public Metrics getMetrics()
         ScanMetricsResult scanMetrics = scanReport.scanMetrics();
         return new Metrics(ImmutableMap.<String, Metric<?>>builder()
                 .put("scanPlanningDuration", new DurationTiming(Duration.succinctDuration(scanMetrics.totalPlanningDuration().totalDuration().toMillis(), MILLISECONDS)))
+                .put("totalProjectedFields", new LongCount(scanReport.projectedFieldIds().size()))
+                .put("projectedFieldIds", new IntList(scanReport.projectedFieldIds().subList(0, Math.min(scanReport.projectedFieldIds().size(), 100)))) // Limit to 100 field ids to avoid reporting huge metrics
```
raunaqmorarka: I'm not sure the limit 100 is needed, queryinfo json and potentially other things in query completion event will definitely carry far heavier structures than this.
If you are truncating, then we need to indicate it is truncated and record the full size.
Probably not worth complicating it. You can just print the whole list

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -537,6 +538,8 @@ public Metrics getMetrics()
         ScanMetricsResult scanMetrics = scanReport.scanMetrics();
         return new Metrics(ImmutableMap.<String, Metric<?>>builder()
                 .put("scanPlanningDuration", new DurationTiming(Duration.succinctDuration(scanMetrics.totalPlanningDuration().totalDuration().toMillis(), MILLISECONDS)))
+                .put("totalProjectedFields", new LongCount(scanReport.projectedFieldIds().size()))
```
raunaqmorarka: Do we need totalProjectedFields, given that we are keeping the whole list ?

PR #28796, opened 2026-03-22
Add Databricks connector

bvolpato:
## Description

Add a new JDBC-based connector for [Databricks](https://www.databricks.com/) SQL warehouses. The connector connects using the [Databricks JDBC driver](https://docs.databricks.com/en/integrations/jdbc/index.html) and supports querying data through Databricks Unity Catalog.

The connector follows the same patterns as the existing Snowflake connector, using `BaseJdbcClient` with:

- Token-based authentication via the Databricks JDBC driver
- Type mappings for Databricks SQL types (BOOLEAN, TINYINT, SMALLINT, INT, BIGINT, FLOAT, DOUBLE, DECIMAL, STRING, VARCHAR, DATE, TIMESTAMP)
- Aggregation pushdown (avg, count, min, max, sum, stddev, variance, covariance, correlation, regression)
- Limit and TopN pushdown
- Backtick quoting for Databricks SQL compatibility
- Unity Catalog support via `databricks.catalog` and `databricks.schema` config properties
- `query()` table function for pass-through native SQL queries

The Databricks JDBC driver does not support `readOnly` mode, so `getConnection()` is overridden to skip `setReadOnly(true)`.

### Files included

- `plugin/trino-databricks/` — connector plugin with Java sources, tests, and pom.xml
- `docs/src/main/sphinx/connector/databricks.md` — connector documentation
- `docs/src/main/sphinx/connector.md` — connector index entry
- `pom.xml` — root POM changes (module, dependency management)

## Testing

- Unit tests: `TestDatabricksPlugin` (plugin smoke test), `TestDatabricksConfig` (config property mapping)
- Tested locally against a live Databricks SQL warehouse on Trino 479:
  - `SHOW SCHEMAS` ✅
  - `SHOW TABLES` ✅
  - `DESCRIBE` ✅
  - `SELECT * FROM ... LIMIT` ✅
  - `SELECT count(*)` ✅
  - `SELECT ... GROUP BY ... ORDER BY` with `sum()` and `avg()` ✅

## Release notes

```markdown
# General
* Add Databricks connector.
```

### DIFF
--- a/plugin/trino-databricks/pom.xml
+++ b/plugin/trino-databricks/pom.xml
@@ -1,0 +1,266 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.trino</groupId>
+        <artifactId>trino-root</artifactId>
+        <version>481-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>trino-databricks</artifactId>
+    <packaging>trino-plugin</packaging>
+    <name>${project.artifactId}</name>
+    <description>Trino - Databricks Connector</description>
+
+    <properties>
+        <air.compiler.fail-warnings>true</air.compiler.fail-warnings>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.databricks</groupId>
+            <artifactId>databricks-jdbc</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.inject</groupId>
+            <artifactId>guice</artifactId>
+            <classifier>classes</classifier>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>configuration</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-base-jdbc</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-plugin-toolkit</artifactId>
+        </dependency>
+
+        <!-- Trino SPI -->
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-annotations</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>slice</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-api</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-api-incubator</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.opentelemetry</groupId>
+            <artifactId>opentelemetry-context</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-spi</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.errorprone</groupId>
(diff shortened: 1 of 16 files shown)


plugin/trino-databricks/pom.xml
```diff
@@ -0,0 +1,266 @@
...
+            </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/TestDatabricksConnectorTest.java</exclude>
+                                <exclude>**/TestDatabricksTypeMapping.java</exclude>
```
ebyhr, line 235: Such tests don't exist in this PR.

We also need a product test to ensure that a class loader issue doesn't exist.

ebyhr:
Could you please clarify the motivation in the PR description, particularly why we want to introduce a _slower_ JDBC-based connector instead of using the Delta Lake connector?

bvolpato:
Thanks for the review! Will take a look and patch it up.

> Could you please clarify the motivation in the PR description, particularly why we want to introduce a _slower_ JDBC-based connector instead of using the Delta Lake connector?

This is required to use their Serverless SQL warehouses, which popped up in a specific use case for our Trino clusters.

Why do you say _slower_? It probably depends on the use case. Queries are likely going to be more expensive (Databricks would charge per DBUs), but based on early tests, performance should be superior.

Praveen2112:
> This is required to use their Serverless SQL warehouses,

But even in case of Serverless SQL warehouse, the metastore i.e unity and storage will be available right ? IIUC Databricks compute layer will be serverless. 

> Queries are likely going to be more expensive (Databricks would charge per DBUs),

Yes so in this case using our Delta lake connector would be cost efficient. 

> based on early tests, performance should be superior.

It depends, for queries which has smaller resultset size, the queries would be faster but as the resultset size grows the queries could take sometime as the JDBC connectors are executed in a single split fashion, it can be extended but there is a data transfer overhead between Databricks and spark.

findepi:
> Why do you say _slower_? It probably depends on the use case. Queries are likely going to be more expensive (Databricks would charge per DBUs), but based on early tests, performance should be superior.

That's interesting use-case.
Can you please describe it in more detail? Is it related to Trino resourcing and DBX resourcing, or something else?

plugin/trino-databricks/pom.xml
bvolpato: Good catch! Added the missing test classes:

- `TestDatabricksConnectorTest` — extends `BaseJdbcConnectorTest` (product test for class loader and connector integration)
- `TestDatabricksTypeMapping` — type mapping round-trip tests for all supported Databricks types
- `TestDatabricksClient` — unit tests for aggregation pushdown (count, sum)

These are modeled after the Snowflake connector equivalents.

bvolpato:
[review-prs-verification]
Head: aa6683eb95bf8c54718fad62b5cbdca4d5f82eb0
Ready to merge: yes

Files reviewed:
- docs/src/main/sphinx/connector.md
- docs/src/main/sphinx/connector/databricks.md
- plugin/trino-databricks/pom.xml
- plugin/trino-databricks/src/main/java/io/trino/plugin/databricks/DatabricksClient.java
- plugin/trino-databricks/src/main/java/io/trino/plugin/databricks/DatabricksClientModule.java
- plugin/trino-databricks/src/main/java/io/trino/plugin/databricks/DatabricksConfig.java
- plugin/trino-databricks/src/main/java/io/trino/plugin/databricks/DatabricksPlugin.java
- plugin/trino-databricks/src/main/java/io/trino/plugin/databricks/ImplementAvgBigint.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/DatabricksQueryRunner.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/TestDatabricksClient.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/TestDatabricksConfig.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/TestDatabricksConnectorTest.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/TestDatabricksPlugin.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/TestDatabricksTypeMapping.java
- plugin/trino-databricks/src/test/java/io/trino/plugin/databricks/TestingDatabricksServer.java
- pom.xml

Tests/checks run and results:
- `./mvnw clean install -B -V -T 1C -DskipTests -Dmaven.source.skip=true -Dair.check.skip-all -pl :trino-spi -am` ✅
- `./mvnw -pl plugin/trino-databricks -am install -B -DskipTests -Dair.check.skip-all` ✅
- `./mvnw -pl plugin/trino-databricks -DskipTests verify -B` ✅
- `./mvnw -pl plugin/trino-databricks test -B -Dair.check.skip-all` ✅

Fixes committed/pushed:
- Updated `plugin/trino-databricks/pom.xml` parent version to `481-SNAPSHOT`.
- Fixed checkstyle blockers in Databricks connector code (`enhanced switch` and unnamed unused lambda parameter).
- Rewrote PR branch history to a single clean commit so per-commit CI check no longer evaluates a broken intermediate commit.
- Pushed head `aa6683eb95bf8c54718fad62b5cbdca4d5f82eb0` to `bvolpato/add-databricks-connector`.

Final CI status:
- `triage` ✅
- `verification/cla-signed` ✅
- No pending checks on current head.

Issue #28727, opened 2026-03-18
Improve check-commit failure message due to webapp compilation error


ebyhr:
Currently, check-commit only leaves the error message below. It's hard to understand what was wrong:
```
Error:  Failed to execute goal com.github.eirslett:frontend-maven-plugin:2.0.0:npm (check (webapp-preview)) on project trino-web-ui: Failed to run task: 'npm run check:clean' failed. org.apache.commons.exec.ExecuteException: Process exited with an error: 1 (Exit value: 1) -> [Help 1]
```

https://github.com/trinodb/trino/actions/runs/23231293617/job/67525721878?pr=28725

PR #28797, opened 2026-03-22
Improve check-commit hint for web UI build failures

kkauy:
Add a hint in `compile-commit` for failures that may come from the web UI preview build, including a local reproduction command.

Related to #28727.

Release notes: not required.

### DIFF
--- a/.github/actions/compile-commit/action.yml
+++ b/.github/actions/compile-commit/action.yml
@@ -47,6 +47,15 @@ runs:
           ${MAVEN_COMPILE_COMMITS} `# defaults, kept in sync with ci.yml` \
           -Dair.check.skip-all=false -Dair.check.skip-basic=true -Dair.check.skip-extended=true -Dair.check.skip-checkstyle=false \
           ${MAVEN_GIB}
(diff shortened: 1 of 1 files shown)


wendigo:
I don't think we need that

kkauy:
Thanks, Do you mean this should be closed or prefer  a different approach here ?

wendigo:
This doesn't address https://github.com/trinodb/trino/issues/28727. Adding information "please do X to understand what's wrong" doesn't improve current situation - it still requires going through the logs

kkauy:
> This doesn't address #28727. Adding information "please do X to understand what's wrong" doesn't improve current situation - it still requires going through the logs

Got it, would you prefer focusing on how the failure is surfaced like clearer error output or direct indication of the failing step ?

ebyhr:
@wendigo is right. The existing error message already indicates that the issue is in the Web UI. What we want to see from the message is **_where_** in the Web UI module the problem occurs.

kkauy:
Thanks, that make sense to me. I will close this PR and work on a separate change focused on making the failing location within the Web UI model more visible.

PR #28799, opened 2026-03-23
Add CREATE OR REPLACE TABLE support for Hive connector

### LINKED ISSUE #21461: CREATE [OR REPLACE] TABLE support for Hive
There exists already `INSERT OVERWRITE` functionality for partition tables in Trino through the use of the session property:

```
SET SESSION hive.insert_existing_partitions_behavior = 'OVERWRITE' 
```

Implement `CREATE OR REPLACE` for (unpartitioned) Hive tables.

## Development information
This functionality builds on top of the scaffolding used for Iceberg & Delta https://github.com/trinodb/trino/issues/13180

See:

- https://github.com/trinodb/trino/pull/17050
- https://github.com/trinodb/trino/pull/13681

laserninja:
Implement CREATE OR REPLACE TABLE for unpartitioned Hive tables, both for direct CREATE TABLE and CREATE TABLE AS SELECT (CTAS).

When replacing an existing table:
- Validates the target is not a view or partitioned table
- Uses the transactional DROP+CREATE->ALTER conversion in SemiTransactionalHiveMetastore for atomic metadata replacement
- Reuses the existing table location for data writes
- Supports both managed and external tables

Also fixes FileHiveMetastore.replaceTable() to handle the case where table metadata is temporarily unavailable during directory rename operations (metadata is stored inside the data directory for file-based metastores).

Resolves: https://github.com/trinodb/trino/issues/21461

### DIFF
--- a/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
+++ b/plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
@@ -269,6 +269,7 @@ import static io.trino.plugin.hive.HiveWritableTableHandle.BucketInfo.createBuck
 import static io.trino.plugin.hive.HiveWriterFactory.computeNonTransactionalBucketedFilename;
 import static io.trino.plugin.hive.HiveWriterFactory.computeTransactionalBucketedFilename;
 import static io.trino.plugin.hive.LocationHandle.WriteMode.DIRECT_TO_TARGET_EXISTING_DIRECTORY;
+import static io.trino.plugin.hive.LocationHandle.WriteMode.STAGE_AND_MOVE_TO_TARGET_DIRECTORY;
 import static io.trino.plugin.hive.PartitionUpdate.UpdateMode.APPEND;
 import static io.trino.plugin.hive.PartitionUpdate.UpdateMode.NEW;
 import static io.trino.plugin.hive.PartitionUpdate.UpdateMode.OVERWRITE;
@@ -1062,15 +1063,30 @@ public class HiveMetadata
     @Override
     public void createTable(ConnectorSession session, ConnectorTableMetadata tableMetadata, SaveMode saveMode)
     {
-        if (saveMode == REPLACE) {
-            throw new TrinoException(NOT_SUPPORTED, "This connector does not support replacing tables");
-        }
         SchemaTableName schemaTableName = tableMetadata.getTable();
         String schemaName = schemaTableName.getSchemaName();
         String tableName = schemaTableName.getTableName();
         List<String> partitionedBy = getPartitionedBy(tableMetadata.getProperties());
         Optional<BucketInfo> bucketInfo = getBucketInfo(tableMetadata.getProperties());
         boolean isTransactional = isTransactional(tableMetadata.getProperties()).orElse(false);
+
+        Optional<Table> existingTable = Optional.empty();
+        if (saveMode == REPLACE) {
+            if (!partitionedBy.isEmpty()) {
+                throw new TrinoException(NOT_SUPPORTED, "CREATE OR REPLACE for partitioned tables is not supported");
+            }
+            existingTable = metastore.getTable(schemaName, tableName);
+            if (existingTable.isPresent()) {
+                Table oldTable = existingTable.get();
+                if (isSomeKindOfAView(oldTable)) {
+                    throw new TrinoException(NOT_SUPPORTED, "Existing table is a view: " + schemaTableName);
+                }
+                if (!oldTable.getPartitionColumns().isEmpty()) {
+                    throw new TrinoException(NOT_SUPPORTED, "CREATE OR REPLACE for partitioned tables is not supported");
+                }
+                metastore.dropTable(session, schemaName, tableName);
+            }
+        }
 
         if (bucketInfo.isPresent() && getAvroSchemaUrl(tableMetadata.getProperties()) != null) {
             throw new TrinoException(NOT_SUPPORTED, "Bucketing columns not supported when Avro schema url is set");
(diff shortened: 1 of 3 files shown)


Praveen2112:
> Reuses the existing table location for data writes

Doesn't this cause a problem when a user queries the table in the mid of a CREATE OR REPLACE TABLE AS SELECT execution, where both the older data and latest data exists ? How do we delete the older files in case of hive ? Unlike Iceberg or DeltaLake - `CREATE OR REPLACE` table implementation is trickier in hive - unless Hive supports it natively. 

Hive creates the splits by iterating over files in a given directory, so if CORTAS (CREATE OR REPLACE TABLE AS SELECT) tries to insert a few files - it will be reflected in a different query during its execution and could cause a correctness issue.

laserninja:
Thanks for raising this. The implementation does use staging directories - it's the same `STAGE_AND_MOVE_TO_TARGET_DIRECTORY` mechanism used by INSERT OVERWRITE. During CORTAS execution, new files are written to a temporary staging path (e.g., `/tmp/trino-<identity>-<uuid>`), not the live table directory. Concurrent readers only see old data during the write.

On commit, `prepareAlterTable()` atomically renames the old data directory to a temp path (`_temp_<tableName>_<queryId>`), then renames the staging directory to the target location. The old data is cleaned up after commit succeeds (or the rename is rolled back on abort).

That said, you're right that this is unsafe when staging directories aren't available — specifically with `DIRECT_TO_TARGET_EXISTING_DIRECTORY` mode (S3, or when `hive.temporary-staging-directory-enabled=false`). In that case, new files would be written directly alongside old files, and the commit would break. I've added an explicit guard that rejects CREATE OR REPLACE when the write mode doesn't support staging, similar to the existing check in `getTableWriteInfo()` for overwrite:

```java
if (locationHandle.getWriteMode() != STAGE_AND_MOVE_TO_TARGET_DIRECTORY) {
    throw new TrinoException(NOT_SUPPORTED, 
        "CREATE OR REPLACE TABLE AS SELECT is not supported when writing directly to target directory");
}
```

This limits CORTAS-with-replace to environments where staging directories work (HDFS, local FS, etc.), which I think is the right tradeoff for now.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/file/FileHiveMetastore.java
```diff
@@ -592,26 +590,42 @@ public synchronized void dropTable(String databaseName, String tableName, boolea
...
-        }
+        // Table metadata may not be found at the expected location when the data directory
+        // (which for FileHiveMetastore also contains metadata) was renamed during ALTER operations
+        // in SemiTransactionalHiveMetastore (e.g., CREATE OR REPLACE TABLE, INSERT OVERWRITE).
+        // The table's existence was already verified in prepareAlterTable before the directory rename.
+        Optional<Table> existingTable = getTable(databaseName, tableName);
 
-        if (isIcebergTable(table) && !Objects.equals(table.getParameters().get("metadata_location"), newTable.getParameters().get("previous_metadata_location"))) {
-            throw new TrinoException(HIVE_CONCURRENT_MODIFICATION_DETECTED, "Cannot update Iceberg table: supplied previous location does not match current location");
+        if (existingTable.isPresent()) {
```
gertjanal: `existingTable.ifPresent(table -> {`

gertjanal: Or even
```
getTable(databaseName, tableName).ifPresent(table -> {

});
```

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/file/FileHiveMetastore.java
```diff
@@ -592,26 +590,42 @@ public synchronized void dropTable(String databaseName, String tableName, boolea
...
 
-        Location tableMetadataDirectory = getTableMetadataDirectory(table);
-        writeSchemaFile(TABLE, tableMetadataDirectory, tableCodec, new TableMetadata(currentVersion, newTable), true);
+        Location tableMetadataDirectory = getTableMetadataDirectory(databaseName, tableName);
 
-        // replace existing permissions
-        deleteTablePrivileges(table);
+        // Ensure the metadata directory exists (it may not if the data directory was renamed)
+        try {
+            if (!fileSystem.directoryExists(tableMetadataDirectory).orElse(false)) {
```
gertjanal, line 614: For non-hierarchical file systems, directoryExists returns an {@code Optional.empty()}. Shouldn't this line be `if (!fileSystem.directoryExists(tableMetadataDirectory).orElseThrow(...)) {`

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
```diff
@@ -1063,16 +1063,32 @@ public void setSchemaAuthorization(ConnectorSession session, String schemaName,
...
+                    throw new TrinoException(NOT_SUPPORTED, "Existing table is a view: " + schemaTableName);
+                }
+                if (!oldTable.getPartitionColumns().isEmpty()) {
+                    throw new TrinoException(NOT_SUPPORTED, "CREATE OR REPLACE for partitioned tables is not supported");
+                }
+                metastore.dropTable(session, schemaName, tableName);
+            }
+        }
+
+        if (saveMode == REPLACE && !partitionedBy.isEmpty()) {
```
gertjanal: Merge this if into the if above, no need to check for `REPLACE` twice?

gertjanal:
Would be a cool feature, thanks!

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveMetadata.java
laserninja: done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/file/FileHiveMetastore.java
laserninja: done

plugin/trino-hive/src/main/java/io/trino/plugin/hive/metastore/file/FileHiveMetastore.java
laserninja: done

findepi:
If i am not mistaken, this is not going to be atomic operation from the perspective of other queries.
This is unlike CREATE OR REPLACE as implemented in e.g. Iceberg and Delta connectors.

If that's the case, I would propose we do not implement this feature, or have a broader conversation about the UX.
cc @trinodb/maintainers

laserninja:
The Hive connector already supports non-atomic table replacement for unpartitioned tables via `INSERT INTO` with `insert_existing_partitions_behavior=OVERWRITE`, which does `dropTable` + `createTable` in `finishInsert` (HiveMetadata.java). CREATE OR REPLACE TABLE provides the same level of atomicity using the same underlying mechanism (`SemiTransactionalHiveMetastore`'s drop+create → ALTER conversion with `STAGE_AND_MOVE_TO_TARGET_DIRECTORY`). We are already using the INSERT INTO + insert_existing_partitions_behavior for some of our clusters. I thought this would be a good addition as it's exposing an existing behavior pattern through a different SQL syntax.

Issue #28803, opened 2026-03-23
OrcSourcePage get block do not handle RuntimeException


YangZhen007:
### Trino version

476

### Please describe the bug

The OrcRecordReader getBlock () method only handles IOException and does not handle RuntimeException

wendigo:
What RuntimeException are you seeing? This isn't enough to justify changes

PR #28804, opened 2026-03-23
exceptionTransform support RuntimeException when read block

YangZhen007:
ref [#28803](https://github.com/trinodb/trino/issues/28803)


wendigo:
What problem does it solve? Can you share stacktrace?

YangZhen007:
> What problem does it solve? Can you share stacktrace?

sorry，it is my misjudgement

PR #28811, opened 2026-03-23
Parquet reader - avoid redundant dictionary reads

ShubhamChaurasia:
## Description

Cannot reopen the previous PR https://github.com/trinodb/trino/pull/27407, opening a new one.

Avoid redundant dictionary reads - earlier the dictionary page is read at the time of rowgroup pruning (`PredicateUtils`) and then again read during the actual file reading(`ParquetReader`). This PR removes this redundant work and moves dictionary based rowgroup pruning logic to `ParquetReader`

Would like to thank @pettyjamesm for his initial suggestions on this.

#### Some numbers with patch

These queries were run on 1TB TPC-DS data(parquet, zstd compressed). 

```
select count(*) from store_sales where ss_store_sk=163;

Without Patch - 26.14 [2.88B rows, 2.08GiB] [110M rows/s, 81.4MiB/s]
With Patch - 21.09 [2.88B rows, 1.09GiB] [137M rows/s, 52.9MiB/s]


select count(*) from store_sales where ss_hdemo_sk < 5226;

Without Patch - 22.23 [2.88B rows, 2.4GiB] [130M rows/s, 111MiB/s]
With Patch - 17.17 [2.88B rows, 1.25GiB] [168M rows/s, 74.6MiB/s]


select count(*) from store_sales where ss_store_sk=163 and ss_hdemo_sk < 5226;

Without Patch - 28.37 [2.88B rows, 5.07GiB] [102M rows/s, 183MiB/s]
With Patch - 18.98 [2.88B rows, 2.94GiB] [152M rows/s, 158MiB/s]


select ss_cdemo_sk, count(*) from store_sales where ss_store_sk=163 and ss_hdemo_sk < 5226 group by ss_cdemo_sk;

Without Patch - 28.87 [2.88B rows, 7.36GiB] [99.8M rows/s, 261MiB/s]
With Patch - 24.23 [2.88B rows, 5.23GiB] [119M rows/s, 221MiB/s]
```

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
@@ -16,20 +16,10 @@ package io.trino.parquet.predicate;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
-import io.airlift.slice.Slice;
-import io.airlift.slice.SliceInput;
-import io.airlift.slice.Slices;
 import io.trino.parquet.BloomFilterStore;
-import io.trino.parquet.DictionaryPage;
 import io.trino.parquet.ParquetCorruptionException;
 import io.trino.parquet.ParquetDataSource;
-import io.trino.parquet.ParquetDataSourceId;
-import io.trino.parquet.ParquetEncoding;
 import io.trino.parquet.ParquetReaderOptions;
-import io.trino.parquet.crypto.AesCipherUtils;
-import io.trino.parquet.crypto.ColumnDecryptionContext;
-import io.trino.parquet.crypto.FileDecryptionContext;
-import io.trino.parquet.crypto.ModuleType;
 import io.trino.parquet.metadata.BlockMetadata;
 import io.trino.parquet.metadata.ColumnChunkMetadata;
 import io.trino.parquet.metadata.ParquetMetadata;
@@ -39,20 +29,12 @@ import io.trino.spi.predicate.TupleDomain;
 import io.trino.spi.type.DecimalType;
 import io.trino.spi.type.Type;
 import org.apache.parquet.column.ColumnDescriptor;
-import org.apache.parquet.column.Encoding;
 import org.apache.parquet.column.statistics.Statistics;
-import org.apache.parquet.format.BlockCipher;
-import org.apache.parquet.format.DictionaryPageHeader;
-import org.apache.parquet.format.PageHeader;
-import org.apache.parquet.format.PageType;
-import org.apache.parquet.internal.column.columnindex.OffsetIndex;
 import org.apache.parquet.internal.filter2.columnindex.ColumnIndexStore;
-import org.apache.parquet.io.ParquetDecodingException;
 import org.apache.parquet.schema.MessageType;
 import org.joda.time.DateTimeZone;
 
 import java.io.IOException;
-import java.io.UncheckedIOException;
 import java.math.BigDecimal;
 import java.util.Arrays;
 import java.util.List;
@@ -61,9 +43,6 @@ import java.util.Optional;
 import java.util.Set;
 
 import static io.trino.parquet.BloomFilterStore.getBloomFilterStore;
-import static io.trino.parquet.ParquetCompressionUtils.decompress;
-import static io.trino.parquet.ParquetReaderUtils.isOnlyDictionaryEncodingPages;
-import static io.trino.parquet.ParquetTypeUtils.getParquetEncoding;
 import static io.trino.parquet.metadata.PrunedBlockMetadata.createPrunedColumnsMetadata;
 import static io.trino.parquet.reader.TrinoColumnIndexStore.getColumnIndexStore;
 import static io.trino.spi.type.BigintType.BIGINT;
@@ -71,20 +50,10 @@ import static io.trino.spi.type.DateType.DATE;
 import static io.trino.spi.type.IntegerType.INTEGER;
 import static io.trino.spi.type.SmallintType.SMALLINT;
 import static io.trino.spi.type.TinyintType.TINYINT;
-import static java.lang.Math.toIntExact;
 import static java.lang.String.format;
-import static java.util.Objects.requireNonNull;
-import static org.apache.parquet.format.Util.readPageHeader;
 
 public final class PredicateUtils
 {
-    // Maximum size of dictionary that we will read for row-group pruning.
-    // Reading larger dictionaries is typically not beneficial. Before checking
-    // the dictionary, the row-group, page indexes and bloomfilters have already been checked
-    // and when the dictionary does not eliminate a row-group, the work done to
-    // decode the dictionary and match it with predicates is wasted.
-    private static final int MAX_DICTIONARY_SIZE = 8096;
-
     private PredicateUtils() {}
 
     public static boolean isStatisticsOverflow(Type type, long min, long max)
@@ -139,7 +108,7 @@ public final class PredicateUtils
         return new TupleDomainParquetPredicate(parquetTupleDomain, columnReferences.build(), timeZone);
     }
 
-    public static boolean predicateMatches(
+    private static PredicateMatchResult predicateMatches(
             TupleDomainParquetPredicate parquetPredicate,
             PrunedBlockMetadata columnsMetadata,
             ParquetDataSource dataSource,
@@ -148,21 +117,20 @@ public final class PredicateUtils
             Optional<ColumnIndexStore> columnIndexStore,
             Optional<BloomFilterStore> bloomFilterStore,
             DateTimeZone timeZone,
-            int domainCompactionThreshold,
-            Optional<FileDecryptionContext> decryptionContext)
+            int domainCompactionThreshold)
             throws IOException
     {
         if (columnsMetadata.getRowCount() == 0) {
-            return false;
+            return PredicateMatchResult.FALSE_WITH_DICT_MATCHING_NOT_NEEDED;
         }
         Map<ColumnDescriptor, Statistics<?>> columnStatistics = getStatistics(columnsMetadata, descriptorsByPath);
         Map<ColumnDescriptor, Long> columnValueCounts = getColumnValueCounts(columnsMetadata, descriptorsByPath);
         Optional<List<ColumnDescriptor>> candidateColumns = parquetPredicate.getIndexLookupCandidates(columnValueCounts, columnStatistics, dataSource.getId());
         if (candidateColumns.isEmpty()) {
-            return false;
+            return PredicateMatchResult.FALSE_WITH_DICT_MATCHING_NOT_NEEDED;
         }
         if (candidateColumns.get().isEmpty()) {
-            return true;
+            return PredicateMatchResult.TRUE_WITH_DICT_MATCHING_NOT_NEEDED;
         }
         // Perform column index, bloom filter checks and dictionary lookups only for the subset of columns where it can be useful.
         // This prevents unnecessary filesystem reads and decoding work when the predicate on a column comes from
@@ -171,21 +139,14 @@ public final class PredicateUtils
 
         // Page stats is finer grained but relatively more expensive, so we do the filtering after above block filtering.
         if (columnIndexStore.isPresent() && !indexPredicate.matches(columnValueCounts, columnIndexStore.get(), dataSource.getId())) {
-            return false;
+            return PredicateMatchResult.FALSE_WITH_DICT_MATCHING_NOT_NEEDED;
         }
 
         if (bloomFilterStore.isPresent() && !indexPredicate.matches(bloomFilterStore.get(), domainCompactionThreshold)) {
-            return false;
+            return PredicateMatchResult.FALSE_WITH_DICT_MATCHING_NOT_NEEDED;
         }
 
-        return dictionaryPredicatesMatch(
-                indexPredicate,
-                columnsMetadata,
-                dataSource,
-                descriptorsByPath,
-                ImmutableSet.copyOf(candidateColumns.get()),
-                columnIndexStore,
-                decryptionContext);
+        return new PredicateMatchResult(true, Optional.of(indexPredicate), Optional.of(ImmutableSet.copyOf(candidateColumns.get())));
     }
 
     public static List<RowGroupInfo> getFilteredRowGroups(
@@ -209,7 +170,7 @@ public final class PredicateUtils
                 Optional<ColumnIndexStore> columnIndex = getColumnIndexStore(dataSource, block, descriptorsByPath, parquetTupleDomain, options, parquetMetadata.getDecryptionContext());
                 Optional<BloomFilterStore> bloomFilterStore = getBloomFilterStore(dataSource, block, parquetTupleDomain, options, parquetMetadata.getDecryptionContext());
                 PrunedBlockMetadata columnsMetadata = createPrunedColumnsMetadata(block, dataSource.getId(), descriptorsByPath);
-                if (predicateMatches(
+                PredicateMatchResult predicateMatchResult = predicateMatches(
                         parquetPredicate,
                         columnsMetadata,
                         dataSource,
@@ -218,9 +179,9 @@ public final class PredicateUtils
                         columnIndex,
                         bloomFilterStore,
                         timeZone,
-                        domainCompactionThreshold,
-                        parquetMetadata.getDecryptionContext())) {
-                    rowGroupInfoBuilder.add(new RowGroupInfo(columnsMetadata, block.fileRowCountOffset(), columnIndex));
+                        domainCompactionThreshold);
+                if (predicateMatchResult.matchesWithoutDictionary()) {
+                    rowGroupInfoBuilder.add(new RowGroupInfo(columnsMetadata, block.fileRowCountOffset(), columnIndex, predicateMatchResult.indexPredicate(), predicateMatchResult.candidateColumnsForDictionaryMatching()));
                     break;
                 }
             }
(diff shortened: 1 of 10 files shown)


ShubhamChaurasia:
@raunaqmorarka @pettyjamesm Could you please check? (I could not re-open the original PR https://github.com/trinodb/trino/pull/27407)
This addresses some of previous comments and also breaks it into multiple commits to make reviewing easy.

wendigo:
I've scheduled some benchmarks

pettyjamesm:
@coderabbitai full review

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -467,39 +467,42 @@ private int nextBatch()
...
-            long rowCount = currentGroupRowRanges.getRowCount();
-            columnIndexRowsFiltered += currentGroupRowCount - rowCount;
-            if (rowCount == 0) {
-                // Filters on multiple columns with page indexes may yield non-overlapping row ranges and eliminate the entire row group.
-                // Advance to next row group to ensure that we don't return a null Page and close the page source before all row groups are processed
-                return advanceToNextRowGroup();
+        while (currentRowGroup < rowGroups.size()) {
+            currentRowGroupMemoryContext.close();
+            currentRowGroupMemoryContext = memoryContext.newAggregatedMemoryContext();
+            freeCurrentRowGroupBuffers();
```
pettyjamesm, line 475: I wasn't sure about the placement of this sequence in the loop on the first commit, but this is indeed the right place to put this logic after the second commit.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -500,11 +502,33 @@ private boolean advanceToNextRowGroup()
             }
             nextRowInGroup = 0L;
             initializeColumnReaders();
+
+            // check dictionary predicate matches, or skip row group
+            if (!dictionaryPredicateMatch(rowGroupInfo)) {
```
pettyjamesm, line 507: Not blocking in this PR, but it seems like there's a future opportunity to leverage `FilteredRowRanges` above to make a more narrow check on the dictionary predicate (i.e.: only dictionary entries referenced within the active ranges should be considered, and alternatively- checking individual dictionary entries could eliminate more sub-ranges in the subsequent scan)

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -500,11 +502,33 @@ private boolean advanceToNextRowGroup()
...
         }
         return false;
     }
 
+    private boolean dictionaryPredicateMatch(RowGroupInfo rowGroupInfo)
+            throws ParquetCorruptionException
+    {
+        for (PrimitiveField field : primitiveFields) {
+            // check presence of indexPredicate and don't eagerly initializePageReader if it's not present
+            if (rowGroupInfo.indexPredicate().isPresent()) {
```
pettyjamesm: AI reviewer mentioned these two items as a a nitpick, but I think it's actually fairly significant for performance:
1. The `rowGroupInfo.indexPredicate().isPresent()` check can be lifted out of the loop
2. Checking `RowGroupInfo#candidateColumnsForDictionaryMatching` to only initialize the page reader for columns where dictionary predicate checks may occur might make sense here since it seems like we're preferring to initialize them lazily when not needed?

lib/trino-parquet/src/main/java/io/trino/parquet/reader/RowGroupInfo.java
```diff
@@ -14,8 +14,16 @@
...
 
 import java.util.Optional;
+import java.util.Set;
 
-public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata, long fileRowOffset, Optional<ColumnIndexStore> columnIndexStore) {}
+public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata,
+                           long fileRowOffset,
+                           Optional<ColumnIndexStore> columnIndexStore,
+                           Optional<TupleDomainParquetPredicate> indexPredicate,
+                           Optional<Set<ColumnDescriptor>> candidateColumnsForDictionaryMatching)
```
pettyjamesm: Why `Optional.empty()` instead of just `Set.of()` ?

lib/trino-parquet/src/main/java/io/trino/parquet/reader/RowGroupInfo.java
```diff
@@ -14,8 +14,16 @@
...
 import org.apache.parquet.internal.filter2.columnindex.ColumnIndexStore;
 
 import java.util.Optional;
+import java.util.Set;
 
-public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata, long fileRowOffset, Optional<ColumnIndexStore> columnIndexStore) {}
+public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata,
+                           long fileRowOffset,
+                           Optional<ColumnIndexStore> columnIndexStore,
+                           Optional<TupleDomainParquetPredicate> indexPredicate,
```
pettyjamesm: Is this the same predicate used for dictionary predicate matching? I would suggest a comment and/or renaming this variable to make it more obvious that this predicate is doing double duty.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
ShubhamChaurasia: I see, will check, thanks for the suggestion.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
ShubhamChaurasia: agreed, and addressed both.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/RowGroupInfo.java
ShubhamChaurasia: yeah, renamed it.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/RowGroupInfo.java
ShubhamChaurasia: Changed

wendigo:
@ShubhamChaurasia please squash commits together

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ColumnReader.java
```diff
@@ -21,6 +23,9 @@ public interface ColumnReader
 
     void setPageReader(PageReader pageReader, Optional<FilteredRowRanges> rowRanges);
 
+    boolean dictionaryPredicateMatch(RowGroupInfo rowGroupInfo)
```
raunaqmorarka, line 26: Adding `dictionaryPredicateMatch(RowGroupInfo)` to the `ColumnReader` interface pushes row-group-level concerns into a column-level abstraction. Before this PR, all methods on this interface are cleanly column-scoped (`hasPageReader`, `setPageReader`, `prepareNextRead`, `readPrimitive`). This new method takes a `RowGroupInfo` and then inside `AbstractColumnReader` reaches back into it to extract the deferred predicate, candidate columns set, and pruned block metadata — all of which the caller (`ParquetReader`) already has.

Consider replacing this with a minimal accessor on the interface:
```java
Optional<DictionaryPage> getDictionaryPage();
```
Then centralize the filtering logic in `ParquetReader.dictionaryPredicateMatch()` where the row-group context already lives. This keeps the column reader focused on column-level reading concerns.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -467,38 +469,64 @@ private int nextBatch()
...
+    }
+
+    private boolean dictionaryPredicateMatch(RowGroupInfo rowGroupInfo)
+            throws ParquetCorruptionException
+    {
+        Set<ColumnDescriptor> columnDescriptors = rowGroupInfo.candidateColumnsForDictionaryMatching();
+        if (rowGroupInfo.deferredPredicate().isPresent() && !columnDescriptors.isEmpty()) {
+            for (PrimitiveField field : primitiveFields) {
+                if (columnDescriptors.contains(field.getDescriptor())) {
+                    initializePageReader(field);
```
raunaqmorarka, line 523: `initializePageReader(field)` is called here *before* `AbstractColumnReader.dictionaryPredicateMatch` checks `isOnlyDictionaryEncodingPages(columnMetaData)`. This means for candidate columns that are NOT fully dictionary-encoded (have mixed dict + plain pages), we pay the cost of creating a `PageReader`, reading the dictionary page, and setting up decoders — only for `AbstractColumnReader.dictionaryPredicateMatch` to return `true` immediately because the column isn't only-dictionary-encoded.

The `isOnlyDictionaryEncodingPages` check should happen *before* `initializePageReader` to avoid this wasted work. This naturally falls out of centralizing the logic in `ParquetReader` (per the `ColumnReader` interface comment):

```java
private boolean dictionaryPredicateMatch(RowGroupInfo rowGroupInfo)
        throws ParquetCorruptionException
{
    Optional<TupleDomainParquetPredicate> deferredPredicate = rowGroupInfo.deferredPredicate();
    Set<ColumnDescriptor> candidateColumns = rowGroupInfo.candidateColumnsForDictionaryMatching();
    if (deferredPredicate.isEmpty() || candidateColumns.isEmpty()) {
        return true;
    }
    for (PrimitiveField field : primitiveFields) {
        ColumnDescriptor descriptor = field.getDescriptor();
        if (!candidateColumns.contains(descriptor)) {
            continue;
        }
        ColumnChunkMetadata columnMetaData = currentBlockMetadata.getColumnChunkMetaData(descriptor);
        if (!isOnlyDictionaryEncodingPages(columnMetaData)) {
            continue;
        }
        initializePageReader(field);
        Statistics<?> columnStatistics = columnMetaData.getStatistics();
        boolean nullAllowed = columnStatistics == null || columnStatistics.getNumNulls() != 0;
        if (!deferredPredicate.get().matches(new DictionaryDescriptor(
                descriptor,
                nullAllowed,
                columnReaders.get(field.getId()).getDictionaryPage()))) {
            return false;
        }
    }
    return true;
}
```

lib/trino-parquet/src/main/java/io/trino/parquet/predicate/PredicateUtils.java
```diff
@@ -253,147 +214,9 @@ private static Map<ColumnDescriptor, Long> getColumnValueCounts(PrunedBlockMetad
...
-        if (dictionaryHeader.getNum_values() > MAX_DICTIONARY_SIZE) {
-            return Optional.empty();
-        }
-        return Optional.of(new PageHeaderWithData(
-                pageHeader,
-                inputStream.readSlice(pageHeader.getCompressed_page_size())));
-    }
-
-    private static DictionaryPage decodeDictionaryPage(ParquetDataSourceId dataSourceId, PageHeaderWithData pageHeaderWithData, ColumnChunkMetadata chunkMetaData, Optional<FileDecryptionContext> decryptionContext)
+    private record PredicateMatchResult(boolean matchesWithoutDictionary, Optional<TupleDomainParquetPredicate> deferredPredicate, Set<ColumnDescriptor> candidateColumnsForDictionaryMatching)
```
raunaqmorarka, line 217: `matchesWithoutDictionary()` reads like a confirmed positive ("it matches, and we didn't need a dictionary"), but when `deferredPredicate` is present, it actually means "tentatively retained, still needs dictionary verification." Consider `passedStatisticsFilters` to make the semantics clearer.

The constant names are also quite verbose. Consider `ELIMINATED` and `RETAINED_NO_DICTIONARY_CHECK_NEEDED` (or `RETAINED` if both are renamed).

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -467,38 +469,64 @@ private int nextBatch()
...
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private boolean dictionaryPredicateMatch(RowGroupInfo rowGroupInfo)
+            throws ParquetCorruptionException
+    {
+        Set<ColumnDescriptor> columnDescriptors = rowGroupInfo.candidateColumnsForDictionaryMatching();
```
raunaqmorarka, line 519: Nit: `columnDescriptors` is too generic — could be `candidateColumns` or `dictionaryMatchCandidates` to convey purpose.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/RowGroupInfo.java
```diff
@@ -14,8 +14,17 @@
...
+import org.apache.parquet.column.ColumnDescriptor;
 import org.apache.parquet.internal.filter2.columnindex.ColumnIndexStore;
 
 import java.util.Optional;
+import java.util.Set;
 
-public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata, long fileRowOffset, Optional<ColumnIndexStore> columnIndexStore) {}
+public record RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata,
+                           long fileRowOffset,
+                           Optional<ColumnIndexStore> columnIndexStore,
```
raunaqmorarka, line 26: Every caller that doesn't use dictionary filtering (ParquetWriter, RedshiftPageSourceProvider) must now pass `Optional.empty(), ImmutableSet.of()`. Consider adding a convenience constructor:

```java
public RowGroupInfo(PrunedBlockMetadata prunedBlockMetadata, long fileRowOffset,
                    Optional<ColumnIndexStore> columnIndexStore) {
    this(prunedBlockMetadata, fileRowOffset, columnIndexStore, Optional.empty(), ImmutableSet.of());
}
```

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -109,6 +109,7 @@ public class ParquetReader
     private static final int BATCH_SIZE_GROWTH_FACTOR = 2;
     public static final String PARQUET_CODEC_METRIC_PREFIX = "ParquetReaderCompressionFormat_";
     public static final String COLUMN_INDEX_ROWS_FILTERED = "ParquetColumnIndexRowsFiltered";
+    public static final String PARQUET_READER_DICTIONARY_FILTERED_ROWGROUPS = "ParquetReaderDictionaryFilteredRowGroups";
```
raunaqmorarka, line 112: Nit: Inconsistent naming convention with the existing metric. `COLUMN_INDEX_ROWS_FILTERED` uses `ParquetColumnIndex...` prefix with `RowsFiltered` suffix, while this uses `ParquetReader...` prefix with `FilteredRowGroups` suffix (different word order). Consider aligning, e.g. `ParquetDictionaryFilteredRowGroups`.

lib/trino-parquet/src/main/java/io/trino/parquet/reader/AbstractColumnReader.java
```diff
@@ -87,6 +100,29 @@ public void setPageReader(PageReader pageReader, Optional<FilteredRowRanges> row
...
+            throws ParquetCorruptionException
+    {
+        checkState(hasPageReader(), "Don't have a pageReader yet, invoke setPageReader() first");
+        Optional<TupleDomainParquetPredicate> deferredPredicate = rowGroupInfo.deferredPredicate();
+        Set<ColumnDescriptor> candidateColumnsForDictionaryMatching = rowGroupInfo.candidateColumnsForDictionaryMatching();
+        if (deferredPredicate.isPresent()) {
+            ColumnDescriptor descriptor = field.getDescriptor();
+            PrunedBlockMetadata prunedBlockMetadata = rowGroupInfo.prunedBlockMetadata();
+            ColumnChunkMetadata columnMetaData = prunedBlockMetadata.getColumnChunkMetaData(descriptor);
+            if (candidateColumnsForDictionaryMatching.contains(descriptor) && isOnlyDictionaryEncodingPages(columnMetaData)) {
```
raunaqmorarka, line 114: `candidateColumnsForDictionaryMatching.contains(descriptor)` is already checked by the caller (`ParquetReader.dictionaryPredicateMatch`) before invoking this method. The `deferredPredicate.isPresent()` guard is also redundant with the caller. These double-checks make the code harder to follow since it's unclear which layer owns the filtering logic.

This is another reason to centralize the logic in `ParquetReader` (per the `ColumnReader` interface comment) — it eliminates the redundancy and makes the responsibility boundary clear.

raunaqmorarka (commented):
Thanks for the PR! The approach of deferring dictionary-based pruning to the reader to avoid redundant dictionary reads is sound and the benchmarks are compelling. I have some suggestions around code structure, naming, and abstraction boundaries.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/TestParquetReaderDictionaryFiltering.java
```diff
@@ -0,0 +1,248 @@
...
+        assertThat(parquetMetadata.getBlocks().size()).isEqualTo(expectedRowGroupCount);
+        for (BlockMetadata blockMetaData : parquetMetadata.getBlocks()) {
+            assertThat(blockMetaData.columns()).hasSize(columnNames.size());
+            for (ColumnChunkMetadata chunkMetadata : blockMetaData.columns()) {
+                assertThat(chunkMetadata.getDictionaryPageOffset()).isGreaterThan(0);
+                int dictionaryPageSize = toIntExact(chunkMetadata.getFirstDataPageOffset() - chunkMetadata.getDictionaryPageOffset());
+                assertThat(dictionaryPageSize).isGreaterThan(0);
+                assertThat(chunkMetadata.getEncodingStats().hasDictionaryPages()).isTrue();
+                assertThat(chunkMetadata.getEncodingStats().hasDictionaryEncodedPages()).isTrue();
+                assertThat(chunkMetadata.getEncodingStats().hasNonDictionaryEncodedPages()).isFalse();
```
raunaqmorarka, line 200: All tests assert `hasNonDictionaryEncodedPages() == false`, meaning every column in the test suite is fully dictionary-encoded. This leaves a gap: **mixed-encoding columns** (dictionary fallback to plain) are never exercised through the new reader-side dictionary filtering path.

When a candidate column has mixed encoding, `isOnlyDictionaryEncodingPages` returns false and the dictionary check should be skipped — but with the current code structure, `initializePageReader` is called *before* that check (see my comment on `ParquetReader.java:523`). A test with enough distinct values to trigger dictionary fallback would verify:
- The row group is **not** incorrectly pruned by dictionary filtering
- Data is read correctly despite the early page reader initialization

Additionally, a test combining a dictionary-encoded column with a mixed-encoding column under a multi-column predicate would cover the heterogeneous case: column A (dictionary-only) doesn't match → row group should still be pruned, even though column B (mixed encoding) can't be dictionary-checked.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/TestParquetReaderDictionaryFiltering.java
```diff
@@ -0,0 +1,248 @@
...
+            throws IOException
+    {
+        try (ParquetReader reader = createParquetReader(dataSource, parquetMetadata, ParquetReaderOptions.defaultOptions(), newSimpleAggregatedMemoryContext(), types, columnNames, predicate)) {
+            SourcePage page = reader.nextPage();
+            int rowsRead = 0;
+            while (page != null) {
+                rowsRead += page.getPositionCount();
+                page = reader.nextPage();
+            }
+            assertThat(rowsRead).isEqualTo(expectedRowsRead);
```
raunaqmorarka, line 223: `assertDictionaryFiltered` only checks row counts, never the actual values returned. Since this PR makes the same `dictionaryPage` field in `AbstractColumnReader` serve double duty — predicate matching in `dictionaryPredicateMatch` *and* data decoding via `dictionaryDecoder` — it would be worth adding at least one test that verifies the returned data values are correct after a row group passes dictionary filtering. This guards against a subtle class of bugs where the shared dictionary page is consumed or corrupted during predicate evaluation.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/TestParquetReaderDictionaryFiltering.java
```diff
@@ -0,0 +1,248 @@
...
+    @Test
+    public void testFilteringWithSingleRowGroup()
+            throws Exception
+    {
+        List<String> columnNames = ImmutableList.of("int_col", "bigint_col", "varchar_col");
+        List<Type> types = ImmutableList.of(INTEGER, BIGINT, VARCHAR);
+        int totalPositions = 1000;
+        int pageCount = 100;
+        // generate multiples of 5 as distinct values in blocks
+        int[] distinctValues = IntStream.range(1, 21).map(i -> i * 5).toArray();
```
raunaqmorarka, line 73: All tests use 20 distinct values. The old code had `MAX_DICTIONARY_SIZE = 8096` which skipped dictionary-based pruning for dictionaries larger than that. This PR removes that limit — dictionaries of any size are now checked. This is the one behavioral change in the PR, and it has no test coverage.

Consider adding a test with > 8096 distinct values (all dictionary-encoded) and a non-matching predicate, verifying the row group *is* pruned. This would have failed under the old code and passes under the new code, directly documenting the behavioral improvement.

raunaqmorarka (commented):
A few test coverage gaps worth addressing:

lib/trino-parquet/src/main/java/io/trino/parquet/reader/ParquetReader.java
```diff
@@ -467,38 +469,64 @@ private int nextBatch()
...
+            if (!dictionaryPredicateMatch(rowGroupInfo)) {
+                dictionaryFilteredRowGroups++;
+                continue;
+            }
+            return true;
+        }
+        return false;
+    }
+
+    private boolean dictionaryPredicateMatch(RowGroupInfo rowGroupInfo)
```
raunaqmorarka, line 516: `initializePageReader` reads from the `ChunkedInputStream`, which materializes the first `ChunkReader` — up to 8MB (`maxBufferSize`). The old code performed a targeted small I/O to read just the dictionary page (~KB) for each candidate column. For row groups that are ultimately pruned by dictionary filtering, this turns a few-KB read into an up-to-8MB read per candidate column.

This compounds with multiple candidate columns. Given `WHERE col_a = 'x' AND col_b = 'y' AND col_c = 'z'`, this method calls `initializePageReader` for each candidate sequentially. If `col_c` finally prunes the row group, we've already materialized up to 8MB per preceding candidate column that gets discarded.

**Suggested fix:** Split the column chunk `DiskRange` at the dictionary/data page boundary for candidate columns in the `ParquetReader` constructor. `FilteredOffsetIndex.calculateOffsetRanges` already does this when column index filtering is active. Extend this to the no-column-index case:

```java
// In the ParquetReader constructor, when filteredOffsetIndex == null
if (isCandidateForDictionaryMatch && firstDataPageOffset > startingPosition) {
    ranges.put(key, new DiskRange(startingPosition, firstDataPageOffset - startingPosition));
    ranges.put(key, new DiskRange(firstDataPageOffset, totalLength - (firstDataPageOffset - startingPosition)));
} else {
    ranges.put(key, new DiskRange(startingPosition, totalLength));
}
```

For large columns (where over-reading matters), `mergeAdjacentDiskRanges` won't re-merge these because their combined length exceeds `maxBufferSize`. The dictionary stays as its own small `ChunkReader`, so `initializePageReader` only reads the dictionary chunk. If the row group is pruned, the data `ChunkReader`s are freed unread via `ChunkedInputStream.close()`. For small columns, the ranges merge back — but total size is small, so there's no meaningful regression.

PPL-TBSK:
@ShubhamChaurasia may I ask what's the status of the PR?

PR #28830, opened 2026-03-24
Deprecate the source/sink provider methods without table credentials and add javadocs for the active methods

findinpath:
## Description

The first commits contains deprecation of the methods that don't have table credentials after the contribution from https://github.com/trinodb/trino/pull/28755

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
+++ b/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
@@ -30,7 +30,22 @@ public interface ConnectorPageSinkProvider
         throw new TrinoException(NOT_SUPPORTED, "This connector does not support creating new tables");
     }
 
-    default ConnectorPageSink createPageSink(ConnectorTransactionHandle transactionHandle, ConnectorSession session, ConnectorOutputTableHandle outputTableHandle, Optional<ConnectorTableCredentials> tableCredentials, ConnectorPageSinkId pageSinkId)
+    /**
+     * Creates a {@link ConnectorPageSink} for writing data to a newly created table.
+     *
+     * @param transactionHandle the transaction handle for this operation
+     * @param session the session in which the write is being performed
+     * @param outputTableHandle the handle identifying the new table being created
+     * @param tableCredentials credentials for accessing the table data
+     * @param pageSinkId unique identifier for this page sink instance
+     * @return a page sink for writing data to the new table
+     */
+    default ConnectorPageSink createPageSink(
+            ConnectorTransactionHandle transactionHandle,
+            ConnectorSession session,
+            ConnectorOutputTableHandle outputTableHandle,
+            Optional<ConnectorTableCredentials> tableCredentials,
+            ConnectorPageSinkId pageSinkId)
     {
         return createPageSink(transactionHandle, session, outputTableHandle, pageSinkId);
     }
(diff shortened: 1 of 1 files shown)


core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
```diff
@@ -21,41 +21,117 @@
 
 public interface ConnectorPageSinkProvider
 {
+    /**
+     * @deprecated Use {@link #createPageSink(ConnectorTransactionHandle, ConnectorSession, ConnectorOutputTableHandle, Optional, ConnectorPageSinkId)} instead.
```
wendigo: s/Use/Implement/g

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
```diff
@@ -21,41 +21,117 @@
...
     }
 
-    default ConnectorPageSink createPageSink(ConnectorTransactionHandle transactionHandle, ConnectorSession session, ConnectorOutputTableHandle outputTableHandle, Optional<ConnectorTableCredentials> tableCredentials, ConnectorPageSinkId pageSinkId)
+    /**
+     * Creates a {@link ConnectorPageSink} for writing data to a newly created table.
+     *
+     * @param transactionHandle the transaction handle for this operation
+     * @param session the session in which the write is being performed
+     * @param outputTableHandle the handle identifying the new table being created
+     * @param tableCredentials optional credentials for accessing the table data
```
wendigo: drop optional, it's known from the type

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
```diff
@@ -21,41 +21,117 @@
...
     }
 
-    default ConnectorMergeSink createMergeSink(ConnectorTransactionHandle transactionHandle, ConnectorSession session, ConnectorMergeTableHandle mergeHandle, Optional<ConnectorTableCredentials> tableCredentials, ConnectorPageSinkId pageSinkId)
+    /**
+     * Creates a {@link ConnectorMergeSink} for executing SQL MERGE operations.
+     *
+     * @param transactionHandle the transaction handle for this operation
+     * @param session the session in which the merge is being performed
+     * @param mergeHandle the handle identifying the merge operation
+     * @param tableCredentials optional credentials for accessing the table data
```
wendigo: drop "optional"

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
```diff
@@ -21,41 +21,117 @@
...
+     * @param pageSinkId unique identifier for this page sink instance
+     * @return a merge sink for executing the MERGE operation
+     */
+    default ConnectorMergeSink createMergeSink(
+            ConnectorTransactionHandle transactionHandle,
+            ConnectorSession session,
+            ConnectorMergeTableHandle mergeHandle,
+            Optional<ConnectorTableCredentials> tableCredentials,
+            ConnectorPageSinkId pageSinkId)
     {
```
wendigo, line 127: verify if the tableCredentials.isEmpty if you delegate to a method not taking table credentials

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
```diff
@@ -21,41 +21,117 @@
...
+     * @param pageSinkId unique identifier for this page sink instance
+     * @return a page sink for writing data during table execute
+     */
+    default ConnectorPageSink createPageSink(
+            ConnectorTransactionHandle transactionHandle,
+            ConnectorSession session,
+            ConnectorTableExecuteHandle tableExecuteHandle,
+            Optional<ConnectorTableCredentials> tableCredentials,
+            ConnectorPageSinkId pageSinkId)
     {
```
wendigo, line 98: verify tableCredentials isEmpty

findinpath:
@wendigo  AC. Pls take another look at the PR.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorPageSinkProvider.java
```diff
@@ -21,41 +21,133 @@
...
+     */
+    default ConnectorPageSink createPageSink(
+            ConnectorTransactionHandle transactionHandle,
+            ConnectorSession session,
+            ConnectorOutputTableHandle outputTableHandle,
+            Optional<ConnectorTableCredentials> tableCredentials,
+            ConnectorPageSinkId pageSinkId)
     {
+        if (tableCredentials.isPresent()) {
+            throw new TrinoException(NOT_SUPPORTED, "This connector does not support table credentials");
```
findepi: The connector clearly does support table credentials, otherwise we would not be on this line. The message is wrong. 

```
Table credentials present but createPageSink taking table credentials is not implemented.
```

findepi: Also, it's OK to ignore this parameter. It's optional parameter that does not matter for correctness.
If a connector creates table credentials it does not actually need, it's not incorrect, so we can continue without this throw.

That's my preferred approach.

findinpath: Removed table credentials check.

To be completely honest - given that @wendigo  was suggesting the opposite, I think this could be reconciled in a small follow-up PR.

PR #28832, opened 2026-03-24
Improve compatibility with JDK 26

wendigo:
JDK 26 introduced an update to JDBC spec that added some new methods that we need to implement

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ForwardingConnection.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ForwardingConnection.java
@@ -14,6 +14,7 @@
 
 package io.trino.plugin.jdbc;
 
+import java.lang.reflect.InvocationTargetException;
 import java.sql.Array;
 import java.sql.Blob;
 import java.sql.CallableStatement;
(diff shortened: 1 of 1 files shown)


losipiuk:
What is the benefit of having that before we go to Java 26?

wendigo:
@losipiuk Trino will run on both jdk 25 and 26 and we won't accumulate changes required to be compatible

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ForwardingConnection.java
```diff
@@ -473,4 +474,52 @@ public boolean isWrapperFor(Class<?> iface)
     {
         return delegate().isWrapperFor(iface);
     }
+
+    @SuppressWarnings("MissingOverride") // Part of the JDK 26, JDBC 4.5 spec
+    public String enquoteLiteral(String val)
+            throws SQLException
+    {
+        try {
+            return (String) delegate().getClass().getMethod("enquoteLiteral", String.class).invoke(delegate(), val);
```
findepi, line 483: If this method was really to be used for anything, then having reflection on every call basis would be prohibitively expensive.
**if** we were to add this method, it should throw UnsupportedException rather than be super slow.

~I understand we need this method if we want Trino to compile with source version with Java 26.~ 
~I am not sure what's the benefit of doing that though.~

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/ForwardingConnection.java
```diff
@@ -473,4 +474,52 @@ public boolean isWrapperFor(Class<?> iface)
     {
         return delegate().isWrapperFor(iface);
     }
+
+    @SuppressWarnings("MissingOverride") // Part of the JDK 26, JDBC 4.5 spec
```
findepi, line 478: `// Part of the JDK 26, JDBC 4.5 spec TODO remove suppression when Trino requires Java 26`

findepi:
This is not needed to **compile** with JDK 26, right?
This is needed to run tests with JDK 26, right?

losipiuk:
> @losipiuk Trino will run on both jdk 25 and 26 and we won't accumulate changes required to be compatible

Yeah true. But we can just throw in all those methods - not delegate. 
We cannot do delegation unless we know that we are on 26. And then we can just do normal delegation without any reflection.

wendigo:
@findepi correct

findepi:
> Yeah true. But we can just throw in all those methods - not delegate.

https://github.com/trinodb/trino/pull/28832#discussion_r2986661358

wendigo:
@findepi how is throwing superior?

findepi:
A method explicitly unimplemented is better than a method implemented poorly.

wendigo:
@findepi how is this method implemented poorly? What's "poor" about this implementation?

wendigo:
I'm fine with either but want to understand 🤷🏻

wendigo:
@findepi gentle ping

losipiuk:
> I'm fine with either but want to understand 🤷🏻

"If this method was really to be used for anything, then having reflection on every call basis would be prohibitively expensive."

Issue #23238, opened 2024-09-02
Iceberg REST Catalog: Support for vended credentials for Azure


c-thiel:
I am currently extending the integration tests for our [iceberg rest catalog implementation](https://github.com/hansetag/iceberg-catalog).

S3 integration with trino works nicely, but I can't get `vended-credentails` up and running.

My configuration looks as follows:
```sh
    CREATE CATALOG test_azure USING iceberg
    WITH (
        "iceberg.catalog.type" = 'rest',
        "iceberg.rest-catalog.uri" = 'https://api.tabular.io/ws/',
        "iceberg.rest-catalog.warehouse" = 'cth-azure',
        "iceberg.rest-catalog.security" = 'OAUTH2',
        "iceberg.rest-catalog.oauth2.token" = '{token}',
        "iceberg.rest-catalog.vended-credentials-enabled" = 'true',
        "fs.native-azure.enabled" = 'true'
    )
```

I can use all endpoints as usual, but data operations fail with the following error:
```python
TrinoExternalError(type=EXTERNAL, name=ICEBERG_FILESYSTEM_ERROR, message="Failed checking new table's location: abfss://<filesystem>@<storage-account-name>.dfs.core.windows.net/0191b3ad-6c11-7fc2-bb2a-12a7648dc115/my_table-a90cff2771dd44d3aa55235cf4f77a47", query_id=20240902_165932_00260_3zngw)
```

I believe the returned config attribute for the table:
```adls.sas-token.<storage-account-name>.dfs.core.windows.net: "skoid=...."``` 
is not being used.
Instead it is trying to load credentials from well-known locations:

```log
2024-09-02T16:59:32.697Z        INFO    Query-20240902_165932_00258_3zngw-817   com.azure.identity.ChainedTokenCredential       Azure Identity => Attempted credential EnvironmentCredential is unavailable.
2024-09-02T16:59:32.697Z        INFO    Query-20240902_165932_00258_3zngw-817   com.azure.identity.ChainedTokenCredential       Azure Identity => Attempted credential WorkloadIdentityCredential is unavailable.
[...]

PR #28855, opened 2026-03-25
Support Azure vended credentials in Iceberg REST catalog

findinpath:
## Description

This change adds support for the following Azure prefixed properties from the vended credentials:

- "adls.sas-token." storage account prefix short-lived OAuth2 access token

Fixes https://github.com/trinodb/trino/issues/23238

Added as well scaffolding for configuring connection/socket timeout for the REST Catalog client because I thought it could be useful to deal with flakiness in the tests.

- `iceberg.rest-catalog.connection-timeout`
  - Maximum time [Duration](prop-type-duration) allowed for socket connection
    requests to complete before timing out.
- `iceberg.rest-catalog.socket-timeout`
  - Maximum time [Duration](prop-type-duration) allowed socket read/write operations
    before timing out.

It apparently wasn't necessary, because the slowness in deleting files on Azure via `iceberg-rest-fixture` was solved through the following settings for the `IcebergAzureRestCatalogBackendContainer`

```
// Increase the size of the worker pool to achieve a higher parallelism for storage operations
.put("ICEBERG_WORKER_NUM_THREADS", "16")
```

https://github.com/apache/iceberg/issues/15760 was found while developing this feature

Logic is pretty similar to what has been done in https://github.com/trinodb/trino/pull/28423

## Release notes

```markdown
## Iceberg
* Add support for Azure vended credentials in Iceberg REST catalog. ({issue}`23238`)
```

### DIFF
--- a/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureAuthSasToken.java
+++ b/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureAuthSasToken.java
@@ -1,0 +1,51 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.filesystem.azure;
+
+import com.azure.storage.blob.BlobContainerClientBuilder;
+import com.azure.storage.file.datalake.DataLakeServiceClientBuilder;
+import com.google.common.collect.ImmutableMap;
+
+import java.util.Map;
+
+public final class AzureAuthSasToken
+        implements AzureAuth
+{
+    private final Map<String, String> sasTokens;
+
+    public AzureAuthSasToken(Map<String, String> sasTokens)
+    {
+        this.sasTokens = ImmutableMap.copyOf(sasTokens);
+    }
+
+    @Override
+    public void setAuth(String storageAccount, BlobContainerClientBuilder builder)
+    {
+        String sasToken = sasTokens.get(storageAccount);
+        if (sasToken == null) {
+            throw new IllegalStateException("No SAS token provided for storage account: " + storageAccount);
+        }
+        builder.sasToken(sasToken);
+    }
+
+    @Override
+    public void setAuth(String storageAccount, DataLakeServiceClientBuilder builder)
+    {
+        String sasToken = sasTokens.get(storageAccount);
+        if (sasToken == null) {
+            throw new IllegalStateException("No SAS token provided for storage account: " + storageAccount);
+        }
+        builder.sasToken(sasToken);
+    }
+}
--- a/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
+++ b/lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
@@ -31,11 +31,15 @@ import io.trino.spi.security.ConnectorIdentity;
 import jakarta.annotation.PreDestroy;
 import reactor.netty.resources.ConnectionProvider;
 
+import java.util.Map;
+import java.util.Map.Entry;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
 
 import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static io.airlift.concurrent.Threads.daemonThreadsNamed;
+import static io.trino.filesystem.azure.AzureFileSystemConstants.EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.Executors.newCachedThreadPool;
 
@@ -130,7 +134,19 @@ public class AzureFileSystemFactory
     @Override
     public TrinoFileSystem create(ConnectorIdentity identity)
     {
-        return new AzureFileSystem(httpClient, concurrencyPolicy, uploadExecutor, tracingOptions, auth, endpoint, readBlockSize, writeBlockSize, maxWriteConcurrency, maxSingleUploadSize, multipart);
+        AzureAuth effectiveAuth = getEffectiveAuth(identity);
+        return new AzureFileSystem(httpClient, concurrencyPolicy, uploadExecutor, tracingOptions, effectiveAuth, endpoint, readBlockSize, writeBlockSize, maxWriteConcurrency, maxSingleUploadSize, multipart);
+    }
+
+    private AzureAuth getEffectiveAuth(ConnectorIdentity identity)
+    {
+        Map<String, String> sasTokens = identity.getExtraCredentials().entrySet().stream()
+                .filter(e -> e.getKey().startsWith(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX))
+                .collect(toImmutableMap(e -> e.getKey().substring(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX.length()), Entry::getValue));
+        if (!sasTokens.isEmpty()) {
+            return new AzureAuthSasToken(sasTokens);
+        }
+        return auth;
     }
 
     public static HttpClient createAzureHttpClient(ConnectionProvider connectionProvider, EventLoopGroup eventLoopGroup, HttpClientOptions clientOptions)
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/azure/AzureFileSystemConstants.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/azure/AzureFileSystemConstants.java
@@ -13,13 +13,9 @@
  */
 package io.trino.filesystem.azure;
 
-import com.azure.storage.blob.BlobContainerClientBuilder;
-import com.azure.storage.file.datalake.DataLakeServiceClientBuilder;
-
-public sealed interface AzureAuth
-        permits AzureAuthAccessKey, AzureAuthDefault, AzureAuthOauth, AzureAuthSasToken
+public final class AzureFileSystemConstants
 {
-    void setAuth(String storageAccount, BlobContainerClientBuilder builder);
+    public static final String EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX = "internal$azure_sas_token_";
 
-    void setAuth(String storageAccount, DataLakeServiceClientBuilder builder);
+    private AzureFileSystemConstants() {}
 }
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
@@ -20,10 +20,13 @@ import io.trino.filesystem.TrinoFileSystemFactory;
 import io.trino.plugin.iceberg.IcebergFileSystemFactory;
 import io.trino.spi.security.ConnectorIdentity;
 import org.apache.iceberg.aws.s3.S3FileIOProperties;
+import org.apache.iceberg.azure.AzureProperties;
 import org.apache.iceberg.gcp.GCPProperties;
 
 import java.util.Map;
+import java.util.Map.Entry;
 
+import static io.trino.filesystem.azure.AzureFileSystemConstants.EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX;
 import static io.trino.filesystem.gcs.GcsFileSystemConstants.EXTRA_CREDENTIALS_GCS_OAUTH_TOKEN_EXPIRES_AT_PROPERTY;
 import static io.trino.filesystem.gcs.GcsFileSystemConstants.EXTRA_CREDENTIALS_GCS_OAUTH_TOKEN_PROPERTY;
 import static io.trino.filesystem.gcs.GcsFileSystemConstants.EXTRA_CREDENTIALS_GCS_PROJECT_ID_PROPERTY;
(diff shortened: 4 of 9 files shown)


ebyhr:
/test-with-secrets sha=4eeadf81784ccd8a9fa6c3b841293891a3852099

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureAuthSasToken.java
```diff
@@ -0,0 +1,52 @@
...
+import static java.util.Objects.requireNonNull;
+
+public final class AzureAuthSasToken
+        implements AzureAuth
+{
+    private final Map<String, String> sasTokens;
+
+    public AzureAuthSasToken(Map<String, String> sasTokens)
+    {
+        this.sasTokens = Map.copyOf(requireNonNull(sasTokens, "sasTokens is null"));
```
ebyhr:
```suggestion
        this.sasTokens = ImmutableMap.copyOf(sasTokens);
```

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
```diff
@@ -130,7 +133,27 @@ public void destroy()
...
+        AzureAuth effectiveAuth = getEffectiveAuth(identity);
+        return new AzureFileSystem(httpClient, concurrencyPolicy, uploadExecutor, tracingOptions, effectiveAuth, endpoint, readBlockSize, writeBlockSize, maxWriteConcurrency, maxSingleUploadSize, multipart);
+    }
+
+    private AzureAuth getEffectiveAuth(ConnectorIdentity identity)
+    {
+        Map<String, String> extraCredentials = identity.getExtraCredentials();
+        ImmutableMap.Builder<String, String> sasTokens = ImmutableMap.builder();
+
+        for (Map.Entry<String, String> entry : extraCredentials.entrySet()) {
```
ebyhr: Consider importing Map.Entry directly https://github.com/trinodb/trino/commit/dd690cb27dbf0bb893e3be260ab6bed3850d7ed8

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
```diff
@@ -130,7 +133,27 @@ public void destroy()
...
+            if (entry.getKey().startsWith(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX)) {
+                String account = entry.getKey().substring(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX.length());
+                sasTokens.put(account, entry.getValue());
+            }
+        }
+
+        Map<String, String> sasTokenMap = sasTokens.buildOrThrow();
```
ebyhr: nit: We could use stream so we can remove a redundant builder variable: 
```java
        Map<String, String> sasToken = extraCredentials.entrySet().stream()
                .filter(e -> e.getKey().startsWith(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX))
                .collect(toImmutableMap(e -> e.getKey().substring(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX.length()), Entry::getValue));
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
```diff
@@ -68,6 +71,19 @@ public TrinoFileSystem create(ConnectorIdentity identity, Map<String, String> fi
                 addOptionalProperty(extraCredentialsBuilder, fileIoProperties, GCPProperties.GCS_PROJECT_ID, EXTRA_CREDENTIALS_GCS_PROJECT_ID_PROPERTY);
             }
 
+            // handle azure
+            for (Map.Entry<String, String> entry : fileIoProperties.entrySet()) {
```
ebyhr: Consider importing Map.Entry directly https://github.com/trinodb/trino/commit/dd690cb27dbf0bb893e3be260ab6bed3850d7ed8

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergAzureVendingRestCatalogConnectorSmokeTest.java
```diff
@@ -0,0 +1,376 @@
...
+import static java.lang.String.format;
+import static org.apache.iceberg.FileFormat.PARQUET;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+final class TestIcebergAzureVendingRestCatalogConnectorSmokeTest
+        extends BaseIcebergConnectorSmokeTest
+{
+    private static final Logger LOG = Logger.get(TestIcebergAzureVendingRestCatalogConnectorSmokeTest.class);
+
+    private final String container;
```
ebyhr: This field can be converted to a local variable.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergAzureVendingRestCatalogConnectorSmokeTest.java
```diff
@@ -0,0 +1,376 @@
...
+    @AfterAll
+    public void removeTestData()
+    {
+        if (fileSystem == null) {
+            return;
+        }
+        try {
+            String location = warehouseLocation.endsWith("/")
+                               ? warehouseLocation.substring(0, warehouseLocation.length() - 1)
+                               : warehouseLocation;
```
ebyhr: The indentation looks wrong. Also, `warehouseLocation` always ends with / in my understanding.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/catalog/rest/TestIcebergAzureVendingRestCatalogConnectorSmokeTest.java
```diff
@@ -0,0 +1,376 @@
...
+    private static final Logger LOG = Logger.get(TestIcebergAzureVendingRestCatalogConnectorSmokeTest.class);
+
+    private final String container;
+    private final String account;
+    private final String accessKey;
+    private final String warehouseLocation;
+
+    private IcebergAzureRestCatalogBackendContainer restCatalog;
+
+    public TestIcebergAzureVendingRestCatalogConnectorSmokeTest()
```
ebyhr: Could you add this test to trino-iceberg pom.xml?

pom.xml
```diff
@@ -1935,6 +1935,12 @@
                 <version>${dep.iceberg.version}</version>
             </dependency>
 
+            <dependency>
+                <groupId>org.apache.iceberg</groupId>
+                <artifactId>iceberg-azure</artifactId>
+                <version>${dep.iceberg.version}</version>
+            </dependency>
```
ebyhr: We avoid defining a dependency in root pom.xml when it's used from a single module.

lib/trino-filesystem/src/main/java/io/trino/filesystem/azure/AzureFileSystemConstants.java
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.filesystem.azure;
+
+public final class AzureFileSystemConstants
+{
+    public static final String EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX = "internal$azure_sas_token_";
+    public static final String EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_EXPIRES_AT_MS_PREFIX = "internal$azure_sas_token_expires_at_ms_";
```
findinpath: Not currently used, but it may be relevant in case of wanting to "refresh" the vended credentials.

ebyhr:
/test-with-secrets sha=922a7a27abad1a24f88df3d552f89b1db6f49f68

ebyhr:
Could you fix TestIcebergAbfsVendingRestCatalogConnectorSmokeTest failures? https://github.com/trinodb/trino/actions/runs/23567816409/job/68623613038

ebyhr:
/test-with-secrets sha=8424686ae9a02b7fd842ee5792d8614e5fc89ad6

lib/trino-filesystem/src/main/java/io/trino/filesystem/azure/AzureFileSystemConstants.java
raphaelsolarski: Can't we just add it in case of need?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
```diff
@@ -68,6 +72,19 @@ public TrinoFileSystem create(ConnectorIdentity identity, Map<String, String> fi
                 addOptionalProperty(extraCredentialsBuilder, fileIoProperties, GCPProperties.GCS_PROJECT_ID, EXTRA_CREDENTIALS_GCS_PROJECT_ID_PROPERTY);
             }
 
+            // handle azure
+            for (Entry<String, String> entry : fileIoProperties.entrySet()) {
+                if (entry.getKey().startsWith(AzureProperties.ADLS_SAS_TOKEN_PREFIX)) {
```
raphaelsolarski, line 76: Can we receive here tokens for multiple accounts? 
I see that `AzureAuthSasToken` in contrast to other `AzureAuth` implementations is modeled as a bag of tokens for multiple accounts, so I want to ensure whether I understand it correctly.

findinpath: Yes, this is my understanding - see https://github.com/apache/iceberg/blob/main/azure/src/main/java/org/apache/iceberg/azure/AzureProperties.java

lib/trino-filesystem/src/main/java/io/trino/filesystem/azure/AzureFileSystemConstants.java
findinpath: this probably makes sense

let me remove it

ebyhr:
/test-with-secrets sha=fb4860036dd1b0b7bda9f98eefa3d47db02e7027

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
```diff
@@ -68,6 +71,14 @@ public TrinoFileSystem create(ConnectorIdentity identity, Map<String, String> fi
...
+                    String account = entry.getKey().substring(AzureProperties.ADLS_SAS_TOKEN_PREFIX.length());
+                    extraCredentialsBuilder.put(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX + account, entry.getValue());
```
chenjian2664: Do we have "expire time" info here?

chenjian2664, line 78: Why not adding the "expire time" info here?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
```diff
@@ -68,6 +71,14 @@ public TrinoFileSystem create(ConnectorIdentity identity, Map<String, String> fi
                 addOptionalProperty(extraCredentialsBuilder, fileIoProperties, GCPProperties.GCS_PROJECT_ID, EXTRA_CREDENTIALS_GCS_PROJECT_ID_PROPERTY);
             }
 
+            // handle azure
+            for (Entry<String, String> entry : fileIoProperties.entrySet()) {
+                if (entry.getKey().startsWith(AzureProperties.ADLS_SAS_TOKEN_PREFIX)) {
+                    String account = entry.getKey().substring(AzureProperties.ADLS_SAS_TOKEN_PREFIX.length());
```
chenjian2664, line 77: I think this behaves a bit differently compared to other filesystems like gcs, we prioritize the vended credentials, whereas in Azure the credentials are tied to the service account. In the latter case, we throw an exception if the credential cannot be found in the map, while for the account probably could success request with other type of auth
It's unclear whether Iceberg is able to return credentials for a different account, could you confirm this?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
findinpath: we're currently not using it

see https://github.com/trinodb/trino/pull/28855#discussion_r2988914764


This will be relevant at the time we'll be implementing vended credentials refreshing.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
findinpath: This seems to be the implementation also in `apache/iceberg`

https://github.com/apache/iceberg/blob/9f2bcf582e5267a860be616ad31c51d67327b792/azure/src/main/java/org/apache/iceberg/azure/adlsv2/VendedAdlsCredentialProvider.java#L92

chenjian2664: Can I understand that the the answer to :
`whether Iceberg is able to return credentials for a different account` is NO.
That means the service account is tied to the table(location) -thus from Iceberg it either returns same account's credentials or either return empty map declare the account don't have vended credentials.

lib/trino-filesystem-azure/src/main/java/io/trino/filesystem/azure/AzureFileSystemFactory.java
```diff
@@ -130,7 +134,19 @@ public void destroy()
...
+
+    private AzureAuth getEffectiveAuth(ConnectorIdentity identity)
+    {
+        Map<String, String> sasTokens = identity.getExtraCredentials().entrySet().stream()
+                .filter(e -> e.getKey().startsWith(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX))
+                .collect(toImmutableMap(e -> e.getKey().substring(EXTRA_CREDENTIALS_AZURE_SAS_TOKEN_PREFIX.length()), Entry::getValue));
+        if (!sasTokens.isEmpty()) {
+            return new AzureAuthSasToken(sasTokens);
+        }
+        return auth;
```
chenjian2664, line 149: @findinpath the fallback logic is use the `auth` here

findinpath: same as for GCS

https://github.com/trinodb/trino/blob/1f100c3409f3da5fa7f9c1e44262fe445f7c22bf/lib/trino-filesystem-gcs/src/main/java/io/trino/filesystem/gcs/GcsStorageFactory.java#L71-L76

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/catalog/rest/IcebergRestCatalogFileSystemFactory.java
chenjian2664: As discussed offline, here's the conclusion:
Azure tables may span multiple storage accounts, which is why multiple credentials are returned.
During vending credentials resolution, we use only one storage account(each client building for request). This account must be present in the credential map returned by Iceberg (the table should return either all accounts or at least the one corresponding to the current storage account). 
Since credential vending is expected to be supported, we [throw an exception](https://github.com/trinodb/trino/pull/28855/changes#diff-07e66e67118da691efd04e60ad50d3be825a99c144ef4d89f96d62fb6923cb9eR36-R38) if not able to find.

chenjian2664:
@findinpath ci fail relates to last update.

ebyhr:
Please update "Release notes" section to mention the new config properties.

PR #28860, opened 2026-03-25
Add unload table function for Hive connector

### LINKED ISSUE #19981: Support for `UNLOAD` functionality for Data Lake connectors
### Introduction

Amazon Athena exposes `UNLOAD` statement which can be used to save to a specified storage location the output of a query.
AWS Athena documentation for `UNLOAD` : https://docs.aws.amazon.com/athena/latest/ug/unload.html

Such a functionality would be beneficial to implement in Trino as well.

### Implementation roadmap

**End state**:
```
    CALL unload(path => …, format => ..., input => TABLE(...))
```

How do we get there:

**First phase**
```
SELECT * FROM TABLE(unload(path => …, input => TABLE (...)))
````

```
SELECT * FROM TABLE(hive.system.unload(
  input => TABLE( <subquery> ),
  location => 's3://bucket/path',
  format => 'parquet'))
```

What would the statement return?

A table function _today_ runs either partitioned or not.
We need to partition the writes and therefore we partition the output - so we produce multiple values for the statement.
The statement would therefore potentially return all the filenames of all the files written.

A table function should be side-effect free.
Calling however this function twice would probably write twice / potentially overwrite and corrupt (in case of Iceberg or Delta Lake tables) content in the specified location.
How would it perform in the context of FTE?

Limitations:
- No idempotency guarantees
- Limited support for controlling partitioning scheme (e.g., exact number of files written)
- Unload table function produces one output row for each file/partitioning element written, which will get displayed to the user as the result of the query

**Second phase**

```
SELECT * FROM TABLE(unload(path => …, input => TABLE (...)))
```
Aspects that can be improved independently from each other:
- Resolve idempotency protocol / commit protocol
- Control for partitioning scheme and scheduling
- Support for partial/final style execution for table functions, so function can return a single summary value (e.g., number of rows written, or files written, etc)

**End state**:
```
    CALL unload(path => …, format => ... , input => TABLE(...))
```

#### Use `CALL` statement 

Initial suggestion from @pettyjamesm 

```
CALL unload(
      location => "s3://bucket/path", 
      format => "parquet", 
      input => TABLE (SELECT * FROM orders))
```

This kind of call would fit the SQL specification. Yes, `CALL` can take table as a parameter (either a table or a query).
Input from @martint 

> This would probably be ideal from a syntactic point of view.

`CALL` is today like a TASK in the coordinator - it is not a "proper" query.
In order to  should be changed to be invoked in a “query”.
This could be sugared behind the scenes as 

```
SELECT * FROM TABLE(system.unload(
  input => TABLE( <subquery> ),
  location => 's3://bucket/path',
  format => 'parquet'))
```

`unload` would return a status value saying e.g.: how many rows were "affected"
The commit protocol should be taken into account.

#### Add support for a `UNLOAD` in the SQL Syntax

It looks attractive, but it doesn’t fit in the Trino model.
The engine wouldn't know to exactly which connector should point `UNLOAD` to.
Moreover, there could be scenarios where depending on the specified location, different connectors would need to be used.

#### Add an extra table parameter to CTAS statement to avoid registering the table

```
CREATE TABLE hive.myschema.mytable 
WITH (
    location=’....’,
    register_table= false) 
AS SELECT * FROM tcph.sf1.nation;
```
The table parameter `register_table` would signal to the connector that the output table of this statement shouldn't be registered to the metastore.
In the example above however `myschema` and `mytable` are both synthetic - they are not used anywhere.

#### Create a special connector with the sole purpose of doing CTAS but don't register the table

```
CREATE TABLE hive.myschema.mytable 
WITH (
    location=’....’) 
AS SELECT * FROM tcph.sf1.nation;
```

Same as the option mentioned previously, the connector would completely disregard  `myschema` and `mytable`.

#### Create a special connector which interprets differently the concept of schema name and table name

```
CREATE TABLE hive.bucket."table/location”
WITH (....)
AS SELECT * FROM tpch.sf1.nation;
```

The connector handling the above query would interpret the schema name as the _bucket name_ and the table name as the path in the bucket to which we'd like to save the table.

#### Use a reserved schema name in a general purpose data lake connector 

```
CREATE TABLE hive.raw."bucket/some/path"
WITH (....)
AS SELECT * FROM tpch.sf1.nation;
```

`raw` would be a reserved schema name in the connector.
This pragmatic choice comes with the side-effect of limiting the user not to use the reserved schema name.
From a UX perspective it feels rather awkward.

#### Insert into a table function the output of a `SELECT`

```
INSERT INTO TABLE(hive.system.unload(
  location => 's3://bucket/path',
  format => 'parquet'))
```

The syntax as it is listed, doesn’t fit in the standard.
This could be misleading for the users  - how should the users distinguish between read-only and writable table functions?

### PR DESCRIPTION
**Pull request by laserninja**
## Description

Implement the `unload` table function for the Hive connector, as described in #19981.

This allows exporting query results to files at a specified location in various formats.

### Usage

```sql
SELECT * FROM TABLE(hive.system.unload(
  input => TABLE(SELECT * FROM nation),
  location => 's3://bucket/path',
  format => 'PARQUET'))
```

### Supported formats
PARQUET (default), ORC, AVRO, CSV, JSON, OPENX_JSON, TEXTFILE

### Returns
| Column | Type | Description |
|--------|------|-------------|
| path | VARCHAR | Full path of the written file |
| rows_written | BIGINT | Number of rows written |
| bytes_written | BIGINT | Number of bytes written |

### Changes

**New files:**
- `UnloadFunction` — Table function definition with `analyze()`
- `UnloadFunctionHandle` — Serializable handle carrying function state
- `UnloadFunctionDataProcessor` — Processes input pages via `HiveFileWriterFactory`
- `UnloadFunctionProvider` — Guice provider with `ClassLoaderSafe` wrapper
- `HiveFunctionProvider` — `FunctionProvider` routing handles to processors
- `TestUnloadFunction` — Integration tests (8 tests)

**Modified files:**
- `HiveModule` — Added `FunctionProvider` and table function bindings
- `HiveConnector` — Added `getTableFunctions()` and `getFunctionProvider()` overrides
- `HiveConnectorFactory` — Wires new dependencies to `HiveConnector`

**Documentation:**
- Added Table functions section to Hive connector docs

Resolves #19981


plugin/trino-hive/src/main/java/io/trino/plugin/hive/functions/unload/UnloadFunction.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Map;
+import java.util.Optional;
+
+import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.function.table.ReturnTypeSpecification.GenericTable.GENERIC_TABLE;
+import static io.trino.spi.type.BigintType.BIGINT;
+import static io.trino.spi.type.VarcharType.VARCHAR;
+import static java.util.Locale.ENGLISH;
+
+public class UnloadFunction
```
ebyhr, line 45: There is currently no access control to restrict the target location. As a cluster administrator, I would prefer not to allow this table function.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/functions/unload/TestUnloadFunction.java
```diff
@@ -0,0 +1,130 @@
...
+
+import java.nio.file.Path;
+
+import static io.trino.plugin.hive.HiveQueryRunner.HIVE_CATALOG;
+import static io.trino.tpch.TpchTable.NATION;
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.jupiter.api.TestInstance.Lifecycle.PER_CLASS;
+
+@TestInstance(PER_CLASS)
+public class TestUnloadFunction
```
ebyhr, line 41: Please improve the test coverage: 
* All supported column types x all file formats
* All unsupported column types x all file formats
* All compression codec
* Column name sensitivity
* Large file
* Unload to non-existing directory
* Unload to non-empty directory
* FTE - we should throw an exception if this function doesn't support FTE

ebyhr (commented):
This requires a syntax review by @martint.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/functions/unload/UnloadFunction.java
laserninja: Makes sense. Will adding a configuration property 'hive.unload-enabled' and defaulting to false help? So administrators must opt in. I can also look into integrating with 'ConnectorAccessControl' for location-level checks. Would you prefer both, or is the config toggle sufficient for the initial version?

laserninja: I am curious how AWS Athena handles it. https://docs.aws.amazon.com/athena/latest/ug/unload.html
We don't use it. Can we loop in someone who does?

laserninja: For now used `hive.unload-enabled`

plugin/trino-hive/src/test/java/io/trino/plugin/hive/functions/unload/TestUnloadFunction.java
laserninja: Done.

PR #27623, opened 2025-12-11
Expose connector-level beans in a separate domain

wendigo:
## Release notes

(X) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Fix some things. ({issue}`issuenumber`)
```

PR #28864, opened 2026-03-25
Require plugins to bind mbeans in namespace

findepi:
`MBeanModule` is used by almost every connector. By default, it exports mbeans into global namespace, which is confusing. Disallow direct use of that module. Instead, require to use a wrapper that verifies an `ObjectNameGenerator` is properly bound.

- alternative to https://github.com/trinodb/trino/pull/27623

### DIFF
--- a/.mvn/modernizer/violations.xml
+++ b/.mvn/modernizer/violations.xml
@@ -259,6 +259,12 @@
     </violation>
 
     <violation>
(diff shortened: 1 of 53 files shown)


findepi:
it's a mess.  We don;'t have a config name generator that could export from base-jdbc and from eg trino-postgresql . Those names are hard to update in tests, hard to export consistently.
Maybe to be continued.

Issue #28835, opened 2026-03-24
Mismatch between HiveMetadata column types and engine statistics precision


anton-kutuzov:
## Problem
In `HiveMetadata.finishChangingTable`, the `columnTypes` map is constructed using the deprecated `getTypeSignature(HiveType)`. This method defaults `TIMESTAMP` to millisecond precision, ignoring the `hive.timestamp-precision` session property.

This creates a mismatch because the engine computes statistics blocks using the actual session precision (micros/nanos), leading to potential data corruption or `VerifyError` in `getTimestampEpochMicro`.

## Proposed Changes
Update `HiveMetadata` to respect session timestamp precision when building `columnTypes`.

PR #28865, opened 2026-03-25
Update HiveMetadata to respect session timestamp precision when build…

anton-kutuzov:
- Fix https://github.com/trinodb/trino/issues/28835

Using session timestamp precision when building columnTypes

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
@@ -6681,6 +6681,31 @@ public abstract class BaseHiveConnectorTest
     }
 
     @Test
+    public void testCollectStatisticsOnInsertTimestampWithPrecision()
+    {
+        for (HiveTimestampPrecision precision : HiveTimestampPrecision.values()) {
+            Session session = withTimestampPrecision(getSession(), precision);
+            String tableName = "test_stats_on_insert_timestamp_precision_" + precision.name() + randomNameSuffix();
+            try {
+                assertUpdate(format("CREATE TABLE %s (c_timestamp TIMESTAMP)", tableName));
+                assertUpdate(session,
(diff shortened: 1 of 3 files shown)


raunaqmorarka (approved):

plugin/trino-hive/src/test/java/io/trino/plugin/hive/BaseHiveConnectorTest.java
```diff
@@ -6680,6 +6680,32 @@ public void testCollectStatisticsOnCreateTableTimestampWithPrecision()
         }
     }
 
+    @Test
+    public void testCollectStatisticsOnInsertTimestampWithPrecision()
+    {
+        // Regression test for https://github.com/trinodb/trino/issues/28835
```
ebyhr: This test passes without this PR's change as far as I tested locally.

anton-kutuzov: Yes, that's right. This is happening because I changed the [timestamp](https://github.com/trinodb/trino/pull/28865/changes#diff-a72390c61d4a74f15d3df3149e26f677ded073b410c473eeb34234e717b6193eR303) retrieval, and it will crash without changes [here](https://github.com/trinodb/trino/pull/28865/changes#diff-1e06d5ca5f4d2b8a7964d589fbe0278d6aa9d99086c4e5ce2c4fb974b9184b20R1920). This test verify that `getTimestampEpochMicro` will not be broken in the future

ebyhr: I'm a bit confused. I would expect the **_regression tests_** to fail before this PR.

anton-kutuzov: I got it what you mean. I removed comment.

Issue #28405, opened 2026-02-20
Migrate all connectors that use DecimalModule to map to NUMBER


findepi:
- requires https://github.com/trinodb/trino/pull/28141
- requires https://github.com/trinodb/trino/issues/28392
- requires https://github.com/trinodb/trino/issues/28401

findepi:
work already done
 
- https://github.com/trinodb/trino/pull/28141
- https://github.com/trinodb/trino/issues/28392
- https://github.com/trinodb/trino/issues/28401
- https://github.com/trinodb/trino/pull/28744
- https://github.com/trinodb/trino/pull/28784

remaining connectors

- ClickHouse
- Redshift (it probably uses DecimalModule unnecessarily, according to docs, Redshift has only DECIMAL type that  directly maps to Trino's)
- Ignite

ningsh7:
Hi @findepi,

I noticed that the Ignite connector has not been migrated yet, and I would like to work on this part.

Ignite‘s `DECIMAL` type supports much larger precision than Trino `DECIMAL`, and even larger than the current practical precision limit of Trino `NUMBER`. Ignite supports `DECIMAL` precision up to 32767 and scale up to 16383.

My proposed approach for read mapping is:

- keep mapping Ignite `DECIMAL` to Trino `DECIMAL` when it can be represented exactly by Trino `DECIMAL`;
- map high-precision Ignite `DECIMAL` to Trino `NUMBER` only when it fits within the current safe precision capability of `NUMBER`. From the current implementation, `NUMBER` appears to keep up to 200 effective decimal digits;
- avoid implicitly mapping even larger Ignite `DECIMAL` declarations to `NUMBER`, since that may silently lose precision. For these cases, we can preserve the legacy `ALLOW_OVERFLOW` behavior when explicitly configured, or leave the type unsupported so users can use `unsupported-type-handling=CONVERT_TO_VARCHAR` if they prefer a string representation.

For write mapping, I checked the existing migrations. PostgreSQL and Oracle added write mappings from Trino NUMBER to unconstrained remote numeric types (`decimal` for PostgreSQL and `number` for Oracle). However, MySQL/MariaDB, SingleStore, and ClickHouse added read mapping only, with write mapping left for future work.

[...]

PR #28873, opened 2026-03-26
Map ClickHouse high-precision DECIMAL to Trino NUMBER

ebyhr:
## Description

Trino's `DECIMAL` has precision of 1-38 decimal digits. ClickHouse's `DECIMAL` supports precision up to 76 digits with scale up to 30. Previously it was impossible to reasonably map high precision ClickHouse `DECIMAL` to Trino. Mapping required agreeing to lossy conversion using `decimal-mapping=ALLOW_OVERFLOW` config and perhaps tuning the behavior further with `decimal-default-scale` and `decimal-rounding-mode` configs. These configs, along with supporting session properties, are provided by `DecimalModule` and reused across multiple connectors.

After the changes, the connector maps high-precision DECIMAL (precision 39-76) to NUMBER, the new Trino high precision decimal type. The decimal configs such as `decimal-mapping`, `decimal-default-scale` are deprecated in the connector. They remain non-deprecated from the perspective of all other connectors using `DecimalModule`.

For backwards compatibility, when `decimal-mapping` is set to `STRICT` or `ALLOW_OVERFLOW`, the mapping to NUMBER is *not* used.

Note on backwards incompatible behavior when `unsupported-type-handling` is used. Previously, when connector was configured with `unsupported-type-handling=CONVERT_TO_VARCHAR` (and without `decimal-mapping`), ClickHouse's high precision decimals were mapped to VARCHAR. Now they are properly supported and mapped to NUMBER, which constitutes a backward incompatible change. This backwards incompatibility is intrinsic nature of `unsupported-type-handling` though and not specific to this new mapping. `unsupported-type-handling` very nature is future incompatibility so anyone configuring this config signs up for a breaking change like this.

This change adds read mapping only. Write mapping may be added separately.

https://clickhouse.com/docs/sql-reference/data-types/decimal

- for https://github.com/trinodb/trino/issues/28405

## Release notes

```markdown
## ClickHouse
* Add support for reading all ClickHouse `DECIMAL` columns. ({issue}`28873`)
```

### DIFF
--- a/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
+++ b/plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
@@ -74,6 +74,7 @@ import io.trino.spi.type.CharType;
 import io.trino.spi.type.DecimalType;
 import io.trino.spi.type.Decimals;
 import io.trino.spi.type.Int128;
+import io.trino.spi.type.NumberType;
 import io.trino.spi.type.StandardTypes;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.TypeManager;
@@ -128,7 +129,6 @@ import static io.trino.plugin.clickhouse.TrinoToClickHouseWriteChecker.UINT16;
 import static io.trino.plugin.clickhouse.TrinoToClickHouseWriteChecker.UINT32;
 import static io.trino.plugin.clickhouse.TrinoToClickHouseWriteChecker.UINT64;
 import static io.trino.plugin.clickhouse.TrinoToClickHouseWriteChecker.UINT8;
-import static io.trino.plugin.jdbc.DecimalConfig.DecimalMapping.ALLOW_OVERFLOW;
 import static io.trino.plugin.jdbc.DecimalSessionSessionProperties.getDecimalDefaultScale;
 import static io.trino.plugin.jdbc.DecimalSessionSessionProperties.getDecimalRounding;
 import static io.trino.plugin.jdbc.DecimalSessionSessionProperties.getDecimalRoundingMode;
@@ -147,6 +147,8 @@ import static io.trino.plugin.jdbc.StandardColumnMappings.integerColumnMapping;
 import static io.trino.plugin.jdbc.StandardColumnMappings.integerWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.longDecimalReadFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.longDecimalWriteFunction;
+import static io.trino.plugin.jdbc.StandardColumnMappings.numberReadFunction;
+import static io.trino.plugin.jdbc.StandardColumnMappings.numberWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.realWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.shortDecimalWriteFunction;
 import static io.trino.plugin.jdbc.StandardColumnMappings.smallintColumnMapping;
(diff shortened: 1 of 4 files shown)


plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
```diff
@@ -886,6 +900,16 @@ private static ObjectWriteFunction uInt64WriteFunction(ClickHouseVersion version
                 });
     }
 
+    private static ColumnMapping numberColumnMapping()
+    {
+        // TODO (https://github.com/trinodb/trino/issues/7100) fix, enable and test decimal pushdown
```
findepi: should this be just before the `DISABLE_PUSHDOWN);` line?

plugin/trino-clickhouse/src/main/java/io/trino/plugin/clickhouse/ClickHouseClient.java
```diff
@@ -706,21 +708,33 @@ public Optional<ColumnMapping> toColumnMapping(ConnectorSession session, Connect
...
+                    return Optional.of(ColumnMapping.mapping(
+                            createDecimalType(precision, max(decimalDigits, 0)),
+                            decimalColumnMapping(createDecimalType(precision, max(decimalDigits, 0))).getReadFunction(),
+                            decimalColumnMapping(createDecimalType(precision, max(decimalDigits, 0))).getWriteFunction(),
+                            // TODO (https://github.com/trinodb/trino/issues/7100) fix, enable and test decimal pushdown
+                            DISABLE_PUSHDOWN));
                 }
-                else {
-                    decimalColumnMapping = decimalColumnMapping(createDecimalType(precision, max(decimalDigits, 0)));
+                // precision > MAX_PRECISION
```
findepi: nit: omit

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -473,6 +474,51 @@ public void testDecimal()
                 .execute(getQueryRunner(), clickhouseCreateAndInsert("tpch.test_nullable_decimal"));
     }
 
+    @Test
+    public void testDecimalExceedingPrecisionMax()
+    {
+        // Test that DECIMAL types with precision > 38 map to NUMBER type
+        // ClickHouse supports Decimal256 for precision from 39 to 76 digits
```
findepi:
```suggestion
        // ClickHouse uses Decimal256 for precision from 39 to 76 digits
```

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -473,6 +474,51 @@ public void testDecimal()
...
+        // Scale range: [0 : P] where P is the precision
```
findepi, line 483: 0-30 ?

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
```diff
@@ -473,6 +474,51 @@ public void testDecimal()
...
+                .addRoundTrip("Nullable(decimal(60, 20))", "NULL", NUMBER, "CAST(NULL AS NUMBER)")
+                .execute(getQueryRunner(), clickhouseCreateAndInsert("tpch.test_decimal_exceeding_precision_max_p60"));
+
+        // Test precision 76 (ClickHouse Decimal256's max), scale 30
+        SqlDataTypeTest.create()
+                .addRoundTrip("decimal(76, 30)", "1234567890123456789012345678901234567890123456.123456789012345678901234567890", NUMBER, "NUMBER '1234567890123456789012345678901234567890123456.123456789012345678901234567890'")
+                .addRoundTrip("decimal(76, 30)", "-1234567890123456789012345678901234567890123456.123456789012345678901234567890", NUMBER, "NUMBER '-1234567890123456789012345678901234567890123456.123456789012345678901234567890'")
+                .addRoundTrip("Nullable(decimal(76, 30))", "NULL", NUMBER, "CAST(NULL AS NUMBER)")
+                .execute(getQueryRunner(), clickhouseCreateAndInsert("tpch.test_decimal_exceeding_precision_max_p76"));
+    }
```
findepi, line 549: Maybe add negative tests
-  scale 31 
- negative scale -1.
- scale > precision (`decimal(10, 20)`)

If clickhouse were to add support for these, we would want to know and update the code appropriately.

ebyhr:
Rebased on master to resolve conflicts.

plugin/trino-clickhouse/src/test/java/io/trino/plugin/clickhouse/BaseClickHouseTypeMapping.java
ebyhr: Sorry, my test case was misleading. The max scale in ClickHouse is 76. Added additional test cases: 
* decimal(76, 76) - max P and S
* decimal(77, 00) - unsupported P
* decimal(5, -1) - negative S
* decimal(76, 77) - unsupported S > P

PR #20555, opened 2024-02-02
Add support for creating and dropping tables using Iceberg object store

jakelong95:
## Description

Currently, Trino is only able to write to and read from tables that use Iceberg's `ObjectStorageLocationProvider`, but is unable to create or drop tables using the location provider. 

This PR enables Trino to create tables using Iceberg's object storage by adding the following properties:
*  `iceberg.object-store.enabled` - Corresponds with Spark's `write.object-storage.enabled`, which enables use of Iceberg's ObjectStorageLocationProvider
*  `iceberg.data-location` - Corresponds to Spark's `write.data.path`, which sets where Iceberg data files will be written

Enabling the object store property and setting the data path will cause Iceberg to provide data file locations prefixed by  a deterministic hash generated from the file name in the location specified by `write.data.path`, which will help reduce throttling from cloud storage systems like S3 by evenly distributing files across multiple prefixes. [Iceberg's own documentation on this feature](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout) has more information.

For example, without the object store enabled you would get the following locations for data files, all under the same prefix `iceberg-tables/myschema/mytable/data`:
* `s3://mybucket/iceberg-tables/myschema/mytable/data/file1.parquet`
* `s3://mybucket/iceberg-tables/myschema/mytable/data/file2.parquet`
* `s3://mybucket/iceberg-tables/myschema/mytable/data/file3.parquet`

But, if you enable the object store and set the data path to  is set to `s3://mybucket/datafiles`, you would get the following locations, each in their own prefix:
* `s3://mybucket/datafiles/<file1 hash>/myschema/mytable/file1.parquet`
* `s3://mybucket/datafiles/<file2 hash>/myschema/mytable/file2.parquet`
* `s3://mybucket/datafiles/<file3 hash>/myschema/mytable/file3.parquet`

## Additional context and related issues

This PR maintains compatibility with Spark by using the table properties `write.object-storage.enabled` and `write.data.path`, which had previously been set up to allow Trino to write to Iceberg tables using the `ObjectStorageLocationProvider` in https://github.com/trinodb/trino/pull/8573

Fixes #8861

## Release notes

(x) Release notes are required, with the following suggested text:

```markdown
# Iceberg
* Add support for [object store file layout](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout). ({issue}`8861`)
```

PR #28887, opened 2026-03-26
Allow data_location without object store layout in Iceberg

Nazarii:
## Description
Previously, setting the `data_location` table property required `object_store_layout_enabled` to be `true`, otherwise Trino would reject the query with an error.

Remove this restriction so users can specify a custom data file location independently of the object store layout. The underlying Iceberg write.data.path property works without object store layout enabled, so the validation was unnecessarily limiting.

Previously, when using the Iceberg object store layout, it was not possible to specify an exact
location for data files — the layout adds a random prefix directory to file paths, making the
final location unpredictable.

The only workaround was to create the table in Spark with explicit data and metadata folder
locations, and then register it in Trino using `register_table`. This required an unnecessary
Spark dependency just for table creation.

With this change, data_location can be set independently of `object_store_layout_enabled`,
allowing tables to be created directly from Trino with a predictable data file location —
removing the need for the Spark workaround entirely.

## Release notes

```markdown
## Iceberg
* Allow configuring `data_location` table property without enabling `object_store_layout_enabled`. ({issue}`28887`)
```

### DIFF
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
@@ -9838,11 +9838,16 @@ public abstract class BaseIcebergConnectorTest
     }
 
     @Test
-    public void testCreateTableWithDataLocationButObjectStoreLayoutDisabled()
+    public void testCreateTableWithDataLocationWithoutObjectStoreLayout()
     {
-        assertQueryFails(
-                "CREATE TABLE test_data_location WITH (data_location = 'local:///data-location/xyz') AS SELECT 1 AS val",
-                "Data location can only be set when object store layout is enabled");
+        String tableName = "test_data_location_no_object_store" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " WITH (data_location = 'local:///data-location/xyz') AS SELECT 1 AS val", 1);
+
+        assertQuery("SELECT * FROM " + tableName, "VALUES 1");
+        String filePath = (String) computeScalar("SELECT file_path FROM \"" + tableName + "$files\"");
+        assertThat(filePath).startsWith("local:///data-location/xyz/");
+
+        assertUpdate("DROP TABLE " + tableName);
     }
 
     @Test
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2576,12 +2576,6 @@ public class IcebergMetadata
         if (properties.containsKey(DATA_LOCATION_PROPERTY)) {
             String dataLocation = (String) properties.get(DATA_LOCATION_PROPERTY)
                     .orElseThrow(() -> new IllegalArgumentException("The data_location property cannot be empty"));
-            boolean objectStoreEnabled = (boolean) properties.getOrDefault(
-                    OBJECT_STORE_LAYOUT_ENABLED_PROPERTY,
-                    Optional.of(Boolean.parseBoolean(icebergTable.properties().get(OBJECT_STORE_ENABLED)))).orElseThrow();
-            if (!objectStoreEnabled) {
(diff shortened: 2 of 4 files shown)


docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1105,7 +1105,8 @@ connector using a {doc}`WITH </sql/create-table-as>` clause.
   - Whether Iceberg's [object store file layout](https://iceberg.apache.org/docs/latest/aws/#object-store-file-layout) is enabled. 
     Defaults to `false`. 
 * - `data_location`
-  - Optionally specifies the file system location URI for the table's data files
+  - Optionally specifies the file system location URI for the table's data files.
+    Can be set independently of `object_store_layout_enabled`.
```
ebyhr: > Can be set independently of `object_store_layout_enabled`.

I don't think this comment is helpful for new readers. Remove.

ebyhr (commented):
Could you fix CI failures?

chenjian2664:
I came from https://github.com/trinodb/trino/pull/20555#issue-2115494156, has the situation changed? The throttling risk still exists, right?

Nazarii:
> Could you fix CI failures?

@ebyhr no, because the CI uses old versions of tests with old logic that I modified in the current PR.

Nazarii:
> I came from [#20555 (comment)](https://github.com/trinodb/trino/pull/20555#issue-2115494156), has the situation changed? The throttling risk still exists, right?

The throttling risk still exists when using `data_location` without `object_store_layout_enabled`, but this is no different from any standard Iceberg table that doesn't use `object store layout` — the risk is tied to `object_store_layout_enabled`, not to `data_location`. Users who need throttling protection should still set
  both properties together. This change simply removes the restriction that forced users to enable object store layout just to specify a custom data path.

docs/src/main/sphinx/connector/iceberg.md
Nazarii: Removed, thanks.

Nazarii:
> Could you fix CI failures?

@ebyhr I fixed tests and now CI fails with other failures - they are not related to our code. They're all caused by Docker Hub rate limiting. Do you have permissions to re-run failed jobs?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9640,11 +9640,16 @@ public void testObjectStoreLayoutEnabledAndDataLocation()
     }
 
     @Test
-    public void testCreateTableWithDataLocationButObjectStoreLayoutDisabled()
+    public void testCreateTableWithDataLocationWithoutObjectStoreLayout()
+            throws Exception
```
ebyhr: Remove redundant `throws Exception`.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
```diff
@@ -9640,11 +9640,16 @@ public void testObjectStoreLayoutEnabledAndDataLocation()
...
     {
-        assertQueryFails(
-                "CREATE TABLE test_data_location WITH (data_location = 'local:///data-location/xyz') AS SELECT 1 AS val",
-                "Data location can only be set when object store layout is enabled");
+        String tableName = "test_data_location_no_object_store" + randomNameSuffix();
+        assertUpdate("CREATE TABLE " + tableName + " WITH (data_location = 'local:///data-location/xyz') AS SELECT 1 AS val", 1);
+
+        String filePath = (String) computeScalar("SELECT file_path FROM \"" + tableName + "$files\"");
```
ebyhr, line 9847: Please verify the table result just in case, `"SELECT * FROM " + tableName`

Nazarii: Added.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergConnectorTest.java
Nazarii: Removed.

ebyhr (approved):
Please squash commits into one.

Nazarii:
> Please squash commits into one.

@ebyhr squashed

PR #28888, opened 2026-03-26
Fix OPTIMIZE failing with uppercase Iceberg column names

kinolaev:
## Description

IcebergMetadata.getColumnHandles() stored data column handles with original-case keys from the Iceberg schema, but ColumnMetadata lowercases all names in its constructor. When getColumnHandlesForOptimize() (introduced in 480) looked up columns by their ColumnMetadata names, the case-insensitive lookup failed for tables with uppercase column names.

The fix lowercases the key in getColumnHandles() to match the convention already enforced by MetadataManager at the engine boundary, and to match what ColumnMetadata.getName() returns.

## Release notes

```markdown
## Iceberg
* Fix failure when column name contains uppercase characters in OPTIMIZE procedure. ({issue}`28888`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2018,32 +2018,23 @@ public class IcebergMetadata
         IcebergTableExecuteHandle executeHandle = (IcebergTableExecuteHandle) tableExecuteHandle;
         IcebergTableHandle icebergTableHandle = (IcebergTableHandle) tableHandle;
         return switch (executeHandle.procedureId()) {
-            case OPTIMIZE -> getColumnHandlesForOptimize(session, icebergTableHandle, icebergTableHandle.getFormatVersion());
+            case OPTIMIZE -> getColumnHandlesForOptimize(icebergTableHandle);
             case OPTIMIZE_MANIFESTS, DROP_EXTENDED_STATS, ROLLBACK_TO_SNAPSHOT, EXPIRE_SNAPSHOTS, REMOVE_ORPHAN_FILES, ADD_FILES, ADD_FILES_FROM_TABLE ->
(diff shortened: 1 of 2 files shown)


ebyhr:
@kinolaev Thanks for detecting this bug quickly. Have you already submitted CLA?

kinolaev:
Yes, around 16:30 utc

chenjian2664:
IMO, update the `getColumnHandles` isn't ideally, it expand the scope - since analyzer and planner depends on the method, though I am not very strong about it.

Instead can we consider update `getColumnHandlesForOptimize` to get the schema directly, sth like:
```
                for (IcebergColumnHandle columnHandle : getTopLevelColumns(SchemaParser.fromJson(table.getTableSchemaJson()), typeManager)) {
                    columnHandles.add(columnHandle);
                }

                if (supportsRowLineage(((IcebergTableHandle) tableHandle).getFormatVersion())) {
                    columnHandles
                            .add(rowIdColumnHandle())
                            .add(lastUpdatedSequenceNumberColumnColumnHandle());
                }
```
cc @ebyhr @dain @findepi

ebyhr:
Perhaps, the default implementation of SPI has the same problem: 
https://github.com/trinodb/trino/blob/0899de1426ffba2cef9e24daff18e303dc94783d/core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorMetadata.java#L150-L164


```sql
# spark
CREATE TABLE test (X int) USING delta LOCATION 's3://test-bucket/test';
INSERT INTO test VALUES 1;
INSERT INTO test VALUES 2;

# trino
ALTER TABLE test EXECUTE optimize;
No handle found for column 'x' in table execute
```

findepi:
> Perhaps, the default implementation of SPI has the same problem

@ebyhr can you please file an issue?
if this is a recent regression, this might need to be a release blocker 🤔

ebyhr:
@findepi Sure, filed #28970 with a release blocker label.

ebyhr:
Let me continue the work at #28972 because it requires more work. I added @kinolaev as the co-author.

PR #28898, opened 2026-03-27
Recognize non-fallible division and modulo operator

losipiuk:
Current logic for division and modulo on integer types assumed it can always fail, which prohibited pushdown of filter expressions which involved those operators.
With new logic we relax the logic so if we see that divisor is constant and non-zero we asume that operator will never fail and pushdown is possible.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
+++ b/core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
@@ -16,13 +16,27 @@ package io.trino.sql.ir;
 import com.google.common.collect.ImmutableList;
 import io.trino.metadata.Metadata;
 import io.trino.metadata.ResolvedFunction;
+import io.trino.spi.type.BigintType;
+import io.trino.spi.type.DecimalType;
+import io.trino.spi.type.DoubleType;
+import io.trino.spi.type.Int128;
+import io.trino.spi.type.IntegerType;
+import io.trino.spi.type.NumberType;
+import io.trino.spi.type.RealType;
 import io.trino.spi.type.RowType;
+import io.trino.spi.type.SmallintType;
+import io.trino.spi.type.TinyintType;
+import io.trino.spi.type.TrinoNumber;
 import io.trino.sql.PlannerContext;
 import io.trino.type.TypeCoercion;
 
+import java.math.BigDecimal;
 import java.util.List;
 
+import static io.trino.metadata.GlobalFunctionCatalog.builtinFunctionName;
 import static io.trino.spi.block.RowValueBuilder.buildRowValue;
+import static io.trino.spi.function.OperatorType.DIVIDE;
+import static io.trino.spi.function.OperatorType.MODULUS;
 import static io.trino.spi.type.BooleanType.BOOLEAN;
 import static io.trino.spi.type.TypeUtils.writeNativeValue;
 import static io.trino.spi.type.VarcharType.VARCHAR;
@@ -72,7 +86,7 @@ public final class IrExpressions
             // These expressions need to verify their operands
             case Array e -> e.elements().stream().anyMatch(element -> mayFail(plannerContext, element));
             case Between e -> mayFail(plannerContext, e.value()) || mayFail(plannerContext, e.min()) || mayFail(plannerContext, e.max());
-            case Call e -> mayFail(e.function()) || e.arguments().stream().anyMatch(argument -> mayFail(plannerContext, argument));
+            case Call e -> mayFail(e) || e.arguments().stream().anyMatch(argument -> mayFail(plannerContext, argument));
             case Case e -> e.whenClauses().stream().anyMatch(clause -> mayFail(plannerContext, clause.getOperand()) || mayFail(plannerContext, clause.getResult())) ||
                     mayFail(plannerContext, e.defaultValue());
             case Cast e -> mayFail(plannerContext, e);
@@ -103,9 +117,45 @@ public final class IrExpressions
         return !cast.type().equals(VARCHAR);
     }
 
-    private static boolean mayFail(ResolvedFunction function)
+    private static boolean mayFail(Call call)
     {
-        return !function.neverFails() && !isDynamicFilterFunction(function.name());
+        ResolvedFunction function = call.function();
+        if (function.neverFails() || isDynamicFilterFunction(function.name())) {
+            return false;
+        }
+        List<Expression> arguments = call.arguments();
+        if (isModulsOrDivide(function) && arguments.get(1) instanceof Constant divisor && !canCauseDivisionByZeroError(divisor)) {
+            return false;
+        }
+        return true;
+    }
+
+    private static boolean isModulsOrDivide(ResolvedFunction function)
+    {
+        return (function.name().equals(builtinFunctionName(MODULUS)) || function.name().equals(builtinFunctionName(DIVIDE))) && function.signature().getArity() == 2;
+    }
+
+    private static boolean canCauseDivisionByZeroError(Constant divisor)
+    {
+        Object value = divisor.value();
+        if (value == null) {
+            return false; // dividing by null is null
+        }
+        return switch (divisor.type()) {
+            case TinyintType _, SmallintType _, IntegerType _, BigintType _ -> (long) value == 0;
+            case DecimalType decimalType -> {
+                if (decimalType.isShort()) {
+                    yield (long) value == 0;
+                }
+                yield ((Int128) value).isZero();
+            }
+            case NumberType _ -> switch (((TrinoNumber) value).toBigDecimal()) {
+                case TrinoNumber.BigDecimalValue(BigDecimal bigdecimal) -> bigdecimal.signum() == 0;
+                case TrinoNumber.Infinity _, TrinoNumber.NotANumber _ -> false;
+            };
+            case RealType _, DoubleType _ -> false; // will return NaN or ±Inf on division by 0
+            default -> true;
+        };
     }
 
     public static Expression not(Metadata metadata, Expression expression)
--- a/plugin/trino-hive/src/test/java/io/trino/plugin/hive/optimizer/TestHivePlans.java
+++ b/plugin/trino-hive/src/test/java/io/trino/plugin/hive/optimizer/TestHivePlans.java
@@ -60,7 +60,6 @@ import static io.trino.sql.ir.Booleans.TRUE;
 import static io.trino.sql.ir.Comparison.Operator.EQUAL;
 import static io.trino.sql.ir.Comparison.Operator.NOT_EQUAL;
 import static io.trino.sql.ir.Logical.Operator.AND;
-import static io.trino.sql.planner.assertions.PlanMatchPattern.any;
 import static io.trino.sql.planner.assertions.PlanMatchPattern.exchange;
 import static io.trino.sql.planner.assertions.PlanMatchPattern.filter;
 import static io.trino.sql.planner.assertions.PlanMatchPattern.join;
@@ -277,16 +276,16 @@ public class TestHivePlans
                 output(
                         join(INNER, builder -> builder
                                 .equiCriteria("L_INT_PART", "R_INT_COL")
-                                .filter(new Comparison(EQUAL, new Call(MODULUS_INTEGER, ImmutableList.of(new Reference(INTEGER, "R_INT_COL"), new Constant(INTEGER, 2L))), new Constant(INTEGER, 0L)))
                                 .left(
                                         exchange(REMOTE, REPARTITION,
-                                                any(
+                                                filter(
+                                                        new Comparison(EQUAL, new Call(MODULUS_INTEGER, ImmutableList.of(new Reference(INTEGER, "L_INT_PART"), new Constant(INTEGER, 2L))), new Constant(INTEGER, 0L)),
                                                         tableScan("table_int_partitioned", Map.of("L_INT_PART", "int_part", "L_STR_COL", "str_col")))))
                                 .right(
                                         exchange(LOCAL,
                                                 exchange(REMOTE, REPARTITION,
                                                         filter(
-                                                                new Between(new Reference(INTEGER, "R_INT_COL"), new Constant(INTEGER, 2L), new Constant(INTEGER, 4L)),
+                                                                new Logical(AND, ImmutableList.of(new In(new Reference(INTEGER, "R_INT_COL"), ImmutableList.of(new Constant(INTEGER, 2L), new Constant(INTEGER, 4L))), new Comparison(EQUAL, new Call(MODULUS_INTEGER, ImmutableList.of(new Reference(INTEGER, "R_INT_COL"), new Constant(INTEGER, 2L))), new Constant(INTEGER, 0L)))),
                                                                 tableScan("table_unpartitioned", Map.of("R_STR_COL", "str_col", "R_INT_COL", "int_col")))))))));
     }
 


core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +112,32 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+        return function.name().equals(builtinFunctionName(MODULUS)) || function.name().equals(builtinFunctionName(DIVIDE));
+    }
+
+    private static boolean canCauseDivisionByZero(Constant divisor)
+    {
+        Object value = divisor.value();
+        return switch (divisor.type()) {
+            case TinyintType _, SmallintType _, IntegerType _, BigintType _ -> Long.valueOf(0L).equals(value);
+            case DecimalType _ -> value instanceof Long l ? l == 0L : Int128.ZERO.equals(value);
+            default -> false;
```
pettyjamesm: Defaulting to `false` seems like the wrong thing to do when we're uncertain of the value type.

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +112,32 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+    private static boolean isModulsOrDivide(ResolvedFunction function)
+    {
+        return function.name().equals(builtinFunctionName(MODULUS)) || function.name().equals(builtinFunctionName(DIVIDE));
+    }
+
+    private static boolean canCauseDivisionByZero(Constant divisor)
+    {
+        Object value = divisor.value();
+        return switch (divisor.type()) {
+            case TinyintType _, SmallintType _, IntegerType _, BigintType _ -> Long.valueOf(0L).equals(value);
```
pettyjamesm: Relying on `Long#equals(Object)` to detect a zero value seems riskier because a zero-value of anything other than `Long` will return `false`. You might try something like this instead: `((Number) value).longValue() == 0`

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +112,32 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+            case DecimalType _ -> value instanceof Long l ? l == 0L : Int128.ZERO.equals(value);
```
pettyjamesm: Same thing here about assuming `value` is an instance of `Int128` and quietly doing the (potentially unsafe) thing if it's not.

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
findepi: Never use the `Number` interface. It backfires. It's on my personal denylist, even before LinkedList.

`Long.equals` works because all these types have long stack representation.
To avoid masking bugs, you can write this as `0 == (long) value`. (if you know value is not null)

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +112,32 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
 
-    private static boolean mayFail(ResolvedFunction function)
+    private static boolean mayFail(Call call)
     {
-        return !function.neverFails() && !isDynamicFilterFunction(function.name());
+        ResolvedFunction function = call.function();
+        List<Expression> arguments = call.arguments();
+        if (isModulsOrDivide(function) && arguments.get(1) instanceof Constant divisor && !canCauseDivisionByZero(divisor)) {
+            return false;
+        }
```
findepi, line 125: Put this after `if (function.neverFails() || isDynamicFilterFunction(function.name())) {`

I wish we could ask the function "can you fail for these arguments?"...

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +112,32 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+        if (function.neverFails() || isDynamicFilterFunction(function.name())) {
+            return false;
+        }
+        return true;
+    }
+
+    private static boolean isModulsOrDivide(ResolvedFunction function)
+    {
+        return function.name().equals(builtinFunctionName(MODULUS)) || function.name().equals(builtinFunctionName(DIVIDE));
```
findepi: `&& arguments.len == 2`

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
findepi: nit:

```
case DecimalType decimalType when decimalType.isShort() -> 0 == (long) value;
            case DecimalType _ -> ((Int128) value).isZero();
```

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
findepi: Add real, double, number.
then we have all numeric types covered.

(agree with James, that `default → true` is safe.)

findepi:
I changed PR title to what I suggest commit title could be.

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
losipiuk: I made it safer

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
losipiuk: Yeah - good catch

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
losipiuk: Yeah - that would be nicer - but I did not want to build whole generic mechanics for this case. 
I am pretty sure @wendigo will generalize :P

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
losipiuk: I did `&& function.signature().getArity() == 2`

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
findepi: > I wish we could ask the function "can you fail for these arguments?"...

sorry, that was not a request to change anything within this PR.




> Put this after `if (function.neverFails() || isDynamicFilterFunction(function.name())) {`

this was more important.

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
findepi: or `case DecimalType decimalType → decimalType.isshort() ? ..................... : ..............`

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +115,48 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
         return !cast.type().equals(VARCHAR);
     }
 
-    private static boolean mayFail(ResolvedFunction function)
+    private static boolean mayFail(Call call)
     {
-        return !function.neverFails() && !isDynamicFilterFunction(function.name());
+        ResolvedFunction function = call.function();
+        List<Expression> arguments = call.arguments();
+        if (isModulsOrDivide(function) && arguments.get(1) instanceof Constant divisor && !canCauseDivisionByZeroError(divisor)) {
```
pettyjamesm: I think @findepi wanted this to come after the `if (function.neverFails() || isDynamicFilterFunction(function.name()))` check below

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +115,48 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+        return (function.name().equals(builtinFunctionName(MODULUS)) || function.name().equals(builtinFunctionName(DIVIDE))) && function.signature().getArity() == 2;
+    }
+
+    private static boolean canCauseDivisionByZeroError(Constant divisor)
+    {
+        Object value = divisor.value();
+        return switch (divisor.type()) {
+            case TinyintType _, SmallintType _, IntegerType _, BigintType _ -> {
+                if (value == null) {
+                    yield false; // dividing by null is fine
```
findepi: Move this before `switch`. It's by necessity same for everyone.

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +115,48 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+                }
+                if (decimalType.isShort()) {
+                    yield (long) value == 0L;
+                }
+                else {
+                    yield ((Int128) value).isZero();
+                }
+            }
+            case RealType _, DoubleType _, NumberType _ -> false; // will return NaN on division by 0
```
findepi: / number 0 throws

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +115,48 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+                yield (long) value == 0L;
+            }
+            case DecimalType decimalType -> {
+                if (value == null) {
+                    yield false; // dividing by null is fine
+                }
+                if (decimalType.isShort()) {
+                    yield (long) value == 0L;
+                }
+                else {
```
findepi: nit: redundant else

findepi:
(we discovered the old truth: small PRs yield higher comment density 😉  )

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
pettyjamesm: Seems like that might be more brittle. Could there be some other type that supports "division or modulus" that would have different `NULL` semantics? (i.e.: the `default` branch)

losipiuk:
> (we discovered the old truth: small PRs yield higher comment density 😉 )

No - this is more about shitty coders.

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
losipiuk: Would be unexpected. Generally args for arithmetic operators are not nullable (so we get null as result if we get null as any arg)

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
losipiuk: Oops.

losipiuk:
Hopefully good now

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +117,45 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+    private static boolean isModulsOrDivide(ResolvedFunction function)
+    {
+        return (function.name().equals(builtinFunctionName(MODULUS)) || function.name().equals(builtinFunctionName(DIVIDE))) && function.signature().getArity() == 2;
+    }
+
+    private static boolean canCauseDivisionByZeroError(Constant divisor)
+    {
+        Object value = divisor.value();
+        if (value == null) {
+            return false; // dividing by null is fine
```
findepi:
```suggestion
            return false; // dividing by null is null
```

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +117,45 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+        return switch (divisor.type()) {
+            case TinyintType _, SmallintType _, IntegerType _, BigintType _ -> (long) value == 0L;
+            case DecimalType decimalType -> {
+                if (decimalType.isShort()) {
+                    yield (long) value == 0L;
+                }
+                yield ((Int128) value).isZero();
+            }
+            case NumberType _ -> switch (((TrinoNumber) value).toBigDecimal()) {
+                case TrinoNumber.BigDecimalValue bigDecimalValue -> bigDecimalValue.value().compareTo(BigDecimal.ZERO) == 0;
```
findepi:
```suggestion
                case TrinoNumber.BigDecimalValue(BigDecimal bigDecimal) -> bigDecimal.compareTo(BigDecimal.ZERO) == 0;
```

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +117,45 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+                if (decimalType.isShort()) {
+                    yield (long) value == 0;
+                }
+                yield ((Int128) value).isZero();
+            }
+            case NumberType _ -> switch (((TrinoNumber) value).toBigDecimal()) {
+                case TrinoNumber.BigDecimalValue(BigDecimal bigdecimal) -> bigdecimal.compareTo(BigDecimal.ZERO) == 0;
+                case TrinoNumber.Infinity _, TrinoNumber.NotANumber _ -> false;
+            };
+            case RealType _, DoubleType _ -> false; // will return NaN on division by 0
```
martint: Small issue, but the comment is not accurate. Only 0/0 will return NaN

losipiuk: True :) Updated

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +117,45 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+        return switch (divisor.type()) {
+            case TinyintType _, SmallintType _, IntegerType _, BigintType _ -> (long) value == 0;
+            case DecimalType decimalType -> {
+                if (decimalType.isShort()) {
+                    yield (long) value == 0;
+                }
+                yield ((Int128) value).isZero();
+            }
+            case NumberType _ -> switch (((TrinoNumber) value).toBigDecimal()) {
+                case TrinoNumber.BigDecimalValue(BigDecimal bigdecimal) -> bigdecimal.compareTo(BigDecimal.ZERO) == 0;
```
martint: BigDecimal.signum() == 0 is a better way to tell if the number is zero.

findepi: But it is also less direct.
Given than compareTo short-circuits using signum, it feels more right to stick to compareTo.

findepi:
CI
- https://github.com/trinodb/trino/issues/28904

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
```diff
@@ -103,9 +117,45 @@ private static boolean mayFail(PlannerContext plannerContext, Cast cast)
...
+                if (decimalType.isShort()) {
+                    yield (long) value == 0;
+                }
+                yield ((Int128) value).isZero();
+            }
+            case NumberType _ -> switch (((TrinoNumber) value).toBigDecimal()) {
+                case TrinoNumber.BigDecimalValue(BigDecimal bigdecimal) -> bigdecimal.signum() == 0;
+                case TrinoNumber.Infinity _, TrinoNumber.NotANumber _ -> false;
+            };
+            case RealType _, DoubleType _ -> false; // will return NaN or +-Inf on division by 0
```
findepi: "+-Inf" does not look pretty. either use `±` or omit `+-` 

```suggestion
            case RealType _, DoubleType _ -> false; // will return NaN or ±Inf on division by 0
```

core/trino-main/src/main/java/io/trino/sql/ir/IrExpressions.java
findepi: i think the idea here is that we understand what function is in the plan, and therefore can reason about its execution properties such as "may this call fail?" or "will this be null"

@pettyjamesm are you concerned that someone adds`/` operator on some other type that returns non-null (or fails) on null input (perhaps abusing `/` to implement path-like concatenation of varchars)
this is not impossible indeed.
to guard against that, we would need to change the `canCauseDivisionByZeroError(Constant)` function to exist with "can fail" on any urecognized type (and change the name). 
also, we would need to verify in `isModulsOrDivide` that operand types are the same (otherwise i could add `/` with heterogenous types and "fool" the rule)...

however, i would say it's an overkill. in thousand other places we don't do such rigorous checks so doing them here would not feel really right to me.

pettyjamesm: Yeah, probably overkill- just pointing it out since we were here.

findepi: yeah, keen eye!

PR #28901, opened 2026-03-27
Add support to NUMBER to Python UDFs

wendigo:
## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
+++ b/plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
@@ -17,6 +17,7 @@ import io.airlift.slice.DynamicSliceOutput;
 import io.airlift.slice.Slice;
 import io.airlift.slice.SliceInput;
 import io.airlift.slice.SliceOutput;
+import io.airlift.slice.Slices;
 import io.trino.spi.TrinoException;
 import io.trino.spi.block.Block;
 import io.trino.spi.block.SqlMap;
@@ -51,6 +52,8 @@ import io.trino.spi.type.TrinoNumber;
 import io.trino.spi.type.Type;
 import io.trino.spi.type.VarcharType;
 
+import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.math.BigDecimal;
 import java.util.List;
 
@@ -110,7 +113,7 @@ final class TrinoTypes
     public static Slice toRowTypeDescriptor(List<Type> types)
     {
         if (types.isEmpty()) {
-            SliceOutput output = new DynamicSliceOutput(8);
+            SliceOutput output = Slices.allocate(8).getOutput();
             output.writeInt(TrinoType.ROW.id());
             output.writeInt(0);
             return output.slice();
@@ -121,9 +124,13 @@ final class TrinoTypes
 
     public static Slice toTypeDescriptor(Type type)
     {
-        SliceOutput output = new DynamicSliceOutput(64);
-        toTypeDescriptor(type, output);
-        return output.slice();
+        try (SliceOutput output = new DynamicSliceOutput(64)) {
+            toTypeDescriptor(type, output);
+            return output.slice();
+        }
+        catch (IOException e) {
+            throw new UncheckedIOException(e);
+        }
     }
 
     private static void toTypeDescriptor(Type type, SliceOutput output)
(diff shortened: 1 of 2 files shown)


plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
```diff
@@ -201,6 +204,18 @@ private static void javaToBinary(Type type, Object value, SliceOutput output)
                         : Decimals.toString((Int128) value, decimalType.getScale());
                 writeVariableSlice(utf8Slice(decimalString), output);
             }
+            case NumberType _ -> {
+                TrinoNumber number = (TrinoNumber) value;
+
+                switch (number.toBigDecimal()) {
+                    case TrinoNumber.BigDecimalValue bigDecimalValue ->
```
findepi:
```suggestion
                    case TrinoNumber.BigDecimalValue(BigDecimal  bigDecimal) ->
```

plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
```diff
@@ -201,6 +204,18 @@ private static void javaToBinary(Type type, Object value, SliceOutput output)
...
+                            writeVariableSlice(utf8Slice(bigDecimalValue.toString()), output);
+                    case TrinoNumber.Infinity infinity ->
```
findepi: You can pull `writeVariableSlice(utf8Slice(` outside of `switch`

plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
```diff
@@ -201,6 +204,18 @@ private static void javaToBinary(Type type, Object value, SliceOutput output)
...
+                            writeVariableSlice(infinity.negative() ? utf8Slice("-Infinity") : utf8Slice("Infinity"), output);
+                    case TrinoNumber.NotANumber _ ->
```
findepi:
```suggestion
                    case TrinoNumber.NotANumber() ->
```

plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
findepi:
```suggestion
                    case TrinoNumber.Infinity(boolean negative) ->
```

plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
```diff
@@ -156,7 +159,7 @@ private static TrinoType singletonType(Type type)
             case StandardTypes.TINYINT -> TrinoType.TINYINT;
             case StandardTypes.DOUBLE -> TrinoType.DOUBLE;
             case StandardTypes.REAL -> TrinoType.REAL;
-            case StandardTypes.DECIMAL -> TrinoType.DECIMAL;
+            case StandardTypes.DECIMAL, StandardTypes.NUMBER -> TrinoType.DECIMAL;
```
findepi, line 169: What is the semantics of TrinoType enum?
I don't see javadoc nor usages on it... Why is it safe to present NUMBER as a decimal?

plugin/trino-functions-python/src/test/java/io/trino/plugin/functions/python/TestPythonFunctions.java
```diff
@@ -751,6 +751,27 @@ SELECT bad_bigint_return()
...
+                WITH (handler = 'multiply')
+                AS $$
+                from decimal import Decimal
+                def multiply(x, y):
+                    return x * y * Decimal("100000000000000000000.000000000000000000000000000000000000001")
+                $$
+                """;
+
+        assertThat(assertions.query(
+                query + "SELECT multiply(NUMBER '1.12345', NUMBER '2.54321')"))
```
findepi, line 772: Can you please add a test with NULL, NaN and ±Infinity?

plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
wendigo: It's a translation between the host and the wasm environment. Types are covered here: https://github.com/trinodb/trino-wasm-python/blob/master/pyhost.h#L10

plugin/trino-functions-python/src/test/java/io/trino/plugin/functions/python/TestPythonFunctions.java
wendigo: NaN and +/- Infinity are not supported by wasm python yet so nothing to test.

plugin/trino-functions-python/src/main/java/io/trino/plugin/functions/python/TrinoTypes.java
findepi: to me this looks like we need to expand that enum when a new trino type is added

plugin/trino-functions-python/src/test/java/io/trino/plugin/functions/python/TestPythonFunctions.java
findepi: the behavior should still be tested, regardless of current behavior

PR #28903, opened 2026-03-27
Reapply "Add support for VARBINARY to JSON conversions"

gertjanal:
This reverts commit 335dc00cc71e4196a24dbc42efb417a0d12b6105.
- https://github.com/trinodb/trino/pull/28635
- https://github.com/trinodb/trino/pull/28234

Thanks to @wendigo for his contribution to this PR

## Description
Casting of `row(greeting varchar, varbinary planet)` to json was not possible.

This PR adds support by converting the varbinary to base64 string:
`{"greeting": "hello", "planet": "d29ybGQ="}`

- Fixes https://github.com/trinodb/trino/issues/10814

## Release notes

```markdown
## General
* Add support for casting `varbinary` to `json` with base64. ({issue}`28234`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/VarbinaryFunctions.java
@@ -19,6 +19,7 @@ import com.google.common.io.BaseEncoding;
 import com.google.common.primitives.Ints;
 import io.airlift.slice.Murmur3Hash128;
 import io.airlift.slice.Slice;
+import io.airlift.slice.SliceOutput;
 import io.airlift.slice.Slices;
 import io.airlift.slice.SpookyHashV2;
 import io.airlift.slice.XxHash64;
@@ -26,6 +27,7 @@ import io.trino.spi.TrinoException;
 import io.trino.spi.function.Description;
 import io.trino.spi.function.LiteralParameters;
 import io.trino.spi.function.ScalarFunction;
+import io.trino.spi.function.ScalarOperator;
 import io.trino.spi.function.SqlType;
 import io.trino.spi.type.StandardTypes;
 
@@ -35,7 +37,11 @@ import java.util.zip.CRC32;
 
 import static io.airlift.slice.Slices.EMPTY_SLICE;
 import static io.trino.operator.scalar.HmacFunctions.computeHash;
+import static io.trino.spi.StandardErrorCode.INVALID_CAST_ARGUMENT;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.function.OperatorType.CAST;
+import static io.trino.spi.type.StandardTypes.JSON;
+import static io.trino.spi.type.StandardTypes.VARBINARY;
 import static io.trino.util.Failures.checkCondition;
 
 public final class VarbinaryFunctions
@@ -500,5 +506,28 @@ public final class VarbinaryFunctions
             reverse.setByte(i, inputSlice.getByte((length - 1) - i));
         }
         return reverse;
+    }
+
+    @SqlType(JSON)
+    @ScalarOperator(CAST)
+    public static Slice castToJson(@SqlType(VARBINARY) Slice binary)
+    {
+        Slice encodedBinary = toBase64(binary);
+        Slice slice = Slices.allocate(2 + encodedBinary.length());
+        SliceOutput output = slice.getOutput();
+        output.writeByte('"');
+        output.writeBytes(encodedBinary);
+        output.writeByte('"');
+        return slice;
+    }
+
+    @SqlType(VARBINARY)
+    @ScalarOperator(CAST)
+    public static Slice castFromJson(@SqlType(JSON) Slice json)
+    {
+        if (json.length() < 2 || json.getByte(0) != '"' || json.getByte(json.length() - 1) != '"') {
+            throw new TrinoException(INVALID_CAST_ARGUMENT, "Expected quoted base64-encoded JSON string");
+        }
+        return fromBase64Varbinary(json.slice(1, json.length() - 2));
     }
 }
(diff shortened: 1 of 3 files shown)


gertjanal:
Considerations by @martint : https://github.com/trinodb/trino/pull/28234#issuecomment-4034912700
```
There are a number of things to consider:

* Why base64? Why not hex-encoded, or SQL-binary-literal encoded?
* How does it relate to SQL/JSON data model and JSON type defined in the latest version of the SQL spec. In fact, there are changes we need to make to the JSON data type to make it compatible with the SQL spec. This could be a step in the wrong direction
* How does it interoperate with the JSON processing functions?
* etc.
```

gertjanal:
> Considerations by @martint : [#28234 (comment)](https://github.com/trinodb/trino/pull/28234#issuecomment-4034912700)
> 
> ```
> There are a number of things to consider:
> 
> * Why base64? Why not hex-encoded, or SQL-binary-literal encoded?
> * How does it relate to SQL/JSON data model and JSON type defined in the latest version of the SQL spec. In fact, there are changes we need to make to the JSON data type to make it compatible with the SQL spec. This could be a step in the wrong direction
> * How does it interoperate with the JSON processing functions?
> * etc.
> ```

* I've choosen base64 as it is the most compact way to encode binary in a string and also a common way to encode binary (hex for example takes way more characters)
* I'm not aware of any binary JSON in SQL, can you elaborate?
* All other JSON processing should be fine as the representation acts as a normal string value

martint:
I’m not convinced we should add `CAST(VARBINARY AS JSON)` with “base64-encode into a JSON string” semantics in the current codebase.

The main issue is that this is not really a JSON-semantic conversion. It is introducing an application-level encoding convention inside an ordinary JSON string. After the cast, the value is just a JSON string, and it is no longer distinguishable from any other string that happens to contain base64 text. That makes the cast ambiguous, especially in the reverse direction, and it gets even harder to reason about once `VARBINARY` appears inside `ROW`, `MAP`, or `ARRAY`. At that point, generic casts to JSON would silently turn binary leaves into strings, and we would lose any principled way to tell “this was binary” from “this was text”.

I think this is particularly problematic given the desire to move Trino’s `JSON` type toward the SQL/JSON data model. In that model, `JSON` is no longer just textual JSON. It is a typed value model with scalars, arrays, and objects, and some scalars (for example datetime-like values) do not round-trip through ordinary RFC 8259 text. If we add `VARBINARY -> JSON` today by mapping it to a base64 JSON string, we are baking in the old “JSON is text” mindset right where we are trying to move away from it. That makes it harder to introduce coherent new JSON semantics later, because we would already have a public cast whose meaning depends on smuggling type information through strings.

The SQL/JSON model is a better foundation here. Under that model, a cast should ideally produce a JSON value in the model, not an encoded representation. So if we ever want to support binary as a Trino extension, the coherent design would be to add binary as an actual SQL/JSON scalar kind, similar in spirit to how the model already has typed scalars beyond plain RFC 8259 text. Then `CAST(VARBINARY AS JSON)` would construct a binary JSON scalar directly, with no base64 conversion at cast time. Arrays, maps, and rows containing `VARBINARY` would naturally contain binary JSON items as well, instead of silently degrading into strings.

Of course, that has implications for the rest of the system: path semantics, `JSON_VALUE(... RETURNING VARBINARY)`, client/protocol handling, and especially serialization. Plain JSON text has no binary literal syntax, so if a binary JSON scalar needs to be rendered as RFC 8259 text, that is where a base64 or tagged-text representation would belong, not in the cast itself. In other words, base64 may be a reasonable serialization convention, but it is not a good `VARBINARY -> JSON` cast semantics.

So my take is:
- adding this cast in the current codebase is a bad fit, especially while we are trying to align `JSON` with the SQL/JSON model
- if we want first-class support long term, binary should be introduced as an actual extension to the SQL/JSON value model, not encoded into JSON strings by cast

gertjanal:
> I’m not convinced we should add `CAST(VARBINARY AS JSON)` with “base64-encode into a JSON string” semantics in the current codebase.
> 
> The main issue is that this is not really a JSON-semantic conversion. It is introducing an application-level encoding convention inside an ordinary JSON string. After the cast, the value is just a JSON string, and it is no longer distinguishable from any other string that happens to contain base64 text. That makes the cast ambiguous, especially in the reverse direction, and it gets even harder to reason about once `VARBINARY` appears inside `ROW`, `MAP`, or `ARRAY`. At that point, generic casts to JSON would silently turn binary leaves into strings, and we would lose any principled way to tell “this was binary” from “this was text”.
> 
> I think this is particularly problematic given the desire to move Trino’s `JSON` type toward the SQL/JSON data model. In that model, `JSON` is no longer just textual JSON. It is a typed value model with scalars, arrays, and objects, and some scalars (for example datetime-like values) do not round-trip through ordinary RFC 8259 text. If we add `VARBINARY -> JSON` today by mapping it to a base64 JSON string, we are baking in the old “JSON is text” mindset right where we are trying to move away from it. That makes it harder to introduce coherent new JSON semantics later, because we would already have a public cast whose meaning depends on smuggling type information through strings.
> 
> The SQL/JSON model is a better foundation here. Under that model, a cast should ideally produce a JSON value in the model, not an encoded representation. So if we ever want to support binary as a Trino extension, the coherent design would be to add binary as an actual SQL/JSON scalar kind, similar in spirit to how the model already has typed scalars beyond plain RFC 8259 text. Then `CAST(VARBINARY AS JSON)` would construct a binary JSON scalar directly, with no base64 conversion at cast time. Arrays, maps, and rows containing `VARBINARY` would naturally contain binary JSON items as well, instead of silently degrading into strings.
> 
> Of course, that has implications for the rest of the system: path semantics, `JSON_VALUE(... RETURNING VARBINARY)`, client/protocol handling, and especially serialization. Plain JSON text has no binary literal syntax, so if a binary JSON scalar needs to be rendered as RFC 8259 text, that is where a base64 or tagged-text representation would belong, not in the cast itself. In other words, base64 may be a reasonable serialization convention, but it is not a good `VARBINARY -> JSON` cast semantics.
> 
> So my take is:
> 
>     * adding this cast in the current codebase is a bad fit, especially while we are trying to align `JSON` with the SQL/JSON model
> 
>     * if we want first-class support long term, binary should be introduced as an actual extension to the SQL/JSON value model, not encoded into JSON strings by cast

Thanks for the extensive reply, I understand

PR #28907, opened 2026-03-28
Fix deleting incorrect rows in Delta Lake

### LINKED ISSUE #28885: Wrong records DELETEed on delta lake parquet files produced by Apache Spark
### Trino version

All recent versions up to 477

### Please describe the bug

Hi all,

we have stumbled upon an issue when using the Delta Lake connector to process tables written by Spark using the Parquet format.

In summary: Trino delete operations seem to completly misinterpret the filter and end up deleting the wrong data.

We have found that disabling parquet column index (`delta.parquet_use_column_index=false`) "fixes" the issue but this is more of a workaround and not a solution.

We also used Claude to analyze the problem and it identifed it as:

> When column index filtering is enabled, Parquet skips entire pages whose min/max stats don't overlap the predicate. This reduces currentGroupRowCount to the filtered count (ParquetReader.java:498). But   
>  lastBatchStartRow() at line 448:
>
>  return firstRowIndexInGroup + nextRowInGroup - batchSize;
>
>  ...uses nextRowInGroup, which tracks position within the filtered output, not within the physical file. After a gap of skipped pages, the row numbers are off by the size of the gap.

To reproduce the problem, we used this minimal [Spark application](https://github.com/razvan/spark-delta-s3/tree/delta) with data stored in a S3 (minio) bucket.

In this example the DELETE operation removes all "text1" records instead of "text5".

After successfully running the Spark application, create the Trino table and select some records.

```sql
CALL delta.system.register_table(
   schema_name => 'default',
   table_name => 'delta_table',
   table_location => 's3a://trino/delta-table'
 );
 
select text, count(*) from delta.default.delta_table where text='text5' group by text;
-- text5	5000
```

Delete all "text5" records.

```sql
delete from delta.default.delta_table where text = 'text5';
-- Updated Rows	5000
-- Execute time	0.929s
-- Start time	Thu Mar 26 13:50:45 CET 2026
-- Finish time	Thu Mar 26 13:50:46 CET 2026
-- Query	delete from delta.default.delta_table where text = 'text5'
```

Unfortunately the records have not been deleted.

```sql
select text, count(*) from delta.default.delta_table where text='text5' group by text;
-- text5	5000
```

But the "text1" records have been deleted instead.

```sql
select text, count(*) from delta.default.delta_table group by text order by text asc;
-- text0	5000
-- text2	5000
-- text3	5000
-- text4	5000
-- text5	5000
-- text6	5000
-- text7	5000
-- text8	5000
-- text9	5000

select text, count(*) from delta.default.delta_table where text='text1' group by text;
-- nothing is returned
```

ebyhr:
## Description

Fixes #28885

## Release notes

```markdown
## Delta Lake
* Fix incorrect results when deleting rows. ({issue}`28885`)
```


lfrancke:
Thanks for the fix. I'd say the release notes understate the importance a bit :)

"Incorrect results" -> "can delete random data"

This issue can/will/did lead to data loss

PR #28911, opened 2026-03-29
Add added_snapshot_id column to Iceberg $files table

### LINKED ISSUE #28910: Iceberg: `$partitions` metadata table missing delete file metrics
## Feature Request / Improvement

Trino's Iceberg `$partitions` metadata table does not expose any information about delete files
per partition. Iceberg's native `PARTITIONS` metadata table includes
`position_delete_record_count`, `position_delete_file_count`, `equality_delete_record_count`,
`equality_delete_file_count`, `spec_id`, `last_updated_at`, and `last_updated_snapshot_id` --
all of which Trino's implementation omits.

This makes it difficult to identify partitions that have accumulated excessive delete files and
need `OPTIMIZE`, without resorting to expensive queries against `$files`.

## Current Trino `$partitions` schema

```sql
SELECT * FROM "my_table$partitions";
```

| Column         | Type   | Description                       |
| -------------- | ------ | --------------------------------- |
| `partition`    | ROW    | Partition values                  |
| `record_count` | BIGINT | Count of data records             |
| `file_count`   | BIGINT | Count of data files               |
| `total_size`   | BIGINT | Total size of data files in bytes |
| `data`         | ROW    | Column-level min/max/null/nan stats |

## Iceberg native `PARTITIONS` schema (available in Spark)

| Column                           | Type      | In Trino?          | Description                                  |
| -------------------------------- | --------- | ------------------ | -------------------------------------------- |
| `partition`                      | STRUCT    | Yes                | Partition values                             |
| `spec_id`                        | INT       | **No**             | Partition spec ID                            |
| `record_count`                   | LONG      | Yes                | Count of data records                        |
| `file_count`                     | INT       | Yes                | Count of data files                          |
| `total_data_file_size_in_bytes`  | LONG      | Yes (as `total_size`) | Total data file size                      |
| `position_delete_record_count`   | LONG      | **No**             | Count of records in positional delete files  |
| `position_delete_file_count`     | INT       | **No**             | Count of positional delete files             |
| `equality_delete_record_count`   | LONG      | **No**             | Count of records in equality delete files    |
| `equality_delete_file_count`     | INT       | **No**             | Count of equality delete files               |
| `last_updated_at`                | TIMESTAMP | **No**             | Commit time of last update to this partition |
| `last_updated_snapshot_id`       | LONG      | **No**             | Snapshot ID that last updated this partition |

### Identifying partitions needing OPTIMIZE

Delete file accumulation is a common operational problem with Iceberg v2 tables using
merge-on-read (MOR) deletes. Each `DELETE` or `MERGE` statement can produce one positional
delete file per affected data file. Over time, partitions accumulate hundreds of delete files,
degrading read performance.

Today, identifying affected partitions requires querying `$files` and grouping manually:

```sql
-- Current workaround: scans ALL file metadata, expensive on large tables
SELECT
    partition,
    count(*) FILTER (WHERE content = 1) AS pos_delete_files,
    count(*) FILTER (WHERE content = 2) AS eq_delete_files,
    sum(record_count) FILTER (WHERE content = 1) AS pos_delete_records,
    sum(record_count) FILTER (WHERE content = 2) AS eq_delete_records
FROM "my_table$files"
GROUP BY partition
HAVING count(*) FILTER (WHERE content IN (1, 2)) > 0
```

While `$files` does have a `partition` column, this approach is expensive on large tables
because it scans every file's metadata (data files + delete files) just to get partition-level
aggregates. The `$partitions` table is designed for exactly this use case -- pre-aggregated
partition-level metrics -- but currently omits delete file information entirely.

### What it should look like

With the proposed columns, identifying problematic partitions becomes trivial:

```sql
-- Proposed: instant, partition-level delete file visibility
SELECT
    partition,
    file_count,
    position_delete_file_count,
    equality_delete_file_count,
    position_delete_record_count
FROM "my_table$partitions"
WHERE position_delete_file_count > 10
ORDER BY position_delete_file_count DESC;
```

### Operational monitoring

The `last_updated_at` and `last_updated_snapshot_id` columns enable time-based monitoring:

```sql
SELECT
    partition,
    position_delete_file_count,
    last_updated_at
FROM "my_table$partitions"
WHERE position_delete_file_count > 0
ORDER BY last_updated_at DESC;
```

## Root cause

Trino's `PartitionsTable` is a custom implementation that does not delegate to Iceberg's native
`MetadataTableType.PARTITIONS`. Other system tables like `$snapshots`, `$entries`,
`$all_manifests` extend `BaseSystemTable` and delegate to Iceberg's metadata table
infrastructure. `PartitionsTable` implements `SystemTable` directly and builds its own
aggregation.

The key gap is in
[`PartitionsTable.getStatisticsByPartition()`](https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java#L175-L194):

```java
private Map<StructLikeWrapperWithFieldIdToIndex, IcebergStatistics> getStatisticsByPartition(TableScan tableScan)
{
    try (CloseableIterable<FileScanTask> fileScanTasks = tableScan.planFiles()) {
        Map<StructLikeWrapperWithFieldIdToIndex, IcebergStatistics.Builder> partitions = new HashMap<>();
        for (FileScanTask fileScanTask : fileScanTasks) {
            DataFile dataFile = fileScanTask.file();
            // ...
            .acceptDataFile(dataFile, fileScanTask.spec());
            // fileScanTask.deletes() is completely IGNORED
        }
        // ...
    }
}
```

The loop iterates `FileScanTask` objects and calls `acceptDataFile(dataFile)` but **never reads
`fileScanTask.deletes()`**. The delete files associated with each scan task are discarded.

Similarly,
[`IcebergStatistics`](https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java#L44-L52)
and its `Builder` only track data file metrics:

```java
public record IcebergStatistics(
        long recordCount,
        long fileCount,
        long size,
        Map<Integer, Object> minValues,
        Map<Integer, Object> maxValues,
        Map<Integer, Long> nullCounts,
        Map<Integer, Long> nanCounts,
        Map<Integer, Long> columnSizes)
```

No fields for delete file counts or records.

<details>
<summary>Comparison with Iceberg's native implementation</summary>

Iceberg's own
[`PartitionsTable.Partition.update()`](https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/PartitionsTable.java)
method handles all file types:

```java
void update(ContentFile<?> file, Snapshot snapshot) {
    switch (file.content()) {
        case DATA:
            this.dataRecordCount += file.recordCount();
            this.dataFileCount += 1;
            this.dataFileSizeInBytes += file.fileSizeInBytes();
            break;
        case POSITION_DELETES:
            this.posDeleteRecordCount += file.recordCount();
            this.posDeleteFileCount += 1;
            break;
        case EQUALITY_DELETES:
            this.eqDeleteRecordCount += file.recordCount();
            this.eqDeleteFileCount += 1;
            break;
    }
}
```

It iterates manifest entries directly (not `FileScanTask`), so it sees all files regardless of
type, including delete files in partitions with no remaining data files.

</details>

###  Add delete file columns to the existing custom implementation

Add delete file metrics to the existing `PartitionsTable` + `IcebergStatistics` flow.

**Files to change:**

1. **`IcebergStatistics.java`** -- Add fields to the record and Builder:
   - `positionDeleteFileCount` (long)
   - `positionDeleteRecordCount` (long)
   - `equalityDeleteFileCount` (long)
   - `equalityDeleteRecordCount` (long)
   - Add `acceptDeleteFile(DeleteFile)` method to Builder

2. **`PartitionsTable.java`** -- Two changes:
   - Add new columns to `columnMetadataBuilder`
   - In `getStatisticsByPartition()`, iterate delete files from `fileScanTask.deletes()`.
     **Note:** The same delete file can appear in multiple `FileScanTask` objects (e.g., an
     equality delete file applies to all data files in a partition). A `Set<String>` of seen
     delete file paths per partition is needed to avoid double-counting.
   - In `buildRecordCursor()`, emit the new column values

3. **Tests** -- Update `$partitions` test assertions to verify new columns

kaveti:
## Summary

- Add `added_snapshot_id` to the Iceberg `$files` system table.
- Populate `added_snapshot_id` from live manifest entry snapshot metadata.
- Preserve upstream `$files` schema additions introduced after this branch diverged.
- Update tests and docs to match the merged schema.

## Release notes

```markdown
## Section
* Fixes https://github.com/trinodb/trino/issues/28910
```

Fixes #28910

### DIFF
--- a/docs/src/main/sphinx/connector/iceberg.md
+++ b/docs/src/main/sphinx/connector/iceberg.md
@@ -1490,9 +1490,9 @@ SELECT * FROM "test_table$files";
 ```
 
 ```text
- content  | file_path                                                                                                                     | record_count    | file_format   | file_size_in_bytes   |  column_sizes        |  value_counts     |  null_value_counts | nan_value_counts  | lower_bounds                |  upper_bounds               |  key_metadata  | split_offsets  |  equality_ids  | file_sequence_number | data_sequence_number | referenced_data_file | pos | manifest_location                                                                                                            | first_row_id | content_offset | content_size_in_bytes
-----------+-------------------------------------------------------------------------------------------------------------------------------+-----------------+---------------+----------------------+----------------------+-------------------+--------------------+-------------------+-----------------------------+-----------------------------+----------------+----------------+----------------+----------------------+----------------------+----------------------+-----+------------------------------------------------------------------------------------------------------------------------------+--------------+----------------+----------------------
- 0        | hdfs://hadoop-master:9000/user/hive/warehouse/test_table/data/c1=3/c2=2021-01-14/af9872b2-40f3-428f-9c87-186d2750d84e.parquet |  1              |  PARQUET      |  442                 | {1=40, 2=40, 3=44}   |  {1=1, 2=1, 3=1}  |  {1=0, 2=0, 3=0}   | <null>            |  {1=3, 2=2021-01-14, 3=1.3} |  {1=3, 2=2021-01-14, 3=1.3} |  <null>        | <null>         |   <null>       | 1                    | 1                    | <null>               | 0   | hdfs://hadoop-master:9000/user/hive/warehouse/test_table/metadata/snap-6116016324956900164-0-3c1b2496-0670-4e37-81f6.avro    | <null>       | <null>         | <null>
+  content  | file_path                                                                                                                     | record_count    | file_format   | file_size_in_bytes   |  column_sizes        |  value_counts     |  null_value_counts | nan_value_counts  | lower_bounds                |  upper_bounds               |  key_metadata  | split_offsets  |  equality_ids  | added_snapshot_id  | file_sequence_number | data_sequence_number | referenced_data_file | pos | manifest_location                                                                                                            | first_row_id | content_offset | content_size_in_bytes
+----------+-------------------------------------------------------------------------------------------------------------------------------+-----------------+---------------+----------------------+----------------------+-------------------+--------------------+-------------------+-----------------------------+-----------------------------+----------------+----------------+----------------+--------------------+----------------------+----------------------+----------------------+-----+------------------------------------------------------------------------------------------------------------------------------+--------------+----------------+----------------------
+ 0        | hdfs://hadoop-master:9000/user/hive/warehouse/test_table/data/c1=3/c2=2021-01-14/af9872b2-40f3-428f-9c87-186d2750d84e.parquet |  1              |  PARQUET      |  442                 | {1=40, 2=40, 3=44}   |  {1=1, 2=1, 3=1}  |  {1=0, 2=0, 3=0}   | <null>            |  {1=3, 2=2021-01-14, 3=1.3} |  {1=3, 2=2021-01-14, 3=1.3} |  <null>        | <null>         |   <null>       | 6116016324956900164 | 1                    | 1                    | <null>               | 0   | hdfs://hadoop-master:9000/user/hive/warehouse/test_table/metadata/snap-6116016324956900164-0-3c1b2496-0670-4e37-81f6.avro    | <null>       | <null>         | <null>
 ```
 
 The output of the query has the following columns:
@@ -1570,6 +1570,15 @@ The output of the query has the following columns:
 * - `readable_metrics`
   - `JSON`
   - File metrics in human-readable form.
+* - `added_snapshot_id`
+  - `BIGINT`
+  - The snapshot ID when the file was first added to the table, as recorded in
+    the selected snapshot's live manifest entry. This makes it possible to join
+    current live files with `$snapshots` to inspect when each file was
+    introduced. If a file is moved to a different manifest by manifest rewrite,
+    the manifest location may change, but `added_snapshot_id` still refers to
+    the snapshot in which the file was originally added. Use `$entries` or
+    `$all_entries` to inspect historical manifest references across snapshots.
 * - `file_sequence_number`
   - `BIGINT`
   - The sequence number of the file, tracking when the file was added.
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
@@ -33,6 +33,7 @@ import io.trino.spi.type.RowType;
 import io.trino.spi.type.TypeManager;
 import org.apache.iceberg.ContentFile;
 import org.apache.iceberg.DeleteFile;
+import org.apache.iceberg.IcebergManifestUtils.FileEntryWithMetadata;
 import org.apache.iceberg.ManifestReader;
 import org.apache.iceberg.MetricsUtil;
 import org.apache.iceberg.PartitionField;
@@ -66,19 +67,28 @@ import static io.trino.plugin.iceberg.IcebergTypes.convertIcebergValueToTrino;
 import static io.trino.plugin.iceberg.IcebergUtil.primitiveFieldTypes;
 import static io.trino.plugin.iceberg.IcebergUtil.readerForManifest;
 import static io.trino.plugin.iceberg.StructLikeWrapperWithFieldIdToIndex.createStructLikeWrapper;
+import static io.trino.plugin.iceberg.system.FilesTable.ADDED_SNAPSHOT_ID_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.COLUMN_SIZES_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.CONTENT_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.CONTENT_OFFSET_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.CONTENT_SIZE_IN_BYTES_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.DATA_SEQUENCE_NUMBER_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.EQUALITY_IDS_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.FILE_FORMAT_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.FILE_PATH_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.FILE_SEQUENCE_NUMBER_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.FILE_SIZE_IN_BYTES_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.FIRST_ROW_ID_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.KEY_METADATA_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.LOWER_BOUNDS_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.MANIFEST_LOCATION_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.NAN_VALUE_COUNTS_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.NULL_VALUE_COUNTS_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.PARTITION_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.POS_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.READABLE_METRICS_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.RECORD_COUNT_COLUMN_NAME;
+import static io.trino.plugin.iceberg.system.FilesTable.REFERENCED_DATA_FILE_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.SORT_ORDER_ID_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.SPEC_ID_COLUMN_NAME;
 import static io.trino.plugin.iceberg.system.FilesTable.SPLIT_OFFSETS_COLUMN_NAME;
@@ -95,6 +105,7 @@ import static io.trino.spi.type.TypeUtils.writeNativeValue;
 import static io.trino.spi.type.VarbinaryType.VARBINARY;
 import static io.trino.spi.type.VarcharType.VARCHAR;
 import static java.util.Objects.requireNonNull;
+import static org.apache.iceberg.IcebergManifestUtils.liveEntriesWithMetadata;
 import static org.apache.iceberg.MetricsUtil.readableMetricsStruct;
 
 public final class FilesTablePageSource
@@ -108,7 +119,7 @@ public final class FilesTablePageSource
     private final List<PartitionField> partitionFields;
     private final Optional<IcebergPartitionColumn> partitionColumnType;
     private final List<Types.NestedField> primitiveFields;
-    private final Iterator<? extends ContentFile<?>> contentIterator;
+    private final Iterator<FileEntryWithMetadata> entryIterator;
     private final Map<String, Integer> columnNameToIndex;
     private final PageBuilder pageBuilder;
     private final long completedBytes;
@@ -137,7 +148,7 @@ public final class FilesTablePageSource
                 .collect(toImmutableList());
         ManifestReader<? extends ContentFile<?>> manifestReader = closer.register(readerForManifest(split.manifestFile(), fileIoFactory.create(trinoFileSystem), idToPartitionSpecMapping));
         // TODO figure out why selecting the specific column causes null to be returned for offset_splits
-        this.contentIterator = closer.register(requireNonNull(manifestReader, "manifestReader is null").iterator());
+        this.entryIterator = closer.register(liveEntriesWithMetadata(requireNonNull(manifestReader, "manifestReader is null")).iterator());
         this.pageBuilder = new PageBuilder(requiredColumns.stream().map(column -> {
             if (column.equals(PARTITION_COLUMN_NAME)) {
                 return split.partitionColumnType().orElseThrow();
@@ -184,10 +195,11 @@ public final class FilesTablePageSource
             return null;
         }
 
-        while (contentIterator.hasNext() && !pageBuilder.isFull()) {
+        while (entryIterator.hasNext() && !pageBuilder.isFull()) {
             pageBuilder.declarePosition();
             long start = System.nanoTime();
-            ContentFile<?> contentFile = contentIterator.next();
+            FileEntryWithMetadata entry = entryIterator.next();
+            ContentFile<?> contentFile = entry.file();
 
             writeValueOrNull(pageBuilder, CONTENT_COLUMN_NAME, () -> contentFile.content().id(), INTEGER::writeInt);
             writeValueOrNull(pageBuilder, FILE_PATH_COLUMN_NAME, contentFile::location, VARCHAR::writeString);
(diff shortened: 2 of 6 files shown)


plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
```diff
@@ -1850,6 +1850,61 @@ private Table updateTableToV2(String tableName)
         return table;
     }
 
+    @Test
```
ebyhr: The purpose of this test class is to verify v2 spec. Move this test to BaseIcebergSystemTables.

ebyhr (commented):
> Iceberg: Add delete file metrics to $partitions metadata table

We don't use "Iceberg: " prefix in a commit title. Please follow https://trino.io/development/process#pull-request-and-commit-guidelines

`Changes: ...` in a commit body isn't so helpful. You can remove it.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergV2.java
kaveti: sure @ebyhr

kaveti:
@ebyhr i have addressed review comments. could you please review again.

findinpath:
@kaveti  is the failure related to your changes?

https://github.com/trinodb/trino/actions/runs/23729005130/job/69119458673?pr=28911

```
 2026-03-30 11:12:45 SEVERE: Failure cause:
tests               | java.lang.AssertionError: Expected row count to be <5>, but was <9>; rows=[[partition, row("a" varchar, "b" varchar)], [record_count, bigint], [file_count, bigint], [total_size, bigint], [position_delete_file_count, bigint], [position_delete_record_count, bigint], [equality_delete_file_count, bigint], [equality_delete_record_count, bigint], [data, row("a" row("min" varchar, "max" varchar, "null_count" bigint, "nan_count" bigint), "c" row("min" varchar, "max" varchar, "null_count" bigint, "nan_count" bigint))]]
tests               | 	at io.trino.tests.product.iceberg.TestIcebergPartitionEvolution.testDroppedPartitionField(TestIcebergPartitionEvolution.java:78)
tests               | 	at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:104)
tests               | 	at java.base/java.lang.reflect.Method.invoke(Method.java:565)
tests               | 	at org.testng.internal.MethodInvocationHelper.invokeMethod(MethodInvocationHelper.java:104)
tests               | 	at org.testng.internal.Invoker.invokeMethod(Invoker.java:645)
tests               | 	at org.testng.internal.Invoker.invokeTestMethod(Invoker.java:851)
tests               | 	at org.testng.internal.Invoker.invokeTestMethods(Invoker.java:1177)
tests               | 	at org.testng.internal.TestMethodWorker.invokeTestMethods(TestMethodWorker.java:129)
tests               | 	at org.testng.internal.TestMethodWorker.run(TestMethodWorker.java:112)
tests               | 	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1090)
tests               | 	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:614)
tests               | 	at java.base/java.lang.Thread.run(Thread.java:1474)
tests               | 
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -29,6 +29,7 @@
 import io.trino.spi.type.RowType;
 import io.trino.spi.type.TypeManager;
 import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DeleteFile;
```
findinpath: > Add `position_delete_file_count`, `position_delete_record_count`, `equality_delete_file_count`, and `equality_delete_record_count` columns to the Iceberg `$partitions` system table.


Could you please adjust `iceberg.md` to have the latest changes reflected in the documentation?

https://trino.io/docs/current/connector/iceberg.html#partitions-table

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -98,7 +100,9 @@ public PartitionsTable(SchemaTableName tableName, TypeManager typeManager, Table
         partitionColumnType.ifPresent(icebergPartitionColumn ->
                 columnMetadataBuilder.add(new ColumnMetadata("partition", icebergPartitionColumn.rowType())));
 
-        Stream.of("record_count", "file_count", "total_size")
+        Stream.of("record_count", "file_count", "total_size",
```
findinpath: nit: use new lines for each item in the stream

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -176,14 +180,26 @@ private Map<StructLikeWrapperWithFieldIdToIndex, IcebergStatistics> getStatistic
...
-                partitions.computeIfAbsent(
+                IcebergStatistics.Builder builder = partitions.computeIfAbsent(
                         structLikeWrapperWithFieldIdToIndex,
-                        _ -> new IcebergStatistics.Builder(icebergTable.schema().columns(), typeManager))
-                        .acceptDataFile(dataFile, fileScanTask.spec());
+                        _ -> new IcebergStatistics.Builder(icebergTable.schema().columns(), typeManager));
+                builder.acceptDataFile(dataFile, fileScanTask.spec());
+
+                // Deduplicate delete files by path since the same delete file can appear
+                // in multiple FileScanTask objects within a partition. For v3 deletion vectors
```
findinpath: > Deduplicate delete files by path since the same delete file can appear in multiple FileScanTask objects within a partition

@kaveti  could you pls create a follow-up PR for $files table as well ?
cc @chenjian2664

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -98,7 +100,9 @@ public PartitionsTable(SchemaTableName tableName, TypeManager typeManager, Table
         partitionColumnType.ifPresent(icebergPartitionColumn ->
                 columnMetadataBuilder.add(new ColumnMetadata("partition", icebergPartitionColumn.rowType())));
```
findinpath, line 100: pls add along this effort also:
- total_data_file_size_in_bytes
- last_updated_at
- last_updated_snapshot_id

https://github.com/apache/iceberg/blob/2abac79fcae94b5ad039bd09f7235be191b0761e/core/src/main/java/org/apache/iceberg/PartitionsTable.java#L51-L93

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
kaveti: https://github.com/trinodb/trino/pull/28933

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -194,6 +230,50 @@ private Map<StructLikeWrapperWithFieldIdToIndex, IcebergStatistics> getStatistic
...
+    {
+        if (snapshotId.isEmpty()) {
+            return;
+        }
+        Snapshot currentSnapshot = icebergTable.snapshot(snapshotId.getAsLong());
+        if (currentSnapshot == null) {
+            return;
+        }
+        Map<Integer, PartitionSpec> specsById = icebergTable.specs();
+        for (ManifestFile manifest : currentSnapshot.allManifests(icebergTable.io())) {
```
chenjian2664: It seems we read the manifests multiple times, 1: for data and delte files , 2: populate time and snapshot id here, is it possible to by reading the manifests only once if we have to?

kaveti: sure, will do it in single pass

findinpath:
There are product tests failures

https://github.com/trinodb/trino/actions/runs/23781856982/job/69296855765?pr=28911

You can run a product test like this:

```
testing/bin/ptl test run --environment singlenode-spark-iceberg  -- -t io.trino.tests.product.iceberg.TestIcebergPartitionEvolution.testDroppedPartitionField
```

In case you have changes, you can rebuild (fast) trino like this
```
mvnd clean install -DskipTests  -Dmaven.javadoc.skip=true -Dmaven.source.skip=true -Dair.check.skip-all=true
```

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
findinpath: gentle reminder

docs/src/main/sphinx/connector/iceberg.md
findinpath, line 1448: You need to update as well the listing

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -164,34 +183,78 @@ public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, Connect
...
+                            if (seen.add(deleteFile.location())) {
+                                builder.acceptDeleteFile(deleteFile);
+                            }
+                        }
+                    }
+
+                    Snapshot snapshot = icebergTable.snapshot(entrySnapshotId);
+                    if (snapshot != null) {
+                        builder.updateLastUpdated(entrySnapshotId, snapshot.timestampMillis());
+                    }
```
findinpath: see that you need to fill in the snapshot only if not present of the update timestamp is more recent than what you have already stored

https://github.com/apache/iceberg/blob/149cc464f9b7df800cc5718af725983473819504/core/src/main/java/org/apache/iceberg/PartitionsTable.java#L300-L305

findinpath: add as well a corresponding test

findinpath: gentle reminder

docs/src/main/sphinx/connector/iceberg.md
raunaqmorarka, line 1447: The example should be updated in the first commit as well

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -194,6 +230,50 @@ private Map<StructLikeWrapperWithFieldIdToIndex, IcebergStatistics> getStatistic
         }
     }
 
+    private void populateLastUpdated(Map<StructLikeWrapperWithFieldIdToIndex, IcebergStatistics.Builder> partitions)
```
raunaqmorarka: It is confusing to add this column in an unoptimized way and then refactor the code in the next commit.
This should be organized as :
Commit 1 : Add whatever fields can be populated without changing the structure of the code and without adding redundant reads of manifests
Commit 2: Re-organize code to support populating additional fields with single pass on manifests
Commit 3: Add remaining fields

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -45,6 +47,12 @@ public record IcebergStatistics(
         long recordCount,
         long fileCount,
         long size,
+        long positionDeleteFileCount,
+        long positionDeleteRecordCount,
+        long equalityDeleteFileCount,
+        long equalityDeleteRecordCount,
+        @Nullable Long lastUpdatedAt,
+        @Nullable Long lastUpdatedSnapshotId,
```
raunaqmorarka: These new fields don't belong here. This class is intended for use in calculating table statistics for the CBO, it just happened to line up with what was needed in PartitionsTable a well.
We should have a new class `io.trino.plugin.iceberg.system.PartitionsTableStatistics` that is composed from IcebergStatistics and plus the addition fields here, and IcebergStatistics should stay unchanged.

kaveti:
@raunaqmorarka  thank you for your review comments . i will keep in mind commit order best practices.  i have addressed your review comments please take a look again.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -174,45 +179,72 @@ public RecordCursor cursor(ConnectorTransactionHandle transactionHandle, Connect
...
+        Snapshot currentSnapshot = icebergTable.snapshot(snapshotId.getAsLong());
+        if (currentSnapshot == null) {
+            return ImmutableMap.of();
+        }
+
+        Map<StructLikeWrapperWithFieldIdToIndex, PartitionsTableStatistics.Builder> partitions = new HashMap<>();
+        Map<StructLikeWrapperWithFieldIdToIndex, Set<String>> seenDeleteFiles = new HashMap<>();
+        Map<Integer, PartitionSpec> specsById = icebergTable.specs();
+
+        for (ManifestFile manifest : currentSnapshot.allManifests(icebergTable.io())) {
```
raunaqmorarka: This is single pass, but its lacking the parallel read of manifests that planFiles does under the covers on `icebergScanExecutor`. So this is going to be much worse for latency on large tables.

raunaqmorarka: Taking a step back, we already have distributed execution for `$files`.
If whatever fields you need can be added as columns to that table, then can everything you get this metadata table just be computed as a query on top of that ?

chenjian2664: The partitions table has been around for some time, it would be worthwhile to add commonly used fields for convenience.

raunaqmorarka: I think the direction proposed in https://github.com/trinodb/trino/pull/28997 is better. We shouldn't have to keep improving these metadata table implementations in isolation if they can be expressed as a query on top of `$files`

kaveti: Agreed that #28997 is a good long-term direction. Today though, $files doesn't expose snapshot IDs or timestamps (needed for last_updated_at), and column-level min/max can't be aggregated in SQL since they're binary-encoded maps. Once $files gains those columns and SystemView lands, $partitions can be migrated.

kaveti: having these fields pre-aggregated in $partitions avoids users needing to run expensive aggregation queries over $files every time they want to check partition health

kaveti: done

raunaqmorarka: I don't understand what is expensive about `$files` compared to `$partitions`. They both read the same manifest files, $partitions is in fact worse because it does not support distributed execution the way $files does.
#28997 is not some long term direction, we can implement that in the short term.

kaveti: i see. i believe this PR can be closed then. let me know if you see any thing from this PR makes  sense. i think view can capture all the info present in this PR.

raunaqmorarka: While the new way is being worked on, you can add the fields you need to `$files` in case they're missing from there

kaveti: @raunaqmorarka i have changed pr context and added the fields req for $files only

plugin/trino-iceberg/src/main/java/org/apache/iceberg/IcebergManifestUtils.java
```diff
@@ -34,4 +34,14 @@ public static <F extends ContentFile<F>> CloseableIterable<ContentFile<F>> liveE
...
         return CloseableIterable.transform(manifestReader.liveEntries(), ManifestEntry::file);
     }
+
+    public static <F extends ContentFile<F>> CloseableIterable<FileEntryWithMetadata> liveEntriesWithMetadata(ManifestReader<F> manifestReader)
+    {
+        return CloseableIterable.transform(manifestReader.liveEntries(),
+                entry -> new FileEntryWithMetadata(entry.file(), entry.snapshotId()));
+    }
+
+    public record FileEntryWithMetadata(
```
chenjian2664: nit: the code can be fix into single line

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1570,6 +1570,9 @@ The output of the query has the following columns:
 * - `readable_metrics`
   - `JSON`
   - File metrics in human-readable form.
+* - `added_snapshot_id`
```
chenjian2664, line 1573: This is derived from the manifest metadata. Could you clarify the scenario? 
I'm trying to understand the semantics at the file level, if a file is referenced by multiple manifests, would this value differ?

kaveti: I have updated the doc for `added_snapshot_id`

@chenjian2664 
$files already tells you what files are live now, but without added_snapshot_id it cannot tell you which commit introduced each live file. With this column, you can answer things like:

“Which snapshot added this current file?”
“When was this current file introduced?” by joining to $snapshots
“Which current files came from snapshot X?”
“When was this partition last written?” by combining $files with partition columns and $snapshots

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1570,6 +1570,15 @@ The output of the query has the following columns:
...
+  - `BIGINT`
+  - ID of the snapshot that added this file, as recorded in the selected
```
chenjian2664:
```suggestion
  - The snapshot ID when the file was first added to the table, as recorded in the selected
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergSystemTables.java
```diff
@@ -473,6 +474,21 @@ public void testFilesPartitionTable()
                         .build());
     }
 
+    @Test
+    public void testFilesTableAddedSnapshotId()
+    {
+        assertQuery(
```
chenjian2664: use `assertThat`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
```diff
@@ -137,7 +148,7 @@ public FilesTablePageSource(
                 .collect(toImmutableList());
         ManifestReader<? extends ContentFile<?>> manifestReader = closer.register(readerForManifest(split.manifestFile(), fileIoFactory.create(trinoFileSystem), idToPartitionSpecMapping));
         // TODO figure out why selecting the specific column causes null to be returned for offset_splits
-        this.contentIterator = closer.register(requireNonNull(manifestReader, "manifestReader is null").iterator());
+        this.entryIterator = closer.register(IcebergManifestUtils.liveEntriesWithMetadata(requireNonNull(manifestReader, "manifestReader is null")).iterator());
```
chenjian2664: static import

chenjian2664 (approved):

kaveti:
@chenjian2664 i have addressed your comments. thank you

plugin/trino-iceberg/src/main/java/org/apache/iceberg/IcebergManifestUtils.java
chenjian2664: I am not sure why I marked it as resolved, please fix it @kaveti

kaveti: done

ebyhr:
```
## Section
* Fixes https://github.com/trinodb/trino/issues/28910
```

Please follow https://trino.io/development/process#release-note-guidelines

PR #28913, opened 2026-03-30
Remove unused write type from `DeltaLakeTableHandle`

chenjian2664:
## Description

The write type `UPDATE` and `DELETE` are not used anywhere

## Release notes

```markdown
## Section
```


ebyhr (approved):
Thanks, I was thinking the same fix in my previous PR, but I was lazy.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeTableHandle.java
```diff
@@ -41,8 +38,6 @@ public class DeltaLakeTableHandle
     public enum WriteType
     {
         MERGE,
-        UPDATE,
-        DELETE
```
findinpath, line 45: What is the purpose of having an enum with one value?

Consider using a boolean field with a meaningful name instead

wendigo (approved):

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeTableHandle.java
chenjian2664: Added  eafc80b

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeTableHandle.java
```diff
@@ -53,15 +41,11 @@ public enum WriteType
     private final ProtocolEntry protocolEntry;
     private final TupleDomain<DeltaLakeColumnHandle> enforcedPartitionConstraint;
     private final TupleDomain<DeltaLakeColumnHandle> nonPartitionConstraint;
-    private final Optional<WriteType> writeType;
+    private final boolean merge;
```
findinpath, line 44: Looking at `isOptimize` which is a flag field same as this one - I see a small inconsistency in naming.

Pls consider `isMerge`

chenjian2664: See `managed` and `timeTravel` in the class :)

Issue #28927, opened 2026-03-30
Cancelled Trino queries keep running on Elasticsearch


lovasoa:
### Trino version

481-SNAPSHOT (master)

### Please describe the bug

Hi ! 

I was investigating the es connector performance characteristics and running benchmarks, and I stumbled on this issue. It seems that expansive queries keep running on elasticsearch after they are cancelled on the trino side.  When Trino cancels a query (timeout or user cancellation), the HTTP connection is never closed, so ES gets no signal to stop.

### Repro

```
./mvnw test -pl plugin/trino-elasticsearch -Dtest=TestElasticsearchQueryTimeout -Dair.check.skip-all=true
```

The test runs an intentionally slow `script_score` query with `query_max_execution_time = 100ms`, waits 5 seconds after Trino returns the timeout error, then asserts that `GET _tasks?actions=*search*` has no non-cancelled search tasks. It currently fails: the tasks are still running with `cancelled: false`.

<details>
<summary>
<code>TestElasticsearchQueryTimeout.java</code>
</summary>

```java
package io.trino.plugin.elasticsearch;

import com.fasterxml.jackson.databind.JsonNode;
import com.fasterxml.jackson.databind.json.JsonMapper;
import io.airlift.json.JsonMapperProvider;
import io.trino.testing.AbstractTestQueryFramework;
import io.trino.testing.QueryRunner;
import org.elasticsearch.client.Request;
import org.elasticsearch.client.RestHighLevelClient;
import org.junit.jupiter.api.Test;

import java.io.IOException;
import java.util.ArrayList;
import java.util.Iterator;
import java.util.List;

import static io.trino.plugin.elasticsearch.ElasticsearchServer.ELASTICSEARCH_8_IMAGE;
import static io.trino.testing.TestingSession.testSessionBuilder;
import static org.assertj.core.api.Assertions.assertThat;

final class TestElasticsearchQueryTimeout
        extends AbstractTestQueryFramework
{
    private static final JsonMapper JSON = new JsonMapperProvider().get();

    private RestHighLevelClient esClient;

    @Override
    protected QueryRunner createQueryRunner()
            throws Exception
    {
        ElasticsearchServer elasticsearch = closeAfterClass(new ElasticsearchServer(ELASTICSEARCH_8_IMAGE));
        esClient = closeAfterClass(elasticsearch.getClient());
        return ElasticsearchQueryRunner.builder(elasticsearch).build();
    }

    @Test
[...]

PR #28935, opened 2026-03-30
ElasticSearch: Add timeout parameter to _search

lovasoa:
## Description

Elasticsearch _search requests now carry the configured request timeout, and Trino fails the query when Elasticsearch reports that the search timed out.

This makes Elasticsearch aware of the timeout, so that it cancels the query on its side even if the Trino node is too slow and does not cut the HTTP connection on time.

Related to #28927.

## Release notes

Release notes are required, with the following suggested text:

```markdown
## fixes
* elasticsearch connector: propagate query timeout information to elasticsearch ({issue}`28927`)
```

### DIFF
--- a/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
+++ b/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
@@ -122,6 +122,7 @@ public class ElasticsearchClient
     private final BackpressureRestHighLevelClient client;
     private final int scrollSize;
     private final Duration scrollTimeout;
+    private final Duration requestTimeout;
 
     private final AtomicReference<Set<ElasticsearchNode>> nodes = new AtomicReference<>(ImmutableSet.of());
     private final ScheduledExecutorService executor = newSingleThreadScheduledExecutor(daemonThreadsNamed("NodeRefresher"));
@@ -146,6 +147,7 @@ public class ElasticsearchClient
         this.ignorePublishAddress = config.isIgnorePublishAddress();
         this.scrollSize = config.getScrollSize();
         this.scrollTimeout = config.getScrollTimeout();
+        this.requestTimeout = config.getRequestTimeout();
         this.refreshInterval = config.getNodeRefreshInterval();
         this.tlsEnabled = config.isTlsEnabled();
     }
@@ -580,7 +582,8 @@ public class ElasticsearchClient
     public SearchResponse beginSearch(String index, int shard, QueryBuilder query, Optional<List<String>> fields, List<String> documentFields, Optional<String> sort, OptionalLong limit)
     {
         SearchSourceBuilder sourceBuilder = SearchSourceBuilder.searchSource()
-                .query(query);
+                .query(query)
+                .timeout(new TimeValue(requestTimeout.toMillis()));
 
         if (limit.isPresent() && limit.getAsLong() < scrollSize) {
             // Safe to cast it to int because scrollSize is int.
@@ -612,7 +615,11 @@ public class ElasticsearchClient
 
         long start = System.nanoTime();
         try {
-            return client.search(request);
+            SearchResponse response = client.search(request);
+            if (response.isTimedOut()) {
+                throw new TrinoException(ELASTICSEARCH_CONNECTION_ERROR, "Elasticsearch query timed out");
+            }
+            return response;
         }
         catch (IOException e) {
             throw new TrinoException(ELASTICSEARCH_CONNECTION_ERROR, e);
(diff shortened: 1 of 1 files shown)


lovasoa:
I signed the cla and sent it to cla@trino.io.

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
```diff
@@ -612,7 +616,11 @@ public SearchResponse beginSearch(String index, int shard, QueryBuilder query, O
 
         long start = System.nanoTime();
         try {
-            return client.search(request);
+            SearchResponse response = client.search(request);
+            if (response.isTimedOut()) {
```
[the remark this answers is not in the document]
lovasoa: @codex This is a good remark, but I think it's a separate issue. Whether we want to clear scrolls and when is not specific to the timeout case.

If your human masters want me to specifically add a clearScroll call before the throw here I'll do it though.

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
```diff
@@ -582,6 +584,8 @@ public SearchResponse beginSearch(String index, int shard, QueryBuilder query, O
         SearchSourceBuilder sourceBuilder = SearchSourceBuilder.searchSource()
                 .query(query);
 
+        sourceBuilder.timeout(new TimeValue(requestTimeout.toMillis(), MILLISECONDS));
```
ebyhr: `sourceBuilder` is a builder. There is no need to separate the call. 

Also, the MILLISECONDS is redundant. We can call the constructor without the parameter, or `timeValueMillis` helper method instead. 
```java
        SearchSourceBuilder sourceBuilder = SearchSourceBuilder.searchSource()
                .query(query)
                .timeout(timeValueMillis(requestTimeout.toMillis()));
```

ebyhr: We set scroll timeout at L610. Can we move this logic there for the consistency? 

https://github.com/trinodb/trino/blob/a4e2d1244548717d730c5966900b4e809e8dbff3/plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java#L607-L611

ebyhr:
> Add ES timeout to _search

Could you change the commit to "Add timeout to _search in Elasticsearch"?

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
Praveen2112: Additionally do we need to configure them for count queries as well ?

Praveen2112: I was looking at ES documentation, setting this timeout doesn't guarantee a query to be terminated right ?

Praveen2112:
@lovasoa But even setting request timeout wouldn't help us much right ? It might return partial result and the queries might still be executing. 
In case of https://github.com/elastic/elasticsearch/issues/47716 

> We have a new way to cancel search requests efficiently from the client in 7.4 (by closing the underlying http channel) and we expose this in our official Java client so we don't need this option on the server if the goal is to trash any requests that take more than a configured threshold

> Wasn't this targeting e-commerce scenarios? Shoppers have a limited attention span and there's a desire to serve something within a given time window.

I think even setting this property might not help us.

lovasoa:
This pr is related to, but does not fix #28927.

This pr prevents queries from executing after the *http timeout* configured in the es plugin settings when trino does not close the connection quick enough (because the node is overloaded for instance). 

We still need a separate fix to handle trino-side cancellations (the issue in #28927)

plugin/trino-elasticsearch/src/main/java/io/trino/plugin/elasticsearch/client/ElasticsearchClient.java
lovasoa: The scroll timeout is set on SearchRequest, but we want to set the timeout on the underlying SearchSourceBuilder

lovasoa: I moved the timeout to the builder chain, and used `new TimeValue(requestTimeout.toMillis())` to match how the scroll timeout is defined

lovasoa:
@cla-bot : no need to repost your message on every commit. I told you I sent the signed cla already.

Praveen2112:
> This pr prevents queries from executing after the http timeout configured in the es plugin settings when trino does not close the connection quick enough (because the node is overloaded for instance).

But setting this property doesn't ensure the query is cancelled right ? I'm trying to understand what we would benefit here ? If our primary goal is support partial result then yes, if not doesn't closing the channel helps us here.

lovasoa:
Closing the channel does the job. But the person responsible for the elasticsearch cluster wants a guarantee that the query will not keep using resources on the elasticsearch side **whatever happens on the client side**. If the trino node is overloaded, it may take several seconds to actually close the connection (and thus for the cancellation to propagate to elasticsearch). This pr does not remove the HTTP timeout, but it propagates the timeout information to elasticsearch, so that the query is always cancelled after the timeout on the ES side, whatever happens on the trino side.

lovasoa:
I opened a separate PR to fix https://github.com/trinodb/trino/issues/28927 : https://github.com/trinodb/trino/pull/28941

martint:
@cla-bot check

PR #28937, opened 2026-03-31
Add NEAREST join support

### LINKED ISSUE #21759: Add support for ASOF join

martint:
Add support for NEAREST in CROSS JOIN and LEFT JOIN ... ON TRUE.

For example:

```sql
...
FROM trades
LEFT JOIN NEAREST (
    FROM quotes
    WHERE quotes.symbol = trades.symbol
    MATCH quotes.ts <= trades.ts
) ON TRUE
```

and 

```sql
...
FROM trades
CROSS JOIN NEAREST (
    FROM quotes
    WHERE quotes.symbol = trades.symbol
    MATCH quotes.ts <= trades.ts
)
```

The current implementation produces a naive plan that leverages existing plan IR  and operators. A more performant implementation will require custom plan nodes and  a dedicated join-like operator.

CROSS JOIN plan shape:

    - Project[left columns..., nearest columns...]
       - TopNRanking[ROW_NUMBER by (left_unique) order by (candidate_key)] 
           - Join[INNER on (WHERE and MATCH)] 
               - AssignUniqueId[left_unique] 
                   - left input 
               - NEAREST source

LEFT JOIN ... ON TRUE plan shape:

    - Project[left columns..., nearest columns...]
       - TopNRanking[ROW_NUMBER by (left_unique) order by (candidate_key)] 
           - Join[LEFT on (WHERE and MATCH)] 
               - AssignUniqueId[left_unique] 
                   - left input 
               - NEAREST source

- Fixes #21759

## Release notes

```markdown
## General
* Add support for the `NEAREST` clause for approximate matches in joins. ({issue}`21759`)
```


core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
```diff
@@ -1231,6 +1258,92 @@ private static Optional<Lateral> getLateral(Relation relation)
         return Optional.empty();
     }
 
+    private RelationPlan planNearestJoin(Join join, RelationPlan leftPlan, Nearest nearest)
```
Praveen2112: This can also be designed as a Optimizer rule right ?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4066,6 +4141,43 @@ private boolean isJsonTable(Relation node)
...
+                return isNearestRelation(aliasedRelation.getRelation());
+            }
+            return node instanceof Nearest;
+        }
+
+        private NearestAnalysis analyzeNearestMatch(Nearest node, Scope nearestScope, Scope leftScope)
+        {
+            if (!(node.getMatch() instanceof ComparisonExpression comparison)) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must be a comparison expression");
+            }
```
Praveen2112, line 4174: Is having a match expression like `colA > colB AND colC > colD` is also a valid syntax in the future ?

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
martint: It can but not in this PR. This is meant to be the minimally possible implementation.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
martint: Not likely. The challenge is that the condition needs to be a total order. For example, given (2,2) on the left side, and (2,1) vs (1,2) on the right side, which one is nearest?

Praveen2112: We can provide an ordering scheme right ?

martint: The ordering scheme is implied by the match expression. Given `left.a > right.a`, it says "given a row in `left`, find the row in `right` with the closest value that's smaller than `left.a`"

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
```diff
@@ -1231,6 +1258,92 @@ private static Optional<Lateral> getLateral(Relation relation)
...
+        RelationPlan candidatePlan = new RelationPlan(candidateRoot, analysis.getScope(nearest), candidateOutputs, outerContext);
+
+        NearestAnalysis nearestAnalysis = analysis.getNearest(nearest);
+        PlanBuilder candidateBuilder = newPlanBuilder(candidatePlan, analysis, lambdaDeclarationToSymbolMap, session, plannerContext)
+                .appendProjections(ImmutableList.of(nearestAnalysis.candidateExpression()), symbolAllocator, idAllocator);
+
+        Symbol orderingSymbol = candidateBuilder.translate(nearestAnalysis.candidateExpression());
+        SortOrder sortOrder = switch (nearestAnalysis.operator()) {
+            case LESS_THAN, LESS_THAN_OR_EQUAL -> SortOrder.DESC_NULLS_LAST;
+            case GREATER_THAN, GREATER_THAN_OR_EQUAL -> SortOrder.ASC_NULLS_LAST;
```
Praveen2112, line 1338: In case of NaN and Infinity, -Infinity, do we need any special handling for the NEAREST join support ?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
Praveen2112: > The challenge is that the condition needs to be a total order.

In case of single comparison expression we could infer from the match expression. If we support something like colA > colB AND colC > colD - we could take in an Ordering scheme to find the nearest expression.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
martint: They will follow normal double/real comparison semantics.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
martint: Not sure I understand what you're suggesting. Can you provide a concrete example?

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
Praveen2112: Should we fail early if we try to find a nearest column on a non-comparable type ? I assume it would fail if there isn't any resolved function for the same. But would it make sense to apply this check here ?

martint: It's already handled by the analyzer. See https://github.com/trinodb/trino/pull/28937/changes/75fa72be5ddb553faf9670c1c70614bfd8b58b39#diff-f7e345d57c9a04d4f8c2409e09245adca19a7fe8a20b3255bbae7a97dc0f4a9dR1746

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
Praveen2112: I was thinking if we need any query patterns 

FROM trades
LEFT JOIN NEAREST (
    FROM quotes
    WHERE quotes.symbol = trades.symbol
    MATCH quotes.ts <= trades.ts AND quotes.col <= trades.col 
    ORDER BY candidate_key_1, candidate_key_2
) ON TRUE


in this format.

martint: Thanks. That gets dangerously close to the general-purpose LATERAL query syntax for which NEAREST is syntactic sugar:

```sql
SELECT *
FROM trades
LEFT JOIN LATERAL (
    SELECT *
    FROM quotes
    WHERE quotes.symbol = trades.symbol
      AND quotes.ts <= trades.ts AND quotes.col <= trades.col
    ORDER BY candidate_key_1, candidate_key_2
    FETCH FIRST 1 ROW ONLY
) ON TRUE
```

Once you get to that point, you might as well use this form directly.

findepi:
This is awesome proof of concept! I am not sure the syntax is ideal from user perspective yet. I have posted syntax considerations to the issue:

- https://github.com/trinodb/trino/issues/21759#issuecomment-4175362019

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -1702,6 +1706,57 @@ protected Scope visitLateral(Lateral node, Optional<Scope> scope)
             return createAndAssignScope(node, scope, queryScope.getRelationType());
         }
 
+        @Override
+        protected Scope visitNearest(Nearest node, Optional<Scope> scope)
+        {
+            if (scope.isEmpty()) {
+                throw semanticException(NOT_SUPPORTED, node, "NEAREST is only supported on the right side of CROSS JOIN or LEFT JOIN");
```
kasiafi: Can we also support IMPLICIT join and INNER ... ON TRUE?

nearest-join-plan.md
kasiafi, line 1: This isn't meant to be merged, is it?

Also, we need a documentation section for `NEAREST`.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -1702,6 +1706,57 @@ protected Scope visitLateral(Lateral node, Optional<Scope> scope)
...
+            }
+
+            Scope leftScope = scope.orElseThrow();
+
+            // NEAREST is treated as a lateral relation by visitJoin(), but in the current implementation only the
+            // top-level WHERE and MATCH clauses may correlate with the left side of the join.
+            // The inner FROM relation is analyzed against the query boundary so it behaves like an
+            // uncorrelated relation body.
+            //
+            // If NEAREST is extended to allow correlation inside FROM <relation>, analyze the
```
kasiafi: add `TODO`.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -3429,13 +3484,19 @@ protected Scope visitJoin(Join node, Optional<Scope> scope)
         {
             JoinCriteria criteria = node.getCriteria().orElse(null);
             if (criteria instanceof NaturalJoin) {
+                if (isNearestRelation(node.getRight())) {
+                    throw semanticException(NOT_SUPPORTED, node, "NATURAL JOIN involving NEAREST is not supported");
+                }
```
kasiafi: I would not special case for NEAREST here. Natural join is generally unsupported, and when we support it, we will have to revise all lateral right sides, not only NEAREST.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -3429,13 +3484,19 @@ protected Scope visitJoin(Join node, Optional<Scope> scope)
...
                 throw semanticException(NOT_SUPPORTED, node, "Natural join not supported");
             }
 
             Scope left = process(node.getLeft(), scope);
             Scope right = process(node.getRight(), isLateralRelation(node.getRight()) ? Optional.of(left) : scope);
 
             if (isLateralRelation(node.getRight())) {
+                if (isNearestRelation(node.getRight()) && node.getType() != Join.Type.CROSS && node.getType() != LEFT) {
+                    throw semanticException(NOT_SUPPORTED, node, "%s JOIN involving NEAREST is not supported", node.getType().name());
+                }
```
kasiafi: The NEAREST case is handled in two places in this block. Can we make it one?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4066,6 +4141,43 @@ private boolean isJsonTable(Relation node)
...
+            return node instanceof Nearest;
+        }
+
+        private NearestAnalysis analyzeNearestMatch(Nearest node, Scope nearestScope, Scope leftScope)
+        {
+            if (!(node.getMatch() instanceof ComparisonExpression comparison)) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must be a comparison expression");
+            }
+
+            boolean leftReferencesSource = hasReferencesToScope(comparison.getLeft(), analysis, nearestScope);
```
kasiafi: Maybe add a comment that this checks whether the expression references FROM.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4066,6 +4141,43 @@ private boolean isJsonTable(Relation node)
...
+            boolean rightReferencesSource = hasReferencesToScope(comparison.getRight(), analysis, nearestScope);
+            if (leftReferencesSource == rightReferencesSource) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must compare one source expression with one non-source expression");
```
kasiafi: We could revise the comments and error messages for consistency: some talk about "source", and some about "FROM". I think that "FROM relation" everywhere would make them easier to understand.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4066,6 +4141,43 @@ private boolean isJsonTable(Relation node)
...
+            }
+
+            Expression candidateExpression = leftReferencesSource ? comparison.getLeft() : comparison.getRight();
+            Expression outerExpression = leftReferencesSource ? comparison.getRight() : comparison.getLeft();
+            if (hasReferencesToScope(candidateExpression, analysis, leftScope) || hasReferencesToScope(outerExpression, analysis, nearestScope)) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must keep source and non-source expressions on opposite sides");
+            }
```
kasiafi: This is easier to express with four booleans: `leftReferencesSource`, `rightReferencesSource`, `leftReferencesOuter`, `rightReferencesOuter`.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
kasiafi: rename to `planJoinNearest`.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
```diff
@@ -1231,6 +1258,92 @@ private static Optional<Lateral> getLateral(Relation relation)
...
+                rightPlanBuilder.getRoot().getOutputSymbols(),
+                false,
+                Optional.of(IrUtils.and(predicates.stream()
+                        .map(expression -> coerceIfNecessary(analysis, expression, candidateTranslations.rewrite(expression)))
+                        .collect(toImmutableList()))),
+                Optional.empty(),
+                Optional.empty(),
+                ImmutableMap.of(),
+                Optional.empty());
+        RelationPlan candidatePlan = new RelationPlan(candidateRoot, analysis.getScope(nearest), candidateOutputs, outerContext);
```
kasiafi, line 1329: I don't get how it works: we're compiling the scope of nearest source relation `analysis.getScope(nearest)` with joined outputs from outer and source relations `candidateOutputs`...

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
```diff
@@ -1231,6 +1258,92 @@ private static Optional<Lateral> getLateral(Relation relation)
...
+        PlanNode candidateRoot = new JoinNode(
+                idAllocator.getNextId(),
+                join.getType() == Join.Type.LEFT ? JoinType.LEFT : JoinType.INNER,
+                leftPlanBuilder.getRoot(),
+                rightPlanBuilder.getRoot(),
+                ImmutableList.of(),
+                leftPlanBuilder.getRoot().getOutputSymbols(),
+                rightPlanBuilder.getRoot().getOutputSymbols(),
+                false,
+                Optional.of(IrUtils.and(predicates.stream()
```
kasiafi, line 1322: `handleSubqueries()` for predicates is missing.

core/trino-main/src/test/java/io/trino/sql/analyzer/TestAnalyzer.java
```diff
@@ -4360,6 +4360,156 @@ public void testJoinLateral()
                 .hasMessage("line 1:63: FULL JOIN involving LATERAL relation is only supported with condition ON TRUE");
     }
 
+    @Test
+    public void testJoinNearest()
```
kasiafi, line 4365: test:
- implicit join
- NEAREST not in join
- WHERE / MATCH not boolean
- WHERE / MATCH involving window or grouping
- WHERE / MATCH involving a subquery
- unsupported comparison operators

kasiafi (approved):

nearest-join-plan.md
martint: Right, that wasn't supposed to be attached to the PR.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
martint: `analysis.getScope(nearest)` is intentional. Unlike a normal join predicate, WHERE and MATCH are analyzed in the NEAREST scope, not the join scope. That scope has the left join input as its local parent and the NEAREST FROM relation as its own relation, so its local field layout is [left fields..., nearest fields...]. That is why it works. 

Having said that, it's not ideal. I'm working on a separate (future) change to add support for the equivalent LATERAL (...) form, so this planning code will eventually reduce to a desugaring to that shape.

martint:
@kasiafi, updated.

Also, added support for implicit ON TRUE.

findepi:
What is implicit ON TRUE? is it only for NEAREST or all the joins?

martint:
For all joins. It’s a similar idea to how the FROM clause is optional.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
```diff
@@ -482,6 +481,10 @@ relationPrimary
         )?
         ((ERROR | EMPTY) ON ERROR)?
       ')'                                                             #jsonTable
+    | NEAREST '('
+        FROM relation
+        (WHERE where=booleanExpression)?
+        MATCH match=booleanExpression ')'                             #nearest
```
findepi, line 487: Why not `NEAREST (relation) MATCH match`?

This would make for a simpler grammar that's more familiar to users. Otherwise there will be questions why I cannot SELECT before the FROM or, why i can have WHERE but not HAVING, etc.

martint: You *need* a WHERE clause to link the left and right sides of the join. I don't think that's more familiar to users. NEAREST is nothing but a simplified (syntactic sugar for) form of:

```sql
LATERAL (
    SELECT *
    FROM <relation>
    WHERE <condition> AND left.match_column <op> right.match_column
    ORDER BY <match column>
    LIMIT 1
)
```

If users want to do something more complicated, they can always use the explicit form.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -3428,6 +3489,9 @@ protected Scope visitJoin(Join node, Optional<Scope> scope)
             if (criteria instanceof NaturalJoin) {
                 throw semanticException(NOT_SUPPORTED, node, "Natural join not supported");
             }
+            if (isNearestRelation(node.getRight())) {
+                validateNearestJoin(node, criteria);
+            }
```
kasiafi: Why is it handled outside the if block
`if (isLateralRelation(node.getRight())) {`
?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -3440,32 +3504,28 @@ protected Scope visitJoin(Join node, Optional<Scope> scope)
...
+                        throw semanticException(
+                                NOT_SUPPORTED,
+                                getJoinErrorLocation(node, criteria),
+                                "%s JOIN involving JSON_TABLE is only supported with condition ON TRUE",
+                                node.getType().name());
                     }
                 }
-                else if (node.getType() == FULL) {
-                    if (!(criteria instanceof JoinOn joinOn) || !joinOn.getExpression().equals(TRUE_LITERAL)) {
+                else if (!isNearestRelation(node.getRight()) && node.getType() == FULL) {
```
kasiafi: It's a pity we can't check if it's a LATERAL with `isLateralRelation`, which would be in symmetry with other cases (NEAREST, UNNEST, JSON_TABLE).
For improved readability, can we inline NEAREST handling here after JSON_TABLE, and then handle LATERAL as
```
else {
  // assert that it's LATERAL
  ...
}
```

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4063,6 +4128,78 @@ private boolean isJsonTable(Relation node)
...
+            return join;
+        }
+
+        private boolean isNearestRelation(Relation node)
+        {
+            if (node instanceof AliasedRelation aliasedRelation) {
+                return isNearestRelation(aliasedRelation.getRelation());
+            }
+            return node instanceof Nearest;
+        }
```
kasiafi, line 4146: Put this method next to  `isJsonTable`.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4063,6 +4128,78 @@ private boolean isJsonTable(Relation node)
...
+
+        private void validateNearestJoin(Join node, JoinCriteria criteria)
+        {
+            if (node.getType() != Join.Type.CROSS && node.getType() != LEFT) {
+                throw semanticException(NOT_SUPPORTED, node, "%s JOIN involving NEAREST is not supported", node.getType().name());
```
kasiafi: IMPLICIT join won't pass here.

kasiafi:
>Also, added support for implicit ON TRUE.

Is there implicit ON TRUE? I think that implicit join takes no criteria.
Also, we could support INNER JOIN ON TRUE.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
kasiafi: Neither will INNER on true.

findepi:
> For all joins. It’s a similar idea to how the FROM clause is optional.

Snowflake's JOIN has optional ON clause, and I remember I had some problems with grammar ambiguities (which JOIN does this ON pertain to?) when working on antlr grammar for Snowflake. I don't remember the specifics though.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
findepi: > You _need_ a WHERE clause to link the left and right sides of the join.

If we're thinking about NEAREST as a special case of a join, we could use a syntax that describes that.

if we're thinking about `NEAREST MATCH` as a way to find the best match, then WHERE clause is no special.
The lateral subquery could be just values, some aggregation or something else. The MATCH is special, the WHERE is not. I don't see why grammar would make WHERE special.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4063,6 +4128,78 @@ private boolean isJsonTable(Relation node)
...
+            if (leftReferencesFromRelation == rightReferencesFromRelation) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must compare one FROM relation expression with one non-FROM expression");
+            }
+
+            if ((leftReferencesFromRelation && leftReferencesOuterRelation) || (rightReferencesFromRelation && rightReferencesOuterRelation)) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must keep FROM relation and non-FROM expressions on opposite sides");
+            }
+
+            Expression candidateExpression = leftReferencesFromRelation ? comparison.getLeft() : comparison.getRight();
+            Expression outerExpression = leftReferencesFromRelation ? comparison.getRight() : comparison.getLeft();
```
kasiafi: We don't require a reference to the outer relation. Is it intentional?

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
```diff
@@ -1231,6 +1259,108 @@ private static Optional<Lateral> getLateral(Relation relation)
...
+        PlanBuilder leftPlanBuilder = newPlanBuilder(leftPlanWithId, analysis, lambdaDeclarationToSymbolMap, session, plannerContext);
+        PlanBuilder rightPlanBuilder = newPlanBuilder(rightPlan, analysis, lambdaDeclarationToSymbolMap, session, plannerContext);
+        Analysis.SubqueryAnalysis subqueries = analysis.getSubqueries(nearest);
+        for (io.trino.sql.tree.Expression predicate : predicates) {
+            Set<QualifiedName> dependencies = NamesExtractor.extractNamesNoSubqueries(predicate, analysis.getColumnReferences());
+            if (dependencies.stream().allMatch(leftPlan.getScope().getRelationType()::canResolve)) {
+                leftPlanBuilder = subqueryPlanner.handleSubqueries(leftPlanBuilder, predicate, subqueries);
+            }
+            else {
+                // For mixed predicates, only uncorrelated subqueries are expected here. Plan them on one side
```
kasiafi: Why? Can't WHERE involve a correlated subquery?

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -4063,6 +4128,78 @@ private boolean isJsonTable(Relation node)
...
+
+            ComparisonExpression.Operator operator = leftReferencesFromRelation ? comparison.getOperator() : comparison.getOperator().flip();
+            if (operator != ComparisonExpression.Operator.LESS_THAN &&
+                    operator != ComparisonExpression.Operator.LESS_THAN_OR_EQUAL &&
+                    operator != ComparisonExpression.Operator.GREATER_THAN &&
+                    operator != ComparisonExpression.Operator.GREATER_THAN_OR_EQUAL) {
+                throw semanticException(NOT_SUPPORTED, node.getMatch(), "NEAREST MATCH clause must use <, <=, >, or >=");
+            }
+
+            return new NearestAnalysis(operator, candidateExpression, outerExpression);
```
kasiafi: I think that we don't need to record `outerExpression`.

core/trino-grammar/src/main/antlr4/io/trino/grammar/sql/SqlBase.g4
martint: NEAREST is not a special kind of join. It is a table-like relation, similar in spirit to UNNEST or JSON_TABLE, whose job is to return the row from its input that is closest to some anchor expression from the left side. In that model, MATCH is special because it defines what "closest" means. WHERE is not special, but it's what's needed for the function to constrain the candidate set based on some row on the left side of the join.

martint:
> Is there implicit ON TRUE? I think that implicit join takes no criteria.
Also, we could support INNER JOIN ON TRUE.

Right, what I meant by implicit was "implicit condition". I'll add support for implicit joins (i.e., `SELECT ... FROM t, NEAREST(...)`)

martint:
> ... some problems with grammar ambiguities ...

Let me look into that. In the meantime, I'll back out that extension. We can re-introduce it later to improve usability.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
martint: Trying to keep things simpler for now. Handling a correlated query is more complicated in the current planning strategy for NEAREST. Once we desugar into the LATERAL (SELECT ... FROM ... WHERE ... ORDER BY ... LIMIT ...) form, we can add that functionality more naturally.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
martint: Right. Left-over from refactoring.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
martint: Yes, that is intentional. MATCH requires one side to come from the FROM relation, but the other side is only required to be non-FROM. It could be a constant, for instance. I'll add some tests for that scenario.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
martint: Good catch. I'm adding support for implicit joins and inner join.

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
martint: No special reason. Fixing.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
kasiafi: I understand the intention, but I think that such a correlation is currently allowed by the Analyzer.

core/trino-main/src/test/java/io/trino/sql/planner/TestLogicalPlanner.java
```diff
@@ -952,6 +952,116 @@ public void testCorrelatedJoinWithTopN()
                                         anyTree(tableScan("nation", ImmutableMap.of("nation_name", "name", "nation_regionkey", "regionkey")))))))));
     }
 
+    @Test
+    public void testCrossJoinNearest()
```
kasiafi: Neither of these tests asserts the full plan shape. I suggest limiting it to two test cases, for example 
- forward + LEFT
- backward + INNER,

and assert the full plan including AssignUniqueId, Join  etc... up to the pruning project on top

kasiafi (approved):
LGTM % LogicalPlanner test and the question about correlation in WHERE.

core/trino-main/src/main/java/io/trino/sql/planner/RelationPlanner.java
martint: Added checks in the analyzer

PR #28963, opened 2026-04-01
Add functions to convert IPADDRESS to 128 bits varchar

gertjanal:
## Description
Add function `ip_to_bits` and `ip_from_bits` to convert an ip address to its 128 bit form and vice versa. These functions can be used to group ip addresses by subnet.
Add function `ip_version(ipaddress)` to get either 4 for ipv4 and 6 for ipv6

```sql
SELECT ip_from_bits('0100111111111110001010010000000001010101010...000010110011111001111');
-- 4ffe:2900:5545:3210:2000:f8ff:fe21:67cf
```

```sql
SELECT ip_from_bits('11000000101011000000000100010100');
-- 192.172.1.20
```

```sql
SELECT ip_to_bits(IPADDRESS '4ffe:2900:5545:3210:2000:f8ff:fe21:67cf');
-- 0100111111111110001010010000000001010101010...000010110011111001111
```

```sql
SELECT ip_to_bits(IPADDRESS 'ffff:ffff:ffff:ffff:ffff:ffff:ffff:ffff', 112);
-- 1111111111111111111111111111111111111111111...111110000000000000000
```

```sql
SELECT ip_to_bits(IPADDRESS '192.172.1.20');
-- 11000000101011000000000100010100
```

```sql
SELECT ip_version(IPADDRESS '192.172.1.20')
-- 4
```

```sql
SELECT ip_version(IPADDRESS '64:ff9b::')
-- 6
```

## Release notes

( X ) Release notes are required, with the following suggested text:

```markdown
## General
* Add functions to convert IPADDRESS to 128 bits varchar representation and vice versa: ip_to_bits and ip_from_bits. Add `ip_version(ipaddress)`
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
@@ -15,19 +15,28 @@ package io.trino.operator.scalar;
 
 import com.google.common.net.InetAddresses;
 import com.google.common.primitives.Ints;
+import io.airlift.slice.DynamicSliceOutput;
 import io.airlift.slice.Slice;
+import io.airlift.slice.SliceOutput;
+import io.airlift.slice.Slices;
 import io.trino.spi.TrinoException;
 import io.trino.spi.function.Description;
 import io.trino.spi.function.ScalarFunction;
 import io.trino.spi.function.SqlType;
 import io.trino.spi.type.StandardTypes;
 
+import java.io.IOException;
 import java.math.BigInteger;
 import java.net.Inet4Address;
 import java.net.InetAddress;
+import java.util.OptionalLong;
 import java.util.regex.Pattern;
 
+import static io.trino.spi.StandardErrorCode.GENERIC_INTERNAL_ERROR;
 import static io.trino.spi.StandardErrorCode.INVALID_FUNCTION_ARGUMENT;
+import static io.trino.spi.type.StandardTypes.INTEGER;
+import static io.trino.spi.type.StandardTypes.IPADDRESS;
+import static io.trino.spi.type.StandardTypes.VARCHAR;
 
 public final class IpAddressFunctions
 {
@@ -117,5 +126,101 @@ public final class IpAddressFunctions
         }
 
         return bytes;
+    }
+
+    @ScalarFunction("ip_to_bits")
+    @Description("Convert the IP address to its IPv4 32-bit or IPv6 128-bit representation")
+    @SqlType(VARCHAR)
+    public static Slice toBits(@SqlType(IPADDRESS) Slice slice)
+    {
+        return toBits(slice, OptionalLong.empty());
+    }
+
+    @ScalarFunction("ip_to_bits")
+    @Description("Convert the IP address to its IPv4 32-bit or IPv6 128-bit representation for a specific subnet")
+    @SqlType(VARCHAR)
+    public static Slice toBits(@SqlType(IPADDRESS) Slice slice, @SqlType(INTEGER) long subnet)
+    {
+        return toBits(slice, OptionalLong.of(subnet));
+    }
+
+    private static Slice toBits(Slice slice, OptionalLong optionalSubnet)
+    {
+        if (slice.length() != 16) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Invalid IP address binary length, expected 16 but was " + slice.length());
+        }
+
+        boolean isIPv4 = ipVersion(slice) == 4;
+        int totalBits = isIPv4 ? 32 : 128;
+        long subnet = optionalSubnet.orElse(totalBits);
+
+        if (isIPv4) {
+            if (subnet < 0 || subnet > 32) {
+                throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Subnet address must be between 0 and 32 for IPv4, but was " + subnet);
+            }
+        }
+        else if (subnet < 0 || subnet > 128) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Subnet address must be between 0 and 128 for IPv6, but was " + subnet);
+        }
+
+        int bit = 0;
+        byte[] result = new byte[totalBits];
+        for (int i = isIPv4 ? 12 : 0; i < 16 && bit < subnet; i++) {
+            int value = slice.getByte(i) & 0xFF;
+            for (int shift = 7; shift >= 0 && bit < subnet; shift--) {
+                result[bit] = (byte) (((value >>> shift) & 1) == 0 ? '0' : '1');
+                bit++;
+            }
+        }
+
+        while (bit < totalBits) {
+            result[bit++] = '0';
+        }
+
+        return Slices.wrappedBuffer(result);
+    }
+
+    @ScalarFunction("ip_from_bits")
+    @Description("Convert an IPv4 32-bit or IPv6 128-bit binary IP address to an IPADDRESS")
+    @SqlType(IPADDRESS)
+    public static Slice fromBits(@SqlType(VARCHAR) Slice slice)
+    {
+        boolean isIPv4 = slice.length() == 32;
+        if (!isIPv4 && slice.length() != 128) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Invalid IP address bits length, expected 32 or 128 but was " + slice.length());
+        }
+
+        try (SliceOutput output = new DynamicSliceOutput(16)) {
+            if (isIPv4) {
+                output.writeZero(10);
+                output.writeShort(0xFFFF);
+            }
+
+            for (int i = isIPv4 ? 12 : 0, n = 0; i < 16; i++) {
+                int value = 0;
+                for (int j = 0; j < 8; j++, n++) {
+                    int bit = switch (slice.getByte(n)) {
+                        case '0' -> 0;
+                        case '1' -> 1;
+                        default -> throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Invalid IP address bits, expected only '0' and '1'");
+                    };
+                    value <<= 1;
+                    value |= bit;
+                }
+                output.writeByte((byte) value);
+            }
+            return output.slice();
+        }
+        catch (IOException e) {
+            throw new TrinoException(GENERIC_INTERNAL_ERROR, e);
+        }
+    }
+
+    @ScalarFunction("ip_version")
+    @Description("Returns 4 for IPv4 and 6 for IPv6")
+    @SqlType(INTEGER)
+    public static long ipVersion(@SqlType(IPADDRESS) Slice slice)
+    {
+        return slice.getLong(0) == 0 && slice.getInt(8) == 0xFFFF_0000 ? 4 : 6;
     }
 }
--- a/docs/src/main/sphinx/functions/ipaddress.md
+++ b/docs/src/main/sphinx/functions/ipaddress.md
@@ -14,3 +14,59 @@ SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0
 SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0:ff00:0042:8328'); -- false
 ```
 :::
+
+:::{function} ip_version(ipaddress) -> integer
+Returns `4` for IPv4 and `6` for IPv6.
+```sql
+SELECT ip_version(IPADDRESS '192.172.1.20');
+-- 4
+```
+:::
+
+:::{function} ip_from_bits(varchar) -> ipaddress
+Convert a 32-bit binary IPv4 address to an IPADDRESS.
+```sql
+SELECT ip_from_bits('11000000101011000000000100010100');
+-- 192.172.1.20
+```
+
+Or convert a 128-bit binary IPv6 address to an IPADDRESS.
+```sql
+SELECT ip_from_bits(concat(
+    '01001111111111100010100100000000',
+    '01010101010001010011001000010000',
+    '00100000000000001111100011111111',
+    '11111110001000010110011111001111'));
+-- 4ffe:2900:5545:3210:2000:f8ff:fe21:67cf
+```
+:::
+
+:::{function} ip_to_bits(ipaddress) -> varchar
+Convert an IPv4 address to its 32-bit representation
+```sql
+SELECT ip_to_bits(IPADDRESS '192.172.1.20');
+-- 11000000101011000000000100010100
+```
+
+Or convert an IPv6 address to its 128-bit representation
+```sql
+SELECT ip_to_bits(IPADDRESS '4ffe:2900:5545:3210:2000:f8ff:fe21:67cf');
+-- 0100111111111110...0000000000111110001111111111111110001000010110011111001111
+```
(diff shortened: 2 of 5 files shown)


electrum:
We should prefix these with `ip_` as "from bits" isn't clear that it will return an IP address. We might want a similar function for other types.

gertjanal:
@electrum I will do that, but I noticed quite some functions use the same format:
```
- {func}`from_base`
- {func}`from_base32`
- {func}`from_base64`
- {func}`from_base64url`
- {func}`from_big_endian_32`
- {func}`from_big_endian_64`
- {func}`from_encoded_polyline`
- `from_geojson_geometry`
- {func}`from_hex`
- {func}`from_ieee754_32`
- {func}`from_ieee754_64`
- {func}`from_iso8601_date`
- {func}`from_iso8601_timestamp`
- {func}`from_iso8601_timestamp_nanos`
- {func}`from_unixtime`
- {func}`from_unixtime_nanos`
- {func}`from_utf8`
```

electrum:
@coderabbitai full review

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +126,86 @@ private static byte[] toBytes(InetAddress address)
 
         return bytes;
     }
+
+    @ScalarFunction("ip_to_bits")
+    @Description("Convert the ip address to its ipv6 128 bit representation")
```
electrum: > Convert the IP address to its IPv6 128-bit representation

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +126,86 @@ private static byte[] toBytes(InetAddress address)
...
+    @SqlType(VARCHAR)
+    public static Slice toBits(@SqlType(IPADDRESS) Slice slice)
+    {
+        return toBits(slice, Optional.empty());
+    }
+
+    @ScalarFunction("ip_to_bits")
+    @Description("Convert the ip address to its ipv6 128 bit representation")
```
electrum: > Convert the IP address to its IPv6 128-bit representation

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +126,86 @@ private static byte[] toBytes(InetAddress address)
...
+                    result.appendByte(bits.charAt(j));
+                }
+            }
+            for (long i = subnet; i < 128; i++) {
+                result.appendByte('0');
+            }
+
+            return result.slice();
+        }
+        catch (IOException e) {
```
electrum: This code doesn't seem to throw `IOException`

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +126,86 @@ private static byte[] toBytes(InetAddress address)
...
+            throw new TrinoException(GENERIC_INTERNAL_ERROR, e);
+        }
+    }
+
+    @ScalarFunction("ip_from_bits")
+    @Description("Convert a 128 bit ip address to a IPADDRESS")
```
electrum: > Convert a 128-bit binary IP address to an IPADDRESS

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +126,86 @@ private static byte[] toBytes(InetAddress address)
...
+            }
+            return output.slice();
+        }
+        catch (IOException e) {
+            throw new TrinoException(GENERIC_INTERNAL_ERROR, e);
+        }
+    }
+
+    @ScalarFunction("ip_version")
+    @Description("Returns 4 for ipv4 and 6 for ipv6")
```
electrum: > Returns 4 for IPv4, or 6 for IPv6

docs/src/main/sphinx/functions/ipaddress.md
```diff
@@ -14,3 +14,35 @@ SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0
...
+Get the ip version of an ipaddress, either `4` or `6`.
+
+:::{function} ip_from_bits(varchar) -> ipaddress
+Convert a varchar of 128 characters (`1` and `0` bits) to an IPADDRESS.
+```sql
+SELECT ip_from_bits('0100111111111110001010010000000001010101010...
+000010110011111001111')
+-- 4ffe:2900:5545:3210:2000:f8ff:fe21:67cf
+```
+:::
```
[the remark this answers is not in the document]
gertjanal: Do a fresh full review

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +126,86 @@ private static byte[] toBytes(InetAddress address)
...
+    @SqlType(INTEGER)
+    public static long ipVersion(@SqlType(IPADDRESS) Slice slice)
+    {
+        for (int i = 0; i < 9; i++) {
```
electrum: This can simplify to
```java
return (slice.getLong(0) == 0 && slice.getInt(8) == 0xFFFF_0000) ? 4 : 6;
```
Note that the original code has an off-by-one error in that it checks the first 9 bytes rather than the first 10 bytes.

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +127,90 @@ private static byte[] toBytes(InetAddress address)
...
+    @SqlType(IPADDRESS)
+    public static Slice fromBits(@SqlType(VARCHAR) Slice slice)
+    {
+        if (slice.length() != 128) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Invalid IP address bits length, expected 128 but was " + slice.length());
+        }
+
+        try (SliceOutput output = new DynamicSliceOutput(16)) {
+            for (int i = 0; i < 16; i++) {
+                output.writeByte((byte) Integer.parseInt(slice.slice(i * Byte.SIZE, Byte.SIZE).toStringAscii(), 2));
```
electrum: Rather than converting to a string and parsing it, you can parse the slice directly. For each byte, read the value and then shift. Something like this:
```java
int n = 0;
for (int i = 0; i < 16; i++) {
    int value = 0;
    for (int j = 0; j < 8; j++)
        int bit = switch (slice.getByte(n)) {
            case '0' -> 0;
            case '1' -> 1;
            default -> throw ...
        }
        value <<= 1;
        value |= bit;
        n++;
    }
    output.writeByte((byte) value);
}

File: core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java [hunk 9]
```diff
@@ -118,4 +127,90 @@ private static byte[] toBytes(InetAddress address)
...
+            }
+            return output.slice();
+        }
+        catch (NumberFormatException e) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Invalid IP address bits, expected only '0' and '1'");
+        }
+        catch (IOException e) {
```
electrum, line 214: Why do we have `IOException`?

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +127,90 @@ private static byte[] toBytes(InetAddress address)
...
+        if (subnet < 0 || subnet > 128) {
+            throw new TrinoException(INVALID_FUNCTION_ARGUMENT, "Subnet address must be between 0 and 128 for IPv6, but was " + subnet);
+        }
+
+        int bit = 0;
+        byte[] result = new byte[128];
+        for (int i = 0; i < 16 && bit < subnet; i++) {
+            int value = slice.getByte(i) & 0xFF;
+            for (int shift = 7; shift >= 0 && bit < subnet; shift--) {
+                result[bit++] = ((value >>> shift) & 1) == 0 ? (byte) '0' : (byte) '1';
```
electrum: Move the `bit++` to a separate statement

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +127,90 @@ private static byte[] toBytes(InetAddress address)
...
+            }
+        }
+
+        while (bit < 128) {
+            result[bit++] = '0';
```
electrum, line 177: Move the `bit++` to a separate statement

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
```diff
@@ -118,4 +127,90 @@ private static byte[] toBytes(InetAddress address)
...
+
+    @ScalarFunction("ip_to_bits")
+    @Description("Convert the ip address to its ipv6 128 bit representation")
+    @SqlType(VARCHAR)
+    public static Slice toBits(@SqlType(IPADDRESS) Slice slice, @SqlType(INTEGER) long subnet)
+    {
+        return toBits(slice, Optional.of(subnet));
+    }
+
+    private static Slice toBits(Slice slice, Optional<Long> optionalSubnet)
```
electrum: Use `OptionalLong`

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
electrum: Can cast to byte on the outside

docs/src/main/sphinx/functions/ipaddress.md
```diff
@@ -14,3 +14,40 @@ SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0
 SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0:ff00:0042:8328'); -- false
 ```
 :::
+
+:::{function} ip_version(ipaddress) -> integer
+Get the ip version of an ipaddress, either `4` or `6`.
```
electrum: Capitalize IP everywhere

docs/src/main/sphinx/functions/ipaddress.md
```diff
@@ -14,3 +14,40 @@ SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0
...
+SELECT ip_to_bits(IPADDRESS '4ffe:2900:5545:3210:2000:f8ff:fe21:67cf')
+-- 0100111111111110001010010000000001010101010...000010110011111001111
+```
+:::
+
+:::{function} ip_to_bits(ipaddress, integer) -> varchar
+:noindex: true
+
+Convert the ip address to its 128 bit representation, with zeroes for the
+subnet range. Note that ipv4 is converted to ipv6.
```
electrum: I think this is the wrong behavior. For IPv4, we should return the natural 32 bit representation, not the mapped representation, just like we return IPv4 strings for IPv4 addresses.

docs/src/main/sphinx/functions/ipaddress.md
```diff
@@ -14,3 +14,40 @@ SELECT contains('2001:0db8:0:0:0:ff00:0042:8329/128', IPADDRESS '2001:0db8:0:0:0
...
+Convert a varchar of 128 characters (`1` and `0` bits) to an IPADDRESS.
+```sql
+SELECT ip_from_bits('0100111111111110001010010000000001010101010...
+000010110011111001111')
+-- 4ffe:2900:5545:3210:2000:f8ff:fe21:67cf
+```
+:::
+
+:::{function} ip_to_bits(ipaddress) -> varchar
+Convert the ip address to its 128 bit representation. Note that ipv4 is
```
electrum: I think this is the wrong behavior. For IPv4, we should return the natural 32 bit representation, not the mapped representation, just like we return IPv4 strings for IPv4 addresses.

electrum (changes requested):

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
gertjanal: Yes, luckily the code rabbit noticed that as well :)

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
gertjanal: nice, updated

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
gertjanal: The try with DynamicOutput needs it: `Unhandled exception from auto-closeable resource: java.io.IOException`

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
gertjanal: ok

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
gertjanal: Nice

core/trino-main/src/main/java/io/trino/operator/scalar/IpAddressFunctions.java
gertjanal: Done

docs/src/main/sphinx/functions/ipaddress.md
gertjanal: Rewritten

docs/src/main/sphinx/functions/ipaddress.md
gertjanal: Rewritten

gertjanal:
@coderabbitai full review

wendigo:
@martint can you review syntax?

wendigo:
@martint PTAL

martint (commented):
This seems like a very niche and low-level representation to expose as an IP-specific SQL function. Representing every bit as an ASCII 0 or 1 in a varchar is inefficient and does not seem like good user-facing UX. It also discards the semantic type information, leaving callers with a string rather than an address or network.

The motivating operation appears to be truncating an address to a subnet for grouping, so we could model that operation directly—for example, `ip_truncate(ipaddress, prefix_length) -> ipaddress`.

If we expect to support broader subnet operations, perhaps we should instead introduce an IPPREFIX type that represents a network address together with its prefix length. We could then provide functions such as `ip_prefix(ipaddress, prefix_length) -> ipprefix`, `contains(ipprefix, ipaddress)`, `network_address(ipprefix)`, and `prefix_length(ipprefix)`. This would give subnet values a canonical representation and allow comparison, grouping, and containment without encoding them as strings.

If textual bit-string conversion is independently useful, it seems more appropriate as a generic varbinary encoding rather than an ipaddress-specific function.

PR #28969, opened 2026-04-01
Evaluate JSON_VALUE defaults lazily

martint:
SQL:2023 subclause 9.48 says JSON_VALUE evaluates and casts `DEFAULT <value expression>` ON EMPTY/ON ERROR only when the selected branch is taken.

Trino currently evaluates JSON_VALUE defaults eagerly because the hidden $json_value function receives them as ordinary arguments. That can fail even when the corresponding branch is never used, and it prevents ON EMPTY default failures from being handled by ON ERROR.

Pass the default expressions to $json_value as zero-argument lambdas, evaluate them only in the selected branch, and cast them to the declared return type there. 

## Release notes

```markdown
## General
* Evaluate {func}`json_value` and {func}`json_table` `ON EMPTY`/`ON ERROR` clauses lazily. ({issue}`TBD`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/operator/scalar/json/JsonValueFunction.java
+++ b/core/trino-main/src/main/java/io/trino/operator/scalar/json/JsonValueFunction.java
@@ -16,6 +16,7 @@ package io.trino.operator.scalar.json;
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.NullNode;
 import com.google.common.collect.ImmutableList;
+import com.google.common.primitives.Primitives;
 import io.airlift.slice.Slice;
 import io.trino.annotation.UsedByGeneratedCode;
 import io.trino.json.JsonPathEvaluator;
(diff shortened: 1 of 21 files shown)


martint:
@coderabbitai review

martint:
@coderabbitai review

martint:
@coderabbitai review

martint:
@coderabbitai review

core/trino-main/src/main/java/io/trino/operator/scalar/json/JsonValueFunction.java
```diff
@@ -309,19 +341,90 @@ public static Object jsonValue(
...
+            case NULL -> null;
+            case ERROR -> throw error.get();
+            case DEFAULT -> {
+                try {
+                    yield evaluateDefault(session, emptyDefaultCoercion, emptyDefault);
+                }
+                catch (TrinoException defaultError) {
+                    // Only handle evaluation errors (e.g., division by zero, cast overflow) by cascading
+                    // to the ON ERROR clause. Other errors indicate unexpected failures and should propagate.
+                    TryFunction.propagateIfUnhandled(defaultError);
```
kasiafi, line 371: This is too weak to capture all errors:
```
trino:tpch> SELECT JSON_VALUE(
         ->     '[]',
         ->     'lax $[1]'
         ->     RETURNING varchar
         ->     DEFAULT JSON_OBJECT('a' : 1, 'a' : 2) ON EMPTY
         ->     NULL ON ERROR);
Query 20260420_080910_00025_h38u6 failed: cannot construct a JSON object with duplicate key
```

```
trino:tpch> SELECT JSON_VALUE(
         ->     '[]',
         ->     'lax $[1]'
         ->     RETURNING bigint
         ->     DEFAULT fail('boom') ON EMPTY
         ->     NULL ON ERROR);
Query 20260420_081126_00026_h38u6 failed: boom
```

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -2820,6 +2824,19 @@ private Type analyzeJsonValueExpression(
             return function.signature().getReturnType();
         }
 
+        private void verifyDefaultCoercion(Expression defaultValue, Type defaultType, Type returnedType, String message)
```
kasiafi, line 2830: Shouldn't coercion failure be deferred to execution?

core/trino-main/src/test/java/io/trino/sql/planner/TestJsonTable.java
```diff
@@ -78,66 +62,37 @@ public class TestJsonTable
...
-                                                                        ImmutableMap.of(
-                                                                                "name", expression(new Constant(createVarcharType(5), Slices.utf8Slice("[ala]"))),
-                                                                                "default_value", expression(new Constant(INTEGER, 5L))),
-                                                                        anyTree(
-                                                                                project(
-                                                                                        ImmutableMap.of(
-                                                                                                "json_col", expression(new Constant(createVarcharType(9), Slices.utf8Slice("[1, 2, 3]"))),
-                                                                                                "int_col", expression(new Constant(INTEGER, 4L))),
-                                                                                        values(1)))))))))));
+        getPlanTester().inTransaction(transactionSession -> {
```
kasiafi: Please keep the test verbose. It is supposed to assert input preprocessing, output layout etc.

kasiafi (commented):
I think we should document the lazy behavior. Also, we should document which parts now support subqueries.

core/trino-main/src/main/java/io/trino/operator/scalar/json/JsonValueFunction.java
martint: Interesting. That `try` itself is to weak for JSON_VALUE failures.

```
SELECT try(JSON_VALUE(...))
```

should return null for the examples above.

martint: Also, the first error is a NOT_SUPPORTED error. Presumably, it will be supported at some point, so it shouldn't be caught by the ON ERROR clause.

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
martint: This only catches OperatorNotFoundException. It's a static type system error, not a per-row cast failure.

core/trino-main/src/main/java/io/trino/operator/scalar/json/JsonValueFunction.java
martint: I added a commit to expand the list of errors supported by `try`

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
kasiafi: The query could still succeed if it didn't need to use the default value.

martint: This is about lazy evaluation, not dynamic typing, though. It's analogous to:

```
SELECT 'x' * 5
FROM t
WHERE false
```

Technically, the `'x' * 5` never gets evaluated per the SQL evaluation model, but it's an invalid expression because of type mismatch.

kasiafi (approved):

PR #28988, opened 2026-04-03
Make `partitionValues` serializable for `IcebergSplit`

chenjian2664:
## Description

The `ConnectorSplit` serves as the data carrier between the coordinator and workers, and therefore should be fully serializable by design. Making `partitionValues` serializable is the first step toward correcting the current implementation.

## Release notes

```markdown
## Section
```


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
```diff
@@ -64,6 +65,7 @@ public IcebergSplit(
             @JsonProperty("fileSize") long fileSize,
             @JsonProperty("fileRecordCount") long fileRecordCount,
             @JsonProperty("fileFormat") IcebergFileFormat fileFormat,
+            @JsonProperty("partitionValues") Optional<List<Block>> partitionValues,
             @JsonProperty("specId") int specId,
             @JsonProperty("partitionDataJson") String partitionDataJson,
```
raunaqmorarka: Can we get rid of `partitionDataJson` through using `partitionValues` ?
It feels redundant to have both

chenjian2664: The `partitionValues` currently is derived from partition + IcebergTablePartitioning, thus we can't remove it directly.
I am going to keep only one between them, do you have suggestions on it?

chenjian2664:
@coderabbitai full review

chenjian2664:
@raunaqmorarka PTAL

raunaqmorarka (commented):
Review comments on partition value serialization changes.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -743,9 +734,8 @@ private IcebergSplit toIcebergSplit(FileScanTaskWithDomain taskWithDomain)
                 task.file().fileSizeInBytes(),
```
raunaqmorarka, line 736: `spec.fields()` includes void transforms but `spec.partitionType()` excludes them. So when `fieldIndex` hits a void transform, `partitionType().field(fieldId)` returns `null` and `.type()` throws NPE.

The old code was safe because `partitionStructFields` was filtered in `applyPartitioning` before reaching `toIcebergSplit`, so void indices were never looked up against `partitionType()`. The new `getPartitionBlockValues` runs unconditionally for every split.

The defensive check in `fromBlocks` (`i >= partitionValues.size()`) suggests you're aware of this — but the bug is on the creation side, not the consumption side.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPartitioningHandle.java
```diff
@@ -59,7 +63,7 @@ public static IcebergPartitioningHandle create(PartitionSpec spec, TypeManager t
                         toTrinoType(spec.schema().findType(field.sourceId()), typeManager)))
                 .collect(toImmutableList());
 
-        return new IcebergPartitioningHandle(false, partitionFields);
+        return new IcebergPartitioningHandle(false, partitionFields, IntStream.range(0, partitionFields.size()).boxed().collect(toImmutableList()));
```
raunaqmorarka, line 66: The identity mapping `[0, 1, ..., n-1]` assumes every spec field has a corresponding entry in the partition struct. Void transforms don't — `partitionType()` skips them, so later fields have shifted indices. This works today only because `applyPartitioning` filters them out before use, but the handle itself carries incorrect indices.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergBucketFunction.java
```diff
@@ -99,15 +106,26 @@ public int applyAsInt(ConnectorSplit split)
         return (int) ((hash & Long.MAX_VALUE) % bucketCount);
     }
 
-    private record HashFunction(List<Integer> dataPath, ValueTransform valueTransform, MethodHandle hashCodeOperator)
+    private List<Object> getPartitionValues(List<Block> partitionBlocks)
+    {
+        List<Object> partitionValues = new ArrayList<>(partitionStructFields.size());
```
raunaqmorarka, line 110: Minor: the rest of the codebase prefers `ImmutableList.Builder` for lists that are built and returned.

chenjian2664: The value could be null, so we can't use `ImmutableList`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
chenjian2664: > spec.fields() includes void transforms but spec.partitionType() excludes them

AFAIK, it's not, the `sepc.partitionType()` doesn't excludes the void transform, let me know if you find the code that excluding it. It's the `spec.isPartitioned()` will filter the void transform - treat it as non-partitioned field, and I didn't find where is the logic that filtering the void for `partitionStructFields` in `applyPartitioning`(I haven't checked the upstream caller though).
See the fixup commit 9105dee I appended - I removed the shortcut that relied on `spec.isUnpartitioned` and added a test to verify if the void transform exist in front of other type partition field doesn't break current logic.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPartitioningHandle.java
chenjian2664: see the above response, the `partitionType` seems not filter the void fields:
https://github.com/apache/iceberg/blob/ec2de913bfea2e1799aa2b72a9a9796418a8d839/api/src/main/java/org/apache/iceberg/PartitionSpec.java#L126-L149

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergBucketFunction.java
chenjian2664: I've added a comment for it

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
```diff
@@ -370,7 +368,7 @@ public ConnectorPageSource createPageSource(
                 length,
                 fileSize,
                 partitionSpec.specId(),
-                partitionDataJson,
+                PartitionData.toJson(partitionData),
```
raunaqmorarka, line 371: I'm wondering why this turns back into json, we seem to be converting back to block later in the usage

raunaqmorarka (commented):
lgtm % minor comment

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java
chenjian2664: we removed the the `partitionDataJson` from `IcebergSplit`, previously it was feed from it and then put into the mergeRowId block

chenjian2664: I don't want to change the existing logic, but please share if you think it should be improved

raunaqmorarka: If it is easy to improve, we should do it.
The JSON round trip would be nice to avoid since we eventually want to use the partition value as a block anyway

chenjian2664: Are you referring to the code on
 https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergPageSourceProvider.java#L1736 

It looks like the entire partition data JSON is written into the block as a single value(instead of writing all field's value into a block), so converting it to JSON is necessary. Later, in the merge sink, we rely on this to deserialize it back into `PartitionData`.

Unless you'd prefer changing the block type from a string to a row type, in that case I doubt if we are worth doing that.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergBucketFunction.java
```diff
@@ -81,8 +88,8 @@ public int getBucket(Page page, int position)
     @Override
     public int applyAsInt(ConnectorSplit split)
     {
-        List<Object> partitionValues = ((IcebergSplit) split).getPartitionValues()
-                .orElseThrow(() -> new IllegalArgumentException("Split does not contain partition values"));
+        List<Object> partitionValues = getPartitionValues(((IcebergSplit) split).getPartitionValues());
+        checkState(!partitionValues.isEmpty(), "Split does not contain partition values");
```
raunaqmorarka: Nit: this `checkState` does not guard against the actual failure mode — if `partitionStructFields` indexes out of bounds on `partitionBlocks`, the error would come from inside `getPartitionValues`, not here. The check only catches the case where `partitionStructFields` itself is empty, which should never happen for a bucketed split. Consider removing it, or replacing it with a more precise assertion.

raunaqmorarka (approved):

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -831,7 +831,7 @@ private Optional<IcebergTablePartitioning> getTablePartitioning(ConnectorSession
             return Optional.empty();
         }
```
findinpath, line 832: "Remove unused parma `partitioningColumns`"
typo `parma` -> `param`

Issue #12920, opened 2022-06-21
System tables not accessible with AS OF syntax 


phd3:
Queries like below (in iceberg connector) fail with

```
SELECT * FROM "test_iceberg_read_versioned_table$partitions" FOR VERSION AS OF ...
```

error message 

```
This connector does not support versioned tables
```

#10768

findepi:
I would call it a missing feature, not a bug.

phd3:
fair, this'd likely be the case for any connector we support

PR #28997, opened 2026-04-05
Replace PartitionsTable with PartitionsView

tbaeg:
## Description
Replace `PartitionsTable` with `PartitionsView`.

### Problem

Currently, Trino's `SystemTable` SPI is designed strictly for flat, scan-level operations. It lacks a mechanism to natively trigger a distributed shuffle (Exchange) for aggregations or post-processing. This presents two hurdles:

1. Coordinator Bottlenecks: While 1:1 metadata projections like Iceberg's `$files` can be scanned in a distributed fashion, `$partitions` requires a global GROUP BY. Because the current API cannot distribute this reduction, the aggregation is forced into memory on a single coordinator node.

2. Code Duplication: Many Iceberg metadata tables analyze the same underlying manifest files. For example, the data required for `$partitions` can be entirely derived by aggregating the columns already exposed by `$files`.

### Solution

Leverage view(s) on top of `SystemTable`s. That allows connectors to define system "tables" using standard SQL queries. This delegates the projection and aggregation logic back to Trino’s execution engine.

By leveraging view(s), connectors can enable engine-native, distributed operations for metadata. Instead of writing code for every `SystemTable` where the information is determined by the same source but projected differently, the connector can simply layer a view on top of its existing distributed system table implementations.

Continuing with the Iceberg example, the connector can parallelize `$partitions` by defining it as a view over `$files`:

```
SELECT 
    partition, 
    COUNT(*) AS file_count,
    SUM(record_count) AS record_count,
    SUM(file_size_in_bytes) AS total_size
    ...
FROM "table$files" 
GROUP BY 1
```

As more connectors adapt this pattern, a formal additional to the SPI (i.e. - `SystemView`) may be beneficial.

## Release notes

```markdown
## Iceberg
*  Improve performance of queries on `$partitions` system table. ({issue}`23147`)
*  {{breaking}} Change lower_bounds and upper_bounds columns in `$files` from `map(integer, varchar)` to `row(...)` with per-column typed fields. Queries using `lower_bounds[1]` or `CAST(lower_bounds AS JSON)` must be updated. ({issue}`23147`)
```


electrum:
Thanks for proposing this. Using a view for the Iceberg `$partitions` makes sense, and doing it as a normal distributed query is an improvement. From an implementation perspective, I don't think we need to add anything new to the SPI. Instead, `IcebergMetadata.getView()` could return the view for the system table, before checking for a normal catalog view.

From the security side, I think it only makes sense to use `INVOKER` security here, as there is no identity (or view "owner") to use for `DEFINER` security.

tbaeg:
> Thanks for proposing this. Using a view for the Iceberg `$partitions` makes sense, and doing it as a normal distributed query is an improvement. From an implementation perspective, I don't think we need to add anything new to the SPI. Instead, `IcebergMetadata.getView()` could return the view for the system table, before checking for a normal catalog view.
> 
> From the security side, I think it only makes sense to use `INVOKER` security here, as there is no identity (or view "owner") to use for `DEFINER` security.

Thanks for the reply! Using `ConnectorMetadata.getView()` is possible, but I think having a separate `SystemView` provides:

1. SPI clarity and intent by reducing code in the relevant methods to expose the system view (i.e. - `listTables`/`listViews`/etc).
2. Related to clarity/intent, but specifically tracing execution path. For example, how a system view lands into `information_schema` becomes muddied coming from the something like `ConnectorMetadata.listTables()`. Identifying problems to, and through `SystemTablesMetadata` is much easier since it's paired down to the "system" resources.
3. Ability to enforce certain rules for all system views (i.e. - requiring `INVOKER` security in all system views). 

@electrum Thoughts?

raunaqmorarka:
Having `$partitions` defined as a view on top of `$files` makes sense, we already have a situation in  https://github.com/trinodb/trino/pull/28911 where having this would avoid maintaining redundant code

electrum:
We shouldn't need to touch any of the listing code, as we don't list these hidden system tables today.

Can you try to implement this using only `getView` for Iceberg and see if there are any issues?

tbaeg:
> We shouldn't need to touch any of the listing code, as we don't list these hidden system tables today.
> 
> Can you try to implement this using only `getView` for Iceberg and see if there are any issues?

Yes, I can try doing it for the dynamic system views.

Is this true for static system views? I believe you'd need to inject listing code if it wasn't part of the `SystemTableMetadata`.

tbaeg:
I added a static and dynamic view to `getView`. It works, but the static view does not show up in `information_schema` without modifying the `listX` method calls.

@electrum

tbaeg:
> Having `$partitions` defined as a view on top of `$files` makes sense, we already have a situation in #28911 where having this would avoid maintaining redundant code

This same principle can be applied to other metadata tables (i.e. - `$entries`/`$all_entries`) and trivializes adding new system tables like `$all_files`/`$delete_files`/`$data_files`/`$all_delete_files`/`$all_data_files`/etc.

I also believe other connectors like `deltalake` has similar issues with with `$partitions`; although I'm not too familiar with `deltalake`.

@raunaqmorarka Did you have any thoughts on the SPI changes?

tbaeg:
@electrum @raunaqmorarka Gentle reminder.

wendigo:
@martint what do you think?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
```diff
@@ -0,0 +1,204 @@
...
+                .formatted(min, max, nullCount, nanCount);
+    }
+
+    private static String buildColumnRowType(String columnName, String trinoTypeDisplayName)
+    {
+        return "%s ROW(min %2$s, max %2$s, null_count BIGINT, nan_count BIGINT)"
+                .formatted("\"" + columnName + "\"", trinoTypeDisplayName);
+    }
+
+    private static String getTypedProjection(String aggregationName, String fieldName, String trinoTypeDisplayName, int fieldId)
```
raunaqmorarka: The string-based type dispatch here is the most visible symptom of a structural issue: `$files.lower_bounds`/`upper_bounds` are `Map<INTEGER, VARCHAR>` of `Transforms.identity().toHumanString(...)` output, so any view layered on top has to reconstruct types from human-strings.

This causes:
- **Session-timezone-dependent output for Iceberg `timestamp` (no-tz).** `from_iso8601_timestamp_nanos` returns `TIMESTAMP(9) WITH TIME ZONE` (`DateTimeFunctions.java:261`); for a no-tz column the human string carries no tz suffix, so the function applies the **session timezone** before the cast back to `TIMESTAMP(6)`. The old `PartitionsTable` decoded bytes via `Conversions.fromByteBuffer` and was timezone-independent.
- **Null-predicate parity divergence** vs. `PartitionsTable`: `PartitionsTable.java:246` nulls when 3 fields are null (ignoring `nanCount`); the view's CASE on line 155 nulls only when all 4 are null. Comment claims compatibility but the predicates differ.
- **Brittle dispatch** via `getDisplayName().startsWith(...)`.

Aligned with the goal of reducing maintenance via views over `$files`: strengthen `$files` to expose typed bounds (per-field-id struct with native Trino types — Iceberg already emits `readable_metrics` internally that's similar). Then `PartitionsView` collapses to a few-line `SELECT partition, ..., MIN(typed_bounds[fid].min) ... GROUP BY 1`, and every future system view layered on `$files` inherits the typed decoding for free.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
```diff
@@ -0,0 +1,204 @@
...
+    {
+        return SchemaTableName.schemaTableName(schemaName, tableName + "$partitions");
+    }
+
+    @Override
+    public ConnectorViewDefinition getViewDefinition()
+    {
+        String viewSql = """
+                SELECT %s SUM(record_count) AS record_count, COUNT(*) AS file_count, SUM(file_size_in_bytes) AS total_size%s
+                FROM "%s"."%s"."%s$files"%s
```
raunaqmorarka: Catalog/schema/table names are interpolated raw into stored SQL that gets re-parsed on every view use; any `"` in a name produces broken syntax. Same concern applies to the column name in `buildColumnRowType` (line 165).

The iceberg plugin already has `IcebergUtil.quotedName` for this — it skips quoting for simple names and escapes `"` → `""` otherwise. One-line replacement per call site.

(Mostly moot once the view body is collapsed via richer `$files` columns — the FROM clause is the only identifier interpolation that remains.)

raunaqmorarka (commented):
Six concerns from a deeper read, posted as inline comments. The overall direction (view-over-`$files` to reduce maintenance burden) is sound — the comments are about strengthening `$files` so views layered on it can stay simple, and tightening the SPI plumbing (handles, version, name-collision) so this scales beyond `$partitions`.

1. **PartitionsView** — strengthen `$files` to expose typed lower/upper bounds; root cause of the type-dispatch fragility, the timestamp-without-tz session-timezone bug, and the null-predicate parity divergence vs. `PartitionsTable`.
2. **SystemTablesMetadata.getTableHandle** — return null for view-only matches; the current behavior produces an unusable handle.
3. **SystemTablesViewsProvider** — enforce the unfixed TODO with `checkArgument`.
4. **MetadataManager** — plumb `startVersion`/`endVersion` through `isView`/`getView` so `$partitions FOR VERSION AS OF X` errors via the same path as `$files FOR VERSION` instead of silently expanding at HEAD.
5. **PartitionsView** — use `IcebergUtil.quotedName` for FROM-clause identifiers.
6. **IcebergUtil** — split the unrelated Variant-type early-return into its own commit.

core/trino-main/src/main/java/io/trino/connector/system/SystemTablesMetadata.java
```diff
@@ -52,35 +55,68 @@
...
     }
 
     @Override
     public ConnectorTableHandle getTableHandle(ConnectorSession session, SchemaTableName tableName, Optional<ConnectorTableVersion> startVersion, Optional<ConnectorTableVersion> endVersion)
     {
-        Optional<SystemTable> table = tables.getSystemTable(session, tableName);
-        if (table.isEmpty()) {
-            return null;
-        }
+        boolean exists = tablesAndViews.getSystemTable(session, tableName).isPresent() ||
```
raunaqmorarka: Returning a `SystemTableHandle` for a name that resolves only as a view is a latent foot-gun: downstream `checkAndGetTable` (line 165) looks up `getSystemTable` only and would throw `NOT_FOUND` for a view's handle. Currently masked because the analyzer takes the view path first via `isView`/`getView`.

Suggest: only return a handle when `getSystemTable(session, tableName).isPresent()`. The view path is already correctly routed via the analyzer, and an unusable handle for view names doesn't add value.

core/trino-main/src/main/java/io/trino/connector/system/SystemTablesViewsProvider.java
```diff
@@ -30,32 +31,51 @@
...
 
-    public SystemTablesProvider(TransactionManager transactionManager, Metadata metadata, String catalogName, Set<SystemTable> systemTables)
+    public SystemTablesViewsProvider(
+            TransactionManager transactionManager,
+            Metadata metadata,
+            String catalogName,
+            Set<SystemTable> systemTables,
+            Set<SystemView> systemViews)
     {
+        // TODO ensure there is no table or view with the same name
```
raunaqmorarka: The TODO is enforceable in one line:

```java
checkArgument(Sets.intersection(systemTablesMap.keySet(), systemViewsMap.keySet()).isEmpty(),
        "System tables and views must have disjoint names");
```

Without it, a connector that registers the same `SchemaTableName` as both gets divergent routing — `getTableHandle` short-circuits on table; `isView` returns true. Iceberg avoids it by construction in the type switch, but the SPI shouldn't rely on that.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -1660,6 +1680,7 @@ private Optional<ConnectorViewDefinition> getViewInternal(Session session, Quali
         Optional<CatalogMetadata> catalog = getOptionalCatalogMetadata(session, viewName.catalogName());
         if (catalog.isPresent()) {
             CatalogMetadata catalogMetadata = catalog.get();
+            // TODO should this always throw when start/end is set?
```
raunaqmorarka: Resolving this TODO is a real correctness fix, not just a cleanup.

Today: `metadata.isView` (line 1614) and `getViewInternal` both pass `Optional.empty(), Optional.empty()` to `getCatalogHandle`. Version is dropped at catalog-handle dispatch, so the existing `NOT_SUPPORTED` throw in `SystemTablesMetadata.getTableHandle` never fires for view names. After this PR, `SELECT * FROM "tbl$partitions" FOR VERSION AS OF X` silently expands at HEAD instead of erroring like `$files FOR VERSION AS OF` does.

Suggested fix:
- Plumb `startVersion`/`endVersion` through `Metadata.isView` and `Metadata.getView`.
- Forward them into `catalogMetadata.getCatalogHandle(session, name, startVersion, endVersion)` here and in `isView`.
- The existing throw in `SystemTablesMetadata.getTableHandle` then fires; `$partitions FOR VERSION` and `$files FOR VERSION` produce the same error via the same path. Resolves both this TODO and the matching one in `getMaterializedViewInternal` (line 1908).

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
```diff
@@ -99,7 +99,7 @@ public void testEndVersionInTableNameAndForClauseShouldFail()
     public void testSystemTables()
     {
         // TODO https://github.com/trinodb/trino/issues/12920
-        assertQueryFails("SELECT * FROM \"test_iceberg_read_versioned_table$partitions\" FOR VERSION AS OF " + v1SnapshotId,
+        assertQueryFails("SELECT * FROM \"test_iceberg_read_versioned_table$files\" FOR VERSION AS OF " + v1SnapshotId,
```
raunaqmorarka, line 104: Switching from `$partitions` to `$files` keeps the assertion green but stops covering the actual subject of issue #12920 — `$partitions FOR VERSION` is now silently expanded at HEAD instead of throwing (see comment on `MetadataManager.getViewInternal`).

Once the version plumbing is fixed, restore the original `$partitions` assertion. Both should error via the same path.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -539,6 +539,10 @@ private static Stream<NestedField> primitiveFields(NestedField nestedField)
                     .map(field -> NestedField.from(field).withName(nestedField.name() + "." + field.name()).build());
         }
 
+        if (type.isVariantType()) {
```
raunaqmorarka, line 544: This `isVariantType()` short-circuit silently drops Variant fields from the `$files` projection and is unrelated to the SystemView change. Worth splitting into its own commit/PR with rationale (or removing) — easier to review the SystemView change in isolation.

tbaeg:
@raunaqmorarka @electrum Just for clarification, are we in favor of the SPI addition? I think there is a consensus for introducing the view based system tables but not necessarily the interface for them.

Would appreciate thoughts from @findinpath and @1fanwang.

raunaqmorarka:
I'd prefer the `getView`-only route. The fact that we have many such metadata tables in iceberg (`$entries`, `$data_files`, `$delete_files`, `$all_files`, ...) is a good argument for a registry, but the registry can live inside the iceberg module. `IcebergMetadata.getView`/`isView` would consult a suffix map before falling through to user views. INVOKER security and naming conventions get enforced once in a shared builder, same as a SystemView SPI would do.

The SPI route adds parallel `SystemTable`/`SystemView` dispatch, which already shows two issues in this PR. `SystemTablesMetadata.getTableHandle` returns a `SystemTableHandle` for view-only matches (masked today by analyzer ordering), and `SystemTablesViewsProvider` has an unenforced TODO for table/view name collisions. Both scale with the number of system views.

For `information_schema` visibility, per-table suffix views shouldn't be enumerated, since that would be combinatorial across every table. With `getView`, the connector simply doesn't list them in `getViews(SchemaTablePrefix)`. With a SystemView SPI, the engine layer has to actively decide what not to list.

I'd revisit a SystemView SPI when a second connector (delta?) needs the same pattern and we see what engine-side coordination is actually missing.

The independent issues from my inline review (typed `$files` bounds, `FOR VERSION` plumbing, identifier escaping, Variant short-circuit split) apply either way.

tbaeg:
@raunaqmorarka Thanks for your feedback!

Given the response(s), I will work toward a `getView` implementation. I put some responses to your feedback for a possible follow up discussion to around `SystemView`s.

> I'd prefer the `getView`-only route. The fact that we have many such metadata tables in iceberg (`$entries`, `$data_files`, `$delete_files`, `$all_files`, ...) is a good argument for a registry, but the registry can live inside the iceberg module. `IcebergMetadata.getView`/`isView` would consult a suffix map before falling through to user views. INVOKER security and naming conventions get enforced once in a shared builder, same as a SystemView SPI would do.

While it's possible to have it in it's own registry, as you mentioned, `delta` is yet another candidate for something very similar.

> The SPI route adds parallel `SystemTable`/`SystemView` dispatch, which already shows two issues in this PR. `SystemTablesMetadata.getTableHandle` returns a `SystemTableHandle` for view-only matches (masked today by analyzer ordering), and `SystemTablesViewsProvider` has an unenforced TODO for table/view name collisions. Both scale with the number of system views.

Yes, the `SystemTableHandle` returns view-only/there are name collisions (hence POC, heh). That said, I think there is a viable path to address this that includes the `SystemView`.

> For `information_schema` visibility, per-table suffix views shouldn't be enumerated, since that would be combinatorial across every table. With `getView`, the connector simply doesn't list them in `getViews(SchemaTablePrefix)`. With a SystemView SPI, the engine layer has to actively decide what not to list.

There was never an intention to expose dynamic tables to `information_schema`. Also, we already actively decide what not to list that for `SystemTable`s (i.e. - dynamic vs static system tables via the `SystemTablesProvider`).

> I'd revisit a SystemView SPI when a second connector (delta?) needs the same pattern and we see what engine-side coordination is actually missing.

Historically, iceberg has been the most demanding (as far as `SystemTable`s go), so, it is the best example of what another connector may need.

> The independent issues from my inline review (typed `$files` bounds, `FOR VERSION` plumbing, identifier escaping, Variant short-circuit split) apply either way.

Agreed. This was very much a POC, so definitely not comprehensive.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
tbaeg: Am I understanding correctly that there should be a new set of columns that provide(s) typed values for the minx/max values from `$files`? If so..

1.  Is it a prerequisite to for this PR?
2. I'd prefer to just replace the previous columns. I know it's a breaking change, but feels odd to have both.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
tbaeg: Separate commit added.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
tbaeg: Should using 'FOR VERSION' with a view always result in an exception?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
raunaqmorarka: I'd rather retype `lower_bounds`/`upper_bounds` in place than add a parallel `typed_bounds` column. `$files` is a system metadata table, not user data, so a one-time schema break is acceptable here, and the result is a cleaner surface than carrying two columns that encode the same information. The shape under `typed_bounds` (`row("1" row("lower" T, "upper" T), ...)`) looks right; I'd just put it on the existing column names.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
raunaqmorarka: Yes, for generic user-defined views `FOR VERSION` should reject. There's no defined way to push a time-travel clause through an arbitrary view body. For `$partitions` specifically we'd ideally want `FOR VERSION` to work by translating to `$files FOR VERSION` (the point of #12920), but until that's wired up, an error is the safe behavior.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -596,7 +600,7 @@ public static String quotedTableName(SchemaTableName name)
         return quotedName(name.getSchemaName()) + "." + quotedName(name.getTableName());
     }
 
-    private static String quotedName(String name)
+    public static String quotedName(String name)
```
raunaqmorarka, line 603: unrelated to this commit

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
tbaeg: Sounds good. I assume we'll want something slightly different like..

`lower_bounds` -> `row("1" T, "2" TT), ...)`
`upper_bounds` -> `row("1" T, "2" TT), ...)`

raunaqmorarka: Yes, that matches Iceberg's DataFile fields and keeps it as a pure type change on the existing columns.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
tbaeg: Ended up adding a separate test since the system view and table are different. I also added the check in the `StatementAnalyzer` that blanket fails versioning on views.

Does seem right or should it be in the `MetadataManager`?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableName.java
```diff
@@ -30,22 +30,33 @@ public final class IcebergTableName
...
         String referencableTableTypes = Stream.of(TableType.values())
                 .filter(tableType -> tableType != DATA)
                 .map(tableType -> tableType.name().toLowerCase(ENGLISH))
                 .collect(Collectors.joining("|"));
+
         TABLE_PATTERN = Pattern.compile("" +
                 "(?<table>[^$@]+)" +
                 "(?:\\$(?<type>(?i:" + referencableTableTypes + ")))?");
+
+        SYSTEM_VIEW_PATTERN = Pattern.compile("" +
```
tbaeg: Likely some clean up around this regex and the corresponding `TABLE_PATTERN` to be done. Since both are used in similar places but cause failures when `PARTITIONS` is filtered out from `TABLE_PATTERN`.

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -681,12 +682,20 @@ public Map<SchemaTableName, ConnectorViewDefinition> getViews(ConnectorSession s
     @Override
     public Optional<ConnectorViewDefinition> getView(ConnectorSession session, SchemaTableName viewName)
     {
+        if (isIcebergSystemViewName(viewName.getTableName())) {
```
tbaeg: While this pattern works, in the longer term, feels better if there was a distinction for a `SystemView`. It would have been nice to encapsulate some of this into something like `PartitionsSystemTableProvider` to avoid this sort of branching.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
raunaqmorarka: Master already supports `FOR VERSION AS OF '<branch>'` on views: `getBranchName` extracts the branch and `createScopeForView` threads it into the view body's table reference, so per-branch access control fires there. `TestBranchingAccessControl#testSelectFromViewBranch{Branch,Table}Privilege` locks that in with both `assertAccessDenied` and `assertAccessAllowed` cases.

Better to drop the analyzer reject and handle `$partitions FOR VERSION` locally in iceberg until #12920 lets us translate it to `$files FOR VERSION`. That keeps the user-view path intact.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableName.java
raunaqmorarka: Worth finishing here. Drop `PARTITIONS` from `TableType` so it's out of `referencableTableTypes`, and route `getRawSystemView`/`getRawBaseTable` through `SYSTEM_VIEW_PATTERN` + `isIcebergSystemViewName` (with a view-side `tableNameFrom` overload). Separates the system-view names from iceberg's system-table names.

tbaeg: I'm not convinced dropping `PARTITIONS` feels right. `isIcebergTableName` is more or less "what is supported as an iceberg system table that is not a data table". In this context, I don't think it necessarily means we need to delineate between views and non-views. Having `isSystemView` similar to `isDataTable`/`isMaterializedViewStorage` is sufficient and we can reassess adjustments later.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
tbaeg: I'm not sure how to handle it only in `iceberg` without changing the SPI for `getTableHandle` or `getView`. Right now the the exception is thrown in `SystemTablesMetadata.getTableHandle`. `getView` nor `getSystemTable` have any access to the version parameters.

Could you provide some additional guidance? Thanks in advance!

tbaeg: In the meantime, I lowered the scope of the rejection in the `StatementAnalyzer` so that the other tests properly pass.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergTableName.java
raunaqmorarka: Fair enough, keeping `PARTITIONS` in `TableType` with `isSystemView` works for now. Can revisit the taxonomy when more system views are added.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergReadVersionedTable.java
raunaqmorarka: The narrowed analyzer rejection (`isBranchVersionReference`) is sufficient for this PR.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -3462,6 +3493,12 @@ public boolean isView(ConnectorSession session, SchemaTableName viewName)
     @Override
     public Optional<ConnectorViewDefinition> getView(ConnectorSession session, SchemaTableName viewName)
     {
+        Optional<ConnectorViewDefinition> systemView = getRawSystemView(session, viewName);
+
+        if (systemView.isPresent()) {
```
raunaqmorarka, line 3498: The `pt (suite-iceberg)` CI failure is a redirect problem.

**Root cause:** `HiveMetadata.redirectTable` redirects `hive.schema."t$partitions"` to `iceberg.schema."t$partitions"`. `getRedirectionAwareTableHandle` resolves the redirect then calls `getTableHandle` on the Iceberg target — which now returns null since `$partitions` is a view. Result: `TABLE_REDIRECTION_ERROR` ("target table does not exist").

**Why the view path doesn't help:** `StatementAnalyzer.visitTable` calls `getView` with the *source* name first. Hive has no view for `$partitions`, so it returns empty. We fall through to `getRedirectionAwareTableHandle` which is table-only.

**Fix:** Add `getRedirectionAwareView` on `Metadata`, mirroring `getRedirectionAwareTableHandle`. It reuses the existing private `getRedirectedTableName` to resolve the chain, then calls `getView` on the target. Return type carries both the view definition and the redirected name (new record similar to `RedirectionAwareTableHandle` but with `Optional<ViewDefinition>`). `StatementAnalyzer.visitTable` calls this instead of plain `getView` and uses the target name for access control — consistent with table redirects at line ~2416.

tbaeg: Thank you for the detailed explanation! I left this as my last fix since it was the least obvious one for me.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
```diff
@@ -0,0 +1,183 @@
...
+                   WHEN %1$s IS NULL AND %2$s IS NULL AND %3$s IS NULL THEN NULL
+                   ELSE ROW(%1$s, %2$s, %3$s, %4$s)
+               END
+               """
+                .formatted(min, max, nullCount, nanCount);
+    }
+
+    private static String buildColumnRowType(String columnName, String trinoTypeDisplayName)
+    {
+        return "%s ROW(min %2$s, max %2$s, null_count BIGINT, nan_count BIGINT)"
```
sopel39, line 150: Manual `"\"" + columnName + "\""` quoting does not escape embedded ". Iceberg permits column names with double quotes; the generated view SQL is then malformed. Reuse `IcebergUtil.quotedName(columnName)` (which is being made public in this PR anyway).

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
```diff
@@ -193,13 +205,27 @@ public static Type getColumnType(String columnName, TypeManager typeManager)
...
-                 UPPER_BOUNDS_COLUMN_NAME -> new MapType(INTEGER, VARCHAR, typeManager.getTypeOperators());
             case KEY_METADATA_COLUMN_NAME -> VARBINARY;
             case SPLIT_OFFSETS_COLUMN_NAME -> new ArrayType(BIGINT);
             case EQUALITY_IDS_COLUMN_NAME -> new ArrayType(INTEGER);
             case READABLE_METRICS_COLUMN_NAME -> typeManager.getType(new TypeSignature(JSON));
             default -> throw new IllegalArgumentException("Unexpected value: " + columnName);
         };
     }
+
+    private static Optional<Type> buildBoundsType(Schema icebergSchema, TypeManager typeManager)
```
sopel39, line 216: lower_bounds / upper_bounds going from `map(integer, varchar)` to `row(...)` will silently break user queries (`lower_bounds[1], CAST(... AS JSON)`, etc.). The PR description / release notes should call this out and the iceberg.md table needs updating (which still says `map(INTEGER, BIGINT)` - already wrong before this PR, doubly wrong after).

core/trino-main/src/main/java/io/trino/metadata/RedirectionAwareView.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.Optional;
+
+import static java.util.Objects.requireNonNull;
+
+public record RedirectionAwareView(
+        Optional<ViewDefinition> viewDefinition,
+        // the target relation name after redirection. Optional.empty() if the view is not redirected.
+        Optional<QualifiedObjectName> redirectedTableName)
+{
+    public static RedirectionAwareView withRedirectionTo(QualifiedObjectName redirectedTableName, ViewDefinition viewDefinition)
```
sopel39: I don't think `withRedirectionTo` it's used

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -889,21 +892,43 @@ private Optional<SystemTable> getRawSystemTable(ConnectorSession session, Schema
...
+    private Optional<ConnectorViewDefinition> getRawSystemView(ConnectorSession session, SchemaTableName tableName)
+    {
+        Optional<BaseTable> baseTable = getRawBaseTable(session, tableName);
+        if (baseTable.isEmpty()) {
+            return Optional.empty();
+        }
+        BaseTable table = baseTable.get();
+        TableType tableType = IcebergTableName.tableTypeFrom(tableName.getTableName());
+        return switch (tableType) {
+            case PARTITIONS -> Optional.ofNullable(new PartitionsView(typeManager, table, catalogName.toString(), tableName.getSchemaName(), tableNameFrom(tableName.getTableName())).get());
```
sopel39: `Optional.ofNullable` -> `Optional.of`

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
```diff
@@ -0,0 +1,183 @@
...
+import static com.google.common.collect.ImmutableSet.toImmutableSet;
+import static io.trino.plugin.iceberg.IcebergUtil.getIdentityPartitions;
+import static io.trino.plugin.iceberg.TypeConverter.toTrinoType;
+import static io.trino.plugin.iceberg.util.SystemTableUtil.getAllPartitionFields;
+import static io.trino.plugin.iceberg.util.SystemTableUtil.getPartitionColumnType;
+import static io.trino.spi.type.BigintType.BIGINT;
+import static java.util.Objects.requireNonNull;
+
+public class PartitionsView
+        implements Supplier<ConnectorViewDefinition>
```
sopel39: I don't think `Supplier` is usefule here. Better to introduce
```
public static ConnectorViewDefinition create(TypeManager typeManager, Table icebergTable, String catalogName, String schemaName, String tableName)
```
method

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
```diff
@@ -348,6 +345,48 @@ private void writePartitionColumns(ContentFile<?> contentFile)
...
+            return;
+        }
+
+        if (bounds == null) {
+            boundsBuilder.appendNull();
+            return;
+        }
+
+        List<RowType.Field> fields = rowType.getFields();
+        // start a new lower/upper bounds entry
```
sopel39: nit: these comments are obvious, remove

sopel39 (commented):
mostly lgtm % comments

Consider adding more test coverage
*  test for the column-rename / schema-evolution scenario, which is the central justification for keying by field ID

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
osscm: +1 , 

This changes lower_bounds/upper_bounds in `$files` from `map(integer, varchar)` to a per-table `row("fieldId" type, ...)`. This is a breaking public API change — any user query doing lower_bounds[1] or element_at(lower_bounds, 1) will break at runtime. The ROW type also varies per table
   schema, which makes generic tooling against `$files` impossible.

osscm: May be its better to have a separate PR to merge  PR for files with explicit release notes, migration guidance will be a good option, 
And I'm thinking another option - old map(integer, varchar) column should remain alongside the new typed column during a deprecation window, or we can keep both.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMetastoreAccessOperations.java
```diff
@@ -384,7 +384,7 @@ public void testSelectSystemTable()
         // select from $partitions
         assertMetastoreInvocations("SELECT * FROM \"test_select_snapshots$partitions\"",
                 ImmutableMultiset.<MetastoreMethod>builder()
-                        .addCopies(GET_TABLE, 1)
+                        .addCopies(GET_TABLE, 2)
```
osscm, line 387: 2 -> is to show that we have 2 HMS/Catalog calls now? should we explain it in the comments and explain that when there are a lot Partition/Catalog based queries, load will be 2x?

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -2055,6 +2055,19 @@ public RedirectionAwareTableHandle getRedirectionAwareTableHandle(Session sessio
         throw new TrinoException(TABLE_REDIRECTION_ERROR, format("Table '%s' redirected to '%s', but the target table '%s' does not exist", tableName, targetTableName, targetTableName));
     }
 
+    @Override
+    public RedirectionAwareView getRedirectionAwareView(Session session, QualifiedObjectName viewName)
+    {
+        QualifiedObjectName targetViewName = getRedirectedTableName(session, viewName, Optional.empty(), Optional.empty());
```
osscm, line 2061: will this could be a problem for the connectors not handle `$-suffixed` names?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
raunaqmorarka: I'm aware that its a breaking change which needs documentation. But the new column type seems more user friendly for writing queries to me. I would like to hear more about what "generic tooling" the change in type would get in the way of. I'm fine with keeping both types if there is a justification for it.

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -897,6 +897,11 @@ default boolean isMaterializedView(Session session, QualifiedObjectName viewName
      */
     RedirectionAwareTableHandle getRedirectionAwareTableHandle(Session session, QualifiedObjectName tableName, Optional<TableVersion> startVersion, Optional<TableVersion> endVersion);
 
+    /**
+     * Get the target view after performing redirection.
+     */
+    RedirectionAwareView getRedirectionAwareView(Session session, QualifiedObjectName viewName);
```
raunaqmorarka, line 903: Please put `getRedirectionAwareView` related changes in separate preparatory commit

core/trino-main/src/main/java/io/trino/metadata/RedirectionAwareView.java
osscm: +1

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/FilesTable.java
tbaeg: Since the intent is to use the new typing for downstream system views, I think this yields a better experience for system view building and general query. The typing for `$partitions` also sets a precedent for dynamic `ROW(...)` in it's `partition` and `data` columns. It is more consistent for `$files` to also provide a dynamic `ROW(...)` type.

For dynamic access/generic tooling, `readable_metrics` seems like a better candidate we can point people toward.

Release notes highlighting the breaking changes feels reasonable enough. Selfishly, I'd prefer to "rip the band-aid off" instead of going through deprecation cycles.

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
tbaeg: AFAIK, no, it should not be a problem. It should walk through the same code path as the previous implementation; this just handles the redirection case.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMetastoreAccessOperations.java
tbaeg: We could, but not sure how much it matters in practice.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsView.java
tbaeg: I think it's useful in the way the code can be encapsulated and structured. My preference would be to keep it in this form if you don't have a strong opinion on it.

sopel39: Trino generally prefers composition over inheritance. In this case there's a single call site (in IcebergMetadata.java) that immediately invokes `.get()`, so extra abstraction layer doesn't add value.

sopel39 (approved):

osscm (approved):
LGTM

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
```diff
@@ -229,16 +234,8 @@ public SourcePage getNextSourcePage()
...
-                    pageBuilder,
-                    LOWER_BOUNDS_COLUMN_NAME,
-                    contentFile::lowerBounds,
-                    this::writeIntegerVarcharInMap);
-            writeValueOrNull(
-                    pageBuilder,
-                    UPPER_BOUNDS_COLUMN_NAME,
-                    contentFile::upperBounds,
-                    this::writeIntegerVarcharInMap);
+            writeTypedBounds(pageBuilder, LOWER_BOUNDS_COLUMN_NAME, contentFile.lowerBounds());
```
raunaqmorarka, line 236: Please update documentation for `$files` in iceberg.md in this commit

core/trino-main/src/main/java/io/trino/metadata/Metadata.java
```diff
@@ -897,6 +897,11 @@ default boolean isMaterializedView(Session session, QualifiedObjectName viewName
      */
     RedirectionAwareTableHandle getRedirectionAwareTableHandle(Session session, QualifiedObjectName tableName, Optional<TableVersion> startVersion, Optional<TableVersion> endVersion);
 
+    /**
+     * Get the target view after performing redirection.
```
raunaqmorarka, line 901: This commit is missing tests for view redirection

tbaeg: Tests added.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
tbaeg: Done.

sopel39:
@raunaqmorarka do you have any more comments? @electrum ?

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMetastoreAccessOperations.java
raunaqmorarka: The second GET_TABLE is avoidable.

**Call 1 (analysis, wasted):** `getRedirectionAwareView` → `getView` → `getViewInternal` → `getCatalogHandle` probes `SystemTablesMetadata.getTableHandle("t$partitions")` → `getSystemTable` → `getRawSystemTable` → `getRawBaseTable` → `catalog.loadTable("t")`. The PARTITIONS case returns empty in `getRawSystemTable` (moved to view), so the probe is wasted but `loadTable` already fired. The subsequent `IcebergMetadata.getView` → `getRawSystemView` → `getRawBaseTable` is a cache hit on the same instance.

**Call 2 (execution, inherent):** The PartitionsView body references `"t$files"`. Executing the view scans `$files` in a separate connector transaction (fresh cache) → `getRawBaseTable` → `catalog.loadTable("t")` again.

Fix: guard `getRawSystemTable` with `isSystemView` so the system-table probe short-circuits before loading the base table:

```java
if (!isIcebergTableName(...) || isDataTable(...) || isMaterializedViewStorage(...) || isSystemView(...)) {
    return Optional.empty();
}
```

This drops it back to 1 GET_TABLE (only the execution-phase `$files` load). Call 2 is inherent to view-over-system-table and would need cross-transaction caching to eliminate.

raunaqmorarka (commented):
lgtm % minor comment about avoiding GET_TABLE call

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergMetastoreAccessOperations.java
tbaeg: I tried adding this guard and it did not prevent 2 `GET_TABLE` calls. 

Maybe I'm missing something, but there is one call is for constructing the `$partitions` definition and another when querying `$files`. I'm not sure how we'd get around this.

This is also inline with the other view related tests like: `testSelectFromView` or `testSelectFromViewWithFilter`.

raunaqmorarka: You're right, both calls are inherent. `PartitionsView.create` needs the Iceberg `Table` to build the view SQL (schema, partition spec), so the analysis-time `loadTable` is unavoidable. This matches the `testSelectFromView` / `testSelectFromViewWithFilter` pattern (both also expect 2 `GET_TABLE` calls).

raunaqmorarka:
@coderabbitai full review

core/trino-main/src/main/java/io/trino/metadata/MetadataManager.java
```diff
@@ -2055,6 +2055,19 @@ public RedirectionAwareTableHandle getRedirectionAwareTableHandle(Session sessio
...
+        Optional<ViewDefinition> view = getView(session, targetViewName);
+
+        if (targetViewName.equals(viewName)) {
+            return RedirectionAwareView.noRedirection(view);
+        }
+
+        return new RedirectionAwareView(view, Optional.of(targetViewName));
+    }
```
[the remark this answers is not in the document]
tbaeg: I may not be understanding the AI review, but having an empty view is a valid case. It is saying that "this is not a view, but a table, fall through to table resolution during redirection".

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -2395,11 +2396,16 @@ else if (shouldFailWhenStale(materializedViewDefinition)) {
...
             if (optionalView.isPresent()) {
-                analysis.addEmptyColumnReferencesForTable(accessControl, session.getIdentity(), name, getBranchName(table));
-                return createScopeForView(table, name, scope, optionalView.get());
+                if (table.getQueryPeriod().isPresent() && !isBranchVersionReference(table)) {
+                    throw semanticException(NOT_SUPPORTED, table, "Views do not support versioning");
+                }
+
+                QualifiedObjectName targetViewName = viewRedirection.redirectedTableName().orElse(name);
+                analysis.addEmptyColumnReferencesForTable(accessControl, session.getIdentity(), targetViewName, getBranchName(table));
+                return createScopeForView(table, targetViewName, scope, optionalView.get());
```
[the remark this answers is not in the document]
tbaeg: `FOR VERSION AS OF $branch` is a valid case that needs to be supported.

docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1551,13 +1551,13 @@ The output of the query has the following columns:
...
+  - `row(...)`
   - Mapping between the Iceberg column ID and its corresponding lower bound in
-    the file.
+    the file (i.e. - `ROW("1" DATE, "2", NUMBER)`).
 * - `upper_bounds`
-  - `map(INTEGER, BIGINT)`
+  - `row(...)`
   - Mapping between the Iceberg column ID and its corresponding upper bound in
-    the file.
+    the file (i.e. - `ROW("1" DATE, "2", NUMBER)`).
```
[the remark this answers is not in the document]
tbaeg: Updated.

plugin/trino-lakehouse/src/main/java/io/trino/plugin/lakehouse/LakehouseMetadata.java
```diff
@@ -681,12 +682,20 @@ public Map<SchemaTableName, ConnectorViewDefinition> getViews(ConnectorSession s
...
     }
 
     @Override
     public boolean isView(ConnectorSession session, SchemaTableName viewName)
     {
+        if (isIcebergTableName(viewName.getTableName()) && isSystemView(viewName.getTableName())) {
+            return icebergMetadata.isView(session, viewName);
+        }
+
         return hiveMetadata.isView(session, viewName);
```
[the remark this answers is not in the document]
tbaeg: This is targeting system view table(s) and is resolved the same way as `LakehouseMetadata.getTableHandle`. Thoughts on this suggestion? @raunaqmorarka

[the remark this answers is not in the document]
raunaqmorarka: The current code is correct. `t$partitions` is syntactically an Iceberg system view name -- Hive has no concept of `$partitions` views, so falling back to Hive is pointless. This matches the `getTableHandle` pattern for `isMaterializedViewStorage` which also short-circuits to Iceberg without Hive fallback.

raunaqmorarka (approved):

PR #29040, opened 2026-04-08
DeltaLake support `optimize` output metrics in result

### LINKED ISSUE #28999: Return execution metrics while running the `OPTIMIZE` command in Delta Lake
We can return metrics such as rewritten data files, removed deletion vectors, added data files in Delta Lake OPTIMIZE procedure. 

https://github.com/trinodb/trino/blob/11bf3c068037352ed9ee6fb8293154c7b9416452/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java#L2940-L2950

* Follow-up of https://github.com/trinodb/trino/pull/28992

Max-Cheng:
## Description

Allow distributed table procedures to output metrics in result.
DeltaLake OPTIMIZE procedure returns the output below:

```
trino> ALTER TABLE deltalake.tpch.region EXECUTE optimize;
ALTER TABLE EXECUTE
        metric_name             | metric_value
--------------------------------+--------------
 rewritten_data_files_count     |            1
 removed_deletion_vectors_count |            1
 added_data_files_count         |            2
(3 rows)
```

- Fixes https://github.com/trinodb/trino/issues/28999

## Release notes

```markdown
## Delta Lake
* Add support for execution metrics while running the `optimize` procedure. ({issue}`28992`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/delta-lake.md
+++ b/docs/src/main/sphinx/connector/delta-lake.md
@@ -731,6 +731,14 @@ EXECUTE <alter-table-execute>`.
 ```{include} optimize.fragment
 ```
 
+```text
+        metric_name             | metric_value
+--------------------------------+--------------
+ rewritten_data_files_count     |            1
+ removed_deletion_vectors_count |            1
+ added_data_files_count         |            2
+```
+
 Use a `WHERE` clause with [metadata columns](delta-lake-special-columns) to filter
 which files are optimized.
 
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
@@ -2943,13 +2943,12 @@ public class DeltaLakeMetadata
         DeltaLakeTableExecuteHandle executeHandle = (DeltaLakeTableExecuteHandle) tableExecuteHandle;
         switch (executeHandle.procedureId()) {
             case OPTIMIZE:
-                finishOptimize(session, executeHandle, fragments, splitSourceInfo);
-                return ImmutableMap.of();
+                return finishOptimize(session, executeHandle, fragments, splitSourceInfo).toMap();
         }
         throw new IllegalArgumentException("Unknown procedure '" + executeHandle.procedureId() + "'");
     }
 
-    private void finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandle executeHandle, Collection<Slice> fragments, List<Object> splitSourceInfo)
+    private OptimizeResult finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandle executeHandle, Collection<Slice> fragments, List<Object> splitSourceInfo)
     {
         DeltaTableOptimizeHandle optimizeHandle = (DeltaTableOptimizeHandle) executeHandle.procedureHandle();
         String tableLocation = executeHandle.tableLocation();
@@ -2958,6 +2957,12 @@ public class DeltaLakeMetadata
         Set<DeltaLakeScannedDataFile> scannedDataFiles = splitSourceInfo.stream()
                 .map(DeltaLakeScannedDataFile.class::cast)
                 .collect(toImmutableSet());
+
+        // delete vector
+        Set<String> filesToDelete = scannedDataFiles.stream()
+                .map(scannedDataFile -> scannedDataFile.deletionVector().map(DeletionVectorEntry::uniqueId))
+                .flatMap(Optional::stream)
+                .collect(Collectors.toSet());
 
         // files to be added
         List<DataFileInfo> dataFileInfos = fragments.stream()
@@ -3001,6 +3006,19 @@ public class DeltaLakeMetadata
                 cleanupFailedWrite(session, optimizeHandle.getCredentialsHandle(), dataFileInfos);
             }
             throw new TrinoException(DELTA_LAKE_BAD_WRITE, "Failed to write Delta Lake transaction log entry", e);
+        }
+        return new OptimizeResult(scannedDataFiles.size(), filesToDelete.size(), dataFileInfos.size());
+    }
+
+    private record OptimizeResult(long rewrittenDataFiles, long removedDeleteFiles, long addedDataFiles)
+    {
+        Map<String, Long> toMap()
+        {
+            return ImmutableMap.<String, Long>builder()
+                    .put("rewritten_data_files_count", rewrittenDataFiles)
+                    .put("removed_delete_files_count", removedDeleteFiles)
+                    .put("added_data_files_count", addedDataFiles)
+                    .buildOrThrow();
         }
     }
 
--- a/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
+++ b/plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
@@ -3153,15 +3153,12 @@ public class DeltaLakeMetadata
     {
         DeltaLakeTableExecuteHandle executeHandle = (DeltaLakeTableExecuteHandle) tableExecuteHandle;
         return switch (executeHandle.procedureId()) {
-            case OPTIMIZE -> {
-                finishOptimize(session, executeHandle, fragments, splitSourceInfo);
-                yield ImmutableMap.of();
-            }
+            case OPTIMIZE -> finishOptimize(session, executeHandle, fragments, splitSourceInfo);
             default -> throw new IllegalArgumentException("Unknown procedure '" + executeHandle.procedureId() + "'");
         };
     }
 
-    private void finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandle executeHandle, Collection<Slice> fragments, List<Object> splitSourceInfo)
+    private Map<String, Long> finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandle executeHandle, Collection<Slice> fragments, List<Object> splitSourceInfo)
     {
         DeltaTableOptimizeHandle optimizeHandle = (DeltaTableOptimizeHandle) executeHandle.procedureHandle();
         String tableLocation = executeHandle.tableLocation();
@@ -3215,6 +3212,16 @@ public class DeltaLakeMetadata
             }
             throw new TrinoException(DELTA_LAKE_BAD_WRITE, "Failed to write Delta Lake transaction log entry", e);
         }
+        long removedDeleteFiles = scannedDataFiles.stream()
+                .map(file -> file.deletionVector().map(DeletionVectorEntry::uniqueId))
+                .flatMap(Optional::stream)
+                .collect(toImmutableSet())
+                .size();
+        return ImmutableMap.<String, Long>builder()
+                .put("rewritten_data_files_count", (long) scannedDataFiles.size())
+                .put("removed_deletion_vectors_count", removedDeleteFiles)
+                .put("added_data_files_count", (long) dataFileInfos.size())
+                .buildOrThrow();
     }
 
     private long commitOptimizeOperation(
--- a/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
+++ b/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
@@ -385,17 +385,14 @@ public abstract class BaseDeltaLakeConnectorSmokeTest
         String tableLocation = getLocationForTable(bucketName, tableName);
         assertUpdate("CREATE TABLE " + tableName + " (key integer, value varchar) WITH (location = '" + tableLocation + "')");
         try {
-            // DistributedQueryRunner sets node-scheduler.include-coordinator by default, so include coordinator
-            int workerCount = getQueryRunner().getNodeCount();
-
             assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'one')", 1);
 
             for (int i = 0; i < 3; i++) {
                 Set<String> initialFiles = getActiveFiles(tableName);
-                computeActual("ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+                computeActual(withSingleWriterPerTask(getSession()), "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
                 Set<String> filesAfterOptimize = getActiveFiles(tableName);
                 assertThat(filesAfterOptimize)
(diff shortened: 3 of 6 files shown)


Max-Cheng:
@ebyhr  OMG your response speed is so fast!

docs/src/main/sphinx/connector/delta-lake.md
```diff
@@ -714,6 +714,13 @@ EXECUTE <alter-table-execute>`.
 
 ```{include} optimize.fragment
 ```
```
findinpath, line 732: `DeltaLake support `optimzie` output metrics in result`

typo `optimzie`  (change to `optimze`) in commit message.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -2943,13 +2943,12 @@ public Map<String, Long> finishTableExecute(ConnectorSession session, ConnectorT
...
             case OPTIMIZE:
-                finishOptimize(session, executeHandle, fragments, splitSourceInfo);
-                return ImmutableMap.of();
+                return finishOptimize(session, executeHandle, fragments, splitSourceInfo);
         }
         throw new IllegalArgumentException("Unknown procedure '" + executeHandle.procedureId() + "'");
     }
 
-    private void finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandle executeHandle, Collection<Slice> fragments, List<Object> splitSourceInfo)
+    private Map<String, Long> finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandle executeHandle, Collection<Slice> fragments, List<Object> splitSourceInfo)
```
findinpath, line 3161: nit: I'd recommend rather returning `OptimizeResult` and doing the `toMap` in the calling context instead.
If you take the suggestion, pls do a preparatory commit as well for `IcebergMetadata`

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
```diff
@@ -385,17 +385,14 @@ public void testOptimizeRewritesTable()
...
         try {
-            // DistributedQueryRunner sets node-scheduler.include-coordinator by default, so include coordinator
-            int workerCount = getQueryRunner().getNodeCount();
-
             assertUpdate("INSERT INTO " + tableName + " VALUES (1, 'one')", 1);
 
             for (int i = 0; i < 3; i++) {
                 Set<String> initialFiles = getActiveFiles(tableName);
-                computeActual("ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+                computeActual(withSingleWriterPerTask(getSession()), "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
```
findinpath: Let's rather do a preparatory commit that makes use of `withSingleWriterPerTask` to explain that we're aiming for stable results.

findinpath (commented):
LGTM % nits

docs/src/main/sphinx/connector/delta-lake.md
```diff
@@ -714,6 +714,13 @@ EXECUTE <alter-table-execute>`.
...
+```text
+        metric_name         | metric_value
+----------------------------+--------------
+ rewritten_data_files_count |            2
+ removed_delete_files_count |            0
```
chenjian2664: We can remove it if it is guaranteed to always be 0

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3002,6 +3001,20 @@ private void finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandl
             }
             throw new TrinoException(DELTA_LAKE_BAD_WRITE, "Failed to write Delta Lake transaction log entry", e);
         }
+        // No files should be removed (this is VACUUM's job)
+        return new OptimizeResult(scannedDataFiles.size(), 0, dataFileInfos.size()).toMap();
```
chenjian2664: Just inline the result - using `ImmutableMap.builder`

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
chenjian2664: but, it seems an un-related to the change, is it required by the current(pr's) change?

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
```diff
@@ -2046,7 +2058,15 @@ public void testOptimizeWithEnforcedRepartitioning()
             Set<String> initialFiles = getActiveFiles(tableName, currentSession);
             assertThat(initialFiles).hasSize(10);
 
-            computeActual(currentSession, "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+            MaterializedResult optimizeResult = computeActual(withSingleWriterPerTask(getSession()), "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE");
+
+            Map<String, Long> metrics = optimizeResult.getMaterializedRows().stream()
+                    .collect(ImmutableMap.toImmutableMap(
```
chenjian2664: static import

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
Max-Cheng: @chenjian2664  @findinpath yep, this is an un-related change: to avoid flaky test results, I’ll split these commits into separate ones.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
ebyhr: We should update DeltaLakeSplitSource to report fully scanned deletion vector files instead of hard-coded 0.

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
ebyhr: Let's avoid touching unrelated modules in this PR. There is no need to modify IcebergMetadata in my opinion.

Max-Cheng: @ebyhr Maybe it's another task?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3002,6 +3007,19 @@ private void finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandl
             }
             throw new TrinoException(DELTA_LAKE_BAD_WRITE, "Failed to write Delta Lake transaction log entry", e);
         }
+        return new OptimizeResult(scannedDataFiles.size(), filesToDelete.size(), dataFileInfos.size());
+    }
+
+    private record OptimizeResult(long rewrittenDataFiles, long removedDeleteFiles, long addedDataFiles)
```
ebyhr: We don't need this record in this connector. Please remove. 

Iceberg connector uses a record class because there are 2 return in `finishOptimize` method.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/BaseDeltaLakeConnectorSmokeTest.java
```diff
@@ -2005,21 +1999,74 @@ public void testOptimizeWithPartitionedTable()
...
+                    .hasSize(7);
             assertThat(getAllDataFilesFromTableDirectory(tableName)).isEqualTo(union(initialFiles, updatedFiles));
         }
         finally {
             assertUpdate("DROP TABLE " + tableName);
         }
     }
 
+    @Test
+    public void testOptimizeWithDeletionVectors()
```
ebyhr: BaseDeltaLakeConnectorSmokeTest is used for running tests on several storages. Please consider reverting changes in this class, and updating TestDeltaLakeConnectorTest instead.

ebyhr (commented):
Please squash commits into one. No need to separate a commit for documentation.

Max-Cheng:
> Please squash commits into one. No need to separate a commit for documentation.

I will squash them after CI completes.

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -934,6 +934,94 @@ public void testAddColumnAndOptimize()
...
+            assertThat(initialFiles).hasSize(6);
+            assertUpdate("DELETE FROM " + tableName + " WHERE nationkey < 5", 30);
+
+            Session singleWriterSession = Session.builder(getSession())
+                    .setSystemProperty("task_min_writer_count", "1")
+                    .build();
+            assertUpdate(
+                    singleWriterSession,
+                    "ALTER TABLE " + tableName + " EXECUTE OPTIMIZE",
+                    "VALUES ('rewritten_data_files_count', 6), ('removed_delete_files_count', 6), ('added_data_files_count', 1)");
```
ebyhr: > ('removed_delete_files_count', 6)

This result looks misleading. Deletion Vector files still exist on the storage, right?

Max-Cheng: @ebyhr  Yep. The DV files remain in storage. This metric counts DVs removed from the active table state. Would removed_deletion_vectors_count make the semantics clearer?

ebyhr: Yes. I don't think we should use "delete files" keyword in Delta Lake connector. That's a concept of Iceberg.

chenjian2664 (approved):
It looks like the current tests only cover deletes that apply to the entire table. Could you update that partition case where the delete affects only a subset of the partitions?

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -934,6 +934,97 @@ public void testAddColumnAndOptimize()
...
+    {
+        try (TestTable table = newTrinoTable(
+                "test_optimize_deletion_vectors",
+                "WITH (deletion_vectors_enabled = true) AS SELECT * FROM tpch.tiny.nation")) {
+            String tableName = table.getName();
+            assertUpdate("INSERT INTO " + tableName + " SELECT * FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT * FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT * FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT * FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT * FROM tpch.tiny.nation", 25);
```
chenjian2664, line 980: nit: no need to write 5 times, just 1 or 2 times is enough for test

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -934,6 +934,97 @@ public void testAddColumnAndOptimize()
...
+        try (TestTable table = newTrinoTable(
+                "test_optimize_partitioned_deletion_vectors",
+                "WITH (deletion_vectors_enabled = true, partitioned_by = ARRAY['regionkey']) AS SELECT nationkey, regionkey FROM tpch.tiny.nation")) {
+            String tableName = table.getName();
+            assertUpdate("INSERT INTO " + tableName + " SELECT nationkey, regionkey FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT nationkey, regionkey FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT nationkey, regionkey FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT nationkey, regionkey FROM tpch.tiny.nation", 25);
+            assertUpdate("INSERT INTO " + tableName + " SELECT nationkey, regionkey FROM tpch.tiny.nation", 25);
+
```
chenjian2664, line 1009: same

plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeConnectorTest.java
```diff
@@ -934,6 +934,97 @@ public void testAddColumnAndOptimize()
...
+            Set<String> initialFiles = getActiveFiles(tableName);
+            assertThat(initialFiles).hasSize(30);
+            assertUpdate("DELETE FROM " + tableName + " WHERE regionkey = 1 AND nationkey < 5", 18);
+
+            Session singleWriterSession = Session.builder(getSession())
+                    .setSystemProperty("task_min_writer_count", "1")
+                    .build();
```
chenjian2664, line 1016: Why does this property need to be pinned to `1`?

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMetadata.java
```diff
@@ -3215,6 +3212,16 @@ private void finishOptimize(ConnectorSession session, DeltaLakeTableExecuteHandl
             }
             throw new TrinoException(DELTA_LAKE_BAD_WRITE, "Failed to write Delta Lake transaction log entry", e);
         }
+        long removedDeleteFiles = scannedDataFiles.stream()
```
chenjian2664, line 3215:
```suggestion
        long removedDeleteFiles = scannedDataFiles.stream()
                .map(file -> file.deletionVector().map(DeletionVectorEntry::uniqueId))
                .flatMap(Optional::stream)
                .distinct()
                .count();
```

PR #29046, opened 2026-04-08
Add VARIANT_JSON and VARIANT_BINARY client capabilities

dain:
## Description

Add support for `VARIANT` in clients.  There are three supported scenarios, based on the settings of client capabilities:
1. nothing set: return type `JSON` with data encoded as `JSON` (legacy clients)
2. `VARIANT`: return type `VARIANT` with data encoded as `JSON`
3. `VARIANT_BINARY`: return type `VARIANT` with data in standard binary variant format.

The cli uses `VARIANT`, and the JDBC driver uses `VARIANT_BINARY`.

The JDBC client uses binary `VARIANT` directly without lowering to JSON and losing all type information.  `VARIANT` columns can be fetched with:
- `rs.getObject(column)` or `rs.getObject(column, Object.class)` to get the corresponding Java object representation. This is the same as `variant.toObject`
- `rs.getObject(column, io.trino.jdbc.Variant.class)` to get an `io.trino.jdbc.Variant`
- `rs.getObject(column, Map.class)` Same as `getObject(Column)` but only works when the result would be a Map. Throws an exception if the variant is not an object.
- `rs.getObject(column, List.class)` Same as `getObject(Column)` but only works when the result would be a Array. Throws an exception if the variant is not an array.
- `rs.getString(column)` or `rs.getObject(column, String.class)` to get the corresponding JSON representation. This is the same as `variant.toJson`, which is the same as the `VARIANT_JSON` encoding.

When mapping to Java object representation, `VARIANT` values map to Java objects as follows:

| VARIANT value | Java object |
| --- | --- |
| null | `null` |
| boolean | `Boolean` |
| int8 | `Byte` |
| int16 | `Short` |
| int32 | `Integer` |
| int64 | `Long` |
| float | `Float` |
| double | `Double` |
| decimal | `BigDecimal` |
| string | `String` |
| binary | `byte[]` |
| date | `LocalDate` |
| time(ntz) | `LocalTime` |
| timestamp(utc) | `Instant` |
| timestamp(ntz) | `LocalDateTime` |
| uuid | `UUID` |
| array | `List<Object>` |
| object | `Map<String, Object>` |

The `io.trino.jdbc.Variant` class is a minimal implementation of VARIANT for decoding data.  This supports getting the raw underlying byte arrays for the value and metdata is the client wants to interact with the raw data (storing in a separate system directly without reencoding).  The SPI Variant class cannot be used because it is compiled with Java 25.

## Release notes

```markdown
## CLI, JDBC
* Add support for VARIANT type. ({issue}`29046`)
```

### DIFF
--- a/client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
+++ b/client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
@@ -34,6 +34,12 @@ public enum ClientCapabilities
     NUMBER,
 
     /**
+     * Whether client supports the `VARIANT` type.
+     * When this capability is not set, the server returns `json` for `VARIANT` columns.
+     */
+    VARIANT,
+
+    /**
      * Whether clients support the session authorization set/reset feature
      */
     SESSION_AUTHORIZATION;
--- a/client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
+++ b/client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
@@ -35,9 +35,14 @@ public enum ClientCapabilities
 
     /**
      * Whether client supports the `VARIANT` type.
-     * When this capability is not set, the server returns `json` for `VARIANT` columns.
+     * When neither this capability nor `VARIANT_BINARY` are set, the server returns `json` for `VARIANT` columns.
      */
     VARIANT,
+
+    /**
+     * Whether client supports the `VARIANT` type encoded as a binary payload on the wire.
+     */
+    VARIANT_BINARY,
 
     /**
      * Whether clients support the session authorization set/reset feature
--- a/client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
+++ b/client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
@@ -69,7 +69,6 @@ import static java.net.HttpURLConnection.HTTP_BAD_METHOD;
 import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import static java.net.HttpURLConnection.HTTP_OK;
 import static java.net.HttpURLConnection.HTTP_UNAUTHORIZED;
-import static java.util.Arrays.stream;
 import static java.util.Objects.requireNonNull;
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
 
@@ -132,13 +131,16 @@ class StatementClientV1
                 .map(Optional::get)
                 .findFirst();
         this.setOriginalRoles.addAll(session.getOriginalRoles());
-        this.clientCapabilities = Joiner.on(",").join(clientCapabilities.orElseGet(() -> stream(ClientCapabilities.values())
+        Set<String> effectiveClientCapabilities = clientCapabilities.orElseGet(() -> Stream.of(ClientCapabilities.values())
                 .map(Enum::name)
-                .collect(toImmutableSet())));
+                .collect(toImmutableSet()));
+        this.clientCapabilities = Joiner.on(",").join(effectiveClientCapabilities);
         this.compressionDisabled = session.isCompressionDisabled();
         this.heartbeatInterval = session.getHeartbeatInterval().toMillis() * 1_000_000;
 
-        this.resultRowsDecoder = new ResultRowsDecoder(new OkHttpSegmentLoader(requireNonNull(segmentHttpCallFactory, "segmentHttpCallFactory is null")));
+        this.resultRowsDecoder = new ResultRowsDecoder(
+                new OkHttpSegmentLoader(requireNonNull(segmentHttpCallFactory, "segmentHttpCallFactory is null")),
+                effectiveClientCapabilities.contains(ClientCapabilities.VARIANT_BINARY.toString()));
 
         Request request = buildQueryRequest(session, query, session.getEncoding());
         // Pass empty as materializedJsonSizeLimit to always materialize the first response
--- a/client/trino-client/src/main/java/io/trino/client/spooling/encoding/JsonQueryDataDecoder.java
+++ b/client/trino-client/src/main/java/io/trino/client/spooling/encoding/JsonQueryDataDecoder.java
@@ -54,9 +54,9 @@ public class JsonQueryDataDecoder
             implements QueryDataDecoder.Factory
     {
         @Override
-        public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes)
+        public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes, boolean supportsVariantBinary)
         {
-            return new JsonQueryDataDecoder(createTypeDecoders(columns));
+            return new JsonQueryDataDecoder(createTypeDecoders(columns, supportsVariantBinary));
         }
 
         @Override
@@ -70,9 +70,9 @@ public class JsonQueryDataDecoder
             extends Factory
     {
         @Override
-        public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes)
+        public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes, boolean supportsVariantBinary)
         {
-            return new ZstdQueryDataDecoder(super.create(columns, queryAttributes));
+            return new ZstdQueryDataDecoder(super.create(columns, queryAttributes, supportsVariantBinary));
         }
 
         @Override
@@ -86,9 +86,9 @@ public class JsonQueryDataDecoder
             extends Factory
     {
         @Override
-        public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes)
+        public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes, boolean supportsVariantBinary)
         {
-            return new Lz4QueryDataDecoder(super.create(columns, queryAttributes));
+            return new Lz4QueryDataDecoder(super.create(columns, queryAttributes, supportsVariantBinary));
         }
 
         @Override
--- a/client/trino-client/src/test/java/io/trino/client/TestResultRowsDecoder.java
+++ b/client/trino-client/src/test/java/io/trino/client/TestResultRowsDecoder.java
@@ -117,7 +117,7 @@ class TestResultRowsDecoder
     {
         AtomicInteger loaded = new AtomicInteger();
         AtomicInteger acknowledged = new AtomicInteger();
-        try (ResultRowsDecoder decoder = new ResultRowsDecoder(new StaticLoader(loaded, acknowledged))) {
+        try (ResultRowsDecoder decoder = new ResultRowsDecoder(new StaticLoader(loaded, acknowledged), false)) {
             assertThat(eagerlyMaterialize(decoder.toRows(fromSegments(spooledSegment(2), spooledSegment(2)))))
                     .hasSize(4)
                     .containsExactly(ImmutableList.of(2137), ImmutableList.of(1337), ImmutableList.of(2137), ImmutableList.of(1337));
@@ -132,7 +132,7 @@ class TestResultRowsDecoder
     {
         AtomicInteger loaded = new AtomicInteger();
         AtomicInteger acknowledged = new AtomicInteger();
-        try (ResultRowsDecoder decoder = new ResultRowsDecoder(new StaticLoader(loaded, acknowledged))) {
+        try (ResultRowsDecoder decoder = new ResultRowsDecoder(new StaticLoader(loaded, acknowledged), false)) {
             assertThat(eagerlyMaterialize(decoder.toRows(fromSegments(spooledSegment(2), spooledSegment(2)))))
                     .hasSize(4)
                     .containsExactly(ImmutableList.of(2137), ImmutableList.of(1337), ImmutableList.of(2137), ImmutableList.of(1337));
@@ -148,7 +148,7 @@ class TestResultRowsDecoder
                 .mapToObj(Integer::toString)
                 .reduce("[", (a, b) -> a + "[" + b + "],", String::concat) + "[1337]]";
         CountingInputStream stream = new CountingInputStream(new ByteArrayInputStream(data.getBytes(UTF_8)));
-        try (ResultRowsDecoder decoder = new ResultRowsDecoder(loaderFromStream(stream))) {
+        try (ResultRowsDecoder decoder = new ResultRowsDecoder(loaderFromStream(stream), false)) {
             Iterator<List<Object>> iterator = decoder.toRows(fromSegments(spooledSegment(2501))).iterator();
             assertThat(stream.getCount()).isEqualTo(0);
             iterator.next();
@@ -174,7 +174,7 @@ class TestResultRowsDecoder
     {
         AtomicInteger loaded = new AtomicInteger();
         AtomicInteger acknowledged = new AtomicInteger();
-        try (ResultRowsDecoder decoder = new ResultRowsDecoder(new StaticLoader(loaded, acknowledged))) {
+        try (ResultRowsDecoder decoder = new ResultRowsDecoder(new StaticLoader(loaded, acknowledged), false)) {
             Iterator<List<Object>> iterator = decoder.toRows(fromSegments(spooledSegment(2), spooledSegment(2)))
                     .iterator();
 
--- a/client/trino-jdbc/pom.xml
+++ b/client/trino-jdbc/pom.xml
@@ -19,6 +19,12 @@
     </properties>
 
     <dependencies>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
+
         <dependency>
             <groupId>com.google.errorprone</groupId>
             <artifactId>error_prone_annotations</artifactId>
@@ -99,6 +105,12 @@
         <dependency>
             <groupId>io.opentelemetry</groupId>
             <artifactId>opentelemetry-context</artifactId>
+            <scope>provided</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.gaul</groupId>
+            <artifactId>modernizer-maven-annotations</artifactId>
             <scope>provided</scope>
         </dependency>
 
@@ -184,6 +196,12 @@
         <dependency>
             <groupId>io.airlift</groupId>
             <artifactId>security</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.airlift</groupId>
+            <artifactId>slice</artifactId>
             <scope>test</scope>
         </dependency>
 
--- a/client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
+++ b/client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
@@ -23,10 +23,11 @@ import io.trino.client.ClientTypeSignature;
 import io.trino.client.ClientTypeSignatureParameter;
 import io.trino.client.CloseableIterator;
 import io.trino.client.Column;
+import io.trino.client.EncodedVariant;
 import io.trino.client.IntervalDayTime;
 import io.trino.client.IntervalYearMonth;
-import io.trino.jdbc.ColumnInfo.Nullable;
 import io.trino.jdbc.TypeConversions.NoConversionRegisteredException;
+import jakarta.annotation.Nullable;
 import org.joda.time.DateTimeZone;
 import org.joda.time.LocalDate;
 import org.joda.time.format.DateTimeFormat;
@@ -150,6 +151,7 @@ abstract class AbstractTrinoResultSet
             .put("interval day to second", TrinoIntervalDayTime.class)
             .put("map", Map.class)
             .put("row", Row.class)
+            .put("variant", Object.class)
             .buildOrThrow();
 
     @VisibleForTesting
@@ -200,6 +202,11 @@ abstract class AbstractTrinoResultSet
                         }
                         return result;
                     })
+                    .add("variant", EncodedVariant.class, Object.class, value -> decodeVariant(value).toObject())
+                    .add("variant", EncodedVariant.class, String.class, value -> decodeVariant(value).toJson())
+                    .add("variant", EncodedVariant.class, Variant.class, AbstractTrinoResultSet::decodeVariant)
+                    .add("variant", EncodedVariant.class, Map.class, value -> variantToMap(decodeVariant(value)))
+                    .add("variant", EncodedVariant.class, List.class, value -> variantToList(decodeVariant(value)))
                     .build();
     protected final CloseableIterator<List<Object>> results;
     private final Map<String, Integer> fieldMap;
@@ -676,8 +683,8 @@ abstract class AbstractTrinoResultSet
         return column(columnIndex);
     }
 
-    @jakarta.annotation.Nullable
-    private static Object convertFromClientRepresentation(ClientTypeSignature columnType, @jakarta.annotation.Nullable Object value)
+    @Nullable
+    private static Object convertFromClientRepresentation(ClientTypeSignature columnType, @Nullable Object value)
             throws SQLException
     {
         requireNonNull(columnType, "columnType is null");
@@ -739,6 +746,29 @@ abstract class AbstractTrinoResultSet
     private static TrinoIntervalYearMonth parseIntervalYearMonth(String value)
     {
         return new TrinoIntervalYearMonth(IntervalYearMonth.parseMonths(value));
+    }
+
+    private static Variant decodeVariant(EncodedVariant value)
+    {
+        return Variant.fromBytes(value.getMetadataBytes(), value.getValueBytes());
+    }
+
+    private static Map<?, ?> variantToMap(Variant variant)
+            throws SQLException
+    {
+        if (variant.valueType() != Variant.ValueType.OBJECT) {
+            throw new SQLException("VARIANT is not an object");
+        }
+        return (Map<?, ?>) variant.toObject();
+    }
+
+    private static List<?> variantToList(Variant variant)
+            throws SQLException
+    {
+        if (variant.valueType() != Variant.ValueType.ARRAY) {
+            throw new SQLException("VARIANT is not an array");
+        }
+        return (List<?>) variant.toObject();
     }
 
     private static TrinoIntervalDayTime parseIntervalDayTime(String value)
@@ -1837,7 +1867,12 @@ abstract class AbstractTrinoResultSet
 
         try {
             T converted = TYPE_CONVERSIONS.convert(columnTypeSignature, object, type);
-            verify(converted != null, "Conversion cannot return null for non-null input, as this breaks wasNull()");
+            boolean variantObjectRepresentation = type == Object.class &&
+                    columnTypeSignature.getRawType().equals("variant");
+            // VARIANT null is a non-SQL-null value whose Java object representation is null.
+            verify(
+                    converted != null || variantObjectRepresentation,
+                    "Conversion cannot return null for non-null input, as this breaks wasNull()");
             return converted;
         }
         catch (NoConversionRegisteredException e) {
@@ -1999,7 +2034,7 @@ abstract class AbstractTrinoResultSet
                     .setColumnLabel(column.getName())
                     .setColumnName(column.getName()) // TODO
                     .setColumnTypeSignature(column.getTypeSignature())
-                    .setNullable(Nullable.UNKNOWN)
+                    .setNullable(ColumnInfo.Nullable.UNKNOWN)
                     .setCurrency(false);
             setTypeInfo(builder, column.getTypeSignature());
             list.add(builder.build());
--- a/client/trino-jdbc/src/main/java/io/trino/jdbc/Variant.java
+++ b/client/trino-jdbc/src/main/java/io/trino/jdbc/Variant.java
@@ -1,0 +1,914 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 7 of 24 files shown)


core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -83,22 +83,25 @@ private JsonEncodingUtils() {}
...
     private static final VarcharEncoder VARCHAR_ENCODER = new VarcharEncoder();
     private static final VarbinaryEncoder VARBINARY_ENCODER = new VarbinaryEncoder();
-    private static final VariantEncoder VARIANT_ENCODER = new VariantEncoder();
 
     public static TypeEncoder[] createTypeEncoders(Session session, List<Type> types)
     {
         verify(!types.isEmpty(), "Columns must not be empty");
 
-        boolean supportsParametricDateTime = requireNonNull(session, "session is null")
+        Session querySession = requireNonNull(session, "session is null");
```
findepi:
```suggestion
        _ clientCapabilities = session.getClientCapabilities();
```

findepi (approved):
LGTM except `TestJsonEncodingUtils` which i didn't review. Seek review from @wendigo .

Add a test in `TestJdbcResultSet` too.

ebyhr:
Please don't remove ``` ({issue}`issuenumber`) ``` placeholder from PR description. You should put the PR number when the relevant issue doesn't exist.

dain:
> Please don't remove `` ({issue}`issuenumber`) `` placeholder from PR description. You should put the PR number when the relevant issue doesn't exist.

There is no relevant issue.  Someone modified the message to say it fixes a bug, but there is no bug since variant is unreleased.

findinpath:
Tested with trino cli 468

```

trino> select CAST(JSON '{"a":1,"b":true}' AS VARIANT);
      _col0       
------------------
 {"a":1,"b":true} 
(1 row)

Query 20260409_160644_00026_n6pci, FINISHED, 1 node
http://localhost:8080/ui/query.html?20260409_160644_00026_n6pci
Splits: 1 total, 1 done (100.00%)
CPU Time: 0.0s total,     0 rows/s,     0B/s, 0% active
Per Node: 0.0 parallelism,     0 rows/s,     0B/s
Parallelism: 0.0
Peak Memory: 256B
0.12 [0 rows, 0B] [0 rows/s, 0B/s]

trino> select CAST(42 AS VARIANT);
 _col0 
-------
 42    
(1 row)
```

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
```diff
@@ -39,8 +44,39 @@ public enum ClientCapabilities
      */
     VARIANT_JSON,
 
+    /**
+     * Whether client supports the `VARIANT` type encoded as a binary payload on the wire.
+     * This capability is opt-in, so clients continue to receive the JSON representation by default.
+     */
+    VARIANT_BINARY(false),
```
findinpath: Is this supposed to be mentioned in `docs/src/main/sphinx/develop/client-protocol.md` ?

I'm seeing in `io.trino.jdbc.TrinoConnection#startQuery` that it is being added without any checks.

dain: I don't see `Client-Capabilities` documented anywhere

client/trino-jdbc/src/main/java/io/trino/jdbc/Variant.java
```diff
@@ -0,0 +1,727 @@
...
+import static com.google.common.base.Preconditions.checkState;
+import static java.lang.Math.floorDiv;
+import static java.lang.Math.floorMod;
+import static java.lang.Math.multiplyExact;
+import static java.nio.ByteOrder.LITTLE_ENDIAN;
+import static java.util.Collections.unmodifiableList;
+import static java.util.Collections.unmodifiableMap;
+import static java.util.Objects.requireNonNull;
+
+public final class Variant
```
findepi, line 62: Document

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -200,6 +202,10 @@ abstract class AbstractTrinoResultSet
                         }
                         return result;
                     })
+                    .add("variant", EncodedVariant.class, Variant.class, AbstractTrinoResultSet::decodeVariant)
+                    .add("variant", EncodedVariant.class, Object.class, value -> decodeVariant(value).toObject())
```
findepi, line 205: > `rs.getObject(column, Object.class)` to get the corresponding Java object representation

i don't think this is how `getObject` should work.
i would expect `rs.getObject(column, Object.class)` to return same thing as `rs.getObject(column)` does.

does JDBC spec define this?

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -150,6 +151,7 @@ abstract class AbstractTrinoResultSet
             .put("interval day to second", TrinoIntervalDayTime.class)
             .put("map", Map.class)
             .put("row", Row.class)
+            .put("variant", Variant.class)
```
findepi: When using old JDBC, user gets VARIANT as JSON.
When using new JDBC, if user wants JSON for the VARIANT, what should they do?

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -661,19 +667,24 @@ public Object getObject(int columnIndex)
...
+        Object value = column(columnIndex);
 
         if (columnInfo.getColumnType() == Types.ARRAY) {
             // Array requires special treatment due to element metadata provided by the Array object
             // TODO (https://github.com/trinodb/trino/issues/6049) consider returning List instead
             return getArray(columnIndex);
         }
 
+        if (columnInfo.getColumnTypeSignature().getRawType().equals("variant") && value instanceof String) {
+            return value;
```
findepi: Remove.
This looks like shadowing DEFAULT_OBJECT_REPRESENTATION entry.

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -728,6 +739,10 @@ private static Object convertFromClientRepresentation(ClientTypeSignature column
             }
         }
 
+        if (columnType.getRawType().equals("variant") && value instanceof String) {
+            return value;
+        }
```
findepi: Remove.
This looks like shadowing DEFAULT_OBJECT_REPRESENTATION entry.

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -200,6 +202,10 @@ abstract class AbstractTrinoResultSet
                         }
                         return result;
                     })
+                    .add("variant", EncodedVariant.class, Variant.class, AbstractTrinoResultSet::decodeVariant)
+                    .add("variant", EncodedVariant.class, Object.class, value -> decodeVariant(value).toObject())
+                    .add("variant", EncodedVariant.class, Map.class, value -> variantToMap(decodeVariant(value)))
+                    .add("variant", EncodedVariant.class, List.class, value -> variantToList(decodeVariant(value)))
```
findepi, line 209: I am not convinced about this additional conversions.
They might be useful, but they also prevent as from adding `rs.getObject(col, String.class)` that would return VARIANT-as-JSON data.

Let's not add them in this PR.
User can always invoke a specific method/cast on the Variant object

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -1831,6 +1869,13 @@ public <T> T getObject(int columnIndex, Class<T> type)
         }
 
         ClientTypeSignature columnTypeSignature = columnInfo(columnIndex).getColumnTypeSignature();
+        if (columnTypeSignature.getRawType().equals("variant") && object instanceof String) {
+            if (type.isInstance(object)) {
+                return type.cast(object);
+            }
+            throw new SQLException(format("Cannot convert VARIANT JSON text to %s", type));
+        }
```
findepi: Remove.
This shadows TYPE_CONVERSIONS entry.

client/trino-jdbc/src/test/java/io/trino/jdbc/BaseTestJdbcResultSet.java
```diff
@@ -570,6 +570,66 @@ public void testNumber()
         }
     }
 
+    @Test
+    public void testVariant()
+            throws Exception
+    {
+        try (ConnectedStatement connectedStatement = newStatement()) {
+            checkRepresentation(connectedStatement.getStatement(), "CAST(NULL AS variant)", Types.JAVA_OBJECT, (rs, column) -> {
```
findepi, line 579: does variant have NULL and NULL-value concepts like JSON?

client/trino-jdbc/src/test/java/io/trino/jdbc/BaseTestJdbcResultSet.java
```diff
@@ -570,6 +570,66 @@ public void testNumber()
...
+                assertThat(rs.wasNull()).isTrue();
+                assertThat(rs.getString(column)).isNull();
+                assertThat(rs.wasNull()).isTrue();
+                assertThat(rs.getBytes(column)).isNull();
+                assertThat(rs.wasNull()).isTrue();
+            });
+
+            checkRepresentation(connectedStatement.getStatement(), "CAST(DATE '2021-02-03' AS variant)", Types.JAVA_OBJECT, (rs, column) -> {
+                Variant variant = rs.getObject(column, Variant.class);
+                assertThat(rs.getObject(column)).isInstanceOf(Variant.class);
```
findepi: swap

client/trino-jdbc/src/test/java/io/trino/jdbc/BaseTestJdbcResultSet.java
```diff
@@ -570,6 +570,66 @@ public void testNumber()
...
+                assertThat(variant.valueType()).isEqualTo(Variant.ValueType.DATE);
+                assertThat(variant.getLocalDate()).isEqualTo(LocalDate.of(2021, 2, 3));
+                assertThat(rs.getObject(column, Object.class)).isEqualTo(LocalDate.of(2021, 2, 3));
```
findepi: Perhaps cool, but let's leave it up to a future PR

findepi (dismissed):

client/trino-client/src/test/java/io/trino/client/TestResultRowsDecoder.java
```diff
@@ -75,6 +75,16 @@ public void testJsonNodeMaterialization()
         }
     }
 
+    @Test
+    public void testVariantJsonNodeMaterialization()
```
wendigo: I don't understand purpose of this test

client/trino-client/src/test/java/io/trino/client/TestResultRowsDecoder.java
```diff
@@ -126,6 +136,20 @@ public void testSpooledJsonMaterialization()
         assertThat(acknowledged.get()).isEqualTo(2);
     }
 
+    @Test
```
wendigo: I don't understand purpose of this test - I'd drop these altogether

core/trino-main/src/test/java/io/trino/server/protocol/TestJsonEncodingUtils.java
```diff
@@ -426,6 +433,82 @@ public void testVariantSerialization()
                 .isEqualTo(column(null, "{\"a\":1,\"b\":[true,null]}", null));
     }
 
+    @Test
+    public void testVariantJsonFallbackSerialization()
+            throws IOException
+    {
+        List<TypedColumn> columns = ImmutableList.of(typed("col0", VARIANT));
+        var blockBuilder = VARIANT.createBlockBuilder(null, 3);
```
wendigo: do not use var

core/trino-main/src/test/java/io/trino/server/protocol/TestJsonEncodingUtils.java
```diff
@@ -578,25 +682,46 @@ public static TypedColumn typed(String name, Type type)
...
     {
         ImmutableList.Builder<Column> columns = ImmutableList.builderWithExpectedSize(types.size());
         for (TypedColumn typedColumn : types) {
-            columns.add(createColumn(typedColumn.name(), typedColumn.type(), true, true));
+            columns.add(createColumn(typedColumn.name(), typedColumn.type(), true, true, supportsVariantJson));
         }
         return createDecoder(columns.build());
     }
 
+    private static Session sessionWithoutCapability(String capability)
```
wendigo: use ClientCapability enum instead of a String

client/trino-client/src/main/java/io/trino/client/spooling/encoding/JsonQueryDataDecoder.java
```diff
@@ -56,7 +57,10 @@ public static class Factory
         @Override
         public QueryDataDecoder create(List<Column> columns, DataAttributes queryAttributes)
         {
-            return new JsonQueryDataDecoder(createTypeDecoders(columns));
+            boolean supportsVariantBinary = queryAttributes.getOptional(DataAttribute.VARIANT_ENCODING, String.class)
```
wendigo: Why is this needed? If the client implements variant - it is supported implicitly as client sends `VARIANT` ClientCapability and server responds with VARIANT encoding. This duplicates client capability. Undo

wendigo:
Why do we need a second commit? Seems complex so what's the benefit

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
dain: That isn't true.  None of this code has shipped, so there is no expectation for preexisting behavior.

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
electrum: Agreed, we can remove the `Object.class` conversion and make `getObject(column)` return the object representation (using the table in the PR description)

dain: I removed the `getObject(column, Object.class)` method, and then made `getObject(column)` return the java object representation.

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
dain: Supporting JSON would require that the system make a decision on exactly the JSON representation of temporal data as JSON.  SQL uses one format and everyone else uses ISO8601.  I think it is better to expose as Java object representation and then if the user wants json they can convert as they want.

client/trino-jdbc/src/test/java/io/trino/jdbc/BaseTestJdbcResultSet.java
dain: Yes.  I aded a test for that.

client/trino-client/src/main/java/io/trino/client/spooling/encoding/JsonQueryDataDecoder.java
dain: I renamed the variable.  This is really client is expecting binary variant encodings.

dain:
> Why do we need a second commit? Seems complex so what's the benefit

This allows to use the full power of VARIANT in JDBC.  VARIANT caries type information that JSON does not, and lowering to JSON removes that.  For example, you don't know which data contains timestamps

client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
```diff
@@ -534,6 +534,12 @@ private boolean isTransient(Throwable exception)
 
     private void processResponse(Headers headers, QueryResults results)
     {
+        String variantEncoding = headers.get(TRINO_HEADERS.responseVariantEncoding());
+        if (variantEncoding == null) {
+            variantEncoding = ResultRowsDecoder.VARIANT_ENCODING_JSON;
+        }
+        resultRowsDecoder.setVariantEncoding(variantEncoding);
```
findepi: StatementClientV1 is thread-safe and this is not


why do we need this at all?
the server sends the response according to the capabilities we submitted. i don't understand why we need `TRINO_HEADERS.responseVariantEncoding()` header.

client/trino-client/src/main/java/io/trino/client/ProtocolHeaders.java
```diff
@@ -88,6 +89,7 @@ enum Headers
         RESPONSE_SET_ROLE("Set-Role"),
         RESPONSE_SET_ORIGINAL_ROLES("Set-Original-Roles"),
         RESPONSE_QUERY_DATA_ENCODING("Query-Data-Encoding"),
+        RESPONSE_VARIANT_ENCODING("Variant-Encoding"),
```
findepi: See comment in StatementClientV1

the server sends representation according to client capabilities.
do we need this header for anything?

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -661,6 +666,7 @@ public Object getObject(int columnIndex)
             throws SQLException
     {
         ColumnInfo columnInfo = columnInfo(columnIndex);
+        Object value = column(columnIndex);
```
findepi: revert?

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -670,14 +676,14 @@ public Object getObject(int columnIndex)
 
         Class<?> defaultRepresentation = DEFAULT_OBJECT_REPRESENTATION.get(columnInfo.getColumnTypeSignature().getRawType());
         if (defaultRepresentation != null) {
-            return getObject(columnIndex, defaultRepresentation);
+            return convertToDefaultObjectRepresentation(columnInfo.getColumnTypeSignature(), value);
```
findepi: revert?

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -670,14 +676,14 @@ public Object getObject(int columnIndex)
...
         }
 
-        return column(columnIndex);
+        return value;
```
findepi: revert?

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
```diff
@@ -728,6 +734,19 @@ private static Object convertFromClientRepresentation(ClientTypeSignature column
...
+    }
+
+    private static Object convertToDefaultObjectRepresentation(ClientTypeSignature columnType, @Nullable Object value)
+            throws SQLException
+    {
+        if (value == null) {
+            return null;
+        }
+        if (columnType.getRawType().equals("variant")) {
+            return decodeVariant((EncodedVariant) value).toObject();
```
findepi: why does variant need special  handling like that?
this logic should go into DEFAULT_OBJECT_REPRESENTATION map.

client/trino-client/src/main/java/io/trino/client/StatementClientV1.java
dain: I removed the response header and then none of this was needed.

client/trino-client/src/main/java/io/trino/client/ProtocolHeaders.java
dain: removed

client/trino-jdbc/src/main/java/io/trino/jdbc/AbstractTrinoResultSet.java
dain: It needed special handling from the prior design, and I missed the clean up.  This is not needed

dain:
@findepi I rethought the JSON decision.  With the redesign to lower VARIANT to JSON for old drivers, we have automatically define the JSON mapping for VARIANT, so that ship sailed.  I added JSON support to the JDBC driver with `getString(col)` and `getObject(col, String.class)`

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
```diff
@@ -33,6 +33,12 @@ public enum ClientCapabilities
      */
     NUMBER,
 
+    /**
+     * Whether client supports the `VARIANT` type encoded as JSON values on the wire.
+     * When this capability is not set, the server returns `json` for `VARIANT` columns.
+     */
+    VARIANT_JSON,
```
findepi: "varian [as] json" can be understood as "send variant values as json type" (what we do for legacy clients) or "send variant values as json objects" (what we do for new clients)

in terms of client capability names, just `VARIANT` would be a better name IMO

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
findepi: Why do we need another client capability?

i thought we need two states

- old clients: no capability being sent, we choose backwards compatible behavior
- new clients: capability being sent, we choose the best possible behavior

now we have four states

- old clients: no capability being sent, we choose backwards compatible behavior
- new clients, sending VARIANT_JSON or VARIANT_BINARY (or both)

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
```diff
@@ -39,8 +44,39 @@ public enum ClientCapabilities
...
+        this.enabledByDefault = enabledByDefault;
+    }
+
+    public boolean enabledByDefault()
+    {
+        return enabledByDefault;
+    }
+
+    public static Set<String> defaultClientCapabilities()
+    {
```
findepi: This requires documentation what does it mean that client capability is enabled by default.
If this is a JDBC thing, the logic should live in trino-jdbc, not here.

Also, I think this is abusing client capabilities concept.
Client capabilities are meant to describe "what client is capable of [handling]".
Here they seem to used for runtime behavior switching.
It would be much better to keep their original meaning. It should be assumed that latest Trino clients have all the client capabilities defined in the protocol.

findepi (dismissed):
First commit LGTM. Naming comment only (important but small)

Second commit I didn't review throughly, because I don't agree with high level mechanics  of the change. 

This PR is a RELEASE-BLOCKER, and if I am reading this correctly, the first commit resolve the release blockade. We should merge the first commit, and move the controversial commit into separate PR.

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
dain: removed

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
dain: Old clients are told the columns are JSON typed, and new clients know the column type is VARIANT.  Types matter and can have different behavior in clients.  Without this you force clients to deal with binary variant data just to know the proper type of a column.

The problem with the 4 states is real, but is a problem of the design of the protocol using flags.  In the latest code, we have VARIANT and VARIANT_BINARY.  that way it is clear what the behavior is when you send both.

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
dain: This is just VARIANT now

core/trino-main/src/main/java/io/trino/server/protocol/JsonEncodingUtils.java
```diff
@@ -331,7 +338,12 @@ public void encode(JsonGenerator generator, Block block, int position)
             }
 
             Variant variant = VARIANT.getObject(block, position);
-            generator.writeRawValue(VariantUtil.asJson(variant).toStringUtf8());
+            String json = VariantUtil.asJson(variant).toStringUtf8();
+            if (supportsVariant) {
+                generator.writeRawValue(json);
+                return;
```
electrum: Make this if/else since it's the end of the method

core/trino-main/src/main/java/io/trino/server/protocol/ProtocolUtil.java
```diff
@@ -105,12 +107,15 @@ private static String formatType(DataType type, boolean supportsParametricDateTi
                 yield ExpressionFormatter.formatExpression(type);
             }
             case RowDataType rowDataType -> rowDataType.getFields().stream()
-                    .map(field -> field.getName().map(name -> name + " ").orElse("") + formatType(field.getType(), supportsParametricDateTime, supportsNumberType))
+                    .map(field -> field.getName().map(name -> name + " ").orElse("") + formatType(field.getType(), supportsParametricDateTime, supportsNumberType, supportsVariant))
```
electrum: I'd wrap after the `+` here

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
```diff
@@ -35,10 +35,15 @@ public enum ClientCapabilities
 
     /**
      * Whether client supports the `VARIANT` type.
-     * When this capability is not set, the server returns `json` for `VARIANT` columns.
+     * When this capability and `VARIANT_BINARY` are not set, the server returns `json` for `VARIANT` columns.
```
electrum: How about

> When neither this capability nor `VARIANT_BINARY` are set, ...

client/trino-jdbc/src/test/java/io/trino/jdbc/BaseTestJdbcResultSet.java
```diff
@@ -570,6 +570,101 @@ public void testNumber()
         }
     }
 
+    @Test
+    public void testVariant()
```
electrum, line 575: Should we test all types here?

client/trino-jdbc/pom.xml
```diff
@@ -19,6 +19,11 @@
     </properties>
 
     <dependencies>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-core</artifactId>
+        </dependency>
```
electrum, line 26: Add blank line after

client/trino-jdbc/pom.xml
```diff
@@ -102,6 +107,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.gaul</groupId>
+            <artifactId>modernizer-maven-annotations</artifactId>
+            <scope>provided</scope>
```
electrum, line 108: I don't think provided is correct here. Optional should be sufficient

electrum (approved):

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
findepi: I don't understand this response.

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
```diff
@@ -33,6 +33,17 @@ public enum ClientCapabilities
...
+    /**
+     * Whether client supports the `VARIANT` type.
+     * When neither this capability nor `VARIANT_BINARY` are set, the server returns `json` for `VARIANT` columns.
+     */
+    VARIANT,
+
+    /**
+     * Whether client supports the `VARIANT` type encoded as a binary payload on the wire.
+     */
+    VARIANT_BINARY,
```
findepi, line 45: Capabilities are named & designed to recognize if "is this client updated". An ideal client should have all capabilities, but these two are clearly conflicting.

Why do you insist on two capabilities for one thing?

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
electrum: There are three states:

- old clients that are told the type is `json`
- new clients that are told the type is `variant` and get the data as JSON
- new clients that are told the type is `variant` and get the data as binary variant

Thus there are two flags:

- I understand the variant type
- I understand (or want to deal with) variant binary

findepi: I am not excited about modeling "new clients that ..." as client capabilities.
[As I said before](https://github.com/trinodb/trino/pull/29046#discussion_r3135530051),  An ideal client should have all capabilities, but these two are clearly conflicting.  It's not clear what should be the behavior when client declares both. Is it well defined, or is it up to server discretion? 

Trino JDBC probably needs binary variant representation.

A client that doesn't want to deal with binary variant representation (perhaps because it doesn't have necessary libraries to decode that), will not send the variant capability and will get data as json string. I don't think json-in-json is so important that it deserves its own 3rd option.

client/trino-client/src/main/java/io/trino/client/ClientCapabilities.java
electrum: Client capabilities describe what a client supports. There's no requirement that a client has every capability. If the intent was to describe if a client is "update to date" in a strictly ordered sense, then we would have used a protocol version number rather independent flags.

These are two features, with some overlap. Supporting binary variant implies support for variant, but supporting the variant type doesn't imply support for binary.

client/trino-jdbc/pom.xml
dain: Provided is needed due to the shading.

PR #29054, opened 2026-04-09
Fix health-check when default http port is used

wendigo:
- Fixes https://github.com/trinodb/trino/issues/29043

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/docker/bin/health-check
+++ b/core/docker/bin/health-check
@@ -2,20 +2,24 @@
 
 set -euo pipefail
 
-function get_property() {
-    grep "^$1=" "$2" | cut -d'=' -f2
+function get_property_value_or_default() {
+  local value
+  if value="$(grep "^$1=" "$2")"; then
+    printf '%s\n' "${value}" | cut -d'=' -f2
+  else
+    # property not found, return default
+    printf '%s\n' "$3"
+  fi
 }
 
 scheme=http
-port=8080
-
 config=/etc/trino/config.properties
 # prefer to use http even if https is enabled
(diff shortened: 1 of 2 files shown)


core/docker/bin/health-check
```diff
@@ -3,12 +3,10 @@
 set -euo pipefail
 
 function get_property() {
-    grep "^$1=" "$2" | cut -d'=' -f2
+    (grep "^$1=" "$2" | cut -d'=' -f2) || echo ""
```
findepi: The `echo` doesn't get to run, right?
The exit code of `grep | cut` is that of `cut`, which should always exit with 0.
It was failing previously because of `set -o pipefail`, which is suppressed in conditional statements (`if`, `||`).

As written now, it's misleading.
a better way to write this would be

```
function get_property_value_or_empty() {
  local value
  if value="$(grep "^$1=" "$2")"; then
    printf '%s\n' "${value}" | cut -d'=' -f2 # or <<< "${value}"
  else 
    # property not found, return empty
    return
  fi
}
```

core/docker/default/etc/config.properties
```diff
@@ -1,6 +1,5 @@
 #single node install config
 coordinator=true
 node-scheduler.include-coordinator=true
-http-server.http.port=8080
```
findepi, line 4: it was removed for testing only, or permanently?

findepi (approved):

core/docker/default/etc/config.properties
wendigo: permanently. default is 8080, there is no point in overriding a default value

findepi (approved):

Issue #27112, opened 2025-10-27
Flaky test `TestLakehouseMotoConnectorSmokeTest`


findepi:
the following failed on `master` https://github.com/trinodb/trino/actions/runs/18831687061/job/53724268251

```
Error:  Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 30.34 s <<< FAILURE! -- in io.trino.plugin.lakehouse.TestLakehouseMotoConnectorSmokeTest
Error:  io.trino.plugin.lakehouse.TestLakehouseMotoConnectorSmokeTest -- Time elapsed: 30.34 s <<< ERROR!
software.amazon.awssdk.core.exception.SdkClientException: Unable to execute HTTP request: The target server failed to respond (SDK Attempt Count: 4)
	at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:130)
	at software.amazon.awssdk.core.exception.SdkClientException$BuilderImpl.build(SdkClientException.java:95)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.utils.RetryableStageHelper.retryPolicyDisallowedRetryException(RetryableStageHelper.java:168)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:73)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.RetryableStage.execute(RetryableStage.java:36)
	at software.amazon.awssdk.core.internal.http.pipeline.RequestPipelineBuilder$ComposingRequestPipelineStage.execute(RequestPipelineBuilder.java:206)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:53)
	at software.amazon.awssdk.core.internal.http.StreamManagingStage.execute(StreamManagingStage.java:35)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.executeWithTimer(ApiCallTimeoutTrackingStage.java:82)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:62)
	at software.amazon.awssdk.core.internal.http.pipeline.stages.ApiCallTimeoutTrackingStage.execute(ApiCallTimeoutTrackingStage.java:43)
[...]

PR #29066, opened 2026-04-10
Remove trino-main compile dependency from trino-faker

### LINKED ISSUE #29065: Remove `trino-main` dependency from `trino-faker`
Plugins should not depend on trino-main other than for tests.

nineinchnick:
## Summary
- Remove compile-scoped `trino-main` and `trino-parser` dependencies from the faker plugin
- Copy datetime parsing methods from `DateTimes` and `DateTimeUtils` into a local `DateTimeParsing` class, simplifying where possible:
  - `parseDate`: java.time `DateTimeFormatter` instead of Joda
  - `parseDayTimeInterval`: manual seconds.millis parsing instead of Joda `PeriodFormatter`
  - `parseYearMonthInterval`: simple `Long.parseLong` instead of Joda `PeriodFormatter`
  - time/timestamp parsing: copied as-is (only depends on trino-spi)
- Replace `instanceof` type checks (`IpAddressType`, `JsonType`, `TDigestType`, `IntervalDayTimeType`, `IntervalYearMonthType`) with `type.getBaseName().equals(StandardTypes.X)` comparisons
- Rewrite `FakerFunctionProvider` to use `FunctionMetadata.scalarBuilder()` and `MethodHandle` instead of `ScalarFromAnnotationsParser`/`SqlScalarFunction` from trino-main

## Test plan
- [x] `mvnd test -pl plugin/trino-faker` — all 48 tests pass
- [x] Tests pass with non-UTC JVM timezone (`-Duser.timezone=Asia/Kathmandu`)
- [x] CI

Fixes #29065

### DIFF
--- a/plugin/trino-faker/pom.xml
+++ b/plugin/trino-faker/pom.xml
@@ -49,22 +49,6 @@
 
         <dependency>
             <groupId>io.trino</groupId>
-            <artifactId>trino-main</artifactId>
-            <exclusions>
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-parser</artifactId>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
             <artifactId>trino-plugin-toolkit</artifactId>
         </dependency>
 
@@ -148,6 +132,12 @@
         <dependency>
             <groupId>io.trino</groupId>
             <artifactId>trino-exchange-filesystem</artifactId>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>io.trino</groupId>
+            <artifactId>trino-main</artifactId>
             <scope>test</scope>
         </dependency>
 
--- a/plugin/trino-faker/src/main/java/io/trino/plugin/faker/DateTimeParsing.java
+++ b/plugin/trino-faker/src/main/java/io/trino/plugin/faker/DateTimeParsing.java
@@ -1,0 +1,410 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.plugin.faker;
+
+import io.trino.spi.type.DateTimeEncoding;
+import io.trino.spi.type.LongTimeWithTimeZone;
+import io.trino.spi.type.LongTimestamp;
+import io.trino.spi.type.LongTimestampWithTimeZone;
+import io.trino.spi.type.TimeType;
+import io.trino.spi.type.TimeWithTimeZoneType;
+import io.trino.spi.type.TimestampType;
+import io.trino.spi.type.TimestampWithTimeZoneType;
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.ZoneOffset;
+import java.time.format.DateTimeFormatter;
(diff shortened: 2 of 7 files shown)


plugin/trino-faker/src/main/java/io/trino/plugin/faker/DateTimeParsing.java
```diff
@@ -0,0 +1,405 @@
...
+import static io.trino.spi.type.TimeZoneKey.getTimeZoneKey;
+import static io.trino.spi.type.TimestampType.MAX_SHORT_PRECISION;
+import static java.lang.Math.multiplyExact;
+import static java.lang.String.format;
+import static java.time.ZoneOffset.UTC;
+
+/**
+ * Duplicated from io.trino.type.DateTimes and io.trino.util.DateTimeUtils
+ * to avoid a dependency on trino-main.
+ * TODO: find a shared module for these methods.
```
nineinchnick: Anyone reviewing this, please advise.

findepi: `lib/trino-plugin-toolkit/src/main/java/io/trino/plugin/base/type`?

nineinchnick: This is a big refactor, I have to do this later.

nineinchnick: I moved only the functions used by the Faker plugin, but this creates a weird split. OTOH moving everything also seems like a bad idea. It's not really the responsibility of the plugin-toolkit to provide parsing methods.

Please confirm if this should be done.

findepi: > It's not really the responsibility of the plugin-toolkit to provide parsing methods.

for trino literals? i agree

core/trino-main/src/main/java/io/trino/type/DateTimes.java
```diff
@@ -345,158 +340,12 @@ private static void formatTimestampIntoBuilder(StringBuilder builder, int precis
 
     public static Object parseTimestamp(int precision, String value)
     {
-        if (precision <= MAX_SHORT_PRECISION) {
-            return parseShortTimestamp(value);
-        }
-
-        return parseLongTimestamp(value);
+        return io.trino.plugin.base.type.DateTimes.parseTimestamp(precision, value);
```
findepi: if we call this new class `Timestamps`, we can get away without qualified names

also, we can import parseTimestamp statically.
what is the point of retaining this method if it is only pass-through?

plugin/trino-faker/src/main/java/io/trino/plugin/faker/DateTimeParsing.java
nineinchnick: Can you confirm we want to refactor the parsing methods to be available to plugins at all, or is that main's private API?

findepi: I think it's fine. You can also copy them (or simplified version thereof).

We could also expose casts from VARCHAR to all the types via ConnectorContext, but that's actually a SPI change, and not as easy.

nineinchnick: I copied all these methods into the Faker plugin. I don't want to do any refactors outside of it.

plugin/trino-faker/pom.xml
```diff
@@ -47,22 +47,6 @@
             <artifactId>units</artifactId>
         </dependency>
 
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-main</artifactId>
```
findepi, line 52: ❤️

plugin/trino-faker/pom.xml
```diff
@@ -47,22 +47,6 @@
...
-                <exclusion>
-                    <groupId>*</groupId>
-                    <artifactId>*</artifactId>
-                </exclusion>
-            </exclusions>
-        </dependency>
-
-        <dependency>
-            <groupId>io.trino</groupId>
-            <artifactId>trino-parser</artifactId>
```
findepi, line 63: ❤️

findepi (approved):

findepi:
please rebase

findepi:
> please rebase

done

also included update of `trino-maven-plugin` to get
- https://github.com/trinodb/trino-maven-plugin/pull/35

findepi:
CI
- https://github.com/trinodb/trino/issues/27112#issuecomment-4286812773

PR #29068, opened 2026-04-10
Fix IS NOT NULL on Glue partition column causing planning timeout

EugeneChung:
## Description

Stop pushing `(col <> '__HIVE_DEFAULT_PARTITION__')` expression to Glue's `GetPartitions` API for varchar partition columns. The not-equals expression prevents Glue from using partition indexes, causing a full partition scan and planning timeouts on large tables.

**Reproducer:**

```sql
CREATE TABLE hive.wheres_etl.airbridge_inapps_v2 (
      ....,
      app_bundle varchar,
      utc_basic_time varchar
  )
  WITH (
      format = 'PARQUET',
      partitioned_by = ARRAY['app_bundle', 'utc_basic_time']
  )
```
The number of partitions would be more than 200,000.

```sql
-- Planning Times out:
SELECT DISTINCT app_bundle FROM wheres_etl.airbridge_inapps_v2
WHERE utc_basic_time = '2026-03-19-02' AND app_bundle IS NOT NULL;

  ⎿  Error: Exit code 1
   Query 20260410_160910_00000_i5rcy failed: The optimizer exhausted the time limit of 180000 ms: Top rules: {
     io.trino.sql.planner.iterative.rule.PushPredicateIntoTableScan: 181613 ms, 1 invocations, 1 applications}

-- Works fine (same query without IS NOT NULL):
SELECT DISTINCT app_bundle FROM wheres_etl.airbridge_inapps_v2
WHERE utc_basic_time = '2026-03-19-02';
```

IS NOT NULL filtering is already handled client-side by `HivePartitionManager`, which correctly filters out null partitions via `parseValuesAndFilterPartition` → `HiveUtil.parsePartitionValue` → `partitionMatches`. This is the same mechanism that already protects non-quoted types (INTEGER, BIGINT).

Non-quoted types already returned `Optional.empty()` at the earlier guard, so this bug only affected quoted types (VARCHAR, CHAR).

## Release notes

```markdown
## Hive connector
* Fix planning timeout when using `IS NOT NULL` on varchar partition columns with Glue metastore. ({issue}`issuenumber`)
```


EugeneChung:
I accidentally removed the branch in my repo. I will open another PR. Sorry to bother you.

Issue #21651, opened 2024-04-22
"Table has exceeded max number of active streams" error happens in BigQuery connector


ebyhr:
https://github.com/trinodb/trino/actions/runs/8751706370/job/24017920223?pr=21628

```
Error:  io.trino.plugin.bigquery.TestBigQueryQueryFailureRecoveryTest.testCreateTable -- Time elapsed: 196.7 s <<< ERROR!
io.trino.testing.QueryFailedException: io.grpc.StatusRuntimeException: FAILED_PRECONDITION: Table has exceeded max number of active streams. Please use existing streams or finalize unused streams. Table=505084745097:tpch.tmp_trino_1516d0b4_12bccd97. ActiveStreamCount=916035320. MaxAllowed=20000. Entity: projects/sep-bq-cicd/datasets/tpch/tables/tmp_trino_1516d0b4_12bccd97
	at io.trino.testing.AbstractTestingTrinoClient.execute(AbstractTestingTrinoClient.java:133)
	at io.trino.testing.DistributedQueryRunner.executeInternal(DistributedQueryRunner.java:501)
	at io.trino.testing.DistributedQueryRunner.executeWithPlan(DistributedQueryRunner.java:490)
	at io.trino.testing.BaseFailureRecoveryTest$FailureRecoveryAssert.execute(BaseFailureRecoveryTest.java:534)
	at io.trino.testing.BaseFailureRecoveryTest$FailureRecoveryAssert.executeExpected(BaseFailureRecoveryTest.java:494)
	at io.trino.testing.BaseFailureRecoveryTest$FailureRecoveryAssert.lambda$failsWithoutRetries$10(BaseFailureRecoveryTest.java:707)
	at io.trino.testing.BaseFailureRecoveryTest$FailureRecoveryAssert.getFailureStageId(BaseFailureRecoveryTest.java:731)
	at io.trino.testing.BaseFailureRecoveryTest$FailureRecoveryAssert.failsWithoutRetries(BaseFailureRecoveryTest.java:707)
	at io.trino.testing.BaseFailureRecoveryTest.testNonSelect(BaseFailureRecoveryTest.java:367)
	at io.trino.testing.BaseFailureRecoveryTest.testTableModification(BaseFailureRecoveryTest.java:307)
	at io.trino.testing.BaseFailureRecoveryTest.testTableModification(BaseFailureRecoveryTest.java:302)
[...]

PR #29092, opened 2026-04-13
Update Minio container image

mosabua:
## Description

Addresses [CVE-2026-39414](https://github.com/advisories/GHSA-h749-fxx7-pwpg) ([GHSA-h749-fxx7-pwpg](https://github.com/advisories/GHSA-h749-fxx7-pwpg)) in the older image. 

This is now using the Chainguard EmeritOSS maintained version of MinIO, which now includes patches for security issues.

Uses https://github.com/chainguard-forks/minio/releases/tag/RELEASE.2026-04-10T21-52-59Z

Related PR for aws-proxy is at https://github.com/trinodb/aws-proxy/pull/210

## Release notes

### DIFF
--- a/testing/trino-testing-containers/src/main/java/io/trino/testing/containers/Minio.java
+++ b/testing/trino-testing-containers/src/main/java/io/trino/testing/containers/Minio.java
@@ -41,8 +41,7 @@ public class Minio
         extends BaseTestContainer
 {
     private static final Logger log = Logger.get(Minio.class);
-
-    public static final String DEFAULT_IMAGE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:66bd82c8fe5e75868ae7d0b2e102d9a0dcf971b270a41bd060a9e6a643476ff8")
+    public static final String DEFAULT_IMAGE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:f2aba4b9c84b61881933142f1cb7cc1a20f8e8456f1f1aa353ddd248f23037f7")
             .asCanonicalNameString();
     public static final String DEFAULT_HOST_NAME = "minio";
 
--- a/testing/trino-testing-containers/src/main/java/io/trino/testing/containers/Minio.java
+++ b/testing/trino-testing-containers/src/main/java/io/trino/testing/containers/Minio.java
@@ -42,7 +42,7 @@ public class Minio
 {
     private static final Logger log = Logger.get(Minio.class);
 
-    public static final String DEFAULT_IMAGE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:66bd82c8fe5e75868ae7d0b2e102d9a0dcf971b270a41bd060a9e6a643476ff8")
+    public static final String DEFAULT_IMAGE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:f767919bd003062ac69713cdce920eb922c9fa3388efe96264e78b763342ca1a")
             .asCanonicalNameString();
     public static final String DEFAULT_HOST_NAME = "minio";
 
--- a/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
+++ b/testing/trino-product-tests-launcher/src/main/java/io/trino/tests/product/launcher/env/common/Minio.java
@@ -40,7 +40,7 @@ public class Minio
 
     private static final String MINIO_ROOT_USER = "minio-access-key";
     private static final String MINIO_ROOT_PASSWORD = "minio-secret-key";
-    private static final String MINIO_RELEASE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:66bd82c8fe5e75868ae7d0b2e102d9a0dcf971b270a41bd060a9e6a643476ff8")
+    private static final String MINIO_RELEASE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:f2aba4b9c84b61881933142f1cb7cc1a20f8e8456f1f1aa353ddd248f23037f7")
             .asCanonicalNameString();
 
     private static final int MINIO_PORT = 9080; // minio uses 9000 by default, which conflicts with hadoop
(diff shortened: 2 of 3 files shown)


ebyhr (dismissed):

testing/trino-testing-containers/src/main/java/io/trino/testing/containers/Minio.java
```diff
@@ -41,8 +41,11 @@ public class Minio
         extends BaseTestContainer
 {
     private static final Logger log = Logger.get(Minio.class);
-
-    public static final String DEFAULT_IMAGE = DockerImageName.parse("cgr.dev/chainguard/minio@sha256:66bd82c8fe5e75868ae7d0b2e102d9a0dcf971b270a41bd060a9e6a643476ff8")
+    
+    // Use specific sha256 digest to avoid accidentally picking up a new 
+    // incompatible image with the same tag
+
```
ebyhr: Remove a redundant empty line.

mosabua: Done

mosabua:
Should this not run more tests @ebyhr ?

mosabua:
Haha .. now it suddenly kicked things off .. thanks @ebyhr

mosabua:
Test failure looks like a false alarm to me. Thoughts @ebyhr @wendigo @electrum ?

mosabua:
Whoa .. this is weird. Earlier run everything passed apart from the referenced Delta Lake test .. now Iceberg with minio-avro is also busting out .. will see how this continues.

findepi:
> Addresses [CVE-2026-39414](https://github.com/advisories/GHSA-h749-fxx7-pwpg) ([GHSA-h749-fxx7-pwpg](https://github.com/advisories/GHSA-h749-fxx7-pwpg)) in the older image.

Does this matter?
it's tests.

mosabua:
> > Addresses [CVE-2026-39414](https://github.com/advisories/GHSA-h749-fxx7-pwpg) ([GHSA-h749-fxx7-pwpg](https://github.com/advisories/GHSA-h749-fxx7-pwpg)) in the older image.
> 
> Does this matter? it's tests.

Ideally we update since these tests also run on CI/CD and developer machines, but we would have to find out why the tests are failing first. I am not sure when I can get into the details.

electrum:
The tests might pass now that we've removed most usages of Minio.

electrum (approved):

mosabua:
Pushed update to latest digest. Waiting on CI.

mosabua:
Dang .. looks like Delta Lake is still failing. Any idea if this is brittle tests or some real issue with Delta Lake and our code or some issue with the minio container @findinpath  @electrum ?

mosabua:
I am finding some issues with the minio container as well as the tests. Working on both in parallel at the moment. Stay tuned.

testing/trino-product-tests/src/main/java/io/trino/tests/product/utils/MinioNotificationsAssertions.java
```diff
@@ -105,24 +108,28 @@ private static String quote(String input)
 
     public static void assertNotificationsCount(String tableName, EventType eventType, String objectName, int expectedCalls)
```
findinpath: @mosabua  I think we should rather drop altogether this scaffolding and use telemetry techniques which we already benefit of in https://github.com/trinodb/trino/blob/c875259b2121ae338f37cf5cd2641c81543e9348/plugin/trino-delta-lake/src/test/java/io/trino/plugin/deltalake/TestDeltaLakeFileOperations.java


Please consider replacing ( if not already covered ) these product tests with integration tests in `TestDeltaLakeFileOperations`


Explicit: Please remove `DELTA_LAKE_MINIO` product test group and associated classes

cc @ebyhr  correct me if I'm wrong.

findinpath: @mosabua  if you consider doing the above recommendation, it would probably make most sense to do it in a separate preparatory PR

mosabua:
Putting this on hold — the `suite-delta-lake-oss` failures are not a Trino issue.

The newer `cgr.dev/chainguard/minio` image in this bump does not deliver S3
bucket notifications to `ListenBucketNotification` subscribers at all. The Delta
Lake product tests `TestDeltaLakeDatabricksMinioReads` and
`TestDeltaLakeOssDeltaLakeMinioReads` rely on those notifications to verify that
Trino issues the expected HEAD and GET requests on the Delta log and data files,
so they fail with no test-side workaround.

Confirmed with a minimal reproducer:
https://github.com/mosabua/minio-listen-repro

- Old digest `sha256:66bd82c8…`, version `DEVELOPMENT.2025-10-15`: all event
  families delivered in about half a second.
- New digest `sha256:a07e6db3…`, version `RELEASE.2026-06-04`: zero events
  delivered, with wildcard or specific subscriptions, and no stream error.

The image is built from the `chainguard-forks/minio` fork, and this is a
regression in that build. I have reverted the exploratory test changes, so this
PR now contains only the image digest bump, and marked it draft until the
notification regression is fixed.

mosabua:
Update on the hold: root cause found and a fix is open upstream.

The `suite-delta-lake-oss` failures were not a notification-delivery problem —
the newer image's MinIO server publishes and writes the events, but the chunked
`text/event-stream` response from `ListenBucketNotification` is never flushed to
the client, so the listener receives nothing. The cause is MinIO commit
`52eee5a2f` (#21651), which wraps every S3 API response in a
`trackingResponseWriter` that does not implement `Flush` or `Hijack`, breaking
the flush chain for streaming responses. The image before that commit worked,
which is why the digest bump exposed it.

Fix PR against the Chainguard fork that builds the image:
https://github.com/chainguard-forks/minio/pull/37

This PR stays a draft until a fixed image is published, after which it is just
the digest bump. A standalone reproducer is at
https://github.com/mosabua/minio-listen-repro

electrum:
@mosabua thanks for working on this. Do you think we should try switching this test to Floci?

mosabua:
> @mosabua thanks for working on this. Do you think we should try switching this test to Floci?

No idea about Floci. I have a PR up that fixes the regression in the minio container hopefully and then we can proceed here. In the longer run however if we find an active project that innovates and moves S3-compatible forward more we should move. The Chainguard minio is purely on maintenance mode.

electrum:
I recently converted all of the AWS tests to use Floci, and even added some local Floci versions that previously required real AWS, after making more than a dozen improvements to Floci. I also added a Floci version of the GCS tests (Azure still needs work).

If there are any incompatibilities, it's easy to fix, and their goal is full compatibility with all clouds while being 100% open source.

mosabua:
Pushed update to the new MinIO image that has the fix. 🤞

mosabua:
Everything passes. Adding to the queue!

PR #29095, opened 2026-04-13
Fix NULL grouping-key rows dropped by MultipleDistinctAggregationsToSubqueries

bvolpato:
## Description

`MultipleDistinctAggregationsToSubqueries` rewrites a grouped multi-distinct aggregation such as

```sql
SELECT k, count(DISTINCT a), count(DISTINCT b)
FROM t
GROUP BY k
```

into an `INNER JOIN` of sub-aggregations joined on the grouping keys. The join used regular equi-criteria, so rows belonging to the `NULL` grouping-key group did not match across sub-aggregations. `GROUP BY` collapses `NULL` values into a single group, but `NULL = NULL` is UNKNOWN, so the join could silently drop the `NULL` group.

Reproducer:

```sql
CREATE TABLE iceberg.test.null_group_sales (region VARCHAR, customer_id BIGINT, product_id BIGINT);
INSERT INTO iceberg.test.null_group_sales VALUES
    ('US', 1, 100),
    ('EU', 2, 200),
    (NULL, 3, 300),
    (NULL, 4, 400),
    (NULL, 3, 500);

WITH SESSION distinct_aggregations_strategy = 'split_to_subqueries'
SELECT region, count(DISTINCT customer_id), count(DISTINCT product_id)
FROM iceberg.test.null_group_sales
GROUP BY region;
```

Expected:

```text
 region | _col1 | _col2
--------+-------+-------
 US     |   1   |   1
 EU     |   1   |   1
 NULL   |   2   |   3
```

Actual before this fix: the `NULL` group is missing.

## Fix

For each grouping key `k`, project two non-null join keys on both sides:

```text
k_is_null   = k IS NULL
k_coalesced = COALESCE(k, <type sentinel>)
```

Then join on both symbols. This keeps the plan as a hashable equi-join and distinguishes real sentinel values from `NULL` rows: `NULL` becomes `(TRUE, sentinel)`, while a real sentinel value becomes `(FALSE, sentinel)`.

If any grouping key type has no native sentinel, skip this rewrite. That avoids replacing the rewrite with a cross join that is not obviously better than the other distinct aggregation strategies. Global aggregations with no grouping keys still use a plain cross join of the single-row sub-aggregations.

- The bug is reachable on the default path. `DistinctAggregationsStrategy` defaults to `AUTOMATIC`, and `DistinctAggregationStrategyChooser#chooseMarkDistinctStrategy` can auto-select `SPLIT_TO_SUBQUERIES` for multi-distinct grouped aggregations over simple sources with low/unknown NDV.
- Existing connector tests in `TestDistinctToSubqueriesAggregations` run against TPC-H tables, where every column is `NOT NULL` by spec, so the `NULL` grouping-key path was not exercised before.
- Added regression coverage in `AbstractTestAggregations#testMultipleDistinctAggregationsPreserveNullGroupKey`, using TPC-H `nation` under `split_to_subqueries`.
- Added coverage for an unsupported/no-sentinel `VARBINARY` grouping key to verify the rule skips the rewrite correctly.
- Updated rule tests to assert sentinel support, unsupported types, and the null-safe equi-join shape.

## Release notes

```markdown
## General
* Fix incorrect results for queries that combine multiple `DISTINCT` aggregations with a `GROUP BY` on a nullable column. Previously, rows with `NULL` in the grouping key could be silently dropped when the `split_to_subqueries` strategy (including automatic selection) was used.
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
+++ b/core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
@@ -16,11 +16,31 @@ package io.trino.sql.planner.iterative.rule;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import io.airlift.slice.Slices;
 import io.trino.cost.TaskCountEstimator;
 import io.trino.matching.Captures;
 import io.trino.matching.Pattern;
 import io.trino.metadata.Metadata;
+import io.trino.spi.type.AbstractIntType;
+import io.trino.spi.type.AbstractLongType;
+import io.trino.spi.type.BooleanType;
+import io.trino.spi.type.CharType;
+import io.trino.spi.type.DecimalType;
+import io.trino.spi.type.DoubleType;
+import io.trino.spi.type.Int128;
+import io.trino.spi.type.LongTimestamp;
+import io.trino.spi.type.LongTimestampWithTimeZone;
+import io.trino.spi.type.SmallintType;
+import io.trino.spi.type.TimeZoneKey;
+import io.trino.spi.type.TimestampType;
+import io.trino.spi.type.TimestampWithTimeZoneType;
+import io.trino.spi.type.TinyintType;
+import io.trino.spi.type.Type;
+import io.trino.spi.type.VarcharType;
+import io.trino.sql.ir.Coalesce;
+import io.trino.sql.ir.Constant;
 import io.trino.sql.ir.Expression;
+import io.trino.sql.ir.IsNull;
 import io.trino.sql.planner.NodeAndMappings;
 import io.trino.sql.planner.PlanCopier;
 import io.trino.sql.planner.Symbol;
@@ -41,10 +61,10 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.IntStream;
 
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.ImmutableList.toImmutableList;
+import static io.trino.spi.type.BooleanType.BOOLEAN;
 import static io.trino.sql.planner.iterative.rule.DistinctAggregationStrategyChooser.createDistinctAggregationStrategyChooser;
 import static io.trino.sql.planner.plan.JoinType.INNER;
 import static io.trino.sql.planner.plan.Patterns.aggregation;
@@ -114,6 +134,17 @@ public class MultipleDistinctAggregationsToSubqueries
         if (!distinctAggregationStrategyChooser.shouldSplitToSubqueries(aggregationNode, context.getSession(), context.getStatsProvider(), context.getLookup())) {
             return Result.empty();
         }
+
+        // Null-safe join on the grouping keys requires a non-null sentinel value per key type
+        // (see buildJoin for details). If any grouping key has a type without a sentinel, skip
+        // the rewrite: a cross join would be required to preserve NULL groups, and that is
+        // not obviously better than the other distinct aggregation strategies.
+        for (Symbol groupingKey : aggregationNode.getGroupingKeys()) {
+            if (nullSentinel(groupingKey.type()).isEmpty()) {
+                return Result.empty();
+            }
+        }
+
         // group aggregations by arguments
         Map<Set<Expression>, Map<Symbol, Aggregation>> aggregationsByArguments = new LinkedHashMap<>(aggregationNode.getAggregations().size());
         // sort the aggregation by output symbol to have consistent join layout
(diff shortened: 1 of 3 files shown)


raunaqmorarka:
@CodeRabbit review

testing/trino-tests/src/test/java/io/trino/tests/TestMultipleDistinctAggregationsWithNullGroupKey.java
```diff
@@ -0,0 +1,116 @@
...
+                        "(NULL, 3, 300), " +
+                        "(NULL, 4, 400), " +
+                        "(NULL, 3, 500)",
+                5);
+
+        String query = "SELECT region, count(DISTINCT customer_id), count(DISTINCT product_id) " +
+                "FROM null_group_sales " +
+                "GROUP BY region";
+
+        MaterializedResult reference = computeActual(singleStepSession(), query);
```
lukasz-stec: Instead of computing the single step, you can explicitly assert the expected results like
`assertQuery(query, "VALUES ('US', 1, 1), ('EU', 1, 1), (NULL, 2, 3)");`

testing/trino-tests/src/test/java/io/trino/tests/TestMultipleDistinctAggregationsWithNullGroupKey.java
```diff
@@ -0,0 +1,116 @@
...
+    {
+        return Session.builder(getSession())
+                .setSystemProperty(DISTINCT_AGGREGATIONS_STRATEGY, "single_step")
+                .build();
+    }
+
+    @Test
+    public void testMultipleDistinctAggregationsPreserveNullGroupKey()
+    {
+        assertUpdate("CREATE TABLE null_group_sales (region VARCHAR, customer_id BIGINT, product_id BIGINT)");
```
lukasz-stec: This is a good test case, generally, so I would move it to the `AbstractTestAggregations`. The same goes for the other test.
For this, you will have to rely only on tpch tables and not create tables (not all connectors support creating tables).
You can use a query like
```
    select regionkey, count(DISTINCT name), count(DISTINCT nationkey)
    from (
        select nationkey, name, case when regionkey > 1 then NULL else regionkey end as regionkey
        from nation) n
    group by regionkey
```

lukasz-stec: Also, please add a test case for type that does not have sentinel

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
```diff
@@ -181,24 +207,162 @@ private AggregationNode buildSubAggregation(AggregationNode aggregationNode, Map
...
+        ImmutableList.Builder<EquiJoinClause> criteriaBuilder = ImmutableList.builder();
+        Assignments.Builder leftExtras = Assignments.builder();
+        Assignments.Builder rightExtras = Assignments.builder();
+        boolean allHaveSentinels = true;
+
+        for (int i = 0; i < leftJoinSymbols.size(); i++) {
+            Symbol lk = leftJoinSymbols.get(i);
+            Symbol rk = rightJoinSymbols.get(i);
+            Object sentinel = nullSentinel(lk.type());
+            if (sentinel == null) {
```
lukasz-stec: For the case without a sentinel, we should skip this rule. It is not clear that a cross join will be better than the alternatives

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
```diff
@@ -181,24 +207,162 @@ private AggregationNode buildSubAggregation(AggregationNode aggregationNode, Map
...
                 ImmutableMap.of(),
                 Optional.empty());
     }
+
+    /**
+     * Returns a non-null sentinel value for {@code type} suitable for use in
+     * {@code COALESCE(k, sentinel)} when building null-safe equi-join keys.
+     * Returns {@code null} if no sentinel is available for this type.
+     */
+    private static Object nullSentinel(Type type)
```
lukasz-stec: return Optional to indicate that not all types are supported

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
```diff
@@ -181,24 +207,162 @@ private AggregationNode buildSubAggregation(AggregationNode aggregationNode, Map
...
+    {
+        if (type instanceof BigintType || type instanceof IntegerType
```
lukasz-stec: Checking for `AbstractIntType` and `AbstractLongType` would simplify the logic.

core/trino-main/src/test/java/io/trino/sql/planner/iterative/rule/TestMultipleDistinctAggregationsToSubqueries.java
```diff
@@ -393,8 +394,11 @@ public void testAutomaticDecisionForAggregationOnTableScan()
                         join(
                                 INNER,
                                 builder -> builder
-                                        .equiCriteria("left_groupingKey", "right_groupingKey")
-                                        .left(aggregation(
+                                        // The join uses null-safe equi-criteria on projected (is_null, coalesced) pairs
+                                        // to preserve NULL grouping-key rows.  We ignore the criteria because the
+                                        // wrapping ProjectNodes use dynamically-allocated symbol names.
+                                        .ignoreEquiCriteria()
```
lukasz-stec: It would be good to not ignroe the criteria. Those are the critical for this rule. If you add the project below, you can specify the symbols in the assignemts and than use it in the join criteria.
For example (taken from `TestJoinNodeFlattener#testPushesProjectionsThroughJoin`)
```
        Symbol c = p.symbol("C");
        Symbol d = p.symbol("D");
        ValuesNode valuesA = p.values(a);
        ValuesNode valuesB = p.values(b);
        ValuesNode valuesC = p.values(c);
        JoinNode joinNode = p.join(
                INNER,
                p.project(
                        Assignments.of(d, new Call(NEGATION_BIGINT, ImmutableList.of(a.toSymbolReference()))),
                        p.join(
                                INNER,
                                valuesA,
                                valuesB,
                                equiJoinClause(a, b))),
                valuesC,
                equiJoinClause(d, c));
```

core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
lukasz-stec: It would be good to add tests for all the types handled here

lukasz-stec (commented):
That's a good catch. Thanks for submitting this. I added a few comments.

bvolpato:
[review-prs-verification]
Head: 3f57cd054f925fc0e5033f37b44347eafa095073
Ready to merge: yes

Files reviewed.
- core/trino-main/src/main/java/io/trino/sql/planner/iterative/rule/MultipleDistinctAggregationsToSubqueries.java
- core/trino-main/src/test/java/io/trino/sql/planner/iterative/rule/TestMultipleDistinctAggregationsToSubqueries.java
- testing/trino-testing/src/main/java/io/trino/testing/AbstractTestAggregations.java

Tests/checks run and results.
- `gh pr checks 29095 --watch`: complete, all listed checks passed at this head.
- `./mvnw -pl core/trino-main -am -DskipTests package`: PASS.
- `./mvnw -pl core/trino-main -am -DskipTests install`: PASS.
- `./mvnw -pl core/trino-main -Dtest=io.trino.sql.planner.iterative.rule.TestMultipleDistinctAggregationsToSubqueries test`: PASS (12 tests, 0 failures).
- `./mvnw -pl testing/trino-testing -DskipTests package`: PASS.
- `./mvnw -pl testing/trino-testing test`: PASS (4 tests, 0 failures).

Fixes committed/pushed, if any.
- None. No file changes were required during this review pass.

Final CI status.
- GitHub PR checks: PASS for head `3f57cd054f925fc0e5033f37b44347eafa095073`.

raunaqmorarka:
@CodeRabbit full review

lukasz-stec (approved):
lgtm, @raunaqmorarka could you take a look

PR #29103, opened 2026-04-14
Add support for running Trino in OpenShift

### LINKED ISSUE #22281: trinodb/trino container image does not work out-of-the-box with OpenShift (e.g. running the container as an arbitrary/random user ID)
When using the provided `trinodb/trino` container image directly in [OpenShift](https://www.redhat.com/en/technologies/cloud-computing/openshift), there are a few issues that need to be stumbled through before it can start to work properly.

As far as I understand it, in a nut-shell, OpenShift normally works by assigning a dynamic range of UIDs to each namespace (or project, as they call it) and then each container is started with a user which has a unique UID number, and that the user is by default a member of the `root` group within the container (but is not the actual `root` user, mind).  There is also a policy which blocks if you try to set `runAsUser`/`runAsGroup` in the `securityContext` to anything which is outside of your namespace's assigned number range (which is exactly what the Trino Helm chart does today -- I will create a separate issue just for this part of chart over on that project 😎 ).

Here is more info on this from the OpenShift docs:  https://docs.openshift.com/container-platform/4.15/openshift_images/create-images.html#use-uid_create-images

There are some "hacks"/"hoops" you can jump through (see https://cookbook.openshift.org/users-and-role-based-access-control/how-can-i-enable-an-image-to-run-as-a-set-user-id.html) to get around this, but it is a bit of a complicated process and is often frowned upon by cluster administrators. So I think it would be better if the Trino image could just run out-of-the box on OpenShift, instead!

For now I have created my own Dockerfile which works around this by essentially just copying the Trino stuff out of the `trinodb/trino` image and running it in its own ubi9 minimal image instead where all of the content is owned by `root:root` and giving the `root` group write access to the `/data/trino` folder.

Here is an example of my Dockerfile which basically does this; it was taken as a copy of https://github.com/trinodb/trino/blob/449/core/docker/Dockerfile and adjusted a bit per the comments.

```docker
# Rebuild Trino image but just allow user to be UBI9's default user so that it will work in OpenShift
# Original Trino Dockerfile: https://github.com/trinodb/trino/blob/master/core/docker/Dockerfile
# How Container User IDs should be handled in OpenShift: https://docs.openshift.com/container-platform/4.15/openshift_images/create-images.html#use-uid_create-images

FROM docker.io/trinodb/trino:449 AS trino

# Use ubi9 minimal as it's more secure
# (but lock the current latest version instead of using "latest" just to reduce unforseen issues; this can and should be adjusted as new versions are built and tested)
FROM registry.access.redhat.com/ubi9/ubi-minimal:9.4-949.1716471857

## Set JAVA_HOME based on what is installed in Trino
## This will need to be updated as changes are made to trinodb/trino
#ARG JDK_VERSION
#ENV JAVA_HOME="/usr/lib/jvm/jdk-${JDK_VERSION}"
ENV JAVA_HOME="/usr/lib/jvm/jdk-jdk-22.0.1+8"

ENV PATH=$PATH:$JAVA_HOME/bin
ENV CATALOG_MANAGEMENT=static
COPY --from=trino $JAVA_HOME $JAVA_HOME

RUN \
    set -xeu && \
    microdnf update -y && \
    microdnf install -y tar less python3 shadow-utils && \
    update-alternatives --install /usr/bin/python python /usr/bin/python3 1 && \
    mkdir -p /usr/lib/trino /data/trino && \
    chmod -R g=u /data/trino && \
    microdnf clean all

## removed these lines since we will use the existing ubi9 user
#    groupadd trino --gid 1000 && \
#    useradd trino --uid 1000 --gid 1000 --create-home && \
#    chown -R "trino:trino" /usr/lib/trino /data/trino

## remove these and just copy all of this from the Trino image
#ARG TRINO_VERSION
#COPY --chown=trino:trino trino-cli-${TRINO_VERSION}-executable.jar /usr/bin/trino
#COPY --chown=trino:trino trino-server-${TRINO_VERSION} /usr/lib/trino
#COPY --chown=trino:trino default/etc /etc/trino
#COPY --chown=trino:trino --from=jvmkill /libjvmkill.so /usr/lib/trino/bin

## and instead copy them all in under root
COPY --from=trino --chown=root:root /usr/bin/trino /usr/bin/trino
COPY --from=trino --chown=root:root /usr/lib/trino /usr/lib/trino
COPY --from=trino --chown=root:root /etc/trino /etc/trino

EXPOSE 8080
CMD ["/usr/lib/trino/bin/run-trino"]
HEALTHCHECK --interval=10s --timeout=5s --start-period=10s \
  CMD /usr/lib/trino/bin/health-check
```

I am not sure 100% what the "right" way to do this is, but just that pretty much anything from Red Hat seems to "just work" (thinking of things like Quarkus-related images, Keycloak, etc) so maybe some better inspiration could be taken from some of those images?  But in theory I do not think it would be so many changes to make it so that `trinodb/trino` can play nicely out-of-the-box with many different container runtimes (Docker, Podman, vanilla Kubernetes, OpenShift, etc ?).

kouzant:
## Description

Trino images do not currently work in Openshift. The core issue is that Openshift run
Pods/Containers as a random UID while Trino image assumes user 'trino' has specific
permissions on certain directories, for instance `/data/trino`

This PR uses Openshift best practices to make files and directories owned by
`trino:0` and make the owner and group permissions the same.

Additionally, this patch installs `nss_wrapper` to fake the Unix login username so
it will always appear as 'trino' from applications instead of the random UID.

- Fixes https://github.com/trinodb/trino/issues/22281

## Release notes

```markdown
## General
* Add support for running Trino in OpenShift. ({issue}`22281`)
```

### DIFF
--- a/core/docker/Dockerfile
+++ b/core/docker/Dockerfile
@@ -44,6 +44,7 @@ RUN \
       curl-minimal grep `# required by health-check` \
       zlib `#required by java` \
       shadow-utils `# required by useradd` \
+      nss_wrapper `# required by run-trino to fake passwd entry on OpenShift` \
       tar `# required to support kubectl cp` && \
       rm -rf /tmp/overlay/var/cache/*
 
--- a/core/docker/Dockerfile
+++ b/core/docker/Dockerfile
@@ -62,13 +62,15 @@ RUN \
     groupadd trino --gid 1000 && \
     useradd trino --uid 1000 --gid 1000 --create-home && \
     mkdir -p /usr/lib/trino /data/trino && \
-    chown -R "trino:trino" /usr/lib/trino /data/trino
+    chown -R "trino:0" /usr/lib/trino /data/trino
 
-COPY --chown=trino:trino trino-cli.jar /usr/bin/trino
-COPY --chown=trino:trino --exclude=bin/darwin-* --exclude=bin/linux-* trino-server /usr/lib/trino
-COPY --chown=trino:trino trino-server/bin/linux-${ARCH} /usr/lib/trino/bin/linux-${ARCH}
-COPY --chown=trino:trino default/etc /etc/trino
-COPY --chown=trino:trino --from=jvmkill /libjvmkill.so /usr/lib/trino/bin
+COPY --chown=trino:0 trino-cli.jar /usr/bin/trino
+COPY --chown=trino:0 --exclude=bin/darwin-* --exclude=bin/linux-* trino-server /usr/lib/trino
+COPY --chown=trino:0 trino-server/bin/linux-${ARCH} /usr/lib/trino/bin/linux-${ARCH}
+COPY --chown=trino:0 default/etc /etc/trino
+COPY --chown=trino:0 --from=jvmkill /libjvmkill.so /usr/lib/trino/bin
+
+RUN chmod -R g=u /usr/lib/trino /data/trino /usr/bin/trino /etc/trino
 
 EXPOSE 8080
 USER trino:trino
--- a/core/docker/bin/run-trino
+++ b/core/docker/bin/run-trino
@@ -2,6 +2,33 @@
 
 set -xeuo pipefail
 
+myuid="$(id -u)"
+mygid="$(id -g)"
+username="${TRINO_USER_NAME:-trino}"
+
+# Force NSS to report $username as the name for our UID, regardless of
+# whether CRI-O has auto-injected a numeric-named entry in /etc/passwd
+# (as happens on OpenShift with random UIDs).
+if [ "$(id -un 2>/dev/null || true)" != "$username" ]; then
+    for wrapper in {/usr,}/lib{64,,/*}/libnss_wrapper.so; do
+        if [ -s "$wrapper" ]; then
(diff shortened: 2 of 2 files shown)


wendigo:
This change is a no-go from my perspective.

kouzant:
> This change is a no-go from my perspective.

Hi @wendigo would you mind elaborating, what is a no-go?

wendigo:
@kouzant introduces changes that are not testable as we don't have infrastructure to even validate whether this works and won't break in the future. Uses LD_PRELOAD which seems a security risk. Generally, you can build a docker image for openshift on a side given a proper infrastructure to test it.

kouzant:
@wendigo I understand your concern on the lack of testing infrastructure, however I believe project adoption will increase as more people are having issues running Trino on Openshift.

As for `nss-wrapper`, to be honest, I don't see it as a security risk since it is a package provided by RedHat, but it is optional for Openshift and I can remove it.

wendigo:
@kouzant non-testable code is problematic to support as it can break any time so there is no value in this change given the increased complexity of the docker file

PR #29104, opened 2026-04-14
User/serafin/unified caching

wendigo:
## Release notes

```markdown
## Section
```


martint (dismissed):
> move cache control to the engine, including cache invalidation, ...

This doesn't make sense in Trino's architecture. The engine is agnostic about underlying connector-specific storage decisions. In particular, it doesn't know what a "file system" is. If we want to teach it about caching, it needs to be done at an appropriate level of abstraction.

martint (changes requested):
Why did you dismiss my review? 

Adding a caching facility makes sense. However, mixing this with the filesystem abstractions that are used within data lake connectors does not. Connectors interact with filesystems. Caching is a higher level and storage/backend agnostic concept.

So, my suggestion is to decouple the caching abstraction from those filesystem abstractions and remove all mentions of "filesystem" from the cache APIs. We don't need to reuse the TrinoFile/Location/etc classes.

wendigo:
@martint block/file cache is inherently different from K/V cache. I don't see a reason to unify these two. Even at the connector level these are two separate set of APIs (LoadingCache vs TrinoFileSystemCache). Block cache allows for random seeks (I want to read a cached header/footer of a file, not a whole file).

wendigo:
@martint `ConnectorFileSystemCache` is storage-agnostic. It operates on `TrinoInput` and `TrinoInputStream` but doesn't know anything about concrete implementations. It just caches some I/O operation, regardless of whether this is file/network/API call or whatnot. I don't want to reimplement Alluxio and introduce some new API for this at this stage.

martint:
I'm not talking about whether it's storage agnostic. I'm talking about the abstraction being a "filesystem" cache (implied by the name ConnectorFileSystemCache). A block cache, where you can read parts of the value, does not automatically mean you're dealing with a file system or files.

wendigo:
@martint current SPI fits use cases we have. I don't have other ideas than that. If you have one - please share one

martint:
> @martint current SPI fits use cases we have. I don't have other ideas than that. If you have one - please share one

1. Don't move the classes from filesystem to SPI. Introduce new ones to represent CacheKey, etc.
2. Don't name the abstraction ConnectorFileSystemCache. Call it BlockCache or BlobCache.
3. Remove all references to "file system" from the cache abstraction we're introducing in the SPI and engine. Specific cache plugin implementations can use a file system, memory, remote distributed cache or whatever they desire under the covers, but that shouldn't leak through the abstraction.

wendigo:
@martint it will duplicate a lot of interfacess with the same API surface. What's the point? What's wrong with the name FileSystem?

martint:
The point is to keep the abstraction at the right level.

I’m not objecting because I think the current API is unusable or because I want duplication for its own sake. It's because this change is moving the cache contract into engine/SPI, and at that level FileSystem is the wrong concept.

Connectors deal with filesystems. The engine should not need to know about “filesystem cache” as the abstraction it exposes. What the engine is managing here is a cache of externally loaded data, with lifecycle, sizing, invalidation, and sharing across catalogs. That is broader than filesystems, even if the first implementation happens to be used by filesystem-backed connectors.

So the problem with the name is that it is not just a name. It leaks the storage model into the SPI and makes the engine-facing contract more specific than it should be. Once this is in SPI, we are effectively saying that engine-managed caching is a filesystem concern.

If the right abstraction requires a small amount of API duplication, that is fine. Reusing the existing filesystem-shaped API is not a benefit if it bakes the wrong abstraction into SPI.

wendigo:
@martint ok please do this work on your own.

PR #29105, opened 2026-04-14
Name iterative optimizer trace spans

electrum:
Use unique, descriptive names such as `Iterative:PushProjectionIntoTableScan` for iterative optimizer trace spans rather than the generic name `IterativeOptimizer` with a list of all the rule names. @nineinchnick found that the list of rule names is huge and can break trace collection systems. The single name is smaller, easier to understand, and directly correlates to the optimizer source code.

My understanding of the optimizer is limited, so the current list of names is the best guess from Codex, but looked reasonable to me. However, the actual name doesn't matter that much. The important thing is that when you are looking at a production trace with a long-running optimizer, the name lets you easily identify it.

## Release notes

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
+++ b/core/trino-main/src/main/java/io/trino/sql/planner/LogicalPlanner.java
@@ -328,13 +328,11 @@ public class LogicalPlanner
         if (!Span.fromContext(Context.current()).isRecording()) {
             return null;
         }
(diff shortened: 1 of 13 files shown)


nineinchnick (approved):

losipiuk (approved):
Are all the names unique - would be nice if they were

electrum:
@losipiuk Good idea. I had that thought originally, but wasn't sure if it was necessary. I'll update this so that the shared/reused optimizers have a separate name per invocation.

electrum:
@coderabbitai full review

PR #29130, opened 2026-04-15
Close Reader in JsonInputFunctions

tejasae-afk:
This in JsonInputFunctions only surfaces in the edge case, so may not be a priority. The resource opened there can leak if JsonInputFunctions exits on an error path. this patch moves the allocation into try-with-resources so cleanup happens on every exit path.

Happy to revise the approach or close this if it doesn’t fit — you know the codebase far better than I do.


martint:
Thanks @tejasae-afk. There's no resource leak in this case. The reader is an ephemeral view over an in-memory byte sequence (Slice).

PR #29131, opened 2026-04-16
Plugins for KingbaseES(mysql、oracle、sqlserver mode）

zbw1231:
## Description
We offered three plugins for KingbaseES which contain three mode（mysql、oracle、sqlserver）.

The formats of these three plug-ins are similar to the MySQL, Oracle, and SQL Server plug-ins, respectively.


ebyhr:
https://en.kingbase.com.cn/ 
> 100% Oracle Compatibility

We already have the Oracle connector. Why do we need a new connector?

zbw1231:
> https://en.kingbase.com.cn/
> 
> > 100% Oracle Compatibility
> 
> We already have the Oracle connector. Why do we need a new connector?

In this commit, the Oracle plugin is intended for connecting Trino to Kingbase running in Oracle compatibility mode, rather than to a native Oracle database. Kingbase provides three compatibility modes: MySQL mode, Oracle mode, and SQL Server mode.

PR #29133, opened 2026-04-16
Avoid static target name in renaming table tests

ebyhr:
## Description

The engine verifies whether the target table already exists before the connector attempts to rename any tables.

SaaS connectors such as BigQuery may contain the same table name in the test environment, 
so this PR randomizes the table name.

https://github.com/trinodb/trino/blob/da5f973f44a9e1a8b71a47134dea8798b6f8a0e5/core/trino-main/src/main/java/io/trino/execution/RenameTableTask.java#L107-L115

## Release notes

### DIFF
--- a/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java
+++ b/testing/trino-testing/src/main/java/io/trino/testing/BaseConnectorSmokeTest.java
@@ -433,7 +433,7 @@ public abstract class BaseConnectorSmokeTest
             throws Exception
     {
         if (!hasBehavior(SUPPORTS_RENAME_TABLE)) {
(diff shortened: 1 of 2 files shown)


utafrali (approved):
The fix is correct and well-targeted. Static target names like `yyyy` can collide with pre-existing tables in shared SaaS environments (e.g. BigQuery), causing the engine-level existence check in `RenameTableTask` to throw before the connector's own rejection, breaking the expected error assertion. Randomizing the target name with `randomNameSuffix()` prevents that collision without requiring any cleanup, since all three call sites are in the `!hasBehavior` early-return path where the query always fails.

wendigo (approved):

PR #29137, opened 2026-04-16
Run only CTAS in Redshift FTE smoke test

ebyhr:
## Description

The purpose of TestRedshiftQueryFailureRecoverySmokeTest is to run CTAS only. 
However, it currently runs all FTE tests. 

- Relates to https://github.com/trinodb/trino/pull/18480

## Release notes

### DIFF
--- a/plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestRedshiftQueryFailureRecoverySmokeTest.java
+++ b/plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestRedshiftQueryFailureRecoverySmokeTest.java
@@ -39,4 +39,48 @@ public class TestRedshiftQueryFailureRecoverySmokeTest
                 .finishesSuccessfully()
                 .cleansUpTemporaryTables();
     }
+
+    @Test
+    @Override
+    protected void testCreateTable() {}
+
+    @Test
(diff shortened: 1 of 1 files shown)


chenjian2664 (approved):
Can we also rename the test class if possible?

wendigo (approved):

plugin/trino-redshift/src/test/java/io/trino/plugin/redshift/TestRedshiftQueryFailureRecoverySmokeTest.java
```diff
@@ -39,4 +39,48 @@ void testCreateTableAsSelect()
                 .finishesSuccessfully()
                 .cleansUpTemporaryTables();
     }
+
+    @Test
```
findinpath, line 43: The commit message says what the code changes do.

However, it doesn't contain a rationale about why they have been done.
In any case, if deleting the ephemeral redshift cluster takes a long time (20min+), it doesn't necessarily bring much if we run just one test because in the end the build job will still fail :( 

Maybe it is worth exploring using Redshift serverless.

ebyhr:
> Can we also rename the test class if possible?

Please feel free to rename as a follow-up. I couldn't think of a better name.

PR #29138, opened 2026-04-16
Add `ConnectorSplitAddressProvider` in SPI

chenjian2664:
## Description

`ConnectorSplit` is intended to be a fully serializable data object
from coordinator to workers. Introduce `ConnectorSplitAddressProvider`,
supplies `addresses` and `isRemotelyAccessible` at scheduling time on
the coordinator, separate from the split data.
Also, add the `IcebergSplitAddressProvider` to support in the Iceberg
connector.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/connector/CatalogServiceProviderModule.java
+++ b/core/trino-main/src/main/java/io/trino/connector/CatalogServiceProviderModule.java
@@ -37,6 +37,7 @@ import io.trino.spi.connector.ConnectorIndexProvider;
 import io.trino.spi.connector.ConnectorNodePartitioningProvider;
 import io.trino.spi.connector.ConnectorPageSinkProvider;
 import io.trino.spi.connector.ConnectorPageSourceProviderFactory;
+import io.trino.spi.connector.ConnectorSplitAddressProvider;
 import io.trino.spi.connector.ConnectorSplitManager;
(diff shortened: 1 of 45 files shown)


core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplit.java
```diff
@@ -40,7 +40,7 @@ default List<HostAddress> getAddresses()
         return List.of();
     }
 
-    default SplitWeight getSplitWeight()
+    default SplitWeight splitWeight()
```
ebyhr: Why do you rename this method? This class has `getAddresses` method. I don't think we should use both `get` prefixed and non-prefixed getters in a class.

chenjian2664: reverted the change

electrum:
Can you move the connector changes to a separate commit from the SPI/engine changes?

Also, are the existing methods on `ConnectorSplit` now deprecated?

raunaqmorarka:
We're working on an alternative to this, its not exactly achieving what I was hoping for

raunaqmorarka:
https://github.com/trinodb/trino/pull/29182 is the approach i'd like to go with

PR #29140, opened 2026-04-16
Expose number of non-finished tasks on worker as metric

losipiuk:
## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/execution/SqlTaskManager.java
+++ b/core/trino-main/src/main/java/io/trino/execution/SqlTaskManager.java
@@ -136,6 +136,7 @@ public class SqlTaskManager
 
     private final long queryMaxMemoryPerNode;
 
+    private volatile long activeTasks;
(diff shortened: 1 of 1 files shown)


findepi (approved):

core/trino-main/src/main/java/io/trino/execution/SqlTaskManager.java
```diff
@@ -136,6 +136,7 @@ public class SqlTaskManager
 
     private final long queryMaxMemoryPerNode;
 
+    private volatile long activeTasks;
```
wendigo, line 139: Wouldn't it be better to use LongAdder and increment on new task and decrement when it's finished aggregating when JMX is fetched?

losipiuk: I do not want to track when it gets finsished. Also we would need to watch for removal from cache without finishing. ... overall this is simpler.

PR #29141, opened 2026-04-16
Test Ignite handling of decimal type with negative scale

### LINKED ISSUE #28416: Test Ignite handling of decimal type with negative scale
The connector claims support for decimal with negative scale

https://github.com/trinodb/trino/blob/dfd75cf4f92c5bbe4929375e701a31e5d8c548c4/plugin/trino-ignite/src/main/java/io/trino/plugin/ignite/IgniteClient.java#L253

it's not tested and may not work

- see https://github.com/trinodb/trino/issues/28388
- see https://github.com/trinodb/trino/pull/28414

emanuelsdsj:
\`IgniteClient.toColumnMapping\` maps \`decimal(p, -s)\` to \`decimal(p+s, 0)\` but
this code path was not tested. Negative scale decimal columns can only be
created directly in Ignite (Trino does not expose negative scale in its type
system), so the test uses \`igniteCreateAndInsert\` to verify that Trino reads
such columns correctly.

Fixes #28416


emanuelsdsj:
Someone is already working on the issue, and i still need to submit the CLA.

PR #29143, opened 2026-04-16
Add cache to ldap group provider

### LINKED ISSUE #29134: trino-ldap-group-provider LDAP lookups cause significant delay for each query
### Trino version

480

### Please describe the bug

The trino-ldap-group-provider is used to fetch groups which are then used in the access control logic.
The problem is that the LdapFilteringGroupProvider or LdapSingleQueryGroupProvider are performing 2 LDAP lookups for each SQL query in Trino.

This causes a delay of multiple seconds, which is annoying when using Trino interactively.

Would it be possible to introduce a cache for the group information retrieved from LDAP ( with a configurable evection time)

mwullink:
## Description

This PR adds caching to the  ldap group provider,

Using the ldap group provider, means every Trino query will result into connection to ldap server and ldap search query. This adds multiple seconds of delay to each query. This is very annoying when using Trino interactively. 

This PR adds caching to the  ldap group provider, and a configuration option for setting the cache TTL.

- Fixes #29134

## Release notes

```markdown
## Section
* Added caching for the ldap group provider and `ldap.group-cache-ttl` configuration option ({issue}`29142 `)
```


mrj006:
This is quite useful actually; I've got Permission Denied when the LDAP server rate-limits the connections (which is quite misleading for users).
@ebyhr I see you interacted with the open issue before. Can you approve this?

plugin/trino-ldap-group-provider/src/test/java/io/trino/plugin/ldapgroup/TestLdapGroupProviderIntegration.java
```diff
@@ -267,6 +268,60 @@ private void assertGetGroupsConcurrently(ConfigBuilder configBuilder)
...
+            // Test with cache enabled (10 minute TTL)
+            Map<String, String> config = configBuilder.apply(new HashMap<>(baseConfig));
+            config.put("ldap.group-cache-ttl", "10m");
+            GroupProvider groupsProvider = factory.create(config);
+
+            // First call - should query LDAP
+            long start = System.nanoTime();
+            Set<String> groups1 = groupsProvider.getGroups("alicea");
+            long firstCallTime = System.nanoTime() - start;
+            assertThat(groups1).containsAll(ImmutableSet.of("clients", "qualityAssurance", "developers"));
```
ebyhr: Use `containsExact*` as much as possible. Same for other places.
```java
assertThat(groups1).containsExactlyInAnyOrder("clients", "qualityAssurance", "developers");
```

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapGroupProviderConfig.java
```diff
@@ -110,4 +115,19 @@ public LdapGroupProviderConfig setLdapUseGroupFilter(boolean ldapUseGroupFilter)
...
     }
+
+    @NotNull
+    @MinDuration("0ms")
+    public Duration getGroupsCacheTtl()
+    {
+        return groupsCacheTtl;
+    }
+
+    @Config("ldap.group-cache-ttl")
```
ebyhr: Please document this property in `group-mapping.md`.

plugin/trino-ldap-group-provider/src/main/java/io/trino/plugin/ldapgroup/LdapFilteringGroupProvider.java
```diff
@@ -61,24 +66,39 @@ public LdapFilteringGroupProvider(
         combinedGroupSearchFilter = filteringConfig.getLdapGroupsSearchFilter()
                 .map(filter -> String.format("(&(%s)(%s={0}))", filter, groupsSearchMemberAttribute))
                 .orElse(String.format("(%s={0})", groupsSearchMemberAttribute));
+
+        groupsCache = EvictableCacheBuilder.newBuilder()
+                .expireAfterWrite(Duration.ofMillis(config.getGroupsCacheTtl().toMillis()))
+                .maximumSize(1000)
```
gertjanal: I think this should be configurable too to get a grip on the memory usage

gertjanal:
I have tested this and it works as expected

mwullink:
if fixed the review comments, but i accidentally made my changes on the master and not on a feature branch, now checks fail.

what to do now? create new PR from feature branch?

mwullink:
this PR is replaced by: https://github.com/trinodb/trino/pull/30436

PR #29144, opened 2026-04-16
Avoid Iceberg fileScanTask.spec() calls, which contains expensive json parsing

jinyangli34:
## Description
Reduce call of Iceberg `task.spec()` if specs map is available.

FileScanTask.spec() call contains potentially expensive json parsing as seen in the code listing below:

https://github.com/apache/iceberg/blob/49839545d0027f17970868c5044f8005abce1267/core/src/main/java/org/apache/iceberg/BaseContentScanTask.java#L68-L77

## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
@@ -803,11 +803,6 @@ public final class IcebergUtil
      * Returns a map from fieldId to serialized partition value containing entries for all identity partitions.
      * {@code null} partition values are represented with {@link Optional#empty}.
      */
-    public static Map<Integer, Optional<String>> getPartitionKeys(FileScanTask scanTask)
-    {
-        return getPartitionKeys(scanTask.file().partition(), scanTask.spec());
-    }
-
     public static Map<Integer, Optional<String>> getPartitionKeys(StructLike partition, PartitionSpec spec)
     {
         ImmutableMap.Builder<Integer, Optional<String>> partitionKeys = ImmutableMap.builder();
@@ -1335,6 +1330,13 @@ public final class IcebergUtil
     public static ManifestReader<? extends ContentFile<?>> readerForManifest(ManifestFile manifest, Table table)
     {
         return readerForManifest(manifest, table.io(), table.specs());
+    }
+
+    public static PartitionSpec getFileScanPartitionSpec(FileScanTask fileScanTask, Map<Integer, PartitionSpec> specs)
+    {
+        int specId = fileScanTask.file().specId();
+        PartitionSpec spec = specs.get(specId);
+        return requireNonNull(spec, "Table specs doesn't contain specId: " + specId);
     }
 
     public static ManifestReader<? extends ContentFile<?>> readerForManifest(ManifestFile manifest, FileIO fileIO, Map<Integer, PartitionSpec> specsById)
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
@@ -321,7 +321,7 @@ public final class FilesTablePageSource
     {
         if (partitionColumnType.isPresent() && columnNameToIndex.containsKey(FilesTable.PARTITION_COLUMN_NAME)) {
             PartitionSpec partitionSpec = idToPartitionSpecMapping.get(contentFile.specId());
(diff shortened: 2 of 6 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -1335,6 +1330,18 @@ public static ManifestReader<? extends ContentFile<?>> readerForManifest(Manifes
         return readerForManifest(manifest, table.io(), table.specs());
     }
 
+    public static PartitionSpec getFileScanPartitionSpec(FileScanTask fileScanTask, Map<Integer, PartitionSpec> specs)
+    {
+        if (fileScanTask.file() != null) {
+            int specId = fileScanTask.file().specId();
+            if (specs.containsKey(specId)) {
```
chenjian2664: Could you please comment when the `specs` don't contain the `specId`?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/files/FilesTablePageSource.java
```diff
@@ -274,7 +274,7 @@ private void writePartitionColumns(ContentFile<?> contentFile)
     {
         if (partitionColumnType.isPresent() && columnNameToIndex.containsKey(FilesTable.PARTITION_COLUMN_NAME)) {
             PartitionSpec partitionSpec = idToPartitionSpecMapping.get(contentFile.specId());
-            StructLikeWrapperWithFieldIdToIndex partitionStruct = createStructLikeWrapper(partitionSpec.partitionType(), contentFile.partition());
+            StructLikeWrapperWithFieldIdToIndex partitionStruct = createStructLikeWrapper(partitionSpec, contentFile.partition());
```
findinpath, line 324: please add to the description of the PR 

> `FileScanTask.spec()` call contains potentially expensive json parsing as seen in the code listing below:

> https://github.com/apache/iceberg/blob/49839545d0027f17970868c5044f8005abce1267/core/src/main/java/org/apache/iceberg/BaseContentScanTask.java#L68-L77

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
raunaqmorarka: use `.get` directly and null check on that instead of the redundant `containsKey`

raunaqmorarka:
@coderabbitai review

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
jinyangli34: I don't think there should be such a case. It's just for safeguard.
Changed to specs.get + null check as suggested [here](https://github.com/trinodb/trino/pull/29144#discussion_r3111761712)

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -1335,6 +1330,19 @@ public static ManifestReader<? extends ContentFile<?>> readerForManifest(Manifes
         return readerForManifest(manifest, table.io(), table.specs());
     }
 
+    public static PartitionSpec getFileScanPartitionSpec(FileScanTask fileScanTask, Map<Integer, PartitionSpec> specs)
+    {
+        if (fileScanTask.file() != null) {
```
raunaqmorarka: `file` cannot be null, pls remove that check

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
```diff
@@ -1335,6 +1330,19 @@ public static ManifestReader<? extends ContentFile<?>> readerForManifest(Manifes
...
+            int specId = fileScanTask.file().specId();
+            PartitionSpec spec = specs.get(specId);
+            if (spec != null) {
```
raunaqmorarka: It's guaranteed to be there, we should be using `checkState` to assert it and return

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
jinyangli34: removed

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergUtil.java
jinyangli34: updated to return requireNonNull

JohnFitzpatrick44 (approved):

raunaqmorarka (approved):

raunaqmorarka:
Please rebase to latest master

PR #29171, opened 2026-04-20
Add CLAUDE.md with guidance for AI coding assistants

raunaqmorarka:
## Summary
- Adds a top-level `CLAUDE.md` documenting Trino-specific conventions for AI coding assistants (Claude Code and similar tools that auto-load `CLAUDE.md` as project context).
- Covers code style do's and don'ts, Airlift import ordering and brace placement, configuration property naming and lifecycle (`@ConfigDescription`, `@LegacyConfig`, `@DefunctConfig`), and Maven build rules.
- Points to `.github/DEVELOPMENT.md` for topics already documented there (Web UI build, release process, Vector API, IDE setup rationale) rather than duplicating them.

## Test plan
- [x] File renders correctly on GitHub

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,40 @@
+# Trino — Claude guidance
+
+**Before writing Java code, you must first read [`.github/DEVELOPMENT.md`](.github/DEVELOPMENT.md)
+in full** — it's the authoritative source for code-style rules (mocks, `var`, switch statements,
+method naming, `format()`, `TrinoException` error codes, AssertJ, Guava immutables, and more).
+This file intentionally does not duplicate those rules; skipping the read means missing them.
+Violations are also caught mechanically by modernizer
+([`.mvn/modernizer/violations.xml`](.mvn/modernizer/violations.xml)), checkstyle (from Airbase),
+and IntelliJ inspections via `mcp__idea__get_file_problems`.
+
+For other topics not covered here (Web UI build, release process, Vector API, IDE setup rationale),
+see the same `DEVELOPMENT.md`.
+
+## JetBrains MCP server
+
+Trino is developed in IntelliJ. If you run Claude Code with the JetBrains MCP server enabled,
+the assistant can drive the IDE directly. Install: https://github.com/JetBrains/mcp-jetbrains.
+
+When available, Claude should prefer these over shell equivalents:
+- `mcp__idea__reformat_file` after Java edits — applies the Airlift formatter; Claude cannot
+  reproduce it by hand.
+- `mcp__idea__get_file_problems` before committing — surfaces IntelliJ inspection results
+  (error-prone, unused imports, nullability) without a full Maven build.
+- `mcp__idea__search_symbol` / `mcp__idea__get_symbol_info` for symbol navigation in a codebase
+  with many overloaded names like `Metadata`, `Session`, `Block`.
+- `mcp__idea__rename_refactoring` for API renames — safer than text substitution.
+
+## Java formatting
+
+Run `mcp__idea__reformat_file` after every Java edit — it applies the Airlift scheme imported into
+IntelliJ. Rules not covered by the formatter:
+
+- No wildcard imports (e.g. `import io.trino.spi.*`) — checkstyle catches these on build; easier
+  to avoid writing them.
+- Braces required around single-statement `if` / `for` / `while` bodies — the formatter does not
+  add missing braces.
+- No `@author` in JavaDoc — commit history is the record.
+
+Topic-specific conventions live under [`.claude/rules/`](.claude/rules/) and auto-load when Claude
+reads matching files (e.g. `*Config.java` triggers the config-properties rule).
--- a/.claude/rules/trino-config-properties.md
+++ b/.claude/rules/trino-config-properties.md
@@ -1,0 +1,37 @@
+---
+paths:
+  - "**/*Config.java"
+  - "**/*SessionProperties.java"
+---
+
+# Trino — configuration and session properties
+
+Naming:
+- Config property names use **dashes**, e.g. `hive.max-partitions-per-scan`.
+- Session property names use **snake_case**, e.g. `max_partitions_per_scan`.
+
+Adding a property:
+- Every `@Config` setter gets an `@ConfigDescription("…")`.
+- Every session property registration includes a description.
+- Credentials and other secrets get `@ConfigSecuritySensitive` so values are redacted in logs
+  and info endpoints.
+
+Renaming a config:
+- Add a `@Deprecated` setter with `@LegacyConfig(value = "old.name", replacedBy = "new.name")`
+  that writes to the same field as the new `@Config("new.name")` setter. The old name keeps
+  working as a backward-compatible alias.
+
+Removing a config:
+- Add the current name **and** any `@LegacyConfig` names to `@DefunctConfig` on the class so
+  startup fails loudly if the config is still set.
+- Remove the matching session property from `SystemSessionProperties.java` (or the connector's
+  session-properties class) if one exists.
+
+Testing a new config:
+- Add a matching `TestMyConfig` using Airlift's `ConfigAssertions` — see existing `Test*Config`
+  classes for the `testDefaults()` / `testExplicitPropertyMappings()` pattern.
+
+Other conventions:
+- Validation annotations (`@NotNull`, `@Min`, `@MinDuration`, etc.) go on setters, not fields.
+- Don't store the config object as a field — read values in the constructor and keep those
+  instead.
--- a/.gitignore
+++ b/.gitignore
@@ -34,4 +34,5 @@ product-test-reports
 **/dependency-reduced-pom.xml
 core/trino-web-ui/src/main/resources/webapp/dist/
 .node
-.claude
+.claude/settings.local.json
+.claude/worktrees/


CLAUDE.md
```diff
@@ -0,0 +1,68 @@
+# Trino — Claude guidance
+
+For topics not covered here (Web UI build, release process, Vector API, IDE setup rationale), see [`.github/DEVELOPMENT.md`](.github/DEVELOPMENT.md).
+
+## Code style — apply on every change
+
+Patterns to avoid:
+- **No mocking libraries.** Hand-write fakes — mocks encourage fragile tests and hide untestable designs.
+- **No `var`.** Use explicit types.
+- **No `default` clause in enum-exhaustive switches.** Handle the unknown case after the switch so Error Prone's `MissingCasesInEnumSwitch` can flag new values.
```
martint: We should be using the new switch syntax and let the compiler flag missing cases.

raunaqmorarka:
@coderabbitai review

CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+- **Avoid the ternary operator** except for trivial expressions.
+
+APIs to prefer:
+- **AssertJ** for complex assertions. Airlift's `Assertions` for cases AssertJ doesn't cover.
+- **Guava immutable collections** (`ImmutableList`, `ImmutableMap`, …) over JDK unmodifiable wrappers — deterministic iteration.
+- **`format()`** (statically imported) over string concatenation for anything beyond trivial appends. Avoid both in performance-critical paths.
+- **`TrinoException` with an error code**, e.g. `new TrinoException(HIVE_TOO_MANY_OPEN_PARTITIONS, …)`.
+
+## Java formatting and tooling
+After non-trivial edits, run `mcp__idea__reformat_file` — the formatter fixes most issues. When IntelliJ IDEA is running with the MCP server, prefer `mcp__idea__*` tools over shell text search for symbol/reference lookups (AST-aware, cheaper).
```
martint: What's this?

CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+
+Imports — groups separated by a single blank line, alphabetical within each, no wildcards:
```
martint: This should follow airlift's code formatting rules.

CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+  3. `javax.`
+  4. `java.`
+  5. static imports
+
+When adding imports to an existing file, match its grouping exactly.
+
+- Opening brace on its **own line** for class, interface, enum, and method declarations.
+- Opening brace on the **same line** for control flow (`if`, `for`, `while`, `try`, lambdas, …).
+- `else`, `catch`, `finally` start a new line after the preceding `}` (not cuddled).
+- Braces required around single-statement `if` / `for` / `while` bodies.
```
martint: Same here. Instead of inventing new rules, this should refer to airlift's code formatting rules

CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+- Add `@LegacyConfig(value = "old.name", replacedBy = "new.name")` on the new setter so users on the old name get a clear migration error instead of a silent default.
+
+Removing a config:
+- Add the current name **and** any `@LegacyConfig` names to `@DefunctConfig` on the class so startup fails loudly if the config is still set.
+- Remove the matching session property from `SystemSessionProperties.java` (or the connector's session-properties class) if one exists.
+
+## Maven builds
+
+Core rules:
+- Use `mvnd`, never `mvn` / `./mvnw` / `mvnw`, even though `DEVELOPMENT.md` shows `./mvnw`.
```
martint: Why? What if mvnd is not installed in the developer's environment?

CLAUDE.md
raunaqmorarka: Replaced the rule with guidance to prefer switch expressions, which give compile-time exhaustiveness on enums natively. Thanks.

CLAUDE.md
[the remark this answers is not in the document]
raunaqmorarka: Dropped the Maven section for now. mvnd saves real compile time over the wrapper but needs some guardrails to be used correctly (worktree install races, stale jar pickup when testing downstream modules, etc.). I still think relying on the mvn wrapper by default is a clunkier experience and mvnd is pretty easy to install, but that's a separate conversation.

CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+- Never run `test` or `surefire:test` alone — compile dependencies first.
```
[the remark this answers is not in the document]
raunaqmorarka: Dropped the Maven section for now. mvnd saves real compile time over the wrapper but needs some guardrails to be used correctly (worktree install races, stale jar pickup when testing downstream modules, etc.). I still think relying on the mvn wrapper by default is a clunkier experience and mvnd is pretty easy to install, but that's a separate conversation.

CLAUDE.md
raunaqmorarka: Dropped the Maven section for now. mvnd saves real compile time over the wrapper but needs some guardrails to be used correctly (worktree install races, stale jar pickup when testing downstream modules, etc.). I still think relying on the mvn wrapper by default is a clunkier experience and mvnd is pretty easy to install, but that's a separate conversation.

CLAUDE.md
raunaqmorarka: These are tool names the JetBrains MCP server exposes to Claude Code — `mcp__idea__reformat_file` invokes IntelliJ's built-in reformat, and `mcp__idea__*` covers AST-level symbol/reference lookups. IntelliJ is already the expected IDE for Trino development (see the Airlift codestyle `IntelliJIdea2019` scheme), so this isn't an exotic dependency. Without the hint, Claude defaults to shell reformatting and grep, which is both slower and less accurate. Keeping it.

CLAUDE.md
raunaqmorarka: Trimmed the section down to three rules the Airlift IntelliJ formatter does not apply on its own: no wildcard imports, missing braces around single-statement bodies, and the `@author` convention. Everything else — import group order, brace placement, else/catch/finally — is left to `mcp__idea__reformat_file`, which the added JetBrains MCP section establishes as a prerequisite.

CLAUDE.md
raunaqmorarka: Trimmed the section down to three rules the Airlift IntelliJ formatter does not apply on its own: no wildcard imports, missing braces around single-statement bodies, and the `@author` convention. Everything else — import group order, brace placement, else/catch/finally — is left to `mcp__idea__reformat_file`, which the added JetBrains MCP section establishes as a prerequisite.

CLAUDE.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Java formatting and tooling
+After non-trivial edits, run `mcp__idea__reformat_file` — the formatter fixes most issues. When IntelliJ IDEA is running with the MCP server, prefer `mcp__idea__*` tools over shell text search for symbol/reference lookups (AST-aware, cheaper).
+
+Imports — groups separated by a single blank line, alphabetical within each, no wildcards:
+  1. `com.`
+  2. `io.` (includes `io.trino`)
+  3. `javax.`
+  4. `java.`
+  5. static imports
```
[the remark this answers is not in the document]
raunaqmorarka: Dropped the inlined import-group list — `mcp__idea__reformat_file` with the Airlift scheme handles ordering correctly, including `jakarta.*` and `org.*`, so duplicating the list here only risks drift.

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+- `mcp__idea__reformat_file` after Java edits — applies the Airlift formatter; Claude cannot reproduce it by hand.
+- `mcp__idea__get_file_problems` before committing — surfaces IntelliJ inspection results (error-prone, unused imports, nullability) without a full Maven build.
+- `mcp__idea__search_symbol` / `get_symbol_info` for symbol navigation in a codebase with many overloaded names like `Metadata`, `Session`, `Block`.
+- `mcp__idea__rename_refactoring` for API renames — safer than text substitution.
+
+For plain text search, `rg` is fine — the MCP search tools are not meaningfully better.
+
+## Code style — apply on every change
+
+Patterns to avoid:
```
martint: These are all covered in DEVELOPMENT.md. No need to repeat here.

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+For topics not covered here (Web UI build, release process, Vector API, IDE setup rationale), see [`.github/DEVELOPMENT.md`](.github/DEVELOPMENT.md).
+
+## JetBrains MCP server
+
+Trino is developed in IntelliJ (see the Airlift codestyle `IntelliJIdea2019` scheme). If you run Claude Code with the JetBrains MCP server enabled, the assistant can drive the IDE directly. Install: https://github.com/JetBrains/mcp-jetbrains.
+
+When available, Claude should prefer these over shell equivalents:
+- `mcp__idea__reformat_file` after Java edits — applies the Airlift formatter; Claude cannot reproduce it by hand.
+- `mcp__idea__get_file_problems` before committing — surfaces IntelliJ inspection results (error-prone, unused imports, nullability) without a full Maven build.
+- `mcp__idea__search_symbol` / `get_symbol_info` for symbol navigation in a codebase with many overloaded names like `Metadata`, `Session`, `Block`.
```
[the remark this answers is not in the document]
raunaqmorarka: Fixed — prefixed `get_symbol_info` with `mcp__idea__` for consistency.

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+- `mcp__idea__rename_refactoring` for API renames — safer than text substitution.
+
+For plain text search, `rg` is fine — the MCP search tools are not meaningfully better.
+
+## Code style — apply on every change
+
+Patterns to avoid:
+- **No mocking libraries.** Hand-write fakes — mocks encourage fragile tests and hide untestable designs.
+- **No `var`.** Use explicit types.
```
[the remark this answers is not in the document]
raunaqmorarka: Softened to "`var` discouraged — prefer explicit types except where inference aids readability" to match `DEVELOPMENT.md`'s wording and the codebase's actual practice (79 occurrences across 39 files).

CLAUDE.md
```diff
@@ -0,0 +1,55 @@
...
+- **`format()`** (statically imported) over string concatenation for anything beyond trivial appends. Avoid both in performance-critical paths.
+- **`TrinoException` with an error code**, e.g. `new TrinoException(HIVE_TOO_MANY_OPEN_PARTITIONS, …)`.
+
+## Java formatting
+
+Run `mcp__idea__reformat_file` after every Java edit — it applies the Airlift scheme imported into IntelliJ. Rules not covered by the formatter:
+
+- No wildcard imports (e.g. `import io.trino.spi.*`) — checkstyle catches these on build; easier to avoid writing them.
+- Braces required around single-statement `if` / `for` / `while` bodies — the formatter does not add missing braces.
+- No `@author` in JavaDoc — commit history is the record.
```
martint: Wouldn't it be better to just point it to where it to look for code style rules? (modernizer rules, checkstyle rules and airlift code style file)

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
+# Trino — Claude guidance
+
+For topics not covered here (Web UI build, release process, Vector API, IDE setup rationale), see [`.github/DEVELOPMENT.md`](.github/DEVELOPMENT.md).
+
+## JetBrains MCP server
+
+Trino is developed in IntelliJ (see the Airlift codestyle `IntelliJIdea2019` scheme). If you run Claude Code with the JetBrains MCP server enabled, the assistant can drive the IDE directly. Install: https://github.com/JetBrains/mcp-jetbrains.
```
findepi: Nit, can you please wrap long lines?
(best with intellij's reformat. my claude doesn't know how to reformat .md correctly)

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+- **AssertJ** for complex assertions. Airlift's `Assertions` for cases AssertJ doesn't cover.
+- **Guava immutable collections** (`ImmutableList`, `ImmutableMap`, …) over JDK unmodifiable wrappers — deterministic iteration.
+- **`format()`** (statically imported) over string concatenation for anything beyond trivial appends. Avoid both in performance-critical paths.
+- **`TrinoException` with an error code**, e.g. `new TrinoException(HIVE_TOO_MANY_OPEN_PARTITIONS, …)`.
+
+## Java formatting
+
+Run `mcp__idea__reformat_file` after every Java edit — it applies the Airlift scheme imported into IntelliJ. Rules not covered by the formatter:
+
+- No wildcard imports (e.g. `import io.trino.spi.*`) — checkstyle catches these on build; easier to avoid writing them.
```
findepi: Isn't that covered by airlift formatter?

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+- Braces required around single-statement `if` / `for` / `while` bodies — the formatter does not add missing braces.
```
findepi: my intellij somehow adds braces for `if` during reformat.

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+- No `@author` in JavaDoc — commit history is the record.
+
+## Configuration properties and session properties
+
+Naming:
+- Config property names use **dashes**, e.g. `hive.max-partitions-per-scan`.
+- Session property names use **snake_case**, e.g. `max_partitions_per_scan`.
+
+Adding a property:
+- Every `@Config` setter gets an `@ConfigDescription("…")`.
```
findepi: worth mentioning that security sentive props should be annotated?

CLAUDE.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Java formatting
+
+Run `mcp__idea__reformat_file` after every Java edit — it applies the Airlift scheme imported into IntelliJ. Rules not covered by the formatter:
+
+- No wildcard imports (e.g. `import io.trino.spi.*`) — checkstyle catches these on build; easier to avoid writing them.
+- Braces required around single-statement `if` / `for` / `while` bodies — the formatter does not add missing braces.
+- No `@author` in JavaDoc — commit history is the record.
+
+## Configuration properties and session properties
```
findepi: It could be better to move config management to separate how to (skill) doc, not to overload the default context.

findepi:
@hashhar PTAL

CLAUDE.md
raunaqmorarka: Dropped both the "Patterns to avoid" and "APIs to prefer" lists, and reworked the preamble to tell Claude explicitly to read `DEVELOPMENT.md` before writing Java code. I also added pointers to modernizer (`.mvn/modernizer/violations.xml`), the checkstyle rules inherited from Airbase, and IntelliJ inspections via `mcp__idea__get_file_problems` so the mechanical enforcement sources are discoverable. One thing to flag: CLAUDE.md is auto-loaded into Claude's context but linked files aren't, so passive "see also" pointers tend to get skipped. The imperative phrasing and named rule categories are there to actually nudge the assistant to open `DEVELOPMENT.md` — we'll see how it holds up in practice.

CLAUDE.md
raunaqmorarka: Dropped both the "Patterns to avoid" and "APIs to prefer" lists, and reworked the preamble to tell Claude explicitly to read `DEVELOPMENT.md` before writing Java code. I also added pointers to modernizer (`.mvn/modernizer/violations.xml`), the checkstyle rules inherited from Airbase, and IntelliJ inspections via `mcp__idea__get_file_problems` so the mechanical enforcement sources are discoverable. One thing to flag: CLAUDE.md is auto-loaded into Claude's context but linked files aren't, so passive "see also" pointers tend to get skipped. The imperative phrasing and named rule categories are there to actually nudge the assistant to open `DEVELOPMENT.md` — we'll see how it holds up in practice.

CLAUDE.md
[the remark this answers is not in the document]
martint: Why even include it? It's in the DEVELOPMENT.md file. If we duplicate the guidance here, we have to keep both in sync. Not ideal.

CLAUDE.md
raunaqmorarka: Tested this with `mcp__idea__reformat_file` on a file with `import java.util.*;` — the wildcard wasn't expanded. The MCP tool invokes "Reformat Code" only (no Optimize Imports, no Cleanup), so it doesn't pick up the Airlift scheme's wildcard-expansion behavior. Checkstyle still catches them at build time, but for Claude's first-draft output it helps to have the rule here. Keeping the bullet.

CLAUDE.md
raunaqmorarka: Also tested — created a file with `if (x > 0) System.out.println("hi");`, ran `mcp__idea__reformat_file`, and the braces were not added. Same reason as the wildcard case: the MCP tool runs plain Reformat Code, not Reformat and Cleanup. Your IDE likely has "Reformat and Cleanup" bound or code cleanup enabled in the reformat dialog, but that doesn't carry over when Claude invokes the formatter through the MCP. Keeping the bullet.

CLAUDE.md
raunaqmorarka: Wrapped at 100 columns to match `.github/CONTRIBUTING.md` and `docs/README.md`'s looser convention. Going tighter (80, per `DEVELOPMENT.md`) produced awkward mid-phrase breaks around long code identifiers like `mcp__idea__get_file_problems`.

CLAUDE.md
raunaqmorarka: Added a bullet: credentials and other secrets get `@ConfigSecuritySensitive` so values are redacted in logs and info endpoints. Good catch.

CLAUDE.md
raunaqmorarka: Don't think this fits a skill. Skills work for task-scoped workflows ("how to add a connector", "how to bisect a perf regression") where there's a clear invocation trigger. Config conventions are declarative rules that should be applied reflexively whenever a `*Config.java` or `SystemSessionProperties.java` file is touched — moving them to an on-demand skill creates the risk that Claude works on a config class without invoking the skill and silently misses the conventions. Fifteen lines of always-loaded context is a reasonable price for reflexive application.

CLAUDE.md
```diff
@@ -0,0 +1,59 @@
...
+
+Adding a property:
+- Every `@Config` setter gets an `@ConfigDescription("…")`.
+- Every session property registration includes a description.
+- Credentials and other secrets get `@ConfigSecuritySensitive` so values are redacted in logs
+  and info endpoints.
+
+Renaming a config:
+- Add `@LegacyConfig(value = "old.name", replacedBy = "new.name")` on the new setter so users on
+  the old name get a clear migration error instead of a silent default.
```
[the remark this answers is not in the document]
raunaqmorarka: Right — verified against Airlift's `ConfigurationMetadata` source: `@LegacyConfig` with `replacedBy` on a `@Config` setter is rejected at startup; the correct pattern is a separate `@Deprecated` legacy setter that aliases to the same field (example: `FeaturesConfig.setWriterMinSize`). Reworded accordingly. Note: your suggestion also placed the annotation on the new `@Config` setter, which Airlift would reject — the legacy setter needs to be separate.

CLAUDE.md
```diff
@@ -0,0 +1,60 @@
...
+
+- No wildcard imports (e.g. `import io.trino.spi.*`) — checkstyle catches these on build; easier
+  to avoid writing them.
+- Braces required around single-statement `if` / `for` / `while` bodies — the formatter does not
+  add missing braces.
+- No `@author` in JavaDoc — commit history is the record.
+
+## Configuration properties and session properties
+
+Naming:
```
dain: Add a testing section that says, "Hey, if you're adding a new one of these, you must add a test. For examples, look at the other config tests."

CLAUDE.md
```diff
@@ -0,0 +1,60 @@
...
+Renaming a config:
+- Add a `@Deprecated` setter with `@LegacyConfig(value = "old.name", replacedBy = "new.name")`
+  that writes to the same field as the new `@Config("new.name")` setter. The old name keeps
+  working as a backward-compatible alias.
+
+Removing a config:
+- Add the current name **and** any `@LegacyConfig` names to `@DefunctConfig` on the class so
+  startup fails loudly if the config is still set.
+- Remove the matching session property from `SystemSessionProperties.java` (or the connector's
+  session-properties class) if one exists.
```
hashhar:
```
Every `@Config` class must have a matching `TestMyConfig` using Airlift's `ConfigAssertions`:
- testDefaults(): verify defaults with `assertRecordedDefaults(recordDefaults(MyConfig.class).setX(defaultValue)...)`
- testExplicitPropertyMappings(): build a `Map<String, String>` of property->value, construct an expected object with non-default values, and call `assertFullMapping(properties, expected)`
```

CLAUDE.md
```diff
@@ -0,0 +1,60 @@
...
+## Configuration properties and session properties
+
+Naming:
+- Config property names use **dashes**, e.g. `hive.max-partitions-per-scan`.
+- Session property names use **snake_case**, e.g. `max_partitions_per_scan`.
+
+Adding a property:
+- Every `@Config` setter gets an `@ConfigDescription("…")`.
+- Every session property registration includes a description.
+- Credentials and other secrets get `@ConfigSecuritySensitive` so values are redacted in logs
```
hashhar:
```
- Validation annotations go on setters (@NotNull, @Min, @Max, @MinDuration, @MaxDuration)
- Never store the config object as a field (it's mutable). Inject it and read values in the constructor.
- Make the config available as `configBinder(binder).bindConfig(MyConfig.class)`
```

CLAUDE.md
```diff
@@ -0,0 +1,60 @@
+# Trino — Claude guidance
+
+Before writing Java code, read [`.github/DEVELOPMENT.md`](.github/DEVELOPMENT.md) — it's the
```
hashhar: If you use `@` to reference the file path I think it forces (at-least Claude) to actually read it. I've had a lot of times where the agent gets lazy and doesn't read the file or just reads the first and last 5 lines.

CLAUDE.md
```diff
@@ -0,0 +1,60 @@
...
+intentionally does not duplicate those rules. Violations are also caught mechanically by modernizer
+([`.mvn/modernizer/violations.xml`](.mvn/modernizer/violations.xml)), checkstyle (from Airbase),
+and IntelliJ inspections via `mcp__idea__get_file_problems`.
+
+For other topics not covered here (Web UI build, release process, Vector API, IDE setup rationale),
+see the same `DEVELOPMENT.md`.
+
+## JetBrains MCP server
+
+Trino is developed in IntelliJ (see the Airlift codestyle `IntelliJIdea2019` scheme). If you run
```
hashhar: can it actually `see the Airlift codestyle`? How, MCP tools? Or do you mean to link to repo here?

And maybe we don't need to tell it about the file, just tell it to use intellij tool to reformat file + apply whatever's not in the file manually.

CLAUDE.md
hashhar: it can be a "rule" to match only `**/*Config.java` but it needs to know this anyway I think. (https://code.claude.com/docs/en/memory#organize-rules-with-claude/rules/)

hashhar (approved):
Good start. Very closely matches what I have as well. I do have to often tell it to rebuild (using `clean` + `-am` or `-amd`) instead of manually trying to `-pl` it's way to a successful build. Sadly haven't managed to find the words to get it to stop doing that.

hashhar:
For testing I have unit test knowledge described in claude.md:

- `-Dair.test.thread-count` as a possible useful thing
- Prefer improving/expanding existing tests over creating new ones. Consolidate related scenarios.
- Prefer clarity over defensive programming in tests. Use direct assertions without defensive checks. Only sanitize what's actually unsafe.

For product tests (click to see full content):

<summary>
<details>

```md
# Product Tests

## Running
```bash
# List available suites and environments
./bin/ptl suite list
./bin/ptl env list

# Run a product test suite
./bin/ptl suite run --suite <suite-name>

# Run a product test group
./bin/ptl test run --environment <env-name> -- -g <group-name1>,<group-name2>

# Run a specific test
./bin/ptl test run --environment <env-name> -- -t <fully-qualified-method-name1>,<fully-qualified-method-name2>
```

## Building
After changing product test code, rebuild **both** modules before running:
```bash
mvnd install -pl :trino-product-tests,:trino-product-tests-launcher -DskipTests -Dair.check.skip-all=true
```
The launcher copies the test jar and server-main into Docker containers, so they must be rebuilt too if required.

## Infrastructure
- Tests run inside a `tests` Docker container on the `ptl-network`
- Trino coordinator container: `presto-master` (port 8080)
- Container names defined in `EnvironmentContainers.java`
- Environment providers live in `trino-product-tests-launcher` under `env/environment/`
- Suite definitions live in `trino-product-tests-launcher` under `suite/suites/`
- Test groups defined in `trino-product-tests-groups/.../TestGroups.java`
- Tempto configuration (JDBC connections, users) in `trino-product-tests-launcher/src/main/resources/docker/trino-product-tests/`

## Framework
- Tests use [Tempto](https://github.com/trinodb/tempto) (TestNG-based, NOT JUnit). See also [Trino product tests README](https://github.com/trinodb/trino/blob/master/testing/trino-product-tests/README.md)
- Tests extend `io.trino.tempto.ProductTest` and use `@Test(groups = {...})` for grouping
- Queries run via JDBC through `QueryExecutor` (e.g. `connectToTrino()`, `onPostgres()`)
- Dependency injection via Guice (`@Inject`, `@Named("databases.trino.server_address")`)
- Multiple user configs (alice, bob, etc.) defined per environment in `tempto-configuration.yaml`

### Test setup: `@BeforeMethodWithContext`
- **DO NOT use `@BeforeClass` or `@BeforeMethod`** - Tempto Guice injection hasn't happened yet at that point
- Use `@BeforeMethodWithContext` (from `io.trino.tempto`) - runs after injection, before each test method
- For one-time setup, guard with `private volatile boolean initialized` (standard Trino pattern):
  ```java
  private volatile boolean initialized;

  @BeforeMethodWithContext
  public void setUp()
  {
      if (initialized) {
          return;
      }
      // ... one-time setup
      initialized = true;
  }
  ```

### Test groups and `PROFILE_SPECIFIC_TESTS`
- `PROFILE_SPECIFIC_TESTS` marks tests that **require a specific environment/profile** to run
- The default test runner **excludes** `profile_specific_tests` - tests in this group are skipped unless explicitly included
- Custom suites that provide the right environment should **NOT** use `PROFILE_SPECIFIC_TESTS` - the suite already guarantees the correct environment via its `EnvironmentProvider`
- Use `PROFILE_SPECIFIC_TESTS` only when the test also belongs to a broader group that runs across multiple environments (e.g. `INSIGHTS` tests that run in both generic and specific profiles)

### Environment providers
- Live in `trino-product-tests-launcher` under `env/environment/`, annotated with `@TestsEnvironment`
- Extend `EnvironmentProvider`, compose base modules (e.g. `StandardMultinode`, `Hadoop`, `Minio`) via constructor injection
- Override `extendEnvironment(Environment.Builder)` to configure containers, copy files, set env vars
- Config files per environment: `conf/environment/<env-name>/config-coordinator.properties` etc., copied into containers via `withCopyFileToContainer()`

### Suite definitions
- Live in `trino-product-tests-launcher` under `suite/suites/`, extend `Suite`
- Override `getTestRuns(EnvironmentConfig)` returning `List<SuiteTestRun>`
- Each `SuiteTestRun` pairs an environment with test groups: `.withGroups(...)`, `.withExcludedGroups(...)`, `.withTests(...)`, `.withExcludedTests(...)`

## Debugging
- View coordinator logs: `docker logs presto-master`
- View test container logs: `docker logs tests`
- Worker logs: `docker logs presto-worker-1` (etc.)
- Filter tests with `-t <fully.qualified.MethodName>` or `-g <group-name>` to iterate quickly on failures
```
</details>
</summary>

CLAUDE.md
raunaqmorarka: Reworked per your pointer — moved the config-properties guidance to `.claude/rules/trino-config-properties.md` with `paths: ["**/*Config.java", "**/*SessionProperties.java"]` so it auto-loads only when Claude reads matching files, verified against the [Claude Code docs](https://code.claude.com/docs/en/memory#organize-rules-with-claude/rules/). `.gitignore` narrowed from the whole `.claude/` to just `settings.local.json` + `worktrees/` so project-level rules are tracked while per-user files stay ignored. Thanks for the pointer — my earlier "always-on context is cheaper" reasoning doesn't hold against path-scoped auto-loading.

CLAUDE.md
raunaqmorarka: Added to the config-properties rule: `TestMyConfig` using Airlift's `ConfigAssertions` is required, pointing at existing `Test*Config` classes for the `testDefaults()` / `testExplicitPropertyMappings()` pattern. The rule auto-loads when Claude reads `*Config.java`, so the testing requirement surfaces in the same context.

CLAUDE.md
raunaqmorarka: Included in the config-properties rule, trimmed to a pointer at existing `Test*Config` classes rather than spelling out `assertRecordedDefaults` / `assertFullMapping` (Claude can crib those from any existing test once it's pointed at the pattern).

CLAUDE.md
raunaqmorarka: Included the setters-not-fields validation rule and the "don't store config as field" convention. Dropped the `configBinder(binder).bindConfig(...)` bullet — that's trivially discoverable from existing `*Module.java`.

CLAUDE.md
raunaqmorarka: Strengthened the imperative — bolded "you must first read", added "in full" to directly counter the reads-first-and-last-5-lines symptom, and spelled out the consequence ("skipping the read means missing them").

On `@path` imports specifically: verified from the [Claude Code docs](https://code.claude.com/docs/en/memory#import-additional-files) that `@` is a full import — it pulls the file's contents into context at launch, not just a hint to read. Applied to `DEVELOPMENT.md` (~300 lines, much of it unrelated to code style like Web UI build, release process, Vector API) that means loading all of it every session.

What I'd actually prefer: split the code-style section out of `DEVELOPMENT.md` into its own file (e.g. `.github/code-style.md`) and `@`-import just that. Keeps the import scoped to what Claude needs. Restructuring `DEVELOPMENT.md` is out of scope for this PR though — separate change.

CLAUDE.md
raunaqmorarka: Good point — the Airlift codestyle parenthetical was aimed at a human reader, not Claude (no URL to fetch; even with one, Claude wouldn't follow external links reliably). Dropped it. The Java formatting section below already tells Claude what to do: run `mcp__idea__reformat_file`, and here are the rules it doesn't auto-fix.

raunaqmorarka:
Thanks for sharing. The testing content fits the `.claude/rules/` pattern we're establishing here — unit-test bullets in `.claude/rules/unit-tests.md` (path-scoped to `**/src/test/**/*.java`), product-test guidance in `.claude/rules/product-tests.md` (scoped to `**/trino-product-tests*/**`). Keeping that out of this PR to avoid scope creep, but happy to open a follow-up that pulls your content across, or defer to you if you'd rather own it directly.

CLAUDE.md
findepi: > The MCP tool invokes "Reformat Code" only (no Optimize Imports, no Cleanup), so it doesn't pick up the Airlift scheme's wildcard-expansion behavior.

We should teach Claude to optimize imports via intellij too then.

CLAUDE.md
findepi: Then there are likely other things we would need to write here, if we cannot have Claude use Intellij's full reformatting power.

CLAUDE.md
raunaqmorarka: Claude can only make intellij do whatever intellij MCP exposes

CLAUDE.md
raunaqmorarka: I imagine Intellij MCP will get better over time, meanwhile it is fine to add more rules here if we need to.
Claude is pretty good about taking cues from existing code though, so we don't need to be exhaustive

PR #29179, opened 2026-04-21
Add branch information to DML access control SPI

ksobolew:
## Description

There are two sides of access control for table branches:

1. Access control for branch management
  a. `CREATE BRANCH ...`, `DROP BRANCH ...` etc.
  b. and associated `GRANT privilege ON BRANCH branch IN TABLE table ...`
2. Access control for accessing data on branches
  a. `SELECT * FROM table FOR VERSION AS OF 'branch'
  b. `INSERT INTO table @ branch ...` etc.
  c. and associated `GRANT SELECT ON BRANCH branch IN TABLE table ...`

Parts 1. and 2.c are already implemented - with some limitations, like inability to manage branch grants on SCHEMA level - but parts 2.a and 2.b were almost entirely missing. This commit implements most of it, which is:

* Adding a `branch` parameter to all relevant access control interface methods
  - covers `SELECT`, `INSERT`, `UPDATE`, `DELETE` and selecting from view with `SECURITY DEFINER` which references a branch in table
  - new methods are added while keeping the old interface for backwards compatibility: the engine (and tests, even in plugins not converted to the new methods) exclusively calls the new methods, which delegate to old methods, so that if the access control plugin implements only the old methods, it will behave the way it used to
  - the minor downside is that the branch information is lost in error messages of default implementations of those methods
* Tracking the branch information in the analyzer - it is already present in the parser, we just need to propagate it

In places where we register a table or columns/fields reference for later, we do that by associating the relevant information with a pair of the fully-qualified table name and branch (represented by `TableAndBranch` record) in place of just the name. This information is then used to call the access control methods.

## Release notes

```markdown
## Section
* Added table branch name parameter to access control SPI functions. ({issue}`29179`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
@@ -24,6 +24,7 @@ import static java.util.Objects.requireNonNull;
 public class Field
 {
     private final Optional<QualifiedObjectName> originTable;
+    private final Optional<String> originBranch;
     private final Optional<String> originColumnName;
     private final Optional<QualifiedName> relationAlias;
     private final Optional<String> name;
@@ -58,27 +59,30 @@ public class Field
                 .build();
     }
 
-    public static Field newUnqualified(Optional<String> name, Type type, Optional<QualifiedObjectName> originTable, Optional<String> originColumn, boolean aliased)
+    public static Field newUnqualified(Optional<String> name, Type type, Optional<QualifiedObjectName> originTable, Optional<String> originBranch, Optional<String> originColumn, boolean aliased)
     {
         requireNonNull(name, "name is null");
         requireNonNull(type, "type is null");
         requireNonNull(originTable, "originTable is null");
+        requireNonNull(originBranch, "originBranch is null");
 
         return builder()
                 .name(name)
                 .type(type)
                 .originTable(originTable)
+                .originBranch(originBranch)
                 .originColumnName(originColumn)
                 .aliased(aliased)
                 .build();
     }
 
-    public static Field newQualified(QualifiedName relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originColumn, boolean aliased)
+    public static Field newQualified(QualifiedName relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originBranch, Optional<String> originColumn, boolean aliased)
     {
         requireNonNull(relationAlias, "relationAlias is null");
         requireNonNull(name, "name is null");
         requireNonNull(type, "type is null");
         requireNonNull(originTable, "originTable is null");
+        requireNonNull(originBranch, "originBranch is null");
 
         return builder()
                 .relationAlias(Optional.of(relationAlias))
@@ -86,17 +90,19 @@ public class Field
                 .type(type)
                 .hidden(hidden)
                 .originTable(originTable)
+                .originBranch(originBranch)
                 .originColumnName(originColumn)
                 .aliased(aliased)
                 .build();
     }
 
-    private Field(Optional<QualifiedName> relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originColumnName, boolean aliased)
+    private Field(Optional<QualifiedName> relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originBranch, Optional<String> originColumnName, boolean aliased)
     {
         requireNonNull(relationAlias, "relationAlias is null");
         requireNonNull(name, "name is null");
         requireNonNull(type, "type is null");
         requireNonNull(originTable, "originTable is null");
+        requireNonNull(originBranch, "originBranch is null");
         requireNonNull(originColumnName, "originColumnName is null");
 
         this.relationAlias = relationAlias;
@@ -104,6 +110,7 @@ public class Field
         this.type = type;
         this.hidden = hidden;
         this.originTable = originTable;
+        this.originBranch = originBranch;
         this.originColumnName = originColumnName;
         this.aliased = aliased;
     }
@@ -116,6 +123,11 @@ public class Field
     public Optional<QualifiedObjectName> getOriginTable()
     {
         return originTable;
+    }
+
+    public Optional<String> getOriginBranch()
+    {
+        return originBranch;
     }
 
     public Optional<String> getOriginColumnName()
@@ -207,6 +219,7 @@ public class Field
         private Type type;
         private boolean hidden;
         private Optional<QualifiedObjectName> originTable = Optional.empty();
+        private Optional<String> originBranch = Optional.empty();
         private Optional<String> originColumnName = Optional.empty();
         private boolean aliased;
 
@@ -253,6 +266,12 @@ public class Field
             return this;
         }
 
+        public Builder originBranch(Optional<String> originBranch)
+        {
+            this.originBranch = requireNonNull(originBranch, "originBranch is null");
+            return this;
+        }
+
         public Builder originColumnName(Optional<String> originColumnName)
         {
             this.originColumnName = requireNonNull(originColumnName, "originColumnName is null");
@@ -268,7 +287,7 @@ public class Field
         public Field build()
         {
             requireNonNull(type, "type is null");
-            return new Field(relationAlias, name, type, hidden, originTable, originColumnName, aliased);
+            return new Field(relationAlias, name, type, hidden, originTable, originBranch, originColumnName, aliased);
         }
     }
 }
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
@@ -2593,6 +2593,7 @@ class StatementAnalyzer
                                 inputField.getType(),
                                 false,
                                 inputField.getOriginTable(),
+                                inputField.getOriginBranch(),
                                 inputField.getOriginColumnName(),
                                 inputField.isAliased());
                         fieldBuilder.add(field);
@@ -2612,6 +2613,7 @@ class StatementAnalyzer
                                 inputField.getType(),
                                 false,
                                 inputField.getOriginTable(),
+                                inputField.getOriginBranch(),
                                 inputField.getOriginColumnName(),
                                 inputField.isAliased());
                         fieldBuilder.add(field);
@@ -2695,6 +2697,7 @@ class StatementAnalyzer
                             getViewColumnType(column, name, table),
                             false,
                             Optional.of(name),
+                            getBranchName(table),
                             Optional.of(column.name()),
                             false))
                     .collect(toImmutableList());
@@ -2809,6 +2812,7 @@ class StatementAnalyzer
                         column.getType(),
                         column.isHidden(),
                         Optional.of(tableName),
+                        getBranchName(table),
                         Optional.of(column.getName()),
                         false);
                 fields.add(field);
@@ -3099,6 +3103,7 @@ class StatementAnalyzer
                             field.getType(),
                             field.isHidden(),
                             field.getOriginTable(),
+                            field.getOriginBranch(),
                             field.getOriginColumnName(),
                             field.isAliased()));
                 }
@@ -3112,6 +3117,7 @@ class StatementAnalyzer
                             field.getType(),
                             field.isHidden(),
                             field.getOriginTable(),
+                            field.getOriginBranch(),
                             field.getOriginColumnName(),
                             field.isAliased()));
                 }
@@ -4886,7 +4892,7 @@ class StatementAnalyzer
                             name = field.getName();
                         }
 
-                        Field newField = Field.newUnqualified(name, field.getType(), field.getOriginTable(), field.getOriginColumnName(), false);
+                        Field newField = Field.newUnqualified(name, field.getType(), field.getOriginTable(), field.getOriginBranch(), field.getOriginColumnName(), false);
                         analysis.addSourceColumns(newField, analysis.getSourceColumns(field));
                         outputFields.add(newField);
                     }
@@ -4896,6 +4902,7 @@ class StatementAnalyzer
                     Optional<Identifier> field = column.getAlias();
 
                     Optional<QualifiedObjectName> originTable = Optional.empty();
+                    Optional<String> originBranch = Optional.empty();
                     Optional<String> originColumn = Optional.empty();
                     QualifiedName name = null;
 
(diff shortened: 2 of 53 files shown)


core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorAccessControl.java
```diff
@@ -387,6 +414,15 @@ default void checkCanTruncateTable(ConnectorSecurityContext context, SchemaTable
...
      */
+    default void checkCanUpdateTableColumns(ConnectorSecurityContext context, SchemaTableName tableName, Optional<String> branch, Set<String> updatedColumns)
+    {
+        checkCanUpdateTableColumns(context, tableName, updatedColumns);
+    }
+
+    /**
+     * @deprecated use {@link #checkCanUpdateTableColumns(ConnectorSecurityContext, SchemaTableName, Optional, Set)}
+     */
+    @Deprecated
```
ssheikin, line 425: `@Deprecated` methods should use new ones, not other way around.

ksobolew: How do you maintain backwards compatibility then? The engine calls the new methods. The plugin implements the old methods, overriding the delegation to the new methods. So the engine never calls the methods implemented by the plugin.

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
```diff
@@ -36,49 +37,53 @@ public static Field newUnqualified(String name, Type type)
...
         requireNonNull(type, "type is null");
         requireNonNull(originTable, "originTable is null");
+        requireNonNull(originBranch, "originBranch is null");
 
-        return new Field(Optional.of(relationAlias), name, type, hidden, originTable, originColumn, aliased);
+        return new Field(Optional.of(relationAlias), name, type, hidden, originTable, originBranch, originColumn, aliased);
     }
 
-    public Field(Optional<QualifiedName> relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originColumnName, boolean aliased)
+    public Field(Optional<QualifiedName> relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originBranch, Optional<String> originColumnName, boolean aliased)
```
kokosing: That class begs for builder...

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
```diff
@@ -631,7 +631,8 @@ protected Scope visitInsert(Insert insert, Optional<Scope> scope)
             analyzeFiltersAndMasks(insert.getTable(), targetTable, new RelationType(tableFields), accessControlScope);
             analyzeCheckConstraints(insert.getTable(), targetTable, accessControlScope, checkConstraints);
             analyzeDefaultColumnValues(insert.getTable(), targetTableHandle.get(), columnHandles.values(), queryScope);
-            analysis.registerTable(insert.getTable(), targetTableHandle, targetTable, session.getIdentity().getUser(), accessControlScope, Optional.empty());
+            // TODO: when redirecting a table, do we always redirect to the same branch, or do we allow customizing this?
```
kokosing: I see two options:
 - I would say when you are redirecting a table (with any branch) you redirect to another table without Ia branch. As branch would need to be a redirection target.
- Or more strict, you can only redirect from a table without a branch to a table without a branch, as branches would need to be specified in redirection source and target.

Let's discuss this with wider community.

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
```diff
@@ -244,6 +246,32 @@ public void testSimpleMask()
         assertThat(assertions.query("SELECT custkey FROM orders WHERE orderkey = 1")).matches("VALUES CAST(NULL AS BIGINT)");
     }
 
+    @Test
+    public void testSimpleMaskOnBranch()
+    {
+        accessControl.reset();
+        accessControl.columnMask(
```
kokosing, line 253: I don't  branch name used here. How does it work? If you specify on a table without any branch is this applied to all branches?

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
```diff
@@ -244,6 +246,32 @@ public void testSimpleMask()
...
+                new QualifiedObjectName(MOCK_CATALOG, "tiny", "nation_with_hidden_column"),
+                "nationkey",
+                USER,
+                ViewExpression.builder()
+                        .identity(USER)
+                        .expression("-nationkey")
+                        .build());
+        assertThat(assertions.query("SELECT nationkey FROM mock.tiny.nation_with_hidden_column FOR VERSION AS OF 'main' WHERE name = 'ARGENTINA'")).matches("VALUES BIGINT '-1'");
```
kokosing: use some other branch name than `main` to make sure you we don't use any defaults.

kokosing: Run these queries for different branches including without a branch, to see differences in behaviour. Same comment applies to all the tests below.

testing/trino-tests/src/test/java/io/trino/security/TestBranchingAccessControl.java
```diff
@@ -0,0 +1,222 @@
...
+import static io.trino.connector.MockConnectorEntities.TPCH_NATION_SCHEMA;
+import static io.trino.testing.TestingAccessControlManager.TestingPrivilegeType.CREATE_VIEW_WITH_SELECT_COLUMNS;
+import static io.trino.testing.TestingAccessControlManager.TestingPrivilegeType.DELETE_TABLE;
+import static io.trino.testing.TestingAccessControlManager.TestingPrivilegeType.INSERT_TABLE;
+import static io.trino.testing.TestingAccessControlManager.TestingPrivilegeType.SELECT_COLUMN;
+import static io.trino.testing.TestingAccessControlManager.TestingPrivilegeType.UPDATE_TABLE;
+import static io.trino.testing.TestingAccessControlManager.branchPrivilege;
+import static io.trino.testing.TestingAccessControlManager.privilege;
+
+final class TestBranchingAccessControl
```
kokosing, line 42: Why not to merge it into `TestAccessControl`?

testing/trino-tests/src/test/java/io/trino/security/TestBranchingAccessControl.java
```diff
@@ -0,0 +1,222 @@
...
+                "INSERT INTO mock.tiny.nation @ main VALUES (101, 'POLAND', 0, 'No comment')",
+                "Cannot insert into branch main in table mock.tiny.nation",
+                branchPrivilege("nation", "main", INSERT_TABLE));
+
+        // INSERT without branch is still allowed
+        assertAccessAllowed(
+                "INSERT INTO mock.tiny.nation VALUES (101, 'POLAND', 0, 'No comment')",
+                branchPrivilege("nation", "main", INSERT_TABLE));
+
+        // INSERT into a different branch is still allowed
```
kokosing, line 126: How do you control what is allowed and what is not? I don't see usage of testing access control manager here.

kokosing (commented):
Let's add more tests to make sure it works between different branches. I see you focused mostly on main branch, and IIRC it is a branch that is used when no branch is used.

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
ksobolew: Is it a recommendation? ;)

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
ksobolew: It's a little tricky with branches and masks, as there is a check in the analyzer that skips analysis of the mask if the column is not accessible *and* is not actually referenced by the query. There is a specific test for that, but this case is just to make sure that masks applied on the table are also applied when we're querying a branch. So the branch is only in the query, there's no way to apply a mask to a branch.

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
ksobolew: I don't think there's a counterpart of git's `master` or `main`, so this name is nothing special. But I can change it to make the intent clearer.

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
ksobolew: > there's no way to apply a mask to a branch

Wait, should there be? :)

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
ksobolew: Not sure what you mean. There's no way to apply mask to a branch, so the only combinations are "mask + query with no branch" (already covered by existing tests) and "mask + query of a branch" (covered by new tests).

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
kokosing: Yes. Separate commit or even PR?

testing/trino-tests/src/test/java/io/trino/security/TestBranchingAccessControl.java
ksobolew: `TestAccessControl` seems to have a different focus - I don't even know what it's supposed to be testing :) There are already test classes like `TestColumnMask` and `TestRowFilter` and I think a couple of others that are focused on some particular aspect. Also, the conventions there are a bit different, like it explicitly manipulates the `TestingAcccessControlManager` instead of using the testing infrastructure like in this new test class.

ksobolew: But maybe it makes sense to merge them

testing/trino-tests/src/test/java/io/trino/security/TestBranchingAccessControl.java
ksobolew: This is using the testing infrastructure: the `assertAccessAllowed` and `assertAccessDenied` methods take a `TestingPrivilege` to be denied and under the hood manipulate the `TestingAcccessControlManager` - deny the privilege, run the query, remove the deny, all under `executeExclusively`.

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
ksobolew: Here you go: #29183

core/trino-main/src/main/java/io/trino/sql/analyzer/StatementAnalyzer.java
ksobolew: In some our discussion it was revealed that redirection as a concept is obsolete and will be removed in the future. Thus we can ignore branches here, that is throw an exception if redirection is attempted on a table with branch.

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
ksobolew: Switched

ksobolew:
I'm done with rebasing for now, PTAL

core/trino-main/src/main/java/io/trino/testing/TestingAccessControlManager.java
```diff
@@ -494,24 +494,24 @@ public void checkCanAlterColumn(SecurityContext context, QualifiedObjectName tab
     }
 
     @Override
-    public void checkCanInsertIntoTable(SecurityContext context, QualifiedObjectName tableName)
+    public void checkCanInsertIntoTable(SecurityContext context, QualifiedObjectName tableName, Optional<String> branch)
     {
         if (shouldDenyPrivilege(context.getIdentity().getUser(), tableName.objectName(), INSERT_TABLE)) {
```
kokosing, line 499: why you are not passing branch to `shouldDenyPrivilege`?

core/trino-main/src/main/java/io/trino/testing/TestingAccessControlManager.java
```diff
@@ -582,7 +582,7 @@ public void checkCanSetSystemSessionProperty(Identity identity, QueryId queryId,
     }
 
     @Override
-    public void checkCanCreateViewWithSelectFromColumns(SecurityContext context, QualifiedObjectName tableName, Set<String> columnNames)
+    public void checkCanCreateViewWithSelectFromColumns(SecurityContext context, QualifiedObjectName tableName, Optional<String> branch, Set<String> columnNames)
     {
         if (!denyIdentityTable.test(context.getIdentity(), tableName.objectName())) {
```
kokosing, line 587: you should also pass branch here.

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
```diff
@@ -244,6 +246,32 @@ public void testSimpleMask()
         assertThat(assertions.query("SELECT custkey FROM orders WHERE orderkey = 1")).matches("VALUES CAST(NULL AS BIGINT)");
     }
 
+    @Test
+    public void testSimpleMaskOnBranch()
```
kokosing, line 250: please do have different maks on different branches including no masking (3 test cases). Do the same for row filtering.

kokosing (approved):
I see that two my first comments were actually addressed in subsequent commit.

Please make testing for masking and filtering more robust and we are good to go.

core/trino-main/src/main/java/io/trino/testing/TestingAccessControlManager.java
ksobolew: Added in the next commit, as you noticed

core/trino-main/src/main/java/io/trino/testing/TestingAccessControlManager.java
ksobolew: Added in the next commit, as you noticed

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
ksobolew: Still not sure what you mean, sorry :) There's no way to apply mask to a branch, not without further changes to the SPI. So currently the mask is applied to the table and it appears regardless of branch. Which is probably as it should be in the first place...

core/trino-main/src/test/java/io/trino/sql/query/TestColumnMask.java
```diff
@@ -933,6 +961,49 @@ public void testNotReferencedAndDeniedColumnMasking()
         assertThat(assertions.query("SELECT orderkey FROM orders WHERE orderkey = 1")).matches("VALUES BIGINT '1'");
     }
 
+    @Test
+    public void testNotReferencedAndDeniedColumnMaskingOnBranch()
```
ssheikin, line 965: testNotReferencedAndDeniedColumnMaskingOnBranch ->
testNotReferencedAnd **NOT** DeniedColumnMaskingOnBranch ?

ksobolew: Well it is denied, just the DENY does not cause the query to fail :)

kokosing:
Thanks!

ebyhr:
Please update "Release notes" section in the PR description.

ksobolew:
> Please update "Release notes" section in the PR description.

Yes, sorry about that. Fixed.

PR #29182, opened 2026-04-21
Route splits by affinity key in the scheduler

raunaqmorarka:
## Description

Move split-to-worker placement for cache-aware scheduling out of connectors and into the engine, behind a new `ConnectorSplit#getAffinityKey()` SPI. Connectors return an opaque key; the engine maps it to preferred workers via a single consistent-hash ring.

### Why

Today, Hive, Delta, and Iceberg each inject a per-catalog `CachingHostAddressProvider` and compute split→worker placement themselves. This puts scheduling logic in connectors, duplicates a hash-ring + refresher thread per catalog over the same worker set, and leaves Hudi out because wiring it requires copying the pattern. The name `CacheKey` also couples the API to filesystem caching, but the underlying mechanism — "splits with the same key run on the same worker across queries" — is a general node-affinity primitive.

### Approach

- SPI: `ConnectorSplit#getAffinityKey(): Optional<String>`. Contract: same key ⇒ same worker(s); only remotely accessible splits may supply a key. The engine's `Split` wrapper enforces the latter.
- Engine: one `ConsistentHashingAddressProvider` bound in `CoordinatorModule`, lazily refreshed on access (no background executor). Scheduler call sites consult the key inline; when empty, `split.getAddresses()` is used unchanged.
- Filesystem library: new `SplitAffinityProvider` interface with a no-op default and a cache-enabled impl bound on the coordinator when \`fs.cache.enabled=true\`.
- Connectors: Hive, Delta, Iceberg, Hudi inject `SplitAffinityProvider` and call `getKey(path, offset, length)` during split creation. Hive keeps its `forceLocalScheduling` + HDFS block-location path.
- Cleanup (folded into the Iceberg commit, the last migrant): delete `CachingHostAddressProvider` + its impls + its test, drop the `allgood-consistent-hash` dep from `lib/trino-filesystem`, remove the old Guice bindings.

Suggested commit-order review:
1. Route splits by affinity key in the scheduler — SPI + engine.
2. Use affinity key in Hive split placement — introduces `SplitAffinityProvider` and its bindings.
3. Use affinity key in Delta Lake split placement.
4. Use affinity key in Iceberg split placement — also deletes the now-unused shared classes.
5. Add affinity key to Hudi split — net-new for Hudi.
6. Convert IcebergSplit to a record — incidental cleanup now that the class is smaller.

## Release notes

```markdown
## Section
```


raunaqmorarka:
@coderabbitai review

core/trino-main/src/main/java/io/trino/execution/scheduler/ConsistentHashingAddressProvider.java
```diff
@@ -11,82 +11,89 @@
...
+    public List<HostAddress> getHosts(String cacheKey)
     {
-        return consistentHashRing.locate(splitKey, replicationFactor)
+        refreshHashRingIfNeeded();
+        return consistentHashRing.locate(cacheKey, preferredHostsCount)
                 .stream()
                 .map(TrinoNode::getHostAndPort)
                 .sorted(hostAddressComparator)
                 .collect(toImmutableList());
     }
```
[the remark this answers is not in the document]
raunaqmorarka: The `HashRing` class is thread-safe. Its source has `private final ReadWriteLock mutex = new ReentrantReadWriteLock(true);` — `add`/`addAll`/`remove` acquire the write lock and `locate` acquires the read lock, so concurrent calls are correctly serialized by the library. The README doesn't highlight this, but the implementation handles it.

chenjian2664:
You're effectively hard-wiring all connectors to a single consistent-hash strategy whenever `getAffinityKey` is implemented. That removes any flexibility to adopt alternative strategies in the future. For example, if we introduce a new `CachingHostAddressProvider` (in today's filesystem cache module) with a different routing policy, there would be no way to opt into it under the current design. Is that an intended constraint?

raunaqmorarka:
> You're effectively hard-wiring all connectors to a single consistent-hash strategy whenever `getAffinityKey` is implemented. That removes any flexibility to adopt alternative strategies in the future. For example, if we introduce a new `CachingHostAddressProvider` (in today's filesystem cache module) with a different routing policy, there would be no way to opt into it under the current design. Is that an intended constraint?

The old `getAddresses` way is still available for connectors which want to use a custom strategy.
There was no need for flexibility here, the connector is just interested in providing a splits placement policy and not in dealing with actual host addresses or running the consistent hashing code itself. Doing it this way prevents duplication of same consistent hashing ring per catalog.

chenjian2664:
> > You're effectively hard-wiring all connectors to a single consistent-hash strategy whenever `getAffinityKey` is implemented. That removes any flexibility to adopt alternative strategies in the future. For example, if we introduce a new `CachingHostAddressProvider` (in today's filesystem cache module) with a different routing policy, there would be no way to opt into it under the current design. Is that an intended constraint?
> 
> The old `getAddresses` way is still available for connectors which want to use a custom strategy. There was no need for flexibility here, the connector is just interested in providing a splits placement policy and not in dealing with actual host addresses or running the consistent hashing code itself. Doing it this way prevents duplication of same consistent hashing ring per catalog.

Once you are rollback to use `getAddresses`, the `getAffinityKey` will be redundant and shouldn't be serialized, the situation will the same as now.
Let's say if the today the `IcebergSplit` use a strategy in fs that using partition range to schedule, how would the current design to support that?

raunaqmorarka:
> > > You're effectively hard-wiring all connectors to a single consistent-hash strategy whenever `getAffinityKey` is implemented. That removes any flexibility to adopt alternative strategies in the future. For example, if we introduce a new `CachingHostAddressProvider` (in today's filesystem cache module) with a different routing policy, there would be no way to opt into it under the current design. Is that an intended constraint?
> > 
> > 
> > The old `getAddresses` way is still available for connectors which want to use a custom strategy. There was no need for flexibility here, the connector is just interested in providing a splits placement policy and not in dealing with actual host addresses or running the consistent hashing code itself. Doing it this way prevents duplication of same consistent hashing ring per catalog.
> 
> Once you are rollback to use `getAddresses`, the `getAffinityKey` will be redundant and shouldn't be serialized, the situation will the same as now. Let's say if the today the `IcebergSplit` use a strategy in fs that using partition range to schedule, how would the current design to support that?

If `getAddresses` is populated, then `getAffinityKey` is enforced to be empty, so there is no extra serialization cost.
Iceberg is still free to populate `getAffinityKey` in different ways, e.g. instead of split path and offsets, it can be partition value. When splits are remotely accessible (as is the case in data lake connectors), there is no real reason for the connector to concern itself with specific host addresses.

core/trino-main/src/main/java/io/trino/execution/scheduler/UniformNodeSelector.java
```diff
@@ -173,8 +179,11 @@ public SplitPlacementResult computeAssignments(Set<Split> splits, List<RemoteTas
             }
             else {
                 // optimizedLocalScheduling enables prioritized assignment of splits to local nodes when splits contain locality information
-                if (optimizedLocalScheduling && !split.getAddresses().isEmpty()) {
-                    candidateNodes = selectExactNodes(nodeMap, split.getAddresses(), includeCoordinator);
+                List<HostAddress> preferredAddresses = split.getConnectorSplit().getAffinityKey()
+                        .map(consistentHashingAddressProvider::getHosts)
+                        .orElseGet(split::getAddresses);
```
chenjian2664, line 184: Can we add a code comment here to make it explicit that we are clear that splits with an affinity key use a consistent hashing algorithm to determine the target workers for scheduling?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplit.java
```diff
@@ -40,6 +41,17 @@ default List<HostAddress> getAddresses()
         return List.of();
     }
 
+    /**
+     * Returns an optional affinity key so splits reading related content are routed to the
+     * same worker(s) across queries. When empty, scheduling falls back to {@link #getAddresses()}.
+     * <p>
+     * Only remotely accessible splits may supply an affinity key (see {@link #isRemotelyAccessible()}).
+     */
+    default Optional<String> getAffinityKey()
```
chenjian2664, line 50: Why put it in the `ConnectorSplit` ? what about put it in the `Split`, `ConnectorSplit` is just a param for getting the key.

raunaqmorarka: Connector has to provide a key per-split somehow
`Split` is in trino-main

core/trino-main/src/main/java/io/trino/execution/scheduler/UniformNodeSelector.java
raunaqmorarka: That would be just narrating what the code does, its written in a straightforward way already

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplit.java
chenjian2664: ok, I think that's the https://github.com/trinodb/trino/pull/29138 does, let's talk about it later.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
```diff
@@ -68,41 +69,9 @@ public IcebergSplit(
             @JsonProperty("deletes") List<DeleteFile> deletes,
             @JsonProperty("splitWeight") SplitWeight splitWeight,
             @JsonProperty("fileStatisticsDomain") TupleDomain<IcebergColumnHandle> fileStatisticsDomain,
+            @JsonProperty("affinityKey") Optional<String> affinityKey,
```
chenjian2664: The key isn't necessary to serialize

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
```diff
@@ -115,16 +84,16 @@ public IcebergSplit(
...
         this.splitWeight = requireNonNull(splitWeight, "splitWeight is null");
         this.fileStatisticsDomain = requireNonNull(fileStatisticsDomain, "fileStatisticsDomain is null");
-        this.addresses = requireNonNull(addresses, "addresses is null");
+        this.affinityKey = requireNonNull(affinityKey, "affinityKey is null");
         this.dataSequenceNumber = dataSequenceNumber;
         this.fileFirstRowId = requireNonNull(fileFirstRowId, "fileFirstRowId is null");
     }
 
-    @JsonIgnore
+    @JsonProperty
```
chenjian2664: The purpose should be fully serializable but I don't see the Iceberg is different from Delta Lake and Hive, they marked the property JsonIgnore
I recommend making the behavior consistent across all (lake house) connectors. For this case, we can add a TODO to note that serialization could be avoided - i,e that we don't want put more "JsonIgnore" in the ConnectorSplit

chenjian2664 (approved):
Please rebase fix the ci failures

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
raunaqmorarka: Going the other direction — DeltaLakeSplit is now a record and serializes affinityKey in a new commit, matching IcebergSplit. Not changing HiveSplit or HudiSplit.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplit.java
raunaqmorarka: Consistency via serialization rather than @JsonIgnore: DeltaLakeSplit is now a record serializing affinityKey in a new commit, matching IcebergSplit. Not changing HiveSplit or HudiSplit.

raunaqmorarka:
@coderabbitai full review

core/trino-main/src/main/java/io/trino/execution/scheduler/faulttolerant/ArbitraryDistributionSplitAssigner.java
```diff
@@ -81,7 +83,8 @@ class ArbitraryDistributionSplitAssigner
             long minTargetPartitionSizeInBytes,
             long maxTargetPartitionSizeInBytes,
             long standardSplitSizeInBytes,
-            int maxTaskSplitCount)
+            int maxTaskSplitCount,
+            ConsistentHashingAddressProvider consistentHashingAddressProvider)
```
wendigo, line 87: nit: we should use interface here `CachingHostAddressProvider`

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/SplitAffinityProvider.java
```diff
@@ -0,0 +1,27 @@
...
+import java.util.Optional;
+
+/**
+ * Produces a stable affinity key for a split, used by the engine to consistently route
+ * splits reading the same content to the same worker(s). A non-empty key opts the split
+ * into cache-aware scheduling via the engine's consistent-hash ring; an empty key leaves
+ * scheduling to the split's own addresses.
+ */
+public interface SplitAffinityProvider
+{
```
wendigo, line 25: Re: String path. Should we use Location here?

wendigo (approved):

core/trino-main/src/main/java/io/trino/execution/scheduler/faulttolerant/ArbitraryDistributionSplitAssigner.java
raunaqmorarka: Leaving as-is. An interface with a single implementation wouldn't buy us flexibility here — the placement strategy is fixed by how `getAffinityKey()` feeds into the hash ring, not by the type the assigner declares. Easy enough to extract an interface later if a second implementation shows up.

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/SplitAffinityProvider.java
raunaqmorarka: Leaving as-is. The argument isn't being treated as a filesystem location — it's opaque key material that gets folded into the affinity hash. Three of the four callers already have a `String` on hand, so switching to `Location` would just mean wrapping it at every callsite and unwrapping it with `toString()` inside the provider.

eskabetxe (approved):

PR #29183, opened 2026-04-21
Convert analyzer's Field class to use the builder pattern

ksobolew:
## Description

It's the most beneficial in places where we construct a copy of the `Field` and only change one of a couple of properties - instead of a soup of positional parameter we have the changed properties clearly marked by name.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
@@ -31,12 +31,20 @@ public class Field
     private final boolean hidden;
     private final boolean aliased;
 
+    public static Builder builder()
+    {
+        return new Builder();
+    }
+
     public static Field newUnqualified(String name, Type type)
     {
         requireNonNull(name, "name is null");
         requireNonNull(type, "type is null");
 
-        return new Field(Optional.empty(), Optional.of(name), type, false, Optional.empty(), Optional.empty(), false);
+        return builder()
+                .name(Optional.of(name))
+                .type(type)
+                .build();
     }
 
     public static Field newUnqualified(Optional<String> name, Type type)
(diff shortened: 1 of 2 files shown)


wendigo (approved):

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
```diff
@@ -53,7 +64,13 @@ public static Field newUnqualified(Optional<String> name, Type type, Optional<Qu
...
+        return builder()
+                .name(name)
+                .type(type)
+                .originTable(originTable)
+                .originColumnName(originColumn)
+                .aliased(aliased)
+                .build();
     }
 
     public static Field newQualified(QualifiedName relationAlias, Optional<String> name, Type type, boolean hidden, Optional<QualifiedObjectName> originTable, Optional<String> originColumn, boolean aliased)
```
kokosing, line 76: should we inline such methods?

kokosing (approved):

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
kokosing: Separate commit :)

ksobolew: Hm, I intended to leave these factory methods alone, but I'll have to make a survey how they are actually used

kokosing: I am fine with leaving methods that accept up 3 parameters. But here if people are passing 3 optionals it is unclear what is what, so I would see a value in such refactor. Maybe a separate PR to make sure we collect opinion from all interested people.

kokosing:
Merged, thanks!

core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
ksobolew: Yes, it totally makes sense to inline that one big factory, I'll prepare a PR

ksobolew: I actually inlined all of them: #29185. Like I suspected, in some cases it can be simplified, or made to make an actual copy where we create a modified instance of a `Field`.

PR #29184, opened 2026-04-21
Implement unified blob cache plugin SPI

wendigo:
The purpose of the unified cache is to move cache control to the engine, including cache invalidation, sizing, etc. Right now every catalog creates and manages its own cache which is suboptimial and hard to manage when dynamic catalogs are used. Instead of having multiple caches, create a single, engine-level cache that is leased to connectors.

## Release notes

```markdown
## Section
```


wendigo:
@coderabbitai review

wendigo:
@coderabbitai full review

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
```diff
@@ -0,0 +1,42 @@
...
+ */
+public interface BlobSource
+{
+    long length()
+            throws IOException;
+
+    void readFully(long position, byte[] buffer, int offset, int length)
+            throws IOException;
+
+    default int readTail(byte[] buffer, int offset, int length)
```
martint: What's the benefit of building this into BlobSource vs having a utility that can do this?

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * A filesystem-agnostic blob cache. Callers identify cached bytes with a
+ * {@link CacheKey} and supply a {@link BlobSource} used to populate the entry
+ * on miss. The returned {@link BlobSource} reads from the cached entry on
+ * subsequent calls.
+ */
+public interface BlobCache
+{
+    BlobSource get(CacheKey key, BlobSource source)
```
martint: BlobSource for input should probably be separate from the output of this method. We may need to add more capability to the source (e.g., read ranges, prefetch, async reads) without polluting what the cache users see as the interface.

martint: Also, the Blob that's returned from the cache may need to keep state, be closeable so that callers can manage lifecycle, etc. We may also want different declared exceptions for the input BlobSource vs the output one.

core/trino-spi/src/main/java/io/trino/spi/cache/CacheTier.java
```diff
@@ -11,9 +11,10 @@
...
+package io.trino.spi.cache;
 
-/**
- * Dummy class needed to preserve the legacy JMX object name.
- */
-public final class TrinoFileSystemCache {}
+public enum CacheTier
+{
+    MEMORY,
+    DISK,
```
martint: This taxonomy is also very limited. What about "distributed memory", "distributed disk", "flash/ssd", etc. In the end, it depends on how they are supposed to be used. Which part of the system cares about which tier is being used and why does the tier matter?

wendigo: Connector cares as tier == latency

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
wendigo: There is no use-case for that one so adding a separate interface with methods that are not used is not reasonable for now. We can revisit this API in the future

wendigo:
@coderabbitai review

wendigo:
@coderabbitai full review

core/trino-spi/src/main/java/io/trino/spi/cache/CacheTier.java
electrum: I agree with @martint that these names are confusing, and it's not clear how they should be used by a connector. Would `METADATA` and `DATA` directly map to how a connector uses it?

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
```diff
@@ -0,0 +1,33 @@
...
+ * {@link CacheKey} and supply a {@link BlobSource} used to populate the entry
+ * on miss. The returned {@link BlobSource} reads from the cached entry on
+ * subsequent calls.
+ */
+public interface BlobCache
+{
+    Blob get(CacheKey key, BlobSource source)
+            throws IOException;
+
+    void invalidate(CacheKey key);
```
dain: maybe add a default implementation

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.IOException;
+
+/**
+ * Lazy source of bytes backing a cached blob. Used by {@link BlobCache} to
+ * populate a cache entry on miss.
+ */
+public interface BlobSource
+{
+    long length()
+            throws IOException;
```
dain, line 36: Document what happens if the length doesn't match what the blob source provides.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
```diff
@@ -0,0 +1,38 @@
...
+
+    void readFully(long position, byte[] buffer, int offset, int length)
+            throws IOException;
```
dain, line 44: For a streaming source, this api forces the data to be full buffered.  I don't know if that is a problem.

core/trino-spi/src/main/java/io/trino/spi/cache/CacheTier.java
wendigo: I'd say that engine doesn't care how connector uses the cache. If connector wants to cache metadata on disk, why it couldn't do it?

wendigo: The enum is more about a storage characteristic the connector picks based on its access pattern

core/trino-spi/src/main/java/io/trino/spi/cache/ConnectorCacheFactory.java
```diff
@@ -11,9 +11,9 @@
...
-package io.trino.plugin.hive.fs;
+package io.trino.spi.cache;
 
-/**
- * Dummy class needed to preserve the legacy JMX object name.
- */
-public final class TrinoFileSystemCache {}
+public interface ConnectorCacheFactory
+{
+    BlobCache createBlobCache(CacheTier tier);
```
dain: I'd consider having only one blob cache and then when caching you provide a hint somehow on what the data is and it's priority.  I'm thinking that you can have a ok implementation for complex mixed usages without have full semantic knowledge of the data that is being cached.  That way, I can have iceberg metadata with high priority, maybe glue second, and then data files lower down.

I don't think it is possible to have a perfect system without hand coding the cache to the usage, but you should be able to have a pretty good design out of the box.

dain (commented):
I think this needs more documentation in the class and likely in docs somewhere.

I have questions like:
* If I have two connectors using Glue, do they share the cache becasue they use the same cache keys.
* Does the connector tell the cache how important data is, or does the connector know the priority?
* What is guidance for the shape of cache keys?

Oh is this really only for files?  I thought I could use this for any data?  I was expecting to see a Glue integration.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
wendigo: Alluxio/Memory cache works that way anyway

wendigo:
@coderabbitai full review

core/trino-spi/src/main/java/io/trino/spi/cache/Blob.java
```diff
@@ -0,0 +1,59 @@
...
+    void readFully(long position, byte[] buffer, int offset, int length)
+            throws IOException;
+
+    /**
+     * Whether this blob holds the storage object's complete content: {@link #length()} is the
+     * object's true size and {@link #readFully} covers it fully. When true, callers may serve
+     * tail reads directly from the blob; otherwise tails come from dedicated tail entries
+     * anchored at the object's actual end.
+     */
+    default boolean hasCompleteContent()
```
losipiuk: hasAllContent?

core/trino-spi/src/main/java/io/trino/spi/cache/Blob.java
```diff
@@ -0,0 +1,59 @@
...
+    {
+        return false;
+    }
+
+    /**
+     * Number of bytes this blob served from cached content so far.
+     */
+    default long cachedBytes()
```
losipiuk: cachedSize

core/trino-spi/src/main/java/io/trino/spi/cache/Blob.java
```diff
@@ -0,0 +1,59 @@
...
+    {
+        return 0;
+    }
+
+    /**
+     * Number of bytes this blob fetched from its backing {@link BlobSource} so far.
+     */
+    default long loadedBytes()
```
losipiuk: loadedSize

core/trino-spi/src/main/java/io/trino/spi/cache/Blob.java
```diff
@@ -0,0 +1,59 @@
...
+ * the engine caches tails as separate entries keyed by a derived {@link CacheKey} component,
+ * so implementations only ever serve exact byte ranges of the content they were populated with.
+ */
+public interface Blob
+        extends Closeable
+{
+    long length()
+            throws IOException;
+
+    void readFully(long position, byte[] buffer, int offset, int length)
```
losipiuk: Just read?

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * A filesystem-agnostic blob cache. Callers identify cached bytes with a
+ * {@link CacheKey} and supply a {@link BlobSource} used to populate the entry
+ * on miss. The returned {@link BlobSource} reads from the cached entry on
+ * subsequent calls.
+ */
+public interface BlobCache
+{
+    Blob get(CacheKey key, BlobSource source)
```
losipiuk, line 34: nit: If you are adding javadoc (good for SPI) add here too.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.Set;
+
+public interface BlobCacheManager
+{
+    /**
+     * Whether the caches this manager creates provide every given capability. The engine only
+     * routes a usage to this manager when it supports the capabilities the usage requires; the
+     * answer may reflect the manager's actual configuration.
+     */
+    boolean supports(Set<CacheCapability> capabilities);
```
losipiuk: nit: Maybe just go with hasCapability(CacheCapability). I doubt multicheck is super crucial here. Let's keep interface simple

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * Creates the cache used by the given catalog. Every {@link CacheKey} passed to the returned
+     * {@link BlobCache} has the catalog name as its first component, so all of a catalog's entries
+     * can be invalidated with the {@code [catalog]} key prefix.
+     */
+    BlobCache createBlobCache(CatalogName catalog);
+
+    void drop(CatalogName catalog);
+
```
losipiuk: make naming of createBlobCache and drop consistent

core/trino-spi/src/main/java/io/trino/spi/cache/CacheKey.java
```diff
@@ -0,0 +1,77 @@
...
+     * Returns a new key with the given component prepended at the head.
+     */
+    public CacheKey withPrefix(String component)
+    {
+        List<String> prefixed = new ArrayList<>(components.size() + 1);
+        prefixed.add(component);
+        prefixed.addAll(components);
+        return new CacheKey(prefixed);
+    }
+
```
losipiuk: Let's stay with append only mode. Instead of this method create

CacheKey append(CacheKey childKey)

core/trino-spi/src/main/java/io/trino/spi/cache/CacheKey.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Returns a new key with the given component appended at the tail, identifying a variant
+     * of this entry that is invalidated together with it by any prefix covering this key.
+     */
+    public CacheKey child(String component)
```
losipiuk: append?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheUsage.java
```diff
@@ -0,0 +1,41 @@
...
+
+import java.util.Set;
+
+/**
+ * Declares who is caching what, and what the cache must be able to do. Implemented as an enum
+ * by each cache consumer: the declaration site is the single place a usage is defined, and
+ * call sites reference the typed constant. Identity is the usage name, so usages survive
+ * classloader boundaries; enum object identity is deliberately not used.
+ */
+public interface CacheUsage
```
losipiuk: I read that javadoc and have no clue what the author had in mind

core/trino-spi/src/main/java/io/trino/spi/cache/CacheCapability.java
```diff
@@ -0,0 +1,34 @@
...
+ * {@link ConnectorCacheFactory#createBlobCache}, and only managers declaring every required
+ * capability are considered. A value is added here only together with the consumer whose
+ * behavior depends on it.
+ */
+public enum CacheCapability
+{
+    /**
+     * Cache hits are served from process memory, with no disk or network I/O.
+     */
+    MEMORY_SPEED_READS,
```
losipiuk: LOW_LATENCY?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheCapability.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Can hold working sets larger than available process memory.
+     */
+    EXCEEDS_MEMORY_CAPACITY,
```
losipiuk: CAN_EXCEED_HEAP_SIZE

lib/trino-filesystem-manager/src/main/java/io/trino/filesystem/manager/FileSystemCacheUsage.java
```diff
@@ -0,0 +1,58 @@
...
+import io.trino.spi.cache.CacheUsage;
+
+import java.util.Set;
+
+import static io.trino.spi.cache.CacheCapability.EXCEEDS_MEMORY_CAPACITY;
+import static io.trino.spi.cache.CacheCapability.MEMORY_SPEED_READS;
+import static java.util.Locale.ENGLISH;
+
+public enum FileSystemCacheUsage
+        implements CacheUsage
```
losipiuk: Why is that interface not just a record?

losipiuk: Actually I am not convinced we need that at all. Why not just pass two args to createBlobCache

core/trino-spi/src/main/java/io/trino/spi/cache/ConnectorCacheFactory.java
```diff
@@ -0,0 +1,33 @@
...
+public interface ConnectorCacheFactory
+{
+    /**
+     * Whether exactly one loaded cache manager satisfies every capability the given usage
+     * requires, guaranteeing that {@link #createBlobCache} succeeds. Consumers for whom
+     * caching is optional check this and run uncached when unsupported; consumers for whom
+     * the usage is required (an explicit operator opt-in) call {@link #createBlobCache}
+     * directly and let it fail.
+     */
+    boolean supports(CacheUsage usage);
```
losipiuk: Can createBlobCache return Optional<BlobCache> instead and we can have single step on caller side?

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
```diff
@@ -0,0 +1,39 @@
...
+     * answer may reflect the manager's actual configuration.
+     */
+    boolean supports(Set<CacheCapability> capabilities);
+
+    /**
+     * Creates the cache used by the given catalog. Every {@link CacheKey} passed to the returned
+     * {@link BlobCache} has the catalog name as its first component, so all of a catalog's entries
+     * can be invalidated with the {@code [catalog]} key prefix.
+     */
+    BlobCache createBlobCache(CatalogName catalog);
```
losipiuk: don't you need to pass requried capabilities here? It seems that single cache implementaino can provide multiple QOS levels depending on what is requested by caller. E.g. spill to disk or not?

losipiuk:
I went through SPI for now - let start with that.

lib/trino-filesystem-cache-alluxio/src/main/java/io/trino/filesystem/alluxio/AlluxioFileSystemCacheModule.java
```diff
@@ -1,62 +0,0 @@
...
-import io.trino.spi.catalog.CatalogName;
-
-import java.util.Properties;
-
-import static com.google.inject.Scopes.SINGLETON;
-import static com.google.inject.multibindings.OptionalBinder.newOptionalBinder;
-import static io.airlift.configuration.ConfigBinder.configBinder;
-import static org.weakref.jmx.guice.ExportBinder.newExporter;
-
-public class AlluxioFileSystemCacheModule
```
raunaqmorarka, line 33: Migration of alluxio (or any other existing cache) should be a separate commit from the one which adds new engine SPI and capabilities; and each migration should be an independant commit

core/trino-spi/src/main/java/io/trino/spi/cache/CacheUsage.java
raunaqmorarka: "usage" doesn't seem the right name here, given that its defining part of cache key and capabilities.
Maybe "CacheRequirements" or something else ?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheUsage.java
```diff
@@ -0,0 +1,41 @@
...
+{
+    /**
+     * Name of the usage, unique among the cache consumers of a catalog, for example
+     * {@code "filesystem.data"}. Becomes a cache key component, so entries of one usage can
+     * be invalidated together. Not called {@code name()} so that enum implementors are not
+     * blocked by the final {@link Enum#name()}.
+     */
+    String usageName();
```
raunaqmorarka: keyPrefix ?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheCapability.java
```diff
@@ -0,0 +1,34 @@
...
+package io.trino.spi.cache;
+
+/**
+ * A behavioral fact about the caches a {@link BlobCacheManagerFactory} creates, objectively
+ * true or false for a given implementation. Consumers pass the capabilities they require to
+ * {@link ConnectorCacheFactory#createBlobCache}, and only managers declaring every required
+ * capability are considered. A value is added here only together with the consumer whose
+ * behavior depends on it.
+ */
+public enum CacheCapability
```
raunaqmorarka, line 23: This is defining the context in which the cache will be used, its saying nothing about concrete capabilities except whether it is an in-memory cache or not, so why not simplify it to a boolean ?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheCapability.java
raunaqmorarka: IN_MEMORY ?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheUsage.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Capabilities the cache serving this usage must have. Only managers declaring every
+     * required capability are considered when fulfilling {@link ConnectorCacheFactory#createBlobCache};
+     * when none qualifies, creation fails — consumers for whom caching is optional check
+     * {@link ConnectorCacheFactory#supports} first.
+     */
+    Set<CacheCapability> requiredCapabilities();
```
raunaqmorarka: Returning a Set looks odd, what if `CacheCapability` are contradictory like MEMORY_SPEED_READS and EXCEEDS_MEMORY_CAPACITY ?

core/trino-spi/src/main/java/io/trino/spi/cache/CacheManagerContext.java
```diff
@@ -0,0 +1,30 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.trino.spi.cache;
+
+import io.opentelemetry.api.OpenTelemetry;
+import io.opentelemetry.api.trace.Tracer;
+
+public interface CacheManagerContext
```
raunaqmorarka, line 19: Do we have something which will allow us to track memory usage of the caches into worker memory pools ?
E.g. io.trino.filesystem.alluxio.AlluxioFileSystemCache#cacheManager or io.trino.parquet.cache.MemoryParquetFooterCache#cache

lib/trino-filesystem-cache-alluxio/src/main/java/io/trino/filesystem/alluxio/AlluxioFileSystemCacheModule.java
wendigo: I've tried that but eventually squashed the work

core/trino-spi/src/main/java/io/trino/spi/cache/CacheCapability.java
wendigo: No. This is a part I've discussed at length with Martin. There will be other knobs in the future like priorities or quotas.

core/trino-spi/src/main/java/io/trino/spi/cache/CacheManagerContext.java
wendigo: Not yet

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.io.IOException;
+import java.util.Collection;
+
+/**
+ * A filesystem-agnostic blob cache. Callers identify cached bytes with a
+ * {@link CacheKey} and supply a {@link BlobSource} used to populate the entry
+ * on miss. The returned {@link BlobSource} reads from the cached entry on
+ * subsequent calls.
+ */
```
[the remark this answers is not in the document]
raunaqmorarka: +1

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
```diff
@@ -0,0 +1,50 @@
...
+ * Lazy source of bytes backing a cached blob. Used by {@link BlobCache} to
+ * populate a cache entry on miss.
+ */
+public interface BlobSource
+        extends Closeable
+{
+    long length()
+            throws IOException;
+
+    void readFully(long position, byte[] buffer, int offset, int length)
```
raunaqmorarka, line 43: javadoc for these 2 as well ?

plugin/trino-blob-cache-alluxio/src/main/java/io/trino/blob/cache/alluxio/AlluxioCache.java
```diff
@@ -0,0 +1,105 @@
...
+import io.trino.spi.cache.Blob;
+import io.trino.spi.cache.BlobSource;
+import io.trino.spi.cache.CacheKey;
+import jakarta.annotation.PreDestroy;
+
+import java.io.IOException;
+
+import static java.util.Objects.requireNonNull;
+
+public class AlluxioCache
```
raunaqmorarka, line 40: Safe to assume that only one of this is instantiated in the server process ?
CacheManager eats half a gig of memory per instance...

plugin/trino-blob-cache-memory/src/main/java/io/trino/blob/cache/memory/MemoryBlobCacheConfig.java
```diff
@@ -1,80 +0,0 @@
...
-    private DataSize maxSize = DEFAULT_CACHE_SIZE;
-    private DataSize maxContentLength = DataSize.of(8, MEGABYTE);
-
-    @NotNull
-    public Duration getCacheTtl()
-    {
-        return ttl;
-    }
-
-    @Config("fs.memory-cache.ttl")
```
raunaqmorarka, line 43: I'm assuming that all existing configs have just moved around in the code and no breaking changes for users

pom.xml
```diff
@@ -1023,12 +1022,6 @@
                 <version>${project.version}</version>
             </dependency>
 
-            <dependency>
-                <groupId>io.trino</groupId>
-                <artifactId>trino-filesystem-cache-alluxio</artifactId>
-                <version>${project.version}</version>
-            </dependency>
-
             <dependency>
```
raunaqmorarka, line 1029: Can we update `File system cache` docs to talk about how multi-catalog case is handled now ?

raunaqmorarka (approved):

plugin/trino-blob-cache-memory/src/main/java/io/trino/blob/cache/memory/MemoryBlobCacheConfig.java
wendigo: It's a breaking change. We will need to document it properly

plugin/trino-blob-cache-alluxio/src/main/java/io/trino/blob/cache/alluxio/AlluxioCache.java
wendigo: Yes. This is the case

plugin/trino-blob-cache-memory/src/main/java/io/trino/blob/cache/memory/MemoryBlobCacheConfig.java
raunaqmorarka: Please document it as breaking in release notes and update docs in this commit

wendigo:
@martint please review SPI part of this PR

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    Blob get(CacheKey key, BlobSource source)
+            throws IOException;
+
+    /**
+     * Invalidates every entry whose key {@link CacheKey#startsWith starts with} the given
+     * components: a full key removes a single entry, a shorter key removes all entries
+     * under that prefix (for example all versions and variants of a file, or a whole catalog).
+     */
+    void invalidate(CacheKey keyOrPrefix);
```
martint: keyOrPrefix -> prefix.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Creates the cache used by the given catalog, providing the given capabilities.
+     * Implementations offering multiple service levels use the capabilities to configure the
+     * returned cache; the engine only passes capabilities this manager
+     * {@link #hasCapability has}. Every {@link CacheKey} passed to the returned
+     * {@link BlobCache} has the catalog name as its first component, so all of a catalog's
+     * entries can be invalidated with the {@code [catalog]} key prefix.
+     */
+    BlobCache createBlobCache(CatalogName catalog, Set<CacheCapability> capabilities);
```
martint: "Blob" in createBlobCache is unnecessary. It's understood it's a blob cache since this is BlobCacheManager

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * Drops all cached state of the given catalog.
+     */
+    void dropBlobCache(CatalogName catalog);
```
martint: Same here

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
martint: Requiring a catalog name means that there cannot be caches that are shared across catalogs.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
```diff
@@ -0,0 +1,50 @@
...
+            throws IOException;
+
+    /**
+     * Read up to {@code length} bytes anchored at the actual end of the underlying object and
+     * return the number of bytes actually read, which is smaller than {@code length} when the
+     * object holds fewer bytes. Implementations must address this read using the source's true
+     * byte range (e.g. an HTTP suffix-range request), independent of any value returned by
+     * {@link #length()} — which may be a caller-supplied hint that does not match the storage object.
+     */
```
martint: I'm not sure I understand what this means. In particular, the part about `independent of any value returned by {@link #length()}`.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
```diff
@@ -0,0 +1,50 @@
...
+    int readTail(byte[] buffer, int offset, int length)
```
martint: What's the use case for this, vs requiring the caller to use readFully() and length() to read from the tail?

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
martint: This contract cannot be honored by the Alluxio implementation in this PR: AlluxioCache.invalidate is an unconditional no-op because prefix invalidation is impossible with its hashed page identifiers. CacheFileSystem nevertheless relies on invalidation after writes, deletes, and renames, and BlobCacheManager.dropBlobCache promises to drop catalog state. The abstraction should either require an explicit prefix-invalidation capability and select only managers that provide it, or stop promising invalidation. Silently accepting the operation is unsafe.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
```diff
@@ -0,0 +1,40 @@
...
+ * subsequent calls.
+ */
+public interface BlobCache
+{
+    /**
+     * Returns the blob cached under the given key, populating the entry from {@code source}
+     * on miss. Implementations that cannot or choose not to cache the entry (for example
+     * when it exceeds a size limit) return a blob reading through to the source instead.
+     */
+    Blob get(CacheKey key, BlobSource source)
```
martint, line 34: The ownership transfer for source needs to cover exceptional return. The documentation says who closes it after a successful cache or pass-through result, but not when get throws. MemoryBlobCache can open the source while loading and then propagate an exception before source.close is reached; the caller receives no Blob and cannot close it. Please define one unconditional ownership rule and require the cache to close the source on every failure path.

core/trino-spi/src/main/java/io/trino/spi/cache/CacheRequirements.java
```diff
@@ -0,0 +1,33 @@
...
+
+import java.util.Set;
+
+/**
+ * What a consumer requires from a blob cache: the usage name declaring who is caching what,
+ * for example {@code "filesystem.data"}, and the capabilities the serving cache manager must
+ * provide. The usage name becomes a cache key component, so it must be unique among the cache
+ * consumers of a catalog, and entries of one usage can be invalidated together.
+ */
+public record CacheRequirements(String usageName, Set<CacheCapability> capabilities)
```
martint, line 26: The documented uniqueness of usageName is a correctness invariant, but it is not enforced anywhere. ScopedBlobCache puts this string directly into the cache namespace, so two unrelated consumers that choose the same usage name and child key silently alias and can receive bytes populated by the other source. The registry should issue or reserve a namespace lease per catalog and usage, make identical repeated requests explicitly idempotent, and reject conflicting reuse.

core/trino-main/src/main/java/io/trino/cache/CacheManagerRegistry.java
```diff
@@ -0,0 +1,182 @@
...
+            requireNonNull(requirements, "requirements is null");
+
+            List<LoadedCacheManager> qualifying = blobCacheManagers.values().stream()
+                    .filter(loaded -> requirements.capabilities().stream().allMatch(loaded.manager()::hasCapability))
+                    .collect(toImmutableList());
+            if (qualifying.isEmpty()) {
+                log.debug("No loaded blob cache manager provides %s required by usage %s of catalog %s", requirements.capabilities(), requirements.usageName(), catalog);
+                return Optional.empty();
+            }
+            if (qualifying.size() > 1) {
```
martint, line 138: Capabilities filter managers but do not provide a selection policy. Loading two implementations with overlapping capabilities, or requesting an empty capability set, makes connector creation fail even though both caches are usable. As more plugins gain capabilities this becomes fragile. There needs to be an operator-configured binding, priority, or explicit manager identity rather than requiring operators to unload every other qualifying manager.

core/trino-spi/src/main/java/io/trino/spi/cache/CacheKey.java
```diff
@@ -0,0 +1,78 @@
...
+    public boolean startsWith(CacheKey prefix)
+    {
+        if (prefix.components.size() > components.size()) {
+            return false;
+        }
+        return components.subList(0, prefix.components.size()).equals(prefix.components);
+    }
+
+    @Override
+    public String toString()
```
martint, line 74: This is not an injective representation of the component list: components ["a|b", "c"] and ["a", "b|c"] produce the same string. It is already used for tracing and descriptions, and SPI implementors will naturally be tempted to use it as a storage identifier. Remove the override or provide an explicitly framed or escaped encoding.

core/trino-spi/src/main/java/io/trino/spi/cache/CacheManagerContext.java
```diff
@@ -0,0 +1,30 @@
...
+{
+    default OpenTelemetry getOpenTelemetry()
```
martint: Why are these default methods? CacheManagerContext is a new interface, so there is no compatibility reason to let an incomplete engine implementation compile and fail later with UnsupportedOperationException. These should be abstract unless there is a usable fallback.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorContext.java
```diff
@@ -100,4 +101,10 @@ default ConnectorExpressionEvaluator getExpressionEvaluator()
     {
         return ConnectorExpressionEvaluator.NO_OP;
     }
+
+    @Unstable
+    default ConnectorCacheFactory getCacheFactory()
```
martint, line 108: A throwing default conflicts with the optional-cache contract. A connector that treats absence of a matching manager as uncached operation still fails on an older or incomplete context before it can observe Optional.empty. This can safely default to a ConnectorCacheFactory that always returns Optional.empty.

core/trino-spi/src/main/java/io/trino/spi/cache/Blob.java
```diff
@@ -0,0 +1,64 @@
...
+     */
+    default boolean hasAllContent()
+    {
+        return false;
+    }
+
+    /**
+     * Number of bytes this blob served from cached content so far.
+     */
+    default long cachedSize()
```
martint: Defaulting cache accounting to zero makes an implementation that forgot to provide metrics indistinguishable from one that genuinely served zero bytes, so query metrics silently under-report. These should either be required methods or represent unsupported accounting explicitly; the same applies to loadedSize.

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
wendigo: Reading footers when the file length is unknown/incorrect (we had cases in the past where manifest files were adverstising incorrect length). Reading footers is using negative offsets

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
wendigo: This was always the case for alluxio. The next step is to drop alluxio (Dain has a PR for that) so I'd prefer this to stay. Memory based caches provide invalidation

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCacheManager.java
wendigo: correct. if we want to introduce deduplication this will be supported on a different level. there is no use case for this right now

core/trino-main/src/main/java/io/trino/cache/CacheManagerRegistry.java
wendigo: today's capability sets are disjoint, ambiguity fails loudly with an actionable message, and a binding mechanism should be designed when a second same-capability manager actually exists (and per cache-kind, per the k/v argument)

wendigo:
@martint AC

core/trino-spi/src/main/java/io/trino/spi/cache/BlobSource.java
martint: If file length is unknown or incorrect, what does length() return?

core/trino-spi/src/main/java/io/trino/spi/cache/BlobCache.java
martint: What if a cache implementation cannot invalidate the entire hierarchy below a key? If alluxio cannot do it, I can imagine there will be other implementations that cannot do it, either.

wendigo: Is documenting this method as best effort will suffice? Or we want stronger guarantees even if we don't have these right now

wendigo:
TailBlobSource, hasAllContent etc went away. We now require that `TrinoInputFile.length` is a correct and known length of a file. Usually it's populated from the Iceberg manifest (there is an escape hatch that does stat() when the written data sizes are incorrect)

wendigo:
@martint please review SPI

martint:
SPI looks good

raunaqmorarka (commented):
Reviewed everything outside the SPI commit. Inline comments below, plus one general point.

This ships a new deployment surface (`cache-manager.config-files`, `cache-manager.name`, per-manager properties files, the memory manager loaded by default when nothing is configured) and retires the catalog-level `fs.cache.*` alluxio properties, but docs are untouched and the release notes section is unfilled. Operators should also be told that the JMX names move (`io.trino.filesystem.alluxio:type=AlluxioCacheStats` becomes `io.trino.blob.cache.alluxio:...`, and the legacy `io.trino.plugin.hive.fs:TrinoFileSystemCache` name is gone) and that existing alluxio on-disk caches are invalidated on upgrade, since page identifiers are now derived from the length-framed `[catalog, usage, ...]` key components.

plugin/trino-blob-cache-memory/src/main/java/io/trino/blob/cache/memory/MemoryBlob.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public long cachedSize()
+    {
+        return cachedSize.get();
+    }
+
+    @Override
+    public long loadedSize()
+    {
+        return data.length();
```
raunaqmorarka: `loadedSize()` returns `data.length()` unconditionally, but the `Blob` javadoc defines it as bytes fetched from the backing `BlobSource`. On a cache hit nothing is fetched, yet `BlobTrinoInput.getMetrics()` reports `bytesReadExternally` equal to the full file size for every hit, so a fully cached query looks like it read everything from storage. `MemoryBlobCache.getOrLoad` could track whether this call populated the entry (a flag set inside the loader) and pass `loaded ? data.length() : 0` into `MemoryBlob`.

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CacheOutputFile.java
```diff
@@ -0,0 +1,95 @@
...
+            {
+                out.write(buffer, offset, length);
+            }
+
+            @Override
+            public void close()
+                    throws IOException
+            {
+                super.close();
+                invalidation.run();
```
raunaqmorarka: `createOrOverwrite` and `createExclusive` invalidate in `finally` because a failed write may still have mutated storage, but this path runs `invalidation.run()` only when `super.close()` succeeds. A close that fails mid-flush can leave partial content on storage with the stale entry still cached. Wrap this in try/finally for consistency.

plugin/trino-blob-cache-alluxio/src/main/java/io/trino/blob/cache/alluxio/AlluxioBlob.java
```diff
@@ -0,0 +1,125 @@
...
+            CacheManager cacheManager,
+            AlluxioConfiguration configuration,
+            AlluxioCacheStats statistics)
+    {
+        this.delegate = requireNonNull(delegate, "delegate is null");
+        this.description = delegate.toString();
+        this.length = requireNonNull(status, "status is null").getLength();
+        this.statistics = requireNonNull(statistics, "statistics is null");
+        // Positioned reads are already correctly sized by the caller, so skip the small-read buffer
+        this.helper = new AlluxioInputHelper(tracer, description, cacheKey, status, cacheManager, configuration, statistics, false);
```
raunaqmorarka: `AlluxioBlob` always constructs the helper with `bufferSmallReads=false`, and it now backs both `TrinoInput` and `TrinoInputStream`. The old `AlluxioInputStream` passed `true` specifically because sequential reads arrive in small arbitrary chunks. With the buffer gone, every small stream `read()` (avro manifests, metadata JSON, transaction logs) costs a tracing span plus a `cacheManager.get` page lookup instead of an arraycopy from the buffer. Either buffer stream-backed reads (doing it engine-side in `BlobTrinoInputStream` would benefit every cache implementation), or if the regression is acceptable, remove the dead code: with the single caller passing `false`, the `readBuffer`, `doBufferRead` and `putBuffer` paths in `AlluxioInputHelper` are unreachable.

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CacheFileSystem.java
```diff
@@ -85,52 +88,56 @@ public TrinoInputFile newEncryptedInputFile(Location location, long length, Inst
...
 
     @Override
     public void deleteDirectory(Location location)
             throws IOException
     {
-        delegate.deleteDirectory(location);
-        cache.expire(location);
+        // Entries must be enumerated while the files still exist; a listing failure fails the
+        // operation before anything is mutated, so the cache stays consistent with storage
+        List<CacheKey> keys = directoryKeys(location);
```
raunaqmorarka: The old code invalidated on delete without touching storage; this version enumerates every file under the directory through `delegate.listFiles` before mutating. On object stores that doubles LIST traffic for a directory drop (the delete implementation lists again internally), a listing failure now blocks the delete entirely, and entries populated concurrently between enumeration and mutation are missed anyway. The benefit is a `tryInvalidate` that the alluxio manager implements as a no-op and that correctness must not rely on, since keys are versioned. Is the guaranteed listing cost worth the best-effort reclaim, versus relying on TTL and eviction as the `AlluxioCache.tryInvalidate` comment argues?

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CacheInputFile.java
```diff
@@ -70,18 +74,23 @@ public TrinoInputStream newStream()
...
+        // populate a whole entry just to learn its size
         if (length.isEmpty()) {
-            Optional<String> key = keyProvider.getCacheKey(delegate);
-            if (key.isPresent()) {
-                length = OptionalLong.of(cache.cacheLength(delegate, key.orElseThrow()));
-            }
-            else {
-                length = OptionalLong.of(delegate.length());
-            }
+            length = OptionalLong.of(delegate.length());
```
raunaqmorarka: This regresses the warm path. The old `cacheLength` served the length from the cached entry, while this goes to the delegate on every call. The updated `TestIcebergMemoryCacheFileOperations` expectations show `InputFile.length(SNAPSHOT)` on the warm-cache assertions where the old flow had no remote access, so this adds a HEAD request per snapshot file per query on the coordinator planning path. A `length()` call is almost always followed by reading the same file, so populating the entry to answer it is not wasted work. Keep the old behavior of answering `length()` through the cache.

plugin/trino-blob-cache-memory/src/main/java/io/trino/blob/cache/memory/InMemoryBlobCacheManager.java
```diff
@@ -0,0 +1,116 @@
...
+                log.warn(e, "Failed to unregister MemoryBlobCacheStats MBean for catalog %s", catalog);
+            }
+        }
+        sharedCache.tryInvalidate(CacheKey.of(catalog.toString()));
+    }
+
+    @Override
+    public void shutdown()
+    {
+        sharedCache.flushCache();
```
raunaqmorarka, line 96: `AlluxioBlobCacheManager.shutdown()` drops all catalogs and unexports their stats MBeans before shutting down the shared cache; this one only flushes, leaving the `catalogs` map and MBean registrations behind. A manager reloaded in the same JVM then fails to export the same names. Mirror the alluxio implementation.

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/TrinoInputFileBlobSource.java
```diff
@@ -0,0 +1,76 @@
...
+        return input;
+    }
+
+    @Override
+    public void close()
+            throws IOException
+    {
+        if (input != null) {
+            input.close();
+            input = null;
```
raunaqmorarka, line 74: `close()` nulls out `input`, so a `readFully` arriving after close lazily opens a fresh input that nothing will ever close. Per the SPI ownership rules that would be a cache bug, but it leaks a connection silently here instead of failing. Consider making close terminal (closed flag, reject subsequent reads) so it fails fast.

plugin/trino-blob-cache-memory/src/test/java/io/trino/blob/cache/memory/TestMemoryBlobCache.java
```diff
@@ -105,6 +110,35 @@ public void testMaxContentLength()
         getFileSystem().deleteFile(location);
     }
 
+    private static class TestingBlobSource
```
raunaqmorarka, line 185: `TestingBlobSource` is unused. Remove it, or add the direct `MemoryBlobCache.get` tests it was presumably written for. Hit and miss `loadedSize` accounting is currently untested, which is where the `MemoryBlob.loadedSize()` issue hides.

wendigo:
@raunaqmorarka I am aware and going to follow up with doc changes

lib/trino-filesystem/src/main/java/io/trino/filesystem/cache/CacheKeyProvider.java
```diff
@@ -13,16 +13,51 @@
...
+    static CacheKey locationKey(Location location)
+    {
+        StringBuilder root = new StringBuilder();
+        location.scheme().ifPresent(scheme -> root.append(scheme).append("://"));
+        location.userInfo().ifPresent(userInfo -> root.append(userInfo).append('@'));
+        location.host().ifPresent(root::append);
+        location.port().ifPresent(port -> root.append(':').append(port));
+        List<String> components = new ArrayList<>();
+        components.add(root.toString());
+        Splitter.on('/').omitEmptyStrings().split(location.path()).forEach(components::add);
```
raunaqmorarka: `omitEmptyStrings` makes `a//b` and `a/b` produce the same key, but those are distinct objects on S3, and double slashes do show up in table locations in practice. The hive provider still appends mtime and length, but the iceberg and delta keys are pure location, so a collision would serve the wrong bytes. Dropping only trailing empty segments is enough for directory prefix matching, interior ones should be kept.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -273,9 +274,20 @@ public CompletableFuture<List<ConnectorSplit>> getNextBatch(int maxSize, Dynamic
...
+        // exceptionally(), which would wrap it in a CompletionException and hide it from callers
+        // that unwrap a single layer
+        CompletableFuture<List<ConnectorSplit>> future = new CompletableFuture<>();
+        toCompletableFuture(nextBatchFuture).whenComplete((splits, throwable) -> {
+            if (throwable == null) {
+                future.complete(splits);
+            }
+            else {
+                Throwable cause = throwable instanceof CompletionException && throwable.getCause() != null ? throwable.getCause() : throwable;
+                future.completeExceptionally(translateMetadataException(cause, tableHandle.getSchemaTableName().toString()));
```
raunaqmorarka: This and the throw to return change in `IcebergExceptions` alter error propagation for non-caching deployments too, and neither has a test. Can you move them to a separate commit with a note on what broke without them?

wendigo: Dropped this. Not needed

raunaqmorarka (approved):

wendigo:
@coderabbitai full review

PR #29185, opened 2026-04-21
Inline analyzer Field factories

ksobolew:
## Description

With the builder added in #29183 it is more readable now to use it instead of the more-or-less-opaque factory methods. This is especially relevant when:

* There are many parameters, in which case it's easier to see which is which instead of a positional parameter soup
* The `Field` constructed is actually a modified copy of some original field, in which case we can only set the fields that change

Follow-up to #29183.

## Release notes

```markdown
## Section
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
@@ -2633,7 +2633,7 @@ public class ExpressionAnalyzer
             for (int i = 0; i < lambdaArguments.size(); i++) {
                 LambdaArgumentDeclaration lambdaArgument = lambdaArguments.get(i);
                 Type type = types.get(i);
-                fields.add(Field.newUnqualified(lambdaArgument.getName().getValue(), type));
+                fields.add(Field.builder().name(lambdaArgument.getName().getValue()).type(type).build());
                 setExpressionType(lambdaArgument, type);
             }
 
--- a/core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
+++ b/core/trino-main/src/main/java/io/trino/sql/analyzer/Field.java
@@ -36,24 +36,6 @@ public class Field
         return new Builder();
     }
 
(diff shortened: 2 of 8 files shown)


core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
```diff
@@ -2633,7 +2633,7 @@ protected Type visitLambdaExpression(LambdaExpression node, Context context)
             for (int i = 0; i < lambdaArguments.size(); i++) {
                 LambdaArgumentDeclaration lambdaArgument = lambdaArguments.get(i);
                 Type type = types.get(i);
-                fields.add(Field.newUnqualified(lambdaArgument.getName().getValue(), type));
+                fields.add(Field.builder().name(lambdaArgument.getName().getValue()).type(type).build());
```
martint, line 2636: One of the benefits of a builder is that values can be built incrementally, across multiple invocations. None of the usages fall in that category, and using a builder inline in this way just makes the call site more cumbersome and harder to read.

martint (changes requested):

core/trino-main/src/main/java/io/trino/sql/analyzer/ExpressionAnalyzer.java
ksobolew: Just these "smaller" methods or all of them?

ksobolew: I can skip the first two but I believe the final two are actually beneficial, especially the last one, where we convert a lot sites where we create a modified instance. As an additional benefit, this automatically handles cases where a new property is added to `Field` that also needs to be copied.

martint: Let me check the other ones. 

However, part of the issue with the builder is that it hides the intent that was explicit in the static method. "newUnqualified" has a specific meaning. It's not obvious that the equivalent builder-based usage intends to do the same. It also makes it hard to know what are valid combinations.

PR #29186, opened 2026-04-21
Delegate encrypted file operations in CacheFileSystem

kaveti:
Previously, invoking newEncryptedInputFile, newEncryptedOutputFile, or encryptedPreSignedUri on CacheFileSystem threw UnsupportedOperationException from the default TrinoFileSystem interface methods. This prevented backing filesystems that support server-side encryption (e.g. S3 SSE-C, GCS CSEK) from being served through the coordinator file cache.

Add overrides that forward the encrypted operations to the underlying filesystem and expire cached entries on writes, mirroring the behavior of the existing unencrypted overrides.

Also add an EncryptionKey.ofAes256(byte[]) factory so callers can construct an AES-256 key from caller-supplied bytes without hardcoding the algorithm string.

Moved changes in to Separate PR from here : https://github.com/trinodb/trino/pull/28793

## Release notes

```markdown
## Section
```

### DIFF
--- a/lib/trino-filesystem/src/main/java/io/trino/filesystem/encryption/EncryptionKey.java
+++ b/lib/trino-filesystem/src/main/java/io/trino/filesystem/encryption/EncryptionKey.java
@@ -31,6 +31,11 @@ public record EncryptionKey(byte[] key, String algorithm)
     {
         byte[] key = new byte[32];
         ThreadLocalRandom.current().nextBytes(key);
+        return ofAes256(key);
+    }
+
+    private static EncryptionKey ofAes256(byte[] key)
+    {
         return new EncryptionKey(key, "AES256");
     }
 
--- a/lib/trino-filesystem/src/test/java/io/trino/filesystem/cache/TestCacheFileSystemEncryption.java
+++ b/lib/trino-filesystem/src/test/java/io/trino/filesystem/cache/TestCacheFileSystemEncryption.java
@@ -1,0 +1,276 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


lib/trino-filesystem/src/main/java/io/trino/filesystem/encryption/EncryptionKey.java
```diff
@@ -27,11 +27,16 @@ public record EncryptionKey(byte[] key, String algorithm)
         requireNonNull(key, "key is null");
     }
 
+    public static EncryptionKey ofAes256(byte[] key)
```
wendigo: can be private, move below `randomAes256`

lib/trino-filesystem/src/test/java/io/trino/filesystem/cache/TestCacheFileSystemEncryption.java
```diff
@@ -0,0 +1,276 @@
...
+ * {@link UnsupportedOperationException} from the default {@link TrinoFileSystem}
+ * interface implementations.
+ */
+final class TestCacheFileSystemEncryption
+{
+    private static final Location LOCATION = Location.of("memory:///file");
+    private static final EncryptionKey KEY = EncryptionKey.randomAes256();
+
+    @Test
+    void newEncryptedInputFileDelegates()
```
ebyhr: Please add `test` prefix. 

https://trino.io/docs/current/develop/tests.html
* Test method must start with test, for example testExplain()

findinpath (approved):

findinpath:
@wendigo  can we land this change?

wendigo (approved):

ebyhr:
Please update "Release notes" section in the PR description.

PR #29120, opened 2026-04-15
Add Doris connector

ningsh7:
## Description

Introduce the initial Doris connector implementation.

The connector uses three Doris interfaces:
- Doris FE HTTP for query planning and split generation
- Doris FE JDBC for metadata access
- Apache Arrow Flight SQL for data reads

It includes support for metadata access, type mapping, split generation, predicate and limit pushdown, and reading Doris tables from Trino.

This change also adds the required configuration, unit tests, integration tests, and product tests to validate the implementation and provide a baseline for future improvements.

## Additional context and related issues

Compared with using the MySQL connector against Doris, this connector adds Doris-specific split planning, Flight SQL-based reads, and Doris-specific type handling.

Related discussions:
- #8653
- #28735
- #24720

Testing includes:
- unit tests for configuration, metadata, split planning, and type mapping
- connector tests following the `BaseConnectorTest` pattern
- smoke and integration tests against a real Doris instance
- product tests with a Docker-based Doris environment

Follow-up changes may further improve performance, high availability, caching, and pushdown coverage.

## Release notes

(x) Release notes are required, with the following suggested text:

```markdown
## Doris connector
* Add initial support for querying Doris tables from Trino.

Issue #28735, opened 2026-03-18
RFC / Interest check: JDBC-based Arrow Flight SQL connector for Doris / StarRocks?


ningsh7:
### Feature Request / Improvement

Hi all,

I’d like to do a small interest check / RFC for the following idea:

**Would the community be interested in a connector for systems like Doris / StarRocks that keeps a JDBC-style integration on the Trino side, but uses the Arrow Flight SQL JDBC driver underneath?**

#### Background

There are already two related open issues:

- #8653 Add apache doris connector
- #17329 Add StarRocks Connector

From our understanding, most existing discussion around Doris / StarRocks connectivity is still centered around traditional JDBC / MySQL wire protocol approaches. At the same time, both systems now support Arrow Flight SQL, and Apache Arrow already provides a standard Flight SQL JDBC driver.

So we are wondering whether the following direction would be useful for Trino as well:

- **On the Trino side**: keep a JDBC-style / base-jdbc-style connector implementation model;
- **On the data source side**: use `jdbc:arrow-flight-sql://...` instead of a traditional protocol such as `jdbc:mysql://...`;
- **Initial target systems**: Doris and StarRocks, since they already implement Flight SQL.

#### What this is *not* about

This is **not** about:

- making Trino expose Arrow Flight SQL as a client/server protocol;
- replacing all JDBC connectors with ADBC;
- redesigning Trino’s external client protocol.

The narrower question is:

> For databases that already support Flight SQL, would it make sense to explore a Trino connector approach that keeps a JDBC-style API/integration model, while switching the underlying transport to Arrow Flight SQL?

#### Potential value of this direction

We think this direction may offer some practical benefits:

- reuse of JDBC-style connector implementation patterns and maintenance model;
- continued use of a standard JDBC driver shape in the Java ecosystem;
- potentially better fit than MySQL wire protocol for large result-set reads because Flight SQL is columnar-oriented;
- less protocol-specific adaptation for systems that already support Flight SQL natively.

That said, we are treating this as an exploration, not assuming it is automatically beneficial in every case.

#### Questions for the community

1. Does this direction seem valuable to the community?
2. If yes, would it be better to:
   - build separate Doris / StarRocks connectors, or
   - build a more generic Flight SQL JDBC connector with small backend-specific adaptations?
3. From a Trino maintainability perspective, would the community prefer:
   - extending the existing JDBC-style / base-jdbc approach, or
   - implementing this as a standalone connector?
[...]

PR #24720, opened 2025-01-16
[WIP] Add StarRocks connector

Max-Cheng:
## Description
Support Trino access Starrocks by connector

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
(x) Release notes are required. Please propose a release note for me.
( ) Release notes are required, with the following suggested text:

```markdown
## Section
* Add Starrocks Connector
```

PR #17330, opened 2023-05-03
Add StarRocks connector

chenjian2664:
## Description
Fix https://github.com/trinodb/trino/issues/17329

## Release notes

( ) This is not user-visible or docs only and no release notes are required.
( ) Release notes are required, please propose a release note for me.
(x) Release notes are required, with the following suggested text:

```markdown
# StarRocks
* Add StarRocks connector. ({issue}`17329`)
```

PR #29189, opened 2026-04-22
Add StarRocks connector

infocusmodereal:
## Description

Add an initial native read-only StarRocks connector.

The connector uses the StarRocks JDBC driver for metadata discovery and Apache Arrow Flight SQL for reads. This avoids MySQL-specific metadata assumptions that can break on StarRocks, such as relying on JDBC `COLUMN_SIZE` being present during type discovery.

The initial implementation includes:
- schema, table, view, and column introspection
- `DESCRIBE` and `SHOW CREATE TABLE` support
- read-only `SELECT` queries over Arrow Flight SQL
- predicate, limit, numeric and temporal TopN, and basic aggregation pushdown
- Flight SQL partitioned split planning with a safe single-split fallback
- optional TLS configuration for Flight SQL connections
- bounded Arrow memory allocation per Flight SQL stream
- StarRocks-aware scalar type mapping with conservative handling for unsupported or partially supported types

Out of scope for v1:
- writes and schema mutations
- connector-managed DDL for complex types
- native decoding for `ARRAY`, `MAP`, and `STRUCT`
- direct reads of StarRocks `HLL` state columns
- optimizations beyond StarRocks Flight SQL partition descriptors and the pushdowns implemented in this PR

This work follows the design direction discussed in #28735 and is informed by the Doris Flight SQL connector in #29120, while remaining a StarRocks-specific implementation rather than extending the generic MySQL connector path.

Prior StarRocks-specific work for context:
- #17330
- #24720

The main StarRocks-specific choices in this PR are:
- metadata comes from the StarRocks JDBC driver and StarRocks `INFORMATION_SCHEMA` rather than MySQL connector metadata code
- reads use Arrow Flight SQL rather than the MySQL wire protocol
- catalog-aware metadata supports `starrocks.catalog-name`, with a fully qualified `SHOW FULL COLUMNS` fallback for StarRocks versions where `INFORMATION_SCHEMA.COLUMNS.TABLE_CATALOG` is absent or not populated
- unsupported aggregate forms, risky ordering semantics, and unproven complex values fall back to Trino execution or conservative type mapping instead of breaking table introspection

Known v1 limitations:
- empty databases inside a configured StarRocks catalog may not be listed if StarRocks does not expose catalog information for empty schemas
- `HLL` state columns are not exposed because StarRocks does not support direct reads of HLL values
- `ARRAY`, `MAP`, and `STRUCT` are mapped conservatively to `VARCHAR`

## Testing

This PR adds coverage for:
- connector configuration and type mapping
- StarRocks metadata fallback behavior
- Arrow value conversion
- SQL generation for pushdown
- split planning and page source behavior
- connector smoke tests following the `BaseConnectorSmokeTest` pattern
- opt-in integration smoke tests against a live StarRocks instance

Validated locally with:
- `./mvnw -pl plugin/trino-starrocks test`
- `./mvnw -pl plugin/trino-starrocks clean compile test-compile -DskipTests -Dair.check.skip-all=true -P errorprone-compiler`
- the gated StarRocks integration smoke test against a live StarRocks instance outside the repository

## Release notes

```markdown
## StarRocks connector
* Add the native read-only StarRocks connector backed by StarRocks JDBC metadata and Arrow Flight SQL reads. ({issue}`28735`)
```

### DIFF
--- a/docs/src/main/sphinx/connector.md
+++ b/docs/src/main/sphinx/connector.md
@@ -41,6 +41,7 @@ Redshift        <connector/redshift>
 SingleStore     <connector/singlestore>
 Snowflake       <connector/snowflake>
 SQL Server      <connector/sqlserver>
+StarRocks       <connector/starrocks>
 System          <connector/system>
 Thrift          <connector/thrift>
 TPC-DS           <connector/tpcds>
--- a/docs/src/main/sphinx/connector/starrocks.md
+++ b/docs/src/main/sphinx/connector/starrocks.md
@@ -1,0 +1,274 @@
+# StarRocks connector
+
+The StarRocks connector allows querying tables and views in an external
+[StarRocks](https://www.starrocks.io/) cluster.
+
+This connector is purpose-built for StarRocks. It does not rely on the generic
+MySQL connector path, because StarRocks metadata and type reporting do not
+always match MySQL connector assumptions.
+
+In v1, the connector is read-only:
+
+- metadata is discovered through the native StarRocks JDBC driver
+- table reads are executed through Arrow Flight SQL
+- writes and schema changes are intentionally out of scope
+
+## Requirements
+
+To connect to StarRocks, you need:
+
+- StarRocks 3.5.1 or higher
+- network access from the Trino coordinator and workers to the StarRocks FE
+  JDBC endpoint and Arrow Flight SQL endpoint
+- a StarRocks user with permission to inspect metadata and run read queries
+
+Arrow Flight SQL must be enabled on both FE and BE nodes, and the FE should be
+reachable by Trino on the configured Flight SQL port.
+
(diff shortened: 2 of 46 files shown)


infocusmodereal:
@cla-bot check

infocusmodereal:
Reopened and rebased onto the current `master`. The connector module passes the full local Maven checks on Temurin 25, and the opt-in integration smoke suite passes against StarRocks 4.0 with Arrow Flight SQL enabled (6/6 scenarios). CI has been restarted. This is ready for maintainer review.

legiondean:
Any plan to make the connector available soon? I see the connector was actually available in 443:
https://ta.thinkingdata.cn/presto-docs/connector/starrocks.html
Not sure why it's removed from the list.

PR #22527, opened 2024-06-27
Add preferred timeout for small dynamic filters

Dith3r:
## Description

Add a CBO rule to estimate if dynamic filter is worth waiting for. Focus on small tables that are most often used as dimension tables and nodes that generate small number of distinct values.

Rule DeterminePreferredDynamicFilterTimeout could allow us to remove 20s forced wait time for every possible DF in JDBC connectors if table presents the least row count statistic.

Query | Baseline duration[s] | PR duration[s]
--- | --- | ---
tpcds/q59 | 12.10 | 4.55
tpcds/q02 | 9.63 | 3.47
tpcds/q36 | 5.82 | 2.60
tpcds/q53 | 8.04 | 4.80 
tpcds/q13 | 7.7 | 5.1

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(X) Release notes are required, with the following suggested text:

```markdown
# General
* Improve performance of queries with selective joins. ({issue}`22527`)
```

PR #29206, opened 2026-04-22
Move dynamic filter wait to engine

chenjian2664:
## Description

Move dynamic filter waiting logic from individual connectors into the engine layer, and remove `DynamicFilter` from the connector SPI entirely.

**Motivation**

Quote from @raunaqmorarka:
> There's no reason why every connector should be hand rolling roughly similar waiting loop on DF. We did it this way in the beginning because connector was the only one benefitting from DFs in the past and it started from a small number of connectors. Since then the engine has gained dynamic row filtering and many more connectors have added similar waiting loop on DF.
> The existing contract between engine and connectors for DFs is also overly complex because the io.trino.spi.connector.DynamicFilter API forces the connector to deal with an object that can update under the covers any time and reason about waiting on it. We want to eventually move io.trino.spi.connector.DynamicFilter out of SPI.

Previously, `IcebergSplitSource`, `DeltaLakeSplitSource`, `HiveSplitSource`, and JDBC each duplicated the same pattern: hold a `DynamicFilter`, poll `isAwaitable()`/`isBlocked()`, manage a timeout, and only then read `getCurrentPredicate()`. This PR centralizes that logic in the engine and makes it structurally impossible for connectors to reimplement it.

**SPI changes:**

- Replace `DynamicFilter` in `ConnectorSplitManager#getSplits` with `Set<ColumnHandle> dynamicFilterColumns` (the static set of columns the DF will cover). Connectors no longer receive a live `DynamicFilter` at split planning time — the predicate isn't stable then anyway.
- Add `DynamicFilterSnapshot(TupleDomain<ColumnHandle> currentPredicate, boolean isComplete)` record as a per-batch snapshot of the engine's DF state, passed to each `getNextBatch` call.
- Replace `ConnectorSplitSource#getNextBatch(int)` with `getNextBatch(int, DynamicFilterSnapshot)` as an abstract method. `ConnectorSplitBatch` is also removed — the method now returns `List<ConnectorSplit>` directly, with `isFinished()` replacing `noMoreSplits`.
- Add `ConnectorSplitSource#getRequestedDynamicFilterWaitTimeoutMillis()` (default `0`) so each connector controls its DF wait budget without managing the wait mechanism.

**Engine changes:**

- `ConnectorAwareSplitSource` owns the DF wait loop: reads `getRequestedDynamicFilterWaitTimeoutMillis()` once at construction, waits on `DynamicFilter.isBlocked()` before the first batch if awaitable and time remains, then passes a `DynamicFilterSnapshot` on every `getNextBatch` call.
- `SplitManager` calls the new `getSplits(..., dynamicFilter.getColumnsCovered(), constraint)`.

**Connector changes:**

- **Iceberg:** removes internal `DynamicFilter`/`Stopwatch` fields and wait loop; removes `iceberg.dynamic-filtering.wait-timeout` config property and `dynamic_filtering_wait_timeout` session property.
- **Delta Lake:** same pattern as Iceberg.
- **Hive:** `HiveSplitSource` receives the per-batch predicate via `getNextBatch`; the partition loader receives the covered-columns set at construction.
- **JDBC:** `DynamicFilteringJdbcSplitSource` uses the per-batch snapshot directly.
- All other connectors implement the new abstract `getSplits` and `getNextBatch` signatures — the compiler enforces migration.


ebyhr:
Could you rebase on master to resolve conflicts?

core/trino-main/src/main/java/io/trino/execution/DynamicFilterConfig.java
```diff
@@ -234,6 +237,20 @@ public DynamicFilterConfig setPartitionedMaxSizePerOperator(DataSize partitioned
         return this;
     }
 
+    @NotNull
+    public Duration getDynamicFilteringWaitTimeout()
+    {
+        return dynamicFilteringWaitTimeout;
+    }
+
+    @Config("dynamic-filtering.wait-timeout")
```
ebyhr: @raunaqmorarka Do you know why the DF timeout and the relevant logic were implemented on the connectors' side? (To avoid SPI change ConnectorSplitSource.getNextBatch, or allow different values per catalog?)

ebyhr (commented):
I hope @raunaqmorarka chimes in this PR.

core/trino-main/src/main/java/io/trino/execution/DynamicFilterConfig.java
raunaqmorarka: DF timeout value should continue to be provided from connectors. There is different trade-off of waiting for DF in each connector (e.g. we want to wait longer for JDBC compared to iceberg).
This can be done by plugging existing connector config values into a new API `ConnectorSplitSource#getDynamicFilterWaitTimeout`.
We shouldn't need new config in engine and no removal of existing configs in connectors.

chenjian2664: updated, PTAL

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -30,6 +32,17 @@
 public interface ConnectorSplitSource
         extends Closeable
 {
+    /**
+     * Returns the next batch of splits. The {@code dynamicFilterPredicate} reflects the
+     * current dynamic filter state resolved by the engine before this call. For connectors
+     * that do not support dynamic filtering, {@link TupleDomain#all()} is passed.
+     */
+    default CompletableFuture<ConnectorSplitBatch> getNextBatch(int maxSize, TupleDomain<ColumnHandle> dynamicFilterPredicate)
```
raunaqmorarka: A few suggestions for the javadoc on this method, since it's now the primary entry point:

- Add `@since <release>` so connector authors can gate compatibility.
- State that implementations should override this method rather than the deprecated `getNextBatch(int)`.
- Clarify the contract on `dynamicFilterPredicate`: it is a snapshot taken before this call, not a live view; the engine may have waited up to `getDynamicFilterWaitTimeoutMillis()` before the first call, and on subsequent calls passes the current predicate without further waiting.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -30,6 +32,17 @@
...
+    {
+        return getNextBatch(maxSize);
+    }
+
+    @Deprecated
```
raunaqmorarka: Consider `@Deprecated(forRemoval = true, since = "<release>")` to make the removal commitment explicit, and a `@deprecated` javadoc tag pointing at the replacement so connectors picking up the warning have a migration hint.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -53,6 +66,16 @@ default Optional<List<Object>> getTableExecuteSplitsInfo()
         return Optional.empty();
     }
 
+    /**
+     * Returns the maximum time in milliseconds the engine should wait for dynamic filters to be
+     * collected before generating splits. Connectors that do not support dynamic filter waiting
+     * should return {@link OptionalLong#empty()}.
+     */
+    default OptionalLong getDynamicFilterWaitTimeoutMillis()
```
raunaqmorarka: Suggest changing this to:

```java
default long getRequestedDynamicFilterWaitTimeoutMillis()
{
    return 0;
}
```

- Rename to `getRequestedDynamicFilterWaitTimeoutMillis` — the current name reads like elapsed time; `Requested` makes it clear this is the connector's preference for the engine to consider.
- Drop the `OptionalLong` wrapper. `0` and `OptionalLong.empty()` are observationally identical in `ConnectorAwareSplitSource` (`.orElse(0L)`), so the wrapper adds no information. Define `0` as "no waiting" in the javadoc.
- Worth documenting the lifecycle: the engine reads this once at split-source construction, so it is fixed for the lifetime of the source (per table scan).

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
raunaqmorarka: With the engine now owning the split-generation-time wait (via `getRequestedDynamicFilterWaitTimeoutMillis()`), the only surface connectors legitimately use at split-generation time is read-only: `getColumnsCovered()` (static), `getCurrentPredicate()` (lazy reads), `isComplete()` (caching short-circuit). Two cleanly orthogonal channels cover all in-tree call sites:

**1. Static scan-wide info on `getSplits`** — replace `DynamicFilter dynamicFilter`:

```java
ConnectorSplitSource getSplits(
        ConnectorTransactionHandle transaction,
        ConnectorSession session,
        ConnectorTableHandle table,
        Set<ColumnHandle> dynamicFilterColumns,
        Constraint constraint);
```

**2. Per-batch snapshot on `getNextBatch`** — bundle predicate + completion into a record:

```java
public record ConnectorDynamicFilter(TupleDomain<ColumnHandle> currentPredicate, boolean isComplete) {}

default CompletableFuture<ConnectorSplitBatch> getNextBatch(int maxSize, ConnectorDynamicFilter dynamicFilter);
```

No connector-facing `getCurrentPredicate()` accessor — exactly one canonical channel for the predicate. Lazy-read sites (Hive's `transform` in `HiveSplitManager:336`, `PartitionMatchSupplier`) stash the latest snapshot in a volatile field on the split source.

Suggest handling this in this PR rather than as a follow-up so the SPI doesn't end up with both channels coexisting. Same deprecation playbook as `getNextBatch` here: deprecate the current `getSplits(..., DynamicFilter, ...)` overload, add the new one default-bridged via an engine-provided adapter, and migrate Iceberg + Delta in this PR. Hive / Mongo / JDBC / Memory keep using the deprecated overload until follow-up PRs migrate them; removal of the deprecated method waits until then.

`DynamicFilter` itself is untouched — page-source-time uses (`*PageSourceProvider`, engine `DynamicPageFilter`) are out of scope here. End-state, the same `ConnectorDynamicFilter` record extends cleanly to the page-source SPI without growth: page-source-time DF blocking is structurally redundant given (a) this PR's split-source wait covers coordinator DFs and (b) `LocalExecutionPlanner.createLookupJoin:2949` already blocks the probe pipeline on build completion (`waitForBuild`) for node-local DFs.

chenjian2664: Replaced the `DynamicFilter` with `ConnectorDynamicFilter` in Hive. When Hive loads splits, it first loads partitions, so the DF should participate in pruning at this stage. 
We now split the static context (covered columns) and non-static context (predicate and isComplete) — we only get the predicate when the upstream calls `getNextBatch` on the connector splits and passes it. This means we need to share the predicate between the Hive partition loader and the connector split source, updating it when the upstream calls `getNextBatch`.
I put a ~~AtomicReference~~ CompleteFuture for sharing such predicate, let me know if you have better alternatives.

chenjian2664:
Updated based on the review feedback:

- **`ConnectorSplitSource`**: Renamed `getDynamicFilterWaitTimeout` → `getRequestedDynamicFilterWaitTimeoutMillis()`, returning `long` (not `OptionalLong`) with `0` meaning "no waiting". Each connector continues to provide its own timeout — no engine-level config added. Added `@since 481` and `@Deprecated(forRemoval = true, since = "481")` javadoc as suggested.
- **`ConnectorSplitManager`**: New `getSplits(..., Set<ColumnHandle> dynamicFilterColumns, ...)` overload replaces the `DynamicFilter` overload for static scan-wide info. Old `DynamicFilter` overload deprecated for removal. Engine (`SplitManager`) calls the new overload.
- **Connectors migrated to new `getSplits` API**: Iceberg, Delta Lake, Hive, JDBC, MongoDB, Hudi, LakeHouse.
- **PR description** updated to reflect the current design.

CI running now. The previous run's `TestDistributedFaultTolerantEngineOnlyQueries` timeout was a pre-existing resource-deadlock flakiness under CI memory pressure (stage 3 held memory waiting for stage 4 output, while stage 4 couldn't get a node because stage 3 held all available memory) — unrelated to this PR. `TestFaultTolerantExecutionDynamicFiltering` (the DF-specific FTE test) passed 18/18.

chenjian2664:
@losipiuk Could you please take a look at [Update engine and classloader wrappers to use new split APIs](https://github.com/trinodb/trino/pull/29206/commits/3497e93246b5a53e8d9d4f25a4a47307242b366e) this one touches FTE relate code paths , I'd appreciate having your eyes on it

losipiuk:
> @losipiuk Could you please take a look at [Update engine and classloader wrappers to use new split APIs](https://github.com/trinodb/trino/pull/29206/commits/3497e93246b5a53e8d9d4f25a4a47307242b366e) this one touches FTE relate code paths , I'd appreciate having your eyes on it

This looks fine, but it does not seem very much related to FTE, at least not directly - did I miss sth obvious?

chenjian2664:
@losipiuk Thanks for reviewing. I noticed the FTE tests were already failing when I started adding changes to this commit — mainly the tests related to `AbstractTestCoordinatorDynamicFiltering` and the Memory connector refactoring. I assumed this might also be affecting the FTE tests. :)

losipiuk:
There may be some timing issues. Can you point me to example failures?

chenjian2664:
@losipiuk https://github.com/trinodb/trino/actions/runs/25434596125/job/74609407569 and https://github.com/trinodb/trino/actions/runs/25430147008/job/74594178701 
Those two, so I modified the Memory connector https://github.com/trinodb/trino/pull/29206/changes/3497e93246b5a53e8d9d4f25a4a47307242b366e#diff-ec6f0bb035faecd81daffeef7eadfb9542ca04ccdfdb5cfd06b5e978bdf7ffadR78-R90
Now ci is green, but I am not confident if it is the right refactoring to resolve it

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
raunaqmorarka: `complete()` is a no-op after the first call, so only the first snapshot is ever observed. Two lazy-read sites regress:

- `getPartitionMetadata`'s `transform` calls `.join().currentPredicate()` per partition batch. It needs the latest predicate so late-resolving DFs still prune.
- `PartitionMatchSupplier.evaluateCurrentDynamicFilter` re-evaluates each call and caches once `isComplete()` flips. A frozen first snapshot means the cache short-circuit never triggers.

Suggest splitting the future into latest-snapshot storage and first-arrival gate:

```java
private volatile ConnectorDynamicFilter latestDynamicFilter =
        new ConnectorDynamicFilter(TupleDomain.all(), false);
private final CompletableFuture<Void> firstSnapshot = new CompletableFuture<>();
```

`getNextBatch` writes `latestDynamicFilter = dynamicFilter; firstSnapshot.complete(null)`. Lazy readers use `latestDynamicFilter` directly. `loadSplits` blocks on `firstSnapshot`.

The gate is still needed with engine-side waiting: `BackgroundHiveSplitLoader.start()` runs inside `getSplits()`, before the engine's wait window opens at first `getNextBatch`. Without it, the loader produces un-pruned splits during the wait.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -286,6 +269,12 @@ private synchronized ConnectorSplitBatch getNextBatchInternal(int maxSize)
                 return NO_MORE_SPLITS_BATCH;
             }
 
+            this.predicatedColumnIds = Stream.concat(
+                            tableHandle.getUnenforcedPredicate().getDomains().orElse(ImmutableMap.of()).keySet().stream(),
+                            dynamicFilterPredicate.getDomains().orElse(ImmutableMap.of()).keySet().stream())
```
raunaqmorarka: `predicatedColumnIds` regresses here. It's now derived from `dynamicFilterPredicate.getDomains().keySet()` (the per-batch resolved predicate at first call) instead of the static set of columns the DF will cover. If the DF hasn't resolved by first `getNextBatch`, the keyset is empty and stats won't be gathered for DF columns.

The static channel `Set<ColumnHandle> dynamicFilterColumns` is exactly for this. It's available at `getSplits` time and stable for the lifetime of the scan. `IcebergSplitManager.getSplits` already receives it but doesn't pass it through. Threading it into the constructor lets `predicatedColumnIds` go back to a `final` field computed once:

```java
this.predicatedColumnIds = Stream.concat(
                tableHandle.getUnenforcedPredicate().getDomains().orElse(ImmutableMap.of()).keySet().stream(),
                dynamicFilterColumns.stream().map(IcebergColumnHandle.class::cast))
        .map(IcebergColumnHandle::getId)
        .collect(toImmutableSet());
```

That also drops the `@Nullable @GuardedBy("this")` on the field, the `verify(predicatedColumnIds != null)` defensive check, and the parameter pass-through to `createFileScanTaskWithDomain`.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergSplitSource.java
```diff
@@ -138,80 +133,15 @@ protected QueryRunner createQueryRunner()
...
-            assertThat(splitSource.isFinished()).isTrue();
-            assertThat(System.currentTimeMillis() - startMillis)
-                    .as("IcebergSplitSource failed to wait for dynamicFilteringWaitTimeout")
-                    .isGreaterThanOrEqualTo(2000);
-        }
-        finally {
-            isBlocked.complete(null);
-        }
+        IcebergSplit split = generateSplit(nationTable, tableHandle, TupleDomain.all());
+        assertThat(split).isNotNull();
```
raunaqmorarka: Asserts only that `split` is non-null. `testFileStatisticsDomain` already exercises the same path with stronger assertions. Suggest dropping this test.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/TestIcebergSplitSource.java
```diff
@@ -222,52 +152,24 @@ public void testFileStatisticsDomain()
...
-            @Override
-            public TupleDomain<ColumnHandle> getCurrentPredicate()
-            {
-                return TupleDomain.all();
-            }
-        });
+        // Stats for regionKey are included when it appears in the dynamic filter predicate passed to getNextBatch
+        TupleDomain<ColumnHandle> dynamicFilterPredicate = TupleDomain.withColumnDomains(
+                ImmutableMap.of(regionKey, Domain.singleValue(BIGINT, 1L)));
+        split = generateSplit(nationTable, tableHandle, dynamicFilterPredicate);
```
raunaqmorarka: The comment and setup describe the regressed behavior. Stats should be gathered when `regionKey` is in the static `dynamicFilterColumns` set passed to `getSplits`, not when the per-batch predicate happens to contain it. After fixing the `predicatedColumnIds` regression, plumb `Set<ColumnHandle> dynamicFilterColumns` through the `generateSplit` helper and assert stats are included for `regionKey` when it's in `dynamicFilterColumns` and the per-batch predicate is `(TupleDomain.all(), false)`. That covers the case where the DF resolves after split-source construction.

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
```diff
@@ -72,7 +83,15 @@ public CatalogHandle getCatalogHandle()
...
-        ListenableFuture<ConnectorSplitBatch> nextBatch = toListenableFuture(source.getNextBatch(maxSize));
+        long timeLeft = dynamicFilteringWaitTimeoutMillis - dynamicFilterWaitStopwatch.elapsed(MILLISECONDS);
+        if (dynamicFilter.isAwaitable() && timeLeft > 0) {
+            CompletableFuture<SplitBatch> emptyBatch = dynamicFilter.isBlocked()
+                    .thenApply(_ -> new SplitBatch(ImmutableList.of(), false))
+                    .completeOnTimeout(new SplitBatch(ImmutableList.of(), false), timeLeft, MILLISECONDS);
+            return toListenableFuture(emptyBatch);
+        }
+        ListenableFuture<ConnectorSplitBatch> nextBatch = toListenableFuture(
+                source.getNextBatch(maxSize, new ConnectorDynamicFilter(dynamicFilter.getCurrentPredicate(), dynamicFilter.isComplete())));
```
raunaqmorarka: Java evaluates left-to-right, so `getCurrentPredicate()` is read before `isComplete()`. The invariant (see `PartitionMatchSupplier:62`) is the reverse: read `isComplete` first, then predicate. If DF resolves between the two reads, the wrapper hands out `(stale partial predicate, isComplete=true)` and consumers that cache on `isComplete` lock in the stale predicate.

```java
boolean isComplete = dynamicFilter.isComplete();
TupleDomain<ColumnHandle> currentPredicate = dynamicFilter.getCurrentPredicate();
```

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DynamicFilteringStats.java
```diff
@@ -57,15 +57,12 @@ public TimeStat getDynamicFilterWaitingTime()
         return dynamicFilterWaitingTime;
     }
 
-    public void processDynamicFilter(DynamicFilter dynamicFilter, Duration waitingTime)
+    public void processDynamicFilter(ConnectorDynamicFilter dynamicFilter, Duration waitingTime)
```
raunaqmorarka: Suggest removing this class. With the engine owning the wait, the metrics no longer measure what their names suggest, and equivalent stats are already collected centrally by `DynamicFilterService` and surfaced on stage/operator stats.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -53,6 +76,19 @@ default Optional<List<Object>> getTableExecuteSplitsInfo()
...
     }
 
+    /**
+     * Returns the maximum time in milliseconds the engine should wait for dynamic filters to be
+     * collected before the first call to {@link #getNextBatch(int, ConnectorDynamicFilter)}.
+     * Return {@code 0} to disable waiting (the default).
+     * <p>
+     * The engine reads this value once when the split source is constructed; it is treated as
+     * fixed for the lifetime of the source (per table scan).
+     */
```
raunaqmorarka, line 83: The engine also stops waiting once the DF is no longer awaitable (`ConnectorAwareSplitSource:86-92`); Memory relies on this by returning `Long.MAX_VALUE` to mean "wait until complete." Suggest making the contract explicit:

```
The engine waits up to this timeout, or until the dynamic filter is fully resolved,
whichever comes first.
```

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: ah, good catch! I misunderstood `complete`, thought it behaved like a mutable set method. Let me fix the usage

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
chenjian2664: You are right, the columns should use dynamicFilterColumns directly

raunaqmorarka:
@coderabbitai full review

findepi:
There is a lot of detail in the PR description but somehow one thing escaped my attention -- "the Why". What problem is this trying to solve?

chenjian2664:
updated the pr description cc @findepi

findepi:
> Move dynamic filter waiting logic from individual connectors into the engine layer by adding a TupleDomain dynamicFilterPredicate parameter to ConnectorSplitSource#getNextBatch, so connectors receive the resolved predicate per batch without needing to hold a DynamicFilter and manage wait/timeout logic themselves.

What if connector doesn't use the predicate (unconditionally, or under some circumstances)?
We add wait, and wait in vein?

chenjian2664:
> What if connector doesn't use the predicate (unconditionally, or under some circumstances)?
We add wait, and wait in vein?

If a connector does not support DF, the engine checks whether the DF is waitable. The behavior is unchanged, so it will not wait in this case.

See:
https://github.com/trinodb/trino/pull/29206/changes/ee0b2d7832eee73630e501b8137f9339d38e3b29#diff-23fa9063c012e48a3511321e20d060a3e0009de6aa1d57b5424cbdc217ef5ecdR86-R87
Connectors can also specify the maximum wait time for the engine.

findepi:
> the engine checks whether the DF is waitable.

isn't that a property of DF itself and its producer? and not the DF consumer?

findepi:
The PR description says that wait control is moved from the connector side to the engine side. That's clear now. Why is that a good change to make? What problem is this solving?

raunaqmorarka:
> The PR description says that wait control is moved from the connector side to the engine side. That's clear now. Why is that a good change to make? What problem is this solving?

It removes duplicated logic from connectors. There's no reason why every connector should be hand rolling roughly similar waiting loop on DF. We did it this way in the beginning because connector was the only one benefitting from DFs in the past and it started from a small number of connectors. Since then the engine has gained dynamic row filtering and many more connectors have added similar waiting loop on DF.
The existing contract between engine and connectors for DFs is also overly complex because the `io.trino.spi.connector.DynamicFilter` API forces the connector to deal with an object that can update under the covers any time and reason about waiting on it. We want to eventually move `io.trino.spi.connector.DynamicFilter` out of SPI.

>> the engine checks whether the DF is waitable.

> isn't that a property of DF itself and its producer? and not the DF consumer?

Whether the DF is awaitable is a property of the query plan, but how much and whether to wait on it is a decision that requires input from both engine and connector. The engine should be involved because it knows estimates about the size of build side and probe side and it applies dynamic row filtering on workers regardless of what the connector is doing. Connector should also be involved because engine today is not able to reason about the latency and cost of a scan in JDBC vs a scan in data lake etc.
A previous attempt to add an optimizer rule in the engine to wait on DFs https://github.com/trinodb/trino/pull/22527 failed because @martint generally did not like the previous approach of engine informing connectors to wait through `io.trino.spi.connector.DynamicFilter#getPreferredDynamicFilterTimeout` + existing wait configs of the connector and wanted engine to take over this part with connector still having the possibility to give an input to the waiting decision.
Current simplification of the SPI has been discussed offline with @martint before this work began.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/PartitionMatchSupplier.java
```diff
@@ -30,16 +32,16 @@
...
     private final List<HiveColumnHandle> partitionColumns;
 
     @Nullable
     private volatile Boolean finalResult; // value is null until the dynamic filter no longer needs to be evaluated
 
-    private PartitionMatchSupplier(DynamicFilter dynamicFilter, HivePartition hivePartition, List<HiveColumnHandle> partitionColumns)
+    private PartitionMatchSupplier(DynamicFilterState dynamicFilterState, HivePartition hivePartition, List<HiveColumnHandle> partitionColumns)
     {
-        this.dynamicFilter = requireNonNull(dynamicFilter, "dynamicFilter is null");
+        this.dynamicFilterState = requireNonNull(dynamicFilterState, "dynamicFilterSupplier is null");
```
raunaqmorarka: Copy-paste leftover: error message says `"dynamicFilterSupplier is null"` but the parameter is `dynamicFilterState`.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/util/PartitionMatchSupplier.java
```diff
@@ -57,9 +59,10 @@ public boolean getAsBoolean()
 
     private boolean evaluateCurrentDynamicFilter()
     {
-        // Must be checked before calling dynamicFilter.getCurrentPredicate()
+        ConnectorDynamicFilter dynamicFilter = dynamicFilterState.get();
+        // Must be checked before calling dynamicFilter.currentPredicate()
```
raunaqmorarka: This comment was load-bearing when `dynamicFilter` was a live mutable `DynamicFilter` (reads could race between `isComplete` and `getCurrentPredicate`). Now that `dynamicFilter` is a `ConnectorDynamicFilter` record (immutable snapshot), both accessors are just field reads on the same object. The ordering constraint no longer exists and the comment is misleading. Suggest removing it.

plugin/trino-hive/src/test/java/io/trino/plugin/hive/TestBackgroundHiveSplitLoader.java
```diff
@@ -323,39 +323,8 @@ public void testIncompleteDynamicFilterTimeout()
...
-
-                        @Override
-                        public TupleDomain<ColumnHandle> getCurrentPredicate()
-                        {
-                            return TupleDomain.all();
-                        }
-                    },
-                    new Duration(1, SECONDS));
+                    ImmutableSet.of(),
+                    DynamicFilterState.completeState());
```
raunaqmorarka: The test is named `testIncompleteDynamicFilterTimeout` but passes `DynamicFilterState.completeState()`, which represents a fully resolved filter. It no longer exercises any incomplete/timeout behavior. The `isBlocked` future on line 323 is also unused (leftover from the old test).

Since the wait logic moved to the engine (`ConnectorAwareSplitSource`), this test should either be renamed to reflect what it actually tests now, or replaced with an engine-level test that exercises the new wait-timeout path.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/DynamicFilterState.java
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    public boolean isReady()
+    {
+        return firstArrival.isDone();
+    }
+
+    public ConnectorDynamicFilter awaitAndGet()
+    {
+        firstArrival.join();
```
raunaqmorarka: `awaitAndGet()` has no callers in the codebase — only `isReady()` + `await()` are used by `BackgroundHiveSplitLoader`, and `get()` is used by `PartitionMatchSupplier`. Consider removing it. It's also a footgun: `firstArrival.join()` blocks indefinitely and doesn't respond to thread interruption, so if query cancellation closes the split source before the first `getNextBatch` arrives, any thread calling this would leak.

chenjian2664: The https://github.com/trinodb/trino/pull/29206/changes#diff-51c6b18c6fabde04e1a54f67ef16e45ec2a48194e19f6a264f90532a25437b3dR335 is using the `awaitAndGet`. 
otherwise we put, `await()` a single line then `get()`, I am not strong about it, I can remove it if you think it should

plugin/trino-hive/src/test/java/io/trino/plugin/hive/TestBackgroundHiveSplitLoader.java
chenjian2664: renamed the test to just simply test the complete filter.

chenjian2664:
@raunaqmorarka comments AC

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
```diff
@@ -14,9 +14,68 @@
...
+                return false;
+            }
+
+            @Override
+            public TupleDomain<ColumnHandle> getCurrentPredicate()
+            {
+                return TupleDomain.all();
+            }
+        }, constraint);
+    }
```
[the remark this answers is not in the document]
raunaqmorarka: @chenjian2664 this one is a genuine problem. Please look into feasibility of "engine needs to dispatch to the deprecated override when it is implemented".
I'm thinking that if its not possible to provide genuine backward compatibility, then we might as well skip the deprecation and do an outright removal. This way connector implementers will be forced to make changes to adjust to new API rather than silently missing DF pruning because the default bridge is hiding the issue.

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcDynamicFilteringSplitManager.java
```diff
@@ -101,44 +96,34 @@ private class DynamicFilteringSplitSource
...
+        public CompletableFuture<ConnectorSplitBatch> getNextBatch(int maxSize, ConnectorDynamicFilter dynamicFilter)
+        {
             log.debug(
-                    "Enumerating splits (query %s, table: %s, waiting time: %s, awaitable: %s, completed: %s)",
+                    "Enumerating splits (query %s, table: %s, waiting time: %s, completed: %s)",
                     session.getQueryId(),
                     table,
-                    waitingTime,
-                    dynamicFilter.isAwaitable(),
+                    succinctNanos(System.nanoTime() - startNanos),
```
raunaqmorarka: I don't think it make sense to log "waiting time" here anymore as it already happened on the engine side

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestJdbcDynamicFilteringSplitManager.java
```diff
@@ -52,57 +52,41 @@ public class TestJdbcDynamicFilteringSplitManager
...
 
-        // verify that getNextBatch() future completes after a timeout
-        CompletableFuture<?> future = splitSource.getNextBatch(100);
-        assertThat(future.isDone()).isFalse();
-        future.get(10, SECONDS);
+        // engine delivers predicate snapshot; split source uses it to build the delegate
+        ConnectorDynamicFilter dynamicFilter = new ConnectorDynamicFilter(TupleDomain.all(), true);
+        ConnectorSplitSource.ConnectorSplitBatch batch = splitSource.getNextBatch(100, dynamicFilter).get();
+        assertThat(batch.isNoMoreSplits()).isTrue();
         assertThat(splitSource.isFinished()).isTrue();
```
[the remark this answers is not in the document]
raunaqmorarka: @chenjian2664 please look into this one as well

plugin/trino-hive/src/main/java/io/trino/plugin/hive/DynamicFilterState.java
```diff
@@ -0,0 +1,69 @@
...
+        latest = requireNonNull(filter, "filter is null");
+        firstArrival.complete(null);
+    }
+
+    public ConnectorDynamicFilter get()
+    {
+        return latest;
+    }
+
+    public CompletableFuture<Void> await()
```
raunaqmorarka: await -> isBlocked

plugin/trino-hive/src/main/java/io/trino/plugin/hive/DynamicFilterState.java
raunaqmorarka: Let's remove it and replace it with a get on the future with a timeout (20s just for safety, infinite wait is risky)

plugin/trino-hive/src/main/java/io/trino/plugin/hive/DynamicFilterState.java
```diff
@@ -0,0 +1,69 @@
...
+        return firstArrival.isDone();
+    }
+
+    public ConnectorDynamicFilter awaitAndGet()
+    {
+        firstArrival.join();
+        return latest;
+    }
+
+    static DynamicFilterState completeState()
```
raunaqmorarka: completeState -> completedState

plugin/trino-hive/src/main/java/io/trino/plugin/hive/HiveSplitManager.java
```diff
@@ -331,8 +332,7 @@ private Iterator<HivePartitionMetadata> getPartitionMetadata(
 
         Iterator<List<HivePartition>> partitionNameBatches = partitionExponentially(hivePartitions, minPartitionBatchSize, maxPartitionBatchSize);
         Iterator<List<HivePartitionMetadata>> partitionBatches = transform(partitionNameBatches, partitionBatch -> {
-            // Use dynamic filters to reduce the partitions listed by getPartitionsByNames
```
raunaqmorarka, line 336: don't remove the comment

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -233,36 +225,28 @@ public IcebergSplitSource(
...
-                .collect(toImmutableSet());
         this.fileModifiedTimeDomain = getFileModifiedTimeDomain(tableHandle.getEnforcedPredicate());
         this.splitAffinityProvider = requireNonNull(splitAffinityProvider, "splitAffinityProvider is null");
         this.metricsReporter = requireNonNull(metricsReporter, "metricsReporter is null");
         this.executor = requireNonNull(executor, "executor is null");
+        this.predicatedColumnIds = Stream.concat(
+                        tableHandle.getUnenforcedPredicate().getDomains().orElse(ImmutableMap.of()).keySet().stream(),
+                        dynamicFilterColumns.stream().map(IcebergColumnHandle.class::cast))
+                .map(IcebergColumnHandle::getId)
+                .collect(toImmutableSet());
```
raunaqmorarka, line 239: This doesn't need to move, can stay where it was

plugin/trino-memory/src/main/java/io/trino/plugin/memory/MemorySplitManager.java
```diff
@@ -79,65 +75,20 @@ public ConnectorSplitSource getSplits(
...
-        @Override
-        public String toString()
-        {
-            return toStringHelper(this)
-                    .add("delay", delay)
-                    .add("delegate", delegate)
-                    .toString();
+        if (enableLazyDynamicFiltering && !dynamicFilterColumns.isEmpty()) {
+            // In fault-tolerant execution, the probe stage must not be scheduled before the build stage
+            // (which collects dynamic filters), otherwise it occupies all available slots and deadlocks.
```
raunaqmorarka: Retain the original code comment

testing/trino-tests/src/test/java/io/trino/execution/AbstractTestCoordinatorDynamicFiltering.java
```diff
@@ -535,37 +530,35 @@ public ConnectorSplitSource getSplits(
...
-                            if (blocked.isDone()) {
-                                splitProduced.set(true);
-                                return completedFuture(new ConnectorSplitBatch(ImmutableList.of(createRemoteSplit()), isFinished()));
+                        @Override
+                        public CompletableFuture<ConnectorSplitBatch> getNextBatch(int maxSize, ConnectorDynamicFilter dynamicFilter)
+                        {
+                            if (!dynamicFilter.isComplete()) {
+                                // Filter not yet complete — return empty batch so the engine retries
+                                // with a fresh predicate snapshot. In standard mode the engine already
+                                // waited; in FTE mode the engine retries until isComplete is true.
```
raunaqmorarka: Can you clarify what is the behaviour you observed in FTE here ?
We need the waiting in ConnectorAwareSplitSource to work effectively for FTE mode as well

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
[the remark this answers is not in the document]
chenjian2664: Make sense, it seems we should remove the old api entirely - I'll do it

plugin/trino-hive/src/main/java/io/trino/plugin/hive/DynamicFilterState.java
chenjian2664: overload with `get(long, TimeUnit)` for exception handling, and replace the logic that waits for a fixed duration

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
chenjian2664: It was added back at 1adb0290780a4eef3b32d0d081f93db7d594486b, will squash into(revert the changes) the original commit for Iceberg.

testing/trino-tests/src/test/java/io/trino/execution/AbstractTestCoordinatorDynamicFiltering.java
chenjian2664: Removed the misleading comment. The behaviour is the same as before — `getNextBatch` checks `!isComplete` (equivalent to the old `!blocked.isDone()`) and returns an empty batch if the filter is not yet ready. The only difference is we now capture the current predicate snapshot via `capturedPredicate.set(dynamicFilter.currentPredicate())`, which is exactly what this PR is designed to deliver.

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestJdbcDynamicFilteringSplitManager.java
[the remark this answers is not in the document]
chenjian2664: Fixed. The test now uses a capturing delegate that records the `ConnectorTableHandle` received by `getSplits` and asserts its constraint equals the DF predicate — proving `intersectedWithConstraint` was called correctly. Note that in JDBC the predicate is pushed down into the table handle (SQL WHERE clause), not used to post-filter splits, so the assertion is on the captured constraint rather than split count.

Also added an optimisation: if the intersected constraint is `TupleDomain.none()`, `getDelegateSplitSource` now short-circuits and returns an empty finished split source immediately, avoiding a database connection entirely. Covered by a second new test that asserts the batch is empty, `isNoMoreSplits()` is true, and the delegate was never called.

@coderabbitai

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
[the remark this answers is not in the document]
chenjian2664: removed the old method directly, and make the new method non-default, required every existing or new connector to implement it. @coderabbitai

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestJdbcDynamicFilteringSplitManager.java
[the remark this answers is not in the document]
chenjian2664: @coderabbitai review again

chenjian2664:
Squashed old fixup, and move the fixup commit close to it target commit.(no changes)

chenjian2664:
@raunaqmorarka I split the migration each connectors to new spi as the independent commit, let me know if you don't think it is necessary. The ci is all fixed, please take a look

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorDynamicFilter.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+package io.trino.spi.connector;
+
+import io.trino.spi.predicate.TupleDomain;
+
+/**
+ * A per-batch snapshot of the engine's dynamic filter state, passed to
+ * {@link ConnectorSplitSource#getNextBatch(int, ConnectorDynamicFilter)}.
+ *
+ * @since 481
```
raunaqmorarka: we don't need `@since` for new SPI

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
```diff
@@ -15,8 +15,30 @@
...
 {
+    /**
+     * Returns splits for a table scan.
+     * <p>
+     * {@code dynamicFilterColumns} is the static set of columns the dynamic filter will cover.
+     * Connectors may use this to make planning-time decisions (e.g. which column statistics to
+     * read). The per-batch resolved predicate arrives separately via
+     * {@link ConnectorSplitSource#getNextBatch(int, ConnectorDynamicFilter)}.
+     *
+     * @since 481
```
raunaqmorarka: drop

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -30,6 +31,24 @@
...
+     * <p>
+     * {@code dynamicFilter} contains a snapshot of the engine's dynamic filter state captured
+     * immediately before this call. The engine waits up to
+     * {@link #getRequestedDynamicFilterWaitTimeoutMillis()} before the <em>first</em> call;
+     * subsequent calls receive the current state without additional waiting.
+     * For connectors that do not use dynamic filtering,
+     * {@link ConnectorDynamicFilter#currentPredicate()} is {@link TupleDomain#all()} and
+     * {@link ConnectorDynamicFilter#isComplete()} is {@code true}.
+     *
+     * @since 481
```
raunaqmorarka: drop

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -30,6 +31,24 @@
...
+    /**
+     * Returns the next batch of splits.
+     * <p>
+     * {@code dynamicFilter} contains a snapshot of the engine's dynamic filter state captured
+     * immediately before this call. The engine waits up to
+     * {@link #getRequestedDynamicFilterWaitTimeoutMillis()} before the <em>first</em> call;
+     * subsequent calls receive the current state without additional waiting.
+     * For connectors that do not use dynamic filtering,
+     * {@link ConnectorDynamicFilter#currentPredicate()} is {@link TupleDomain#all()} and
+     * {@link ConnectorDynamicFilter#isComplete()} is {@code true}.
```
raunaqmorarka: skip this line

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -53,6 +72,20 @@ default Optional<List<Object>> getTableExecuteSplitsInfo()
         return Optional.empty();
     }
 
+    /**
+     * Returns the maximum time in milliseconds the engine should wait for dynamic filters to be
```
raunaqmorarka: The contract here is that this is a hint to the engine, the engine is still free to make it's own decision based on the input from connector here and potentially its own reasoning based on some optimizer rule.
We shouldn't commit to exactly following the timeout here in the javadoc

plugin/trino-cassandra/src/test/java/io/trino/plugin/cassandra/TestCassandraSplitManager.java
```diff
@@ -96,8 +97,8 @@ PRIMARY KEY(partition_key, clustering_key))
 
         CassandraTableHandle tableHandle = new CassandraTableHandle(
                 new CassandraNamedRelationHandle(KEYSPACE, tableName, Optional.of(partitions.build()), ""));
-        try (ConnectorSplitSource splitSource = splitManager.getSplits(null, null, tableHandle, DynamicFilter.EMPTY, null)) {
-            List<ConnectorSplit> splits = splitSource.getNextBatch(100).get().getSplits();
+        try (ConnectorSplitSource splitSource = splitManager.getSplits(null, null, tableHandle, ImmutableSet.of(), null)) {
+            List<ConnectorSplit> splits = splitSource.getNextBatch(100, new ConnectorDynamicFilter(TupleDomain.all(), true)).get().getSplits();
```
raunaqmorarka: ConnectorDynamicFilter.EMPTY ?

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -234,36 +226,28 @@ public IcebergSplitSource(
...
```
raunaqmorarka, line 239: This code block doesn't need to move

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergSplitSource.java
```diff
@@ -161,7 +157,6 @@ public class IcebergSplitSource
     private final Domain pathDomain;
     private final Domain fileModifiedTimeDomain;
     private final OptionalLong limit;
-    private final Set<Integer> predicatedColumnIds;
```
raunaqmorarka, line 168: This doesn't need to move

plugin/trino-base-jdbc/src/test/java/io/trino/plugin/jdbc/TestingDatabase.java
```diff
@@ -14,6 +14,7 @@
 package io.trino.plugin.jdbc;
```
raunaqmorarka, line 15: This should squash with the next commit for jdbc connectors

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DynamicFilteringJdbcSplitSource.java
```diff
@@ -37,25 +39,25 @@ public class DynamicFilteringJdbcSplitSource
...
     private final JdbcTableHandle tableHandle;
+    private final long dynamicFilterWaitTimeoutMillis;
 
-    DynamicFilteringJdbcSplitSource(ConnectorSplitSource delegateSplitSource, DynamicFilter dynamicFilter, JdbcTableHandle tableHandle)
+    DynamicFilteringJdbcSplitSource(ConnectorSession session, ConnectorSplitSource delegateSplitSource, JdbcTableHandle tableHandle)
     {
         this.delegateSplitSource = requireNonNull(delegateSplitSource, "delegateSplitSource is null");
-        this.dynamicFilter = requireNonNull(dynamicFilter, "dynamicFilter is null");
         this.tableHandle = requireNonNull(tableHandle, "tableHandle is null");
+        this.dynamicFilterWaitTimeoutMillis = getDynamicFilteringWaitTimeout(session).toMillis();
```
raunaqmorarka: I think the waiting was already requested by `DynamicFilteringSplitSource`, does this one still need to ask for it ?

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DynamicFilteringJdbcSplitSource.java
raunaqmorarka, line 51: There seems to be a lot of overlap in DynamicFilteringSplitSource and this class now, earlier that other class had the extra responsibility of doing the waiting.
Try consolidating to just this class now

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: updated, please help to take a look

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DynamicFilteringJdbcSplitSource.java
chenjian2664: You are right, the timeout config already convered by `DynamicFilteringSplitSource`

chenjian2664:
@raunaqmorarka comments addressed, the Iceberg fixup I squashed, others left as the fixup

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/DynamicFilteringJdbcSplitSource.java
```diff
@@ -26,46 +37,65 @@
...
                 .thenApply(batch -> {
                     TupleDomain<JdbcColumnHandle> dynamicFilterPredicate = dynamicFilter.currentPredicate()
                             .transformKeys(JdbcColumnHandle.class::cast);
                     return new ConnectorSplitBatch(
                             batch.getSplits().stream()
-                                    // attach dynamic filter constraint to JdbcSplit
                                     .map(split -> {
                                         JdbcSplit jdbcSplit = (JdbcSplit) split;
-                                        // If split was a subclass of JdbcSplit, there would be additional information
-                                        // that we would need to pass further on.
```
raunaqmorarka, line 66: keep these comments

raunaqmorarka (approved):

electrum:
Since we're doing all the work to migrate everything to a new API, we should change the new API to return the splits directly, rather than going through the obsolete `ConnectorSplitBatch`. There's no reason to have a separate `isNoMoreSplits()` which is redundant and confusingly overlaps with `isFinished()`.

electrum:
For easier review, please squash all the commits that have trivial changes into a single commit. For example, Thrift, Blackhole, TPC-H, etc.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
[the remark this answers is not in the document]
findepi: ( pending discussion in maintainers channel )

chenjian2664:
@electrum PTAL

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorDynamicFilter.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package io.trino.spi.connector;
+
+import io.trino.spi.predicate.TupleDomain;
+
+/**
+ * A per-batch snapshot of the engine's dynamic filter state, passed to
+ * {@link ConnectorSplitSource#getNextBatch(int, ConnectorDynamicFilter)}.
+ */
+public record ConnectorDynamicFilter(TupleDomain<ColumnHandle> currentPredicate, boolean isComplete)
```
findepi: We already have `io.trino.spi.connector.DynamicFilter` being "dynamic filter passed to Connector"
this new class has different name (obviously), but   name difference doesn't even allude to semantics difference

rename

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
```diff
@@ -15,17 +15,24 @@
...
+    ConnectorSplitSource getSplits(
             ConnectorTransactionHandle transaction,
             ConnectorSession session,
             ConnectorTableHandle table,
-            DynamicFilter dynamicFilter,
-            Constraint constraint)
-    {
-        throw new UnsupportedOperationException();
-    }
+            Set<ColumnHandle> dynamicFilterColumns,
```
findepi, line 34: Why provide dynamicFilterColumns without providing dynamicFilter?

Or bigger Q, why not provide dynamicFilter to a connector that _wants_ to use it?

the PR desc explains why add DF to next-batch

> Move dynamic filter waiting logic from individual connectors into the engine layer by adding a TupleDomain dynamicFilterPredicate parameter to ConnectorSplitSource#getNextBatch, so connectors receive the resolved predicate per batch without needing to hold a DynamicFilter and manage wait/timeout logic themselves.


but i can't find explanation why we should stop providing DF to getSplits

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -71,6 +72,20 @@ default Optional<List<Object>> getTableExecuteSplitsInfo()
         return Optional.empty();
     }
 
+    /**
+     * Hints to the engine the maximum time in milliseconds to wait for dynamic filters to be
+     * collected before the first call to {@link #getNextBatch(int, ConnectorDynamicFilter)}.
+     * The engine may wait up to this timeout, or until the dynamic filter is fully resolved,
+     * whichever comes first. Return {@code 0} to disable waiting (the default).
```
findepi: "Return {@code 0} to disable waiting (the default)." is pure redundancy

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -71,6 +72,20 @@ default Optional<List<Object>> getTableExecuteSplitsInfo()
         return Optional.empty();
     }
 
+    /**
+     * Hints to the engine the maximum time in milliseconds to wait for dynamic filters to be
+     * collected before the first call to {@link #getNextBatch(int, ConnectorDynamicFilter)}.
+     * The engine may wait up to this timeout, or until the dynamic filter is fully resolved,
```
findepi, line 78: "or until the dynamic filter is fully resolved" is pure redundancy

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -71,6 +72,20 @@ default Optional<List<Object>> getTableExecuteSplitsInfo()
...
+     * whichever comes first. Return {@code 0} to disable waiting (the default).
+     * <p>
+     * The engine reads this value once when the split source is constructed; it is treated as
+     * fixed for the lifetime of the source (per table scan).
+     */
+    default long getRequestedDynamicFilterWaitTimeoutMillis()
```
findepi, line 84: The idea behind the change is to free connectors from having to think about DF waits... but they still need to think about DF waits. Are we sure we really gained what we wanted to gain?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -81,26 +96,4 @@ default Metrics getMetrics()
     {
         return Metrics.EMPTY;
     }
-
-    class ConnectorSplitBatch
-    {
-        private final List<ConnectorSplit> splits;
-        private final boolean noMoreSplits;
```
findepi, line 88: Why is this class removed? I understand the new API doesn't have noMoreSplits flag and the wrapper class is not needed, but this seems orthogonal to DFs.
It it obvious that `noMoreSplits` was not useful for anything?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
```diff
@@ -37,21 +35,24 @@ public interface ConnectorSplitSource
...
      * a normal batch result.
+     * {@code dynamicFilter} contains a snapshot of the engine's dynamic filter state captured
+     * immediately before this call. The engine waits up to
+     * {@link #getRequestedDynamicFilterWaitTimeoutMillis()} before the <em>first</em> call;
+     * subsequent calls receive the current state without additional waiting.
+     * <p>
+     * When the returned list is empty and {@link #isFinished()} returns {@code false}, the caller
+     * should retry. No more batches will be produced once {@link #isFinished()} returns {@code true}.
      */
-    default CompletableFuture<ConnectorSplitBatch> getNextBatch(int maxSize)
```
findepi, line 51: What prevents maintaining backward compatibility for getNextBatch?

or are you saying backward compatibility is passé?

findepi (commented):
(just SPI)

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorDynamicFilter.java
chenjian2664: Any suggestions on the naming?
It essentially acts as a "snapshot" of the current state, I'm thinking of going with "DynamicFilterSnapshot"

Renamed, let me know if you have different naming idea

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
chenjian2664: If DynamicFilter stays in `getSplits`, connectors can still implement their own wait loop there, which is exactly the pattern this PR is trying to eliminate, that is leaving it in place would make the `getNextBatch` change optional rather than structural, well-behaved connectors would adopt it, but others could keep the old approach indefinitely.

At getting splits time the predicate isn't stable yet (the DF hasn't been collected from build side operators). So `dynamicFilter.getCurrentPredicate()` would return `TupleDomain.all()` for most of the waiting period anyway, the only useful thing a connector can do there is wait, which is the behavior we're moving to the engine. `Set<ColumnHandle>` gives connectors the one thing they legitimately need at plan time: which columns will be filtered (e.g. to decide which partition statistics to load), without exposing the live interface that tempts them to wait on it.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: removed

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: removed

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: Yes. Before this PR, each connector that wanted DF-aware split generation had to implement the full async wait loop: hold a DynamicFilter reference, call `isBlocked().get()` with a timeout, handle early resolution, then call `getCurrentPredicate()`. That logic was duplicated across Hive, Iceberg, Delta Lake, and others, all slightly differently, all error-prone.
Now engine owns that mechanism entirely. `getRequestedDynamicFilterWaitTimeoutMillis()` is just a policy expression, return a number from config, nothing more. Most connectors don't override it at all (default 0) and still benefit from DF filtering on subsequent batches once the filter is ready.
The connectors that do override it (Hive, Iceberg) still need to say how long they're willing to wait, but that knowledge is inherently connector-specific.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: Same reasoning as the earlier getSplits [discussion](https://github.com/trinodb/trino/pull/29206#discussion_r3239011032) a backward-compat bridge would be behavior-incompatible, not just source-compatible.
A bridge from `getNextBatch(int)` to the new `getNextBatch(int, ConnectorDynamicFilter)` would have to pass a dummy `ConnectorDynamicFilter`, silently feeding connectors `TupleDomain.all()` on every batch. Third-party connectors that override the old signature expecting real predicate values would lose DF filtering with no compile error or warning. On top of that, Keeping a bridge that returns a deleted type isn't viable.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: Following the [suggestion](https://github.com/trinodb/trino/pull/29206#issuecomment-4623860519) — the return type cleanup felt like a natural opportunity while touching `ConnectorSplitSource`. 
I agree it may be too aggressive to bundle both the SPI behavioral change and the `ConnectorSplitBatch` removal in the same commit. I can split them, one commit for the new `getNextBatch` signature and one for the return type refactoring, to keep the history easier to review, WDYT?

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
```diff
@@ -72,14 +84,23 @@ public CatalogHandle getCatalogHandle()
     public ListenableFuture<SplitBatch> getNextBatch(int maxSize)
     {
         checkState(source != null, "Already finished or closed");
```
marcinsbd, line 86: is it feasible that concurrent thread calls close() between checking null state and calling source.getNextBatch() ?
Shouldn't we capture source in local variable?
```
ConnectorSplitSource currentSource = source; 
checkState(currentSource != null, "Already finished or closed");
....
currentSource.getNextBatch(maxSize, new ConnectorDynamicFilter(currentPredicate, isComplete)));
```

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
```diff
@@ -72,14 +84,23 @@ public CatalogHandle getCatalogHandle()
...
-        ListenableFuture<ConnectorSplitBatch> nextBatch = toListenableFuture(source.getNextBatch(maxSize));
-        return Futures.transform(nextBatch, splitBatch -> {
-            List<ConnectorSplit> connectorSplits = splitBatch.getSplits();
+        long timeLeft = dynamicFilteringWaitTimeoutMillis - dynamicFilterWaitStopwatch.elapsed(MILLISECONDS);
```
marcinsbd, line 87: Could we get here negative value?  If scheduler takes more time to make first call to this method than the wait timeout?
Shouldn't we initialize dynamicFilterWaitStopwatch rather here lazily than in the ctor

```
if (dynamicFilterWaitStopwatch == null) {
            dynamicFilterWaitStopwatch = Stopwatch.createStarted();
}
```

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
findepi: > If DynamicFilter stays in `getSplits`, connectors can still implement their own wait loop there, which is exactly the pattern this PR is trying to eliminate

the PR tries to forbid this (make impossible),
or tries to make it unnecessary?

findepi: The PR title is: `Pass dynamic filter predicate to ConnectorSplitSource#getNextBatch`
This sounds like "hey connector, we got new options for you"
it does not sound like "Move dynamic filter waits to engine"

Which one was your intent?

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
```diff
@@ -72,14 +84,23 @@ public CatalogHandle getCatalogHandle()
...
+        if (dynamicFilter.isAwaitable() && timeLeft > 0) {
+            CompletableFuture<SplitBatch> emptyBatch = dynamicFilter.isBlocked()
+                    .thenApply(_ -> new SplitBatch(ImmutableList.of(), false))
+                    .completeOnTimeout(new SplitBatch(ImmutableList.of(), false), timeLeft, MILLISECONDS);
```
marcinsbd, line 91: Are we ok with the common executor here?

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorDynamicFilter.java
findepi: DynamicFilterSnapshot SGTM, thanks

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
findepi: > as the earlier getSplits [discussion](https://github.com/trinodb/trino/pull/29206#discussion_r3239011032) a backward-compat bridge would be behavior-incompatible, not just source-compatible.

behavior-incompatible backward-compat is bad, agreed

behavior-incompatibility is direct consequence of the design to make the change backwards incompatible (https://github.com/trinodb/trino/pull/29206#discussion_r3377520871), not something inherent to the change.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
findepi: I think you cannot cleanup in separate commit, because you can only cleanup when you introduce a new API method.
The right place to capture the raionale for the change is the commit message mentioning why ConnectorSplitBatch is not used for the new method.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitManager.java
chenjian2664: Thanks for the question!
I've updated the pr title - that's I am moving the dynamic filter wait to engine

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: Actually we could first do the refactoring changing return type - removing `ConnectorSplitBatch` then applying the refactoring for moving DynamicFilter wait to engine, it's clean but looks like overkill since a bit.
I am going to update the commit message as you suggested.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
chenjian2664: The goal of this PR is to move DF waiting logic entirely into the engine. Keeping `DynamicFilter` in `getSplits` - even for backward compatibility, would allow connectors to still implement their own wait loop there, which directly contradicts that goal. Passing `DynamicFilterSnapshot` instead of `DynamicFilter` is a deliberate choice, it exposes only the resolved predicate, not the wait interface, since `DynamicFilter` in `getSplits` is only useful for waiting, not for reading. At `getSplits` time the build side hasn't executed yet, so `getCurrentPredicate()` returns `TupleDomain.all()`. The predicate doesn't become meaningful until after the engine has waited for DF collection. The only thing a connector can usefully do with `DynamicFilter` there is call `isBlocked().get()`, which is exactly what this PR moves to the engine.
`Set<ColumnHandle>` gives connectors the one thing that IS known at planning time (which columns will be filtered) without exposing the wait interface.

Even if we kept `DynamicFilter` in `getSplits` for backward compatibility, we'd end up with two concurrent behaviors: migrated connectors rely entirely on the engine's wait loop in `ConnectorAwareSplitSource`, while un-migrated connectors still manage their own wait in `getSplits`. For migrated connectors the behavior is clean and predictable. For un-migrated connectors it's unclear who dominates - the engine waits before getNextBatch, the connector waits in `getSplits`, and the interaction between the two timeouts is undefined. Having a single well-defined place where DF waiting happens is the goal, a soft migration path permanently undermines it.

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
chenjian2664: This is a pre-existing condition, not introduced by this PR. The class is annotated `@NotThreadSafe` and the class comment already acknowledges that `close()` may be called concurrently while a batch request is outstanding, the caller i.e, `SourcePartitionedScheduler` is responsible for not calling `getNextBatch` after or when `closing`.

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
chenjian2664: The negative value case is handled correctly. The condition `timeLeft > 0` guards the wait block, if timeLeft is negative (the timeout already elapsed before the first call), the engine skips waiting entirely and proceeds straight to `getNextBatch` with the current predicate, we don't wait if time has already run out.
Initializing the stopwatch lazily on the first `getNextBatch` call would change the semantic, the timeout would always start fresh when batches are first requested, regardless of how long construction took. The current design intentionally starts the clock at construction time so the total wall-clock wait is bounded by `getRequestedDynamicFilterWaitTimeoutMillis`, not just the time spent inside `getNextBatch` calls.

core/trino-main/src/main/java/io/trino/split/ConnectorAwareSplitSource.java
chenjian2664: Do you have suggestions, or do you mean we should using `thenApplyAysnc(..., directExector())`? if so I am thinking the callback is trivial, running the completing thread is acceptable.

core/trino-spi/src/main/java/io/trino/spi/connector/ConnectorSplitSource.java
findepi: > The goal of this PR is to move DF waiting logic entirely into the engine.

That makes sense (that is, at least is coherent). I was confused by your earlier message expressing concerns about backwards incompatibility. The backwards incompatibility is here by design. It should be surfaced (PR title 👍 ), that's it.



> Even if we kept `DynamicFilter` in `getSplits` for backward compatibility, we'd end up with two concurrent behaviors: migrated connectors rely entirely on the engine's wait loop in `ConnectorAwareSplitSource`, while un-migrated connectors still manage their own wait in `getSplits`. 

I don't see this as a big problem.

>  For un-migrated connectors it's unclear who dominates - the engine waits before getNextBatch, the connector waits in `getSplits`, and the interaction between the two timeouts is undefined.

max of two timeouts, what's unclear about that?

findepi: btw Are you planning on removing other `DynamicFilter` direct usages from SPI?

findepi:
Given that this is intentionally breaking change, with no backward compatibility layer, I don't fully follow the commit structure. 

```
...
* 112d5a28ae7 - Use ConnectorDynamicFilter in Jdbc connectors
* 2b9c54eee94 - Use ConnectorDynamicFilter in LakeHouse
* f38ec9e9e57 - Use ConnectorDynamicFilter in Hive
* 90428f194e1 - Use ConnectorDynamicFilter in Iceberg
* 11207ba90a0 - Use ConnectorDynamicFilter in Delta Lake
* 2525387a942 - Migrate Memory connector to new split SPI
* 83c8197aa81 - Migrate built-in connectors to new split SPI
....
```

Are you sure all tests pass at each commit?

chenjian2664:
> Given that this is intentionally breaking change, with no backward compatibility layer, I don't fully follow the commit structure.
> 
> ```
> ...
> * 112d5a28ae7 - Use ConnectorDynamicFilter in Jdbc connectors
> * 2b9c54eee94 - Use ConnectorDynamicFilter in LakeHouse
> * f38ec9e9e57 - Use ConnectorDynamicFilter in Hive
> * 90428f194e1 - Use ConnectorDynamicFilter in Iceberg
> * 11207ba90a0 - Use ConnectorDynamicFilter in Delta Lake
> * 2525387a942 - Migrate Memory connector to new split SPI
> * 83c8197aa81 - Migrate built-in connectors to new split SPI
> ....
> ```
> 
> Are you sure all tests pass at each commit?

Let me check

chenjian2664:
@findepi I've tested, now the plan for the final shape of commit will be 3+ commits:

* Introduce commit — purely additive, zero behavioral change:
  - Adds `DynamicFilterSnapshot`, the new `getNextBatch(int, DynamicFilterSnapshot)` and `getSplits(Set<ColumnHandle>)` signatures as defaults with backward-compat bridges
  - SplitManager is unchanged — still calls getSplits(DynamicFilter), no `ConnectorAwareSplitSource` yet

* Migrate simple connectors — one commit per connector (JMX, TPC-H/DS, Kafka, Thrift, BlackHole, etc.):
  - These connectors have no DF waiting logic — migration is a pure signature change
  - The engine still calls the old API; bridges route to the new overrides
 
* Remove commit — the single behavioral change:
  - Migrates the DF-aware connectors (Hive, Iceberg, Delta Lake, JDBC, MongoDB) in the same commit as the engine switch
  - Introduces to switches SplitManager to `getSplits(Set<ColumnHandle>),` removes old bridges and ConnectorSplitBatch
  - These connectors are migrated and the engine switch lands atomically, so they immediately receive real `DynamicFilterSnapshot` values 

For now I will let the commit as is - just for reviewing, once we agree with the implementation, I will squash them

findepi:
> Migrate simple connectors — one commit per connector (JMX, TPC-H/DS, Kafka, Thrift, BlackHole, etc.):
> 
> * These connectors have no DF waiting logic — migration is a pure signature change
> * The engine still calls the old API; bridges route to the new overrides

At this (interim) state we have effectively no waiting whatsoever, right?
assuming there is a test that needs DF to be effective for it to pass, would such test(s) pass?

chenjian2664:
@findepi right. It should still pass, the waiting logic stays the same at this step

Issue #4131, opened 2020-06-22
Admit EXPLAIN queries when required worker count not satisfied


findepi:
Given coordinator with

```
node-scheduler.include-coordinator=false
query-manager.required-workers=1
```

query e.g.
```
EXPLAIN (TYPE IO) ....
```

will get into `WAITING_FOR_RESOURCES` state, waiting for a worker.
However, once the worker shows up and query gets admitted, the query executes on coordinator only.
We should not wait for a worker in this case.

findepi:
I used `EXPLAIN (TYPE IO)` as an example, because the results do not depend on workers in any way.
`EXPLAIN` factors-in current worker count, so the results may be different once the required number
of workers shows up.

PR #29208, opened 2026-04-22
Skip worker availability check for coordinator-only queries

### LINKED ISSUE #4130: Admit coordinator-only "system queries" when required worker count not satisfied
Given coordinator with

```
node-scheduler.include-coordinator=false
query-manager.required-workers=1
```

query
```
select * from system.runtime.nodes;
```

will get into `WAITING_FOR_RESOURCES` state, waiting for a worker.
However, once the worker shows up and query gets admitted, the query executes on coordinator only.
We should not wait for a worker in this case.

stefan-toubia:
## Description

On coordinator-only clusters (`node-scheduler.include-coordinator=false`, zero workers), admin queries like `SELECT * FROM system.runtime.queries` stall indefinitely in `WAITING_FOR_RESOURCES`. `LocalDispatchQuery.waitForMinimumWorkers` unconditionally waits for at least one countable node before dispatch, even though `SqlQueryExecution.shouldWaitForMinWorkers` already reports `false` for queries that only reference internal/system catalogs.

This short-circuits the minimum-worker wait when `shouldWaitForMinWorkers()` returns `false`, matching the existing intent of that predicate.

Added an integration test in `TestMinWorkerRequirement` that runs `SELECT * FROM system.runtime.queries` and `SELECT * FROM system.runtime.nodes` against a coordinator-only cluster with zero workers and asserts they dispatch successfully.

Fixes #4130 (open since 2020). Related: #4131 (same issue for `EXPLAIN`, not addressed here).

## Release notes

```markdown
## General
* Fix admin queries against `system.runtime` tables getting stuck in `WAITING_FOR_RESOURCES` on coordinator-only clusters. ({issue}`4130`)
```

### DIFF
--- a/core/trino-main/src/main/java/io/trino/dispatcher/LocalDispatchQuery.java
+++ b/core/trino-main/src/main/java/io/trino/dispatcher/LocalDispatchQuery.java
@@ -38,6 +38,7 @@ import java.util.concurrent.Executor;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Consumer;
 
+import static com.google.common.util.concurrent.Futures.immediateVoidFuture;
 import static com.google.common.util.concurrent.Futures.nonCancellationPropagating;
 import static io.airlift.concurrent.MoreFutures.addExceptionCallback;
 import static io.airlift.concurrent.MoreFutures.addSuccessCallback;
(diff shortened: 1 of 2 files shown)


stefan-toubia:
The test failures reveal this is not a straightforward change. The issue is that some system table queries like `SELECT DISTINCT connector_name FROM system.metadata.catalogs` still require workers due to the aggregation. Making those queries coordinator-local requires a planner change.

PR #29209, opened 2026-04-22
Add support for match_only_text field type in OpenSearch

### LINKED ISSUE #29165: OpenSearch connector: Add support for match_only_text field type

The OpenSearch connector does not recognize the `match_only_text` field type introduced in OpenSearch 2.12. Fields with this type are silently excluded from the table schema, making them invisible to Trino queries:

    DESCRIBE opensearch.default.my_index;

    -- only shows keyword, float, etc. fields
    -- match_only_text fields are missing

The `match_only_text` type is a storage-optimized variant of `text` that omits positions, frequencies, and norms. From a read perspective, it is identical to `text` — both return a plain string value from `_source`.
OpenSearch documentation: https://opensearch.org/blog/Optimize-storage-and-performance-using-MatchOnlyText-field

The `text` type is already mapped to `VARCHAR` in the connector. The same mapping should apply to `match_only_text`.

**Use case:** `match_only_text` is commonly used for large text fields such as log messages, document bodies,... where storage savings of up to 25% over `text` are significant. Users of these fields are currently forced to use `raw_query` to access them, which returns a single JSON blob instead of individual rows and is capped at the `max_result_window` limit (10,000 by default). This prevents standard SQL operations like JOINs and aggregations over these fields.

**Proposed approach:**

* Add `match_only_text` alongside `text` in the type mapping logic that resolves OpenSearch field types to Trino types (likely a one-line change in the decoder/metadata layer).
* Predicate pushdown is not expected for this type, consistent with the existing `text` behavior.
* Add a test case with a `match_only_text` field to confirm it is exposed as `VARCHAR` and readable via standard SQL.
* Update the type mapping table in the OpenSearch connector docs.

pedroluislopez:
## Description

Map the `match_only_text` OpenSearch field type to `VARCHAR`, alongside the
existing `text` and `keyword` mappings. Fields of this type are now exposed in
the table schema and readable from Trino.

`match_only_text` is a storage-optimized variant of `text` introduced in
OpenSearch 2.12. It returns identical values to `text` (plain strings from
`_source`), so the existing `VarcharDecoder` can be reused as-is.

- Fixes #29165

Before this change, columns declared as `match_only_text` were silently dropped
from the table schema, forcing users to fall back to `raw_query` (which returns
JSON blobs and is capped at 10,000 rows) and preventing standard SQL operations
such as JOINs and aggregations.

The new `testMatchOnlyTextDataType` test is `@Disabled` in
`TestOpenSearchConnectorTest` because the field type is not available in
OpenSearch 1.1.0; it runs on the 2.19.4, 3.0.0, and `latest` images.

## Release notes

```markdown
## OpenSearch connector
* Add support for the `match_only_text` field type. ({issue}`29165`)

### DIFF
--- a/docs/src/main/sphinx/connector/opensearch.md
+++ b/docs/src/main/sphinx/connector/opensearch.md
@@ -215,6 +215,9 @@ according to the following table:
 * - `TEXT`
   - `VARCHAR`
   -
+* - `MATCH_ONLY_TEXT`
+  - `VARCHAR`
+  -
 * - `DATE`
   - `TIMESTAMP`
   - For more information, see [](opensearch-date-types).
--- a/plugin/trino-opensearch/src/main/java/io/trino/plugin/opensearch/OpenSearchMetadata.java
+++ b/plugin/trino-opensearch/src/main/java/io/trino/plugin/opensearch/OpenSearchMetadata.java
@@ -296,7 +296,7 @@ public class OpenSearchMetadata
                 case "short" -> new TypeAndDecoder(SMALLINT, new SmallintDecoder.Descriptor(path));
                 case "integer" -> new TypeAndDecoder(INTEGER, new IntegerDecoder.Descriptor(path));
                 case "long" -> new TypeAndDecoder(BIGINT, new BigintDecoder.Descriptor(path));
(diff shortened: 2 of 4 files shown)


ebyhr:
@pedroluislopez Have you already submitted CLA?

pedroluislopez:
@ebyhr Yes, I sent that this morning (UTC).

PR #29210, opened 2026-04-22
Add support for full text queries in OpenSearch

### LINKED ISSUE #29164: OpenSearch connector: Add full text query support

The Elasticsearch connector supports combining Trino SQL with Elasticsearch full text queries by embedding the query string in the table name:

    SELECT * FROM es.default."myindex:status:active AND message:error"

This uses the Elasticsearch query_string syntax and pushes the query down natively, returning results as regular rows compatible with JOINs, aggregations, and all standard SQL operations. The feature was introduced in #1662 and is documented under "Full text queries" in the Elasticsearch connector docs.

The OpenSearch connector, created as a dedicated connector in #11377, does not include this feature. OpenSearch supports the same query_string syntax: https://opensearch.org/docs/latest/query-dsl/full-text/query-string/

**Use case:** Without full text query support, OpenSearch connector users are limited to `raw_query`, which returns a single JSON blob rather than individual rows and is capped at the `index.max_result_window` limit (10,000 by default). This makes it impossible to efficiently combine full text search results with JOINs against other data sources in Trino, which is the primary reason for using a federated query engine.

**Proposed approach:**

* Port the table name parsing logic from `ElasticsearchClient.java` that   splits `"index:query"` into index name and query string.
* Port the `QueryStringQueryBuilder` integration from   `ElasticsearchQueryBuilder.java` into the OpenSearch equivalent.
* Wire the query through `OpenSearchMetadata.java` and the page source,   following the same pattern as the Elasticsearch connector.
* Add corresponding tests mirroring   `TestElasticsearchIntegrationSmokeTest.java` full text query coverage.
* Document the feature under a new "Full text queries" section in the   OpenSearch connector docs.

The implementation should be straightforward since the query_string API is identical in both Elasticsearch and OpenSearch.

pedroluislopez:
## Description

Allow combining Trino SQL queries with OpenSearch full text queries by embedding
a [query_string](https://opensearch.org/docs/latest/query-dsl/full-text/query-string/) expression in the table name after a colon:

```sql
SELECT * FROM "tweets: +trino SQL^2"
```

The colon-separated fragment is parsed in `OpenSearchMetadata#getTableHandle`,
stored in `OpenSearchTableHandle.query`, and pushed down as a
`QueryStringQueryBuilder` by the existing `OpenSearchQueryBuilder`. This mirrors
the behaviour already available in the Elasticsearch connector.

Fixes #29164.

The downstream machinery was already in place: `OpenSearchTableHandle` carries
an `Optional<String> query` and `OpenSearchQueryBuilder` already wraps that
value in a `QueryStringQueryBuilder`. Only `getTableHandle` needed to parse the
table name and propagate the query fragment, so this PR is a small,
self-contained port of the Elasticsearch feature.

## Release notes

```markdown
## OpenSearch
* Add support for full text queries by appending a Lucene `query_string`
  expression to the table name after a colon, e.g.
  `SELECT * FROM "tweets: +trino SQL^2"`. ({issue}`29164`)
```

### DIFF
--- a/docs/src/main/sphinx/connector/opensearch.md
+++ b/docs/src/main/sphinx/connector/opensearch.md
@@ -390,6 +390,16 @@ The following hidden columns are available:
   - The source of the original document.
 :::
 
+(opensearch-full-text-queries)=
+## Full text queries
+
+Trino SQL queries can be combined with OpenSearch queries by providing the [full text query]
+as part of the table name, separated by a colon. For example:
+
+```sql
+SELECT * FROM "tweets: +trino SQL^2"
+```
+
 (opensearch-sql-support)=
 ## SQL support
 
--- a/plugin/trino-opensearch/src/main/java/io/trino/plugin/opensearch/OpenSearchMetadata.java
+++ b/plugin/trino-opensearch/src/main/java/io/trino/plugin/opensearch/OpenSearchMetadata.java
@@ -180,8 +180,15 @@ public class OpenSearchMetadata
         requireNonNull(tableName, "tableName is null");
 
         if (tableName.getSchemaName().equals(schemaName)) {
-            if (client.indexExists(tableName.getTableName()) && !client.getIndexMetadata(tableName.getTableName()).schema().fields().isEmpty()) {
-                return new OpenSearchTableHandle(OpenSearchTableHandle.Type.SCAN, schemaName, tableName.getTableName(), Optional.empty());
+            String[] parts = tableName.getTableName().split(":", 2);
+            String table = parts[0];
+            Optional<String> query = Optional.empty();
+            if (parts.length == 2) {
+                query = Optional.of(parts[1]);
+            }
+
+            if (client.indexExists(table) && !client.getIndexMetadata(table).schema().fields().isEmpty()) {
+                return new OpenSearchTableHandle(OpenSearchTableHandle.Type.SCAN, schemaName, table, query);
             }
         }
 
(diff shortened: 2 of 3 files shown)


ebyhr (changes requested):
Thanks for your contribution. However, I don't think we want to support this syntax in Opensearch connector. It was intentionally excluded when we port the code from Elasticsearch connector. We should consider modifying `raw_query` table function, or introducing a new table function.

pedroluislopez:
Thank you for the feedback. A table function approach sounds good. OpenSearch's own SQL plugin already has a well-established syntax for this using functions like `match()`, `match_phrase()`, and `query_string()` inside WHERE clauses:

```
SELECT * FROM my_index WHERE match(message, 'this is a test')
SELECT * FROM my_index WHERE match_phrase(address, '880 Holmes Lane')
SELECT * FROM my_index WHERE query_string('status:active AND type:error')
```
Docs: https://opensearch.org/docs/latest/search-plugins/sql/full-text/

A similar approach in Trino could be a table function that accepts the index, query, and optional parameters, returning individual rows (not a single JSON blob like `raw_query`), so the results can be used in JOINs and standard SQL operations without the `max_result_window` limit.

We could propose another issue/PR using this approach.

pedroluislopez:
It could also be resolved by modifying the existing `raw_query` function to return individual rows instead of a single JSON blob, and using scroll internally.

However, this would be a breaking change for existing users who currently parse the single-row JSON response.

pedroluislopez:
Closing in favor of https://github.com/trinodb/trino/pull/29216 with a new table function approach.

Issue #24775, opened 2025-01-23
`$entries` and `$all_entries` system tables throw exceptions after deleted rows


ebyhr:
```sql
TABLE "region$entries"; -- OK
DELETE FROM region WHERE regionkey = 0;
TABLE "region$entries";
```

```sh
Cannot invoke "org.apache.iceberg.types.Type.typeId()" because "type" is null
java.lang.NullPointerException: Cannot invoke "org.apache.iceberg.types.Type.typeId()" because "type" is null
	at org.apache.iceberg.types.Conversions.internalFromByteBuffer(Conversions.java:141)
	at org.apache.iceberg.types.Conversions.fromByteBuffer(Conversions.java:127)
	at io.trino.plugin.iceberg.EntriesTable.lambda$appendIntegerVarcharMap$7(EntriesTable.java:268)
	at java.base/java.util.Map.forEach(Map.java:733)
	at io.trino.plugin.iceberg.EntriesTable.lambda$appendIntegerVarcharMap$8(EntriesTable.java:265)
	at io.trino.spi.block.MapBlockBuilder.buildEntry(MapBlockBuilder.java:132)
	at io.trino.plugin.iceberg.EntriesTable.appendIntegerVarcharMap(EntriesTable.java:265)
	at io.trino.plugin.iceberg.EntriesTable.lambda$appendDataFile$3(EntriesTable.java:207)
	at io.trino.spi.block.RowBlockBuilder.buildEntry(RowBlockBuilder.java:111)
	at io.trino.plugin.iceberg.EntriesTable.appendDataFile(EntriesTable.java:152)
	at io.trino.plugin.iceberg.EntriesTable.addRow(EntriesTable.java:143)
	at io.trino.plugin.iceberg.BaseSystemTable.lambda$addRows$2(BaseSystemTable.java:105)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at io.trino.plugin.iceberg.BaseSystemTable.addRows(BaseSystemTable.java:105)
	at io.trino.plugin.iceberg.BaseSystemTable.lambda$buildPages$1(BaseSystemTable.java:93)
	at java.base/java.lang.Iterable.forEach(Iterable.java:75)
	at io.trino.plugin.iceberg.BaseSystemTable.buildPages(BaseSystemTable.java:93)
	at io.trino.plugin.iceberg.BaseSystemTable.pageSource(BaseSystemTable.java:78)
[...]

PR #29212, opened 2026-04-22
Fix $entries NPE on null sort_order_id

### LINKED ISSUE #29211: Iceberg $entries / $all_entries system tables NPE on data files with null sort_order_id (spec-valid state)
### Trino version

Latest (480)

## Summary

Reading the Iceberg `$entries` or `$all_entries` system table fails with a `GENERIC_INTERNAL_ERROR` / `NullPointerException` whenever the scanned manifests contain any data-file entry whose `sort_order_id` is `null`. A single null value anywhere in the selected manifests is sufficient to fail the entire query.

```
io.trino.spi.TrinoException: Cannot invoke "java.lang.Integer.intValue()" because "sortOrderId" is null
```

This makes `$entries` / `$all_entries` effectively unusable for any Iceberg table written by a client that leaves `sort_order_id` unset on data files. PyIceberg is the common case in the wild: every PyIceberg data-file write sets `sort_order_id = null`, regardless of whether the table has a sort order. The Iceberg spec explicitly permits this null state, so the failure is a reader bug, not writer corruption.

## Affected versions

- **Trino 469** (first release containing `$entries`) **through current `master`** (480+).
- `$entries` was added in commit [`4db417903d`](https://github.com/trinodb/trino/commit/4db417903d) (2024-11-19, released in 469).
- `$all_entries` was added in commit [`47720cbf04`](https://github.com/trinodb/trino/commit/47720cbf04) (2024-12-20, released in 469).
- The NPE pattern has been present since those commits and was carried through the refactor in [`3efd8349ab1`](https://github.com/trinodb/trino/commit/3efd8349ab1) (fix for #24775, released in 470). It is **not** addressed by the writer-side fix in PR [#28091](https://github.com/trinodb/trino/pull/28091), which only ensures tables *written by Trino* populate `sort_order_id`; it does nothing for tables written by other spec-compliant clients.

## Why this is a Trino bug, not writer corruption — per the Iceberg spec

The Iceberg Table Spec (<https://iceberg.apache.org/spec/>) defines field `140 sort_order_id` on `data_file` as **optional** in all versions (v1, v2, and v3). The spec explicitly describes the null state and how readers must interpret it (spec note [3] on the data-file schema):

> **`140 sort_order_id`** — `int` — *optional* — "ID representing sort order for this file [3]."
>
> [3] *"If sort order ID is missing or unknown, then the order is assumed to be unsorted. Only data files and equality delete files should be written with a non-null order id. Position deletes are required to be sorted by file and position, not a table order, and should set sort order id to null. Readers must ignore sort order id for position delete files."*

Source: Apache Iceberg Table Spec, *"Manifests → Data File schema"* section — <https://iceberg.apache.org/spec/#manifests>.

So `sort_order_id = null` on a data file is a valid spec state that readers must tolerate (assume "unsorted"). Trino currently doesn't.

## Root cause

`plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/EntriesTable.java` dereferences the boxed `Integer` into a primitive without a null check, in both the `DATA` and `EQUALITY_DELETE` branches:

https://github.com/trinodb/trino/blob/master/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/EntriesTable.java

```java
// DATA branch (line ~253-254)
Integer sortOrderId = dataFile.get(++position, Integer.class);
INTEGER.writeLong(fieldBuilders.get(position), Long.valueOf(sortOrderId));

// EQUALITY_DELETE branch (line ~268-269)
Integer sortOrderId = dataFile.get(++position, Integer.class);
INTEGER.writeLong(fieldBuilders.get(position), Long.valueOf(sortOrderId));
```

`Long.valueOf(long)` auto-unboxes the `Integer`; when `sortOrderId` is `null`, the unbox throws `NullPointerException`. The exception propagates up through `BaseSystemTable.buildPages` and fails the whole query.

Stack trace (abbreviated, from a production Trino 476 deployment):

```
io.trino.plugin.iceberg.system.EntriesTable.lambda$appendDataFile$...(EntriesTable.java:254)
io.trino.spi.block.RowBlockBuilder.buildEntry(...)
io.trino.plugin.iceberg.system.EntriesTable.addRow(...)
io.trino.plugin.iceberg.BaseSystemTable.addRows(...)
io.trino.plugin.iceberg.BaseSystemTable.buildPages(...)
Caused by: java.lang.NullPointerException:
    Cannot invoke "java.lang.Integer.intValue()" because "sortOrderId" is null
```

## Reproduction

The cheapest repro uses PyIceberg, whose data-file write path unconditionally sets `sort_order_id = null`. PyIceberg's behavior is spec-compliant; Java/Spark writers happen to default to `0` = "unsorted", which masks the bug for tables written by those clients only.

### Step 1 — create and populate a table with PyIceberg

```python
# Python — requires: pip install "pyiceberg[pyarrow,s3fs]" pyarrow
import pyarrow as pa
from pyiceberg.catalog import load_catalog
from pyiceberg.schema import Schema
from pyiceberg.types import LongType, NestedField, TimestampType

catalog = load_catalog("rest", **{
    "type": "rest",
    "uri": "http://localhost:8181",          # your Iceberg REST catalog
    "warehouse": "s3://test-warehouse/",
    "s3.endpoint": "http://localhost:9000",
    "s3.access-key-id": "admin",
    "s3.secret-access-key": "password",
})

schema = Schema(
    NestedField(1, "id", LongType(), required=True),
    NestedField(2, "ts", TimestampType(), required=True),
)

catalog.create_namespace_if_not_exists("repro")
tbl = catalog.create_table("repro.pyiceberg_no_sort_order", schema=schema)

arrow_tbl = pa.table({
    "id": pa.array([1, 2, 3], type=pa.int64()),
    "ts": pa.array([
        "2026-01-01T00:00:00",
        "2026-01-01T00:01:00",
        "2026-01-01T00:02:00",
    ], type=pa.timestamp("us")),
})
tbl.append(arrow_tbl)   # <-- writes a data file with sort_order_id = null
```

### Step 2 — query `$entries` / `$all_entries` from Trino

Point Trino at the same Iceberg catalog, then run any of:

```sql
-- all three fail identically, regardless of projection or predicate:
SELECT count(*) FROM iceberg.repro."pyiceberg_no_sort_order\$entries";
SELECT count(*) FROM iceberg.repro."pyiceberg_no_sort_order\$all_entries";
SELECT sort_order_id FROM iceberg.repro."pyiceberg_no_sort_order\$entries";
```

Expected: result set.
Actual:

```
Query ... failed: Cannot invoke "java.lang.Integer.intValue()" because "sortOrderId" is null
```

## Metadata-table compatibility on an affected table

Tested on a 853 M-row PyIceberg-written table against Trino 476. Only the views that per-entry project `sort_order_id` fail; everything at the manifest-file / snapshot / partition / file level works:

| View             | Status | Notes                                                                 |
|------------------|:------:|-----------------------------------------------------------------------|
| `$properties`    | OK     | table-level properties                                                |
| `$history`       | OK     | snapshot ancestry                                                     |
| `$snapshots`     | OK     | includes `summary` map (added-data-files, added-records, …)           |
| `$manifests`     | OK     | manifest files of the current snapshot                                |
| `$partitions`    | OK     | per-partition stats including `readable_metrics`                      |
| `$files`         | OK     | per-file stats + `readable_metrics`                                   |
| `$all_manifests` | OK     | every historical manifest                                             |
| `$refs`          | OK     | branch/tag refs                                                       |
| **`$entries`**     | **FAIL**   | `sortOrderId is null` — per-entry projection fails                    |
| **`$all_entries`** | **FAIL**   | `sortOrderId is null` — per-entry projection fails                    |

## Proposed fix

Null-check the value and either emit SQL `NULL` (preserving the on-disk semantics) or coerce to `0` (the reserved id for the unsorted order, per the spec). The former is strictly more faithful to what was written:

```java
Integer sortOrderId = dataFile.get(++position, Integer.class);
if (sortOrderId == null) {
    fieldBuilders.get(position).appendNull();
} else {
    INTEGER.writeLong(fieldBuilders.get(position), sortOrderId.longValue());
}
```

Applied in both the `DATA` (line ~253-254) and `EQUALITY_DELETE` (line ~268-269) branches of `EntriesTable#appendDataFile`.

### Regression test

Trino's own writer always populates `sort_order_id`, and `DataFiles.Builder` from iceberg-java also hard-defaults it to `SortOrder.unsorted().orderId()` (== `0`) in its no-arg constructor — so neither of them can reach the null state from a test. A pure-Java regression test therefore has to go around the builder. The simplest way is to build a `DataFile` normally, then reflectively null the inherited `BaseFile.sortOrderId` field before committing:

```java
DataFile dataFile = DataFiles.builder(icebergTable.spec())
        .withPath(path).withFormat(format)
        .withFileSizeInBytes(1234).withRecordCount(1)
        .build();
// DataFiles.Builder defaults sort_order_id to SortOrder.unsorted().orderId() (== 0).
// Reach the spec-valid null state by clearing the inherited BaseFile.sortOrderId
// field so the manifest row is written with the column unset.
Field f = dataFile.getClass().getSuperclass().getDeclaredField("sortOrderId");
f.setAccessible(true);
f.set(dataFile, null);
icebergTable.newAppend().appendFile(dataFile).commit();
```

With that single file committed, `SELECT count(*) FROM <table>$entries` throws the exact NPE above on unfixed Trino, and returns a row with `sort_order_id IS NULL` after the fix.

## Related

- PR [#28091](https://github.com/trinodb/trino/pull/28091) — *Populate `sort_order_id` in iceberg metadata during writes* (merged 2026-02-04). Covers the **writer** side only; tables written by Trino no longer trip this NPE when later read, but tables written by any other spec-compliant client still do.
- Issue [#24775](https://github.com/trinodb/trino/issues/24775) — neighboring NPE in the same `EntriesTable` code, different code path (null `Type` on bound deserialization after a `DELETE`). Fixed in [`3efd8349ab1`](https://github.com/trinodb/trino/commit/3efd8349ab1), released in 470. Suggests `EntriesTable` would benefit from a general null-tolerance pass.

jonasbrami:
## Description

Fix `NullPointerException` when reading the Iceberg `$entries` or `$all_entries` system table on a table whose manifest contains any data file or equality delete file with `sort_order_id = null`. A single null anywhere in the selected manifests currently fails the whole query with `Cannot invoke "java.lang.Integer.intValue()" because "sortOrderId" is null`.

The Iceberg table spec marks `data_file.sort_order_id` as optional in v1, v2, and v3: *"If sort order ID is missing or unknown, then the order is assumed to be unsorted"*. Trino's writer now always populates `0` (PR #28091), but tables written by spec-compliant clients such as PyIceberg — which unconditionally sets `sort_order_id = null` on every data-file write — still trip the NPE when read through `$entries` / `$all_entries`.

`EntriesTable#appendDataFile` auto-unboxed the boxed `Integer` via `Long.valueOf(sortOrderId)` in both the `DATA` and `EQUALITY_DELETE` branches. This PR null-checks both sites and emits SQL `NULL` when the field is unset, matching the existing `key_metadata` idiom a few lines above in the same method.

A regression test (`BaseIcebergSystemTables#testEntriesTableWithNullSortOrderId`, run on both the Parquet and ORC subclasses) commits a `DataFile` directly via the Iceberg Java API. Since both Trino's writer and iceberg-java's `DataFiles.Builder` hard-default `sort_order_id` to `0` — `Builder`'s no-arg constructor sets `SortOrder.unsorted().orderId()` — the test reflectively clears the inherited `BaseFile.sortOrderId` field before `appendFile(...).commit()` to reach the spec-valid null state. Before this fix the test throws the exact NPE above; after, it returns a row with `data_file.sort_order_id IS NULL`.

- Fixes #29211.
- `$entries` was introduced in `4db417903d` (released in 469) and `$all_entries` in `47720cbf04` (released in 469). The NPE has been present in both since those commits and was carried through the refactor in `3efd8349ab1` (fix for #24775, released in 470). All Trino releases **469 through current master (480+)** are affected.
- Related but distinct: PR #28091 addressed the **writer** side (Trino's own writer now populates `sort_order_id`); this PR addresses the **reader** side for manifests produced by any other spec-compliant writer.
- Related prior NPE in the same file on a different code path: #24775 (fixed in `3efd8349ab1`). This PR takes the same minimal null-tolerance approach.

## Release notes

```markdown
## Iceberg connector
* Fix `NullPointerException` when reading the `$entries` or `$all_entries`. ({issue}`29211`)
```

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/EntriesTable.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/EntriesTable.java
@@ -242,8 +242,14 @@ public class EntriesTable
                     // data files don't have equality ids
                     fieldBuilders.get(++position).appendNull();
 
+                    // sort_order_id is optional per the Iceberg spec — null means "unsorted"
                     Integer sortOrderId = dataFile.get(++position, Integer.class);
-                    INTEGER.writeLong(fieldBuilders.get(position), Long.valueOf(sortOrderId));
+                    if (sortOrderId == null) {
+                        fieldBuilders.get(position).appendNull();
+                    }
+                    else {
+                        INTEGER.writeLong(fieldBuilders.get(position), sortOrderId.longValue());
+                    }
                 }
                 case POSITION_DELETE -> {
                     // position delete files don't have equality ids
@@ -258,7 +264,12 @@ public class EntriesTable
                     appendIntegerArray((ArrayBlockBuilder) fieldBuilders.get(position), equalityIds);
 
                     Integer sortOrderId = dataFile.get(++position, Integer.class);
-                    INTEGER.writeLong(fieldBuilders.get(position), Long.valueOf(sortOrderId));
+                    if (sortOrderId == null) {
+                        fieldBuilders.get(position).appendNull();
+                    }
+                    else {
+                        INTEGER.writeLong(fieldBuilders.get(position), sortOrderId.longValue());
+                    }
                 }
             }
         });
--- a/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergSystemTables.java
+++ b/plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergSystemTables.java
@@ -25,7 +25,10 @@ import io.trino.testing.MaterializedRow;
 import io.trino.testing.QueryRunner;
 import io.trino.testing.sql.TestTable;
 import org.apache.iceberg.BaseTable;
+import org.apache.iceberg.DataFile;
+import org.apache.iceberg.DataFiles;
 import org.apache.iceberg.FileContent;
+import org.apache.iceberg.FileFormat;
 import org.apache.iceberg.Snapshot;
 import org.apache.iceberg.Table;
 import org.intellij.lang.annotations.Language;
@@ -34,6 +37,7 @@ import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.TestInstance;
 
+import java.lang.reflect.Field;
 import java.time.LocalDate;
 import java.util.ArrayList;
 import java.util.List;
@@ -789,6 +793,45 @@ public abstract class BaseIcebergSystemTables
                             "dt":{"column_size":null,"value_count":null,"null_value_count":null,"nan_value_count":null,"lower_bound":null,"upper_bound":null},\
                             "id":{"column_size":51,"value_count":1,"null_value_count":0,"nan_value_count":null,"lower_bound":1,"upper_bound":1}\
                             }""");
+        }
+    }
+
+    @Test
+    void testEntriesTableWithNullSortOrderId()
+            throws Exception
+    {
+        // Iceberg spec (table spec v1/v2/v3) marks data_file.sort_order_id as optional: a null value
+        // is a valid on-disk state meaning "unsorted", and readers must tolerate it. Writers such as
+        // PyIceberg emit null here on default appends. Trino's own writer populates 0, which is why
+        // reaching the code path from SQL is impossible — commit a DataFile directly through the
+        // Iceberg Java API to produce a manifest entry with sort_order_id unset.
+        try (TestTable table = new TestTable(getQueryRunner()::execute, "test_entries_null_sort_order", "AS SELECT 1 id")) {
+            Table icebergTable = loadTable(table.getName());
+            DataFile dataFile = DataFiles.builder(icebergTable.spec())
+                    .withPath(icebergTable.location() + "/data/synthetic-no-sort-order." + format.name().toLowerCase(ENGLISH))
+                    .withFormat(FileFormat.fromString(format.name()))
+                    .withFileSizeInBytes(1234)
+                    .withRecordCount(1)
+                    .build();
+            // DataFiles.Builder defaults sort_order_id to SortOrder.unsorted().orderId() (== 0).
+            // The spec-valid null state is what PyIceberg emits; reach it by clearing the inherited
+            // BaseFile.sortOrderId field so the manifest row is written with the column unset.
+            Field sortOrderIdField = dataFile.getClass().getSuperclass().getDeclaredField("sortOrderId");
+            sortOrderIdField.setAccessible(true);
+            sortOrderIdField.set(dataFile, null);
+            assertThat(dataFile.sortOrderId()).isNull();
+
+            icebergTable.newAppend().appendFile(dataFile).commit();
+            long appendedSnapshotId = icebergTable.currentSnapshot().snapshotId();
+
+            // Reading $entries must not throw and must surface the null as SQL NULL.
+            assertThat(computeScalar("SELECT data_file.sort_order_id FROM \"" + table.getName() + "$entries\"" +
+                    " WHERE snapshot_id = " + appendedSnapshotId))
+                    .isNull();
+            // $all_entries shares the same code path and must also tolerate the null.
+            assertThat(computeScalar("SELECT data_file.sort_order_id FROM \"" + table.getName() + "$all_entries\"" +
+                    " WHERE snapshot_id = " + appendedSnapshotId))
+                    .isNull();
         }
     }
 


wendigo:
The question is whether we should hardcore 0 or use NULL in that case.

cc @ebyhr @raunaqmorarka

raunaqmorarka:
> The question is whether we should hardcore 0 or use NULL in that case.
> 
> cc @ebyhr @raunaqmorarka

Returning `SortOrder.unsorted().orderId()` or NULL is fine, I don't think it matters which one we pick here

ebyhr (approved):

PR #29217, opened 2026-04-22
Improve Parquet file and row group size estimates

dain:
## Description

Estimate pending Parquet writer bytes using compressed page history instead of
raw buffered bytes. The estimate now uses per-column compression ratios when
enough data exists, falls back to aggregate file compression stats for small
samples, and caps compression expansion at 1.5x. Row group rollover and
file-size reporting now share the same compression-aware estimate.

Experimented with a 350 MB Parquet file using ZSTD. The tables show average
non-tail output sizes in MB before and after this change. I tested one run
with a very large row group target so each output file had one row group,
and another with row groups set to one quarter of the target file size.
Non-tail files and row groups are typically within a couple percent of the
target after the change; the smallest row group target tested, 8 MB, averaged
about 12% too small.

Single-row-group output:

| Target size | RG target | Before file | After file |
|-------------|-----------|------------:|-----------:|
| 32 MB       | 1024 MB   |      4.8 MB |    31.4 MB |
| 64 MB       | 1024 MB   |     21.5 MB |    63.9 MB |
| 128 MB      | 1024 MB   |     87.2 MB |   126.9 MB |
| 256 MB      | 1024 MB   |    210.8 MB |   254.6 MB |

Quarter-size row group target:

| Target size | RG target | Before file | After file | Before RG | After RG |
|-------------|-----------|------------:|-----------:|----------:|---------:|
| 32 MB       | 8 MB      |     30.0 MB |    32.6 MB |    0.9 MB |   7.1 MB |
| 64 MB       | 16 MB     |     55.9 MB |    64.6 MB |    2.1 MB |  16.2 MB |
| 128 MB      | 32 MB     |    105.1 MB |   128.5 MB |    4.8 MB |  31.3 MB |
| 256 MB      | 64 MB     |    221.4 MB |   256.6 MB |   21.4 MB |  63.9 MB |

Also, there is a small initial commit to make the parquet footer read size configurable.  The hard coded value is too small for wide tables.

## Release notes

```markdown
## Section
* Improve Parquet file and row group size estimates. ({issue}`issuenumber`)
* Add `parquet.footer-read-size` configuration property. ({issue}`issuenumber`)
```

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
@@ -13,6 +13,7 @@
  */
 package io.trino.parquet.reader;
 
+import com.google.common.annotations.VisibleForTesting;
 import io.airlift.slice.Slice;
 import io.airlift.slice.SliceInput;
 import io.airlift.slice.Slices;
@@ -20,6 +21,7 @@ import io.airlift.units.DataSize;
 import io.trino.parquet.ParquetCorruptionException;
 import io.trino.parquet.ParquetDataSource;
 import io.trino.parquet.ParquetDataSourceId;
+import io.trino.parquet.ParquetReaderOptions;
 import io.trino.parquet.ParquetWriteValidation;
 import io.trino.parquet.crypto.AesCipherUtils;
 import io.trino.parquet.crypto.AesGcmEncryptor;
@@ -41,6 +43,7 @@ import java.util.Arrays;
 import java.util.Optional;
 
 import static io.trino.parquet.ParquetMetadataConverter.fromParquetStatistics;
+import static io.trino.parquet.ParquetReaderOptions.DEFAULT_FOOTER_READ_SIZE;
 import static io.trino.parquet.ParquetValidationUtils.validateParquet;
 import static io.trino.parquet.ParquetValidationUtils.validateParquetCrypto;
 import static io.trino.parquet.crypto.AesCipherUtils.GCM_TAG_LENGTH;
@@ -50,6 +53,7 @@ import static java.lang.Boolean.TRUE;
 import static java.lang.Math.min;
 import static java.lang.Math.toIntExact;
 import static java.lang.System.arraycopy;
+import static java.util.Objects.requireNonNull;
 import static org.apache.parquet.format.Util.readFileCryptoMetaData;
 import static org.apache.parquet.format.Util.readFileMetaData;
 
@@ -58,24 +62,24 @@ public final class MetadataReader
     private static final Slice MAGIC = Slices.utf8Slice("PAR1");
     private static final Slice EMAGIC = Slices.utf8Slice("PARE");
     private static final int POST_SCRIPT_SIZE = Integer.BYTES + MAGIC.length();
-    // Typical 1GB files produced by Trino were found to have footer size between 30-40KB
-    private static final int EXPECTED_FOOTER_SIZE = 48 * 1024;
 
     private MetadataReader() {}
 
+    @VisibleForTesting
     public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<FileDecryptionProperties> fileDecryptionProperties)
             throws IOException
     {
-        return readFooter(dataSource, Optional.empty(), Optional.empty(), fileDecryptionProperties);
+        return readFooter(dataSource, DEFAULT_FOOTER_READ_SIZE, Optional.empty(), Optional.empty(), fileDecryptionProperties);
     }
 
-    public static ParquetMetadata readFooter(ParquetDataSource dataSource, DataSize maxFooterReadSize, Optional<FileDecryptionProperties> fileDecryptionProperties)
+    public static ParquetMetadata readFooter(ParquetDataSource dataSource, ParquetReaderOptions options, Optional<ParquetWriteValidation> parquetWriteValidation, Optional<FileDecryptionProperties> fileDecryptionProperties)
             throws IOException
     {
-        return readFooter(dataSource, Optional.of(maxFooterReadSize), Optional.empty(), fileDecryptionProperties);
+        requireNonNull(options, "options is null");
+        return readFooter(dataSource, options.getFooterReadSize(), Optional.of(options.getMaxFooterReadSize()), parquetWriteValidation, fileDecryptionProperties);
     }
 
-    public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<DataSize> maxFooterReadSize, Optional<ParquetWriteValidation> parquetWriteValidation, Optional<FileDecryptionProperties> fileDecryptionProperties)
+    public static ParquetMetadata readFooter(ParquetDataSource dataSource, DataSize footerReadSize, Optional<DataSize> maxFooterReadSize, Optional<ParquetWriteValidation> parquetWriteValidation, Optional<FileDecryptionProperties> fileDecryptionProperties)
             throws IOException
     {
         // Parquet File Layout:
@@ -90,7 +94,9 @@ public final class MetadataReader
 
         // Read the tail of the file
         long estimatedFileSize = dataSource.getEstimatedSize();
-        long expectedReadSize = min(estimatedFileSize, EXPECTED_FOOTER_SIZE);
+        long footerReadSizeInBytes = footerReadSize.toBytes();
+        validateParquet(footerReadSizeInBytes >= POST_SCRIPT_SIZE, dataSource.getId(), "Parquet footer read size must be at least %s bytes", POST_SCRIPT_SIZE);
+        long expectedReadSize = min(estimatedFileSize, footerReadSizeInBytes);
         Slice buffer = dataSource.readTail(toIntExact(expectedReadSize));
 
         Slice magic = buffer.slice(buffer.length() - MAGIC.length(), MAGIC.length());
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/writer/StructColumnWriter.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/writer/StructColumnWriter.java
@@ -84,14 +84,24 @@ public class StructColumnWriter
     }
 
     @Override
-    public long getBufferedBytes()
+    public long getEstimatedBufferedBytes(CompressionStats compressionStats)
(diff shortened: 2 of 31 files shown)


dain:
@coderabbitai full review

dain:
@coderabbit review

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -58,25 +59,29 @@ public final class MetadataReader
...
-    // Typical 1GB files produced by Trino were found to have footer size between 30-40KB
-    private static final int EXPECTED_FOOTER_SIZE = 48 * 1024;
 
     private MetadataReader() {}
 
     public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<FileDecryptionProperties> fileDecryptionProperties)
             throws IOException
     {
-        return readFooter(dataSource, Optional.empty(), Optional.empty(), fileDecryptionProperties);
+        return readFooter(dataSource, ParquetReaderOptions.DEFAULT_FOOTER_READ_SIZE, Optional.empty(), Optional.empty(), fileDecryptionProperties);
```
electrum: Static import

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -58,25 +59,29 @@ public final class MetadataReader
     private static final Slice MAGIC = Slices.utf8Slice("PAR1");
     private static final Slice EMAGIC = Slices.utf8Slice("PARE");
     private static final int POST_SCRIPT_SIZE = Integer.BYTES + MAGIC.length();
-    // Typical 1GB files produced by Trino were found to have footer size between 30-40KB
-    private static final int EXPECTED_FOOTER_SIZE = 48 * 1024;
 
     private MetadataReader() {}
 
     public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<FileDecryptionProperties> fileDecryptionProperties)
```
electrum, line 69: After this change, this method should only be used by tests (only production use was `DeltaLakeMergeSink`), so add `@VisibleForTesting`

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -58,25 +59,29 @@ public final class MetadataReader
...
             throws IOException
     {
-        return readFooter(dataSource, Optional.empty(), Optional.empty(), fileDecryptionProperties);
+        return readFooter(dataSource, ParquetReaderOptions.DEFAULT_FOOTER_READ_SIZE, Optional.empty(), Optional.empty(), fileDecryptionProperties);
     }
 
     public static ParquetMetadata readFooter(ParquetDataSource dataSource, DataSize maxFooterReadSize, Optional<FileDecryptionProperties> fileDecryptionProperties)
```
electrum: Is this version of the method used anymore? I think it could be deleted

lib/trino-parquet/src/main/java/io/trino/parquet/reader/MetadataReader.java
```diff
@@ -58,25 +59,29 @@ public final class MetadataReader
...
             throws IOException
     {
-        return readFooter(dataSource, Optional.of(maxFooterReadSize), Optional.empty(), fileDecryptionProperties);
+        return readFooter(dataSource, ParquetReaderOptions.DEFAULT_FOOTER_READ_SIZE, Optional.of(maxFooterReadSize), Optional.empty(), fileDecryptionProperties);
     }
 
     public static ParquetMetadata readFooter(ParquetDataSource dataSource, Optional<DataSize> maxFooterReadSize, Optional<ParquetWriteValidation> parquetWriteValidation, Optional<FileDecryptionProperties> fileDecryptionProperties)
```
electrum: The only remaining usage of this method is `ParquetWriter`, which could pass a value here, then we could delete this method. It could record the size of the footer when writing and pass the exact size.

plugin/trino-hive/src/main/java/io/trino/plugin/hive/parquet/ParquetReaderConfig.java
```diff
@@ -191,6 +191,23 @@ public DataSize getMaxFooterReadSize()
...
+    public ParquetReaderConfig setFooterReadSize(DataSize footerReadSize)
+    {
+        options = ParquetReaderOptions.builder(options)
+                .withFooterReadSize(footerReadSize)
+                .build();
+        return this;
+    }
+
+    @NotNull
+    @MinDataSize("8B")
```
electrum, line 205: Add a max here?
```java
@MaxDataSize("128MB")
```

plugin/trino-delta-lake/src/main/java/io/trino/plugin/deltalake/DeltaLakeMergeSink.java
```diff
@@ -400,7 +400,7 @@ private Slice writeMergeResult(Slice path, FileDeletion deletion)
         }
         TrinoInputFile inputFile = fileSystem.newInputFile(Location.of(path.toStringUtf8()));
         try (ParquetDataSource dataSource = new TrinoParquetDataSource(inputFile, parquetReaderOptions, fileFormatDataSourceStats)) {
-            ParquetMetadata parquetMetadata = MetadataReader.readFooter(dataSource, Optional.empty());
+            ParquetMetadata parquetMetadata = MetadataReader.readFooter(dataSource, parquetReaderOptions.getFooterReadSize(), Optional.of(parquetReaderOptions.getMaxFooterReadSize()), Optional.empty(), Optional.empty());
```
electrum: We might want a version of the method that takes `ParquetReaderOptions`

lib/trino-parquet/src/main/java/io/trino/parquet/writer/valuewriter/BloomFilterValuesWriter.java
```diff
@@ -55,6 +55,14 @@ public long getBufferedSize()
         return writer.getBufferedSize();
     }
 
+    public long getEstimatedBufferedSize()
+    {
+        if (writer instanceof DictionaryFallbackValuesWriter dictionaryFallbackValuesWriter) {
```
electrum: Could use a switch expression
```java
return switch (writer) {
    case DictionaryFallbackValuesWriter dictionary -> dictionary.getEstimatedBufferedSize();
    default -> writer.getBufferedSize();
};

File: lib/trino-parquet/src/main/java/io/trino/parquet/writer/valuewriter/PrimitiveValueWriter.java [hunk 8]
```diff
@@ -58,6 +58,17 @@ public long getBufferedSize()
         return valuesWriter.getBufferedSize();
     }
 
+    public long getEstimatedBufferedSize()
+    {
+        if (valuesWriter instanceof DictionaryFallbackValuesWriter dictionaryFallbackValuesWriter) {
```
electrum: Could use switch expression
```java
return switch (valuesWriter) {
    case DictionaryFallbackValuesWriter dictionary -> dictionary.getEstimatedBufferedSize();
    case BloomFilterValuesWriter filter -> filter.getEstimatedBufferedSize();
    default -> valuesWriter.getBufferedSize();
};

File: lib/trino-parquet/src/main/java/io/trino/parquet/writer/StructColumnWriter.java [hunk 9]
```diff
@@ -84,16 +84,25 @@ public List<BufferData> getBuffer()
     }
 
     @Override
-    public long getBufferedBytes()
+    public long getEstimatedBufferedBytes(CompressionStats compressionStats)
     {
-        // Avoid using streams here for performance reasons
```
electrum, line 89: Could keep the comment

lib/trino-parquet/src/test/java/io/trino/parquet/writer/TestParquetWriter.java
```diff
@@ -339,6 +341,78 @@ public void testWriterMemoryAccounting()
...
+    public void testEstimatedWrittenBytesUsesEncodedCompressedSize()
+            throws IOException
+    {
+        List<String> columnNames = ImmutableList.of("column");
+        List<Type> types = ImmutableList.of(VARCHAR);
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
+        ParquetWriter writer = createParquetWriter(
+                outputStream,
+                ParquetWriterOptions.builder()
+                        .setMaxPageSize(DataSize.ofBytes(2 * 1024 * 1024))
```
electrum: You can use `DataSize.of(2, MEGABYTE)`

Same for the rest

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergParquetConnectorTest.java
```diff
@@ -185,8 +185,8 @@ void testTableChangesOnMultiRowGroups()
             String filePath = getOnlyTableFilePath(table.getName());
             ParquetMetadata parquetMetadata = getParquetFileMetadata(fileSystem.newInputFile(Location.of(filePath)));
             int blocksSize = parquetMetadata.getBlocks().size();
-            int splitBatchSize = new QueryManagerConfig().getScheduleSplitBatchSize();
-            assertThat(blocksSize > splitBatchSize && blocksSize % splitBatchSize != 0).isTrue();
```
electrum, line 189: The mod check is not needed?

electrum (approved):

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergParquetConnectorTest.java
dain: I removed this intentionally. The old check required the row-group count to be greater than the split batch size and also not an exact multiple of it, which means it was testing for a final partial split batch. The behavior this test is trying to cover is that table_changes processes splits across more than one batch, and blocksSize > splitBatchSize is enough for that. Keeping the modulo check makes the test depend on the exact number of generated row groups, which is brittle now that compression-aware row-group sizing can move that count around.

PR #29228, opened 2026-04-24
Use getBranch to get the branch name for table and view accounting

ksobolew:
## Description

This is not used for access control, it's only meant to properly account for tables referenced by the query - and the view. If we improperly register the branch, then it won't be found in the `tableColumnReferences` collection in `Analysis#getReferencedTables` and the query will be reported in `QueryStateMachine#referencedTables` as querying no columns in the affected table or view.

Fixes #29179

## Release notes

  * This is a fix to unreleased code

```markdown
## Section
```


kokosing (commented):
Please provide tests for both cases.

ksobolew:
> Please provide tests for both cases.

I dug deeper into this and it looks like they don't affect access control - the difference is in the `QueryInfo` for the query. If the branch is not properly passed, then the `QueryInfo` will report that no columns were queried in the table. I'm not sure this can be easily tested...

kokosing (approved):

PR #29231, opened 2026-04-24
Replace exception-driven view detection in Hive and Iceberg metadata layers

### LINKED ISSUE #29163: Replace exception-driven view detection in Hive and Iceberg metadata layers

Several places in the Hive and Iceberg connectors currently detect whether a metastore relation is a view by attempting a table-shaped operation and catching the resulting exception:

- HiveMetadata.java — `// it is not a table (e.g. it's a view) (TODO remove exception-driven logic for this case)`
- GlueIcebergTableOperations.java — `// TODO table operations should not be constructed for views (remove exception-driven code path)`
- An equivalent check is reachable through HiveMetastoreTableOperations.java.

**Use case:** Exception-driven control flow obscures the intended logic, makes stack-trace-based debugging noisy, and pays the cost of throwable construction on every view access. A small SPI refinement would let all three call sites disappear together.

**Proposed approach:**
- Add a method such as `Optional<RelationKind> getRelationKind(SchemaTableName)` (or extend the existing relation-listing API) on the metastore abstraction so callers can branch explicitly.
- Convert the three call sites to the new API.
- Remove the now-dead `try/catch` paths.
- Confirm no behavior change for unknown relations.

laserninja:
## Description

Move view/materialized-view detection from `getRefreshedLocation()` to `loadTable()` in both `TrinoHiveCatalog` and `TrinoGlueCatalog` so that views are rejected early with a clean `TableNotFoundException` instead of relying on catching exceptions thrown deep inside Iceberg table loading.

In `HiveMetadata.streamTableColumns()`, pre-check the metastore before entering the try block so views are skipped without triggering (and catching) a `TrinoException`.

Fixes #29163

The core insight is that `getRefreshedLocation()` is only reachable through the `loadTable()` → `loadIcebergTable()` → `refresh()` call chain, so moving the view check into `loadTable()` makes the downstream check dead code.

### Changes

**Iceberg – HMS catalog**
- `TrinoHiveCatalog.loadTable()`: Fetch the metastore table upfront; throw `TableNotFoundException` if absent or any kind of view.
- `AbstractMetastoreTableOperations.getRefreshedLocation()`: Remove now-dead view detection block and unused imports.

**Iceberg – Glue catalog**
- `TrinoGlueCatalog.loadTable()`: Check `isTrinoView`/`isTrinoMaterializedView` before creating `TableOperations`; throw `TableNotFoundException` early.
- `GlueIcebergTableOperations.getRefreshedLocation()`: Remove now-dead view detection block and unused imports/variables.

**Hive**
- `HiveMetadata.streamTableColumns()`: Pre-check the metastore table before the try block; return `Stream.empty()` when the table is missing or is a view (unless it is a Hive view and `translateHiveViews` is enabled).

## Release notes


laserninja:
@findepi @ebyhr can you review this please

PR #29239, opened 2026-04-27
Add delete file metrics to Iceberg $partitions table

### LINKED ISSUE #28910: Iceberg: `$partitions` metadata table missing delete file metrics

1fanwang:
## Description

Adds four delete-file metric columns to the Iceberg `$partitions` metadata table:

- `position_delete_record_count` (BIGINT)
- `position_delete_file_count` (BIGINT)
- `equality_delete_record_count` (BIGINT)
- `equality_delete_file_count` (BIGINT)

The same columns are already exposed by Iceberg's Spark connector — see [Spark `$partitions`](https://iceberg.apache.org/docs/1.10.0/spark-queries/#partitions).

Today, finding partitions that have accumulated merge-on-read delete files (and would benefit from `OPTIMIZE`) requires aggregating `$files`, which scans every file metadata entry on every query. The `$partitions` table is designed for exactly this pre-aggregated visibility, but currently omits delete files entirely.

Implementation:

- `IcebergStatistics.Builder` gains an `acceptDeleteFile(DeleteFile)` method that branches on `FileContent.{POSITION_DELETES,EQUALITY_DELETES}` and increments the corresponding counters.
- `PartitionsTable.getStatisticsByPartition()` iterates `fileScanTask.deletes()` and feeds each entry to the builder. The same delete file (and the same DV inside a Puffin file in V3) can apply to many data files in the partition, so it shows up across multiple `FileScanTask`s — dedup by `(location, referencedDataFile)` so each logical delete entry counts once per partition, matching Iceberg native `PartitionsTable`'s manifest-based counting.
- The new columns are appended after `data` to preserve positional access for existing JDBC clients reading the previous columns by index.

`TableStatisticsReader` is unaffected: it only iterates DataFiles, so the new fields default to `0` there (no behaviour change).

Closes #28910.

`spec_id`, `last_updated_at`, and `last_updated_snapshot_id` are also mentioned in the issue. They are out of scope for this PR — the four delete metric columns are the operational pain point, and the snapshot-tracking columns require additional manifest-walking logic that is cleaner as a follow-up.

## Test coverage

- `BaseIcebergSystemTables.testPartitionsTable` — schema assertion updated; sums of the four delete columns are 0 when there are no DML deletes.
- `BaseIcebergSystemTables.testPartitionsTablePositionDeleteMetrics` (new) — uses the existing `test_table_with_dml` fixture (an UPDATE and a DELETE) and asserts `position_delete_*` are positive on the affected partitions and `0` on untouched partitions.
- `BaseIcebergSystemTables.testPartitionsTableEqualityDeleteMetrics` (new) — writes an equality delete file via `EqualityDeleteUtils` for one partition of a V2 table and asserts `equality_delete_*` is positive on that partition and `0` elsewhere; position columns stay at `0` across the board.
- `BaseIcebergConnectorTest.testCreatePartitionedTable` — `information_schema.columns` assertion updated for the new column list.

## Release notes

```markdown
## Iceberg connector
* Add `position_delete_record_count`, `position_delete_file_count`, `equality_delete_record_count`, and `equality_delete_file_count` columns to the `$partitions` metadata table to surface merge-on-read delete file accumulation per partition. ({issue}`28910`)
```


docs/src/main/sphinx/connector/iceberg.md
```diff
@@ -1475,6 +1475,22 @@ The output of the query has the following columns:
 * - `data`
   - `ROW(... ROW (min ..., max ... , null_count BIGINT, nan_count BIGINT))`
```
findinpath, line 1476: Add to the description of the PR https://iceberg.apache.org/docs/1.10.0/spark-queries/?h=partitions#partitions for other maintainers to have context that these capabilities exist already in the Spark implementations as well.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergPartitionEvolution.java
```diff
@@ -83,7 +83,11 @@ public void testDroppedPartitionField(boolean dropFirst)
                         row("data", "row(" +
                                 // A/B is now partitioning column in the first partitioning spec, and non-partitioning in new one
                                 (dropFirst ? "\"a\"" : "\"b\"") + " row(\"min\" varchar, \"max\" varchar, \"null_count\" bigint, \"nan_count\" bigint), " +
-                                "\"c\" row(\"min\" varchar, \"max\" varchar, \"null_count\" bigint, \"nan_count\" bigint))"));
+                                "\"c\" row(\"min\" varchar, \"max\" varchar, \"null_count\" bigint, \"nan_count\" bigint))"),
```
findinpath, line 86: squash in the  previous commit pls

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
```diff
@@ -121,6 +122,13 @@ public PartitionsTable(SchemaTableName tableName, TypeManager typeManager, Table
             this.columnMetricTypes = ImmutableList.of();
         }
 
+        Stream.of(
```
findinpath:
```suggestion
        columnMetadataBuilder.add(new ColumnMetadata("position_delete_record_count", BIGINT));
        columnMetadataBuilder.add(new ColumnMetadata("position_delete_file_count", BIGINT));
        columnMetadataBuilder.add(new ColumnMetadata("equality_delete_record_count", BIGINT));
        columnMetadataBuilder.add(new ColumnMetadata("equality_delete_file_count", BIGINT));
```

nit consider rather the non-streamed approach

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -133,6 +148,27 @@ public void acceptDataFile(DataFile dataFile, PartitionSpec partitionSpec)
             }
         }
 
+        public void acceptDeleteFile(DeleteFile deleteFile)
+        {
+            String referenced = deleteFile.referencedDataFile();
+            String dedupKey = deleteFile.location() + "::" + (referenced == null ? "" : referenced);
```
findinpath: use a record as key instead of patching together a string key

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -133,6 +148,27 @@ public void acceptDataFile(DataFile dataFile, PartitionSpec partitionSpec)
...
+            if (!seenDeleteFiles.add(dedupKey)) {
+                return;
+            }
+            long records = deleteFile.recordCount();
+            switch (deleteFile.content()) {
+                case POSITION_DELETES -> {
+                    positionDeleteFileCount++;
+                    positionDeleteRecordCount += records;
```
findinpath, line 160: if we have one deletion vector applying for multiple partitions, won't it skew the values shown for the partition?

oskar-szwajkowski: I don't think such thing should be a concern:
```
A deletion vector must be applied to a data file when all of the following are true:
...
...
The data file's partition (both spec and partition values) is equal [4] to the deletion vector's partition
```

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergSystemTables.java
```diff
@@ -228,6 +241,42 @@ public void testPartitionsTableOnDropColumn()
...
+                .collect(toImmutableMap(row -> (LocalDate) row.getField(0), Function.identity()));
+
+        // DELETE FROM ... WHERE _date = '2022-02-02' AND _varchar = 'b2' removes one row.
+        assertThat((Long) rowsByPartition.get(LocalDate.parse("2022-02-02")).getField(1)).isPositive();
+        assertThat((Long) rowsByPartition.get(LocalDate.parse("2022-02-02")).getField(2)).isPositive();
+        // UPDATE on _date = '2022-01-01' rewrites one row, producing position deletes for the prior copy.
+        assertThat((Long) rowsByPartition.get(LocalDate.parse("2022-01-01")).getField(1)).isPositive();
+        assertThat((Long) rowsByPartition.get(LocalDate.parse("2022-01-01")).getField(2)).isPositive();
+        // Equality deletes are not produced by DML on Iceberg tables — they are written by external
+        // engines (e.g. Flink) — so the equality columns stay at zero here.
```
oskar-szwajkowski, line 268: Could you use `EqualityDeleteUtils` to write an equality delete file and test that it is reflected in `$partitions` table output?

docs/src/main/sphinx/connector/iceberg.md
1fanwang: Done in 8e560d03d36 — linked Spark `$partitions` from the description.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergPartitionEvolution.java
1fanwang: Squashed into the main commit (force-pushed as 8e560d03d36).

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/system/PartitionsTable.java
1fanwang: Done in 8e560d03d36.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
1fanwang: Done in 8e560d03d36 — added a `DeleteFileKey(String location, @Nullable String referencedDataFile)` record.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
1fanwang: Right — per the spec a DV is partition-scoped. The dedup set is also per-partition (one `IcebergStatistics.Builder` per partition struct in `getStatisticsByPartition`), so each entry counts at most once per partition.

plugin/trino-iceberg/src/test/java/io/trino/plugin/iceberg/BaseIcebergSystemTables.java
1fanwang: Done in 8e560d03d36 — added `testPartitionsTableEqualityDeleteMetrics` that writes an equality delete via `EqualityDeleteUtils.writeEqualityDeleteForTable` for one partition of a V2 table and asserts `equality_delete_*` is positive on that partition and `0` elsewhere.

1fanwang:
looks like flaky ci tests?
https://github.com/trinodb/trino/actions/runs/25124221187/job/73633504544?pr=29239

oskar-szwajkowski (approved):

oskar-szwajkowski:
please make sure to squash commits before merging

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
```diff
@@ -133,6 +148,27 @@ public void acceptDataFile(DataFile dataFile, PartitionSpec partitionSpec)
...
+                }
+                case EQUALITY_DELETES -> {
+                    equalityDeleteFileCount++;
+                    equalityDeleteRecordCount += records;
+                }
+                case DATA -> throw new IllegalStateException("Unexpected data file in deletes: " + deleteFile);
+            }
+        }
+
+        private record DeleteFileKey(String location, @Nullable String referencedDataFile) {}
```
findinpath, line 170: `@Nullable String referencedDataFile` -> `Optional<String>`

Add also null checks in the constructor

findinpath (approved):
LGTM % comment

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
raunaqmorarka: These new fields don't belong here. This class is intended for use in calculating table statistics for the CBO, it just happened to line up with what was needed in PartitionsTable a well.
We should have a new class io.trino.plugin.iceberg.system.PartitionsTableStatistics that is composed from IcebergStatistics and plus the addition fields here, and IcebergStatistics should stay unchanged.
On the whole, I prefer the direction of https://github.com/trinodb/trino/pull/28997 rather than continuing to add to `$partitions` in its current form because it carries risk of OOM on coordinator or high latency on large tables because it runs on coordinator-only mode and it doesn't make sense to repeat work of making it distributed like `$files`. Querying `$files` can serve as a workaround here even today (add fields there if you need to).

1fanwang:
After checking #28997, I'm convinced the view-over-`$files` direction is the right shape — closing this in favor of that path.

A few notes from the digging here, in case useful for whoever picks up the view conversion:

- The four delete columns map cleanly to a `GROUP BY partition` over `$files` filtered by `content`. `FilesTable` already exposes `content`, `record_count`, and `partition` per row, so the view definition is mechanical — no new aggregation code in the connector.
- Spark's `$partitions` parity is wider than these four columns: it also has `spec_id`, `last_updated_at`, and `last_updated_snapshot_id`. The view direction picks those up naturally via joins on `$snapshots`, which is a much cleaner story than incrementally adding columns to the single-coordinator path.
- `TableStatisticsReader` only reads the data-file fields (`recordCount`, `fileCount`, `size`, `minValues`/`maxValues`/`nullCounts`/`columnSizes`) — it never calls `acceptDeleteFile`. CBO and partition-table concerns are mixed into one class today, and the view conversion lets that separation happen for free.
- The dedup-by-`(location, referencedDataFile)` logic in `acceptDeleteFile` exists because the same DV can appear in many `FileScanTask.deletes()` lists. In the view world this dedup is implicit — one delete file is one row in `$files` regardless of how many data files reference it — so that complexity disappears too.

For users on #28910 who need the data today, the workaround is short:

```sql
SELECT partition,
       count(*) FILTER (WHERE content = 1) AS position_delete_file_count,
       sum(record_count) FILTER (WHERE content = 1) AS position_delete_record_count,
       count(*) FILTER (WHERE content = 2) AS equality_delete_file_count,
       sum(record_count) FILTER (WHERE content = 2) AS equality_delete_record_count
FROM "tbl$files"
WHERE content IN (1, 2)
GROUP BY partition
```

We could instead send a short docs PR adding this to the iceberg connector page so the workaround is discoverable while #28997 lands.

Thanks @findinpath and @oskar-szwajkowski for the review, even if the diff isn't landing in this shape. And @raunaqmorarka, thanks for the architectural design comment.

plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergStatistics.java
1fanwang: thanks @raunaqmorarka , closed with comment https://github.com/trinodb/trino/pull/29239#issuecomment-4362548858

Issue #20270, opened 2024-01-03
Unable to query on tables without full read access in MySQL connector


AndresLostaunau:
I'm using Trino 435 on docker and I found an issue that whenever I try to query a table on a MySQL DB where I don't have read access on every column (on the MySQL end, not by Trino AccessControl) I get denied, even when having read access in all the columns I'm doing the SELECT on. It prompts the following error:

Failed to get case sensitivity for columns. SELECT command denied to user '<db_user>@'<db_host>' for table '<table>'.

The second sentence is what MySQL prompts when trying to read an inaccessible column, but I can't find any documentation regarding first sentence. It's worth noting that I can read all other tables just fine, and that this isn't an issue querying with other applications where I use the same credentials, and also that the columns don't have any type of uppercase letters. My configuration is pretty much the one the page guide offers.

ebyhr:
cc: @wendigo @hashhar

hashhar:
Thanks for the bug report @AndresLostaunau. Unfortunately there is no workaround for this (other than allowing access on MySQL end and doing the access control on Trino side) and will require a code change.

hhsnow:
Adding a data point in case the cause of the bug is unknown -- SELECTS with specific column grants worked in release 419. I haven't tried versions between 419-435 to determine the release that it broke in.

PR #29244, opened 2026-04-27
Fix column-level permissions in SQL Server connector

LBPeraza:
## Description

When the connection user has a DENY permission on some columns in a particular table, any query from that table fails, even when querying only allowed columns. Errors look like:

```
Failed to get case sensitivity for columns.
The SELECT permission was denied on the column 'sensitive_data'
```

This happens because `getCaseSensitivityForColumns` tries to get metadata for the query `SELECT * FROM <table>`.

To fix this, we can query `sys.columns` instead.

- https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-columns-transact-sql
- Same issue in MySQL connector: https://github.com/trinodb/trino/issues/20270

## Release notes

```markdown
## SQL Server connector
* Fix permission failure when reading tables. ({issue}`29244`)
```

### DIFF
--- a/plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
+++ b/plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
@@ -80,7 +80,6 @@ import io.trino.spi.connector.JoinType;
 import io.trino.spi.connector.RelationColumnsMetadata;
 import io.trino.spi.connector.RelationCommentMetadata;
 import io.trino.spi.connector.SchemaTableName;
-import io.trino.spi.connector.TableNotFoundException;
 import io.trino.spi.expression.ConnectorExpression;
 import io.trino.spi.predicate.Domain;
 import io.trino.spi.predicate.Range;
@@ -137,6 +136,7 @@ import java.util.stream.Stream;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.base.Throwables.throwIfInstanceOf;
+import static com.google.common.collect.ImmutableMap.toImmutableMap;
 import static com.google.common.collect.MoreCollectors.toOptional;
 import static com.google.common.util.concurrent.MoreExecutors.directExecutor;
 import static io.airlift.slice.Slices.utf8Slice;
@@ -571,27 +571,25 @@ public class SqlServerClient
     @Override
     protected Map<String, CaseSensitivity> getCaseSensitivityForColumns(ConnectorSession session, Connection connection, SchemaTableName schemaTableName, RemoteTableName remoteTableName)
     {
-        PreparedQuery preparedQuery = new PreparedQuery(format("SELECT * FROM %s", quoted(remoteTableName)), ImmutableList.of());
+        return Jdbi.open(connection).createQuery("""
+                                                 SELECT c.name AS column_name, c.collation_name FROM sys.columns c
+                                                 INNER JOIN sys.tables t ON c.object_id = t.object_id
+                                                 INNER JOIN sys.schemas s ON s.schema_id = t.schema_id
+                                                 WHERE s.name = :schema_name AND t.name = :table_name
+                                                 """)
+                .bind("schema_name", remoteTableName.getSchemaName().orElseThrow())
+                .bind("table_name", remoteTableName.getTableName())
+                .collectRows(toImmutableMap(rowView -> rowView.getColumn("column_name", String.class),
+                        rowView -> toCaseSensitivity(rowView.getColumn("collation_name", String.class))));
+    }
 
-        try (PreparedStatement preparedStatement = queryBuilder.prepareStatement(this, session, connection, preparedQuery, Optional.empty())) {
-            ResultSetMetaData metadata = preparedStatement.getMetaData();
-            ImmutableMap.Builder<String, CaseSensitivity> columns = ImmutableMap.builder();
-            for (int column = 1; column <= metadata.getColumnCount(); column++) {
-                String name = metadata.getColumnName(column);
-                columns.put(name, metadata.isCaseSensitive(column) ? CASE_SENSITIVE : CASE_INSENSITIVE);
-            }
-            return columns.buildOrThrow();
+    private static CaseSensitivity toCaseSensitivity(String collation)
+    {
+        if (collation == null || collation.isEmpty()) {
+            return CASE_INSENSITIVE;
         }
-        catch (SQLException e) {
-            if (e instanceof SQLServerException sqlServerException && sqlServerException.getSQLServerError().getErrorNumber() == 208) {
-                // The 208 indicates that the object doesn't exist or lack of permission.
-                // Throw TableNotFoundException because users shouldn't see such tables if they don't have the permission.
-                // TableNotFoundException will be suppressed when listing information_schema.
-                // https://learn.microsoft.com/sql/relational-databases/errors-events/mssqlserver-208-database-engine-error
-                throw new TableNotFoundException(schemaTableName);
-            }
-            throw new TrinoException(JDBC_ERROR, "Failed to get case sensitivity for columns. " + requireNonNullElse(e.getMessage(), e), e);
-        }
+        // https://learn.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support
+        return collation.contains("_BIN") || collation.contains("_BIN2") || collation.contains("_CS") ? CASE_SENSITIVE : CASE_INSENSITIVE;
     }
 
     @Override
--- a/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
+++ b/plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
@@ -1,0 +1,83 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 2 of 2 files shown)


plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
```diff
@@ -571,27 +571,24 @@ protected void copyTableSchema(ConnectorSession session, Connection connection,
     @Override
     protected Map<String, CaseSensitivity> getCaseSensitivityForColumns(ConnectorSession session, Connection connection, SchemaTableName schemaTableName, RemoteTableName remoteTableName)
     {
-        PreparedQuery preparedQuery = new PreparedQuery(format("SELECT * FROM %s", quoted(remoteTableName)), ImmutableList.of());
+        return Jdbi.open(connection).createQuery("""
```
raphaelsolarski, line 574: this handle from `Jdbi.open(Connection)` should not be closed?

LBPeraza: I think `collectRows` closes the handle when it's done, but it should also be safe to close it more than once so I'll add it to make it clearer (and in case `collectRows` doesn't in fact close its handle).

raphaelsolarski: handle.close() afaik closes also connection which in your case comes outside the method so for me it wasn't so obvious.

LBPeraza: Yep, it seems like it does - explicitly closing the handle makes some of the other tests fail. Reverted.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
```diff
@@ -0,0 +1,73 @@
...
+import io.trino.testing.AbstractTestQueryFramework;
+import io.trino.testing.QueryRunner;
+import io.trino.testing.sql.TestTable;
+import org.junit.jupiter.api.Test;
+
+import java.util.Map;
+
+import static io.trino.testing.TestingNames.randomNameSuffix;
+
+public class TestSqlServerColumnLevelPermissions
```
ebyhr: https://trino.io/docs/current/develop/tests.html
> * Test classes should be defined as package-private and final.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
```diff
@@ -0,0 +1,73 @@
...
+
+        return SqlServerQueryRunner.builder(sqlServer)
+                .addConnectorProperties(Map.of(
+                        "connection-user", limitedUser,
+                        "connection-password", password))
+                .build();
+    }
+
+    @Test
+    public void testSelectPermittedColumn()
```
ebyhr: https://trino.io/docs/current/develop/tests.html
> * Test methods should be defined as package-private.

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
```diff
@@ -571,27 +571,24 @@ protected void copyTableSchema(ConnectorSession session, Connection connection,
...
-        catch (SQLException e) {
-            if (e instanceof SQLServerException sqlServerException && sqlServerException.getSQLServerError().getErrorNumber() == 208) {
-                // The 208 indicates that the object doesn't exist or lack of permission.
-                // Throw TableNotFoundException because users shouldn't see such tables if they don't have the permission.
-                // TableNotFoundException will be suppressed when listing information_schema.
-                // https://learn.microsoft.com/sql/relational-databases/errors-events/mssqlserver-208-database-engine-error
-                throw new TableNotFoundException(schemaTableName);
-            }
-            throw new TrinoException(JDBC_ERROR, "Failed to get case sensitivity for columns. " + requireNonNullElse(e.getMessage(), e), e);
+    private static CaseSensitivity getCaseSensitivityForCollation(String collation)
```
ebyhr:
```suggestion
    private static CaseSensitivity toCaseSensitivity(String collation)
```

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
```diff
@@ -571,27 +571,24 @@ protected void copyTableSchema(ConnectorSession session, Connection connection,
     @Override
     protected Map<String, CaseSensitivity> getCaseSensitivityForColumns(ConnectorSession session, Connection connection, SchemaTableName schemaTableName, RemoteTableName remoteTableName)
     {
-        PreparedQuery preparedQuery = new PreparedQuery(format("SELECT * FROM %s", quoted(remoteTableName)), ImmutableList.of());
+        return Jdbi.open(connection).createQuery("""
+                                                 SELECT c.name AS column_name, c.collation_name FROM sys.columns c
```
ebyhr, line 575: What if a collation is configured at the database level? Does `sys.columns` still return the expected value?

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
```diff
@@ -571,27 +571,24 @@ protected void copyTableSchema(ConnectorSession session, Connection connection,
...
-                // https://learn.microsoft.com/sql/relational-databases/errors-events/mssqlserver-208-database-engine-error
-                throw new TableNotFoundException(schemaTableName);
-            }
-            throw new TrinoException(JDBC_ERROR, "Failed to get case sensitivity for columns. " + requireNonNullElse(e.getMessage(), e), e);
+    private static CaseSensitivity getCaseSensitivityForCollation(String collation)
+    {
+        if (collation == null || collation.isEmpty()) {
+            return CASE_INSENSITIVE;
         }
+        return collation.endsWith("BIN") || collation.endsWith("BIN2") || collation.contains("_CS_") ? CASE_SENSITIVE : CASE_INSENSITIVE;
```
ebyhr: Can we add a link to the SQLServer's documentation? e.g. https://learn.microsoft.com/en-us/sql/relational-databases/collations/collation-and-unicode-support

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
```diff
@@ -0,0 +1,73 @@
...
+    {
+        assertQuery("SELECT allowed_column FROM " + testTable.getName(), "VALUES ('foo')");
```
ebyhr, line 59: DESCRIBE statement contain denied_column column, right? Could you add another assertion with DESCRIBE just in case?

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
```diff
@@ -571,27 +571,24 @@ protected void copyTableSchema(ConnectorSession session, Connection connection,
...
+        return Jdbi.open(connection).createQuery("""
+                                                 SELECT c.name AS column_name, c.collation_name FROM sys.columns c
+                                                 INNER JOIN sys.tables t ON c.object_id = t.object_id
+                                                 INNER JOIN sys.schemas s ON s.schema_id = t.schema_id
+                                                 WHERE s.name = :schema_name AND t.name = :table_name
+                                                 """)
+                    .bind("schema_name", remoteTableName.getSchemaName().orElseThrow())
+                    .bind("table_name", remoteTableName.getTableName())
+                    .collectRows(toImmutableMap(rowView -> rowView.getColumn("column_name", String.class),
+                            rowView -> getCaseSensitivityForCollation(rowView.getColumn("collation_name", String.class))));
```
ebyhr: The indentation looks wrong.

ebyhr (approved):

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
LBPeraza: Yes, `DESCRIBE` still returns the denied columns. Added a test for that.

plugin/trino-sqlserver/src/main/java/io/trino/plugin/sqlserver/SqlServerClient.java
LBPeraza: Yes, the `collation_name` in `sys.columns` is only NULL if the column isn't character based - https://learn.microsoft.com/en-us/sql/relational-databases/system-catalog-views/sys-columns-transact-sql

LBPeraza:
@ebyhr thanks for the review - I've responded to all the comments so this should be ready to merge.

plugin/trino-sqlserver/src/test/java/io/trino/plugin/sqlserver/TestSqlServerColumnLevelPermissions.java
```diff
@@ -0,0 +1,73 @@
...
+    private TestTable testTable;
+
+    @Override
+    protected QueryRunner createQueryRunner()
+            throws Exception
+    {
+        TestingSqlServer sqlServer = closeAfterClass(new TestingSqlServer());
+
+        String limitedUser = "limited_user" + randomNameSuffix();
+        String password = "A_sTr0nG_p4s$word";
```
chenjian2664, line 38: is it intentional?

PR #29245, opened 2026-04-27
Support extensibility through newHandle factory method on JdbcTableHandle

findinpath:
## Release notes

```markdown
## Section
```

### DIFF
--- a/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
+++ b/plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
@@ -36,7 +36,7 @@ import static com.google.common.base.Preconditions.checkState;
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
-public final class JdbcTableHandle
+public class JdbcTableHandle
         extends BaseJdbcConnectorTableHandle
 {
     private final JdbcRelationHandle relationHandle;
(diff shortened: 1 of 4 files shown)


plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
```diff
@@ -36,7 +36,7 @@
 import static com.google.common.collect.ImmutableList.toImmutableList;
 import static java.util.Objects.requireNonNull;
 
-public final class JdbcTableHandle
```
chenjian2664, line 39: Could you clarify if this change is necessary, I'm wondering about the reasoning behind it

plugin/trino-base-jdbc/src/main/java/io/trino/plugin/jdbc/JdbcTableHandle.java
```diff
@@ -105,9 +105,34 @@ public JdbcTableHandle(
...
+            int nextSyntheticColumnId,
+            Optional<String> authorization,
+            List<JdbcAssignmentItem> updateAssignments)
     {
         return new JdbcTableHandle(
+                relationHandle,
+                constraint,
+                constraintExpressions,
+                sortOrder,
+                limit,
```
wendigo, line 125: you are not using any of the existing fields from JdbcTableHandle so why is this a method instead of a static one?

findinpath:
Closing the PR because this doesn't seem to be a good idea.

PR #29246, opened 2026-04-27
Use DELTA_LENGTH_BYTE_ARRAY encoding in the parquet writer

raunaqmorarka:
## Summary

Make the Trino Parquet writer use `DELTA_LENGTH_BYTE_ARRAY` (Parquet encoding 6) instead of `PLAIN` as the non-dictionary fallback for `BYTE_ARRAY` columns, behind a new `parquet.writer.delta-length-byte-array-encoding-enabled` config flag. The Parquet specification recommends this encoding over `PLAIN`: lengths are encoded with `DELTA_BINARY_PACKED` and the byte data is concatenated separately, typically yielding smaller encoded size and better post-compression ratio.

## Default per connector

| Connector | Default | Override |
|---|---|---|
| Hive | enabled | `parquet.writer.delta-length-byte-array-encoding-enabled` config property |
| Delta Lake | enabled | same |
| Iceberg | disabled¹ | per-catalog config; per-query session property `iceberg.parquet_writer_delta_length_byte_array_encoding_enabled` |

¹ The Iceberg connector overrides the default to `false` because iceberg-arrow's vectorized parquet reader does not support `DELTA_LENGTH_BYTE_ARRAY` pages in iceberg ≤ 1.10.x. The fix is in [apache/iceberg#15362](https://github.com/apache/iceberg/pull/15362), expected to land in iceberg 1.11+. The default should be flipped to `true` for Iceberg once the dependency is bumped — `TestIcebergSparkCompatibility#testTrinoSparkParquetDeltaLengthByteArrayCompatibility` is a sentinel that will start failing when the iceberg-arrow path supports the encoding.

Note: Some versions of Apache Spark contain a bug in their vectorized parquet reader which cause failure to read columns with DELTA_LENGTH_BYTE_ARRAY encoding, this has been fixed by https://issues.apache.org/jira/browse/SPARK-52240 in versions 4.1.0, 4.0.1 and 3.5.7

### `BenchmarkBinaryColumnReader` — read throughput, ops/s

Realistic data shapes only (the synthetic CHAR-with-padding cases are absent from typical workloads).

| positionLength | type | data shape | PLAIN | DELTA_LENGTH | speedup |
|---|---|---|--:|--:|--:|
| VARIABLE_0_100 | UNBOUNDED | unbounded binary, length ≤ 100 B | 64.9 | 452.9 | **6.97×** |
| VARIABLE_0_100 | VARCHAR_ASCII_BOUND_EXACT | bounded VARCHAR, length ≤ 100 chars | 55.7 | 505.8 | **9.08×** |
| VARIABLE_0_1000 | UNBOUNDED | unbounded binary, length ≤ 1000 B | 15,976 | 182,538 | **11.42×** |
| VARIABLE_0_1000 | VARCHAR_ASCII_BOUND_EXACT | bounded VARCHAR, length ≤ 1000 chars | 14,440 | 205,584 | **14.24×** |
| FIXED_10 | VARCHAR_ASCII_BOUND_EXACT | bounded VARCHAR, fixed length 10 chars | 92 | 1,006 | **10.92×** |
| FIXED_100 | VARCHAR_ASCII_BOUND_EXACT | bounded VARCHAR, fixed length 100 chars | 4,189 | 121,255 | **28.95×** |

20 samples per combo (5 warmup + 10 measurement iterations × 2 forks).

### `BenchmarkParquetFormat` — write LINEITEM with SNAPPY

| `useDeltaLengthByteArrayEncoding` | ops/s | bytes/file | MB/s |
|---|--:|--:|--:|
| `true` (DELTA_LENGTH) | **5.97 ± 0.45** | 12,326,852 | **298.4 ± 22.4** |
| `false` (PLAIN) | 4.86 ± 0.98 | 12,738,816 | 243.1 ± 49.1 |

20 samples (10 warmup + 10 measurement iterations × 2 forks). +23% throughput, −3% file size on the write path.

## Release notes

```markdown
## Hive connector
* Improve compression and read performance for high-cardinality string columns in Parquet files by using `DELTA_LENGTH_BYTE_ARRAY` encoding. This can be disabled via the `parquet.writer.delta-length-byte-array-encoding-enabled` config property. ({issue}`29246`)

## Delta Lake connector
* Improve compression and read performance for high-cardinality string columns in Parquet files by using `DELTA_LENGTH_BYTE_ARRAY` encoding. This can be disabled via the `parquet.writer.delta-length-byte-array-encoding-enabled` config property. ({issue}`29246`)
```

### DIFF
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriterOptions.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriterOptions.java
@@ -33,6 +33,7 @@ public class ParquetWriterOptions
     public static final int DEFAULT_BATCH_SIZE = 10_000;
     public static final DataSize DEFAULT_MAX_BLOOM_FILTER_SIZE = DataSize.of(1, MEGABYTE);
     public static final double DEFAULT_BLOOM_FILTER_FPP = 0.05;
+    public static final boolean DEFAULT_USE_DELTA_LENGTH_BYTE_ARRAY_ENCODING = true;
 
     public static ParquetWriterOptions.Builder builder()
     {
@@ -47,6 +48,7 @@ public class ParquetWriterOptions
     private final double bloomFilterFpp;
     // Set of column dot paths to columns with bloom filters
     private final Set<String> bloomFilterColumns;
+    private final boolean useDeltaLengthByteArrayEncoding;
 
     private ParquetWriterOptions(
             DataSize maxBlockSize,
@@ -55,7 +57,8 @@ public class ParquetWriterOptions
             int batchSize,
             DataSize maxBloomFilterSize,
             double bloomFilterFpp,
-            Set<String> bloomFilterColumns)
+            Set<String> bloomFilterColumns,
+            boolean useDeltaLengthByteArrayEncoding)
     {
         this.maxRowGroupSize = Ints.saturatedCast(maxBlockSize.toBytes());
         this.maxPageSize = Ints.saturatedCast(maxPageSize.toBytes());
@@ -64,6 +67,7 @@ public class ParquetWriterOptions
         this.maxBloomFilterSize = Ints.saturatedCast(maxBloomFilterSize.toBytes());
         this.bloomFilterFpp = bloomFilterFpp;
         this.bloomFilterColumns = ImmutableSet.copyOf(bloomFilterColumns);
+        this.useDeltaLengthByteArrayEncoding = useDeltaLengthByteArrayEncoding;
         checkArgument(this.bloomFilterFpp > 0.0 && this.bloomFilterFpp < 1.0, "bloomFilterFpp should be > 0.0 & < 1.0");
     }
 
@@ -102,6 +106,11 @@ public class ParquetWriterOptions
         return bloomFilterFpp;
     }
 
+    public boolean isUseDeltaLengthByteArrayEncoding()
+    {
+        return useDeltaLengthByteArrayEncoding;
+    }
+
     public static class Builder
     {
         private DataSize maxBlockSize = DEFAULT_MAX_ROW_GROUP_SIZE;
@@ -111,6 +120,7 @@ public class ParquetWriterOptions
         private DataSize maxBloomFilterSize = DEFAULT_MAX_BLOOM_FILTER_SIZE;
         private Set<String> bloomFilterColumns = ImmutableSet.of();
         private double bloomFilterFpp = DEFAULT_BLOOM_FILTER_FPP;
+        private boolean useDeltaLengthByteArrayEncoding = DEFAULT_USE_DELTA_LENGTH_BYTE_ARRAY_ENCODING;
 
         public Builder setMaxBlockSize(DataSize maxBlockSize)
         {
@@ -154,6 +164,12 @@ public class ParquetWriterOptions
             return this;
         }
 
+        public Builder setUseDeltaLengthByteArrayEncoding(boolean useDeltaLengthByteArrayEncoding)
+        {
+            this.useDeltaLengthByteArrayEncoding = useDeltaLengthByteArrayEncoding;
+            return this;
+        }
+
         public ParquetWriterOptions build()
         {
             return new ParquetWriterOptions(
@@ -163,7 +179,8 @@ public class ParquetWriterOptions
                     batchSize,
                     maxBloomFilterSize,
                     bloomFilterFpp,
-                    bloomFilterColumns);
+                    bloomFilterColumns,
+                    useDeltaLengthByteArrayEncoding);
         }
     }
 }
--- a/lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriters.java
+++ b/lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriters.java
@@ -182,7 +182,7 @@ final class ParquetWriters
             ParquetWriterOptions writerOptions,
             Optional<DateTimeZone> parquetTimeZone)
     {
-        TrinoValuesWriterFactory valuesWriterFactory = new TrinoValuesWriterFactory(writerOptions.getMaxPageSize(), DEFAULT_DICTIONARY_PAGE_SIZE);
+        TrinoValuesWriterFactory valuesWriterFactory = new TrinoValuesWriterFactory(writerOptions, DEFAULT_DICTIONARY_PAGE_SIZE);
         WriteBuilder writeBuilder = new WriteBuilder(
                 messageType,
                 trinoTypes,
--- a/lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
+++ b/lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
@@ -53,7 +53,6 @@ import java.util.OptionalLong;
 import static io.airlift.slice.Slices.EMPTY_SLICE;
 import static io.trino.memory.context.AggregatedMemoryContext.newSimpleAggregatedMemoryContext;
 import static io.trino.parquet.ParquetEncoding.BIT_PACKED;
-import static io.trino.parquet.ParquetEncoding.PLAIN;
 import static io.trino.parquet.ParquetEncoding.PLAIN_DICTIONARY;
 import static io.trino.parquet.ParquetEncoding.RLE;
 import static io.trino.parquet.ParquetEncoding.RLE_DICTIONARY;
@@ -239,13 +238,13 @@ public abstract class AbstractColumnReaderTest
         PrimitiveField field = createField(format, true);
         ColumnReader reader = createColumnReader(field);
         // Write data
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getValuesWriter(version);
         T[] values1 = format.write(writer, new Integer[] {1});
-        DataPage page1 = createDataPage(version, PLAIN, writer, 1);
+        DataPage page1 = createDataPage(version, writer, 1);
         T[] values2 = format.resetAndWrite(writer, new Integer[] {2, 3});
-        DataPage page2 = createDataPage(version, PLAIN, writer, 2);
+        DataPage page2 = createDataPage(version, writer, 2);
         T[] values3 = format.resetAndWrite(writer, new Integer[] {4});
-        DataPage page3 = createDataPage(version, PLAIN, writer, 1);
+        DataPage page3 = createDataPage(version, writer, 1);
         // Read and assert
         reader.setPageReader(getPageReaderMock(List.of(page1, page2, page3), null), Optional.empty());
         Block actual1 = readBlock(reader, 1); // Parquet/Trino page size the same
@@ -271,11 +270,11 @@ public abstract class AbstractColumnReaderTest
         ColumnReader reader = createColumnReader(field);
         // Write data
         DictionaryValuesWriter dictionaryWriter = format.getDictionaryWriter();
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getValuesWriter(version);
         T[] values1 = format.write(dictionaryWriter, new Integer[] {1, 1});
         DataPage page1 = createDataPage(version, RLE_DICTIONARY, dictionaryWriter, 2);
         T[] values2 = format.write(writer, new Integer[] {2});
-        DataPage page2 = createDataPage(version, PLAIN, writer, 1);
+        DataPage page2 = createDataPage(version, writer, 1);
         DictionaryPage dictionaryPage = getDictionaryPage(dictionaryWriter);
         // Read and assert
         reader.setPageReader(getPageReaderMock(List.of(page1, page2), dictionaryPage), Optional.empty());
@@ -296,11 +295,11 @@ public abstract class AbstractColumnReaderTest
         PrimitiveField field = createField(format, false);
         ColumnReader reader = createColumnReader(field);
         // Write data
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getValuesWriter(version);
         T[] values1 = format.write(writer, new Integer[] {null});
-        DataPage page1 = createNullableDataPage(version, PLAIN, writer, field, true);
+        DataPage page1 = createNullableDataPage(version, writer, field, true);
         T[] values2 = format.write(writer, new Integer[] {null, null});
-        DataPage page2 = createNullableDataPage(version, PLAIN, writer, field, true, true);
+        DataPage page2 = createNullableDataPage(version, writer, field, true, true);
         // Read and assert
         reader.setPageReader(getPageReaderMock(List.of(page1, page2), null), Optional.empty());
         // Deliberate mismatch between Trino/Parquet page sizes
@@ -324,13 +323,13 @@ public abstract class AbstractColumnReaderTest
         PrimitiveField field = createField(format, false);
         ColumnReader reader = createColumnReader(field);
         // Write data
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getValuesWriter(version);
         T[] values1 = format.write(writer, new Integer[] {null, 1});
-        DataPage page1 = createNullableDataPage(version, PLAIN, writer, field, true, false);
+        DataPage page1 = createNullableDataPage(version, writer, field, true, false);
         T[] values2 = format.resetAndWrite(writer, new Integer[] {null, 2, null});
-        DataPage page2 = createNullableDataPage(version, PLAIN, writer, field, true, false, true);
+        DataPage page2 = createNullableDataPage(version, writer, field, true, false, true);
         T[] values3 = format.resetAndWrite(writer, new Integer[] {3, null, 4});
-        DataPage page3 = createNullableDataPage(version, PLAIN, writer, field, false, true, false);
+        DataPage page3 = createNullableDataPage(version, writer, field, false, true, false);
         // Read and assert
         reader.setPageReader(getPageReaderMock(List.of(page1, page2, page3), null), Optional.empty());
         // Deliberate mismatch between Trino/Parquet page sizes
@@ -401,9 +400,9 @@ public abstract class AbstractColumnReaderTest
         PrimitiveField field = createField(format, false);
         ColumnReader reader = createColumnReader(field);
         // Write data
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getValuesWriter(version);
         T[] values1 = format.write(writer, new Integer[] {1, 2});
-        DataPage page1 = createNullableDataPage(version, PLAIN, writer, field, false, false);
+        DataPage page1 = createNullableDataPage(version, writer, field, false, false);
         // Read and assert
         reader.setPageReader(getPageReaderMock(List.of(page1), null), Optional.empty());
         // Deliberate mismatch between Trino/Parquet page sizes
@@ -422,9 +421,9 @@ public abstract class AbstractColumnReaderTest
         PrimitiveField field = createField(format, false);
         ColumnReader reader = createColumnReader(field);
         // Write data
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getValuesWriter(version);
         T[] values1 = format.write(writer, new Integer[] {1, 2});
-        DataPage page1 = createNullableDataPage(version, PLAIN, writer, field, false, false);
+        DataPage page1 = createNullableDataPage(version, writer, field, false, false);
         // Read and assert
         reader.setPageReader(getPageReaderMock(List.of(page1), null, true), Optional.empty());
         // Deliberate mismatch between Trino/Parquet page sizes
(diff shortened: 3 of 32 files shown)


raunaqmorarka:
@coderabbitai review

lib/trino-parquet/src/test/java/io/trino/parquet/writer/TestParquetWriter.java
```diff
@@ -606,6 +609,51 @@ void testBloomFilterWithDictionaryFallback()
...
+
+        // 50,000 UUIDs (~1.8 MB raw) exceed the 1 MB dictionary page size and force fallback
+        // out of the dictionary encoder, which writes DELTA_LENGTH_BYTE_ARRAY pages.
+        ParquetDataSource dataSource = new TestingParquetDataSource(
+                writeParquetFile(
+                        ParquetWriterOptions.builder().build(),
+                        types,
+                        columnNames,
+                        generateInputPages(types, 1000, sliceValues)),
+                ParquetReaderOptions.defaultOptions());
```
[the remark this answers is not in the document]
raunaqmorarka: Set `setUseDeltaLengthByteArrayEncoding(true)` on the writer options so the test intent is independent of the default.

testing/trino-product-tests/src/main/java/io/trino/tests/product/hive/TestHiveSparkCompatibility.java
```diff
@@ -361,6 +362,44 @@ public void testTrinoSparkParquetBloomFilterCompatibility()
...
+            // Force the DELTA_LENGTH_BYTE_ARRAY fallback by inserting many high-cardinality
+            // varchar/varbinary values. ~5000 rows of UUID-derived strings overflow the default
+            // 1 MB dictionary page and trigger the fallback path.
+            StringBuilder values = new StringBuilder();
+            for (int i = 0; i < 5000; i++) {
+                String uuid = UUID.randomUUID().toString();
+                if (i > 0) {
+                    values.append(", ");
+                }
+                values.append(format("(%d, '%s', X'%s')", i, uuid, uuid.replace("-", "")));
```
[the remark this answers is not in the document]
raunaqmorarka: 5000 UUIDs is sufficient. Fallback fires at first-page flush via the per-page compression-ratio check in `DictionaryFallbackValuesWriter.getBytes`, not the 1 MB dictionary cap; for unique values the dict-encoded size matches raw size and the check fails immediately. Empirical threshold is ~2000 values, pinned by `TestParquetWriter#testDeltaLengthByteArrayFallbackIsWritten`, which uses the same UUID generator and asserts `DELTA_LENGTH_BYTE_ARRAY` appears in the column-chunk encodings.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeInsertCompatibility.java
```diff
@@ -436,6 +438,43 @@ public Object[][] compressionCodecs()
...
+            // UUID-derived rows. The unique values overflow the default 1 MB dictionary page
+            // and trigger the encoding fallback path.
+            StringBuilder values = new StringBuilder();
+            for (int i = 0; i < 5000; i++) {
+                String uuid = UUID.randomUUID().toString();
+                if (i > 0) {
+                    values.append(", ");
+                }
+                values.append(format("(%d, '%s', X'%s')", i, uuid, uuid.replace("-", "")));
+            }
```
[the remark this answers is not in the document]
raunaqmorarka: 5000 UUIDs is sufficient. Fallback fires at first-page flush via the per-page compression-ratio check in `DictionaryFallbackValuesWriter.getBytes`, not the 1 MB dictionary cap; for unique values the dict-encoded size matches raw size and the check fails immediately. Empirical threshold is ~2000 values, pinned by `TestParquetWriter#testDeltaLengthByteArrayFallbackIsWritten`. Also added `count(DISTINCT bin)` so the binary read path is decoded too.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -3584,6 +3585,38 @@ public void testTrinoReadingSparkParquetBloomFilters()
...
+
+        // Force DELTA_LENGTH_BYTE_ARRAY fallback with high-cardinality UUID-derived data
+        // overflowing the default 1 MB dictionary page cap.
+        StringBuilder values = new StringBuilder();
+        for (int i = 0; i < 5000; i++) {
+            String uuid = UUID.randomUUID().toString();
+            if (i > 0) {
+                values.append(", ");
+            }
+            values.append(format("(%d, '%s', X'%s')", i, uuid, uuid.replace("-", "")));
```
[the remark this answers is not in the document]
raunaqmorarka: 5000 UUIDs is sufficient. Fallback fires at first-page flush via the per-page compression-ratio check in `DictionaryFallbackValuesWriter.getBytes`, not the 1 MB dictionary cap; for unique values the dict-encoded size matches raw size and the check fails immediately. Empirical threshold is ~2000 values, pinned by `TestParquetWriter#testDeltaLengthByteArrayFallbackIsWritten`. Also added `count(DISTINCT bin)` so the binary read path is decoded too.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -3584,6 +3585,38 @@ public void testTrinoReadingSparkParquetBloomFilters()
...
+        }
+        onTrino().executeQuery(format("INSERT INTO %s VALUES %s", trinoTableName, values));
+
+        assertThat(onSpark().executeQuery(format("SELECT count(*) FROM %s", sparkTableName)))
+                .containsOnly(row(5000L));
+        assertThat(onSpark().executeQuery(format("SELECT count(DISTINCT str) FROM %s", sparkTableName)))
+                .containsOnly(row(5000L));
+
+        onTrino().executeQuery("DROP TABLE " + trinoTableName);
```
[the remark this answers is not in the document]
raunaqmorarka: Wrapped the test body in try/finally with `DROP TABLE` cleanup, matching the convention used by other tests in this file.

testing/trino-product-tests/src/main/java/io/trino/tests/product/iceberg/TestIcebergSparkCompatibility.java
```diff
@@ -3584,6 +3585,38 @@ public void testTrinoReadingSparkParquetBloomFilters()
...
+                values.append(", ");
+            }
+            values.append(format("(%d, '%s', X'%s')", i, uuid, uuid.replace("-", "")));
+        }
+        onTrino().executeQuery(format("INSERT INTO %s VALUES %s", trinoTableName, values));
+
+        assertThat(onSpark().executeQuery(format("SELECT count(*) FROM %s", sparkTableName)))
+                .containsOnly(row(5000L));
+        assertThat(onSpark().executeQuery(format("SELECT count(DISTINCT str) FROM %s", sparkTableName)))
+                .containsOnly(row(5000L));
```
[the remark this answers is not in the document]
raunaqmorarka: 5000 UUIDs is sufficient. Fallback fires at first-page flush via the per-page compression-ratio check in `DictionaryFallbackValuesWriter.getBytes`, not the 1 MB dictionary cap; for unique values the dict-encoded size matches raw size and the check fails immediately. Empirical threshold is ~2000 values, pinned by `TestParquetWriter#testDeltaLengthByteArrayFallbackIsWritten`. Added `count(DISTINCT bin)` so the binary read path is decoded too.

raunaqmorarka:
/test-with-secrets sha=7e334b9826dce312f0f2a8296f979347b24bbfce

findepi:
> Iceberg	**forced off**	iceberg-arrow's vectorized parquet reader fails on `DELTA_LENGTH_BYTE_ARRAY` pages in iceberg ≤ 1.10.x; fix in [apache/iceberg#15362](https://github.com/apache/iceberg/pull/15362), expected in 1.11+. Re-enable once the dependency is bumped.

what is the motivation for a  bug in iceberg-arrow's reader to be a hard  blocker for this feature, especially that it is an opt-in thing?

lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriterOptions.java
```diff
@@ -111,6 +119,7 @@ public static class Builder
         private DataSize maxBloomFilterSize = DEFAULT_MAX_BLOOM_FILTER_SIZE;
         private Set<String> bloomFilterColumns = ImmutableSet.of();
         private double bloomFilterFpp = DEFAULT_BLOOM_FILTER_FPP;
+        private boolean useDeltaLengthByteArrayEncoding = true;
```
findepi: not sure why we have these DEFAULT_ constants, but might be worth aligning with pre-existing style and defining such for the new config

lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriters.java
```diff
@@ -182,7 +182,10 @@ static List<ColumnWriter> getColumnWriters(
             ParquetWriterOptions writerOptions,
             Optional<DateTimeZone> parquetTimeZone)
     {
-        TrinoValuesWriterFactory valuesWriterFactory = new TrinoValuesWriterFactory(writerOptions.getMaxPageSize(), DEFAULT_DICTIONARY_PAGE_SIZE);
+        TrinoValuesWriterFactory valuesWriterFactory = new TrinoValuesWriterFactory(
+                writerOptions.getMaxPageSize(),
+                DEFAULT_DICTIONARY_PAGE_SIZE,
+                writerOptions.isUseDeltaLengthByteArrayEncoding());
```
findepi: writerOptions is immutable
we could pass the whole writerOptions into TrinoValuesWriterFactory

(opt)

lib/trino-parquet/src/test/java/io/trino/parquet/reader/decoders/AbstractValueDecodersTest.java
```diff
@@ -361,54 +335,4 @@ private static <T> List<Runner<T>> generateRunnerForBatchSize(Function<Integer,
                 .map(runnerProvider)
                 .collect(toImmutableList());
     }
-
-    private static ValuesWriter getValuesWriter(ParquetEncoding encoding, PrimitiveTypeName typeName, OptionalInt typeLength)
-    {
-        if (encoding.equals(ParquetEncoding.RLE)) {
-            if (typeName.equals(BOOLEAN)) {
```
findepi, line 368: Quite some testing code being removed in a commit that adds support for new encoding. I suppose the code might be simply being moved somewhere else, but that's not apparent.
If this is indeed a code move, please extract to a prep commit.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
```diff
@@ -324,13 +323,13 @@ public <T> void testReadNullable(DataPageVersion version, ColumnReaderFormat<T>
         PrimitiveField field = createField(format, false);
         ColumnReader reader = createColumnReader(field);
         // Write data
-        ValuesWriter writer = format.getPlainWriter();
+        ValuesWriter writer = format.getPlainWriter(version);
         T[] values1 = format.write(writer, new Integer[] {null, 1});
-        DataPage page1 = createNullableDataPage(version, PLAIN, writer, field, true, false);
+        DataPage page1 = createNullableDataPage(version, getParquetEncoding(writer.getEncoding()), writer, field, true, false);
```
findepi: One could expect `format.getPlainWriter(version)` to produce "a PLAIN writer", not  variable one.

this is test-only thing. possible to name better?

findepi: nit: `getParquetEncoding(writer.getEncoding())`  could be a variable, initialized after the `writer` is initialized, making the lines shorter

lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
```diff
@@ -732,7 +732,9 @@ private DataPage createDataPage(DataPageVersion version, ParquetEncoding encodin
     private DataPage createDataPage(DataPageVersion version, ParquetEncoding encoding, ValuesWriter writer, int valueCount, OptionalLong firstRowIndex)
             throws IOException
     {
-        Slice slice = Slices.wrappedBuffer(writer.getBytes().toByteArray());
```
findepi, line 738: this change looks redundant, revert

lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
```diff
@@ -732,7 +732,9 @@ private DataPage createDataPage(DataPageVersion version, ParquetEncoding encodin
...
+        byte[] valueBytes = writer.getBytes().toByteArray();
+        writer.reset();
```
findepi, line 762: why is reset added here that wasn't needed before?
(or maybe it was needed => separate commit?)

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeInsertCompatibility.java
```diff
@@ -436,6 +438,47 @@ public Object[][] compressionCodecs()
         };
     }
 
+    @Test(groups = {DELTA_LAKE_DATABRICKS, DELTA_LAKE_DATABRICKS_143, DELTA_LAKE_DATABRICKS_154, DELTA_LAKE_DATABRICKS_164, DELTA_LAKE_OSS, PROFILE_SPECIFIC_TESTS})
+    @Flaky(issue = DATABRICKS_COMMUNICATION_FAILURE_ISSUE, match = DATABRICKS_COMMUNICATION_FAILURE_MATCH)
+    public void testTrinoCreatedTableWithDeltaLengthByteArrayCompatibility()
```
findepi, line 437: Great that you thought about adding compatibility tests with Delta.
It would be sad, however, if the non-dictionary writer was not really tested before the changes. Were we missing test coverage?

findepi (commented):
LGTM. Would be great to get more eyes 👀   on the change, especially from folks closer to Parquet.

raunaqmorarka:
Switched to `bindConfigDefaults` in `IcebergModule` so the default for the Iceberg connector is `false`, while keeping the property user-configurable per catalog. Hive and Delta Lake still default to `true`. Documented the limitation in the file-formats docs.

lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriterOptions.java
raunaqmorarka: Added `DEFAULT_USE_DELTA_LENGTH_BYTE_ARRAY_ENCODING` constant and use it as the Builder default.

lib/trino-parquet/src/main/java/io/trino/parquet/writer/ParquetWriters.java
raunaqmorarka: Done. `TrinoValuesWriterFactory` now takes `ParquetWriterOptions` directly and reads `maxPageSize` and `useDeltaLengthByteArrayEncoding` from it.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/decoders/AbstractValueDecodersTest.java
raunaqmorarka: Split into a prep commit (`Extract TestingValuesWriters utility from AbstractValueDecodersTest`, pure code move) and the main commit, whose body now explicitly states it widens the test matrix without adding new encoding support.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
raunaqmorarka: Renamed to `getValuesWriter(version)`.

raunaqmorarka: Added overloads of `createDataPage` and `createNullableDataPage` that derive the encoding from the writer; non-dictionary call sites use them and drop the redundant `getParquetEncoding(writer.getEncoding())` argument.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
raunaqmorarka: Not redundant — V2 writers (`RunLengthBitPackingHybridValuesWriter`, `DeltaBinaryPackingValuesWriter*`, `DeltaByteArrayWriter`) throw `IllegalArgumentException("You cannot call toBytes() more than once without calling reset()")` on a second `getBytes()` call. `PlainValuesWriter` tolerated double-`getBytes`, which is why this line wasn't needed before. Verified by removing locally: `TestNestedColumnReader.testReadOnlyNulls[BOOLEAN+V2]` fails because the test reuses the same writer across two `createDataPage` calls.

lib/trino-parquet/src/test/java/io/trino/parquet/reader/AbstractColumnReaderTest.java
raunaqmorarka: Same as the comment on the line above — required by the V2 writers introduced in this commit. PLAIN tolerated double `getBytes()`, but RLE/DELTA throw, breaking tests like `testReadOnlyNulls` that reuse the writer. Splitting into a prep commit doesn't make sense: on master with PLAIN-only writers the reset is genuinely unneeded.

testing/trino-product-tests/src/main/java/io/trino/tests/product/deltalake/TestDeltaLakeInsertCompatibility.java
raunaqmorarka: Yes — there was no dedicated cross-engine compatibility test for the non-dictionary fallback writer before this PR. The dictionary path is well-exercised by existing product tests (small string columns), but the PLAIN fallback for high-cardinality BYTE_ARRAY data wasn't covered.

raunaqmorarka:
**ClickBench, PLAIN fallback vs DLBA fallback (4 runs each on a stable remote machine, 43 queries)**

Headline: **DLBA fallback gives -5.3% total runtime** (175.7s → 166.4s).

Per-run variance was tight — coefficient of variation under 2.1% on every query, median ~0.5% — so the per-query deltas are real, not measurement noise.

In the rewritten file, the high-cardinality string columns end up DLBA-encoded after the dictionary overflows: URL (98% of pages DLBA), Title (97%), Referer (97%), SearchPhrase (83%). Lower-cardinality string columns (MobilePhoneModel, PageCharset, etc.) fit entirely in the dictionary and stay RLE-dictionary-encoded, unaffected by the encoding change.

**Wins on queries that read at least one DLBA-encoded column:**

| Query | PLAIN fallback (ms) | DLBA fallback (ms) | Δ | DLBA columns read |
|---|---:|---:|---:|---|
| q21 | 3760 | 2852 | **-24.2%** | URL, SearchPhrase |
| q22 | 5736 | 4568 | -20.4% | URL, Title, SearchPhrase |
| q20 | 3849 | 3119 | -19.0% | URL |
| q27 | 4083 | 3417 | -16.3% | URL |
| q23 | 13013 | 11184 | -14.1% | URL (+ all columns via SELECT *) |
| q25 | 1081 | 948 | -12.3% | SearchPhrase |
| q24 | 1439 | 1274 | -11.5% | SearchPhrase |
| q26 | 1430 | 1275 | -10.9% | SearchPhrase |
| q30 | 2491 | 2344 | -5.9% | SearchPhrase |
| q31 | 2695 | 2549 | -5.4% | SearchPhrase |
| q05 | 3417 | 3233 | -5.4% | SearchPhrase |
| q14 | 3376 | 3197 | -5.3% | SearchPhrase |
| q28 | 23546 | 22366 | -5.0% | Referer |
| q12 | 3177 | 3041 | -4.3% | SearchPhrase |
| q34 | 17240 | 16543 | -4.0% | URL |
| q33 | 16594 | 16060 | -3.2% | URL |
| q13 | 4903 | 4746 | -3.2% | SearchPhrase |
| q17 | 6844 | 6640 | -3.0% | SearchPhrase |

**No real regressions.** The largest movement in the wrong direction is q35 at +4.3%, a pure-numeric query (`ClientIP` arithmetic) that doesn't touch any DLBA-encoded column. Other small positives (q09 +1.9%, q32 +0.8%, q40 +1.4%, q41 +0.2%, q42 +0.7%, q36-q38 +0.4 to +0.8%) are also numeric-dominated, or filter so selectively on date/counter that the encoding has minimal surface area to act on.

File size on the ClickBench hits dataset is essentially unchanged (1020 MB → 1022 MB, +0.2%) — Snappy is doing the heavy lifting on both encodings; DLBA's per-row uncompressed savings don't translate to compressed-bytes savings on this corpus, but they don't bloat the file either.

**Summary**: enabling DLBA as the dict-overflow fallback gives a clear 10-25% speedup on URL/Title-heavy aggregations, a 3-6% speedup on a broad second tier of SearchPhrase/Referer-heavy queries, and is essentially neutral elsewhere on ClickBench. No regression observed.

raunaqmorarka:
@coderabbitai full review

sopel39 (approved):

Issue #29243, opened 2026-04-27
Add drop-with-purge-by-catalog property to Iceberg REST catalog


vkft:
### Feature Request / Improvement

Proposal
Add new drop-with-purge-by-catalog property to Iceberg Rest catalog:

> iceberg.rest-catalog.drop-with-purge-by-catalog-enabled=true

This property should control if PURGE operation should be handled by REST Catalog (default behavior) or by Trino

Rationale
Some Iceberg REST catalog are not able to remove table's file from storage that causes leaving a lot of orphan folders/files or causes incompatibility issues (as an example Apache Polaris 1.4.0 has such issue and it's not possible to drop huge table from Trino with Polaris Rest catalog connected).
Trino already has PURGE implementation for AWS Glue and Iceberg REST catalog for Google security and property should enable this flow (unregisterTable, dropTableData, deleteTableDirectory) for rest of catalog implementation

Issue #26798, opened 2025-10-02
Add a property to Iceberg tables to not remove data files when the table is dropped.


sopel39:
By default `DROP TABLE` on Iceberg tables removes metadata as well as data files (even if files were added via `add-files` procedure). This might lead to unintentional (and non-reversible) loss of data. To mitigate that risk and provide safer default behavior for users a table property could be added to Iceberg tables to not remove data files when the table is dropped. Other engines (Spark, Starrocks) lean on this safer behavior by default.

The plan is as follows:
1. Add config property `iceberg.default-purge-data-on-delete.enabled` (by default `true`)
2. Make a corresponding table property `iceberg_purge_data_on_delete_enabled`
3. When the property is not set (on an older table), then default to `iceberg.default-purge-data-on-delete.enabled` value.
4. For new tables use `iceberg.default-purge-data-on-delete.enabled` as default table property
5. `DROP TABLE` will remove table data only when table property `iceberg_purge_data_on_delete_enabled` is set to `true`
6. To remove table with data user has to alter table property
```
ALTER TABLE table SET PROPERTIES iceberg_purge_data_on_delete_enabled=true
```
And then drop the table
```
DROP TABLE table
```

sopel39:
cc @raunaqmorarka @findepi @ebyhr @raunaqmorarka @electrum

tbaeg:
@sopel39 What would the behavior be for other metadata operations that delete files (i.e. - `expire_snapshots`)? 

https://github.com/trinodb/trino/pull/26213

sopel39:
> @sopel39 What would the behavior be for other metadata operations that delete files (i.e. - expire_snapshots)?

@tbaeg I didn't plan for behavior change in other operations. What do you suggest?

tbaeg:
@sopel39 I don't necessarily think there has to be a behavior change, but probably worth at least considering. At the minimum, I do think it's worth changing the properties to use the word `drop` instead of `delete`.

`iceberg.default-purge-data-on-delete.enabled` -> `iceberg.default-purge-data-on-drop.enabled`

raunaqmorarka:
I don't think we should change the current default behaviour. Most users would be caught unawares and incur increased storage costs if we started to silently stop deleting data.
I think users can already use the iceberg table property `gc.enabled` to avoid deleting data on drop. When that is set, running `expire_snapshots` should throw an exception.
We could add a catalog configuration property which when set will always create new tables with `gc.enabled` set to false.

[...]

PR #29248, opened 2026-04-27
Add drop with purge by catalog property to iceberg rest catalog

vkft:
## Description
Addressing feature proposal: https://github.com/trinodb/trino/issues/29243.
Add new Iceberg Rest catalog property that controls whether REST Catalog should handle drop with purge operation. Defaults to `true`. This property must only be set as a workaround for https://github.com/apache/polaris/issues/3695.

https://github.com/apache/polaris/issues/3695
It's hard to use Trino with Apache Polaris 1.4.0 release if 'drop table' is required.
Root cause is in Apache Polaris TableCleanupTaskHandler implementation that does not work well with few GB tables and more.
Apache Polaris contributors verdict for now:

> It's likely purge workload saturates storage/credential HTTP clients, Polaris stops answering liveness in time, Kubernetes restarts it. We will need a delegation service for drop-by-purge, which isn't there yet. We have to build it. As a workaround, try not to have Polaris handle it, use engines(spark/trino) for purging table instead.

This PR is addressing their suggestion.

PR details: table purge is reusing existing logic from REST catalog + Security == GOOGLE. To don't duplicate code but make method name consistent with when it's executing I rename purgeBigLakeTable to purgeTableInternally (means purge internally by Trino). Not sure that naming is good so please feel free to make suggestions here.
Thanks!

## Release notes

```markdown
## Section
* Fix some things. ({issue}`29243`)
```

### DIFF
--- a/docs/src/main/sphinx/object-storage/metastores.md
+++ b/docs/src/main/sphinx/object-storage/metastores.md
@@ -536,7 +536,10 @@ following properties:
 * - `iceberg.rest-catalog.case-insensitive-name-matching.cache-ttl`
   - [Duration](prop-type-duration) for which case-insensitive namespace, table, 
     and view names are cached. Defaults to `1m`.
-  :::
+* - `iceberg.rest-catalog.drop-with-purge-by-catalog-enabled`
+  - Controls whether REST Catalog should handle drop with purge operation. Defaults to `true`. This property must only be set as a workaround for [this
+    issue](https://github.com/apache/polaris/issues/3695).   
+    :::
 
 The following example shows a minimal catalog configuration using an Iceberg
 REST metadata catalog:
--- a/docs/src/main/sphinx/object-storage/metastores.md
+++ b/docs/src/main/sphinx/object-storage/metastores.md
@@ -538,8 +538,8 @@ following properties:
     and view names are cached. Defaults to `1m`.
 * - `iceberg.rest-catalog.drop-with-purge-by-catalog-enabled`
   - Controls whether REST Catalog should handle drop with purge operation. Defaults to `true`. This property must only be set as a workaround for [this
-    issue](https://github.com/apache/polaris/issues/3695).   
-    :::
+      issue](https://github.com/apache/polaris/issues/3695).
+  :::
 
 The following example shows a minimal catalog configuration using an Iceberg
 REST metadata catalog:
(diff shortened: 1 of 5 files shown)


docs/src/main/sphinx/object-storage/metastores.md
```diff
@@ -536,6 +536,9 @@ following properties:
 * - `iceberg.rest-catalog.case-insensitive-name-matching.cache-ttl`
   - [Duration](prop-type-duration) for which case-insensitive namespace, table, 
     and view names are cached. Defaults to `1m`.
+* - `iceberg.rest-catalog.drop-with-purge-by-catalog-enabled`
+  - Controls whether REST Catalog should handle drop with purge operation. Defaults to `true`. This property must only be set as a workaround for [this
+      issue](https://github.com/apache/polaris/issues/3695).
```
ebyhr, line 541: I'd prefer not to introduce a new property solely for a single REST catalog implementation. The Polaris-specific issue should be addressed within Polaris itself; otherwise, other query engines are likely to encounter the same problem.

vkft: Hello @ebyhr 
Thanks for your feedback! I would agree with you, but:
- We already have the same workaround for Google-handled security. So, we are not reinventing anything new, but just enable feature that already exists.
- If I'm not mistaken Apache Spark purges files/directories by itself. At least it is so on 3.5.6.
- REST Catalog should provide purge functionality for sure, but query engine should also set a flag if table should be dropped with purge or without. Currently we have drop-with-purge implemented only and no property to change behavior. Use unregister instead of drop does not solve an issue once, for instance, we have big analytics team that are using DBT intensively and drop operation is standard for a lot of cases (or event is generated by DBT itself). So, if we don't want to just enable such option like let Trino purge data, probably we should introduce property to set if drop should be done with purge or without. What do you think?
- And yes, I'm just trying to solve my production blocker :) 

Will be happy to hear your feedback. Thanks!

ebyhr: The logic for GOOGLE isn't workaround. I don't expect they'll change their behavior in the foreseeable future.

> we should introduce property to set if drop should be done with purge or without.

We already have an existing issue. https://github.com/trinodb/trino/issues/26798

> I'm just trying to solve my production blocker

My question is why you want to introduce a workaround in Trino instead of fixing Polaris.

vkft: > My question is why you want to introduce a workaround in Trino instead of fixing Polaris.
Well... Trino is doing this purge great. But it's not a answer, I know. I'm not an Apache Polaris contributor that's why trying to solve it in easiest way. But agree, it should be solved in Polaris. Thanks

PR #28814, opened 2026-03-23
Extend IcebergInputInfo

oskar-szwajkowski:
## Description

Add separate information for equality and position deletes count

## Additional context and related issues

This will enhance visibility over table's delete file structure

## Release notes

( ) This is not user-visible or is docs only, and no release notes are required.
( ) Release notes are required. Please propose a release note for me.
(X) Release notes are required, with the following suggested text:

Add information about position and equality delete files count to `IcebergInputInfo`

Issue #25963, opened 2025-06-09
Iceberg REST catalog: can Trino honor `snapshot-loading-mode=refs`?


anechitaev:
Hello,
The Iceberg REST API supports a `snapshot-loading-mode=refs` flag that returns only the snapshots referenced by branches or tags, reducing the payload for long-lived tables. I tried adding the same flag to a Trino Iceberg REST catalog  as a catalog property but Trino reports
`Error: Configuration property 'snapshot-loading-mode' was not used`

Is there a specific reason Trino’s Iceberg connector ignores `snapshot-loading-mode` today, or did I miss something?  
Are there plans (or blockers) to support this optimisation in a future release?

Thanks!

PR #29262, opened 2026-04-28
Add totalSnapshots to IcebergInputInfo

grantatspothero:
## Description
Tracks snapshot count for tables for better observability

Related PR: https://github.com/trinodb/trino/pull/28814

## Release notes

Add information about snapshot count to IcebergInputInfo

### DIFF
--- a/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
+++ b/plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
@@ -2745,9 +2745,11 @@ public class IcebergMetadata
                         .collect(toImmutableList()) : ImmutableList.of();
 
         Map<String, String> summary = ImmutableMap.of();
+        long totalSnapshots = 0;
         if (icebergTableHandle.getSnapshotId().isPresent()) {
             Table table = catalog.loadTable(session, icebergTableHandle.getSchemaTableName());
             summary = table.snapshot(icebergTableHandle.getSnapshotId().getAsLong()).summary();
+            totalSnapshots = Iterables.size(table.snapshots());
         }
         Optional<String> totalRecords = Optional.ofNullable(summary.get(TOTAL_RECORDS_PROP));
         Optional<String> deletedRecords = Optional.ofNullable(summary.get(DELETED_RECORDS_PROP));
@@ -2761,6 +2763,7 @@ public class IcebergMetadata
                 icebergTableHandle.getSnapshotId(),
                 partitionFields,
                 getFileFormat(icebergTableHandle.getStorageProperties()).name(),
+                String.valueOf(totalSnapshots),
                 totalRecords,
                 deletedRecords,
                 totalDataFiles,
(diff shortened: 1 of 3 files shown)


plugin/trino-iceberg/src/main/java/io/trino/plugin/iceberg/IcebergMetadata.java
```diff
@@ -2745,9 +2745,11 @@ public Optional<Object> getInfo(ConnectorSession session, ConnectorTableHandle t
                         .collect(toImmutableList()) : ImmutableList.of();
 
         Map<String, String> summary = ImmutableMap.of();
+        long totalSnapshots = 0;
         if (icebergTableHandle.getSnapshotId().isPresent()) {
             Table table = catalog.loadTable(session, icebergTableHandle.getSchemaTableName());
             summary = table.snapshot(icebergTableHandle.getSnapshotId().getAsLong()).summary();
+            totalSnapshots = Iterables.size(table.snapshots());
```
raunaqmorarka, line 2752: This is internally loading snapshots in TableMetadata.
Safe to assume that by the time we get here snapshots are already loaded and we won't be the ones to trigger it ?

grantatspothero: The table is loaded by TableOperations which for non-rest catalogs uses `TableMetadataParser`: https://github.com/apache/iceberg/blob/main/core/src/main/java/org/apache/iceberg/TableMetadataParser.java#L557-L583
TableMetadataParser always loads the snapshots array in full/sets `snapshotsSupplier` to null and sets `snapshotsLoaded` to true. So there is no lazy loading of snapshots happening when using file-backed TableOperations. 

Referenced Iceberg lazy loading commits:
- Adds support for lazy loading of snapshots, but file backed table metadata is never lazily loaded: https://github.com/apache/iceberg/pull/6811
- Adds lazy loading support for rest catalog backed table metadata: https://github.com/apache/iceberg/pull/6850


I'll confirm for REST catalogs, but it appears if someone used a REST catalog with `snapshot-loading-mode=REFS` then yes this would trigger snapshot loading which is bad. 

I also need to confirm if `table.snapshot(icebergTableHandle.getSnapshotId().getAsLong())` would trigger the same snapshot loading. It depends if the referenced commit exists in iceberg `refs`. The common case is it wouldn't trigger snapshot loading, but it might if querying a historical snapshot not in refs.

grantatspothero: The above is safe today for REST catalogs only because trino does not allow configuring `snapshot-mode=REFS`, but if this is allowed through trino (https://github.com/trinodb/trino/issues/25963) then yes this could trigger a full snapshots load.

Don't want to introduce a performance footgun for the future if we ever allow configuring snapshot-mode.

grantatspothero:
Closing, slightly better observability isn't worth potential future performance footguns if trino ever supports `snapshot-loading-mode=REFS`

